PR #33, opened 2020-04-28
Doc/add documentation

kyazdani42:


doc/nvim-treesitter.txt
```diff
@@ -0,0 +1,90 @@
+*nvim-treesitter* 
+
+Minimum version of neovim: nightly
+
+Authors: Yazdani Kiyan <yazdani.kiyan@protonmail.com>, Vigouroux Thomas <@EMAIL?>
```
vigoux, line 5: I should really have a protonmail too...
```suggestion
Authors: Yazdani Kiyan <yazdani.kiyan@protonmail.com>, Vigouroux Thomas <tomvig38@gmail.com>
```

doc/nvim-treesitter.txt
```diff
@@ -0,0 +1,90 @@
...
+
+==============================================================================
+INTRODUCTION                                       *nvim-treesitter-intro*
+
+nvim-treesitter wraps the neovim treesitter api to provide functionnalities such
+as highlighting and incremental selection, and a command to easily install parsers.
+
```
vigoux, line 12: Something like this should be added.
```suggestion
This plugin also provides a some utility function to abstract out treesitter so that plugin are easily writable.

```

kyazdani42: maybe we should add this when it is available

PR #34, opened 2020-04-30
Add function 'nvim-treesitter/install'.ensure_installed

theHamsta:
Hello,

I was trying to create a function that users could put into their config to ensure that their favorite languages are parsers installed or get installed if not yet available.

Currently, you can do:
```
require'nvim-treesitter/install'.ensure_installed({'html', 'python'})
```
or just one language
```
require'nvim-treesitter/install'.ensure_installed('cpp')
```
or all available languages
```
require'nvim-treesitter/install'.ensure_installed('all')
```
I am not sure where to put this function in your repo structure and whether really all requested parsers get installed (the status only shows when a particular languages got installed). So I am looking forward for feedback!

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -325,6 +325,9 @@ function M.setup(user_data)
       if config[mod].keymaps and type(data.keymaps) == 'table' then
         config[mod].keymaps = data.keymaps
       end
+      if mod == 'ensure_installed' then
+        require'nvim-treesitter/install'.ensure_installed(data)
+      end
     end
   end
 end
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -3,6 +3,7 @@ local fn = vim.fn
 local luv = vim.loop
 local configs = require'nvim-treesitter/configs'
 local parsers = configs.get_parser_configs()
+local has_parser = require'nvim-treesitter/parsers'.has_parser
 
 local M = {}
 
@@ -142,6 +143,24 @@ local function install(ft)
 
   run_install(cache_folder, package_path, ft, install_info)
 end
+
+
+M.ensure_installed = function(languages)
+  if type(languages) == 'string' then
+    if languages == 'all' then
+      languages = configs.available_parsers()
+    else
+      languages = {languages}
+    end
+  end
+
+  for _, ft in ipairs(languages) do
+    if not has_parser(ft) then
+      install(ft)
+    end
+  end
+end
+
 
 M.commands = {
   TSInstall = {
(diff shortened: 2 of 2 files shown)


lua/nvim-treesitter/install.lua
```diff
@@ -143,6 +143,25 @@ local function install(ft)
...
+  if type(languages) == 'string' then
+    if languages == 'all' then
+      languages = configs.available_parsers()
+    else
+      languages = {languages}
+    end
+  end
+
+  for _, ft in ipairs(languages) do
+    local is_installed = #api.nvim_get_runtime_file('parser/'..ft..'.so', false) > 0
```
kyazdani42, line 157: you could use the `has_parser` function

theHamsta:
Like so:

```lua
      highlight = {
          enable = true,             -- false will disable the whole extension
          disable = {}, -- list of language that will be disabled
      },
      textobj = {                        -- this enables incremental selection
          enable = true,
          disable = {},
      },
      ensure_installed = 'all' -- or {'python', 'cpp', 'rust'}
  }
```
Maybe rename it to `ensure_parser_installed`?

I also fixed a small bug in `iter_cmd` that was using a global which prevented installing multiple parsers in parallel.

lua/nvim-treesitter/configs.lua
```diff
@@ -166,6 +166,7 @@ local config = {
     },
     is_supported = function() return true end
   },
+  ensure_installed = {},
```
kyazdani42, line 169: config keys are actually used by `:TSModuleInfo`, maybe we should rename it to 'modules', i'll do that in a PR. So you can remove ensure_installed from here.

lua/nvim-treesitter/install.lua
```diff
@@ -39,9 +40,11 @@ local function iter_cmd(cmd_list, i, ft)
   local attr = cmd_list[i]
   if attr.info then print(attr.info) end
 
+  local handle
+
```
kyazdani42, line 44: cool thanks.

lua/nvim-treesitter/configs.lua
```diff
@@ -325,6 +326,10 @@ function M.setup(user_data)
       if config[mod].keymaps and type(data.keymaps) == 'table' then
         config[mod].keymaps = data.keymaps
       end
+      if mod == 'ensure_installed' then
+        config.ensure_installed = data
+        require'nvim-treesitter/install'.ensure_installed(data)
+      end
```
kyazdani42, line 332: i think as this function will only run once, we don't need to add it to `config.ensure_installed`.
Though i wonder if you had something in mind by storing this in the config object ?

theHamsta: not, really. I only thought it would be strange if all other data gets stored and this one not.

kyazdani42: hum well remove it for now, i'll do a small refactoring to make sure this can be stored properly.

theHamsta: ok. it's removed.

PR #36, opened 2020-05-01
Add some more community parsers

theHamsta:
I added some more the parsers, some of them maintained by the community. Not sure whether to add them right now or wait a bit longer for this project to be even more mature.

The haskell one really takes long to compile...

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -145,6 +145,62 @@ parsers.tsx = {
   }
 }
 
+parsers.scala = {
+  install_info = {
+    url = "https://github.com/tree-sitter/tree-sitter-scala",
+    files = { "src/parser.c", "src/scanner.c" },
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/configs.lua
```diff
@@ -145,6 +145,64 @@ parsers.tsx = {
...
+
+parsers.nix = {
+  install_info = {
+    url = "https://github.com/cstrahan/tree-sitter-nix",
+    files = { "src/parser.c", "src/scanner.cc" },
+  }
+}
+
+
+
```
kyazdani42, line 205: could you remove these 2 empty lines ?
Thanks for the PR again :+1: 
I think they are good to have, so i'll merge when you update this little detail.

theHamsta: Oops. :stuck_out_tongue:

PR #37, opened 2020-05-01
Add 'nvim-treesitter/node-movement'

theHamsta:
First draft for tree-sitter-base movement:

Those mappings let you move in the documents node hierarchy:
```
nmap <a-k> :lua require'nvim-treesitter/node-movement'.move_up()<cr>
nmap <a-h> :lua require'nvim-treesitter/node-movement'.move_left()<cr>
nmap <a-l> :lua require'nvim-treesitter/node-movement'.move_right()<cr>
nmap <a-j> :lua require'nvim-treesitter/node-movement'.move_down()<cr>
```

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -223,6 +223,17 @@ local config = {
       },
       is_supported = function() return true end
     },
+    node_movement = {
+      enable = false,
+      disable = {},
+      is_supported = function() return true end,
+      keymaps = {
+        move_up = "<a-k>",
+        move_down = "<a-j>",
+        move_left = "<a-h>",
+        move_right = "<a-l>",
+      },
+    },
     -- folding = {
     --   enable = false,
     --   disable = {},
--- a/lua/nvim-treesitter/node_movement.lua
+++ b/lua/nvim-treesitter/node_movement.lua
@@ -1,0 +1,91 @@
+local api = vim.api
+local parsers = require'nvim-treesitter.parsers'
+local utils = require'nvim-treesitter.utils'
+local M = {}
+
+
+M.NodeMovementKind = {
+  up = 'up',
+  down = 'down',
+  left = 'left',
+  right = 'right',
+}
+
+M.current_node = {}
+
+local function node_start_to_vim(node)
+  if not node then return end
+
+  local row, col = node:start()
+  local exec_command = string.format('call cursor(%d, %d)', row+1, col+1)
+  api.nvim_exec(exec_command, false)
+end
+
+M.do_node_movement = function(kind)
+  local buf, line, col = unpack(vim.fn.getpos("."))
+
+  local current_node = M.current_node[buf]
+
+  if current_node then
+    local node_line, node_col = current_node:start()
+    if line-1 ~= node_line or  col-1 ~= node_col then
+      current_node = nil
+    end
+  end
+  local destination_node
+
+  if parsers.has_parser() then
+    local root = parsers.get_parser():parse():root()
+    if not current_node then
+      current_node = root:named_descendant_for_range(line-1, col-1, line-1, col)
+    end
+
+    if kind == M.NodeMovementKind.up then
+       destination_node = current_node:parent()
+    elseif kind == M.NodeMovementKind.down then
+      if current_node:named_child_count() > 0 then
+        destination_node = current_node:named_child(0)
+      else
+        local next_node = utils.get_next_node(current_node)
+        if next_node and next_node:named_child_count() > 0 then
+          destination_node = next_node:named_child(0)
+        end
+      end
+    elseif kind == M.NodeMovementKind.left then
+      destination_node = utils.get_previous_node(current_node, true, true)
+    elseif kind == M.NodeMovementKind.right then
+      destination_node = utils.get_next_node(current_node, true, true)
+    end
+    M.current_node[buf] = destination_node or current_node
+  end
+
+  if destination_node then
+    node_start_to_vim(destination_node)
+  end
+end
+
+M.move_up = function() M.do_node_movement(M.NodeMovementKind.up) end
+M.move_down = function() M.do_node_movement(M.NodeMovementKind.down) end
+M.move_left = function() M.do_node_movement(M.NodeMovementKind.left) end
+M.move_right = function() M.do_node_movement(M.NodeMovementKind.right) end
+
+function M.attach(bufnr)
+  local buf = bufnr or api.nvim_get_current_buf()
+
+  local config = require'nvim-treesitter.configs'.get_module('node_movement')
+  for funcname, mapping in pairs(config.keymaps) do
+    api.nvim_buf_set_keymap(buf, 'n', mapping,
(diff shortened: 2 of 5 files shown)


theHamsta:
I need to get some experience how those mappings transfer to different languages but would be cool to kind of create a paredit kind of editing experience where you can swap function arguments or arms of if/else blocks.

theHamsta:
I find this useful for navigating in markdown documents where you can move section-wise/sub-section-wise or per paragraph.

kyazdani42:
This is a great idea, we should add it as a module. Though the `textobj` modules which handles incremental selection now doesn't only relate to the incremental selection but also the navigation.
Should we rename `textobj` as `incremental_selection` and leave textobj as a utils module, just using utility functions from there when we need and bind the incremental selection to an actual config module ? @vigoux

vigoux:
Wow awesome PR !
I agree with @kyazdani42, we should move it in a separate module, and rename `textobj` to `incremental_selection`.

theHamsta:
I realized that we probably should highlight the current node below cursor after a movement then the user you easier create a visual selection from it.

Also 'down' should probably not always go to first child but rather to the last child visited (if available).

I need also to set the movement to the jumplist.

Well, I still have no clue how to do most things here, I guess some experimenting in necessary.

lua/nvim-treesitter/node-movement.lua
```diff
@@ -0,0 +1,125 @@
...
+      local next_node = get_next_node(node:parent())
+      if next_node and next_node:named_child_count() > 0 then
+        destination_node = next_node:named_child(0)
+      elseif next_node then
+        destination_node = next_node
+      end
+    end
+  end
+  return destination_node
+end
```
kyazdani42, line 75: Could you move those two functions in the utils file ? It might be reused by other modules

lua/nvim-treesitter/node-movement.lua
```diff
@@ -0,0 +1,125 @@
...
+    node_start_to_vim(destination_node)
+  end
+end
+
+M.move_up = function() M.do_node_movement(M.NodeMovementKind.up) end
+M.move_down = function() M.do_node_movement(M.NodeMovementKind.down) end
+M.move_left = function() M.do_node_movement(M.NodeMovementKind.left) end
+M.move_right = function() M.do_node_movement(M.NodeMovementKind.right) end
+
+return M
```
kyazdani42, line 125: also add an `attach` and `detach` function as other modules do (you could look how its done inside `textobj`)

kyazdani42: so you can provide default keymaps, and it can be enabled/disabled directly from the `configs.setup` function

vigoux:
One crazy cool thing we may went to do is moving nodes too, first just reordering while staying at the same level, them moving them up or down the tree.

With this it would be easy to move functions, arguments or whatever.

theHamsta:
It would also be very cool to have a module that switches a buffer to lisp mode for editing and back when finished.

vigoux:
> It would also be very cool to have a module that switches a buffer to lisp mode for editing and back when finished.

Don't know if it would have a real use case though, editing the tree blindly is not that great ;-)

theHamsta:
Is there any easy possibility to make the cursor movement also work in visual mode? Or do I have to check in which mode I am and then use a code similar to `node_range_to_vim` in `textobj.lua`?

kyazdani42:
i don't know about that, try `api.nvim_win_set_cursor`

vigoux:
@theHamsta the only way I found was as in `node_range_to_vim` it seems that doing qpu calls from lua gets you to normal mode however I tried. I'm not sure of my solution though.

lua/nvim-treesitter/node-movement.lua
theHamsta: Ok. I did that I tried to hide the extra magic (module-specific behavior) into optional parameters.

theHamsta:
I think this can be reviewed. There's a lot left to do though. But I think it would be better to have separate PRs for that.

In particular, I would love to have:

[ ] functions for moving nodes (i.e. changing the text, could be easily integrated into `do_node_movement` function)
[ ] highlight for current_node after move
[ ] function to convert current_node to visual selection (-> this could have the same mapping in normal mode as `increment_selection` in visual mode)
[ ] make node movements also work for visual selection (this is different from `incremental_selection` since it would expand to selection only on one side with the same movements as in normal mode)

theHamsta:
Maybe one more open question. What should happen if the user is between node e.g. in empty space between functions. Currenty, behavior: do nothing

kyazdani42:
we should add @theHamsta  suggestions to the roadmap @vigoux

lua/nvim-treesitter/utils.lua
```diff
@@ -93,4 +93,63 @@ function M.smallest_containing_scope(node, bufnr)
...
+  local parent = node:parent()
+  if parent then
+    local found_pos = 0
+    for i = 0,parent:named_child_count()-1,1 do
+      if parent:named_child(i) == node then
+        found_pos = i
+        break
+      end
+    end
+    if 0 <= found_pos - 1 then
```
kyazdani42, line 141: `if 0 < found_pos`

lua/nvim-treesitter/node_movement.lua
```diff
@@ -0,0 +1,96 @@
...
+      else
+        local next_node = utils.get_next_node(current_node)
+        if next_node and next_node:named_child_count() > 0 then
+          destination_node = next_node:named_child(0)
+        end
+      end
+    -- LEFT
+    elseif kind == M.NodeMovementKind.left then
+      destination_node = utils.get_previous_node(current_node, true, true)
+    -- RIGHT
```
kyazdani42, line 60: i dont think those comments are useful, the if statement already is by itself

lua/nvim-treesitter/node_movement.lua
```diff
@@ -0,0 +1,96 @@
...
+    local root = parsers.get_parser():parse():root()
+    if not current_node then
+      current_node = root:named_descendant_for_range(line-1, col-1, line-1, col)
+    end
+
+
+     --UP
+    if kind == M.NodeMovementKind.up then
+       destination_node = current_node:parent()
+    -- DOWN
```
kyazdani42, line 47: remove those comments too

lua/nvim-treesitter/configs.lua
```diff
@@ -223,6 +223,17 @@ local config = {
       },
       is_supported = function() return true end
     },
+    node_movement = {                        -- this enables incremental selection
```
kyazdani42, line 226: I think this comment should not be there

lua/nvim-treesitter/utils.lua
theHamsta: Why? 0 is a valid index for a node. It's c indices not lua.

theHamsta: Sorry I am confused. Now I get what you mean! :smile:

vigoux:
Tried the PR a bit, it is crazy neat. One little thing is that is is hard to determine on which dpeth you are, guess that's the objective of #42.

theHamsta:
Ok. Updated README. Also for renaming: `textobj` -> `incremental_selection`

PR #39, opened 2020-05-02
Move textobj module to incremental_selection.

vigoux:
As suggested in #37, rename the textobj module to incremental_selection.
Also adds a utility function to get the config of a module.


lua/nvim-treesitter/configs.lua
```diff
@@ -403,4 +403,8 @@ function M.available_modules()
   return vim.tbl_keys(config.modules)
 end
 
+function M.get_config(mod)
+  return config.modules[mod]
+end
```
kyazdani42, line 408: i think it should be call `get_module` instead

PR #41, opened 2020-05-02
Fix documentation after #35

theHamsta:


theHamsta:
Oops. I was wrong the old configuration is still valid.

PR #42, opened 2020-05-03
[WIP] Node movement hightlight

theHamsta:
Add optional hightlights for current node in `node_movement`. This PR is on top of #37 

TODO:

- [ ] clear autocmd on detach
- [ ] define Highlight group somewhere

I've never developed vim plugins, so I don't know what's the best practice in defining highlights.
Maybe place something like `highlight NvimTreesitterCurrentNode guibg=#444400
` here: https://github.com/nvim-treesitter/nvim-treesitter/blob/0207836eb2aaa876d52ffddd1ec89ec347c67996/plugin/nvim-treesitter.vim#L7

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -232,6 +232,7 @@ local config = {
         move_down = "<a-j>",
         move_left = "<a-h>",
         move_right = "<a-l>",
+        select_current_node = "grn",
       },
     },
     -- folding = {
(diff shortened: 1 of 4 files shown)


vigoux:
One thing you may want to do rather than defining raw colors is using 
```vim
highlight link NvimTreesitterCurrentNode {SomeOtherGroup}
```
Check `:h hi-link` for more infos but I think that's the better way.

vigoux:
One neat thing could also be kind of a statusline indicator for the current selected node.
Highlighting is great but some people don't like it that much, so a statusline indicator or a simple function which evaluates to the current node type would be awesome !

theHamsta:
Hi. With #43 users can use  `nvim-treesitter.node_sitter.get_current_node()` to display infos about it in statusline, e.g. its name (`function`, `argument`, `indentifier`, start/end).

kyazdani42:
Also this should not be merged before #45, the code organization will change a bit and it'll be easier to get the current node.

theHamsta:
Sure, I can finish this this evening. Was busy the whole week.

PR #43, opened 2020-05-03
Node movement selection

theHamsta:
Add function to select current node in `node_movement` this plays nicely with `incremental_selection` since you can grow the selection afterwards.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ require'nvim-treesitter.configs'.setup {
           move_down = "<a-j>",
           move_left = "<a-h>",
           move_right = "<a-l>",
+          select_current_node = "<leader>e",-- select current node
+                                            -- ideally same as incremental_selection.node_incremental
         }
     },
     ensure_installed = 'all' -- one of 'all', 'language', or a list of languages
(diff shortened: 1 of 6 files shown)


vigoux:
The way `incremental_selection` works is that is selects the current node you are on when called for the first time. Correct me if I'm wrong but I think that the `select_current_node` thing is already implemented.

theHamsta:
Actually, yes. It's aleady implemented. This is why the added function only has 3 lines. I was only missing the normal mode binding to get your first visual selection. `incremental_selection` can not know the depth you're currently at after cursor movement.

theHamsta:
Maybe you're right and I should just add a normal mode mapping for `incremental_selection.increasing_node`. `incremental_selection` already has mappings `operator_pending` and `visual_mode`.

vigoux:
I did a bit of testing and it seems that `node_movement.current_node` has only one key, which is `0`. 
It might be an error, and you may want to check if `buf == 0` you should maybe get `nvim_get_current_buf()` instead. I don't really know for now but you may want to investigate.

kyazdani42:
@vigoux  @theHamsta i think this PR will be useless after i merged the refactoring (i can basically do a 2 line function for this thing). I'm closing this

PR #44, opened 2020-05-05
Provide a statusline indicator

vigoux:
It will show the current branch at the cursor going the tree as such.
root->node->subnode->leaf

If an argument is provided to `statusline`, then the tree will be
truncated as follows :
..->subnode->subnode

Also fixes an issue with `expression_at_point`.

### DIFF
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -97,5 +97,16 @@ A list of languages can be found at |:TSInstallInfo|
 
 List modules state for the current session.
 
- vim:tw=78:ts=8:noet:ft=help:norl:
+==============================================================================
+FUNCTIONS~
+						   *nvim-treesitter-functions*
 
+|nvim_treesitter#statusline(size)|
+						*nvim_treesitter#statusline()*
+
+Returns a string describing the current position in the syntax tree. This
+could be used as a statusline indicator.
+Note: The `size` argument is optionnal. When specified, the string will not be
+      longer than `size`.
+
+ vim:tw=78:ts=8:noet:ft=help:norl:
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -5,11 +5,12 @@ local M = {}
 
 function M.has_parser(lang)
   local lang = lang or api.nvim_buf_get_option(0, 'filetype')
+  if not lang or #lang == 0 then return false end
   return #api.nvim_get_runtime_file('parser/' .. lang .. '.*', false) > 0
 end
 
 function M.get_parser(bufnr, lang)
-  if M.has_parser() then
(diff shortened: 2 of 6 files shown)


kyazdani42:
should we make a vim function binding so plugins like lightline or airline can use that also or should we leave it as is ?

vigoux:
Yes you are right, should be added in last push.

lua/nvim-treesitter/parsers.lua
```diff
@@ -5,11 +5,12 @@ local M = {}
 
 function M.has_parser(lang)
   local lang = lang or api.nvim_buf_get_option(0, 'filetype')
+  if not lang or #lang == 0 then return false end
```
kyazdani42, line 8: why checking for the string length ?

vigoux: When opening a new buffer with `:enew` it is created with no filetype. Calling `:lua print(vim.api.nvim_buf_get_option(0, "ft")) ` on it will return an empty string instead of `nil`. Then, creating the parsers will cause error. I really think that this case should be handled upstream though, we may want to open a PR for this in the future.

kyazdani42: i dont see why it would create an error, an empty string should behave as any random string, but if you encountered some issue then i think we should leave it as is.

vigoux: It had created errors actually. Errors while loading the shared library within treesitter. I might look a bit into this soon.

kyazdani42: i think the `has_parser` function is not working properly because it doesn't look for the `.so` extension but it looks for `.*` so obviously it fails

kyazdani42: i'm gonna fix this on the current refactoring i'm working on, should be more stable and the api will be available to external plugins too.

doc/nvim-treesitter.txt
```diff
@@ -97,5 +97,16 @@ A list of languages can be found at |:TSInstallInfo|
 
 List modules state for the current session.
 
- vim:tw=78:ts=8:noet:ft=help:norl:
+==============================================================================
+FUNCTIONS~
+						   *nvim-treesitter-functions*
+
+|require'nvim-treesitter'.statusline(size)|
```
kyazdani42, line 104: didn't you provide a vim function ?

vigoux: Oh yeah you are right, completely forgot about it...

kristijanhusak:
Any chance to improve this to return names for some parts of the structure? 
For example, instead of this:
```
program->class_declaration->declaration_list->method_declaration
```

to have this:
```
program->MyClass->declaration_list->MyMethod
```

PR #45, opened 2020-05-06
refacto/feat: better handling of parser updates

kyazdani42:
Notes:
I didn't add the textobj module because i want this to be merged first.

Also the node_movement now works by moving between scopes, but will be disabled for languages without locals definitions.

To move with node movement, you need to define `parent_scope` `child_scope` `next_scope` and `previous_scope`.

For the node_movement to work as expected, the `locals.scm` should be describing the scopes as properly as they can, otherwise, it might be confusing for users to move around (right now, lua scopes are not complete at all). Also this will probably make it easier to create scopes as you can see what happens when you move from scope to scope.

I will add the documentation once this will be reviewed and tested properly.

features:
- [x]  node_movement is moving between scopes.
- [x] add selection initialization from normal mode
- [x] add a decremental selection

improvements:
- [x] attach to buffer to run tree parsing on change
- [x] run state update on `CursorMoved`
- [x] the buffer state is:
```lua
{
  cursor_pos = { row=row, col=col },
  current_node = node_under_cursor,
  selection = {
      range = nil, -- activates when starting a selection
      nodes = {} -- filling up when starting an incremental selection
  },
  parser = parser, -- parser for current buffer
}
```
- [x] refacto all the modules reliant on parsing the tree, update the current nodes, get the current nodes...

fixes:
- [x] fix `has_parser` to look for `.so` libraries
- [x] fix should select the whole file when selection root in selection

to roadmap:
- add proper textobj module
- buffer updates should run after a specified timeout (that might not be necessary)
- previous node/scope and next node/scope for selection

### DIFF
--- a/lua/nvim-treesitter.lua
+++ b/lua/nvim-treesitter.lua
@@ -1,15 +1,14 @@
 local api = vim.api
+
 local install = require'nvim-treesitter.install'
-local locals = require'nvim-treesitter.locals'
 local utils = require'nvim-treesitter.utils'
 local info = require'nvim-treesitter.info'
 local configs = require'nvim-treesitter.configs'
+local state = require'nvim-treesitter.state'
 
 local M = {}
 
--- This function sets up everythin needed for a given language
--- this is the main interface through the plugin
-function M.setup(lang)
+function M.setup()
   utils.setup_commands('install', install.commands)
   utils.setup_commands('info', info.commands)
   utils.setup_commands('configs', configs.commands)
@@ -18,33 +17,34 @@ function M.setup(lang)
     for _, mod in pairs(configs.available_modules()) do
       if configs.is_enabled(mod, ft) then
         local cmd = string.format("lua require'nvim-treesitter.%s'.attach()", mod)
-        api.nvim_command(string.format("autocmd FileType %s %s", ft, cmd))
+        api.nvim_command(string.format("autocmd NvimTreesitter FileType %s %s", ft, cmd))
       end
     end
+    local cmd = string.format("lua require'nvim-treesitter.state'.attach_to_buffer(%s)", ft)
+    api.nvim_command(string.format('autocmd NvimTreesitter FileType %s %s', ft, cmd))
   end
+
+  state.run_update()
 end
 
 function M.statusline(indicator_size)
-  local indicator_size = indicator_size or 1000
-  local expr = require"nvim-treesitter.utils".expression_at_point()
-  local current_node =
-    require'nvim-treesitter.node_movement'.current_node[api.nvim_get_current_buf()]
+  local indicator_size = indicator_size or 100
+  local bufnr = api.nvim_get_current_buf()
+  local buf_state = state.get_buf_state(bufnr)
+  if not buf_state then return "" end
 
-  local indicator = ""
+  local current_node = buf_state.current_node
+  if not current_node then return "" end
+
+  local expr = current_node:parent()
+  local prefix = ""
+  if expr then
+    prefix = "->"
+  end
+
+  local indicator = current_node:type()
   while expr and (#indicator + #(expr:type()) + 5) < indicator_size do
-
-    local prefix = ""
-    if expr:parent() then
-      prefix = "->"
-    end
-
-
-    if expr == current_node then
-      indicator = string.format("%s[%s]%s", prefix, expr:type(), indicator)
-    else
-      indicator = prefix .. expr:type() .. indicator
-    end
-
+    indicator = expr:type() .. prefix .. indicator
     expr = expr:parent()
   end
 
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -1,6 +1,8 @@
 local api = vim.api
+
 local queries = require'nvim-treesitter.query'
-local parser_utils = require'nvim-treesitter.parsers'
+local utils = require'nvim-treesitter.utils'
+
 local parsers = {}
 
 parsers.javascript = {
@@ -218,28 +220,28 @@ local config = {
       enable = false,
       disable = {},
       keymaps = {
+        init_selection="gnn",
         node_incremental="grn",
-        scope_incremental="grc"
+        scope_incremental="grc",
+        node_decremental="grm"
       },
-      is_supported = function() return true end
+      is_supported = function(ft)
+        return queries.get_query(ft, 'locals')
+      end
     },
     node_movement = {
       enable = false,
       disable = {},
-      is_supported = function() return true end,
+      is_supported = function(ft)
+        return queries.get_query(ft, 'locals')
+      end,
       keymaps = {
-        move_up = "<a-k>",
-        move_down = "<a-j>",
-        move_left = "<a-h>",
-        move_right = "<a-l>",
+        parent_scope = "<a-k>",
+        child_scope = "<a-j>",
+        next_scope = "<a-l>",
+        previous_scope = "<a-h>",
       },
-    },
-    -- folding = {
-    --   enable = false,
-    --   disable = {},
-    --   keymaps = {},
-    --   is_supported = function() return false end
-    -- }
+    }
   },
   ensure_installed = nil
 }
@@ -279,12 +281,12 @@ local function enable_all(mod, ft)
     end
   end
   if ft then
-    if parser_utils.has_parser(ft) then
+    if utils.has_parser(ft) then
       enable_mod_conf_autocmd(mod, ft)
     end
   else
     for _, ft in pairs(M.available_parsers()) do
-      if parser_utils.has_parser(ft) then
+      if utils.has_parser(ft) then
         enable_mod_conf_autocmd(mod, ft)
       end
     end
@@ -364,7 +366,7 @@ M.commands = {
 -- @param mod: module (string)
 -- @param ft: filetype (string)
 function M.is_enabled(mod, ft)
-  if not M.get_parser_configs()[ft] or not parser_utils.has_parser(ft) then
+  if not M.get_parser_configs()[ft] or not utils.has_parser(ft) then
     return false
   end
 
@@ -378,6 +380,7 @@ function M.is_enabled(mod, ft)
   for _, parser in pairs(module_config.disable) do
     if ft == parser then return false end
   end
+
   return true
 end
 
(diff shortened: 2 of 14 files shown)


kyazdani42:
@theHamsta @vigoux 
I've tested node_movement a bit, and the behaviour is a little bit unexpected. It's hard to guess which node the cursor is going to jump on, and sometimes it is blocked for no good reason.
I've tried to make the behaviour more consistent, but i did not manage to achieve it.

What should we do ? from a user's perspective i wouldn't use this feature as it feels kind of broken and the default vim behaviour to move around is easier to understand.

I wish we could find something to make this feature really useful, i'm pretty sure it has some potential we have yet to discover, but in the current state, i'm not all in to keep it.
if i can think of something, i'll implement it. If you can think of something, feel free to share, and we will discuss implementation details.

theHamsta:
One part one the problem is, that the grammars for the different languages differ drastically in granularity. So you have to go deeper in node-hierarchy for some languages than for other. It's a lot better with highlighting, when you see at what level you currently are.

Also setting the jump list is a must for this feature since you often want to go back where you were after going `up`. None of these tips was working for me: https://stackoverflow.com/questions/19195160/push-a-location-to-the-jumplist.

Maybe it is also worth it to turn off the extra magic of node_movement or make it configurable (extra magic is the two `true` when calling `get_next_node`). Switching parents while keeping the level might be unexpected.

kyazdani42:
The jump list would actually not be too hard to implement, i already did something like that for the selection (i have a list of ranges to go back if i want to).

So jumping from scope to scope, or jumping from important node to important node would be better.
Maybe the behaviour should be changed, instead of having 'up' or 'down' movement, we could have something like `inner_scope` `outer_scope` `next_scope` `previous_scope` ?

Also we might want to disable the feature if the `locals.scm` file is not present for a language, thus focusing the feature on the specifity of each language.

kyazdani42:
also from the documentation:
`You can explicitly add a jump by setting the ' mark with "m'".  Note
that calling setpos() does not do this.`
But as i recall it didn't work for us

lua/nvim-treesitter/node_movement.lua
```diff
@@ -1,88 +1,76 @@
...
+      end
+    end
+  end,
+  [NodeMovementKind.left] = function(node)
+    return ts_utils.get_previous_node(node, true, true)
+  end,
+  [NodeMovementKind.right] = function(node)
+    return ts_utils.get_next_node(node, true, true)
+  end,
+}
```
theHamsta, line 36: :+1:

lua/nvim-treesitter/node_movement.lua
```diff
@@ -1,88 +1,76 @@
...
-      destination_node = utils.get_next_node(current_node, true, true)
-    end
-    M.current_node[buf] = destination_node or current_node
-  end
+  local destination_node = get_node_fn[kind](current_node)
 
   if destination_node then
-    node_start_to_vim(destination_node)
+    local row, col = destination_node:start()
+    vim.fn.setpos(".", { buf, row+1, col+1, 0 })
```
theHamsta, line 51: Could you set `m'`? Apparently, this is the offical ways to set jumps though it was not working in every case for me

lua/nvim-treesitter/node_movement.lua
```diff
@@ -1,88 +1,76 @@
...
+    return ts_utils.get_previous_node(node, true, true)
+  end,
+  [NodeMovementKind.right] = function(node)
+    return ts_utils.get_next_node(node, true, true)
+  end,
+}
 
 M.do_node_movement = function(kind)
-  local buf, line, col = unpack(vim.fn.getpos("."))
+  local buf = api.nvim_get_current_buf()
```
theHamsta, line 39: Finally, this is fixed :yum: ! My bad that I didn't realize this in the beginning.

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -0,0 +1,103 @@
...
+--- Gets the actual text content of a node
+-- @param node the node to get the text from
+-- @param bufnr the buffer containing the node
+-- @return list of lines of text of the node
+function M.get_node_text(node, bufnr)
+  local bufnr = bufnr or api.nvim_get_current_buf()
+  if not node then return {} end
+
+  -- We have to remember that end_col is end-exclusive
+  local start_row, start_col, end_row, end_col = node:range()
```
theHamsta, line 14: I really hate `node:range()`. We could think about using the LSP ranges in another refactor PR:
```lua
function M.node_to_lsp_range(node)
  local start_line, start_col, end_line, end_col = node:range()
  local rtn = {}
  rtn.start = { line = start_line, character = start_col }
  rtn['end'] = { line = end_line, character = end_col }
  return rtn
end
```

theHamsta:
The refactoring look nice for me so far. It really cleans up the code!

lua/nvim-treesitter/node_movement.lua
kyazdani42: in the documentation they explicitly say setpos doesn't allow setting marks. i'll try it though.

lua/nvim-treesitter/ts_utils.lua
kyazdani42: i agree it is quite disturbing, we should do something like that indeed.

lua/nvim-treesitter/node_movement.lua
kyazdani42: actually the node_movement is not working properly, i'm currently rewriting it

theHamsta:
Regarding the node_movement / node shifting. It's really nice for lisp where it's obvious on which node you are. Node shifting for other languages makes only at some levels of the node hierarchy sense. 

lisp

nvim-treesitter finally enables a hack-free paredit/vim-sexp!

kyazdani42:
> nvim-treesitter finally enables a hack-free paredit/vim-sexp!

This is really cool ! I don't know if it will match the new way of moving between nodes though ... I'm gonna push some modifications soon, you'll be able to try it out by yourself and tell me if it is still convenient

lua/nvim-treesitter/ts_utils.lua
vigoux: Or we should modify the way `node:range` works on neovim's side.

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -0,0 +1,211 @@
...
+  local iter_node = node
+
+  while iter_node ~= nil do
+    local row_, col_ = iter_node:start()
+    if vim.tbl_contains(scopes, iter_node) and (row_+1 ~= row or col_ ~= col) then
+      return iter_node
+    end
+    iter_node = iter_node:parent()
+  end
+end
```
vigoux, line 122: Why not using this in `scope_incremental` ?

lua/nvim-treesitter/state.lua
```diff
@@ -0,0 +1,114 @@
...
+    selection = {
+      range = nil,
+      nodes = {}
+    },
+    parser = parser,
+  }
+
+  M.update()
+  api.nvim_buf_attach(bufnr, false, {
+      -- TODO(kyazdani): on lines should only parse the changed content
```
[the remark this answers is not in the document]
kyazdani42: yes, we could actually look at bfredl's bytetrack implementation

lua/nvim-treesitter/ts_utils.lua
kyazdani42: parent_scope behave relative to the start of the cursor position. It is a little bit disturbing, but the incremental selection behaves relative to selection ranges. Though scope_incremental could use the `containing scope` function

lua/nvim-treesitter/incremental_selection.lua
```diff
@@ -1,77 +1,108 @@
...
 end)
 
 M.scope_incremental = select_incremental(function(node)
-  if node then
-    return utils.smallest_containing_scope(node:parent() or node)
+  local scopes = locals.get_scopes(bufnr)
+  local current = node:parent()
+  while current ~= nil and not vim.tbl_contains(scopes, current) do
+    current = current:parent()
   end
```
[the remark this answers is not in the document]
kyazdani42: well containing_scope already does that.

lua/nvim-treesitter/configs.lua
```diff
@@ -218,28 +220,28 @@ local config = {
...
       keymaps = {
-        move_up = "<a-k>",
-        move_down = "<a-j>",
-        move_left = "<a-h>",
-        move_right = "<a-l>",
+        parent_scope = "tk",
+        child_scope = "tj",
+        next_scope = "tl",
+        previous_scope = "th",
       },
```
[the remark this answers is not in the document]
kyazdani42: Well I tried to use the node movement, and it really made no sense to me, i am used to move around with the regular vim motions which is still the best way to move in vim imho. 

But in most languages (aka lua, c, rust, c++, java,(*script)...) you are moving a lot between scopes. I'm not a python developer, but it seems to me that `if` and `for` still kind of represent scope, delimited by tabulations though.

Maybe node movement would be better for something like switching nodes as you showed earlier. That's why i left those functions around.

The issue with the nodes representing multiple nodes at the same cursor position is actually pretty disturbing, that's why i decided to just use the smallest named node under the cursor, so we have more control over what we want to do. Because you can expand from there to a bigger scope.

But i'd have to say, even with the scope movement, i find it a little bit hard to understand how to move as quickly as i'm used to.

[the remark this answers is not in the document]
theHamsta: Ok. I get that. Then I move that stuff simply to my personal configuration.

I made Python locals.scm where I cheated also a little bit and made `for`/`while` scopes. Tt's possible but then you have to be careful whether people define scope by something they would like to move around or by the actual scoping rules variable.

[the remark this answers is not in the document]
theHamsta: Maybe it's then better to define movements for specific syntax nodes then? Like `next_function`/`previous_function`... vim has `]]` but that's only working on text not the real syntax.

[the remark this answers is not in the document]
kyazdani42: i actually thought about this. Scope movement is quite disturbing, you never really know where you are going.

I think that instead of providing the bindings ourselves, we should let the users decide and provide utility functions so they can basically do what they want with it. 
I already started stabilizing the api with this PR, i will work on providing this api as simple as possible in my next PR.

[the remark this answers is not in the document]
vigoux: Pretty low level concern : the `t...` mappings are already used by default neovim. We may want to keep the previous ones.

lua/nvim-treesitter.lua
```diff
@@ -18,41 +17,25 @@ function M.setup(lang)
...
-
-    expr = expr:parent()
-  end
-
-  if expr then
-    return "..." .. indicator
-  else
-    return indicator
-  end
+  return current_node:sexpr()
```
vigoux, line 38: Why changing this ? It was useful to see the full path to the current node.

kyazdani42: Mmhh that's why i asked you to review this. `sexpr` already gives you the full node, i don't really know if getting the whole path will hold in a statusline. But i haven't tried it.

vigoux: The way it worked was displaying the path from root to the current node, truncating if it is too long, like so : 
```
...->block->call_expression->identifier
```
Or if everything fits:
```
source_file->function_item->block
```

(Examples using the Rust parser)

PR #48, opened 2020-05-09
Fix invalid cursor position for tree root

vigoux:
This happens when trying to select the whole program using incramental selection.

Also adds the `program` node to lua locals.

### DIFF
--- a/lua/nvim-treesitter/incremental_selection.lua
+++ b/lua/nvim-treesitter/incremental_selection.lua
@@ -8,6 +8,11 @@ local function node_range_to_vim(node)
 
   local start_row, start_col, end_row, end_col = node:range()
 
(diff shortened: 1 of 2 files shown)


lua/nvim-treesitter/incremental_selection.lua
```diff
@@ -8,6 +8,11 @@ local function node_range_to_vim(node)
 
   local start_row, start_col, end_row, end_col = node:range()
 
+  if end_row + 1 > vim.fn.line('$') then
+    end_row = vim.fn.line('$') - 1
+    end_col = #vim.fn.getline('$') - 1
+  end
```
kyazdani42, line 14: you can remove this, i will add it in the refactoring

PR #49, opened 2020-05-09
[WIP]: Node shifting

theHamsta:
In the current state this is just a hack. The feature should be waiting for #45 #42 and on a decision on the general future of node_movement.

Just placing it here for some of the utility functions that might be handy.

Works well for lisp, where almost all swapping of parens is admissible 
lisp

For other languages you have to be careful to not destroy the syntax while moving the node
java
In this gif everything works fine. However if I were to move a node after the position of a line comment, the line comment annihilates this node from the syntax tree and it can not be moved further (since it got merged with the comment).

For those languages, only swapping of functions, parameters or statements could make sense.

### DIFF
--- a/lua/nvim-treesitter/utils.lua
+++ b/lua/nvim-treesitter/utils.lua
@@ -130,7 +130,7 @@ function M.replace_node_text(buf, node, replacement_lines)
 end
 
 
---- Swaps the contents of two nodes returning new range of destination (LSP range)
+--- Swaps the contents of two nodes returning new ranges (LSP ranges)
 -- @param buf           buffer number
 -- @param source        first node
 -- @param destination   second node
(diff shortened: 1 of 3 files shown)


lua/nvim-treesitter/utils.lua
```diff
@@ -64,6 +64,193 @@ function M.is_parent(dest, source)
   return false
 end
 
+function M.string_to_lines(str)
+  local t={}
+  local line_breaks = 0
+  for line in str:gmatch("(.-[\n(\r\n)\r])") do
+    table.insert(t, line)
+    line_breaks = line_breaks + 1
+  end
```
vigoux, line 73: That seems quite complicated, why not using `vim.fn.split` here ?

theHamsta: This function is not used anymore. All the editing part is down by `vim.lsp.utils.apply_textedits`. As I said, I just piled up random code here. Should be fixed after the big refactoring.

theHamsta:
Ok, since node movement is removed I am closing this. I try to come up with a PR only containing reusable utility functions for editing nodes (change text, swap them, reparse) without direct end-user functionality.

PR #50, opened 2020-05-09
Export configs.parsers to enable users to register their own parsers

theHamsta:
Export configs.parsers to enable users to register their own parsers from the internet or override the repo if they prefer an alternative parser.

E.g.

```lua
    require "nvim-treesitter.configs".parsers.lisp = {
        install_info = {
            url = "https://github.com/theHamsta/tree-sitter-clojure",
            files = {"src/parser.c"}
        }
    }
```

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -245,6 +245,7 @@ local config = {
 }
 
 local M = {}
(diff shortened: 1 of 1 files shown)


theHamsta:
There is alread `require "nvim-treesitter.configs".get_parser_configs()`

Issue #52, opened 2020-05-15
No locals.scm query found for c, cpp and rust


yashLadha:
```
health#nvim_treesitter#check
========================================================================
## Installation
  - OK: `git` executable found.
  - OK: `cc` executable found.

## c parser healthcheck
  - OK: c parser found.
  - WARNING: No `locals.scm` query found for c
    - ADVICE:
      - Open an issue at https://github.com/nvim-treesitter/nvim-treesitter
  - WARNING: No `highlights.scm` query found for c
    - ADVICE:
      - Open an issue at https://github.com/nvim-treesitter/nvim-treesitter

## cpp parser healthcheck
  - OK: cpp parser found.
  - WARNING: No `locals.scm` query found for cpp
    - ADVICE:
      - Open an issue at https://github.com/nvim-treesitter/nvim-treesitter
  - WARNING: No `highlights.scm` query found for cpp
    - ADVICE:
      - Open an issue at https://github.com/nvim-treesitter/nvim-treesitter

## lua parser healthcheck
  - OK: lua parser found.
  - OK: `locals.scm` found.
  - OK: `highlights.scm` found.

## rust parser healthcheck
  - OK: rust parser found.
  - WARNING: No `locals.scm` query found for rust
    - ADVICE:
      - Open an issue at https://github.com/nvim-treesitter/nvim-treesitter
  - WARNING: No `highlights.scm` query found for rust
    - ADVICE:
      - Open an issue at https://github.com/nvim-treesitter/nvim-treesitter

## Missing parsers
  - WARNING: Some parsers are not installed:
    csharp
    html
    typescript
    swift
    java
    python
    nix
    yaml
    elm
    vue
    ruby
    ocaml
    go
    scala
    haskell
    json
    toml
    javascript
    css
    julia
    markdown
    php
    bash
    tsx
    - ADVICE:
      - Install them using `:TSInstall language

```

```
NVIM v0.5.0-504-g55b62a937
Build type: Release
LuaJIT 2.1.0-beta3
[...]

PR #53, opened 2020-05-15
Add C queries

vigoux:
Also add a missing `field` capture in locals query.

Fixes partially #52.

### DIFF
--- a/queries/c/locals.scm
+++ b/queries/c/locals.scm
@@ -1,0 +1,36 @@
+;; Functions definitions
+(function_declarator
+  declarator: (identifier) @definition.function) @scope
+(preproc_function_def
+  name: (identifier) @definition.macro) @scope
+
+(preproc_def
+  name: (identifier) @definition.macro)
+(pointer_declarator
+  declarator: (identifier) @definition.variable)
+(parameter_declaration
+  declarator: (identifier) @definition.variable)
+(init_declarator
+  declarator: (identifier) @definition.variable)
+(array_declarator
+  declarator: (identifier) @definition.variable)
+(declaration
+  declarator: (identifier) @definition.variable)
+(enum_specifier
+  name: (*) @definition.type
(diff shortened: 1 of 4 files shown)


queries/c/locals.scm
```diff
@@ -0,0 +1,31 @@
...
+    (enumerator name: (identifier) @definition.variable)))
+
+;; Type / Struct / Enum
+(field_declaration
+  declarator: (field_identifier) @definition.field)
+(type_definition
+  declarator: (type_identifier) @definition.type)
+
+;; References
+(identifier) @reference
```
kyazdani42, line 31: shouldn't we define scopes for this too ?

queries/c/locals.scm
```diff
@@ -0,0 +1,34 @@
...
+
+;; Scope
+(for_statement) @scope
```
kyazdani42, line 34: we'll add more scopes another time i guess.
You should also update the README though.

vigoux: Actually, C is quite poor in scopes, even `for_statement` is added for convenience, the only real scope used in C is function based I think.

kyazdani42: well to me every condition blocks, while loops, do for loops, switch statement and so on are scopes. But that's ok we'll leave it for another time as i said :)

vigoux: That's the weird part of C, nothing despite functions and files are scopes ;-)

kyazdani42: well i'm not sure but if you do something like (that you shouldnt do though)
```c
if (something) {
  int a = 0;
...
}
```
the a is stored in a register and removed from the register when you are going out of the `if` scope ?

vigoux: You are right, how dumb am I. I'm adding the scopes right now.

vigoux:
Should be better by now

PR #54, opened 2020-05-15
Expose internal api.

kyazdani42:
- add `exposed_state` to expose 'current_node' and 'cursor_pos'
for a current buffer to the user.
- add `get_buf_state` and `get_node_api` for users.
- add documentation about api functions.
- remove `node_movement` module which should be done in user side.


vigoux:
I am quite misled by the `cursor_pos` thing. Is it the current cursor position in the buffer ? Isn't this the same as `vim.fn.getpos(".")` ?

kyazdani42:
It's the same, but i exposed it for convenience, one call of `getpos` should be sufficient, there are no point to make multiple ones.

PR #56, opened 2020-05-15
Add python locals.scm

theHamsta:
In Python, it's a bit tricky to define scopes.

Normal control structures like `if`, `for`, `while` don't define scopes
```python
i = 5
for i in range(3):
    pass
print(i)
#>>> 2
```

### DIFF
--- a/queries/python/highlights.scm
+++ b/queries/python/highlights.scm
@@ -103,7 +103,7 @@
 
 ; Keywords
 
-"as" @keyword
+"as" @include
 "assert" @keyword
 "async" @keyword
 "await" @keyword
@@ -118,10 +118,10 @@
 "exec" @keyword
 "finally" @keyword
 "for" @repeat
-"from" @keyword
+"from" @include
 "global" @keyword
 "if" @conditional
-"import" @keyword
+"import" @include
 "lambda" @keyword
 "nonlocal" @keyword
 "pass" @keyword
--- a/queries/python/highlights.scm
+++ b/queries/python/highlights.scm
@@ -1,0 +1,176 @@
+;; From tree-sitter-python licensed under MIT License
+; Copyright (c) 2016 Max Brunsfeld
+
+; Identifier naming conventions
+
+
+((import_from_statement
+   name: (dotted_name
+           (identifier)) @type)
+ (match? @type "^[A-Z]"))
+
+((identifier) @constant
+ (match? @constant "^[A-Z][A-Z_]*$"))
+
+; Function calls
+
+(decorator) @function
+
+(call
+  function: (attribute
+              attribute: (identifier) @method))
+
+(call
+  function: (identifier) @function)
+
+((call
+   (identifier) @constructor)
+ (match? @constructor "^[A-Z]"))
+
+;; Builtin functions
+
+((call
+  function: (identifier) @function.builtin)
+ (match?
+   @function.builtin
+   "^(abs|all|any|ascii|bin|bool|breakpoint|bytearray|bytes|callable|chr|classmethod|compile|complex|delattr|dict|dir|divmod|enumerate|eval|exec|filter|float|format|frozenset|getattr|globals|hasattr|hash|help|hex|id|input|int|isinstance|issubclass|iter|len|list|locals|map|max|memoryview|min|next|object|oct|open|ord|pow|print|property|range|repr|reversed|round|set|setattr|slice|sorted|staticmethod|str|sum|super|tuple|type|vars|zip|__import__)$"))
+
+;; Function definitions
+
+(function_definition
+  name: (identifier) @function)
+
+(identifier) @variable
+(attribute attribute: (identifier) @property)
+(type (identifier) @type)
+(parameters
+  (identifier) @parameter) 
+
+; Literals
+
+(none) @constant.builtin
+(true) @boolean
+(false) @boolean
+((identifier) @constant.builtin
+              (match? @constant.builtin "self"))
+
+(integer) @number
+(float) @float
+
+(comment) @comment
+(string) @string
+(escape_sequence) @escape
+
+(interpolation
+  "{" @punctuation.special
+  "}" @punctuation.special) @embedded
+
+; Tokens
+
+"-" @operator
+"-=" @operator
+"!=" @operator
+"*" @operator
+"**" @operator
+"**=" @operator
+"*=" @operator
+"/" @operator
+"//" @operator
+"//=" @operator
+"/=" @operator
+"&" @operator
+"%" @operator
+"%=" @operator
+"^" @operator
+"+" @operator
+"+=" @operator
+"<" @operator
+"<<" @operator
+"<=" @operator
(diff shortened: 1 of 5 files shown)


vigoux:
Do you want to add the `highlights.scm` while you are at it ? You could use the query provided by the parser for this [here](https://github.com/tree-sitter/tree-sitter-python/blob/master/queries/highlights.scm).

vigoux:
Oh and you should update the README to notify that you maintain the python queries.

theHamsta:
Had `highlights.scm` already in my working tree, had only to add it. Does it realy make sense to maintain own versions of the highlights instead of just downloading the ones from the respective repos?

At least for the unmaintained repos we could automatically install queries on `:TSInstall`

kyazdani42:
well the ones from the repos are just basic highlights, they can be extended big time. And you might want to highlight your code in a different way. 
What we will provide is a way to load user highlight queries, so it'll be easier for people to make advanced colorschemes.

queries/python/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+(call
+  function: (identifier) @function)
+
+;; Builtin functions
+
+;((call
+  ;function: (identifier) @function.builtin)
+ ;(#match?
+   ;@function.builtin
+   ;"^(abs|all|any|ascii|bin|bool|breakpoint|bytearray|bytes|callable|chr|classmethod|compile|complex|delattr|dict|dir|divmod|enumerate|eval|exec|filter|float|format|frozenset|getattr|globals|hasattr|hash|help|hex|id|input|int|isinstance|issubclass|iter|len|list|locals|map|max|memoryview|min|next|object|oct|open|ord|pow|print|property|range|repr|reversed|round|set|setattr|slice|sorted|staticmethod|str|sum|super|tuple|type|vars|zip|__import__)$"))
```
vigoux, line 27: Why keeping this commented ? It sounds like a great addition

queries/python/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+"lambda" @keyword
+"nonlocal" @keyword
+"pass" @keyword
+"print" @keyword
+"raise" @keyword
+"return" @keyword
+"try" @keyword
+"while" @keyword
+"with" @keyword
+"yield" @keyword
```
vigoux, line 123: Some of them should be marked as `conditional` and `repeat`.
```suggestion
"as" @keyword
"assert" @keyword
"async" @keyword
"await" @keyword
"break" @repeat
"class" @keyword
"continue" @repeat
"def" @keyword
"del" @keyword
"elif" @conditional
"else" @conditional
"except" @keyword
"exec" @keyword
"finally" @keyword
"for" @repeat
"from" @keyword
"global" @keyword
"if" @conditional
"import" @keyword
"lambda" @keyword
"nonlocal" @keyword
"pass" @keyword
"print" @keyword
"raise" @keyword
"return" @keyword
"try" @keyword
"while" @repeat
"with" @keyword
"yield" @keyword
```

queries/python/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+  (identifier) @parameter) 
+
+; Literals
+
+(none) @constant.builtin
+(true) @constant.builtin
+(false) @constant.builtin
+
+(integer) @number
+(float) @number
```
vigoux, line 47: We have a special capture group for this : `float`.
```suggestion
(float) @float
```

queries/python/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+
+;((identifier) @constant
+ ;(#match? @constant "^[A-Z][A-Z_]*$"))
+
+; Function calls
+
+(decorator) @function
+
+(call
+  function: (attribute attribute: (identifier) @function.method))
```
vigoux, line 17: For us this is only `method`.
```suggestion
  function: (attribute attribute: (identifier) @method))
```

queries/python/highlights.scm
```diff
@@ -0,0 +1,150 @@
+;; From tree-sitter-python licensed under MIT License
+; Copyright (c) 2016 Max Brunsfeld
+
+; Identifier naming conventions
+
+;((identifier) @constructor
+ ;(#match? @constructor "^[A-Z]"))
+
+;((identifier) @constant
+ ;(#match? @constant "^[A-Z][A-Z_]*$"))
```
vigoux, line 10: Why commenting these too ?

queries/python/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+(integer) @number
+(float) @number
+
+(comment) @comment
+(string) @string
+(escape_sequence) @escape
+
+(interpolation
+  "{" @punctuation.special
+  "}" @punctuation.special) @embedded
```
vigoux, line 55: `embedded` groups are not yet supported, this will not have any effect for now but hopefully it will be supported shortly.

queries/python/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+"yield" @keyword
+
+; Additions for nvim-treesitter
+"(" @punctuation.bracket
+")" @punctuation.bracket
+"[" @punctuation.bracket
+"]" @punctuation.bracket
+
+"," @punctuation.delimiter
+"." @punctuation.delimiter
```
vigoux, line 132: Don't you think that we can also add `:` as a delimiter ?

queries/python/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+(attribute attribute: (identifier) @property)
+(type (identifier) @type)
+(parameters
+  (identifier) @parameter) 
+
+; Literals
+
+(none) @constant.builtin
+(true) @constant.builtin
+(false) @constant.builtin
```
vigoux, line 44: why is there not `self` here ? I know it is actually not a constant but having it highlighted could be some "recommandation" to use it.

queries/python/highlights.scm
theHamsta: The queries from tree-sitter-python cause an invalid-query-syntax error.
I changed the `#match?` to `match?`.

queries/python/highlights.scm
theHamsta: Same as above. `#match?` is invalid.

queries/python/highlights.scm
theHamsta: `;` is at least not. Have to look up what it is.

queries/python/highlights.scm
theHamsta: I know. I also put a reminder to embed RST in documentation and `>>>` as Python blocks in documentation.

queries/python/highlights.scm
vigoux: We will need to change this again when the treesitter update will be merged.

theHamsta:
I added an `@include` to use the `Include` highlight like the built-in Python syntax (`python.vim`).

theHamsta:
TODO: make `as` a @include

PR #58, opened 2020-05-17
[WIP]: Add @definition.*.doc to CONTRIBUTING.md

theHamsta:
We should add doc to definitions as soon as more advanced query syntax is available.
```scheme
(function_definition
  name: (identifier) @definition.function
  body: (block (expression_statement (string) @definition.function.doc)?)) @scope
```
And instructions that strip away the quotes. (strip is used by official treesitter).

```lisp
    (function_definition
      name: (identifier) @name
      body: (block . (expression_statement (string) @doc))) @function
    (#strip! @doc "(^['\"\\s]*)|(['\"\\s]*$)")
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -94,6 +94,8 @@ are optional and will not have any effect for now.
   `type`
   `field`
 
+If you can extract documentation adjacent to a definition use `@definition.*.doc` (e.g. `@definition.function.doc`).
+
 `@scope`
 
 `@reference`


theHamsta:
I guess the strip of quotes is not absolutely necessary. But we should wait until `?` is allowed. Treesitter was also using a `adjacent!` instruction to combine two queries, like a group of comments above any kind of definition.

vigoux:
Now that the runtime has been updated, we should add the appropriate predicates to the queries lua file

theHamsta:
Already added it to the Go PR.
But as @definition.doc in a @definition.function query.

PR #62, opened 2020-05-22
Some improvements to c/locals.scm

theHamsta:
- Fix function scope (was declaration only without body)
- Use @definition.var like the other local files (lua, ruby)
- Add translation_unit scope
~~- Add variable definitions by function parameters~~ not necessary! Already included in available queries.

I was experimenting with resolving variables for GCC and accidentally adapted the C locals.scm a bit. @vigoux, maybe some changes of this PR are useful...

### DIFF
--- a/queries/c/locals.scm
+++ b/queries/c/locals.scm
@@ -1,25 +1,25 @@
 ;; Functions definitions
 (function_declarator
-  declarator: (identifier) @definition.function) @scope
+  declarator: (identifier) @definition.function)
 (preproc_function_def
   name: (identifier) @definition.macro) @scope
 
 (preproc_def
   name: (identifier) @definition.macro)
 (pointer_declarator
-  declarator: (identifier) @definition.variable)
+  declarator: (identifier) @definition.var)
 (parameter_declaration
-  declarator: (identifier) @definition.variable)
+  declarator: (identifier) @definition.var)
 (init_declarator
-  declarator: (identifier) @definition.variable)
+  declarator: (identifier) @definition.var)
 (array_declarator
-  declarator: (identifier) @definition.variable)
+  declarator: (identifier) @definition.var)
 (declaration
-  declarator: (identifier) @definition.variable)
+  declarator: (identifier) @definition.var)
 (enum_specifier
   name: (*) @definition.type
   (enumerator_list
-    (enumerator name: (identifier) @definition.variable)))
+    (enumerator name: (identifier) @definition.var)))
 
 ;; Type / Struct / Enum
 (field_declaration
@@ -34,3 +34,5 @@
 (for_statement) @scope
 (if_statement) @scope
 (while_statement) @scope
+(translation_unit) @scope
+(function_definition) @scope


queries/c/locals.scm
```diff
@@ -1,25 +1,44 @@
...
+  (parameter_declaration 
+    declarator: (pointer_declarator
+                  declarator: (identifier) @definition.var)))
+
+; pointers to pointers
+(parameter_list
+  (parameter_declaration 
+    declarator: (pointer_declarator
+                  declarator: (pointer_declarator
+                                declarator: (identifier) @definition.var))))
```
vigoux, line 41: The problem with C is those pointers, I think that we may want to use something like
```suggestion
(parameter_list
  (parameter_declaration 
    declarator: (.*) @declaration.var))
```
 And then let the user use it

theHamsta: But then you will have the whole variable declaration, not only the type. Those to declarations strip away everything for the user (also possible const, volatile madness).

I admit that there's has to be a more elegant way to write this but if not repeating the nested structure won't hurt us to much (assuming that triple pointers are very rare).

vigoux: Maybe this shows that we may want to add a predicate which would take the first child matching a given query, something like `(submatch? @decl "((identifier) @declaration.var)"`

theHamsta: Another, possibility would to use the `#strip!` instruction that they use in the treesitter repo:
```((parameter_list
  (parameter_declaration 
    declarator: (*) @definition.var))
    (#strip!  @definition.var "^.*\\s"))
```
Then one could only select the right most word of the query.

theHamsta: Ok. I saw that the thing with the parameter decelerator is not necessary any more when the function block now belongs to the function scope. The query: 
```scheme
(pointer_declarator
  declarator: (identifier) @definition.var)
```
already defines the pointer correctly.

theHamsta:
I guess the queries for c are really useful. Will you extend them for C++? I guess the queries should be already useful for C++ as they are right now.

theHamsta:
C queries working fine for resolving variable definitions while debugging :yum: :
image

vigoux:
> Will you extend them for C++?
I don't right C++ at all, but if anyone wants to add it based on the C query, it will be highly appreciated !

vigoux:
> C queries working fine for resolving variable definitions while debugging 😋 :
> image

Whoa ! Is this a plugin you are writing now ?

theHamsta:
It's just an extension to nvim-dap https://github.com/mfussenegger/nvim-dap/pull/13

I use nvim-treesitter to resolve the definition position for the variables in scope.

theHamsta:
I never write C only C++. I can do it. But we should probably introduce the concept of one language extending another for the queries. The treesitter-grammar for C++ also extends the C grammar.

So the C++ locals/highlights could include your definitions, plus some extra queries. 

Looking forward for your Rust queries!

theHamsta:
Maybe it's also better to use "@defintion.variable"... It's only a lot easier for users when all languages have  the same captures.

lua/nvim-treesitter/query.lua
```diff
@@ -7,10 +7,30 @@ local function read_query_file(fname)
...
+
+  for _, base_language in pairs(M.base_language_map[ft] or {}) do
+    local base_files = api.nvim_get_runtime_file(string.format('queries/%s/%s.scm', base_language, query_name), false)
+    if base_files and #base_files > 0 then
+      query_string = read_query_file(base_files[1])..'\n'..query_string
+    end
+  end
+
+  if #query_string > 0 then
+    return ts.parse_query(ft, query_string)
```
vigoux, line 33: sounds great !

lua/nvim-treesitter/query.lua
```diff
@@ -71,4 +91,6 @@ function M.iter_prepared_matches(query, qnode, bufnr, start_row, end_row)
   end
 end
 
+
+
```
kyazdani42, line 95: great stuff :) Can you remove these ?

theHamsta: Ooops. wrong PR with an outdated commit. That happens during rebase. Also fixed this in #63

vigoux:
Sounds great @kyazdani42, discussion will continue on #63

PR #63, opened 2020-05-23
 Introduce base languages for queries (for C++ locals)

theHamsta:
This PR should wait for #62 .

Some treesitter grammars just extend another treesitter grammar.
This enables us to use the C queries also for C++.
We only need to put additional queries in the C++ files to support additional C++ features (#52 ).

I would also like to include those additional queries for C++ in this PR to see whether this concept works well. Right now, C++ files have out-of-the-box syntax hightlighting from the C queries.

### DIFF
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -13,11 +13,31 @@ local function read_query_files(filenames)
   return table.concat(contents, '\n')
 end
 
+-- Some treesitter grammars extend others.
+-- We can use that to import the queries of the base language
+M.base_language_map = {
+  cpp = {'c'},
+  typescript = {'javascript'},
+  tsx = {'typescript', 'javascript'},
+}
+
 function M.get_query(ft, query_name)
   local query_files = api.nvim_get_runtime_file(string.format('queries/%s/%s.scm', ft, query_name), true)
+  local query_string = ''
 
   if #query_files > 0 then
-    return ts.parse_query(ft, read_query_files(query_files))
+    query_string = read_query_files(query_files)..query_string
+  end
+
+  for _, base_lang in ipairs(M.base_language_map[ft] or {}) do
+    local base_files = api.nvim_get_runtime_file(string.format('queries/%s/%s.scm', base_lang, query_name), false)
+    if base_files and #base_files > 0 then
+        query_string = read_query_files(base_files)..query_string
+    end
+  end
+
+  if #query_string > 0 then
+    return ts.parse_query(ft, query_string)
   end
 end
 
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ List of currently supported languages:
 - [x] ruby (maintained by @TravonteD)
 - [x] c (maintained by @vigoux)
 - [x] go (maintained by @theHamsta)
-- [ ] cpp
+- [x] cpp (maintained by @theHamsta, extends C queries)
 - [ ] rust
 - [x] python (maintained by @theHamsta)
 - [ ] javascript
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -94,10 +94,10 @@
 (sized_type_specifier) @type
 
 ((identifier) @type
- (match? @type "^[A-Z]"))
+ (#match? @type "^[A-Z]"))
 
 ((identifier) @constant
- (match? @constant "^[A-Z][A-Z\\d_]+$"))
+ (#match? @constant "^[A-Z][A-Z\\d_]+$"))
 
 (comment) @comment
 
@@ -109,3 +109,4 @@
   (identifier)) @parameter
 
 (ERROR) @error
+
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -31,7 +31,6 @@
 
 "--" @operator
 "-" @operator
-"-=" @operator
 "->" @operator
 "!=" @operator
 "*" @operator
@@ -40,7 +39,6 @@
 "&&" @operator
 "+" @operator
 "++" @operator
-"+=" @operator
 "<" @operator
 "<=" @operator
 "==" @operator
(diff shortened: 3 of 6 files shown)


theHamsta:
Seems to work fine. Most of the logic for C++ is already in the C queries.

vigoux:
We will merge this after #62.

lua/nvim-treesitter/query.lua
```diff
@@ -7,10 +7,30 @@ local function read_query_file(fname)
   return table.concat(vim.fn.readfile(fname), '\n')
 end
 
+-- Some treesitter grammars extend others.
+-- We can use that to import the queries of the base language
+M.base_language_map = {
+  cpp = { 'c' },
+  tsx = { 'js' },
+}
```
kyazdani42, line 15: could you only make key to strings mapping ? Or could a language have multiple base languages ?

theHamsta: I though this would be a more general solution if the situation of multiple base languages ever arrises. If a user creates their own mash-up language parser they can import multiple queries.

But in general this problem is tackled by embedded languages. The feature is a bit newer so I don't know whether all web-stuff parsers already use it or whether they just import another grammar and create a single AST.

theHamsta: Also this is currently the only way to extend extensions. For my fantasy-cpp-parser I could import both `c` and `cpp`.

kyazdani42: To me a base language is a language from which you extend to other ones. Embedding another language is another issue (C in CPP vs HTML in python is definitely not the same thing).
I don't know how embedding solutions will work though, maybe parser definitions should have associated filetypes like one extended language and potential embedded languages. @vigoux what's you opinion on this matter ?

theHamsta: I guess we can forget about the one-language-embedding-multiple-others case.

The only question is whether dependency chains are possible: C extends B and B extends A. Then, C would require queries A and B.

theHamsta: Or should we just simplify to pure string until we encounter the problem for the first time?

kyazdani42: > I guess we can forget about the one-language-embedding-multiple-others case.

Are you sure ? I mean html does have css and javascript but embedding will probably mean different trees, different parsers and so different highlight files.

For now, we should simplify to pure string. But it might be problematic for stuff like javascript in typescript in react. Not sure, but if this is the only case, i think we can forget about multiple highlighting files and create one for this specific purpose.

theHamsta: But in the case of injected languages, we will just have complete distinct set of queries. We can not simply concatenate..

theHamsta:
I made some research through some of the official tree-sitter parser I only found cpp, tsx and typescript extend another grammar (where tsx, and typescript are identical). Extending works by concatenating the dict of the grammar rules with the rules of the base language. Didn't see anything more advanced.

theHamsta:
Does anyone have an issue with opening parens not being highlighted? It only happens on one of my machines (with identical config), only obvious difference is the Neovim revision. The working one is about two weeks behind. It also happens if you just add those two queries to current master:

```scheme
"(" @punctuation.bracket ; not highlighted ???
")" @punctuation.bracket ; highlighted

"[" @punctuation.bracket ; highlighted
"]" @punctuation.bracket ; highlighted
```
Also happens with a minimal init.vim.

theHamsta:
Please do not merge yet! I have encountered some error messages with this highlighting when editing some nodes. I need to find out whether it's a nvim issue or an issue with the queries.

theHamsta:
Ok, seems to be a Neovim problem. To reproduce:

- use only one query:
```scheme
((identifier) @constant
 (match? @constant "^[A-Z][A-Z\\d_]+$"))
```

- edit this expression (e.g. `::resizeRequest,` 
```cpp
    connect(_session, &Konsole::Session::resizeRequest, this, &Konsole::SessionController::sessionResizeRequest);
```
- observe nvim pasting error messages in your code

image

vigoux:
Hmmm.. So that error cam up again.. What type of edit did you do ?

theHamsta:
As I said, for instance changing: `resizeRequest` to `ResizeRequest` by pressing `U`.

theHamsta:
With the new runtime I observe more errors like that even without edits.

kyazdani42:
this is a know issue, which should be solved with the byte resolution update PR if im not mistaken. We will probably need to finish this because the author has not given any news for quite some time

vigoux:
As this is a blocker for some other issues, I think that we should merge this, as the issue mentionned  at https://github.com/nvim-treesitter/nvim-treesitter/pull/63#issuecomment-640045503 is upstream.

vigoux:
@theHamsta can you rebase this on master so that we are sure to have the intended behaviour ?

theHamsta:
Should I return to the dictionary solution? #79 could profit from multiple base languages.

I'll rebase it now "as is".

vigoux:
> Should I return to the dictionary solution? #79 could profit from multiple base languages.

I don't think so. My problem is that having this would point to a feature which does not exist yet : language injection. Thus keeping it like this avoids advertising for things that we can't use yet.

One thing could be to have a little `TODO` comment here though.

theHamsta:
This has nothing to do with language injection. Typescript and tsx are **exactly** the same grammar rules (https://github.com/tree-sitter/tree-sitter-typescript/blob/master/common/define-grammar.js) but split up into to parsers.

So `common.js` imports the JS rules and typescript and tsx imports `common.js` and could at least in theory diverge in future.

vigoux:
Oh okay I don't how bad I understood this sorry, I think I scrambled some PRs in my head.
Let's merge that then

theHamsta:
Without the dictionary solution we would have to duplicate the queries for `tsx` and `typescript`, which should be the same.

kyazdani42:
lets roll back to the dictionnary solution, now that we can merge queries it might be even more useful. Ill read the pr once again but atm, it lgtm

kyazdani42:
Could you update the code so #80 is still relevant ? Thanks !

theHamsta:
Done!

PR #64, opened 2020-05-25
Treesitter based code folding.

vigoux:
This PR brings a new lightweight `fold` module. As folding is only line based for now, how folding works can be quite distrurbing for now (but useful in many cases). When https://github.com/neovim/neovim/pull/12194 will land and some further lua apis will be available, it will be easy to have syntax-based column folding.

### DIFF
--- a/lua/nvim-treesitter/fold.lua
+++ b/lua/nvim-treesitter/fold.lua
@@ -1,0 +1,29 @@
+local api = vim.api
+local parsers = require'nvim-treesitter.parsers'
+
+local M = {}
+
+function M.get_fold_indic(lnum)
+  if not parsers.has_parser() or not lnum then return '0' end
+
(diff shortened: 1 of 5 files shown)


lua/nvim-treesitter/fold.lua
```diff
@@ -0,0 +1,32 @@
+local api = vim.api
+local parsers = require'nvim-treesitter.parsers'
+
+local M = {}
+
+function M.get_fold_indic(lnum)
+  if parsers.has_parser() and lnum then
```
kyazdani42, line 7: `if not parser.has_parser() or not lnum then return '0' end`
then you wouldn't need to have to put everything else in the else ;)

kyazdani42: lgtm to me otherwise, nice job :)

PR #67, opened 2020-05-30
Add css highlight queries

TravonteD:
Adds basic queries for highlighting standard css files. Some of the captures will most likely need changing in the future for better uniformity with other filetypes


vigoux:
Thanks, can you also change the `README.md` to notify that you are maintaining this language ?

PR #69, opened 2020-05-31
Go highlights

theHamsta:
I never do anything in Go, but when I looked at the syntax highlighting of random Go file in vim I had to do something. Mostly copied from tree-sitter-go.

Right now most tree-sitter-repos get a file `tags.scm` added. They include information on how to extract documentation #58 and defininition for call references. The question is whether we want to have call references.

### DIFF
--- a/queries/go/highlights.scm
+++ b/queries/go/highlights.scm
@@ -1,0 +1,136 @@
+;; Forked from tree-sitter-go 
+;; Copyright (c) 2014 Max Brunsfeld (The MIT License)
+
+;;
+; Function calls
+
+(call_expression
+  function: (identifier) @function)
+
+(call_expression
+  function: (selector_expression
+    field: (field_identifier) @method))
+
+; Function definitions
+
+(function_declaration
+  name: (identifier) @function)
+
+(method_declaration
+  name: (field_identifier) @method)
+
+; Identifiers
+
+(type_identifier) @type
+(field_identifier) @property
+(identifier) @variable
+
+(parameter_declaration (identifier) @parameter)
+(variadic_parameter_declaration (identifier) @parameter)
+
+((identifier) @constant
+ (#eq? @constant "_")) 
+
+((identifier) @constant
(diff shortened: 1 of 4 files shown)


queries/go/highlights.scm
```diff
@@ -0,0 +1,138 @@
...
+(parameter_declaration (identifier) @parameter)
+(variadic_parameter_declaration (identifier) @parameter)
+
+((identifier) @constant
+ (eq? @constant "_")) 
+
+;; TODO: This variable is so prominent in Go
+;; it deserves a special hightlight. Which?
+;((identifier) @constant
+ ;(eq? @constant "err"))
```
vigoux, line 37: For this, is using `constant` or `constant.builtin` the kind of solution you are searching for ?
Otherwise, adding a special `variable.builtin` can be a good idea too.

queries/go/locals.scm
```diff
@@ -0,0 +1,62 @@
...
+    ;(#strip! @doc "^//\\s*")
+    ;(#set-adjacent! @doc @function)
+)
+
+(
+    ;(comment)* @doc
+    (method_declaration
+        name: (field_identifier) @definition.method); @method
+    ;(#strip! @doc "^//\\s*")
+    ;(#set-adjacent! @doc @method)
```
vigoux, line 14: We might want to implement these too.
What `#set-adjacent!` is meant to do ?

theHamsta: I have no idea. I guess since it's a tag query this should format the final content of the tag. Or system works differently so we could probably just omit it.

queries/go/highlights.scm
theHamsta: Looking a bit at it. I found it confusing for it to have a different. I will just omit that highlight.

theHamsta:
Updated to new query syntax.

vigoux:
Yep that will be needed for other languages as well now that the runtime has been updated

PR #70, opened 2020-05-31
Add tree-sitter-regex

theHamsta:
This might be interesting for injected highlighting. So we could wait for that to come.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,7 @@ List of currently supported languages:
 - [ ] yaml
 - [ ] nix
 - [ ] markdown
+- [x] regex (maintained by @theHamsta)
 
 ## Troubleshooting
 Before doing anything run `:checkhealth nvim_treesitter`. This will help you find where the bug might come from.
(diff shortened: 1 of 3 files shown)


vigoux:
Oh that's nice !
Language injection is on our todo list, but it requires a bit of work on neovim's side, so we might want to dig into this first.

vigoux:
Can we add this to the `README.md` too ?

theHamsta:
You mean the name of the parser of that it may be useful for language injection?

vigoux:
The name of the parser I think, but also you should add it to the `CONTRIBUTING.md` that one can use this parser for regexes

theHamsta:
I didn't add it to `CONTRIBUTING.md` since I don't know how the language injection stuff works exactly. Maybe you can add a language injection section once your PR is merge in Neovim?

vigoux:
Yep, I need to open it before though...

vigoux:
@theHamsta would you like to be the maintainer of this language too ?

theHamsta:
Would be ok with that.

PR #72, opened 2020-06-02
Updated python highlights

doubleloop:
Probably best to review each commit separately.

### DIFF
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -13,6 +13,7 @@ local hlmap = vim.treesitter.TSHighlighter.hl_map
 hlmap.error = "Error"
 hlmap["punctuation.delimiter"] = "Delimiter"
 hlmap["punctuation.bracket"] = "Delimiter"
+hlmap["punctuation.special"] = "Delimiter"
 
 -- Constants
 hlmap["constant"] = "Constant"
--- a/queries/python/highlights.scm
+++ b/queries/python/highlights.scm
@@ -3,31 +3,32 @@
 
 ; Identifier naming conventions
 
-
-((import_from_statement
-   name: (dotted_name
-           (identifier)) @type)
- (match? @type "^[A-Z]"))
-
+(identifier) @Normal
 ((identifier) @type
  (match? @type "^[A-Z]"))
-
 ((identifier) @constant
- (match? @constant "^[A-Z][A-Z_]*$"))
+ (match? @constant "^[A-Z][A-Z_0-9]*$"))
 
 ; Function calls
 
 (decorator) @function
+((decorator (dotted_name (identifier) @function))
+ (match? @function "^([A-Z])@!.*$"))
+
+(call
+  function: (identifier) @function)
 
 (call
   function: (attribute
               attribute: (identifier) @method))
 
-(call
-  function: (identifier) @function)
+((call
+   function: (identifier) @constructor)
+ (match? @constructor "^[A-Z]"))
 
 ((call
-   (identifier) @constructor)
+  function: (attribute
+              attribute: (identifier) @constructor))
  (match? @constructor "^[A-Z]"))
 
 ;; Builtin functions
@@ -43,10 +44,9 @@
 (function_definition
   name: (identifier) @function)
 
-(identifier) @variable
-(attribute attribute: (identifier) @property)
 (type (identifier) @type)
-((call 
+
+((call
   function: (identifier) @isinstance
   arguments: (argument_list
     (*)
@@ -55,7 +55,7 @@
 
 ; Normal parameters
 (parameters
-  (identifier) @parameter) 
+  (identifier) @parameter)
 ; Default parameters
 (keyword_argument
   name: (identifier) @parameter)
@@ -65,10 +65,10 @@
 ; Variadic parameters *args, **kwargs
 (parameters
   (list_splat ; *args
-    (identifier) @parameter)) 
+    (identifier) @parameter))
 (parameters
   (dictionary_splat ; **kwargs
-    (identifier) @parameter)) 
+    (identifier) @parameter))
 
 
 ; Literals
@@ -84,11 +84,7 @@
 
 (comment) @comment
 (string) @string
-(escape_sequence) @escape
-
-(interpolation
-  "{" @punctuation.special
-  "}" @punctuation.special) @embedded
+(escape_sequence) @string.escape
 
 ; Tokens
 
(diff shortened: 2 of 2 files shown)


vigoux:
Thanks for contributing !
LGTM, let's have know the views of @theHamsta on this too, as he is the language maintainer.

lua/nvim-treesitter/highlight.lua
```diff
@@ -48,6 +49,7 @@ hlmap["type"] = "Type"
 hlmap["type.builtin"] = "Type"
 hlmap["structure"] = "Structure"
 hlmap["include"] = "Include"
+hlmap["variable"] = "Identifier"
```
vigoux, line 52: i don't really agree with this, as "variable" will be the default highlighting for everything then.
What is your opinion on this @kyazdani42 ?

lua/nvim-treesitter/highlight.lua
```diff
@@ -13,6 +13,7 @@ local hlmap = vim.treesitter.TSHighlighter.hl_map
 hlmap.error = "Error"
 hlmap["punctuation.delimiter"] = "Delimiter"
 hlmap["punctuation.bracket"] = "Delimiter"
+hlmap["punctuation.special"] = "Delimiter"
```
vigoux, line 16: You might want to add this to the `CONTRIBUTING.md` file too.

lua/nvim-treesitter/highlight.lua
kyazdani42: well `Identifier` is defined as `any variable name` in vim documentation. not sure what you mean by 'the default highlighting

theHamsta: Now, everything is red with my colorscheme. "Everything" should by rather neutral for my taste, so I would need to make `variable` to no hihglight after this change. How does this @vigoux @kyazdani42 look on your computers? If this change would make the highlighting unusable for most colorschemes we should not do it.

queries/python/highlights.scm
```diff
@@ -3,31 +3,33 @@
...
-           (identifier)) @type)
- (match? @type "^[A-Z]"))
-
+(identifier) @variable
 ((identifier) @type
  (match? @type "^[A-Z]"))
-
 ((identifier) @constant
- (match? @constant "^[A-Z][A-Z_]*$"))
+ (match? @constant "^[A-Z][A-Z_0-9]*$"))
```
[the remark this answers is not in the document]
theHamsta: Could you add a built-in highlight for `assert` and a highlight for `__dunder__` identifiers?

lua/nvim-treesitter/highlight.lua
theHamsta: Also add a highlight `punctuation.bracket` for normal `{` and `}` (e.g. for dicts)

theHamsta:
You should also squash your commits.

lua/nvim-treesitter/highlight.lua
kyazdani42: basically identifier should not be used for variables as red is definitively not a good color. Identifier applies well on vimscript, but not on most languages, specially not python

doubleloop:
>The only issue that I have is that variable highlight. I would prefer to remove the (identifier) @variable query.

the issue I wanted to solve wit 538a697 was this highlight:
Screenshot from 2020-06-03 00-06-55

(here I use orange for `Identifier`, green for `Function` and yellow for `String`)

Screenshot from 2020-06-03 00-06-31

Issues here:
1. `foo` in line 1 should have the same color as foo inside format string
1. It would be desirable to be able to customize color of `foo` in line 1, and because `:h group-name` states that::
>         *Identifier         any variable name
if not `Identifier`, what is reasonable choice here?

kyazdani42:
i guess normal would be more appropriate. Most colorscheme decide to make identifier red because it matches a lot of stuff in vim (and tags for instance)

theHamsta:
We need to make sure that the unspecific `(identifier) @variable` does not override more specific highlights. How is the precedence of highlighting handled right now. Right now @variable was overwriting the built-in highlight of `assert`. This you also be a bug of neovim or by the order we specify the highlights.

vigoux:
I see now, and that's a tricky one.
I think that we can find a way to fix that using those `@embedded` captures you see in some of atom queries.
I don't really know how we could do this but as we investigate language injection (which is the next item in my todo list after https://github.com/neovim/neovim/pull/12321 and updating the queries), we will dive a bit into how we can use multiple parsers, and highlight things correctly.

theHamsta:
Hi. I guess with @variable on top of the file everything is working as it should (`assert` is not overwritten). No special treatment of @embedded is necessary.

The only open question is how @variable should be hightlighted. I am open for any decision but that decision should be propageted to the other languages. 

Thanks again for improving the highlights!

theHamsta:
Could you check why constanst no longer get highlighted properly with your PR?

Without PR
image

With PR:
image

queries/python/highlights.scm
```diff
@@ -169,41 +170,23 @@
...
-    (match? @constant "^[A-Z][A-Z_]*$"))
-
-((attribute
-    attribute: (identifier) @type)
-    (match? @type "^[A-Z][a-z_]+"))
-
-((attribute
-    object: (identifier) @type)
-    (match? @type "^[A-Z][a-z_]+"))
-
```
theHamsta, line 192: Why delete those?

vigoux:
>  No special treatment of `@embedded` is necessary.

If I remenber how python works, `f"4 = {var}"` are interpolated strings, and thus a whole new tree can appear within `{}`, which requires `@embedded`

theHamsta:
In theory yes, but in practice the stuff contained interpolated string is already contained in the main-AST.

At least, it works out-of-the-box right now. But I didn't try out really complicated expressions.

vigoux:
A better way to do it then : 
```
(string) @string
;; Reset highlighting for string interpolation
(interpolation) @Normal

;; Other highlighting afterward
```

theHamsta:
`@embedded` will rock for C macros! I already put a TODO in the C highlights.

(those poor Github users with `@embedded` and `@variable` handles! They get mentioned all the time you forget to escape the `@something`)

theHamsta:
I'm getting the correct hightlights with your PR if I move the following queries further down the file.

- use  "^[A-Z][A-Z_\\d]+$" for constants like in the other repos

```
((identifier) @type
 (match? @type "^[A-Z]"))
((identifier) @constant
 (match? @constant "^[A-Z][A-Z_\\d]+$"))
```

image

After that and squashing this PR should be ready to go. Except @kyazdani42 @vigoux "@variable" should `Normal`? I am ok with `Identifier`. It's as fine as `Normal` for me.

vigoux:
Is	`@varlable` still needed if you do the modification I suggested above?
Having `@variable` to Normal is the same as ignoring it.

theHamsta:
> Is `@varlable` still needed if you do the modification I suggested above?
> Having `@variable` to Normal is the same as ignoring it.
Except that it fixes the F-string issue.

vigoux:
Quick side note: for now the `#match?` predicate uses vim regexes (in very-magic mode), so you can use vim's character classes if needed

theHamsta:
@doubleloop do you want to finish this? or should we just wrap up your changes?

doubleloop:
@theHamsta sry for late response, I was afk

>Could you check why constanst no longer get highlighted properly with your PR

I fixed it in #a4f1081 . I used negation (although vim regex has strange syntax for this), to avoid repetitions (for example uppercase identifiers are constants so instead of writing this rule everywhere we make more specific rule, this rules can have exceptions for constants)

> You should also squash your commits.

If the fixups are ok, I will squash all.

> We need to make sure that the unspecific (identifier) @variable does not overwrite more specific highlights. How is the precedence of highlighting handled right now?

AFAIK if there are conflicting for the highlight, the last one is used. Thus if more specific cases are defined below, it will be ok.

It would be nice if ordering would not matter and this could also work:
```
(call
  function: (identifier) @function)
(identifier) @Normal
```
but I guess it would require some changes in vim.

> With this PR `@variable` was overwriting the built-in highlight of assert

It is not unless on top

> I don't really agree with this, as "variable" will be the default highlighting for everything

I am guessing that interpolation did not work because in case of no rule, vim probably used te color of previous token instead of resetting it. If this is true, setting default highlighting of identifier to Normal at the top of scm file is sane workaround IMO.

Still, naming this default as `variable` was bad idea (misleading and too specific) so I linked it directly to `@Normal` and removed it from `hlmap`

theHamsta:
Well, I still don't know how but I had the fealing that the queries coming first are painted before the ones coming after that. For me, it made a huge difference in which order I ordered the queries. Not sure, if this is desired or just an effect of implementation details in Neovim.

The official treesitter queries have both `(identifier) @variable` and rules for more specific identifiers. Not sure, whether they prioritize by the order of the queries or if they have a priority of the `@` highlights.

vigoux:
Highlighting priority is on its way within neovim, and the priority we are coming for is : the latest match within the query wins.
We can call that painting order if you want : latest applied wins.

doubleloop:
Just squashed the commits

theHamsta:
@vigoux apparently I have no write-access to merge it.

vigoux:
> @vigoux apparently I have no write-access to merge it.

It should be okay now

theHamsta:
Yeah!

vigoux:
Welcome to the team!!
Can we open an issue regarding the debate we had here though?
Some kind of rfc to flesh out how we are going to tackle `@embedded` and `@injection`?

stsewd:
Thanks everyone! I've been waiting for this!

Issue #76, opened 2020-06-07
Update queries to use the new query syntax.


vigoux:
Since https://github.com/neovim/neovim/pull/12321, the query syntax has changed a bit, and thus we need to change our queries here. 

TODO:
 - [x] lua
 - [x] ruby
 - [x] c
 - [x] go
 - [x] python
 - [x] html
 - [x] css

PR #77, opened 2020-06-07
- #76 update the predicate syntax for ruby and css

TravonteD:


queries/ruby/highlights.scm
```diff
@@ -88,7 +88,7 @@
 (keyword_parameter (identifier) @parameter)
 
 ((identifier) @function
- (is-not? local))
+ (#is-not? local))
```
vigoux, line 91: I just see this predicate, and it is not supported (yet)

vigoux:
Thanks for updating these queries !

queries/ruby/highlights.scm
TravonteD: Do we want to remove it for now?

vigoux: Yes maybe, more like commenting it as a TODO for me for example

PR #79, opened 2020-06-12
feat(queries): add typescript and javascript queries

steelsojka:
First stab at typescript/javascript. The typescript queries just include all the queries from javascript but with some typescript specific queries. Is there a way to "inject" or import the javascript queries instead of duplicating them? I included the injection files, even though they are not supported. I can remove them if they are not wanted at this time.

Most of these were copied from https://github.com/tree-sitter/tree-sitter-javascript and https://github.com/tree-sitter/tree-sitter-typescript with a couple modifications.


steelsojka:
This should wait until https://github.com/nvim-treesitter/nvim-treesitter/pull/63 so we can just extend javascript/jsx/typescript/tsx

theHamsta:
Right now, tsx and typescript filetypes have the same grammar rules. It would make sense for #63 to revert to the dictionary solution. Then, `typescript` could extend `javascript` and `tsx` could extend `typescript` (without any new queries. Just copy the typescript queries).

steelsojka:
@theHamsta That was my thought. I will keep an eye on that pr and adjust this one accordingly once it goes in.

queries/javascript/highlights.scm
```diff
@@ -0,0 +1,181 @@
...
+"switch" @keyword
+"target" @keyword
+"throw" @keyword
+"try" @keyword
+"typeof" @keyword
+"var" @keyword
+"void" @keyword
+"while" @keyword
+"with" @keyword
+"yield" @keyword
```
vigoux, line 179: Everything is factorable into one group like this :
```
[
"else",
...
] @keyword
```

Furthermore it might be also good to split those keywords depending on their type (like `@conditionnal` and such). please refer to the `CONTRIBUTING.md` for more informations on that

queries/javascript/locals.scm
```diff
@@ -0,0 +1,35 @@
...
+    (identifier) @local.definition))
+
+(variable_declarator
+  name: (identifier) @local.definition)
+
+; References
+;------------
+
+(identifier) @local.reference
+
```
vigoux, line 35: In our versions of `locals.scm`, we removed the unnecessary `local.` prefix, could you please change that too ?

queries/typescript/highlights.scm
```diff
@@ -0,0 +1,209 @@
...
+
+((shorthand_property_identifier) @constant
+ (#match? @constant "^[A-Z_][A-Z\\d_]+$"))
+
+((identifier) @constructor
+ (#match? @constructor "^[A-Z]"))
+
+((identifier) @variable.builtin
+ (#match? @variable.builtin "^(arguments|module|console|window|document)$")
+ (#is-not? local))
```
vigoux, line 18: `#is?` and `#is-not?` predicates are not yet supported, thus you might not want to keep these matches.

kyazdani42:
yes, with query merging and #63 tsx queries could be used for typescript files

steelsojka:
@kyazdani42 Now that #63 is merged, I'll update this to use it.

steelsojka:
@kyazdani42 This works great now. The only thing I have noticed is sometimes `tsx` files will have the filetype `typescriptreact` instead of tsx, which it won't find the parser for. I thought about making an alias map but there are way to many calls to account for. My short term solution is to just symlink the `tsx` parser to `typescriptreact` or set the filetype to `tsx` instead. Maybe something to look into down the road.

kyazdani42:
ive seen that earlier indeed. Im going to push a fix very soon because there are a few things i also wanted to do relating to javascript parser which will require some discussion.

PR #80, opened 2020-06-12
feat(queries): allow user query overrides

steelsojka:
This allows the users to override built in queries. The use can provide their own file in their configuration at `queries/<filetype>/<query_name>.scm`. These files extend the existing query file, not replace it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,13 @@ List of currently supported languages:
 - [ ] nix
 - [ ] markdown
 
+## User Query Extensions
+
+You can add your own query files by placing a query file in vim's runtime path after `nvim-treesitter` is sourced.
+If the language has a built in query file, that file will be appended to or it will be used (useful for languages not yet supported).
+For example, you can add files to `<vim-config-dir>/after/queries/lua/highlights.scm` to add more queries to lua highlights.
+You can also manually add query paths to the runtime path by adding this to your vim config `set rtp+='path/to/queries'`.
+
 ## Troubleshooting
 Before doing anything run `:checkhealth nvim_treesitter`. This will help you find where the bug might come from.
 
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -3,14 +3,21 @@ local ts = vim.treesitter
 
 local M = {}
 
-local function read_query_file(fname)
-  return table.concat(vim.fn.readfile(fname), '\n')
+local function read_query_files(filenames)
+  local contents = {}
+
+  for _,filename in ipairs(filenames) do
+    vim.list_extend(contents, vim.fn.readfile(filename))
+  end
+
+  return table.concat(contents, '\n')
 end
 
 function M.get_query(ft, query_name)
-  local query_files = api.nvim_get_runtime_file(string.format('queries/%s/%s.scm', ft, query_name), false)
+  local query_files = api.nvim_get_runtime_file(string.format('queries/%s/%s.scm', ft, query_name), true)
+
   if #query_files > 0 then
-    return ts.parse_query(ft, read_query_file(query_files[1]))
+    return ts.parse_query(ft, read_query_files(query_files))
   end
 end
 


steelsojka:
Note, this also still requires an existing query file in order to override, since features are enabled by the presence of a query file. So just adding a buffer variable with a query for a language that doesn't have a query file in this repo won't work.

kyazdani42:
hi.
I wonder if we should just provide additionnal query and highlight files by filetype instead of by buffer ? It would also allow users to locally create / extend queries without modifying the plugin repository directly. 
@vigoux @theHamsta what are you thoughts about this ?

steelsojka:
I like that idea. Where would the files live (structure-wise)?

kyazdani42:
exactly, we would not care where the files live if these are user defined. The map ft => path is the way i would expect it to be (as a user), so if you wish to do it, go for it, it was on the roadmap anyway so that feature will be gladly welcomed !

steelsojka:
@kyazdani42 awesome. I will implement those changes.

kyazdani42:
well i'd expect them to overwrite. If you can manage to make an option to either override or append, it would be even more awesome though, if it's too hard, just do the overwrite.
With options, you could do something like
```
{ 
  ft = {
   highlight = {
     path = '/some/file',
     action = 'overwrite',
  },
   query = {
    path = '/some/other/file',
   action = 'extend'
 }
}
```
As you wish :)

steelsojka:
@kyazdani42  What about adding it as part of the config instead of a global variable?

```
require "nvim-treesitter.configs".setup {
  highlight = {
    enable = true;
    queries = {
      lua = {
        path = "/some/filepath/lua.scm",
        action = 'extend'
      }
    }
  };
}
```

kyazdani42:
that would work better indeed, go for it !

steelsojka:
Since the query files are by query type and NOT by module, I put a top level `queries` key on the config.

vigoux:
Sounds great!
One thing to note though is that queries a searched as runtime files, thus they can be overwritten by adding a query in a path like `.config/nvim/queries/foo/locals.scm`.

steelsojka:
@vigoux So should this just be for extensions then since you can already override them by putting the file in the correct location?

vigoux:
@steelsojka yep, absolutely, actually, we may even change the behavior and have extending as the default behavior, as overwriting is only a special type of extension (like prepending the extension with `(_*) @Normal` for highlights)

vigoux:
I think that the modification becomes much simpler if we have extension as the only behavior, it is just a matter of concatenating the files in `query.lua`.
Haven't tried it though...

steelsojka:
@vigoux I got it working with just adding the runtime file and it works great. Much simpler solution.

kyazdani42:
this simple and i just works ? great job. I will test this tomorrow

steelsojka:
@kyazdani42 I tested with a pretty simple example. I'm not sure about a more "advanced" usage, but I don't foresee any issues.

kyazdani42:
well i just tested trying to override a highlight property, and it does not work as expected.
I think the issue is that if one value has been declared once, it will not be overriden by another one.
To reproduce:
```
mkdir -p ~/tmp/queries/lua
echo "local @operator" > ~/tmp/queries/lua/highlights.scm
echo "set rtp+=~/tmp >> $NVIM_CONFIG"
```
open a lua file with treesitter highlight enabled, the `local` keyword should be @operator but is actually @keyword as declared in the main syntax.
Note that if you have gui options on @operator it will add it to the current highlight, so your highlight would be @keyword + gui from @operator.

kyazdani42:
if i iterate in normal order, the highlighting 'overrides' properly. Although it is not really expected behaviour, as rules are not properly overriding themselves but merging themselves.
I don't really know how this could be solved. I didn't test to override the `queries.scm` files, but behaviour should be the same as i think the `ts.parse_query` function call is merging the rules. @vigoux what do you think about this ?

steelsojka:
@kyazdani42  I think the call to `vim.get_runtime_files` returns the file paths in non predictable order which is why it was iterating in reverse. In my testing my override file would show up first in the list. IT must depend on. I'm not 100% familiar on how the queries merge exactly. I tested `(nil) @boolean` for lua highlights and `nil` was colored the same as boolean, so it "worked" in that sense.

kyazdani42:
i will take a look at the code to unreavel this. It shouldn't be too hard to fix imho.

kyazdani42:
i don't think it returns in non predictable order. It will return the file depending on where you set them on your `runtimepath` (basically like `$PATH` works for a shell).
It would then depend on the user. If you set the runtimepath of your query before the runtimepath of nvim-treesitter, going backward would work for highlights, but if you put it after, going forward would work.
I would expect users to set the runtimepath for their queries after nvim-treesitter runtimepath, so going forward would be the appropriate solution. Although this would only be the case for highlighting as rules are applied one after another.

steelsojka:
@kyazdani42 That makes sense. I put the file in `~/.vim/queries/lua/highlights.scm` which was sourced BEFORE my plugins, which is why I did it in reverse. I'll change the loop behavior.

PR #86, opened 2020-06-15
feat(lang): add json highlights

steelsojka:


queries/json/highlights.scm
```diff
@@ -0,0 +1,13 @@
+(true) @constant.builtin
+(false) @constant.builtin
```
kyazdani42, line 2: isn't there `@boolean` @vigoux  ?

steelsojka: Oops... don't know why I did constant...

steelsojka: @kyazdani42 Updated

steelsojka:
> This highlighting might have issue due to the overlapping matches (`label` and `string`).

Should be fixed now. `@string` is only for the value of a pair instead just a string node.

PR #89, opened 2020-06-19
refacto: remove buf state and api

kyazdani42:
after some discussion with @vigoux, we decided to remove the `buf_state` api that was too complicated and use a simple `get_node_at_cursor` function. Also the incremental selection should be easier to extend.


README.md
```diff
@@ -131,19 +131,13 @@ Some of these features are :
...
+local ts_utils = require 'nvim-treesitter.ts_utils'
 ```
-
-You can also retrieve the current state of the current buffer with:
-```lua
-local buf_state = require'nvim-treesitter'.get_buf_state()
-```
-
-More information is available in neovim documentation (`:help nvim-treesitter-api`).
+More information is available in neovim documentation (`:help nvim-treesitter-utils`).
```
vigoux, line 140: I am a little bothered with this line...
```suggestion
More information is available in the help file (`:help nvim-treesitter-utils`).
```

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -205,4 +206,10 @@ function M.previous_scope(node)
   end
 end
 
+function M.get_node_at_cursor()
+  local cursor = api.nvim_win_get_cursor(0)
+  local root = parsers.get_parser().tree:root()
+  return root:named_descendant_for_range(cursor[1]-1,cursor[2],cursor[1]-1,cursor[2])
+end
```
vigoux, line 213: Can we just change that to be :
```suggestion
function M.get_node_at_cursor(winnr)
  local cursor = api.nvim_win_get_cursor(winnr or 0)
  local root = parsers.get_parser().tree:root()
  return root:named_descendant_for_range(cursor[1]-1,cursor[2],cursor[1]-1,cursor[2])
end
```

kyazdani42:
indeed, but i don't think the incremental selection has changed in term of usability

vigoux:
Lets merge that then.
We should update the roadmap, and find new feature ideas for the plugin, like tiny little things that can be useful

PR #90, opened 2020-06-19
refacto/fix: filetype / parser name distinction

### LINKED ISSUE #84: parsers not recognized/working for tsx files and csharp files
a first easy way to fix this is to check for the filetype and change it when needed.

kyazdani42:
fixes #84.

- move parser list in `parsers.lua`. All the code related to the parser is now basically in parsers.lua
- most `ft` variables have been renamed to `lang`, and the switch between those is easier to do with two function:
   - `parsers.ft_to_lang(ft) -> lang`
   - `parsers.lang_to_ft(lang) -> ft`
- filetypes are now only used when binding autocmd to Filetype on start and on commands like `TSBufEnable` and such. Otherwise, we use the `lang` to match the parser name we want. In the parser list, if a parser name does not match its filetype, we define a key `filetype`.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -61,9 +61,9 @@ parserlist.ruby = {
   }
 }
 
-parserlist.bash = {
+parserlist.sh = {
   install_info = {
-    url = "https://github.com/tree-sitter/tree-sitter-bash",
+    url = "https://github.com/nvim-treesitter/tree-sitter-bash",
     files = { "src/parser.c", "src/scanner.cc" },
   }
 }
@@ -124,9 +124,9 @@ parserlist.swift = {
   }
 }
 
-parserlist.csharp = {
+parserlist.cs = {
   install_info = {
-    url = "https://github.com/tree-sitter/tree-sitter-c-sharp",
+    url = "https://github.com/nvim-treesitter/tree-sitter-c-sharp",
     files = { "src/parser.c", "src/scanner.c" },
   }
 }
@@ -139,9 +139,9 @@ parserlist.typescript = {
   }
 }
 
-parserlist.tsx = {
+parserlist.typescriptreact = {
   install_info = {
-    url = "https://github.com/tree-sitter/tree-sitter-typescript",
+    url = "https://github.com/nvim-treesitter/tree-sitter-typescript",
     files = { "src/parser.c", "src/scanner.c" },
     location = "tree-sitter-tsx/tsx"
   }
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -3,213 +3,6 @@ local api = vim.api
 local queries = require'nvim-treesitter.query'
 local parsers = require'nvim-treesitter.parsers'
 
-local parserlist = {}
-
-parserlist.javascript = {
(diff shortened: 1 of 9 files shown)


lua/nvim-treesitter/configs.lua
```diff
@@ -139,9 +139,9 @@ parserlist.typescript = {
   }
 }
 
-parserlist.tsx = {
+parserlist.typescriptreact = {
   install_info = {
```
theHamsta, line 143: Needs to be changed also in base language map.

kyazdani42: indeed, this is a very much WIP PR until i can make some changes into neovim's core !

kyazdani42: actually found that neovims core might just work quite fine. I will fix the issues in this PR instead

vigoux:
LGTM, it is good to have some consistency, I will review this more heavily later, but it looks good

lua/nvim-treesitter/parsers.lua
```diff
@@ -1,17 +1,255 @@
...
+        url = "https://github.com/tree-sitter/tree-sitter-regex",
+        files = { "src/parser.c" }
+    }
+}
+
+local M = {
+  list = list
+}
+
+local ft_to_parsername = {}
```
theHamsta, line 218: This dictionary should be made somehow accessible for configuration for users. In case, they want to use a parser from a mainstream language for niche-dialect.

kyazdani42: Maybe the users could setup their own parsers ? they don't need to access this particular variable if they setup new parsers directly in the `list`

kyazdani42: well this should not be available actually, because you cannot setup new parsers just editing this dict. I'm going to add some code to make this possible though.

kyazdani42:
I think adding this should be in another PR as it will involve a little bit of code and thinking, i don't want to pollute this PR with other features :)

lua/nvim-treesitter/configs.lua
```diff
@@ -242,81 +35,89 @@ local config = {
...
+  if not lang then
+    return
+  end
+
+  if not parsers.list[lang] or not config.modules[mod] then
     return
   end
 
   local loaded_mod = require(string.format("nvim-treesitter.%s", mod))
   loaded_mod.detach(bufnr, ft)
```
theHamsta, line 98: @kyazdani42 there are still a few occurrences of  'ft' left! ft is undefined here.

theHamsta: Also in `M.is_enabled`

theHamsta: Also `locals.lua` needs to be updated (ft -> lang). Just seeing this now when playing a bit with nvim-treesitter.

kyazdani42: woops sorry guys i thought you reviewed the code before that !
I will push an update thanks for notifying

theHamsta: Reviewing was looking at the code and test some basic functionality. Disabling is not triggered that often. But the linters see it. With proper CI this will not happen in future.

PR #91, opened 2020-06-19
Refactor: utilize new consolidated syntax

TravonteD:


queries/css/highlights.scm
```diff
@@ -1,72 +1,88 @@
...
+ (attribute_name)
+ ] @property
 
 
 ((property_name) @type
- (#match? @type "^--"))
+                 (#match? @type "^--"))
 ((plain_value) @type
- (#match? @type "^--"))
+               (#match? @type "^--"))
```
kyazdani42, line 61: are those tabs required ? i dont mind but it looks weird on github

TravonteD: Good catch, it was just vim trying to be smart with the indentation :smile:

queries/ruby/highlights.scm
```diff
@@ -39,32 +48,45 @@
...
+   ])
+
 (method_call
-  method: (constant) @function)
+  [
+   receiver: (constant) @constant
+   method: [
+            (identifier) @function
+            (constant) @function
+            ]
```
vigoux, line 65: Isn't this refectorable even further in this ?
```suggestion
   method: [
            (identifier)
            (constant)
            ] @function
```

TravonteD: I did a quick test and yes that works! I've updated accordingly

vigoux:
That's all good for me, your queries are really an example though ! Can you add that to the `CONTRIBUTING.md` ? Something saying that the reference in term of complicated but accurate queries are the one you maintain

TravonteD:
Added a line to let new contributers know that they can use the ruby queries as an example.

queries/css/highlights.scm
```diff
@@ -1,72 +1,88 @@
...
-"{" @punctuation.bracket
-")" @punctuation.bracket
-"(" @punctuation.bracket
-"}" @punctuation.bracket
+[
+ "#"
+ ","
+ "."
+ ":"
+ "::"
```
vigoux, line 80: Just noticed that you have lost the `@punctuation.delimiter` here

TravonteD: Updated with a fix

PR #92, opened 2020-06-19
fix(install): remove .gitignore from parser directory

steelsojka:
I kept getting this error when opening a buffer without a filetype (terminal buffers).

Screen Shot 2020-06-19 at 10 59 50 AM

Seems like it's trying to read the .gitignore file as a parser. This change removes the file from that directory and moves it the root of the project. I also had to create the parser directory on install since git won't track in anymore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+parser/*.so
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -30,6 +30,12 @@ local function run_install(cache_folder, package_path, ft, repo)
   local parser_lib_name = package_path.."/parser/"..ft..".so"
   local command_list = {
     {
+      cmd = 'mkdir',
+      opts = {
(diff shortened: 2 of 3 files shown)


kyazdani42:
very strange, but maybe there is a bug somewhere else then ? It shouldn't try to read something else than libraries. Could you try to fix your issue in a different way ? If you can't we'll merge this but i'm not too happy to add code that shouldn't be necessary

steelsojka:
@kyazdani42 I checked the code in completion-treesitter and it seemed like this was the best place for the fix. It's trying to get the parser tied to the filetype but for terminal buffers or buffers that didn't have a filetype, it would try and read the directory for parsers. I think the change would need to go into neovim core since it's the call to `vim.treesitter.get_parser`. I'll investigate it further and report back.

steelsojka:
@kyazdani42 Ended up finding it. You were spot on. Here's the PR in completion-treesitter https://github.com/nvim-treesitter/completion-treesitter/pull/8

PR #93, opened 2020-06-20
Capture parameters in declaration

PitcherTear22:
Idk if it is needed in invocation, @vigoux please comment


vigoux:
It seems right to me as they are declarations.
You can look at @TravonteD open PR as an example on how to refactor queries.

PitcherTear22:
Idk about this https://github.com/PitcherTear22/nvim-treesitter/blob/a188d4ff91a324fe3cd80914ba731b5430fee757/queries/java/highlights.scm#L74

PitcherTear22:
Idk what to do with this:
```scheme

((field_access
  object: (identifier) @type)
 (#match? @type "^[A-Z]"))
((scoped_identifier
  scope: (identifier) @type)
 (#match? @type "^[A-Z]"))
```

vigoux:
I don't really know, leave this as is for now I think.

PitcherTear22:
Merge this then?

vigoux:
I would like to review this more in depth, I will merge this after that 😉

queries/java/highlights.scm
```diff
@@ -2,20 +2,33 @@
...
-  name: (identifier) @method)
+ name: (identifier))
+] @method
+
+[
+(method_declaration
+ parameters: (formal_parameter))
+(method_declaration
+ parameters: (formal_parameter))
+] @parameter
```
theHamsta, line 17: Could you check whether you're also capturing lambda parameters? Probably, this requires an additional query.

theHamsta: Also check the parameters of constructors:

image

theHamsta: I guess: this should be 

```
(method_declaration
   parameters: (formal_parameter) @parameter)
```
Or it will capture the whole method as `@parameter`.

image

PitcherTear22: > Could you check whether you're also capturing lambda parameters? Probably, this requires an additional query.

Idk what are lamda parameters

PitcherTear22: > I guess: this should be
> 
> ```
> (method_declaration
>    parameters: (formal_parameter) @parameter)
> ```
> 
> Or it will capture the whole method as `@parameter`.
> 
> image

Isn't it like this already?

vigoux: > Idk what are lamda parameters

I think he meant lambda function (or closure) parameters

theHamsta: > > I guess: this should be
> > ```
> > (method_declaration
> >    parameters: (formal_parameter) @parameter)
> > ```
> > 
> > 
> > Or it will capture the whole method as `@parameter`.
> > image
> 
> Isn't it like this already?

No, your query doesn't match anything. I guess you want something like:

```

(method_declaration
 parameters: 
 	(formal_parameters
 		(formal_parameter
        	name: (identifier) @parameter))) 
```
Or simpler:
```
(formal_parameter
        	name: (identifier) @parameter))) 
```
Or:
```
(formal_parameter) @parameter 
```

theHamsta: You can test your queries here: https://tree-sitter.github.io/tree-sitter/playground

vigoux: > ```
> (formal_parameter
>         	name: (identifier) @parameter))) 
> ```

I suggest using this one, as you capture the actual thing you want to capture

queries/java/highlights.scm
```diff
@@ -2,20 +2,29 @@
...
 ; Methods
 
+[
 (method_declaration
-  name: (identifier) @method)
+ name: (identifier))
 (method_invocation
-  name: (identifier) @method)
+ name: (identifier))
+] @method
```
vigoux, line 10: With this kind of this you are going to capture to whole `method_delcaration` group, which for sure is not right, you will have to keep it capturing the `(identifier)` node alone.

queries/java/highlights.scm
```diff
@@ -2,20 +2,29 @@
...
 ; Annotations
 
+[
 (annotation
-  name: (identifier) @attribute)
+ name: (identifier))
 (marker_annotation
-  name: (identifier) @attribute)
+ name: (identifier))
+] @attribute
```
vigoux, line 24: Same here

queries/java/highlights.scm
```diff
@@ -45,12 +54,18 @@
...
+ name: (identifier))
 (class_declaration
-  name: (identifier) @type)
+ name: (identifier))
 (enum_declaration
-  name: (identifier) @type)
+ name: (identifier)
+(constructor_declaration
+ name: (identifier))
+(type_identifier)
```
vigoux, line 66: Same here

queries/java/highlights.scm
PitcherTear22: > you will have to keep it capturing the (identifier) node alone

That would capture all identifiers instead of method identefiers only, no?

queries/java/highlights.scm
```diff
@@ -54,17 +54,16 @@
...
 
-[
 (interface_declaration
- name: (identifier))
+ name: (identifier) @type)
 (class_declaration
- name: (identifier))
+ name: (identifier) @type)
 (enum_declaration
  name: (identifier)
```
vigoux: There is no closing paren here ?

queries/java/highlights.scm
```diff
@@ -54,17 +54,16 @@
...
 (constructor_declaration
- name: (identifier))
-(type_identifier)
-] @type
+ name: (identifier) @type)
+(type_identifier @type)
```
vigoux, line 65: That seem wrong to
```suggestion
(type_identifier) @type
```

queries/java/highlights.scm
```diff
@@ -46,11 +55,16 @@
...
 
 (interface_declaration
-  name: (identifier) @type)
+ name: (identifier) @type)
 (class_declaration
-  name: (identifier) @type)
+ name: (identifier) @type)
 (enum_declaration
-  name: (identifier) @type)
+ name: (identifier)
```
vigoux, line 62: Missing paren here
```suggestion
 name: (identifier) @type)
```

queries/java/highlights.scm
```diff
@@ -150,25 +167,30 @@
 "case"
 ] @conditional
 
-; 
+;
```
vigoux, line 170: Why keeping this ?

PitcherTear22: To separate sections, I don't know what to name that section

theHamsta:
I tested the hightlighting and all queries look good to me.

You could add a query to capture lambda parameters:
image

So 
```scheme
(inferred_parameters (identifier) @parameter)
```

queries/java/highlights.scm
```diff
@@ -59,32 +73,35 @@
...
+] @type.builtin
 
 ; Variables
 
 ((identifier) @constant
  (#match? @constant "^_*[A-Z][A-Z\d_]+"))
 
 (identifier) @variable
-
 (this) @variable.builtin
```
theHamsta, line 90: @vigoux we have no hightlight for this yet. In python, we use `@constant.builtin`

queries/java/highlights.scm
```diff
@@ -59,32 +73,35 @@
...
+(floating_point_type)
+(void_type)
+] @type.builtin
 
 ; Variables
 
 ((identifier) @constant
  (#match? @constant "^_*[A-Z][A-Z\d_]+"))
 
 (identifier) @variable
```
theHamsta, line 89: This is not highlighted.

theHamsta:
Also `->` is missing in the operators.

queries/java/highlights.scm
vigoux: No`@variable.bultin`? I thought we had... We will need to fix that later...

queries/java/highlights.scm
vigoux: Non we don't this highlighting is not useful, could you remove it?

PitcherTear22: ok

vigoux: No modification here ?

queries/java/highlights.scm
vigoux: This needs to be at least commented out with a `TODO`...

queries/java/highlights.scm
```diff
@@ -50,21 +50,21 @@
...
- name: (identifier))
+ name: (identifier) @type)
 (enum_declaration
- name: (identifier)
+ name: (identifier) @type)
 (constructor_declaration
- name: (identifier))
-(type_identifier)
-] @type
+ name: (identifier) @type)
```
TravonteD, line 65:
```suggestion
   name: (identifier) @type)
```

PitcherTear22: The original file used 1 space, I have used 1 space everywhere. Should I change it everywhere?

PitcherTear22: Just realised you have an editorconfig, done

queries/java/highlights.scm
```diff
@@ -150,25 +168,30 @@
...
 "for"
 "while"
 ] @repeat
 
-;
+; Includes
 
 "import" @include
 
 ; Punctuation
```
theHamsta, line 182: We could also add "package" to includes.

queries/java/highlights.scm
```diff
@@ -150,25 +168,30 @@
 "case"
 ] @conditional
 
-; 
+;
 
 [
 "for"
 "while"
 ] @repeat
```
theHamsta, line 176: Is there a do...while loop in Java? So do would be also repeat.

queries/java/highlights.scm
theHamsta: Don't want to block this PR. Just ideas for more highlights...

PitcherTear22: > Don't want to block this PR. Just ideas for more highlights...

yeah no worry

queries/java/highlights.scm
PitcherTear22: idk it didn't save properly

queries/java/highlights.scm
PitcherTear22: done (TODO)

queries/java/highlights.scm
PitcherTear22: i deleted those

queries/java/highlights.scm
PitcherTear22: done

queries/java/highlights.scm
PitcherTear22: done

queries/java/highlights.scm
PitcherTear22: It isn't there, i think gh messed up. Please see https://raw.githubusercontent.com/PitcherTear22/nvim-treesitter/master/queries/java/highlights.scm

PitcherTear22: Find says phrase not found https://imgur.com/O9N6DIM

theHamsta: I was referring to `@variable`

PitcherTear22: `identifier` or `this`?
`this` has been commented out as TODO

theHamsta: Just delete `(identifier) @variable`. We have no highlight for `@variable`. This is what @vigoux was trying to say.

queries/java/highlights.scm
```diff
@@ -51,40 +72,47 @@
...
-(boolean_type) @type.builtin
-(integral_type) @type.builtin
-(floating_point_type) @type.builtin
-(floating_point_type) @type.builtin
-(void_type) @type.builtin
+[
+(boolean_type)
+(integral_type)
+(floating_point_type)
+(floating_point_type)
```
theHamsta, line 92: Floating_point_type is doubled in this query.

PitcherTear22: done

PR #95, opened 2020-06-21
Adapt Python highlights to new query syntax

theHamsta:
#76

Maybe we should add `@defintion.class`. It's now also in the `tags.scm` of many treesitter repos.


vigoux:
> Maybe we should add `@defintion.class`. It's now also in the `tags.scm` of many treesitter repos.

If we don't already have one, we should definitely.

PR #97, opened 2020-06-21
Deactivate highlighting of @error

theHamsta:
This can be annoying while typing  #78  while in some cases this might be
useful to detect syntax errors. We could change the default to no
highlighting of errors. We can keep it in the code to highlight that
users can still opt-in to an highlighting of `@error`.

I think languages should still include the `@error` query so the feature to highlight error can still be activated.

### DIFF
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -10,7 +10,7 @@ local M = {
 local hlmap = vim.treesitter.TSHighlighter.hl_map
 
 -- Misc
-hlmap.error = "Error"
+hlmap.error = nil -- = "Error" activates highlighting of syntax errors
 hlmap["punctuation.delimiter"] = "Delimiter"
 hlmap["punctuation.bracket"] = "Delimiter"
 hlmap["punctuation.special"] = "Delimiter"


lua/nvim-treesitter/highlight.lua
```diff
@@ -10,7 +10,7 @@ local M = {
 local hlmap = vim.treesitter.TSHighlighter.hl_map
 
 -- Misc
-hlmap.error = "Error"
+hlmap.error = nil
```
theHamsta, line 13: Equivalent to deleting this line, but it still has some documenting character.

lua/nvim-treesitter/highlight.lua
```diff
@@ -10,7 +10,7 @@ local M = {
 local hlmap = vim.treesitter.TSHighlighter.hl_map
 
 -- Misc
-hlmap.error = "Error"
+hlmap.error = nil -- = "Error" activates highlighting of syntax errors
```
theHamsta, line 13: Equivalent to deleting this line. Keep it for its self-documenting character.

vigoux:
I am not that pleased with that... I actually though about linking this to a new highlight group (see #99 ) and to include it by default in all queries (by concatenation)

theHamsta:
Superseded by #99.  But I still think highlighting of syntax errors should be off by default.

PR #98, opened 2020-06-21
WIP: Add disclaimer to README.md

theHamsta:
Should we add some kind of disclaimer to the README to warn users that development of TS highlighting in nvim is not finished?

I am unsure about the text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,9 @@
 # nvim-treesitter
 Treesitter configurations and abstraction layer for Neovim.
 
+**Warning: Treesitter and Treesitter highlighting are an experimental feature of nightly versions of Neovim.
+Please consider the experience with this plug-in as experimental until Neovim 0.5 is released!**
+
 # Quickstart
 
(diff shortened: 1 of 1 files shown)


theHamsta:
Shit: CSS is not rendered by Github.

theHamsta:
Discovered a bug in the scanner of the Markdown parser. If it causes a SEGFAULT, nvim will crash :cry:

TravonteD:
> Shit: CSS is not rendered by Github.

We could just make it large and bold. A header would probably fine?

vigoux:
> Discovered a bug in the scanner of the Markdown parser. If it causes a SEGFAULT, nvim will crash cry

Well.. That is quite a problem... We should report that upstream... Did you try using the `tree-sitter` cli ?

theHamsta:
If I can reproduce it, I can report in on `tree-sitter-markdown`. Since it's a custom `scanner.cc` all kind of bad things can happen.

PR #99, opened 2020-06-21
Use custom highlight groups

vigoux:
Pretty little one to use custom highlight groups, this will allow more customisability on users side.


lua/nvim-treesitter/highlight.lua
```diff
@@ -12,44 +12,44 @@ local hlmap = vim.treesitter.TSHighlighter.hl_map
 
 -- Misc
 hlmap.error = "Error"
```
theHamsta, line 14: And what about linking `Error`?

vigoux: Oh I forgot that actually

theHamsta:
I am ok with that solution. As a user, I would always prefer manipulating the dictionary instead of changing the highlight groups.
The dictionary makes it a lot easier to add own queries (`hl_map[fantasy] = 'CrazyAnimatedBoldLetters'`).

vigoux:
But this requieres going deep into the plugin, instead, using highlight groups only requires some lines in your `init.vim`, but in any case, we should have that anyway

theHamsta:
But the dictionary is not from the plugin but directly from the Neovim API.

vigoux:
Actually, we are cheating, this dictionnary is not meant to be exposed, we are more exploiting it than anything

theHamsta:
I have seen this dict in configs that are not using our plugin.

theHamsta:
Maybe you could document the changes? Also how to deactivate the `Error` highlight. Does not have to be in the same PR.

vigoux:
Yeah I will add that, and document every highlight group

doc/nvim-treesitter.txt
```diff
@@ -166,4 +166,135 @@ Note: This is highly experimental, and folding can break on some types of
...
+							    *hl-TSConditional*
+For keywords related to conditionnals.
+
+`TSRepeat`
+								 *hl-TSRepeat*
+For keywords related to loops.
+
+`TSLabel`
+								  *hl-TSLabel*
+For labels: `labe:` in C and `:label:` in Lua.
```
theHamsta, line 269: typo!

doc/nvim-treesitter.txt
```diff
@@ -166,4 +166,135 @@ Note: This is highly experimental, and folding can break on some types of
       edits. If you encounter such breakage, hiting `zx` should fix folding.
       In any case, feel free to open an issue with the reproducing steps.
 
+==============================================================================
+HIGHLIGHTS                                        *nvim-treesitter-highlights*
+
+TSError
```
theHamsta, line 172: TSError: .. for syntax/parsing errors

Maybe you can add an introduction to this section. Especially, with an example how you could overwrite a highlighting. E.g. how to avoid highlighting of syntax errors.

doc/nvim-treesitter.txt
```diff
@@ -166,4 +166,135 @@ Note: This is highly experimental, and folding can break on some types of
...
+`TSStructure`
+							      *hl-TSStructure*
+This is left as an exercise for the reader.
+
+`TSInclude`
+								*hl-TSInclude*
+For includes: `#include` in C, `use` or `extern crate` in Rust, or `require`
+in Lua
+
  vim:tw=78:ts=8:noet:ft=help:norl:
```
theHamsta, line 300: Thanks for adding the documentation :+1:

PR #100, opened 2020-06-21
User textobj

theHamsta:
I thought it would be cool if users could define their own text object by queries.

Everything in this PR is not implemented the proper way. The module is just activated for easier review and faster testing.

With the given hard-coded config you can delete a function text object in Python/C++. Maybe the queries should also come from `locals.scm` but this is a more direct way for a user to achieve that for a specific query.

PS: Why was the operator pending mapping removed from `incremental_selection`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,34 @@ require'nvim-treesitter.configs'.setup {
         }
       }
     },
+    textobjects = { -- syntax-aware textobjects
+	enable = true,
+	disable = {},
+	keymaps = {
+	    ["iL"] = { -- you can define your own textobjects directly here
+		python = "(function_definition) @function",
+		cpp = "(function_definition) @function",
+		c = "(function_definition) @function",
+		java = "(method_declaration) @function"
+	    },
+	    -- or you use the queries from supported languages with textobjects.scm
+	    ["af"] = "@function.outer",
+	    ["if"] = "@function.inner",
+	    ["aC"] = "@class.outer",
+	    ["iC"] = "@class.inner",
+	    ["ac"] = "@conditional.outer",
+	    ["ic"] = "@conditional.inner",
+	    ["ae"] = "@block.outer",
+	    ["ie"] = "@block.inner",
+	    ["al"] = "@loop.outer",
+	    ["il"] = "@loop.inner",
+	    ["is"] = "@statement.inner",
+	    ["as"] = "@statement.outer",
+	    ["ad"] = "@comment.outer",
+	    ["am"] = "@call.outer",
+	    ["im"] = "@call.inner"
+	}
+    },
     ensure_installed = 'all' -- one of 'all', 'language', or a list of languages
 }
 EOF
@@ -171,6 +199,7 @@ The roadmap and all features of this plugin are open to change, and any suggesti
 - `refactor.navigation`: Syntax based definition listing and navigation.
   * List all definitions
   * Go to definition
+- `textobjects`: Vim textobjects defined by treesitter queries
 
 ## Defining Modules
 
@@ -217,7 +246,7 @@ More information is available in the help file (`:help nvim-treesitter-utils`).
 
 ## Supported Languages
 
-For treesitter to work, we need to use query files such as those you can find in
+For `nvim-treesitter` to work, we need to use query files such as those you can find in
 `queries/{lang}/{locals,highlights,textobjects}.scm`
 
 We are looking for maintainers to write query files for their languages.
--- a/README.md
+++ b/README.md
@@ -217,7 +217,8 @@ More information is available in the help file (`:help nvim-treesitter-utils`).
 
 ## Supported Languages
 
-For treesitter to work, we need to use query files such as those you can find in `queries/{lang}/{locals,highlights}.scm`
+For treesitter to work, we need to use query files such as those you can find in
+`queries/{lang}/{locals,highlights,textobjects}.scm`
 
 We are looking for maintainers to write query files for their languages.
 
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -4,6 +4,19 @@ local queries = require'nvim-treesitter.query'
 local parsers = require'nvim-treesitter.parsers'
 local utils = require'nvim-treesitter.utils'
 
+local M = {}
+
+local function has_some_textobject_mapping(lang)
+  for _, v in pairs(M.get_module('textobjects').keymaps) do
+    if type(v) == 'table' then
+      if v[lang] then
+        return true
+      end
+    end
+  end
+  return false
+end
+
 local config = {
   modules = {},
   ensure_installed = nil
@@ -65,12 +78,21 @@ local builtin_modules = {
         list_definitions = "gnD"
       }
     }
+  },
+  textobjects = {
+    module_path = 'nvim-treesitter.textobjects',
+    enable = false,
+    disable = {},
+    is_supported = function(lang)
+      return has_some_textobject_mapping(lang) or queries.has_textobjects(lang)
+    end,
+    keymaps = {
+      inverse_mappings = true
+    }
   }
 }
 
 local special_config_keys = {'enable', 'disable', 'keymaps'}
-
-local M = {}
 
 -- Resolves a module by requiring the `module_path` or using the module definition.
 local function resolve_module(mod_name)
@@ -279,12 +301,17 @@ function M.setup_module(mod, data, mod_name)
       mod.disable = data.disable
     end
     if mod.keymaps and type(data.keymaps) == 'table' then
-      for f, map in pairs(data.keymaps) do
-        if mod.keymaps[f] then
-          mod.keymaps[f] = map
+      if mod.keymaps.inverse_mappings then
+        mod.keymaps = data.keymaps
+      else
+        for f, map in pairs(data.keymaps) do
+          if mod.keymaps[f] then
+            mod.keymaps[f] = map
+          end
         end
       end
     end
+
     for k, v in pairs(data) do
       -- Just copy all non-special configuration keys
       if not vim.tbl_contains(special_config_keys, k) then
--- a/lua/nvim-treesitter/incremental_selection.lua
+++ b/lua/nvim-treesitter/incremental_selection.lua
@@ -8,32 +8,32 @@ local M = {}
 
 local selections = {}
 
-local function update_selection(buf, node)
-  local start_row, start_col, end_row, end_col = node:range()
-
-  if end_row == vim.fn.line('$') then
-    end_col = #vim.fn.getline('$')
-  end
-
-  vim.fn.setpos(".", { buf, start_row+1, start_col+1, 0 })
-  vim.fn.nvim_exec("normal v", false)
-  vim.fn.setpos(".", { buf, end_row+1, end_col+1, 0 })
-end
-
 function M.init_selection()
   local buf = api.nvim_get_current_buf()
   local node = ts_utils.get_node_at_cursor()
   selections[buf] = { [1] = node }
-  update_selection(buf, node)
+  ts_utils.update_selection(buf, node)
+end
+
+-- moves 0-based node position by one character
+local function inclusive_pos_to_exclusive(row, col)
+  local line = vim.fn.getline(row + 1)
+
+  -- move by one character changes row?
+  if #line == col + 1  then
+    return row + 1, 0
+  else
+    return row, col + 1
+  end
 end
 
 local function visual_selection_range()
   local _, csrow, cscol, _ = unpack(vim.fn.getpos("'<"))
   local _, cerow, cecol, _ = unpack(vim.fn.getpos("'>"))
-  if csrow < cerow then
-    return csrow-1, cscol-1, cerow-1, cecol-1
+  if csrow < cerow or (csrow == cerow  and cscol <= cecol) then
+    return csrow-1, cscol-1, inclusive_pos_to_exclusive(cerow-1, cecol-1)
   else
-    return cerow-1, cecol-1, csrow-1, cscol-1
+    return cerow-1, cecol-1, inclusive_pos_to_exclusive(csrow-1, cscol-1)
   end
 end
 
@@ -53,7 +53,7 @@ local function select_incremental(get_parent)
       local csrow, cscol, cerow, cecol = visual_selection_range()
       local root = parsers.get_parser().tree:root()
       local node = root:named_descendant_for_range(csrow, cscol, cerow, cecol)
-      update_selection(buf, node)
+      ts_utils.update_selection(buf, node)
       selections[buf] = { [1] = node }
       return
     end
@@ -65,7 +65,7 @@ local function select_incremental(get_parent)
       table.insert(nodes, node)
     end
 
-    update_selection(buf, node)
+    ts_utils.update_selection(buf, node)
   end
 end
 
@@ -84,7 +84,7 @@ function M.node_decremental()
 
   table.remove(selections[buf])
   local node = nodes[#nodes]
-  update_selection(buf, node)
+  ts_utils.update_selection(buf, node)
 end
 
 function M.attach(bufnr)
--- a/lua/nvim-treesitter/textobjects.lua
+++ b/lua/nvim-treesitter/textobjects.lua
@@ -1,0 +1,96 @@
+local api = vim.api
+local ts = vim.treesitter
+
+local configs = require "nvim-treesitter.configs"
+local parsers = require "nvim-treesitter.parsers"
+local queries = require'nvim-treesitter.query'
+local locals = require'nvim-treesitter.locals'
+local ts_utils = require'nvim-treesitter.ts_utils'
+
+local M = {}
+
+function M.select_textobject(query_string)
+  local bufnr = vim.api.nvim_get_current_buf()
+  local ft = api.nvim_buf_get_option(bufnr, "ft")
+  if not ft then return end
+  local lang = parsers.ft_to_lang(ft)
+
+  local row, col = unpack(vim.api.nvim_win_get_cursor(0))
+  row = row - 1
+
+  local matches = {}
+
+  if string.match(query_string, '^@.*') then
+    matches = locals.get_capture_matches(bufnr, query_string, 'textobjects')
+  else
+    local parser = parsers.get_parser(bufnr, lang)
+    local root = parser:parse():root()
+    local start_row, _, end_row, _ = root:range()
+
+    local nested = {}
+    local query = ts.parse_query(lang, query_string)
+    for m in queries.iter_prepared_matches(query, root, bufnr, start_row, end_row) do
+      for _, n in pairs(m) do
+        if n.node then
+          table.insert(matches, n.node)
+        end
+      end
+    end
+  end
+
+  local match_length
+  local smallest_range
+
+  for _, m in pairs(matches) do
+    if ts_utils.is_in_node_range(m, row, col) then
+      local length = ts_utils.node_length(m)
+      if not match_length or length < match_length then
+        smallest_range = m
+        match_length = length
+      end
+    end
+  end
+
+  if smallest_range then
+    ts_utils.update_selection(bufnr, smallest_range)
+  end
+end
+
+function M.attach(bufnr, lang)
+  local buf = bufnr or api.nvim_get_current_buf()
+  local config = configs.get_module("textobjects")
+  local lang = lang or parsers.ft_to_lang(api.nvim_buf_get_option(bufnr, "ft"))
+
+  for mapping, query in pairs(config.keymaps) do
+    if type(query) == 'table' then
+      query = query[lang]
+    elseif not queries.get_query(lang, 'textobjects') then
+      query = nil
+    end
+    if query then
+      local cmd = ":lua require'nvim-treesitter.textobjects'.select_textobject('"..query.."')<CR>"
+      api.nvim_buf_set_keymap(buf, "o", mapping, cmd, {silent = true})
+      api.nvim_buf_set_keymap(buf, "v", mapping, cmd, {silent = true})
+    end
+  end
+end
+
+function M.detach(bufnr)
+  local buf = bufnr or api.nvim_get_current_buf()
+  local config = configs.get_module("textobjects")
+  local lang = parsers.ft_to_lang(api.nvim_buf_get_option(bufnr, "ft"))
+
+  for mapping, query in pairs(config.keymaps) do
+    if type(query) == 'table' then
+      query = query[lang]
+    elseif not queries.get_query(lang, 'textobjects') then
+      query = nil
+    end
+    if query then
+      api.nvim_buf_del_keymap(buf, "o", mapping)
+      api.nvim_buf_del_keymap(buf, "v", mapping)
+    end
+  end
+end
+
+return M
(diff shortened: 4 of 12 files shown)


theHamsta:
I thought direct queries are easier if the user wants to get a specific thing in their source code.

Probably we should also allow this:

`['if'] = '@defintion.function'`

theHamsta:
I guess `update_selection` should be
```lua
local function update_selection(buf, node)
  local start_row, start_col, end_row, end_col = node:range()

  if end_row == vim.fn.line('$') then
    end_col = #vim.fn.getline('$')
  end

  vim.fn.setpos(".", { buf, start_row + 1, start_col, 0 })
  vim.fn.nvim_exec("normal v", false)
  vim.fn.setpos(".", { buf, end_row + 1, end_col, 0 })
end
```
instead of 
```lua
local function update_selection(buf, node)
  local start_row, start_col, end_row, end_col = node:range()

  if end_row == vim.fn.line('$') then
    end_col = #vim.fn.getline('$')
  end

  vim.fn.setpos(".", { buf, start_row + 1, start_col + 1, 0 })
  vim.fn.nvim_exec("normal v", false)
  vim.fn.setpos(".", { buf, end_row + 1, end_col + 1, 0 })
end
```

col is 0 based.

kyazdani42:
i removed it back when i refactored the incremental selection a while ago. It was on the roadmap to add this back.
I think update selection works this way so the incremental selection can grow (i think there was an issue otherwise, but it might not be related to that change you suggest, im not sure)

theHamsta:
Maybe this +1 was related to the issue? Just maybe keep that in mind the next time you're working on it.

theHamsta:
You now can do things like 

```lua
      keymaps = {
        ['il'] = { -- put your queries right in the config
          python = '(function_definition) @function',
          cpp = '(function_definition) @function',
          c = '(function_definition) @function'
        },
        ['af'] = '@function-obj.outer', -- refer to queries in locals.scm
        ['if'] = '@function-obj.inner',
        ['aC'] = '@scope'
    }
```

Especially for "inner" text-object some whitespace control would be necessary to get the right vim feeling.

vigoux:
I am in favor of using a different query for this feature, so that it can be tuned easily if one wants to. Actually I think that 'early every feature should hav an associated query that defines its behavior. It allows easy customization, because these can be overwritten, and keeps things clean and separated.

theHamsta:
Ok. Then, I niceify this PR as soon as I have time with its own `textobj.scm`. With this new context the query names could just be `@function.inner`/`@function.outer`, `@class.inner`/`@class.outer`, `@block.inner`/`@block.outer`, `@conditional.inner`/`@conditional.outer`. `@comment.inner`/`@comment.outer`.

vigoux:
We can evan add `@loop` too

lua/nvim-treesitter/configs.lua
```diff
@@ -28,6 +28,21 @@ local config = {
       is_supported = function(lang)
         return queries.get_query(lang, 'locals')
       end
+    },
+    textobj = {
+      enable = true,
+      disable = {},
+      is_supported = function(_) return true end,
```
kyazdani42, line 35: maybe this should check the the query for textObj exists (either the locals.scm or the textObj.scm) ?

lua/nvim-treesitter/textobj.lua
```diff
@@ -0,0 +1,158 @@
...
+  local ft = api.nvim_buf_get_option(bufnr, "ft")
+  if not ft then return end
+  local lang = parsers.ft_to_lang(ft)
+
+  local row, col = unpack(vim.api.nvim_win_get_cursor(0))
+  row = row - 1
+
+  local matches = {}
+
+
```
kyazdani42, line 67: empty line. Also move the functions you need to ts_utils :)

lua/nvim-treesitter/textobj.lua
```diff
@@ -0,0 +1,158 @@
...
+    update_selection(bufnr, smallest_range)
+  end
+end
+
+function M.attach(bufnr)
+    local buf = bufnr or api.nvim_get_current_buf()
+    local config = configs.get_module("textobj")
+
+    local lang = parsers.ft_to_lang(api.nvim_buf_get_option(bufnr, "ft"))
+    for mapping, queries in pairs(config.keymaps) do
```
kyazdani42, line 124: maybe just check `config.keymaps` is a table, if its not return, so then you dont have to check if the query exists because you'd know you would have something in your keymap :) You could do the same for the detach function.
This is just a suggestion, i like to have less control flow in loops as possible as it makes the code easier to understand (to me at least).

theHamsta:
Ok, should now be more or less fine.

- instead of `locals.scm` I use `textobjects.scm`
- queries also work for languages that have no `textobjects.scm` as long as the user has set one mapping for this language (maybe I should remove this feature and only allow `textobjects.scm`)
- there's a problem for `@function.inner` with the c-queries (temporary solution is to match with the curly braces):
```scheme
(function_definition
  body:  (compound_statement
                          ("{" (_)* @function.inner "}"))?) @function.outer
```
The query works as expected using the treesitter playground but Neovim only matches nodes if `(_)*` is exactly one node.
Neovim not match a group of nodes.

Alternatives:
 - use some regex capturing/text processing to remove "{" "}" (I don't like that)
 - use the query without the star, then I have multiple matches and I have to glue them together. (Problem: when are to adjecent nodes indented to be fused and when should they be separate?).
 - allow groups of nodes for a capture (prefered solution since it resembles the behavior of the official tree sitter runtime)

Test config
```lua
            textobjects = {
                enable = true,
                disable = {},
                keymaps = {
                    ['il'] = {
                        python = '(function_definition) @function',
                        cpp = '(function_definition) @function',
                        c = '(function_definition) @function',
                        java = '(method_declaration) @function' -- has no textobjects.scm
                    },
                    ['af'] = '@function.outer', -- use textobjects.scm
                    ['if'] = '@function.inner',
                }
            },
```

lua/nvim-treesitter/configs.lua
```diff
@@ -192,7 +208,7 @@ function M.setup(user_data)
       end
       if config.modules[mod].keymaps and type(data.keymaps) == 'table' then
         for f, map in pairs(data.keymaps) do
-          if config.modules[mod].keymaps[f] then
+          if config.modules[mod].keymaps[f] or mod == 'textobjects' then
```
theHamsta, line 211: This module use the inverse of the keymap. Maybe we should remove the feature that users define textobjects directly in config.

lua/nvim-treesitter/locals.lua
```diff
@@ -89,4 +101,38 @@ function M.get_references(bufnr)
...
+--- Return all nodes in locals corresponding to a specific capture (like @scope, @reference)
+-- Works like M.get_references or M.get_scopes except you can choose the capture
+-- Can also be a nested capture like @definition.function to get all nodes defining a function
+function M.get_capture_matches(bufnr, capture_string, locals_kind)
+    if not string.sub(capture_string, 1,2) == '@' then
+      print('capture_string must start with "@"')
+      return
+    end
+
+    -- Build capture path: @definition.function -> ["definition", "function"]
```
theHamsta, line 114: Could also be ["definition", "function", "node"] but not when we allow node groups for a capture.

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -212,4 +212,52 @@ function M.get_node_at_cursor(winnr)
   return root:named_descendant_for_range(cursor[1]-1,cursor[2],cursor[1]-1,cursor[2])
 end
 
+
+-- Set visual selection to node
+function M.update_selection(buf, node)
```
theHamsta, line 217: Please check whether this is correct and also works with `incremental  selection`

vigoux:
Oh that's a worrying issue... I will do some testing... Maybe what is going on is that treesitter tries to match *the same node* when you specify that `(_)` I don't know if that's an expected behavior...

lua/nvim-treesitter/incremental_selection.lua
```diff
@@ -53,19 +41,19 @@ local function select_incremental(get_parent)
...
       local root = parsers.get_parser().tree:root()
       local node = root:named_descendant_for_range(csrow, cscol, cerow, cecol)
-      update_selection(buf, node)
+      ts_utils.update_selection(buf, node)
       selections[buf] = { [1] = node }
       return
     end
 
-    node = get_parent(nodes[#nodes])
+    local node = get_parent(nodes[#nodes])
```
theHamsta, line 49: This was a global before

lua/nvim-treesitter/locals.lua
```diff
@@ -2,20 +2,26 @@
 -- Locals are a generalization of definition and scopes
 -- its the way nvim-treesitter uses to "understand" the code
 local api = vim.api
-local ts = vim.treesitter
```
theHamsta, line 5: Unused.

lua/nvim-treesitter/textobjects.lua
```diff
@@ -0,0 +1,102 @@
...
+      for _, n in pairs(m) do
+        if n.node then
+          table.insert(matches, n.node)
+        end
+      end
+    end
+  end
+
+  local match_length
+  local smallest_range
```
theHamsta, line 46: A possible extension for a later PR could be to add an option to also search for nearest match if cursor is not within any  matched range (behavior of targets.vim)

theHamsta:
I have the  suspicion that also treesitter doesn't allow group of nodes for captures either but just returns multiple matches.

vigoux:
Did you tried that in the playground?

theHamsta:
Ok, what I told bullshit. The query will work as expected. I just have to glue together the individual matches when they are adjacent.
```
(function_definition
  body:  (compound_statement
                          ("{" (_)* @function.inner "}"))?) @function.outer
```

lua/nvim-treesitter/ts_utils.lua
theHamsta: It seems that `incremental_selection` has problems to increment some selections without the last exclusive character.

kyazdani42: I had this issue before that's why i added one columns to the selection each time. If you have some time, look into it, or i'll do it.

vigoux:
I'd rather use your second solution, as is only relates to the nodes and not the actual text they represent.
The thing is that it would require some upstream changes. Tnhe `strip!` predicate does not suit me that well. But if it is the only way to get this done, then we should add this. It will not be that easy though I think.
We can alos bring that issue upstream (treesitter) in my opinion, as it would be interesting to have their views on this.

theHamsta:
This is how `strip!`,`select-adjacent!` is expected to work. But I didn't find their implementation.

   https://github.com/tree-sitter/tree-sitter/blob/de53b82e2c48d6221af9881e50cc57d961de3365/cli/src/tests/tags_test.rs#L122-L166 

 I guess both operations wouldn't be difficult to implement on the C side, just change the start/end bytes of the nodes based on a regex or adjacent nodes.

lua/nvim-treesitter/incremental_selection.lua
```diff
@@ -8,32 +8,31 @@ local M = {}
...
+    return row, col + 1
+  end
 end
 
 local function visual_selection_range()
   local _, csrow, cscol, _ = unpack(vim.fn.getpos("'<"))
   local _, cerow, cecol, _ = unpack(vim.fn.getpos("'>"))
-  if csrow < cerow then
-    return csrow-1, cscol-1, cerow-1, cecol-1
+  if csrow <= cerow then
```
theHamsta, line 32: Only comparing row is not sufficient!

theHamsta:
I think the changes should be ready now for review. Since this is a PR affects two modules I would prefer to have some more days for testing. 

- A big blocker (?) for this PR could be that right at the moment textobjects can only be single nodes. Probably, users want to capture multiple nodes and be able to trim some characters for inner textobjects.
- I adapted `incremental_nodes` to still work with my changes. I assume that visual selection is always a inclusive range while the ranges needed for treesitter are exclusive. Everything seems to be working when the corresponding conversion is performed. I discovered a bug (?) that could swap the end and start of a range.
- I extended locals to work with any named query file name. So users, can also have their own query group after this PR (`fantasy.scm` queryable by `get_locals(bufnr, 'fantasy')`
- I added `textobject.scm` for C/C++ and Python. They are useful for testing but should maybe be their own PR.

lua/nvim-treesitter/locals.lua
```diff
@@ -2,20 +2,35 @@
...
+    return rawget(table, key)
+  end
 }
 
-function M.collect_locals(bufnr)
+local query_chache
+if not query_chache then
+  query_chache = {}
+  setmetatable(query_chache, default_dict)
+end
```
kyazdani42, line 23: i think you meant `cache` ?

lua/nvim-treesitter/locals.lua
```diff
@@ -2,20 +2,35 @@
...
+    if not exists then
+      table[key] = {}
+    end
+    return rawget(table, key)
+  end
 }
 
-function M.collect_locals(bufnr)
+local query_chache
+if not query_chache then
```
kyazdani42, line 20: wont this module global objects be initialized only on first `require` ?

lua/nvim-treesitter/locals.lua
```diff
@@ -89,4 +106,39 @@ function M.get_references(bufnr)
...
+    for _, match in pairs(M.get_locals(bufnr, locals_kind)) do
+      local insert = match
+      -- Follow capture path while we still have a table
+      for _, capture_segment in pairs(capture_path) do
+        if type(insert) == 'table' then
+          insert = insert[capture_segment]
+        end
+      end
+
+      -- Following the capture path was successful
```
kyazdani42, line 133: i think you can remove the comments in this function, the code is already pretty explicit :)

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -212,4 +212,52 @@ function M.get_node_at_cursor(winnr)
...
+  if end_row == vim.fn.line('$') then
+    end_col = #vim.fn.getline('$')
+  end
+
+  vim.fn.setpos(".", { buf, start_row + 1, start_col + 1, 0 })
+  vim.fn.nvim_exec("normal v", false)
+
+  -- Convert exclusive end position to inclusive
+  if end_col == 0 then
+    vim.fn.setpos(".", { buf, end_row, -1, 0 })
```
kyazdani42, line 229: why is `end_row` not changed here but changed in the `else` part ?

lua/nvim-treesitter/locals.lua
theHamsta: Isn't that the goal of a cache to keep results from previous requires? I wanted to initialize the dict only once and then reuse it.

lua/nvim-treesitter/locals.lua
theHamsta: I hope I can use the function from the highlight-definition PR which is more general than this one.

lua/nvim-treesitter/ts_utils.lua
theHamsta: This condition code for two things:
- convert from 1-based indices to 0-based indices
- subtract one character, this changes the row of it's the first character

lua/nvim-treesitter/locals.lua
kyazdani42: what i meant was that you probably doesn't need to check query_cache exists, as this is the global scope of the module and this might not be initialized twice anyway.

lua/nvim-treesitter/ts_utils.lua
kyazdani42: oh i see, my bad.

lua/nvim-treesitter/locals.lua
theHamsta: Query_cache should be the default_dict to support arbitrary query kinds. Query_cachd holds ticks and cache for each query kind.  Maybe you know a better name? I initialize with {} whenever a new local type is requested for the first time

lua/nvim-treesitter/ts_utils.lua
theHamsta: But I should simplify this part to make it obvious

lua/nvim-treesitter/locals.lua
vigoux: We already have this somewhere into `queries.lua`, it is called `split`, we should move it to `utils` then.

lua/nvim-treesitter/locals.lua
kyazdani42: i meant `cache` not `chache`

vigoux:
Now that #110 is merged, can we rebase that ?

theHamsta:
Need to refactor some things. I have time for it this weekend.

vigoux:
No rush, as this is a new feature, it is better to take time

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -234,7 +233,7 @@ function M.find_definition(node, bufnr)
   for _, def in ipairs(locals.get_definitions(bufnr)) do
     for _, def_node in ipairs(M.get_local_nodes(def)) do
       if M.get_node_text(def_node)[1] == node_text then
-        table.insert(matching_def_nodes, def_node)
```
theHamsta, line 237: @steelsojka please use a linter! Your refactoring introduced a lot of post-spaces. We should probably add a linter job as a CI test.

lua/nvim-treesitter/locals.lua
theHamsta: I would rather replace the `split` in queries.lua by `vim.split`. For this I used the implementation of @steelsojka

README.md
```diff
@@ -169,7 +198,8 @@ More information is available in the help file (`:help nvim-treesitter-utils`).
 
 ## Supported Languages
 
-For treesitter to work, we need to use query files such as those you can find in `queries/{lang}/{locals,highlights}.scm`
+For treesitter to work, we need to use query files such as those you can find in
+`queries/{lang}/{locals,highlights,textobjects}.scm`
```
vigoux, line 202: Can we change that while we are at it 
```suggestion
For `nvim-treesitter` to work, we need to use query files such as those you can find in
`queries/{lang}/{locals,highlights,textobjects}.scm`
```

lua/nvim-treesitter/textobjects.lua
```diff
@@ -0,0 +1,99 @@
...
+    for prepared_match in queries.iter_prepared_matches(query, root, bufnr, start_row, end_row) do
+      table.insert(nested, prepared_match)
+    end
+    for _, m in pairs(nested) do
+      for _, n in pairs(m) do
+        if n.node then
+          table.insert(matches, n.node)
+        end
+      end
+    end
```
vigoux, line 41: I think that this can be done in one go :
```suggestion
    for m in queries.iter_prepared_matches(query, root, bufnr, start_row, end_row) do
      for _, n in pairs(m) do
        if n.node then
          table.insert(matches, n.node)
        end
      end
    end
```

theHamsta:
@vigoux should this be possible:
```lua
        ['il'] = { -- put your queries right in the config (more convenient for user extensions)
          python = '(function_definition) @function',
          cpp = '(function_definition) @function',
          c = '(function_definition) @function'
        },
```
pro: users can configure a single addition in a single paragraph instead of creation three files in the correct locations
contra: additional complexity

vigoux:
Hmmm interesting proposal...
Our goal when providing query files was to avoid people to get into how the plugin actually works, and separate clearly the p'ugin from the queries, this way you have three actors : nvim-treesitter contributors, module writers, and query writers.
But in this case, as the queries are really small, I really don't know...
Can you ask that on gitter to have other people's opinion?
@steelsojka what is you opinion on this?

steelsojka:
@vigoux I'm a fan of very flexible apis that allow for lots of user customization only IF they want to go down that path. Can we support both the query files and user defined queries in the config? I think the config is easier for really small changes that a user might like to have, but the query files allows for much larger changes.

theHamsta:
So I leave "as is" and if it is very annoying we can still remove it later. Now both paths are available.

lua/nvim-treesitter/textobjects.lua
```diff
@@ -0,0 +1,95 @@
...
+  if not ft then return end
+  local lang = parsers.ft_to_lang(ft)
+
+  local row, col = unpack(vim.api.nvim_win_get_cursor(0))
+  row = row - 1
+
+  local matches = {}
+
+  if string.match(query_string, '^@.*') then
+    matches = locals.get_capture_matches(bufnr, query_string, 'textobjects')
```
vigoux, line 24: As discussed, this will have to be moved to `query` instead of `locals`.

lua/nvim-treesitter/textobjects.lua
```diff
@@ -0,0 +1,95 @@
...
+  for mapping, query in pairs(config.keymaps) do
+    if type(query) == 'table' then
+      query = query[lang]
+    elseif not queries.get_query(lang, 'textobjects') then
+      query = nil
+    end
+    if query then
+      local cmd = ":lua require'nvim-treesitter.textobjects'.select_textobj('"..query.."')<CR>"
+      api.nvim_buf_set_keymap(buf, "o", mapping, cmd, {silent = true})
+      api.nvim_buf_set_keymap(buf, "v", mapping, cmd, {silent = true})
```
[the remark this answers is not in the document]
theHamsta: I think this is difficult. The function `select_textobject` already selects an textobject which can be directly used for mappings. How the mappings performed and under which circumstances is highlighly user dependent. I can maybe add a doc entry that you could put something like this in your `ftplugin/rust.vim`

```vim
onoremap <silent> aa :lua require'nvim-treesitter.textobjects'.select_textobj('(attribute_item) @rust')<CR>
vnoremap <silent> aa :lua require'nvim-treesitter.textobjects'.select_textobj('(attribute_item) @rust')<CR>
```

lua/nvim-treesitter/textobjects.lua
```diff
@@ -0,0 +1,95 @@
...
+
+  if smallest_range then
+    ts_utils.update_selection(bufnr, smallest_range)
+  end
+end
+
+function M.attach(bufnr, lang)
+  local buf = bufnr or api.nvim_get_current_buf()
+  local config = configs.get_module("textobjects")
+  local lang = lang or parsers.ft_to_lang(api.nvim_buf_get_option(bufnr, "ft"))
```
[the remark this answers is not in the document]
theHamsta: https://github.com/nvim-treesitter/nvim-treesitter/pull/161

lua/nvim-treesitter/textobjects.lua
theHamsta: https://github.com/nvim-treesitter/nvim-treesitter/pull/161

You're right. But we should try not to change to much stuff at once. Better let users test that feature before touching more modules. This PR already affect `textobjects` and `incremental_selection`

theHamsta:
You can try out how textobjects feel like with `strip!` (it's able to remove `{` `}` and white spaces from inner C textobjects. It's not clear whether this is the correct architectural decision but at least it is a good draft of how it should feel like: https://github.com/theHamsta/nvim-treesitter/tree/strip

theHamsta:
@vigoux Rebase #161 or do this PR first?

vigoux:
I think we will go for this pr first.

vigoux:
There is some conflicts that must be resolved before merging, can this be rebased on latest master ?

PR #105, opened 2020-06-22
add highlight queries for bash

TravonteD:
Some things still aren't highlighted properly, but they seem to be known issues in treesitter-bash. 
I'll be keeping track of any fixes that will affect highlighting.

### DIFF
--- a/queries/bash/highlights.scm
+++ b/queries/bash/highlights.scm
@@ -1,0 +1,110 @@
+[
+ "("
+ ")"
+ "{"
+ "}"
+ "["
+ "]"
+ ] @punctuation.bracket
+
+[
+ ";"
+ ";;"
+ (heredoc_start)
+ ] @punctuation.delimiter
+
+[
+ ">"
+ "<"
+ "&"
+ "&&"
+ "|"
+ "||"
+ "="
+ "=="
+ "!="
+ ] @operator
+
+[
+ (string) 
+ (raw_string)
+ (heredoc_body)
+] @string
+
+[
+ "if"
+ "then"
+ "else"
+ "elif"
+ "fi"
+ "case"
+ "in"
+ "esac"
+ ] @conditional
+
+[
+ "for"
+ "do"
+ "done"
+ "while"
+ ] @repeat
+
+[
+ "declare"
+ "export"
+ "local"
+ "readonly"
+ "unset"
+ ] @keyword
+
+[
+ (special_variable_name) 
+ ("$" (special_variable_name))
+ ] @constant
+
+((word) @constant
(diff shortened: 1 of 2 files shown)


queries/bash/highlights.scm
```diff
@@ -0,0 +1,110 @@
...
+ "declare"
+ "export"
+ "local"
+ "readonly"
+ "unset"
+ ] @keyword
+
+[
+ (special_variable_name) 
+ ("$" (special_variable_name))
```
vigoux, line 62: That seems weird.. Doesn't they overlap ?

queries/bash/highlights.scm
```diff
@@ -0,0 +1,110 @@
...
+
+(command_name (word)) @function
+
+(command 
+  argument: [
+             (word) @parameter
+             ((word) @number
+              (#match? @number "^\d*$"))
+             (concatenation (word) @parameter)
+             ])
```
vigoux, line 92: Wow that's a tricky one ! Isn't the `@number` case already covered ?
And the `@parameter` seem to be convered below too

queries/bash/highlights.scm
```diff
@@ -0,0 +1,110 @@
...
+
+(file_redirect
+  descriptor: (file_descriptor) @operator
+  destination: (word) @parameter)
+
+
+("$" (variable_name)) @identifier
```
vigoux, line 99: There is a `(variable_name)` capture below, they seem to overlap

queries/bash/highlights.scm
TravonteD: Nope the parser is weird about some of the matches so they will look a bit "odd" here's an example with an without the second pair.
before
after

you can see here that in the first screenshot the "0" in the special variable is highlighted but not the "$" that goes with it. Surprisingly only the "$" get highlighted if you only use the pair.

queries/bash/highlights.scm
TravonteD: Number has to be explicit here because "(word)" has already been set to "@parameter". The more specific the query the more precidence it seems to have. so if you override in a specific query like this one. you will need to specify again if you have predicates. The difficult thing about the bash parser is that I had regex out the numbers myself :/

queries/bash/highlights.scm
TravonteD: Same issue as above the sibling query is necessary for the "$" to considered the same as the "(variable_name)"

queries/bash/highlights.scm
vigoux: But that seems to work so its okay

PR #107, opened 2020-06-24
Python walrus

theHamsta:
- Add ":=" operator (also to upstream https://github.com/tree-sitter/tree-sitter-python/pull/72)
-  Reset highlighting in f-string interpolation
       This solution is preferable to `(identifier) @Normal` #72 since otherwise
       everything without highlight will use string-highlighting (`(identifier) @Normal` solves this only for identifier).


theHamsta:
Not that I deactivated the highlighting of brackets in my config to use `rainbow_parenthesis`. So brackets have no highlight and should have the default highlighting. The "{" interpolation curly braces still have a highlighting group since they have another query. So the should be highlighted.

Before:

image

After:
image

vigoux:
Can we try using this as an example in the README ?

theHamsta:
As an example for what?

vigoux:
I think that we can show some things any way, otherwise I will try we an example in C or Rust, chich changed quite a lot

theHamsta:
Apparently me memory was bad:
image

With treesitter:
![image](https://user-images.githubusercontent.com/7189118/85690675-17646980-b6d4-11ea-9995-3f2f3d168b51.png

All the intelligent syntax highlighting was done by the Semshi plugin that I used before.

vigoux:
Coool ! Can you add that to the README ?

theHamsta:
There's a reason why we don't use C++:
image
Problem is somewhere in the lua code performing the regexes.

vigoux:
Does it happen each time or after some editing ?

theHamsta:
On some files, without editing. But I only observe it after the runtime update. The messages clearly point to some Lua regex handling code.

vigoux:
Can you raise an issue for that ? With some minimal reproducing steps, and ping me in this, I will dig into this.

kyazdani42:
i'm pretty sure this is still the same issue related to buffer updates, highlighting is not yet stable at all, editing code with it doesn't work, it happens in any language when you operate with `~` or `r`

vigoux:
> i'm pretty sure this is still the same issue related to buffer updates, highlighting is not yet stable at all, editing code with it doesn't work, it happens in any language when you operate with `~` or `r`

If it happens without edits, then I am pretty sure it is not related to that

kyazdani42:
I read wrong sorry

theHamsta:
I think this can be merged. Screenshots can go to separate PR.

theHamsta:
I fixed the cpp queries: #112

PR #113, opened 2020-06-26
C highlights: Make ? an operator in c highlights

theHamsta:
Question: Should ":" also be highlighted as an operator if used in the context of "?"? For me, `@delimiter` is also ok.

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -48,6 +48,7 @@
 ">=" @operator
 "!" @operator
 "||" @operator
+"?" @operator
 
 "-=" @operator
 "+=" @operator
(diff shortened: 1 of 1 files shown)


vigoux:
I think that highlighting the `:` makes more sense as an `operator`.

kyazdani42:
i would have considered it more like a `@conditional` thing, but operator would work fine too.

queries/c/highlights.scm
```diff
@@ -48,6 +48,7 @@
 ">=" @operator
 "!" @operator
 "||" @operator
+"?" @operator
```
vigoux, line 51: I agree with @kyazdani42 , I would prefer it to be as a conditional...
```suggestion
(conditional_expression [ "?" ":" ] @conditional)
```

PR #114, opened 2020-06-26
Add is? predicate

vigoux:
As it is harder than I thought, I am opening yet another WIP PR...
We need to carefully tackle this because we will have to separate internal queries and how we expose them.
We might need to break some things along the way for locals, which is sad now that `completion-nvim` is finally based on `nvim-treesitter`...

My proposal is :
 - [ ] to move some things like `get_query` to an internal file called `query_common`
 - [ ] refactor the locals to use this
 - [ ] actually implement de new predicate

As this is our first implemented "matching" predicate, it should be easier afterwards... Hopefully...

### DIFF
--- a/lua/nvim-treesitter/utils/query.lua
+++ b/lua/nvim-treesitter/utils/query.lua
@@ -1,0 +1,107 @@
+local api = vim.api
+local fn = vim.fn
+local luv = vim.loop
+local ts = vim.treesitter
+
+local M = {}
+
+local function read_query_files(filenames)
+  local contents = {}
+
+  for _,filename in ipairs(filenames) do
+    vim.list_extend(contents, vim.fn.readfile(filename))
+  end
+
+  return table.concat(contents, '\n')
+end
+
+-- Some treesitter grammars extend others.
+-- We can use that to import the queries of the base language
+M.base_language_map = {
+  cpp = {'c'},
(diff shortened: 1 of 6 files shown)


kyazdani42:
i wonder if we should move those predicates thing inside neovim's api ? this is highly related to the treesitter api right ? If they are used in the highlighter, the should be available also for locals as a function of some kind. (enlighten me if i'm mistaken somehow)

vigoux:
Actually yes but we can't because neobim doesn't have locals, which is required for the `is?` predicate.

vigoux:
Little update on this : this is really harder than what I thought, and it will be quite hard to use it within our highlighter.
Implementing the `is?` predicate in itself requires us to rely on the `locals.scm` query, which is problematic in itself.

lua/nvim-treesitter/utils/query.lua
```diff
@@ -0,0 +1,107 @@
...
+end
+
+
+function M.iter_prepared_matches(match_func, query, qnode, bufnr, start_row, end_row)
+  -- A function that splits  a string on '.'
+  local function split(string)
+    local t = {}
+    for str in string.gmatch(string, "([^.]+)") do
+      table.insert(t, str)
+    end
```
theHamsta, line 68: Can't we use `vim.split` instead?

PR #117, opened 2020-06-27
Extend documentation for TSError

theHamsta:
I would love to add also a documentation here how to deactivate the error highlighting here. But I don't know how.

### DIFF
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -170,7 +170,13 @@ Note: This is highly experimental, and folding can break on some types of
 ==============================================================================
 HIGHLIGHTS                                        *nvim-treesitter-highlights*
 
-TSError
+`TSError`
+							          *hl-TSError*
+For syntax/parser errors.
+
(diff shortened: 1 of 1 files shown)


doc/nvim-treesitter.txt
```diff
@@ -170,7 +170,9 @@ Note: This is highly experimental, and folding can break on some types of
 ==============================================================================
 HIGHLIGHTS                                        *nvim-treesitter-highlights*
 
-TSError
+`TSError`
+							          *hl-TSError*
+For syntax/parser errors.
```
kyazdani42, line 175: is this highlight group disabled by default ? If so, you could write about that here.

theHamsta: It's enabled. I would prefer disabled.

vigoux: Then we can easily disable by default, can you change this in this PR ?

theHamsta: You can change it to default=Off if you like.

Issue #94, opened 2020-06-21
How to highlight JSX elements?


JoosepAlviste:
**Describe the bug**
I'm trying to add a custom highlighting color to JSX elements, but am failing to do so. I have the following query in `after/queries/javascript/highlights.scm`:

```
(jsx_text) @function
```

But when I open Neovim, then I get this error:

```
Error detected while processing <home dir>/.local/share/nvim/plugged/nvim-treesitter/plugin/nvim-treesitter.vim:
line   20:
E5108: Error executing lua ...m/HEAD-a0a84fc/share/nvim/runtime/lua/vim/treesitter.lua:127: query: invalid node type at position 3054
```

I saw that the `tree-sitter-javascript` parser includes the JSX queries in [`highlights-jsx.scm`](https://github.com/tree-sitter/tree-sitter-javascript/blob/master/queries/highlights-jsx.scm) not `highlights.scm`. Could that be the reason why it's failing for me? I tried to move this query to `highlights-jsx.scm` but it doesn't look like that file is read.

**To Reproduce**
Here is a minimal case to reproduce the problem:

`init.vim`:

```vim
call plug#begin(stdpath('data') . '/plugged')
Plug 'nvim-treesitter/nvim-treesitter'
call plug#end()

syntax off

lua <<EOF
require'nvim-treesitter.configs'.setup {
    highlight = {
        enable = true,
    },
    incremental_selection = {
        enable = false,
    },
    ensure_installed = {'javascript'}
}
EOF
```

`after/queries/javascript/highlights.scm`:

```
(jsx_text) @function
```

An example JSX file (`MyComponent.js`):

```js
import React from 'react';

const MyComponent = () => (
  <div>Hello World</div>
);

export const MyComponent;
```

**Expected behavior**
I would expect there to not be an error and the JSX text element should be highlighted with the highlight group I specified.

**Output of `:checkhealth nvim_treesitter`**

```
health#nvim_treesitter#check
========================================================================
  - ERROR: Failed to run healthcheck for "nvim_treesitter" plugin. Exception:
    function health#check[21]..health#nvim_treesitter#check, line 1
    Vim(lua):E5108: Error executing lua ...m/HEAD-a0a84fc/share/nvim/runtime/lua/vim/treesitter.lua:127: query: invalid node type at position 3054
```

Here is the health check results if I remove the query:

```
health#nvim_treesitter#check
========================================================================
## Installation
  - OK: `git` executable found.
[...]

PR #125, opened 2020-06-29
add used_by key to parsers

kyazdani42:
Enables the use of multiple filetypes for one parser.
You can now say that `zsh` uses the parser for `bash`, or any other ones.
This way you can setup parsers like
```lua
local parserlist = require'nvim-treesitter.parsers'.list
parserlist.some_parser.used_by = {'some_filetype', 'some_other_filetype'}
```
this doesn't fix any issue that i'm aware of, there is still work to do around the language integrations.
this could also be used as a temporary fix to #94 by disabling the typescript parser and use only the tsx one by commenting the typescript parser and adding the `used_by` key in the tsx config with `{ 'typescript' }` in it.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -49,9 +49,10 @@ end
 local function enable_mod_conf_autocmd(mod, lang)
   if not config.modules[mod] or M.is_enabled(mod, lang) then return end
 
-  local ft = parsers.lang_to_ft(lang)
   local cmd = string.format("lua require'nvim-treesitter.%s'.attach()", mod)
-  api.nvim_command(string.format("autocmd FileType %s %s", ft, cmd))
+  for _, ft in pairs(parsers.lang_to_ft(lang)) do
+    api.nvim_command(string.format("autocmd NvimTreesitter FileType %s %s", ft, cmd))
+  end
   for i, parser in pairs(config.modules[mod].disable) do
     if parser == lang then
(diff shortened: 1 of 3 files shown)


lua/nvim-treesitter/configs.lua
```diff
@@ -101,15 +102,18 @@ end
 local function disable_mod_conf_autocmd(mod, lang)
   if not config.modules[mod] or not M.is_enabled(mod, lang) then return end
 
-  local ft = parsers.lang_to_ft(lang)
-  api.nvim_command(string.format("autocmd! FileType %s", ft))
+  local cmd = string.format("lua require'nvim-treesitter.%s'.attach()", mod)
+  -- TODO(kyazdani): detach the correct autocmd... doesn't work when using %s, cmd
+  for _, ft in pairs(parsers.lang_to_ft(lang)) do
+    api.nvim_command(string.format("autocmd! NvimTreesitter FileType %s", ft))
+  end
```
vigoux, line 109: Can we have some comments on why that change ?

kyazdani42: hum basically before it didn't work because the augroup wasn't specified. Now it does, but i cannot remove a particular autocmd in the NvimTreesitter group, which causes TSDIsableAll to disable all modules autocmds...

kyazdani42:
we might need to add some docs for that.

vigoux:
We will need some big `tidy all and document` PR soon...

PR #143, opened 2020-07-02
feat(configs): dynamic module defintions

steelsojka:
This reworks the module system to allow module definitions to be rendered dynamically.

## Why?
The module system takes care of a lot of boilerplate (language detection, enable/disable, query detection, buffer attachment, etc...) that could be leveraged if made public. These are the benefits this feature provides:

* Another way to consume this plugin. Plugin authors can just use the utils or right a module to take advantage of the interface this plugin provides.
* Module definitions no longer depend on the ordering of files that are loaded. We can load modules after the plugin is initialized and before or after the user has configured it (setup call). For example, if a plugin creates a module, that plugin could be loaded dynamically (vim-plug or others) without any issues and still get registered correctly.
* Simpler modules for plugins/users to write. No query file detection or language support logic has to be implemented by the plugin author or user.
* Builtin modules no longer have a reliance on directory structure of the project.
* User modules use the exact same api as the internal built in modules.

Here is an example of how a module can be registered.

```lua
require "nvim-treesitter".define_modules {
  some_cool_plugin = {
    feature_one = {
      module_path = "some_cool_plugin.feature_one"
    }
  }
}
```

These modules are just like any other module, so these commands work `:TSModuleInfo some_cool_plugin.feature_one`

Optionally, modules can provide `attach` and `detach` directly.

```lua
require "nvim-treesitter".define_modules {
  some_cool_plugin = {
    feature_one = {
      attach = function(bufnr, lang) end,
      detach = function(bufnr, lang) end
    }
  }
}
```

- [x] Docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ $ git clone https://github.com/nvim-treesitter/nvim-treesitter.git
 
 ## Adding parsers
 
-Treesitter is using a different _parser_ for every language. It can be quite a pain to install, but fortunately `nvim-treesitter` 
+Treesitter is using a different _parser_ for every language. It can be quite a pain to install, but fortunately `nvim-treesitter`
 provides two command to tackle this issue:
   - `TSInstall` to install one or more parser. You can use `TSInstall all` to download all parsers.
   - `TSInstallInfo` to know which parser is installed.
@@ -158,6 +158,41 @@ The roadmap and all features of this plugin are open to change, and any suggesti
 - `refactor.navigation`: Syntax based definition listing and navigation.
   * List all definitions
   * Go to definition
+
+## Defining Modules
+
+Users and plugin authors can take advantage of modules by creating their own. Modules provide:
+
+- Treesitter language detection support
+- Attach and detach to buffers
+- Works with all nvim-treesitter commands
+
+You can use the `define_modules` function to define one or more modules or module groups.
+
+```lua
+require 'nvim-treesitter'.define_modules {
+  my_cool_plugin = {
+    attach = function(bufnr, lang)
+      -- Do cool stuff here
+    end,
+    detach = function(bufnr)
+      -- Undo cool stuff here
+    end,
+    is_supported = function(lang)
+      -- Check if the language is supported
+    end
+  }
+}
+```
+
+Modules can consist of the following properties:
+
+- `module_path`: A require path (string) that exports a module with an `attach` and `detach` function. This is not required if the functions are on this definition.
+- `enable`: Determines if the module is enabled by default. This is usually overridden by the user.
+- `disable`: A list of languages that this module is disabled for. This is usually overridden by the user.
+- `is_supported`: A function that takes a language and determines if this module supports that language.
+- `attach`: A function that attachs to a buffer. This is required if `module_path` is not provided.
+- `detach`: A function that detaches from a buffer. This is required if `module_path` is not provided.
 
 ## Utils
 
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -4,76 +4,95 @@ local queries = require'nvim-treesitter.query'
 local parsers = require'nvim-treesitter.parsers'
 local utils = require'nvim-treesitter.utils'
 
--- @enable can be true or false
--- @disable is a list of languages, only relevant if enable is true
--- @keymaps list of user mappings for a given module if relevant
--- @is_supported function which, given a ft, will return true if the ft works on the module
 local config = {
-  modules = {
-    highlight = {
-      enable = false,
-      disable = {},
-      is_supported = queries.has_highlights
+  modules = {},
+  ensure_installed = nil
+}
+local user_setup_data = nil
+-- List of modules that need to be setup on initialization.
+local queued_modules_defs = {}
+-- Whether we've initialized the plugin yet.
+local is_initialized = false
+local builtin_modules = {
+  highlight = {
+    module_path = 'nvim-treesitter.highlight',
+    enable = false,
+    disable = {},
+    is_supported = queries.has_highlights
+  },
+  incremental_selection = {
+    module_path = 'nvim-treesitter.incremental_selection',
+    enable = false,
+    disable = {},
+    keymaps = {
+      init_selection="gnn",
+      node_incremental="grn",
+      scope_incremental="grc",
+      node_decremental="grm"
     },
-    incremental_selection = {
+    is_supported = queries.has_locals
+  },
+  refactor = {
+    highlight_definitions = {
+      module_path = 'nvim-treesitter.refactor.highlight_definitions',
       enable = false,
       disable = {},
-      keymaps = {
-        init_selection="gnn",
-        node_incremental="grn",
-        scope_incremental="grc",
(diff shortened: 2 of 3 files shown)


lua/nvim-treesitter/configs.lua
```diff
@@ -286,10 +379,23 @@ function M.get_module(mod_path)
...
+-- A module should contain an attach and detach function.
 -- @param mod the module table
 function M.is_module(mod)
-  return type(mod) == 'table' and type(mod.is_supported) == 'function' 
+  return type(mod) == 'table' 
+    and ((type(mod.attach) == 'function' and type(mod.detach) == 'function') 
+      or type(mod.module_path) == 'string')
+end
+
+-- Initializes builting modules and any queued modules
```
theHamsta, line 390: built-in

lua/nvim-treesitter/configs.lua
```diff
@@ -286,10 +379,23 @@ function M.get_module(mod_path)
 end
 
 -- Determines whether the provided table is a module.
--- A module should contain an 'is_supported' function.
+-- A module should contain an attach and detach function.
 -- @param mod the module table
 function M.is_module(mod)
-  return type(mod) == 'table' and type(mod.is_supported) == 'function' 
+  return type(mod) == 'table'
```
theHamsta, line 385: postspace

lua/nvim-treesitter/configs.lua
```diff
@@ -286,10 +379,23 @@ function M.get_module(mod_path)
 end
 
 -- Determines whether the provided table is a module.
--- A module should contain an 'is_supported' function.
+-- A module should contain an attach and detach function.
 -- @param mod the module table
 function M.is_module(mod)
-  return type(mod) == 'table' and type(mod.is_supported) == 'function' 
+  return type(mod) == 'table' 
+    and ((type(mod.attach) == 'function' and type(mod.detach) == 'function')
```
theHamsta, line 386: postspace

lua/nvim-treesitter/configs.lua
```diff
@@ -286,10 +379,23 @@ function M.get_module(mod_path)
...
+    and ((type(mod.attach) == 'function' and type(mod.detach) == 'function') 
+      or type(mod.module_path) == 'string')
+end
+
+-- Initializes builting modules and any queued modules
+-- registered by plugins or the user.
+function M.init()
+  is_initialized = true
+  M.define_modules(builtin_modules)
+
```
theHamsta, line 395: line containing only spaces

lua/nvim-treesitter/configs.lua
```diff
@@ -260,18 +278,93 @@ function M.setup_module(mod, data, mod_name)
...
+    table.insert(queued_modules_defs, mod_defs)
+    return
+  end
+
+  recurse_modules(function(key, mod, _, group)
+    group[key] = vim.tbl_extend("keep", mod, {
+      enable = false,
+      disable = {},
+      is_supported = function() return true end
+    })
```
theHamsta, line 324: a lot of post spaces.

lua/nvim-treesitter/configs.lua
```diff
@@ -4,76 +4,95 @@ local queries = require'nvim-treesitter.query'
 local parsers = require'nvim-treesitter.parsers'
 local utils = require'nvim-treesitter.utils'
 
--- @enable can be true or false
--- @disable is a list of languages, only relevant if enable is true
--- @keymaps list of user mappings for a given module if relevant
--- @is_supported function which, given a ft, will return true if the ft works on the module
 local config = {
```
theHamsta, line 7: @vigoux aren't those locals not overwritten on every require. Also before it was not possible to get the current config after a user call of `setup` since it was always set back to the default values. `:lua print(vim.inspect(require'nvim-treesitter.configs'.get_module('incremental_selection') ` returns `nil`

steelsojka: I'll look into that. Does lua run the file on every require or is it cached after the first require of it?

steelsojka: Also, this depends when you get the module. The built-in modules don't get defined until `require('nvim-treesitter').setup()` is called.

theHamsta: I guess lua runs the whole file on every require. You can check that by placing a small print statement next to the local definition. It is at least called again when a user runs it. Don't whether it's cached when the same thread requires it again.

vigoux:
This needs a rebase now, and style checking will run automatically. Can you add some docs on this too ?

lua/nvim-treesitter/configs.lua
steelsojka: I added a some config to remove trailing whitespace... have it configured for js/ts/html but not lua. Sorry about that.

lua/nvim-treesitter/configs.lua
steelsojka: @theHamsta What's the desired result? If I run `:lua print(require'nvim-treesitter.configs'.get_module('incremental_selection'))` I get a table. Lua doesn't run the file for each require (I tried what you suggested with the print statements) so the config state shouldn't be affected.

I think the one key difference with this implementation vs the previous is modules are now defined async. They could be defined at any time. So you can't "get all modules", since a plugin may not have defined it yet, but you can get all registered at that time.

theHamsta: Nice! Now its also working for me. Thanks for checking!

lua/nvim-treesitter/configs.lua
theHamsta: I think you only need to fix that typo.

lua/nvim-treesitter/configs.lua
```diff
@@ -4,76 +4,96 @@ local queries = require'nvim-treesitter.query'
...
-  modules = {
-    highlight = {
+  modules = {},
+  ensure_installed = nil
+}
+local user_setup_data = nil
+print('run')
```
theHamsta, line 12: Thanks for answering me question. Now you can remove this.

steelsojka: lol oops. Sorry about that.

README.md
```diff
@@ -159,6 +159,41 @@ The roadmap and all features of this plugin are open to change, and any suggesti
...
+```
+
+Modules can consist of the following properties:
+
+- `module_path`: A require path (string) that exports a module with an `attach` and `detach` function. This is not required if the functions are on this definition.
+- `enable?`: Determines if the module is enabled by default.
+- `disable?`: A list of languages that this module is disabled for. This is usually overridden by the user.
+- `is_supported?`: A function that takes a language and determines if this module supports that language.
+- `attach?`: A function that attachs to a buffer. This is required if `module_path` is not provided.
+- `detach?`: A function that detaches from a buffer. This is required if `module_path` is not provided.
```
vigoux, line 195: Why adding `?` here ? Because they are optional ?

steelsojka: Yeah... Sorry. I write a lot of typescript and that's the notation for optional. What do you prefer to mark them as optional.

vigoux: Hmmm, maybe just add a notice before all of that, the notation is fine once this is explained.

PR #144, opened 2020-07-04
Add Travis CI with luacheck

theHamsta:
Some style checking would be useful. Some issues are already fixed in #100 

Luacheck is happy now: https://travis-ci.org/github/theHamsta/nvim-treesitter/builds/704953331

Edit: ooops, it was only checking one file on bash.

### DIFF
--- a/lua/nvim-treesitter/refactor/navigation.lua
+++ b/lua/nvim-treesitter/refactor/navigation.lua
@@ -7,18 +7,6 @@ local api = vim.api
 
 local M = {}
 
-local function node_to_qf(node, kind)
-  local lnum, col, _ = def.node:start()
-
-  return {
-    bufnr = bufnr,
-    lnum = lnum + 1,
-    col = col + 1,
-    text = ts_utils.get_node_text(def.node)[1] or '',
-    kind = kind
-  }
-end
-
 function M.goto_definition(bufnr)
   local bufnr = bufnr or api.nvim_get_current_buf()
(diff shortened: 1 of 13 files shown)


theHamsta:
Now everything is fine: https://travis-ci.org/github/theHamsta/nvim-treesitter/builds/704954174

theHamsta:
Currently, I cannot activate travis-ci.org for the main repo (I only see repos for my other organistation). But it should be possible for @vigoux when you're owner of the organization.

lua/nvim-treesitter/refactor/navigation.lua
```diff
@@ -7,18 +7,6 @@ local api = vim.api
...
-
-  return {
-    bufnr = bufnr,
-    lnum = lnum + 1,
-    col = col + 1,
-    text = ts_utils.get_node_text(def.node)[1] or '',
-    kind = kind
-  }
-end
-
```
theHamsta, line 21: Probably useful but unused.

vigoux:
Awesome! I will review that more in depth later, but enabling it is cool!
But can we use GH actions instead? So that we have everything centralized in here?

theHamsta:
I learned Travis, I learned Gitlab CI. I don't want to learn GH actions.

I don't know what kind of integration you're missing from Travis. If you enable it for this repo it will be displayed as a automatically for each PR.

theHamsta:
We can switch to GH actions later if you want to. But someone has to look into how it works. Maybe when we're adding unit tests.

PR #147, opened 2020-07-05
Add strip! directive for queries

theHamsta:
The strip! directive is implemented in the tree-sitter tags crate in the main tree-sitter repo
(https://github.com/tree-sitter/tree-sitter/blob/de53b82e2c48d6221af9881e50cc57d961de3365/tags/src/lib.rs#L167).
It is used specifically for tag extraction to shrink the range of a node (or better the text of a node).

For our textobjects queries the strip! directive might be use full since it allows us to strip aways whitespaces and leading/trailing characters. I implemented `strip!` to behave the same way as in tree-sitter/tags. Except that I am calculating a new `processed_range` that can be used by applications that expect the query to be somehow post-processed. For textobjects a `expand!` directive would also be useful to add trailing whitespaces to the node range. I am using `strip!` on this branch to strip away "{" "}" and white spaces from blocks.

```scheme
((compound_statement) @block.inner
  (strip! @block.inner "(^\{?\s*)|(\s*\}?$)"))
```
While implementing this I discovered a bug in Neovim https://github.com/neovim/neovim/issues/12595 (you cannot escape " at the moment. Since `\"` gets escape to `\\"`).

Since its a application-specific extension I don't think it needs to be be pushed to upstream Neovim (and if yes we should get some experiences with it locally).

I don't know if this the right way to go since it introduces a new concept of a processed range that does not need to correspond to a tree-sitter node.

### DIFF
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -1,5 +1,6 @@
 local api = vim.api
 local ts = vim.treesitter
+local utils = require 'nvim-treesitter.utils'
 
 local M = {}
 
@@ -71,6 +72,56 @@ function M.get_query(lang, query_name)
   end
 end
 
+ --This function is copied from Neovims treesitter.lua to enforece same behavior
+local magic_prefixes = {['\\v']=true, ['\\m']=true, ['\\M']=true, ['\\V']=true}
+local function check_magic(str)
+  if string.len(str) < 2 or magic_prefixes[string.sub(str,1,2)] then
+    return str
+  end
+  return '\\v'..str
+end
+
+local function strip_beginning(lines, regex)
+  local current_col = 0
+  local current_line = 0
+  local re = vim.regex(check_magic('^('..regex..')'))
+  for linenr, line in ipairs(lines) do
+    current_line = linenr
+    local match_start, match_end = re:match_str(line)
+    if match_start ~= 0 then
+      break
+    else
+      current_col = match_end
+    end
+    if match_end ~= vim.str_byteindex(line, #line) then
+      break
+    end
+    current_line = current_line + 1
+  end
+  return current_line - 1, current_col
+end
+
+local function strip_end(lines, regex)
+  local current_col = 0
+  local line_diff = 0
+  local re = vim.regex(check_magic('('..regex..')$'))
+  for linenr=#lines, 1, -1 do
+    local line = lines[linenr]
+    line_diff = #lines - linenr
+    local match_start, match_end = re:match_str(line)
+    if match_end ~= vim.str_byteindex(line, #line) then
+      break
+    else
+      current_col = match_start
+    end
+    if match_start ~= 0 then
+      break
+    end
+    line_diff = line_diff + 1
+  end
+  return line_diff, current_col
+end
+
 function M.iter_prepared_matches(query, qnode, bufnr, start_row, end_row)
   -- A function that splits  a string on '.'
   local function split(string)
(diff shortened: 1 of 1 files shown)


theHamsta:
For performance reasons, I should cache the compiled regexes I created in a table.

theHamsta:
Probably, the stripping could also be deferred and only executed when `processed_range` is ever requested. So set a `process_range` function instead.

vigoux:
I don't really see what use this predicate has, can you pull us up a little example use case ?

theHamsta:
https://github.com/theHamsta/nvim-treesitter/tree/strip just try out the @block.inner query for C.
It gives you the range of a block between the braces.

lua/nvim-treesitter/query.lua
```diff
@@ -120,6 +185,39 @@ function M.iter_prepared_matches(query, qnode, bufnr, start_row, end_row)
           if pred[1] == "set!" and type(pred[2]) == "string" then
             insert_to_path(prepared_match, split(pred[2]), pred[3])
           end
+          if pred[1] == "strip!" and #pred == 3 then
+            local query_name = query.captures[pred[2]]
```
vigoux, line 189:
```suggestion
            local capture_name = query.captures[pred[2]]
```

lua/nvim-treesitter/query.lua
```diff
@@ -120,6 +185,39 @@ function M.iter_prepared_matches(query, qnode, bufnr, start_row, end_row)
...
+            local regex = pred[3]
+
+            local node = utils.get_at_path(prepared_match, query_name..'.node')
+            local ts_utils = require 'nvim-treesitter.ts_utils'
```
vigoux, line 193: Please move that above, otherwise it will be critical in terms of performances

lua/nvim-treesitter/query.lua
```diff
@@ -120,6 +185,39 @@ function M.iter_prepared_matches(query, qnode, bufnr, start_row, end_row)
...
+            else
+              start_col =strip_col
+            end
+            local strip_line, strip_col = strip_end(node_lines, regex)
+            end_line = end_line - strip_line
+            if strip_line == 0 then
+              end_col = end_col - strip_col
+            else
+              end_col = strip_col
+            end
```
vigoux, line 210: That name redefinition makes things less clear, I think that luacheck will be triggered by this

theHamsta: Actually not, this is the way luacheck likes. For the redefinition of the local as @kyazdani42 does it all over the repo I had to disable the warning.

Strip transforms the range so I found it natural to modify the range variables. But I can rename it to processed_end_line.

lua/nvim-treesitter/query.lua
```diff
@@ -71,6 +72,70 @@ function M.get_query(lang, query_name)
...
+    M._regex_cache = {}
+  end
+  if not M._regex_cache[regex] then
+    local re = vim.regex(check_magic(regex))
+    M._regex_cache[regex] = re
+    return re
+  else
+    return M._regex_cache[regex]
+  end
+end
```
vigoux, line 95: Are we sure that regexes are not compiled in the query already ? If that's the case, then we should use it, if not, we should post an PR upstream to fix that.

lua/nvim-treesitter/query.lua
vigoux: Ok, that's fine for me

lua/nvim-treesitter/query.lua
theHamsta: As far as I could see it it is performed only for `match?` right now (https://github.com/neovim/neovim/blob/bd5f0e9695cb21c8b97f844ce21432ee8d06b7ed/runtime/lua/vim/treesitter.lua#L140). Neovim/Treesitter do not know anything of the semantics of the directives so they cannot decide whether a string is a regex or something else.

lua/nvim-treesitter/query.lua
theHamsta: This is needed to avoid a circular dependency. But maybe I can avoid `get_node_text` by using `match_line`

lua/nvim-treesitter/query.lua
vigoux: Oh you are right...
Then that is a problem, hen we will need for quite some time I guess...
On thing that might be worth investigating too (not really related, just want to throw that idea here) is trying to override `Query:match_preds` which is the part of the `Query` metatable which matches the predicates.

Another thing is maybe upstreaming some of our modifications to neovim, but that's another discussion.

theHamsta: I think there's a difference between directives for choosing matches and application-specific annotations. I guess its even ok if a nvim-treesitter module has its own post-processing directives (they could then be implemented even in the module itself). Then, users can configure the behavior of the module on a query basis.

For textobjects it makes sense to trim/extend a matched node. Whitespaces will never be captured by the syntax tree but they are important for this feature. Also, to strip leading `// ` of comments when you want to change a inner comment.

Matching sub-regions of a node could also get relevant for highlighting of injected inline code in comments: You have to merge multiple comment nodes, remove leading `// ` and determine a region where the example code starts. We have to see which parts are useful in a greater context and which not.

lua/nvim-treesitter/query.lua
theHamsta: Circular dependency was fixed in master by style checking.

theHamsta:
Ok, I need to think how multiline behavior should be like. Probably, it would be enough to just strip first and last line of match.

theHamsta:
Ok, the problem with muliline regexes still exist:

 - vim.fn.match_str only supports single line regexes

You can get the right feeling for multi-line regexes if you add a separate query for continuation lines (current darf).

- lua regexes support multiline regex but they have a different syntax compared to the currently used vim regexes.

theHamsta:
There's now a better solution for this #590

PR #148, opened 2020-07-05
Allow arbitrary query files in locals.lua

theHamsta:
This is a preparation for `textobject` queries. Maybe a bit easier to review #100 in smaller pieces.

### DIFF
--- a/lua/nvim-treesitter/locals.lua
+++ b/lua/nvim-treesitter/locals.lua
@@ -5,16 +5,30 @@ local api = vim.api
 
 local queries = require'nvim-treesitter.query'
 local parsers = require'nvim-treesitter.parsers'
+local utils = require'nvim-treesitter.utils'
 
-local M = {
-  locals = {}
+local default_dict = {
+  __index = function(table, key)
+    local exists = rawget(table, key)
+    if not exists then
+      table[key] = {}
+    end
+    return rawget(table, key)
+  end
 }
 
-function M.collect_locals(bufnr)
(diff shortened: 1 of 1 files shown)


vigoux:
I don't really understand what you mean by `locals_kind` here, do you mean `scope` and such or am I misled ?

theHamsta:
The kind of query files we are caching. Locals is using `locals.scm`. The textobjects module uses the same infrastructure for `textobjects.scm`. Users can create their own collection of queries if the want to.

vigoux:
But that's not locals then ?

vigoux:
> Edit: maybe query_type is a better name?

My question in more : why using `locals` instead of `query`, locals are really meant to extract "semantic" information from the syntax tree, why not refactoring this into something separate ?

theHamsta:
You mean renaming `locals.lua`? It keeps its original functionality and API stays the same. No functionality is added.

Everything in `locals.lua` is about getting semantic regions from the current buffer, so its also needed for textobjects (even though its rather an internal detail instead of a change for user API)

vigoux:
As you want, let's take that modification in for now, I'll do a refactor later anyway to add the `is?` predicate.

theHamsta:
You were right! Probably, all the stuff that is shared with textobjects should go to a own lua file and scope/definition/reference functionality from `ts_utils` should go `locals.lua`

vigoux:
That's what I am planning to do soon, no worry, hopefully I can get that out later next week

PR #149, opened 2020-07-05
Lua fix overlap

vigoux:
This avoids those annoying flickering highlights for now

### DIFF
--- a/queries/lua/highlights.scm
+++ b/queries/lua/highlights.scm
@@ -2,25 +2,31 @@
 
 ;;; Builtins
 ;; Keywords
+(
+  [
+  "if"
+  "then"
+  "else"
+  "elseif"
+   ] @conditional
+  "end" @confitional
+)
+
+
+(
+  [
+  "do"
+  "while"
+  "repeat"
+  "for"
+  ] @repeat
+
+  "end" @repeat
+)
 
 [
-"if"
-"then"
-"else"
-"elseif"
- ] @conditional
-
-[
-"do"
-"while"
-"repeat"
-"for"
 "in"
-] @repeat
-
-[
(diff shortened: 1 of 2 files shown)


theHamsta:
I guess this fixed some problems with flickering. However with this`end` is not highlighted in many occasions or this outer function

image

vigoux:
Yeah, that bug is quite turning me mad, I don't know why it happens... Trying it in the TS playground gives correct results, so I guess to problem comes from or implementation...

theHamsta:
The Lua parser itself has some issues.
If you think this is an improvement over the previous state we can merge. I had the feeling that at least the flickering was reduced.

queries/lua/highlights.scm
```diff
@@ -2,25 +2,31 @@
...
-"in"
-] @repeat
+(
+  [
+  "do"
+  "while"
+  "repeat"
+  "for"
+  ] @repeat
+
```
theHamsta, line 23: not (_)* between keyword and end?

queries/lua/highlights.scm
```diff
@@ -2,25 +2,31 @@
 
 ;;; Builtins
 ;; Keywords
+(
+  [
+  "if"
+  "then"
+  "else"
+  "elseif"
+   ] @conditional
```
theHamsta, line 11: same here

queries/lua/highlights.scm
vigoux: No need for this because of the semantic of this kind of matches : it matches sibling node, thus if if will match both anyway.

vigoux:
It should be fixed soon when highlighting get fixed upstream

PR #150, opened 2020-07-05
Add pre-push script

theHamsta:
It denies pushing when you're not passing luacheck. Has to be copied to `.git/hooks` to activate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,4 @@ before_install:
   - sudo luarocks install luacheck
 
 script:
-  - luacheck **/**/*.lua --codes
+  - ./scripts/style-check.sh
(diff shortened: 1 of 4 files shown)


theHamsta:
Btw. is there a ways to get all subfolders in bash without using `find`. `**` was doing this job with zsh.

vigoux:
Not sure pre-push is really a gain in this, but adding this a some kind of CI scripts would be great. I like the idea !

vigoux:
> Btw. is there a ways to get all subfolders in bash without using `find`. `**` was doing this job with zsh.

Using find is fine, and that's more shell-agnostic, so we will keep it.

theHamsta:
I like pre-push scripts, especially if they are directly in the repo. For me its that subset of checks that can be run below two seconds so that it saves you time by avoiding pushing something that would fail 20secs later on CI.

Should I rename it to `style-check.sh`?

vigoux:
I quite like your argument of avoiding to wait for pr builds... Let's call it this way and add a notice to the CONTRIBUTING

vigoux:
Can we creatz a `scripts` dit for that?

PR #153, opened 2020-07-06
Fix C and add a screenshot

vigoux:
Let's add a Screenshot to the repo, so that everybody has a quick overview of how treesitter highlighting will change everything !
This screenshot will evolve with the plugin for sure.


queries/c/highlights.scm
```diff
@@ -96,12 +101,13 @@
...
-(sized_type_specifier) @type
 
-((identifier) @type
- (#match? @type "^[A-Z]"))
+[
+(type_identifier)
+(primitive_type)
+(sized_type_specifier)
+ ] @type
+(binary_expression left: (identifier) @type)
```
vigoux, line 110: Strangely enough, this is the type

queries/c/highlights.scm
```diff
@@ -96,12 +101,13 @@
 
 (field_identifier) @property
 (statement_identifier) @label
-(type_identifier) @type
-(primitive_type) @type
-(sized_type_specifier) @type
 
-((identifier) @type
```
theHamsta, line 103: Why remove this? This is the only chance how we can detect types when their are used (I know this convention is often only used for C++). What else would a upper case identifier mean (even in C)?

C could use additionally `foo_t` with regex "_t$" for types.

If you think this is to C++-like, move this to C++-highlights?

vigoux: No not at all but that actually creates wrong highlights, please refer to my other comment at the same point.

vigoux: This version : `(binary_expression left: (identifier) @type)` actually does highlight the types too in declarations, and also avoid some highlighting errors.
I think that I have misses the same highlight in the `cast` nodes though.

theHamsta: What expression would cause a binary_expression for example?
Using the naming convention is better (query is copied from C++ upstream) since we can not cover all situations where a type can occur. In C and especially C++ a type can occur in virutally any identifier position.

queries/c/highlights.scm
```diff
@@ -71,6 +71,11 @@
   "&="
```
theHamsta, line 71: Would you add operators "<<" ">>"

queries/c/highlights.scm
vigoux: > in declarations

As I said in my previous comment

> In C and especially C++ a type can occur in virtually any identifier position.

I agree on that, but anyway, I can find examples in nvim codebase (actually the screenshot...) where this way of highlighting type is actually wrong, which, in my opinion is a problem. I can revert that, but I think that we will want a better way of highlighting types.

queries/c/highlights.scm
vigoux: Oh yeah I missed those too. Will do.

queries/c/highlights.scm
theHamsta: Maybe you're using a old parser version of the C parser. I can not achieve any highlight with your query. Can you give an example expression?

Well I am not arguing that you should remove your more sophisticated expression, but as a naming convention can serve as a basic guess (they are a bit more applied for C++ code. So moving this to C++ would be an option)

vigoux: > Maybe you're using a old parser version of the C parser. I can not achieve any highlight with your query. Can you give an example expression? I tried various types of declarations.

That's weird... It worked for me and outputed the screenshot, for example `TSNode *node = foo();` should work.

theHamsta: image

theHamsta: TSInstall C ?

vigoux: > TSInstall C ?

Hehe I am not that dumb ;-)

I might have missed something, in any case, this way should work and is better in my opinion than relying on naming conventions:
```
(declaration type: (_) @type)
(cast_expression type: (_) @type)
```

We have a syntax tree in our hands, relying too much on naming conventions is the job of regex based syntax highlighting.

theHamsta: Maybe the built-in C parser of Nvim is old.

theHamsta: https://medium.com/@mujjingun_23509/full-proof-that-c-grammar-is-undecidable-34e22dd8b664

theHamsta:
```scheme
(declaration type: (_) @type)
(cast_expression type: (_) @type)
```
that's not necessary a type as a whole. In C++ you can have an expression that evaluates to a type or you have `foo::Xer::Type_T` where only the last part is the type. So `(declaration type: (identifier) @type)` would be ok. C++ relies on naming conventions to decide whether the last part of a namespace experssion is a type `foo::Xer::Type_T`

vigoux: But hey, that's C++ not C 😉
Don't mix things up, the goal here is to fix C highlights, I understand that you use it for C++, and I can add that very same line in the C++ highlights query, but actually, C does not need something this general.

vigoux: > Maybe the built-in C parser of Nvim is old.

I think that this was a mistake on my side.

theHamsta: > But hey, that's C++ not C wink
> Don't mix things up, the goal here is to fix C highlights, I understand that you use it for C++, and I can add that very same line in the C++ highlights query, but actually, C does not need something this general.

This is why I am suggesting to move that query to the C++ highlights.

vigoux:
Little note:
```c
TSLanguage *(*lang_parser)(void);
```
This is wrongly parsed as an expression while it is  declaration, I will report an issue on treesitter-c parser.

queries/c/highlights.scm
vigoux: Should be all good for now, I use something that will for sure be fine for C but not for C++, the `binary_expression` thing was actually a parsing error (see the comment below) I will open an issue rn to report that.

theHamsta:
Also simple things like `int foo* = NULL`

queries/cpp/highlights.scm
```diff
@@ -101,3 +101,5 @@
 ] @keyword
 
 "::" @operator
+
+((identifier) @type (#match? @type "^[A-Z]"))
```
theHamsta, line 105: to the top of the file? this query is only a first guess and in some situations we know better.

queries/c/highlights.scm
```diff
@@ -96,12 +110,14 @@
...
 
-((identifier) @type
- (#match? @type "^[A-Z]"))
+[
+(type_identifier)
+(primitive_type)
+(sized_type_specifier)
+ ] @type
+
+(_ type: (_) @type)
```
theHamsta, line 120: This highlights stuff like `(typeof(foo))` wrong. Either be more specific here (identifier pointer ...) or reset the highlighting for things like typeof.

theHamsta: would favor an enumration for things where the highlighting is ok, like `identifier` and `pointer_expression` `type_descriptor`.

vigoux: Looks like `typeof` is considered as a function call by the parser, and thus it is a special case of it.

theHamsta: but the argument of typeof will then be highlighted as a type

vigoux: Heh no they wont, because `typeof` is considered as a function.

theHamsta: Does not matter. `typeof` is highlighted a function. The arguments as type, because (_) sets the hightlight for the whole region.

vigoux: But why? Do you have a screens hot of that?

The match means :
Any node that has a child tagged `type` should highlight this child as a type.
As `typeof`is a function call, it does not have a `type`-tagged child, and thus nothing is highlighted as a type.

theHamsta: image

The tag type means only that it's at the position where the type should go. (_) highlights the whole node as type as type. 

image

vigoux: This should be better now !

PR #154, opened 2020-07-07
fix(refactor): highlight def perf issue fix

steelsojka:
The key api change here was adding the ability to run the locals query on a subset of the tree instead of from the root. This saves us from having to check against all references.


theHamsta:
After a 10 sec test, I can confirm that it's a lot better now :smile: Will have a closer look tomorrow.

PR #156, opened 2020-07-08
Introduce @definition.import for Python/Javascript

theHamsta:
We could introduce a new kind of definition: `import`
Alternatively, I could also declare Pythons imports as `@definition.var`

Problem: `import skimage.io` defines `skimage` and `io` where it should only define `skimage.io`. 

Edit: maybe this can be solved by a new query for reference.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,7 +131,8 @@ are optional and will not have any effect for now.
   macro
   type
   field
-  doc for documentation adjecent to a definition. E.g.
+  import for imported names
+  doc for documentation adjacent to a definition. E.g.
 ```
 
 ```scheme
--- a/queries/javascript/locals.scm
+++ b/queries/javascript/locals.scm
@@ -29,7 +29,7 @@
   name: (identifier) @definition)
 
 (import_specifier
-  (identifier) @definition)
+  (identifier) @definition.import)
 
 ; References
 ;------------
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -131,6 +131,7 @@ are optional and will not have any effect for now.
   macro
   type
   field
+  import -- for imported names
   doc for documentation adjecent to a definition. E.g.
```
vigoux, line 135: Can we fix this line too ?

theHamsta: like removing `--`?

vigoux: That damn GH diffs...
```suggestion
  doc for documentation adjacent to a definition. E.g.
```

CONTRIBUTING.md
```diff
@@ -131,8 +131,8 @@ are optional and will not have any effect for now.
   macro
   type
   field
-  doc for documentation adjecent to a definition. E.g.
-```
+  import for imported names
+  doc for documentation adjacent to a definition. E.g.
```
vigoux, line 135:
```suggestion
  doc for documentation adjacent to a definition. E.g.
  ```
```

theHamsta: A this was in ```

theHamsta: Sorry.

PR #161, opened 2020-07-10
Post user textobject refactor

theHamsta:
Currently based on #100 

Clean up some stuff after #100 to separate adding a new feature from a refactoring API change

- [x] add `parsers.get_buf_lang` (every module uses a similar statement to get the language of a buffer)
- [x] refactor locals module
     - [x] remove stuff that is shared between all query groups (now locals/textobjects) -> query.lua
     - [x] add stuff from ts_utils that is strongly related to locals (get_definition, scopes, etc.)

### DIFF
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -1,25 +1,21 @@
 local api = vim.api
 local ts = vim.treesitter
+local utils = require'nvim-treesitter.utils'
+local parsers = require'nvim-treesitter.parsers'
 
 local M = {}
 
-local function read_query_files(filenames)
-  local contents = {}
-
-  for _,filename in ipairs(filenames) do
-    vim.list_extend(contents, vim.fn.readfile(filename))
+local default_dict = {
+  __index = function(table, key)
+    local exists = rawget(table, key)
+    if not exists then
+      table[key] = {}
+    end
+    return rawget(table, key)
   end
+}
 
-  return table.concat(contents, '\n')
-end
-
--- Creates a function that checks whether a certain query exists
--- for a specific language.
-local function get_query_guard(query)
-  return function(lang)
-    return M.get_query(lang, query) ~= nil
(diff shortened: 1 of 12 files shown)


theHamsta:
@steelsojka some additional changes were necessary after your review after rebasing. Some API changes could break your doc plugin.

theHamsta:
Please check carefully, if all modules are still working (maybe test a day with this). Refactoring is a bit risky in Lua.

lua/nvim-treesitter/query.lua
```diff
@@ -1,25 +1,22 @@
...
-
--- Creates a function that checks whether a certain query exists
--- for a specific language.
-local function get_query_guard(query)
-  return function(lang)
-    return M.get_query(lang, query) ~= nil
-  end
-end
+local query_cache = {}
+setmetatable(query_cache, default_dict)
```
vigoux, line 19: Isn't that more :
```suggestion
local query_cache = setmetatable({}, default_dict)
```

lua/nvim-treesitter/query.lua
```diff
@@ -34,10 +31,42 @@ M.query_extensions = {
...
+
+  for _,filename in ipairs(filenames) do
+    vim.list_extend(contents, vim.fn.readfile(filename))
+  end
+
+  return table.concat(contents, '\n')
+end
+
+local function update_cached_matches(bufnr, changed_tick, query_group)
+  query_cache[query_group][bufnr] = {tick=changed_tick, cache=( M.collect_group_results(bufnr, query_group) or {} )}
```
vigoux, line 57: My worry is that a given buffer can have multiple queries of the same kind (for injection purposes mainly), and thus this we not be viable in the long term.

theHamsta: Then we could just an additional key for the parser range or just do bit more work on re-collecting the query_group for all parsers.

PR #57, opened 2020-05-17
Add Rust queries

vigoux:
Also slightly changes the way constructors are highlighted to better match their highlighting by default.

This will partially fix #52. 

TODO:
- [x] `highlights.scm`
- [ ] `locals.scm`
- [ ] Update `README`

PR #162, opened 2020-07-10
Add rust locals.scm

theHamsta:
- [x] definitions (started)
- [x] scopes

I will try to use the existing work in the discussion #57

### DIFF
--- a/queries/cpp/locals.scm
+++ b/queries/cpp/locals.scm
@@ -23,6 +23,13 @@
 (alias_declaration
   name: (type_identifier) @definition.type)
 
+;; Namespaces
+(namespace_definition 
+  name: (identifier) @definition.namespace
+  body (_) @scope)
+
+(namespace_identifier) @reference.namespace
+
 ;; Function defintions
 (template_function
   name: (identifier) @definition.function) @scope
--- a/queries/rust/locals.scm
+++ b/queries/rust/locals.scm
@@ -1,0 +1,102 @@
+; Imports
+(extern_crate_declaration
+    name: (identifier) @definition.import)
+
+(use_declaration 
+  argument: (scoped_identifier 
+              name: (identifier) @definition.import))  
+
+(use_as_clause 
+  alias: (identifier) @definition.import)
+
+(use_list
+    (identifier) @definition.import) ; use std::process::{Child, Command, Stdio};
+
+; Functions
+(function_item 
+    name: (identifier) @definition.function)
+
+(function_item 
+  name: (identifier) @definition.method
+  parameters: (parameters 
+                (self_parameter)))
+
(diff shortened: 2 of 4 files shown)


theHamsta:
Probably still incomplete, will play around with it another day.

queries/rust/locals.scm
```diff
@@ -0,0 +1,100 @@
...
+(field_declaration
+  name: (field_identifier) @definition.field)
+
+(enum_variant
+  name: (identifier) @definition.field) 
+
+; References
+(identifier) @reference
+(type_identifier) @reference
+(field_identifier) @reference
```
theHamsta, line 74: Same issue as in #164. I think we should also consider references to types and fields.

vigoux: I agree on this in the end

queries/cpp/locals.scm
```diff
@@ -20,6 +20,13 @@
   name: (scoped_type_identifier
           name: (type_identifier) @definition.type))
 
+;; Namespaces
+(namespace_definition 
+  name: (identifier) @definition.namespace
+  body (_) @scope)
+
+(namespace_identifier) @reference
+
```
vigoux, line 29: Same for this

queries/cpp/locals.scm
```diff
@@ -49,3 +56,4 @@
 
 (destructor_name
   name: (_) @constructor)
+
```
vigoux, line 59: Same

queries/rust/locals.scm
```diff
@@ -0,0 +1,100 @@
...
+(for_expression) @scope
+(loop_expression) @scope
+(if_expression) @scope
+(if_let_expression) @scope
+(match_expression) @scope
+(match_arm) @scope
+
+(struct_item) @scope
+(enum_item) @scope
+(impl_item) @scope
```
vigoux, line 100: Can we refactor that in  a `[ ]` block ?

queries/rust/locals.scm
```diff
@@ -0,0 +1,100 @@
...
+(field_identifier) @reference
+
+
+; Macros
+(macro_definition
+  name: (identifier) @definition.macro)
+
+; Module
+(mod_item 
+  name: (identifier) @definition.namespace)
```
vigoux, line 83: Now I see why you added those namespaces :wink:

queries/rust/locals.scm
theHamsta: Should it be `@reference.type`?

vigoux: Yes, I like that more !

vigoux:
Awesome ! That is one step further in #52  !

PR #164, opened 2020-07-10
C/C++ highlights: Small improvements for type definitions

theHamsta:
Definition of types by:
 - using (C++)
 - plain struct

Reference of types

### DIFF
--- a/queries/c/locals.scm
+++ b/queries/c/locals.scm
@@ -26,9 +26,12 @@
   declarator: (field_identifier) @definition.field)
 (type_definition
   declarator: (type_identifier) @definition.type)
+(struct_specifier
+  name: (type_identifier) @definition.type)
 
 ;; References
 (identifier) @reference
+(type_identifier) @reference
 
 ;; Scope
 [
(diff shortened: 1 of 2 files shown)


queries/c/locals.scm
```diff
@@ -26,9 +26,12 @@
   declarator: (field_identifier) @definition.field)
 (type_definition
   declarator: (type_identifier) @definition.type)
+(struct_specifier
+  name: (type_identifier) @definition.type)
 
 ;; References
 (identifier) @reference
+(type_identifier) @reference
```
vigoux, line 34: I don't think that type_identifiers are actually references.

theHamsta: What is a reference for you? It references a type identifier when we can have type definitions we should also have type references. Or maybe add reference.type?

theHamsta: I'll have a look again at which positions that node type can occur

theHamsta: `type_reference` is the `Foo` in `Foo a = create_foo();` So references to type names.

PR #171, opened 2020-07-12
Highlight current scope

theHamsta:
Highlights current scope. Should this be a built-in module or should I publish it as separate plugin? Then, I would just try to merge the first commit of this PR that adds a `highlight_node` utility function.

### DIFF
--- a/lua/nvim-treesitter/ts_utils.lua
+++ b/lua/nvim-treesitter/ts_utils.lua
@@ -124,8 +124,8 @@ function M.parent_scope(node, cursor_pos)
   end
 end
 
-function M.containing_scope(node)
-  local bufnr = api.nvim_get_current_buf()
+function M.containing_scope(node, bufnr)
+  local bufnr = bufnr or api.nvim_get_current_buf()
 
   local scopes = locals.get_scopes(bufnr)
   if not node or not scopes then return end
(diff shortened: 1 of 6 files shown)


theHamsta:
Would be probably more useful to highlight the scope at the sign-column.

theHamsta:
highlight_scope

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -313,4 +313,17 @@ function M.find_usages(node, scope_node, bufnr)
   return usages
 end
 
+function M.highlight_node(node, buf, hl_namespace, hl_group)
+  if not node then return end
+  M.highlight_range({node:range()}, buf, hl_namespace, hl_group)
+end
+
+function M.highlight_range(range, buf, hl_namespace, hl_group)
```
theHamsta, line 321: Should probably use *vim.highlight.range()*

theHamsta:
It's also possible to do an inverse highlighting (dim everything but the current scope)

inverse_highlight

but this conflicts with treesitter highlighting.

vigoux:
Ouch, if that's not a killer feature, I don't know what is ! Do you have some docs on how to have the current scope within the sign column too ?

theHamsta:
Sign column is not implemented. I don't know how to do it without setting signs.

PR #177, opened 2020-07-14
checkhealth for all query groups

theHamsta:


theHamsta:
Or defer this? Right now everything is really green :smiley:

lua/nvim-treesitter/query.lua
```diff
@@ -37,6 +37,7 @@ M.query_extensions = {
 M.has_locals = get_query_guard('locals')
 M.has_textobjects = get_query_guard('textobjects')
 M.has_highlights = get_query_guard('highlights')
```
vigoux, line 39: Can we refactor that too using `M.built_in_query_groups` too then ? Something like : 
```suggestion
for _, query in ipairs(M.built_in_query_groups) do
  M["has_" .. query] = get_query_guard(query)
end
```

PR #186, opened 2020-07-15
Use a build matrix to check queries

vigoux:
This will allow use to produce parsers artifacts for macos and linux at least for now.

### DIFF
--- a/.github/workflows/check_query_files.yml
+++ b/.github/workflows/check_query_files.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [unbuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest]
     steps:
       - uses: actions/checkout@v1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/check_query_files.yml
```diff
@@ -5,7 +5,10 @@ on: [push, pull_request]
 jobs:
   luacheck:
     name: Check Query Files
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [unbuntu-latest, macos-latest]
```
vigoux, line 11:
```suggestion
        os: [ubuntu-latest, macos-latest]
```

theHamsta:
Shouldn't this be a separate testing workflow?

I tried to separate Lua style checking from the syntax of the query files. The query files (typo in queries) should be unrelated to OS but further testing should be done for multiple OSes.

vigoux:
Then I think that we need a workflow to just compile the parsers on every platform maybe every day or so.

theHamsta:
I mean having a workflow for tests on every platform (automatically compiles the parsers) and one workflow for the queries. But one big workflow that tests everything would also be ok.

vigoux:
Yeah I see what you mean, but I will to do that in another PR, this one is a mess.

PR #188, opened 2020-07-15
feat(textobjects): Add `start` to include preceding things like documentation

theHamsta:
Idea shamelessly stolen from: https://github.com/nvim-treesitter/nvim-tree-docs

This allows to include starting positions for textobjects like:
```scheme
(template_declaration
  (function_definition
    body:  (compound_statement) @function.inner) @function.outer) @function.outer.start
```
when we have
```cpp
template <typename T>
T* givePtr( T& t )
{
    return &t;
}
```
also the following c query will match
```scheme
(function_definition
    body:  (compound_statement) @function.inner) @function.outer
```
both matches have the same length, though refer to the same thing.
We prefer the query that has earliest start (if any).
Users can use this to attach documentation to text objects.

### DIFF
--- a/lua/nvim-treesitter/locals.lua
+++ b/lua/nvim-treesitter/locals.lua
@@ -126,7 +126,7 @@ function M.get_capture_matches(bufnr, capture_string, query_kind)
 
     local matches = {}
     for _, match in pairs(M.get_locals(bufnr, query_kind)) do
-      local insert = utils.get_at_path(match, capture_string..'.node')
+      local insert = utils.get_at_path(match, capture_string)
 
       if insert then
         table.insert(matches, insert)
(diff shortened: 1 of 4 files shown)


lua/nvim-treesitter/locals.lua
```diff
@@ -126,7 +126,7 @@ function M.get_capture_matches(bufnr, capture_string, query_kind)
 
     local matches = {}
     for _, match in pairs(M.get_locals(bufnr, query_kind)) do
-      local insert = utils.get_at_path(match, capture_string..'.node')
+      local insert = utils.get_at_path(match, capture_string)
```
vigoux, line 129: Why is this needed ?

theHamsta: I need the whole capture without node. It also contains my starts. I had this also changed for query things like `strip!`
A @function.outer match would look like that
```lua 
{
    node = <userdata>,
    start = {
              node = <userdata>
     }
}

PR #189, opened 2020-07-15
Python textobjects: Include @decorators and improve @conditional.inner

theHamsta:
Include decorators to functions and improve @conditional.inner for conditions of elif statements


theHamsta:
The only thing that sucks when copying and pasting in Python is the indentation. At some point, we probably need some whitespace control for textobjects.

PR #193, opened 2020-07-16
Add jsdoc parser

theHamsta:
I did not register as a maintainer since it's probably only interesting for JS (@steelsojka) and after implementation of language injection.


steelsojka:
@theHamsta I'll be the maintainer for this. No problem.

PR #195, opened 2020-07-16
ruby: fix references and definitions in locals

TravonteD:
The highlighting when highlighting references and definitions in ruby
were off. This corrects that by assigning them to their correct properties.

### DIFF
--- a/queries/ruby/locals.scm
+++ b/queries/ruby/locals.scm
@@ -20,17 +20,26 @@
 ; OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 ; SOFTWARE.
 
-;;; DESCLARATIONS AND SCOPES
+;;; DECLARATIONS AND SCOPES
 ((method) @scope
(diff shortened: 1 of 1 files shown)


queries/ruby/locals.scm
```diff
@@ -20,17 +20,26 @@
...
  ] @scope
 
-(method_parameters (identifier) @definition.function)
-(lambda_parameters (identifier) @definition.function)
+(identifier) @reference
+(constant) @reference
+(instance_variable) @reference
+
+(module name: (constant) @definition.namespace)
+(class name: (constant) @definition.namespace)
```
theHamsta, line 38: @definition.type ?

TravonteD: Is that what we are using for classes? If so than yes, in Ruby they have little difference to modules.

theHamsta: All other languages use @defintion.type for classes

TravonteD: Noted, I've updated the query.

PR #201, opened 2020-07-18
Python highlights: highlight functions with self as methods

theHamsta:
Also highlight __init__/__new__ as constructor

### DIFF
--- a/queries/python/highlights.scm
+++ b/queries/python/highlights.scm
@@ -48,6 +48,19 @@
 (function_definition
   name: (identifier) @function)
 
+((function_definition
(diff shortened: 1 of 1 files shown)


theHamsta:
Interesting bug: For neovim `(eq? @self "self")` `(match? @self "self")` are not the same. But `(match? @self "^self$")` behaves as expected.

vigoux:
> Interesting bug: For neovim `(eq? @self "self")` `(match? @self "self")` are not the same. But `(match? @self "^self$")` behaves as expected.

I think that this is a feature.

PR #202, opened 2020-07-18
Fixed to use normal! instead of normal when selecting text objects

svermeulen:


vigoux:
Hi ! Thanks for your contribution and nice catch !

PR #203, opened 2020-07-18
Added first attempt at highlighting and text objects for csharp

svermeulen:


queries/c_sharp/highlights.scm
```diff
@@ -0,0 +1,137 @@
...
+ "public"
+ "readonly"
+ "sealed"
+ "static"
+ "virtual"
+ "volatile"
+ "async"
+ "await"
+] @StorageClass
+
```
theHamsta, line 96: StorageClass?

svermeulen: Oh is that not a valid syntax type?  Ok I've changed this to just `keyword` which seems to be more in line with what the other files are doing

theHamsta:
You could add a highlight for `this`

queries/c_sharp/highlights.scm
```diff
@@ -0,0 +1,132 @@
...
+ ":"
+] @punctuation.delimiter
+
+[
+ "["
+ "]"
+ "{"
+ "}"
+ "("
+ ")"
```
theHamsta, line 93: "<" ">"

queries/c_sharp/highlights.scm
```diff
@@ -0,0 +1,132 @@
...
+ "finally"
+] @exception
+
+[
+ ";"
+ "."
+ ","
+ ":"
+] @punctuation.delimiter
+
```
theHamsta, line 86: could you add all the operator stuff? You should be able to copy them from languages like C or Java.

queries/c_sharp/highlights.scm
```diff
@@ -0,0 +1,132 @@
...
+ "["
+ "]"
+ "{"
+ "}"
+ "("
+ ")"
+] @punctuation.bracket
+
+[
+ "using"
```
theHamsta, line 97: You could use @include for keywords related to importing packages.

queries/c_sharp/highlights.scm
```diff
@@ -0,0 +1,132 @@
...
+ "virtual"
+ "volatile"
+ "async"
+ "await"
+ "class"
+ "delegate"
+ "enum"
+ "interface"
+ "namespace"
+ "struct"
```
theHamsta, line 130: Could it be that "in" (foreach loop) is missing (maybe also in upstream)?

svermeulen:
Ok I've addressed all your feedback except for the following:

> You could highlight calls to functions. They are usally the "name" of a member_access_expression in a invocation_expression:
> 
> ```
>  invocation_expression [45, 15] - [45, 70])
>               member_access_expression [45, 15] - [45, 28])
>                 expression: identifier [45, 15] - [45, 21])
>                 name: identifier [45, 22] - [45, 28])
> ```

I played around with this but could not find a way to highlight methods calls without also highlighting all identifiers (eg. a local variable).  Also, what would be the highlight group for method calls?

> 
> You could also highlight the parameters of Lambda expressions.

In my tests this appears to already be covered by the current `@parameter` lines

theHamsta:
This seems to be working fine for me:
```scheme
(invocation_expression
              (member_access_expression
                                name: (identifier) @method))
```

Could you add yourself as a maintainer for C# (in case you want to be that...)?

svermeulen:
Ok, I added myself as a c# maintainer, and added those lines you suggested (which do seem to work for me)

theHamsta:
Try out the following file in online playground with my queries:
```csharp
using System;
using System.Linq;
using System.Collections.Generic;

public class Anagram
{
    private string baseWord;

    public Anagram(string baseWord)
    {
        this.baseWord = baseWord;
    }

    public string[] FindAnagrams(string[] potentialMatches)
    {
        List<string> matches = new List<string>();

        foreach (string word in potentialMatches)
        {
            if (IsWordAnagramOfBaseWord(word))
            {
                matches.Add(word);
            }
        }

        return matches.ToArray();
    }

    private bool IsWordAnagramOfBaseWord(string word)
    {
        return (IsNotTheSameWordAsBaseWord(word) && HasSameLettersAsBaseWord(word));
    }

    private bool IsNotTheSameWordAsBaseWord(string word)
    {
        return !baseWord.Equals(word,StringComparison.OrdinalIgnoreCase);
    }

    private bool HasSameLettersAsBaseWord(string word)
    {
        return SortedCharArrayForString(baseWord).Equals(SortedCharArrayForString(word));
    }

    private string SortedCharArrayForString (string word)
    {
        return String.Concat(word.ToLower().OrderBy(letter => letter));
    }
}
```
Problematic is `!` in front of the call expression and the `&&` between. The parsed syntax tree look very stange.

queries/c_sharp/highlights.scm
```diff
@@ -0,0 +1,206 @@
...
+((identifier) @field
+ (#match? @field "^m_"))
+
+(parameter_list
+  (parameter
+   name: (identifier) @parameter))
+
+[
+ (integer_literal)
+ (real_literal)
```
theHamsta, line 24: there is also @float

svermeulen:
Ok yeah true, I see a lot of highlight errors in that file.  I just commented out the attempt to highlight method calls for now

svermeulen:
Just want to point out that after using this for awhile, I regularly get tree sitter errors.  So I don't think this is ready yet.  Not sure if the errors are due to this change, this plugin, the c# parser file, neovim, tree sitter, etc. but here's what they look like.  The errors actually over-write the text of the file somehow.  If I undo and redo it goes away though.

image

svermeulen:
Actually those same errors occur when using other python treesitter highlight as well, so seem independent of this c# change

svermeulen:
Current theory is that the above errors are specific to my setup, which involved a custom status line that displayed treesitter data at the current cursor.  After disabling this part of my status line, I have not seen the errors, so I think I can confirm it's unrelated to this change

theHamsta:
@svermeulen I think this is a really good start. Most of the refinement can only be done after using the highlights for a while. Thank you!

PR #204, opened 2020-07-19
java locals.scm

PitcherTear22:


queries/java/locals.scm
```diff
@@ -1,10 +1,36 @@
 ; CREDITS @maxbrunsfeld (maxbrunsfeld@gmail.com)
 
-(class_declaration
-  name: (identifier) @name) @class
+;(class_declaration
+;  name: (identifier) @name) @class
```
theHamsta, line 4: @defintiion.class?

queries/java/locals.scm
```diff
@@ -1,10 +1,36 @@
...
+;(method_declaration
+;  name: (identifier) @name) @method
+;
+;(method_invocation
+;  name: (identifier) @name) @call
 
-(method_declaration
-  name: (identifier) @name) @method
+; SCOPES
+(class_declaration) @scope
```
theHamsta, line 13: I liked your : `(class_declaration: body (_)) @scope` better

queries/java/locals.scm
PitcherTear22: ok

theHamsta:
Would also be nice if you could add @reference and @definition.var

queries/java/locals.scm
PitcherTear22: that part is from tree-sitter/tree-sitter-java

PitcherTear22: commented out

PitcherTear22:
what is "@reference"

PitcherTear22:
nvm got it

PitcherTear22:
It is recognised as var, not reference
image

theHamsta:
> what is @reference

any usage of an identifier. I how we find usages of a definition.

Thanks for your contribution! You contributed everything for Java. Do you want to register yourself in the README.md as the maintainer for Java?

PitcherTear22:
How do you differentiate? both definition and usage are labelled `identifier`

PitcherTear22:
I think declaration being at the top in all `locals.scm` does that? Definition is definition, everything else is reference

PitcherTear22:
other languages have `(field_identifier) @reference`, but for java it doesnt seem to differentiate:
image

PitcherTear22:
> Do you want to register yourself in the README.md as the maintainer for Java?

Do it, my fork is old so I'll only use it for java stuff, so can't include this in the PR

queries/java/locals.scm
```diff
@@ -1,10 +1,34 @@
...
+(local_variable_declaration) @definition.var
+(field_declaration) @definition.field
+(import_declaration) @definition.import
 
-(method_invocation
-  name: (identifier) @name) @call
+;; REFERENCE
+(identifier) @reference
+(type_identifier) @reference
+;(field_identifier) @reference
```
vigoux, line 34: This node type does not exist, could you remove this ?

queries/java/locals.scm
```diff
@@ -1,10 +1,37 @@
-; CREDITS @maxbrunsfeld (maxbrunsfeld@gmail.com)
-
+; SCOPES
 (class_declaration
-  name: (identifier) @name) @class
+   body: (_)) @scope
```
vigoux, line 3: Maybe you meant ?
```suggestion
   body: (_) @scope)
```

queries/java/locals.scm
```diff
@@ -1,10 +1,37 @@
...
+(method_declaration) @scope
+(block) @scope
+; 3 captures for `if` - if+else, if only, else only
+(if_statement) @scope
+(if_statement
+  body: (_)) @scope
```
vigoux, line 9: Same here 
```suggestion
(if_statement
  body: (_) @scope)
```

queries/java/locals.scm
```diff
@@ -1,10 +1,37 @@
...
+(if_statement
+  alternative: (_)) @scope
```
vigoux, line 11: Same here
```suggestion
(if_statement
  alternative: (_) @scope)
```

queries/java/locals.scm
```diff
@@ -1,10 +1,37 @@
...
+; 3 captures for`try` - try+catch, try only, catch only
+(try_statement) @scope
+(try_statement
+  body: (_) @scope)
```
vigoux, line 15: Why having it twice ?

queries/java/locals.scm
```diff
@@ -1,10 +1,37 @@
...
+(catch_clause) @scope
+
+(for_statement) @scope
+(do_statement) @scope
```
vigoux, line 19: These dont have a `body` field ?

queries/java/locals.scm
PitcherTear22: `(try_statement) @scope` captures try+catch
`(try_statement
  body: (_) @scope)`  captures body of try

queries/java/locals.scm
PitcherTear22: nope i want to capture the whole thing. i tagged you in the relevant message on gitter.

queries/java/locals.scm
vigoux: Not that pleased by that then, why not capturing the `try` on one side and `catch` on the other side ?

PitcherTear22: wait

PitcherTear22: yanking the whole exception handling would be nice

so try+catch together

theHamsta: But all the body queries are unnecessary. Just use `(block) @scope` and you are done (there's only a few exceptions, e.g. the curly braces of a class are not `block`). You only need to capture the whole thing if it is allowed to introduce own variables like `for` or `catch`.

PitcherTear22: https://imgur.com/z6pjf8Y

PitcherTear22: how about
```scheme
(if_statement) @scope ; deleted "if body" and "else", "if" is needed in additon to "block" for if+else
(try_statement) @scope ; deleted "try body", "try" is needed in additon to "block" for try+catch
(catch_clause) @scope ; "catch" is needed in additon to "block" because it has its own Exception variable

(for_statement) @scope
(while_statement) @scope; deleted "do" but "catch" is needed in additon to "block" because it has its loop variable
```

PitcherTear22: https://imgur.com/z6pjf8Y

PitcherTear22: wait `do` is also needed because it has loop variable in
```java
do{

} while(i-->0);
```

vigoux: That's not a declaration ?

theHamsta: What about `do int i; while(true);`? Is this allowed?

PitcherTear22: > What about `do int i; while(true);`? Is this allowed?

yes

PitcherTear22: in this case `(do_statement)` covers the whole thing,
```scheme
(do_statement
  body: (_) @scope)
```
covers the `int i`
image

PitcherTear22: which one should i keep?

queries/java/locals.scm
```diff
@@ -1,26 +1,34 @@
...
+  consequence: (_) @scope) ; if body in case there are no braces
 (if_statement
-  alternative: (_) @scope)
-(try_statement) @scope
-(catch_clause) @scope
-(do_statement) @scope
+  alternative: (_) @scope) ; else body in case there are no braces
+
+; try/catch
+(try_statement) @scope ; covers try+catch, individual try and catch are covered by (block)
```
vigoux, line 20: Why is the whole `try_statement` necessary if individual blocks are covered ?

PitcherTear22: to grab the whole exception handling logic - [this](https://imgur.com/z6pjf8Y) would be nice to have

PitcherTear22:
Closing in favour of #219

PR #208, opened 2020-07-20
Predicates

theHamsta:


theHamsta:
reminder to me: use cases c# imports and lambdas / Java / Python imports

theHamsta:
You can now just drop functions in the predicates module, e.g. `foo_bar` and it will enable the predicates `foo-bar?` and `not-foo-bar?`.

Question: should I remove `lispify` and use the lisp names in the module, e.g. `M['has-ancestor']`?

A test CI could be useful for stuff like this.

The Java query for the import can now be implemented as:
```scheme
((scoped_identifier
  (identifier) @definition.import)
 (has-ancestor? @definition.import import_declaration))
```
The `has-ancestor?` predicate is less powerful than a possible `submatch?`. However, it is simple to implement, fast to check and does not require any changes to the syntax of the predicates.

lua/nvim-treesitter/query_predicates.lua
```diff
@@ -0,0 +1,79 @@
+local utils = require'nvim-treesitter.utils'
+
+local M = {}
+
+local function get_nth_child(node, n)
+  if node:named_child_count() >= n then
+    return node:named_child(n)
+  end
+end
```
vigoux, line 9: I think that child indexing is `0` based.

theHamsta:
@vigoux I guess upstream will just provide us a way to register predicates. We should still have a file with our custom predicates.

PR #209, opened 2020-07-21
add rest_parameters as definition in javascript

steelsojka:


kyazdani42:
i'm not too sure but aren't rest parameters `@parameters` ?

vigoux:
Welcome back @kyazdani42 !

steelsojka:
@kyazdani42 Yes, but it also a definition like other parameters.

steelsojka:
I should also add them as a parameter for highlighting though...

kyazdani42:
go for it

queries/javascript/locals.scm
```diff
@@ -25,6 +25,10 @@
   (array_pattern
     (identifier) @definition))
 
+(formal_parameters
+  (rest_parameter
+    (identifier) @definition))
```
theHamsta, line 30: @definition.var ? Also for other variable definitions.

queries/javascript/highlights.scm
```diff
@@ -74,6 +74,10 @@
 
 (formal_parameters (identifier) @variable.parameter)
 
+(formal_parameters
+  (rest_parameter
+    (identifier) @variable.parameter))
```
theHamsta, line 79: @parameter

Issue #198, opened 2020-07-17
hl_map['method'] ignores color


fdschmidt93:
**Describe the bug**
When defining a highlight scheme for the method the hlmap["method"] group ignores the color being set and seems to revert to the global default.

![tree-sitter_issue](https://user-images.githubusercontent.com/39233597/87828905-478bcc00-c87e-11ea-8339-778dd8220c81.png)
On the left-hand side, I have set the color bold_red for hlmap["method"]. However, as seen for torch.arange on the right-hand side only the bold font is being applied. I played a lot with the highlighting groups (still have not found how I quite like it perfectly :) and everything else seemed to have worked fine.

**To Reproduce**
Set hlmap["method"] to anything else but the standard highlighting group. For instance, see "[4.8] nvim-treesitter in my [init.vim](https://github.com/fdschmidt93/dotfiles/blob/master/nvim/.config/nvim/init.vim). I define highlighting groups for my colorscheme gruvbox further below manually to have more control and hopefully facilitates to reproduce the issue.

**Expected behavior**
Color highlighting for method group as set in config.

**Output of :checkhealth nvim_treesitter**
health -- nvim_treesitter -- check
-- Installation
  - OK: `git` executable found.
  - OK: `cc` executable found.

-- python parser 
  - OK: python parser found.
  - OK: `highlights.scm` found.
  - OK: `locals.scm` found.
  - OK: `textobjects.scm` found.

**Output of `nvim --version`**
NVIM v0.5.0-594-ga02a267f8
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
[...]

PR #211, opened 2020-07-22
fixes golang method highlights

WinWisely268:
Method highlights is overwritten by property because of the ordering.

something like:

Before:

screenshot_2020-07-22-085154

After:

screenshot_2020-07-22-085116

Possibly connected to #198 but on golang instead of python


theHamsta:
Your PR also fixes the problem on my side! Nice work!

PR #215, opened 2020-07-22
Parsers: Add parser for dart

theHamsta:


theHamsta:
I added some basic highlights and parsed a few files with tree-sitter CLI. The parser seems to work fairly well

theHamsta:
TODO: upstream repo has no license. Can we use the highlighting queries? @UserNobody14

queries/dart/highlights.scm
```diff
@@ -0,0 +1,104 @@
...
+    name: (identifier) @parameter)
+
+;; Literals
+
+(hex_integer_literal) @number
+(decimal_integer_literal) @number
+(decimal_floating_point_literal) @float
+;; Those node do not seem to exist
+;(octal_integer_literal) @number
+;(hex_floating_point_literal) @float
```
vigoux, line 68: Then we can remove this completely

queries/dart/highlights.scm
```diff
@@ -0,0 +1,104 @@
+(dotted_identifier_list) @string
+
+; Methods
+
+(function_type
+  name: (identifier) @method)
+(super) @function
+
+; Annotations
+; TODO: highlight??
```
vigoux, line 10: Are those todos from upstream ?

theHamsta: no from me

theHamsta: @Akin909 seems to have interest in fine-tuning those highlights.

vigoux: Will be done in another PR I guess, let's merge that quickly to allow him to enhance Dart and maintain it.

queries/dart/highlights.scm
```diff
@@ -0,0 +1,104 @@
...
+"final" @keyword
+"implements" @keyword
+"is" @keyword
+"as" @keyword
+"mixin" @keyword
+"external" @keyword
+"new" @keyword
+"return" @keyword
+"static" @keyword
+; "rethrow" @keyword
```
Akin909, line 98: This is a valid keyword much like the ones above not sure if it's causing some kind of error but just FYI

queries/dart/highlights.scm
theHamsta: But they should exist as something different... If it's possible to write octals they should appear as something different (`08` at least is parsed a decimal)

queries/dart/highlights.scm
Akin909: I will try and have a look when I get a chance although I'm pretty unfamiliar with tree sitter syntax so any link to similar previous PRs or areas of the code base would be great other wise will just try stuff and see what I get 😆

queries/dart/highlights.scm
theHamsta: as well as "async", "var" "await" they need to be captured as something different. Would be happy if someone could take over who is actually using dart.

queries/dart/highlights.scm
Akin909: I'd primarily just be starting with seeing if I can get methods and classes picked up and highlighted

queries/dart/highlights.scm
Akin909: @theHamsta is it worth merging what you have so far (obviously up to you and the other maintainers) and I can make a new branch later to add some more stuff so the surface area of my PR is smaller. I think what this PR has is already going to be really useful

queries/dart/highlights.scm
theHamsta: confirmed. they do not exist in `grammar.js`.

queries/dart/highlights.scm
theHamsta: I would suggest to checkout the `tree-sitter-dart` repo and do a `npm install`. You can the do `tree-sitter parse <filename>` to see a parsed tree that you can correlate with the source file or follow the instructions how to run the tree-sitter web-ui locally.

In CONTRIBUTING.txt you should find an overview on the highlight categories that we have currently (of course @attribute is missing and could be added or replaced by something else). For experimenting with the syntax the tree-sitter web-ui is best: https://tree-sitter.github.io/tree-sitter/playground (but dart is missing. So you could build it locally: `tree-sitter build-wasm`

theHamsta:
@Akin909 has made some improvements to the highlights. Should we just merge or have a new PR that supersedes this one?

vigoux:
Let's merge this to let @Akin909 what he wants to do.

PR #217, opened 2020-07-23
Improve dart highlights

Akin909:
This PR adds a few more highlights for dart files, most notably tweaking the regex to match more classes, adding operators and punctuation.

I'd at some later date like to see if the upstream grammar can provide nodes/groups (is that the terminology) for methods calls and class instantiation properly but with these changes the highlighting is much improved (at least when working in flutter, it's already pretty good for a standard dart file)

### Before
dart-treesitter

### After
treesitter-regex-dart

### DIFF
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -28,7 +28,7 @@ hlmap["character"] = "TSCharacter"
 hlmap["number"] = "TSNumber"
 hlmap["boolean"] = "TSBoolean"
 hlmap["float"] = "TSFloat"
-hlmap["attribute"] = "TSAttribute"
+hlmap["annotation"] = "TSAnnotation"
 
 -- Functions
 hlmap["function"] = "TSFunction"
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -28,6 +28,7 @@ hlmap["character"] = "TSCharacter"
 hlmap["number"] = "TSNumber"
 hlmap["boolean"] = "TSBoolean"
 hlmap["float"] = "TSFloat"
+hlmap["attribute"] = "TSAttribute"
 
 -- Functions
 hlmap["function"] = "TSFunction"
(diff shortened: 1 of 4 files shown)


Akin909:
This is definitely not everything just a small subset of things to start

vigoux:
Hi and thank you for contributing ! As you as going to contribute quite a lot `dart` support, can you add you as the `dart` maintainer in the `README` ? Ans also feel free to join : [Gitter](https://gitter.im/nvim-treesitter/develop?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge) !

Akin909:
👍 @vigoux I'll happy to look after dart support. I'll tweak the `README`

Akin909:
@theHamsta I get a `Bad node name` error when I try and use `attribute` in the treesitter playground for dart. Did you mean the labelled arguments? (If you did I was thinking about tackling that later as I'm not sure how to write that query atm and won't have time till next week to try)

theHamsta:
I was referring to the `@attribute` capture:
```scheme
; Annotations
; TODO: highlight??
(annotation
  name: (identifier) @attribute)
(marker_annotation
  name: (identifier) @attribute)
```
they will have no effect as long as there is no "attribute" connected with a hlgroup here:

https://github.com/theHamsta/nvim-treesitter/blob/24a81902d3f75ca9f549048a77e6d27a8555fabe/lua/nvim-treesitter/highlight.lua#L15-L53

lua/nvim-treesitter/highlight.lua
```diff
@@ -28,6 +28,7 @@ hlmap["character"] = "TSCharacter"
 hlmap["number"] = "TSNumber"
 hlmap["boolean"] = "TSBoolean"
 hlmap["float"] = "TSFloat"
+hlmap["attribute"] = "TSAttribute"
```
vigoux, line 31: Hmm.. We already have `field` and `property` what would be the difference with them ?

theHamsta: I guess this kind of attribute is rather a code annotation. Like C++ attributes `[[maybe_unsed]]` or go annotation that tell you how specified field should be serialized. Maybe change to "@annotation"?

vigoux: I rather like `@annotation` as they will be used in java and rust too then.

Akin909:
Renamed `attribute` to `annotation` @theHamsta, happy for me to merge this?

PR #218, opened 2020-07-23
Swap textobject

theHamsta:
Swap textobject (thinking mainly of function parameters, statements and functions).

What needs to be done first:
- add @object.inner.end (to catch the big enemy of this feature: line comments)
- don't swap with next node but next textobject of same kind (e.g. also a parameter), take documentation and line comments with it.
- to include adjacent comments into the textobject we need the `adjacent?` predicate

But if you want to swap just parameters it already works now (only failure case: there's a line comment behind a parameter)

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,31 @@ require'nvim-treesitter.configs'.setup {
         ["ad"] = "@comment.outer",
         ["am"] = "@call.outer",
         ["im"] = "@call.inner"
-      }
+      },
+      -- swap parameters (keymap -> textobject query)
+      swap_next = {
+        ["<a-p>"] = "@parameter.inner",
+      },
+      swap_previous = {
+        ["<a-P>"] = "@parameter.inner",
+      },
+      -- set mappings to go to start/end of adjacent textobjects (keymap -> textobject query)
+      goto_previous_start = {
+        ["[m"] = "@function.outer",
+        ["[["] = "@class.outer",
+      },
+      goto_previous_end = {
+        ["[M"] = "@function.outer",
+        ["[]"] = "@class.outer",
+      },
+      goto_next_start = {
+        ["]m"] = "@function.outer",
+        ["]]"] = "@class.outer",
+      },
+      goto_next_end = {
+        ["]M"] = "@function.outer",
+        ["]["] = "@class.outer",
+      },
     },
     ensure_installed = "all" -- one of "all", "language", or a list of languages
 }
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -86,8 +86,12 @@ local builtin_modules = {
       return has_some_textobject_mapping(lang) or queries.has_textobjects(lang)
     end,
     keymaps = {},
-    swap_next_keymaps = {},
-    swap_previous_keymaps = {}
+    swap_next = {},
+    swap_previous = {},
+    goto_next_start = {},
+    goto_next_end = {},
+    goto_previous_start = {},
+    goto_previous_end = {}
   }
 }
 
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -85,7 +85,9 @@ local builtin_modules = {
     is_supported = function(lang)
       return has_some_textobject_mapping(lang) or queries.has_textobjects(lang)
     end,
-    keymaps = {}
+    keymaps = {},
+    swap_next_keymaps = {},
+    swap_previous_keymaps = {}
   }
 }
 
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -212,6 +212,36 @@ function M.get_capture_matches(bufnr, capture_string, query_group)
     return matches
 end
 
+function M.find_best_match(bufnr, capture_string, query_group, filter_predicate, scoring_function)
+    if not string.sub(capture_string, 1,2) == '@' then
+      api.nvim_err_writeln('capture_string must start with "@"')
+      return
+    end
+
+    --remove leading "@"
+    capture_string = string.sub(capture_string, 2)
+
+    local best
+    local best_score
+
+    for maybe_match in M.iter_group_results(bufnr, query_group) do
+      local match = utils.get_at_path(maybe_match, capture_string)
+
+      if match and filter_predicate(match) then
+        local current_score = scoring_function(match)
+        if not best then
+          best = match
+          best_score = current_score
+        end
+        if current_score > best_score then
+          best = match
+          best_score = current_score
+        end
+      end
+    end
+    return best
+end
+
 -- Iterates matches from a query file.
 -- @param bufnr the buffer
 -- @param query_group the query file to use
--- a/lua/nvim-treesitter/textobjects.lua
+++ b/lua/nvim-treesitter/textobjects.lua
@@ -8,7 +8,7 @@ local ts_utils = require'nvim-treesitter.ts_utils'
 
 local M = {}
 
-local function get_textobject_at_point(query_string)
+local function textobject_at_point(query_string)
   local bufnr = vim.api.nvim_get_current_buf()
   local lang = parsers.get_buf_lang(bufnr)
   if not lang then return end
@@ -74,59 +74,69 @@ local function get_textobject_at_point(query_string)
 end
 
 function M.select_textobject(query_string)
-  local bufnr, textobject = get_textobject_at_point(query_string)
+  local bufnr, textobject = textobject_at_point(query_string)
   if textobject then
     ts_utils.update_selection(bufnr, textobject)
   end
 end
 
 local function swap_textobject(query_string, direction)
-  local bufnr, textobject_range, node = get_textobject_at_point(query_string)
+  local bufnr, textobject_range, node = textobject_at_point(query_string)
   local step = direction > 0 and 1 or -1
   if not node then return end
   for _ = 1, math.abs(direction), step do
       if direction > 0 then
-        ts_utils.swap_nodes(textobject_range, M.next_textobject(node, query_string, true, bufnr), bufnr, "yes, set cursor!")
+        ts_utils.swap_nodes(textobject_range,
+                            M.next_textobject(node, query_string, true, bufnr),
+                            bufnr,
+                            "yes, set cursor!")
       else
-        ts_utils.swap_nodes(textobject_range, M.previous_textobject(node, query_string, true, bufnr), bufnr, "yes, set cursor!")
+        ts_utils.swap_nodes(textobject_range,
+                            M.previous_textobject(node, query_string, true, bufnr),
+                            bufnr,
+                            "yes, set cursor!")
       end
   end
 end
 
-function M.swap_textobject_next(query_string)
+function M.swap_next(query_string)
   swap_textobject(query_string, 1)
 end
 
-function M.swap_textobject_previous(query_string)
+function M.swap_previous(query_string)
   swap_textobject(query_string, -1)
 end
 
-function M.goto_adjacent_textobejct(query_string, forward, start, same_parent)
-  local bufnr, _, node = get_textobject_at_point(query_string)
-  local ajacent_textobject
+function M.goto_adjacent(query_string, forward, start, same_parent)
+  local bufnr, _, node = textobject_at_point(query_string)
+  local adjacent_textobject
   if forward then
-    ajacent_textobject = M.next_textobject(node,  query_string, same_parent, bufnr)
+    adjacent_textobject = M.next_textobject(node,  query_string, same_parent, bufnr)
   else
-    ajacent_textobject = M.previous_textobject(node,  query_string, same_parent, bufnr)
+    adjacent_textobject = M.previous_textobject(node,  query_string, same_parent, bufnr)
   end
 
-  if ajacent_textobject then
-    local adjacent_textobject_range = {ajacent_textobject:range()}
+  if adjacent_textobject then
+    local adjacent_textobject_range = {adjacent_textobject:range()}
     if start then
-      api.nvim_win_set_cursor(api.nvim_get_current_win(), { adjacent_textobject_range[1] + 1, adjacent_textobject_range[2] })
+      api.nvim_win_set_cursor(api.nvim_get_current_win(),
+                              { adjacent_textobject_range[1] + 1, adjacent_textobject_range[2] })
     else
-      api.nvim_win_set_cursor(api.nvim_get_current_win(), { adjacent_textobject_range[3] + 1, adjacent_textobject_range[4] })
+      api.nvim_win_set_cursor(api.nvim_get_current_win(),
+                              { adjacent_textobject_range[3] + 1, adjacent_textobject_range[4] })
     end
   end
 end
 
-M.goto_next_textobject_start = function(query_string) M.goto_adjacent_textobejct(query_string, 'forward', 'start', false) end
-M.goto_next_textobject_end = function(query_string) M.goto_adjacent_textobejct(query_string, 'forward', false, false) end
-M.goto_previous_textobject_start = function(query_string) M.goto_adjacent_textobejct(query_string, false, 'start', false) end
-M.goto_previous_textobject_end = function(query_string) M.goto_adjacent_textobejct(query_string, false, false, false) end
+-- luacheck: push ignore 631
+M.goto_next_start = function(query_string) M.goto_adjacent(query_string, 'forward', 'start', false) end
+M.goto_next_end = function(query_string) M.goto_adjacent(query_string, 'forward', false, false) end
+M.goto_previous_start = function(query_string) M.goto_adjacent(query_string, false, 'start', false) end
+M.goto_previous_end = function(query_string) M.goto_adjacent(query_string, false, false, false) end
+-- luacheck: pop
 
-function M.goto_next_textobject_end(query_string)
-  local bufnr, _, node = get_textobject_at_point(query_string)
+function M.goto_next_end(query_string)
+  local bufnr, _, node = textobject_at_point(query_string)
   if not node then return end
   local next_textobject = M.next_textobject(node,  query_string, false, bufnr)
   local next_textobject_range = next_textobject:range()
@@ -189,6 +199,13 @@ function M.previous_textobject(node, query_string, same_parent, bufnr)
   return previous_node and previous_node.node
 end
 
+local normal_mode_functions = { "swap_next",
+                                "swap_previous",
+                                "goto_next_start",
+                                "goto_next_end",
+                                "goto_previous_start",
+                                "goto_previous_end"}
+
 function M.attach(bufnr, lang)
   local buf = bufnr or api.nvim_get_current_buf()
   local config = configs.get_module("textobjects")
@@ -206,26 +223,17 @@ function M.attach(bufnr, lang)
       api.nvim_buf_set_keymap(buf, "v", mapping, cmd, {silent = true, noremap = true })
     end
   end
-  for mapping, query in pairs(config.swap_next_keymaps) do
-    if type(query) == 'table' then
-      query = query[lang]
-    elseif not queries.get_query(lang, 'textobjects') then
-      query = nil
-    end
-    if query then
-      local cmd = ":lua require'nvim-treesitter.textobjects'.swap_textobject_next('"..query.."')<CR>"
-      api.nvim_buf_set_keymap(buf, "n", mapping, cmd, {silent = true})
-    end
-  end
-  for mapping, query in pairs(config.swap_previous_keymaps) do
-    if type(query) == 'table' then
-      query = query[lang]
-    elseif not queries.get_query(lang, 'textobjects') then
-      query = nil
-    end
-    if query then
-      local cmd = ":lua require'nvim-treesitter.textobjects'.swap_textobject_previous('"..query.."')<CR>"
-      api.nvim_buf_set_keymap(buf, "n", mapping, cmd, {silent = true})
+  for _, function_call in pairs(normal_mode_functions) do
+    for mapping, query in pairs(config[function_call] or {}) do
+      if type(query) == 'table' then
+        query = query[lang]
+      elseif not queries.get_query(lang, 'textobjects') then
+        query = nil
+      end
+      if query then
+        local cmd = ":lua require'nvim-treesitter.textobjects'."..function_call.."('"..query.."')<CR>"
+        api.nvim_buf_set_keymap(buf, "n", mapping, cmd, {silent = true, noremap = true })
+      end
     end
   end
 end
@@ -246,14 +254,16 @@ function M.detach(bufnr)
       api.nvim_buf_del_keymap(buf, "v", mapping)
     end
   end
-  for mapping, query in pairs(config.swap_next_keymaps) or pairs(config.swap_previous_keymaps) do
-    if type(query) == 'table' then
-      query = query[lang]
-    elseif not queries.get_query(lang, 'textobjects') then
-      query = nil
-    end
-    if query then
-      api.nvim_buf_del_keymap(buf, "n", mapping)
+  for _, function_call in pairs(normal_mode_functions) do
+    for mapping, query in pairs(config[function_call] or {}) do
+      if type(query) == 'table' then
+        query = query[lang]
+      elseif not queries.get_query(lang, 'textobjects') then
+        query = nil
+      end
+      if query then
+        api.nvim_buf_del_keymap(buf, "n", mapping)
+      end
     end
   end
 end
--- a/lua/nvim-treesitter/textobjects.lua
+++ b/lua/nvim-treesitter/textobjects.lua
@@ -8,7 +8,7 @@ local ts_utils = require'nvim-treesitter.ts_utils'
 
 local M = {}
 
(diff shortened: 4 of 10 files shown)


theHamsta:
swap_comment

I changed the feature in that way that I try to swap with the same kind of textobject with same parent. This facilitates also the feature requested in #250 . I added the corresponding functions but without the necessary user configuration.

theHamsta:
I need to check again whether I am handling swapping of multiline nodes correctly.

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +66,41 @@ function M.select_textobject(query_string)
...
+function M.select_textobject(query_string)
+  local bufnr, textobject = get_textobject_at_point(query_string)
+  if textobject then
+    ts_utils.update_selection(bufnr, textobject)
+  end
+end
+
+function M.swap_textobject(query_string, direction)
+  local bufnr, textobject_range, node = get_textobject_at_point(query_string)
+  local step = (direction > 0 and 1 or -1)
```
kyazdani42, line 85: not sure you need the braces here

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +66,41 @@ function M.select_textobject(query_string)
...
 end
 
+function M.select_textobject(query_string)
+  local bufnr, textobject = get_textobject_at_point(query_string)
+  if textobject then
+    ts_utils.update_selection(bufnr, textobject)
+  end
+end
+
+function M.swap_textobject(query_string, direction)
```
kyazdani42, line 83: should we export this if we export swap_next and swap_prev ?

queries/c/textobjects.scm
```diff
@@ -53,3 +53,6 @@
 
 (preproc_else
   (_) @statement.outer)
+
+(parameter_list
+  (parameter_declaration) @parameter.inner)
```
kyazdani42, line 58: is this needed in this PR ?

lua/nvim-treesitter/textobjects.lua
theHamsta: I guess the linter was happier with it, but will investiage.

queries/c/textobjects.scm
theHamsta: Usually you want to swap parameters.

theHamsta:
@stsewd can you test please? I guess your feature should already work perfectly since it does not involve any editing.

 
image

e.g. by replaceing "@parameter.inner" with "@function.inner"

theHamsta:
With the changes in ts_utils I though that it would be great win if we supported "virtual nodes" at some point (arbitrary ranges that behave like nodes so that you can use the same API). You could use this to process a node range in queries or in own Lua code.

queries/c/textobjects.scm
theHamsta: Should I add it for all languages?

lua/nvim-treesitter/textobjects.lua
theHamsta: made it local

theHamsta:
Really cool would be a swap mode like in `Plug 'machakann/vim-swap'`. But it's quite risky since it implies deleting all keymaps and restoring them when finished swapping.

stsewd:
@theHamsta `@function.outer` works!

theHamsta:
@stsewd  You can now do
```lua
            textobjects = {
                enable = true,
                disable = {},
                keymaps = {
                    ["iL"] = {
                        python = "(function_definition) @function",
                        cpp = "(function_definition) @function",
                        c = "(function_definition) @function",
                        java = "(method_declaration) @function"
                    },
                    ["af"] = "@function.outer",
                    ["if"] = "@function.inner",
                    ["aC"] = "@class.outer",
                    ["iC"] = "@class.inner",
                    ["ac"] = "@conditional.outer",
                    ["ic"] = "@conditional.inner",
                    ["ae"] = "@block.outer",
                    ["ie"] = "@block.inner",
                    ["al"] = "@loop.outer",
                    ["il"] = "@loop.inner",
                    ["is"] = "@statement.inner",
                    ["as"] = "@statement.outer",
                    ["ad"] = "@comment.outer",
                    ["id"] = "@comment.inner",
                    ["am"] = "@call.outer",
                    ["im"] = "@call.inner"
                },
                swap_textobject_next = {
                    ["<a-n>"] = "@parameter.inner",
                    gpf = "@function.outer"
                },
                swap_textobject_previous = {
                    ["<a-p>"] = "@parameter.inner",
                    gpF = "@function.outer"
                },
                goto_next_textobject_start = {
                    ["öö"] = "@function.inner",
                },
            },
```
to configure mappings. öö was obviously not mapped in my config ;-)

theHamsta:
Goto node does not work if cursor is not within a textobject and `get_node_at_cursorpos` returns nil

stsewd:
> EDIT: how should the mappings be to replace Vims ]], [[ ?

@theHamsta they would be something like

```lua
    goto_previous_textobject_start = {
      ["[m"] = "@function.outer",
      ["[M"] = "@function.outer.end",
      ["[["] = "@class.outer",
      ["[]"] = "@class.outer.end",
    },
    goto_next_textobject_start = {
      ["]m"] = "@function.outer",
      ["]M"] = "@function.outer.end",
      ["]]"] = "@class.outer",
      ["]["] = "@class.outer.end",
    },

```

theHamsta:
I blindly copied your mapping to the README.md example... I never understood these kind of mappings. Especially, that they always went in the opposite direction that I was expecting.

README.md
```diff
@@ -146,7 +146,31 @@ require'nvim-treesitter.configs'.setup {
...
+        ["[]"] = "@class.outer",
+      },
+      goto_next_textobject_start = {
+        ["]m"] = "@function.outer",
+        ["]]"] = "@class.outer",
+      },
+      goto_next_textobject_end = {
+        ["]M"] = "@function.outer",
+        ["]["] = "@class.outer",
+      },
```
vigoux, line 173: Why including `textobject` everywhere ? I find that easier if we remove them all.

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +66,146 @@ function M.select_textobject(query_string)
...
+
+function M.swap_textobject_next(query_string)
+  swap_textobject(query_string, 1)
+end
+
+function M.swap_textobject_previous(query_string)
+  swap_textobject(query_string, -1)
+end
+
+function M.goto_adjacent_textobeject(query_string, forward, start, same_parent)
```
vigoux, line 110: Typo ? 
```suggestion
function M.goto_adjacent_textobject(query_string, forward, start, same_parent)
```

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +66,146 @@ function M.select_textobject(query_string)
...
+          previous_node = m
+          _, _, previous_node_end = previous_node.node:end_()
+        end
+        if other_end > previous_node_end then
+          previous_node = m
+          _, _, previous_node_end = previous_node.node:end_()
+        end
+      end
+    end
+  end
```
vigoux, line 197: Can't these be refactored a bit ? Using something like a closure to find the appropriate node ?

queries/c/textobjects.scm
vigoux: Is that possible for you ?

README.md
theHamsta: well, swap_previous, swap_next was their previous name that's ok. but how name the goto function. goto_next_start?

vigoux: Yep.

vigoux: We are in the textobject module anyway.

theHamsta:
> Is that possible for you ?

@vigoux  Well, swap textobject was specifically made for parameters.

lua/nvim-treesitter/textobjects.lua
theHamsta: But we have nothing like filter or sort... Lua is copy paste programming. Well we could add the minimum for a more functional programming.

README.md
theHamsta: but what start?

theHamsta: changed it.

lua/nvim-treesitter/textobjects.lua
theHamsta: We could add a generic function to filter and sort nodes.

theHamsta: Is the new version better. At least it could be reused in more situations.

README.md
```diff
@@ -146,7 +146,31 @@ require'nvim-treesitter.configs'.setup {
...
         ["am"] = "@call.outer",
         ["im"] = "@call.inner"
-      }
+      },
+      -- swap parameters (keymap -> textobject query)
+      swap_next = {
+        ["<a-p>"] = "@parameter.inner",
+      },
+      swap_previous = {
+        ["<a-p>"] = "@parameter.inner",
```
stsewd, line 155: I guess this should be different

```suggestion
        ["<a-P>"] = "@parameter.inner",
```

queries/c/textobjects.scm
theHamsta: Yes, tomorrow.

README.md
theHamsta: fixed

queries/c/textobjects.scm
stsewd: Added the ones for python at https://github.com/nvim-treesitter/nvim-treesitter/pull/252

lua/nvim-treesitter/textobjects.lua
theHamsta: As we saw in #252, many languages require a new predicate that allows excluding certain node types (here we need to exclude comments).

So adding @parameter should be separate PR.

queries/c/textobjects.scm
theHamsta: As we saw in #252, many languages require a new predicate that allows excluding certain node types (here we need to exclude comments).

So adding @parameter should be separate PR.

lua/nvim-treesitter/configs.lua
```diff
@@ -85,7 +85,13 @@ local builtin_modules = {
     is_supported = function(lang)
       return has_some_textobject_mapping(lang) or queries.has_textobjects(lang)
     end,
-    keymaps = {}
+    keymaps = {},
+    swap_textobject_next = {},
+    swap_textobject_previous = {},
```
vigoux, line 90: Still some `textobject` here ?

lua/nvim-treesitter/query.lua
```diff
@@ -192,6 +192,36 @@ function M.get_capture_matches(bufnr, capture_string, query_group)
     return matches
 end
 
+function M.find_best_match(bufnr, capture_string, query_group, filter_predicate, scoring_function)
+    if not string.sub(capture_string, 1,2) == '@' then
+      print('capture_string must start with "@"')
```
vigoux, line 197: Meh, why printing here if it's an error ? `nvim_err_writln` should be used.

lua/nvim-treesitter/query.lua
```diff
@@ -192,6 +192,36 @@ function M.get_capture_matches(bufnr, capture_string, query_group)
...
+      return
+    end
+
+    local best
+    local best_score
+
+    --remove leading "@"
+    capture_string = string.sub(capture_string, 2)
```
vigoux, line 205: You should move the check and this part even closer for readability.

lua/nvim-treesitter/query.lua
theHamsta: We always you print in this repo. But I fixed it.

lua/nvim-treesitter/textobjects.lua
vigoux: Yep, looks better.

theHamsta:
> As a closing thought, the textobject module is growing pretty big, and that starts to bother me. I tryllu think the features embedded here are unbelievably useful, but maybe we should start moving things in submodules, as for the refactor module.

You're right. I can do some refactoring later. I would mostly try to reduce code by out-factor shared functionality. `find_best_match` can be done also in `textobject_at_point`. I'll have look whether it makes sense to have a textobject/{movement, selection,swap} module

theHamsta:
We can improve it later to work in regions without treesitter nodes (emtpy lines between functions)

theHamsta:
Swapping also works great for statements. But we maybe also need to exclude comments here...

theHamsta:
Change the README.md example to swap functions/statements?
```
                swap_next = {
                    ["<a-p>"] = "@parameter.inner",
                    ["<a-f>"] = "@function.outer",
                    ["<a-s>"] = "@statement.outer",
                },
                swap_previous = {
                    ["<a-P>"] = "@parameter.inner",
                    ["<a-F>"] = "@function.outer",
                    ["<a-S>"] = "@statement.outer",
                },
```

theHamsta:
Should we do the refactor in submodules now before users add this feature to their config? It makes sense to share attach/detach and current/adjacent textobject. I guess we should only refactor internally by moving the functionalities to separate files while keeping it one module for the users.

vigoux:
the refactor can happen later, internally, but the interface should not change, so either we go for the appropriate interface now, or we refactor now.

theHamsta:
I'd argue for a internal refactor. I want to keep detach/attach in it's current file and then have a subfolder folder with the actual functions: textobject/{shared, swap, move, select}

theHamsta:
I encountered a cursor position issue. Would like to investigate before merging

stsewd:
Some things I have noticed:

- If you are in an empty line, it doesn't jump to the next node (like the empty line between the function and the class)
- It skips nested functions (function inside a function)

```python
def bar(b, a):

    print(b + a)


class Foo:

    def bar(b, a):
        print(b + a)

        def nested():
            pass
```

Otherwise it's working great! I think those issues are pretty minor and can be fixed later.

theHamsta:
> if you are in an empty line, it doesn't jump to the next node (like the empty line between the function and the class)
known issue. need to replace `get_node_at_cursor`

> It skips nested functions (function inside a function)
intended behavior it will only accept textobjects after the current one

With the cursor issue: char_delta is not calculated correctly when `line_delta > 0` will fix that tonight.

stsewd:
> intended behavior. it will only accept textobjects after the current one. The behavior could be changed by an additional option.

Yes please, vim's `[m` check for nested functions by default.

theHamsta:
> Yes please, vim's [m check for nested functions by default.

because Vim does not understand the syntax. I allowed overlaps for goto_next for now.

theHamsta:
> Yes please, vim's [m check for nested functions by default.

because Vim does not understand the syntax. I allowed overlaps for goto_next for now.

theHamsta:
The cursor column after swap should now be also correct if `line_delta ~= 0`. Could not fix it yesterday because I was too dumb to write inequality yesterday (tried `line_delta` and `line_delta >0`)

theHamsta:
```python
def bar(b, a):

    print(b + a)


class Foo:

    def bar(b, a):
        print(b + a)

        def nested():
            pass
```
Your example should work now. Still don't know whether that's a good idea.

theHamsta:
I guess my byte counting for cursor positioning could also be wrong when someone uses non-ascii characters. Should convert the string length to bytes to by correct.

stsewd:
> Your example should work now. Still don't know whether that's a good idea.

Thanks, for me it feels natural to move to the next node you see in the code, there can be more complex things like a meta class.

I tried pulling this branch, but  I can still see that node is being skipped

theHamsta:
>Thanks, for me it feels natural to move to the next node you see in the code, there can be more complex things like a meta class.

>I tried pulling this branch, but I can still see that node is being skipped

Because I forgot to push.

theHamsta:
Ooops, I wanted to do a rebase and force pushed a old version of this PR from a different machine.  54bb9de  should be the correct revision. Sorry! Can revert on the weekend when I am at my other machine.

theHamsta:
`git fetch fork 54bb9de` did the necessary magic. Glad that git never forgets :smile:

theHamsta:
But when I just tested the behavior of the goto command I had to assume that for next/previous different nodes may have the same end, but never the same start to get the nested function behavior correctly.

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +66,147 @@ function M.select_textobject(query_string)
...
+  local bufnr, _, node = textobject_at_point(query_string)
+  local adjacent_textobject
+  if forward then
+    adjacent_textobject = M.next_textobject(node,  query_string, same_parent, overlapping_range_ok, bufnr)
+  else
+    adjacent_textobject = M.previous_textobject(node,  query_string, same_parent, overlapping_range_ok, bufnr)
+  end
+
+  if adjacent_textobject then
+    local adjacent_textobject_range = {adjacent_textobject:range()}
```
stsewd, line 122:
```suggestion
    local adjacent_textobject_range = {adjacent_textobject:range()}
    -- Set the item in jump list
    vim.cmd "normal! m'"
```

vigoux: We should have something to jump to a given node / range and set's that.

theHamsta: :+1: I changed it. Put the extra function to the refactor TODOs.

theHamsta:
@kyazdani42 is this ready? A split-up of the textobjects module in smaller files will happen after this PR (user invisble change).

kyazdani42:
I'll review this tonight, but it should be good :)

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +66,150 @@ function M.select_textobject(query_string)
...
+  local bufnr, textobject = textobject_at_point(query_string)
+  if textobject then
+    ts_utils.update_selection(bufnr, textobject)
+  end
+end
+
+local function swap_textobject(query_string, direction)
+  local bufnr, textobject_range, node = textobject_at_point(query_string)
+  local step = direction > 0 and 1 or -1
+  if not node then return end
```
kyazdani42, line 86: i think you can do this before the step declaration (even though it doesn't matter much)

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +66,150 @@ function M.select_textobject(query_string)
...
+  local overlapping_range_ok = false
+  local same_parent = true
+  for _ = 1, math.abs(direction), step do
+      if direction > 0 then
```
kyazdani42, line 90: this for loop is not properly indented

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +66,150 @@ function M.select_textobject(query_string)
...
+  local bufnr, _, node = textobject_at_point(query_string)
+  local adjacent_textobject
+  if forward then
+    adjacent_textobject = M.next_textobject(node,  query_string, same_parent, overlapping_range_ok, bufnr)
+  else
+    adjacent_textobject = M.previous_textobject(node,  query_string, same_parent, overlapping_range_ok, bufnr)
+  end
+
+  if adjacent_textobject then
+    -- Set the item in jump list
```
kyazdani42, line 122: i think this comment is not required ? a little `:help m'` gives the information

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +66,150 @@ function M.select_textobject(query_string)
...
+  if overlapping_range_ok then
+    _, _, search_end = node:end_()
+    search_end = search_end + 1
+  else
+    _, _, search_end = node:start()
+  end
+  local previous_node = queries.find_best_match(bufnr,
+                                                query_string,
+                                                'textobjects',
+                                                function(match)
```
kyazdani42, line 191: maybe you could define this function outside and make a closure to instantiate parameters that comes from the upper function ? it would be easier to read and understand ^^

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -187,12 +182,68 @@ function M.is_in_node_range(node, line, col)
...
+
+  local text1 = M.get_node_text(node_or_range1)
+  local text2 = M.get_node_text(node_or_range2)
+
+  local edit1 = { range = range1, newText = table.concat(text2, '\n') }
+  local edit2 = { range = range2, newText = table.concat(text1, '\n') }
+  vim.lsp.util.apply_text_edits({edit1, edit2}, bufnr)
+
+  if cursor_to_second then
+    -- Set the item in jump list
```
kyazdani42, line 214: this is not needed either imo

theHamsta: before we had discussion with @vigoux to create a helper function for this.

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +67,158 @@ function M.select_textobject(query_string)
...
+      ts_utils.swap_nodes(textobject_range,
+                          M.next_textobject(node, query_string, same_parent, overlapping_range_ok, bufnr),
+                          bufnr,
+                          "yes, set cursor!")
+    else
+      ts_utils.swap_nodes(textobject_range,
+                          M.previous_textobject(node, query_string, same_parent, overlapping_range_ok, bufnr),
+                          bufnr,
+                          "yes, set cursor!")
+    end
```
kyazdani42, line 102: now i come back to this, i would suggest a more readable way of writing this:
```lua
local fn = direction > 0 and M.next_textobject or M.previous_textobject
local t_obj = fn(node, query_string, same_parent, overlapping_range_ok, bufnr)
ts_utils.swap_nodes(textobject_range, t_obj, bufnr, "yes, set cursor!")
```

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +67,158 @@ function M.select_textobject(query_string)
...
+end
+
+function M.goto_adjacent(query_string, forward, start, same_parent, overlapping_range_ok)
+  local bufnr, _, node = textobject_at_point(query_string)
+  local adjacent_textobject
+  if forward then
+    adjacent_textobject = M.next_textobject(node,  query_string, same_parent, overlapping_range_ok, bufnr)
+  else
+    adjacent_textobject = M.previous_textobject(node,  query_string, same_parent, overlapping_range_ok, bufnr)
+  end
```
kyazdani42, line 121: i think you could do the same thing here, caching the function that takes the exact same arguments so you can do a 2 line function call instead of a full 6 lines if/else statement :)

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +67,158 @@ function M.select_textobject(query_string)
...
+    utils.set_jump()
+
+    local adjacent_textobject_range = {adjacent_textobject:range()}
+    if start then
+      api.nvim_win_set_cursor(api.nvim_get_current_win(),
+                              { adjacent_textobject_range[1] + 1, adjacent_textobject_range[2] })
+    else
+      api.nvim_win_set_cursor(api.nvim_get_current_win(),
+                              { adjacent_textobject_range[3] + 1, adjacent_textobject_range[4] })
+    end
```
kyazdani42, line 133: same here, you could cache the object in a variable and call the function only once. Control flow would be easier to understand

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +67,158 @@ function M.select_textobject(query_string)
...
+                              if not same_parent or node:parent() == match.node:parent() then
+                                local _, _, start = match.node:start()
+                                local _, _, end_ = match.node:end_()
+                                return start > search_start and end_ >= node_end
+                              end
+                            end
+  local filter_function = function(match)
+                            local _, _, node_start = match.node:start()
+                            return -node_start
+                          end
```
kyazdani42, line 167: i'm not too found of this kind of code because of this weird indentation but we'll keep it like this for the moment.

lua/nvim-treesitter/textobjects.lua
```diff
@@ -66,13 +67,158 @@ function M.select_textobject(query_string)
...
+  local filter_function = function(match)
+                            local _, _, node_end = match.node:end_()
+                            return node_end
+                          end
+
+  local previous_node = queries.find_best_match(bufnr,
+                                                query_string,
+                                                'textobjects',
+                                                scoring_function,
+                                                filter_function)
```
kyazdani42, line 210: couldn't this be written in one line ?

lua/nvim-treesitter/utils.lua
```diff
@@ -67,4 +67,8 @@ function M.print_warning(text)
   api.nvim_command(string.format([[echohl WarningMsg | echo "%s" | echohl None]], text))
 end
 
+function M.set_jump()
+  vim.cmd "normal! m'"
+end
```
kyazdani42, line 72: cool

lua/nvim-treesitter/textobjects.lua
theHamsta: Will fix it. Probably as a local function

theHamsta:
@kyazdani42 ping

kyazdani42:
@theHamsta I still have the same remarks as the last review i made (because the code hasn't changed ?)

theHamsta:
@kyazdani42 sorry for that, apparently I did revert the changes a few days ago when I tried to rebase on changed master :facepalm: Should be updated now.

PR #219, opened 2020-07-23
Java locals

theHamsta:
This #204 by @PitcherTear22 , rebased (was very outdate), squashed and extended.

Now, imports are also definitions. The online playground cannot be fully trusted for Java. The parsers version seems to be outdated. The parsed tree for imports was drastically different. Stupid me was testing `first?` `last?` on this :-(

Using the `highlight_current_scope` with this scopes hints for a bug in finding the current scope (e.g. )... Need to investigate more, it works fine for the other languages.

scope

Maybe we need to wait for a simple version of submatch that only requires a specific parent node.

### DIFF
--- a/queries/java/highlights.scm
+++ b/queries/java/highlights.scm
@@ -103,7 +103,7 @@
 ((identifier) @constant
   (#match? @constant "^_*[A-Z][A-Z\d_]+"))
 
-
+(this) @constant.builtin
 
 ; Literals
 
--- a/queries/java/highlights.scm
+++ b/queries/java/highlights.scm
@@ -13,6 +13,8 @@
 ; Parameters
 (formal_parameter
   name: (identifier) @parameter)
+(catch_formal_parameter
+  name: (identifier) @parameter)
 
 ;; Lambda parameter
 (inferred_parameters (identifier) @parameter) ; (x,y) -> ...
--- a/queries/java/locals.scm
+++ b/queries/java/locals.scm
@@ -1,10 +1,69 @@
-; CREDITS @maxbrunsfeld (maxbrunsfeld@gmail.com)
-
+; SCOPES
+; declarations
+(program) @scope
 (class_declaration
-  name: (identifier) @name) @class
+  body: (_) @scope)
+(enum_declaration
+  body: (_) @scope)
+(method_declaration) @scope ; whole method_declaration because arguments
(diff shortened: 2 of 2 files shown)


queries/java/highlights.scm
```diff
@@ -103,7 +105,7 @@
 ((identifier) @constant
   (#match? @constant "^_*[A-Z][A-Z\d_]+"))
 
-
+(this) @constant.builtin
```
vigoux, line 108: I think that we went for `variable.builtin` in other languages.

queries/java/locals.scm
```diff
@@ -1,10 +1,104 @@
...
+          (scoped_identifier
+            ((identifier) @definition.import)))))))
+(import_declaration
+  (scoped_identifier
+    (scoped_identifier
+      (scoped_identifier
+        (scoped_identifier
+          (scoped_identifier
+            (scoped_identifier
+              ((identifier) @definition.import))))))))
```
vigoux, line 94: We really need submatch, can't we use something like `(scoped_identifier (identifier))` alone ?

queries/java/highlights.scm
theHamsta: `variable.builtin` never happened. Other languages, use `constant.builtin`. Could change this for all languages.

queries/java/locals.scm
theHamsta: This would also match usages of the classes like `google.FooClass()`. I can finish the predicates PR today with a simplified version of submatch (only looks for a specific node type as parent).

A real submatch is difficult to implement with current query syntax.

queries/java/highlights.scm
vigoux: Nope, keep it this way if it is consistent

queries/java/locals.scm
vigoux: No rush, maybe we should ask the question in upstream treesitter.

vigoux: Something like a new wildcard, or a new meaning for `_`.

theHamsta: I removed this for now. Can be solved e.g. with the predicates PR #208

theHamsta:
We now use `has-ancestor?` to match the nested structure of imports.

PR #222, opened 2020-07-26
add highlight for @variable

ray-x:
Add highlight for variable and variable.builtin


ray-x:
Failed to pass lua check

PR #223, opened 2020-07-26
Dart highlighs: disable invalid nodes after parser update to fix CI

theHamsta:
Dart highlights currently break our CI. Apparently, the grammar has changed upstream. I did some bisecting to deactivate all invalid nodes: @Akin909 

Related: #222 @ray-x

### DIFF
--- a/queries/dart/highlights.scm
+++ b/queries/dart/highlights.scm
@@ -112,14 +112,14 @@
 
 [
  "abstract"
- "assert"
- "break"
+ ;"assert"
+ ;"break"
  "on"
  "class"
  "default"
  "enum"
  "extends"
- "final"
+ ;"final"
  "implements"
  "as"
  "mixin"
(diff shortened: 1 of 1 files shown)


theHamsta:
Interesting: the parser recognizes $-escaped identifiers in string literals. But the parsed range is not correct (only one char long):

image

theHamsta:
All the affected keyword seem to be hidden now. So we can no longer capture them: https://github.com/UserNobody14/tree-sitter-dart/commit/c0774755b82ae5f714b7788123f8b9961fe42b1c
```javascript
        _break_builtin: $ => token('break'),
        _assert_builtin: $ => token('assert'),
        _case_builtin: $ => token('case'),
```
@TimWhiting

vigoux:
Something to note : we should pin the cplit of the parsers to avoid such breakage

PR #227, opened 2020-07-27
configs: attach on unknown parsers too

### LINKED ISSUE #226: Impossible to add a custom parser
**Describe the bug**
When building and installing manually a custom parser, `nvim-treesitter` does not let you use it easily.

**To Reproduce**
Build a custom parser or get `tree-sitter-viml` without installing it.

Try to use it with the playground.

Nothing happens and the buffer is not attached to.

**Expected behavior**
It should be possible to add your own parser / queries and use nvim-treesitter like with any other parser.

**Additional context**
I think the problem comes from how the module system handles languages.

vigoux:
Fixes #226

### DIFF
--- a/lua/nvim-treesitter/ts_utils.lua
+++ b/lua/nvim-treesitter/ts_utils.lua
@@ -113,6 +113,7 @@ function M.get_named_children(node)
 end
 
 function M.get_node_at_cursor(winnr)
+  if not parsers.has_parser() then return end
   local cursor = api.nvim_win_get_cursor(winnr or 0)
   local root = parsers.get_parser():parse():root()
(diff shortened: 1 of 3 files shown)


lua/nvim-treesitter/ts_utils.lua
```diff
@@ -113,6 +113,7 @@ function M.get_named_children(node)
 end
 
 function M.get_node_at_cursor(winnr)
+  if not parsers.has_parser() then return end
```
theHamsta, line 116: Do we have to add this check to more functions. I guess not (?). This here is only relevant because of the statusline function, right?

vigoux: Yes that's right.

PR #230, opened 2020-07-27
install: pin parser versions for stability

vigoux:
This will avoid the Ci problems we encoutered before. Some of them are not done yet because of lazyness, it will be needed to pin them later.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -5,21 +5,21 @@ local list = {}
 
 list.javascript = {
   install_info = {
-    url = "https://github.com/tree-sitter/tree-sitter-javascript",
+    url = "https://github.com/tree-sitter/tree-sitter-javascript/tree/0711e93eb265f729c88c5555989d60bee3011873",
     files = { "src/parser.c", "src/scanner.c" },
   }
 }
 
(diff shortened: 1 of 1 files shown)


vigoux:
Wrong way to do it, still need to find a better way.
Maybe we can just avoid using git, but use things like unzip instead.

theHamsta:
You can read the versions from a file that is auto-generated by a bash script. But I don't think that's necessary.

The actual problem is that we load all the query files of all the languages at start-up and don't give a helpful error message.
Our CI guarantees that up-to-date parsers work. Users should never see an error for Dart queries when the do not use dart. If they do then the error message should be `error loading dart highlight queries (files: ...). Syntax error in query files or outdated parser? (try :TSInstall dart)`.

vigoux:
I quite like the idea. Having better messages can be done easily too imo. And it might be something worth upstreaming.

theHamsta:
For the pin feature: I would just add a "git reset --hard <commit>" to the install pipeline. The commits could come from a textfile with the format "lang commit" per line. We would have a bash/python script to update the textfile when bumping the versions.

vigoux:
My problem is that the `git reset --hard` way forces us the clone the whole repo each time, with its history, which can take time.

PR #237, opened 2020-07-28
feat(parsers): add scheme parser

steelsojka:


README.md
```diff
@@ -296,6 +296,7 @@ List of currently supported languages:
 - [ ] jsdoc
 - [x] dart (maintained by @Akin909)
 - [x] rst (maintained by @stsewd)
+- [x] scheme (maintained by @steelsojka)
```
theHamsta, line 299: Tree-sitter query language? It's not real scheme.

steelsojka: I wasn't sure exactly. What should the official "language" name be? ts-query?

steelsojka: I'll rename the parser as well...

steelsojka: Does the filetype need to change? `.scm` files are detected as 'scheme' (at least with my setup)

theHamsta: Should be enough not make that clear in the README.md. I guess you can keep all other usages of scheme. Until someone wants to parse real scheme... :smile:

PR #238, opened 2020-07-28
rename query parser to query from scheme

steelsojka:


lua/nvim-treesitter/parsers.lua
```diff
@@ -236,9 +236,9 @@ list.jsdoc = {
   }
 }
 
-list.scheme = {
+list.query = {
   install_info = {
-    url = "https://github.com/nvim-treesitter/tree-sitter-scheme",
+    url = "https://github.com/nvim-treesitter/tree-sitter-query",
     files = { "src/parser.c" }
```
theHamsta, line 242: @steelsojka Missing used_for ! or you have to define a query ftype

steelsojka: I think the filetype will just be `query`. I haven't added a way to intelligently detect a query file vs a scheme file yet.

PR #241, opened 2020-07-31
feat(predicates): add adjacent predicate

steelsojka:
Adds the adjacent predicate. Allows to match nodes that are directly followed by one or more match or type. Here are some examples.

Matches a comment node that is directly followed by a function definition.
```scheme
(
  (comment) @function.doc
  (function_declaration) @function.definition
  (adjacent? @function.doc @function.definition)
)
```

Matches any comment that is direclty followed by a variable declaration or function declaration
```scheme
(
  (comment) @comment
  (adjacent? @comment "function_declaration" "lexical_declaration")
)
```

Matches a block of comments adjacent to a function declaration.
```scheme
(
  (comment) @function.doc
  (function_declaration) @function.definition
  (adjacent_block? @function.doc @function.definition)
)
```

The above would match this.

```javascript
// This comment is matched
// So is this one!
function test() {}
```

This predicate takes any number of captures and/or string node names.

### DIFF
--- a/lua/nvim-treesitter/query_predicates.lua
+++ b/lua/nvim-treesitter/query_predicates.lua
@@ -1,4 +1,5 @@
 local utils = require'nvim-treesitter.utils'
+local ts_utils = require'nvim-treesitter.ts_utils'
 
 local M = {}
 
@@ -10,6 +11,42 @@ end
 
 local function get_node(query, match, pred_item)
   return utils.get_at_path(match, query.captures[pred_item]..'.node')
+end
+
+local function create_adjacent_predicate(match_successive_nodes)
+  return function(query, match, pred)
+    if #pred < 3 then error("adjacent? must have at least two arguments!") end
+    local node = get_node(query, match, pred[2])
+    if not node then return true end
+
+    local adjacent_types = {unpack(pred, 3)}
+    local adjacent_node = ts_utils.get_next_node(node)
+
+    if match_successive_nodes then
+      -- Move to the last node in a series that doesn't match the node type
+      -- and use that node to compare with.
+      while adjacent_node and adjacent_node:type() == node:type() do
+        node = adjacent_node
+        adjacent_node = ts_utils.get_next_node(node)
+      end
+    end
+
+    if not adjacent_node then return false end
+
+    for _, adjacent_type in ipairs(adjacent_types) do
+      if type(adjacent_type) == "number" then
+        if get_node(query, match, adjacent_type) == adjacent_node then
+          return true
+        end
+      elseif type(adjacent_type) == "string" then
+        if adjacent_node:type() == adjacent_type then
+          return true
+        end
+      end
+    end
+
+    return false
+  end
 end
 
 function M.check_predicate(query, match, pred)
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/query_predicates.lua
```diff
@@ -71,4 +72,30 @@ M['has_ancestor?'] = function(query, match, pred)
   return false
 end
 
+M['adjacent?'] = function(query, match, pred)
+  if #pred < 3 then error("adjacent? must have at least two arguments!") end
+  local node = get_node(query, match, pred[2])
+  if not node then return true end
+
+  local adjacent_types = {unpack(pred, 3)}
+  local adjacent_node = ts_utils.get_next_node(node, false, false)
```
theHamsta, line 81: You can leave away the optional parameters. Maybe we should remove them altogether from the function since additional crazyness for node movement is no longer required (node movement is also with `false,false` better.

lua/nvim-treesitter/query_predicates.lua
```diff
@@ -71,4 +72,30 @@ M['has_ancestor?'] = function(query, match, pred)
...
+
+  if not adjacent_node then return false end
+
+  for _, adjacent_type in ipairs(adjacent_types) do
+    print(vim.inspect(adjacent_type))
```
theHamsta, line 86: remove

lua/nvim-treesitter/query_predicates.lua
```diff
@@ -71,4 +72,30 @@ M['has_ancestor?'] = function(query, match, pred)
...
+    if type(adjacent_type) == "number" then
+      if get_node(query, match, adjacent_type) == adjacent_node then
```
theHamsta, line 88: This is nice to read but implies a full loop to all children on each call. Should be tolerable.

lua/nvim-treesitter/query_predicates.lua
steelsojka: lol. my bad

lua/nvim-treesitter/query_predicates.lua
steelsojka: Are you referring to the loop of adjacent_types?

steelsojka:
@theHamsta I think we can do that. I'll update this pr with it

steelsojka:
@theHamsta treesitter didn't like predicate name with `*` in it. So I did `adjacent_block` since it matches a block of nodes against an adjacent node. I can change the name if you have a better one 😉

lua/nvim-treesitter/query_predicates.lua
theHamsta: get_node loops over all children each call. would be enough to just check the next node.

lua/nvim-treesitter/query_predicates.lua
```diff
@@ -71,4 +108,7 @@ M['has_ancestor?'] = function(query, match, pred)
   return false
 end
 
+M['adjacent?'] = create_adjacent_predicate(false)
+M['adjacent_block?'] = create_adjacent_predicate(true)
```
theHamsta, line 112: use lisp naming convention: `adjacent-block?` or `adjacent-same-node-groups?`

lua/nvim-treesitter/query_predicates.lua
steelsojka: sounds good. I'll make that change

steelsojka: This get_node call just seems to do this

```lua
local function get_node(query, match, pred_item)
  return utils.get_at_path(match, query.captures[pred_item]..'.node')
end
```

Are you thinking about a different call?

PR #244, opened 2020-08-01
Add queries for PHP, TOML and HTML

tk-shirasaka:
I added query for some languages.
If you would like, can you merge this pull request?

### DIFF
--- a/queries/javascript/locals.scm
+++ b/queries/javascript/locals.scm
@@ -6,6 +6,9 @@
 (arrow_function) @scope
 (function_declaration) @scope
 (method_definition) @scope
+(for_statement) @scope
+(for_in_statement) @scope
+(catch_clause) @scope
 
 ; Definitions
 ;------------
--- a/queries/php/highlights.scm
+++ b/queries/php/highlights.scm
@@ -1,0 +1,191 @@
+; Types
+
+[
+ (primitive_type)
+ (cast_type)
+ ] @type.builtin
+(type_name (name) @type)
+
+; Functions
+
+(array_creation_expression "array" @function.builtin)
+(list_literal "list" @function.builtin)
+
+(method_declaration
+  name: (name) @method)
+
+(function_call_expression
+  function: (qualified_name (name)) @function)
+
+(scoped_call_expression
+  name: (name) @function)
+
+(member_call_expression
+  name: (name) @method)
+
+(function_definition
+  name: (name) @function)
+
+; Member
+
+(property_element
+  (variable_name) @property)
+
+(member_access_expression
+  name: (variable_name (name)) @property)
+(member_access_expression
+  name: (name) @property)
+
+; Variables
+
+(relative_scope) @variable.builtin
+
+((name) @constant
+ (#match? @constant "^_?[A-Z][A-Z\d_]+$"))
+
+((name) @constructor
+ (#match? @constructor "^[A-Z]"))
+
+((name) @variable.builtin
+ (#eq? @variable.builtin "this"))
+
+(variable_name) @variable
+
+; Basic tokens
+
+[
+ (string)
+ (heredoc)
+ ] @string
+
+(boolean) @boolean
+(null) @constant.builtin
+(integer) @number
+(float) @float
+(comment) @comment
+
+; Keywords
+
+[
+ "$"
+ "abstract"
+ "as"
+ "break"
+ "class"
+ "const"
+ "continue"
+ "declare"
+ "default"
+ "echo"
+ "enddeclare"
+ "extends"
+ "final"
+ "function"
+ "global"
+ "implements"
+ "insteadof"
+ "interface"
+ "namespace"
+ "new"
+ "private"
+ "protected"
+ "public"
+ "return"
+ "static"
+ "trait"
+ ] @keyword
+
+[
+ "case"
+ "else"
+ "elseif"
+ "endif"
+ "endswitch"
+ "if"
+ "switch"
+ ] @conditional
+
+[
+ "do"
+ "endfor"
+ "endforeach"
+ "endwhile"
+ "for"
+ "foreach"
+ "while"
+ ] @repeat
+
+[
+ "catch"
+ "finally"
+ "throw"
+ "try"
+ ] @exception
+
+[
+ "include_once"
+ "include"
+ "require_once"
+ "require"
+ "use"
+ ] @include
+
+[
+ ","
+ ";"
+ "."
+ ] @punctuation.delimiter
+
+[
+ (php_tag)
+ "?>"
+ "("
+ ")"
+ "["
+ "]"
+ "{"
+ "}"
+ ] @punctuation.bracket
+
+[
+  "="
+
+  "-"
+  "*"
+  "/"
+  "+"
+  "%"
+
+  "~"
+  "|"
+  "&"
+  "<<"
+  ">>"
+
+  "->"
+
+  "<"
+  "<="
+  ">="
+  ">"
+  "=="
+  "!="
+  "==="
+  "!=="
+
+  "!"
+  "&&"
+  "||"
+
+  "-="
+  "+="
+  "*="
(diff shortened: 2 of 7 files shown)


queries/php/highlights.scm
```diff
@@ -0,0 +1,114 @@
...
+"require_once" @keyword
+"require" @keyword
+"return" @keyword
+"static" @keyword
+"switch" @keyword
+"throw" @keyword
+"trait" @keyword
+"try" @keyword
+"use" @keyword
+"while" @keyword
```
theHamsta, line 114: Could you use the bracket notation for multiple nodes of one type, like [ "private" "protected" ...] @keyword

Also please read contributing.md for different categories of keywords

queries/toml/highlights.scm
```diff
@@ -0,0 +1,10 @@
+(bare_key) @type.builtin
+
+(pair
+  (bare_key) @property)
+
+(string) @string
+(boolean) @constant.builtin
+(integer) @number
```
theHamsta, line 8: Please distinguish floats -> Contributing.md

queries/typescript/locals.scm
```diff
@@ -1,2 +1,13 @@
+; Scopes
+;-------
+
+(method_definition) @scope
+(if_statement) @scope
+(for_statement) @scope
+(try_statement) @scope
+
```
theHamsta, line 8: Isn't this already included in the is queries?

tk-shirasaka: I added this scope because it wasn't highlighted in a particular statement
But if this is not the correct behavior, I'm roll it back.

Should I roll back?

queries/toml/highlights.scm
tk-shirasaka: Thank you for your advice!

Differentiated between floating point numbers and integers.

queries/php/highlights.scm
tk-shirasaka: Thank you for advice!

I'm fixed it.

queries/typescript/locals.scm
theHamsta: All those things also exist in JS. So you can add them to the JS queries and have them in typescript automatically.
It's probably good to add `statement_block` as scope since curly braces introduce a scope. 

@steelsojka  could you review please?

tk-shirasaka: Is that so.
Thank you for teaching me
I rolled back the scope query, so check again

tk-shirasaka: Is that so.
Thank you for teaching me
I rolled back the scope query, so check again

theHamsta: why not add them to JS instead?

queries/javascript/locals.scm
```diff
@@ -6,6 +6,10 @@
 (arrow_function) @scope
 (function_declaration) @scope
 (method_definition) @scope
+(if_statement) @scope
+(for_statement) @scope
+(switch_statement) @scope
+(try_statement) @scope
```
steelsojka, line 12: try_statement doesn't actually create a new scope (it's already covered by statement_block), however `catch_clause` does. I would just change this to `catch_clause`

queries/javascript/locals.scm
```diff
@@ -6,6 +6,10 @@
 (arrow_function) @scope
 (function_declaration) @scope
 (method_definition) @scope
+(if_statement) @scope
+(for_statement) @scope
```
steelsojka, line 10: Can you also add `for_in_statement` to this as well?

queries/javascript/locals.scm
```diff
@@ -6,6 +6,10 @@
...
+(switch_statement) @scope
```
steelsojka, line 11: I don't believe a switch statement actually creates a scope. You can create one in the case but that is covered by a `statement_block`

```js
case a: {
   // stuff
}
```

queries/javascript/locals.scm
```diff
@@ -6,6 +6,10 @@
 (arrow_function) @scope
 (function_declaration) @scope
 (method_definition) @scope
+(if_statement) @scope
```
steelsojka, line 9: This one is already covered by the `statement_block` in the consequence.

tk-shirasaka: I understood as follows.
- if_statement: Not required (because it is covered by statement_block)
- for_statement: add for_in_statement
- switch_statement: Not required (because the scope is not actually created)
- try_statement: changed to catch_clause (because try_statement doesn't actually create scope, but catch_clause does)

Is this perception correct?

tk-shirasaka: I fixed it as above
Please check again.

queries/php/highlights.scm
```diff
@@ -0,0 +1,191 @@
...
+ ] @include
+
+[
+ ","
+ ";"
+ ] @punctuation.delimiter
+
+[
+ (php_tag)
+ "?>"
```
vigoux, line 141: Those two look weird to me, I guess `php_tag` is the whole node, and you might want to capture only the start and end of the tag using `(php_tag ... @punctuation.bracket)`.

tk-shirasaka: Thank you for your advice!
I tried your advice, but the tree-sitter got the following error:
```
Error detected while proccessing /path/to/nvim-treesitter.vim
line 23:
E5108: Error executing lua .../runtime/lua/vim/treesitter.lua:151: query: invalid syntax as position 1700
```

And when I checked the sample php program and its tree-sitter structure, it became as follows.

Sample program:
```php
<?php
$title = "sample";
?>

<html>
    <head>
        <title><?= $title ?></title>
    </head>
</html>
```

Tree sitter structure:
```
program [0, 0] - [0, 9])
  php_tag [0, 0] - [5, 0])
  expression_statement [0, 1] - [18, 1])
    assignment_expression [0, 1] - [17, 1])
      variable_name [0, 1] - [6, 1])
        name [1, 1] - [6, 1])
      string [9, 1] - [17, 1])
  text_interpolation [0, 2] - [18, 6])
    text [0, 4] - [15, 6])
    php_tag [15, 6] - [18, 6])
  expression_statement [19, 6] - [25, 6])
    variable_name [19, 6] - [25, 6])
      name [20, 6] - [25, 6])
  text_interpolation [26, 6] - [0, 9])
    text [28, 6] - [0, 9])
```

I think that "php_tag" is not whole node, isn't it?
I could be wrong, and if so, please point me out.

theHamsta: I just tested it "php_tag" is just the start, not the whole node.

tk-shirasaka:
Thank you for taking the valuable time to review this.
I may not be enough, but I'd like to be a PHP maintainer until someone better comes along.

vigoux:
Then feel free to chenge the `php` item is the `README` and add you as the maintainer, and we're glad to count you as part of the maintainer team !

queries/php/highlights.scm
```diff
@@ -0,0 +1,192 @@
...
+ (heredoc)
+ ] @string
+
+(boolean) @boolean
+(null) @constant.builtin
+(integer) @number
+(float) @float
+(comment) @comment
+
+"$" @constant
```
vigoux, line 67: Meh `@constant`, can we add something more meaningful like `@keyword` or something ?

tk-shirasaka: You are right.
I will fix this.

By the way, since I've added a toml query this time, should I change the toml field in the readme as well?

vigoux: I guess you can add this too if you want to maintain toml too !

tk-shirasaka: Thanks for the confirmation!
I've fixed the above issue and changed the readme.

If all is well, can you merge this PR?

tk-shirasaka:
Sorry for my late reply.
Please merge.

vigoux:
Hey no rush dude 😉

PR #261, opened 2020-08-05
PHP: Update locals query

tk-shirasaka:
Reference and definition queries have been improved as follows

Added definitions for methods, functions, and classes, and all specific names are now referenced.

### DIFF
--- a/queries/php/locals.scm
+++ b/queries/php/locals.scm
@@ -2,21 +2,42 @@
 ;-------
 
 (class_declaration) @scope
-(property_declaration) @scope
-(method_declaration) @scope
-(function_definition) @scope
 (while_statement) @scope
 (foreach_statement) @scope
 (if_statement) @scope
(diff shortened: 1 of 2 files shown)


queries/php/locals.scm
```diff
@@ -2,21 +2,42 @@
...
+  (qualified_name
+    (name) @definition.class))
+
+(class_declaration
+  name: (name) @definition.class)
 
 ; References
 ;------------
 
-(variable_name) @reference
```
vigoux, line 22: Why ?

tk-shirasaka: (variable_name) contains "$" and (name), and class names, function names, etc. do not contain "$".
"$", if present, will not match all of the names except (variable_name), so I changed (variable_name) to (name).

tk-shirasaka:
Thank you for review and merge!

Issue #142, opened 2020-07-02
Incorrect definition highlighting in for different scopes


steelsojka:
**Describe the bug**
Definition is incorrect when the focused symbol is the definition.

**To Reproduce**
Given this javascript file

![nvim-ts-error-1](https://user-images.githubusercontent.com/2062154/86373096-a3344380-bc48-11ea-97e5-3d745a79a4e0.png)

When the cursor is on the variable `a` (not the function parameter), the function parameter `a` is highlighted as the definition.
If the parenthesis are removed from the arguments the variable definition is flagged as the definition of the argument.'

![nvim-ts-error-2](https://user-images.githubusercontent.com/2062154/86373289-e2fb2b00-bc48-11ea-9754-cd127805675e.png)

**Expected behavior**
Neither one should be highlighted since they are in different scopes and are definitions them self. This may just be an issue with the queries for javascript or the algorithm used for finding definitions.

**Output of `:checkhealth nvim_treesitter` ***
```
Paste the output here
```

**Output of `nvim --version`**
```
Paste your output here
```

**Additional context**

vigoux:
Can this be a `locals.scm` issue ?

steelsojka:
@vigoux It may be. I need to look into a little bit further.

vigoux:
Sorry deleted the coment, but it might be related to locals.

theHamsta:
I think the find_definition should always find declarations before the current cursor position.

Consider the following example:
```cpp
void normalReactionTest(cr::HCSITS& cr, SphereID sp)
{
   contactThreshold = Thresholds<real_t>::contactThreshold();
   // plane at 5,5,5
   // radius 1.1
   sp->setPosition(  Vec3(5,5,6) );
   sp->setLinearVel( Vec3(0,0,0) );
   cr.setErrorReductionParameter( real_t(1.0) ); // <----------------- cursor at beginning of line
   cr.timestep( real_c( real_t(1.0) ) );
}

int main( int argc, char** argv )
{
   cr::HCSITS cr(globalBodyStorage, forest->getBlockStoragePointer(), storageID, hccdID, fcdID); // <-- will find this definition
   return EXIT_SUCCESS;
}
} // namespace walberla

```

Without #145 it can't recognize the definition of `cr` as an arguments and finds cr in a completely different scope (because also the scope definition for function bodies is missing

vigoux:
[...]

PR #264, opened 2020-08-06
feat(predicates): is predicate

steelsojka:
Adds the `is?` predicate.

This will match against `kinds` defined in the locals queries.

test.js
```
function(a) {
  console.log(a);
}
```
locals.scm
```
(formal_parameters (identifier) @definition.parameter)
```
query
```
(
  (identifier) @id
  (is? @id parameter)
)
```

This would match the two identifiers with text 'a', but not the `console` identifier.

This is how I interpreted this feature. Let me know if I got this wrong.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -59,6 +59,8 @@ local function iter_cmd_sync(cmd_list)
 end
 
 local function run_install(cache_folder, package_path, lang, repo, with_sync)
+  parsers.reset_cache()
+
   local project_name = 'tree-sitter-'..lang
   local project_repo = cache_folder..'/'..project_name
   -- compile_location only needed for typescript installs.
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -284,11 +284,25 @@ function M.get_parser_configs()
   return M.list
 end
 
+local parser_files
+
+function M.reset_cache()
+  parser_files = setmetatable({}, {
+    __index = function(tbl, key)
+      rawset(tbl, key, api.nvim_get_runtime_file('parser/' .. key .. '.*', false))
+      return rawget(tbl, key)
+    end
+  })
+end
+
+M.reset_cache()
+
 function M.has_parser(lang)
   local buf = api.nvim_get_current_buf()
   local lang = lang or M.get_buf_lang(buf)
+
   if not lang or #lang == 0 then return false end
-  return #api.nvim_get_runtime_file('parser/' .. lang .. '.*', false) > 0
+  return #parser_files[lang] > 0
 end
 
 function M.get_parser(bufnr, lang)


lua/nvim-treesitter/query_predicates.lua
```diff
@@ -1,8 +1,19 @@
 local utils = require'nvim-treesitter.utils'
 local ts_utils = require'nvim-treesitter.ts_utils'
 
+local _ts_locals
 local M = {}
 
+-- Lazily have to load this to avoid circular dependencies.
+-- Cache it once it's loaded to a
+local function get_locals_api()
```
steelsojka, line 9: Not sure how else to handle this one without moving a bunch of stuff around. Let me know if there might be a better way.

theHamsta: I guess you can just `require` here performance implications will only be minimal. Require only runs the required file once: https://www.lua.org/pil/8.1.html

But for me this solution is also fine.

lua/nvim-treesitter/query_predicates.lua
```diff
@@ -108,6 +119,33 @@ M['has-ancestor?'] = function(query, match, pred)
...
 end
 
+M['is?'] = function(query, match, pred, bufnr)
+  if #pred ~= 3 then error("is? must have exactly two arguments!") end
+  local node = get_node(query, match, pred[2])
+  local def_type = pred[3]
+  if not node then return true end
+
+  local ts_locals = get_locals_api()
+  local definitions = ts_locals.get_definitions(bufnr)
```
theHamsta, line 129: why not `find_definition`?

vigoux:
Switching that to a draft as bugs were found.

lua/nvim-treesitter/query_predicates.lua
steelsojka: I think I misunderstood the original use case of this. So this would find the definition of the node and if the node is of the type specified, then it would match?

steelsojka:
I think I got a better working implementation, but there's an issue with finding the correct definition that throws the results off. I'm gonna investigate that one.

lua/nvim-treesitter/query_predicates.lua
theHamsta: this was my understanding. What was yours?

steelsojka: Not sure honestly lol. But now I think I got it!

steelsojka:
I think this also fixes an issue with find_definition #142 . I will retest and close the issue if this fixes it.

lua/nvim-treesitter/query_predicates.lua
theHamsta: Independent of the meaning. There should not be a stack overflow on any input.

steelsojka:
@theHamsta With the automatic inverse predicate of this, it ends up being `not-is?`. Is this ok or do we need some alternative way to do inverse predicates?

vigoux:
It is perfect.

theHamsta:
I've already thought about inverting the word order but `is-adjacent-not` sound strange.

lua/nvim-treesitter/locals.lua
```diff
@@ -177,7 +195,7 @@ function M.containing_scope(node, bufnr)
     iter_node = iter_node:parent()
   end
 
-  return iter_node or node
+  return iter_node or nil
```
steelsojka, line 198: I needed this change in order to be able to determine between a scope being found verses the node being a scope. I update all usages to account for this.

lua/nvim-treesitter/query_predicates.lua
steelsojka: This should be fixed now.

lua/nvim-treesitter/locals.lua
tjdevries: You could add third parameter `allow_scope` or something similar. if `nil` or `true`, then you return the `node`, else you just retun `nil` at the end.

tjdevries: (and you could call it with `false` passed)

steelsojka: I like this. I'll update it.

steelsojka:
Note, in order for this to be used with highlight queries, we need this pr merged in upstream. https://github.com/neovim/neovim/pull/12639

PR #273, opened 2020-08-09
configs: actually enable and disable on setup

vigoux:
Attempt to fix #272 .

PR #278, opened 2020-08-11
pass full module path in recurse accumulator to enable/disable sub mo…

laxman20:
Hello. I was experiencing the issue #273 attempted to fix. The fix worked for the top level modules (highlight and textobjects) but not any submodules (all the refactoring ones). I was able to fix it for the submodules as well.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -279,13 +279,13 @@ function M.setup(user_data)
     else
       config.modules[name] = vim.tbl_deep_extend('force', config.modules[name] or {}, data)
 
-      recurse_modules(function(mod_name)
+      recurse_modules(function(_, _, new_path)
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/configs.lua
```diff
@@ -279,13 +279,13 @@ function M.setup(user_data)
     else
       config.modules[name] = vim.tbl_deep_extend('force', config.modules[name] or {}, data)
 
-      recurse_modules(function(mod_name)
+      recurse_modules(function(mod_name, module, new_path)
```
steelsojka, line 282: Since `module` is unused can you change it to `_`

laxman20: Done. I'm new to lua, but I think I can apply the same change to `mod_name` as well?

PR #281, opened 2020-08-11
perf: cache parser file list for performance

vigoux:
This speeds up starting time by quite a lot.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -284,11 +284,25 @@ function M.get_parser_configs()
   return M.list
 end
 
+local parser_files
+
+function M.reset_cache()
+  parser_files = setmetatable({}, {
+    __index = function(tbl, key)
+      rawset(tbl, key, api.nvim_get_runtime_file('parser/' .. key .. '.*', false))
+      return rawget(tbl, key)
+    end
+  })
+end
+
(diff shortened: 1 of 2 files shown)


steelsojka:
Do we want a method to invalidate this cache? What after a user runs `TSInstall lua`. Wouldn't this be cached to have no parsers?

vigoux:
Maybe we can invalidate the cache in this case

vigoux:
I just started to think about it, it will be fine if someone installs or updates a parser, as we are only caching file paths and actual parser.

vigoux:
Invalidating the cache is done every time a parser is installed, but not when it is updated, as this does not change the path of the parser.

lua/nvim-treesitter/parsers.lua
```diff
@@ -284,11 +284,25 @@ function M.get_parser_configs()
...
+function M.reset_cache()
+  parser_files = setmetatable({}, {
+    __index = function(tbl, key)
+      rawset(tbl, key, api.nvim_get_runtime_file('parser/' .. key .. '.*', false))
+      return rawget(tbl, key)
+    end
+  })
+end
+
+M.reset_cache()
```
kyazdani42, line 298: should we call this `init_parser_cache()` ? i'm not too sure, maybe it doesn't really matter.

PR #283, opened 2020-08-11
feat: intuitive runtime queries

vigoux:
Starting now, runtime queries will be sourced in this order :

- Queries that are not in any `after` folder, will serve as a base, with
each occurence overwriting the others (that is, .config/nvim/queries has
the highest priority)
- Queries within the `after` directory will be sourced one after the
other.

The rationale is that this reminds all the `.vim` files (ftplugin) for
example, and this allows both to experiment and to override queries
easily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,10 +281,21 @@ List of currently supported languages:
 
 ## User Query Extensions
 
-You can add your own query files by placing a query file in vim's runtime path after `nvim-treesitter` is sourced.
-If the language has a built in query file, that file will be appended to or it will be used (useful for languages not yet supported).
-For example, you can add files to `<vim-config-dir>/after/queries/lua/highlights.scm` to add more queries to lua highlights.
-You can also manually add query paths to the runtime path by adding this to your vim config `set rtp+='path/to/queries'`.
+Queries are what `nvim-treesitter` uses to extract informations from the syntax tree, and they are
+located in the `queries/{lang}/*` runtime directories (like the `queries` folder of this plugin).
+
+`nvim-treesitter` considers queries as any runtime file (see `:h rtp`), that is :
+
+- if the file is in any `after/queries/` folder, then it will be used to extend the already defined
+  queries.
+- Otherwise, it will be used as a base to define the query, the first query found (with the highest
+  priority) will be the only one to be used.
+
+This hybrid approach is the most standard way, and according to that, here is some ideas on how to
+use is :
+- If you want to rewrite (or write) a query, don't use `after/queries`.
+- If you want to override a part of a query (only one match for example), use the `after/queries`
+  directory.
 
 ## Troubleshooting
 
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -104,15 +104,21 @@ By default, everything is disabled. To enable support for features, in your `ini
 ==============================================================================
 USER QUERY EXTENSIONS			    *nvim-treesitter-query-extensions*
 
-You can add your own query files by placing a query file in vim's runtime path
-after `nvim-treesitter` is sourced.  If the language has a built in query file,
-that file will be appended to or it will be used (useful for languages not yet
-supported).
+Queries are what `nvim-treesitter` uses to extract informations from the syntax tree, and they are
+located in the `queries/{lang}/*` runtime directories (like the `queries` folder of this plugin).
 
-For example, you can add files to `<vim-config-dir>/after/queries/lua/highlights.scm`
-to add more queries to lua highlights.  You can also manually add query paths
-to the runtime path by adding this to your vim config `set rtp+='path/to/queries'`.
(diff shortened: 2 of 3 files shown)


lua/nvim-treesitter/query.lua
```diff
@@ -69,6 +69,28 @@ function M.get_matches(bufnr, query_group)
   return query_cache[query_group][bufnr].cache
 end
 
+local function filter_files(file_list)
+  local main = {}
+  local after = {}
+
+  for _, fname in ipairs(file_list) do
+    if string.find(fname, "after", 1, true) then
```
theHamsta, line 77: Is the occurence of "after" really the only possibility to detect after directories?

vigoux: Hehe that was a bit hacky I admit

vigoux: It should be better now, still needs documenting.

kyazdani42:
Looks good, but shouldn't queries be merged instead of replaced ?

vigoux:
Should be all good now

theHamsta:
@vigoux  Will this change automatically overwrite future built-in queries by Neovim? Don't plugin managers append to rtp path? It could potentially go wrong when we update the c/c++ but c queries lack behind and we get a old/new c-c++ chimera

vigoux:
Built-in queries in neovim is blocked for some time, waiting for us to find a proper solution to the overwriting problem (thus this pr here).

Regarding the version problem, I think that pining the parsers version and updating everything at once will help us keeping things stable.

theHamsta:
> Regarding the version problem, I think that pining the parsers version and updating everything at once will help us keeping things stable.

That's not the problem. We can pin our parsers and queries. But problems start when we have parsers and queries in different places (repos that develop independently from each other).

Goal should be: having the latest and greatest with our plugin installed and falling back to stable default with built-in.

vigoux:
Oh I see what you mean now.

I guess it is parsers responsibility to keep backward compatibility. But we can try (in a later PR) to have a fallback mechanism at some point.

doc/nvim-treesitter.txt
```diff
@@ -104,15 +104,21 @@ By default, everything is disabled. To enable support for features, in your `ini
...
+Queries are what `nvim-treesitter` uses to extract informations from the syntax tree, and they are
+located in the `queries/{lang}/*` runtime directories (like the `queries` folder of this plugin).
 
-For example, you can add files to `<vim-config-dir>/after/queries/lua/highlights.scm`
-to add more queries to lua highlights.  You can also manually add query paths
-to the runtime path by adding this to your vim config `set rtp+='path/to/queries'`.
+`nvim-treesitter` considers queries as any runtime file (see `:h rtp`), that is :
 
+- if the file is in any `after/queries/` folder, then it will be used to extend the already defined
+  queries
```
theHamsta, line 113: dot: `.`

doc/nvim-treesitter.txt
```diff
@@ -104,15 +104,21 @@ By default, everything is disabled. To enable support for features, in your `ini
...
+- Otherwise, it will be used as a base to define the query, the first query found (with the highest
+  priotiry) will be the only one to be used.
```
theHamsta, line 115: typo

doc/nvim-treesitter.txt
```diff
@@ -104,15 +104,21 @@ By default, everything is disabled. To enable support for features, in your `ini
...
+
+This hybrid approach is the most standard way, and according to that, here is some ideas on how to
+use is :
+- If you want to rewrite (or write) a query, don't use `after/queries`.
+- If you want to override a part of a query (only one match for example), use the `after/queries`
```
theHamsta, line 120: I guess override is not really possible with `after`. Only extensions/additions, right?

README.md
```diff
@@ -281,10 +281,21 @@ List of currently supported languages:
...
-You can also manually add query paths to the runtime path by adding this to your vim config `set rtp+='path/to/queries'`.
+Queries are what `nvim-treesitter` uses to extract informations from the syntax tree, and they are
+located in the `queries/{lang}/*` runtime directories (like the `queries` folder of this plugin).
+
+`nvim-treesitter` considers queries as any runtime file (see `:h rtp`), that is :
+
+- if the file is in any `after/queries/` folder, then it will be used to extend the already defined
+  queries
+- Otherwise, it will be used as a base to define the query, the first query found (with the highest
+  priotiry) will be the only one to be used.
```
theHamsta, line 292: see comments for doc.txt

doc/nvim-treesitter.txt
vigoux: Override is possible if you just use the very same match.

theHamsta: Can you give an example? Like deactivting a query for @foo by repeating it with @foo?

Issue #282, opened 2020-08-11
Usage highlighting does not work with variable shadowing


steelsojka:
**Describe the bug**

When variables are defined with the same name in a lower scope usage highlighting highlights them even though they aren't correct usages.

![Screen Shot 2020-08-11 at 12 59 09 PM](https://user-images.githubusercontent.com/2062154/89932137-c6043100-dbd2-11ea-9ca0-d8a1c765ee43.png)

**To Reproduce**
Use this code:

```javascript
import { test } from 'test';

const a = 'test';

function(a) {
  console.log(a);

  return function(a) {
    return a;
  }
}

a = 'test'
```

Put cursor on the first `a` definition

**Expected behavior**
Only `a = 'test'` should have it's usage highlighted

**Output of `:checkhealth nvim_treesitter`**

<details>
<code>

Paste the output here

</code>
</details>

**Output of `nvim --version`**
```
Paste your output here
```

**Additional context**

steelsojka:
I've looked into this issue and it seems a hard one to solve. We need to determine the definition for each reference and see if it's the same as the definition of the node we have highlighted. This get's VERY expensive since `find_definition` has to check each definition and it's scope.

vigoux:
I think we need to to some kind of sprint focusing on locals, definitions and so on.
Starting from scratch may help us to have better results afterwards.

steelsojka:
I have a refactor that may help here with definitions and references... I'll see how it turns out.

theHamsta:
A note for this: two definitions in same scope should probably not count as shadowing (or should it?)

In python and rust you can do:
```python
a = 1
a = a + 1
```
```rust
let a = 1;
let a = a + 1;
```

vigoux:
I guess that the problem occurs for Python only, because. the `let` keyword in Rust really means shadowing here.

theHamsta:
In Rust, it should probably count as shadowing.

Maybe I can add "set!" attribute to show @steelsojka that a python "definition" will not shadow if it's in same scope (except `global`, `nonlocal` is used)

steelsojka:
I'll have to experiment with that one. I'm gonna test this with #284 and see if the issue still exists... I'm curious how it will behave.

steelsojka:
@theHamsta So would the definition be the one that occurs first? If definitions collide within the same scope we could just take the one that occurs first.

vigoux:
T would say take the closest before the current reference (in the current scope of course...).

[...]

PR #284, opened 2020-08-12
fix(definitions): optimize and fix definition highlighting

steelsojka:
This will fix https://github.com/nvim-treesitter/nvim-treesitter/issues/282

This will also improve definition performance on successive definition and usage lookups.


lua/nvim-treesitter/locals.lua
```diff
@@ -158,6 +140,42 @@ function M.recurse_local_nodes(local_def, accumulator, full_match, last_match)
   end
 end
 
+--- Get a single dimension table to look definition nodes by scope and text.
+-- This is memoized by buffer tick to avoid computing on every lookup.
+M.get_defintions_lookup_table = ts_utils.memoize_by_buf_tick(function(bufnr)
```
vigoux, line 145: Little typo here

vigoux: Furthermore, I don't see the point of exposing that if we don't expose the `get_definition_id` too.

vigoux: Last comment I promise, can you add a bit more doc regarding what represents the return value of this function, and maybe the idea behind it ? I don't find that especially easy to get.

steelsojka: Yeah good point. I will update this.

theHamsta: and it has a typo

lua/nvim-treesitter/locals.lua
```diff
@@ -24,6 +24,16 @@ function M.get_locals(bufnr)
...
 end
 
+--- Creates unique id for a node based on text and range
+-- @param scope: the scope node of the definition
+-- @param bufnr: the buffer
+-- @param node_text: the node text to use
+-- @returns a string id
+function M.get_definition_id(scope, node_text)
+  -- Add a vaild starting character in case node text doesn't start with a valid one.
+  return table.concat({ 'k', node_text, scope:range() }, '_')
```
theHamsta, line 34: why not node.id? Is node.id not unique when compared over time?

theHamsta:
Nice work. This is no fast enough to run on `CursorMoved`. Maybe we can offer that as an option?

lua/nvim-treesitter/locals.lua
steelsojka: Honestly I didn't even nodes had IDs. I don't think I saw it in the docs. I'm also not 100% sure if the node id will be the same across different trees...

theHamsta: I guess the ID should change on reparsing. I am however not sure. You could  test it.

PR #287, opened 2020-08-13
Fix #286: Highlight C++ raw_string_literal

### LINKED ISSUE #286: Raw string literals should be treated the same as string literals while highlighting in c++
**Describe the bug**
Raw string literals get a different color than string literals. 

**To Reproduce**
Screen Shot 2020-08-13 at 12 27 30 AM

**Expected behavior**
Both the string literal and raw string literal should be green in color. 

<details>
<code>

## cpp parser healthcheck
  - OK: cpp parser found.
  - OK: `highlights.scm` found.
  - OK: `locals.scm` found.
  - OK: `textobjects.scm` found.

</code>
</details>

```
NVIM v0.5.0-627-g94b7ff730
Build type: Debug
LuaJIT 2.1.0-beta3
```

**Additional context**
I played around with this a bit on the [treesitter playground](https://tree-sitter.github.io/tree-sitter/playground), and it correctly identifies the string literal and raw string literal.

theHamsta:


theHamsta:
btw. should be also added to upstream.

queries/cpp/highlights.scm
```diff
@@ -108,3 +108,5 @@
 
 "::" @operator
 "..." @operator
+
+(raw_string_literal)  @string
```
vigoux, line 112: Can that be in the `Constants` section of the file ?

Issue #280, opened 2020-08-11
Local variable use 'Identifier' as group highlight.


PangPangPangPangPang:
In the latest version, it seems like local variable se 'Identifier' as group highlight. It looks like as below, but previous version looks normal.
![image](https://user-images.githubusercontent.com/8207475/89882407-5f463e00-dbf9-11ea-8930-abc1e4d55448.png)

vigoux:
Hi, what is the issue here ?

PangPangPangPangPang:
> Hi, what is the issue here ?

I think local variable should use default white color.

vigoux:
Then I highly suggest adding the following to your config :
```
highlight! link TSVariable Normal
```

I think that this is more a colorscheme issue tbh.

theHamsta:
I guess most colorscheme use a very aggressive highlighting for Identifier. We had a discussion over that before. Back then, we decided to use Normal over Identifier. I guess we have to leave that to user configuration as there's no easy way to find a compromise.

PangPangPangPangPang:
> I guess most colorscheme use a very aggressive highlighting for Identifier. We had a discussion over that before. Back then, we decided to use Normal over Identifier. I guess we have to leave that to user configuration as there's no easy way to find a compromise.

Yes. I test several colorscheme, all of them can't distinguish between variables and attributes.

vigoux:
Do you want us to do a specific action regarding this issue, or do you think leaving that to user configuration is fine ?

PangPangPangPangPang:
> Do you want us to do a specific action regarding this issue, or do you think leaving that to user configuration is fine ?

I think the alternative configuration of the plug-in should be friendly to most users. For me, I can accept two treatments

theHamsta:
I would also prefer returning to the old defaults. Simply because highlighting means givening different things different colors. Would be a pity if all the other stuff with Identifier would get lost.

vigoux:
The old default meaning `Normal` here ?

PangPangPangPangPang:
Maybe `None` is better.

PR #288, opened 2020-08-13
fix(highlight): link TSVariable to Normal

vigoux:
After some discussion in #280 , we came to the conslusion that `Normal` should be use to highlight variables instead of `Identifier`.

@akinsho , as you were the previous commiter there, can you give us your opinion on this too ?

### DIFF
--- a/plugin/nvim-treesitter.vim
+++ b/plugin/nvim-treesitter.vim
@@ -1,4 +1,4 @@
-" Last Change: 2020 Jun 29
+" Last Change: 2020 Aug 13
 
 if exists('g:loaded_nvim_treesitter')
   finish
@@ -62,7 +62,7 @@ highlight default link TSTypeBuiltin Type
 highlight default link TSStructure Structure
 highlight default link TSInclude Include
 
-highlight default link TSVariable Identifier
+highlight default link TSVariable Normal
 highlight default link TSVariableBuiltin Special
 
 highlight default link TSDefinitionUsage Visual


akinsho:
@vigoux I think being able to highlight variables in certain contexts is pretty useful. I was primarily trying to add highlighting for assignments to dart and saw that we were using `@variable` in some other queries but it wasn't attached to any highlight. I actually think the case in #280 (highlighting local variables is pretty useful, and was going to be one of the advantages of treesitter) but since maybe it's a matter of personal taste I'm fine with the default being hidden and you can set it yourself etc 🤷🏿

theHamsta:
@akinsho it's becaus treesitter's @variable is our @identifier

theHamsta:
Can we add @variable to all languages? We could open a issue for that.

vigoux:
Please be careful and quote when you talk about capture groups, we will turn some people crazy otherwise.

akinsho:
I definitely agree that `Identifier` was probably not the right highlight there but the base highlights don't seem like enough (range of colors 🤷🏿 ) from what I can see.

IMO think `Identifier` is overused since in a lot of filetypes where I use treesitter most things are `red` which is my colorscheme's highlight for identifier which is jarring. I definitely think having `@variable` highlights is useful but not sure what default vim highlight group would be good for that. I've overwritten the color for those in my config since it was a bit much.

vigoux:
The default highlight for something is `Normal`, the thing is that variables are likely the most common thing in program (I mean... I am sure they are the most likely), thus, they should be highlighted in a neutral way, and `Normal` is this neutral way.

PR #313, opened 2020-08-16
Add highlights for OCaml.

undu:
It's missing function usages and arguments, but they are represented by the same type of node in the same list, it wasn't clear to me how to match the first item of the list only, and the others.

For the rest it should be pretty complete but I might have missed some advanced syntax, especially regarding types.

### DIFF
--- a/queries/ocaml/hightlights.scm
+++ b/queries/ocaml/hightlights.scm
@@ -1,0 +1,147 @@
+(module_name) @structure
+
+[
+  (module_type_name)
+  (type_constructor)
+  (type_variable)
+  (constructor_name)
+  (tag)
+] @type
+(type_binding (type_constructor) @type)
+
+(exception_definition (constructor_declaration (constructor_name) @exception))
+(application_expression (value_path (value_name) @_raise)
+                        (#eq? @_raise "raise")
+                        (constructor_path) @exception)
+(application_expression (value_path (value_name) @_raise)
+                        (#eq? @_raise "raise")
+                        (parenthesized_expression
+                        (application_expression (constructor_path) @exception)))
+(infix_expression (value_path (value_name) @_raise)
+                  (#eq? @_raise "raise")
+                  (application_expression (constructor_path) @exception))
+
+(attribute_id) @property
+(field_name) @field
+
+(let_binding
+  (value_name) @function
+  (parameter))
+(let_binding
+  (value_name) @function
+  (function_expression))
+
+(application_expression . (value_path (value_name) @function))
+(application_expression . (value_path (value_path (value_name) @function)))
+(application_expression . (value_path (value_path (value_path (value_name) @function))))
+(value_specification (value_name) @function (function_type))
+(infix_expression (value_path (value_name) @function)
+                  (infix_operator) @operator
+                  (#eq? @operator "@@"))
+(infix_expression (value_path (value_path (value_name) @function))
+                  (infix_operator) @operator
+                  (#eq? @operator "@@"))
+(infix_expression (value_path (value_path (value_path (value_name) @function)))
+                  (infix_operator) @operator
+                  (#eq? @operator "@@"))
+(infix_expression (infix_operator) @operator
+                  (#eq? @operator "|>")
+                  (value_path (value_name) @function))
+(infix_expression (infix_operator) @operator
+                  (#eq? @operator "|>")
+                  (value_path (value_path (value_name) @function)))
+(infix_expression (infix_operator) @operator
+                  (#eq? @operator "|>")
+                  (value_path (value_path (value_path (value_name) @function))))
+
+
+(let_binding (parameter (label (label_name) @parameter)))
+(let_binding (parameter (label_name) @parameter))
+(let_binding (parameter (value_name) @parameter))
+(let_binding (parameter (typed_pattern (value_name) @parameter)))
+(function_type (typed_label (label_name) @parameter))
+
+;; Literals
+(unit) @constant
+(boolean) @boolean
+(number) @number
+(character) @character
+
+(comment) @comment
+
+[(string) (quoted_string)] @string
+(escape_sequence) @string.escape
+(conversion_specification) @string.special
+
+(infix_operator) @operator
+
+[
+  (let_operator)
+  (and_operator)
+  "let"
+  "rec"
+  "nonrec"
+  "of"
+  "in"
+  "type"
+  "val"
+  "and"
+  "struct"
+  "sig"
+  "functor"
+  "external"
+  "module"
+  "match"
+  "with"
+  "when"
+  "begin"
+  "end"
+  "fun"
+  "function"
+  "assert"
+  "lazy"
+  "as"
+  "method"
+  "new"
+  "object"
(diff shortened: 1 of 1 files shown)


steelsojka:
@undu If you want to match the first item in a list you would use the `.` operator. For example:

```scheme
(formal_parameters . (identifier))
```

This would match the first parameter of a list of parameters.

undu:
Thanks for the tip, now function usages with module paths paths up to 3 components are highlighted as such.

I'd prefer a way to generalize the highlighting of those for paths with more elements, but module paths with more than 3 nested modules are exceedingly rare.

I've also added some more elements for types, fields and separators; I've also removed a couple highlight groups which were too greedy.

queries/ocaml/hightlights.scm
```diff
@@ -0,0 +1,146 @@
...
+  "val"
+  "and"
+  "struct"
+  "sig"
+  "functor"
+  "external"
+  "module"
+  "match"
+  "with"
+  "when"
```
theHamsta, line 97: Is this a conditional?

queries/ocaml/hightlights.scm
```diff
@@ -0,0 +1,146 @@
...
+(infix_expression (infix_operator) @operator
+                  (#eq? @operator "|>")
+                  (value_path (value_name) @function))
+(infix_expression (infix_operator) @operator
+                  (#eq? @operator "|>")
+                  (value_path (value_path (value_name) @function)))
+(infix_expression (infix_operator) @operator
+                  (#eq? @operator "|>")
+                  (value_path (value_path (value_path (value_name) @function))))
+
```
theHamsta, line 56: This is not ideal. But I don't see a better solution for now since it would require combining two queries. A predicate for this situation would need to be really specialized. Can we extend the has-ancenstor to accept a Lua function as a predicate for the parent? Or a predicate just for that situation?

queries/ocaml/hightlights.scm
```diff
@@ -0,0 +1,146 @@
+(module_name) @structure
```
theHamsta, line 1: Structure? I dont think we have this highlight (too lazy to check on mobile). Maybe `@namespace`?

TravonteD: We do have a structure highlight `TSStructure`

queries/ocaml/hightlights.scm
```diff
@@ -0,0 +1,146 @@
...
+  (module_type_name)
+  (type_constructor)
+  (type_variable)
+  (constructor_name)
+  (tag)
+] @type
+(type_binding (type_constructor) @type)
+
+(exception_definition (constructor_declaration (constructor_name) @exception))
+(application_expression (value_path (value_name) @_raise)
```
TravonteD, line 13: What is the purpose of the `@_raise` query?

queries/ocaml/hightlights.scm
```diff
@@ -0,0 +1,146 @@
...
+(field_name) @field
+
+(let_binding
+  (value_name) @function
+  (parameter))
+(let_binding
+  (value_name) @function
+  (function_expression))
+
+(application_expression . (value_path (value_name) @function))
```
TravonteD, line 34: It looks like you set `(value_path (value_name) @function)` everywhere, but above where it's `@_raise` you can shortcut this by extracting it and having it at the top-level. The highlighter will then default it to `@function` unless a more specific query is used.

queries/ocaml/hightlights.scm
```diff
@@ -0,0 +1,146 @@
...
+ "#"
+ ":"
+ "="
+ "|"
+ "->"
+ "::"
+ "~"
+ "?"
+ "|"
+ "+"
```
TravonteD, line 138: the plus, minus, star, etc. should probably be highlighted as `@operator`

queries/ocaml/hightlights.scm
```diff
@@ -0,0 +1,146 @@
...
+(infix_expression (value_path (value_name) @_raise)
+                  (#eq? @_raise "raise")
+                  (application_expression (constructor_path) @exception))
+
+(attribute_id) @property
+(field_name) @field
+
+(let_binding
+  (value_name) @function
+  (parameter))
```
TravonteD, line 29: Should this be highlighted as `@parameter`?

queries/ocaml/hightlights.scm
undu: They are used for type annotations as well, I need to add context for these.

queries/ocaml/hightlights.scm
undu: I don't think so: only the `value_name` in the `parameter` subtree should be captured as a parameter. This is done elsewhere.

queries/ocaml/hightlights.scm
undu: The problem is that `(value_path (value_name))` matches any value, function or not. So I have to depend on the outermost context. I've raised the issue https://github.com/tree-sitter/tree-sitter-ocaml/issues/40 to prevent this awkwardness.

queries/ocaml/hightlights.scm
undu: `@_raise` is used to capture the function `raise`, which is always followed by a constructor of an exception. This is because `raise` is a normal function (and value) instead of a keyword.

queries/ocaml/hightlights.scm
undu: I could use either, but structure works.

queries/ocaml/hightlights.scm
undu: I've raised an issue to the ocaml syntax, I think the tree structure can be changed to avoid this uglyness. https://github.com/tree-sitter/tree-sitter-ocaml/issues/40

queries/ocaml/hightlights.scm
undu: Good catch!

theHamsta (approved):

queries/ocaml/hightlights.scm
theHamsta: Namespace does not exist yet. Structure was this what nobody knew what it actually is. Now I remember :-)

theHamsta:
You can also be the maintainer in the README if you want to

queries/ocaml/hightlights.scm
```diff
@@ -0,0 +1,147 @@
...
+  "functor"
+  "external"
+  "module"
+  "match"
+  "with"
+  "begin"
+  "end"
+  "fun"
+  "function"
+  "try"
```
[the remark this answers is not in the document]
vigoux: Agreed on this, @undu is this exception related ?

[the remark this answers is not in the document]
undu: yes, it's part of the `try ... with <capture exception> -> ...` block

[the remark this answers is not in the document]
vigoux: Just `@exception`

vigoux (approved):

vigoux:
Awesome ! Thanks again for your contribution !

PR #321, opened 2020-08-17
fix issue with typescript parser definition

kyazdani42:
remove the used_by key that has no meaning in this context except create bugs.


kyazdani42:
my bad forgot typescript.tsx was another filetype possibility for tsx.

PR #325, opened 2020-08-19
Highlight operators + improve consistency with Vim syntax

Avi-D-coder:


queries/rust/highlights.scm
```diff
@@ -97,7 +97,7 @@
 
 (lifetime (identifier) @label)
 
-(self) @variable.builtin
+(self) @keyword
```
kyazdani42, line 100: i believe self should stay a builtin variable. If you really want this to be a keyword, you could override it in one of your own query files :)

kyazdani42 (changes requested):

queries/rust/highlights.scm
theHamsta: Or link variable.builtin to keyword. It's mostly `this` and `self`

queries/rust/highlights.scm
```diff
@@ -165,8 +165,42 @@
...
+"-"
+"-="
+"+"
+"+="
+"/"
+"/="
+">"
+"<"
+">="
+"<="
```
TravonteD, line 197: This is duplicated from line 171
```suggestion
```

queries/rust/highlights.scm
Avi-D-coder: Fair enough. I opened #327 in this regard.

kyazdani42 (approved):
looks good now :) thanks

PR #336, opened 2020-08-22
feat(refactor.navigation): add navigation.goto_{next,previous}_usage

theHamsta:
I like `*` and `#` to consider only other usages of the identifier under cursor (if it's an identifier). Overwriting the default mappings is however not a good idea. Any suggestion what default key mapping we could use for this? We could also not set any default mappings.

TODO:

 - [x] docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,9 @@ EOF
 
 Provides "go to definition" for the symbol under the cursor,
 and lists the definitions from the current file.
+goto_next_usage/goto_previous_usage go to the next usage of the
+identifier under the cursor.
+
 
 ```lua
 lua <<EOF
@@ -195,6 +198,8 @@ require'nvim-treesitter.configs'.setup {
       keymaps = {
         goto_definition = "gnd",
         list_definitions = "gnD",
+        goto_next_usage = "<a-*>",
+        goto_previous_usage = "<a-#>",
       },
     },
   },
--- a/lua/nvim-treesitter/refactor/navigation.lua
+++ b/lua/nvim-treesitter/refactor/navigation.lua
@@ -12,14 +12,10 @@ function M.goto_definition(bufnr)
   local bufnr = bufnr or api.nvim_get_current_buf()
   local node_at_point = ts_utils.get_node_at_cursor()
 
-  utils.set_jump()
-
   if not node_at_point then return end
 
   local definition, _ = locals.find_definition(node_at_point, bufnr)
-  local start_row, start_col, _ = definition:start()
-
-  api.nvim_win_set_cursor(0, { start_row + 1, start_col })
+  ts_utils.goto_node(definition)
 end
 
 function M.list_definitions(bufnr)
@@ -47,6 +43,25 @@ function M.list_definitions(bufnr)
   vim.fn.setqflist(qf_list, 'r')
   api.nvim_command('copen')
 end
+
+function M.goto_adjacent_usage(bufnr, delta)
+  local bufnr = bufnr or api.nvim_get_current_buf()
+  local node_at_point = ts_utils.get_node_at_cursor()
+  if not node_at_point then return end
+
+  local def_node, scope = locals.find_definition(node_at_point, bufnr)
+  local usages = locals.find_usages(def_node, scope, bufnr)
+
+  local index = utils.index_of(usages, node_at_point)
+  if not index then return end
+
+  local target_index = (index + delta + #usages - 1) % #usages + 1
+  ts_utils.goto_node(usages[target_index])
+  return usages[target_index]
+end
+
+function M.goto_next_usage(bufnr) return M.goto_adjacent_usage(bufnr, 1) end
+function M.goto_previous_usage(bufnr) return M.goto_adjacent_usage(bufnr, -1) end
 
 function M.attach(bufnr)
   local config = configs.get_module('refactor.navigation')
--- a/lua/nvim-treesitter/ts_utils.lua
+++ b/lua/nvim-treesitter/ts_utils.lua
@@ -276,4 +276,24 @@ function M.swap_nodes(node_or_range1, node_or_range2, bufnr, cursor_to_second)
   end
 end
 
+function M.goto_node(node, goto_end, avoid_set_jump)
+  if not node then return end
+  if not avoid_set_jump then
+    utils.set_jump()
+  end
+  local range = {node:range()}
+  local position
+  if not goto_end then
+    position = { range[1], range[2] }
+  else
+    -- ranges are exclusive: -1 character!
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -195,6 +198,8 @@ require'nvim-treesitter.configs'.setup {
       keymaps = {
         goto_definition = "gnd",
         list_definitions = "gnD",
+        goto_next_usage = "<a-*>",
+        goto_previous_usage = "<a-#>",
```
vigoux, line 202: Can use use `Shift` here instead ? just to be q bit less opinionated 😉

lua/nvim-treesitter/utils.lua
```diff
@@ -71,4 +71,12 @@ function M.set_jump()
...
 end
 
+function M.index_of(tbl, obj)
+  for i, o in ipairs(tbl) do
+    if o == obj then
+      return i
+    end
+  end
+end
+
```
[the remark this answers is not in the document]
vigoux: Isn't there a `vim.tbl_*`thing for that ?

README.md
theHamsta: On my German keyboard "*" already implies shift. I guess also on US keyboards.

vigoux: Oh let's leave it like that then

theHamsta:
@vigoux there's *vim.tbl_add_reverse_lookup()* but that mutates the table.

lua/nvim-treesitter/utils.lua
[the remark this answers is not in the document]
theHamsta: *vim.tbl_add_reverse_lookup()* but it mutates the table.

README.md
theHamsta: I really don't care about the defaults. Any suggestion is ok for me.

We're already using scheme-like syntax, don't you realize that nvim is slowly becoming emacs? We should  set <c-r><c-a><m-*> :stuck_out_tongue:

lua/nvim-treesitter/utils.lua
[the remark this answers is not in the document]
theHamsta: I guess the reason why there is not a helper for that is that it almost always hints for the mistake that you switched keys and values. It will always be inefficient. But if we get usages as an ordered list it should be ok.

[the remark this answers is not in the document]
vigoux: I think that they should be ordered

lua/nvim-treesitter/refactor/navigation.lua
```diff
@@ -48,6 +44,26 @@ function M.list_definitions(bufnr)
...
+  local index = utils.index_of(usages, node_at_point)
+  if not index then return end
+
+  local target_index = (index + delta + #usages - 1) % #usages + 1
+  ts_utils.goto_node(usages[target_index])
+  return usages[target_index]
+end
+
+function M.goto_next_usage(bufnr) return M.goto_adjacent_usage(bufnr, 1) or vim.cmd('normal! *') end
+function M.goto_previous_usage(bufnr) return M.goto_adjacent_usage(bufnr, -1) or vim.cmd('normal! #') end
```
stsewd, line 65: Not sure if falling back to the original `*` is something expected. Since you may believe you are jumping to the next usage when you aren't

lua/nvim-treesitter/utils.lua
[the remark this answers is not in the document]
steelsojka: The order depends on how treesitter captures them in the query. I would assume from top to bottom. IF NOT then we would have to order them ourselves 😢

lua/nvim-treesitter/refactor/navigation.lua
theHamsta: Yeah, I can remove the fall back

lua/nvim-treesitter/utils.lua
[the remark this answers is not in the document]
theHamsta: Until now everything was ordered. Should we hope that this stays like that in future?

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -276,4 +276,24 @@ function M.swap_nodes(node_or_range1, node_or_range2, bufnr, cursor_to_second)
...
+  local range = {node:range()}
+  local position
+  if not goto_end then
+    position = { range[1], range[2] }
+  else
+    -- ranges are exclusive: -1 character!
+    if range[4] == 0 then
+      position = { range[3] - 1, -1 }
+    else
+      position = { range[3], range[4] - 1 }
```
theHamsta, line 293: This wasn't done in textobjects.move but it is necessary to set cursor on last character of the node

theHamsta:
Can I add this function to this PR? A faster `vim.lsp.buf.definition`...

```
function M.goto_definition(bufnr, or_else)
  local bufnr = bufnr or api.nvim_get_current_buf()
  local node_at_point = ts_utils.get_node_at_cursor()

  if not node_at_point then return end

  local definition, _ = locals.find_definition(node_at_point, bufnr)

  if or_else and definition.id == node_at_point.id then
    or_else()
  else
    ts_utils.goto_node(definition)
  end
end

function M.goto_definition_lsp_fallback(bufnr) M.goto_definition(bufnr, vim.lsp.buf.definition) end
```

vigoux:
> Can I add this function to this PR? A faster `vim.lsp.buf.definition`...

Meh.... Another PR ?

theHamsta:
rollback...

lua/nvim-treesitter/refactor/navigation.lua
```diff
@@ -48,6 +44,26 @@ function M.list_definitions(bufnr)
   api.nvim_command('copen')
 end
 
+function M.goto_adjacent_usage(bufnr, delta)
+  local bufnr = bufnr or api.nvim_get_current_buf()
+  local node_at_point = ts_utils.get_node_at_cursor()
+
+  local def_node, scope = locals.find_definition(node_at_point, bufnr)
+
+  if not node_at_point then return end
```
vigoux, line 53:
```suggestion
  if not node_at_point then return end
  
  local def_node, scope = locals.find_definition(node_at_point, bufnr)

```

lua/nvim-treesitter/utils.lua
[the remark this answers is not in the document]
steelsojka: I think that's fair to assume.

lua/nvim-treesitter/refactor/navigation.lua
theHamsta: :+1:

vigoux (approved):

PR #339, opened 2020-08-23
fix: add query extensions lastly, so they always override

tobiwild:
I have the following override in `.vim/after/queries/javascript/jsx.scm` (to use a different highlight group for JSX tags):
```
(jsx_closing_element name: (identifier) @tag)
(jsx_opening_element name: (identifier) @tag)
(jsx_self_closing_element name: (identifier) @tag)
```
This works for Javascript but not for Typescript, as it has some entry in `M.base_language_map`. With this change it seems to work in Typescript too.

### DIFF
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -89,6 +89,11 @@ function M.get_query_files(lang, query_name)
   local lang_files = filtered_runtime_queries(lang, query_name)
   vim.list_extend(query_files, lang_files)
 
+  for _, base_lang in ipairs(M.base_language_map[lang] or {}) do
+    local base_files = filtered_runtime_queries(base_lang, query_name)
+    vim.list_extend(query_files, base_files)
+  end
+
   for _, ext in ipairs(extensions) do
     local l = lang
     local e = ext
@@ -98,11 +103,6 @@ function M.get_query_files(lang, query_name)
     end
     local ext_files = filtered_runtime_queries(l, e)
     vim.list_extend(query_files, ext_files)
-  end
-
-  for _, base_lang in ipairs(M.base_language_map[lang] or {}) do
-    local base_files = filtered_runtime_queries(base_lang, query_name)
-    vim.list_extend(query_files, base_files)
   end
 
   return query_files


vigoux:
@theHamsta wha is your opinion on this ? Does this breaks the c / cpp order ?

theHamsta:
I really wanted @kyazdani42 or @steelsojka to answer this since they refactored this piece of code explicitly for jsx.

When I wrote the code the base languages were loaded first. I don't why this was changed and how they should interact with "extensions". For the goal to have user extension always last, this PR only helps for this specific JSX case and is not a general solution.

For c++ this should have no effect since there aren't any extensions. So the change is purely jsx related.

steelsojka:
Shouldn't the order be base lang -> lang -> extensions? For example TSX I would expect (from first loaded to last, based on our current configuration).

javascript -> typescript -> javascript.jsx

Judging by these changes it's

typescript -> javascript -> javascript.jsx

And the code before these changes is

typescript -> javascript.jsx -> javascript

So, getting the base files should be moved above where we get the standard queries for the language

theHamsta:
Our original code was like:

baselang -> lang

Now it's like:

lang -> base -> extension

I guess it should be

base -> lang -> extension

I guess what @tobiwild  wants is something like this:

base -> lang -> extension -> user base -> user lang -> user extension

steelsojka:
@theHamsta I don't think that way really makes sense. You can do this currently:

base -> user base -> lang -> user lang -> extensions -> user extensions

If you wanted everything to apply after the extensions you would just add a query for the extension language.

lua/nvim-treesitter/query.lua
```diff
@@ -89,6 +89,11 @@ function M.get_query_files(lang, query_name)
   local lang_files = filtered_runtime_queries(lang, query_name)
   vim.list_extend(query_files, lang_files)
 
+  for _, base_lang in ipairs(M.base_language_map[lang] or {}) do
```
steelsojka, line 92: I think this should move above the `lang_files` so base files are loaded before the standard lang files.

tobiwild:
Issue seems to be fixed in current master, I moved the file to `.vim/after/queries/jsx/highlights.scm` and now it correctly overrides.

PR #341, opened 2020-08-25
fix(locals): add nil guard for definition id creation

steelsojka:


vigoux (approved):

vigoux:
Can you link the Zulip discussion in this pr too?

theHamsta:
I removed the setting to nil in the last day and didn't observe any problems. So I guess this should fix the problem. Thanks for submitting!

theHamsta:
Here is the link: https://nvim-treesitter.zulipchat.com/#narrow/stream/252271-general/topic/Nil.20in.20locals

PR #345, opened 2020-08-25
Add Windows support (mingw)

theHamsta:
This PR adds Windows support for parser installation.

What did I do?

 -   install git
 -  open powershell (admin)
 -   install chocoletey by pasting the following line

```
Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))
```

 - open powershell (admin)
 - run `choco install mingw`
 - `refreshenv`
 - run Nvim-qt from nightly neovim releases
  - google how to install vim plug
  - `TSInstall cpp` installs stuff. *.so is loaded and plugin functionality works (highlighting, go to definition, list locals).

It was a lot of pain to program this on Windows. Will fix lua style on Linux :stuck_out_tongue_closed_eyes:

#6 helps to compile the parsers using the mingw toolchain. TODO: MSVC (CL.EXE)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Please consider the experience with this plug-in as experimental until Neovim 0.
 
 - Neovim [nightly](https://github.com/neovim/neovim#install-from-source)
 - `git` in your path.
-- A C compiler in your path.
+- A C compiler in your path ([Windows users please read this!](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Windows-support)).
 
 ## Installation
 
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -58,6 +58,10 @@ local function iter_cmd_sync(cmd_list)
   return true
 end
 
+local function select_executable(executables)
+  return vim.tbl_filter(function(c) return fn.executable(c) == 1 end, executables)[1]
+end
+
 local function run_install(cache_folder, package_path, lang, repo, with_sync)
   parsers.reset_cache()
 
@@ -66,6 +70,14 @@ local function run_install(cache_folder, package_path, lang, repo, with_sync)
   -- compile_location only needed for typescript installs.
   local compile_location = cache_folder..'/'..(repo.location or project_name)
   local parser_lib_name = package_path.."/parser/"..lang..".so"
+
+  local compilers = { "cc", "gcc", "clang" }
+  local cc = select_executable(compilers)
+  if not cc then
+    api.nvim_err_writeln('No C compiler found! "'..table.concat(compilers, '", "')..'" are not executable.')
+    return
+  end
+
   local command_list = {
     {
       cmd = 'rm',
@@ -83,7 +95,7 @@ local function run_install(cache_folder, package_path, lang, repo, with_sync)
       },
     },
     {
-      cmd = 'cc',
+      cmd = cc,
       info = 'Compiling...',
       err = 'Error during compilation',
       opts = {
@@ -148,12 +160,8 @@ local function install_lang(lang, ask_reinstall, cache_folder, package_path, wit
   run_install(cache_folder, package_path, lang, install_info, with_sync)
 end
 
--- TODO(kyazdani): this should work on windows too
 local function install(with_sync, ask_reinstall)
   return function (...)
-    if fn.has('win32') == 1 then
-      return api.nvim_err_writeln('This command is not available on windows at the moment.')
-    end
 
     if fn.executable('git') == 0 then
       return api.nvim_err_writeln('Git is required on your system to run this command')
--- a/lua/nvim-treesitter/utils.lua
+++ b/lua/nvim-treesitter/utils.lua
@@ -19,7 +19,9 @@ end
 
 function M.get_package_path()
   for _, path in pairs(api.nvim_list_runtime_paths()) do
-    if string.match(path, '.*/nvim%-treesitter') then
+    local last_segment = vim.fn.fnamemodify(path, ":t")
+    local penultimate_segment = vim.fn.fnamemodify(path, ":t:t")
+    if last_segment == "nvim-treesitter" or (last_segment == "" and penultimate_segment == "nvim-treesitter") then
       return path
     end
   end
@@ -28,20 +30,15 @@ function M.get_package_path()
 end
 
 function M.get_cache_dir()
-  local home = fn.get(fn.environ(), 'HOME')
-  local xdg_cache = fn.get(fn.environ(), 'XDG_CACHE_HOME')
+  local cache_dir = fn.stdpath('data')
 
-  if xdg_cache == 0 then
-    xdg_cache = home .. '/.cache'
-  end
-
-  if luv.fs_access(xdg_cache, 'RW') then
-    return xdg_cache
+  if luv.fs_access(cache_dir, 'RW') then
+    return cache_dir
   elseif luv.fs_access('/tmp', 'RW') then
     return '/tmp'
   end
 
-  return nil, 'Invalid cache rights, $XDG_CACHE_HOME or /tmp should be read/write'
+  return nil, 'Invalid cache rights, '..fn.stdpath('data')..' or /tmp should be read/write'
 end
 
 -- Gets a property at path


theHamsta:
Would be glad if a real Windows user could confirm this. Also if the current solution is enough or whether parser download or CL.exe support is desired. Probably a few command line tools used during the installation process only worked because mingw was installed.

theHamsta:
`cl.exe /Fe: parser.so /Isrc src/parser.c /Os /LD` `"stdbool.h": No such file or directory`. I get the same error with `stdint.h`. Apparently, there's no flag to explicitly compile as C. On their website they claim they support `stdbool.h`.

theHamsta:
An alternative to manually determine the cache directory would be to use 
```
stdpath({what})					*stdpath()* *E6100*
		Returns |standard-path| locations of various default files and
		directories.

		{what}       Type    Description ~
		cache        String  Cache directory. Arbitrary temporary
		                     storage for plugins, etc.
		config       String  User configuration directory. The
		                     |init.vim| is stored here.
		config_dirs  List    Additional configuration directories.
		data         String  User data directory. The |shada-file|
		                     is stored here.
		data_dirs    List    Additional data directories.

		Example: >
			:echo stdpath("config")
```
but then we would need to call mkdir first. This might complicate thing on Windows. Because there's no `-p` option so we need to check whether it exists.

theHamsta:
`stdpath('data')` should usually exist since it hold the `shada` file.

vigoux:
> Installation instructions to wiki page?
Yes, that would be a good idea

README.md
```diff
@@ -82,7 +82,7 @@ so you'll need to activate them by putting this in your `init.vim` file:
 ```lua
 lua <<EOF
-  ensure_installed = "all"      -- one of "all", "language", or a list of languages
+  ensure_installed = "all",     -- one of "all", "language", or a list of languages
```
vigoux, line 85: Not related to this PR ?

doc/nvim-treesitter.txt
```diff
@@ -36,7 +36,7 @@ To enable supported features, put this in your `init.vim` file:
 >
   lua <<EOF
   require'nvim-treesitter.configs'.setup {
-    ensure_installed = "all"      -- one of "all", "language", or a list of languages
+    ensure_installed = "all",     -- one of "all", "language", or a list of languages
```
vigoux, line 39: Same

lua/nvim-treesitter/utils.lua
```diff
@@ -19,7 +19,7 @@ end
 
 function M.get_package_path()
   for _, path in pairs(api.nvim_list_runtime_paths()) do
-    if string.match(path, '.*/nvim%-treesitter') then
+    if string.match(path, '.*/nvim%-treesitter') or string.match(path, '.*\\nvim%-treesitter') then
```
vigoux, line 22: Can we use `fnamemodify` here instead ? That would avoid us some headaches

lua/nvim-treesitter/utils.lua
```diff
@@ -29,7 +29,14 @@ end
...
+  local xdg_cache
+  if fn.has("win32") then
+    xdg_cache = fn.get(fn.environ(), 'LOCALAPPDATA')
+  else
+    xdg_cache = fn.get(fn.environ(), 'XDG_CACHE_HOME')
+  end
+
 
   if xdg_cache == 0 then
     xdg_cache = home .. '/.cache'
```
vigoux, line 42: As mentionned in the comments, it would be better to use `stdpath` here.

vigoux (changes requested):

README.md
theHamsta: so what? but related to setting up this plugin for Windows in a manner that's working.

vigoux: I' really like separating PRs by their meaning rather than `By the way I changed that too.`

theHamsta: I won't open a PR for a comma. If this change is already in master a rebase would remove this.

vigoux: Let's keep it given the size of this change, but the fact that the only change is this file is a comma bothers me.

theHamsta:
Brave people can try out this for `cl.exe`

It worked for load c but for cpp the anti-virus scanner went crazy.

install.lua
```lua
local api = vim.api
local fn = vim.fn
local luv = vim.loop

local info = require'nvim-treesitter.info'

local M = {}

function M.iter_cmd(cmd_list, i, lang, success_message)
  if i == #cmd_list + 1 then return print(success_message) end

  local attr = cmd_list[i]
  if attr.info then print(attr.info) end

  local handle

  handle = luv.spawn(attr.cmd, attr.opts, vim.schedule_wrap(function(code)
    handle:close()
    if code ~= 0 then return api.nvim_err_writeln(attr.err) end
    M.iter_cmd(cmd_list, i + 1, lang, success_message)
  end))
end

local function get_command(cmd)
  local ret = ''
  if cmd.opts and cmd.opts.cwd then
    ret = string.format('cd %s;\n', cmd.opts.cwd)
  end

  ret = string.format('%s%s ', ret, cmd.cmd)

  local options = ""
  if cmd.opts and cmd.opts.args then
    for _, opt in ipairs(cmd.opts.args) do
      options = string.format("%s %s", options, opt)
    end
  end

  return string.format('%s%s', ret, options)
end

local function iter_cmd_sync(cmd_list)
  for _, cmd in ipairs(cmd_list) do
    if cmd.info then
      print(cmd.info)
    end

    vim.fn.system(get_command(cmd))
    if vim.v.shell_error ~= 0 then
      api.nvim_err_writeln(cmd.err)
      return false
    end

  end

  return true
end

local function select_executable(executables)
  return vim.tbl_filter(function(c) return fn.executable(c) == 1 end, executables)[1]
end

local function run_install(cache_folder, package_path, lang, repo, with_sync)
  parsers.reset_cache()

  local project_name = 'tree-sitter-'..lang
  local project_repo = cache_folder..'/'..project_name
  -- compile_location only needed for typescript installs.
  local compile_location = cache_folder..'/'..(repo.location or project_name)
  local parser_lib_name = package_path.."/parser/"..lang..".so"

  local compilers = { "cc", "gcc", "cl" }
  local cc = select_executable(compilers)
  if not cc then
    api.nvim_err_writeln('No C compiler found! "'..table.concat(compilers, '", "')..'" are not executable.')
    return
  end

  local command_list = {
    {
      cmd = 'rm',
      opts = {
        args = { '-rf', project_repo },
      },
    },
    {
      cmd = 'git',
      info = 'Downloading...',
      err = 'Error during download, please verify your internet connection',
      opts = {
        args = { 'clone', '--single-branch', '--branch', 'master', '--depth', '1', repo.url, project_name },
        cwd = cache_folder,
      },
    },
    {
      cmd = cc,
      info = 'Compiling...',
      err = 'Error during compilation',
      opts = {
        args = (string.match(cc, 'cl$') or string.match(cc, 'cl.exe$'))
        and vim.tbl_flatten({
            '/Fe:',
            'parser.so',
            '/Isrc',
            repo.files,
            '-Os',
            '/LD',
        })
        or vim.tbl_flatten({
            '-o',
            'parser.so',
            '-I./src',
            repo.files,
            '-shared',
            '-Os',
            '-lstdc++',
            '-fPIC',
        }),
        cwd = compile_location
      }
    },
    {
      cmd = 'mv',
      opts = {
        args = { compile_location..'/parser.so', parser_lib_name }
      }
    },
    {
      cmd = 'rm',
      opts = {
        args = { '-rf', project_repo },
      },
    },
  }

  if with_sync then
    if iter_cmd_sync(command_list) == true then
      print('Treesitter parser for '..lang..' has been installed')
    end
  else
    M.iter_cmd(command_list, 1, lang, 'Treesitter parser for '..lang..' has been installed')
  end
end

local function install_lang(lang, ask_reinstall, cache_folder, package_path, with_sync)
  if #api.nvim_get_runtime_file('parser/'..lang..'.so', false) > 0 then
    if ask_reinstall ~= 'force' then
      if not ask_reinstall then return end

      local yesno = fn.input(lang .. ' parser already available: would you like to reinstall ? y/n: ')
      print('\n ') -- mandatory to avoid messing up command line
      if not string.match(yesno, '^y.*') then return end
    end
  end

  local parser_config = parsers.get_parser_configs()[lang]
  if not parser_config then
    return api.nvim_err_writeln('Parser not available for language '..lang)
  end

  local install_info = parser_config.install_info
  vim.validate {
    url={ install_info.url, 'string' },
    files={ install_info.files, 'table' }
  }

  run_install(cache_folder, package_path, lang, install_info, with_sync)
end

local function install(with_sync, ask_reinstall)
  return function (...)

    if fn.executable('git') == 0 then
      return api.nvim_err_writeln('Git is required on your system to run this command')
    end

    local package_path, err = utils.get_package_path()
    if err then return api.nvim_err_writeln(err) end

    local cache_folder, err = utils.get_cache_dir()
    if err then return api.nvim_err_writeln(err) end

    local languages
    local ask
    if ... == 'all' then
      languages = parsers.available_parsers()
      ask = false
    else
      languages = vim.tbl_flatten({...})
      ask = ask_reinstall
    end

    for _, lang in ipairs(languages) do
      install_lang(lang, ask, cache_folder, package_path, with_sync)
    end
  end
end

function M.update(lang)
  if lang then
    install(false, 'force')(lang)
  else
    local installed = info.installed_parsers()
    for _, lang in pairs(installed) do
      install(false, 'force')(lang)
    end
  end
end

function M.uninstall(lang)
  if lang == 'all' then
    local installed = info.installed_parsers()
    for _, lang in pairs(installed) do
      M.uninstall(lang)
    end
  elseif lang then
    local package_path, err = utils.get_package_path()
    if err then
      print(err)
      return
    end
    local parser_lib = package_path.."/parser/"..lang..".so"

    local command_list = {
        {
          cmd = 'rm',
          opts = {
            args = { parser_lib },
          },
          info = "Uninstalling parser for "..lang,
          err = "Could not delete "..parser_lib,
        },
      }
      M.iter_cmd(command_list, 1, lang, 'Treesitter parser for '..lang..' has been uninstalled')
  end
end

M.ensure_installed = install(false, false)

M.commands = {
  TSInstall = {
    run = install(false, true),
    args = {
      "-nargs=+",
      "-complete=custom,v:lua.ts_installable_parsers"
    }
  },
  TSInstallSync = {
    run = install(true, true),
    args = {
      "-nargs=+",
      "-complete=custom,v:lua.ts_installable_parsers"
    }
  },
  TSUpdate = {
    run = M.update,
    args = {
      "-nargs=*",
      "-complete=custom,v:lua.ts_installed_parsers"
    }
  },
  TSUninstall = {
    run = M.uninstall,
    args = {
      "-nargs=+",
      "-complete=custom,v:lua.ts_installed_parsers"
    }
  }
}

return M
```

vigoux:
Lol can't we have a patch instead of that full file ?

theHamsta:
Virus scanner went crazy and deleted the whole plugin folder so I posted the text. Can you use computers with Windows? I will create a branch back on Linux.

lua/nvim-treesitter/utils.lua
theHamsta: Maybe fnamemodifiy brings new headaches? You can check whether the new version is better.

README.md
theHamsta: Ok, I create a PR. Then, rebasing makes this disappear.

lua/nvim-treesitter/utils.lua
```diff
@@ -29,14 +31,14 @@ end
 
 function M.get_cache_dir()
   local home = fn.get(fn.environ(), 'HOME')
-  local xdg_cache = fn.get(fn.environ(), 'XDG_CACHE_HOME')
+  local cache_dir = fn.stdpath('data')
```
vigoux, line 34: Why not `stdpath('cache')` ?

theHamsta: Because then we have to make sure the cache path exists. You gave :+1:  to `stdpath('data')`

vigoux: Okay, sound good then, I wasn't sure in the end, now I am.

vigoux (approved):
Let's include that tiny doc change too in the end.

theHamsta:
I can make a separate doc commit in this PR that updates README to link to wiki page.

vigoux:
> I can make a separate doc commit in this PR that updates README to link to wiki page.

Good idea, and we should move some parts of the documentation to the wiki page, to keep the readme simple.

theHamsta:
@vigoux I removed the check `cache_dir == 0` the docs reads as if `stdpath('data')` always returns a value.

lua/nvim-treesitter/install.lua
```diff
@@ -66,6 +70,14 @@ local function run_install(cache_folder, package_path, lang, repo, with_sync)
   -- compile_location only needed for typescript installs.
   local compile_location = cache_folder..'/'..(repo.location or project_name)
   local parser_lib_name = package_path.."/parser/"..lang..".so"
+
+  local compilers = { "cc", "gcc", "clang" }
```
theHamsta, line 74: Should we export the compilers of this module, so that users can set their favorite compiler in a weird location?

README.md
```diff
@@ -19,7 +19,7 @@ Please consider the experience with this plug-in as experimental until Neovim 0.
 
 - Neovim [nightly](https://github.com/neovim/neovim#install-from-source)
 - `git` in your path.
-- A C compiler in your path.
+- A C compiler in your path ([Windows users please read this!](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Windows-support)).
 
 ## Installation
```
theHamsta, line 24: #348 got merged separately.

lua/nvim-treesitter/install.lua
vigoux: Hmmm.. Maybe at some point, but let's keep it like that for now

vigoux:
Can you change the `Check query files` action to be a matrix and run tests on multiple platforms or is it a huge pain ?

theHamsta:
It was a huge pain when I tried it last time because I'm not yet familiar with GH syntax. It wouldn't accept an additional Multi-Platform Test job. But as I said I prepare to do this maybe this weekend.

theHamsta:
I would prefer having the check query files platform specific. But test on all platforms. If you find a quick solution, just push on this branch.

theHamsta:
I could copy one of my old cross platform Travis jobs though. ,😅

vigoux:
Otherwise we could just have the parser installation cross platform

vigoux (commented):
LGTM, can we merge that @theHamsta ?

PR #351, opened 2020-08-26
Add lockfile and make lockfile default install revision

theHamsta:
For introducing lockfiles I would suggest a two PR strategy. In this PR, we offer the option to install the parser from a revision of a lockfile instead of master (this is the default). The behavior can be controller by a config variable.

```
setup {
   update_strategy = 'lockfile' 
}
```

A second PR can track the currently installed parser revision to check whether they match with the lockfile (which can be updated on nvim-treesitter update) or the newest master. This will require some async iter_cmd magic to be non-blocking.

TODO:
 - [ ] docs


theHamsta:
This is still WIP. But we need to test this PR because it could help users to use Java, Go, Python parsers for highlighting again.

kyazdani42 (approved):
LGTM

theHamsta:
Probably the easiest option to continue from here is to just dump the used lockfile at the moment of user installation to a user installation file. We need to update a specific parser whenever revisions in user installation and lockfile differ.

PR #356, opened 2020-08-27
fix(health): only check installed parsers

vigoux:
This is to avoid awfully long checkhealths. And not installed parsers can be infered from the fact that they are not listed.

### DIFF
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -2,7 +2,7 @@ local api = vim.api
 local fn = vim.fn
 
 local queries = require'nvim-treesitter.query'
-local parsers = require'nvim-treesitter.parsers'
+local info = require'nvim-treesitter.info'
 
 local health_start = vim.fn["health#report_start"]
 local health_ok = vim.fn['health#report_ok']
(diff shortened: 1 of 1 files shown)


theHamsta:
I thought this long ago: wouldn't it be better to just show one summary for all the languages? E.g.

x highlights.scm: missing for foolang, bar, baz

lua/nvim-treesitter/health.lua
```diff
@@ -44,38 +45,24 @@ end
 
 function M.checkhealth()
   -- Installation dependency checks
-  health_start('Installation')
   install_health()
-
-  local missing_parsers = {}
   -- Parser installation checks
   for _, parser_name in pairs(parsers.available_parsers()) do
```
theHamsta, line 50: Just change to installed_parsers?

theHamsta, line 50: Just use installed_parsers?

theHamsta (approved):

lua/nvim-treesitter/health.lua
vigoux: Will do, give me a sec

vigoux: done

theHamsta (approved):

PR #357, opened 2020-08-28
fix(fold): revamp fold into a module

vigoux:
Move fold into a module, and update it a bit to fix some issues.

### DIFF
--- a/lua/nvim-treesitter/fold.lua
+++ b/lua/nvim-treesitter/fold.lua
@@ -5,6 +5,8 @@ local parsers = require'nvim-treesitter.parsers'
 
 local M = {}
 
+-- This is cached on buf tick to avoid computing that multiple times
+-- Especially not for every line in the file when `zx` is hit
 local folds_levels = utils.memoize_by_buf_tick(function(bufnr)
   local lang = parsers.get_buf_lang(bufnr)
 
@@ -38,10 +40,12 @@ local folds_levels = utils.memoize_by_buf_tick(function(bufnr)
   local levels = {}
   local current_level = 0
 
+  -- We now have the list of fold opening and closing, fill the gaps and mark where fold start
   for lnum=0,api.nvim_buf_line_count(bufnr) do
     local prefix= ''
     local shift = levels_tmp[lnum] or 0
 
+    -- Determine if it's the start of a fold
     if levels_tmp[lnum] and shift >= 0 then
       prefix = '>'
     end
--- a/lua/nvim-treesitter/fold.lua
+++ b/lua/nvim-treesitter/fold.lua
@@ -1,28 +1,60 @@
+local api = vim.api
+local utils = require'nvim-treesitter.ts_utils'
+local query = require'nvim-treesitter.query'
 local parsers = require'nvim-treesitter.parsers'
 
 local M = {}
 
+local folds_levels = utils.memoize_by_buf_tick(function(bufnr)
+  local lang = parsers.get_buf_lang(bufnr)
+
+  local matches
+  if query.has_fold(lang) then
+    matches = query.get_capture_matches(bufnr, "@fold", "fold")
+  elseif query.has_locals(lang) then
+    matches = query.get_capture_matches(bufnr, "@scope", "locals")
+  else
+    return {}
+  end
+
+  local levels_tmp = {}
+
+  for _, node in ipairs(matches) do
+    local start, _, stop, stop_col = node.node:range()
+
+    if stop_col > 0 then
+      stop = stop + 1
+    end
+
+    -- This can be folded
+    -- Fold only multiline nodes that are not exactly the same as prevsiously met folds
+    if start ~= stop and not (levels_tmp[start] and levels_tmp[stop]) then
+      levels_tmp[start] = (levels_tmp[start] or 0) + 1
+      levels_tmp[stop] = (levels_tmp[stop] or 0) - 1
+    end
+
+  end
+
+  local levels = {}
+  local current_level = 0
+
+  for lnum=0,api.nvim_buf_line_count(bufnr) do
+    current_level = current_level + (levels_tmp[lnum] or 0)
+    levels[lnum + 1] = current_level
+  end
+
+  return levels
+end)
+
 function M.get_fold_indic(lnum)
   if not parsers.has_parser() or not lnum then return '0' end
 
-  local function smallest_multiline_containing(node, level)
-    for index = 0,(node:named_child_count() -1) do
-      local child = node:named_child(index)
-      local start, _, stop, _ = child:range()
+  local buf = api.nvim_get_current_buf()
 
-      if start ~= stop and start <= (lnum -1) and stop >= (lnum -1) then
-        return smallest_multiline_containing(child, level + 1)
-      end
-    end
+  local levels = folds_levels(buf) or {}
 
-    return node, level
-  end
+  return tostring(levels[lnum] or 0)
 
-  local parser = parsers.get_parser()
-
-  local _, level = smallest_multiline_containing(parser:parse():root(), 0)
-
-  return tostring(level)
 end
 
 return M
--- a/lua/nvim-treesitter/fold.lua
+++ b/lua/nvim-treesitter/fold.lua
@@ -39,8 +39,15 @@ local folds_levels = utils.memoize_by_buf_tick(function(bufnr)
   local current_level = 0
 
   for lnum=0,api.nvim_buf_line_count(bufnr) do
-    current_level = current_level + (levels_tmp[lnum] or 0)
-    levels[lnum + 1] = current_level
+    local prefix= ''
+    local shift = levels_tmp[lnum] or 0
+
+    if levels_tmp[lnum] and shift >= 0 then
+      prefix = '>'
+    end
+
+    current_level = current_level + shift
+    levels[lnum + 1] = prefix .. tostring(current_level)
   end
 
   return levels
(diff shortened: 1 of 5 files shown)


lua/nvim-treesitter/fold.lua
```diff
@@ -1,28 +1,60 @@
...
+local folds_levels = utils.memoize_by_buf_tick(function(bufnr)
+  print("hey")
+  local levels_tmp = {}
 
-  local function smallest_multiline_containing(node, level)
-    for index = 0,(node:named_child_count() -1) do
-      local child = node:named_child(index)
-      local start, _, stop, _ = child:range()
+  for _, node in ipairs(query.get_capture_matches(bufnr, "fold", "fold")) do
+    print("yo")
```
theHamsta, line 13: Yo!

lua/nvim-treesitter/fold.lua
```diff
@@ -1,28 +1,60 @@
...
+  return levels
+end)
+
+function M.get_fold_indic(lnum)
+  if not parsers.has_parser() or not lnum then return '0' end
 
-  local _, level = smallest_multiline_containing(parser:parse():root(), 0)
+  local buf = api.nvim_get_current_buf()
+
+  print(vim.inspect(folds_levels))
```
theHamsta, line 41: debug

lua/nvim-treesitter/fold.lua
```diff
@@ -1,28 +1,60 @@
...
+
+  local levels = folds_levels(buf) or {}
+
+  return tostring(levels[lnum] or 0)
+end
+
+function M.attach(bufnr)
+  local win = vim.fn.win_findbuf(bufnr)[1]
```
theHamsta, line 49: A buffer can have multiple windows.

queries/lua/fold.scm
```diff
@@ -0,0 +1,6 @@
+[
+ (for_in_statement)
+ (for_statement)
+ (if_statement)
+ (function_definition)
+ ] @fold
```
theHamsta, line 6: do we really want to start a new group of query files? We should at least be able to fall back to old behavior. Since it will take ages until all languages are supported and it will take more effort for people who maintain custom languages.

The only problem with the old folding was that one line could have multiple folds. I guess this can still happen when you write everything in one line.

lua/nvim-treesitter/fold.lua
theHamsta: Will the foldmethod be inherited? When the window switches to a different buffer it will still have the folding marker. This why I gave up my try to make folding a module. But thanks that you are tackling the problem.

queries/lua/fold.scm
vigoux: Hopefully it should not, but I am diging into the issue rn

lua/nvim-treesitter/fold.lua
vigoux: Oh, that's an interesting issue

queries/lua/fold.scm
vigoux: Should be good now

lua/nvim-treesitter/fold.lua
vigoux: Maybe we can have the module enabled globally by default, and let users fine tune it using `nvim_treesitter#foldexr()` otherwise

lua/nvim-treesitter/fold.lua
```diff
@@ -1,28 +1,69 @@
...
+  api.nvim_win_set_option(win, 'foldmethod', 'expr')
+  api.nvim_win_set_option(win, 'foldexpr', 'nvim_treesitter#foldexpr()')
+end
 
-  return tostring(level)
+function M.detach(bufnr)
+  local win = vim.fn.win_findbuf(bufnr)[1]
+  api.nvim_win_set_option(win, 'foldmethod', nil)
+  api.nvim_win_set_option(win, 'foldexpr', nil)
```
theHamsta, line 66: Detach/attach will probably not work:

 - here it's not guaranteed that the buffer has still a window
 - it is not clear whether detaching from this buffer should deactivate the foldexpr. It might be that the new buffer still supports `foldexpr`

vigoux: Then let's not have fold as a module.
It will keep the thing simple and still allow users to use it.

lua/nvim-treesitter/fold.lua
```diff
@@ -1,28 +1,69 @@
...
+    levels[lnum + 1] = prefix .. tostring(current_level)
   end
 
-  local parser = parsers.get_parser()
+  return levels
+end)
 
-  local _, level = smallest_multiline_containing(parser:parse():root(), 0)
+function M.get_fold_indic(lnum)
+  if not parsers.has_parser() or not lnum then return '0' end
```
theHamsta, line 48: does this guard allow to use treesitter foldexpr globally activated? Then we could remove the attach/detach. We could allow the user to configure a fallback if treesitter folding is not available.

vigoux:
Removed the module part as it is not that useful, it is now just a basic rework

queries/lua/fold.scm
vigoux: Regarding the new query group thing, I think we should.
With runtime files, vim always had a tednedcy to split things in specialized files, and I think that we should do the same. Plus this allows us to fold, indent, or wathever, on different nodes and allow users to customize however they want

theHamsta: We had a feature that was working pretty good without any queries (which was a pretty elegant solution from you). You always have to bear in mind that users have to maintain and write those queries and users starting a new language won't have the time to do that in the first place. It's nice that you can have a better experience with *carefully* tuned configs but we shouldn't force it. A simpler solution is always better.

We will have a feature regression if we don't provide the old fallback. If we force all language maintainers to implement the fold queries in a rush (so not *carefully* tuned). We will have bugs of type: cannot fold for-loops because the maintainer forgot that single node in the query. Would be great if we can provide a nice feature without configuration and perfect one if people are willing to tune it for that specific language. I don't think that people will provide those queries for any language. Right now we have languages with all queries and some with partial support and I guess it'll stay that way for a time.

vigoux: I added a fallback to `locals` scopes, so that is uses that as a default.

vigoux: And in any case, the old solution was buggy, and fixing it would take too much performance, the solution as it is right now, that is `fold.scm` + fallback to `locals.scm` will be good enough.

lua/nvim-treesitter/fold.lua
```diff
@@ -1,28 +1,71 @@
...
-        return smallest_multiline_containing(child, level + 1)
+  for _, node in ipairs(matches) do
+    local start, _, stop, _ = node.node:range()
+    stop = stop + 1
+
+    -- This can be folded
+    -- Fold only multiline nodes that are not exactly the same as prevsiously met folds
+    if start ~= stop and not (levels_tmp[start] and levels_tmp[stop]) then
+      if levels_tmp[start] and levels_tmp[stop] then
+        print("folding twice :", start, stop)
```
theHamsta, line 32: debug

lua/nvim-treesitter/fold.lua
```diff
@@ -1,28 +1,67 @@
...
-      if start ~= stop and start <= (lnum -1) and stop >= (lnum -1) then
-        return smallest_multiline_containing(child, level + 1)
-      end
+  for _, node in ipairs(matches) do
+    local start, _, stop, _ = node.node:range()
+    stop = stop + 1
+
+    -- This can be folded
+    -- Fold only multiline nodes that are not exactly the same as prevsiously met folds
+    if start ~= stop and not (levels_tmp[start] and levels_tmp[stop]) then
```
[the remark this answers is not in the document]
vigoux: It indeed does not work...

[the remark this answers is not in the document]
vigoux: I guess we can have that version for now.
The thing is that it is already an improvement over `classic` folding.
This PR also brings a big performance improvement too, so I think we should merge that asap.

vigoux:
@theHamsta can we go ahead and merge that first version ? Enhancements will come later

theHamsta:
Hi. I'm at the moment a bit short of time. So didn't test yet. If you feel confident you can merge it's for sure an improvement. 

Next steps should be how you distribute the lines among different children if the share lines. Also C++ folds need to be created since the only work with c queries after this PR.

lua/nvim-treesitter/fold.lua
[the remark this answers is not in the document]
theHamsta: Probably you just have to use  an arbitrary criterion to decide that some folds end earlier when they are in shared line. A decision whether a shared line where to set splits (fold out and in) between the children.

lua/nvim-treesitter/fold.lua
```diff
@@ -1,28 +1,70 @@
...
-      local child = node:named_child(index)
-      local start, _, stop, _ = child:range()
+  for _, node in ipairs(matches) do
+    local start, _, stop, stop_col = node.node:range()
 
-      if start ~= stop and start <= (lnum -1) and stop >= (lnum -1) then
-        return smallest_multiline_containing(child, level + 1)
-      end
+    if stop_col > 0 then
+      stop = stop + 1
```
theHamsta, line 28: A this marks also an exclusive end of the fold... The stop col *might* not be necessary.

theHamsta (approved):

lua/nvim-treesitter/fold.lua
vigoux: Maybe ... We'll try later on

PR #360, opened 2020-08-29
Lsp fallback for goto_definition

theHamsta:
Depends on #336


vigoux:
@theHamsta , this needs a rebase, and then we will be good to go

vigoux (approved):

PR #362, opened 2020-08-30
feat(CI): auto-update README.md

theHamsta:
Experiment to make the README CI rather auto-update the PRs instead of making them fail. See: https://github.com/neovim/nvim-lspconfig/blob/master/.github/workflows/docgen.yml

### DIFF
--- a/.github/workflows/update-readme.yml
+++ b/.github/workflows/update-readme.yml
@@ -1,11 +1,12 @@
 name: Check README parser info
 
-on: [push, pull_request]
+on: push
 
 jobs:
-  luacheck:
+  update-readme:
     name: Check README parser info
     runs-on: ubuntu-latest
+    if: github.ref != 'master'
     steps:
       - uses: actions/checkout@v1
 
@@ -18,8 +19,17 @@ jobs:
           mkdir -p ~/.local/share/nvim/site/pack/nvim-treesitter/start
           ln -s $(pwd) ~/.local/share/nvim/site/pack/nvim-treesitter/start
 
-      - name: Compile parsers
-        run: ./nvim.appimage --headless -c "TSInstallSync all" -c "q"
-
+      # inspired by nvim-lspconfigs
       - name: Check README
-        run: ./nvim.appimage --headless -c "luafile ./scripts/update-readme.lua" -c "q"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COMMIT_MSG: |
+            [docgen] Update README.md
+            skip-checks: true
+        run: |
+          git config user.email "actions@github"
+          git config user.name "Github Actions"
+          git remote set-url origin https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
+          ./nvim.appimage --headless -c "luafile ./scripts/update-readme.lua" -c "q" || git add README.md
+          # Only commit and push if we have changes
+          git diff --quiet && git diff --staged --quiet || (git commit -m "${COMMIT_MSG}"; git push origin HEAD:${GITHUB_REF})
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ We are looking for maintainers to write query files for their languages.
 
 List of currently supported languages:
 
-<!--Update this section by running `nvim --headless -c "luafile ./scripts/update-readme.lua"`-->
+<!--This section of the README is automatically updated by a CI job-->
 <!--parserinfo-->
 - [x] [bash](https://github.com/tree-sitter/tree-sitter-bash) (maintained by @TravonteD)
 - [x] [c](https://github.com/tree-sitter/tree-sitter-c) (maintained by @vigoux)
--- a/scripts/update-readme.lua
+++ b/scripts/update-readme.lua
@@ -37,8 +37,5 @@ if string.find(readme_text, generated_text, 1, 'plain') then
 else
   print("New README.md was written. Please commit that change! Old text was: ")
   print(string.sub(readme_text, string.find(readme_text, "<!%-%-parserinfo%-%->.*<!%-%-parserinfo%-%->")))
-  print("\n")
-  print('If you see this on CI, please run `nvim --headless -c "luafile ./scripts/update-readme.lua"` '
-      ..'locally and commit the changes!')
   vim.cmd('cq')
 end


vigoux:
Damn, looks like it worked

theHamsta:
Probably, GH-actions need rights to push on the branch.

vigoux:
> Probably, GH-actions need rights to push on the branch.

Are you sure ? It seems that it worked ?

theHamsta:
> Are you sure ? It seems that it worked ?

Saw this only later, the actual commit failed though.

vigoux:
That's weird...

.github/workflows/update-readme.yml
```diff
@@ -21,5 +21,17 @@ jobs:
       - name: Compile parsers
         run: ./nvim.appimage --headless -c "TSInstallSync all" -c "q"
```
theHamsta, line 22: Skip that? Or do we want to add information to the README/docs that requires parser information?

vigoux: I don't think we need that, you can remove it

.github/workflows/update-readme.yml
```diff
@@ -21,5 +21,17 @@ jobs:
...
 
+      # inspired by nvim-lspconfigs
       - name: Check README
-        run: ./nvim.appimage --headless -c "luafile ./scripts/update-readme.lua" -c "q"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
vigoux, line 27: I think we need to set that

.github/workflows/update-readme.yml
```diff
@@ -21,5 +21,17 @@ jobs:
...
+          COMMIT_MSG: |
+            [docgen] Update README.md
+            skip-checks: true
```
vigoux, line 30: We will need to filter that, maybe rename that to `no-readme-update` and filter that in the action.

.github/workflows/update-readme.yml
```diff
@@ -18,8 +19,17 @@ jobs:
...
+        run: |
+          git config user.email "actions@github"
+          git config user.name "Github Actions"
+          git remote set-url origin https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
+          ./nvim.appimage --headless -c "luafile ./scripts/update-readme.lua" -c "q" || git add README.md
```
vigoux, line 33: This should be an `&&` ?

vigoux (commented):
Looks like it works pretty great

.github/workflows/update-readme.yml
theHamsta: Only add, when update fails.

vigoux:
This makes the action fail though, while we would like it to succeed

theHamsta:
Maybe just give him the permissions?
```
!remote: Permission to nvim-treesitter/nvim-treesitter.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/nvim-treesitter/nvim-treesitter.git/': The requested URL returned error: 403
##[error]Process completed with exit code 128.
``

**Comment by theHamsta** [comment]
I just fixed the CI of nvim-lspconfigs and they have a similar problem with `skip-checks` now.

File: .github/workflows/update-readme.yml (hunk 2 above)
Comment (theHamsta) [reply .github/workflows/update-readme.yml:25]: You need to do that on Travis. But it would be strange if you need Github to authenticate it self on Github. There's also a magic environment variable to do this on Gitlab. It gives the current user the rights of the triggered. But apparently the jobs in not executed with the rights of the user but with the rights of the GH Actions bot.

**Comment by theHamsta** [comment]
`docgen` causes the CI to be skipped. We probably want the CI to be rerun to have the nice button on each PR request.

File: .github/workflows/update-readme.yml [hunk 5]
```diff
@@ -1,11 +1,12 @@
 name: Check README parser info
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   luacheck:
     name: Check README parser info
     runs-on: ubuntu-latest
+    if: github.ref != 'master'
```
vigoux, line 9: Those two conditions seems to make this action never run...

vigoux: We should remove the `if` i guess

theHamsta: It was like this before and we are not on master. What's strange is that the GH actions commit does not seem to be associated with any branch. Was it only a shallow checkup?

vigoux: > The action is run + all checks are run on the actual commit of the user. But not after the push of the GH bot.

Can you try that once just to see ?

.github/workflows/update-readme.yml
```diff
@@ -1,11 +1,12 @@
 name: Check README parser info
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   luacheck:
```
vigoux, line 6: This should be renamed btw

vigoux (changes requested):

vigoux:
Looks like it works...

.github/workflows/update-readme.yml
theHamsta: tried. made no difference. nvim-lspconfig has the same problem right now. Something changed with GH actions.

theHamsta:
It stucks forever in a state where it cannot determine the status. Like without the if.

vigoux:
It looks like it is happening in your repo rather than on this one though

theHamsta:
> It looks like it is happening in your repo rather than on this one though
> Do you have Actions setup in your fork ?

There's nothing to set up. GH just happen automatically. The most recent commit is always tested but not the updates README.md commit (with skip or without).

vigoux:
Then that's what we want right, so let's merge that version and enhance afterwards

.github/workflows/update-readme.yml
```diff
@@ -1,11 +1,12 @@
 name: Check README parser info
 
-on: [push, pull_request]
+on: [push]
```
vigoux, line 3:
```suggestion
on: push
```

theHamsta:
I'd just wait how nvim-lspconfigs handles this. This looks like a GH bug though, since it doesn't show the status of the PR checks anymore. undefined here - no display in nvim-lspconfigs: https://github.com/neovim/nvim-lspconfig/pull/330

vigoux (commented):
To have the `if` working : https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions

theHamsta:
I guess it should be ok when we still see the test status with big buttons (not only small tick) when there's not doc change. Will test that case now.

.github/workflows/update-readme.yml
vigoux: Our secrets store is empty though so this env varialbe will not be expanded to anything

theHamsta:
Maybe we can generate the module default config table automatically automatically to the doc/*.txt or the README.md

vigoux:
> Maybe we can generate the module default config table automatically automatically to the doc/*.txt or the README.md

Can that be embedded in this PR ?

theHamsta:
This is a bit more complicated: after doing some hacks (without proper formatting and without adapting to correct usage semantic suggests that it could be possible:

# Available Modules


## nvim-treesitter.textobjects.swap

Define your own mappings to swap the node under the cursor with the next or previous one,
like function parameters or arguments.

```lua
{
  enable = true,
  swap_next = {
    ["<a-f>"] = "@function.outer",
    ["<a-l>"] = "@parameter.inner",
    ["<a-s>"] = "@statement.outer"
  },
  swap_previous = {
    ["<a-F>"] = "@function.outer",
    ["<a-L>"] = "@parameter.inner",
    ["<a-S>"] = "@statement.outer"
  }
}
```

## nvim-treesitter.textobjects.select

Define your own text objects mappings
similar to `ip` (inner paragraph) and `ap` (a paragraph).

```lua
{
  enable = true,
  keymaps = {
    aC = "@class.outer",
    ac = "@conditional.outer",
    ad = "@comment.outer",
    ae = "@block.outer",
    af = "@function.outer",
    al = "@loop.outer",
    am = "@call.outer",
    as = "@statement.outer",
    iC = "@class.inner",
    iL = {
      c = "(function_definition) @function",
      cpp = "(function_definition) @function",
      java = "(method_declaration) @function",
      python = "(function_definition) @function"
    },
    ic = "@conditional.inner",
    id = "@comment.inner",
    ie = "@block.inner",
    if = "@function.inner",
    il = "@loop.inner",
    im = "@call.inner",
    is = "@statement.inner"
  }
}
```

## nvim-treesitter.textobjects.move

Define your own mappings to jump to the next or previous text object.
This is similar to `]m`, `[m`, `]M`, `[M` Neovim's mappings to jump to the next
or previous function.

```lua
{
  enable = true,
  goto_next_end = {},
  goto_next_start = {
    ["öö"] = "@function.inner"
  },
  goto_previous_end = {},
  goto_previous_start = {
    ["üü"] = "@function.inner"
  }
}
```

## nvim-tree-docs.internal


```lua
{
  enable = true,
  keymaps = {
    doc_all_in_range = "gdd",
    doc_node_at_cursor = "gdd"
  }
}
```

## nvim-treesitter.incremental_selection

Incremental selection based on the named nodes from the grammar.

```lua
{
  enable = true,
  keymaps = {
    init_selection = "<enter>",
    node_decremental = "grm",
    node_incremental = "<enter>",
    scope_incremental = "Ts"
  }
}
```

## nvim-treesitter-playground.internal


```lua
{
  enable = true,
  persist_queries = false,
  updatetime = 25
}
```

## nvim-treesitter.refactor.highlight_current_scope

Highlights the block from the current scope where the cursor is.

```lua
{
  enable = false,
  inverse_highlighting = true
}
```

## nvim-treesitter.refactor.navigation

Provides "go to definition" for the symbol under the cursor,
and lists the definitions from the current file.

```lua
{
  enable = true,
  keymaps = {
    goto_definition = "gnd",
    goto_next_usage = "<a-*>",
    goto_previous_usage = "<a-#>",
    list_definitions = "gnD"
  }
}
```

## nvim-treesitter.refactor.smart_rename

Renames the symbol under the cursor within the current scope (and current file).

```lua
{
  enable = true,
  keymaps = {
    smart_rename = "grr"
  }
}
```

## nvim-treesitter.refactor.highlight_definitions


```lua
{
  enable = true
}
```

## nvim-treesitter.highlight

Consistent syntax highlighting.

```lua
{
  custom_captures = {},
  enable = true
}
```

vigoux:
Hmmm... I am not sure we need that in the end... Because we don't make that much changes to the modules anyway

theHamsta:
Better leave that idea for another PR, then it can be discussed whether it's necessary at all.

.github/workflows/update-readme.yml
vigoux: I don't think that we need the `skip-checks: true` here.

.github/workflows/update-readme.yml
vigoux: Can you remove that `GITHUB_TOKEN` ?

vigoux (changes requested):

.github/workflows/update-readme.yml
theHamsta: GITHUB_TOKEN is necessary for pushing.

.github/workflows/update-readme.yml
theHamsta: It doesn't have any effects right now. But usually it would skip the checks for the docgen commit (tests and recursive call to it itself, though there shouldn't be any changes then)

vigoux (approved):

PR #364, opened 2020-08-30
fix get_package_path function for paths with trailing slash

errx:
This function is not working with paths with trailing slash

so the :p modifier will add path serparator in the end (if missing) but only if directory exists
:h modifier will remove trailing path serparator

### DIFF
--- a/lua/nvim-treesitter/utils.lua
+++ b/lua/nvim-treesitter/utils.lua
@@ -20,8 +20,7 @@ end
 function M.get_package_path()
   for _, path in pairs(api.nvim_list_runtime_paths()) do
     local last_segment = vim.fn.fnamemodify(path, ":p:h:t")
-    local penultimate_segment = vim.fn.fnamemodify(path, ":p:h:t:t")
-    if last_segment == "nvim-treesitter" or (last_segment == "" and penultimate_segment == "nvim-treesitter") then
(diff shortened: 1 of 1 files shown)


theHamsta:
Shouldn't the first case just be
```
fnamemodify("foo/bar", ":t") -> bar
```
It was for the case that the path ends without "7".

For readability `match(path, `/nvim%-treesitter/*`) or `match(path, `\\nvim%-treesitter\\*`) would be a lot better.

errx:
I was thinking that "last_segment"  is for "some_path/nvim-treesitter" and "penultimate_segment" is for "some_path/nvim-treesitter/something"
is it correct?

with :p:h modifiers your path always will be without trailing slash so :t modifier will work

theHamsta:
No, I just wanted to catch the case that the path ends with `/`

This should also work by modifying the path of the containing lua file.
```lua
local function script_path()
    return debug.getinfo(1, "S").source:sub(2)
end
print(vim.fn.fnamemodify(script_path(), ':p:h:h:h'))
```

errx:
I've removed second check.

`script_path` is great too.

So I guess its up to you which way is better

theHamsta (approved):

vigoux (approved):
I like this way of doing it, and, @theHamsta i insist on using `fnamemodify` to avoid portability headaches.

PR #365, opened 2020-08-31
Simplify configs.setup

theHamsta:
This PR tries to simplify `configs.setup` to avoid a few problems. The main problem: attach/detach should be strictly inverse to each other and if they are not it shouldn't cause a problem for disabled modules.

- try to avoid enabling modules just to disable them afterwards for languages in `disable`. This could avoid problems when attach/detach of certain modules are not inverse to each other
- make attach/detach of `highlight_current_scope` inverse to each other
- avoid calling `enable_all` with disable/enable data that is actually for a different module

This solves the bug that executing this lua file (once all modules are loaded) automatically activates `highlight_current_scope` though it's deactivated. Happens with every top-level module set to true.

```lua
    require "nvim-treesitter.configs".setup(
        {
            highlight = {
                enable = true, -- false will disable the whole extension
                disable = {"html", "lua"} -- list of language that will be disabled
            },
            refactor = {
                highlight_current_scope = {
                    enable = false,
                    inverse_highlighting = true,
                    disable = {"python", "markdown"}
                },
                highlight_definitions = {
                    enable = true,
                    disable = {"markdown"}
                },
            },
            ensure_installed = "all",
        }
    )
```

This is still a draft. There were a few bugs that I discovered on the way. I found the function how it was before very confusing and I could not understand it fully. So please handle those changes with care.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -149,7 +149,9 @@ local function enable_mod_conf_autocmd(mod, lang)
   end
 end
 
-local function enable_all(mod, lang)
+local function enable_all(mod, lang, blocklist)
+  blocklist = blocklist or {}
+
   local config_mod = M.get_module(mod)
 
   if not config_mod then return end
@@ -157,7 +159,9 @@ local function enable_all(mod, lang)
   for _, bufnr in pairs(api.nvim_list_bufs()) do
     local ft = api.nvim_buf_get_option(bufnr, 'ft')
     if not lang or parsers.lang_match_ft(lang, ft) then
-      enable_module(mod, bufnr, lang)
+      if not vim.tbl_contains(blocklist, lang) then
+        enable_module(mod, bufnr, lang)
+      end
     end
   end
   if lang then
@@ -166,7 +170,7 @@ local function enable_all(mod, lang)
     end
   else
     for _, lang in pairs(parsers.available_parsers()) do
-      if parsers.has_parser(lang) then
+      if parsers.has_parser(lang) and not vim.tbl_contains(blocklist, lang) then
         enable_mod_conf_autocmd(mod, lang)
       end
     end
@@ -304,12 +308,9 @@ function M.setup(user_data)
 
       recurse_modules(function(_, _, new_path)
         if data.enable then
-          enable_all(new_path)
+          enable_all(new_path, nil, data.disable)
         end
 
-        for _, lang in ipairs(data.disable or {}) do
-          disable_mod_conf_autocmd(new_path, lang)
-        end
       end, config.modules)
     end
   end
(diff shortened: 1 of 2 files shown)


vigoux:
Is that really a draft ?

lua/nvim-treesitter/configs.lua
```diff
@@ -295,24 +299,17 @@ end
...
-      recurse_modules(function(_, _, new_path)
-        if data.enable then
-          enable_all(new_path)
-        end
-
-        for _, lang in ipairs(data.disable or {}) do
-          disable_mod_conf_autocmd(new_path, lang)
-        end
-      end, config.modules)
+  config.modules = vim.tbl_deep_extend('force', config.modules, user_data)
```
theHamsta, line 302: Why do we have modules here? This causes our table structure to be slightly different from the one the user provides us and it's only for one boolean value.

vigoux: We need that because of the use of the recurse module below

lua/nvim-treesitter/configs.lua
```diff
@@ -295,24 +299,17 @@ end
...
+
+  config.ensure_installed = user_data.ensure_installed
```
vigoux, line 304: I don't think we need the line anymore

lua/nvim-treesitter/configs.lua
```diff
@@ -295,24 +299,17 @@ end
...
+  require'nvim-treesitter.install'.ensure_installed(config.ensure_installed)
```
theHamsta, line 305: We could also just not store `ensure_installed`. We don't need it any more. Actually, I should also delete `enure_installed` from `config.modules`.

lua/nvim-treesitter/refactor/highlight_current_scope.lua
```diff
@@ -37,8 +37,8 @@ function M.attach(bufnr)
 end
 
 function M.detach(bufnr)
-  M.clear_usage_highlights(bufnr)
```
theHamsta, line 40: This function does not exist anymore. Lol.

lua/nvim-treesitter/configs.lua
theHamsta: Recurse modules does not care about a boolean because it will do a `is_module` check.

vigoux: Then you can even remove the boolean afterwards to save us some time

theHamsta:
Would be good to have @steelsojka a look on this. I only know slightly how the module system works and I really don't want to break something.

theHamsta:
The blocklist does not work yet... :-(

vigoux (commented):
Please just remove the `ensure_installed` key from `config.modules` and we will be good to go.

vigoux (approved):

theHamsta:
Add this?
```lua
if user_data.ensure_installed then
  require'nvim-treesitter.install'.ensure_installed(user_data.ensure_installed)
end
```
`ensure_installed` will also work with nil but the require would not be executed.

vigoux (approved):

PR #366, opened 2020-08-31
feat(smart_rename): set cursor to start of renamed code after renaming

theHamsta:
Sometimes when I'm shorting a identifier with smart rename, my cursor will be on some of the nodes after the identifier. So I was wondering whether we should set the cursor to start of renamed node after renaming, or should this be configurable?

Peek 2020-08-31 23-16
If I would not change the cursor position, my cursor would be on the next argument or even the closing parenthesis.

### DIFF
--- a/lua/nvim-treesitter/refactor/smart_rename.lua
+++ b/lua/nvim-treesitter/refactor/smart_rename.lua
@@ -43,6 +43,7 @@ function M.smart_rename(bufnr)
     table.insert(edits, text_edit)
   end
   vim.lsp.util.apply_text_edits(edits, bufnr)
+  ts_utils.goto_node(node_at_point, false, true)
 end
 
 function M.attach(bufnr)


vigoux:
Moving to the start of the renamed node is a good idea

lua/nvim-treesitter/refactor/smart_rename.lua
```diff
@@ -43,6 +43,7 @@ function M.smart_rename(bufnr)
     table.insert(edits, text_edit)
   end
   vim.lsp.util.apply_text_edits(edits, bufnr)
+  ts_utils.goto_node(node_at_point, false, true)
```
vigoux, line 46: Can you do that before the edits are applied just to be sure the node is valid ?

vigoux (changes requested):

lua/nvim-treesitter/refactor/smart_rename.lua
theHamsta: It's lua so the node shouldn't be garbage collected though it's already invalid. But it is still not completely correct because due to renaming of the same reference in the same line the node position might have changed.

PR #367, opened 2020-09-01
feat: add query filetype

vigoux:
Add a little `ftdetect` just to trigger `query` filetype for queries.

### DIFF
--- a/ftdetect/query.vim
+++ b/ftdetect/query.vim
@@ -1,0 +1,11 @@
+" Last Change: 2020 Sep 01
+
+function! s:shouldFt(path)
+  let l:q_dir = fnamemodify(a:path, ":p:h:h:t")
+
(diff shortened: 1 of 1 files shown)


ftdetect/query.vim
```diff
@@ -0,0 +1,11 @@
+" Last Change: 2020 Sep 01
+
+function! s:shouldFt(path)
+  let l:q_dir = fnamemodify(a:path, ":p:h:h:t")
```
theHamsta, line 4: Also a check for the `*.scm` ending?

vigoux: Already did in the autocmd

theHamsta: Sorry :smile:

theHamsta (approved):

PR #370, opened 2020-09-01
use xmap instead of vmap in textobjects

kuator:


kuator:
`vmap` works for visual and selecte mode. `xmap`  works only for visual mode. Snippet plugins use select mode for editing placeholders, and it's extremely annoying when instead of inserting a character, a text object gets selected.

vigoux:
Hey you, did you looked a bit elsewhere to see if we have other vmaps?

kuator:
> Hey you, did you looked a bit elsewhere to see if we have other vmaps?

I have found one more place, my lua-fu is weak, so I just grepped for places where mappings are set.

vigoux (approved):
LGTM, @theHamsta this is textobjects related so...

lua/nvim-treesitter/incremental_selection.lua
```diff
@@ -100,7 +100,7 @@ function M.attach(bufnr)
     if funcname == "init_selection" then
       mode = 'n'
     else
-      mode = 'v'
+      mode = 'x'
```
theHamsta, line 103: You missed changing the corresponding detach. But everything else LGTM! :+1:  Thanks!

kuator: sorry, fixed it.

vigoux (approved):

PR #371, opened 2020-09-01
feat(config): add global disable option

Konan11221:


lua/nvim-treesitter/configs.lua
```diff
@@ -300,17 +300,22 @@ function M.setup(user_data)
...
+      end
       config.modules[name] = vim.tbl_deep_extend('force', config.modules[name] or {}, data)
 
       recurse_modules(function(_, _, new_path)
         if data.enable then
           enable_all(new_path)
         end
-
-        for _, lang in ipairs(data.disable or {}) do
+        for _, lang in ipairs(config.modules[name].disable) do
```
theHamsta, line 315: Isn't this the old buggy version? Just tbl_merge it with the module disable in the new version.

@vigoux  We should remove markdown since it's only SEGFAULTing

theHamsta:
@Konan11221 would something like this work for you?
```lua
-- Setup call for users to override module configurations.
-- @param user_data module overrides
function M.setup(user_data)
  config.modules = vim.tbl_deep_extend('force', config.modules, user_data)

  require'nvim-treesitter.install'.ensure_installed(user_data.ensure_installed)

  recurse_modules(function(_, _, new_path)
    local data = utils.get_at_path(config.modules, new_path)
    local disable = {}
    vim.list_extend(disable, user_data.disable)
    vim.list_extend(disable, data.disable)
    if data.enable then
      enable_all(new_path, nil, disable)
    end
  end, config.modules)
end
```

lua/nvim-treesitter/configs.lua
theHamsta: Or with @Konan11221 solution, we can even keep it for people who like to experiment. Could you extend the README with a note which parsers are disabled by default. Thank you for your contribution.

steelsojka:
@theHamsta I think your solution is the way to go. I think the current implementation is flawed as it is.

steelsojka:
I'm also working on a PR that reworks the autocommand for modules. Right now there's an issue where disabling a module causes the autocommand to stop working. So it may overlap with this.

kyazdani42:
maybe just augrouping every autocommands of a module in its own scope would help when disabling, we would just disable the whole group.

theHamsta:
@Konan11221 to disable a language globally:
```lua
  require "nvim-treesitter.parsers".get_parser_configs().markdown = nil
```
before you call `setup`.

PR #373, opened 2020-09-01
docs: Add CODEOWNERS file

vigoux:
This will allow to trigger reviewers automatically, we will have to think about it and add people as
we go.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,7 @@
+/lua/nvim-treesitter/textobjects/ @theHamsta
+/lua/nvim-treesitter/incremental_selection.lua @theHamsta
+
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,7 @@
+/lua/nvim-treesitter/textobjects/ @theHamsta
+/lua/nvim-treesitter/incremental_selection.lua @theHamsta
```
theHamsta, line 2: doesn't @kyazdani42 want to take care of this?

theHamsta (approved):

.github/CODEOWNERS
vigoux: We'll discuss that later, that's just a draft

Issue #322, opened 2020-08-18
module 'vim.treesitter.query' not found


twnaing:
**Describe the bug**
module 'vim.treesitter.query' not found:

**To Reproduce**
1. open nvim
2. See error

```
Error detected while processing /home/user/.local/share/nvim/plugged/nvim-treesitter/plugin/nvim-treesitter.vim:
line   23:
E5108: Error executing lua ...nvim-treesitter/lua/nvim-treesitter/query_predicates.lua:1: module 'vim.treesitter.query' not found:
        no field package.preload['vim.treesitter.query']
        no file '/home/user/.config/nvim/lua/vim/treesitter/query.lua'
        no file '/home/user/.config/nvim/lua/vim/treesitter/query/init.lua'
```

**Expected behavior**
There should be no error and Neovim should be opened.

**Output of `:checkhealth nvim_treesitter`**

<details>
<code>

only blank page opened

</code>
</details>

**Output of `nvim --version`**
```
NVIM v0.5.0-627-g94b7ff730
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc-5 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/travis/build/neovim/bot-ci/build/neovim/build/config -I/home/travis/build/neovim/bot-ci/build/neovim/src -I/home/travis/build/neovim/bot-ci/build/neovim/.deps/usr/include -I/usr/include -I/home/travis/build/neovim/bot-ci/build/neovim/build/src/nvim/auto -I/home/travis/build/neovim/bot-ci/build/neovim/build/include
Compiled by travis@travis-job-9b5ea932-4f57-417a-8d5e-0a80f7a2a426

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/share/nvim"
```

[...]

PR #374, opened 2020-09-01
Add warning about required Neovim version

theHamsta:
I am not sure how to tackle #322 since `lua-match?` will introduce the next breaking change. But maybe we can catch with this people who try the plugin with Neovim stable.

### DIFF
--- a/lua/nvim-treesitter.lua
+++ b/lua/nvim-treesitter.lua
@@ -1,9 +1,15 @@
+--TODO(theHamsta): remove once stabilized!
+if not pcall(require,"vim.treesitter.query") then
+  error("nvim-treesitter requires a more recent Neovim nightly version!")
+end
+
 local install = require'nvim-treesitter.install'
 local utils = require'nvim-treesitter.utils'
 local ts_utils = require'nvim-treesitter.ts_utils'
 local info = require'nvim-treesitter.info'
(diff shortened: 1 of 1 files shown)


vigoux:
I was careful and tried not to introduce braking changes with `Lua-match?` so normally it should be good.

lua/nvim-treesitter.lua
```diff
@@ -1,10 +1,16 @@
+--TODO(theHamsta): remove once stabilized!
+if not pcall(require,"vim.treesitter.query") then
```
vigoux, line 2: I am sure we can have a better check than just that.

vigoux: Maybe using the output of `:version`, or `has()` ?

vigoux (changes requested):

lua/nvim-treesitter.lua
theHamsta: Can you suggest an improvement? With `has` we could only check for vim-patches and version can vary a lot depending on the built enviroment. Also: how can we compare Git SHAs?

This tests for that specific commit which I suppose is the oldest working version, but I can't tell whether we rely on even newer features.

vigoux (approved):

PR #376, opened 2020-09-02
fix(locals): js add arrow function parameter as definition

steelsojka:


theHamsta:
image

theHamsta (approved):

steelsojka:
Yeah, this one is weird. `formal_parmeters` is only for parameters wrapped in `()`

PR #379, opened 2020-09-02
Add a "Performance" section to docs

kontheocharis:
This pull request adds a "Performance" section to docs stating that the way nvim-treesitter checks `rtp` might result in slow startup times.

### DIFF
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -755,11 +755,12 @@ the cursor.
                                                             *hl-TSCurrentScope*
 Used by refactor.highlight_current_scope to highlight the current scope.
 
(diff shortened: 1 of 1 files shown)


vigoux (commented):
Correction : we search rtp for parsers and queries 😉

kontheocharis:
Noted, will update. I'm not too familiar with the code, I kind of skimmed through it just now.

vigoux (commented):
Last comment, we don't actually install them, maybe just "indexing" is the right word here.

vigoux (approved):
Awesome, looks like an awesome first contribution! Glad catched taht issue with the rtp!

vigoux:
CI is failing but that's not related to those changes.

PR #386, opened 2020-09-04
Fix filename

314eter:
`hightlights.scm` should be `highlights.scm`


vigoux (approved):

vigoux:
Thanks for that, nice catch !

PR #391, opened 2020-09-05
Support TSInstall on Windows with Clang

R-Michelsen:
I've made some changes to the install script so tree sitter languages can be installed on Windows with the clang compiler installed. I had some trouble getting #347 to work I think due to the cl compiler being quite a bit different from the gcc/clang counterparts. With clang most commands remain the same only some considerations have to be made like path separators atleast when using cmd.exe. Maybe it'd be preferable to write the command_list changes with powershell instead, not sure. 

Either way this PR works on my Windows 10 machine.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -62,14 +62,82 @@ local function select_executable(executables)
   return vim.tbl_filter(function(c) return fn.executable(c) == 1 end, executables)[1]
 end
 
+local function select_args(repo)
+  if fn.has('win32') then
+    return {
+      '-o',
+      'parser.so',
+      '-I./src',
+      repo.files,
+      '-shared',
+      '-Os',
+      '-lstdc++',
+    }
+  else
+    return {
+      '-o',
+      'parser.so',
+      '-I./src',
+      repo.files,
+      '-shared',
+      '-Os',
+      '-lstdc++',
+      '-fPIC'
+    }
+  end
+end
+
+local function select_install_rm_cmd(cache_folder, project_name)
+  if fn.has('win32') then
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/install.lua
```diff
@@ -62,14 +62,81 @@ local function select_executable(executables)
...
+      'parser.so',
+      '-I./src',
+      repo.files,
+      '-shared',
+      '-Os',
+      '-lstdc++',
+      '-fPIC'
+    }
+  end
+end
```
kyazdani42, line 88:
```lua
local args = {
      '-o',
      'parser.so',
      '-I./src',
      repo.files,
      '-shared',
      '-Os',
      '-lstdc++',
}
if not fn.has('win32') then
  table.insert('-fPIC')
end
return args
```

kyazdani42 (changes requested):
almost LGTM :)

R-Michelsen:
There is something odd going on that is causing the compilations to fail on the Linux side now, Im gonna try and investigate, its a bit difficult from the log and a Windows PC though

R-Michelsen (commented):
Seems reasonable, I've changed the code accordingly

vigoux:
CI reports an error, that will be blocking for this PR.

R-Michelsen:
> CI reports an error, that will be blocking for this PR.

I'm installing WSL to try and figure out what is going wrong on the Linux side when compiling parsers, the intended behaviour is to not change anything on the Linux side at all so Im not sure what is going on.

R-Michelsen:
Alright, everything seems good now. I messed up the fn.has('win32') checks, should be good now.

kyazdani42 (approved):
LGTM, let's wait for someone else to validate :)

Issue #378, opened 2020-09-02
Extending query files are applied for every query types


kyazdani42:
in `query.lua`, `base_language_map` and `query_extensions` are loaded for every type of queries (highlights, locals...), and i think it could be invalid in some cases (for instance, the `javascript/jsx.scm` file is highlight specific, but is loaded for the locals).
I don't have too much time to invest lately, so i put this here in case someone wants to tackle that.

vigoux:
This is actually a bug we need to work on, how about splitting those files and use something like `queries/jsx/*.scm`instead ?

kyazdani42:
mhhh that could actually be a good solution ! we then could provide folds for jsx also and not just highlighting

PR #404, opened 2020-09-06
feat(folds): add ts and js folds

steelsojka:


kyazdani42 (approved):
would typescript use javascript on this one too ? LGTM otherwise :)
PS: could you also add jsx :) ? would be great to fold groups of nested components !

steelsojka:
There will inherit the JS folds. I'll add jsx to this as well

steelsojka:
@kyazdani42 How does the jsx queries work? Are they just a part of the regular js files? I see a separate highlight file for jsx but I'm not sure how that is picked up.

vigoux (approved):

kyazdani42:
@steelsojka i think there might result in bugs see #378

PR #415, opened 2020-09-07
Cpp folds

theHamsta:
- add C++ folds
 - delete C++ locals contained in C locals

### DIFF
--- a/queries/c/fold.scm
+++ b/queries/c/fold.scm
@@ -4,4 +4,10 @@
  (while_statement)
  (translation_unit)
  (function_definition)
+ (struct_specifier)
+ (comment)
+ (preproc_if)
+ (preproc_elif)
+ (preproc_else)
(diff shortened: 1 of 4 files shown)


theHamsta:
Wouldn't it be easier to just use curly braces ` (compound_statement)` as fold marker?

queries/c/fold.scm
```diff
@@ -4,4 +4,9 @@
  (while_statement)
  (translation_unit)
  (function_definition)
+ (comment)
+ (preproc_if)
+ (preproc_elif)
+ (preproc_else)
+ (preproc_ifdef)
```
theHamsta, line 11: The hierarchy might be a bit of a problem if you have multiple `#else`

- if
 - else (folding the first else will fold all other else since they are children in the parse tree)
  - else
    - else

vigoux (approved):

PR #416, opened 2020-09-07
fix using highlight query extension for every query type

### LINKED ISSUE #378: Extending query files are applied for every query types

kyazdani42:
fixes #378


theHamsta (approved):

vigoux (approved):
Thanks mate, feels good to have you back!

PR #417, opened 2020-09-07
Add cross platform CI

theHamsta:


theHamsta:
Yeah! nvim comes bundled with `curl` on Windows. So we could download pre-built parsers.

theHamsta:
Adding this *might* push the built parsers to a pre-release
```yml
      - name: Compile parsers
        if: branch == 'master'
        run: |
           zip --junk-paths pre-built-parsers parser

      - name: Create Release
        if: branch == 'master'
        id: create_release
        uses: actions/create-release@v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          tag_name: parsers
          release_name: Pre-compiled parsers
          draft: false
          prerelease: true

      - name: Upload Release Asset
        if: branch == 'master'
        id: upload-release-asset 
        uses: actions/upload-release-asset@v1
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          upload_url: ${{ steps.create_release.outputs.upload_url }} # This pulls from the CREATE RELEASE step above, referencing it's ID to get its outputs object, which include a `upload_url`. See this blog post for more info: https://jasonet.co/posts/new-features-of-github-actions/#passing-data-to-future-steps 
          asset_path: ./pre-built-parsers.zip
          asset_name: pre-built-parsers.zip
          asset_content_type: application/zip
```

theHamsta:
`:TSInstallSync` does not work for Windows the moment. This is partially because the the `cmd` in `get_command` is not escaped like that: https://github.com/theHamsta/nvim-treesitter/blob/ci/lua/nvim-treesitter/install.lua but also because the clone command return immediately without being finished which causes `cd` of the compile command to fail.

lua/nvim-treesitter/install.lua
```diff
@@ -47,7 +47,7 @@ local function iter_cmd_sync(cmd_list)
       print(cmd.info)
     end
 
-    vim.fn.system(get_command(cmd))
+    print(vim.fn.system(vim.fn.escape(get_command(cmd), '\\')))
```
theHamsta, line 50: Maybe leave this print here? Otherwise, users will never see the output of the command. Could also be useful for finding errors when `:TSInstall` fails...

.github/workflows/tests.yml
```diff
@@ -0,0 +1,51 @@
...
+          ln -s $(pwd) ~/.local/share/nvim/site/pack/nvim-treesitter/start
+
+      - name: Prepare and compile Parsers (Windows)
+        if: matrix.os == 'windows-latest'
+        run: |
+          choco install neovim --pre
+          mkdir ~/AppData/Local/nvim/pack/nvim-treesitter/start
+          mkdir ~/AppData/Local/nvim-data
+          cp -r $(pwd) ~/AppData/Local/nvim/pack/nvim-treesitter/start
+          C:\\tools\\neovim\\Neovim\\bin\\nvim.exe --headless -c "TSInstall all" -c "sleep 120" -c "q"
```
theHamsta, line 41: We need probably a start_nvim script...

theHamsta:
Just had a request on reddit regarding binaries for Windows parsers. The Windows pipeline seems to be working. Only MacOS never terminates for some reason.

PR #421, opened 2020-09-08
fix(highlights): add missing js operators

steelsojka:


vigoux (approved):

queries/javascript/highlights.scm
```diff
@@ -113,23 +113,40 @@
...
+  ">"
+  ">="
+  ">>"
+  "||"
+  "??"
+  "%"
+  "%="
+] @operator
+
+(ternary_expression ["?" ":"] @operator)
```
kyazdani42, line 149: `?` is also an operator now. You can do `object?.value` (if object is null/undefined it will not throw and return undefined)

steelsojka: Yeah, but the javascript parser doesn't support it so it throws an error :(. It would technically be a delimiter though right?

steelsojka: `obj?.someProp` and null coalesing operator `??`

kyazdani42: it throws really ...? lets update the javascript parser then :) maybe issue this on the parser repo.

steelsojka: I opened an issue on the typescript parser a while back but then realized it should go in the JS parser.

It says the nodes are unrecognized so the query fails to parse.

steelsojka: I need to remove `??` for this to parse correctly :cry:

kyazdani42: jeez :(

steelsojka:
@vigoux Are the parsers pinned to a specific versions? I don't know if that got implemented yet or not.

vigoux:
Nope, not yet.

steelsojka:
I can't figure out why the query is failing in CI. It works fine locally. A couple of the operators were just added in the JS grammer that requires the latest parser.

PR #428, opened 2020-09-10
Add folding for local functions in lua

parmort:
I noticed that local functions weren't folding with the current head, so I poked around a bit and came up with this.


vigoux:
CI failed because of the ocaml parser update, I am rerunning the checks.

vigoux:
And welcome here by the way ! Thanks for this !

vigoux (approved):

Issue #363, opened 2020-08-30
"Plugin runtime path not found" error


errx:
**Describe the bug**
"Plugin runtime path not found" when trying to install parser

**To Reproduce**
TSInstall go

**Expected behavior**
parser should be installed

**Output of `:checkhealth nvim_treesitter`**

<details>
<code>

  health#nvim_treesitter#check
  ========================================================================
  ## Installation
    - OK: `git` executable found.
    - OK: `cc` executable found.
  
</code>
</details>

**Output of `nvim --version`**
```
NVIM v0.5.0-91109ffda
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/bin/clang -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNDEBUG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/tmp/neovim-20200828-46044-19k8zbb/build/config -I/tmp/neovim-20200828-46044-19k8zbb/src -I/usr/local/include -I/tmp/neovim-20200828-46044-19k8zbb/deps-build/include -I/usr/local/Cellar/unibilium/2.0.0/include -I/usr/local/Cellar/libtermkey/0.22/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include -I/usr/local/opt/gettext/include -I/tmp/neovim-20200828-46044-19k8zbb/build/src/nvim/auto -I/tmp/neovim-20200828-46044-19k8zbb/build/include
Compiled by o.matrohin@matrohin.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/Cellar/neovim/HEAD-91109ff/share/nvim"

Run :checkhealth for more info
```

**Additional context**
[...]

PR #432, opened 2020-09-10
Find package path based on the Lua source file

toupeira:
This is more robust compared to the previous method where we walked
up the tree and matched on the directory name, which also required
that the repository was cloned in a directory named `nvim-treesitter`.

Also see previous discussion at https://github.com/nvim-treesitter/nvim-treesitter/issues/363#issuecomment-687211730

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -200,9 +200,7 @@ local function install(with_sync, ask_reinstall)
       return api.nvim_err_writeln('Git is required on your system to run this command')
     end
 
-    local package_path, err = utils.get_package_path()
-    if err then return api.nvim_err_writeln(err) end
-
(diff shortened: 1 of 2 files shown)


lua/nvim-treesitter/install.lua
```diff
@@ -267,11 +265,7 @@ function M.uninstall(lang)
       M.uninstall(lang)
     end
   elseif lang then
-    local package_path, err = utils.get_package_path()
-    if err then
-      print(err)
-      return
-    end
+    local package_path = utils.get_package_path()
```
toupeira, line 268: I removed the error handling since this should only fail under very exceptional circumstances now, but let me know if you prefer keeping it for consistency.

vigoux (approved):
Awesome, decreasing reliance on the rtp is always a good thing for performances...

PR #453, opened 2020-09-12
C++ highlights: fix highlighting of methods

theHamsta:


queries/c/highlights.scm
```diff
@@ -116,7 +113,9 @@
 ]  @function.macro
 ; TODO (preproc_arg)  @embedded
 
-(field_identifier) @property
+((field_identifier) @property
+ (not-has-ancestor? @property call_expression)
+ (not-has-ancestor? @property function_declarator))
```
theHamsta, line 118: this is wrong. :-(

theHamsta:
Is adding a new predicate cheating? I somehow can't get a direct parent when I don't specify a node type.

queries/c/highlights.scm
```diff
@@ -103,9 +103,6 @@
 
 (call_expression
   function: (identifier) @function)
-(call_expression
-  function: (field_expression
-    field: (field_identifier) @function))
```
theHamsta, line 108: @vigoux This can only happen in C++, right? Or should we consider the case of a function pointer that can cause a expression of type `foo.bar()` in C.

theHamsta: I better leave that query for that case.

vigoux: `foo.bar()` can happen in C too indeed

lua/nvim-treesitter/query_predicates.lua
```diff
@@ -64,5 +74,16 @@ query.add_predicate('is?', function(match, pattern, bufnr, pred)
...
+  if not valid_args(pred[1], pred, 2) then return end
+
+  local node = match[pred[2]]
+  local types = {unpack(pred, 3)}
+
+  if not node then return true end
+
+  return vim.tbl_contains(types, node:type())
+end)
+
```
vigoux, line 87: No no no, use `[]` syntax for that.

lua/nvim-treesitter/query_predicates.lua
```diff
@@ -32,22 +32,32 @@ query.add_predicate("nth?", function(match, pattern, bufnr, pred)
...
+      node = node:parent()
+    end
   end
   return false
-end)
+end
+
+query.add_predicate('has-ancestor?', has_ancestor)
+
+query.add_predicate('has-parent?', has_ancestor)
```
vigoux, line 60: Why renaming here ?

vigoux (changes requested):

lua/nvim-treesitter/query_predicates.lua
theHamsta: what do you mean?

lua/nvim-treesitter/query_predicates.lua
theHamsta: two queries implemented by one function. the name determines the functionality

lua/nvim-treesitter/query_predicates.lua
theHamsta: @vigoux I didn't not quite understand what your opinion was to solve this. It's about having the same children tree but with different types as parents vs `[ ]` where you can have alternate subtrees but need to specify each subtree completely.

vigoux: Maybe we can have that as a workaround for some time, but this should be raised as an issue upstream. For me, the whole point of queries is to avoid strange predicates matching the type of a node.

theHamsta: Usually, the purpose of of using Lisp-syntax is to have no syntax at all. I find it much more elegant to define a predicate for the current application on hand (only local to this plugin) than to pollute upstream with dozens of special cases.

`has-type` (not used in this PR) could be expressed by `[ ]` https://github.com/tree-sitter/tree-sitter/issues/759#issuecomment-706551332 There is however some problems with how `[ ]` was designed which make it difficult where to place the common children. We can easily transition to the official solution.

theHamsta:
@vigoux can we have another attempt to merge this?

https://github.com/nvim-treesitter/nvim-treesitter/issues/662 had a issue about the type heuristics. I eliminated the case for uppercase functions but to find this out for scoped namespace highlights will be much more difficult because we have not real hint whether it's a constructor or a function or just a deeply namespace nested identifier. We've inherited the uppercase == typedefinition heursitics from upstream tree-sitter. I find it quite useful.

About the `[]` notation. It cannot be used for our purposes, I had a discussion with Max Brunsfield that the recommendation is to just copy multiple expressions with same subtree since it can't be expressed with the `[ ]` syntax without amibguities even if the  query code would be updated or extended.

queries/cpp/highlights.scm
```diff
@@ -23,15 +23,18 @@
 (template_function
   name: (identifier) @function)
 
-(template_method
-  name: (field_identifier) @method)
+(((field_expression
+     (field_identifier) @method)) @_parent
+ (has-parent? @_parent template_method function_declarator call_expression))
```
vigoux, line 28:
```suggestion
 (#has-parent? @_parent template_method function_declarator call_expression))
```

vigoux (changes requested):

queries/c/highlights.scm
```diff
@@ -119,7 +119,9 @@
 ]  @function.macro
 ; TODO (preproc_arg)  @embedded
 
-(field_identifier) @property
+(((field_expression
+     (field_identifier) @property)) @_parent
+ (not-has-parent? @_parent template_method function_declarator call_expression))
```
theHamsta, line 124: need to fix this.

vigoux (approved):

PR #454, opened 2020-09-12
do not highlight jsx_text as normal

kyazdani42:
highlighting with `@Normal` is conflicting with other groups, because normal also highlights the background color.
What happens then is that the CursorLine don't get properly highlighted on nodes where `@Normal` is applied.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,7 +72,8 @@ effect on highlighting. We will work on improving highlighting in the near futur
 
 ```
 @comment
-@error for error (ERROR` nodes.
+@error for error `ERROR` nodes.
+@none to disable completely the highlight
 @punctuation.delimiter for `;` `.` `,`
 @punctuation.bracket for `()` or `{}`
 @punctuation.special for symbols with special meaning like `{}` in string interpolation.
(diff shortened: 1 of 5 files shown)


theHamsta:
Couldn't you create a new highlight group for resetting on some things. Python also uses `@Normal`

kyazdani42:
i guess that would be ideal, not sure what name to give it though. Something like none would work ?

steelsojka (approved):
👍

theHamsta:
You need to document your query to make CI pass.

theHamsta (approved):

theHamsta (changes requested):
could you also use it for Python?

vigoux (approved):

stsewd (commented):
We should document this in CONTRIBUITING.md

theHamsta (approved):

kyazdani42:
@theHamsta I'll leave you handle the python side, i'm not too sure on which node i should apply this group.

stsewd (approved):

theHamsta:
@kyazdani42 the one that is `@Normal`

```scheme
; Reset highlighing in f-string interpolations
(interpolation) @Normal
```

theHamsta:
Thanks!

PR #472, opened 2020-09-15
feat: add parser for ocaml interface files

undu:
Currently ocaml interface files (.mli) are interpreted as standard ml files, which generates erroneous trees.

This PR activates the correct parsing of mli files when using the filetype `ocamlinterface`. This filetype does not exist in vim or neovim out of the box. The filetyppe can be manually set, there is a branch of [vim-ocaml](https://github.com/undu/vim-ocaml/tree/ocaml_interface) sets it out of the box.

Some upstream changes for the ocaml highlights are included. These currently don't provide the best highlighting possible because captures rule precedence in neovim doesn't work the same way as treesitter assumes.

I'd like to get some comments on what do people think it's the best approach for adding a new parser that neovim doesn't support out of the box because parser selection is more granular than existing filetypes.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -49,9 +49,10 @@ local function iter_cmd_sync(cmd_list)
       print(cmd.info)
     end
 
-    vim.fn.system(get_command(cmd))
+    local ret = vim.fn.system(get_command(cmd))
     if vim.v.shell_error ~= 0 then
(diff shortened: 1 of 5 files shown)


lua/nvim-treesitter/install.lua
```diff
@@ -49,9 +49,9 @@ local function iter_cmd_sync(cmd_list)
       print(cmd.info)
     end
 
-    vim.fn.system(get_command(cmd))
+    print(vim.fn.system(get_command(cmd)))
```
theHamsta, line 52: maybe print the output only on error

theHamsta (approved):

Issue #463, opened 2020-09-14
[Ruby highlight] require should be highlighted as TSInclude not TSFunction


bluz71:
Hello,

With [this Ruby file](https://github.com/bluz71/vim-nightfly-guicolors/blob/master/samples/ruby.rb) the `require` call at the top of the file is being highlighted as `TSFunction` yet it should be `TSInclude`. `require` is basically a Ruby code importation statement which is what the `TSInclude` highlight group exists for.

This is set in the `ruby/highlights.scm` file:

```
((identifier) @function
 (#eq? @function "require"))

"defined?" @function
```

Compare current Ruby Treesitter highlighting which is using `TSFunction` blue for the `require`:

<img width="551" alt="Screen Shot 2020-09-14 at 1 30 48 pm" src="https://user-images.githubusercontent.com/11382509/93040920-84ade800-f68e-11ea-8d88-b6fa2d743c7d.png">

Against Python Treesitter which is correctly highlighting imports in `TSInclude` dark red:

<img width="493" alt="Screen Shot 2020-09-14 at 1 31 53 pm" src="https://user-images.githubusercontent.com/11382509/93040972-aeffa580-f68e-11ea-9100-469d4991072e.png">

Note, Ruby also provides `require_relative` and `load` as less common code loading statements. You may also want to highlight those as `TSInclude` as well for consistency.

As best as I can tell, most Neovim Treesitter languages (Go, Rust, C, C++, Python) are now correctly using `TSInclude`, Ruby is the one that is out-of-step.

Best regards.

theHamsta:
@TravonteD 
@bluz71 thanks again for this issue! Your color theme looks really nice!

bluz71:
I appreciate the compliment.

I maintain two (Neo)vim color schemes: [moonfly](https://github.com/bluz71/vim-moonfly-colors) and [nightfly](https://github.com/bluz71/vim-nightfly-guicolors).

My aim is to have Neovim Treesitter look really nice since I genuinely believe it is the future of semantic editing and highlighting. I watched Max Brunsfeld's Treesitter talks a while ago and quickly became a believer in the technology.

Thanks to the team of coders doing the Neovim Treesitter work.

bluz71:
I just tested current `master` branch which contains #477.

Unfortunately it does not work with [this Ruby file](https://github.com/bluz71/vim-nightfly-guicolors/blob/master/samples/ruby.rb).

[...]

PR #477, opened 2020-09-16
update require statements to use TSInclude

TravonteD:
This should clean up #463

### DIFF
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -45,8 +45,8 @@
 
 ; Function calls
 
-((identifier) @function
- (#eq? @function "require"))
+((identifier) @include
+ (#vim-match? @include "^(require|require_relative|load)$"))
 
 "defined?" @function
 


queries/ruby/highlights.scm
```diff
@@ -45,8 +45,8 @@
 
 ; Function calls
 
-((identifier) @function
- (#eq? @function "require"))
+((identifier) @include
+ (#vim-match? @include "^(require|require_relative|load)$"))
```
vigoux, line 49: I don't really think it is worth using vim-match here
```suggestion
 (#lua-match? @include "^(require|require_relative|load)$"))
```

theHamsta: lua does not support `|`

vigoux: Huh, you are right, I still need to work on my lua-fu

vigoux (approved):

PR #486, opened 2020-09-18
[Ruby highlighting] Revert @include highlighting for require

bluz71:
In #463 I advocated that require/require_relative & load should be
highlighted as `@include`.

Turns out that is not a good idea.

require is not a keyword, it is an actual function, and it can be
overriden as a class method.

For example in Rails strong params is invoked via a require method, an
example:

```ruby
  def album_params
    params.require(:album).permit(:title, :genre_id, :year, :track_list,
                                  :cover, :cover_cache)
  end
```

With the PR #477 the the require above would be highlighted as
TSInclude, not what we want.

In the end it is best to stay with the original function highlighting
for require.

I apologize for needlessly going down this rabbit hole.

### DIFF
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -45,8 +45,8 @@
 
 ; Function calls
 
-((identifier) @include
- (#vim-match? @include "^(require|require_relative|load)$"))
+((identifier) @function
+ (#eq? @function "require"))
 
 "defined?" @function
 


vigoux:
Little suggestion.
Maybe the `require` that you want to highlight is a different node than the`require` called as method, and you can play on that to fix you issue?

Thanks for contributing by the way!

bluz71:
In Ruby, importation via something like:

```ruby
require "pry"
```

That actually is a method/function call, noting that parenthesis is optional in Ruby, so you could also do:

```ruby
require("pry")
```

I think we end up on a slippery slope by highlighting some methods in one color depending on context compared with another. We may end up with spaghetti code.

Note, I would like `require` highlighted in `TSInclude`, but I would not want to make Neovim Treesitter needlessly ugly for edge cases like this.

I still favour reverting the code since at the end of the day `require` is a method/function call which has the side effect of loading code.

If `require` was a Ruby keyword, then we could absolutely highlight it in `TSInclude`, but its not a keyword.

vigoux:
I am not should it is actually an edge case.
`require` here can actually have two different _meanings_ which is what we aim for.

bluz71:
> I am not should it is actually an edge case. 
> require here can actually have two different meanings which is what we aim for.

I was playing devil's advocate and being conservative.

Your suggestion would produce a nicer result.

I myself don't know how to do that. So the way forward is to close this, and have `require` have two meanings.

bluz71:
Closing, @vigoux wants `require` to have two meanings, either as a top-level `TSInclude` or as a `TSFunction`.

Treesitter experts will need to figure that out, I am out of my depth in that regard.

PR #488, opened 2020-09-18
Add case builtin for dart

akinsho:
Based on https://github.com/UserNobody14/tree-sitter-dart/pull/10 to unhide the `case` node
This PR also shows `Function` as a type rather than a keyword. `Function` is used as a type annotation but I've previously incorrectly included it as a keyword which isn't correct.

Lastly switch to using the existing annotation highlight for annotations (duh) rather than the no longer existing `attribute` annotation

### DIFF
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -29,6 +29,7 @@ hlmap["number"] = "TSNumber"
 hlmap["boolean"] = "TSBoolean"
 hlmap["float"] = "TSFloat"
 hlmap["annotation"] = "TSAnnotation"
+hlmap["attribute"] = "TSAttribute"
 
 -- Functions
 hlmap["function"] = "TSFunction"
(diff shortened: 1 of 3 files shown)


vigoux:
I don't think that `@annotation` is supported anymore hehe.

akinsho:
@vigoux what should the correct highlight be for something like 

```
@override
existingClassMethod()
```

I think it should map to `PreProc` but I don't know what the correct TS highlight for that would be. Annotation still seems to be in the code base and worked locally

vigoux:
I think that `@attribute` is good, as we made it this was.

akinsho:
@vigoux `TSAttribute` doesn't actually exist anymore 🤔 if you search the code base there are no longer any usages. I think it was removed at some point since I think it existed when I last checked. I assumed someone had removed it because it wasn't wanted anymore?

akinsho:
I can re-add it and link it to `PreProc` and remove the annotation highlight?

Issue #502, opened 2020-09-19
gcc fails to compile parsers on Windows: not a valid WIN32 application


osherbz:
**Is your feature request related to a problem? Please describe.**
On windows system setting-up nvim-treesitter was a little problem for me, because I got some error like :
1. uv_dlopen cant find module
2. *.so is not valid WIN32 application.

I fixed it by removing gcc/g++ compilers from path and using clang.

**Describe the solution you'd like**
Add this information to troubleshooting on windows or windows wiki.

**Describe alternatives you've considered**
When Using `:TSUpdate` etc... I would love if a little menu will shown up and show available compilers and choose from them.
like this for example:
:TSUpdate 
-- compilers from path
-- 1. gcc/g++ (default)
--  2. clang 
-- Enter compiler number:

**Additional context**
I am using gcc/g++ many times, every time that I want to TSUpdate I will need to remove it from path.
It will be very comfortable if there will be menu like i mentioned above.

Thanks in advance for any help you are able to provide.

theHamsta:
You can edit the Windows wiki entry to add this information. After all it's a a wiki where everybody can share their knowledge and recommendations.

Should we prefer clang over gcc? `-fPIC` was removed for gcc on Windows, have you tried compiling with gcc with `-fPIC`.

theHamsta:
I would prefer fixing the issue with gcc over asking the user every time for a compiler. In a ideal world, any compiler should work for nvim-treesitter.

osherbz:
> You can edit the Windows wiki entry to add this information. After all it's a a wiki where everybody can share their knowledge and recommendations. Could you add instruction on how to install clang on Windows? Probably you can describe best since you are using Windows. When I was writing the article I had to borrow a foreign computer with Windows installed.
> 
> Should we prefer clang over gcc? `-fPIC` was removed for gcc on Windows, have you tried compiling with gcc with `-fPIC`?

Yes I will add information to the wiki and also add clang install instructions.
Now I am checking the -fPIC flag.

osherbz:
@theHamsta  I added the clang installation and troubleshooting on Windows wiki, but I dont know how to push it.
I tried to fork https://github.com/nvim-treesitter/nvim-treesitter.wiki.git and didnt work.

Should I Send the updated Windows-support.md here?

theHamsta:
@osherbz sorry! I didn't know that the Wiki is read only for contributers. You can post your updated file here. We try to give everyone permission ASAP.

[...]

PR #503, opened 2020-09-19
feat(install): make compiler selectable

theHamsta:
Compilers are selectable via `'nvim-treesitter.install'.compilers` (list of compilers) or environment variable CC

Addresses #502

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -8,6 +8,8 @@ local info = require'nvim-treesitter.info'
 
 local M = {}
 
+M.compilers = { vim.fn.getenv('CC'), "cc", "gcc", "clang" }
+
 function M.iter_cmd(cmd_list, i, lang, success_message)
   if i == #cmd_list + 1 then return print(success_message) end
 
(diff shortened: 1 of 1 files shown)


theHamsta:
Documentation for this will go to Windows wiki. On Linux/MacOS it shouldn't really matter which C compiler is used. `CC` is mostly the favorite compiler of the user.

kyazdani42 (approved):
lgtm :)

theHamsta:
@osherbz FYI, with this you can just use 

```lua
require 'nvim-treesitter.install'.compilers = { "clang" }
```

The new default would be
```
M.compilers = { vim.fn.getenv('CC'), "cc", "gcc", "clang" }
```
Do you think we should prefer `clang` due to the problems you experienced? gcc does not install `cc` but it's currently preferred over clang.

osherbz:
@theHamsta 

No, I personally think `require 'nvim-treesitter.install'.compilers = { "clang" }` will be better. gcc is more common.
We should add this information into windows-support or to README.md.
And updating window-support troubleshooting  (instead of removing compilers form path we will specify the compiler by your command).

osherbz:
@theHamsta  Can you make the wiki editable? please, I want to update the windows-support.md

theHamsta:
@kontheocharis I don't have the rights to make the wiki editable. I already asked `@vigoux`.

kontheocharis:
I assume you tagged me by accident... @theHamsta

PR #506, opened 2020-09-19
Install parsers with `repo.url` as a local path

theHamsta:
This is mainly thought as an internal feature for parser developers to install them from a local path

E.g. when you are working on a Lisp parser:

```
  require "nvim-treesitter.parsers".get_parser_configs().lisp = {
    install_info = {
      url = "~/projects/tree-sitter-lisp",
      files = {"src/parser.c"}
    }
  }
```
and then `:TSInstall lisp`/`:TSUpdate lisp` to compile and install the parser and `:TSInstallFromGrammar lisp` to run `tree-sitter generate` before compiling.

Depends on #503 (only because I edited the same code region).

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -101,16 +101,31 @@ local function iter_cmd_sync(cmd_list)
   return true
 end
 
-local function run_install(cache_folder, install_folder, lang, repo, with_sync)
+local function run_install(cache_folder, install_folder, lang, repo, with_sync, generate_from_grammar)
   parsers.reset_cache()
 
   local path_sep = utils.get_path_sep()
 
   local project_name = 'tree-sitter-'..lang
+  local maybe_local_path = vim.fn.expand(repo.url)
+  local from_local_path = vim.fn.isdirectory(maybe_local_path) == 1
+  if from_local_path then
+    repo.url = maybe_local_path
+  end
+
   -- compile_location only needed for typescript installs.
-  local compile_location = cache_folder..path_sep..(repo.location or project_name)
+  local compile_location
+  if from_local_path then
+    compile_location = repo.url
+  else
+    compile_location = cache_folder..path_sep..(repo.location or project_name)
+  end
   local parser_lib_name = install_folder..path_sep..lang..".so"
 
+  if generate_from_grammar and vim.fn.executable('tree-sitter') ~= 1 then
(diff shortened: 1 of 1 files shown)


theHamsta:
Would be even better if we would `uv_close` the parser shared object and reload on update.

lua/nvim-treesitter/install.lua
```diff
@@ -101,16 +101,30 @@ local function iter_cmd_sync(cmd_list)
...
-local function run_install(cache_folder, install_folder, lang, repo, with_sync)
+local function run_install(cache_folder, install_folder, lang, repo, with_sync, generate_from_grammar)
   parsers.reset_cache()
 
   local path_sep = utils.get_path_sep()
 
   local project_name = 'tree-sitter-'..lang
+  local from_local_path = vim.fn.isdirectory(vim.fn.expand(repo.url)) == 1
+  if from_local_path then
+    repo.url = vim.fn.expand(repo.url)
```
vigoux, line 112: Could you avoid calling `expand` twice here ?

lua/nvim-treesitter/install.lua
```diff
@@ -121,8 +135,23 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync)
...
-  vim.list_extend(command_list, shell.select_download_commands(repo, project_name, cache_folder, revision))
+  if not from_local_path then
+    vim.list_extend(command_list, { shell.select_install_rm_cmd(cache_folder, project_name) })
+    vim.list_extend(command_list, shell.select_download_commands(repo, project_name, cache_folder, revision))
+  end
+  if generate_from_grammar then
+    vim.list_extend(command_list, {
+      {
+        cmd = 'tree-sitter',
+        info = 'Generating source files from grammer.js...',
```
vigoux, line 146:
```suggestion
        info = 'Generating source files from grammar.js...',
```

lua/nvim-treesitter/install.lua
theHamsta: fixed

lua/nvim-treesitter/install.lua
theHamsta: fixed

theHamsta:
An upgrade for this PR would be to offer the installation of tree-sitter-cli if not installed. But this would be another PR.

vigoux (approved):

PR #507, opened 2020-09-19
fix(python): use `@Normal` again for f-strings `@none` is not aggressive enough

theHamsta:
We really want to reset all potential highlighting from the string here.

An alternative would be to add explicit highlighting of `@variable` but this wouldn't reset potential background highlighting of the string.


stsewd (approved):
Cursor line bug is back with this :/

PR #512, opened 2020-09-21
Improve OCaml highlighting and add locals and folds

314eter:
Some small tweaks to the highlighting queries:
- Change `#contains?` to `#match?` (otherwise every type that contains "int" will be considered builtin)
- Do not capture `label_name` as `@parameter` (when necessary, like in `let f ~x = x`, it's a `value_pattern`)

I expanded the local queries. Is it correct that the `reference.kind` is currently ignored?


theHamsta (approved):

theHamsta:
It's ignored bat the moment but could be used to only search for a definition of same kind.

PR #514, opened 2020-09-21
Python locals: account for expression_list -> left_hand_side

theHamsta:
https://github.com/tree-sitter/tree-sitter-python/pull/83 (was only merged a few minutes ago)

BREAKING CHANGE!

### DIFF
--- a/queries/python/locals.scm
+++ b/queries/python/locals.scm
@@ -78,15 +78,13 @@
 
 ;;; Assignments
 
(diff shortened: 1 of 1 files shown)


vigoux (approved):

theHamsta:
Wasn't that a bit fast? I meant merging this is a breaking change. The change of the parser only caused definitions to look different.

PR #515, opened 2020-09-21
Remove textobjects module

theHamsta:
Fix https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/7


lua/nvim-treesitter/query.lua
```diff
@@ -111,9 +111,13 @@ function M.has_query_files(lang, query_name)
   return false
 end
 
-function M.get_query(lang, query_name)
+function M.get_query_string(lang, query_name)
```
theHamsta, line 114: exposing this could be handy for some situations. I'm using it so see what queries are used for which language,

vigoux: Why do you need to know that ? Why not using `has_*`functions ?

theHamsta: Just docgen. Which subset of the  queries are available in which language. Has a languages for instance the `@class.outer` textobject

vigoux: Then you can use `query.captures` with and already parsed query, this is the table of defined captures within the query.

theHamsta: But then I need the parsers available. But would also be possible to compile them beforehand. Right now it's a plain find on a string.

stsewd (approved):

lua/nvim-treesitter/query.lua
theHamsta: New version uses `query.captures` but it requires compiling the parsers beforehand.

vigoux:
Is this ready?

theHamsta:
Yes! Just merge it together with "remove-refactor" and a note where to find the modules. E.g. in readme, zulip.

vigoux:
Can you add those information already? So that the pr is self contained

theHamsta:
The README section for external modules is in https://github.com/nvim-treesitter/nvim-treesitter/pull/487. So this PR kind of depends on it. You could just add slug for `nvim-treesitter-textobjects` to https://github.com/nvim-treesitter/nvim-treesitter/pull/487

theHamsta:
I updated the README for #515 #487 . Maybe this will #487 needs to be updated for that.

theHamsta:
Should I just merge both PRs?

Issue #516, opened 2020-09-22
[Highlight Proposal] Create new TSNamespace highlight group for 'namespace' language concept


bluz71:
Hello Neovim Treesitter team,

At the moment C++ namespaces such as `std` are highlighted as `TSConstant`. With my [moonfly](https://github.com/bluz71/vim-moonfly-colors) and [nightfly](https://github.com/bluz71/vim-nightfly-guicolors) themes the results are not ideal.

Recently we had an issue where HTML tags where being highlighted as `TSType` which also was not ideal. New `TSTag` and `TSTagDelimiter` highlight groups resulted from that discussion. This was a big improvement. I believe we have another similar issue here.

Namespaces in programming languages provide a scope to partition logic and functionality under a chosen name to avoid collisions. The most well-known example would be C++, here is an example:

```cpp
using std::cout;
using std::endl;

namespace box1 {
    int box_side = 4;
}

namespace box2 {
    int box_side = 12;
}

int main() {
    int box_side = 42;
    cout << box1::box_side << endl;  // Outputs 4.
    cout << box2::box_side << endl;  // Outputs 12.
    cout << box_side << endl;  // Outputs 42.
}
```

`std`, `box1` and `box2` are namespaces. Ideally in Treesitter I would like to highlight those in a custom color, not as `TSConstant`. In my opinion namespaces are not constants.

Namespaces are not limited to C++.

Ruby has an equivalent mechanism with `module` as seen in this snippet:

```ruby
require 'digest'

module Encryption
  def encrypt(string)
    Digest::SHA2.hexdigest(string)
  end
end
```

`Encryption` and `Digest` are effectively namespaces.

I am not a Java person, but it may be the case the Java packages could be considered namespaces.

C# has namespaces:

```c#
using System;

Console.WriteLine("Hello World!");
int i = Convert.ToInt32("123");
```

`System` is considered a namespace.

Python modules might also be considered namespaces:

```python
import modulea

modulea.func1()
modulea.func2()
a = modulea.class1()
```

`modulea` may be considered a namespace?

Lua may also be a candidate for possible `TSNamespace` highlighting:

```lua
local M = {}

local function test() print(123) end
function M.test1() test() end
function M.test2() M.test1(); M.test1() end

return M
```

`M` may be considered a namespace.

I guess this issue is a request for comment much like `TSTag` was. Feedback welcome.

[...]

PR #518, opened 2020-09-22
fix(highlights): Add TSNamespace highlight

theHamsta:
Start adding highlights for
     - C++
     - Rust (including other scoped_identifier/scoped_type_identifier fixes)
     - JS (only namespace_import)

Addresses #516

Vim has no dedicated hl-group for modules/namespace continued to use Constant (open for better suggestions)

#453 could be useful for Rust highlights. In the queries it can almost always be `scoped_identifier` or `scoped_type_identifier` as a parent.

### DIFF
--- a/plugin/nvim-treesitter.vim
+++ b/plugin/nvim-treesitter.vim
@@ -41,6 +41,7 @@ highlight default link TSProperty Identifier
 highlight default link TSConstructor Special
 highlight default link TSAnnotation PreProc
 highlight default link TSAttribute PreProc
+highlight default link TSNamespace Include
 
 highlight default link TSConditional Conditional
 highlight default link TSRepeat Repeat
@@ -53,7 +54,6 @@ highlight default link TSException Exception
 
 highlight default link TSType Type
 highlight default link TSTypeBuiltin Type
-highlight default link TSStructure Structure
 highlight default link TSInclude Include
 
 highlight default link TSVariable TSNone
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -13,11 +13,15 @@
 (type_identifier) @type
 (primitive_type) @type.builtin
 (field_identifier) @field
-
+(mod_item
+ name: (identifier) @namespace)
 
 ; Function calls
 (call_expression
   function: (identifier) @function)
+(call_expression
+  function: (scoped_identifier 
+              (identifier) @function .))
 (call_expression
   function: (field_expression
     field: (field_identifier) @function))
@@ -36,6 +40,16 @@
  (#match? @constant "^[A-Z]"))
 
 ; Assume that uppercase names in paths are types
+(scoped_identifier
+  path: (identifier) @namespace)
+(scoped_identifier
+ (scoped_identifier
+  name: (identifier) @namespace))
+(scoped_type_identifier
+  path: (identifier) @namespace)
+(scoped_type_identifier
+ (scoped_identifier
+  name: (identifier) @namespace))
 ((scoped_identifier
   path: (identifier) @type)
  (#match? @type "^[A-Z]"))
@@ -43,6 +57,13 @@
     name: (identifier) @type)
  (#match? @type "^[A-Z]"))
 
+(crate) @namespace
+(scoped_use_list
+  path: (identifier) @namespace)
+(scoped_use_list
+  path: (scoped_identifier
+            (identifier) @namespace))
+(use_list (scoped_identifier (identifier) @namespace . (_)))
 (use_list (identifier) @type (#match? @type "^[A-Z]"))
 (use_as_clause alias: (identifier) @type (#match? @type "^[A-Z]"))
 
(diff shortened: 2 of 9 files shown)


bluz71:
> Vim has no dedicated hl-group for modules/namespace continued to use Constant (open for better suggestions)

I am now of the opinion that `Structure` is the appropriate generic highlight group to use as the default for `TSNamespace`.

From the [Namespace](https://en.wikipedia.org/wiki/Namespace) Wikipedia entry:

> Namespaces are commonly **structured** as hierarchies to allow reuse of names in different contexts.

> It is possible to use anonymous **structs** as namespaces in C since C99

Yes, I am stretching it, but we can justify the use of `Structure` for namespaces via the above Wikipedia notes. I feel it is a better group than `Constant` for namespaces.

bluz71:
I just test @theHamsta's new `TSNamespace` branch and it works very well for C++ and Rust.

A definite improvement when theming Treesitter.

Many thanks.

theHamsta:
The documentation for `TSStructure` was not really ideal. Nobody knew what is was for. So it only had one usage.

bluz71:
> The documentation for `TSStructure` was not really ideal. Nobody knew what is was for. So it only had one usage.

`TSStructure` probably shouldn't even exist if it only has one use in OCaml files. It may well be the case the OCaml module names (currently the only use of `TSStructure`) could go over to `TSNamespace`? I don't know OCaml well enough to know if that is a good or terrible idea.

Anyway, `TSNamespace` defaulting to `Structure` seems a reasonable choice.

theHamsta:
Vim doc reads:

```
	 Structure	struct, union, enum, etc.
```

bluz71:
Those Vim docs about syntax groups where likely written 20 years ago when Bram only cared about the C language.

The docs for `Constant` state:

`Constant  any constant`

It's a choice about which default makes the makes sense for `TSNamespace` and will look pretty good with most existing Vim themes. Using the same color for `struct, union, enum` will produce a good result.

theHamsta:
Most colorschemes I experiment with just use keyword highlighting for structure. Not a good choice for an identifier also because it will be the same color as `namespace` itself.

My choice came mainly from using the same highlights as `vim-modern-cpp`
```
hi def link cppSTLnamespace    Constant
```

bluz71:
Ok, that’s a reasonable point, I step back.

Note, with my themes I am already setting custom `TSNamespace` colors in anticipation of this enhancement. Result is nice.

Best regards.

plugin/nvim-treesitter.vim
```diff
@@ -41,6 +41,7 @@ highlight default link TSProperty Identifier
 highlight default link TSConstructor Special
 highlight default link TSAnnotation PreProc
 highlight default link TSAttribute PreProc
+highlight default link TSNamespace Constant
```
vigoux, line 44: Why adding twice?

theHamsta: maybe that was why it was still a draft :-)

queries/rust/highlights.scm
```diff
@@ -60,8 +81,10 @@
   (#eq? @type "derive"))
 
 (macro_invocation
-  macro: (identifier) @function.macro
-  "!" @function.macro)
```
theHamsta, line 64: "!" was overwritten anyways in the current queries.

queries/rust/highlights.scm
```diff
@@ -36,13 +40,30 @@
...
+(scoped_identifier
+  path: (identifier) @namespace)
+(scoped_identifier
+ (scoped_identifier
+  name: (identifier) @namespace))
+(scoped_type_identifier
+  path: (identifier) @namespace)
+(scoped_type_identifier
+ (scoped_identifier
+  name: (identifier) @namespace))
```
theHamsta, line 52: It's possible that it was a draft to wait for #453 .
As far as I understood you can do
```lisp
[(scoped_identifier
  path: (identifier) @namespace)
(scoped_type_identifier
  path: (identifier) @namespace)]
```
but not 

```lisp
([scoped_identifier scoped_type_identifier]
  path: (identifier) @namespace)
```
after the emacs tree-sitter creator complained. This is partially addressed by matching on supernodes.

vigoux (approved):

undu:
Regarding ocaml and the usage of structure: with regex-based highlighting module names are regarded as includes, structures matched that in the colorscheme I used, seeing that it kept the colour and more-or-less matched the namespace use-case I I ended up using in ocaml.

I would ask to use `Include` instead of `Constant`, as namespaces/modules are more related to including than to a constant. At least for ocaml code choosing the same highlight for namespaces and constants would make them stand out less in the code, when they are both important to recognise.

plugin/nvim-treesitter.vim
vigoux, line 44: As @undu suggested, it sounds better to have that highlighted as `Include` instead of `Constant`.

theHamsta: tested the last 24h with different color schemes. `Include` looks generally good although it is close to functions in some color schemes.

PR #522, opened 2020-09-23
feat: add TSConfigInfo to display current config

theHamsta:
https://nvim-treesitter.zulipchat.com/#narrow/stream/252271-general/topic/Random/near/210929394

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -223,6 +223,18 @@ local function recurse_modules(accumulator, root, path)
   end
 end
 
+-- Shows current configuration of all nvim-treesitter modules
+-- @param process_function function used as the `process` parameter
+--        for vim.inspect (https://github.com/kikito/inspect.lua#optionsprocess)
+local function config_info(process_function)
+  process_function = process_function or function(item, path)
+     if path[#path] == vim.inspect.METATABLE then return end
+     if path[#path] == "is_supported" then return end
+     return item
+  end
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/configs.lua
```diff
@@ -222,6 +222,15 @@ local function recurse_modules(accumulator, root, path)
...
 end
 
+function M.config_info(process_function)
+  process_function = process_function or function(item, path)
+     if path[#path] == vim.inspect.METATABLE then return end
+     if path[#path] == "is_supported" then return end
+     return item
+  end
+  print(vim.inspect(config, {process = process_function}))
+end
```
theHamsta, line 232: could also make this a local function

vigoux: I'd rather have in as a local function.

kyazdani42:
i was wondering if using a buffer to display stats about the plugin would be better.

vigoux:
> i was wondering if using a buffer to display stats about the plugin would be better.

I don't think so, just having everything dumped in the message window is fine IMO.

theHamsta:
I considered using a read-only floating window..

kyazdani42:
dumping everything is just not working for me on small screens, we cannot see anything at all, it's all messed up, that's why i was suggesting otherwise.

kyazdani42:
take the vim plug way of displaying status, it just works great for me on any screen :)

kyazdani42 (approved):

vigoux (approved):

PR #525, opened 2020-09-23
dart: improve the highlighting for class instantiation

akinsho:
So it finds `_HiddenClasses` as well as `OtherClasses`.

This also adds highlighting for the punctuation in a template literal, previously this wasn't working when specifying `"${"` as one node which causes the parser to error but if each character is captured separately that works 🤷🏿‍♂️. I've asked on the upstream repo but it's pretty inactive.

### DIFF
--- a/queries/dart/highlights.scm
+++ b/queries/dart/highlights.scm
@@ -1,25 +1,25 @@
 (dotted_identifier_list) @string
 
 ; Methods
-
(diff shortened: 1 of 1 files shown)


theHamsta (approved):

theHamsta:
Thanks for the contribution! 👍

PR #527, opened 2020-09-24
tell the user when installation is complete

sanga:
For me the final message I get is `Treesitter parser for haskell has been installed` which somehow doesn't give me a warm fuzzy feeling that everything installation is done for all parsers. So this PR just scratches my OCD itch. Totally untested btw, this is a quick drive-by PR.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -231,6 +231,7 @@ local function install(with_sync, ask_reinstall)
     for _, lang in ipairs(languages) do
       install_lang(lang, ask, cache_folder, install_folder, with_sync)
     end
+    print('Treesitter installation complete')
   end
 end
 


stsewd (approved):
Thanks! Failure is unrelated to this PR, looks like a change in a grammar.

vigoux:
Hmm... `Treesitter installation completed` isn't what I'd go for as treesitter is actually already installed.

kyazdani42:
i agree with vigoux, maybe something like `treesitter parsers were installed property` sounds better imo

lua/nvim-treesitter/install.lua
```diff
@@ -231,6 +231,7 @@ local function install(with_sync, ask_reinstall)
     for _, lang in ipairs(languages) do
       install_lang(lang, ask, cache_folder, install_folder, with_sync)
     end
+    print('Treesitter installation complete')
```
kyazdani42, line 234: if this run asynchronously, wouldn't this print before the parsers are all installed ?

theHamsta:
Maybe the installer could just count the number of total parsers to be compiled. We could the display " haskel parser has installed [25/29]" "29 parsers have been installed!".

sanga:
Apologies for this (note to self: maybe don't try to code anything with the flu). Yes, "treesitter parsers were installed property" sounds much better to me also. Or I like @theHamsta suggestion even better

theHamsta:
@sanga implemented the progress display in #351

sanga:
Cheers @theHamsta :bow:

PR #532, opened 2020-09-26
Add Kotlin parser

theHamsta:
https://github.com/QthCN/tree-sitter-kotlin is a fork of https://github.com/fwcd/tree-sitter-kotlin with some recent updates. The parser is working quite well on all the files that I tried. Thought, that it you be a nice task for Hacktober fest to add some queries.

fwcd made a Kotlin language server and a debug adapter for Kotlin.
Probably, the language server is using this parser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [jsdoc](https://github.com/tree-sitter/tree-sitter-jsdoc) (maintained by @steelsojka)
 - [x] [json](https://github.com/tree-sitter/tree-sitter-json) (maintained by @steelsojka)
 - [ ] [julia](https://github.com/tree-sitter/tree-sitter-julia)
+- [ ] [kotlin](https://github.com/QthCN/tree-sitter-kotlin)
 - [x] [lua](https://github.com/nvim-treesitter/tree-sitter-lua) (maintained by @vigoux)
 - [x] [nix](https://github.com/cstrahan/tree-sitter-nix) (maintained by @leo60228)
 - [x] [ocaml](https://github.com/tree-sitter/tree-sitter-ocaml) (maintained by @undu)
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -111,6 +111,13 @@ list.java = {
   maintainers = {"@p00f"},
 }
 
+list.kotlin = {
+  install_info = {
+    url = "https://github.com/QthCN/tree-sitter-kotlin",
+    files = {"src/parser.c"}
+  }
+}
+
 list.html = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-html",


Happy-Dude:
If these parsers are readily available on GitHub, what do y'all think of adding them via git submodule and using `git clone --recursive` during clone or `git submodule update -init` to pull them down once cloned?

theHamsta:
I thought about it, also for pinning the dependencies. In general, it is easier for users and for us to configure (you can add and remove parsers just by changing your config) and inspect which parsers are available and to set extra arguments on how to compile them (later information is not available when setting a git submodule only). This was why a Lua config was sufficient for us in the beginning.

The version dependency management could be an argument for submodules...

Happy-Dude:
Forgive my ignorance on this; where does nvim-treesitter save the pulled-down parsers to?

If it's similar to how [vim-plug](https://github.com/junegunn/vim-plug/blob/master/plug.vim) does a `git clone` for install, then the user themselves can take it upon themselves to make these submodules, like how I did [here](https://github.com/Happy-Dude/dotfiles/tree/master/vim/.vim/pack/plugged/opt). The more simple it is, the more "best of all worlds" can be left to either the addon or user to manage, or both!

I agree with the concern of version/dependency management though, which I hadn't thought of earlier.

theHamsta:
`nvim-treesitter` was designed to be flexible about this. `nvim` and `nvim-treesitter` accepts all parsers called `language.so` in a directory `parser` somewhere in your runtime path.

This means you can install a parser with vim-plug
```
  Plug "tree-sitter/tree-sitter-c", {"do:" {"mkdir -p parser && cc src/parser.c -I/src -o  parser/c.so -fPIC -Os -lstdc++"}
``` 
I actually did it this way before `nvim-treesitter` existed.

tormodatt:
can we get this merged? If so, I have a somewhat working highlights.scm-query for this parser (current fwcd version)

theHamsta:
Nice! @tormodatt you can also take the changes of this PR for your own. Would really love to have Kotlin highlights.

README.md
```diff
@@ -141,15 +141,16 @@ We are looking for maintainers to add more parsers and to write query files for
 - [x] [dart](https://github.com/UserNobody14/tree-sitter-dart) (maintained by @Akin909)
 - [ ] [elm](https://github.com/razzeee/tree-sitter-elm)
 - [x] [fennel](https://github.com/travonted/tree-sitter-fennel) (maintained by @TravonteD)
-- [x] [graphql](https://github.com/bkegley/tree-sitter-graphql) (maintained by @bkegley)
 - [x] [go](https://github.com/tree-sitter/tree-sitter-go) (maintained by @theHamsta, @WinWisely268)
+- [x] [graphql](https://github.com/bkegley/tree-sitter-graphql) (maintained by @bkegley)
```
kyazdani42, line 145: why is there a diff here ^^' ?

kyazdani42 (changes requested):

README.md
theHamsta: It's on all PRs merged after the graphql PR. It's from the GH actions bot. Needs a rebase.

tormodatt:
will do! I might need to do some changes if we are going to use the fork though (QthCN/tree-sitter-kotlin). do you know what changes that are made there compared to fwcd @theHamsta ? maybe we should try to get the changes upstream so that we could use the fwcd-version that probably will be more maintained?

theHamsta:
Choose the version you like (upstream or fork). I had the impression the upstream is dead and that the fork did some reasonable although minor changes.

theHamsta:
closing since @tormodatt is interested in completing this PR by adding queries for a kotlin parser.

tormodatt:
I found the QthCN to be the best parser for highlighting as well after some more investigation. created a PR here: https://github.com/nvim-treesitter/nvim-treesitter/pull/729

PR #534, opened 2020-09-27
Installation of maintained parsers only

theHamsta:
Unmaintained parsers only give users little benefit but take sometimes a
a long time to install (e.g. Markdown, Julia, Haskell parser). We could
recommend to only install maintained parsers by default.

So instead of `ensure_installed = 'all'` or `:TSInstall all` you can now also install `'maintained'` parsers only.

This PR is on top of #351


kyazdani42 (approved):
lgtm, should we merge this?

PR #536, opened 2020-09-28
chore: actually use 'cache' for cache_dir

theHamsta:
Argument for 'data' was that is always exists.
But we could just try mkdir if it does not exist or try out some other directory.


clason:
superseded by rewrite on `main` branch

PR #540, opened 2020-09-29
fix(highlights): use new highlighter interface

vigoux:
Should be merged after https://github.com/neovim/neovim/pull/12953.

I'll try my best to port breaking changes before they actually land on nvim master, and merge them at the same time to limit breakage.


theHamsta:
Can you check for this patch (e.g. check for `Parser:input_source()` instead of the old check) and warn users that don't have. Don't we instantiate parsers in more places?

vigoux:
I'll add that check in the version check, and maybe automate the process a bit. 

Parser changes are transparent for us.
The problem only covers the highlighter itself.

theHamsta (approved):

PR #541, opened 2020-09-29
Add CodeQL to the parser list

pwntester:


pwntester:
will send along with the queries

PR #542, opened 2020-09-29
initial support for CodeQL

pwntester:


queries/ql/highlights.scm
```diff
@@ -0,0 +1,103 @@
...
+  "then"
+  "where"
+
+  (false)
+  (predicate)
+  (result)
+  (specialId)
+  (super)
+  (this)
+  (true)
```
theHamsta, line 45: Thanks for the contribution. A few notes on the different group of keywords:
true, false -> `@boolean`
this, super -> `@variable.builtin`
sum,rank,instanceof,avg... -> `@function.builtin`
import -> `@include`
if, then... -> `@conditional`
and, or,select... -> `@keyword.operator`

queries/ql/highlights.scm
```diff
@@ -0,0 +1,103 @@
...
+  ","
+  "|"
+] @punctuation.delimiter
+
+(className) @type
+
+(varName) @variable
+
+(integer) @number
+(float) @number
```
theHamsta, line 92: `@float`

theHamsta (approved):

theHamsta:
Strange the Github CI didn't add the parsers to the README...

theHamsta:
Ah, it does not update README because you're on your master branch. We will get the README update on the next PR automatically that comes from another branch.

vigoux:
Hmm that really needs a fix...

theHamsta:
I mentioned "master" as a exception in the workflow. lsp-config workflow has the same problem.

pwntester:
Thanks guys! will use a new branch in the next PR (locals.scm) so the README is generated

theHamsta:
I already pushed that autogenerated commit

PR #549, opened 2020-10-01
Add ocamllex highlights

undu:
The parser emits opaque `ocaml` nodes, since these are not yet handled to the ocaml parser they remain unhighlighted. The ocaml nodes appear inside action nodes. I've tagged those with `@embedded` since that's how it seems to be handled in other scm files. is this how it's supposed to be used?
 
I've added the hash to the lockfile, although it's not clear in the diff since it's all in the same line.

Like it happens to highlight mli files, this PR is needed: https://github.com/ocaml/vim-ocaml/pull/61 (or set manually the filetype to ocamllex)


theHamsta (approved):
Thanks again!

PR #559, opened 2020-10-04
fix: readme link to nvim-treesitter-textobjects

tonycheang:
Hey! Really interesting work going on here, been following and updating regularly. Saw that `textobjects` was moved to a new module and found a small typo while updating my dotfiles.


theHamsta (approved):
Thanks for finding that!

PR #563, opened 2020-10-05
Remove dependency on unzip

SuhasHebbar:
Currently `unzip` is needed to extract the parsers after download. `unzip` may not be available by default on some OSes/Distros. `tar` will be a better choice here due to it being available by default on many UNIXes and Windows.

This PR also changes over the download format from zip to tar.gz. `bsdtar` which is available by default on Windows (since build 17063) , supports extracting zip files while GNU tar which is default on most Linux distros does not (as far as I can see).

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -37,6 +37,30 @@ local function get_revision(lang)
   return (lockfile[lang] and lockfile[lang].revision)
 end
 
+local function select_mkdir_cmd(directory, cwd, info_msg)
+  if fn.has('win32') == 1 then
+    return {
+      cmd = 'cmd',
+      opts = {
+        args = { '/C', 'mkdir', directory},
+	cwd = cwd,
+      },
+      info = info_msg,
+      err = "Could not create "..directory,
+    }
+  else
+    return {
+      cmd = 'mkdir',
+      opts = {
+        args = { directory },
+	cwd = cwd,
+      },
+      info = info_msg,
+      err = "Could not create "..directory,
+    }
+  end
+end
+
 local function select_rm_file_cmd(file, info_msg)
   if fn.has('win32') == 1 then
     return {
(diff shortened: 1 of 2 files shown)


lua/nvim-treesitter/install.lua
```diff
@@ -37,6 +37,30 @@ local function get_revision(lang)
...
+      opts = {
+        args = { directory },
+	cwd = cwd,
+      },
+      info = info_msg,
+      err = "Could not create "..directory,
+    }
+  end
+end
+
```
theHamsta, line 63: you could also create this with `vim.fn.mkdir` beforehand. But this should be also ok. Could you move all these `select-*` functions to a separate file. `select_compiler_args` would also be a better name for `select_args`.

theHamsta:
Will be really awesome when it works for everyone out-of-the box!

lua/nvim-treesitter/install.lua
SuhasHebbar: Do you have a preference for the filename? Would command_selectors.lua be acceptable?

theHamsta: `install_shell_commands.lua`?, `shell_command_selection.lua`?, `shell_command_selectors.lua`? I would add the `shell_` to bring some context which kind of commands it refers to.

SuhasHebbar: I've gone with `shell_command_selectors.lua`.

lua/nvim-treesitter/install.lua
```diff
@@ -6,6 +6,7 @@ local utils = require'nvim-treesitter.utils'
 local parsers = require'nvim-treesitter.parsers'
 local info = require'nvim-treesitter.info'
 local configs = require'nvim-treesitter.configs'
+local shell_command_selectors = require'nvim-treesitter.shell_command_selectors'
```
theHamsta, line 9: you could use the nickname `shell` for this modules in this file...

theHamsta (approved):
Thanks a lot for this contribution!

lua/nvim-treesitter/install.lua
SuhasHebbar: > you could also create this with `vim.fn.mkdir` beforehand.
I tried using fn.mkdir for creating a new directory instead of defining a separate function for it.

But it didn't work since the newly created directory got deleted again by `M.select_install_rm_cmd(cache_folder, project_name..'-tmp')`.

theHamsta: when you delete the `-tmp` directory before `mkdir` the first `select_install_rm_cmd` is not needed. But it should be ok as it is right now.

vigoux (approved):
I would like to have @kyazdani42 's opinion on this too, as he worked on this a lot too.

kyazdani42:
looks good to me :)

PR #567, opened 2020-10-09
Add background and foreground colors to TSNone

akinsho:
This fixes the issue where colors highlighted as TSNone were not
reverting to the background and foreground color. It follow us up #551.

### DIFF
--- a/plugin/nvim-treesitter.vim
+++ b/plugin/nvim-treesitter.vim
@@ -11,7 +11,7 @@ let g:loaded_nvim_treesitter = 1
 
 lua require'nvim-treesitter'.setup()
 
-highlight default TSNone term=NONE cterm=NONE gui=NONE guifg=foreground guibg=background ctermfg=fg ctermbg=bg
+highlight default TSNone term=NONE cterm=NONE gui=NONE guifg=foreground ctermfg=fg
 
 highlight default link TSError TSNone
 
--- a/plugin/nvim-treesitter.vim
+++ b/plugin/nvim-treesitter.vim
@@ -11,17 +11,18 @@ let g:loaded_nvim_treesitter = 1
 
 lua require'nvim-treesitter'.setup()
 
-function s:has_attr(attr)
-  return strlen(synIDattr(hlID('Normal'), a:attr)) > 0
+function s:has_attr(attr, mode)
+  let norm_color = synIDattr(hlID('Normal'), a:attr, a:mode)
+  return strlen(norm_color) > 0
 endfunction
 
 " if the ctermfg or guifg is not known by nvim then using the
 " fg or foreground highlighting value will cause an E419 error
-if s:has_attr('fg')
-  highlight default TSNone term=NONE cterm=NONE gui=NONE guifg=foreground ctermfg=fg
-else
-  highlight default TSNone term=NONE cterm=NONE gui=NONE
-endif
+" so we check to see if either highlight has been set if not default to NONE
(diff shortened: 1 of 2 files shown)


theHamsta:
image

akinsho:
Ah didn't notice the `cterm` bits erroring. I'll push a fix, yeah it does seem to remove the cursorline highlight 😞 I'm not. Does not setting the `gui{fg/bg}` work for you in a dart file? maybe the current solution is right and there's something in my setup. Also no idea why this removes the cursorline anyway

akinsho:
Thinking about it I'm not sure that only setting `gui/cterm` could work since these are for applying attributes like `bold/italic` etc, but wouldn't affect the highlighting. Anyway on my end running `hi TSNone` without this PR returns `TSNone cleared`

akinsho:
Sorry this is turning into a stream of conciousness. I managed to fix the cursorline issue by not highlighting the background. I just realised there's no reason to do that in this case.

theHamsta:
This is vim. It's just how Bram wanted to have his favorite editor. No need to understand it. I had your solution before but saw no difference to removing both `fg` highlights.

theHamsta (approved):

plugin/nvim-treesitter.vim
```diff
@@ -11,7 +11,7 @@ let g:loaded_nvim_treesitter = 1
 
 lua require'nvim-treesitter'.setup()
 
-highlight default TSNone term=NONE cterm=NONE gui=NONE
+highlight default TSNone term=NONE cterm=NONE gui=NONE guifg=foreground ctermfg=fg
```
stsewd, line 14: I'm getting an error with this change 

```
Error detected while processing .../nvim-treesitter/plugin/nvim-treesitter.vim:                                                                                                                                                 
line   14:                                                                                                                                                                                                                                      
E419: FG color unknown
```

akinsho: Hmm 🤔 so found an explanation for this in the vim docs
```
							*E419* *E420*
	When Vim knows the normal foreground and background colors, "fg" and
	"bg" can be used as color names.  This only works after setting the
	colors for the Normal group and for the Windows console.
```

I'm guessing from this that in your setup @stsewd you don't have the normal foreground set or maybe it's set dynamically or something. what does `hi Normal` output for you?

akinsho: I've tried adding a check to see if `Normal` foreground is set before using this now, can you see if that works for you?

theHamsta:
Actually, setting `fg` should not be necessary. Does moving `@none` to the end of the file make this change necessary for you?

plugin/nvim-treesitter.vim
stsewd: I have

```
:hi Normal                                                                                                              
Normal         xxx guifg=#D9D7CE guibg=#212733
```

The error is still present with your change

kyazdani42 (changes requested):
having the E419 error even with the synId check.

akinsho:
I think the issue with the `synID` check might be an ordering issue. If this plugin is setup before the colour scheme then foreground will be empty. I'm gonna try changing this later today but in the interim if either of you get a chance @kyazdani42 or @stsewd can you see if you setup treesitter before your colorscheme is set? And then moving the colorscheme command to before you call setup on this plugin to test if the error goes away

I'll try changing the code to only apply the highlight on colorscheme autocmd

kyazdani42 (approved):
seems to work just fine on my end now, it doesn't override the cursorline background so that's perfect

stsewd (approved):
works for me

PR #581, opened 2020-10-13
C++ highlights: highlight all kinds of optional parameters

theHamsta:
declarator can be a identifier, a reference or a pointer

This is a intended mirco-improvement to highlight parameters like `foo(TypeT* t = nullptr)` except that it doesn't.

All reference, pointer and member highlights (`m_foo`) do not work on my machine for a longer time. I don't understand why.

### DIFF
--- a/queries/cpp/highlights.scm
+++ b/queries/cpp/highlights.scm
@@ -14,10 +14,10 @@
 ; function(Foo ...foo)
 (variadic_parameter_declaration
   declarator: (variadic_declarator
-                (identifier) @parameter))
+                (_) @parameter))
 ; int foo = 0
 (optional_parameter_declaration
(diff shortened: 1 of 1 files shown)


queries/cpp/highlights.scm
```diff
@@ -17,7 +17,7 @@
                 (identifier) @parameter))
 ; int foo = 0
 (optional_parameter_declaration
-    declarator: (identifier) @parameter)
+    declarator: (_) @parameter)
```
theHamsta, line 20: either `identifier`, `reference_declarator`, `pointer_declarator`

vigoux (approved):

theHamsta:
Was a config error on my side (those bug report ruin your config ;-))

PR #585, opened 2020-10-15
js: add query for @parameter

fffed:
Before:
image

After:
image

Only second case is left

### DIFF
--- a/queries/javascript/highlights.scm
+++ b/queries/javascript/highlights.scm
@@ -107,7 +107,7 @@
   (array_pattern
     (identifier) @parameter))
 
-; ( a ) => null
+; a => null
 (variable_declarator                                                                                                                                                                                        
     value: (arrow_function                                                                                                                                                                                    
       parameter: (identifier) @parameter))
--- a/queries/javascript/highlights.scm
+++ b/queries/javascript/highlights.scm
@@ -91,6 +91,23 @@
   (rest_parameter
     (identifier) @parameter))
 
+; ({ a }) => null
+(formal_parameters
+  (object_pattern
+    (shorthand_property_identifier) @parameter))
+
+; ({ a: b }) => null
+(formal_parameters
+  (object_pattern
+    (pair
+      value: (identifier) @parameter)))
+
+; ([ a ]) => null
+(formal_parameters
+  (array_pattern
+    (identifier) @parameter))
+
+; ( a ) => null
 (variable_declarator                                                                                                                                                                                        
     value: (arrow_function                                                                                                                                                                                    
       parameter: (identifier) @parameter))
--- a/queries/javascript/highlights.scm
+++ b/queries/javascript/highlights.scm
@@ -95,10 +95,6 @@
     value: (arrow_function                                                                                                                                                                                    
       parameter: (identifier) @parameter))
 
-(formal_parameters                                                                                                                                                                                          
-    (object_pattern                                                                                                                                                                                           
-      (shorthand_property_identifier) @parameter))
-
 ; Variables
 ;----------
 (namespace_import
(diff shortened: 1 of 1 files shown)


queries/javascript/highlights.scm
```diff
@@ -91,6 +91,14 @@
   (rest_parameter
     (identifier) @parameter))
 
+(variable_declarator                                                                                                                                                                                        
+    value: (arrow_function                                                                                                                                                                                    
+      parameter: (identifier) @parameter))
+
+(formal_parameters                                                                                                                                                                                          
+    (object_pattern                                                                                                                                                                                           
+      (shorthand_property_identifier) @parameter))
```
kyazdani42, line 100: @steelsojka i don't think this should be a property, object destructuring refers to object properties.

queries/javascript/highlights.scm
```diff
@@ -91,6 +91,10 @@
   (rest_parameter
     (identifier) @parameter))
 
+(variable_declarator                                                                                                                                                                                        
+    value: (arrow_function                                                                                                                                                                                    
+      parameter: (identifier) @parameter))
```
steelsojka, line 96: I don't think this solves the issue presented in the description. The syntax tree for the expression is

```
            value: arrow_function [37, 10] - [37, 33]
              parameters: formal_parameters [37, 10] - [37, 27]
                identifier [37, 11] - [37, 16]
                object_pattern [37, 18] - [37, 26]
                  shorthand_property_identifier [37, 20] - [37, 24]
```

A) You would need to change it to `parameters`
B) You would need to change `identifier` to `shorthand_property_identifier` to target the destructered property.

steelsojka (changes requested):

queries/javascript/highlights.scm
steelsojka: I don't know. If it's destructured in the parameter list it's pretty much a parameter....  It's kind of like named parameters in a way. I'm kind of on the fence about it.

queries/javascript/highlights.scm
kyazdani42: ive already asked to change the shorthand property as its not a parameter by itself, the object is.

kyazdani42: he removed it after my previous comment. But i didn't expect it to be removed, i just noticed that a shorthand_property_identifier is not a parameter but an object property.
I don't know if assigning to parameter would be a good idea. What do you think about that ?

steelsojka: It's an object property, but since it's only in the context of function parameters it might be ok. Would this be better for a custom query extension on the users end?

fffed: I added it(object property destructuring) only because I used to it looks like this in Sublime editor, but @kyazdani42 had a good point that it is should be what it is in a first place)))

steelsojka:
I think we should highlight destructuring within an argument list as parameters, but if you we do it we need to account for array destructuring and aliasing as well. These should cover all the scenarios:

```scheme
; ({ a }) => null
(arrow_function
  (formal_parameters
    (object_pattern
      (shorthand_property_identifier) @parameter)))

; ({ a: b }) => null
(arrow_function
  (formal_parameters
    (object_pattern
      (pair
        value: (identifier) @parameter))))

; ([ a ]) => null
(arrow_function
  (formal_parameters
    (array_pattern
      (identifier) @parameter)))
```

Add these in and then this should be ready to go.

steelsojka (changes requested):
Include all parameter variations and this should be good to go!

fffed:
> Include all parameter variations and this should be good to go!

Done. I also leave comments for cases should we keep them or what?

steelsojka (approved):

Issue #521, opened 2020-09-23
[RFC] Range metatable?


theHamsta:
**Is your feature request related to a problem? Please describe.**
Not everything we want to do with tree-sitter can be represented by a single node. Sometimes we want to have a subrange of a node and sometimes we want to group together multiple nodes (sometimes of same kind, sometimes with a start and end node).

Situation where we want to express subranges of a node or group nodes:
- textobjects
- group of comments for doc generation
- ranges for language injection

**Describe the solution you'd like**
It would be cool to have a Lua metatable that duck-types to API of nodes (it can give you start/end byte, a parent node, it could have a type "virtual node", etc.). This would enable kind of virtual nodes that can be used in functions that usually expects nodes. Such ranges could also be produced by directives in queries that trim/join/intersect node ranges.

**Describe alternatives you've considered**
We could just work with nodes and work with ranges. Sometimes we want functions that can work on both. Then, we have to distinguish the cases or coerce the nodes to ranges.

**Additional context**
We could incubate this concept here. If it proves useful we could upstream it to the Neovim main repository. I don't know if it might intersect with planed solutions for language injection.

This is just a random thought. There might be better solutions and it might be a bad idea to have something that's similar to a node but actually isn't one.

theHamsta:
Fixed via #590

PR #590, opened 2020-10-18
Add virtual range for creating node-like objects from multiple nodes

theHamsta:
#521 

Capturing range that are not equal to the range of exactly one node via queries is not easy right now. But is some times necessary e.g. for textobjects, setting ranges for language injection, getting the doc-strings for @steelsojka 's docgen plugin.

[This PR](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/14) enables defining multi-node textobjects defined defined by start/end node pairs. But this makes the code unnecessary complicated because we have to check a lot of cases.

This PR enables defining range object that implement the same API as regular nodes by queries. Then, multiple-node textobjects just work without any code change. A similar directive could be introduced that trims away `// ` from comments or similar tasks that require range not equal not tree-sitter nodes or finding `TODO:` in a comment string.

```scheme
((
  [(method_signature) (function_signature)] @start
  .
  (function_body) @end
)
(make-range! "function.outer" @start @end)) 
```

image

You can see in this example that the range of a function in flutter is actually defined by two nodes `void main` and the body. With this query, the plugin can use `function.object` as if it were a real tree-sitter node.

### DIFF
--- a/lua/nvim-treesitter/tsrange.lua
+++ b/lua/nvim-treesitter/tsrange.lua
@@ -1,0 +1,145 @@
+local M = {}
+local TSRange = {}
+TSRange.__index = TSRange
+
+local api = vim.api
+local parsers = require'nvim-treesitter.parsers'
+
+local function get_byte_offset(buf, row, col)
+  return api.nvim_buf_get_offset(buf, row)
+         + vim.fn.byteidx(api.nvim_buf_get_lines(buf, row, row + 1, false), col)
+end
+
+function TSRange.new(buf, start_row, start_col, end_row, end_col)
+  return setmetatable(
+    {
+      start_pos = {start_row, start_col, get_byte_offset(buf, start_row, start_col)},
+      end_pos = {end_row, end_col, get_byte_offset(buf, end_row, end_col)},
+      buf = buf,
+      [1] = start_row,
+      [2] = start_col,
+      [3] = end_row,
+      [4] = end_col,
+    },
+    TSRange)
+end
+
+function TSRange.from_nodes(buf, start_node, end_node)
+  TSRange.__index = TSRange
+  local start_pos = {start_node:start()}
+  local end_pos = {end_node:end_()}
+  return setmetatable(
+    {
+      start_pos = {start_pos[1], start_pos[2], start_pos[3]},
+      end_pos = {end_pos[1], end_pos[2], end_pos[3]},
+      buf = buf,
+      [1] = start_pos[1],
+      [2] = start_pos[2],
+      [3] = end_pos[1],
+      [4] = end_pos[2],
+    },
+    TSRange)
+end
+
+function TSRange.from_table(buf, range)
+  return setmetatable(
+    {
+      start_pos = {range[1], range[2], get_byte_offset(buf, range[1], range[2])},
+      end_pos = {range[3], range[4], get_byte_offset(buf, range[3], range[4])},
+      buf = buf,
+      [1] = range[1],
+      [2] = range[2],
+      [3] = range[3],
+      [4] = range[4],
+    },
+    TSRange)
+end
+
+function TSRange:parent(range)
+  local parser = parsers.get_parser(self.buf, parsers.get_buf_lang(range))
+  local root = parser.tree:root()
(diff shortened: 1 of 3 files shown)


lua/nvim-treesitter/tsrange.lua
```diff
@@ -0,0 +1,145 @@
...
+      [1] = range[1],
+      [2] = range[2],
+      [3] = range[3],
+      [4] = range[4],
+    },
+    TSRange)
+end
+
+function TSRange:parent(range)
+  local parser = parsers.get_parser(self.buf, parsers.get_buf_lang(range))
```
theHamsta, line 59: needs to be adapted when we have the LanguageTree implementation

lua/nvim-treesitter/tsrange.lua
```diff
@@ -0,0 +1,145 @@
...
+function TSRange:has_error()
+  return #self:collect_children(function(c) return c:has_error() end) > 0 and true or false
+end
+
+function TSRange:sexpr()
+  return table.concat(vim.tbl_map(function(c) return c:sexpr() end, self:collect_children()), ' ')
+end
+
+M.TSRange = TSRange
+return M
```
kyazdani42, line 145: LGTM, but are methods like range:named,missing and such not implemented yet ?

theHamsta: What do you mean? `TSRange` as such is neither a named nodes nor a missing node. So I just returned `false`

kyazdani42: I think this is not worth adding then.

theHamsta: but when these methods are called and missing, it will cause an error in code that assumes mostly real nodes (usually we have only code that is aware of queries using virtual nodes)

kyazdani42: better to have an error than undefined behavior imo

theHamsta: alright, then I remove it. Though stating that this is not a filler node or a named node is for me the opposite of defined behavior.

kyazdani42: well this is not a node, its a range of nodes right ?

steelsojka:
From a usability stand point, is this meant to be used as a "real" node? Meaning, if I have a function that takes a node, it could receive one of these range nodes transparently?

theHamsta:
> From a usability stand point, is this meant to be used as a "real" node? Meaning, if I have a function that takes a node, it could receive one of these range nodes transparently?

Yes, that's the idea. We've already some duplicated implementation for ranges and nodes. The idea is that if someone writes functions for nodes you should also be able to use this functionality if the "thing" isn't exactly one node in one language.

theHamsta:
I could try to implement Python injection for docstrings with this

```python
"""

>>> python_code.here()
"""
```

lua/nvim-treesitter/tsrange.lua
steelsojka: I think these methods shouldn't throw errors... If I wanted to use this in place of a node for I now need to ALWAYS wrap these calls in a `pcall` because there is no guarantee that it's not a virtual node. IMO it should be a noop or return false. Can this behavior be configurable? This makes it kind of hard to use in a plugin... The convenience is kind of lost.

theHamsta: I would also prefer to never throw an error. those methods are for selecting special kinds of nodes. We can return with confidence that we are not one of the special nodes the user is asking for.

kyazdani42: well i agree, i didn't understood the purpose of this at first, but it indeed integrates smoothly with nodes.

kyazdani42 (approved):

kyazdani42:
looks like it's mergeable to me, @steelsojka  @theHamsta is this fully ready ? I think we should merge asap so we can go forward with language injection. Great work @theHamsta :)

theHamsta:
Removed the errors. Should be fine now.

steelsojka (approved):

PR #592, opened 2020-10-18
Fix #591: Add verilog parser

### LINKED ISSUE #591: Is it possible to add support for Verilog ?
There's a github page for [Verilog](https://github.com/tree-sitter/tree-sitter-verilog) and another one for [SystemVerilog](https://github.com/rfdonnelly/tree-sitter-sv). It would be nice if these could be supported.

theHamsta:
@medwatt have you tested which of the parsers is better. I could not make the other parser working with nvim-treesitter though I re-generated the grammar from source. The official parsers had some parsing errors on the next best file I tested.

The other one had even more errors. Are they parsing different dialects of Verilog?

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,7 @@ List of currently supported languages:
 - [x] [toml](https://github.com/ikatyang/tree-sitter-toml) (maintained by @tk-shirasaka)
 - [ ] [tsx](https://github.com/tree-sitter/tree-sitter-typescript)
 - [x] [typescript](https://github.com/tree-sitter/tree-sitter-typescript) (maintained by @steelsojka)
+- [ ] [verilog](https://github.com/tree-sitter/tree-sitter-verilog)
 - [ ] [vue](https://github.com/ikatyang/tree-sitter-vue)
 - [ ] [yaml](https://github.com/ikatyang/tree-sitter-yaml)
 <!--parserinfo-->
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -285,6 +285,13 @@ list.ql = {
   maintainers = {'@pwntester'},
 }
 
+list.verilog = {
+  install_info = {
+    url = "https://github.com/tree-sitter/tree-sitter-verilog",
+    files = { "src/parser.c" },
+  },
+}
+
 -- Parsers for injections
 list.regex = {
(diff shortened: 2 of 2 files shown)


medwatt:
@theHamsta I don't know which one is better as I haven't tested them. Verilog and SystemVerilog are like C and C++ in the sense that C++ can be considered to be a superset of C. Basically, all Verilog syntax/constructs are supported by SystemVerilog.

theHamsta:
The official parser was updated only 5 days ago. So we should stay with it.

theHamsta:
Well, the last real commit for tree-sitter-verilog (apart from runtime update and adding the generated files was  Mar 16)

medwatt:
> Well, the last real commit for tree-sitter-verilog (apart from runtime update and adding the generated files was Mar 16)
> 
> @zegervdv added the source files. Do you have any experience using the parser together with nvim-treesitter?

I only started using treesitter in neovim a couple of days ago and liked it so much that I went searching for a parser for Verilog or VHDL. So, unfortunately, I don't have much experience.

zegervdv:
I have been using the tree-sitter-verilog for a couple of weeks now. It has a few bugs, some constructs that are not being parsed correctly, but overall it seems to work fine.

I've been trying to write some queries for highlighting (see [branch](https://github.com/zegervdv/nvim-treesitter/tree/verilog-support)) but I am not really familiar with the concepts :)

theHamsta:
@zegervdv would you be interested in creating a PR from your changes? As you see, there are more people interested in having a Verilog parser. We can help you with all the details. User usually push fixes when they detect small paper cuts. So it's much more easy than to maintain the queries alone.

zegervdv:
Sure, was planning to but not feeling ready yet. I'll rebase and create a PR soon.

theHamsta:
superseded by #943

medwatt:
@zegervdv I added the code below and the verilog parser installed but I don't notice any difference in the syntax highlighting compared to the regex-based syntax when I enable and disable treesitter. Is there something else I need to do ?

```lua
list.verilog = {
  install_info = {
    url = "https://github.com/tree-sitter/tree-sitter-verilog",
    files = { "src/parser.c" },
  }
}
```

theHamsta:
It needs highlights.scm. Basically a mapping which syntax node should get which highlight.

PR #603, opened 2020-10-19
feat(ocamllex): use injections for ocaml

undu:
This allows the big blocks of ocaml code to be highlighted correctly.

Example:
before

after

### DIFF
--- a/queries/ocamllex/highlights.scm
+++ b/queries/ocamllex/highlights.scm
@@ -3,10 +3,12 @@
 
 (lexer_entry_name) @function
 
-["as" "let" "parse" "rule"] @keyword
+["as" "let" "and" "parse" "rule"] @keyword
 
 [(eof) (character)] @character
 (string) @string
+
+(ocaml) @none
 
 (character_range "-" @operator)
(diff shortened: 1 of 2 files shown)


theHamsta (approved):

queries/ocamllex/highlights.scm
```diff
@@ -16,7 +18,7 @@
 (regexp_repetition ["*"] @operator)
 (regexp_strict_repetition ["+"] @operator)
 
-(action ["{" "}"] @punctuation.special) @embedded
+(action ["{" "}"] @punctuation.special)
```
theHamsta, line 21: What is this action? Might `@none` apply here to reset the highlighting or is this always never highlighted.

vigoux (approved):
Yay injections for more languages now !!

queries/ocamllex/highlights.scm
undu: action contains only the brackets and the ocaml node, so resetting the highlighting for the ocaml node is enough, I think.

PR #605, opened 2020-10-20
Add lua params

p00f:


theHamsta:
Shouldn't it be
```
(parameters
    (indentifier))
```
(function (identifier)) is the function name and only for exported/global functions and not for local ones.

p00f:
The query file uses the same syntax for function names which confused me too https://github.com/nvim-treesitter/nvim-treesitter/blob/3b8efe7a1ebfd5cc84bc0b43e8e5142eb4ef9de8/queries/lua/highlights.scm#L110.

But then I tried it and it worked - my query captures the parameter names correctly:
image
(Colorscheme highlights parameters in orange italic)

theHamsta:
Please change it to
```
(parameters
    (indentifier))
```
This means, highlight a child of function that's either an identifier (the function name as you showed above) or `parameters`
```
(function [(parameters) (identifier)] @parameter)
```

p00f:
Done

p00f:
But in the case of line 110, `identifier` is a child of `function_name` 
image

Shouldn't line 110 be `(function [(function_name (identifier)) @function]`?

theHamsta:
You can probably remove the `identifier` from this query. `function_name` is enough
```
function [34, 0] - [44, 3]
  function_name [34, 9] - [34, 22]
    function_name_field [34, 9] - [34, 22]
      object: identifier [34, 9] - [34, 10]
      property_identifier [34, 11] - [34, 22]
```

p00f:
Thanks, understood. Should I touch that if I didn't write it?

p00f:
Similarly in line 113:
https://github.com/nvim-treesitter/nvim-treesitter/blob/3b8efe7a1ebfd5cc84bc0b43e8e5142eb4ef9de8/queries/lua/highlights.scm#L113
 `local_function` doesn't seem to have function name:
image

queries/lua/highlights.scm
```diff
@@ -121,6 +121,9 @@
 (function_call (identifier) @function . (arguments))
 (function_call (field_expression (property_identifier) @function) . (arguments))
 
+;; Parameters
+(function [(parameters (identifier))] @parameter)
```
vigoux, line 125: Why having only one alternative in here?

theHamsta: why having `function` here? It's redundant.

p00f: > Why having only one alternative in here? 

Wdym ?

p00f: > why having `function` here? It's redundant. Also it does not apply to local_function

Done :+1:

vigoux (approved):

PR #607, opened 2020-10-20
handle jsx components with dot in them

davidroeca:
I noticed some issues with components that had `.` in them - SUI React has a lot of components defined in this way, such as [Table](https://react.semantic-ui.com/collections/table/) etc.
 
Before:
image

After:
image

### DIFF
--- a/queries/jsx/highlights.scm
+++ b/queries/jsx/highlights.scm
@@ -18,19 +18,19 @@
  (#match? @type "^[A-Z]")))
 
 ; Handle the dot operator effectively - <My.Component>
-(jsx_opening_element ((nested_identifier (identifier) @type (identifier) @type)))
+(jsx_opening_element ((nested_identifier (identifier) @tag (identifier) @type)))
 
 (jsx_closing_element ((identifier) @type
  (#match? @type "^[A-Z]")))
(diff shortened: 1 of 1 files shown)


kyazdani42:
@steelsojka not sure this is expected behavior. names with uppercase first letter are always highlighted as TSConstructor for js/jsx/ts/tsx.

davidroeca:
@kyazdani42 is `<little.Uppercase>` even acceptable react?

kyazdani42:
sometimes when your import styles as default, you could do `<styles.SomeStyle>`

davidroeca:
There could be multiple approaches here. One is to always identify the first `identifier` in `nested_identifier` as `@tag`. Another would be to handle `PascalCase`/`camelCase` separately via another `#match?` expression

steelsojka:
@kyazdani42 I'm fine with this change. In theory you can have multiple levels of nested properties.

```javascript
<ns.accounts.AccountProfile />
```

But it's an edge so I'm not to concerned about it. The proposed approach works best imo since it's case agnostic just for tags (which case shouldn't matter).

steelsojka (approved):

PR #611, opened 2020-10-23
Remove textobjects from contributing

theHamsta:
Another thing that went wrong when merging remove-refactor and remove-textobjects.


theHamsta:
Oops that was too much

vigoux (approved):

PR #612, opened 2020-10-23
fix(javascript): fix constant variable highlight

JoosepAlviste:
Hey!

I found that in JavaScript, the `@constant` query gets overwritten by the `@constructor` query. They both query for variables starting with an uppercase letter, but the `@constant` query took effect first.

Here's how it looks like:

**Before:**

image

**After:**

image


steelsojka (approved):
Looks good. Thanks for the contribution!

PR #614, opened 2020-10-24
show [nvim-treesitter] during install

kyazdani42:
it can be disturbing on startup when treesitter is updating, you don't really know which plugin is printing the [Compiling], so i added [nvim-treesitter] to inform users treesitter parser is being updated/installed.


theHamsta (approved):
Any reason why you are using two brackets?

theHamsta:
But it should be nvim-treesitter

PR #615, opened 2020-10-24
Fixes Issue #597.

saadparwaiz1:
Since, the [nvim-treesitter-textobject](https://github.com/nvim-treesitter/nvim-treesitter-textobjects) has been moved to another repository. The checkhealth script needs to be updated so it doesn't check for indent.scm.


theHamsta:
Indents is unrelated to nvim-treesitter-textobjects. Indentexpr is still handled by this repo.

Issue #543, opened 2020-09-30
What would it take to use my lua grammar :)


tjdevries:
Hi @vigoux  :)

Just wondering what it would take to switch to this grammar: https://github.com/tjdevries/tree-sitter-nlua

I'd like to iterate on it and use it w/ tree-sitter, but it's pretty hard right now because of the default query appending and not using the same names...

If change the names of my repo would make things easy / you would accept it as the new lua grammar, I could do that. Just wondering what your thoughts are :smile:

theHamsta:
use https://github.com/nvim-treesitter/nvim-treesitter/pull/506 to install from local paths or else
```lua
  require "nvim-treesitter.parsers".get_parser_configs().lisp = {
    install_info = {
      url = "~/projects/tree-sitter-lisp",
      files = {"src/parser.c"}
    }
  }
  --require "nvim-treesitter.parsers".get_parser_configs().viml = {
  --install_info = {
  --url = "https://github.com/vigoux/tree-sitter-viml",
  --files = {"src/parser.c"}
  --}
  --}
  require "nvim-treesitter.parsers".get_parser_configs().markdown = nil
  require "nvim-treesitter.parsers".get_parser_configs().zig = {
    install_info = {
      url = "https://github.com/GrayJack/tree-sitter-zig",
      files = {"src/parser.c"}
    }
  }
  require "nvim-treesitter.parsers".get_parser_configs().kotlin = {
    install_info = {
      url = "https://github.com/QthCN/tree-sitter-kotlin",
      files = {"src/parser.c"}
    }
  }

```

tjdevries:
One question I had related to this. Any way to get the query files from the parser repo? I'd like to keep the files there if possible and iterate on them.

vigoux:
I think the easiest way would be to symlink them manually... We don't keep the parsers files around (to save some roome on users side).

theHamsta:
@vigoux he wants us to adopt his queries.

theHamsta:
If you make the language of your parser `nlua` both parsers you even co-exist (or better their queries). A setting could determine whether to load the one or the other.

vigoux:
The trees generated by the `lua` parser and the `nlua` parser differs by a fair bit, I think having them more close one to the other would be a good selling point (as switching from one to the other would basically be noop).

tjdevries:
[...]

PR #619, opened 2020-10-27
Feat/graphql support

kyazdani42:
- finish highlight
- ask the maintainer if it expects upgrading the parser to 2020 spec or fork the parser
- remove template substitution from injection (should not highlight ${...} stuff)
- maybe make a plugin to add injection support in this case as it's mostly framework related (and its still a little niche)

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,7 @@ List of currently supported languages:
 - [ ] [elm](https://github.com/razzeee/tree-sitter-elm)
 - [x] [fennel](https://github.com/travonted/tree-sitter-fennel) (maintained by @TravonteD)
 - [x] [go](https://github.com/tree-sitter/tree-sitter-go) (maintained by @theHamsta, @WinWisely268)
+- [x] [graphql](https://github.com/dralletje/tree-sitter-graphql) (maintained by @kyazdani42)
 - [ ] [haskell](https://github.com/tree-sitter/tree-sitter-haskell)
 - [x] [html](https://github.com/tree-sitter/tree-sitter-html) (maintained by @TravonteD)
 - [x] [java](https://github.com/tree-sitter/tree-sitter-java) (maintained by @p00f)
--- a/ftdetect/graphql.vim
+++ b/ftdetect/graphql.vim
@@ -1,0 +1,1 @@
+autocmd BufEnter,BufNewFile *.gql,*.graphql setlocal ft=graphql
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -320,6 +320,14 @@ list.query = {
   maintainers = {"@steelsojka"},
 }
 
+list.graphql = {
+  install_info = {
+    url = "https://github.com/dralletje/tree-sitter-graphql",
+    files = { "src/parser.c"}
+  },
+  maintainers = {"@kyazdani42"},
+}
+
 local M = {
   list = list
 }
--- a/queries/graphql/highlights.scm
+++ b/queries/graphql/highlights.scm
@@ -1,0 +1,6 @@
+(Type) @type
+(Variable) @variable
+
+(OperationDefinition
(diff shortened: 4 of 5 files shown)


theHamsta:
Don't forget to also inject the regex parser into JS ;-)

kyazdani42:
i think i might not add the graphql at the end, because it's framework related (i've discussed that with @steelsojka).
I think we should make a plugin repo to group all the js framework/libs that alter the views. So maybe @steelsojka your treesitter angular sould be a treesitter-addons :)

steelsojka:
> Don't forget to also inject the regex parser into JS ;-)

I actually tried this and it currently doesn't work.... I think there is an issue with the ranges when it does highlighting. The first line of injected region doesn't get highlighted. I'm gonna investigate it further later today.

steelsojka:
@kyazdani42 I'm good with that. I'll just keep working in the current repo and we move it where ever we need to.

steelsojka:
@kyazdani42 So, I'm creating a treesitter parser for the angular expression syntax (it's a small subset of javascript.... kind of), I'm going to use this in the angular add on. Should I add the parser to the main nvim-treesitter package (like you did with graphql) or should it be added through the plugin `nvim-treesitter-angular`?

kyazdani42:
i think you can add the parser in nvim treesitter. After all, jsx has its own right ?

kyazdani42:
although if it will be only used through iniection based on the extensions brought by the plugin, you could add the parser to the plugin repo.

kyazdani42:
hi @bkegley, i wonder if your parser is working well enough ?

bkegley:
@kyazdani42 just opened #686 to add the grammar. I'll be adding remaining highlights tonight.

Should `highlights.scm` live in `queries/graphl` in `nvim-treesitter` or in my grammar repo?

theHamsta:
@bkegley was a nice stream when you created the parser.

kyazdani42:
so cool, not sure where to put the queries though, but here would be fine imo(or maybe both)

theHamsta:
In #543 @tjdevries has the same problem. Couldn't we just have a key in the parser info that tells us to grab the queries of the parser repo during installation.

bkegley:
It would be nice if they could be pulled in during installation so that queries wouldn't have to be maintained in 2 places. How is the lockfile pulling in hashes for pinning the parser?

theHamsta:
We have a script that retrieves the current version via Git. At the moment you would probably set only the revision of the graphql server manually since some queries need to be updated.

steelsojka:
I think the queries should live in nvim-treesitter. Highlight queries may have predicates that are implemented in and specific to nvim-treesitter implementations. Our capture groups aren't guaranteed to match the captures of the highlights in the grammar repo as well. From what I understand, the highlight queries in the grammar repos are specific to the RUST implementation of the highlighter, which doesn't match nvims implementation.

kyazdani42:
closing this because its already done :) thanks @bkegley

tjdevries:
@steelsojka what about for grammars written for neovim? If like to maintain all the Lua stuff in working on on one repo, particularly while things are in Flux on my end.

steelsojka:
@tjdevries That's perfectly fine. I was mostly referring to the highlight queries that are included in the treesitter grammar repos that are tested using the rust highlighter for testing. For example, the highlights in tree-sitter-javascript would not be compatible with the ones maintained in nvim-treesitter. If it's specifically for neovim, then it can live in a plugin repo or wherever.

tjdevries:
Sorry, I mean, what if I want to write all my queries for neovim in my repo. That way they are always perfectly in sync in case I change any names. The repo is built for neovim. I do not care if no other editors use it at the moment. I will also add things like adding more keyword support like `vim.api.*` items and more.

theHamsta:
@tjdevries just include the parser installation into your PR. Your parser could have the filetype for now nlua to coexist with the other parser.

PR #624, opened 2020-10-30
fix(languagetree): highlight children regardless of starting column

steelsojka:
This fixes some issues with language injection I've noticed.

Currently, we only get the lines that changed (not including the column), but injections that don't start at 0 never match the range given for the line change. This runs all highlighters that exist for the given line. This fixes code like this:

Screen Shot 2020-10-30 at 8 53 14 AM


vigoux (approved):
LGTM

steelsojka:
That was fast

vigoux:
Yeah no time to waste, it is so cool to have you contribute to this

PR #625, opened 2020-10-31
Add folds for Go

ClaudiaJ:
The default behavior in which `@scope` captures were folded resulted in
toplevel fold for an entire source file, which was mildly irritating.

related: #597


stsewd (approved):
Thanks!

PR #628, opened 2020-11-03
Add initial clojure support

sogaiu:
There is a preliminary highlights.scm file and a blank locals.scm file (to enable incremental selection to work).

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -20,6 +20,14 @@ list.c = {
   maintainers = {"@vigoux"},
 }
 
+list.clojure = {
+  install_info = {
+    url = "https://github.com/sogaiu/tree-sitter-clojure",
+    files = { "src/parser.c" }
+  },
+  maintainers = {"@sogaiu"},
+}
+
 list.cpp = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-cpp",
--- a/queries/clojure/locals.scm
+++ b/queries/clojure/locals.scm
@@ -1,0 +1,1 @@
+; placeholder file to get incremental selection to work
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -1,0 +1,73 @@
+(dis_expr) @comment
+
+(kwd_lit) @type
+
+(str_lit) @string
+
+(num_lit) @number
+
+(bool_lit) @boolean
+
+(nil_lit) @constant.builtin
+
+(comment) @comment
+
+(regex_lit) @string.regex
+
+;; metadata experiment
+(meta_lit
+ marker: "^" @punctuation.special)
+
+;; dynamic variables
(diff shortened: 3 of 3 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -20,6 +20,14 @@ list.c = {
   maintainers = {"@vigoux"},
 }
 
+list.clojure = {
+  install_info = {
+    url = "https://github.com/tree-sitter/tree-sitter-clojure",
```
vigoux, line 25:
```suggestion
    url = "https://github.com/sogaiu/tree-sitter-clojure",
```

sogaiu: Thanks!

vigoux:
@theHamsta no `README` update ?

theHamsta:
:man_shrugging: in nvim-treesitter-textobject it always pushes to master even when it was told to always push except on master.

queries/clojure/locals.scm
```diff
@@ -0,0 +1 @@
+; placeholder file to get incremental selection to work
```
theHamsta, line 1: I guess it would be better if you could just replace this line
https://github.com/sogaiu/nvim-treesitter/blob/908e87216e48385e420367b6e3715d4abb8ee049/lua/nvim-treesitter/configs.lua#L37

by `parsers.has_parser`

theHamsta: It's not that simple... But I would still prefer to have this feature for all languages if an empty file is enough.

sogaiu: I don't understand how the incremental selection works nor it scope.

Is it easy to explain why locals.scm is necessary for incremental selection?  Or perhaps there is something I might read?

theHamsta: The behavior is to use scopes (https://github.com/sogaiu/nvim-treesitter/blob/908e87216e48385e420367b6e3715d4abb8ee049/lua/nvim-treesitter/locals.lua#L261 which reads query files https://github.com/sogaiu/nvim-treesitter/blob/908e87216e48385e420367b6e3715d4abb8ee049/lua/nvim-treesitter/locals.lua#L52) or alternatively the parent.

I guess it could be fixed here: https://github.com/sogaiu/nvim-treesitter/blob/908e87216e48385e420367b6e3715d4abb8ee049/lua/nvim-treesitter/incremental_selection.lua#L86 to only do the call to containing_scope if `query.has_locals(parsers.get_buf_lang()` and else just use `node:parent() or node`.

theHamsta: The incremental selection would normally operate on variable scopes e.g. `{ }` of a c++ file

sogaiu: Thanks for the explanation.

theHamsta: But this isn't in the scope of this PR. Just keep the empty file.

theHamsta (approved):

PR #531, opened 2020-09-26
Python locals.scm: variables -> left_hand_side

theHamsta:
This shouldn't be merged yet! The commits that require this change come after the update runtime commit of the Python repository which currently makes the Python parser unusable.

PR #629, opened 2020-11-03
Update lockfile and parser queries

theHamsta:
This is #531


theHamsta:
We need to wait for AppImage to build with new runtime.

vigoux (approved):

vigoux (approved):

vigoux:
Awesome, are we good then ?

theHamsta:
If it's green, it's green. Was red before :smile:

theHamsta:
Now it should be fine.

theHamsta:
Users need now newest nightly. Or they gonna have a bad time.

PR #635, opened 2020-11-04
JS optional parameters

theHamsta:
I have no idea how to write a function in JS but googling told me that you can have optional parameters that are not highlighted right now https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/19#pullrequestreview-523806600

```javascript
(function connect(hostname = "localhost",
                 port = 80,
                 method = "HTTP") {

})
```
The dots of `rest_pattern` `...` are not highlighted right now. How could the be highlighted. Punctuation? Parameter?


kyazdani42:
The ... thing is basically like vararg but in js. I think highlighting as parameter would make sense

theHamsta:
The vararg identifier is highlighted as parameter. I am referring to he dots.
```
(...foo
    ^^^ 
    this is highlighted
```

kyazdani42:
well these are just dots i dont think it matters too much

kyazdani42 (approved):

PR #639, opened 2020-11-09
Julia highlights

theHamsta:
Maybe I'm gonna do this Advent of Code in Julia. I already added some highlights and plan to iteratively add more.

Julia beautiful tree sitter: https://www.sarahbeekmans.com/julia-hill/

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -1,0 +1,177 @@
+(identifier) @variable
+;; In case you want type highlighting based on Julia naming conventions (this might collide with mathematical notation)
+;((identifier) @type ; exception: mark `A_foo` sort of identifiers as variables
+  ;(match? @type "^[A-Z][^_]"))
+((identifier) @constant
+  (match? @constant "^[A-Z][A-Z_]{2}[A-Z_]*$"))
+
+[
+  (triple_string)
+  (string)
+] @string
+
+(string
+  prefix: (identifier) @constant.builtin)
+
+(macro_identifier) @function.macro
+(macro_identifier (identifier) @function.macro) ; for any one using the variable highlight
+(macro_definition
+  name: (identifier) @function.macro
+  ["macro" "end" @keyword])
+
+(field_expression
+  (identifier)
+  (identifier) @field .)
+
+(function_definition
+  name: (identifier) @function)
+(call_expression
+  (identifier) @function)
+(call_expression
+  (field_expression (identifier) @method .))
+(broadcast_call_expression
+  (identifier) @function)
+(broadcast_call_expression
+  (field_expression (identifier) @method .))
+(parameter_list
+  (identifier) @parameter)
+(parameter_list
+  (optional_parameter .
+    (identifier) @parameter))
+(typed_parameter
+  (identifier) @parameter
+  (identifier) @type)
+(type_parameter_list
+  (identifier) @type)
+(typed_parameter
+  (identifier) @parameter
+  (parameterized_identifier) @type)
+(function_expression
+  . (identifier) @parameter)
+(spread_parameter) @parameter
+(spread_parameter
+  (identifier) @parameter)
+(named_argument
+    . (identifier) @parameter)
+(argument_list
+  (typed_expression
+    (identifier) @parameter
+    (identifier) @type))
+(argument_list
+  (typed_expression
+    (identifier) @parameter
+    (parameterized_identifier) @type))
+
+;; Symbol expressions (:my-wanna-be-lisp-keyword)
+;; Same highlight as in clojure
+(quote_expression
+ (identifier)) @type
+
+;; Parsing error! foo (::Type) get's parsed as two quote expressions
+(argument_list 
+  (quote_expression
+    (quote_expression
+      (identifier) @type)))
+
+(type_argument_list
+  (identifier) @type)
+(parameterized_identifier (_)) @type
+(argument_list
+  (typed_expression . (identifier) @parameter))
+
+(typed_expression
+  (identifier) @type .)
+(typed_expression
+  (parameterized_identifier) @type .)
+
+(struct_definition
+  name: (identifier) @type)
+
+(number) @number
+(range_expression
+    (identifier) @number
+      (eq? @number "end"))
+(range_expression
+  (_
+    (identifier) @number
+      (eq? @number "end")))
+(coefficient_expression
+  (number)
+  (identifier) @constant.builtin)
+
+;; TODO: operators.
+;; Those are a bit difficult to implement since the respective nodes are hidden right now (_power_operator)
+;; and heavily use Unicode chars (support for those are bad in vim/lua regexes)
+;[;
+    ;(power_operator);
+    ;(times_operator);
+    ;(plus_operator);
+    ;(arrow_operator);
+    ;(comparison_operator);
+    ;(assign_operator);
+;] @operator ;
+
+"end" @keyword
+
+(if_statement
+  ["if" "end"] @conditional)
+(elseif_clause
+  ["elseif"] @conditional)
+(else_clause
+  ["else"] @conditional)
+(ternary_expression
+  ["?" ":"] @operator)
+
+(function_definition ["function" "end"] @keyword.function)
+
+(comment) @comment
+
+[
+  "const"
+  "return"
+  "macro"
+  "struct"
+] @keyword
+
+((identifier) @keyword (#match? @keyword "^(global|local)$"))
+
+(compound_expression
+  ["begin" "end"] @keyword)
+(try_statement
+  ["try" "end" ] @exception)
+(finally_clause
+  "finally" @exception)
+(catch_clause
+  "catch" @exception)
+(quote_statement
+  ["quote" "end"] @keyword)
+(let_statement
+  ["let" "end"] @keyword)
+(for_statement
+  ["for" "end"] @repeat)
+(while_statement
+  ["while" "end"] @repeat)
+(break_statement) @repeat
+(continue_statement) @repeat
+(for_binding
+  "in" @repeat)
+(for_clause
+  "for" @repeat)
+(do_clause
+  ["do" "end"] @keyword)
+
+(export_statement
+  ["export"] @include)
+
+[
+  "using"
+  "module"
+  "import"
+] @include
+
+((identifier) @include (#eq? @include "baremodule"))
+
+(((identifier) @constant.builtin) (match? @constant.builtin "^(nothing|Inf|NaN)$"))
+(((identifier) @boolean) (eq? @boolean "true"))
+(((identifier) @boolean) (eq? @boolean "false"))
+
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [javascript](https://github.com/tree-sitter/tree-sitter-javascript) (maintained by @steelsojka)
 - [x] [jsdoc](https://github.com/tree-sitter/tree-sitter-jsdoc) (maintained by @steelsojka)
 - [x] [json](https://github.com/tree-sitter/tree-sitter-json) (maintained by @steelsojka)
-- [ ] [julia](https://github.com/tree-sitter/tree-sitter-julia)
+- [x] [julia](https://github.com/tree-sitter/tree-sitter-julia) (maintained by @mroavi, @theHamsta)
 - [x] [kotlin](https://github.com/QthCN/tree-sitter-kotlin) (maintained by @tormodatt)
 - [x] [lua](https://github.com/nvim-treesitter/tree-sitter-lua) (maintained by @vigoux)
 - [x] [nix](https://github.com/cstrahan/tree-sitter-nix) (maintained by @leo60228)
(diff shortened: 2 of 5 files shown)


queries/julia/highlights.scm
```diff
@@ -0,0 +1,131 @@
...
+  (number)
+  (identifier) @constant.builtin)
+
+[
+    ;(_power_operator)
+    ;(_times_operator)
+    ;(_plus_operator)
+    ;(_arrow_operator)
+    ;(_comparison_operator)
+    ;(_assign_operator)
```
theHamsta, line 58: this does not work, since they are not strictly supertypes (they would be if grammar.js would drop the token in front of choice).

theHamsta:
https://github.com/tree-sitter/tree-sitter-julia/pull/19 would make it easy to match against different kinds of operators.

theHamsta:
This is a collaboration with @mroavi. We tuned the highlights a bit and agreed that they are ready for a first review.

Operators are missing. We left them for a follow-up PR since a solution for those can be quite convoluted.

Possible solutions
- expose now hidden operator nodes (might require forking the parser if respective PR doesn't get accepted)
- try to list the operator with a `#is-julia-operator?` predicate (regexes are quite complicated and don't properly support all the involved Unicode chars)

theHamsta:
@clason since you gave us a like here... I'm wondering whether you would be available for a review?

clason:
Sure, I'll give it a whirl!

clason:
It looks to me that the highlighting of variables is a bit inconsistent. Is that a style guide I'm not aware of?

Screenshot 2020-12-30 at 18 15 45

clason:
I also noticed

Screenshot 2020-12-30 at 18 23 51

(I believe the `quux` should be highlighted as function, not the struct `bar`?)

mroavi:
> It looks to me that the highlighting of variables is a bit inconsistent. Is that a style guide I'm not aware of?

You are right. In fact, the problem arises because of not following the style guide. We can remove this query to fix this:

https://github.com/theHamsta/nvim-treesitter/blob/4f27c7787c9865f26106496811fa94d8e9636952/queries/julia/highlights.scm#L2-L3

mroavi:
> I also noticed
> 
> Screenshot 2020-12-30 at 18 23 51
> 
> (I believe the `quux` should be highlighted as function, not the struct `bar`?)

Right. In fact, it should be highlighted as a method. Here is the buggy query:

https://github.com/theHamsta/nvim-treesitter/blob/4f27c7787c9865f26106496811fa94d8e9636952/queries/julia/highlights.scm#L23-L24

theHamsta:
The method issue should be fixed now by only highlighting the last identifier in `foo.bar.baz()` call.

Is there any way how we can distinguish types from local variables without semantic analysis? Can you assign to types in Julia? I got the impression that in most cases we are right with the convention that PascalCase identifiers are types. There is really few positions where grammar alone makes it unique whether we have a type or a local variable. When we use the PascalCase convention we usually only have the ambiguity Modul <-> Type.

Can we vote whether we want the naming convention solution like in other languages or whether we should leave PascalCase identifiers without highlight?

I added a capture for struct definitions. In this situation we can be sure  that it's a definition.

Also added variadic parameter support-

clason:
> Can we vote whether we want the naming convention solution like in other languages or whether we should leave PascalCase identifiers without highlight? (right now I out-commented the highlight in case you want to test)

I for one would vote strongly against such naming conventions. Julia is often (but of course not only) used for scientific computing, where variable naming follows mathematical notation. I'd rather have variables not highlighted than highlighted inconsistently unless I deviated from the "natural" notation.

clason:
> The method issue should be fixed now by only highlighting the last identifier in `foo.bar.baz()` call.

Hmm, now I don't get any part of `foo.bar()` highlighted?

mroavi:
> > The method issue should be fixed now by only highlighting the last identifier in `foo.bar.baz()` call.
> 
> Hmm, now I don't get any part of `foo.bar()` highlighted?

In my case it does get highlighted. It is being captured by this query:
https://github.com/theHamsta/nvim-treesitter/blob/4f27c7787c9865f26106496811fa94d8e9636952/queries/julia/highlights.scm#L54-L56

Try assigning a color to `TSField:`.

clason:
Good call, that works.

(Is there a Wiki or help section that lists all the `TS*` fields a color scheme should handle?)

theHamsta:
No, the field query is also wrong (fixed it now). It should be handled by
```
(call_expression
  (field_expression (identifier) @method .))
```

theHamsta:
@clason Yes, the vim doc should list all highlight groups and nvim-treesitter.vim defines them.

theHamsta:
Really a pitty that we don't have a (non-segfaulting) markdown parser right now. Or else we could inject Markdown into the docstrings.

theHamsta:
I added a constant highlight for all UPPERCASE identifiers with more than 3 letters. Saw a few constants in Julia code. I hope this will not collide with any mathematical notation.

clason:
> I added a constant highlight for all UPPERCASE identifiers with 3 or more letters. Saw a few constants in Julia code. I hope this will not collide with any mathematical notation.

Well.... (I could imagine the SVD of a matrix, for example?) Such assumptions are unlikely to be safe.

clason:
What _might_ be interesting to highlight differently is `foo!` for mutating functions? (That's a pretty firm convention in Julia.)

theHamsta:
Well, you should name that local `svd` when following the naming conventions of Julia. Each function or type `SVD` will still be highlighted as function.

clason:
Yeah, maybe I'm overthinking it. (I was also thinking of precomputed products like `ACB` of three matrices `A`, `B`,  and `C`...) Three characters is probably a good compromise.

(The [https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/](style guide) doesn't mention constants, though.)

mroavi:
Just realized that `global` and `local` should be added as keywords.

theHamsta:
I'm working the Pattern Recognition Lab of my uni. 95% correct is quite good. I checked 5 random projects from awesome-julia and they all used that convention for constants. We can refine that later if this will cause problems.

mroavi:
I was quickly browsing through the Julia source code and comparing our syntax highlighting with the one used in Atom.

In Julia, it is allowed to define a parameter with no name, just its type. We are not highlighting that properly. Here is an example from `abstractarray.jl`:

image

theHamsta:
image
Looks good on my machine. I updated the PR a lot. You're sure you're on the most recent version?

mroavi:
I fetched the latest version and still have the problem. Then it is probably getting captured by a node that is highlighted white in my pc or that has no highlight assigned. Could you check which query is capturing that highlight?

theHamsta:
It's a parsing error:
```
    argument_list [10, 9] - [10, 42]
      quote_expression [10, 10] - [10, 23]
        quote_expression [10, 11] - [10, 23]
          identifier [10, 12] - [10, 23]
```
I can try to highlight that error as type. Some times it's highlighted as `@variable`.

mroavi:
One other difference btwn our highlights and those of Atom. We are not highlighting [symbols](https://docs.julialang.org/en/v1/manual/metaprogramming/#Symbols):

Atom:
image

Nvim:
image

theHamsta:
image

Should not get type highlight

theHamsta:
They get parsed as quote expresion of identifier. Really strange that clojure does not have keyword highlighting (they use `@type`?)

mroavi:
image

In the first instance of `::LinearIndex`, the grammar reports no error, and the highlight is correct.
In the second case, the grammar does report an error and it is not highlighted.

theHamsta:
Please don't send screenshots! Better the parsed result and the source code so that I can copy an paste.

The variadic case typed case was not added i guess.

I'll use type highlighting for Symbols for now to be in consistent with clojure. I'm strongly for opening a new category for this.

mroavi:
Code:
```
eachindex(::IndexLinear, A::AbstractVector) = (@_inline_meta; axes1(A))

function eachindex(::IndexLinear, A::AbstractArray, B::AbstractArray...)
    indsA = eachindex(IndexLinear(), A)
end
```
Parser results:
```
assignment_expression [267, 0] - [267, 71]
  call_expression [267, 0] - [267, 43]
    identifier [267, 0] - [267, 9]
    argument_list [267, 9] - [267, 43]
      quote_expression [267, 10] - [267, 23]
        quote_expression [267, 11] - [267, 23]
          identifier [267, 12] - [267, 23]
      typed_expression [267, 25] - [267, 42]
        identifier [267, 25] - [267, 26]
        identifier [267, 28] - [267, 42]
  parenthesized_expression [267, 46] - [267, 71]
    macro_expression [267, 47] - [267, 60]
      macro_identifier [267, 47] - [267, 60]
        identifier [267, 48] - [267, 60]
    call_expression [267, 62] - [267, 70]
      identifier [267, 62] - [267, 67]
      argument_list [267, 67] - [267, 70]
        identifier [267, 68] - [267, 69]
function_definition [269, 0] - [271, 3]
  name: identifier [269, 9] - [269, 18]
  parametere: parameter_list [269, 18] - [270, 39]
    ERROR [269, 19] - [269, 21]
    identifier [269, 21] - [269, 32]
    typed_parameter [269, 34] - [269, 50]
      identifier [269, 34] - [269, 35]
      identifier [269, 37] - [269, 50]
    optional_parameter [269, 52] - [270, 39]
      identifier [269, 52] - [269, 53]
      ERROR [269, 53] - [270, 9]
        field_expression [269, 55] - [270, 9]
          identifier [269, 55] - [269, 68]
          ERROR [269, 69] - [269, 72]
          identifier [270, 4] - [270, 9]
      call_expression [270, 12] - [270, 39]
        identifier [270, 12] - [270, 21]
        argument_list [270, 21] - [270, 39]
          call_expression [270, 22] - [270, 35]
            identifier [270, 22] - [270, 33]
            argument_list [270, 33] - [270, 35]
          identifier [270, 37] - [270, 38]
```
In the first instance of `::IndexLinear`, the grammar does give an error, and the highlight is correct.
In the second instance, the grammar does report an error and it is not highlighted.

theHamsta:
You can see that it's clearly parsed wrong. I added a query for this misinterpretation.
```
argument_list [267, 9] - [267, 43]
      quote_expression [267, 10] - [267, 23]
        quote_expression [267, 11] - [267, 23]
          identifier [267, 12] - [267, 23]
```

theHamsta:
Also interpolations would be cool for highlighting. In case anyone wants to fork the grammar.

image

mroavi:
Your query for the misinterpretation didn't catch this case:
Code:
```julia
@noinline function throw_eachindex_mismatch(::IndexLinear, A...)
    throw(DimensionMismatch("all inputs to eachindex must have the same indices, got $(join(eachindex.(A), ", ", " and "))"))
end
```
Parser results:
```
macro_expression [213, 0] - [215, 3]
  macro_identifier [213, 0] - [213, 9]
    identifier [213, 1] - [213, 9]
  macro_argument_list [213, 10] - [215, 3]
    function_definition [213, 10] - [215, 3]
      name: identifier [213, 19] - [213, 43]
      parametere: parameter_list [213, 43] - [213, 64]
        ERROR [213, 44] - [213, 46]
        identifier [213, 46] - [213, 57]
        spread_parameter [213, 59] - [213, 63]
          identifier [213, 59] - [213, 60]
      call_expression [214, 4] - [214, 125]
        identifier [214, 4] - [214, 9]
        argument_list [214, 9] - [214, 125]
          call_expression [214, 10] - [214, 124]
            identifier [214, 10] - [214, 27]
            argument_list [214, 27] - [214, 124]
              string [214, 28] - [214, 108]
              ERROR [214, 110] - [214, 118]
                string [214, 110] - [214, 114]
                identifier [214, 115] - [214, 118]
              string [214, 119] - [214, 123]
```

mroavi:
`global` and `local` should be marked as keywords.

mroavi:
`struct` should also be captured as a keyword.

mroavi:
I found a way to print all Julia reserved words ([source](https://stackoverflow.com/a/52341209/1706778))

```
❯ julia --lisp

> reserved-words
(begin while if for try return break continue function macro quote let local
 global const do struct module baremodule using import export end else elseif
 catch finally true false)
```
These all should be captured as keywords.

mroavi:
> I found a way to print all Julia reserved words ([source](https://stackoverflow.com/a/52341209/1706778))
> 
> ```
> ❯ julia --lisp
> 
> > reserved-words
> (begin while if for try return break continue function macro quote let local
>  global const do struct module baremodule using import export end else elseif
>  catch finally true false)
> ```
> 
> These should be captured as keywords if not already captured as other node 'types'.

theHamsta:
Added all supported keywords (unsupported by matching (identifier) for the keyword text). Please use review comments so we have a thread per issue that can be marked as resolved!

This PR discussion is getting really long!

theHamsta:
Markdown injections:

image

Looks like as if the markdown parser is coming back: https://github.com/ikatyang/tree-sitter-markdown/issues/14#issuecomment-753304375

vigoux (approved):

theHamsta:
It probably has still flaws, but fixing them one by one by whoever spots them should make it easier to make incremental improvements.

PR #656, opened 2020-11-15
CSS: Move property_name, feature_name to @field

rockerBOO:
Move property_name and feature_name to @field

I feel this is a more appropriate location and allows better distinction. 

My colors (sorry for the very slight color change, I can adjust if necessary)
2020-11-15-141903_4480x1440_scrot

### DIFF
--- a/queries/css/highlights.scm
+++ b/queries/css/highlights.scm
@@ -49,11 +49,13 @@
  (class_name)
  (id_name)
  (namespace_name)
- (property_name)
- (feature_name)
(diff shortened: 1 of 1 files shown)


rockerBOO:
Hmm i think this is counter intuitive so i'm going to make some adjustments. Some distinction there but maybe a better identifier of the `id/class` names?

The `id/class` names feel like variables but there are other things as variables (css variables). They are selectors but of a certain name.

PR #661, opened 2020-11-17
add gdscript

tjdevries:
First pass at gdscript

### DIFF
--- a/queries/gdscript/highlights.scm
+++ b/queries/gdscript/highlights.scm
@@ -1,0 +1,27 @@
+(identifier) @variable
+
+(name) @variable
+(type) @type
+(comment) @comment
+
+(enum_definition
+  (name) @constant) @Statement
+
+(function_definition
+  (name) @function
+  (parameters
+    (identifier)* @parameter) @punctuation.bracket) @Statement
+
+(return_statement) @Statement
+
+(variable_statement
+  (inferred_type)? @operator) @Statement
+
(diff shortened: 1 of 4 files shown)


theHamsta:
Our CI is unhappy with you using `@function.call` (not documented). I guess we replaced that by just `@function`. Would really welcome `@function.call`

queries/gdscript/highlights.scm
```diff
@@ -0,0 +1,27 @@
+(identifier) @variable
+
+(name) @variable
+(type) @type
+(comment) @comment
+
+(enum_definition
+  (name) @constant) @Statement
```
theHamsta, line 8: We don't have `@Statement`

vigoux: We actually do, it is a shortcut to set highlights directly from their groups, not recommended though...

theHamsta: I was not trying to say that it won't work. It's just not in our `TSHighlights`. If there is a strong reason to use it that is very language specific I'd say it ok.

rockerBOO: I think this would set `TSConstant` as well as the `Statement` groups.

theHamsta:
Just rebased and fixed the CI for this PR.

vigoux (approved):

vigoux:
This needs a rebase (again).

PR #667, opened 2020-11-18
Initial Nix queries

leo60228:
I've added pretty detailed comments, since I'm assuming most people working on nvim-treesitter don't use Nix.

I tried to add an indents.scm as well, though I got absolutely awful results. Since that currently only exists for Lua, I'm guessing that indent support is a work in progress?

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,6 @@ List of currently supported languages:
 
 <!--This section of the README is automatically updated by a CI job-->
 <!--parserinfo-->
-
 - [x] [bash](https://github.com/tree-sitter/tree-sitter-bash) (maintained by @TravonteD)
 - [x] [c](https://github.com/tree-sitter/tree-sitter-c) (maintained by @vigoux)
 - [x] [c_sharp](https://github.com/tree-sitter/tree-sitter-c-sharp) (maintained by @svermeulen)
@@ -245,7 +244,7 @@ List of currently supported languages:
 - [x] [json](https://github.com/tree-sitter/tree-sitter-json) (maintained by @steelsojka)
 - [ ] [julia](https://github.com/tree-sitter/tree-sitter-julia)
 - [x] [lua](https://github.com/nvim-treesitter/tree-sitter-lua) (maintained by @vigoux)
-- [ ] [nix](https://github.com/cstrahan/tree-sitter-nix)
+- [x] [nix](https://github.com/cstrahan/tree-sitter-nix) (maintained by @leo60228)
 - [x] [ocaml](https://github.com/tree-sitter/tree-sitter-ocaml) (maintained by @undu)
 - [x] [ocaml_interface](https://github.com/tree-sitter/tree-sitter-ocaml) (maintained by @undu)
 - [x] [ocamllex](https://github.com/atom-ocaml/tree-sitter-ocamllex) (maintained by @undu)
--- a/queries/nix/highlights.scm
+++ b/queries/nix/highlights.scm
@@ -66,6 +66,9 @@
 ; import
 ((identifier) @_i (#eq? @_i "import")) @include
 
+; null
+((identifier) @_i (#eq? @_i "import")) @constant.builtin
+
 ; these are technically functions but they act more like keywords (abort and throw are control flow, derivation is a core language construct)
 ((identifier) @_i (#match? @_i "^(abort|derivation|throw)$")) @keyword
 
--- a/queries/nix/highlights.scm
+++ b/queries/nix/highlights.scm
@@ -15,7 +15,7 @@
   ("else")
 ] @conditional
 
-; undocumented, I'm not actually sure what this does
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -245,7 +245,7 @@ List of currently supported languages:
 - [x] [json](https://github.com/tree-sitter/tree-sitter-json) (maintained by @steelsojka)
 - [ ] [julia](https://github.com/tree-sitter/tree-sitter-julia)
 - [x] [lua](https://github.com/nvim-treesitter/tree-sitter-lua) (maintained by @vigoux)
-- [ ] [nix](https://github.com/cstrahan/tree-sitter-nix)
+- [x] [nix](https://github.com/cstrahan/tree-sitter-nix)
 - [x] [ocaml](https://github.com/tree-sitter/tree-sitter-ocaml) (maintained by @undu)
```
theHamsta, line 249: Are you aware that this part of the readme is autogenerated. You have to add yourself as the maintainer in parsers.lua

leo60228: No, thank you.

leo60228:
What does the CI failure mean? The error message isn't very helpful.

kyazdani42:
indent support is very very much not ready for public use imo. It works better than normal indentation for lua but still has some quircks. The whole implementation needs to be rethinked ^^

theHamsta:
You seem to use a node that's not known to the nix parser. You can check nodes_types.json to see  which node types are available. You should see the same error message in Neovim when loading the queries.

leo60228:
Where would that file be? I can't find any node_types.json in https://github.com/cstrahan/tree-sitter-nix, my Neovim installation, or nvim-treesitter.

theHamsta:
https://github.com/cstrahan/tree-sitter-nix/blob/master/src/grammar.json apparently it had a different name on a older tree sitter cli

leo60228:
`grammar.json` seems to be different from `node_types.json`. I reran `tree-sitter generate` and it both updated `grammar.json` and created `node_types.json`. I'm going to see if this fixes the issue and then make a PR.

leo60228:
Wait, I misunderstood. The PR doesn't seem like it'll be necessary.

queries/nix/locals.scm
```diff
@@ -0,0 +1,15 @@
...
+
+; rec attrsets
+(rec_attrset (bind . (attrpath) @definition.field)) @scope
+
+; functions and parameters
+(function . [
+    (identifier) @definition.parameter
+    (formals (formal . (identifier) @definition.parameter))
+]) @scope
+((formals) "@" (identifier) @definition.parameter) ; I couldn't get this to work properly inside the (function)
```
theHamsta, line 12: We don't have (cheep) semantic highlighting yet

theHamsta (approved):
This looks all very good. Thanks! Can we merge?

queries/nix/highlights.scm
```diff
@@ -0,0 +1,114 @@
...
+
+; if/then/else
+[
+  ("if")
+  ("then")
+  ("else")
+] @conditional
+
+; undocumented, I'm not actually sure what this does
+("or") @keyword.operator
```
theHamsta, line 19: some wanted different highlights when operators were words instead of +,/,*,- . Default highlighting that "or" will have the same highlighting as `+`. `-`

leo60228: `@keyword.operator` is documented. The `or` operator is in Nix's lexer, and I recall having seen it used, but I can't find any documentation for what it actually does.

queries/nix/locals.scm
leo60228: What do you mean? I've been using `nvim-treesitter-refactor` for testing this. The issue I was encountering was that for some reason if I combined the two `(formals)` instances then the identifiers wouldn't get `@definition.parameter` if there was an `@ identifier`.

queries/nix/highlights.scm
leo60228: ...it's with all the other operators. I don't know how I missed it. It's used to specify a default for a field access. Normally, `a.b` will error if the field doesn't exist, but `a.b or c` will instead evaluate as `c`. I think `@keyword.operator` is fitting for that.

queries/nix/locals.scm
theHamsta: ah, sorry for my confusion.

vigoux (approved):
Looks all good to me, is this PR ready ?

leo60228:
This should be ready to merge. I've noticed that syntax errors often cause catastrophic parsing failures. I'd like someone with more tree-sitter experience to give me pointers on fixing that, though I'm pretty sure that's an issue with tree-sitter-nix and not this PR.

vigoux:
This is indeed an issue with the parser itself, but maybe we could help, please reach out on the Zulip chat, so that we can talk a bit over there.

Issue #668, opened 2020-11-18
Class members are highligted no more


GGCristo:
Since the last commits, member and attributes are no longer highlighted in c++.
Sorry for not using the template but I think in this case I can be straight forward, if you still want I can fill it.
If I do TSHIglightUnderCursor it return "No tree-sitter matches found!"

theHamsta:
Sorry, probably something went wrong when merging this old PR. I'll try to fix it tomorrow

GGCristo:
#672

Issue #446, opened 2020-09-12
[C++ highlight] some function calls are being highlighted as TSProperty but should be TSFunction


bluz71:
Hello,

With [this C++ file](https://github.com/bluz71/vim-nightfly-guicolors/blob/master/samples/cplusplus.cc) the function invocations in the `parseStory` function definition are being highlighted as `TSProperty` but should be highlighted as `TSFunction`.

Here is nvim-treesitter highlighting that segment (look at the crimson highlights such as `firstChild` and compare with the blue color `parseStory` itself which is correctly styled as `TSFunction`):

<img width="563" alt="Screen Shot 2020-09-12 at 11 41 03 am" src="https://user-images.githubusercontent.com/11382509/92984348-dcace900-f4ec-11ea-9569-2919089918fc.png">

Contrast with the Atom editor highlighting the same segment, notice the same color for `parseStory`, `firstChild`:

<img width="594" alt="Screen Shot 2020-09-12 at 11 42 09 am" src="https://user-images.githubusercontent.com/11382509/92984373-049c4c80-f4ed-11ea-8c13-ce204e0c4f07.png">

Best regards.

theHamsta:
Are you sure that the highlight is TSProperty and not TSMethod?

theHamsta:
You were right it's TSProperty. I am trying to fix the prioritization of those highlights...

bluz71:
You are right, `TSMethod` is a more appropriate highlight group than `TSFunction` for the items in crimson in the first screenshot; it has been a while since I did C++.

I look forward to testing the updates when they land.

Many thanks.

bluz71:
For my own record keeping, this issue is waiting upon PR #453

bluz71:
Tested, issue is confirmed as fixed:

![cplusplus](https://user-images.githubusercontent.com/11382509/99615794-262b2e80-2a70-11eb-9b2c-d5a61d1972f4.png)

Many thanks!

PR #672, opened 2020-11-19
c/cpp: fix field and initializer highlights

jdrouhard:
After a recent fix for #446, declarations in class/struct definitions stopped being marked as properties or methods. This fix will add property highlights to field declarations, and method highlight to field function declarations.

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -122,6 +122,11 @@
 (((field_expression
      (field_identifier) @property)) @_parent
  (#not-has-parent? @_parent template_method function_declarator call_expression))
+
+(((field_identifier) @property)
+ (#has-ancestor? @property field_declaration)
+ (#not-has-ancestor? @property function_declarator))
+
 (statement_identifier) @label
 
 [
--- a/queries/cpp/highlights.scm
+++ b/queries/cpp/highlights.scm
@@ -27,6 +27,12 @@
      (field_identifier) @method)) @_parent
  (#has-parent? @_parent template_method function_declarator call_expression))
 
+(field_initializer
+ (field_identifier) @property)
+
+(function_declarator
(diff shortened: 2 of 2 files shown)


theHamsta (approved):
LGTM. @GGCristo @bluz71 could you have a look on this?

theHamsta:
@vigoux is there any difference between `@property` and `@field`?

GGCristo:
It works partially as you can see:


It doesn't work for pointer nor operator when there is a space between "operator" and the symbol.

theHamsta:
Operator is a known parser error.

jdrouhard:
There's a PR (https://github.com/tree-sitter/tree-sitter-cpp/pull/72) in tree-sitter-cpp to fix the space problem. PRs there don't get merged very often.

jdrouhard:
I added support for reference and pointer (and other) declarators that wrap the identifier. Let me know if that fixes it fully (other than the space after `operator`) @GGCristo. Ping @theHamsta

GGCristo:
Yep, that solved it completely.

jdrouhard:
Good point. What do you suggest? We can't go back to just plain `(field_identifier) @property` because methods use that too, which was the whole reason behind the original change.

bluz71:
> LGTM. @GGCristo @bluz71 could you have a look on this? Does this fix #668 ?

I gave the nominated branch a quick look and it seems fine with my very simple C++ sample file. Members variables are being highlighted as properties.

However, I am not best to confirm if C++ is being highlighted correctly since I am not a C++ person.

Cheers.

jdrouhard:
@theHamsta I have something that generalizes the `field_identifier` a bit more. Now it'll match a property if it has a `field_declaration` ancestor but not a `function_declarator` ancestor.

vigoux (approved):

PR #681, opened 2020-11-21
README: Explain how to add unsupported parsers

clason:
This adds a brief explanation how to add unsupported parsers to nvim-treesitter (assuming the grammar is already written) via the snippet @theHamsta provided on Gitter.

It might be good to also add more details how to add the corresponding queries; I've left this vague since I don't know the details myself (is it enough to have a `queries` directory somewhere in `rtp`?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,22 +70,25 @@ $ git clone https://github.com/nvim-treesitter/nvim-treesitter.git
 ## Adding parsers
 
 Treesitter uses a different _parser_ for every language, which needs to be generated via `tree-sitter-cli` from a `grammar.js` file, then compiled to a `.so` library that needs to be placed in neovim's `runtimepath` (typically under `parser/{lang}.so`). To simplify this, `nvim-treesitter`
-provides commands to automate this process:
+provides commands to automate this process. If the language is already [supported by `nvim-treesitter`](#supported), you can install it with
+```vim
+:TSInstall {language}
+```
+This command supports tab expansion. You can also get a list of all available languages and their installation status with `:TSInstallInfo`.
 
-- `TSInstall {language}` to install one or more parsers from a generated `c` file. (This requires a `C` compiler in your path.)
-- `TSInstallFromGrammar {language}` to install one or more parsers from the original `grammar.js`. (In addition to a `C` compiler, this requires the `tree-sitter-cli` executable in your path; see https://tree-sitter.github.io/tree-sitter/creating-parsers#installation for installation instructions.)
+If you update `nvim-treesitter` and want to make sure the parser is at the latest compatible version (as specified in `nvim-treesitter`'s `lockfile.json`), use `:TSUpdate {language}`. To update all parsers unconditionally, use `:TSUpdate all` or just `:TSUpdate`.
 
-`TSInstall <tab>` and `TSInstallFromGrammar <tab>` will give you a list of supported languages; you can also use `TSInstall all` to install every parser on the list. To show which languages are available together with their installation status, use `TSInstallInfo`.
+### Adding unsupported parsers
 
-If your language is not yet included in this list, you can add it locally as follows:
+If you have a parser that is not on the list (either from a repository on Github or a local directory), you can add it manually for use by `nvim-treesitter` as follows:
 
-1. Clone the repository or [create a new project](https://tree-sitter.github.io/tree-sitter/creating-parsers#project-setup) in, say, `~/projects/tree-sitter-zimbu`.
-2. Run `tree-sitter generate` in this directory (followed by `tree-sitter test`, for good measure).
+1. Clone the repository or [create a new project](https://tree-sitter.github.io/tree-sitter/creating-parsers#project-setup) in, say, `~/projects/tree-sitter-zimbu`. Make sure that the `tree-sitter-cli` executable is installed and in your path; see https://tree-sitter.github.io/tree-sitter/creating-parsers#installation for installation instructions.
+2. Run `tree-sitter generate` in this directory (followed by `tree-sitter test` for good measure).
 3. Add the following snippet to your `init.vim`:
 
 ```vim
 lua <<EOF
-local parser_config =  require "nvim-treesitter.parsers".get_parser_configs()
+local parser_config = require "nvim-treesitter.parsers".get_parser_configs()
 parser_config.zimbu = {
   install_info = {
     url = "~/projects/tree-sitter-zimbu", -- local path or git repo
@@ -94,13 +97,14 @@ parser_config.zimbu = {
   filetype = "zu", -- if filetype does not agrees with parser name
   used_by = {"bar", "baz"} -- additional filetypes that use this parser
 }
+EOF
 ```
 
-4. Start `nvim` and run `TSInstall zimbu` (or `TSInstallFromGrammar zimbu` if you skipped step 2).
+4. Start `nvim` and `:TSInstall zimbu`.
+
+You can also skip step 2 and use `:TSInstallFromGrammar zimbu` to install straight from `grammar.js`. Once the parser is installed, you can update it (from the latest revision of the `main` branch if `url` is a Github repository) with `:TSUpdate zimbu`.
 
 Note that this only installs the parser itself; using it for, e.g., highlighting also requires corresponding queries that need to be written and placed in the appropriate directory (e.g., as `queries/zimbu/highlights.scm`).
-
-Once a parser is installed, you can update it via `TSUpdate {language}`. If the parser is supported, this will checkout and install the revision specified in `nvim-treesitter`'s `lockfile.json`; otherwise it will use the latest revision of the repo or directory given in the `url` field above. Like `TSInstall`, you can get a list of possible arguments with `TSUpdate <tab>` or update every installed parser with `TSUpdate all` (or just `TSUpdate` for short).
 
 ## Setup
 
@@ -217,7 +221,7 @@ Each feature can be enabled or disabled by different means:
 
 Check [`:h nvim-treesitter-commands`](doc/nvim-treesitter.txt) for a list of all available commands.
 
-# Supported Languages
+# <a name="supported"></a>Supported Languages
 
 For `nvim-treesitter` to work, we need to use query files such as those you can find in
 `queries/{lang}/{locals,highlights,textobjects}.scm`
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@
   </p>
 </div>
 
-The goal of `nvim-treesitter` is both to provide a simple and easy way to use the interface for [tree-sitter](https://github.com/tree-sitter/tree-sitter) in Neovim and to provide some basic functionality such as highlighting based on on it:
+The goal of `nvim-treesitter` is both to provide a simple and easy way to use the interface for [tree-sitter](https://github.com/tree-sitter/tree-sitter) in Neovim and to provide some basic functionality such as highlighting based on it:
 
 ![cpp example](assets/example-cpp.png)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -69,35 +69,38 @@ $ git clone https://github.com/nvim-treesitter/nvim-treesitter.git
 
 ## Adding parsers
 
-Treesitter uses a different _parser_ for every language. It can be quite a pain to install, but fortunately `nvim-treesitter`
-provides two command to tackle this issue:
+Treesitter uses a different _parser_ for every language, which needs to be generated via `tree-sitter-cli` from a `grammar.js` file, then compiled to a `.so` library that needs to be placed in neovim's `runtimepath` (typically under `parser/{lang}.so`). To simplify this, `nvim-treesitter`
```
theHamsta, line 72: Isn't it getting to complicated when you explain to newbies how to install or generate custom parsers? We could out-source this into a install your own parser section. But the instructions are quite clear so it could be also fine here.

README.md
```diff
@@ -69,35 +69,38 @@ $ git clone https://github.com/nvim-treesitter/nvim-treesitter.git
...
+provides commands to automate this process:
 
-- `TSInstall {language}` to install one or more parsers.
-  `TSInstall <tab>` will give you a list of supported languages, or select `all` to install them all.
-- `TSInstallInfo` to know which parser is installed.
+- `TSInstallInfo` to know which parsers are available and installed.
+- `TSInstall {language}` to install one or more parsers from a generated `c` file. (This requires a `C` compiler in your path.)
+- `TSInstallFromGrammar {language}` to install one or more parsers from the original `grammar.js`. (In addition to a `C` compiler, this requires the `tree-sitter-cli` executable in your path; see https://tree-sitter.github.io/tree-sitter/creating-parsers#installation for installation instructions.)
 - `TSUpdate` to update already installed parsers
```
theHamsta, line 78: Could you also add what update means in that context (install the revision of `lockfile.json` if parser contained in this file or else newest master).

clason: With pleasure, if you explain in a bit more detail what it actually does ;)

README.md
clason: That would be an alternative, yes; I thought of putting it in the wiki, but 

1. the instructions are quite short (effectively replacing the not really necessary example output that was there initially) and
2. especially at the moment, you want to make it as easy as possible for people to write and contribute parsers, so having that as visible as possible makes sense.

Later we can replace this with a link to a Wiki page (especially if the instructions get more extensive with queries etc.), but there should be an easy-to-find place that points to this information.

README.md
theHamsta: Ok, this is a bit more complicated... Probably that's the reason why it's not mentioned here.

- `:TSUpdate` is just `:TSInstall` without the `X is already installed. Reinstall?` query.
- The version that `:TSInstall` uses is described in `lockfile.json`. We pin the parsers because the parser revisions must be compatible with our queries.
- When you update, lockfile.json might change. TSUpdate reinstalls the new version in that case (thus #682). We recommend to run `:TSUpdate` when updating nvim-treesitter

See #682.

clason: How about this? I restructured the explanation a bit to avoid the forward reference. (It _is_ getting a bit long now, but I think explicit is good at this stage.)

theHamsta: Yes, your explanation sounds good!

README.md
```diff
@@ -69,35 +69,39 @@ $ git clone https://github.com/nvim-treesitter/nvim-treesitter.git
...
-provides two command to tackle this issue:
+Treesitter uses a different _parser_ for every language, which needs to be generated via `tree-sitter-cli` from a `grammar.js` file, then compiled to a `.so` library that needs to be placed in neovim's `runtimepath` (typically under `parser/{lang}.so`). To simplify this, `nvim-treesitter`
+provides commands to automate this process:
 
-- `TSInstall {language}` to install one or more parsers.
-  `TSInstall <tab>` will give you a list of supported languages, or select `all` to install them all.
-- `TSInstallInfo` to know which parser is installed.
-- `TSUpdate` to update already installed parsers
+- `TSInstallInfo` to know which parsers are available and installed.
+- `TSInstall {language}` to install one or more parsers from a generated `c` file. (This requires a `C` compiler in your path.)
```
theHamsta, line 75: > from a generated `c` file

I guess it's a bit unclear to what file this refers. It sounds a bit as if the user should provide that file. 

> to install one or more parsers from the project's Github repository

Do we have to repeat the compiler requirement that we mentioned in the first README section?

theHamsta:
image

It needs a whole screen page for the message: run `:TSInstall X`  for each language you need

README.md
clason: True (although that section doesn't state the requirement for `TSInstallFromGrammar` -- so maybe that section should be "minimal requirements" instead?)

Agreed on the unclear wording; I'll improve it.

clason:
Well, I have a 27" screen ;)

But yes, now it makes sense to split it into subsections, and make clear that most users can stop reading after `TSInstall {lang}`.

theHamsta:
I really mean just adding a subsection header for the advanced section. I don't think you have to change the explanation. Just make a lazy first-time viewer aware that they don't need that section for now.

clason:
Nah, anything worth doing is worth doing well. 

(Incidentally, is there a `TSUpdateFromGrammar` as well? Or should I omit `TSInstallFromGrammar` -- ideally you should `tree-sitter test` any changes, anyway ;)?)

clason:
So I ended up restructuring the whole README in **Quickstart** and **Advanced setup**, with lots of crosslinks. Let me know what you think!

theHamsta:
For me this makes sense! We should ask @vigoux and @pwntester. They've put a lot of effort in the documentation. Link to rendered REAMDE: https://github.com/nvim-treesitter/nvim-treesitter/blob/04cdb43129baefdb9bcde0d097d13691c14f1685/README.md

README.md
```diff
@@ -194,97 +225,71 @@ set foldexpr=nvim_treesitter#foldexpr()
...
-`queries/{lang}/{locals,highlights,textobjects}.scm`
+1. Clone the repository or [create a new project](https://tree-sitter.github.io/tree-sitter/creating-parsers#project-setup) in, say, `~/projects/tree-sitter-zimbu`. Make sure that the `tree-sitter-cli` executable is installed and in your path; see https://tree-sitter.github.io/tree-sitter/creating-parsers#installation for installation instructions.
+2. Run `tree-sitter generate` in this directory (followed by `tree-sitter test` for good measure).
+3. Add the following snippet to your `init.vim`:
 
-We are looking for maintainers to write query files for their languages.
+```vim
+lua <<EOF
+local parser_config = require "nvim-treesitter.parsers".get_parser_configs()
+parser_config.zimbu = {
```
theHamsta, line 264: Let's just hope that nobody creates a zimbu parser :smile:

README.md
```diff
@@ -116,12 +103,80 @@ require'nvim-treesitter.configs'.setup {
...
+
+For `nvim-treesitter` to support a specific feature for a specific language requires both a parser for that language and an appropriate language-specific query file for that feature.
+
+The following is a list of languages for which a parser can be installed through `:TSInstall`; a checked box means that `nvim-treesitter` also contains queries at least for the `highlight` module.
+
+We are looking for maintainers to add more parsers and to write query files for their languages.
+
+<!--This section of the README is automatically updated by a CI job-->
+<!--parserinfo-->
+
```
theHamsta, line 128: If you're wondering why the README CI fails... Don't worry. I fixed it in another PR.

README.md
```diff
@@ -29,13 +29,32 @@
...
 ![cpp example](assets/example-cpp.png)
 
 Traditional highlighting (left) vs Treesitter-based highlighting (right).
-See more examples in [our gallery](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Gallery).
+More examples can be found in [our gallery](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Gallery).
+
+**Warning: Treesitter and nvim-treesitter highlighting are an experimental feature of nightly versions of Neovim.
+Please consider the experience with this plug-in as experimental until Neovim 0.5 is released!
+You can find the current roadmap [here](https://github.com/nvim-treesitter/nvim-treesitter/projects/1).
```
stsewd, line 41: I think we can drop the roadmap project, doesn't look like is used/maintained. /cc @vigoux

README.md
```diff
@@ -29,13 +29,32 @@
...
+The roadmap and all features of this plugin are open to change, and any suggestion will be highly appreciated!**
+
+Nvim-treesitter is based on three interlocking features: [**language parsers**](#language-parsers), [**queries**](#adding-queries), and [**modules**](#available-modules), where *modules* provide features -- such as highlighting -- based on *queries* for syntax objects extracted from a given buffer by *language parsers*. Users will generally only need to interact with parsers and modules as explained in the next section. For more detailed information on setting these up, see ["Advanced setup"](#advanced-setup).
```
stsewd, line 44: What about using real em-dashes?

```suggestion
Nvim-treesitter is based on three interlocking features: [**language parsers**](#language-parsers), [**queries**](#adding-queries), and [**modules**](#available-modules), where *modules* provide features —such as highlighting— based on *queries* for syntax objects extracted from a given buffer by *language parsers*. Users will generally only need to interact with parsers and modules as explained in the next section. For more detailed information on setting these up, see ["Advanced setup"](#advanced-setup).
```

README.md
```diff
@@ -116,12 +103,80 @@ require'nvim-treesitter.configs'.setup {
...
+- [ ] [yaml](https://github.com/ikatyang/tree-sitter-yaml)
+<!--parserinfo-->
+
 
 # Available modules
 
-## Highlight
+Modules provide the top-level features of `nvim-treesitter`. These can be implemented either as part of `nvim-treesitter` or as an external plugin.
+
+## Included modules
```
stsewd, line 175: I think here we can just list the included modules, and have a link to https://github.com/nvim-treesitter/nvim-treesitter/wiki/Extra-modules-and-plugins for the external modules. That way we avoid having so many nested sections

README.md
```diff
@@ -194,97 +240,72 @@ set foldexpr=nvim_treesitter#foldexpr()
...
 
-Check [`:h nvim-treesitter-commands`](doc/nvim-treesitter.txt) for a list of all available commands.
+## Adding parsers
 
-# Supported Languages
+If you have a parser that is not on the list of supported languages (either as a repository on Github or in a local directory), you can add it manually for use by `nvim-treesitter` as follows:
 
-For `nvim-treesitter` to work, we need to use query files such as those you can find in
-`queries/{lang}/{locals,highlights,textobjects}.scm`
+1. Clone the repository or [create a new project](https://tree-sitter.github.io/tree-sitter/creating-parsers#project-setup) in, say, `~/projects/tree-sitter-zimbu`. Make sure that the `tree-sitter-cli` executable is installed and in your path; see https://tree-sitter.github.io/tree-sitter/creating-parsers#installation for installation instructions.
```
stsewd, line 260:
```suggestion
1. Clone the repository or [create a new project](https://tree-sitter.github.io/tree-sitter/creating-parsers#project-setup) in, say, `~/projects/tree-sitter-zimbu`. Make sure that the `tree-sitter-cli` executable is installed and in your path; see <https://tree-sitter.github.io/tree-sitter/creating-parsers#installation> for installation instructions.
```

README.md
```diff
@@ -304,16 +325,25 @@ require'nvim-treesitter'.define_modules {
...
-- `module_path`: A require path (string) that exports a module with an `attach` and `detach` function. This is not required if the functions are on this definition.
-- `enable`: Determines if the module is enabled by default. This is usually overridden by the user.
-- `disable`: A list of languages that this module is disabled for. This is usually overridden by the user.
-- `is_supported`: A function that takes a language and determines if this module supports that language.
-- `attach`: A function that attaches to a buffer. This is required if `module_path` is not provided.
-- `detach`: A function that detaches from a buffer. This is required if `module_path` is not provided.
+# Extra features
 
-# Utils
+### Statusline indicator
```
stsewd, line 339:
```suggestion
## Statusline indicator
```

README.md
```diff
@@ -304,16 +325,25 @@ require'nvim-treesitter'.define_modules {
...
+
+```vim
+echo nvim_treesitter#statusline(90)  " 90 can be any length
+module->expression_statement->call->identifier
+```
+
+### Utilities
```
stsewd, line 346:
```suggestion
## Utilities
```

README.md
```diff
@@ -323,61 +353,41 @@ local ts_utils = require 'nvim-treesitter.ts_utils'
...
-  directory.
-
 # Troubleshooting
 
-Before doing anything make sure you have the latest version of this plugin and run `:checkhealth nvim_treesitter`.
-This will help you find where the bug might come from.
+Before doing anything, make sure you have the latest version of this plugin and run `:checkhealth nvim_treesitter`. It can also help to update the parsers via `:TSUpdate`.
 
-## Feature `X` does not work for `{language}`...
+#### Feature `X` does not work for `{language}`...
```
stsewd, line 360: Looks like these should be h2 `##`

stsewd (commented):
Thanks! This looks great. I left some notes. Also, I tend separate long paragraphs in one line per sentence to avoid having long lines and makes it easy to see the diff when a sentence is changed :)

README.md
clason: Ah, I didn't know that this page was there (and maintained). (Several of these sections can be moved to the wiki, in fact, although wiki pages tend to be much less visible.)

README.md
clason: I'd have hoped Markdown to render these ;) I don't like em-dashes, though (prefer continental en-dashes with spaces).

README.md
clason: Not sure I prefer that. It makes sense hierarchically, but doesn't look as good typographically, especially for such short sections?

README.md
clason: same as above; it's less of a heading and more of a structured list...

clason: (should probably be consistent with the module list, though)

vigoux (approved):

stsewd (approved):

README.md
stsewd: yeah, no strong opinion there.

PR #682, opened 2020-11-21
Only update outdated_parsers on TSUpdate

theHamsta:
I have no idea why Github decided to close #640 but here is an updated version. It should now also work when `lockfile.json` changed after a plugin update and when `parser-info` is not writable.

Please also test. E.g. by

-  `:TSUpdate` (now it should write all installed parser revisions, will run for all parsers a last time)
- Change `lockfile.json` (only one revision -> should only update that parser on TSUpdate, or execute `write-lockfile.json` -> :TSUpdate should  update all parsers with new revision available)

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,6 @@ List of currently supported languages:
 
 <!--This section of the README is automatically updated by a CI job-->
 <!--parserinfo-->
-
 - [x] [bash](https://github.com/tree-sitter/tree-sitter-bash) (maintained by @TravonteD)
 - [x] [c](https://github.com/tree-sitter/tree-sitter-c) (maintained by @vigoux)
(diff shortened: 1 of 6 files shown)


theHamsta:
Bonus: fixes README.md CI

README.md
```diff
@@ -226,7 +226,6 @@ List of currently supported languages:
 
 <!--This section of the README is automatically updated by a CI job-->
 <!--parserinfo-->
-
```
theHamsta, line 229: Ok, this was the reason? Someone changed that section manually while it still was the same for the Lua script?

theHamsta: Exeriment. Will CI fail when this is fixed manually?

theHamsta: Result: it will work when REAMDE needs update and also if not.

vigoux (approved):

PR #684, opened 2020-11-22
javascript: : in pairs as punctuation.delimiter

rockerBOO:
Set : to be operator in javascript.

punctuation.delimiter

[old-operator-example](https://user-images.githubusercontent.com/15027/99914664-e6ad4e00-2ccc-11eb-8640-df957f6d3a62.png)

### DIFF
--- a/queries/javascript/highlights.scm
+++ b/queries/javascript/highlights.scm
@@ -49,6 +49,8 @@
   key: (property_identifier) @method
   value: (arrow_function))
 
+(pair ":" @punctuation.delimiter)
+
 (assignment_expression
   left: (member_expression
     property: (property_identifier) @method)
--- a/queries/javascript/highlights.scm
+++ b/queries/javascript/highlights.scm
@@ -49,8 +49,6 @@
   key: (property_identifier) @method
   value: (arrow_function))
 
-(pair ":" @punctuation.delimiter)
-
 (assignment_expression
   left: (member_expression
     property: (property_identifier) @method)
@@ -151,6 +149,8 @@
 ";" @punctuation.delimiter
 "." @punctuation.delimiter
 "," @punctuation.delimiter
+
+(pair ":" @punctuation.delimiter)
 
 [
   "--"
--- a/queries/javascript/highlights.scm
+++ b/queries/javascript/highlights.scm
@@ -49,7 +49,7 @@
   key: (property_identifier) @method
   value: (arrow_function))
 
-(pair ":" @punctuation.delimiter)
+(pair ":" @operator)
 
 (assignment_expression
   left: (member_expression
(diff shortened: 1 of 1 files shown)


kyazdani42:
in rust we do not highlight this, in lua we highlight `=` as operator in a table. Is this really a punctuation delimiter or a assignment character ^^ ?

rockerBOO:
mmm I like operator better here. I was replicating the matching from typescript for object type definition `type X {x: Type}` and type delimiter `x: MyType`. I still like operator but should all these  match or are they different?

rockerBOO:
Changed `:` to be an operator.

steelsojka:
I don't think I would consider this an operator at all. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#Table 

If you want a it colored a specific way I would extend the javascript query and give a custom highlight group. I think this might add confusion for some colorschemes.

rockerBOO:
https://tc39.es/ecma262/#sec-punctuators
This has `:` in the list of punctuators. I think returning this back to `(pair ":" @punctuation.delimiter` would be appropriate.

kyazdani42:
well depends, but for me `:` in an object is an operator for variable assignment kind of. Just to keep consistency between language, but i wouldn't mind highlighting as a punctuation delimiter (although `:` is not a delimiter, `;` is, or maybe my english is just too limited in this area to understand properly :) )

rockerBOO:
https://tc39.es/ecma262/#sec-object-initializer

This is the object initializer spec. This spec has these punctuations `{` `}` `,` `:`. 
`{`, `}` are `punctuation.bracket`. 
`,` is `punctuation.delimiter`

> [delimiter](https://www.merriam-webster.com/dictionary/delimiter): a character that marks the beginning or end of a unit of data

Using this definition, the delimiter of a key: value is `:`. 

I feel now that `punctuation.delimiter` matches throughout and matches the spec.

rockerBOO:
To further confuse things Atom sets it as a operator. (If I am reading properly)

https://github.com/atom/language-javascript/blob/master/grammars/javascript.cson#L386

queries/javascript/highlights.scm
```diff
@@ -49,6 +49,8 @@
   key: (property_identifier) @method
   value: (arrow_function))
 
+(pair ":" @punctuation.delimiter)
```
steelsojka, line 52: Can you move this to the section with the other punctuation?

steelsojka (changes requested):

rockerBOO:
Updated the positioning of this query.

steelsojka:
> To further confuse things Atom sets it as a operator. (If I am reading properly)
> 
> https://github.com/atom/language-javascript/blob/master/grammars/javascript.cson#L386

I tried to look in the text mate grammar for vscode's highlighting to try and compare and I couldn't navigate it very well. If you find another case this is tagged as an operator, we can just change it in another PR. I'm good this for now if you are.

steelsojka (approved):

PR #687, opened 2020-11-23
fix: update interface following languagetree merge

vigoux:
Quickfix to adapt to latest nvim master changes.

@steelsojka do you want to do this this way or do you have and alternative formulation ?


theHamsta (approved):
Can confirm that it works. Maybe change the requirement warning to require language tree?

steelsojka:
@vigoux Looks good. We'll need to remove the language tree configuration flag though in the highlight module right?

steelsojka:
@vigoux We also need to update the current injection queries to the new capture groups.

theHamsta:
Can you update the "your nvim is too old warning"? This saves us from answering a few issues that would be raised otherwise.

vigoux:
Updated the message, it should be better now

lua/nvim-treesitter/indent.lua
```diff
@@ -40,7 +40,7 @@ function M.get_indent(lnum)
   local parser = parsers.get_parser()
   if not parser or not lnum then return -1 end
 
-  local node = get_node_at_line(parser:parse():root(), lnum-1)
+  local node = get_node_at_line(parser:parse()[1]:root(), lnum-1)
```
kyazdani42, line 43: cool so i can use the language tree for indent ?

vigoux:
Huh, CI failure is painful... There is no nightly artifacts anymore, so I think we should download and build by hand ?

kyazdani42 (commented):
thats perfect, but is it possible to do just `parser:parse():root()` to get the main root maybe ? just a suggestion

lua/nvim-treesitter/highlight.lua
```diff
@@ -81,19 +81,14 @@ function M.attach(bufnr, lang)
...
-
-    M.highlighters[bufnr] = ts.highlighter.new(parser, query)
+  for k, v in pairs(config.custom_captures) do
+    hlmap[k] = v
   end
+
+  local query = queries.get_query(lang, "highlights")
+  if not query then return end
+
+  M.highlighters[bufnr] = ts.highlighter.new(parser, query)
```
steelsojka, line 91: The second argument to this is now an options object... You actually don't need to provide the query at all. It will get looked up by nvim lazily.

vigoux: The process here is just mimicking what is happening upstream, I don't think it should bother that much to have that here too ?

steelsojka:
@kyazdani42 The issue with that approach is, there is no main root tree now. Each language can have multiple trees that can occupy different regions of the file. I though about adding that call for 'legacy' usage, but I fell like it would end up as a bad practice for plugins and modules. We should be accounting for each tree that a language spans instead of just the first one.

lua/nvim-treesitter/indent.lua
vigoux: I think so, but that'll need some more work beforehand

theHamsta:
@vigoux maybe you can just wrap the error message with a `if vim.fn.getenv('CI') == 0 then`

theHamsta:
Probably it should be getent('CI') ~= vim.NIL

steelsojka (approved):

PR #689, opened 2020-11-23
javascript: Add operator to / for division

rockerBOO:
This currently has no highlights for `/` in division.

This adds `/` for division punctuation. https://tc39.es/ecma262/#prod-DivPunctuator

2020-11-23-115514_4480x1440_scrot

And doesn't match for `/` in tags (jsx/tsx)

2020-11-23-115555_4480x1440_scrot

### DIFF
--- a/queries/javascript/highlights.scm
+++ b/queries/javascript/highlights.scm
@@ -180,6 +180,7 @@
   "%="
 ] @operator
 
+(binary_expression "/" @operator)
 (ternary_expression ["?" ":"] @operator)
 
 "(" @punctuation.bracket


queries/javascript/highlights.scm
```diff
@@ -180,6 +180,7 @@
   "%="
 ] @operator
 
+(binary_expression "/" @operator)
```
steelsojka, line 183: Can this just move up into the list of operators above it?

rockerBOO: I do not know if this would be 100% accurate. In the spec its its own delineation from the other punctuation's. So I wanted it to match the case stated. I can move it to simplify but I am not sure.

steelsojka (approved):

PR #693, opened 2020-11-23
RFC: C/C++ macro injections

theHamsta:
With injections, it would be possible to self-inject C into macros
image
or C++
image

It would be good to have some kind of background highlighting to signalize the use that this is macro code (how to do this for all colorschemes?). The advantage of the current highlighting is that it makes clear were macro code starts and were it ends.
image

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -114,9 +114,11 @@
 (preproc_function_def
   name: (identifier) @function.macro)
 [
- (preproc_arg)
  (preproc_defined)
 ]  @function.macro
+
+
+ ;(preproc_arg)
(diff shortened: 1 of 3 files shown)


vigoux:
I think that we'll have this kind of changing background afterwards, for now let's experiment with language injection capabilities

vigoux (approved):

theHamsta:
Numpy header:
image

without injection:
image

Can't tell which one is better. Without injection really warns from the danger of macros and makes their range clear.

vigoux:
Huh I agree...

theHamsta:
Will we have blending of two highlights at some point? Or alpha colors? Then one could have the macro highlight with slight variations in the colors.

PR #695, opened 2020-11-23
lua: fix "then" highlighting

vigoux:


theHamsta:
Is this because of a newer version of the Lua parser (if yes you could increment the version)? It is not broken for me.

vigoux:
No it is only hapening in `elseif`s, and that with current parser version.

theHamsta (approved):

PR #696, opened 2020-11-23
c: fix parameter captures

vigoux:
Leveraging @theHamsta changes.

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -157,11 +157,8 @@
 (comment) @comment
 
 ;; Parameters
-(parameter_declaration
-  declarator: (identifier) @parameter)
-
-(parameter_declaration
-  declarator: (pointer_declarator) @parameter)
+((identifier) @parameter
+  (#has-ancestor? @parameter "parameter_declaration"))
 
 (preproc_params
   (identifier)) @parameter
--- a/queries/c/locals.scm
+++ b/queries/c/locals.scm
@@ -8,8 +8,12 @@
   name: (identifier) @definition.macro)
 (pointer_declarator
   declarator: (identifier) @definition.var)
-(parameter_declaration
-  declarator: (identifier) @definition.parameter)
+; (parameter_declaration
+;   declarator: (identifier) @definition.parameter)
+
+((identifier) @definition.parameter
+  (#has-ancestor? @definition.parameter "parameter_declaration"))
+
 (init_declarator
   declarator: (identifier) @definition.var)
 (array_declarator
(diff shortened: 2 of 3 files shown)


theHamsta:
Doesn't this make also the `reference_declarator` captures obsolete (they work like `pointer_declarator`). C++ highlights/locals.

theHamsta:
@vigoux can you update this PR for C++? The same change you did for `pointer_declarator` should apply for `address_declarator`

vigoux:
@theHamsta I added the C++ parameter thing, do you think it's good now ?

queries/c/locals.scm
```diff
@@ -8,8 +8,12 @@
...
 (pointer_declarator
   declarator: (identifier) @definition.var)
-(parameter_declaration
-  declarator: (identifier) @definition.parameter)
+; (parameter_declaration
+;   declarator: (identifier) @definition.parameter)
+
+((identifier) @definition.parameter
+  (#has-ancestor? @definition.parameter "parameter_declaration"))
+
```
theHamsta, line 16: I guess removing is sufficient since this query already comes from C

theHamsta (approved):
I guess this is good to go! We have to be careful no to use #has-... In combination with identifier too often. There're really a lot of identifiers in a file and you have to check all parents till the root

queries/c/locals.scm
theHamsta: Wornf position. I mean in the copy file. I'm on mobile right now. Too Little screen to big fingers...

queries/cpp/highlights.scm
```diff
@@ -19,6 +17,9 @@
 (optional_parameter_declaration
     declarator: (_) @parameter)
 
+((identifier) @parameter
+
+  (#has-ancestor? @parameter "parameter_declaration"))
```
theHamsta, line 22: Meant this one. It's already in C highlights

vigoux: Oh yeah sorry huh

queries/c/highlights.scm
```diff
@@ -157,11 +157,8 @@
 (comment) @comment
 
 ;; Parameters
-(parameter_declaration
-  declarator: (identifier) @parameter)
-
-(parameter_declaration
-  declarator: (pointer_declarator) @parameter)
+((identifier) @parameter
+  (#has-ancestor? @parameter "parameter_declaration"))
```
theHamsta, line 161: We probably want to avoid `#has-ancestor` since checking the parent has become expensive.

theHamsta (changes requested):

theHamsta:
Closing since this PR is stale for a long time and we want to avoid `has-ancestor?`. But I oberved a similar problem today with references to variadic template arguments

PR #710, opened 2020-11-26
Capture anonymous functions assigned to variables

### LINKED ISSUE #709: `queries/lua/locals.scm` does not capture anonymous functions assigned to variables
**Describe the bug**
```local my_fn_handle = function() some_code_here end``` could return a `@definintion.function` capture for `my_fn_handle`

### PR DESCRIPTION
**Pull request by ElPiloto**
Closes #709

### DIFF
--- a/queries/lua/locals.scm
+++ b/queries/lua/locals.scm
@@ -29,6 +29,11 @@
    (identifier) @definition.function) @scope)
 (function_definition) @scope
 
+(local_variable_declaration
+  (variable_declarator
+    (identifier) @definition.function)
+  (function_definition) @scope)
+
 (program) @scope
 ((if_statement) @scope)
 ((for_in_statement) @scope)
--- a/queries/lua/locals.scm
+++ b/queries/lua/locals.scm
@@ -27,12 +27,14 @@
 
 ((local_function
    (identifier) @definition.function) @scope)
-(function_definition) @scope
-
+[
 (local_variable_declaration
   (variable_declarator
     (identifier) @definition.function)
   (function_definition) @scope)
+(function_definition) @scope
+]
+
 
 (program) @scope
 ((if_statement) @scope)

### CODE REVIEW

File: queries/lua/locals.scm [hunk 1]
```diff
@@ -29,6 +29,11 @@
    (identifier) @definition.function) @scope)
 (function_definition) @scope
 
+(local_variable_declaration
+  (variable_declarator
+    (identifier) @definition.function)
+  (function_definition) @scope)
```
theHamsta, line 35: Isn't that @scope already mentioned earlier?

ElPiloto: It is helpful for my use case and, more importantly, in line with the other queries in locals.scm to have the `@definition.function` and `@scope` defined within the same query where possible.  This lets you link the different captures to each other e.g. this scope goes with this function.

But I could put my new query before the standalone `(function_definition) @scope` query if that's what you were thinking?

encode-this: I see - the issue is that the `function_definition` scope would show up twice since it would match two queries.  I was under the impression treesitter queries had precedence rules, but I am doubting that after reading `:h nvim-treesitter-query-extensions`. I'll use an alternation to get rid of the duplicate scope.

ElPiloto: I see - the issue is that the `function_definition` scope would show up twice since it would match two queries.  I was under the impression treesitter queries had precedence rules, but I am doubting that after reading `:h nvim-treesitter-query-extensions`. I'll use an alternation to get rid of the duplicate scope.

theHamsta: sorry, `(function_definition) @scope` was later. I don't know whether capturing a node twice is bad or not. @vigoux is the maintainer for Lua.

vigoux: Capturing a node twice is not really good for locals, for highlights it is actually fine though.

vigoux:
Little interrogation though, why not just adding :
```
(local_variable_declaration
  (variable_declarator
    (identifier) @definition.function)
  (function_definition))
```

That would match your function definition correctly, and still leverage the `function_definition` scope.

theHamsta:
That's what I was saying. But @ElPiloto prefers scopes and corresponding captures in one match.

ElPiloto:
Yeah the issue for me is how would you know what the correspondence is between scope and definition unless scope and definition live in the same query.  You could know that the definition is under the scope, but you could not know that the scope is "named" by the definition.

The updated version (with the [ ]-based alternation) should not yield duplicate scopes, so that should make everyone happy.

If that doesn't work for you guys, let me know and I'll just add it without the alternation and without the duplicate scope.  Won't help my personal use case, but still something that should be added to locals.lua. Let me know :)

vigoux:
Hi, sorry for the delay, could you rebase your branch on to latest master ? Thanks !

vigoux (approved):

PR #715, opened 2020-11-27
graphql: add injection to javascript and add comment query

carlitux:
@bkegley please review this

### DIFF
--- a/queries/javascript/injections.scm
+++ b/queries/javascript/injections.scm
@@ -1,1 +1,6 @@
 ((comment) @jsdoc)
+
+(call_expression
+	function: ((identifier) @_name
+		(#eq? @_name "gql"))
+	arguments: ((template_string) @graphql))
--- a/queries/graphql/highlights.scm
+++ b/queries/graphql/highlights.scm
@@ -98,6 +98,8 @@
 
 (description) @comment
 
+(comment) @comment
+
 (directive_location
   (executable_directive_location) @type.builtin)
 


queries/javascript/injections.scm
```diff
@@ -1 +1,6 @@
 ((comment) @jsdoc)
+
+(call_expression
+	function: ((identifier) @_name
+		(#eq? @_name "gql"))
+	arguments: ((template_string) @graphql))
```
kyazdani42, line 6: i did that on the draft PR for graphql, but this would also match for `${thing}` variables injected in the template string right ? so if you have
```javascript
gql`
 ${queries}

  ...some graphql types
`
```
the ${queries} would be highlighted in white.
not sure this is an issue we can resolve right now though

bkegley: image

It appears to be matching correctly for me. Although `:TSHighlightUnderCursor` seems to be matching the javascript grammar in the template string instead of the graphql grammar. e.g. `directive` is TSString instead of TSKeyword

theHamsta: The code of the playground has not be updated for multi-parsers. I just applied some duct tape that it will somehow run with the new API. No idea what it will show you within injected languages.

theHamsta: @steelsojka plans an update. He knows better than I how the language-tree is working: https://github.com/nvim-treesitter/playground/issues/11

kyazdani42 (approved):

queries/javascript/injections.scm
steelsojka: @theHamsta I'm working on updating the API's in nvim-treesitter first before I tackle the playground changes. I have a branch created and plan on getting it completed this weekend.

Issue #31, opened 2020-04-27
Way to find the correct parser


carlitux:
Hi,

I was playing with the plugin and I found these:

- javascript parser works well but not support for flowtype, not related to this plugin

- typescript has support for flowtype on files that has filetype javascript - https://github.com/tree-sitter/tree-sitter-typescript/pull/67

- on research atom do a mix of filetype - content regexp https://github.com/atom/language-typescript/pull/14/files

kyazdani42:
Hi,
I'm not too sure what you are talking about. Flow is a type system for javascript right ?
It shouldn't be used in typescript files as typescript already is a type system that wraps javascript.

If you could elaborate what you were trying to suggest here it would be nice, thanks.

carlitux:
Sorry I didn't explain to you well.

Yes, flow is a type system for javascript but **tree-sitter-typescript** supports it please see https://github.com/tree-sitter/tree-sitter-typescript/pull/67, **tree-sitter-javascript** doesn't support flow.

You can confirm this in this pull request of **atom** that uses **tree-sitter-typescript** for javascript files that have flow content.

How can I use **tree-sitter-typescript** in **nvim-treesitter** for javascript files that also have **(/\\*([^*]|\\*[^/])*|//.*)@flow** content ?

kyazdani42:
Well i really don't see the point of parsing flow inside typescript as they are different type systems.
To resolve that you could change the filetype of your javascript files to typescript so you can use the typescript parser instead.
I will investigate more on my end to find the proper way to handle that particular case.

I must also inform you that at the moment, we do not yet have query files for javascript and typescript. Those will come along soon, informations about that will be added to the README.

carlitux:
> To resolve that you could change the filetype of your javascript files to typescript so you can use the typescript parser instead.

I thought that also but doing that but issues raise with completion, diagnostic, and all the stuff that starts using tsserver instead of flow.

Only tree-sitter typescript is required.

> I must also inform you that at the moment, we do not yet have query files for javascript and typescript. Those will come along soon, informations about that will be added to the README.

Yes, I know I added some to test

Thank you very much for the plugin!

kyazdani42:
Yes you are right, i didn't think about other plugins.
Well i'm gonna work on fixing that in the next days.

vigoux:
I think that this issue will be fixed along with #63 , as it will provide a way to trigger various parsers for some filetypes. Am I wrong ?

[...]

PR #716, opened 2020-11-27
Allow configurations

carlitux:
Trying to use typescript for **flow typed** [issue related](https://github.com/nvim-treesitter/nvim-treesitter/issues/31) I set up as describes Readme file but as this lines 

https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/parsers.lua#L344-L354 

runs at require, any new setup has no effect because `ft_to_parsername` is not repopulated, and any `used_by` value is not used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,19 @@ EOF
 You can also skip step 2 and use `:TSInstallFromGrammar zimbu` to install directly from a `grammar.js` in the top-level directory specified by `url`.
 Once the parser is installed, you can update it (from the latest revision of the `main` branch if `url` is a Github repository) with `:TSUpdate zimbu`.
 
+## Update parsers used_by
+
+Sometimes needs to use some parser for different filetype.
+
+Add the following snippet to your `init.vim`:
+
+```vim
+lua <<EOF
+local parser_config = require "nvim-treesitter.parsers".get_parser_configs()
+parser_config.typescript.used_by = "javascriptflow"
+EOF
+```
+
 ## Adding queries
 
 Queries are what `nvim-treesitter` uses to extract informations from the syntax tree; they are
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1,7 +1,33 @@
 local api = vim.api
 local ts = vim.treesitter
 
-local list = {}
+local ft_to_parsername = {}
+
+local function update_ft_to_parsername(name, parser)
+  if type(parser.used_by) == 'table' then
+    for _, ft in pairs(parser.used_by) do
+      ft_to_parsername[ft] = name
+    end
+  end
+  ft_to_parsername[parser.filetype or name] = name
+end
+
+local list = setmetatable({}, {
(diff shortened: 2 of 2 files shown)


lua/nvim-treesitter/parser_configs.lua
```diff
@@ -0,0 +1,358 @@
...
+  M.list[parser] = options
+end
+
+function M.use_by(parser, filetype)
+  if (M.list[parser] ~= nil) then
+    local old_used_by = M.list[parser].used_by or {}
+    table.insert(old_used_by, filetype)
+    M.list[parser].used_by = old_used_by
+                print('updated the config')
+  end
```
theHamsta, line 355: This can be implemented by a meta table. It could also update the config after the call of setup. Can't we just remove those three methods when the table is now public?

theHamsta: Maybe the use_by can be a shortcut for newbies. But then it could also invalidate the filetype to parser cache and not print anything.

carlitux: Sorry, I thought I removed all print line code.

About meta table I was thinking the same, I will work soon on that

theHamsta: I think a better solution would be if you function would just set the mapping in `ft_to_parsername` in parser.lua. This also works after setup.

carlitux: @theHamsta I think I simplified the changes with metatables and more clear interface to configure. Waiting for your feedback

I cleaned history as well to avoid many commits.

theHamsta: Hi. Sorry for all the efforts. What I actually meant to say was that you could just do 

```
function M.set_parser_for_filetype(ft, parser_lang) 
     ft_to_parsername[ft] = lang
end
```
because most people like you want to set the parser for their filetype and not vice-versa. I came with this solution very late.

But your solution is maybe a bit more consistent (only one way to configure things).

theHamsta (approved):

vigoux (approved):
Elegant and simple, that's all I need

README.md
```diff
@@ -274,6 +274,19 @@ EOF
 You can also skip step 2 and use `:TSInstallFromGrammar zimbu` to install directly from a `grammar.js` in the top-level directory specified by `url`.
 Once the parser is installed, you can update it (from the latest revision of the `main` branch if `url` is a Github repository) with `:TSUpdate zimbu`.
 
+## Update parsers used_by
+
+Sometimes needs to use some parser for different filetype.
+
+Add the following snippet to your `init.vim`:
```
fabiomcosta, line 281: This is actually super helpful for me, I appreciate it!
The text on this section seems to be lacking some context and why people would want to do this.

After reading this: https://github.com/nvim-treesitter/nvim-treesitter/issues/31

Wouldn't people need to have this code on their vim configs in order to make this change work?
```
au! BufRead,BufNewFile *.js if getline(1) =~ '//\s*@flow' | setf javascriptflow | endif
```

PR #722, opened 2020-11-29
graphql: add missing highlights

bkegley:
* enum values
* input parameters
* 'implements' keyword
* include '@' in directive name
* fragment definition and spread

I included comments here which is a duplicate of #715 as there seems to still be discussion about merging the injection query. I can remove that if we want.

### DIFF
--- a/queries/graphql/highlights.scm
+++ b/queries/graphql/highlights.scm
@@ -22,6 +22,9 @@
 (directive_definition
   (name) @type)
 
+(directive_definition
+  "@" @type)
+
 (scalar_type_extension
   (name) @type)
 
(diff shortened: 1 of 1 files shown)


queries/graphql/highlights.scm
```diff
@@ -22,6 +22,9 @@
 (directive_definition
   (name) @type)
 
+(directive_definition
+  "@" @type)
```
theHamsta, line 26: The "@" should have type highlighting?

theHamsta (approved):

queries/graphql/highlights.scm
bkegley: Yeah. It's part of the directive name.

queries/graphql/highlights.scm
```diff
@@ -98,6 +111,8 @@
 
 (description) @comment
 
+(comment) @comment
```
kyazdani42, line 114: i've merged the other PR which handles the `(comment)` node, you can remove it :)

bkegley: Done!

vigoux (approved):

PR #729, opened 2020-11-30
Add kotlin

tormodatt:
does have the most of the basics, but has some issues as well. I'll look at scoping next. any tips are welcome :)

Skjermbilde 2020-11-30 kl  23 09 17

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -116,6 +116,7 @@ list.kotlin = {
     url = "https://github.com/QthCN/tree-sitter-kotlin",
     files = { "src/parser.c" },
   },
+  maintainers = {"@tormodatt"},
 }
 
 list.html = {
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -111,6 +111,13 @@ list.java = {
   maintainers = {"@p00f"},
 }
 
+list.kotlin = {
+  install_info = {
+    url = "https://github.com/QthCN/tree-sitter-kotlin",
+    files = { "src/parser.c" },
+  },
+}
+
 list.html = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-html",
--- a/queries/kotlin/highlights.scm
+++ b/queries/kotlin/highlights.scm
@@ -129,13 +129,17 @@
 [
 "["
 "]"
-"$"
-"${"
 "{"
 "}"
 "("
 ")"
 ] @punctuation.bracket
+
+[
+"$"
+"${"
+"}"
+] @punctuation.special
 
 ;; Comments
 (comment) @comment
--- a/queries/kotlin/highlights.scm
+++ b/queries/kotlin/highlights.scm
@@ -1,0 +1,158 @@
+;;; Kotlin 
+;; Issues
+
+; mapOf with type definition AND multiple entries are not highligted correctly. Issue in parser. Eg.
+; a = mapOf<String, String>(AA.value to AA.value, AA.value to AA.value)
+
+
+(function_modifier) @type
+
+;; String
+
+(line_string_literal) @string
+(multi_line_string_literal) @string
+(line_string_literal (interpolated_identifier)) @string
+(line_string_literal (interpolated_expression (line_string_literal))) @string
+
+;; Constants
+; Assume all-caps names are constants
+((simple_identifier) @constant
+ (#vim-match? @constant "^[A-Z][A-Z_]+"))
+
+
+;; Variables
+(class_parameter (simple_identifier) @variable.builtin)
+
+; Class level
+(property_declaration (variable_declaration (simple_identifier) @variable.builtin))
+
+(statements (call_expression (navigation_expression (simple_identifier) @variable.builtin)))
+
+(directly_assignable_expression (navigation_expression (simple_identifier) @variable.builtin))
+
+(directly_assignable_expression (simple_identifier) @variable.builtin)
+
+
+; TODO not supported yet
+; (navigation_expression (simple_identifier) @type
+;   (#is-not? local))
+; (directly_assignable_expression (simple_identifier) @type
+;   (#is-not? local))
+
+
+;; Property access syntax
+; TODO
+; a = get().value etc
+(assignment (navigation_expression (navigation_suffix) @type ))
+; val a = A.get().value etc
+(property_declaration  (navigation_expression (navigation_suffix) @type ))
+; value in `mapOf(A.value to B.value)` etc.
+(infix_expression  (navigation_expression (navigation_suffix) @type ))
+; java in `Car::class.java`
+(value_argument (navigation_expression (navigation_suffix) @type ))
+
+
+;; Interpolated
+
+(interpolated_identifier) @type
+(interpolated_expression) @type
+
+
+;; Builtin functions
+((simple_identifier) @function.builtin
+ (#vim-match? @function.builtin "(^let$|^listOf$|^mapOf$|^map$)"))
+
+
+((simple_identifier) @function.builtin
+ (#vim-match? @function.builtin "^to$"))
+
+
+(modifiers (annotation (user_type (type_identifier)))) @attribute
+
+(annotation) @attribute
+
+(integer_literal) @number
+
+(boolean_literal) @boolean
+
+
+;; it
+(indexing_expression (simple_identifier) @type
+  (#vim-match? @type "^it$"))
+
+
+;; Keywords
+
+[
+ "enum"
+ "as"
+ "class"
+ "object"
+ "data"
+ "val"
+ "init"
+ "private"
+ "var"
+ "break"
+ "by"
+ "fun"
+ "companion"
+ "return"
+ "constructor"
+] @keyword
+; const does not work
+
+
+;; Conditionals
+[
+"if"
+"is"
+"else"
+"when"
+] @conditional
+
+
+;; Includes
+
+"import" @include
+"package" @include
+
+
+;; Punctuation
+
+[
+";"
+"."
+","
+] @punctuation.delimiter
+
+[
+"["
+"]"
+"$"
+"${"
+"{"
+"}"
+"("
+")"
+] @punctuation.bracket
+
+;; Comments
+(comment) @comment
+
+; Methods
+
+(function_declaration (simple_identifier) @function)
+
+
+[
+"="
+"-"
+"->"
+"+"
+"++"
+"--"
+"*"
+] @operator
+
+(ERROR) @error
--- a/queries/kotlin/highlights.scm
+++ b/queries/kotlin/highlights.scm
@@ -25,7 +25,7 @@
 
 ;; Variables/fields
 ; attribute in data classes etc.
-(class_parameter (simple_identifier) @attribute)
+(class_parameter (simple_identifier) @field)
 ; variable in normal classes
 (property_declaration (variable_declaration (simple_identifier) @variable))
 ; accessed field in blocks. `logger` in `logger.info("")`
@@ -54,7 +54,7 @@
 (boolean_literal) @boolean
 
 ;; Annotations
-(annotation (single_annotation) @annotation)
+(annotation (single_annotation) @attribute)
 
 ;; Types
 ; TODO more type definitions
--- a/queries/kotlin/highlights.scm
+++ b/queries/kotlin/highlights.scm
@@ -1,19 +1,21 @@
 ;;; Kotlin
-;; Issues
-
-; mapOf with type definition AND multiple entries are not highligted correctly. Issue in parser. Eg.
-; a = mapOf<String, String>(AA.value to AA.value, AA.value to AA.value)
-
-;; parser does not recognice `in` and `downto` in `for (i in 5..1) print(i)`
 
 ;; Strings
-
 (line_string_literal) @string
 (multi_line_string_literal) @string
 ; Interpolated
 (interpolated_identifier) @none
 (interpolated_expression) @none
 
+
+;; Suffixes
+;; Issues:
+;; - `doors` in `c = car.doors` not reconized
+(navigation_expression (navigation_suffix) @property ) ; TODO to method and resetable italics
+(assignment (navigation_expression (navigation_suffix) @none ))
+(call_expression (navigation_expression (navigation_suffix) @none ))
+(navigation_expression (navigation_expression (navigation_suffix) @none ))
+(parenthesized_expression (navigation_expression (navigation_suffix) @property))
 
 ;; Constants
 ; Assume all-caps names are constants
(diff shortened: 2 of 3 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -111,6 +111,13 @@ list.java = {
   maintainers = {"@p00f"},
 }
 
+list.kotlin = {
+  install_info = {
+    url = "https://github.com/QthCN/tree-sitter-kotlin",
+    files = { "src/parser.c" },
```
theHamsta, line 117: You want to be the maintainer?

queries/kotlin/highlights.scm
```diff
@@ -0,0 +1,162 @@
...
+; a = mapOf<String, String>(AA.value to AA.value, AA.value to AA.value)
+
+
+(function_modifier) @type
+
+;; String
+
+(line_string_literal) @string
+(multi_line_string_literal) @string
+(line_string_literal (interpolated_identifier)) @string
```
theHamsta, line 14: interpolated_identifier can be `@none` and `@variable`. See dart highlights.

theHamsta: how is this different from line 12? But you can give highlights to `interpolated_identifier`

queries/kotlin/highlights.scm
```diff
@@ -0,0 +1,162 @@
...
+(line_string_literal (interpolated_expression (line_string_literal))) @string
```
theHamsta, line 15: `(interpolated_expression) @none` ? None undoes the string highlighting for the range of the interpolation. Is this something like Python f-strings or dart $ interpolations `"three: ${ 1 + 2}"`?

queries/kotlin/highlights.scm
```diff
@@ -0,0 +1,162 @@
...
+; Assume all-caps names are constants
+((simple_identifier) @constant
+ (#vim-match? @constant "^[A-Z][A-Z_]+"))
+
+
+;; Variables
+(class_parameter (simple_identifier) @variable.builtin)
+
+; Class level
+(property_declaration (variable_declaration (simple_identifier) @variable.builtin))
```
theHamsta, line 27: Can you explain what simple_identifier is that it is `@variable.builtin`? `@variable.builtin` is usually for keyword like variables like `this` or `self`

queries/kotlin/highlights.scm
```diff
@@ -0,0 +1,162 @@
...
+(interpolated_identifier) @type
+(interpolated_expression) @type
+
+
+;; Builtin functions
+((simple_identifier) @function.builtin
+ (#vim-match? @function.builtin "(^let$|^listOf$|^mapOf$|^map$)"))
+
+
+((simple_identifier) @function.builtin
```
theHamsta, line 66: Ok, simple is just everything builtin? Sorry, for previous doubts...

queries/kotlin/highlights.scm
```diff
@@ -0,0 +1,162 @@
...
+ (#vim-match? @function.builtin "^to$"))
+
+
+(modifiers (annotation (user_type (type_identifier)))) @attribute
```
theHamsta, line 70: Why all this inner matching when you capture all of it? Isn't every annotation `@attribute`?

theHamsta (commented):
Seems to be a very complete start! Thank you for working on this!

queries/kotlin/highlights.scm
```diff
@@ -0,0 +1,162 @@
...
+; (navigation_expression (simple_identifier) @type
+;   (#is-not? local))
+; (directly_assignable_expression (simple_identifier) @type
+;   (#is-not? local))
+
+
+;; Property access syntax
+; TODO
+; a = get().value etc
+(assignment (navigation_expression (navigation_suffix) @type ))
```
theHamsta, line 46: Which of `a = get().value` would be highlighted type?

queries/kotlin/highlights.scm
```diff
@@ -0,0 +1,162 @@
...
+ (#vim-match? @constant "^[A-Z][A-Z_]+"))
+
+
+;; Variables
+(class_parameter (simple_identifier) @variable.builtin)
+
+; Class level
+(property_declaration (variable_declaration (simple_identifier) @variable.builtin))
+
+(statements (call_expression (navigation_expression (simple_identifier) @variable.builtin)))
```
theHamsta, line 29: No, simple_indentifier is any identifier and not only built-in

queries/kotlin/highlights.scm
```diff
@@ -0,0 +1,162 @@
+;;; Kotlin 
+;; Issues
+
+; mapOf with type definition AND multiple entries are not highligted correctly. Issue in parser. Eg.
+; a = mapOf<String, String>(AA.value to AA.value, AA.value to AA.value)
+
```
theHamsta, line 6: You could add a `@type` highlighting for the typical upperclass naming convention. Like in C++/Java/Python

queries/kotlin/highlights.scm
```diff
@@ -0,0 +1,162 @@
...
+"="
+"-"
+"->"
+"+"
+"++"
+"--"
+"*"
+] @operator
+
+(ERROR) @error
```
theHamsta, line 162: Maybe add
```
     ( call_expression
        simple_identifier @function)
```

queries/kotlin/highlights.scm
theHamsta: also
```
(type_identifier) @type
```

queries/kotlin/highlights.scm
```diff
@@ -0,0 +1,162 @@
...
+
+[
+"="
+"-"
+"->"
+"+"
+"++"
+"--"
+"*"
+] @operator
```
theHamsta, line 160: `@parameter` could also be include. Also fine to add in later PR.

queries/kotlin/highlights.scm
theHamsta: so `(simple_identifier) @variable` at first line

lua/nvim-treesitter/parsers.lua
tormodatt: for the hightlight-queries and/or the parser? I don't have experiance any with writing languageparsers (yet)

queries/kotlin/highlights.scm
tormodatt: ah, nice. didn't know about `@none`. I'll look into that

queries/kotlin/highlights.scm
tormodatt: sorry for abusing some of the highlight categories. for the first version I just wanted to highlight the different categories in different colors. I'll go through these once more. this query will hit `variableA` in things like `private val variableA: Int = 2`. maybe `@field` is the correct one to use here?

I plan to move this to support local scoping if I find out how that works

queries/kotlin/highlights.scm
tormodatt: correct! forgot to remove this. will fix

queries/kotlin/highlights.scm
tormodatt: `value`. will update comment. and the the `@type` is wrong to use here as well. maybe `@property`?

lua/nvim-treesitter/parsers.lua
theHamsta: for the highlights.

queries/kotlin/highlights.scm
theHamsta: yes, `@field` or `@property`

queries/kotlin/highlights.scm
tormodatt: clean up this part a bit in the latest commit

queries/kotlin/highlights.scm
tormodatt: changed in latest commit

queries/kotlin/highlights.scm
```diff
@@ -1,36 +1,44 @@
...
-
-;; String
+;; Strings
 
 (line_string_literal) @string
 (multi_line_string_literal) @string
-(line_string_literal (interpolated_identifier)) @string
-(line_string_literal (interpolated_expression (line_string_literal))) @string
+; Interpolated
+(interpolated_identifier) @none
```
theHamsta, line 14: can be `(interpolated_identifier) @none @variable` (default has not highlighting set for variable, but some users set colors for basic identifiers)

queries/kotlin/highlights.scm
tormodatt: annotations are actually superclass of type_identifier, so I needed to be bit more specific

tormodatt:
thank you for the feedback!

I tried to get the hightlights as simliar to Intellij as possible this evening, just to test the limits of nvim-treesitter. I think a came pretty close (Intellij on the left, nvim on the right):

Skjermbilde 2020-12-02 kl  21 17 14
Skjermbilde 2020-12-02 kl  21 18 37
(set to the same font, color scheme etc)

there are still some issues though, regarding consistancy of fields, short classes beeing reconized as enums etc. and I need to find a list of builtin functions somewhere

queries/kotlin/highlights.scm
```diff
@@ -1,36 +1,44 @@
...
+(property_declaration (variable_declaration (simple_identifier) @field))
+; accessed field in blocks. `logger` in `logger.info("")`
+(statements (call_expression (navigation_expression (simple_identifier) @field)))
 
-; Class level
-(property_declaration (variable_declaration (simple_identifier) @variable.builtin))
+(statements (navigation_expression (simple_identifier) @field))
 
-(statements (call_expression (navigation_expression (simple_identifier) @variable.builtin)))
+(call_expression (navigation_expression (simple_identifier) @field))
```
theHamsta, line 34: is that rather a `@method` when you can call it?

queries/kotlin/highlights.scm
```diff
@@ -1,36 +1,44 @@
...
 
-(directly_assignable_expression (navigation_expression (simple_identifier) @variable.builtin))
+; TODO check if this is needed
+(directly_assignable_expression (navigation_expression (simple_identifier) @string))
+; `variable` in `variable = car.doors`
+(directly_assignable_expression (simple_identifier) @field)
 
-(directly_assignable_expression (simple_identifier) @variable.builtin)
+(lambda_parameters) @field
```
theHamsta, line 41: `@parameter` ?

queries/kotlin/highlights.scm
```diff
@@ -41,44 +49,55 @@
...
+;; Booleans
 (boolean_literal) @boolean
 
+;; Annotations
+(annotation (single_annotation) @annotation)
+
+;; Types
+; TODO more type definitions
+(variable_declaration (user_type (type_identifier) @type))
+(callable_reference (type_identifier) @type)
```
theHamsta, line 88: can't this be reduced to `(type_identifier) @type`

theHamsta:
Fairly complete now. Nice work! Our CI is failing to load your highlights...

theHamsta:
> here are still some issues though, regarding consistancy of fields

I think we can iterate on that. You (and other users) will see more things to improve when you actually use the highlights.

queries/kotlin/highlights.scm
tormodatt: that would include annotations and classnames, eg.: `Throws` in `@Throws` and `Test` in `class Test(...)`

tormodatt: could probably override that though

theHamsta:
Maybe useful: https://github.com/nvim-treesitter/playground/pull/13

tormodatt:
`./scripts/check-queries.lua:41: Invalid capture @annotation in highlights for kotlin.`

Shouldn't that be recognized? I'm able to change the hightlight by changing `TSAnnotation` locally.

https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/highlight.lua#L31

theHamsta:
`@annotation` is not it CONTRIBUTING.md. I guess that concept changed to `@attribute` though I like annotation better.

from C++
```
; Annotations (not fully supported by parser)

((ERROR) @attribute
         (#vim-match? @attribute  "\[?\[.*\]\]?.*$"))
(attribute) @attribute
```

theHamsta:
Can you add `@variable` for `(interpolated_identifier)` and normal identifiers (normal identifiers best in first line. Like in https://github.com/nvim-treesitter/nvim-treesitter/blob/407ff057db0216afd3b36d8564980a13de9614e6/queries/java/highlights.scm#L5)? Almost all languages have it. Then we can merge! :tada:

tormodatt:
Looks like this now:
(with "TSProperty" = purple, "TSVariable" = white/none):
Skjermbilde 2020-12-06 kl  01 17 27
Skjermbilde 2020-12-06 kl  01 17 38

(with "TSProperty" = white/none, "TSVariable" = purple):
Skjermbilde 2020-12-06 kl  01 22 45
Skjermbilde 2020-12-06 kl  01 22 53

theHamsta (approved):

queries/kotlin/highlights.scm
```diff
@@ -0,0 +1,186 @@
...
...o|chunked|chunkedSequence|count|drop|dropLast|dropLastWhile|dropWhile|elementAtOrElse|elementAtOrNull|filter|filterIndexed|filterIndexedTo|filterNot|filterNotTo|filterTo|find|findLast|first|firstOrNull|flatMap|flatMapIndexed|flatMapIndexedTo|flatMapTo|fold|foldIndexed|foldRight|foldRightIndexed|forEach|forEachIndexed|getOrElse|getOrNull|groupBy|groupByTo|groupingBy|indexOfFirst|indexOfLast|last|lastOrNull|map|mapIndexed|mapIndexedNotNull|mapIndexedNotNullTo|mapIndexedTo|mapNotNull|mapNotNullTo|mapTo|max|maxBy|maxByOrNull|maxOf|maxOfOrNull|maxOfWith|maxOfWithOrNull|maxOrNull|maxWith|maxWithOrNull|min|minBy|minByOrNull|minOf|minOfOrNull|minOfWith|minOfWithOrNull|minOrNull|minWith|minWithOrNull|none|onEach|onEachIndexed|partition|random|randomOrNull|reduce|reduceIndexed|reduceIndexedOrNull|reduceOrNull|reduceRight|reduceRightIndexed|reduceRightIndexedOrNull|reduceRightOrNull|reversed|runningFold|runningFoldIndexed|runningReduce|runningReduceIndexed|scan|scanIndexed|scanReduce|scanReduceIndexed|single|singleOrNull|slice|sumBy|sumByDouble|sumOf|take|takeLast|takeLastWhile|takeWhile|toCollection|toHashSet|toList|toMutableList|toSet|windowed|windowedSequence|withIndex|zip|zipWithNext)$"))
```
theHamsta, line 54: Is there any specific difference between this category and the latter (with print, println)? They are not real marcos, right?

tormodatt: correct! the difference is that these are builtin extention functions; like `someString.trim()` where .trim() must be called on a string, whilst the other category are standalone builtin functions that can be called anywhere. like `println(...)` or `val a = listOf(1, 2, 3)`.

the reason I divided them is that that is what Intellj does, and that is what I'm used to.

tormodatt: so it's not really macros like in C. if this is to far a stretch, let me know and I will change it

theHamsta: If you like the highlighting, you could add a new category `@method.builtin` and default link it to a highlight of your choice. You'd need to add it to CONTRIBUTING.md to make the linter happy.

theHamsta: @tormodatt I've seen a few advent of code solutions for kotlin. I guess it's really a bit macro-like. So you could also leave it this way. Would you consider the PR as ready to merge?

tormodatt: yeah, one could argue that this is kotlin answer to macros, but method.builtin is definetly correct as well.

maybe we can merge this now, and consider adding a category later.

theHamsta: image

Maybe fix the constant regex later. I guess you need to add `^...$` or else it will be a constant with two uppercase letter only.
 
Your highlight for the built-in methods is really handy.

tormodatt: eksisting is `(#vim-match? @constant "^[A-Z][A-Z_]+"))`, the `+` makes it match more than two letters. I agree that we should add `$`, but then we need to match numbers as well. I'm not able to make that work though. Tried both:
`^[A-Z][A-Z_\\d]+$` and `^[A-Z][A-Z_\d]+$` but none of those matches `PLATE2` in `private const val PLATE2 = "asdf$PLATE"`. 

what regex rules should work here?

tormodatt: > Your highlight for the built-in methods is really handy.

thanks! I made a simple script for fetching the functions from kotlin stdlib (just the most used classes for now). are there any suitable place to put this, or should I just put it in my own repo for now?

theHamsta: I would just copy the regex from Python, C++ or Java for constants.
E.g. in c++ `((identifier) @constant
                        (#match? @constant "^[A-Z][A-Z_0-9]*$"))`

theHamsta: It's possible that `\` do not completely work the way the should in match. they are fed like that into `vim.fn.match`

`vim.fn.match(vim.fn.escape('\\v'..your_text, '\\'))`

PR #731, opened 2020-12-02
css: add indent support

wingyplus:
When you enter a new line after `{` or new line `o` in command mode, it'll enter a new line with auto indent. For example:

```
a {< enter
  |
}
```

The indent when enter a new line after properties declaration is also support:

```
a {
  color: black;< enter
  |
}
```

NOTE: `|` is a cursor.

### DIFF
--- a/queries/css/indents.scm
+++ b/queries/css/indents.scm
@@ -1,0 +1,9 @@
+[
+  (block)
+  (declaration)
+] @indent
+
+[
+  "}"
+] @branch
+


queries/css/indents.scm
```diff
@@ -0,0 +1,9 @@
+[
+  (block)
+  (declaration)
+] @indent
+
+[
+  "}"
+] @branch
```
kyazdani42, line 8: shouldn't `{` also be a branch ?
```css
.some-class
{
  |# you want the cursor here
}

kyazdani42: although this shouldn't be happening too often ^^

wingyplus: I've been trying to add `"{"` to `@branch` but it's not works. :(

kyazdani42: maybe its not a node, we'll merge this as it is and iterate if we find unwanted behaviors.

kyazdani42 (approved):

PR #736, opened 2020-12-03
Update js ts

theHamsta:
PR just for reference.

Problem:

typescript does not inherit the parameter rule from JS. So all `@definition.parameter` queries will not be accepted by the typescript parser.

How to continue? 

- Not let typescript inherit from JS
- Let both JS and TS inherit from a common subset
- Update JS only (does not harm old TS parser)

@steelsojka

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -105,10 +105,10 @@
     "revision": "02e774c911d123ea3fbe5273cf9d987fa88dd3fb"
   },
   "tsx": {
-    "revision": "73afadbd117a8e8551758af9c3a522ef46452119"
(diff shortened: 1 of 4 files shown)


theHamsta:
@kyazdani42 but the ts-parser can import arbitrary grammar rules from JS and decide to have different categories for parameters.

TS: will parse required_parameter will it's just `identifier` in JS.

kyazdani42:
i didn't think of the locals thing indeed. Not sure which solution we should aim for, i'll leave that for you guys to decide :) I believe inheriting a common query file would be ideal.

theHamsta:
This is just not relevant anymorre

PR #737, opened 2020-12-04
fix the problem 'error during compilation'

jerrywang1981:
when install typescript, I got error 'error during compilation', this line of code change fixed my problem.

the error was

`for loop initial declarations are only allowed in c99 mode`

### DIFF
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -62,6 +62,7 @@ function M.select_compiler_args(repo)
         '-shared',
         '-Os',
         '-lstdc++',
+        '-std=gnu99',
   }
   if fn.has('win32') == 0 then
     table.insert(args, '-fPIC')


theHamsta:
Interesting to see that setting the standard to gnu99 does not break C++ compilation as C99 did. We'll need to test this also on other compilers. Also look look into commit history but we had this flag in the past but it got removed (remember discussions around this).

Which compiler/OS version are you using that does not have C99 as a default?

jerrywang1981:
I'm on CentOS 6, I tested in Mac but not had the problem, maybe this change should only apply to linux? 
better reject this PR.

```
[jerry@VM-0-13-centos ~]$ gcc --version
gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

jerrywang1981:
@theHamsta feel free to reject the PR. I upgrade gcc to `gcc (GCC) 7.3.1 20180303` in my CentOs and not having the problem.

vigoux:
Thanks for this pr and investigating though!
This version thing of the compiler is good to know.

theHamsta:
Adding this flag lead to compilation errors with clang when compiling the C++ files.

PR #739, opened 2020-12-04
Add "raise" to ruby keywords

kassio:


theHamsta:
@TravonteD 

Raise does not exist as token. Can you check what it is instead? I guess you could do

```
((method_call
   (identifier) @keyword)
   (#eq? @keyword "raise")

**Comment by kassio** [comment]
@theHamsta you don't use ruby and this is my first contact with treesitter 😄 

> Is raise a built in function function.builtin?

It seems to be an identifier:
node


I did as you suggested. What do you think?

**Comment by theHamsta** [comment]
@TravonteD the ruby syntax uses Exception highlighting for raise

  syn keyword rubyException raise fail catch throw
 hi def link rubyException		Exception

**Comment by kassio** [comment]
@theHamsta, @TravonteD again sorry, I don't know much about treesitter queries, but I tried to add the `@exception` values:

```diff
@@ -14,10 +14,8 @@
  "def"
  "do"
  "end"
- "ensure"
  "module"
  "next"
- "rescue"
  "retry"
  "return"
  "then"
@@ -45,14 +43,20 @@
  "while"
  ] @repeat

+[
+ "cache"
+ "ensure"
+ "fail"
+ "rescue"
+ "rescue"
+ "throw"
+ ] @exception
+
 (constant) @constant

 ((identifier) @keyword
  (#vim-match? @keyword "^(private|protected|public)$"))

-((identifier) @keyword
- (#eq? @keyword "raise"))
-
 ; Function calls
```

But this makes the check fail:
```console
Checking ruby highlights
Check failed:
...-0c8d6ab/share/nvim/runtime/lua/vim/treesitter/query.lua:119: query: invalid node type at position 403
./scripts/check-queries.lua: line 1: --: command not found
./scripts/check-queries.lua: line 3: syntax error near unexpected token `('
```

queries/ruby/highlights.scm
```diff
@@ -50,6 +50,9 @@
 ((identifier) @keyword
  (#vim-match? @keyword "^(private|protected|public)$"))
 
+((identifier) @keyword
+ (#eq? @keyword "raise"))
```
TravonteD, line 54:
```suggestion
((identifier) @exception
 (#eq? @exception "raise"))
```
It should be done this way

theHamsta:
@kassio maybe this will help you with the queries: https://github.com/nvim-treesitter/playground/pull/13

The parser defines which nodes are valid. `"raise"` does not work since it isn't a separate node in the syntax tree, instead it's an identifier. `node-types.json` in tree-sitter-ruby gives you a list of nodes (or string literals in grammar.js).

theHamsta (approved):

PR #742, opened 2020-12-07
fix: find parsers correctly Explicitely installed parsers will not be found implicitely

maisiliym:


theHamsta (approved):
@vigoux can you remember that we would use internal APIs here (in case it changes in future)?

vigoux:
I will remember it, otherwise, please leave a little `HACK` there so that we remember.

maisiliym:
> I will remember it, otherwise, please leave a little `HACK` there so that we remember.
done

vigoux (approved):

PR #673, opened 2020-11-19
docs: Update docs for missing groups, re-order alphabetically

rockerBOO:
- Adding missing groups. TSTag, TSTagDelimiter, TSComment (maybe a couple others)
- Reorder the docs/code to be alphabetical to make it easier to scan.

PR #746, opened 2020-12-07
Fix some missing highlight and add placeholder in docs

mandos:
This is a little blind fix, I found there was missing highlight for TSVariable and digging deeper I found missing documentation for some other highlight too. I just fix it as I could with placeholders in documentation. @theHamsta could you review it, I same that you are mostly author of this part of plugin.


mandos:
Closed because of #673

PR #747, opened 2020-12-07
Update ruby parser

theHamsta:
Includes fixes for float rationals and keyword parameters

I guess it would be nice to have a highlight for rational.

image

Maybe operator?
image

fyi: @TravonteD


vigoux:
I'd use `@type`.

vigoux:
And I can't see the fixes, is that from my side?

theHamsta:
Fixes were done in tree-sitter-ruby

vigoux (approved):

PR #749, opened 2020-12-09
Update Elm parser

theHamsta:
No new nodes only a fix for trailing spaces

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -21,7 +21,7 @@
     "revision": "9432021ba6e96124ff1b6bd91f79c0a2e3c87cf4"
   },
   "elm": {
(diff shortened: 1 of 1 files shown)


vigoux:
I think we should set the auto updater to update parsers with no changes and report parsers that need changes.
That would allow us to avoid doing updates one by one.

vigoux (approved):

theHamsta:
It's automated as soon as we can update JS/TS parser.

theHamsta:
I thought about doing one PR per parser but that would involve generating the workflow file to have a per language matrix. Would require additional engineering to find out how to do this.

theHamsta:
We could have one PR for failing parsers and one PR for parsers without failing tests. This would be totally feasible.

PR #752, opened 2020-12-10
lockfile.json: add erlang parser revision to lockfile

wingyplus:
Currently, nvim-treesitter always show error cannot download tar file because erlang revision isn't in the lockfile list.


wingyplus:
Oh, I just see that it's duplicated with #726. Close the PR then.

PR #756, opened 2020-12-10
Enable automatic parser updateing with a hack (skip typescript for now)

theHamsta:
When we skip javascript/typescript it should be possible to update the parsers automatically.

This is a lazier solution than updating all parsers one after another and test with another nvim instance whether this causes a break.


theHamsta:
I also tried to have two PRs (one for working updates, one for not supported updates). It was too much noise.

kyazdani42 (approved):

PR #764, opened 2020-12-13
Fix doc typo

lunjon:
Found a minor typo in the `ts_utils` section.


theHamsta (approved):
Thanks for fixing that!

PR #770, opened 2020-12-16
first draft of YAML highlighting

dimbleby:
A first draft of highlighting for YAML.  

Some comments:

- This is my first time with treesitter, I've very likely done things wrong, corrections welcome!
 
- In particular there are a couple of things that aren't working, in ways that I don't understand.

First:
```
(flow_pair key: (flow_node) @field)
```
is supposed to catch the `foo` in `mapping: { foo: bar }`, but seems to have no effect.  I don't see how this is different from the handling of `block_mapping_pair`, which seems to work fine

Second, including `"+"` in the list of punctuation characters causes errors.  I've no idea how this is different from the other punctuation characters.

- I have treated keys as `field`, whereas the JSON highlighter treats them as `label`.  IMO `field` makes more sense both here and in JSON, but if you have considered this possibility in the JSON context, and decided otherwise, and value consistency, perhaps we should go `label` here too.

- YAML is complicated and has some obscure corners and I wouldn't be at all surprised to find that I've missed some of them.

### DIFF
--- a/queries/yaml/highlights.scm
+++ b/queries/yaml/highlights.scm
@@ -1,5 +1,5 @@
-(block_mapping_pair key: (flow_node) @field)
-(flow_pair key: (flow_node) @field) ; This one doesn't seem to work, no idea what's wrong with it.
+(block_mapping_pair key: (_) @field)
+(flow_mapping (_ key: (_) @field))
 (boolean_scalar) @boolean
 (null_scalar) @constant.builtin
 (double_quote_scalar) @string
--- a/queries/yaml/highlights.scm
+++ b/queries/yaml/highlights.scm
@@ -15,7 +15,6 @@
 (yaml_directive) @keyword
 (ERROR) @error
 [
-; "+" ; including this causes "invalid node type" errors?!?
 ","
 "-"
 ":"
--- a/queries/yaml/highlights.scm
+++ b/queries/yaml/highlights.scm
@@ -4,8 +4,7 @@
 (null_scalar) @constant.builtin
 (double_quote_scalar) @string
 (single_quote_scalar) @string
-(double_quote_scalar (escape_sequence)) @string.escape
-(single_quote_scalar (escape_sequence)) @string.escape
+(escape_sequence) @string.escape
 (integer_scalar) @number
 (float_scalar) @number
 (comment) @comment
--- a/queries/yaml/highlights.scm
+++ b/queries/yaml/highlights.scm
@@ -1,0 +1,31 @@
+(block_mapping_pair key: (flow_node) @field)
+(flow_pair key: (flow_node) @field) ; This one doesn't seem to work, no idea what's wrong with it.
+(boolean_scalar) @boolean
+(null_scalar) @constant.builtin
+(double_quote_scalar) @string
+(single_quote_scalar) @string
+(double_quote_scalar (escape_sequence)) @string.escape
+(single_quote_scalar (escape_sequence)) @string.escape
+(integer_scalar) @number
+(float_scalar) @number
+(comment) @comment
+(anchor) @type
+(alias) @type
+(tag) @type
+(yaml_directive) @keyword
+(ERROR) @error
+[
+; "+" ; including this causes "invalid node type" errors?!?
+","
+"-"
+":"
+">"
+"?"
+"|"
+] @punctuation.delimiter
+[
+"["
+"]"
+"{"
+"}"
+] @punctuation.bracket


queries/yaml/highlights.scm
```diff
@@ -0,0 +1,31 @@
...
+(integer_scalar) @number
+(float_scalar) @number
+(comment) @comment
+(anchor) @type
+(alias) @type
+(tag) @type
+(yaml_directive) @keyword
+(ERROR) @error
+[
+; "+" ; including this causes "invalid node type" errors?!?
```
theHamsta, line 18: It's not in that list: https://github.com/ikatyang/tree-sitter-yaml/blob/master/grammar.js#L511-L530 Btw. what is + supposed to do in yaml?

dimbleby: https://yaml.org/spec/1.2/spec.html#id2794534

> Keeping is specified by the “+” chomping indicator.

dimbleby: but I hadn't realized that literal characters needed to correspond to things described in the grammar, so I suppose I should just take this out for now

theHamsta: You can experiment with this PR: https://github.com/nvim-treesitter/playground/pull/13

queries/yaml/highlights.scm
```diff
@@ -0,0 +1,30 @@
+(block_mapping_pair key: (flow_node) @field)
+(flow_pair key: (flow_node) @field) ; This one doesn't seem to work, no idea what's wrong with it.
+(boolean_scalar) @boolean
+(null_scalar) @constant.builtin
+(double_quote_scalar) @string
+(single_quote_scalar) @string
+(double_quote_scalar (escape_sequence)) @string.escape
+(single_quote_scalar (escape_sequence)) @string.escape
```
vigoux, line 8:
```suggestion
(double_quote_scalar (escape_sequence) @string.escape) 
(single_quote_scalar (escape_sequence) @string.escape)
```

Or even 
```suggestion
(escape_sequence) @string.escape
```

dimbleby: ha, yes, the final simplification certainly looks easier.

I don't understand what's going on in the variation where you move the bracket, I guess it doesn't matter if we go to the simpler one anyway, but if you have time to explain I'd appreciate it.

vigoux: The `@capture` will always capture the node _at the direct left_ of the `@` symbole, and because of that, what you'd actually capture here is "`double_quote_scalar` with an `escape_sequence` child in them

queries/yaml/highlights.scm
theHamsta: You can also watch tree-sitter-yaml to support "+" as soon as they support it.

queries/yaml/highlights.scm
```diff
@@ -0,0 +1,29 @@
+(block_mapping_pair key: (flow_node) @field)
+(flow_pair key: (flow_node) @field) ; This one doesn't seem to work, no idea what's wrong with it.
```
vigoux, line 2: Did you tried looking in the playground to see if it captures what you want it to capture ?

dimbleby: Yes, that's what I tried to do (that's how I came up with all of these!).

dimbleby: specifically the playground tells me that I have something like this:
```
                      flow_mapping [34, 9] - [34, 22]
                        flow_pair [34, 11] - [34, 20]
                          key: flow_node [34, 11] - [34, 15]
                            plain_scalar [34, 11] - [34, 15]
                              string_scalar [34, 11] - [34, 15]
                          value: flow_node [34, 17] - [34, 20]
                            plain_scalar [34, 17] - [34, 20]
                              string_scalar [34, 17] - [34, 20]
```

vigoux: Could you test your query in the query tester of the playground too ?

dimbleby: I'm not sure what this means - and I'm about to stop for the day.

If you can leave instructions I might get to it later in the week: or it should be super-easy to try yourself, just a yaml document containing something like `mapping: { foo: bar }` where we're trying to capture the foo.

vigoux: When using the playgroung, you can hit `o` to open a query debugger, that might be a good way to see why this query does not match anything.

theHamsta: This could be a bug of the parser? It will never match that specific pattern and  when you use (_ key: flow_node) nvim freezes sometimes.

From the maker of tree-sitter-markdown. The complete parser is written in C.

dimbleby: I don't know what the query debugger is supposed to do, it doesn't obviously do anything at all for me.

(Except, if I close the playground window containing the parse tree while the debugger is open, I get a segfault).

However, I found that I could get the highlighting to work by wildcarding:

```
(block_mapping_pair key: (_) @field)
(flow_mapping (_ key: (_) @field))
```

Here the `(_)` are just intended as an improvement / future-proofing on what I already had: let's treat any key as a `@field`.

But the first wilcard in the flow_mapping line is because I just don't know what to write there to make it work.  It seems to me that it should be `flow_pair`, but apparently it isn't.  Ho hum.

theHamsta: I'm pretty sure that the parser causes the segfault or a incompatibility of the tree-sitter runtimes. When the maintainer creates his parser in C in can of course segfault.

theHamsta (approved):

queries/yaml/highlights.scm
theHamsta: [video.webm.zip](https://github.com/nvim-treesitter/nvim-treesitter/files/5717304/video.webm.zip)

Should actually look like this? (I thought Github annouced video support? Could not upload the webm)

PR #771, opened 2020-12-16
fix(install): account for NIL executables

steelsojka:
This occured when trying to add a parser  with a plugin. On my system at least, this list contained a `vim.NIL` which caused an error during install.


theHamsta:
I already fixed this the same way as you.

steelsojka:
Nice! I'll close this.

PR #772, opened 2020-12-16
Create indents.scm for python

esn89:
Create indents.scm for python

### DIFF
--- a/queries/python/indents.scm
+++ b/queries/python/indents.scm
@@ -11,4 +11,6 @@
 [
   "elif"
   "else"
+  "finally"
+  "except"
 ] @branch


queries/python/indents.scm
```diff
@@ -0,0 +1,14 @@
...
+  (try_statement)
+  (for_statement)
+  (for_in_clause)
+  (class_definition)
+] @indent
+
+[
+  "elif"
+  "else"
+] @branch
```
kyazdani42, line 14: could there be things like finally clause here ? or catch ?

esn89: I have added finally and "except", since python doesn't have catch

theHamsta (approved):
Looks good to me. I can not really judge this since the feature is not working on my machine.

esn89:
@theHamsta @kyazdani42 

Thanks.  I just updated the plugin via vim-plug and Notice:

```
Error detected while processing /Users/esn89/.config/nvim/plugged/nvim-treesitter/plugin/nvim-treesitter.vim:                                                                                                                             
line   12:                                                                                                                                                                                                                               
E5108: Error executing lua ...nvim-treesitter/lua/nvim-treesitter/query_predicates.lua:89: attempt to call field 'add_directive' (a nil value)                                                                                           
Press ENTER or type command to continue                                                                                        
```

Was wondering, if this is due to my latest change or somethign else

doubleloop:
Anyone tested this? Should it work? Just recompiled master vim, updated plugin and parsers then started some python programming and immediately had to disable treesitter indent in config because it is difficult to write code with this (nearly every indentation is either too big or there is no indentation at all, or typing `:` makes properly indented line more indented..)
Is it known issue or did I screw up configuration? All this indenting issues occur with minimal configuration:

```
call plug#begin('~/.local/share/nvim/plugged')
Plug 'nvim-treesitter/nvim-treesitter'
call plug#end()

lua << EOF
    highlight = {
        enable = true
    },
    indent = {
        enable = true,
    },
}
EOF
```

theHamsta:
Indentation support is only alpha right now.

PR #775, opened 2020-12-18
Ruby constants highlighted as Type

kassio:
## Problem

Currently nvim-treesitter is highlighting ruby constants and types differently, but in ruby they are basically the same thing

## Solution

Based on [vim-ruby highlight](https://github.com/vim-ruby/vim-ruby/blob/master/syntax/ruby.vim#L502), highlight most ruby constants as types.

| before | after |
| --- | --- |
| _private_tmp_a_rb | _private_tmp_a_rb |

#### vim-ruby (for reference)

_private_tmp_a_rb

### DIFF
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -45,7 +45,7 @@
  "while"
  ] @repeat
 
-(constant) @constant
+(constant) @type
 
 ((identifier) @keyword
  (#vim-match? @keyword "^(private|protected|public)$"))
@@ -82,12 +82,12 @@
 
 (method name: [
                (identifier) @function
-               (constant) @constant
+               (constant) @type
                ])
 
 (singleton_method name: [
                          (identifier) @function
-                         (constant) @constant
+                         (constant) @type
                          ])
 
 (class name: (constant) @type)
@@ -103,8 +103,8 @@
 ((identifier) @constant.builtin
  (#vim-match? @constant.builtin "^__(callee|dir|id|method|send|ENCODING|FILE|LINE)__$"))
 
-((constant) @constant.macro
- (#vim-match? @constant.macro "^[A-Z\\d_]+$"))
+((constant) @type
+ (#vim-match? @type "^[A-Z\\d_]+$"))
 
 [
  (self)


queries/ruby/highlights.scm
```diff
@@ -45,7 +45,7 @@
  "while"
  ] @repeat
 
-(constant) @constant
+(constant) @type
```
vigoux, line 48: I think this should stay as `@constant` ?

kassio: to be honest I'm still learning this query syntax, so I'm not sure. I applied the suggested change and it didn't changed the highlight, so I'm ok with it:

_HOME_src_ruby-highlight_rb

vigoux: Do you have a constant defined here (I am not a Ruby expert) ?

kassio: Oh, thanks for asking, your question made me noticed that `::Some::High::Name::Space` is not being highlighted as type/constant. 😬 

In ruby anything all upcased is a constant, but types (stuff that starts with an Upcase) is usually also highlighted as a constant.

vigoux: I am not sure what you mean, so I guess I'd ask the opinion of @TravonteD :wink:

theHamsta: If found this also strange when I updated the queries. Before the query was just never activated due to a syntax error.

tree-sitter-ruby unified `call` and `method_call`

kassio: @theHamsta, sorry, it's not clear to me. Should I make any change on these queries?

theHamsta: No worries. It just mean that there is only `call` now. `method_call` no longer exists.

WhyNotHugo:
image

There's something off here, in that one class name is highlighted as one thing, but another is highlighted different.

The tree looks sane though, I'm not sure why the highlight is misbehaving.

theHamsta:
```scheme
;; Assume all-cap names are constants.
((identifier) @constant
 (.match? @constant "^[A-Z_][A-Z_\\d]*$"))
```
could probably be fixed by
```scheme
;; Assume all-cap names are constants.
((identifier) @constant
 (.match? @constant "^[A-Z_][A-Z_\\d]+$"))
```

WhyNotHugo:
I believe so. I guess things like `C` are kinda ambiguous, since it could either be a class name or constant names.

OTOH, it's very poor variable naming either way :shrug:

TravonteD (approved):
I'm good with this change. Given the way type is used it looks fine.

kassio:
@theHamsta is there anything missing to get this one merged?

theHamsta (approved):

PR #776, opened 2020-12-18
fix(query): add noop handler for make-range

steelsojka:


vigoux (approved):

kassio:
@vigoux this seems to be breaking in my machine, reverting remove the error:

```
Error detected while processing /Users/kassioborges/.local/share/nvim/site/pack/packer/start/nvim-treesitter/plugin/nvim-treesitter.vim:
line   12:
E5108: Error executing lua ...nvim-treesitter/lua/nvim-treesitter/query_predicates.lua:89: attempt to call field 'add_directive' (a nil value)
Error detected while processing /Users/kassioborges/.local/share/nvim/site/pack/packer/start/playground/plugin/nvim-treesitter-playground.vim:
line    3:
E5108: Error executing lua ...cker/start/playground/lua/nvim-treesitter-playground.lua:9: loop or previous error loading module 'nvim-treesitter'
```

steelsojka:
@kassio You'll need to update your neovim to the latest nightly version.

vkaracic:
@steelsojka I've updated neovim to the latest master version (`NVIM v0.5.0-dev+965-gd0668b36a`) and I still get the `query_predicates.lua:89: attempt to call field 'add_directive' (a nil value)` error. Is there something else that I need to do here?

theHamsta:
We had once the case that `make` was only doing build without updating the runtime queries. Could you grep your neovim source directory and your installed runtime for `add_directive`? The last user that reported a similar problem could solve it using `make clean` && `make`

vkaracic:
The [`src`](https://github.com/neovim/neovim/tree/d0668b36a3e2d0683059baead45bea27e2358e9c/src) directory does not mention `add_directive`, and in the vim runtime it is only mentioned in the `nvim-treesitter/lua/nvim-treesitter/query_predicates.lua:89` line. I've tried to recompile it with `make clean` (I usually compile with `make distclean; make`) but still the same error happens.

vkaracic:
I think I've figured it out. `/usr/local/share/nvim/runtime` (the fall-back for $VIM) never gets updated when compiling from source, when I copied over the `neovim/runtime` directory from the repository to that fall-back directory I don't get the error anymore.
Seems weird, shouldn't that be updated if it's obviously a dependency?

PR #778, opened 2020-12-18
Fix ruby include highlights

kassio:
Fix `require`, `require_relative`, `load` and `include`, `extend`, `preprend` to the includes group.

| before | after | vim-ruby for reference |
| --- | --- | --- |
| _HOME_src_ruby-highlight_rb | _HOME_src_ruby-highlight_rb | _HOME_src_ruby-highlight_rb |

### DIFF
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -72,8 +72,12 @@
 
 (program
  (call
-  (identifier) @include)
- (#vim-match? @include "^(require|require_relative|load)$"))
+   (identifier) @include)
+   (#vim-match? @include "^(require|require_relative|load)$"))
+
+ (call
+   (identifier) @include
+   (#vim-match? @include "^(include|extend|prepend)$"))
 
 ; Function definitions
 


queries/ruby/highlights.scm
```diff
@@ -70,11 +73,6 @@
             ] @function
    )
 
-(program
- (call
-  (identifier) @include)
- (#vim-match? @include "^(require|require_relative|load)$"))
-
```
theHamsta, line 77: There was a discussion whether not-top-level requires should be include: https://github.com/nvim-treesitter/nvim-treesitter/issues/463

@bluz71

kassio: @theHamsta, Thanks for pointing me to that. I changed to be a `method_call`, it's looking like that in my machine?

_HOME_src_ruby-highlight_rb

theHamsta: image

No, it's now all a `call`. `method_call` was removed. Sorry if I was confusing you

kassio: Ohh, gotcha now. Code updated. Thanks!

kassio: But unfortunately the updated code, using `call` instead of `method_call` does not produce the expected results:
_HOME_src_ruby-highlight_rb

theHamsta: Did you `:TSUpdate`? You should on every update of the plugin.

theHamsta: image

I have those three highlights with the old queries.

kassio: @theHamsta you're right! `TSUpdate` did the trick. Sorry for the noise.

By the way, how do you show this hover with the highlight names?

kassio:
At the end, I think the current query is the best one for now. Sorry for the noise.

vigoux:
Are you sure you are happy with it? We are fine working on that with you!

kassio:
@vigoux, I'm happy enough.

I wasn't able to define a query to get only the inlcude macros (`include`, `prepend`, `extend`) like vim-ruby does:

_HOME_src_ruby-highlight_rb


But to be honest I'm not sure we should as well. 😬

PR #782, opened 2020-12-19
fixes to bash highlighting

dimbleby:
Various fixes to the bash highlighting:

- top and tail matches with `^` and `$` so that they don't match more than intended
- switch to lua matches - are these faster? - not sure that this matters
- move capture of number parameters to a separate block: not that I know what I'm doing, but this seems to work where the original did not
- remove a redundant capture (which I suspect may have been compensating for the non-workingness per the other change)

Before:

before

- `3` is not recognized as a number
- all of `bar=false` is treated as a boolean
- only `baz="baz"` is highlighted as expected

After:

after

### DIFF
--- a/queries/bash/highlights.scm
+++ b/queries/bash/highlights.scm
@@ -29,6 +29,7 @@
  "|"
  "||"
  "="
+ "=~"
  "=="
  "!="
  ] @operator
@@ -117,3 +118,5 @@
 
 (case_item
   value: (word) @parameter)
+
+(regex) @string.regex
--- a/queries/bash/highlights.scm
+++ b/queries/bash/highlights.scm
@@ -70,13 +70,13 @@
 (special_variable_name) @constant
 
 ((word) @constant.builtin
- (#vim-match? @constant.builtin "SIG(INT|TERM|QUIT|TIN|TOU|STP|HUP)"))
+ (#match? @constant.builtin "^SIG(INT|TERM|QUIT|TIN|TOU|STP|HUP)$"))
 
 ((word) @boolean
-  (#vim-match? @boolean "true|false"))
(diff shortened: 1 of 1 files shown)


dimbleby:
also added captures for `=~` and `(regex)`, which is what you get on the right hand side of `=~`.

queries/bash/highlights.scm
```diff
@@ -116,4 +119,4 @@
 (case_item
   value: (word) @parameter)
 
-(concatenation (word) @parameter)
```
theHamsta, line 119: @TravonteD

theHamsta (approved):

queries/bash/highlights.scm
```diff
@@ -96,11 +97,13 @@
...
              (word) @parameter
-             ((word) @number
-              (#vim-match? @number "^\d*$"))
              (concatenation (word) @parameter)
              ])
 
+(command
+  argument: ((word) @number
+             (#match? @number "^[0-9]+$")))
+
```
vigoux, line 106: Can't you "just" match the `word` here? I don't feel the `command` node is useful.

dimbleby: There's already a query to match a number as just word, up at line 80.  But if we remove this extra query, then parameters that are also numbers get caught as parameters.

What's the precedence rule?  Last query wins?  It looks as though moving the number query to be after the parameter query works: is it safe to rely on that?  Maybe a bit more fragile to future refactoring?

vigoux: You can safely rely on "last capture wins" as this is the precedence rule we will be going for (for now).

vigoux (approved):

PR #783, opened 2020-12-20
Remove TSStructure

delphinus:
These definitions exist in highlight.lua, but does not exist here.


theHamsta:
For TSError and TSIndentifier this is on purpose (no highlight, except if set by user).

`Structure` is not used anymore you can delete it from the lua file and the documentation (it's only in the vim doc)?

delphinus:
I see. I reverted that and removed code about `TSStructure`.

theHamsta:
Thanks a lot! And sorry for the confusion.

theHamsta (approved):

PR #795, opened 2020-12-28
Python regex injection

theHamsta:
This add regex injections for Python expression of type `re.regex_function(r"regex-string")`

Additionally for regex highlighting:

- Give unescaped regex characters normal highlighting. In a normal string escape means special meaning. In regex escaping a special characters means it removes the special meaning and should thus be highlighted with normal regex highlights. This gives the escaped chars a different highlight as normal special chars.
- Give class_character constant highlighting to have a some colors in regexes for different kinds of chars
- add "?" operator

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,6 +114,7 @@ are optional and will not have any effect for now.
 @constant.macro
 @string
 @string.regex
+@string.regex-classchar
 @string.escape
 @string.special
 @character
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -114,6 +114,7 @@ are optional and will not have any effect for now.
 @constant.macro
 @string
 @string.regex
+@string.regex-classchar
```
vigoux, line 117: Is this needed ?

theHamsta: removed.

vigoux (approved):

PR #813, opened 2021-01-03
Do not use builtin tar on Windows

AdnoC:
Several parsers ([typescript](https://github.com/tree-sitter/tree-sitter-typescript/blob/master/tsx/corpus/common) and [ocaml](https://github.com/tree-sitter/tree-sitter-ocaml/blob/master/interface/corpus/module-items.txt) specifically) contain symlinks in their repo. The tar that comes with Windows cannot handle them.

So, on Windows just use git to download parsers.

Once I got this settled everything else worked fine when compiling with clang.

### DIFF
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -109,10 +109,13 @@ function M.select_mv_cmd(from, to, cwd)
 end
 
 function M.select_download_commands(repo, project_name, cache_folder, revision)
-  local has_tar = vim.fn.executable('tar') == 1 and fn.hs('win32') ~= 1
+  local is_windows = fn.has('win32') == 1
+
+  revision = revision or repo.branch or "master"
+
+  local has_tar = vim.fn.executable('tar') == 1 and not is_windows
   if has_tar and vim.fn.executable('curl') == 1 and repo.url:find("github.com", 1, true) then
 
-    revision = revision or repo.branch or "master"
     local path_sep = utils.get_path_sep()
     return {
       M.select_install_rm_cmd(cache_folder, project_name..'-tmp'),
@@ -152,22 +155,46 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
       M.select_install_rm_cmd(cache_folder, project_name..'-tmp')
     }
   else
+    local git_folder
+    if is_windows then
+      git_folder = cache_folder ..'\\'.. project_name
+    else
+      git_folder = cache_folder ..'/'.. project_name
+    end
+
+    local clone_error = 'Error during download, please verify your internet connection'
+    if is_windows then
+      clone_error = clone_error .. ". If on Windows you may need to enable Developer mode"
+    end
+
     return {
       {
         cmd = 'git',
         info = 'Downloading...',
-        err = 'Error during download, please verify your internet connection',
+        err = clone_error,
         opts = {
           args = {
             'clone',
+            '-c', 'core.symlinks=true',
             '--single-branch',
             '--branch', repo.branch or 'master',
-            '--depth', '1',
             repo.url,
             project_name
           },
           cwd = cache_folder,
         },
+      },
+      {
+        cmd = 'git',
+        info = 'Checking out locked revision',
+        err = 'Error while checking out revision',
+        opts = {
+          args = {
+            '-c', 'core.symlinks=true',
+            'checkout', revision
+          },
+          cwd = git_folder
+        }
       }
     }
   end
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -109,7 +109,8 @@ function M.select_mv_cmd(from, to, cwd)
 end
 
 function M.select_download_commands(repo, project_name, cache_folder, revision)
-  if vim.fn.executable('tar') == 1 and vim.fn.executable('curl') == 1 and repo.url:find("github.com", 1, true) and not fn.hs('win32') then
+  local has_tar = vim.fn.executable('tar') == 1 and not fn.hs('win32')
+  if has_tar and vim.fn.executable('curl') == 1 and repo.url:find("github.com", 1, true) then
 
     revision = revision or repo.branch or "master"
     local path_sep = utils.get_path_sep()
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -109,7 +109,7 @@ function M.select_mv_cmd(from, to, cwd)
 end
 
 function M.select_download_commands(repo, project_name, cache_folder, revision)
-  if vim.fn.executable('tar') == 1 and vim.fn.executable('curl') == 1 and repo.url:find("github.com", 1, true) then
+  if vim.fn.executable('tar') == 1 and vim.fn.executable('curl') == 1 and repo.url:find("github.com", 1, true) and not fn.hs('win32') then
 
     revision = revision or repo.branch or "master"
     local path_sep = utils.get_path_sep()
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -155,13 +155,7 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
       M.select_install_rm_cmd(cache_folder, project_name..'-tmp')
     }
   else
-    local git_folder
-    if is_windows then
-      git_folder = cache_folder ..'\\'.. project_name
-    else
-      git_folder = cache_folder ..'/'.. project_name
-    end
-
+    local git_folder = utils.join_path(cache_folder, project_name)
     local clone_error = 'Error during download, please verify your internet connection'
     if is_windows then
       clone_error = clone_error .. ". If on Windows you may need to enable Developer mode"
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -109,7 +109,7 @@ function M.select_mv_cmd(from, to, cwd)
 end
 
 function M.select_download_commands(repo, project_name, cache_folder, revision)
-  local has_tar = vim.fn.executable('tar') == 1 and not fn.hs('win32')
+  local has_tar = vim.fn.executable('tar') == 1 and fn.hs('win32') ~= 1
   if has_tar and vim.fn.executable('curl') == 1 and repo.url:find("github.com", 1, true) then
 
     revision = revision or repo.branch or "master"


vigoux (approved):
Sounds reasonable, @theHamsta what's your opinion on this?

theHamsta:
Neither tree-sitter-typescript nor tree-sitter-ocaml contain symlinks. I would not know how Git could fix that if they had.

Out git solution does not have pinning and using the newest version of tree-sitter-typescript. If symlinks were the problem we could just change the compile commands of the affected parsers to use the actual file location.

Could you investigate what the actual problem is? It's true that we append the name of the subfolder to install tsx and typescript at the same time.

theHamsta (changes requested):
typo: should be `fn.has`

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -109,7 +109,8 @@ function M.select_mv_cmd(from, to, cwd)
 end
 
 function M.select_download_commands(repo, project_name, cache_folder, revision)
-  if vim.fn.executable('tar') == 1 and vim.fn.executable('curl') == 1 and repo.url:find("github.com", 1, true) then
+  local has_tar = vim.fn.executable('tar') == 1 and not fn.hs('win32')
```
theHamsta, line 112: *not will not work here. You have to check for `fn.hs('win32') ~= 1`

AdnoC:
The links in my original post were to the symlink files in the typescript and ocaml repos (once again, the ocaml one: https://github.com/tree-sitter/tree-sitter-ocaml/blob/master/interface/corpus/module-items.txt). The issue is getting the source code of the parser out of the tarball. 

When I try, I get this output:

```cmd
> tar -xvf .\tree-sitter-ocaml-0.16.0.tar.gz
x tree-sitter-ocaml-0.16.0/
x tree-sitter-ocaml-0.16.0/.gitattributes
x tree-sitter-ocaml-0.16.0/.gitignore
x tree-sitter-ocaml-0.16.0/.travis.yml
x tree-sitter-ocaml-0.16.0/README.md
x tree-sitter-ocaml-0.16.0/binding.gyp
x tree-sitter-ocaml-0.16.0/examples/
x tree-sitter-ocaml-0.16.0/examples/.gitkeep
x tree-sitter-ocaml-0.16.0/interface/
x tree-sitter-ocaml-0.16.0/interface/binding.gyp
x tree-sitter-ocaml-0.16.0/interface/corpus/
x tree-sitter-ocaml-0.16.0/interface/corpus/module-items.txt: Can't create '\\\\?\\C:\\Users\\Adam Cutler\\Downloads\\tmp\\tree-sitter-ocaml-0.16.0\\interface\\corpus\\module-items.txt'
x tree-sitter-ocaml-0.16.0/interface/corpus/module-specification.txt
x tree-sitter-ocaml-0.16.0/interface/grammar.js
x tree-sitter-ocaml-0.16.0/interface/index.js
x tree-sitter-ocaml-0.16.0/interface/src/
x tree-sitter-ocaml-0.16.0/interface/src/binding.cc
x tree-sitter-ocaml-0.16.0/interface/src/grammar.json
x tree-sitter-ocaml-0.16.0/interface/src/node-types.json
x tree-sitter-ocaml-0.16.0/interface/src/parser.c
x tree-sitter-ocaml-0.16.0/interface/src/scanner.cc
x tree-sitter-ocaml-0.16.0/interface/src/tree_sitter/
x tree-sitter-ocaml-0.16.0/interface/src/tree_sitter/parser.h
x tree-sitter-ocaml-0.16.0/ocaml/
x tree-sitter-ocaml-0.16.0/ocaml/binding.gyp
x tree-sitter-ocaml-0.16.0/ocaml/corpus/
x tree-sitter-ocaml-0.16.0/ocaml/corpus/attributes.txt
x tree-sitter-ocaml-0.16.0/ocaml/corpus/comments.txt
x tree-sitter-ocaml-0.16.0/ocaml/corpus/expressions.txt
x tree-sitter-ocaml-0.16.0/ocaml/corpus/module-expressions.txt
x tree-sitter-ocaml-0.16.0/ocaml/corpus/module-implementation.txt
x tree-sitter-ocaml-0.16.0/ocaml/corpus/module-items.txt
x tree-sitter-ocaml-0.16.0/ocaml/corpus/module-types.txt
x tree-sitter-ocaml-0.16.0/ocaml/corpus/patterns.txt
x tree-sitter-ocaml-0.16.0/ocaml/corpus/toplevel.txt
x tree-sitter-ocaml-0.16.0/ocaml/corpus/types.txt
x tree-sitter-ocaml-0.16.0/ocaml/grammar.js
x tree-sitter-ocaml-0.16.0/ocaml/index.js
x tree-sitter-ocaml-0.16.0/ocaml/src/
x tree-sitter-ocaml-0.16.0/ocaml/src/binding.cc
x tree-sitter-ocaml-0.16.0/ocaml/src/grammar.json
x tree-sitter-ocaml-0.16.0/ocaml/src/node-types.json
x tree-sitter-ocaml-0.16.0/ocaml/src/parser.c
x tree-sitter-ocaml-0.16.0/ocaml/src/scanner.cc
x tree-sitter-ocaml-0.16.0/ocaml/src/tree_sitter/
x tree-sitter-ocaml-0.16.0/ocaml/src/tree_sitter/parser.h
x tree-sitter-ocaml-0.16.0/package.json
x tree-sitter-ocaml-0.16.0/queries/
x tree-sitter-ocaml-0.16.0/queries/highlights.scm
x tree-sitter-ocaml-0.16.0/queries/locals.scm
x tree-sitter-ocaml-0.16.0/queries/tags.scm
x tree-sitter-ocaml-0.16.0/script/
x tree-sitter-ocaml-0.16.0/script/known_failures.txt
x tree-sitter-ocaml-0.16.0/script/parse-examples
x tree-sitter-ocaml-0.16.0/test/
x tree-sitter-ocaml-0.16.0/test/highlight/
x tree-sitter-ocaml-0.16.0/test/highlight/attributes.ml
x tree-sitter-ocaml-0.16.0/test/highlight/constants.ml
x tree-sitter-ocaml-0.16.0/test/highlight/functions.ml
x tree-sitter-ocaml-0.16.0/test/highlight/locals.ml
x tree-sitter-ocaml-0.16.0/test/highlight/modules.ml
x tree-sitter-ocaml-0.16.0/test/highlight/operators.ml
x tree-sitter-ocaml-0.16.0/test/highlight/types.ml
tar.exe: Error exit delayed from previous errors.
```

I was getting a similar error for the typescript parser. Also for a symlink file. I tried passing several flags to `tar` but it would not extract without errors.

Part of the cause is that in Windows, you need administrator permissions to create symlinks (though you can now allow non-elevated users in Windows 10 through settings (https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/)).

theHamsta:
Sorry, I checked only the files we are using for symlinks. Since switching to  Git would mean loosing the locking functionality, could you please check

- would the old solution with `unzip` solve the problem: https://github.com/nvim-treesitter/nvim-treesitter/pull/563 (unzip is Unix only but typically comes with Git)?
- does the tsx/typescript parser work with the current master when we use your current approach (I guess JS and TS are locked to an older version for compatibility reasons)?
- would it be possible to do a `git pull --unshallow && git checkout ${revision}` to ensure the locked version is installed (requires full checkout on Windows)?

AdnoC:
Hmm. I tested ocaml by creating a buffer containing the code from the ocaml page of "learn x in y minutes". Didn't test typescript, which is an issue because I'm getting an error on this line https://github.com/neovim/neovim/blob/9223d1450d391b06cd59f8469c53291281899c9f/runtime/lua/vim/treesitter/query.lua#L258. 

After some investigation: it appears that git does not resolve symlinks either (by default). It just creates a file containing the path to the file it is linking as text. It doesn't error, but I assume it will cause something to misbehave.

If you enable symlinks in Windows 10 (by turning on "Developer mode" as described in https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/) AND enable symlinks when cloning (via the configuration option `core.symlinks=true`) git will properly create symlinks. If you set `core.symlinks=true` but do NOT enable "Developer mode" you will get the following output and error:

```cmd
> git clone -c core.symlinks=true --single-branch --branch master --depth 1 "https://github.com/tree-sitter/tree-sitter-ocaml" ocaml
Cloning into 'ocaml'...
remote: Enumerating objects: 61, done.
remote: Counting objects: 100% (61/61), done.
remote: Compressing objects: 100% (49/49), done.
remote: Total 61 (delta 11), reused 32 (delta 6), pack-reused 0
Unpacking objects: 100% (61/61), done.
error: unable to create symlink interface/corpus/module-items.txt: Permission denied
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'
```

Testing on a Mac, git resolves symlinks regardless of the value of `core.symlinks`. So setting the value is safe regardless of OS.

I don't know if you have to set `core.symlinks` when you checkout. I assume you do if you change the symlink file though.

I tried the port of `unzip` for Windows (obtained from http://gnuwin32.sourceforge.net/packages/unzip.htm). It unarchives the `.zip` downloaded from the releases page, but does not resolve symlinks. it just creates a text file containing the path to the linked file.


My proposed solution:
* `git clone` the full repo (no `--depth 1` but yes `--single-branch`), passing `-c core.symlinks=true`
* `git checkout <REVISION>`
* If on Windows, adjust the `git clone` error message to mention symlinks and/or developer mode

I've applied my changes and both re-installed all the parsers several times and tested typescript and ocaml. Seems to work.

theHamsta:
Ocaml is guaranteed to work. We are with all parsers on the newest revision except Typescript and Javascript. We hold that update back as it would cause that JS and TS can no longer share queries. On CI newest TS fails. Any progress on this @steelsojka ?

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -151,22 +155,46 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
...
     return {
       {
         cmd = 'git',
         info = 'Downloading...',
-        err = 'Error during download, please verify your internet connection',
+        err = clone_error,
         opts = {
           args = {
             'clone',
+            '-c', 'core.symlinks=true',
```
theHamsta, line 178: symlinks should not be needed.

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -151,22 +155,46 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
       M.select_install_rm_cmd(cache_folder, project_name..'-tmp')
     }
   else
+    local git_folder
+    if is_windows then
+      git_folder = cache_folder ..'\\'.. project_name
+    else
+      git_folder = cache_folder ..'/'.. project_name
```
theHamsta, line 162: there should be already a method giving you that folder.

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -151,22 +155,46 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
...
+    end
+
+    local clone_error = 'Error during download, please verify your internet connection'
+    if is_windows then
+      clone_error = clone_error .. ". If on Windows you may need to enable Developer mode"
```
theHamsta, line 167: we should not need any logic around the symlinks.

theHamsta:
Hmm, your solution should work. Thanks for the efforts. I guess Windows users can wait that extra time that a full checkout costs.

lua/nvim-treesitter/shell_command_selectors.lua
vigoux: Please use `utils.join_path` here.

vigoux (changes requested):

vigoux (dismissed):

lua/nvim-treesitter/shell_command_selectors.lua
vigoux: I think you should remove this an let people handle symlinks the way they want ?
This worked well without this change before I think, could you confirm that it is the case ?

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -151,22 +155,40 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
       M.select_install_rm_cmd(cache_folder, project_name..'-tmp')
     }
   else
+    local git_folder = utils.join_path(cache_folder, project_name)
+    local clone_error = 'Error during download, please verify your internet connection'
+    if is_windows then
+      clone_error = clone_error .. ". If on Windows you may need to enable Developer mode"
```
theHamsta, line 161: I would add a link or not mention it. As a user I would have no idea what developer mode is and how I can enable it.

theHamsta (approved):

lua/nvim-treesitter/shell_command_selectors.lua
vigoux: Yeah I agree on that point too, maybe you could even suggest to open an issue.

vigoux (commented):
Still not pleased by the symlink part, what is the rationale behind that ?

theHamsta:
Wait, 7-zip is bundled with nvim? We could use this to make both git and tar work https://github.com/neovim/neovim/issues/14078

koernepr:
`7za.exe` is contained in `win32tools.zip` and also listed in the respective comment in `CMakeLists.txt`, but the binary does not seem to be included in the release .zips (I don't know why this is the case). I've improved the issue description to point this out more prominently; sorry for the confusion.

theHamsta:
@AdnoC do you want to finish this PR? We also have a Windows CI now.

theHamsta:
Closed via https://github.com/nvim-treesitter/nvim-treesitter/issues/1052#issuecomment-820869981

PR #819, opened 2021-01-04
feat: added more fold queries

elianiva:
Added fold queries for several simple languages. I didn't add HTML because I find it quite inconsistence. Here's an example.

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>Something</title>
  </head>
  <body>
    <div>  <- cursor on this line
      <span>somethin</span>
    </div>
  </body>
</html>
```

If you try to fold the `div`, it will fold up until the last `</html>`.  But this won't happend if you expand `<span>` tag to this.

```html
<span>
  somethin
</span>
```

There might be some stuff that hasn't been covered yet so any advice would be appreciated :)

### DIFF
--- a/queries/css/folds.scm
+++ b/queries/css/folds.scm
@@ -1,0 +1,3 @@
+[
+  (rule_set)
+] @fold
--- a/queries/json/folds.scm
+++ b/queries/json/folds.scm
@@ -1,0 +1,4 @@
+[
+  (object)
+  (array)
+] @fold
--- a/queries/toml/folds.scm
+++ b/queries/toml/folds.scm
@@ -1,0 +1,4 @@
+[
+  (table)
+  (array)
+] @fold
--- a/queries/yaml/folds.scm
+++ b/queries/yaml/folds.scm
@@ -1,0 +1,3 @@
+[
+  (block_mapping_pair)
+] @fold


queries/css/folds.scm
```diff
@@ -0,0 +1,3 @@
+[
+  (rule_set)
```
elianiva, line 2: I'm a bit confused if we should use `block` or `rule_set`. I personally prefer it to be folded based on `rule_set` but I don't know if anyone other than me prefer it this way

queries/json/folds.scm
```diff
@@ -0,0 +1,3 @@
+[
+  (object)
```
vigoux, line 2: Are there nodes for `tables` too ?

queries/css/folds.scm
vigoux: As you are the first one adding this, we'll go for your version, and enhance afterwards.

queries/json/folds.scm
elianiva: I think I missed the array, will check it out in a minute

elianiva: ok I've added the `array` and I couldn't find any `tables` node.

vigoux (approved):

Issue #738, opened 2020-12-04
Select highlight group base on longest matching theme key


alemuller:
Nvim-tree-sitter could have the same behavior on matching theme key as Tree-Sitter.

Excerpt from Tree Sitter 's [Syntax Highlighting](https://tree-sitter.github.io/tree-sitter/syntax-highlighting) documentation:

> #### Highlight Names
> 
> A theme can contain multiple keys that share a common subsequence. Examples:
> 
>  * `variable` and `variable.parameter`
>  * `function`, `function.builtin`, and `function.method`
> 
> For a given highlight produced, styling will be determined based on the **longest matching theme key**. For example, the highlight `function.builtin.static` would match the key `function.builtin` rather than function.

This feature would make easier for the end users to customize their highlighting without the need to add or change rules on `highlights.scm`. At same time, matching the longest valid theme key, would provide a sane default (instead of failing to apply any highlight).

vigoux:
Oh that's interesting... This would also allow to directly use parsers highlight queries, which is great.

alemuller:
I found an workaround that might be useful for others struggling with the same thing.

It seems nvim-ts allows to set more than one theme key per rule. For example:

```scm
(entity_declaration
   (identifier) @constant @constant.entity)
```

This will identifier with  `@constant` and `@constant.entity` in that order.

PR #820, opened 2021-01-05
Add VHDL parser

alemuller:


queries/vhdl/highlights.scm
```diff
@@ -0,0 +1,2397 @@
...
+      at_end: [(simple_name) (extended_simple_name)] @_t @error @error.repeated.name .))
+   (#not-eq? @_h @_t))
+; }}}
+; 12.4 Use clauses {{{
+(selected_name
+   prefix: (selected_name
+             prefix: ((simple_name) @_p
+                      (#vim-match? @_p "^[iI][eE]{3}$"))
+             suffix: ((simple_name) @_s @error @error.illegal.package.ieee
+                      (#not-vim-match? @_s "^\%([nN][uU][mM][eE][rR][iI][cC]_\%([bB][iI][tT]|[sS][tT][dD]\)\%(_\%([uU][nN]\)?[sS][iI][gG][nN][eE][dD]\)?|\%([fF][iI][xX][eE][dD]|[fF][lL][oO][aA][tT]\)\%(_[gG][eE][nN][eE][rR][iI][cC]\)?_[pP][kK][gG]|[fF][iI][xX][eE][dD]_[fF][lL][oO][aA][tT]_[tT][yY][pP][eE][sS]|[sS][tT][dD]_[lL][oO][gG][iI][cC]_\%(1164|[tT][eE][xX][tT][iI][oO]|[aA][rR][iI][tT][hH]|[mM][iI][sS][cC]|\%([uU][nN]\)?[sS][iI][gG][nN][eE][dD]\)|[mM][aA][tT][hH]_\%([rR][eE][aA][lL]|[cC][oO][mM][pP][lL][eE][xX]\)\)$"))
```
theHamsta, line 2293: wouldn't setting the case-insensitve flag `\C` or having  dedicated predicate make in more readable (can't tell whether this will work with all the crazy escaping happening in nvim)

queries/vhdl/highlights.scm
```diff
@@ -0,0 +1,2397 @@
...
+
+(ERROR) @error @error.parser
+
+[
+   (identifier)
+   (extended_identifier)
+   (PSL_Identifier)
+   (simple_name)
+   (extended_simple_name)
+] @variable
```
theHamsta, line 2397: :scream:

theHamsta: must have been a huge effort to port the whole standard. impressive!

vigoux (commented):
This is huuuuuuuge, I fear that the performance might a bit off, did you encountered some performance issues ?

queries/vhdl/highlights.scm
alemuller: I tried using the case insensitive flag but it didn't work. Now, after doing some investigation, I think there is a bug on [neovim/runtime/lua/vim/treesitter/query.lua:175](https://github.com/neovim/neovim/blob/885c2f7281c2f064961e10fc7e55110dff718dc9/runtime/lua/vim/treesitter/query.lua#L175).  The `magic_prefixes[string.sub(str,1,2)]` on line 167 can't be true. 

I used `\%(` instead of `(` because some regex expressions on this query (1) exceeds the hardcoded limit of captures; (2)  don't need the capture -- and `\%(` should be faster.

By the way, if you think that expressions is bad have a look at line 1531.

I will try do some more investigation on the `#match` to see if is actually a bug and then update the regex expressions on the query to increase legibility.

queries/vhdl/highlights.scm
alemuller: Yeah. Writing the grammar and the query took way longer than I'd admit. But I'm happy of the result. The parser with the highlight query can highlight common syntax errors very precisely.

The grammar is still missing a couple of things from the standard, such PSL OBE Expression and protected tool directives (used for encryption). I intend to implement the protected directives, but not the OBE expressions -- their syntax are not suitable for tree-sitter format, and isn't supported by many tools anyway.

There is a few other uncommon syntax errors that I intended to add to query.

Once all the standard queries (locals/tags/fold/indent) are implement I intend to extend the grammar to the latest language revision.

alemuller:
> This is huuuuuuuge, I fear that the performance might a bit off, did you encountered some performance issues ?

The query takes about 900ms to load regardless of the size of the file. Once is loaded, it works flawlessly, even with massive files.

The plugin performance might be an issue thought, the ts rainbows takes about 2 seconds to update the buffer after every keystroke.

queries/vhdl/highlights.scm
alemuller: Actually, I think creating a new predicate might help writing clearer query and avoid the regex compile overhead.

The code above could look like this:

```
(selected_name
   prefix: (selected_name
             prefix: ((simple_name) @_p
                      (#foobar? @_p "ieee"))
             suffix: ((simple_name) @_s @error @error.illegal.package.ieee
                      (#foobar? @_s "fixed_float_types"
                                    "fixed_pkg"
                                    "float_pkg"
                                    "fixed_generic_pkg"
                                    "float_generic_pkg"
                                    "math_complex"
                                    "math_real"
                                    "numeric_bit"
                                    "numeric_std"
                                    "numeric_bit_unsigned"
                                    "numeric_std_unsigned"
                                    "std_logic_1164"
                                    "std_logic_textio"
                                    "std_logic_misc"
                                    "std_logic_arith"
                                    "std_logic_signed"
                                    "std_logic_unsigned"))) @error.region)
```

Any idea for a name to such predicate? Also, how to tell case sensitive and case insensitive apart?

vigoux:
CI I failing because of some unknown highlight captures, could you fix that ?

alemuller:
@vigoux I think this is an issue with the CI. Custom capture are documented on [Readme](https://github.com/nvim-treesitter/nvim-treesitter#highlight) and work just fine.

I used custom captures to enable users to customize the syntax highlighting without the having to write a query. VHDL objects have many attributes (kind, direction, subtype, and others). With the custom captures the user can easily highlight variable with one color and a signal with another, or; highlight the input signal with one color and output signal with another and so on.

The custom captures follows the #738 naming scheme. They are also used to document errors and warning, for instance `@error.missing.parenthesis`). 

**NOTE**: Since #738 is not supported, this query also includes default fallback on the predefined captures.

---

A MIW for the CI failure:
```
; query
[
  "entity"
  "is"
  "begin"
  "end"
] @keyword

(entity_declaration  [
  "entity"
  "is"
  "begin"
  "end"
]  @keyword.unit.primary)

(foobar
   "entity" @keyword.foobar)
```

I tried defining both captures at the specification:

```
; query
(entity_declaration  [
  "entity"
  "is"
  "begin"
  "end"
]  @keyword @keyword.unit.primary)

(foobar
   "entity" @keyword @keyword.foobar)
```

But sometimes tree-sitter only matches the keyword without the context and fails to apply any highlight:

```
; parsed s-expression
(ERROR
    "entity")
```

So, the best approach seems to be defining all the keywords as @keyword and specialize them in the patterns.

theHamsta:
Custom captures are thought for user configuration. The captures for the languages should be the same in this repo. This is a self-imposed restriction.

alemuller:
@theHamsta The query includes both the predefined capture and custom capture. All nodes fallback to predefined capture.

Does it need to support *only* the predefined captures?

queries/vhdl/highlights.scm
theHamsta: I think it would be worth to try to fix the bug in the regex escaping.
```
((identifier) @foo
(#match? @foo "^\cres$"))
```

Maybe `any-of?` with a boolean parameter for case-sensitive or not?

```
((identifier) @foo
(#any-of? nil "foo" "bar" "baz"))

((identifier) @foo
(#any-of? t "foo" "bar" "baz"))

``

File: lua/nvim-treesitter/query_predicates.lua [hunk 3]
```diff
@@ -85,5 +89,32 @@ query.add_predicate('has-type?', function(match, pattern, bufnr, pred)
...
+      fun = noop
+  end
+
+  local ts_utils = require'nvim-treesitter.ts_utils'
+  local node = match[pred[2]]
+  local node_text = fun(ts_utils.get_node_text(node, bufnr)[1])
+
+  local dict = {}
+  for i=3,#pred do
+      dict[fun(pred[i])] = true
```
theHamsta, line 110: why do you need to create a dictionary when you can compare directly to match? Lua string comparisons are only pointer comparisons since all strings are interned.

alemuller: Ops. Originally I was trying to do something else and left this as mistake. I will fix it.

alemuller: I think `any-of` is a better name. Also, prefixing with `case-insensitive`  makes `not-case-insensitive-foobar` semantically ambiguous.

pidgeon777:
Hi, how could the VHDL treesitter be enabled in Neovim?

`:TSInstall vhdl` doesn't work, unfortunately.

vigoux:
> Hi, how could the VHDL treesitter be enabled in Neovim?

Hi, to enable the `VHDL` parser we would need to merge this PR, but for now it seems a bit dead.
I am not a `VHDL` expert so I can't tell, but if you want to take this PR over, we can help you in the process.

SethGower:
> Hi, how could the VHDL treesitter be enabled in Neovim?
> 
> `:TSInstall vhdl` doesn't work, unfortunately.

I am trying to add it as a custom parser. I was able to get it installed, but I am not sure what's wrong, since it doesn't seem to be providing highlighting (I ran `syntax off` and nothing was highlighted. This wasn't the case for other files, such as lua and c). 

```lua
parser_config.vhdl = {
  install_info = {
    url = "~/.local/share/tree-sitter/tree-sitter-vhdl",
    files = {"src/parser.c"}
  },
  filetype = "vhdl"
}
```
I cloned the `tree-sitter-vhdl` repo into `~/.local/share/tree-sitter/tree-sitter-vhdl` and followed the instructions in the [readme](https://github.com/nvim-treesitter/nvim-treesitter#adding-parsers). It installs correctly (doesn't fail and shows up in `:TSInstallInfo`. And I can manually parse a `vhdl` file using `tree-sitter parse <file>` on the terminal. So I am not sure what's wrong.

theHamsta:
@SethGower did you add any highlight queries? E.g. the huge file in this PR in case it is still up-to-date

SethGower:
> @SethGower did you add any highlight queries? E.g. the huge file in this PR in case it is still up-to-date

Ah, no I didn't. I am new to this all, so I didn't fully understand the requirements. That makes sense.

theHamsta:
The playground helps you to understand what tree-sitter is doing: https://github.com/nvim-treesitter/playground and how you can write queries. Queries basically select specific parts of the AST and assign a highlight.

SethGower:
@theHamsta thanks! I will take a look at that. I don't have a ton of time to do a bunch of stuff with this, but I am still trying to learn and contribute best I can.

pidgeon777:
Unfortunately, I do not have the necessary knowledge to add the VHDL parser to the repository files, but I wonder if it could be possible to add it to the list of parsers installable through `:TSInstall`. Until then, how could I effectively add it manually?

theHamsta:
The README documents how to add parsers. You can also just rebase this PR on current master. For highlighting queries you need to have the `vhdl` folder from this repo in queries folder of nvim-treesitter or any other runtime path

SethGower:
@pidgeon777 see my previous chunk to add it manually. That doesn't currently have highlighting for VHDL with TS, but it does have the parser itself (you can see it in action if you have [the Playground](https://github.com/nvim-treesitter/playground) installed, as @theHamsta stated earlier). Currently I am able to use the parser for showing me [context](https://github.com/romgrk/nvim-treesitter-context) in VHDL files which is nice. I will be looking into getting the highlighting working.

SethGower:
For clarification. This is in my [`init.lua`](https://github.com/SethGower/dotfiles/blob/master/neovim/.config/nvim/init.lua#L352):
```lua
parser_config.vhdl = {
  install_info = {
    url = "~/.local/share/tree-sitter/tree-sitter-vhdl",
    files = {"src/parser.c"}
  },
  filetype = "vhdl"
}
```
I cloned [`tree-sitter-vhdl`](https://github.com/alemuller/tree-sitter-vhdl) into `~/.local/share/tree-sitter/tree-sitter-vhdl` (if you put it somewhere else, change in the above config). Then simply reload `nvim` and run `TSInstall vhdl`

pidgeon777:
> For clarification. This is in my [`init.lua`](https://github.com/SethGower/dotfiles/blob/master/neovim/.config/nvim/init.lua#L352):
> 
> ```lua
> local parser_config = require "nvim-treesitter.parsers".get_parser_configs()
> parser_config.vhdl = {
>   install_info = {
>     url = "~/.local/share/tree-sitter/tree-sitter-vhdl",
>     files = {"src/parser.c"}
>   },
>   filetype = "vhdl"
> }
> ```
> 
> I cloned [`tree-sitter-vhdl`](https://github.com/alemuller/tree-sitter-vhdl) into `~/.local/share/tree-sitter/tree-sitter-vhdl` (if you put it somewhere else, change in the above config). Then simply reload `nvim` and run `TSInstall vhdl`

I tried your suggestions, and it seems to have installed it.

How could I check if it is actually working?

SethGower:
@pidgeon777 if you have the playground installed (see links in previous comments), you can run `:TSPlaygroundToggle` (or something similar, just do `:TSP<Tab>` when you have a `.vhd` file open and it will open a split on the left, and should show the tree.

PR #822, opened 2021-01-05
Update teal syntax for 0.10

euclidianAce:
accommodates the changes in syntax for teal introduced in 0.10


stsewd (commented):
You can update the commit in the lockfile so tests pass

https://github.com/nvim-treesitter/nvim-treesitter/blob/f2c219cddcaf6d7237bdc957c84ef4cdb9072a89/lockfile.json#L113-L115

vigoux (approved):
Awesome to see you are maintaining this !! Thanks !

PR #824, opened 2021-01-05
asynchronous buffer highlighting and attach (using libuv async handle)

oberblastmeister:
I have recently created a [pr](https://github.com/nvim-telescope/telescope.nvim/pull/386) for telescope.nvim that makes the `nvim_buf_attach` callback asynchronous using `uv_async_t`. It boosted performance a lot so I thought to try it with treesitter because it takes a little long to open a file compared to without it. Doing this boosted my startup performance when opening a file from ~400ms to ~300ms. Because it is asynchronous it also has the really cool effect of asynchronously highlighing the buffer similar to other editors like vscode.


vigoux:
This is simply sick

kyazdani42 (approved):
That's real cool man thanks

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -213,7 +213,7 @@ function M.memoize_by_buf_tick(fn)
       cache[bufnr] = {}
       api.nvim_buf_attach(bufnr, false,
         {
-          on_changedtick = function() cache[bufnr] = fn(bufnr) end,
```
vigoux, line 216: Not sure about this one, I think we want that to stay sync, since in some case we rely on the lazyness of this

oberblastmeister: I put that in there because it makes the typing latency much lower and makes everything generally feel smoother, especially on files where tree sitter is slower like rust and markdown.

oberblastmeister: Maybe I can remove that for now and put that in a separate pull request? The other async function seems fine and should improve performance a lot when opening a file. I have tested it on a rust file and it opened from ~700ms to ~300ms.

vigoux (approved):

rafikdraoui:
I don't really notice any speedup (~55-60ms loading `scripts/check-includes.py` file from neovim's repo, as measured by https://github.com/tweekmonster/startuptime.vim, for both before and after this change). 

But I do notice the delay between opening a buffer and having its text highlighted. I find it a bit jarring, similar to a [flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) on a web page.

I tried recording it with `asciinema`, but the neovim colors don't show up at all in the recording...

oberblastmeister:
I am not familiar with that plugin. Can you test with `vim <file> --startuptime vim.log` and then check the log?

bfredl:
> Because it is asynchronous it also has the really cool effect of asynchronously highlighing the buffer similar to other editors like vscode.

I don't think this is true in any meaningful way. The highest cost of highlighting is parsing which is still done synchronously each redraw after edit. Dynamically async parsing like atom does would be cool but will require a thread safe memline (or some COW hack)

rafikdraoui:
> I am not familiar with that plugin. Can you test with vim <file> --startuptime vim.log and then check the log?

Right, I think that plugin doesn't consider lua code (or bundle it with vimrc and only surfaces timings for plugins). When I use `--startuptime` directly, I do see a big speedup.

So the first screen update does happen faster, but it can kind of _feel_ slower because of the delayed syntax highlighting...

I don't think there's anything that can be done about this (other than hoping that tree-sitter parsing gets faster in the future), that's just the trade-off we have to make one way or another when choosing sync vs async. :shrug:

oberblastmeister:
@bfredl I should have used better wording. I only meant asynchronous as in when you open the buffer it allows you to open it first and then it will highlight because the only part that I changed is the autocmd that attaches to the buffer. The treesitter highlight should popup kind of like what @rafikdraoui said. You are right that the parsing that is done after each edit is still synchronous. Dynamically async parsing would be really cool but it would definately require more work than what I did.

vigoux:
We have issues starting to appear with this (namely #838) which may lead to a revert of this PR.
@oberblastmeister where do you think the issue may come from ?

tami5:
revert revert revert :(

PR #825, opened 2021-01-06
feat: added more indent queries

elianiva:
I've added indent queries for some languages but I might have not covered some cases, would be awesome if anyone would try it and give some advice :)

List of languages that I've added
- HTML
- JSON
- YAML
- TOML
- Javascript
- Typescript
- JSX
- TSX
- Go
- Ruby
- GraphQL
- Rust
- Teal
- C
- C++

Known issues: 
- Last `/>` on self closing HTML tag doesn't get outdented properly **(Fixed after #837)**
- Last `</>` on jsx fragment doesn't get outdented properly
- Last `}` on Rust `else_clause` doesn't get outdented properly **(Fixed)**
- Last `end` on Teal sometimes doesn't get outdented properly. e.g if you use anonymous function for an argument.
- Some block doesn't get indented properly when you press `o` on the first line of the block to add a new line. Hopefully these illustrations make it clearer.
```javascript
return(| <- if you press `o` here
| <- your cursor will go here
  | <- instead of here, notice the indentation 
)
```
another case is the Rust's else clause
```rust
} else {| <- if you press `o` here
| <- it will go here, no indent
  println!("something");| <- if you press `o` here
  | <- it will go here, the indentation is correct
}
```

note: this branch will only receives force push commits

### DIFF
--- a/queries/go/indents.scm
+++ b/queries/go/indents.scm
@@ -1,0 +1,17 @@
+[
+  (import_declaration)
+  (function_declaration)
+  (const_declaration)
+  (var_declaration)
+  (type_declaration)
+  (composite_literal)
+  (func_literal)
+  (block)
+] @indent
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+] @branch
--- a/queries/javascript/indents.scm
+++ b/queries/javascript/indents.scm
@@ -1,0 +1,26 @@
+; inherits: (jsx)
+
+[
+  (object)
+  (array)
+  (arguments)
+  (statement_block)
+  (object_pattern)
+  (class_body)
+  (method_definition)
+  (named_imports)
+  (binary_expression)
+  (return_statement)
+  (template_substitution)
+  (expression_statement (call_expression))
+  (export_clause)
+] @indent
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+  "["
+  "]"
+] @branch
--- a/queries/tsx/indents.scm
+++ b/queries/tsx/indents.scm
@@ -1,0 +1,1 @@
+; inherits: typescript,jsx
--- a/queries/typescript/indents.scm
+++ b/queries/typescript/indents.scm
@@ -1,0 +1,6 @@
+; inherits: javascript
+
+[
+  (interface_declaration)
+  (object_type)
+] @indent
--- a/queries/c/indents.scm
+++ b/queries/c/indents.scm
@@ -1,0 +1,16 @@
+[
+  (init_declarator)
+  (compound_statement)
+  (preproc_arg)
+  (field_declaration_list)
+  (case_statement)
+] @indent
+
+
+[
+  "#define"
+  "#ifdef"
+  "#endif"
+  "{"
+  "}"
+] @branch
--- a/queries/cpp/indents.scm
+++ b/queries/cpp/indents.scm
@@ -1,0 +1,16 @@
+[
+  (enumerator_list)
+  (struct_specifier)
+  (compound_statement)
+  (case_statement)
+  (condition_clause)
+  (conditional_expression)
+] @indent
+
+[
+  (statement_identifier)
+  "#ifdef"
+  "#endif"
+  "{"
+  "}"
+] @branch
--- a/queries/graphql/indents.scm
+++ b/queries/graphql/indents.scm
@@ -1,0 +1,9 @@
+[
+  (definition)
+  (selection)
+] @indent
+
+[
+  "{"
+  "}"
+] @branch
--- a/queries/html/indents.scm
+++ b/queries/html/indents.scm
@@ -1,0 +1,9 @@
+[
+  (element)
+] @indent
+
+[
+  (end_tag)
+  ">"
+  "/>"
+] @branch
--- a/queries/json/indents.scm
+++ b/queries/json/indents.scm
@@ -1,0 +1,9 @@
+[
+  (object)
+  (array)
+] @indent
+
+[
+  "}"
+  "]"
+] @branch
--- a/queries/jsx/indents.scm
+++ b/queries/jsx/indents.scm
@@ -1,0 +1,10 @@
+[
+  (jsx_fragment)
+  (jsx_element)
+  (jsx_self_closing_element)
+] @indent
+
+[
+  (jsx_closing_element)
+  ">"
+] @branch
--- a/queries/ruby/indents.scm
+++ b/queries/ruby/indents.scm
@@ -1,0 +1,12 @@
+[
+  (class)
+  (method)
+  (module)
+  (call)
+  (if)
+] @indent
+
+[
+  (elsif)
+  "end"
+] @branch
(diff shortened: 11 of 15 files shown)


vigoux (approved):
Terrific work again :wink:
I'll let @steelsojka review javascript ones.
But for `rust` is seems okay, appart for the `else` dedenting, maybe we can implement a `@dedent` capture to force this behavior somehow.

queries/javascript/indents.scm
```diff
@@ -0,0 +1,25 @@
+; inherits: (jsx)
```
yzia2000, line 1: jsx indents are not being imported in react js files. Works if you add indent and branch queries from jsx to javascript index.scm instead of inheritting. Might be due to some overriding

queries/javascript/indents.scm
```diff
@@ -0,0 +1,25 @@
...
+  (return_statement)
+] @indent
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+  "["
+  "]"
```
yzia2000, line 24: Indentation issue with else block:
```
if () {
} else {
    }
```
can be fixed by appending @ branch with statement_block

queries/tsx/indents.scm
```diff
@@ -0,0 +1 @@
+; inherits: typescript,jsx
```
yzia2000, line 1: jsx components are working perfectly here

queries/typescript/indents.scm
```diff
@@ -0,0 +1,6 @@
+; inherits: javascript
```
yzia2000, line 1: Indentation issue with else block:
```
if () {
} else {
    }
```

queries/go/indents.scm
```diff
@@ -0,0 +1,17 @@
...
+  (if_statement)
+  (composite_literal)
+  (func_literal)
+] @indent
+
+[
+  "("
+  ")"
+  "{"
+  "}"
```
yzia2000, line 16: Indentation issue with else block:
```
if () {
} else {
    }
```
can be fixed by appending @ branch with block

yzia2000 (changes requested):
jsx indents dont seem to be working when inheritted in js files. Also adding (statement_block) to javascript/indents and (block) to go/indents in @ branch fixes the indentation issue:
```
if () {
} else {
    }
```
Overall great work. Much needed for main branch.

yzia2000:
Hi I have found some issues after reviewing this branch with go and javascript. Thanks for the work. This pr is much needed for development.

queries/javascript/indents.scm
elianiva: hmm, it works on my end though. Let me double check it with a minimal init.vim

yzia2000: Sure

queries/javascript/indents.scm
elianiva: that does seem to work, but I found another issue with it. It gives the `else` block a double indentation like this.
```javascript
if (2 > 1) {
  console.log("A")
} else {
    console.log("c")
}
```

yzia2000:
> Terrific work again 
> I'll let @steelsojka review javascript ones.
> But for `rust` is seems okay, appart for the `else` dedenting, maybe we can implement a `@dedent` capture to force this behavior somehow.

Rust dedenting seems to be working for me. I think the issue arises because alternative in js and go files are blocks or statement_blocks instead of else_blocks. Seems to work if we append @ branch with (block) and (statement_block) in the respective languages.

queries/javascript/indents.scm
yzia2000: Really. It seems to work on my end. Let me debug again

yzia2000:
```
        if (!this.state.toggled) {
            this.props.toggleOn();
        } else {
            this.props.toggleOff();
        }
```
I am getting this kind of indentation. Let me use minimal nvim

queries/javascript/indents.scm
elianiva: it does seem to work using this minimal config https://paste.sh/OIGOYy-6#0Djftco4hcAZ2dgfN2BA0kTI, don't forget to set the `indentexpr` to `nvim_treesitter#indent()`. Not sure why it has to be done it manually.

Also, the last `</>` for jsx fragment seems to be broken, it doesn't get outdented properly. I thought by adding `"</>"` to `@branch` would fix it but it didn't.

queries/javascript/indents.scm
elianiva: yep, tried your query and it still doesn't work unfortunately.

queries/javascript/indents.scm
yzia2000: setting indentexpr made it work haha

queries/javascript/indents.scm
yzia2000: actually I realized that it works without setting indentxpr and doesnt when indentxpr is explicitly set

elianiva: I don't know why it doesn't set it automatically when `indent.enable` is set to true. I believe it was set automatically in the past but then it stops doing that at some point (or I just misremember it, idk)

elianiva: ahhhh, so that's the issue. Fixed it, thanks!

elianiva:
Question: does Teal has the same indentation as Lua? As far as I know, Teal is just Lua but strongly typed (and a simpler one than  JS compared TS). If it's the same then I can just copy the existing Lua indents for Teal.

elianiva:
Huh, strange. Teal indentation doesn't work at all. I'll push them and see if anyone also has this issue.

queries/go/indents.scm
```diff
@@ -0,0 +1,18 @@
...
+  (const_declaration)
+  (var_declaration)
+  (type_declaration)
+  (if_statement)
+  (composite_literal)
+  (func_literal)
+] @indent
+
+[
+  (block)
```
yzia2000, line 13: I think this should be put in indent

queries/javascript/indents.scm
```diff
@@ -0,0 +1,25 @@
...
+  (class_body)
+  (method_definition)
+  (named_imports)
+  (binary_expression)
+  (return_statement)
+  (expression_statement (call_expression))
+] @indent
+
+[
+  (statement_block)
```
yzia2000, line 18: Not needed anymore I think

yzia2000 (changes requested):
I think these two can be reverted now that if statement has been removed.

elianiva:
@vigoux I tried fixing the outdent for `}` on else clause and I think I've fixed it. Does it work for you?

vigoux:
We'll wait after #837 is merged to merge this, as we shoudl first fix the core mechanism before trying to find work-arounds in the queries.

vigoux:
Could you try rebasing ? And is this PR ready, I think we can go incrementally afterwards.

elianiva:
Yup, I think this PR is ready. I've been using it daily and have no issue except the ones I've mentioned that we can improve overtime.

vigoux:
We merged #837 so maybe most of these are actually fixed now.

vigoux:
Could you remove the WIP tags, I'll merge that once it's done/

elianiva:
done

PR #829, opened 2021-01-06
Rust locals: Make `const_block` a scope

theHamsta:
Requires #828


theHamsta:
Actually not needed. const_block includes a `block` that is already a scope.

PR #835, opened 2021-01-07
indent: fix wrong line number passed to descendant_for_range

jedrzejboczar:
Currently the number passed is 1-based, but `descendant_for_range` expects 0-based number. It's most likely due to the fact that this piece of code was previously inside the `get_node_at_line` which receives `lnum - 1`.


vigoux (approved):

vigoux:
I don't think this is needed anymore. Am I right ?

jedrzejboczar:
Yes, this was included in https://github.com/nvim-treesitter/nvim-treesitter/pull/837. Just made that PR earlier in case you wanted to include only that bugfix.

PR #836, opened 2021-01-07
C++ highlights: Fix destructor highlighting

theHamsta:
Field is not "name"

or should the highlight be "constructor" or stay "function"?


vigoux:
I would say that using method is great.

vigoux (approved):

PR #837, opened 2021-01-08
Indent improvements

jedrzejboczar:
This PR adds some improvements to indent functionality.
Note that it includes the commit from https://github.com/nvim-treesitter/nvim-treesitter/pull/835.

00a63f0 Uses `vim.fn.prevnonblank` to match indentation of preceding non-blank line after moving to new line with operations like `o` in normal mode or `<cr>` in insert mode. Without this change `root:descendant_for_range` was being used to get the wrapper node. This worked for languages with braces like C, but failed in e.g. Python (now `root:descendant_for_range` is still used as a fallback in some situations, see comments in code).

Following commits add `@return` capture that can be used to mark "last" nodes in a block. It differs from `@branch` in that the `@return` node itself keeps indentation of its block, but when it is the node found by the "prevnonblank" step, than the previous indentation won't be kept. 

I added example `@return` queries for Python. However particularly for Python, it has a corner case, when this won't work when `@return` is in the first line of a block. 
E.g. here `@return` won't be applied (after hitting `o` on the line with "return"):
```python
def foo():
    return
```
This is because `get_node_at_line` will find `block` instead of `return_statement`, because they both start on the same line.
However `@return` will be applied correctly in the following example:
```python
def foo():
    x = 1
    return
```

But I was also able to test `@return` for C in code like:
```c
if (1)
    return;
```
and it worked.
But in C on the other hand I experience an issue with `else` clauses (`@branch` doesn't work properly with `else`). I'm not including any indent queries for C/C++ because they are not really usable with this issue.

The final commits add another capture called `@ignore` which can be used for things like multi-line strings and multi-line comments. When an `@ignore` node is encountered when moving up the tree, no indentation will be performed.

I would be grateful if someone could test if these changes do not introduce any regressions. I was mainly testing on Python, Lua and C++ (with my `indents.scm`) and it seemed ok aside from the things mentioned above.

### DIFF
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -55,10 +55,23 @@ function M.get_indent(lnum)
   local indent = 0
   local indent_size = get_indent_size()
 
-  -- if we are on a new line (for instance by typing `o` or `O`)
-  -- we should get the node that wraps the line our cursor sits in
-  -- and if the node is an indent node, we should set the indent level as the indent_size
-  -- and we set the node as the first child of this wrapper node or the wrapper itself
+  -- to get corret indetation when we land on an empty line (for instance by typing `o`), we try
+  -- to use indentation of previous nonblank line, this solves the issue also for languages that
+  -- do not use @branch after blocks (e.g. Python)
+  if not node then
+    local prevnonblank = vim.fn.prevnonblank(lnum)
+    if prevnonblank ~= lnum then
+      local prev_node = get_node_at_line(root, prevnonblank-1)
+      -- we take that node only if ends before lnum, or else we would get incorrect indent
+      -- on <cr> in positions like e.g. `{|}` in C (| denotes cursor position)
+      if prev_node and (prev_node:end_() < lnum-1) then
+        node = prev_node
+      end
+    end
+  end
+
+  -- if the prevnonblank fails (prev_node wraps our line) we need to fall back to taking
+  -- the first child of the node that wraps the current line, or the wrapper itself
   if not node then
     local wrapper = root:descendant_for_range(lnum-1, 0, lnum-1, -1)
     node = wrapper:child(0) or wrapper
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -35,6 +35,7 @@ local get_indents = utils.memoize_by_buf_tick(function(bufnr)
     indents = get_map('@indent.node'),
     branches = get_map('@branch.node'),
     returns = get_map('@return.node'),
+    ignores = get_map('@ignore.node'),
   }
 end)
 
@@ -85,15 +86,24 @@ function M.get_indent(lnum)
     node = node:parent()
   end
 
+  local first = true
   local prev_row = node:start()
 
   while node do
-    node = node:parent()
-    local row = node and node:start() or prev_row
-    if q.indents[node_fmt(node)] and prev_row ~= row then
+    -- do not indent if we are inside an @ignore block
+    if q.ignores[node_fmt(node)] and node:start() < lnum-1 and node:end_() > lnum-1 then
+      return -1
+    end
+
+    -- do not indent the starting node, do not add multiple indent levels on single line
+    local row = node:start()
+    if not first and q.indents[node_fmt(node)] and prev_row ~= row then
       indent = indent + indent_size
       prev_row = row
     end
+
+    node = node:parent()
+    first = false
   end
 
   return indent
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -22,20 +22,20 @@ local function node_fmt(node)
 end
 
 local get_indents = utils.memoize_by_buf_tick(function(bufnr)
-  local indents = queries.get_capture_matches(bufnr, '@indent.node', 'indents') or {}
-  local branches = queries.get_capture_matches(bufnr, '@branch.node', 'indents') or {}
-
-  local indents_map = {}
-  for _, node in ipairs(indents) do
-    indents_map[tostring(node)] = true
+  local get_map = function(capture)
+    local matches = queries.get_capture_matches(bufnr, capture, 'indents') or {}
+    local map = {}
+    for _, node in ipairs(matches) do
+      map[tostring(node)] = true
+    end
+    return map
   end
 
-  local branches_map = {}
-  for _, node in ipairs(branches) do
-    branches_map[tostring(node)] = true
-  end
-
-  return { indents = indents_map, branches = branches_map }
+  return {
+    indents = get_map('@indent.node'),
+    branches = get_map('@branch.node'),
+    returns = get_map('@return.node'),
+  }
 end)
 
 local function get_indent_size()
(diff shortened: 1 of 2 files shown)


vigoux:
@kyazdani42 I thin you'll be interested

elianiva:
tested this with all of indents that I've added on #825, it doesn't seem to break anything. It even made HTML indent better, well done :)
Will try to play around with `@ignore` capture.

lua/nvim-treesitter/indent.lua
```diff
@@ -22,20 +22,21 @@ local function node_fmt(node)
...
-  end
-
-  local branches_map = {}
-  for _, node in ipairs(branches) do
-    branches_map[tostring(node)] = true
+  local get_map = function(capture)
+    local matches = queries.get_capture_matches(bufnr, capture, 'indents') or {}
+    local map = {}
+    for _, node in ipairs(matches) do
+      map[tostring(node)] = true
```
vigoux, line 29: I'd rather use `node:id()` here :
```suggestion
      map[node:id()] = true
```

lua/nvim-treesitter/indent.lua
```diff
@@ -46,40 +47,63 @@ function M.get_indent(lnum)
...
   if not node then
-    local wrapper = root:descendant_for_range(lnum, 0, lnum, -1)
+    local prevnonblank = vim.fn.prevnonblank(lnum)
+    if prevnonblank ~= lnum then
+      local prev_node = get_node_at_line(root, prevnonblank-1)
+      -- we take that node only if ends before lnum, or else we would get incorrect indent
+      -- on <cr> in positions like e.g. `{|}` in C (| denotes cursor position)
+      local use_prev = prev_node and (prev_node:end_() < lnum-1)
+      -- nodes can be marked @return to prevent using them
+      use_prev = use_prev and not q.returns[node_fmt(prev_node)]
```
vigoux, line 68:
```suggestion
      use_prev = use_prev and not q.returns[prev_node:id()]
```

lua/nvim-treesitter/indent.lua
```diff
@@ -46,40 +47,63 @@ function M.get_indent(lnum)
...
+    end
+  end
+
+  -- if the prevnonblank fails (prev_node wraps our line) we need to fall back to taking
+  -- the first child of the node that wraps the current line, or the wrapper itself
+  if not node then
+    local wrapper = root:descendant_for_range(lnum-1, 0, lnum-1, -1)
     node = wrapper:child(0) or wrapper
-    if indents[node_fmt(wrapper)] ~= nil and wrapper ~= root then
+    if q.indents[node_fmt(wrapper)] ~= nil and wrapper ~= root then
```
vigoux, line 80:
```suggestion
    if q.indents[wrapper:id()] ~= nil and wrapper ~= root then
```

lua/nvim-treesitter/indent.lua
```diff
@@ -46,40 +47,63 @@ function M.get_indent(lnum)
...
       indent = indent_size
     end
   end
 
-  while node and branches[node_fmt(node)] do
+  while node and q.branches[node_fmt(node)] do
```
vigoux, line 85:
```suggestion
  while node and q.branches[node:id()] do
```

lua/nvim-treesitter/indent.lua
```diff
@@ -46,40 +47,63 @@ function M.get_indent(lnum)
...
 
+  local first = true
   local prev_row = node:start()
 
   while node do
-    node = node:parent()
-    local row = node and node:start() or prev_row
-    if indents[node_fmt(node)] and prev_row ~= row then
+    -- do not indent if we are inside an @ignore block
+    if q.ignores[node_fmt(node)] and node:start() < lnum-1 and node:end_() > lnum-1 then
```
vigoux, line 94:
```suggestion
    if q.ignores[node:id()] and node:start() < lnum-1 and node:end_() > lnum-1 then
```

lua/nvim-treesitter/indent.lua
```diff
@@ -46,40 +47,63 @@ function M.get_indent(lnum)
...
+      return -1
+    end
+
+    -- do not indent the starting node, do not add multiple indent levels on single line
+    local row = node:start()
+    if not first and q.indents[node_fmt(node)] and prev_row ~= row then
```
vigoux, line 100:
```suggestion
    if not first and q.indents[node:id()] and prev_row ~= row then
```

vigoux (commented):
Left some suggestion, otherwise awesome additions !

jedrzejboczar:
@vigoux That would be nice as it would allow to use more complex queries than single-node ones. I wanted to try something like this in the future, as I think a more complicated query could solve my issues with `else` in C/C++. 

I tired following your suggestions, but it fails when I indent multiple lines at once (e.g. with `gg=G`) - when there is any change applied during indentation, all the lines after that changed one are indented incorrectly. 

After some investigation I found that this is due to `get_node_at_line` being wrapped in `memoize_by_buf_tick`. When vim indents subsequent lines it seems to apply the changes after each line, but `on_changedtick` is only called once after all the lines have been changed. So the same cache is used for all the lines, but if there is any change, treesitter seems to recompute the tree, then all the node IDs change their values and as a result we fail to identify any nodes thereafter, because the cache is too old. Basically my debug log looks like this:
```
using cache (buf 1)
missing = 0, new = 0
using cache (buf 1)
missing = 0, new = 0
using cache (buf 1)
(...)
missing = 0, new = 0
using cache (buf 1)
missing = 84, new = 84
using cache (buf 1)
missing = 0, new = 0
(...)
using cache (buf 1)
missing = 0, new = 0
Recomputing get_indets(1)
139 lines indented 
```

After removing `memorize_by_buf_tick` I get correct results, but indenting becomes noticeably slower (like an order of magnitude). Is there a way to achieve what `memorize_by_buf_tick` does, but to trigger the recomputation after any change in the buffer (or at least do it in sync with treesitter)?

jedrzejboczar:
I just checked that `nvim_buf_get_changedtick` returns the same value for all lines indented in a single `gg=G` but treesitter node IDs do change, e.g.
```
using cache (buf 1) changedtick 39
missing = 0, new = 0
using cache (buf 1) changedtick 39
missing = 253, new = 253
using cache (buf 1) changedtick 39
missing = 0, new = 0
using cache (buf 1) changedtick 39
missing = 198, new = 198
using cache (buf 1) changedtick 39
missing = 0, new = 0
```
`missing` is the number of unique node IDs that disappeared and `new` is the number of new ones.

vigoux:
So from what I understand, the issue is because of `gg=G` not incrementing the `changedtick` ?
If that's the case, we should open an issue upstream, as this does not seem to be right.

kyazdani42:
the issue is that changedtick is not updated for each parser iteration, which makes the indent cache out of sync. 
This might be the expected behavior though, I'm not totally sure how changedtick is usually updated

jedrzejboczar:
I tried to force `SIN_CHANGED` in [set_indet](https://github.com/neovim/neovim/blob/master/src/nvim/indent.c#L96). This makes the `changedtick` increment correctly, but `on_changedtick` does not seem to get called. 

Digging deeper I found that `op_reindent` (called for `=`) calls `changed_lines` just once [at the end](https://github.com/neovim/neovim/blob/master/src/nvim/ops.c#L673-L675). I tried adding:
```c
changed_lines(curwin->w_cursor.lnum, 0, curwin->w_cursor.lnum + 1, 0L, true);
```
[here](https://github.com/neovim/neovim/blob/master/src/nvim/ops.c#L654). This also makes `changedtick` be updated correctly, but `on_changedtick` is still not being called.
Now I'm not really sure where the callback is triggered, I'm guessing it's [this](https://github.com/neovim/neovim/blob/master/src/nvim/buffer_updates.c#L219)? Or am I wrong? I am not sure about it.

I think I'll open an issue about it later in neovim upstream. For now we can probably keep using `tostring(node)` as it should still work?

vigoux:
Okay, let's open an issue upstream, and merge that, you already brought a ton of improvements, so will go with that.
Could you open an issue upstream linking with the discussion here ? Thanks !

vigoux (approved):

Issue #838, opened 2021-01-08
highlighting works as unexpected after attach async, e0b49a9a1885708d3565e9a59847b14a42faef2c


kevinhwang91:
**Describe the bug**
see reproduce.

**To Reproduce**
1. `nvim foo.lua +'e bar.lua' +'b #' `

`bar.lua` can't be highlighted.

**Expected behavior**
`bar.lua` must be highlighted.

vigoux:
Indeed the highlights differ, but I am not sure this actually comes from the async part.

vigoux:
Did a git bisect, and indeed the first bad commit is e0b49a9a1885708d3565e9a59847b14a42faef2c.

PR #840, opened 2021-01-08
Bugfix: fixed async highlighting when opening multiple files

oberblastmeister:
I have fixed the async highlighting on buf attach. The issue was that we need to get the buffer number synchronously and *then* do everything async or the buffer number is wrong by the time the async kicks in.


lua/nvim-treesitter/configs.lua
```diff
@@ -299,8 +299,8 @@ end
 -- @param bufnr the bufnr
 -- @param lang the language of the buffer
 function M.attach_module(mod_name, bufnr, lang)
-  local bufnr = bufnr or api.nvim_get_current_buf()
-  local lang = lang or parsers.get_buf_lang(bufnr)
```
vigoux, line 303: Why having a global here ?

lua/nvim-treesitter/configs.lua
```diff
@@ -311,7 +311,16 @@ function M.attach_module(mod_name, bufnr, lang)
   end
 end
 
-M.attach_module_async = utils.async(M.attach_module)
+-- must have custom async function or bufnr will be wrong
+function M.attach_module_async(mod_name, bufnr)
+  bufnr = bufnr or api.nvim_get_current_buf() -- get bufnr before doing async, or bufnr will be wrong
```
vigoux, line 316: same here, I think you mean `local`

vigoux (changes requested):
Great, thanks for fixing the issue.

vigoux:
@kevinhwang91 does that fix #838  ?

lua/nvim-treesitter/configs.lua
oberblastmeister: When you use the local keyword the variable will be redefined, and the langserver will give a warning. Since the variable comes from the function argument it is already local. It is like doing
```lua
local bufnr = <user arg> -- from the function parameter
bufnr = bufnr or api.nvim_get_current_buf() -- not global, mutating local
```

oberblastmeister: Its kind of a small change, mostly just to make the language server not give out a warning.

vigoux: Oh yeah sorry for that, is see now, and that's better indeed

kevinhwang91:
Yes, it does.

vigoux (approved):

rafikdraoui:
This change introduces a new issue. I don't exactly know what in my workflow trigger this error, but I found a reproducible case that can be bisected to commit https://github.com/nvim-treesitter/nvim-treesitter/commit/28f2ba475ff7324cd840b8130d7817dcd393e1ed

```vim
cexpr getline(1, '$')
copen
cexpr getline(1, '$')
```

This will throw several (identical) error messages:

```
Error executing vim.schedule lua callback: .../nvim-treesitter/lua/nvim-treesitter/parsers.lua:487: Invalid buffer id: 2
```

theHamsta:
Probably a race condition. We could check on attachment whether the buffer still exists. When we launch async the buffer may be deleted in the mean time

f12markovic:
I'm getting this `Invalid buffer id error` whenever i jump to c/cpp function definition in system headers provided by ALE

PR #843, opened 2021-01-09
added query enhancements for json and toml

oberblastmeister:


theHamsta (approved):

queries/json/locals.scm
```diff
@@ -0,0 +1,2 @@
+(object) @scope
+(array) @scope
```
vigoux, line 2: Could you use an alternation for that ?
```suggestion
[
  (object)
  (array) 
] @scope
```

vigoux (approved):

PR #845, opened 2021-01-12
Revert "added attach async"

vigoux:
Following issue #838 , https://github.com/nvim-treesitter/nvim-treesitter/pull/824#issuecomment-758056977, I think it is better to revert #824 as we did not gained that much performance.

I'd rather focus on optimizing things than making them async.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -78,7 +78,7 @@ local function enable_mod_conf_autocmd(mod)
     return
   end
 
-  local cmd = string.format("lua require'nvim-treesitter.configs'.attach_module_async('%s')", mod)
+  local cmd = string.format("lua require'nvim-treesitter.configs'.attach_module('%s')", mod)
   api.nvim_command(string.format("autocmd NvimTreesitter FileType * %s", cmd))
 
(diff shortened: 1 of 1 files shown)


theHamsta:
Why? The async makes a huge difference when you're loading a big parser with a lot of queries on a large file (thinking of C++).

I understand that you want to revert now, but a long term goal should be to have a working async implementation. It feels so much better and the bugs are not inherent to an async implementation.

tami5:
> Why? 

This

theHamsta (approved):

vigoux:
> Why? The async makes a huge difference when you're loading a big parser with a lot of queries on a large file (thinking of C++).

I'd rather focus on incrementally parsing the file for that, using timeouts, so that we avoid blocking operations. 
Furthermore, I'd also optimize both the queries (maybe limit regexes and so on) to make them run faster.

vigoux:
Also fixes #846

steelsojka:
I agree with @theHamsta. We can revert it now for stability but it should be a goal for the apis to handle async processes.

srillia:
vscode async processes is perfect ,without async processes ,treesitter is too weak.

PR #850, opened 2021-01-14
fix(ci): only track lockfile.json

vigoux:


.github/workflows/update-parsers-pr.yml
```diff
@@ -35,6 +35,11 @@ jobs:
           cp lockfile.json /tmp/lockfile.json
           cat /tmp/lockfile.json | /tmp/jq --sort-keys > lockfile.json
 
+      - name: Commit changes
+        run: |
+          git add lockfile.json
+          git commit -m "Update lockfile.json"
+
```
theHamsta, line 42: maybe we could also add a cron trigger? Don't know how to formulate it that it only get's executed on master.
```
on:  
  push:
  pull_request:
  schedule:
    - cron: '0 0 * * *'  # every day at midnight
```

theHamsta: probably also unrelated to this PR.

vigoux: Yeah I think it's unrelated, and I'd like to keep it like this for now, to synchronise updates with actual activity periods of the plugin.

theHamsta, line 42: how does this change anything in the logic? Won't the create pull request action do another commit?

vigoux: Apparently not, given the example showed in the `README`

theHamsta: :shrug: ok

vigoux: Lol yeah, let's try that, it wont hurt anyway :wink:

theHamsta (approved):

PR #858, opened 2021-01-16
Use function keyword highlight group for Rust's fn

cbarrete:
The Rust grammar doesn't have a special token for fn, so it has to remain hardcoded, but that's still an improvement.

### DIFF
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -140,7 +140,6 @@
 "dyn"
 "enum"
 "extern"
-"fn"
 "impl"
 "let"
 "macro_rules!"
@@ -164,6 +163,8 @@
 (attribute_item)
 (inner_attribute_item)
  ] @keyword
+
+"fn" @keyword.function
 
 (use_list (self) @keyword)
 (scoped_use_list (self) @keyword)


theHamsta:
Using `"fn"` is not hard coding. It will only match the token `"fn"` not when you name happen to use an identifier "fn" (so when there is a verbatim "fn" in grammar.js.
image

cbarrete:
Of course, what I meant by that is that I had to put `("fn" @keyword.function)` rather than ` (fn @keyword.function)`.
I'm not sure what the right TS terminology is, but essentially there's no lambda for it in its `grammar.js`.

cbarrete:
Of course, what I meant is that I had to use `("fn" @keyword.function)` rather than `(fn @keyword.function)`.
Not sure what the right terminology is but there's no lambda for `fn` in Rust's `grammar.js`.

queries/rust/highlights.scm
```diff
@@ -165,6 +164,8 @@
 (inner_attribute_item)
  ] @keyword
 
+ ("fn" @keyword.function)
```
vigoux, line 167: It should be valid to change this to 
```suggestion
 "fn" @keyword.function
```

vigoux:
By the way, the terminolgy for those kind of node is `terminal` as they are leafs in the syntax-tree.

queries/rust/highlights.scm
cbarrete: Nice

vigoux (approved):

PR #877, opened 2021-01-26
feat(javascript): move shorthand prop to property

steelsojka:


queries/javascript/highlights.scm
```diff
@@ -11,13 +11,11 @@
 ;-----------
 
 (property_identifier) @property
-(shorthand_property_identifier) @variable
+(shorthand_property_identifier) @property
 
 ; Special identifiers
 ;--------------------
 
-(identifier) @variable
```
steelsojka, line 19: This was duplicated from above

vigoux (approved):

PR #879, opened 2021-01-27
Add cache for get_query function

oberblastmeister:
Is putting the cache inside of a `do ... end` block the right choice?

### DIFF
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -75,7 +75,7 @@ do
         end
       end
     else
-      error("Must have cannot have query_name by itself!")
+      error("Cannot have query_name by itself!")
     end
   end
 end
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -81,7 +81,7 @@ do
       end
     elseif not lang and not query_name then
       for lang, _ in pairs(cache) do
-        for query_name in pairs(cache[lang]) do
+        for query_name, _ in pairs(cache[lang]) do
           M.reload_file_cache(lang, query_name)
         end
       end
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -55,7 +55,7 @@ do
     return rawget(tbl, key)
   end
 
-  -- cache will have will auto set the table for each lang if it is nil
+  -- cache will auto set the table for each lang if it is nil
   local cache = setmetatable({}, mt)
 
   --- Same as `vim.treesitter.query` except will return cached values
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -25,7 +25,10 @@ do
   local query_cache = caching.create_buffer_cache()
 
   local function update_cached_matches(bufnr, changed_tick, query_group)
-    query_cache.set(query_group, bufnr, {tick=changed_tick, cache=( M.collect_group_results(bufnr, query_group) or {} )})
+    query_cache.set(query_group, bufnr, {
+        tick = changed_tick,
+        cache= M.collect_group_results(bufnr, query_group) or {}
+    })
   end
 
   function M.get_matches(bufnr, query_group)
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -7,8 +7,6 @@ local caching = require'nvim-treesitter.caching'
 
 local M = {}
 
-local query_cache = caching.create_buffer_cache()
-
 M.built_in_query_groups = {'highlights', 'locals', 'folds', 'indents'}
 
 -- Creates a function that checks whether a given query exists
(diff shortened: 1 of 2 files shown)


lua/nvim-treesitter/query.lua
```diff
@@ -46,8 +46,20 @@ function M.has_query_files(lang, query_name)
...
+    if cache[lang] == nil then
+      cache[lang] = {}
+    end
+
+    if cache[lang][query_name] == nil then
+      cache[lang][query_name] = tsq.get_query(lang, query_name)
+    end
+
+    return cache[lang][query_name]
+  end
```
theHamsta, line 62: when I'm proposed this everyone wanted to have a version that can load the changes from disk. But I'm also for caching the queries as using the queries is a much more frequent use-case than editing and changing the queries

oberblastmeister: I agree because changing the queries is done less often. Also it is slow because if you are running this a lot it runs `io.open` synchronously and then re-parses the whole thing. I talked with bfredl and vigoux on gitter and they suggested that it be cached so thats why i made this pr.

oberblastmeister: Maybe we can have a reload function if people are editing the queries.

theHamsta: I think a autocmd for query buffers on write could invalidate the cache

lua/nvim-treesitter/query.lua
```diff
@@ -46,8 +46,38 @@ function M.has_query_files(lang, query_name)
...
+        M.reload_file_cache(lang, query_name)
+      end
+    elseif not lang and not query_name then
+      for lang, _ in pairs(cache) do
+        for query_name in pairs(cache[lang]) do
+          M.reload_file_cache(lang, query_name)
+        end
+      end
+    else
+      error("Must have cannot have query_name by itself!")
```
theHamsta, line 78: Something got swapped in this message

lua/nvim-treesitter/query.lua
oberblastmeister: I am not sure what you mean?

p00f: write an autocmd to invalidate cache when you save query buffers

p00f: `query` is a recognized filetype after installing nvim-treesitter

oberblastmeister: thank you, makes sense now

lua/nvim-treesitter/query.lua
```diff
@@ -61,9 +62,22 @@ do
...
+        cache[lang] = {}
+      end
+    end
+
+    -- nil means query is not cached yet. "invalid" means that query was cached and was not a valid one
     if lang and query_name then
-      cache[lang][query_name] = tsq.get_query(lang, query_name)
+      local got = tsq.get_query(lang, query_name)
+      if got == nil then
+        cache[lang][query_name] = "invalid"
```
vigoux, line 77: I am not sure we should do the disctinction here, if there was an error, there is a lot of chances the error will be fixed, then the query will be resourced, so using nil is fine here.

lua/nvim-treesitter/query.lua
```diff
@@ -80,6 +94,12 @@ do
   end
 end
 
+--- This function is meant for an autocommand and not to be used. Only use if file is a query file.
+function M.reload_file_cache_on_write()
+  print('reloading cache')
+  M.reload_file_cache(vim.fn.expand('%:p:h:t'), vim.fn.expand('%:t:r'))
+end
```
vigoux: This should rather use a filename, and `fnamemodify`.

plugin/nvim-treesitter.vim
```diff
@@ -5,6 +5,9 @@ if exists('g:loaded_nvim_treesitter')
 endif
 
 augroup NvimTreesitter
+  " on every query file write we want to set and autocommand that will reload the cache
+  autocmd FileType query
+      \ autocmd! NvimTreesitter BufWritePost <buffer> lua require'nvim-treesitter.query'.reload_file_cache_on_write()
```
vigoux, line 10: This should pass the argument to the function.

lua/nvim-treesitter/query.lua
```diff
@@ -46,8 +46,56 @@ function M.has_query_files(lang, query_name)
...
+do
+  local mt = {}
+  mt.__index = function(tbl, key)
+    if rawget(tbl, key) == nil then
+      rawset(tbl, key, {})
+    end
+    return rawget(tbl, key)
+  end
+
+  -- cache will have will auto set the table for each lang if it is nil
```
vigoux, line 58:
```suggestion
  -- cache will auto set the table for each lang if it is nil
```

lua/nvim-treesitter/query.lua
```diff
@@ -46,8 +46,56 @@ function M.has_query_files(lang, query_name)
...
-function M.get_query(lang, query_name)
-  return tsq.get_query(lang, query_name)
+do
+  local mt = {}
+  mt.__index = function(tbl, key)
+    if rawget(tbl, key) == nil then
+      rawset(tbl, key, {})
+    end
+    return rawget(tbl, key)
+  end
```
vigoux, line 56: Don't we have utilities for caching in `caching.lua` ?

lua/nvim-treesitter/query.lua
```diff
@@ -46,8 +46,56 @@ function M.has_query_files(lang, query_name)
...
+
+    return cache[lang][query_name]
+  end
+
+  --- Reloads the query file cache.
+  --- If lang and query_name is both present, will reload for only the lang and query_name.
+  --- If only lang is present, will reload all query_names for that lang
+  --- If none are present, will reload everything
+  function M.reload_file_cache(lang, query_name)
+    -- nil means query is not cached yet. "invalid" means that query was cached and was not a valid one
```
vigoux, line 75:
```suggestion
```

lua/nvim-treesitter/query.lua
```diff
@@ -46,8 +46,56 @@ function M.has_query_files(lang, query_name)
...
+    end
+  end
+end
+
+--- This function is meant for an autocommand and not to be used. Only use if file is a query file.
+function M.reload_file_cache_on_write(fname)
+  local fnamemodify = vim.fn.fnamemodify
+  print('reloading cache')
+  M.reload_file_cache(fnamemodify(fname, ':p:h:t'), fnamemodify(fname, ':t:r'))
 end
```
vigoux, line 99: Out of curiosity: why having that out of the `do ... end` block ?
This block really helps reading btw, so that's great !

lua/nvim-treesitter/query.lua
oberblastmeister: That seems to be for caching on each buffer while this is for a global cache of the queries files.  For example the function that uses `caching.lua` is`get_matches`. That will run `update_cached_matches` which runs `iter_group_results` which will get the query files for the current buffer using `get_query` which is the function that I am implementing caching for.

lua/nvim-treesitter/query.lua
oberblastmeister: I put it out of the `do ... end` block for privacy so it cannot directly access the cache and instead has to call a function that will access the cache. That function pretty much only modifies the file so I thought that it should remain separate from the caching stuff.

oberblastmeister:
`get_capture_matches` seems to use `iter_group_results` which `get_matches` does the same and is also cached. Would it make sense to change it?

vigoux:
No I don't think it is worth changing now, or rather, we'll look into it later.

vigoux (approved):

PR #881, opened 2021-01-28
Add ledger parser, folds and highlights queries

cbarrete:


queries/ledger/highlights.scm
```diff
@@ -0,0 +1,16 @@
+((comment) @comment)
+
+((date) @number)
+((account) @field) ; TODO variable would be more fitting, but nvim-treesitter doesn't have a special color for it yet (see TSVariable)
```
vigoux, line 4: I think that if it is a variable, then you should highlight that as a variable.

Tweaking highlights for a language will be possible soon, and will allow you to change the color of `TSVariable` for `ledger` only for example.

cbarrete: Yeah, I should have cleaned that up, good catch :)

vigoux (approved):

PR #890, opened 2021-01-30
Update lockfile

theHamsta:
Copied the fix from the tree-sitter-teal repo.


vigoux (approved):

theHamsta:
Oops the took the wrong CI commit. Will update soon

PR #895, opened 2021-01-31
Implement toggle commands

### LINKED ISSUE #259: TSBufToggle
When i work in some files, i sometime want to enable a plugin that uses the default neovim syntax (like styled-components stuff in js files).
Mapping TSBufDisable and TSBufEnable would work but it's a hassle and it would be easier to just map `TSBufToggle highlight`.

echasnovski:
This PR adds toggle (enable if disabled, disable if enabled) commands `TSBufToggle` and `TSToggleAll`. They should act exactly like enable and disable commands depending on the state of module.

I also took the liberty of making small fixes to documentation of existing commands (sync command names, tweak style).

Fixes #259.

### DIFF
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -231,6 +231,13 @@ A list of modules can be found at |:TSModuleInfo|
 Disable {module} on the current buffer
 A list of modules can be found at |:TSModuleInfo|
 
+                                                                *:TSBufToggle*
+:TSBufToggle {module}~
+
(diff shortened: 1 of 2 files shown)


vigoux (approved):
Yes yes yes yes yes yes yes !

Thank you soo much for that 😉 

I'll let others approve that too to be sure, but thanks thanks thanks for that and welcome in the contributors team !

theHamsta (approved):

theHamsta:
Thanks a lot for that!

PR #900, opened 2021-01-31
Add devicetree parser with queries

jedrzejboczar:
This adds Linux [Device Tree](https://www.kernel.org/doc/html/latest/devicetree/usage-model.html) files parser (https://github.com/joelspadin/tree-sitter-devicetree).

Fold queries are straightforward. For highlights I'm not entirely sure of the captures that should be used, I tried to select the ones that seemed sensible (and to end up with roughly similar colors as with vim's `syntax/dts.vim`, at least for gruvbox). I am not really sure what `locals.scm` are, maybe someone could check/improve/remove these queries. Indents work fine for me, there is some issue with indenting the first node in the file, but couldn't find what causes it.

Some `.dts`/`.dtsi` files to test these queries can be found e.g. [here](https://github.com/torvalds/linux/tree/master/arch/arm/boot/dts).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,7 @@ Each of these `scheme` files contains a *tree-sitter query* for a given purpose.
 Before going any further, we highly suggest that you [read more about tree-sitter queries](https://tree-sitter.github.io/tree-sitter/using-parsers#pattern-matching-with-queries).
 
 Each query has an appropriate name, which is then used by modules to extract data from the syntax tree.
-For now the following types of queries are used by `nvim-treesitter`:
+For now two types of queries are used by `nvim-treesitter`:
 
 - `highlights.scm`: used for syntax highlighting, using the `highlight` module.
 - `locals.scm`: used to extract keyword definitions, scopes, references, etc, using the `locals` module.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,7 @@ Each of these `scheme` files contains a *tree-sitter query* for a given purpose.
 Before going any further, we highly suggest that you [read more about tree-sitter queries](https://tree-sitter.github.io/tree-sitter/using-parsers#pattern-matching-with-queries).
 
 Each query has an appropriate name, which is then used by modules to extract data from the syntax tree.
-For now two types of queries are used by `nvim-treesitter`:
+For now the following types of queries are used by `nvim-treesitter`:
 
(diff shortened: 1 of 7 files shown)


theHamsta:
locals.scm try to give semantic meaning to the nodes by determining which nodes are definitions, scopes or usages of a variables. I don't think this makes sense for config language.

jedrzejboczar:
Ok, so maybe `@tag` or `@label` then? This is a capture for things like [version](https://github.com/torvalds/linux/blob/master/arch/arm/boot/dts/am335x-bone.dts#L5) and [memreserve](https://github.com/torvalds/linux/blob/master/arch/arm/boot/dts/bcm283x.dtsi#L11).

theHamsta:
No, `@attribute` is totally fine. We just have to clean up our docs.

jedrzejboczar:
I replaced `@annotation` with `@attribute` and changed the capture group for `reference` as it references `labeled_item.identifier` so both now have `@label` group. Now, when you said what `@attribute` is used for I also changed the group for `node.identifier` to `@namespace` as it seems to make more sense (`node.identifier` kind-of introduces a new scope).

CONTRIBUTING.md
```diff
@@ -47,7 +47,7 @@ Each of these `scheme` files contains a *tree-sitter query* for a given purpose.
 Before going any further, we highly suggest that you [read more about tree-sitter queries](https://tree-sitter.github.io/tree-sitter/using-parsers#pattern-matching-with-queries).
 
 Each query has an appropriate name, which is then used by modules to extract data from the syntax tree.
-For now two types of queries are used by `nvim-treesitter`:
+For now the following types of queries are used by `nvim-treesitter`:
```
vigoux, line 50: That seems like an unrelated change.
I tend to like keeping PRs into their scopes.

lua/nvim-treesitter/parsers.lua
```diff
@@ -416,6 +416,15 @@ list.turtle = {
...
 
+list.devicetree = {
+  install_info = {
+    url = "https://github.com/joelspadin/tree-sitter-devicetree",
+    files = { "src/parser.c" },
+    branch = "main",
+  },
+  filetype = 'dts',
+}
+
```
vigoux, line 427: So you will not maintain the queries for this language ?

jedrzejboczar: Probably I could

vigoux (approved):

Issue #902, opened 2021-02-01
LR: SuperCollider support


madskjeldgaard:
**Language informations**
Hi!
It would be nice to add support for the supercollider language at some point. It works for filetype `supercollider` and suffices `.scd` and `.sc`.

I'm working on a grammar etc here: https://github.com/madskjeldgaard/tree-sitter-supercollider

Parser config could look like this:
```lua
parser_config.supercollider = {
	install_info = {
		-- url = "~/code/tree-sitter-supercollider",
		url = "https://github.com/madskjeldgaard/tree-sitter-supercollider",
		files = {"src/parser.c"},
		maintainer = "@madskjeldgaard"
	},
	filetype = "supercollider", -- if filetype does not agrees with parser name
}
```

theHamsta:
Do you want to create a PR for this? You only have to add your snippet to `parsers.lua`

PR #906, opened 2021-02-02
Supercollider

madskjeldgaard:
This adds support for the supercollider programming language. See https://github.com/nvim-treesitter/nvim-treesitter/issues/902

### DIFF
--- a/queries/supercollider/highlights.scm
+++ b/queries/supercollider/highlights.scm
@@ -75,7 +75,7 @@
 ; "super"
 ; "this"
 "var"
-] @type.builtin
+] @keyword
 
 ; Brackets
 [
--- a/queries/supercollider/highlights.scm
+++ b/queries/supercollider/highlights.scm
@@ -98,7 +98,7 @@
 ; control structure
 (control_structure) @conditional
 
-(escape_sequence) @escape
+(escape_sequence) @string.special
 
 ; SinOsc.ar()!2
 (duplicated_statement) @repeat
--- a/queries/supercollider/highlights.scm
+++ b/queries/supercollider/highlights.scm
@@ -6,8 +6,6 @@
 (block_comment) @comment
 
 ; Argument definition
-
-; Fixme: Should be attribute but nvim-treesitter doesn't supprt it at the moment
 (argument name: (identifier) @parameter) 
 
 ; Variables
@@ -101,8 +99,3 @@
 
 ; SinOsc.ar()!2
 (duplicated_statement) @repeat
-
-; Builtins
-; ((identifier) @variable.builtin
-;  (#match? @variable.builtin "^(inf|nil|thisFunction|thisFunctionDef|thisMethod|thisProcess|thisThread|currentEnvironment|topEnvironment)$")
-;  (#is-not? local))
--- a/queries/supercollider/highlights.scm
+++ b/queries/supercollider/highlights.scm
@@ -9,7 +9,6 @@
 
 ; Fixme: Should be attribute but nvim-treesitter doesn't supprt it at the moment
 (argument name: (identifier) @parameter) 
-;(argument name: (identifier) @parameter)  
 
 ; Variables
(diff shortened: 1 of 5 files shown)


madskjeldgaard:
I had to run the lockfile update script to be able to extract the tar ball. I am confused about this (since the CONTRIBUTING.md file doesn't seem to mention this) but should I upload my updated lockfile as well?

theHamsta:
No, the lockfile is a automated CI job. Usually, you should be able to install directly from master when there's no lockfile. Maybe it did not respect that you put `branch=main` there?

theHamsta (approved):

madskjeldgaard:
> No, the lockfile is an automated CI job. Usually, you should be able to install directly from master when there's no lockfile. Maybe it did not respect that you put `branch=main` there?
> 
> On my machine, it works without doing anything with the lockfile.

That's great to hear. Thanks!

queries/supercollider/highlights.scm
```diff
@@ -0,0 +1,109 @@
...
+
+; comments
+(line_comment) @comment
+(block_comment) @comment
+
+; Argument definition
+
+; Fixme: Should be attribute but nvim-treesitter doesn't supprt it at the moment
+(argument name: (identifier) @parameter) 
+;(argument name: (identifier) @parameter)
```
vigoux, line 12: I think there was a discussion of using `property` instead of attribute.

Plu a minor duplication : 
```suggestion
; Fixme: Should be attribute but nvim-treesitter doesn't supprt it at the moment
(argument name: (identifier) @parameter) 
```

queries/supercollider/locals.scm
```diff
@@ -0,0 +1,36 @@
+; The local variables query is different from the highlights query in that, while the highlights query uses arbitrary capture names which can then be mapped to colors, the locals variable query uses a fixed set of capture names, each of which has a special meaning.
+; The capture names are as follows:
+; @local.scope - indicates that a syntax node introduces a new local scope.
+; @local.definition - indicates that a syntax node contains the name of a definition within the current local scope.
+; @local.reference - indicates that a syntax node contains the name which may refer to an earlier definition within some enclosing scope.
```
vigoux, line 5: These are not valid anymore.

queries/supercollider/highlights.scm
```diff
@@ -0,0 +1,109 @@
...
+
+(escape_sequence) @string.special
+
+; SinOsc.ar()!2
+(duplicated_statement) @repeat
+
+; Builtins
+; ((identifier) @variable.builtin
+;  (#match? @variable.builtin "^(inf|nil|thisFunction|thisFunctionDef|thisMethod|thisProcess|thisThread|currentEnvironment|topEnvironment)$")
+;  (#is-not? local))
```
vigoux, line 109: Do you really need the `is-not?` predicate here ? If not, you may just remove it and only use the match predicate.

queries/supercollider/locals.scm
```diff
@@ -0,0 +1,36 @@
...
+
+(variable_definition 
+	name: (variable (environment_var (identifier) @definition.var))
+	(set! "definition.var.scope" "global"))
+
+(function_definition name: (variable) @definition.var
+	 (set! "definition.var.scope" "parent")
+)
+;(variable_definition name: (variable (local_var(identifier))) @local.definition) 
+;(variable) @local.definition
```
vigoux, line 34: I think it is a duplication.

```suggestion
```

vigoux (commented):
Thanks for the contribution !
I left a few comments here and there, nothing really major, just some things.

queries/supercollider/locals.scm
madskjeldgaard: Thanks I cleaned these up

vigoux (approved):
Looks like we're good to go, thanks for this awesome addition!

madskjeldgaard:
Thank you for this awesone plugin :)

PR #916, opened 2021-02-04
Update lockfile.json and `amount` query for ledger

cbarrete:
Cleanly solve the issue in https://github.com/nvim-treesitter/nvim-treesitter/pull/910.

@theHamsta this should be fine :)


theHamsta (approved):
Thanks you :slightly_smiling_face:

PR #920, opened 2021-02-05
feat: add support for svelte

elianiva:
~~we'll need to wait until https://github.com/Himujjal/tree-sitter-svelte/pull/3 got merged but I thought I'll put this here so everyone could comment on this.~~ It has been merged

also, I don't know who to put in the maintainer field, is it someone who maintains the query or is it the one who maintains the parser? or both? I can only maintain the queries since I don't know much about C to make the scanner for the parser :laughing:

~~edit: should wait for #926 because I added some stuff from there (the reason why the test failed), Svelte is like HTML afterall so they have similar queries~~. It's ready now.


vigoux:
@elianiva the maintainer part is indeed the maintainer of the queries 😉

elianiva:
> @elianiva the maintainer part is indeed the maintainer of the queries

Thanks for the quick reply! I'll change it.

elianiva:
This is ready if anyone wants to merge this. There's still https://github.com/Himujjal/tree-sitter-svelte/pull/4 but that's trivial and will be fixed soon.

theHamsta (approved):

PR #932, opened 2021-02-07
Add `selection_mode` argument to `ts_utils.update_selection()`

echasnovski:
This update allows setting a visual selection mode when calling `ts_utils.update_selection()`. Essentially, behavior is equivalent to the following steps: go to start position of the node, invoke certain visual selection mode, go to end position of the node.

Note for possible update: for blockwise mode select block not from start to end positions, but block containing the whole node. The drawback is that it will not be aligned with default Vim's behavior: after selecting region, user can change selection mode (by typing `v`, `V`, or `<C-v>`). So with current approach two actions are equivalent: `update_selection(..., "blockwise")` and `update_selection(..., "charwise")` followed by `<C-v>`.

Related PR: nvim-treesitter/nvim-treesitter-textobjects/pull/37

### DIFF
--- a/lua/nvim-treesitter/ts_utils.lua
+++ b/lua/nvim-treesitter/ts_utils.lua
@@ -131,7 +131,8 @@ function M.highlight_range(range, buf, hl_namespace, hl_group)
 end
 
 -- Set visual selection to node
--- @param selection_mode One of "charwise" (default), "linewise", "blockwise"
+-- @param selection_mode One of "charwise" (default) or "v", "linewise" or "V",
+--   "blockwise" or "<C-v>" (as a string with 5 characters or a single character)
 function M.update_selection(buf, node, selection_mode)
   selection_mode = selection_mode or "charwise"
   local start_row, start_col, end_row, end_col = M.get_node_range(node)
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/ts_utils.lua
```diff
@@ -145,7 +147,15 @@ function M.update_selection(buf, node)
   end_col = end_col + 1
 
   vim.fn.setpos(".", { buf, start_row, start_col, 0 })
-  vim.fn.nvim_exec("normal! v", false)
+
+  -- Start visual selection in appropriate mode
+  local v_table = {charwise = "v", linewise = "V", blockwise = "<C-v>"}
```
theHamsta, line 152: you could just use `v_table[selction_mode] or selction_mode` to also allow `<c-v>` in the optional parameter.

echasnovski: Do you suggest that possible values of `seleciton_mode` are **only** "v", "V" and "\<C-v\>", or **all** "charwise", "linewise", "blockwise", "v", "V" and "\<C-v\>"? I didn't like the first approach because it introduces cryptic values, and second one because it introduces duplicates.

But this has nothing to do with `nvim-treesitter-textobjects` PR, because there mapping is done from different outputs of Vim's functions to appropriate values of `selection_mode`. And those are different depending on keymapping mode.

theHamsta: > all "charwise", "linewise", "blockwise", "v", "V" and "<C-v>"

all of course. first three are for human usage and the rest to pipe vims abbreviation of these modes as keymaps from other functions.

> introduces duplicates

so what? `v_table[selction_mode] or selction_mode` is much shorter than the alternative that would be necessary when you want to update from vim's abbreviation. Selection mode can either be of the three string codes that you suggested or the corresponding key abbreviation used by vim.

echasnovski: Sure, no problem. I'll just make sure that "\<C-v\>" can be supplied either as 5 character string or as a single character.

echasnovski:
Is there something I can do to help merge this? This PR will enable nvim-treesitter/nvim-treesitter-textobjects/pull/37 , which will increase comfort of package's day-to-day usage (at least for me).

theHamsta:
Thanks for reminding me on that. For me it looks good to merge. I'll just do a bit of testing later today.

theHamsta (approved):

PR #937, opened 2021-02-08
Initial support for Vue files

### LINKED ISSUE #936: No `highlights.scm` query found for vue
I'm trying to add highlighting for `vue`. I've followed the [instructions on the README](https://github.com/nvim-treesitter/nvim-treesitter#advanced-setup), and have reached a point where `:checkhealth` suggests I open an issue.

1. I cloned `https://github.com/ikatyang/tree-sitter-vue.git` into `/home/hugo/workspace/Forks/tree-sitter-vue`.
2. I ran `tree-sitter generate && tree-sitter test`
3. I added this to my config:

```lua
parser_config.vue = {
  install_info = {
    url = "/home/hugo/workspace/Forks/tree-sitter-vue",
    files = {"src/parser.c", "src/scanner.cc"}
  },
  filetype = "vue"
}
```

4. I ran `:TSInstall`.

Syntax highlighting for files where `filetype=vue` doesn't work. `:checkhealth` shows this:

```
## vue parser healthcheck
  - OK: vue parser found.
  - WARNING: No `highlights.scm` query found for vue
    - ADVICE:
      - Open an issue at https://github.com/nvim-treesitter/nvim-treesitter
  - WARNING: No `locals.scm` query found for vue
    - ADVICE:
      - Open an issue at https://github.com/nvim-treesitter/nvim-treesitter
  - WARNING: No `folds.scm` query found for vue
    - ADVICE:
      - Open an issue at https://github.com/nvim-treesitter/nvim-treesitter
  - WARNING: No `indents.scm` query found for vue
    - ADVICE:
      - Open an issue at https://github.com/nvim-treesitter/nvim-treesitter
```

Am I missing something?

WhyNotHugo:
Due to limitations in tree-sitter-vue, TypeScript and JavaScript are indistinguishable. Since the former is a superset of the latter, always treating them as TypeScript should result in valid highlighting in all
cases.

Support for LESS and other formats requires support from the tree-sitter itself.

Closes #936

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -300,7 +300,8 @@ list.vue = {
   install_info = {
     url = "https://github.com/ikatyang/tree-sitter-vue",
     files = { "src/parser.c", "src/scanner.cc" },
-  }
+  },
+  maintainers = {"@WhyNotHugo"},
 }
 
 list.jsonc = {
--- a/queries/vue/folds.scm
+++ b/queries/vue/folds.scm
@@ -1,0 +1,6 @@
+[
+  (element)
+  (template_element)
+  (script_element)
+  (style_element)
+] @fold
--- a/queries/vue/highlights.scm
+++ b/queries/vue/highlights.scm
@@ -1,0 +1,38 @@
+[
+  (component)
+  (template_element)
+  (start_tag)
+  (tag_name)
+  (directive_attribute)
+  (directive_dynamic_argument)
+  (directive_dynamic_argument_value)
+  (component)
+  (end_tag)
+] @tag
+
+(erroneous_end_tag_name) @error
+(attribute_name) @property
+(attribute_value) @string
+(quoted_attribute_value) @string
+(comment) @comment
+
+(text) @none
+(element) @string
+(interpolation) @punctuation.special
+(interpolation
+  (raw_text) @none)
+
+[
+  (directive_modifier)
+  (directive_name)
+  (directive_argument)
+] @method
+
+"=" @operator
+
+[
+ "<"
+ ">"
+ "</"
+ "/>"
+ ] @tag.delimiter
--- a/queries/vue/injections.scm
+++ b/queries/vue/injections.scm
@@ -1,0 +1,19 @@
+((style_element
+  (raw_text) @css))
+
+; There's no queries for less, sass nor postcss.
+
+((script_element
+  (raw_text) @javascript))
+
+(
+  (script_element
+    (start_tag
+      (attribute
+        (quoted_attribute_value (attribute_value) @_lang)))
+    (raw_text) @typescript)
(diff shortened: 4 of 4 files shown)


WhyNotHugo:
I tried adding folds by including this in `vue/folds.scm`:

```
[
  (template_element)
  (script_element)
  (style_element)
] @fold
```

But Vim did not seem to pick this up. I don't actually use folds myself at all, so maybe I just mistested.

theHamsta:
You have to explicitly set the fold method:
```viml
set foldmethod=expr foldexpr=nvim_treesitter#foldexpr()
```
or in lua
```lua
  vim.cmd("set foldmethod=expr foldexpr=nvim_treesitter#foldexpr()")
```

WhyNotHugo:
Duh, right, I never set it up!

Updated the PR with folding changes too. I personally don't use it, but it's low-effort and will likely be useful to someone eventually.

elianiva:
You might want to check out my Svelte queries for injecting Javascript or Typescript based on the `lang` property on the `script` tag. But I still think it's weird that Typescript/Javascript has different queries since the former is a superset of the latter.

WhyNotHugo:
> But I still think it's weird that Typescript/Javascript has different queries since the former is a superset of the latter.

I agree. I guess the big difference is that TypeScript-specific syntax would throw an error if the content is supposed to be JS.

Regrettably, we have no queries for any of the supported style languages, so I've just left that as CSS until those are present (e.g.: `sass`, `less`, `postcss`, etc).

WhyNotHugo:
Any news here? AFAIK, there's nothing missing on my end here.

theHamsta:
Sorry for the delay! I'm quite busy atm. From my side this PR looks good. But I have no experience with web development. It would be great if @elianiva or @steelsojka  could also have a look.

theHamsta (approved):

elianiva:
might be worth to add `element` to fold and group the `@tag` queries like this

```clojure
[
  (component)
  (template_element)
  (start_tag) 
  (directive_attribute) 
  (directive_dynamic_argument) 
  (directive_dynamic_argument_value) 
  (component) 
  (end_tag) 
] @tag

```

Everything else looks good to me.

steelsojka:
My only concern is with the injections. I believe it would inject
JavaScript and typescript for the blocks for the same node. I will have to
look at the injection code in nvim to confirm that, but that can be changed
after merging if need be.

WhyNotHugo:
@elianiva Makes sense. I've applied those suggestions and also grouped the `@method ` ones.

Would you prefer I rebase / squash a bit?

WhyNotHugo:
I believe all concerns have been resolved.

theHamsta:
Nice! Thanks for the contribution!

cloggier:
@WhyNotHugo Thank you so much for this!

PR #940, opened 2021-02-10
Fix nix compilation

theHamsta:
This is the update-lockfile-pr

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -3,10 +3,10 @@
     "revision": "8ece09ca4c0b5e59b124cd19fa92c76b1a9e9dd4"
   },
   "c": {
-    "revision": "80b6b748e612d5831a3d5fe296c5a39b95ecb7bb"
+    "revision": "8d4631b673f8a9d968d1d93393c0a965194abac0"
   },
   "c_sharp": {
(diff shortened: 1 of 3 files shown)


theHamsta:
@leo60228 there might be more things that have changed in the nix parser. Most of them are just fixes and additional fields.

leo60228:
It looks like `tree-sitter-highlight` queries were added upstream. Do we have some way to make use of those?

theHamsta:
I had the idea that we could have a tag in the parser config to autoimport the queries. Other's had the opinion to host the nvim queries here to avoid confusion with other consumers of the parsers

vigoux (approved):

cstrahan:
Feel free to send any change requests upstream! :slightly_smiling_face: 

I didn't realize anyone was using `tree-sitter-nix` yet -- hope the recent changes haven't caused too much trouble. Glad to see it's of use! I write `tree-sitter-nix` under the premise that it would enable others to do neat things, though I never had a particular use in mind (yet, anyway); as such, it's really important to me that `tree-sitter-nix` be suitable for everyone else's purposes -- it's you all that I wrote it for, after all :slightly_smiling_face:.

PR #941, opened 2021-02-10
Remove debug print statement

theHamsta:
The message does not even provide any context where it's coming from
and what it is exactly doing and has to be acknowledged when printed in
conjunction with the "wrote xxx bytes message"


vigoux (approved):

vigoux:
Could you try rebasing on master?

theHamsta:
The trick is close and reopen ;-). Because it already was completely rebased.

PR #953, opened 2021-02-15
Add support for JSONC

WhyNotHugo:
`jsonc` is basically JSON with inline comments. It's used a lot for configuring desktop apps, including `waybar`, `coc`, `vscode` and others.

I've written a parser which is about 95% based on the `json` parser, with comment support added on top. The parser itself has some tests, as well as some manual testing on file I have lying around.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -303,6 +303,13 @@ list.vue = {
   }
 }
 
+list.jsonc = {
+  install_info = {
+    url = "https://gitlab.com/WhyNotHugo/tree-sitter-jsonc.git",
+    files = { "src/parser.c" },
+  }
+}
+
 list.elm = {
   install_info = {
     url = "https://github.com/elm-tooling/tree-sitter-elm",
--- a/queries/jsonc/folds.scm
+++ b/queries/jsonc/folds.scm
@@ -1,0 +1,1 @@
+; inherits: json
--- a/queries/jsonc/highlights.scm
+++ b/queries/jsonc/highlights.scm
@@ -1,0 +1,3 @@
+; inherits: json
+
+(comment) @comment
--- a/queries/jsonc/indents.scm
+++ b/queries/jsonc/indents.scm
@@ -1,0 +1,1 @@
+; inherits: json
--- a/queries/jsonc/locals.scm
+++ b/queries/jsonc/locals.scm
@@ -1,0 +1,1 @@
+; inherits: json


stsewd:
Looks like there is a way to inherit from other grammars

https://github.com/tree-sitter/tree-sitter-cpp/blob/a35a275df92e7583df38f2de2562361f2b69987e/grammar.js#L1

https://github.com/tree-sitter/tree-sitter-cpp/blob/a35a275df92e7583df38f2de2562361f2b69987e/grammar.js#L9

So you don't have to update yours on each change.

stsewd (approved):

WhyNotHugo:
Oh, nice, thanks for pointing that out!

I've updated my tree-sitter to extend that one and dropped all the copy pasta.

image

stsewd:
Going to force a re-check to check with the new update

theHamsta:
I guess pinning the parser version does not work with gitlab (the PR adding this feature for git is not merged). Maybe you could somehow deduce what the tar link for gitlab is?  Maybe they have the same folder structure for their tar archives as Github has.

WhyNotHugo:
@theHamsta Sorry, I don't quite understand what's failing.

theHamsta:
Try to set an old revision of your parser in lockfile.json to see whether it can retrieve it (probably will fall back to git installation from master and not the pinned version). Our default Installation method is downloading a versioned tar from Github ( url is hardcoded for Github).

We are pinning parsers to protect users from potential incompatible tree-sitter versions or incompatible queries.

Since the json parser will probably not change any soon we might  not need that feature for jsonc.

WhyNotHugo:
Oooh! I hadn't even noticed that lockfile. Shall I also pin the `jsonc` parser? I don't see it changing much anytime soon though, especially not in incompatible ways.

> Our default Installation method is downloading a versioned tar from Github ( url is hardcoded for Github).

Oh, that explains a lot of issues I had setting up git URLs and stuff. Wouldn't it make more sense to actually use git under the hood, rather than something github-specific?

In any case, any commit previous can be fetched via:

- https://gitlab.com/WhyNotHugo/tree-sitter-jsonc/-/archive/master/tree-sitter-jsonc-master.tar
- https://gitlab.com/WhyNotHugo/tree-sitter-jsonc/-/archive/14ff22d8e4856e690a85b3a7a225211533c0dadf/tree-sitter-jsonc-master.tar

Would you rather I point to the URL to the tar file of a specific version?

theHamsta:
> Oh, that explains a lot of issues I had setting up git URLs and stuff. Wouldn't it make more sense to actually use git under the hood, rather than something github-specific?

Please don't start that discussion again! Git is still supported as a alternative installation method. Both should work. We switched to tar because git does not support shallow checkouts of a non-tree revision. There is also a open pull request to make Git support the checkout after clone (but without shallow cloning. So you have to wait a bit longer to clone).

The pinning is done automatically by a CI job. So you don't have to do anything.

theHamsta:
I guess we can tackle the lockfile issue still later.

WhyNotHugo:
Thanks!

> Please don't start that discussion again!

Sorry to hit a never here! I'll check out the mentioned PR.

theHamsta:
Both works: Git and tar. So it should be possible for everyone to install it.

lua/nvim-treesitter/parsers.lua
```diff
@@ -303,6 +303,13 @@ list.vue = {
   }
 }
 
+list.jsonc = {
+  install_info = {
+    url = "https://gitlab.com/WhyNotHugo/tree-sitter-jsonc.git",
+    files = { "src/parser.c" },
```
stsewd, line 309: You should add yourself as maintainer of the jsonc with `maintainers = {"@foo"}` so it's listed as maintained in the readme :)

theHamsta: Also `  readme_name = "JSON with comments",` would be helpful to explain what `jsonc` is.

WhyNotHugo: Done here: https://github.com/nvim-treesitter/nvim-treesitter/pull/957

David-Else:
@WhyNotHugo Hi, how do I actually use a JSONC file? If I create a `.json` with comments and I am using the JSON LSP (https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#jsonls)  it says:
> comments are not permitted in JSON

It is the VS Code LSP which does allow JSONC (somehow in VS Code), not sure how to proceed. Cheers!

theHamsta:
Of i remember correctly there is a config key for the lap to allow comments (can check in an hour or so).

For the highlight one has to switch the ft to jsonc or to change the parsers mapping to use jsonc parser for json

theHamsta:
I did not find a documentation. But Google has cached this piece of text

> The JSON language server supports requests on documents of language id json and jsonc . json documents are parsed and validated following the JSON specification. jsonc documents additionally accept single line ( // ) and multi-line comments ( /* ... */ ).

So it might be a contribution neovim/lsp-config to use jsonc when the filetype is set to jsonc.

WhyNotHugo:
Vim is auto-detecting your file as `json`, instead of `jsonc` (likely due to the extension?). Hence, you're getting the wrong highlighting and the lsp thinks this is `json`.

> Of i remember correctly there is a config key for the lap to allow comments (can check in an hour or so).

Add this to tell vim that the file is `jsonc`.

    // vim: ft=jsonc

David-Else:
Thanks! `// vim: ft=jsonc` works.

Can I add something to my config to say all JSON files are JSONC? That would make it like VS Code, and I can't think of any disadvantage other than allowing comments in all JSON files which doesn't bother me.

theHamsta:
Well, this sets the filetype always to jsonc. But does the LSP still get activated?
```
    au! BufRead,BufNewFile *.json set filetype=jsonc
```

> allowing comments in all JSON files which doesn't bother me

It's non-standard and some parsers could just reject the json if it has comments. It's sad that comments are not officially allowed.

WhyNotHugo:
> I can't think of any disadvantage other than allowing comments in all JSON files which doesn't bother me.

Application that expect a JSON file will fail in you provide them with a JSONC file. Just to be very clear `json` and `jsonc` are different formats, with different rules. `json` **does not** allow comments.

Whatever application is using `.json` but expects `jsonc` is being really confusing -- it's like asking for `config.js` but allowing typescript code inside of it

David-Else:
Looks like we need a `.jsonc` format, things are confusing with the same file extension. I will just stick to using the comment to activate JSONC, thanks to everyone for getting it included!

WhyNotHugo:
There is a `jsonc` format. I've some files named just `config` that are jsonc, and neovim recognises them fine and sets`ft=jsonc` by itself.

However, if the file is `something.json`, neovim will assume it's a json file, **not** json (and with good reason). In that case, you should leave the above comment. You can also tell vim to recognise filetypes based on their path (e.g.: `.config/myapp/*.json`).

PR #962, opened 2021-02-19
docs: add additional steps for Fedora/CentOS

### LINKED ISSUE #961: Installing nvim-treesitter on Fedora fails compiling
Same error as here: https://www.reddit.com/r/neovim/comments/jbsfx2/cant_install_treesitter_error_during_compilation/
Solved by
```bash
sudo dnf install gcc-c++
```
I'll fork the repo and add that to documentation.

gbaranski:
closes #961

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ For more detailed information on setting these up, see ["Advanced setup"](#advan
 - Neovim [nightly](https://github.com/neovim/neovim#install-from-source)
 - `tar` and `curl` in your path (or alternativly `git`)
 - A C compiler in your path and libstdc++ installed ([Windows users please read this!](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Windows-support)).
+- For Fedora/CentOS users, install `gcc-c++`, [more info](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Linux-Support#961-error-during-compilation)
 
 ## Installation
 


theHamsta:
I don't think that the hint specific for Fedora is very helpful since most people will ask: then, what do I have to do for Ubuntu? Probably, better when you change it as follows
```
- A C compiler in your path and libstdc++ installed ([Windows users please read this!](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Windows-support)).
```
to 
```
- A C compiler in your path ([Windows users please read this!](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Windows-support)).
- libstdc++ installed (`sudo apt install g++`, ` sudo dnf install gcc-c++`) 
```

theHamsta:
@gbaranski you don't want to pursue this PR further? You had a good point explaining this, but we should add some hint that is not specific to Fedora but also holds for Ubuntu, Windows, MacOS

gbaranski:
> @gbaranski you don't want to pursue this PR further? You had a good point explaining this, but we should add some hint that is not specific to Fedora but also holds for Ubuntu, Windows, MacOS

what do you think about creating new wiki page called `Platform specfic` which will contain any platform specific notes?

theHamsta:
Until now we only had platform-specific hints for Windows (which is the only platform having a different install logic in nvim-treesitter). The danger with not having such a small detail not in the README is that most people will not read it. Maybe have the requirement in the README and linked platform specific instructions in wiki right after the requirement?
```
- libstd++ install (How to install?)
```

libstdc++ is only needed for some parsers that use C++ due to the link flag `-lstdc++`. Wouldn't it be enough to make libstdc++ a separate bullet point and name the RPM and DEB package names in parenthesis (`gcc-cpp` for RPM-based distributions (Fedora/CentOS), `g++` on Ubuntu/Debian). For Windows stdc++ is included in MinGW or llvm as well as for Mac users in `brew install llvm` or `brew install gcc`

PR #965, opened 2021-02-20
feat: add java indent

elianiva:


kyazdani42:
@p00f could you check that this is right ?

p00f:
sure just a sec

p00f:
for argument list i couldn't find an authoritative style guide but shouldn't it be in one line?
image
The rest is nice!

elianiva:
I don't know about that either :laughing:
I just add the indent query because I recently started writing java at my school so I still don't know much about it.

p00f:
this is better than 
image
though, so I'm fine

p00f (approved):

queries/java/indents.scm
```diff
@@ -0,0 +1,19 @@
+[
+  (class_declaration)
+  (class_body)
+  (constructor_declaration)
+  (constructor_body)
+  (block)
+  (switch_block)
+  (array_initializer)
+  (argument_list)
```
p00f, line 9:
```suggestion
  (argument_list)
  (formal_parameters)
```
Indentation of formal_parameters behaves the same way (not ideal that is), but better than nothing

p00f (changes requested):

queries/java/indents.scm
p00f: @theHamsta is the cpp maintainer, can you shed light?

p00f:
Same with array_initializer
image

p00f (approved):

elianiva:
I have a question, not related to indent but still related to Java. Do we want to inject jsdoc for java comments? They both are very similar, but I'm not sure if it's a good idea.

kyazdani42 (approved):

kyazdani42:
@elianiva i don't know java much, but if they do not use jsdoc, i don't think we should enable injection for it :)

elianiva:
That makes sense, I'll just do it locally then.

theHamsta:
Maybe this could work. Jsdoc and javadoc pretty much look the same.

PR #969, opened 2021-02-22
chore: rewrite healthcheck

### LINKED ISSUE #908: make checkhealth output more compact
I think treesitter helpers could be clearer than they are and less verbose:

For instance TSInstallInfo doesn't sort the grammars so it's difficult to look for a specific grammar. 
Similarly `:checkhealth nvim_treesitter` output could be more compact. checkhealth should let you see in a glance what's wrong, the more output, the harder it gets to parse it.

Here 2 examples:
```
  - ERROR: `cc` executable not found.
    - ADVICE:
      - Install `gcc` with your package manager.
      - Install `clang` with your package manager.
      - Check that your `$PATH` is set correctly.
```
the 3 lines could be arranged as `check that either gcc or clang is in your $PATH`.

```
## lua parser healthcheck
  - OK: lua parser found.
  - OK: `highlights.scm` found.
  - OK: `locals.scm` found.
  - OK: `folds.scm` found.
  - OK: `indents.scm` found.
```
I have a feeling the first   `- OK: lua parser found.` is obvious since otherwise there would be no healthcheck for this parser ?
The rest could be written as a one-liner  ``` `highlights.scm`, `locals.scm` `indents.scm` found.```

Likewise I find TSModuleInfo hard to read since the difference between the cross and the tick is quite small. Maybe remove crosses to make it easier to see ?

Also as treesitter is supposed to have some native support in neovim, it would be nice to split the healthcheck into 2 parts and upstream a very basic healtcheck into the core.

Sorry for the complaints, loving what you do keep up the excellent work.

kyazdani42:
fixes #908

### DIFF
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -60,7 +60,8 @@ function M.checkhealth()
     end
   end
   print([[ Legend: H[ighlight], L[ocals], F[olds], I[ndents]
-         *) multiple parsers found, only one will be used]])
+         *) multiple parsers found, only one will be used
+         x) errors found in the query, try to run :TSUpdate {lang}]])
 end
 
 return M
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -6,7 +6,6 @@ local info = require'nvim-treesitter.info'
 
 local health_start = vim.fn["health#report_start"]
 local health_ok = vim.fn['health#report_ok']
-local health_warn = vim.fn['health#report_warn']
 local health_error = vim.fn['health#report_error']
 
 local M = {}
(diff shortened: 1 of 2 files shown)


theHamsta (approved):
Nice!

lua/nvim-treesitter/health.lua
```diff
@@ -24,51 +23,44 @@ local function install_health()
...
 
-local function query_health(lang, query_group)
+local function query_status(lang, query_group)
   local ok, found_query = pcall(queries.get_query, lang, query_group)
   if not ok then
-    health_error("Error in `"..query_group..".scm` query found for "..lang..'\n'..found_query, {
-      "Try `:TSUpdate "..lang.."` (queries might have been adapted for a upstream parser change)",
-      "Try to find the syntax error/invalid node type in above file."
-    })
+    return "x"
```
theHamsta, line 36: legend for `x` would be cool

theHamsta (approved):

PR #970, opened 2021-02-22
Chore: ident files and remove some useless code

kyazdani42:
This PR is not that useful but at least code is indented properly.
I also removed some code in `parsers.lua` that wasn't needed/used anymore.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -17,16 +17,12 @@ local list = setmetatable({}, {
 
     rawset(table, parsername, setmetatable(parserconfig, {
       __newindex = function(parserconfigtable, key, value)
-        if key == "used_by" then
-          ft_to_parsername[value] = parsername
-        else
-          rawset(parserconfigtable, key, value)
-        end
+        rawset(parserconfigtable, key, value)
       end
     }))
 
     update_ft_to_parsername(parsername, parserconfig)
-   end
+  end
 })
 
 list.javascript = {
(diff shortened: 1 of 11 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -17,16 +17,12 @@ local list = setmetatable({}, {
 
     rawset(table, parsername, setmetatable(parserconfig, {
       __newindex = function(parserconfigtable, key, value)
-        if key == "used_by" then
-          ft_to_parsername[value] = parsername
```
theHamsta, line 21: why is `used_by` not needed anymore?

lua/nvim-treesitter/parsers.lua
```diff
@@ -454,20 +450,6 @@ function M.ft_to_lang(ft)
   return ft_to_parsername[ft] or ft
 end
 
-function M.lang_to_ft(lang)
```
theHamsta, line 457: this code is used

lua/nvim-treesitter/parsers.lua
kyazdani42: i believe used_by is already set after the rawset call, removing this code did not affect highlighting in jsx files so i believed it was ok to remove it.

theHamsta: It is set in the update_ft_to_parser name that is called two lines after the removed code.

theHamsta (approved):

PR #973, opened 2021-02-23
feat: indent additions

oberblastmeister:
- added indent for query files
- indent for `rust` match
- toml indent fixes

### DIFF
--- a/queries/query/indents.scm
+++ b/queries/query/indents.scm
@@ -1,0 +1,12 @@
+[
+  (list)
+] @indent
+
+[
+  (list)
+]
+
+[
+  "["
+  "]"
+] @branch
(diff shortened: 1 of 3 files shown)


queries/query/indents.scm
```diff
@@ -0,0 +1,12 @@
+[
+  (list)
+] @indent
+
+[
+  (list)
+]
```
kyazdani42, line 7: duplicate entry ?

kyazdani42 (changes requested):

queries/query/indents.scm
oberblastmeister: woops

kyazdani42 (approved):

vigoux (approved):

PR #980, opened 2021-02-26
Add CI and Windows support for cl.exe

JoeyGrajciar:
I tried #347 and changes were looking fine(`cd` does not like `;`), but since the branch was too old I decided to do it this way.

Changes are working tested on windows machine. Also to make this work `nvim` has to be started from `Developer Command Prompt` (`cl.exe` has some environmental needs)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,53 @@
+name: Parser compilation and tests
+
+on: [push, pull_request]
+
+jobs:
+  check_compilation_unix_like:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        cc: [ gcc, clang ]
+        exclude:
+          - os: macos-latest
+            cc: gcc
+
+        include:
+          - os: windows-latest
+            cc: cl
+
+          - os: macos-latest
+            cc: gcc-10
+
+    name: Parser compilation
+    runs-on: ${{ matrix.os }}
+    env:
+      CC: ${{ matrix.cc }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ilammy/msvc-dev-cmd@v1.5.0
+
+      - name: Install and prepare Neovim
+        run: |
+          bash ./scripts/ci-install-${{ matrix.os }}.sh
+
+      - name: Compile parsers Unix like
+        if: matrix.os != 'windows-latest'
+        run: |
+          nvim --headless -c "TSInstallSync all" -c "q"
+
+      - name: Compile parsers Windows
+        if: matrix.os == 'windows-latest'
+        run: |
+          C:\\tools\\neovim\\Neovim\\bin\\nvim.exe --headless -c "TSInstallSync all" -c "q"
+
+      - name: Post compile Windows
+        if: matrix.os == 'windows-latest'
+        run: cp -r ~/AppData/Local/nvim/pack/nvim-treesitter/start/nvim-treesitter/parser/* parser
+        shell: bash
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: parsers-${{ matrix.os }}-${{ matrix.cc }}-x86_64
+          path: parser/*
+
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -11,7 +11,7 @@ local shell = require'nvim-treesitter.shell_command_selectors'
 local M = {}
 local lockfile = {}
 
-M.compilers = { vim.fn.getenv('CC'), "cc", "gcc", "clang" }
+M.compilers = { vim.fn.getenv('CC'), "cc", "gcc", "clang", "cl" }
 
 local started_commands = 0
 local finished_commands = 0
@@ -93,13 +93,6 @@ function M.iter_cmd(cmd_list, i, lang, success_message)
 end
 
 local function get_command(cmd)
-  local ret = ''
-  if cmd.opts and cmd.opts.cwd then
-    ret = string.format('cd %s;\n', cmd.opts.cwd)
-  end
-
-  ret = string.format('%s%s ', ret, cmd.cmd)
-
   local options = ""
   if cmd.opts and cmd.opts.args then
     for _, opt in ipairs(cmd.opts.args) do
@@ -107,7 +100,11 @@ local function get_command(cmd)
     end
   end
 
-  return string.format('%s%s', ret, options)
+  local final = string.format('%s %s', cmd.cmd, options)
+  if cmd.opts and cmd.opts.cwd then
+    final = shell.make_directory_change_for_command(cmd.opts.cwd, final)
+  end
+  return final
 end
 
 local function iter_cmd_sync(cmd_list)
@@ -151,7 +148,8 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync, 
   if from_local_path then
     compile_location = repo.url
   else
-    compile_location = cache_folder..path_sep..(repo.location or project_name)
+    local repo_location = string.gsub(repo.location or project_name, '/', path_sep)
+    compile_location = cache_folder..path_sep..repo_location
   end
   local parser_lib_name = install_folder..path_sep..lang..".so"
 
@@ -192,7 +190,7 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync, 
       info = 'Compiling...',
       err = 'Error during compilation',
       opts = {
-        args = vim.tbl_flatten(shell.select_compiler_args(repo)),
+        args = vim.tbl_flatten(shell.select_compiler_args(repo, cc)),
         cwd = compile_location
       }
     },
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -53,20 +53,31 @@ function M.select_executable(executables)
   return vim.tbl_filter(function(c) return c ~= vim.NIL and fn.executable(c) == 1 end, executables)[1]
 end
 
-function M.select_compiler_args(repo)
-  local args = {
-    '-o',
-    'parser.so',
-    '-I./src',
-    repo.files,
-    '-shared',
-    '-Os',
-    '-lstdc++',
-  }
-  if fn.has('win32') == 0 then
-    table.insert(args, '-fPIC')
+function M.select_compiler_args(repo, compiler)
+  if (string.match(compiler, 'cl$') or string.match(compiler, 'cl.exe$')) then
+    return {
+      '/Fe:',
+      'parser.so',
+      '/Isrc',
+      repo.files,
+      '-Os',
+      '/LD',
+    }
+  else
+    local args = {
+      '-o',
+      'parser.so',
+      '-I./src',
+      repo.files,
+      '-shared',
+      '-Os',
+      '-lstdc++',
+    }
+    if fn.has('win32') == 0 then
+     table.insert(args, '-fPIC')
+    end
+    return args
   end
-  return args
 end
 
 function M.select_install_rm_cmd(cache_folder, project_name)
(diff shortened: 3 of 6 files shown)


theHamsta:
Just asking: is the a possibility that we can test this functionality on CI? 

I've started to add Windows to the CI pipelines (but with clang). It would be interesting for us to see whether the parsers compile fine on `cl.exe` or with `llvm` on Windows.

https://github.com/nvim-treesitter/nvim-treesitter/pull/417/files

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -172,4 +183,20 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
   end
 end
 
+function M.select_temp_directory_change_command_start(dir)
```
theHamsta, line 186: is there maybe a more pregnant same for this? `select_push_dir_command`/``select_pop_dir_command`

lua/nvim-treesitter/install.lua
```diff
@@ -107,7 +106,11 @@ local function get_command(cmd)
     end
   end
 
-  return string.format('%s%s', ret, options)
+  local final = string.format('%s%s', ret, options)
+  if cmd.opts and cmd.opts.cwd then
+    final = final .. shell.select_temp_directory_change_command_end()
+  end
+  return final
```
theHamsta, line 113: It would be probably better to just have one function to set the temporary directory that can be called at the end of this function to surround the actual command.

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -172,4 +183,20 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
...
+  if fn.has('win32') == 1 then
+    return string.format("pushd %s & ", dir)
```
theHamsta, line 188: Will this always work on win32? Even when using bash?

lua/nvim-treesitter/install.lua
JoeyGrajciar: yeah you are right :) it will be nicer :)

lua/nvim-treesitter/shell_command_selectors.lua
JoeyGrajciar: hmm I guess not. this one is tricky

JoeyGrajciar:
CI should be possible I guess :) should I add it in this PR?

lua/nvim-treesitter/shell_command_selectors.lua
theHamsta: Maybe Developer Command Prompt sets some env variables? Does something like `$SHELL` exists on Windows.

theHamsta:
> CI should be possible I guess :) should I add it in this PR?

If have time for it you can include it in this PR. But CI usually involves more efforts than one might think beforehand. If you think it will be too much work we can also include it in a follow-up PR. Windows CI would help us (all maintainers are Linux or sometimes Mac users) and Windows users a lot since sometimes Windows users submit patches but they will not work when other users use another shell/compiler combination.

lua/nvim-treesitter/shell_command_selectors.lua
JoeyGrajciar: hmm so `pushd` and `popd` do work locally on Mac, but not on CI in ubuntu.

JoeyGrajciar: checked those commands in Git Bash and they should work. was able to compile `cpp` with clang.

JoeyGrajciar:
> If have time for it you can include it in this PR.

Will check this later today.

theHamsta:
image
Classic CI work :smile: . So glad that I'm not an dev opt engineer.

.github/workflows/test.yml
```diff
@@ -0,0 +1,20 @@
...
+      - uses: actions/checkout@v2
+      - uses: ilammy/msvc-dev-cmd
+
+      - name: Install and prepare Neovim
+        run: |
+          bash ./scripts/ci-install-windows-latest.sh
+          echo "C:\\tools\\neovim\\Neovim\\bin\\" >> $GITHUB_PATH
+
+      - name: Compile parsers
+        run: nvim --headless -c "TSInstall all" -c "sleep 120" -c "q"
```
theHamsta, line 19: Does `:TSSync all` work? Then `sleep` would not be necessary.

.github/workflows/test.yml
```diff
@@ -0,0 +1,20 @@
...
+    name: Parser compilation using MSVC cl.exe
+    runs-on: 'windows-latest'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ilammy/msvc-dev-cmd@v1.5.0
+
+      - name: Install and prepare Neovim
+        run: |
+          bash ./scripts/ci-install-windows-latest.sh
+          $env:Path += "C:\\tools\\neovim\\Neovim\\bin\\"
```
theHamsta: maybe it requires a `refreshenv`?

.github/workflows/test.yml
JoeyGrajciar: apparently it works :)

.github/workflows/test.yml
JoeyGrajciar: currently went with complete path to `nvim.exe`, but seems there are some failures for some languages and also `gcc` appeared in the wild :)

JoeyGrajciar: okay so issue was with those "special" ones with location specified in parser config

theHamsta:
Looks nice! Do the uploaded parsers from CI work on you machine? In the past some users said that the shared libraries would not be accepted by their OS when loading.

JoeyGrajciar:
> Do the uploaded parsers from CI work on you machine?

what uploaded parsers?

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -172,4 +183,12 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
   end
 end
 
+function M.make_directory_change_for_command(dir, command)
+  if fn.has('win32') == 1 then
+    return string.format("pushd %s & %s & popd", dir, command)
```
theHamsta, line 188: We did not solve that issue yet, whether to use pushd or cd on Windows, didn't we?

JoeyGrajciar: `cd %s;\n` does not work on Windows :) It does not like that `;`

JoeyGrajciar: I mean now we can test it quite easily in CI :)

JoeyGrajciar:
Should I try to add also macOS CI?

theHamsta:
This will upload the compiled parsers when CI finishes
```
      - uses: actions/upload-artifact@v2
        with:
          name: parsers
          path: parser/*
```

theHamsta:
I think MacOS is not so different from Linux. So there is no urgent need to add it but it won't harm.

JoeyGrajciar:
Autosquash history, renamed CI commit (its not windows only anymore) :)
Will try to test those artifacts today.

lua/nvim-treesitter/shell_command_selectors.lua
theHamsta: But bash on windows should behave like bash on Linux?

JoeyGrajciar: you mean running this installation command from `bash` in CI?

theHamsta: I mean when users on windows use bash instead of powershell. Will they have a problem? I can't really tell what's the common eniviroment for running nvim on windows (probably neovim-qt) but users seem to have managed to install parsers also with cd;

JoeyGrajciar: I mean probably most of the shell commands which do check `win32` will not work on bash.

JoeyGrajciar:
> In the past some users said that the shared libraries would not be accepted by their OS when loading.

gcc on windows for some reason generates some libraries with dependencies on `libstdc++-6.dll` and `libgcc_s_seh-1.dll`
other issue might be trying to load x86_64 on x86 system and vice versa (same for arm)

theHamsta:
> gcc on windows for some reason generates some libraries with dependencies on libstdc++-6.dll and libgcc_s_seh-1.dll
other issue might be trying to load x86_64 on x86 system and vice versa (same for arm)

@JoeyGrajciar  I don't think that this was the reason. The binaries depend on libstdc++ due to the linker flag `-lstdc++`. We need this flag because some scanners use C++ for `scanner.cc`. `cl.exe` made historically never a real difference between C and C++

theHamsta (approved):

JoeyGrajciar:
> The binaries depend on libstdc++ due to the linker flag -lstdc++. We need this flag because some scanners use C++ for scanner.cc. cl.exe made historically never a real difference between C and C++

I meant that some parsers built on windows-latest with gcc are not getting loaded on windows due to those dependencies.

theHamsta:
Probably we can update the Windows wiki entry a bit after merging this. If someone cannot get the parsers compiled at least they can download them from CI.

JoeyGrajciar:
Okay so after some playing with that compilation locally I found the issue :)
gcc that is available during compilation (PATH) on `windows-latest` is from Mingw 8.1.0 and it creates these dependencies
image

After upgrading mentioned Mingw to latest 10.2.0 it looks like this for the same file
image

information about them is the key
```
gcc -v
Using built-in specs.
COLLECT_GCC=C:\ProgramData\Chocolatey\lib\mingw\tools\install\mingw64\bin\gcc.exe
COLLECT_LTO_WRAPPER=C:/ProgramData/Chocolatey/lib/mingw/tools/install/mingw64/bin/../libexec/gcc/x86_64-w64-mingw32/8.1.0/lto-wrapper.exe
Target: x86_64-w64-mingw32
Configured with: ...
Thread model: posix
gcc version 8.1.0 (x86_64-posix-seh-rev0, Built by MinGW-W64 project) 
```

vs

```
gcc -v
Using built-in specs.
COLLECT_GCC=C:\ProgramData\chocolatey\lib\mingw\tools\install\mingw64\bin\gcc.exe
COLLECT_LTO_WRAPPER=c:/programdata/chocolatey/lib/mingw/tools/install/mingw64/bin/../libexec/gcc/x86_64-w64-mingw32/10.2.0/lto-wrapper.exe
Target: x86_64-w64-mingw32
Configured with: ...
Thread model: win32
Supported LTO compression algorithms: zlib
gcc version 10.2.0 (GCC)
```

AFAIK they are not going to upgrade this Mingw to latest version. Instead there is Msys2 installed which also contains gcc.
Will investigate option to use that one instead or we could try to install latest Mingw as part of setup.

theHamsta:
@JoeyGrajciar I think it's ok when artifacts from cl.exe are working or gcc/clang builds locally

JoeyGrajciar:
@theHamsta do you mean that I should disable creation of artifact for gcc on `windows-latest`?

theHamsta:
@JoeyGrajciar you could, but I would just link to the CL.exe binaries in the wiki. I'd say everyone using the artifacts is at their own risk.

theHamsta:
I fear that the MacOS brew method might be broken because tree-sitter has been updated to 0.19.0. Could merge since the button is still green 😄

JoeyGrajciar:
> I fear that the MacOS brew method might be broken because tree-sitter has been updated to 0.19.0. Could merge since the button is still green

for now it is good 😄 rebased on top of master and looks fine even though it was downloading 0.18.2

theHamsta:
@vigoux @kyazdani42 any comments to the new cross-platform CI? It's a lot of jobs but now every build configuration is tested. Great work to have our plugin work on all platforms with all compilers!

JoeyGrajciar:
> It's a lot of jobs but now every build configuration is tested.

That is very bold statement 😁it is probably most used common combinations 🙂

kyazdani42 (approved):
good job indeed

theHamsta:
@JoeyGrajciar can you update the wiki with whatever magic that was necessary to make this work and how to download parsers from CI?

JoeyGrajciar:
@theHamsta wiki updated.

PR #985, opened 2021-02-28
Support the glimmer parser when the `hbs` template literal tag is used

NullVoxPopuli:
Idk if this is correct, but I know that gql usage is similar:
```js
const query = gql`
 query stuff
`;
```
so I copy-pasted that for glimmer, which looks like

```js
cost template = hbs`
  <glimmer-stuff />
  <html-like />
`;  
```

Here is the WIP syntax support for glimmer: https://github.com/alexlafroscia/tree-sitter-glimmer

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -296,6 +296,7 @@ list.glimmer = {
   install_info = {
     url = "https://github.com/alexlafroscia/tree-sitter-glimmer",
     files = { "src/parser.c", "src/scanner.c" },
+    branch = 'main',
   },
   readme_name = "Glimmer and Ember",
   maintainers = { "@alexlafroscia" },
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -292,6 +292,16 @@ list.toml = {
   maintainers = {"@tk-shirasaka"},
 }
 
+list.glimmer = {
+  install_info = {
+    url = "https://github.com/alexlafroscia/tree-sitter-glimmer",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  readme_name = "Glimmer and Ember",
+  maintainers = { "@alexlafroscia" },
+  filetype = "handlebars"
+}
+
 list.vue = {
   install_info = {
     url = "https://github.com/ikatyang/tree-sitter-vue",
--- a/queries/glimmer/highlights.scm
+++ b/queries/glimmer/highlights.scm
@@ -14,7 +14,7 @@
 ; == Mustache Statements ===
 
 ; Hightlight the whole statement, to color brackets and separators
-(mustache_statement) @tag
+(mustache_statement) @tag.delimiter
 
 ; Generic identifiers are variables
 (identifier) @variable
@@ -33,4 +33,4 @@
   ">"
   "</"
   "/>"
-] @tag
+] @tag.delimiter
--- a/queries/glimmer/highlights.scm
+++ b/queries/glimmer/highlights.scm
@@ -1,3 +1,7 @@
+;; By @alexlafroscia (https://github.com/alexlafroscia/tree-sitter-glimmer)
+;; Licensed under MIT
+;; Modified for nvim-treesitter
+
 ; Tags that start with a lower case letter are HTML tags
 ((tag_name) @tag
   (#match? @tag "^[a-z]"))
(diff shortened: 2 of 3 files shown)


theHamsta:
Do you want to add the glimmer parser, too (`parsers.lua`)? You would also need to add glimmer highlights to make it work. Probably, you can copy most stuff from html (github.com/nvim-treesitter/playground is useful for that).

NullVoxPopuli:
Just added an attempt at the parser -- hopefully it works? I'm not quite set up to debug tree-sitter itself quite yet.

Also, is there a way to upstream this sort of stuff to tree-sitter? shouldn't tree-sitter has these types of configs? rather than nvim-treesitter? (cause then every editor plugin needs the same customizations?)

theHamsta:
@NullVoxPopuli every editor (Atom, Neovim, Emacs) that uses tree-sitter has currently it's own queries they all highlight different things and have different queries implemented. Maybe we will have some Atom-style query fallback at some time but not at the moment. Neovim and Emacs basically forked because the tree-sitter ones very too basic and didn't map well to the editor's highlight groups.

theHamsta:
> Just added an attempt at the parser -- hopefully it works? 

You can test it with https://github.com/nvim-treesitter/playground

NullVoxPopuli:
I mean, the parser works, the stuff in question is just what's in _this_ PR, #985

theHamsta (approved):

lua/nvim-treesitter/parsers.lua
```diff
@@ -292,6 +292,16 @@ list.toml = {
   maintainers = {"@tk-shirasaka"},
 }
 
+list.glimmer = {
+  install_info = {
+    url = "https://github.com/alexlafroscia/tree-sitter-glimmer",
+    files = { "src/parser.c", "src/scanner.c" },
```
theHamsta, line 298:
```lua
    branch = 'main',
```
does not work without setting the branch to main

theHamsta, line 298:
```lua
    branch = 'main',
```
does not work without setting the branch to main

theHamsta (changes requested):
Branch needs to be set to `main`

theHamsta:
Seems to work
image

NullVoxPopuli:
> Branch needs to be set to main

I don't see that as an option?
image

NullVoxPopuli:
oh! you mean on the parser plugin, gotchya :+1:

theHamsta (approved):

theHamsta:
Ready to merge? Or do you want some highlights?

theHamsta:
```
 Tags that start with a lower case letter are HTML tags
((tag_name) @tag
  (#match? @tag "^[a-z]"))
; Tags that start with a capital letter are Glimmer components
((tag_name) @constructor
  (#match? @constructor "^[A-Z]"))

(attribute_name) @property

(string_literal) @string
(number_literal) @number
(boolean_literal) @boolean

; == Mustache Statements ===

; Hightlight the whole statement, to color brackets and separators
(mustache_statement) @tag

; Generic identifiers are variables
(identifier) @variable
; Helpers are functions
(helper_identifier) @function

(comment_statement) @comment

(attribute_node "=" @operator)

(block_params "as" @keyword)
(block_params "|" @operator)

[
  "<"
  ">"
  "</"
  "/>"
] @tag
```

NullVoxPopuli:
> Ready to merge or do you want some highlights? You can probably adapt the queries from their repo and use the tags from our html or other template languages.

I have highlights already with my manual / local config: 
image

image

How would highlights be added to nvim-treesitter?

theHamsta:
Just create a file `queries/glimmer/highlights.scm`. I made above some changes `tag.delimiter` -> `tag`. `attribute` -> `property`. Feel free to add more highlights if necessary (with tags documented in `CONTRIBUTING.md`).

theHamsta:
@NullVoxPopuli oops. We have `tag.delimiter`. Then, only `attribute` -> `property` as in Html. Now I understand what this `attribute` was that we removed.

theHamsta:
Thanks for this contribution! Maybe we need to add an auto-import feature for the highlight files...

NullVoxPopuli:
legit!

Now I have hbs highlighting in js/ts/hbs! (without configuring local repo references)

image

NullVoxPopuli:
@alexlafroscia ^^ :D

alexlafroscia:
Sweet! Thanks for doing this @NullVoxPopuli -- I was going to wait until I had put it through it's paces more, but I'll up-stream fixes/improvements as they get made to the parser on my end.

PR #987, opened 2021-03-01
feat: use @ignore for comments and multiline strings

elianiva:
It doesn't work 100% though, for example this jsdoc comment
```javascript
// Original
/**
 * Some description
 * @param {string} foo - bar
 * @return foo
 */
const foo = (foo: string): string => foo

// Without this PR
/**
* Some description
* @param {string} foo - bar
* @return foo
*/
const foo = (foo: string): string => foo

// After this PR (incorrect)
/**
 * Some description
 * @param {string} foo - bar
 * @return foo
*/
const foo = (foo: string): string => foo

// Shouldn't get affected by the indent rule (correct)
/**
 * Some description
 * @param {string} foo - bar
 * @return foo
 */
const foo = (foo: string): string => foo
```

### DIFF
--- a/queries/c/indents.scm
+++ b/queries/c/indents.scm
@@ -14,3 +14,8 @@
   "{"
   "}"
 ] @branch
+
+[
+  (comment)
+  (preproc_function_def)
+] @ignore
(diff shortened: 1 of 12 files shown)


kyazdani42 (approved):
seems weird the @ignore directive does not work for the last line only. Maybe there is an offset issue with the algorithm (but it doesn't show up in other places so i'm not sure)

PR #995, opened 2021-03-02
feat: Respect user queries

tjdevries:
Now you can do this:

```lua
local read_query = function(filename)
  return table.concat(vim.fn.readfile(vim.fn.expand(filename)), "\n")
end

  -- ensure_installed = {'lua'}, -- one of 'all', 'language', or a list of languages
  ensure_installed = { 'go', 'rust', 'toml', 'query', },

  parsers = {
    lua = {
      injections = read_query("~/plugins/tree-sitter-lua/queries/lua/injections.scm"),
    }
  },

  highlight = {
    enable = enabled, -- false will disable the whole extension
    use_languagetree = false,
    disable = {"json"},
    custom_captures = custom_captures,
    queries = {
      lua  = read_query("~/plugins/tree-sitter-lua/queries/lua/highlights.scm"),
      rust = read_query("~/.config/nvim/queries/rust/highlights.scm"),
    }
  },
}

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -270,6 +270,7 @@ end
 -- @param user_data module overrides
 function M.setup(user_data)
   config.modules = vim.tbl_deep_extend('force', config.modules, user_data)
+  config.parsers = user_data.parsers or {}
 
   local ensure_installed = user_data.ensure_installed or {}
   if #ensure_installed > 0 then
@@ -409,6 +410,10 @@ end
 
 function M.get_update_strategy()
   return config.update_strategy
+end
+
+function M.get_parsers_config()
+  return config.parsers
 end
 
 return M
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -97,11 +97,13 @@ function M.attach(bufnr, lang)
     hlmap[k] = v
   end
 
-  local opts = {}
+  local opts = nil
   if config.queries then
     if config.queries[lang] then
-      opts.queries = {
-        [lang] = config.queries[lang]
+      opts = {
+        queries = {
+          [lang] = config.queries[lang]
+        }
       }
     end
   end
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by tjdevries** [comment]
Basically this will allow you to be able to specify _exactly_ what you'd like to be in your query.

Probably could use extension for other modules, but tbh, I don't use the other ones as often, so it's not as high priority for me.

**Comment by tjdevries** [comment]
@nvim-treesitter Any thoughts?

**Comment by theHamsta** [comment]
I would prefer to have this kind of config in nvim core for all the types of queries. The function to map `lang x querytype x injection context` to the actual query text should be overwritable (better by Lua config than runtime logic). Would be more elegant to have one thing how queries are handled across all nvim land.

If highlight/injection or this kind of configuration are sufficiently special we could also have them here. But we already have some quite complicated runtime query logic in Neovim core, so it may be better to adapt/make more configurable/simplify the logic there instead of adding more config and logic here.

**Comment by tjdevries** [comment]
But the thing is you don't provide any way to pass in the configuration in nvim-treesitter. you ALWAYS look up all the items.

The mechanism already exists in core -- you pass in the query if you want to override. It's just that nvim tree sitter passes `nil` to those values, instead of allowing users to use them.

**Comment by theHamsta** [comment]
I know that you can pass the queries to the highlighter directly. This is why we can totally merge this. Especially since the mother language is only important for the queries of injected languages.

The fallback is to call this function. Wouldn't it be possible to configure the behavior of this function directly? To also configure locals the same way as you did here? (query type x language -> what to do?)

```
function M.get_query(lang, query_name)
  local query_files = M.get_query_files(lang, query_name)
  local query_string = read_query_files(query_files)

  if #query_string > 0 then
    return M.parse_query(lang, query_string)
  end
end
```

theHamsta (approved):

tjdevries:
Ah, very very very smart.

Let me reconsider this before we go further. Ty

steelsojka:
We should also consider passing custom injection queries to the parser as well when it's created. It also allows a map of queries by language.

tjdevries:
What about this PR: https://github.com/neovim/neovim/pull/14252

This might make things much much better than my hax here?

vigoux:
Yep upstream pr is indeed better, shall I close this one then?

theHamsta:
@tjdevries I think this is superseded by the upstream PR.

PR #1003, opened 2021-03-05
add treesitter version in checkhealth

glepnir:
cc @vigoux @theHamsta   treesitter 0.19 has some problem with neovim,so add the version in checkhealth will be useful. it works like this

image

### DIFF
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -12,6 +12,16 @@ local M = {}
 
 local function install_health()
   health_start('Installation')
+  if fn.executable('tree-sitter') == 0 then
+    health_error('`tree-sitter` executable not found')
+  else
+    local handle = io.popen('tree-sitter  -V')
+    local result = handle:read("*a")
+    handle:close()
+    local version = vim.split(result,'\n')[1]:match('[^tree%psitter].*')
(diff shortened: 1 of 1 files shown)


vigoux (commented):
I find that misleading that the tree-sitter version showed here may be different than the one nvim has been compiled against (for exemple if you only use bundled deps).

glepnir:
aha right. need enhance.

theHamsta:
It should be possible to get the ABI version of the bundled tree-sitter runtime and the ABI for the installed parsers. It is used for the version check in nvim core `tslua_add_language`.

vigoux:
Version checking can be done (and is done today in neovim) using the version specified in `treesitter/api/h`.

I think what we really need here is a proper checkhealth upstream, with version checks for both parsers and the runtime.

glepnir:
also think get the version from upsream will be good.

theHamsta (approved):
Sorry thought this was an issue. The PR looks good. Maybe add an explanation what the `tree-sitter` binary is used for (only for generating parsers from JS) because most end users won't need that functionality.

```
`tree-sitter` found 0.19.1 (used for :TSInstallFromGrammar)
```

glepnir:
done.

theHamsta:
Thanks!

PR #1008, opened 2021-03-07
feat(dart): add indents

elianiva:
The `case` block can't be indented because the query doesn't group them (same with Java). Also, there's this scenario because I couldn't define *how many* it should get indented using the query.

```dart
// correct
static Future<Nvim> spawn(
    {String nvimBinary = 'nvim',
    List<String> commandArgs = const ['--embed'],
    NvimHandler onNotify,
    NvimHandler onRequest}) async {
  var nvim = Nvim._spawn(nvimBinary, commandArgs);
  nvim._nvim = await nvim._nvimFut;
  // ....
}

// after indenting
static Future<Nvim> spawn(
  {String nvimBinary = 'nvim',
  List<String> commandArgs = const ['--embed'],
  NvimHandler onNotify,
  NvimHandler onRequest}) async {
  var nvim = Nvim._spawn(nvimBinary, commandArgs);
  nvim._nvim = await nvim._nvimFut;
  // ....
}
```

ps: the code was stolen from [smolck's dart-nvim-api](https://github.com/smolck/dart-nvim-api/blob/master/lib/src/neovim.dart)

### DIFF
--- a/queries/dart/indents.scm
+++ b/queries/dart/indents.scm
@@ -1,0 +1,27 @@
+[
+  (class_body)
+  (function_body)
+  (function_expression_body)
+  (declaration (initializers))
+  (switch_block)
+  (if_statement)
+  (formal_parameter_list)
+  (formal_parameter)
+  (list_literal)
+  (return_statement)
+  (arguments)
+] @indent
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+  "["
+  "]"
+] @branch
+
+; this one is for dedenting the else block
+(if_statement (block) @branch)
+
+(comment) @ignore


kyazdani42:
based on what i see:
```
// correct should be (with indent spaces 2)
static Future<Nvim> spawn(
  {String nvimBinary = 'nvim',
    List<String> commandArgs = const ['--embed'],
    NvimHandler onNotify,
    NvimHandler onRequest}) async {
  var nvim = Nvim._spawn(nvimBinary, commandArgs);
  nvim._nvim = await nvim._nvimFut;
}
```
I'm not sure how Dart indent works but this formatting looks weird anyway.

queries/dart/indents.scm
```diff
@@ -0,0 +1,21 @@
+[
+  (class_body)
+  (function_body)
+  (declaration (initializers))
+  (switch_block)
+  (if_statement)
+  (formal_parameter_list)
+  (list_literal)
```
kyazdani42, line 8: I think i'd add `formal_parameter` to this list ? i really have no idea how dart indenting works though

elianiva:
> I'm not sure how Dart indent works but this formatting looks weird anyway.

yeah, I usually put `}` and `)` on a new line but that doesn't seem to be the case with Dart.

akinsho:
Ah this is interesting I hadn't looked into treesitter for indentation yet. In the code snippet you posted, `dartfmt` will do that if user hasn't followed  [the guidance](https://flutter.dev/docs/development/tools/formatting#:~:text=The%20guideline%20for%20adding%20a,breaks%20for%20Flutter%2Dstyle%20code.) To append a trailing comma to a list off arguments. I believe the tooling can't do this for some complex reason relating you darts syntax so the user is expected to. If you are a coma after the final parameter in the example and reformatted it. It would be indented like you'd expect

akinsho:
Here's a note that explains some of the indentation behaviour better than I can (from [`dart-lang/dart-vim`](https://github.com/dart-lang/dart-vim-plugin#faq)
> The indentation capabilities within vim are limited and it's not easy to fully express the indentation behavior of dartfmt. The major area where this plugin differs from dartfmt is indentation of function arguments when using a trailing comma in the argument list. When using a trailing comma (as is common in flutter widget code) dartfmt uses 2 space indent for argument parameters. In all other indentation following an open parenthesis (argument lists without a trailing comma, multi-line assert statements, etc) dartfmt uses 4 space indent. This plugin uses 4 space indent to match the most cases.

akinsho:
Just tested this out, could be missing something but not sure this works as is. When I try and enter a new line in a dart project the indentation level is wrong, this was tested on a simple widget tree.

elianiva:
> When I try and enter a new line in a dart project the indentation level is wrong, this was tested on a simple widget tree.

yeah, I think I missed the `arguments` node which is used for the widget tree. I think I've fixed it.

akinsho:
That's definitely better but a few edge cases still I think for example
```dart
  Widget _loginOptions(LoginViewModel model) { // <-- press <CR> here goes
  | <-- should be indented by one more
    return Column(
      children: [
        Expanded(
          child: Column(
            mainAxisAlignment: MainAxisAlignment.center,
            children: [
              SizedBox(
                width: double.infinity,
                child: ElevatedButton(
                  onPressed: () async { // <-- press <CR> here
				  | <-- goes here, one indent missing
					| <-- should be here					
                    await model.signUpAnon();
                  },
                  child: Text('Sign up as Guest'),
                ),
              ),
              SizedBox(height: 16),
              SizedBox(
                width: double.infinity,
                child: ElevatedButton(
                  onPressed: () {
                    model.setIsRegistering(true);
                  },
                  child: Text('Create an account'),
                ),
              ),
            ],
          ),
        ),
      ],
    );
  }

```

elianiva:
```dart
Widget _loginOptions(LoginViewModel model) { // <-- press <CR> here goes
| <-- should be indented by one more
```

this is also the case for some languages like Rust and JS/TS. I *think* this is due to how the algorithm works (or my query just sucks :laughing:). I haven't read any of the algorithm yet, I made most of the indent queries in this repo but I never read how it actually works.

```dart
                  onPressed: () async { // <-- press <CR> here
				  | <-- goes here, one indent missing
					| <-- should be here		
		    | <-- not here?     ^ instead of here		
                    await model.signUpAnon();
                  },
```

akinsho:
@elianiva sorry my formatting came out wrong there so weird, the `onPressed` example should look like
```lua
 onPressed: () async { // <-- press <CR> here
 | <-- goes here, one indent missing
    | <-- should be here			
    await model.signUpAnon();
 },
```
It's basically just slightly off by one when entering a new scope but tbh I think this sort of thing can be iterated on and isn't  broken so 🤷🏿

elianiva:
ah, so same issue with the other ones I've mentioned. I kinda used to that at this point because it only happens when it's exactly the next line that needs 1 indent deeper. If you press `<CR>` on `..UpAnon(); <- here` then it gives the correct result.

akinsho:
Just tried what you described and that works fine, if it's an issue with other languages then that's fine I'm sure when a solution is found it can be ported to dart in the future. LGTM for what it's worth 🤷🏿‍♂️ , thanks for adding this 👍🏿

kyazdani42:
yes i've noticed the new line stuff is wrong for js/ts too. The algorithm might not be perfect, after my first initial attempts, one dude took it over and made it much better but this new line thing is not working as expected. I might fix that when i have some time because it's kind of disturbing. Anyhow, looks good

kyazdani42 (approved):

akinsho:
@kyazdani42 do you know if this indentation issue with it not starting at the correct level after pressing enter was ever fixed for js/ts, and I can apply the same fixes to dart? Been using this for a while and has started to get on my nerves 😆

PR #1012, opened 2021-03-09
Support for R language

echasnovski:
This PR adds support for R language. Maintainer of r-lib/tree-sitter-r, @jimhester, agreed to add it to the list here (see r-lib/tree-sitter-r#7). Queries are taken in their current state from repository.

### DIFF
--- a/queries/r/highlights.scm
+++ b/queries/r/highlights.scm
@@ -1,0 +1,86 @@
+; highlights.scm
+
+(call function: (identifier) @function)
+(namespace_get function: (identifier) @function.method)
+(namespace_get_internal function: (identifier) @function.method)
+
+; Literals
+
+(integer) @number
+
+(float) @number
+
+(complex) @number
+
+(string) @string
+
+(comment) @comment
+
+(formal_parameters (identifier) @variable.parameter)
+
+(identifier) @variable
+
+; Operators
+[
+ "="
+ "<-"
+ "<<-"
+ "->"
+] @operator
+
+(unary operator: [
+  "-"
+  "+"
+  "!"
+  "~"
+] @operator)
+
+(binary operator: [
+  "-"
+  "+"
+  "*"
+  "/"
+  "^"
+  "<"
+  ">"
+  "<="
+  ">="
+  "=="
+  "!="
+  "||"
+  "|"
+  "&&"
+  "&"
+  ":"
+  "~"
+] @operator)
+
+(special) @operator
+
+[
+ "("
+ ")"
+ "["
+ "]"
+ "{"
+ "}"
+] @punctuation.bracket
+
+[
+ "while"
+ "if"
+ "else"
+ "function"
+ "repeat"
+ "for"
+ "in"
+ (dots)
+ (true)
+ (false)
+ (break)
+ (next)
+ (inf)
+ (nan)
+ (na)
+ (null)
+] @keyword
--- a/queries/r/locals.scm
+++ b/queries/r/locals.scm
@@ -1,0 +1,11 @@
+; locals.scm
+
+(function_definition) @local.scope
+
+(formal_parameters (identifier) @local.definition)
+
+(left_assignment name: (identifier) @local.definition)
+(equals_assignment name: (identifier) @local.definition)
+(right_assignment name: (identifier) @local.definition)
+
+(identifier) @local.reference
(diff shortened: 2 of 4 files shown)


queries/r/locals.scm
```diff
@@ -0,0 +1,11 @@
+; locals.scm
+
+(function_definition) @local.scope
+
+(formal_parameters (identifier) @local.definition)
```
theHamsta, line 5: we have different captures `@definition.parameter`, `@reference`, @definition.var`

queries/r/highlights.scm
```diff
@@ -0,0 +1,86 @@
...
+
+(float) @number
+
+(complex) @number
+
+(string) @string
+
+(comment) @comment
+
+(formal_parameters (identifier) @variable.parameter)
```
theHamsta, line 19: CI is unhappy about that -> `@parameter`

queries/r/highlights.scm
```diff
@@ -0,0 +1,86 @@
+; highlights.scm
+
+(call function: (identifier) @function)
+(namespace_get function: (identifier) @function.method)
```
theHamsta, line 4: `@method`

theHamsta: We get this question over and over. You could create an issue to unifiy the qurey system but right now Emacs, Atom and nvim-treesitter have slightly different queries. The ones that live in their original repo are typically in the Atom-style (even if there is no exact definition what that is). Emacs and Nvim have a bit more highlight groups that often correspond to already existing highlights. That's the reason why we have our own queries here in this repo.

There is also https://github.com/nvim-treesitter/nvim-treesitter/issues/904 and a suggestion to allow more specific quries. E.g. `@function,method` would fallback to `@function` if the more specific one does not exist.

Someone would have to take the effort and change all queries if everyone agrees that having one query for Atom and nvim would be better.

theHamsta: See CONTRIBUTING.md

echasnovski: > We get this question over and over.

Sorry for trouble. Didn't find anywhere the actual answer to this question, so decided to double check.

theHamsta: @echasnovski I was not trying to say that we are annoyed but rather that this is a real issue that many people have (that we should solve?) 

summary of discussion:
- own queries: be more flexible, be more nvim-specific
- use upstream: less maintenance

Maybe we can find a good solution or compromise. PRs and issues to help us resolving that are welcome.

queries/r/locals.scm
```diff
@@ -1,11 +1,11 @@
 ; locals.scm
 
-(function_definition) @local.scope
+(function_definition) @scope
 
-(formal_parameters (identifier) @local.definition)
+(formal_parameters (identifier) @definition)
```
theHamsta, line 5: `@definition.parameter`?

queries/r/locals.scm
```diff
@@ -1,11 +1,11 @@
...
 
-(left_assignment name: (identifier) @local.definition)
-(equals_assignment name: (identifier) @local.definition)
-(right_assignment name: (identifier) @local.definition)
+(left_assignment name: (identifier) @definition)
+(equals_assignment name: (identifier) @definition)
+(right_assignment name: (identifier) @definition)
```
theHamsta, line 9: these are probable `@definition.var`

queries/r/locals.scm
echasnovski: Yea, probably. Thanks.

Off topic observation: having many capture groups [in CONTRIBUTING.md](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#locals) without explanations of what they actually mean creates a need for guessing.

queries/r/locals.scm
theHamsta: Ok, then it should be fine. One could still argue that all of this are still variables even if you can assign functions to it.

theHamsta (approved):

theHamsta:
Could you pin this parser to the commit before the 0.19.2 release? We will use the newest revision in #1000 or we wait for this commit until #1000 has been merged.

theHamsta:
I'll merge your PR into #1000 since 0.19.2 for nvim will be merged soon.

PR #1014, opened 2021-03-10
Update svelte parser

jupblb:
Make svelte parser depend on `src/scanner.c` instead of `src/scanner.cc` because of https://github.com/Himujjal/tree-sitter-svelte/commit/43817cfe3e936859959a897fce411e3982fd2273

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -460,7 +460,7 @@ list.devicetree = {
 list.svelte = {
   install_info = {
     url = "https://github.com/Himujjal/tree-sitter-svelte",
-    files = { "src/parser.c", "src/scanner.cc" },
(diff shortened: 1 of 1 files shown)


theHamsta:
Already done in #1000. The old version in the lockfile is not affected right?

theHamsta:
As CI is failing, I assume the locked version still has scanner.cc

jupblb:
Thanks for this information. I guess we can close this PR then. But it's worth underlining that svelte parser revision from lockfile relates to https://github.com/Himujjal/tree-sitter-svelte/commit/5d27580b15d778fd329c59bf6bca900bb736c17c which is not part of the main repository. I *guess* in my case it just fetches the newest commit.

theHamsta:
Are you using Windows? When you fall back to git there is currently to regard of the lockfile (there's also a PR for it).

PR #1020, opened 2021-03-12
fix: indent on new line

kyazdani42:
Fixes #1019
Fixes #1074
Fixes #802

One issue with this PR is that python is still not working as expected.
I'm not sure why but python behaves quite strangely:
typing `def a():` gives a node without errors (although it's an empty function) thus indenting like the function start row
typing
```python
def a():
    print(1)
    |
    |
```
it will indent on every new line (not sure if this is expected behavior.
I'm pretty sure the singularity of the python language makes it hard to make it work with other languages. (EDIT: maybe this could be solved with more advanced indent queries)
If someone has an idea on how to fix that, i'd be happy to hear them.

@jedrzejboczar i'd like your input on this. Newline indent is working super well with this PR, but for python i did not manage to make it work properly. 

EDIT: i've found some strange behaviors in HTML indents just by nesting divs.

PR #1024, opened 2021-03-13
Add initial indent plugin unit tests

jedrzejboczar:
Currently indentation has many corner cases that fail and trying to fix them often breaks something else. This PR is an initial sketch of unit tests of indentation. It uses [plenary.nvim](https://github.com/nvim-lua/plenary.nvim) and I mostly based the setup on [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim)'s tests.

The basic idea is to create files per language in `lua/tests/indent/LANG/` that will be short code snippets with correct indentation. Then the test runner opens them and just runs `normal gg=G`. I then compares file contents before vs after. 

I added only some Python tests, from which most fail (there are some tests that follow quite specific PEP8 guidelines, but maybe some could be removed). I also added minimal examples of C and Rust files. I hope that the test base will be expanded to include test cases for all the supported languages. This would help to find all the corner cases and improve the implementation of `indent.lua`. For now most tests are failing but building a solid test base first should make working on the indent plugin much easier. And when we fix the issues we can run the tests in CI.

Currently there are only tests with `normal gg=G` but it would be good to also add tests for hitting newline in code snippets like:
```python
def foo():
    print()|  # <- cursor here
```
The file would be loaded, then we could do `normal osomething` on that line and check if we get correctly indented line. 

This could be helpful for working on https://github.com/nvim-treesitter/nvim-treesitter/pull/1020 and any other modifications of the indent plugin, as testing everything manually after changes is a slow and cumbersome.


jedrzejboczar:
Ah, forgot to add, to run the tests just execute `./scripts/run_tests.sh`. There should be 4 successes/6 failures for Python, 1/1 for C and 1/0 for Rust.

lua/nvim-treesitter/test_utils.lua
```diff
@@ -0,0 +1,38 @@
...
+end
+
+function M.indent_whole_file(file, opts)
+  assert.are.same(1, vim.fn.filereadable(file))
+
+  -- load reference file
+  vim.cmd(string.format('edit %s', file))
+  local reference = vim.api.nvim_buf_get_lines(0, 0, -1, true)
+
+  -- FIXME: why this is not set automatically?
```
elianiva, line 21: There's a FileType autocmd that *should* fix this but I still need to do this manually. Seems like it's not just me that still has this issue 😆

jedrzejboczar:
I implemented the new line type of tests and added more realistic tests for C, C++, Lua, Rust and Python. I thing this should now be in a state where it can be actually useful.

To summarize, tests are defined as:

* Proper indentation is defined in given language files (e.g. `lua/tests/indent/c/*.c`). These files should be formatted the way we want indentation to work in that language.
* Each language has a specification file (e.g. `lua/tests/indent/c_spec.lua`) with definitions of the "new line" tests.

There are currently two types of tests:

1. "Whole file" - these ones automatically discover all the files from `indent/LANG/`. Each file is first opened in a buffer, its content is stored, then we run `normal gg=G` and compare the content of the file with the original one. If there is any difference the test fails.
2. "New line" - these have to be defined manually in the specification files. Each test loads the file, then inserts some text at given line using the command `normal! o`. Then we check the resulting indent of the inserted line and compare it with desired value.

There is a simple script to run the tests: `./scripts/run_tests.sh`. In case of multiple tests failing there is quite a lot of output so I added very basic summary using grep (enabled with `./scripts/run_tests.sh --summary`). In the future it would be nice to have some filtering options to e.g. filter by language/test name. I attach example files with the currently expected output: [full.txt](https://github.com/nvim-treesitter/nvim-treesitter/files/6353662/full.txt), [summary.txt](https://github.com/nvim-treesitter/nvim-treesitter/files/6353663/summary.txt).

I was able to improve the assertion messages a little bit, so now the errors are displayed as a side-by-side diff:
```
Success ||      indent C++: new line: c/ternary.c[4]
Fail    ||      indent C++: new line: c/no_braces.c[4]
            ./lua/nvim-treesitter/test_utils.lua:62: Incorrect indentation
                       Found:                                       Expected:
                        |int foo(int x) {                            |int foo(int x) {
                        |    if (x > 10)                             |    if (x > 10)
                        |        return 10;                          |        return 10;
                        |    else                                    |    else
              4 vs 8 => |    x++;                                    |        x++;
                        |        return x;                           |        return x;
                        |                                            |
                        |    while (1)                               |    while (1)
                        |        x++;                                |        x++;
                        |                                            |
                        |    for (int i = 0; i < 3; ++i)             |    for (int i = 0; i < 3; ++i)
                        |        x--;                                |        x--;
                        |}                                           |}

            stack traceback:
                ./lua/nvim-treesitter/test_utils.lua:62: in function 'compare_indent'
```

In the tests I added, there are a lot of failures, but I believe this will be helpful when working on the current implementation of indentation. These are the current results I get (success/all):

* C: 19 / 32
* C++: 20 / 38
* Lua: 17 / 22
* Rust: 40 / 50
* Python: 16 / 31

Some of the tests I created may have disputable indentation expectations (like aligning the hanging indent in Python), so this is to be check and decided what we actually want to support.

Currently there are only tests for 5 languages, but I hope others could help adding tests for different languages. This should be as easy as writing some properly indented files in that language and adding a spec file based on the existing ones.

There are a lot of new files added (most are test case definitions) so here is a little summary:
```
├── lua
│   ├── nvim-treesitter
│   │   └── test_utils.lua
│   └── tests
│       └── indent
│           ├── c_spec.lua
│           ├── cpp_spec.lua
│           ├── lua_spec.lua
│           ├── python_spec.lua
│           └── rust_spec.lua
│           ├── c
│           │   ├── ...
│           ├── cpp
│           │   ├── ...
│           ├── lua
│           │   ├── ...
│           ├── python
│           │   ├── ...
│           └── rust
│               └── ...
└── scripts
    ├── minimal_init.lua
    └── run_tests.sh
```

Could you please take a look at this @kyazdani42? I'd be grateful for a review and any suggestions.

lua/nvim-treesitter/test_utils.lua
jedrzejboczar: Now when I disabled this it works as expected, so I guess it has already been fixed :)

lua/tests/indent/cpp_spec.lua
```diff
@@ -0,0 +1,62 @@
...
+    end
+  end)
+
+  describe('new line:', function()
+    local run = function(file, spec, title)
+      title = title and title or tostring(spec.on_line)
+      it(string.format('%s[%s]', get_name(file), title), function()
+        new_line(file, spec, opts)
+      end)
+    end
```
kyazdani42, line 36: i think you could factor out this run function :)

kyazdani42 (commented):
Thanks a lot for doing this, this is greatly appreciated.
I don't have much to say about the code style which is quite clean imho, and this way we will be able to improve the indent algorithm bit by bit without breaking stuff ! Thanks :)

vigoux:
Awesome!

Just a tiny thing : I'd quite like this to rather go in a `tests` directory directly rather than the `lua` directory.

The reason is that when nvim searches for Lua file (on `require`) they'll be searched too, which is not really great.

lua/tests/indent/cpp_spec.lua
jedrzejboczar: You mean to have it as top-level local function? Or to move it to `test_utils.lua`?

kyazdani42: test_utils it, it's common to every spec.lua file you wrote in some details which you could generate with a closure.

jedrzejboczar:
@vigoux I moved all the tests to top level `tests/` directory. There is still `lua/nvim-treesitter/test_utils.lua` remaining in the `lua/` directory. I used it to share code between the `*_spec.lua` files. Do you also want it to be moved out of `lua`? Then I guess I'd have to load it bypassing the regular `require`?

@kyazdani42 I moved the `run` functions out of the `describe` blocks, which looks a tiny bit cleaner imo. There is still some code duplication between the `*_spec.lua` files, I didn't factor it all out to `test_utils.lua` because it uses these global functions `describe` and `it` that are somehow set by busted. But if you think it would be good I may try to factor it out anyway, it would make the specification files cleaner.

kyazdani42:
@jedrzejboczar you could factor it out i think it'll be easier to add support to new languages then ! Could you move the test_utils.lua in the test directory and somehow add this to the runtimepath through the minimal init file ?

vigoux:
Maybe we can do something similar to what we do to check tu loading of queries, with a Lua script we load in a headless neovim too.

jedrzejboczar:
Ok, I managed to move everything into `tests/` and also to refactor it to avoid code duplication so the spec files are a lot cleaner now.

vigoux:
Awesome, I'm going to test that later today, just to get a grasp of how it works and so, and merge it right after that.

For now though we will not add it to the CI, in order to wait for the indents to be fixed. @kyazdani42 do you agree on the CI part?

kyazdani42:
very cool
I agree, would fail the CI each time which we might not want

theHamsta:
But you can have allowed failures in CI

vigoux (approved):
It is working great ! I am 100% in favor of this, and I'll start using it for fold tests too !

kyazdani42:
@theHamsta yep but still the CI would show a red light each time ^^

kyazdani42:
ok let's merge this, it will enable us to do more testing overall. Thanks a lot again @jedrzejboczar !

PR #1025, opened 2021-03-14
use health#report_warn for treesitter version check

glepnir:
cc @theHamsta

### DIFF
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -7,19 +7,21 @@ local info = require'nvim-treesitter.info'
 local health_start = vim.fn["health#report_start"]
 local health_ok = vim.fn['health#report_ok']
 local health_error = vim.fn['health#report_error']
+local health_warn = vim.fn['health#report_warn']
 
 local M = {}
 
 local function install_health()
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/health.lua
```diff
@@ -7,13 +7,14 @@ local info = require'nvim-treesitter.info'
...
 local health_error = vim.fn['health#report_error']
+local health_warn = vim.fn['health#report_warn']
 
 local M = {}
 
 local function install_health()
   health_start('Installation')
   if fn.executable('tree-sitter') == 0 then
-    health_error('`tree-sitter` executable not found')
+    health_warn('`tree-sitter` executable not found')
```
theHamsta, line 17: but there is still the explanation missing what `tree-sitter` is used for

glepnir: space had add. the explanation of  tree-sitter like what?

vigoux: I'd say he meant `the command-line is required when running `TSInstallFromGrammar` and for the installation of some parsers`

theHamsta (approved):

PR #1039, opened 2021-03-16
Graphql, Turtle, SparQL 0.19.3

theHamsta:
@bkegley just updated his parser


vigoux:
Why are all those failing?

theHamsta:
Because apart from building from source (MacOS the other still have 12 ABI). Also the only required test no longer exists.

kyazdani42 (approved):

PR #1042, opened 2021-03-17
Syntax plus treesitter

theHamsta:
I thought this is a better solution than giving an option to `ts.highlighter` (the culprit that deactivates syntax) because it seems to handle also the case no syntax was loaded (though in a weird way by deactivating syntax and then checking whether it's deactivated).

Fixes https://github.com/nvim-treesitter/nvim-treesitter/issues/959

 @sperezconesa can you check whether this works?

You would add the option `additional_vim_regex_highlighting = true` to you setup call of nvim-treesitter in the highlight section.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -22,7 +22,8 @@ local builtin_modules = {
     enable = false,
     disable = {'markdown'}, -- FIXME(vigoux): markdown highlighting breaks everything for now
     custom_captures = {},
-    is_supported = queries.has_highlights
+    is_supported = queries.has_highlights,
+    additional_vim_regex_highlighting = false,
   },
   incremental_selection = {
     module_path = 'nvim-treesitter.incremental_selection',
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -104,7 +104,9 @@ function M.attach(bufnr, lang)
   end
 
   ts.highlighter.new(parser, {})
-  if config.additional_vim_regex_highlighting then
+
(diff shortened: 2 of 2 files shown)


cloggier:
What are the benefits/downsides of having `syntax` `on` with tree-sitter's highlighting?

theHamsta:
@cloggier plugins that rely on `syntax` like vim-matchit or fold plugins will continue to work. Also for broken parsers like Julia the regex syntax was a good backup. It's here because it was requested.

cloggier:
@theHamsta Thanks for explanation. Just to be sure, will ts' highlighting take precedence incase syntax is not off?

theHamsta:
When I tried the last time it had precedence. This may have changed now. You can try with `:set syntax=1` in a tree-sitter buffer.

lua/nvim-treesitter/configs.lua
```diff
@@ -22,7 +22,8 @@ local builtin_modules = {
     enable = false,
     disable = {'markdown'}, -- FIXME(vigoux): markdown highlighting breaks everything for now
     custom_captures = {},
-    is_supported = queries.has_highlights
+    is_supported = queries.has_highlights,
+    additional_vim_regex_highlighting = false,
```
theHamsta, line 26: Maybe have this true only for some languages?

theHamsta: This will now allow either `true` or `{ 'python', 'cpp' }` if you only want to have this for some languages.

kyazdani42 (approved):

vigoux (approved):

AckslD:
Awesome! Was just looking for something like this since it makes having `spell` on much nicer :)

Issue #1046, opened 2021-03-17
Coordinate with tree-sitter to establish convention for auto-loading common files from syntax plugins


NullVoxPopuli:
**Is your feature request related to a problem? Please describe.**

Today, we must specify stuff like: https://github.com/NullVoxPopuli/dotfiles/blob/master/home/.config/nvim/lua/plugin-config/treesitter.lua#L15-L30 
```lua
 local parser_config = require "nvim-treesitter.parsers".get_parser_configs()

  parser_config.glimmer = {
    install_info = {
      url = "~/Development/OpenSource/tree-sitter-glimmer",
      files = {
        "src/parser.c",
        "src/scanner.c"
      }
    },
    filetype = "hbs",
    used_by = {
      "handlebars",
      "html.handlebars"
    }
  }
```
in order to tell treesitter what files to use.

This then needs to be duplicated across development, nvim-treesitter, and likely other editors as well. 

**Describe the solution you'd like**

A convention established with the tree-sitter folks so that syntax parsing plugins can define this config, and reduce duplication. Maybe an `onLoad` callback/hook or something for each plugin to add their own config?

It'd still be nice to override a single property for debugging though.
Like, let's say plugins _can_ configure their own parser_config, and I wanted to debug something locally -- instead of doing the above code snippet, I'd only do:
```lua
 local parser_config = require "nvim-treesitter.parsers".get_parser_configs()

 parser_config.glimmer.install_info.url = "~/path/to/local/copy/of/plugin"
```

**Describe alternatives you've considered**

Leaving things as is isn't great for DX / esp when wanting to get tree-sitter used across multiple editors / distribution platforms. I'd like to iterate on using tree-sitter in a light web editor, but configuring all the parser_configs per language _even after_ having all the syntax plugins seems a little.. redundant? idk

**Additional context**

n/a

steelsojka:
I don't think there is enough overlap in how parsers are consumed to condense it into something all editors could share. A lot of these properties are neovim specific, like filetype and filetype aliasing.

> This then needs to be duplicated across development, nvim-treesitter, and likely other editors as well.

I'm not sure what you mean by this. Usually this is only configured once in nvim-treesitter or a plugin that is registering a parser. The user consuming it wouldn't have to setup any of this information.

> Leaving things as is isn't great for DX / esp when wanting to get tree-sitter used across multiple editors / distribution platforms.

[...]

PR #1047, opened 2021-03-18
Update glimmer queries from plugin

NullVoxPopuli:
Maybe there is a way to read this file from the plugin directly?

### DIFF
--- a/queries/glimmer/highlights.scm
+++ b/queries/glimmer/highlights.scm
@@ -1,10 +1,9 @@
-;; By @alexlafroscia (https://github.com/alexlafroscia/tree-sitter-glimmer)
-;; Licensed under MIT
-;; Modified for nvim-treesitter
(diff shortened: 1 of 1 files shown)


theHamsta (approved):
You could add a PR for importing queries. The copy could be done with a similar command as the copy of the parser after compilation

NullVoxPopuli:
> You could add a PR for importing queries. The copy could be done with a similar command as the copy of the parser after compilation

ooo, what's that mean?
Sounds like it could maybe be a step towards https://github.com/nvim-treesitter/nvim-treesitter/issues/1046 ? :D 
I'm still fairly new to tree-sitter, and haven't looked at how nvim-tree-sitter works

alexlafroscia:
Thanks for updating this @NullVoxPopuli!😀

PR #1058, opened 2021-03-19
go: add builtin types

110y:
As same as https://github.com/nvim-treesitter/nvim-treesitter/pull/1034 which adds built-in functions, this PR adds built-in `types`.
ref: https://golang.org/pkg/builtin/


theHamsta (approved):

theHamsta:
Thanks!

PR #1068, opened 2021-03-20
kotlin: updated parser, and added annotation highlight again

tormodatt:


theHamsta (approved):

queries/kotlin/highlights.scm
```diff
@@ -70,10 +70,9 @@
 (type_identifier) @type
 
 ;; Annotations
-;; TODO:
-;(annotation (single_annotation) @attribute)
-;(single_annotation (user_type (type_identifier) @attribute))
-;(single_annotation (constructor_invocation (user_type (type_identifier) @attribute)))
+(annotation (single_annotation) @annotation)
```
theHamsta, line 73: I don't know how this is called at the moment. CI says it's not in contributing at the moment. It should use the same highlights as dart annotations an C++ attributes.

PR #1075, opened 2021-03-21
add support for beancount files

polarmutex:
I have written a tree-sitter parser for beancount that I would like added to this repo, thanks

### DIFF
--- a/queries/beancount/highlights.scm
+++ b/queries/beancount/highlights.scm
@@ -1,0 +1,26 @@
+
+(date) @field
+(txn) @annotation
+
+(account) @type
+
+(amount) @number
+(incomplete_amount) @number
+(amount_tolerance) @number
+
+(currency) @property
+
+(key) @label
+(string) @string
+
+(tag) @constant
+(link) @constant
+
+(comment) @comment
+
+[
+    (BALANCE) (OPEN) (CLOSE) (COMMODITY) (PAD)
+    (EVENT) (PRICE) (NOTE) (DOCUMENT) (QUERY)
+    (CUSTOM) (PUSHTAG) (POPTAG) (PUSHMETA)
+    (POPMETA) (OPTION) (INCLUDE) (PLUGIN)
+] @keyword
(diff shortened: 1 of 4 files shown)


queries/beancount/highlights.scm
```diff
@@ -0,0 +1,26 @@
+
+(date) @field
+(txn) @annotation
+
```
theHamsta, line 4: @vigoux @steelsojka should we use `@anotation` and `@attribute` the same way as the Atom queries? (Source code annotations/attributes and `@attribute` for html/xml attributes)

queries/beancount/locals.scm
```diff
@@ -0,0 +1,3 @@
+(transaction
+    (date) @definition.date
+    (txn_strings) @definition.payee) @definition.transaction
```
theHamsta, line 3: At the moment, we only have a fixed set of definitions. You would have to edit contributing.md to add new queries.

polarmutex: removed the locals file since this language does not have normal language mechanics

queries/beancount/highlights.scm
polarmutex: changed it to attribute

theHamsta (approved):

queries/beancount/highlights.scm
```diff
@@ -0,0 +1,26 @@
...
+(tag) @constant
+(link) @constant
+
+(comment) @comment
+
+[
+    (BALANCE) (OPEN) (CLOSE) (COMMODITY) (PAD)
+    (EVENT) (PRICE) (NOTE) (DOCUMENT) (QUERY)
+    (CUSTOM) (PUSHTAG) (POPTAG) (PUSHMETA)
+    (POPMETA) (OPTION) (INCLUDE) (PLUGIN)
```
vigoux, line 25: Tiny tiny comment, which is absolutely not blocking: there is ways to have keywords always matched as terminals, using the `alias` function in the grammar, feel free to reach out to me on Zulip to talk about that!

vigoux (approved):
Great, thanks for this addition!

PR #1085, opened 2021-03-22
Add basic queries for bibtex

clason:
What it says on the tin: a first attempt at adding queries for bibtex:

* highlights (can be improved, suggestions for better captures appreciated)
* indents
* folds

@theHamsta


queries/bibtex/highlights.scm
```diff
@@ -0,0 +1,49 @@
...
+
+[
+  (brace_word)
+  (quote_word)
+] @string
+
+[
+  (key_brace)
+  (key_paren)
+] @attribute
```
theHamsta, line 37: `attribute` has the moment the meaning of `code annotation`. We should probably change that... memo to myself. I would leave it that way.

clason: yeah, I don't know the proper highlight groups for tree-sitter, I'm afraid. That's why the checks are failing 😞 

I'd  be grateful for suggestions for appropriate groups!

queries/bibtex/highlights.scm
```diff
@@ -0,0 +1,49 @@
...
+
+(command) @function.builtin
+
+(number) @number
+
+(field
+  name: (identifier) @variable.builtin)
+
+(token
+  (identifier) @variable.parameter)
```
theHamsta, line 27: This is `@parameter` for us. Should we add `@variable.parameter` as a comp mapping (needs to be added to hl_map in highlight.lua and contributing).

queries/bibtex/highlights.scm
theHamsta: no, attribute exists. The failure is due to `@variable.parameter`

clason: I mean "the failure is due to the fact that I don't know what I'm doing, generally" :]

queries/bibtex/indents.scm
```diff
@@ -0,0 +1,10 @@
+[
+ (entry)
+] @indent
+
+[
+ "{"
+ "}"
+] @branch
+
+(comment) @ignore
```
theHamsta, line 10: so you could add yourself as maintainer for bibtex? Or we add an auto-import feature

clason: "I would prefer not to" :)

clason: (see above: not really knowing what I'm doing)

clason: the indents and folds were easy enough, though -- those I can manage :)

clason: But on a more serious note, I don't think we need an auto-import here -- the highlights queries need tuning, and the other ones are not from upstream.

theHamsta (approved):

README.md
```diff
@@ -134,7 +134,7 @@ We are looking for maintainers to add more parsers and to write query files for
 <!--parserinfo-->
 - [x] [bash](https://github.com/tree-sitter/tree-sitter-bash) (maintained by @TravonteD)
 - [x] [beancount](https://github.com/polarmutex/tree-sitter-beancount) (maintained by @polarmutex)
-- [ ] [bibtex](https://github.com/latex-lsp/tree-sitter-bibtex)
+- [x] [bibtex](https://github.com/latex-lsp/tree-sitter-bibtex)
```
theHamsta, line 137: this is automatically edited. the cross will only be added if there is a maintainer.

clason: Sorry, I was going by the text before the list.

(Which I wrote, so...)

vigoux (approved):
Great! Thanks!

theHamsta:
Looks really nice!

Issue #1067, opened 2021-03-20
Error executing lua Vim:E484: Can't open file lock file.json


richban:
Everytime I open `nvim`, `nvim-treesitter` fails to compile. I can't run neither of the commands: `TSUpdate` `TSUpdate all` etc. receive the following error message:

```
E5108: Error executing lua Vim:E484: Can't open file /Users/rbanyi/.local/share/nvim/site/pack/packer/start/nvim-treesitter/lock
file.json

```

```
   19 health#nvim_treesitter#check
   18 ========================================================================
   17 ## Installation
   16   - OK: `tree-sitter` found  0.19.4(parser generator, used for :TSInstallFromGrammar)
   15   - OK: `git` executable found.
   14   - OK: `cc` executable found.
   13
   12 ## Parser/Features H L F I
   11   - supercollider  ✓ ✓ ✓ ✓
   10   - glimmer        ✓ . . .
    9   - json           ✓ ✓ ✓ ✓
    8   - yaml           ✓ ✓ ✓ ✓

```

```tree-sitter 0.19.4```

```
NVIM v0.5.0-dev+1161-gfa4ee00fa
Build type: Release
LuaJIT 2.1.0-beta3
[...]

PR #1092, opened 2021-03-23
Address #1067: check whether lockfile exists

theHamsta:
@richban could you try which path this function actually returns on your machine: https://github.com/theHamsta/nvim-treesitter/blob/b8c2128e7954fd7916c2004d92aa831b8f2d9c32/lua/nvim-treesitter/utils.lua#L40-L46 ?

This PR should at least allow you to use nvim-treesitter without lockfile when it does not exist.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -33,7 +33,8 @@ end
 
 local function get_revision(lang)
   if #lockfile == 0 then
-    lockfile = vim.fn.json_decode(vim.fn.readfile(utils.join_path(utils.get_package_path(), 'lockfile.json')))
+    local filename = utils.join_path(utils.get_package_path(), 'lockfile.json')
+    lockfile = vim.fn.filereadable(filename) == 1 and vim.fn.json_decode(vim.fn.readfile(filename)) or {}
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):

richban:
```
"/Users/richban/.local/share/nvim/site/pack/packer/start/nvim-treesitter/lua/nvim-treesitter/utils.lua"
"/Users/richban/.local/share/nvim/site/pack/packer/start/nvim-treesitter"
"/Users/richban/.local/share/nvim/site/pack/packer/start/nvim-treesitter/lua/nvim-treesitter/utils.lua"
"/Users/richban/.local/share/nvim/site/pack/packer/start/nvim-treesitter"
```

richban:
Also I have found out that the error for compilation is from `tree-sittter-haskell` in the path `/Users/richban/.local/share/nvim/tree-sitter-haskell`. It does not compile but why does it download it in the first place.

theHamsta:
The Haskell parser does not compile on our CI when using Apple clang or an old gcc or on all configurations on windows. It needs to be compiled with c++11 (the compiler must have c++11 as a default)

PR #1094, opened 2021-03-23
Update teal highlights

euclidianAce:


lua/nvim-treesitter/parsers.lua
```diff
@@ -385,7 +385,6 @@ list.teal = {
   install_info = {
     url = "https://github.com/euclidianAce/tree-sitter-teal",
     files = { "src/parser.c", "src/scanner.c" },
-    requires_generate_from_grammar  = true,
```
theHamsta, line 388: you would also need to bump the parser revision in lockfile.json

theHamsta (approved):
yes, I checked the versions

PR #1098, opened 2021-03-24
feat(install): ignore parser install option

steelsojka:
This allows a list of parsers to be ignored from install. This is useful when you want 'all' parsers except maybe one that may be having issues or that you just don't want. For example, my scenario, three parsers fail to compile on macos so I want to exclude them.


theHamsta (approved):

steelsojka:
This commit is messed up. Let me fix it

theHamsta:
Just merge it yourself when you think that it's ready

PR #1105, opened 2021-03-26
refactor(all): language tree adaption

steelsojka:
This integrates the language tree into most of the features. This also fixes some other issues I discovered. Memoizing was really messed up and performing a crazy amount of unneeded computations. This will greatly effect the performance of definition lookup, indents, and folds. Indents still aren't working yet. I will continue looking at it.

### DIFF
--- a/lua/nvim-treesitter/fold.lua
+++ b/lua/nvim-treesitter/fold.lua
@@ -1,5 +1,5 @@
 local api = vim.api
-local utils = require'nvim-treesitter.ts_utils'
+local tsutils = require'nvim-treesitter.ts_utils'
 local query = require'nvim-treesitter.query'
 local parsers = require'nvim-treesitter.parsers'
 
@@ -7,18 +7,19 @@ local M = {}
 
 -- This is cached on buf tick to avoid computing that multiple times
 -- Especially not for every line in the file when `zx` is hit
-local folds_levels = utils.memoize_by_buf_tick(function(bufnr)
-  local lang = parsers.get_buf_lang(bufnr)
+local folds_levels = tsutils.memoize_by_buf_tick(function(bufnr)
   local max_fold_level = api.nvim_win_get_option(0, 'foldnestmax')
+  local parser = parsers.get_parser(bufnr)
 
-  local matches
-  if query.has_folds(lang) then
-    matches = query.get_capture_matches(bufnr, "@fold", "folds")
-  elseif query.has_locals(lang) then
-    matches = query.get_capture_matches(bufnr, "@scope", "locals")
-  else
-    return {}
-  end
+  if not parser then return {} end
+
+  local matches = query.get_capture_matches_recursively(bufnr, function(lang)
+    if query.has_folds(lang) then
+      return "@fold", "folds"
+    elseif query.has_locals(lang) then
+      return "@scope", "locals"
+    end
+  end)
 
   local levels_tmp = {}
 
@@ -35,7 +36,6 @@ local folds_levels = utils.memoize_by_buf_tick(function(bufnr)
       levels_tmp[start] = (levels_tmp[start] or 0) + 1
       levels_tmp[stop] = (levels_tmp[stop] or 0) - 1
     end
-
   end
 
   local levels = {}
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -333,12 +333,16 @@ Swaps the nodes or ranges.
 set `cursor_to_second`  to true to move the cursor to the second node
 
 						  *ts_utils.memoize_by_buf_tick*
-memoize_by_buf_tick(fn)~
+memoize_by_buf_tick(fn, options)~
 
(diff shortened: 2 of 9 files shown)


steelsojka:
@kyazdani42 I need someone to check that indents still work... I'm not sure how they work (I don't use them), but I need to make sure it didn't break.

kyazdani42:
I think it would need more for indent to work, i'm testing this out and coming back to you

kyazdani42:
just found out this breaks indent for some filetypes (like graphql), with `set debug=throw` i got `lua/nvim-treesitter/indent.lua:10: att
empt to index local 'root' (a nil value)`.
I was working on some indent improvments too, i'll wait for this PR to be merged before i finish mine.

steelsojka:
I'll check it out

steelsojka:
@kyazdani42 Can you give me an example file that causes that error?

kyazdani42:
```graphql
query getOrder($_id: String) {
  getOrder(_id: $_id) {
    ...OrderAttributes
  }
}

query ordersSearch($currentPage: Int, $perPage: Int, $search: String, $filters: OrderFilters) {
  ordersSearch(
    currentPage: $currentPage
    perPage: $perPage
    search: $search
    filters: $filters
  ) {
    total
    orders {
      ...OrderAttributes
    }
  }
}
```
simple but enough :)

steelsojka:
@kyazdani42 I couldn't reproduce the scenario with the error using graphql. I tried the same thing before my changes and with my changes and got the same results. Is there anything else I may be missing?

lua/nvim-treesitter/fold.lua
```diff
@@ -1,24 +1,29 @@
...
+  local matches = {}
+  parser:for_each_tree(function(tree, lang_tree)
+    local lang = lang_tree:lang()
+
+    if query.has_folds(lang) then
+      vim.list_extend(matches, query.get_capture_matches(bufnr, "@fold", "folds", tree:root(), lang))
+    elseif query.has_locals(lang) then
+      vim.list_extend(matches, query.get_capture_matches(bufnr, "@scope", "locals", tree:root(), lang))
+    end
+  end)
```
theHamsta, line 26: can we have that section as reusable function? Like, get_capture_matches for all trees. I would otherwise just copy this code to nvim-treesitter-textobjects

steelsojka: Sure can. 👍  I'll get that in sometime tomorrow.

steelsojka: Done!

theHamsta:
This would be the implementation for `find_best_match` to make the textobject module fully multi-language capable
```lua
function M.find_best_match(bufnr, capture_string, query_group, filter_predicate, scoring_function)
  if not string.sub(capture_string, 1,2) == '@' then
    api.nvim_err_writeln('capture_string must start with "@"')
    return
  end

  --remove leading "@"
  capture_string = string.sub(capture_string, 2)

  local best
  local best_score

  local parser = parsers.get_parser(bufnr)
  parser:for_each_tree(function(tree, lang_tree)
    for maybe_match in M.iter_group_results(bufnr, query_group, tree:root(), lang_tree:lang()) do
      local match = utils.get_at_path(maybe_match, capture_string)

      if match and filter_predicate(match) then
        local current_score = scoring_function(match)
        if not best then
          best = match
          best_score = current_score
        end
        if current_score > best_score then
          best = match
          best_score = current_score
        end
      end
    end
  end)
  return best
end
```

Don't you think the handling of the language tree is too much for normal users? I mean it could easily be integrated into `iter_group_results` so that we only offer language-tree aware functions (take root into account as well as all injected language that are within root range). The refactor was only very little in the textobject so it's also ok to do the `for each language` loop for each language.

I need to rewrite `find_best_match`, `iter_group_results` anyways to avoid `nested.query.structures` and add a `find_best_matches`

theHamsta (approved):

steelsojka:
@theHamsta I agree. Since you have to rework a little bit of that anyways, lets leave this as is for now and we can handle that in another PR.

theHamsta:
@steelsojka I accidentally pushed my version of find_best_match. Just revert/rebase if needed. I don't want to force-push on your branch.

steelsojka:
@theHamsta Is this good to merge as it? Anymore changes you want me to make?

kyazdani42:
sorry for the delays :) well it works a little better but i can now fix the indents :) thanks for the work !

theHamsta:
Another problem is now that textobjects/indents/folds could be activated because the injected language supports it. E.g. in a svelte file JS textobjects would work but you have to create at least one svelte query in order to have the `has_query('svelte', 'textobjects')` working

steelsojka:
Since the languages are dynamic, you won't be able to detect it during the attach method. The language would be pure svelte until you started typing some javascript. We could tap into the child changes of the root language tree to determine whether we should attach or not. We would have to remove the initial guard though.

PR #1109, opened 2021-03-27
feat: add scss parser

elianiva:
The parser from this tree-sitter-scss repo seems to be working good enough, haven't use it for daily though since I barely work with scss.

Also, please suggest me the proper capture group, I just picked whatever makes sense to me and make the result looks colourful enough :laughing: 

cc @cloggier @euoia @timsofteng

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [ruby](https://github.com/tree-sitter/tree-sitter-ruby) (maintained by @TravonteD)
 - [x] [rust](https://github.com/tree-sitter/tree-sitter-rust) (maintained by @vigoux)
 - [ ] [scala](https://github.com/tree-sitter/tree-sitter-scala)
-- [ ] [scss](https://github.com/serenadeai/tree-sitter-scss)
+- [ ] [scss](https://github.com/elianiva/tree-sitter-scss)
 - [x] [sparql](https://github.com/BonaBeavis/tree-sitter-sparql) (maintained by @bonabeavis)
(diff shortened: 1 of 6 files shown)


timsofteng:
@elianiva how can I test it?

elianiva:
> @elianiva how can I test it?

the simplest way is to install the plugin from this branch and do `:TSInstallFromGrammar scss` :)

theHamsta:
Can't you ask whether they could regenerate the parser? I wouldn't add any parsers that require generate from source and rather remove all those who don't provide C files for their parsers. If the original project is dead you may also create a fork for it.

elianiva:
sure, I'll ask for that and create a fork if I got no reply probably in a week or so. I was originally didn't expect that it would even work so I created this as some sort of proof-of-concept and see what people thinks about it

cloggier:
@elianiva How to test it with Vue parser?

theHamsta:
@cloggier you need to define in `queries/vue/injections.scm` which nodes should be `@scss`

cloggier:
Thanks @theHamsta! Did change it unconditionally to `scss` and it works pretty well (as well on separate scss files), @elianiva.

elianiva:
I'm also thinking of making `postcss` falling back to `scss` since it's possible to do some nesting stuff in `postcss`.

theHamsta:
Any updates on adding the parser without `requires_generate_from_grammar`?

elianiva:
we can use my fork for now, there doesn't seem to be any response from their side

theHamsta:
Merge?

elianiva:
sure

theHamsta (approved):

PR #1120, opened 2021-03-31
Add an alternative implementation to find definitions

theHamsta:
I made an alternative implementation to resolve definitions. I hope it's faster. I did not measure it. So it might be a lie. But when I tested it on larger Python files I didn't encounter any typing problems.

The current implementation does not take advantage of incremental parsing and never releases memory. Incremental updating could be done easily by updating the new region. There is no need to evict old entries so eventually garbage will aggregate in the scope tree and we could do a garbage collection from time to time or just delete the whole tree.

~~EDIT: I think it still has bug. Not all definitions are correct.~~
EDIT: not all local query files are correct :smile:

### DIFF
--- a/lua/nvim-treesitter/locals.lua
+++ b/lua/nvim-treesitter/locals.lua
@@ -5,9 +5,93 @@
 local queries = require "nvim-treesitter.query"
 local ts_utils = require "nvim-treesitter.ts_utils"
 local ts_query = vim.treesitter.query
+local parsers = require'nvim-treesitter.parsers'
 local api = vim.api
+local caching = require'nvim-treesitter.caching'
+local queries = require'nvim-treesitter.query'
 
 local M = {}
+
+local scope_tree = caching.create_buffer_cache()
+
+local function update_scope_tree(cache, buf, parse_node, start_row, end_row)
+  local parser = parsers.get_parser(buf)
+
+  --TODO: do that in a refining manner only updating a specific region?
+  cache.scopes = {}
+  parser:parse()
+  parser:for_each_tree(function(tree, lang_tree)
+    local lang = lang_tree:lang()
+
+    local query = queries.get_query(lang, 'locals')
+
+    local definitions = {}
+    local range = tree:root() and {tree:root():range()}
+    local matches = query
+       and query:iter_matches(parse_node or tree:root(), buf, start_row or range[1], end_row or (range[3] + 1))
+       or function() end
+    for _, match in matches do
+      for id, node in pairs(match) do
+        local name = query.captures[id]
+        if name == "scope" then
+          local found = cache.scopes[node:id()]
+          if found then
+            found.node = node
+          else
+            cache.scopes[node:id()] = { node = node, definitions = {} }
+          end
+        elseif name:find("^definition") then
+          local text = ts_utils.get_node_text(node, buf)[1]
+          table.insert(definitions, {node = node, type = name, text = text})
+        end
+      end
+    end
+
+    for _, d in ipairs(definitions) do
+      local parent = d.node:parent()
+      local found
+      while not found and parent do
+        found = cache.scopes[parent:id()]
+        parent = parent:parent()
+      end
+      if found then
+        found.definitions[d.text] = d
+      end
+    end
+  end)
+end
+
+function M.get_scope_tree_fast(buf)
+  local buf = buf or api.nvim_get_current_buf()
+  local cache = scope_tree.get('scopes', buf)
+  local new_tick = api.nvim_buf_get_changedtick(buf)
+
+  if not cache then
+    cache = {}
+    cache.scopes = {}
+  end
+  if not cache.tick or new_tick > cache.tick then
+    update_scope_tree(cache, buf)
+    cache.tick = new_tick
+    scope_tree.set('scopes', buf, cache)
+  end
+  return cache.scopes
+end
+
+function M.find_definition_fast(buf, node)
+  local scopes = M.get_scope_tree_fast(buf)
+  local search = ts_utils.get_node_text(node, buf)[1]
+  local parent = node:parent()
+  local found, found_scope
+  while not found and parent do
+    found_scope = scopes[parent:id()]
+    if found_scope then
+      found = found_scope.definitions[search]
+    end
+    parent = parent:parent()
+  end
+  return found, found_scope
+end
 
 function M.collect_locals(bufnr)
   return queries.collect_group_results(bufnr, "locals")
--- a/lua/nvim-treesitter/locals.lua
+++ b/lua/nvim-treesitter/locals.lua
@@ -2,7 +2,6 @@
 -- Locals are a generalization of definition and scopes
 -- its the way nvim-treesitter uses to "understand" the code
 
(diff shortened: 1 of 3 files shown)


theHamsta:
In any case, those additional queries will be a opt-in module. Normal people don't have to fear any impact by semantic queries.

steelsojka:
@theHamsta What's the main change in the algorithm? Is there any downside compared to the current implementation?

theHamsta:
- avoiding `iter_prepared_matches` and `is_definition` (avoid allocation and destructuring of unnecessary tables, this should be the main speed up)
- using dictionaries instead of lists for look up
- effectively only scopes and their definitions are stored. They are arranged in a tree before searching for definitions. The old algorithm did not build up a scope tree and just cached the results of exhaustive searches.The cache stored the relation between references to definitions. With the new approach, the references are not stored in any cache but finding a definition is only dictionary look ups:
``` while not found: look up parent by node:id until we have a scope, then look up definitions in scope by own node text```
with string being interned in Lua we're just looking up pointers in dictionaries, no other function calls or logic involved 

 - When we would build up the scope tree on each changed tick, then it would work similar as the old implementation. When we would do partial updates of the trees then we would have to do some sort of garbage collection from time to time 
- the current cache only contains the definitions and scopes in a dictionary, definitions and references could be stored when we want to iterate over them. Right now the tree is not iterable and only made for look up of the definition of a particular node

We should probably deprecate `iter_prepared_matches`.

**Comment by steelsojka** [comment]
Awesome! Thank you for that rundown. This will indeed be much faster. Maybe I'll be able to turn back on reference highlighting....

**Comment by steelsojka** [comment]
So once this goes in we can update the refactor module to use this api correct?

**Comment by theHamsta** [comment]
@steelsojka we could, I tried to start this. But there is so much that relies on `iter_prepared_matches` and how it arranges data. So it could involve some work. It would also be possible to just use this for the purpose of "find definition" and "find usages"

**Comment by theHamsta** [comment]
Using it on a  80k C++ file with highlight_definition and updatetime=100 no problem :sunglasses: .
```
function M.highlight_usages(bufnr)
  M.clear_usage_highlights(bufnr)

  local node_at_point = ts_utils.get_node_at_cursor()
  local references = locals.get_references(bufnr)

  if not node_at_point or not vim.tbl_contains(references, node_at_point) then
    return
  end

  local def, scope = locals.find_definition_fast(node_at_point, bufnr)
  if not def or not scope then return end
  local usages = locals.find_usages_fast(def.node, scope.node, bufnr)

  for _, usage_node in ipairs(usages) do
    if usage_node ~= node_at_point then
      ts_utils.highlight_node(usage_node, bufnr, usage_namespace, 'TSDefinitionUsage')
    end
  end

  if def_node ~= node_at_point then
    ts_utils.highlight_node(def_node, bufnr, usage_namespace, 'TSDefinition')
  end
end
```

**Comment by theHamsta** [comment]
Am advantage of this implementation is that the last definition (if multiple) is taken... needs to be fixed

**Comment by vigoux** [comment]
I am in favor of replacing the current implementation to use this scope tree, there's no need to have both of them.

Updating the other plugins could be done simultaneously.

**Comment by steelsojka** [comment]
Might be a good idea to just deprecate the other api and remove it when the
others are migrated.

**Comment by theHamsta** [comment]
I think this needs some time to find a good solution. It's also easier to find a good data structure for a specific purpose than to find something to can be used in a more general setting. Also the "tree" is a bit of a lie, since you can walk from all children to the parent nodes but I did not add the back references to use it as a tree. Again, that would only be necessary for some applications like a document tree. I don't know a good solution yet.

**Comment by ahelwer** [comment]
@theHamsta could you please rebase this PR on top of the latest changes in master branch so I can start working off of it?

**Comment by theHamsta** [comment]
@ahelwer I've rebased it. But I have not tested it. I don't know whether a rewrite would be easier.

**Comment by ahelwer** [comment]
@theHamsta any sort of incremental scope update will be very difficult to do. Consider a tla+ file like:
```tla
---- MODULE Test ----
def1 == x
def2 == y
def3 == z
====
```
We then add four characters somewhere in the middle:
```tla
---- MODULE Test ----
def1 == x
====
def2 == y
def3 == z
====
```
The `====` marks the end of the module and thus affects the parsing of the entire second half of the source code, which is parsed as syntactically-meaningless freeform text. So really we're concerned with what aspect of the parse tree is updated, rather than what aspect of the text. I don't know whether tree-sitter provides any sort of tree-diffing functionality.

Incidentally, the snippet highlighting above is provided by the TLA+ tree-sitter grammar, now built into github!

lua/nvim-treesitter/locals.lua
```diff
@@ -2,13 +2,104 @@
...
+    cache.scopes = {}
+    cache.references = {}
+  end
+  if not cache.tick or new_tick > cache.tick then
+    update_scope_tree(cache, buf)
+    cache.tick = new_tick
+    scope_tree.set('scopes', buf, cache)
+  end
+  return cache.scopes, cache.references
+end
```
vigoux, line 85: Why not using one of our `memoize_by_buf_tick` here ?

lua/nvim-treesitter/locals.lua
```diff
@@ -2,13 +2,104 @@
...
 
 local M = {}
 
+local scope_tree = caching.create_buffer_cache()
+
+local function update_scope_tree(cache, buf, parse_node, start_row, end_row)
+  local parser = parsers.get_parser(buf)
+
+  --TODO: do that in a refining manner only updating a specific region?
+  cache.scopes = {}
```
vigoux, line 20: I think that, in order to do so, we'll have to use a better data structure, indexed by lines.

vigoux (commented):
Looks pretty nice, I am going to try it.
Why not using it be default?

vigoux:
@ahelwer there is a thing in tree-sitter that is some kind of "syntax-changed event".
I'll try to see if I can invalidate parts of the buffer based on that.

lua/nvim-treesitter/locals.lua
ahelwer: @vigoux don't think that would work; see comment above.

theHamsta: or something like this https://en.wikipedia.org/wiki/Interval_tree

theHamsta:
> ny sort of incremental scope update will be very difficult to do. Consider a TLA+ file like:

we'll catastrophic invalidation can happen. They also happen today for highlighting. But they shouldn't happen on every keystroke, plus a long calculation should always have a  timeout.

theHamsta:
> Looks pretty nice, I am going to try it.
> Why not using it be default?

This PR is now a year old. It was before `:parent` was discouraged. I would probably build the data structure top-down now instead of only calling `:parent()`.

lua/nvim-treesitter/locals.lua
vigoux: Did not know about it, but that looks like something we'd use, yes

clason:
superseded by rewrite on `main` (which drops all non-install functionality); such functionality should be refactored into a separate plugin (or included in `nvim-treesitter-refactor`?)

PR #1122, opened 2021-04-01
Add node requirement

theHamsta:
- Error when `node` is not available on `:TSInstallFromGrammar`
- exclude `requires_install_from_grammar` parsers from `all`/`maintained` when user has no tree-sitter CLI

Will the parsers be updated to 0.19.X at some point? @cbarrete @undu  @ostera @madskjeldgaard @leo60228

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -84,7 +84,7 @@
     "revision": "ea43db6830632fd3531b9cbc34a93502b0d4339a"
   },
   "ledger": {
-    "revision": "609d5e5ab5955823b3faeaec8d2afc91860c639a"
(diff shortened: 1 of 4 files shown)


kyazdani42 (approved):

cbarrete:
The version of https://github.com/cbarrete/treesitter-ledger that nvim-treesitter has been using for a while is on 0.19.2. Is something wrong with it?

theHamsta:
It had a while only 0.19.2 in package.json. But now I see that also the C file was updated. Thanks for the update I will remove the `requires_generate_from_grammar`

Issue #1110, opened 2021-03-27
I'm having a problem when installing treesitter


briones-gabriel:
**Describe the bug**

When trying to install treesitter, there is an ABI version mismatch.

**To Reproduce**

1. Install neovim
2. Add treesitter with vim-plug
3. Install a language
4. Enable the highlight module

**Expected behavior**

**Output of `:checkhealth nvim_treesitter`**

<details>
<code>

1 health#nvim_treesitter#check                                                                                            
  2 ========================================================================                                                
  3 ## Installation                                                                                                         
  4   - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
  5   - OK: `git` executable found.                                                                                         
  6   - OK: `cc` executable found.                                                                                          
  7                                                                                                                         
  8 ## Parser/Features H L F I                                                                                              
  9   - c              ✓ ✓ ✓ ✓                                                                                              
 10   - java           x x . x                                                                                              
 11                                                                                                                         
 12  Legend: H[ighlight], L[ocals], F[olds], I[ndents]                                                                      
 13          +) multiple parsers found, only one will be used                                                               
 14          x) errors found in the query, try to run :TSUpdate {lang}                                                      
 15                                                                                                                         
 16 The following errors have been detected:                                                                                
 17   - ERROR: java(highlights): /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:33: ABI version mismatch for /home/gabriel/.vim/plugged/nvim-treesitter/parser/java.so: supported          between 9 and 12, found 13
 18   - ERROR: java(locals): /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:33: ABI version mismatch for /home/gabriel/.vim/plugged/nvim-treesitter/parser/java.so: supported              between 9 and 12, found 13
 19   - ERROR: java(indents): /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:33: ABI version mismatch for /home/gabriel/.vim/plugged/nvim-treesitter/parser/java.so: supported             between 9 and 12, found 13

</code>
</details>

**Output of `nvim --version`**

<details>
<code>

NVIM v0.5.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
[...]

PR #1123, opened 2021-04-01
ABI healthcheck

theHamsta:
This should help users @przepompownia 

Ref: https://github.com/nvim-treesitter/nvim-treesitter/issues/1110#issuecomment-811775871

### DIFF
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -11,6 +11,8 @@ local health_warn = vim.fn['health#report_warn']
 
 local M = {}
 
+local NVIM_TREESITTER_MINIMUM_ABI = 13
+
 local function install_health()
   health_start('Installation')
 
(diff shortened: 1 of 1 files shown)


vigoux:
Hmmm, not sure that's the way we should go... Maybe doing the check for ABI compatibility in Lua side for each parser would be better.

theHamsta:
But the ABI check is already done (also shown in the individual error messages for each parser). The issue is that in case of a miss-match users are confused what to do. This part of the health-check is rather about our plugin providing the Git SHAs for 13 ABI. So the user should rather upgrade Neovim instead of downgrading the parsers.

In future, we can avoid ABI chaos by just versioning our lockfile. So when a new ABI comes our we will just open a new section in our lockfile with the new ABI version while the old one remains unchanged and can still be used without problems.

lua/nvim-treesitter/health.lua
```diff
@@ -25,6 +27,17 @@ local function install_health()
...
+  if fn.executable('node') == 0 then
+    health_warn('`node` executable not found (only needed for :TSInstallFromGrammar,'..
+                ' not required for :TSInstall)')
+  else
+    local handle = io.popen('node --version')
+    local result = handle:read("*a")
+    handle:close()
+    local version = vim.split(result,'\n')[1]
+    health_ok('`node` found '..version..' (only needed for :TSInstallFromGrammar)')
+  end
```
kyazdani42, line 39: thinks i've seen that in another PR :)

kyazdani42 (commented):
hum looks like this PR has some node check that is in another PR you've made, could you either close the other or fix this one :) ?

lua/nvim-treesitter/health.lua
theHamsta: rebased! :smile:

theHamsta: not yet. remove the wrong commits in the wrong PR

kyazdani42 (approved):
:+1:

vigoux (approved):

PR #1127, opened 2021-04-01
fix: indents

kyazdani42:
Well let me say one thing:
It is quite hard to make indent work for multiple languages. So many edge cases.

ok so what i've noted so far:
- by getting the last node on the last non empty line, it fixes rust else branch because the matched node was the `}` from the previous node
- but it doesn't fix the `else` branch for go
- python works on proper files, but opening a new one and entering `def main():<CR>` does not properly indent
- for this little hack to work, i had to make a special node table for xml like file (html, jsx) because last nodes are always `text` which wraps at the next line.
- also unrelated to this PR, but on typescript files, doing `o` after function calls like:
```typescript
func()
  .doSomething()
  .doOtherThing(' ');
```
will indent just like the previous line although we want to indent like the `func` line.
- oh and open a 3K lines file and typing `o` is terribly slow now. This whole algorithm takes too much time. Maybe it is too complex and we did not find the best alternative, although i'm not sure if there is one :(

### DIFF
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -36,6 +36,7 @@ local get_indents = tsutils.memoize_by_buf_tick(function(bufnr, root, lang)
     branches = get_map('@branch.node'),
     returns = get_map('@return.node'),
     ignores = get_map('@ignore.node'),
+    ignores_end = get_map('@ignore_end.node'),
   }
 end, {
   -- Memoize by bufnr and lang together.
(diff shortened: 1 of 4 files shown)


CaptainKO:
> * also unrelated to this PR, but on typescript files, doing `o` after function calls like:
> 
> ```ts
> func()
>   .doSomething()
>   .doOtherThing(' ');
> ```
> 
> will indent just like the previous line although we want to indent like the `func` line.
> 
> * oh and open a 3K lines file and typing `o` is terribly slow now. This whole algorithm takes too much time. Maybe it is too complex and we did not find the best alternative, although i'm not sure if there is one :(

@kyazdani42, Thank u for confirming it. I thought there's something wrong with my config file :(

theHamsta:
@kyazdani42 is this still relevant?

kyazdani42:
I'll close this, this is really old and the implementation was clunky and slow :smile:

PR #1137, opened 2021-04-05
Improve :TSModuloInfo output

mhinz:
Just a few small changes to make it more pleasant to read (IMHO):

- 2 spaces between modules in the header
- always same module order in the header (random order is super confusing)
- marks are exactly below the start of the module name in the header
- most importantly.. sort the languges by name. :P

Before:

After:

### DIFF
--- a/lua/nvim-treesitter/info.lua
+++ b/lua/nvim-treesitter/info.lua
@@ -37,15 +37,20 @@ local function print_info_module(sorted_languages, mod)
   end
 end
 
-local function print_info_modules(sorted_languages)
-  local max_str_len = #sorted_languages[1]
+local function print_info_modules(parserlist)
+  table.sort(parserlist, function(a, b) return #a > #b end)
+  local max_str_len = #parserlist[1]
+
   local header = string.rep(' ', max_str_len + 2)
-  for _, mod in pairs(configs.available_modules()) do
+  local mods = configs.available_modules()
+  table.sort(mods)
+  for _, mod in pairs(mods) do
     header = string.format('%s%s ', header, mod)
   end
   api.nvim_out_write(header..'\n')
 
-  for _, lang in pairs(sorted_languages) do
+  table.sort(parserlist)
+  for _, lang in pairs(parserlist) do
     local padding = string.rep(' ', max_str_len - #lang)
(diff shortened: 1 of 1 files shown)


theHamsta:
I thought that it might be good to just list the activated module per line. `cpp: indent ✓, highlight ✗`

Would it be possible to summarize submodules like `refactor.{smart_rename,highlight_current_scope,...}` in one table cell? This is how it looks on my machine 
image

theHamsta (approved):

lua/nvim-treesitter/info.lua
```diff
@@ -37,28 +37,30 @@ local function print_info_module(sorted_languages, mod)
...
-local function print_info_modules(sorted_languages)
-  local max_str_len = #sorted_languages[1]
-  local header = string.rep(' ', max_str_len + 2)
-  for _, mod in pairs(configs.available_modules()) do
-    header = string.format('%s%s ', header, mod)
-  end
-  api.nvim_out_write(header..'\n')
+local function print_info_modules(parserlist)
+  table.sort(parserlist, function(a, b) return #a > #b end)
+  local max_str_len = #parserlist[1]
```
theHamsta, line 42: maybe limit `max_str_len`? `incremental_selection` -> `incremental_sel...`

theHamsta (approved):

theHamsta:
Especially the sorting per language was missing. Thank you a lot!

mhinz:
Oh, wow. That's a lot of modules. I didn't expect that.

To be honest, my main intention in the beginning was to add green/red colors, to make it easier to see differences immediately, because distinguishing between ✓/✗ can be hard with a small font. But I got weird behaviour switching between using `nvim_out_write()` and `nvim_command()` (for `echohl HLGroup | echon "text" | echohl NONE`) etc.

Instead of filling the message history, maybe it would be better to use an own buffer for this all. Scrollable, with proper highlighting. Then we would have multiple tables. The first one for "only top-level" modules followed by tables for all the submodules, so `refactor`, `textobjects`, etc. in your case.

So, maybe let's merge this for now and then I'll open a new PR for completely refactoring `:TSModuleInfo`.

PR #1139, opened 2021-04-05
feat: add omnifunc for query filetype

theHamsta:
A simple omnifunc for the query filetype.
noidea

### DIFF
--- a/autoload/nvim_treesitter.vim
+++ b/autoload/nvim_treesitter.vim
@@ -21,3 +21,14 @@ endfunction
 function! nvim_treesitter#indent() abort
 	return luaeval(printf('require"nvim-treesitter.indent".get_indent(%d)', v:lnum))
 endfunction
+
+function! nvim_treesitter#query_omnifunc(findstart, base) abort
+  if a:findstart
+    " locate the start of the word
+    let start = col('.') - 2 " -2 necessary to exclude current cursor position and to transfer into 0-based indexing
+    let result = matchstrpos(getline('.')[0:start], '\v("|\\|\-|\w)*$')
+    return result[1]
+  else
+    return luaeval(printf("require'nvim-treesitter.query_filetype'.omnifunc([[%s]])", a:base))
+  endif
+endfunction
--- a/after/ftplugin/query.vim
+++ b/after/ftplugin/query.vim
@@ -1,1 +1,2 @@
 setlocal commentstring=;\ %s
+setlocal omnifunc=nvim_treesitter#query_omnifunc
--- a/lua/nvim-treesitter/query_filetype.lua
+++ b/lua/nvim-treesitter/query_filetype.lua
@@ -1,0 +1,36 @@
+local M = {}
+
+function M.omnifunc(base)
+  local filename = vim.api.nvim_buf_get_name(vim.api.nvim_get_current_buf())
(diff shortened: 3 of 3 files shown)


autoload/nvim_treesitter.vim
```diff
@@ -21,3 +21,17 @@ endfunction
 function! nvim_treesitter#indent() abort
 	return luaeval(printf('require"nvim-treesitter.indent".get_indent(%d)', v:lnum))
 endfunction
+
+function! nvim_treesitter#query_omnifunc(findstart, base) abort
+  if a:findstart
+    " locate the start of the word
+    let line = getline('.')
+    let start = col('.') - 1
+    while start > 0 && ((line[start - 1] =~ '\a' || line[start - 1] =~ '-' || line[start - 1] =~ '"') && line[start - 1] !~ '(' )
```
theHamsta, line 30: should probably be a `strmatchpos`

autoload/nvim_treesitter.vim
```diff
@@ -21,3 +21,14 @@ endfunction
 function! nvim_treesitter#indent() abort
 	return luaeval(printf('require"nvim-treesitter.indent".get_indent(%d)', v:lnum))
 endfunction
+
+function! nvim_treesitter#query_omnifunc(findstart, base) abort
+  if a:findstart
+    " locate the start of the word
+    let start = col('.') - 2
```
theHamsta, line 28: -2. Why?

theHamsta: Ok, without the current cursor position and transform into 0-based indexing

autoload/nvim_treesitter.vim
```diff
@@ -21,3 +21,14 @@ endfunction
...
 endfunction
+
+function! nvim_treesitter#query_omnifunc(findstart, base) abort
+  if a:findstart
+    " locate the start of the word
+    let start = col('.') - 2 " -2 necessary to exclude current cursor position and to transfer into 0-based indexing
+    let result = matchstrpos(getline('.')[0:start], '\v("|\\|\-|\w)*$')
+    return result[1]
+  else
+    return luaeval(printf("require'nvim-treesitter.query_filetype'.omnifunc([[%s]])", a:base))
```
theHamsta, line 32: can this be done in a way that will not break when you enter `]]`?

vigoux:
I did some things in `architext.nvim` that I actually never ported, mayde that'll help a bit.

theHamsta:
@vigoux then, we should use your code! Maybe you can take the `vim.fn.escape` part of my PR to escape backslashes?

vigoux:
Yup, that looks like a great idea.
I am starting to think that we should host all the query-related things into a custom plugin, which we could then reuse.

theHamsta:
We are also missing a `:TSEditQuery` which opens the appropriate query file and prompts user to select if there are multiple.

theHamsta:
Should we leave this open until we have a proper solution? Maybe move this to playground? I think that this feature is so minor that most users won't bother to install a plugin for it but are just happy if it just works.

vigoux:
Maybe moving that to the playground is better, this makes the plugin more like a dev-tools plugin for nvim-treesitter, and we can point to that in the contributing and things.

PR #1143, opened 2021-04-05
:TSModuleInfo rewrite

mhinz:
The new version uses its own buffer for listing the modules state for the current session.

Since there can be many modules and many using submodules, the code sorts them into namespaces. A list with the following modules..

    {'indent', 'highlight', 'refactor.smart_rename', 'refactor.navigation'}

..will be changed to..

    { default = {'indent', 'highlight'}, refactor = {'smart_rename', 'navigation'} }

There is one table per namespace.

The colors are up for debate, but I added the highlight groups to the doc, in case people don't like the defaults.

### DIFF
--- a/lua/nvim-treesitter/info.lua
+++ b/lua/nvim-treesitter/info.lua
@@ -21,28 +21,12 @@ local function install_info()
   end
 end
 
-local function print_info_module(sorted_languages, mod)
-  local max_str_len = #sorted_languages[1]
-  local header = string.format('%s%s', string.rep(' ', max_str_len + 2), mod)
-  api.nvim_out_write(header..'\n')
-  for _, lang in pairs(sorted_languages) do
-    local padding = string.rep(' ', max_str_len - #lang + #mod / 2 + 1)
-    api.nvim_out_write(lang..":"..padding)
-    if configs.is_enabled(mod, lang) then
-      api.nvim_out_write('✓')
-    else
-      api.nvim_out_write('✗')
-    end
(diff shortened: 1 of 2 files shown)


theHamsta:
Beautiful :heart: 

But the status for non-builtin modules will always be false...

mhinz:
> But the status for scoped modules will always be false

Fixed!

theHamsta (approved):

lua/nvim-treesitter/info.lua
```diff
@@ -21,57 +21,113 @@ local function install_info()
...
+local function print_info_modules(parserlist, module)
+  api.nvim_command('enew')
+  local curbuf = api.nvim_get_current_buf()
+
+  local modules
+  if module then
+    modules = namespace_modules({module})
+  else
+    modules = namespace_modules(configs.available_modules())
+  end
```
kyazdani42, line 98: cool cool.
a little lua trick here:
```lua
local modules = module and { module } or config.availabe_modules
```
works more like a ternary but if/else is fine too :)

kyazdani42 (approved):

lua/nvim-treesitter/info.lua
mhinz: Oh, right. I even used it [here](https://github.com/nvim-treesitter/nvim-treesitter/pull/1143/files#diff-971a54c46b5207639b94c6a7603138567fcbaa268e87439eab2723371bea2519R72).

Still getting used to Lua and its idoms and best practices. :)

PR #1160, opened 2021-04-07
feat: Allow to enabled_langs for a module (inverse of disable)

theHamsta:


vigoux:
Don't `enable` already does that? In any case, it should.

theHamsta:
It doesn't. I thought about reusing `enable` for that but our code base is to messy to allow that without any major refactorings. Had this approach before.

The issue is that `enable`  has boolean behavior for toggle which means that reusing `enable` won't be persistent after a `:TSDisableAll` `:TSEnableAll`. I can move the initial value of enable to `enable_langs` under the hood to persist this setting.

theHamsta:
Look for example at this code. It looks important and busy. It is never executed
https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/configs.lua#L284-L289

lua/nvim-treesitter/configs.lua
```diff
@@ -280,13 +287,6 @@ function M.setup(user_data)
   end
 
   config.modules.ensure_installed = nil
-
-  recurse_modules(function(_, _, new_path)
-    local data = utils.get_at_path(config.modules, new_path)
-    if data.enable then
-      enable_all(new_path)
-    end
-  end, config.modules)
```
theHamsta, line 289: If this code would be important it would be actually executed. Might be in some configs. But each module defines an autocmd at init to actually enable it.

lua/nvim-treesitter/configs.lua
```diff
@@ -261,8 +264,12 @@ function M.is_enabled(mod, lang)
     return false
   end
 
-  for _, parser in pairs(module_config.disable) do
-    if lang == parser then return false end
+  if type(module_config.enable) == 'table' then
```
theHamsta, line 267: the root of the problem is that we use `enable` both as a config variable and a state variable. I guess there is already some logic that assumes that `enable` is a state, so it's difficult to decouple

theHamsta:
Be aware that any change in the `config` module is very likely to break something for some specific config.

lua/nvim-treesitter/configs.lua
```diff
@@ -135,6 +135,9 @@ local function disable_all(mod)
   end
 
   disable_mod_conf_autocmd(mod)
+  if type(config_mod.enable) == 'table' then
+    config_mod.enabled_langs = config_mod.enable
+  end
```
theHamsta, line 140: lets move the config value, to a separate value to protect it from state changes. This is not nice but is due to the fact that after this change `enable` will be a config and a state variable in user config but our code requires a separations of those concerns.

theHamsta: Alternative would be to use `currently_enabled` for the state. This might be confusing for everyone relying on the old behavior and has to risk of confusing state and config in some places.

theHamsta: We can not do that in setup or define_modules because we don't know which of the ones happens first and setup does not know which of the modules actually end up being modules.

kyazdani42:
i agree with you, at some point i wanted to refactor this config module but it was so big and messy i got lost in the process x)

theHamsta:
@kyazdani42 The module is well tested by so many users. So it seems to work. I'm just saying that I would avoid making too many changes in that file.

kyazdani42:
i understand :) the module has many bugs though, usually using the module enable/disable command fails on my end and the UX behind it is not very logical. The setup stuff is working fine though.

kyazdani42 (approved):
looks right, i'm approving this although i'd like @vigoux to review this as well before we merge :)

vigoux:
Although I agree on the functionality, I find the implementation not nice.
I'd rather go for :
1. A big set of tests for the config module
2. A big refactor of the config module

theHamsta:
@vigoux then we probably should do the refactor first.

vigoux:
Nope, the tests first, to be at least a bit sure everything work as we intended.

theHamsta:
I meant refactor before PR.

seblj:
Any updates on this?

theHamsta:
From my side we would integrate this even if the current situation is not ideal.

I created project board for tasks that we're left open for quite some time and that we need to tackle sooner or latter. There are quite some issues connected to the handling of modules. Some simplification of the module logic would help here.

clason:
superseded by rewrite on `main` branch (which removes modules altogether)

PR #1162, opened 2021-04-07
feat: Add TSEditQuery

theHamsta:
~~This is currently affected by the runtime file duplication bug:
https://github.com/neovim/neovim/issues/14294~~

EDIT: used the upstream method that does the dedup and include inherited languages.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -1,6 +1,7 @@
 local api = vim.api
 
 local queries = require'nvim-treesitter.query'
+local nvim_query = require'vim.treesitter.query'
 local parsers = require'nvim-treesitter.parsers'
 local utils = require'nvim-treesitter.utils'
 local caching = require'nvim-treesitter.caching'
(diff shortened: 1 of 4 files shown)


lua/nvim-treesitter/configs.lua
```diff
@@ -245,6 +265,13 @@ M.commands = {
       "-nargs=0",
     },
   },
+  TSEditQuery = {
+    run = M.edit_query_file,
+    args = {
+      "-nargs=+",
+      "-complete=custom,nvim_treesitter#available_query_groups",
+    },
+  },
```
theHamsta, line 274: Maybe a version of this command to add/edit a query extension in the users config directory?

vigoux: I think the default version of the command should ask if the user wants to create a new extension if it does not already exists.

lua/nvim-treesitter/configs.lua
```diff
@@ -1,6 +1,7 @@
 local api = vim.api
 
 local queries = require'nvim-treesitter.query'
+local nvim_query = require'vim.treesitter.query'
```
vigoux, line 4: WTF we need clearer names

lua/nvim-treesitter/configs.lua
```diff
@@ -196,6 +197,30 @@ local function config_info(process_function)
...
+      vim.cmd(':edit '..files[choice])
+    end
+  end
+end
+
+function M.edit_query_file_user_after(query_group, lang)
+  lang = lang or parsers.get_buf_lang()
+  local folder = utils.join_path(vim.fn.stdpath('config'), 'after', 'queries', lang)
+  local file = utils.join_path(folder, query_group..'.scm')
+  pcall(vim.fn.mkdir, folder, "p", "0755")
```
vigoux, line 220: Error report here ?

lua/nvim-treesitter/query.lua
```diff
@@ -23,6 +23,19 @@ for _, query in ipairs(M.built_in_query_groups) do
...
+  for _, f in ipairs(query_files) do
+    groups[vim.fn.fnamemodify(f, ':t:r')] = true
+  end
+  local list = {}
+  for k, _ in pairs(groups) do
+    table.insert(list, k)
+  end
+  return list
+end
+
```
vigoux, line 38: Shouldn't we just use the default query groups ?

lua/nvim-treesitter/configs.lua
theHamsta: Edit does not really create a file. So I think we should only ask for the folder.

lua/nvim-treesitter/query.lua
theHamsta: Why? It works for all query groups of all plugins.

vigoux: Yup, I just thought about that after hitting enter, it sounds better this way.

vigoux (approved):

PR #1186, opened 2021-04-14
rebased AdnoC/windows-tar-workaround on master

Thanood:
This PR is a rebase of https://github.com/AdnoC/nvim-treesitter/tree/windows-tar-workaround as requested in #813.
This is a quick rebase, I've tried to keep the intended logic in place but I'm not exactly a git guru. 😃

There was one change introduced in 60e67de18c69d57691b09738585d58515458a281 which introduces the use of  `'-c', 'core.symlinks=true'` which caused the clone to fail again so I deleted that line.

I hope, that's okay. 😃 
At least I could now run `:TSUpdate`  and successfully update tree-sitter-typescript.

### DIFF
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -187,7 +187,6 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
         opts = {
           args = {
             'clone',
-            '-c', 'core.symlinks=true',
             '--single-branch',
             '--branch', repo.branch or 'master',
             repo.url,
@@ -202,7 +201,6 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
         err = 'Error while checking out revision',
         opts = {
           args = {
-            '-c', 'core.symlinks=true',
             'checkout', revision
           },
           cwd = git_folder
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -121,15 +121,13 @@ end
 
 function M.select_download_commands(repo, project_name, cache_folder, revision)
 
-  local has_tar = vim.fn.executable('tar') == 1
-  local has_curl = vim.fn.executable('curl') == 1
-  local is_github = repo.url:find("github.com", 1, true)
-  local is_gitlab = repo.url:find("gitlab.com", 1, true)
+  local can_use_tar = vim.fn.executable('tar') == 1 and vim.fn.executable('curl') == 1
+  local is_github_or_gitlab = repo.url:find("github.com", 1, true) or repo.url:find("gitlab.com", 1, true)
   local is_windows = fn.has('win32') == 1
 
   revision = revision or repo.branch or "master"
 
-  if has_tar and has_curl and (is_github or is_gitlab) and not is_windows then
+  if can_use_tar and is_github_or_gitlab and not is_windows then
 
     local path_sep = utils.get_path_sep()
     local url = repo.url:gsub('.git$', '')
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -185,8 +185,6 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
         opts = {
           args = {
             'clone',
-            '--single-branch',
-            '--branch', repo.branch or 'master',
             repo.url,
             project_name
           },
@@ -199,7 +197,7 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
         err = 'Error while checking out revision',
         opts = {
           args = {
-            'checkout', revision
+            'checkout', revision,
           },
           cwd = git_folder
         }
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -175,9 +175,6 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
   else
     local git_folder = utils.join_path(cache_folder, project_name)
     local clone_error = 'Error during download, please verify your internet connection'
-    if is_windows then
-      clone_error = clone_error .. ". If on Windows you may need to enable Developer mode"
-    end
 
     return {
       {
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -141,7 +141,7 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
         opts = {
           args = {
             '-L', -- follow redirects
-            is_github and url.."/archive/"..revision..".tar.gz"
+            is_github_or_gitlab and url.."/archive/"..revision..".tar.gz"
                       or url.."/-/archive/"..revision.."/"..project_name.."-"..revision..".tar.gz",
             '--output',
             project_name..".tar.gz"
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -122,12 +122,14 @@ end
 function M.select_download_commands(repo, project_name, cache_folder, revision)
 
   local can_use_tar = vim.fn.executable('tar') == 1 and vim.fn.executable('curl') == 1
-  local is_github_or_gitlab = repo.url:find("github.com", 1, true) or repo.url:find("gitlab.com", 1, true)
+  local is_github = repo.url:find("github.com", 1, true)
+  local is_gitlab = repo.url:find("gitlab.com", 1, true)
+
   local is_windows = fn.has('win32') == 1
 
   revision = revision or repo.branch or "master"
 
-  if can_use_tar and is_github_or_gitlab and not is_windows then
+  if can_use_tar and (is_github or is_gitlab) and not is_windows then
 
     local path_sep = utils.get_path_sep()
     local url = repo.url:gsub('.git$', '')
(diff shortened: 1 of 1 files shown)


Thanood:
Some locked revision checkouts fail with `fatal: reference is not a tree: 02b01653c8a1c198ae7287d566efa86a135b30d5` - but the check succeeds anyway.

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -169,22 +173,38 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
...
         cmd = 'git',
         info = 'Downloading...',
-        err = 'Error during download, please verify your internet connection',
+        err = clone_error,
         opts = {
           args = {
             'clone',
             '--single-branch',
             '--branch', repo.branch or 'master',
-            '--depth', '1',
```
kyazdani42, line 182: depth 1 was used to only download the commit required, this will basically download a whole branch of commits ? what was the aim of removing that ?

theHamsta: To make the pinning work. `--depth=1` only works for branches

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -121,12 +121,16 @@ end
...
+  local has_tar = vim.fn.executable('tar') == 1
+  local has_curl = vim.fn.executable('curl') == 1
   local is_github = repo.url:find("github.com", 1, true)
   local is_gitlab = repo.url:find("gitlab.com", 1, true)
+  local is_windows = fn.has('win32') == 1
 
-  if vim.fn.executable('tar') == 1 and vim.fn.executable('curl') == 1 and (is_github or is_gitlab) then
+  revision = revision or repo.branch or "master"
+
+  if has_tar and has_curl and (is_github or is_gitlab) and not is_windows then
```
theHamsta, line 132: Could this logic be a bit more encapsulated? It's a bit difficult to understand when the windows conditions is added additionally. Maybe just introduce for `has_tar and has_curl and (is_github or is_gitlab)` a new variable, e.g. `can_use_tar`?

`if can_use_tar and not is_windows` is a bit more compact.

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -169,22 +173,38 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
       M.select_install_rm_cmd(cache_folder, project_name..'-tmp')
     }
   else
+    local git_folder = utils.join_path(cache_folder, project_name)
+    local clone_error = 'Error during download, please verify your internet connection'
+    if is_windows then
+      clone_error = clone_error .. ". If on Windows you may need to enable Developer mode"
```
theHamsta, line 179: This developer mode was a bit off in the other PR. Wasn't this developer mode only to have the symlinks in Git but actually nobody needs the symlinks? Only the corpus is symlinked. The symlinks were only a problem because tar would crash because of them. For me, L178-180 are not really needed. First, because based on the discussion in the previous PR git won't error and second because this is probably not the likely cause why a download might have failed on Windows (also I wouldn't know how this dev mode can be enabled).

> After some investigation: it appears that git does not resolve symlinks either (by default). It just creates a file containing the path to the file it is linking as text. Git doesn't error, but I assume it caused typescript treesitter to misbehave.

theHamsta:
Thanks for taking this effort. This will save Windows users a lot of time!

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -169,22 +173,38 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
...
       {
         cmd = 'git',
         info = 'Downloading...',
-        err = 'Error during download, please verify your internet connection',
+        err = clone_error,
         opts = {
           args = {
             'clone',
             '--single-branch',
             '--branch', repo.branch or 'master',
```
theHamsta, line 191: Can you try whether removing `--single-branch` and `--branch` fixes the installation problems with Gitlab? They should not be needed when the version is selected later with the checkout.

theHamsta: If possible, only try remove `--single-branch` i guess this could be the culprit why all `main` repos fail on windows CI.

theHamsta: But thinking about it... the branch was only needed to enable `--depth=1`

lua/nvim-treesitter/shell_command_selectors.lua
Thanood: https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/#oTepts0M8148xRoj.97

The reason seems to be that with developer mode you don't need Admin permissions to create symlinks.

lua/nvim-treesitter/shell_command_selectors.lua
Thanood: I've changed this in my latest commit according to your suggestions. Thanks. 👍 
I've added `is_github_or_gitlab`, though.

lua/nvim-treesitter/shell_command_selectors.lua
Thanood: To be honest, I'm a little confused regarding what to do here. Especially considering the first review comment. 😉 

is `--branch` not needed anymore because `--depth=1` was deleted anyway?

lua/nvim-treesitter/shell_command_selectors.lua
Thanood: > also I wouldn't know how this dev mode can be enabled

Update and Security/or developers, second option 😃 
I can try if this setting solves the problem.

But I'm not sure if I like that solution tbh.
"A 3rd party package is probably using symlinks in its repository, please turn on developer mode to install your desired parser" ... sounds a little strange to me. 😉 

As this only happens on updates, maybe we can instead (or as an alternative) tell the user to start neovim in an elevated shell for the update?

theHamsta: > '-c', 'core.symlinks=true'

Why? It only causes problems. For the user it's absolutely irrelevant whether symlinks work or not. So why tell them to look for dev mode when it has absolutely no benefit for them?

lua/nvim-treesitter/shell_command_selectors.lua
theHamsta: `--depth=1` was needed because we wanted to clone only the commit we need. This is much faster than doing a full-clone and then a checkout of the commit we want. But `--branch` only works with branches and not with individual commits that are not tree.

If the longer download times are a problem for you, you could use the git method only for parsers with symlinks (they could be marked as such)

lua/nvim-treesitter/shell_command_selectors.lua
Thanood: I just thought because of your quote in the first comment:

> Git doesn't error, but I assume it caused typescript treesitter to misbehave.

lua/nvim-treesitter/shell_command_selectors.lua
Thanood: I don't care for longer download times. 😃 
Changed it right now, works for me. I'm waiting for CI and have a look.

lua/nvim-treesitter/shell_command_selectors.lua
theHamsta: > Git doesn't error,

so it does not make sense to check for the error

>     but I assume it caused typescript treesitter to misbehave.

That was just bullshit. One reason why the other PR was never merged was because he wanted to make symlinks work on Windows while breaking stuff for other users.

This is the symlink in the typescript repo:

https://github.com/tree-sitter/tree-sitter-typescript/blob/master/typescript/corpus/common

the file is absolutely not necessary to compile the parser but is just part of their test suite.

lua/nvim-treesitter/shell_command_selectors.lua
theHamsta: Yes, we want to make stuff work for users :slightly_smiling_face: . We want to get to a code that causes least problems on most machines. :tada: CI seems to work now!

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -139,7 +141,7 @@ function M.select_download_commands(repo, project_name, cache_folder, revision)
         opts = {
           args = {
             '-L', -- follow redirects
-            is_github and url.."/archive/"..revision..".tar.gz"
+            is_github_or_gitlab and url.."/archive/"..revision..".tar.gz"
```
theHamsta, line 144: Have to change that back, because  we have to use the first url for github and the second for gitlab

theHamsta: It crashes the Linux/MacCI
```
{
  cmd = "tar",
  err = "Error during tarball extraction.",
  info = "Extracting...",
  opts = {
    args = { "-xvf", "tree-sitter-jsonc.tar.gz", "-C", "tree-sitter-jsonc-tmp" },
    cwd = "/home/runner/.local/share/nvim"
  }
}
```

lua/nvim-treesitter/shell_command_selectors.lua
Thanood: I will remove the dev-mode message then.

lua/nvim-treesitter/shell_command_selectors.lua
Thanood: Just as an additional note, this is ta's response:
`tar: Error opening archive: Unrecognized archive format`

theHamsta: It's unrecognized because Gitlab will post html when the URL is invalid.

theHamsta (approved):

theHamsta:
Nice work! When all parsers install on CI this looks fine to me :tada:

PR #1197, opened 2021-04-17
HTML/JSX: Highlight attributes as attributes

lmlorca:
Make HTML and JSX attributes use `@attribute` instead of `@property`


kyazdani42 (approved):

theHamsta:
I'm ok with that change. You would need to change all previous usages of `@attribute` to `@annotation`. This would be more in line with the usages in the Atom queries. One would also have to update CONTRIBUTE.md and maybe the highlights...

theHamsta:
This needs to be changed probably also for svelte. Is css `(attribute_name)` an `@attribute`? :man_shrugging:

lmlorca:
@theHamsta Alright it seems simple enough. Just to be sure, stuff like Java annotations and C# attributes, would be `@annotation` correct? (In that case I wonder if python decorators shouldn't be also annotations).

About svelte, I'm not familiar but I will fire up a dummy project later to test. If that element follows a XML-like syntax (`attribute_name=attribute_value`), then it would be `@attribute` probably, otherwise, not sure. If it's a css syntax, I guess whatever is coherent with css. However, since you mention Atom queries, I wonder if there's a way to just follow those as a guideline for everything instead of guessing...?

theHamsta:
I guess Python attributes are fine as functions, since they are really always executed (also they must be defined as functions).

Java/C++/C# attributes are more for introspection and by default are not executed (though they might be used to generated some code in the background).

If only XML/HTML attributes are `@attribute` then it is ok. Maybe we should change the highlight to whatever highlight html/xml attributes had before? Svelte highlight will soon inherit from HTML directly

lmlorca:
Then I guess `@attribute for e.g. Python decorators` from the contributing has to go as well...

Honestly `@attributes` isn't very used anyway, way less than `@property`. It didn't seem that the fact that the html/jsx attributes are currently marked as `@property` required changing that keyowrd in every other language. Much less `@attribute`. 

I don't even belive only the xmls have `@attributes`, the upstream queries uses them for example, for both Java annotations and HTML attributes.

And it's funny cause `tree-sitter/tree-sitter-c-sharp` doesn't even use `@attribute` even though in C# annotations are officially named attributes. It uses `@type` for its "attributes" instead, which makes sense as well in terms of what they are rather than what they do (usually C# attributes are color-coded as types too in IDE's and text editors, including Visual Studio).

Finally since the xmls are a very idiosyncratic family (an exception rather than a norm), what about some kind of `@tag.attribute` (if that's possible anyway). But honestly `@attributes` isn't very used anyway, way less than `@property` so changing the other uses just because of this and possibly breaking existing colorschemes, I don't know about that...

theHamsta:
> Then I guess @attribute for e.g. Python decorators from the contributing has to go as well...

In the past we decided for function, just to have consistent highlighting in the following case ( and have `attribute/annotation` (we already swapped them before and now only use `attribute`)). Annotation should only be code annotations and not any executable parts of the code
```
def foo_decorator():
     pass

@foo_decorator # have same highlighting here since function `foo_decorator` is just placed here
def bar():
      pass

```

antonk52:
@lmlorca are you planning to continue working on this? If not I am happy finish the intent of this PR 🙂

lmlorca:
@antonk52 my bad, totally forgot about this. @theHamsta I added @tag.attribute/TSTagAttribute, similar to @tag.delimiter/TSTagDelimiter

theHamsta:
Nice one! Can you add `@tag.attribute` to CONTRIBUTING.md and the vim documentation so that tests can pass??

theHamsta (approved):

lmlorca:
@theHamsta I hope its all ok now!

theHamsta:
Yes it is. I hoped that one of our JS-Ninjas would have a look on this.  But now I see that kyazdani42 already approved

theHamsta:
Tried to merge, but a rebase is needed!

lmlorca:
Alright this is such a simple change it's just simpler to redo the PR.

Issue #1187, opened 2021-04-14
Add vimscript/vimL support


reportaman:
**Language informations**

Add support for vimscript/vimL syntax coloring using nvim-treesitter.

kyazdani42:
@vigoux started a viml parser but apparently was very hard to do because vimL is quite badly designed, not sure if he's gonna finish it :)

vigoux:
There is indeed a VimL parser but it is lacking a lot of language features still, I'll try to resume work on it as soon as I can, and add support for VimL.

PR #1198, opened 2021-04-18
feat: add support for viml

vigoux:
As asked in #1187 this PR will be tracking the progress in supporting
VimL.

As the parser is not yet finished, this is really just a draft branch,
so that people can subscribe to it and have updates on the progress.

### DIFF
--- a/queries/vim/highlights.scm
+++ b/queries/vim/highlights.scm
@@ -1,0 +1,105 @@
+;; Keywords
+
+[
+  "if"
+  "else"
+  "elseif"
+  "endif"
+] @conditional
+
+[
+  "try"
+  "catch"
+  "finally"
+  "endtry"
+  "throw"
+] @exception
+
+[
+  "for"
+  "endfor"
+  "in"
+  "while"
+  "endwhile"
+] @repeat
+
+[
+  "function"
+  "endfunction"
+] @keyword.function
+
+;; Function related
+(function_declaration name: (_) @function)
+(call_expression function: (identifier) @function)
+
+[ (bang) (spread) ] @punctuation.special
+
+[ (no_option) (inv_option) (default_option) (option_name) ] @variable.builtin
+[ (scope) "a:" ] @namespace
+
+(ternary_expression ["?" ":"] @conditional)
+
+;; Commands and user defined commands
+
+[
+  "let"
+  "unlet"
+  "call"
+  "execute"
+  "normal"
+  "set"
+  "silent"
+  "echo"
+  "autocmd"
+  "return"
+  "lua"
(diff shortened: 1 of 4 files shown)


queries/vim/injections.scm
```diff
@@ -0,0 +1,6 @@
+(lua_statement . (_) @lua)
+(ruby_statement . (_) @ruby)
+(python_statement . (_) @python)
```
stsewd, line 3:
```suggestion
(python_statement . (_) @python)
(comment) @comment
```

queries/vim/highlights.scm
```diff
@@ -0,0 +1,91 @@
...
+] @repeat
+
+[
+  "function"
+  "endfunction"
+] @keyword.function
+
+[ (bang) (spread) ] @punctuation.special
+
+[ (no_option) (inv_option) (default_option) (option_name) ] @constant.builtin
```
theHamsta, line 33: are options constant when you can set them? `@variable.builtin`?

vigoux: Huh, yup, that seems better.

vigoux:
After quite some time, I think this is pretty much ready.

shadmansaleh:
@vigoux I think this is this a bug.
Screenshot

```
" .. a
```
is highlighted as comment . While this " is end of string

Also if I comment out the let statement . It recognises the string correctly but endf looses highlight.
Screenshot

and endf is highlighted again if I change a to ab 😕
Screenshot_20210812-190458_Termux

Can you reproduce this ?

vigoux:
@shadmansaleh can reproduce in the parser too, let me check what the problem is.

vigoux:
Found a fix, pushing in few minutes.

vigoux:
And pushed !  (sorry for the noise)

shadmansaleh:
> And pushed ! (sorry for the noise)

Works now confirmed. Thanks for the quick fix . I'll open issue in the parser instead of commenting here if I see something funny. I'm Sorry too :)

vigoux:
Don't worry about that, and keep those bug reports coming, I need it !

vigoux:
I am in the process of adding things, so please report any missing highlight

stsewd:
The windows CI is broken since this change, not sure why it fails after installing some parsers (there are like 3 PRs with the same problem)

jdhao:
@vigoux  Some highlights are missing strangely and `"` in mapping is treated as comments. See image below:

image

vigoux:
Hey, this is a known, bug that has been reported into `tree-sitter-viml` please report bugs there.

PR #1202, opened 2021-04-18
Add Markdown again

theHamsta:
I will test this for a while...

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -305,13 +305,12 @@ list.haskell = {
   }
 }
 
--- FIXME(vigoux): markdown is broken for now
--- list.markdown = {
---   install_info = {
---     url = "https://github.com/ikatyang/tree-sitter-markdown",
---     files = { "src/parser.c", "src/scanner.cc" },
---   }
--- }
+ list.markdown = {
+   install_info = {
+     url = "https://github.com/ikatyang/tree-sitter-markdown",
+     files = { "src/parser.c", "src/scanner.cc", "-DTREE_SITTER_MARKDOWN_AVOID_CRASH=1" },
+   }
+ }
 
 list.toml = {
   install_info = {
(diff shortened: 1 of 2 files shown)


theHamsta:
If someone has time, they can do some fuzzing of the new verison...

lua/nvim-treesitter/parsers.lua
```diff
@@ -295,13 +295,12 @@ list.haskell = {
...
--- list.markdown = {
---   install_info = {
---     url = "https://github.com/ikatyang/tree-sitter-markdown",
---     files = { "src/parser.c", "src/scanner.cc" },
---   }
--- }
+ list.markdown = {
+   install_info = {
+     url = "https://github.com/ikatyang/tree-sitter-markdown",
+     files = { "src/parser.c", "src/scanner.cc", "-DTREE_SITTER_MARKDOWN_AVOID_CRASH=1" },
```
theHamsta, line 301: wow, this works even for `cl.exe`

tami5:
Thanks a lot @theHamsta I'll test it out in an hour

theHamsta:
Best would be to fuzz the parser. I still don't trust it. But using it a week or so might be useful also.

tami5:
> Best would be to fuzz the parser. I still don't trust it. But using it a week or so might be useful also.

Okay how do I fuzz the parser :$ :D

theHamsta:
You can read an instruction in the original SEGFAULT issue. You need to find out how to compile the parser in the test ASAN script with exception and this AVOID_CRASH definition

theHamsta:
I had one situation where i might had a infinite loop when I had a markdown file open. This might also had a different cause...

aktau:
Especially since you have a repro case for such an incredibly long parse, it'd be good to file this as an issue to https://github.com/ikatyang/tree-sitter-markdown (cc @ikatyang).

theHamsta:
@aktau I can't reproduce with the current nvim. Maybe there was something wrong with a former nvim revision.

We still don't know whether this parser is reliable. Did you do some testing or fuzzing?

theHamsta:
Having the infinite loop issue with another README file. I guess the simplest solution would be to just take the external rules of the HTML scanner and start with a new grammar.js (maybe with indent/dedent from Python). At least it would be impossible for me to identify any problems with the current implementation.

dimbleby:
https://github.com/ikatyang/tree-sitter-markdown/issues/32

theHamsta:
At the moment we don't have any plans for this

PR #1204, opened 2021-04-19
Remove Python treesitter docstring injections

pappasam:
These result in some really annoying docstrings at present. Two problems:

1. Not all Python docstrings are in RST
2. The current implementation is inconsistent across docstrings, yielding inconsistent colors for the docstring headers.

See:

- <https://github.com/nvim-treesitter/nvim-treesitter/discussions/958>
- <https://github.com/Thanood/nvim-treesitter/commit/4039741dfa297b71e45c58e0aa9948a7f87d0a78>
- <https://github.com/nvim-treesitter/nvim-treesitter/issues/806>

## With injections

python-treesitter-bad

## Without injections

treesitter-python-good

We can re-add injections like this if/when the bugs have been worked out. There's currently no way to override the buggy implementation at present, so it's probably best to remove them until they can be disabled by users and/or they at least provide consistent highlighting.


pappasam:
Note: I'm using <https://github.com/pappasam/papercolor-theme-slim>, but inconsistent highlighting should theoretically be present for any colorscheme that's treesitter-compatible.

pappasam:
@stsewd Sounds good to me! That said, if for any reason it turns out to be harder to implement than we currently think, it might be a good idea to temporarily remove the injections / move them back to a development branch until we've figured things out? I'll trust your judgment here, just a suggestion.

pappasam:
And thanks for all your work here, I'm over the moon about how far treesitter support has evolved for Neovim!

theHamsta:
I don't see why there should be title at all. Doesn't RST require ========= or ---------------------- for titles?

stsewd:
> I don't see why there should be title at all. Doesn't RST require ========= or ---------------------- for titles?

yeah, I don't know how the title happens, I haven't been able to reproduce it outside the injections. I think something to do with `"""` being captured somehow.

theHamsta:
It's interesting that the case of ```"""text"""` is not a title highlight. It is parsed a `"text"` node. But `"text"` is only an alias. Could it be that there is an error within the aliased node so that you have multiple highlights mixed
image

How do you get the indentation with external scanner? In this case, the bold text is parsed as a block quote.
image

I would be ok for deactivating injections for now. Maybe @steelsojka can say more on what's happening here

steelsojka:
I'd have to look at the tree to pinpoint. A lot of times it has to do with offsetting the injection correctly. I don't code python so I'm not 100% sure what the expected result is in the doc string.

stsewd:
@pappasam I implemented a way to disable the docstrings injections or choose other format for them https://github.com/nvim-treesitter/nvim-treesitter/pull/1240

pappasam:
@stsewd amazing!!! I'll follow progress on that PR and close this once it's merged

rafikdraoui:
The only injections I find problematic are the docstring ones. I think it still makes sense to keep the regexp and comment injections.

One way to do this is to override the injections query with `vim.treesitter.set_query`:

```lua
-- nvim-treesitter/queries/python/injections.scm, with docstring
-- injections removed
local py_injections = [[
((call
  function: (attribute object: (identifier) @_re)
  arguments: (argument_list (string) @regex))
 (#eq? @_re "re")
 (#match? @regex "^r.*"))

(comment) @comment
]]
vim.treesitter.set_query('python', 'injections', py_injections)
```

However, I think it would be better to flip things around, where the `queries/python/injections.scm` only contains the regexp and comment injections, and people who really want the docstring rst injections can use `set_query` to append them.

I'm biased because I don't want the docstring injections, but still, I think it makes sense to put the burden of amending queries on the minority(?) of people who are willing to use the currently WIP/half-broken rst injections, rather than on everyone else who would rather have consistent highlighting for docstrings.

pappasam:
@rafikdraoui you are my hero, thanks!

theHamsta:
I also think the RST highlighting shouldn't be the default

yochem:
Would it be a good idea to inject [`@comment`](https://github.com/stsewd/tree-sitter-comment) in docstrings instead? to highlight keywords like TODO, WARNING, etc.?

clason:
Not really, no. The right approach is to capture them as `@string.documentation` and let users decide how to treat them.

amaanq:
No, keywords like TODO, WARNING, etc should not be in docstrings, docstrings are **documentation**, those are better suited for comments.

yochem:
Seems fair!

PR #1205, opened 2021-04-19
feat(query): directives for multi injections per match

steelsojka:
This feature adds directives for injecting at multiple different nodes within the same match. There's been some limitations on where the injection site can be within a single match. The `combined` capture combines across all matches of the same pattern. These allow for users to combine injection nodes for a single match. Here's a screen shot of an example where two JS template strings are injected with embedded javascript but they are parsed in the same context together, BUT only as part of that match. If `combined` was used it would parse with every other binary expression with the same form.

Screen Shot 2021-04-19 at 2 05 05 PM

The query for the above is
```scheme
((binary_expression
  left: (template_string) @_one
  right: (template_string) @_two)
 (#set! "language" "javascript")
 (#inject-at! @_one @_two)
 (#inject-offset-each! 0 1 0 -1))
```


steelsojka:
Still a work in progress. I will also probably incorporate fenced language injection into this as well, since it all kind of overlaps.

theHamsta:
Superseded by #2557

PR #1215, opened 2021-04-22
Some fixes with folds

vigoux:
Do actually two things :
- fix(c): fold switch statements
- fix(folds): always fold more than one line

### DIFF
--- a/lua/nvim-treesitter/fold.lua
+++ b/lua/nvim-treesitter/fold.lua
@@ -30,9 +30,11 @@ local folds_levels = tsutils.memoize_by_buf_tick(function(bufnr)
       stop = stop + 1
     end
 
+    local should_fold = start + 1 > stop -- Only fold for 2+ lines
+
     -- This can be folded
     -- Fold only multiline nodes that are not exactly the same as previously met folds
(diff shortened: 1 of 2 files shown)


stsewd (approved):

theHamsta:
Really nice! Was thinking about this yesterday and you just fixed it 👍

lua/nvim-treesitter/fold.lua
```diff
@@ -30,9 +30,11 @@ local folds_levels = tsutils.memoize_by_buf_tick(function(bufnr)
       stop = stop + 1
     end
 
+    local should_fold = start + 1 > stop -- Only fold for 2+ lines
```
Gelio, line 33: I came here from https://github.com/nvim-treesitter/nvim-treesitter/issues/1226 trying to see what's wrong. This should probably be `start + 1 < stop`, should it not? I'll try it out locally and see it that works

vigoux: Lol yeah, that's definitely it

Gelio: Yup, that's it :smile: I've raised https://github.com/nvim-treesitter/nvim-treesitter/pull/1234 with the fix

PR #1219, opened 2021-04-22
fix(lua): correctly highlight field names

### LINKED ISSUE #1180: Highlighting of Lua Fields Applies to Any Identifier Inside {}
**Language informations**

Any identifier that is a direct child of field gets highlighted as `Field`. I don't know whether this is intended or whether it should only apply to fields of type `{ field = value }` (so only to the lhs `field`).

It looks a bit strange when some of the identifiers get highlighted but not when you add a ` + 1` to it.
image

vigoux:
This correctly handles field names in lua.
The query is a bit bizarre then, but that's because the parser does
not handle fields so well...

Fixes #1180

### DIFF
--- a/queries/lua/highlights.scm
+++ b/queries/lua/highlights.scm
@@ -138,7 +138,8 @@
 (string) @string
 (number) @number
 (label_statement) @label
-(field (identifier) @field)
+; A bit of a tricky one, this will only match field names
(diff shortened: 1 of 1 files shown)


theHamsta (approved):
I suppose this is a in `a = b` when I read the query. :+1:

PR #1223, opened 2021-04-22
Allow failures for defined parsers

theHamsta:
@vigoux  I don't know how to handle this best. I guess @elianiva needs some more time with the Haskell parser and it was expected to fail for a long time. We could have this to actually check whether CI is not failing for parsers we would not expect that it would. We could also just wait until Haskell can be installed properly-

A PR adding queries for Haskell should find a proper solution.


vigoux:
Maybe we could just ignore the haskell parser for now, and add it back with the Haskell pr

vigoux:
Maybe your way will ease debugging locally for the parser, and just make CI more compliant.

vigoux (approved):

PR #1224, opened 2021-04-22
fix: use correct name in FUNDING.yml

vigoux:
Maybe there is an issue with our dash in the name, we'll see.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,1 @@
-open_collective: nvim-treesitter
+open_collective: "nvim-treesitter"


theHamsta (approved):

vigoux:
Why is it not showing up ?

stsewd:
@vigoux maybe you need to activate the sponsorship option in the settings of the repo. This is from a project I have
Screenshot_2021-04-22 pythonecuador pythonecuador github io

vigoux:
I'll push that one first, and then directly commit to master to have the final tweaks

PR #1230, opened 2021-04-23
codeql indents

pwntester:


kyazdani42 (approved):
don't know much about codeql so i trust that should work :)

PR #1244, opened 2021-04-24
Add GDScript filetype and use BufNewFile,BufRead

Shatur95:
I used `BufNewFile,BufRead` as suggested in `:h ftdetect`. I also noticed that we have `query.vim` that use `BufEnter,BufNewFile`. Probably we should change it to `BufNewFile,BufRead` too.

### DIFF
--- a/ftdetect/gdscript.vim
+++ b/ftdetect/gdscript.vim
@@ -1,0 +1,1 @@
+autocmd BufNewFile,BufRead *.gd set ft=gdscript
--- a/ftdetect/query.vim
+++ b/ftdetect/query.vim
@@ -8,4 +8,4 @@ function! s:shouldFt(path)
   endif
 endfunction
 
-autocmd BufEnter,BufNewFile *.scm call s:shouldFt(expand("%"))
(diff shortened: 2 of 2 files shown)


vigoux:
> I used `BufNewFile,BufRead` as suggested in `:h ftdetect`. I also noticed that we have `query.vim` that use `BufEnter,BufNewFile`. Probably we should change it to `BufNewFile,BufRead` too.

I agree on changing the query filetype too to the suggested values.

vigoux (approved):

Shatur95:
@vigoux, pushed!

ftdetect/gdscript.vim
```diff
@@ -0,0 +1 @@
+autocmd BufNewFile,BufRead *.gd set ft=gdscript
```
theHamsta, line 1: I'm wondering whether we should do this for the other filetypes that aren't supported by nvim too?

Shatur95: I think so.

theHamsta: fish, svelte, maybe julia?

PR #1259, opened 2021-04-27
Add go-mod tree-sitter config

camdencheek:
Adds a tree-sitter config for go.mod files as well as highlighting
queries.

A couple of notes:
- This only adds highlighting queries (not indentation, folds, etc.). I couldn't find any documentation on how those are interpreted. Happy to add those as well with some help. 
- This requires a `ftdetect` rule for go.mod files that sets `filetype=gomod`. It seems that graphql also doesn't set its filetype, so I wasn't sure if this is something that should be handled in this plugin. 

Example of it in action:
image

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -104,6 +104,16 @@ list.go = {
   maintainers = {"@theHamsta", "@WinWisely268"},
 }
 
+list.gomod = {
+  install_info = {
+    url = "https://github.com/camdencheek/tree-sitter-go-mod",
+	branch = "main",
+    files = { "src/parser.c" },
+  },
+  maintainers = {"@camdencheek"},
+  filetype = "gomod",
+}
+
 list.graphql = {
   install_info = {
     url = "https://github.com/bkegley/tree-sitter-graphql",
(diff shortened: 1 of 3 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -104,6 +104,16 @@ list.go = {
   maintainers = {"@theHamsta", "@WinWisely268"},
 }
 
+list.gomod = {
+  install_info = {
+    url = "https://github.com/camdencheek/tree-sitter-go-mod",
+	branch = "main",
```
theHamsta, line 110: indentation.

theHamsta (approved):
Very nice work! :tada:

vigoux (approved):
Bam, simple and efficient, in master it goes, thanks for this contribution!

camdencheek:
Awesome! RE: indents and folds, is there documentation somewhere about how that works? The files look pretty simple, but I'm not confident I fully understand the semantics.

theHamsta:
Folds just enumerates all foldable nodes. I don't fully understand indent. But there should be nodes that cause an indent for their whole length and `@branch` that dedents for this special nodes (e.g. a if/else node would cause an indent but for the line with else you would dedent for this one line.

camdencheek:
Ah, nice. Much simpler than I as expecting. Thanks!

PR #1262, opened 2021-04-28
chore(indents): use node:id() instead of node_fmt

theHamsta:
@kyazdani42 I think the code should be roughly equivalent except that `:id()` doesn't have the nil check. It appears that the outer code already checks for `nil`

### DIFF
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -16,17 +16,12 @@ local function get_node_at_line(root, lnum)
   end
 end
 
-local function node_fmt(node)
-  if not node then return nil end
-  return tostring(node)
-end
-
 local get_indents = tsutils.memoize_by_buf_tick(function(bufnr, root, lang)
   local get_map = function(capture)
     local matches = queries.get_capture_matches(bufnr, capture, 'indents', root, lang) or {}
     local map = {}
     for _, node in ipairs(matches) do
-      map[tostring(node)] = true
+      map[node:id()] = true
     end
     return map
   end
@@ -73,7 +68,7 @@ function M.get_indent(lnum)
       end
 
       -- nodes can be marked @return to prevent using them
-      if prev_node and not q.returns[node_fmt(prev_node)] then
+      if prev_node and not q.returns[prev_node:id()] then
         local row = prev_node:start()
         local end_row = prev_node:end_()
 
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/indent.lua
```diff
@@ -107,13 +102,13 @@ function M.get_indent(lnum)
 
   while node do
```
theHamsta, line 103: node ~=nil

lua/nvim-treesitter/indent.lua
```diff
@@ -93,12 +88,12 @@ function M.get_indent(lnum)
   if not node then
     local wrapper = root:descendant_for_range(lnum-1, 0, lnum-1, -1)
     node = wrapper:child(0) or wrapper
```
theHamsta, line 90: `wrapper ~= nil`

lua/nvim-treesitter/indent.lua
```diff
@@ -93,12 +88,12 @@ function M.get_indent(lnum)
   if not node then
     local wrapper = root:descendant_for_range(lnum-1, 0, lnum-1, -1)
```
theHamsta, line 89: root ~= nil

lua/nvim-treesitter/indent.lua
```diff
@@ -73,7 +68,7 @@ function M.get_indent(lnum)
       end
 
       -- nodes can be marked @return to prevent using them
-      if prev_node and not q.returns[node_fmt(prev_node)] then
+      if prev_node and not q.returns[prev_node:id()] then
```
theHamsta, line 71: prev_node ~= nil

lua/nvim-treesitter/indent.lua
```diff
@@ -16,17 +16,12 @@ local function get_node_at_line(root, lnum)
...
-  return tostring(node)
-end
-
 local get_indents = tsutils.memoize_by_buf_tick(function(bufnr, root, lang)
   local get_map = function(capture)
     local matches = queries.get_capture_matches(bufnr, capture, 'indents', root, lang) or {}
     local map = {}
     for _, node in ipairs(matches) do
-      map[tostring(node)] = true
+      map[node:id()] = true
```
theHamsta, line 24: node ~= nil

kyazdani42:
is this ready for review ? i see you've left some comments with `~= nil`

theHamsta:
@kyazdani42 I just wanted to document that it should still be correct and equivalent to the former code

kyazdani42:
i will run the tests and if it works like before I'll merge this then

kyazdani42:
running the tests shows diff of regression :/

theHamsta:
Then better leave it the old way

PR #1264, opened 2021-04-30
highlights(c/c++): "..." also exists in C (variadic macros)

theHamsta:


queries/c/highlights.scm
```diff
@@ -93,7 +93,7 @@
  (false)
 ] @boolean
 
-[ "." ";" ":" "," ] @punctuation.delimiter
+[ "." ";" ":" "," "..."] @punctuation.delimiter
```
theHamsta, line 96: hmm, it is not really a delimiter.

stsewd (approved):

queries/c/highlights.scm
```diff
@@ -95,6 +95,8 @@
 
 [ "." ";" ":" "," ] @punctuation.delimiter
 
+"..." @punctuation.special
+
```
vigoux, line 99: This is not really a punctuation, is it? Maybe using the same that what we used for Lua is better here?

theHamsta: it's points so it could be punctuation like in JS. Before it was `@operator` which represents a bit the aspect that it kind of works and combines parameters.

Lua uses `@constant` which is a bit weird because it not really holds a values and mixes highlight groups for punctuation and alphanumeric identifiers.

vigoux: But the problem is not the fact that it is punctuation, this actually has the meaning of a parameter, so considering this at the same level as say the `.` in java is simply wrong.

I would be interested in the opinion of other members of the team around that, but I am not really in favor of this capture for this.

And we'd need to allign everything on this.

For me `@parameter` sounds better.

theHamsta: We can't simply mix `?!.` with identifiers. In JS and C++ `...` actually modifies the parameter so you would have `foo...` where punctuation is pretty accurate (`@parameter` would merge it with the actual parameter name). The `...` will then also be used in the actual code and in other types of ellipsis and is not always a parameter. Base categories like identifiers, keywords and special characters like `?%+*/.(){}`

This is also the reason because there was a problem with `@operator` and `@keyword.operator` because it mixed the characters with builtin identifiers and keywords.

> at the same level as say the . in java is simply wrong.

it would be in the `special` :star:  category and not in the delimiter category. I think in almost all editor highlighting schemes `.` appear the same the same.

theHamsta: Btw. also Rust has this for calling variadic C functions: https://rust-lang.github.io/rfcs/2137-variadic.html#guide-level-explanation

vigoux (approved):
I changed my mind, I agree it is better

PR #1265, opened 2021-04-30
chore: show errors output during async installation

theHamsta:
Collect stdout/stderr of commands and only show them on command failure

Looks like this:
image

image

Screenshot_20210430_165907

I don't know why the users has to press `<cr>` an additional time after we show the error

Before:
image

Disclaimer: I have very little experience with `luv`

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -16,6 +16,8 @@ M.compilers = { vim.fn.getenv('CC'), "cc", "gcc", "clang", "cl" }
 local started_commands = 0
 local finished_commands = 0
 local failed_commands = 0
+local complete_std_output = {}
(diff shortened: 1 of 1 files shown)


kyazdani42:
is this ready ?

theHamsta:
From my side yes. Maybe one could change the curl command to omit progress.

kyazdani42 (approved):
code looks allright to me, i could not test it's erroring output though

theHamsta:
You can just insert an error in an url or install from a grammar.js with errors.

theHamsta:
@kyazdani42 is the double "press enter" a blocker for this PR?

kyazdani42:
@theHamsta i don't think so, go for it the errors are better than before :)

PR #1269, opened 2021-05-01
parsers: add tree-sitter-fish

theHamsta:
Addresses #1268

PR #1272, opened 2021-05-01
Add fish queries

### LINKED ISSUE #1268: Add support for fish
**Language informations**

The only remaining language for me that doesn't have a treesitter parser is fish. I found a treesitter project for fish here:
https://github.com/krnik/tree-sitter-fish

However, it just gives me errors and I am not knowledgeable enough about treesitter to be able to understand why. I have created a seperate issue on the respective repository.

ram02z:
Closes #1268.
One small issue I have which I have not been able to figure out is how to highlight all command options like `--description`. As you can see from the highlight snippet below, only the first command option is highlighted (`--argument-names`) and following option on the same line (`--description`) is ignored. The following syntax highlight rule is from line 93:

```
(command
  argument: [
             (word) @parameter (#vim-match? @parameter "^-")
            ]
)
```

> Highlight

image

> Indent

Works well for every statement listed in `indents.scm` apart from `if_statement`. When adding `else` or `else if` clauses the indenation doesn't work and the `end` keyword indents on enter. See gif below.
if_indent_bug

> Folds

fish_foldr

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -145,6 +145,7 @@ list.fish = {
     url = "https://github.com/krnik/tree-sitter-fish",
     files = { "src/parser.c", "src/scanner.c" },
   },
+  maintainers = {"@krnik", "@ram02z"},
 }
 
 list.php = {
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -140,6 +140,13 @@ list.bash = {
   maintainers = {"@TravonteD"},
 }
 
+list.fish = {
+  install_info = {
+    url = "https://github.com/krnik/tree-sitter-fish",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+}
+
 list.php = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-php",
--- a/queries/fish/highlights.scm
+++ b/queries/fish/highlights.scm
@@ -1,0 +1,155 @@
+;; Fish highlighting
+
+;; Operators
+
+[
+ "&&"
+ "||"
+ "|"
+ "&"
+ "="
+ "!="
+ ".."
+ "!"
+ (direction)
+ (stream_redirect)
+] @operator
+
+[
+ "not"
+ "and"
+ "or"
+] @keyword.operator
+
+;; Conditionals
+
+(if_statement
+[
+ "if"
+ "end"
+] @conditional)
+
+(switch_statement
+[
+ "switch"
+ "end"
+] @conditional)
+
+(case_clause
+[
+ "case"
+] @conditional)
+
+(else_clause 
+[
+ "else"
+] @conditional)
+
+(else_if_clause 
+[
+ "else"
+ "if"
+] @conditional)
+
+;; Loops/Blocks
+
+(while_statement
+[
+ "while"
+ "end"
+] @repeat)
+
+(for_statement
+[
+ "for"
+ "end"
+] @repeat)
+
+(begin_statement
+[
+ "begin"
+ "end"
+] @repeat)
+
+;; Keywords
+
+[
+ "in"
+ "return"
+ (break)
+ (continue)
+] @keyword
+
+;; Punctuation
+
+[
+ "["
+ "]"
+ "{"
+ "}"
+ "("
+ ")"
+] @punctuation.bracket
+
+"," @punctuation.delimiter
+
+;; Commands
+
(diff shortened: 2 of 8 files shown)


krnik:
> One small issue I have which I have not been able to figure out is how to highlight all command options like `--description`. As you can see from the highlight snippet below, only the first command option is highlighted (`--argument-names`) and following option on the same line (`--description`) is ignored. The following syntax highlight rule is from line 93:
> 
> ```
> (command
>   argument: [
>              (word) @parameter (#vim-match? @parameter "^-")
>             ]
> )
> ```

Fixed! I made a mistake when defining `command` rule.

ram02z:
> [Fixed](https://github.com/krnik/tree-sitter-fish/commit/a83dad3344d640e2543b9df45ea8132ddcc581b9)! I made a mistake when defining `command` rule.

@krnik It still doesn't work on the function block, but works on other commands. I realised that line 93 was not for the function block since `function_definition` isn't a command😅.
I have also updated the `function_definition` highlight in https://github.com/ram02z/nvim-treesitter/commit/e90354b3ab2695196893bc119e4956823445e683 but I think the grammer needs an update also. 

image

queries/fish/highlights.scm
```diff
@@ -0,0 +1,154 @@
...
+  option: [
+          (word)
+          (concatenation (word))
+          ] @parameter (#match? @parameter "^-")
+)
+
+;; Switch statements
+
+(switch_statement
+    value: (word) @parameter)
```
vigoux, line 128: Not sure the `parameter` is right here, as it is not exactly the case.
This should be something like `constant`.

queries/fish/highlights.scm
```diff
@@ -0,0 +1,154 @@
...
+
+(case_clause ["case"] @keyword.function)
```
vigoux, line 130: Here too, this is not a function related keyword.

vigoux (commented):
Awesome, thanks for handling that!!
If you want you can also add you as the fish maintainer in the repo!

I left a few comments btw.

krnik:
@ram02z Added fixes for other incorrect fields (function and for loop)

krnik:
Let's also add the query `(command name: (word) @function)` to highlight non-builtin command names.
Like in bash https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/bash/highlights.scm#L90

vigoux:
I think what you should do is rebase on #1269 just to get the parser set up, and that'll make the CI run on your addition.

queries/fish/highlights.scm
ram02z: I removed this. Referenced languages like golang and bash and none of them highlight the case value.

queries/fish/highlights.scm
ram02z: Moved `case_clause` to `@conditional`.

ram02z:
> 
> Let's also add the query `(command name: (word) @function)` to highlight non-builtin command names.
> Like in bash https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/bash/highlights.scm#L90

Added this along with the changes requested by @vigoux .

ram02z:
> 
> 
> I think what you should do is rebase on #1269 just to get the parser set up, and that'll make the CI run on your addition.

I rebased on #1269. Not sure why it added 2873d1efdaedf562af202d13594f1022def1fa49 along with the commits but I hope that is still fine.

lua/nvim-treesitter/parsers.lua
```diff
@@ -140,6 +140,13 @@ list.bash = {
   maintainers = {"@TravonteD"},
 }
 
+list.fish = {
+  install_info = {
+    url = "https://github.com/krnik/tree-sitter-fish",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+}
+
```
vigoux, line 149: If you want to maintain the queries for fish, you should add your name here in the `maintainers` key (see other entries in the parser list for examples).

vigoux (approved):

lua/nvim-treesitter/parsers.lua
ram02z: Will do! @krnik Should I add your name as well?

krnik: I'll be glad to help if something breaks so yes, please add me too.

ram02z:
Added the maintainer names for the fish parser list.

kidonng:
The README seems to need updating as well.

vigoux:
Hmmm yep it should be... Maybe in the next README update, I'll check that !

PR #1289, opened 2021-05-11
Add filetype detection for go.mod files

camdencheek:
A parser for go.mod files was added in #1259, but neovim does not set the correct filetype for go.mod files by default. This adds a ftdetect rule to set `ft=gomod`.


vigoux (approved):
Not sure what the best practices are, but I am in favor, let's get this merged.

PR #1290, opened 2021-05-11
highlights(latex): add more commands

theHamsta:
I was wondering whether we should add the commands from tree-sitter-latex/commands.json to the latex highlights. Since they are anonymous nodes they can't be easily matches with fields or with the `.` operator. Not sure whether this is a good idea.


cionx:
Regarding dca982244954aba823671cc9e8744ebfca7db175:

Don’t the last lines of `queries/latex/highlights.scm` also need to be adjusted for `alignat`, `gather` and `flalign`?

```scm
(begin
 name: (_) @text.environment.name
 (#not-match? @text.environment.name "^(displaymath|equation|multline|eqnarray|align|array|split)[*]?$"))
(end
 name: (_) @text.environment.name
 (#not-match? @text.environment.name "^(displaymath|equation|multline|eqnarray|align|array|split)[*]?$"))
```

theHamsta:
Any-of PR now merged. So this PR needs to be updated

queries/latex/highlights.scm
```diff
@@ -42,9 +39,223 @@
...
+  "\\Glsentrylong"
+  "\\glsentrylongpl"
+  "\\Glsentrylongpl"
+  "\\glsentryshort"
+  "\\Glsentryshort"
+  "\\glsentryshortpl"
+  "\\Glsentryshortpl"
+  "\\glsentryfullpl"
+  "\\Glsentryfullpl"
+]
```
vigoux, line 258: Are you missing a capture here ?

vigoux (approved):

PR #1294, opened 2021-05-14
Elixir support

nifoc:
This PR adds support for Elixir, thanks to [this awesome parser](https://github.com/ananthakumaran/tree-sitter-elixir).

I have been dogfooding this for quite some time (via the custom parser configuration) and didn't run into any issues.

A couple of important points:

* Makes use of the new `#any-of?` predicate
* I'm not sure if I have actually added `elixir` to alle the right places (mainly just `README.md` and `parsers.lua`)
* Will the entry in `lockfile.json` be generated automatically?
* I've been installing the parser via `grammer.js`. I'm not sure if I have to add this to the configuration or if having `elixir` shipped in `nvim-treesitter` will automatically pre-compile stuff?
* Since I intend to maintain the queries, I have added myself as the maintainer 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,6 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [dart](https://github.com/UserNobody14/tree-sitter-dart) (maintained by @Akin909)
 - [x] [devicetree](https://github.com/joelspadin/tree-sitter-devicetree) (maintained by @jedrzejboczar)
 - [x] [dockerfile](https://github.com/camdencheek/tree-sitter-dockerfile) (maintained by @camdencheek)
-- [x] [elixir](https://github.com/ananthakumaran/tree-sitter-elixir) (maintained by @nifoc)
 - [ ] [elm](https://github.com/elm-tooling/tree-sitter-elm)
 - [x] [erlang](https://github.com/AbstractMachinesLab/tree-sitter-erlang) (maintained by @ostera)
 - [x] [fennel](https://github.com/travonted/tree-sitter-fennel) (maintained by @TravonteD)
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [dart](https://github.com/UserNobody14/tree-sitter-dart) (maintained by @Akin909)
 - [x] [devicetree](https://github.com/joelspadin/tree-sitter-devicetree) (maintained by @jedrzejboczar)
 - [x] [dockerfile](https://github.com/camdencheek/tree-sitter-dockerfile) (maintained by @camdencheek)
+- [x] [elixir](https://github.com/ananthakumaran/tree-sitter-elixir) (maintained by @nifoc)
 - [ ] [elm](https://github.com/elm-tooling/tree-sitter-elm)
 - [x] [erlang](https://github.com/AbstractMachinesLab/tree-sitter-erlang) (maintained by @ostera)
 - [x] [fennel](https://github.com/travonted/tree-sitter-fennel) (maintained by @TravonteD)
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -233,6 +233,7 @@ list.elixir = {
   install_info = {
     url = "https://github.com/ananthakumaran/tree-sitter-elixir",
     files =  { "src/parser.c", "src/scanner.cc" },
+    requires_generate_from_grammar  = true,
   },
   maintainers = { '@nifoc' },
 }
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -229,6 +229,14 @@ list.erlang = {
   maintainers = { '@ostera' },
 }
 
+list.elixir = {
+  install_info = {
+    url = "https://github.com/ananthakumaran/tree-sitter-elixir",
+    files =  { "src/parser.c", "src/scanner.cc" },
+  },
+  maintainers = { '@nifoc' },
+}
+
 list.ocaml = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-ocaml",
--- a/queries/elixir/highlights.scm
+++ b/queries/elixir/highlights.scm
@@ -1,0 +1,190 @@
+(identifier) @variable
+
+(interpolation
+ "#{" @punctuation.special
+ "}" @punctuation.special) @none
+
+[
+  "after"
+  "do"
+  "end"
+] @keyword
+
+[
+  "and"
+  "in"
+  "not"
+  "not in"
+  "or"
+] @keyword.operator
+
+(call (function_identifier) @method)
+
+(call (function_identifier) @keyword
+ (#any-of? @keyword
+  "quote"
+  "receive"
+  "self"
+  "super"
+  "unquote"
+  "unquote_splicing"
+  "with"
+  "assert"
+  "assert_in_delta"
+  "assert_raise"
+  "assert_receive"
+  "assert_received"
+  "catch_error"
+  "catch_exit"
+  "catch_throw"
+  "flunk"
+  "refute"
+  "refute_in_delta"
+  "refute_receive"
+  "refute_received"))
+
+"fn" @keyword.function
+
+(call (function_identifier) @keyword.function
+ (#lua-match? @keyword.function "^def%a*$"))
+
+(call (function_identifier) @keyword.function
+ (#any-of? @keyword.function "describe" "doctest" "on_exit" "setup" "setup_all" "test"))
+
+"else" @conditional
+
+(call (function_identifier) @conditional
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -145,6 +145,7 @@ We are looking for maintainers to add more parsers and to write query files for
 - [x] [dart](https://github.com/UserNobody14/tree-sitter-dart) (maintained by @Akin909)
 - [x] [devicetree](https://github.com/joelspadin/tree-sitter-devicetree) (maintained by @jedrzejboczar)
 - [x] [dockerfile](https://github.com/camdencheek/tree-sitter-dockerfile) (maintained by @camdencheek)
+- [x] [elixir](https://github.com/ananthakumaran/tree-sitter-elixir) (maintained by @nifoc)
```
theHamsta, line 148: Readme and lock file are autogenerated

lua/nvim-treesitter/parsers.lua
```diff
@@ -229,6 +229,14 @@ list.erlang = {
   maintainers = { '@ostera' },
 }
 
+list.elixir = {
+  install_info = {
+    url = "https://github.com/ananthakumaran/tree-sitter-elixir",
+    files =  { "src/parser.c", "src/scanner.cc" },
+  },
+  maintainers = { '@nifoc' },
+}
```
theHamsta, line 238: You have to add a requires _generate_from_source. If it's not in the repo it needs to be generated. You also have to check whether npm is required (e.g. for depdencies or whether tree-sitter-cli is sufficient.

queries/elixir/highlights.scm
```diff
@@ -0,0 +1,190 @@
...
+           [(identifier) @parameter
+            (binary_op
+             left: (identifier) @parameter)]))
+         (identifier) @function]
+        operator: "when")
+       (binary_op
+        left: (identifier)
+        operator: _ @function
+        right: (identifier))]
+      (#lua-match? @keyword.function "^def%a*$"))
```
theHamsta, line 182: A very long expression

theHamsta (changes requested):
Looks very nice. Probably  the config needs to be adapted so that the installation works automatically

queries/elixir/locals.scm
```diff
@@ -0,0 +1,95 @@
...
+     (keyword_list
+      (identifier) @definition.parameter))
+    (struct
+     (keyword_list
+      (identifier) @definition.parameter))
+    (binary_op
+     left: (identifier) @definition.parameter
+     operator: "\\\\")]
+   (#lua-match? @definition.parameter "^[^_]")))
+ (#lua-match? @_call-name "^defp?$"))
```
theHamsta, line 42: Do you really need all this context

queries/elixir/highlights.scm
nifoc: Would it be better to split this up into multiple expressions (i.e. one expression per "case")?

Since you can do pattern matching in function heads in Elixir, function definitions can get quite complicated …

theHamsta (approved):

queries/elixir/highlights.scm
theHamsta: You are the maintainer. When you think this is the best way to work with it I support you :+1:

theHamsta (approved):

theHamsta:
We had the discussion whether we should add filetype detect commands for languages not supported natively by nvim. What's your opinion on adding something like this?
https://github.com/theHamsta/nvim-treesitter/blob/efbb1c66d27eb5b4bfbcc1f59d3384e0641c8214/ftdetect/gdscript.vim#L1

Should this be in plugins/personal configs or nvim-treesitter?

nifoc:
Just a quick thought: Maybe we should hold off merging this for a few more days, because the `any-of?` predicate is so new? (I saw you mentioned this in another PR)

Give people a bit more to install a newer nightly.

theHamsta:
I also thought this #1295

nifoc:
> We had the discussion whether we should add filetype detect commands for languages not supported natively by nvim. What's your opinion on adding something like this?
> https://github.com/theHamsta/nvim-treesitter/blob/efbb1c66d27eb5b4bfbcc1f59d3384e0641c8214/ftdetect/gdscript.vim#L1
> 
> Should this be in plugins/personal configs or nvim-treesitter?

Oh, right. I didn't think about that. When I uninstalled `vim-elixir` I added a similar line to my personal config.
But I can see why adding adding ftdetect for unsupported languages to nvim-treesitter might be beneficial.

Right now I don't have a strong opinion either way … But I will definitely give some thought to this!

queries/elixir/highlights.scm
vigoux: I am not sure it is better to split it in mutiple parts?

nifoc: I was playing around with this today and I feel like multiple parts are way more readable 👍 

I will update the PR accordingly.

nifoc:
I'm sorry for the large amount of changes in my latest push. There was an update to the parser that simplified some things (regarding unused and _special_ variables) and I took the time to improve other queries a bit more.

I also added comments above some of the more complex queries, hopefully making it easier for other people to contribute.

theHamsta:
At the moment your queries will not be tested. If you want you can install the tree-sitter CLI on CI to install also this parser.

scripts/ci-install-ubuntu-latest.sh
```diff
@@ -1,5 +1,8 @@
 sudo apt-get update
 sudo add-apt-repository universe
+wget -O - https://github.com/tree-sitter/tree-sitter/releases/download/v0.19.4/tree-sitter-linux-x64.gz | gunzip -c > tree-sitter
+sudo cp ./tree-sitter /usr/bin/tree-sitter
+sudo chmod uog+rwx /usr/bin/tree-sitter
```
theHamsta, line 5: :+1: 

Very strange though...
` Warning: parser for elixir is not installed`

scripts/ci-install-macos-latest.sh
```diff
@@ -1,6 +1,7 @@
 brew update
 brew install llvm --HEAD
 brew install neovim --HEAD
+brew install tree-sitter
```
theHamsta, line 4: This seems to be included already in `brew install neovim --HEAD`

nifoc: Have removed it locally

theHamsta: https://github.com/theHamsta/nvim-treesitter/blob/62bda244c9bf7e88f31d119f72bceae0e7116443/lua/nvim-treesitter/install.lua#L157

It does not work due to this line. Sorry for putting it there in the first place. I guess the logic should be if `not requires_generate_from_source or (not vim.fn.has('win32') and not generate_requires_npm)`

scripts/ci-install-ubuntu-latest.sh
nifoc: It looks like the script thinks `npm` is required:

> Skipping language elixir on CI (requires npm)!

I haven't looked into why that is yet. I will take a look tomorrow.

scripts/ci-install-macos-latest.sh
theHamsta: @nifoc maybe I was wrong...   maybe just `requires_generate_from_grammar` -> `generate_requires_npm` as the message says.

theHamsta: I hate CIs... :sweat_smile:

nifoc: hmmm … If it change the line to `repo.generate_requires_npm` I think it would not install `typescript` (+ others)?
I only checked for `typescript` and it's currently installed in CI.

theHamsta: You're right... Can this check be removed completely?

scripts/ci-install-ubuntu-latest.sh
theHamsta: That's the error message right after this check. But it does not fit very well to the check.

scripts/ci-install-macos-latest.sh
theHamsta: @nifoc this seems to have worked. All parsers are now either generatable without npm or installable directly from C files.

theHamsta:
Nice work!

dkarter:
Thanks for adding Elixir support @nifoc !

For anyone coming here because your Elixir syntax highlighting is suddenly broken with the "any-of" message, try upgrading Neovim 0.5 to the latest head version.

https://github.com/nvim-treesitter/nvim-treesitter/issues/1301

duduvz14:
I noticed considerable lag when opening elixir files, and narrowed it down to when the highlight module is enabled, this is not happening with other files being parsed by treesitter, do you think this is something in my particular config? How could I help? Thanks for adding elixir support by the way!

theHamsta:
@duduvz14 We observed some lags in some languages since 0.19.0 for parsing the queries in some languages https://github.com/tree-sitter/tree-sitter/issues/973 . Maybe you can try to reproduce that without Neovim. Would be great to pinpoint where the problem is.

PR #1299, opened 2021-05-16
parsers: Fortran/Supercollider no longer requires_generate_from_grammar

theHamsta:
Supercollider no longer requires generate from source

### DIFF
--- a/queries/fortran/highlights.scm
+++ b/queries/fortran/highlights.scm
@@ -34,7 +34,7 @@
 [
 "implicit"
 (none)
-] @annotation
+] @attribute
 
 [
   "function"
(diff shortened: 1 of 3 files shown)


theHamsta:
Supercollider does not seem to compile when not generated from source atm.

queries/fortran/highlights.scm
```diff
@@ -34,7 +34,7 @@
 [
 "implicit"
 (none)
-] @annotation
+] @attribute
```
theHamsta, line 37: there is still the PR to do `@attribute` -> `@annotation` for all queries that are not html attributes.

vigoux (approved):

PR #1306, opened 2021-05-17
Update fish parser

vigoux:
Updates the lock file and update fish highlight query accordingly.


theHamsta (approved):

theHamsta:
Thanks for keeping an eye on the parsers :+1:

PR #1311, opened 2021-05-17
Fish highlights move `test` to `test_command`

ram02z:
https://github.com/krnik/tree-sitter-fish/commit/1c25205cd7a68434bd1e576972193b3994404307 moved `test` under `test_command`, so the `command` highlighting no longer applies.


theHamsta (approved):

theHamsta:
Thank you for the contribution!

PR #1312, opened 2021-05-17
feat(dart): add async/sync* to highlights

akinsho:
Relates to https://github.com/UserNobody14/tree-sitter-dart/issues/18
Not sure what the outcome of that issue will be but this attempts to update/fix some highlights in the interim.

### Todo
- [x] fix async/sync*
- [x] Update type regex to only match Uppercase classes not CONSTANTS

### DIFF
--- a/queries/dart/highlights.scm
+++ b/queries/dart/highlights.scm
@@ -184,6 +184,9 @@
     "abstract"
     "as"
     "async"
+    "async*"
+    "yield"
+    "sync*"
     "await"
     "covariant"
     "deferred"
@@ -205,7 +208,7 @@
 
 ; when used as an identifier:
 ((identifier) @variable.builtin
- (#vim-match? @variable.builtin "^(abstract|as|covariant|deferred|dynamic|export|external|factory|Function|get|implements|import|interface|library|operator|mixin|part|set|static|typedef)$"))
+ (#match? @variable.builtin "^(abstract|as|covariant|deferred|dynamic|export|external|factory|Function|get|implements|import|interface|library|operator|mixin|part|set|static|typedef)$"))
 
 ["if" "else" "switch" "default"] @conditional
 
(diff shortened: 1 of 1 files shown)


queries/dart/highlights.scm
```diff
@@ -205,7 +208,7 @@
 
 ; when used as an identifier:
 ((identifier) @variable.builtin
- (#vim-match? @variable.builtin "^(abstract|as|covariant|deferred|dynamic|export|external|factory|Function|get|implements|import|interface|library|operator|mixin|part|set|static|typedef)$"))
+ (#match? @variable.builtin "^(abstract|as|covariant|deferred|dynamic|export|external|factory|Function|get|implements|import|interface|library|operator|mixin|part|set|static|typedef)$"))
```
theHamsta, line 211: You could switch this to use `any-of?`. `#any-of? @variable.builtin "abstract" "as" ...`

vigoux (approved):

queries/dart/highlights.scm
```diff
@@ -105,13 +105,11 @@
 (final_builtin) @constant.builtin
 
 ((identifier) @type
- (#match? @type "^_?[A-Z]"))
+ (#match? @type "^_?[A-Z][a-z]"))
```
theHamsta, line 108: I used `[A-Z].*[a-z]` in one of the other languages to also catch stuff like `IBlock` and `VBuffer`

akinsho: Thanks for the pointer just tweaked it

theHamsta:
Thank you!

queries/dart/highlights.scm
```diff
@@ -90,7 +87,6 @@
   name: (identifier) @type)
 (enum_constant
   name: (identifier) @type)
-(type_identifier) @type
```
theHamsta, line 93: Why is `type_identifier` not needed anymore?

akinsho: It was a duplicate, the exact same thing is present at line 97. There was a bit of messiness generally so some of the other changes are tidying up.

akinsho:
Sorry for that last force push @theHamsta there were some typos in the original commit so just wanted to fix those up

PR #1323, opened 2021-05-19
feat(dart): add late & required keywords

akinsho:
Since a new version of dart was released a few keywords were added. This PR adds the two main ones I'm aware of. There might be more but I'll add those as I discover them in the future

### DIFF
--- a/queries/dart/highlights.scm
+++ b/queries/dart/highlights.scm
@@ -163,6 +163,8 @@
     ; of the assertion not just the keyword
     ; assert
     (case_builtin)
+    "late"
+    "required"
     "extension"
     "on"
     "class"


theHamsta:
But aren't the keywords dependent on what is in grammar.js or are you adding all the keywords to tree-sitter-dart right now?

theHamsta (approved):

theHamsta (approved):

akinsho:
@theHamsta I could be wrong but I think treesitter dart has supported this for a while since this release has been planned for months I probably could have added it all the way back then but I didn't know much about the new dart version at the time so didn't

theHamsta:
You can use `lua D(vim.treesitter.inspect_language('dart').symbols)` to list all symbols from the dart parser (probably filter by false to get anonymous nodes)

PR #1353, opened 2021-06-01
Add injections for Elixir

dbernheisel:
1. Add markdown placeholder injection for when Markdown is re-added. This is enabled in the screenshot
2. Add regex injection

image

### DIFF
--- a/queries/elixir/injections.scm
+++ b/queries/elixir/injections.scm
@@ -1,1 +1,10 @@
 (comment) @comment
+
+; TODO: re-add when markdown is added
+; (heredoc
+;   (heredoc_content) @markdown)
+
+((sigil
+   (sigil_start) @_start
+   (sigil_content) @regex)
+ (#match? @_start "~(r|R)[/</\\\"[({|]"))


vigoux (approved):
Hey, thank you very much for this first contribution and welcome !

Everything looks great, I am okay to merge !

nifoc:
I think you also have to include `(escape_sequence)` in your `@regex` match, because the parser generates such a node for `\.`.

I added the following query to `~/.config/nvim/after/queries/elixir/injections.scm` to verify:

```
((sigil
  (sigil_start) @_sigil-type
  [(sigil_content) (escape_sequence)] @regex)
 (#lua-match? @_sigil-type "^~[rR].$"))
```

(I don't know if using `lua-match` instead of `match` actually makes a performance difference in this case … probably not)

dbernheisel:
@nifoc Your regex will exclude heredoc delimiters and include more delimiters than allowed in the language; but I do like it if we can adjust to include heredocs. I didn't see `lua-match?` in the tree-sitter query docs. Is it a nvim-treesitter only thing? the elixir highlights.scm seems to be the only place its used.

I didn't notice earlier the inclusion of escape_sequence in there and [it looks like it can also include interpolation too](https://github.com/ananthakumaran/tree-sitter-elixir/blob/74024384dd09a34908d534de320d0bc30f26ae8f/grammar.js#L598-L603). I don't see a way to combine the sibling nodes before sending to regex. I tried `@combined` but that seems to combine it with the base document entirely, not for each sigil. I would also want the combination to exclude the interpolated piece, so maybe there's not a perfect way to get it done?

image

vigoux:
>  I didn't see `lua-match?` in the tree-sitter query docs. Is it a nvim-treesitter only thing? the elixir highlights.scm seems to be the only place its used.

This is an `nvim` specific thing, it is listed in `:h lua-treesitter-predicates`.

For the injection part, it looks like it will be a bit difficult, the grammar produces a very weird tree, having that wrapped into a node would be better.
I'll try to find a solution.

nifoc:
> Your regex will exclude heredoc delimiters and include more delimiters than allowed in the language

Oh, right. I totally forgot about heredoc delimiters. I think we should be able to get away with `"^~[rR]"`, because the parser should already take care of the delimiters when creating the `sigil_start` node: https://github.com/ananthakumaran/tree-sitter-elixir/blob/cb7c3cbfe5d5747bc305be02d4d4f6d9a8fd57db/src/scanner.cc#L1097

dbernheisel:
@nifoc cool, I love the simpler regex. I'll get a PR up if you don't beat me to it.

nifoc:
Feel free to go ahead and submit a PR :)

PR #1351, opened 2021-05-31
Improve c sharp

theHamsta:
Address #1349

PR #1356, opened 2021-06-03
More C# improvements

Luxed:
Improved highlights and folds. Addresses #1349.
Here's a before and after for the highlights:

<details>
<summary>before:</summary>

image

</details>

<details>
<summary>after:</summary>

image

</details>

<details>
<summary>folds in action:</summary>

image

</details>

Some notable changes:

- `var` is now highlighted as a keyword. This felt like a more logical highlight than type.
- The first identifier in a series of identifiers (like `Console.WriteLine()`) is no longer highlighted as a type. This change was reverted (from #1351). The reasoning behind this is that this identifier isn't necessarily a type, it could also be a namespace. Since we don't really have any idea of which it could be, I think it's better to leave it blank for now.
- Field declarations are no longer restricted to starting with `_` or `m_`.

Suggestions are welcome!

### DIFF
--- a/queries/c_sharp/highlights.scm
+++ b/queries/c_sharp/highlights.scm
@@ -9,16 +9,16 @@
 (interpolation) @none
 
 (invocation_expression
-     (member_access_expression
-            name: (identifier) @method))
+  (member_access_expression
+    name: (identifier) @method))
+
+(invocation_expression
+  function: (conditional_access_expression
+    (member_binding_expression
+      name: (identifier) @method)))
 
 (namespace_declaration
   name: [(qualified_name) (identifier)] @namespace)
-
-((member_access_expression
-  (identifier) @type
-  (_) .)
-  (#match? @type "^[A-Z].*[a-z]"))
 
 (qualified_name
   (identifier) @type)
@@ -26,11 +26,14 @@
 (invocation_expression
       (identifier) @method)
 
-((identifier) @field
- (#match? @field "^_"))
+(field_declaration
+  (variable_declaration
+    (variable_declarator
+      (identifier) @field)))
 
-((identifier) @field
- (#match? @field "^m_"))
+(initializer_expression
+  (assignment_expression
+    left: (identifier) @field))
 
 (parameter_list
   (parameter
@@ -56,9 +59,10 @@
 
 [
  (predefined_type)
- (implicit_type)
  (void_keyword)
 ] @type.builtin
+
+(implicit_type) @keyword
 
 (comment) @comment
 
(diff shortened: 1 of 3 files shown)


vigoux (approved):

queries/c_sharp/highlights.scm
```diff
@@ -9,28 +9,31 @@
...
-  (#match? @type "^[A-Z].*[a-z]"))
-
 (qualified_name
   (identifier) @type)
 
 (invocation_expression
       (identifier) @method)
 
-((identifier) @field
- (#match? @field "^_"))
```
theHamsta, line 30: @svermeulen I generally like the C++ member conventions. Not sure how popular they are in C#

svermeulen:
@theHamsta or @Luxed would either of you be willing to replace me as the designated c# maintainer on the readme?  Been awhile since I've been in there and never really understood it well to begin with

queries/c_sharp/highlights.scm
theHamsta: All fields that are used within a method without `this.*` can be found in the same file. So we could remove the check since we will be able to find all those with an efficient `is?` in future

Luxed: > All fields that are used within a method without `this.*` can be found in the same file.

Unfortunately, this is not the case. They could be found in another file as part of a `partial class`, `abstract class` or even an "extended" class (overridden might be a better word?).

I guess I could revert this change since an identifier that starts with a `_` should always be a field. It is a convention, but is not enforced at all. I personally encounter both daily.

Luxed:
> @theHamsta or @Luxed would either of you be willing to replace me as the designated c# maintainer on the readme? Been awhile since I've been in there and never really understood it well to begin with

I would be happy to take your place if you are not willing to work on it anymore :stuck_out_tongue:.
Can I just put my name in the README or is there something else to do?

theHamsta:
@Luxed you have to do it in `parsers.lua` README is updated automatically.

queries/c_sharp/highlights.scm
theHamsta: In C++ I made it fused it to a single regex. Decide yourself whether you think this rule should be kept or not.

```
((identifier) @field
 (#match? @field "(^_|^m_|_$)"))
```

svermeulen:
@Luxed Yep that'd be great.  Thanks!

Luxed:
I am done with this Pull Request.
I will keep in mind the conversation about the field declaration, but for now, I think this is good enough for every day usage.
Locals will be my next task. I'll make a PR as soon as I have understand how they work and have a simple, working, query.

theHamsta (approved):
Nice work and thanks again!

PR #1357, opened 2021-06-03
jsx query - Components as constructor instead of type

napmn:
PR links to [this issue](https://github.com/nvim-treesitter/nvim-treesitter/issues/1028).
JSX queries for highlighting are updated so that components with capital first letter are highlighted as constructor instead of type which should be the correct behavior.

### Before
Screenshot 2021-06-03 at 09 57 04

### After
Screenshot 2021-06-03 at 09 58 47

### DIFF
--- a/queries/jsx/highlights.scm
+++ b/queries/jsx/highlights.scm
@@ -14,23 +14,23 @@
 (jsx_self_closing_element
   name: (identifier) @tag)
 
-(jsx_opening_element ((identifier) @type
- (#match? @type "^[A-Z]")))
+(jsx_opening_element ((identifier) @constructor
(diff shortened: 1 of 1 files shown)


steelsojka:
@kyazdani42 Didn't we have a discussion about this in the past? Wasn't `type` the agreed upon capture for components or am I thinking of something else?

kyazdani42:
I think we discussed this on the linked issue, constructor makes more sense than type right ?

steelsojka:
For sure. I'm good with constructor.

kyazdani42 (approved):
LGTM thanks :)

PR #1362, opened 2021-06-04
c++ highlights: don't capture all identifiers with uppercase initials as types

Cnly:
Current highlighting queries capture all identifiers with uppercase initials as types, which may work very poorly on some codebases, e.g. LLVM. This PR removes these heuristics and let (I guess) the inherited c highlighting queries do their work. Below are a comparison before and after the changes:

CleanShot 2021-06-04 at 20 01 05@2x
CleanShot 2021-06-04 at 20 01 15@2x

### DIFF
--- a/queries/cpp/highlights.scm
+++ b/queries/cpp/highlights.scm
@@ -32,10 +32,6 @@
     name: (identifier) @function))
 
 
-((identifier) @type
- (#match? @type "^[A-Z].*[a-z]")
- (#not-has-parent? @type function_declarator))
-
 (namespace_identifier) @namespace
 ((namespace_identifier) @type
                         (#match? @type "^[A-Z]"))


theHamsta:
Ideally I would like to have such decisions customizable. https://github.com/neovim/neovim/pull/14196 would be a solution for that because the queries could have arbitrary specialization. As an alternative would could also have predicate to activate the heuristics for all of nvim-treesitter.

I personally would like to keep those regexes since the C++ grammar only has very few places where only a type is admissible. But I can try whether it's possible to live without them in complicated code bases.

Cnly:
@theHamsta Currently this sounds more like a problem of correctness to me. Could you provide some examples where keeping these is beneficial?

theHamsta:
I was first also very skeptical when I saw those naming convention queries in the Atom queries but the grammar itself can't really tell you in most places what category an identifier belongs to and in this situations you have to make the decision whether you would guess based on naming conventions or just don't do the highlight.

I guess many of the situations where you can have a type are nowadays also handled by this query (things like Foo::Bar or `Foo::bar`). So other query might be not so important any more. I will try a while without it.
```
((namespace_identifier) @type
                        (#match? @type "^[A-Z]"))
```

Can you give me an example where you would use an uppercase identifier when it isn't a type? We thought that there are code bases that use upper case functions, this is why functions can overwrite this heuristics. The LLDB code base is "special" in many ways though...

Cnly:
Thanks for the explanation @theHamsta. I myself haven't seen other codebases that use uppercase variables either. And when I tried a while without the heuristics, I saw two cases where types are not highlighted correctly. The first is `throw MyError()` which we can fix by highlighting `throw_statement`. The other is constructors like `MyObject()` which I'm not yet sure how to fix or whether it's important.

theHamsta:
I tried without the heuristics and it does not change to much. Maybe we could just use the heuristics. E.g. template invokations
Screenshot_20210604_214234

It would be good in general to have fewer regexes and parent checks. Let's just sleep a night over it.

theHamsta:
From my side I can also move that query to my personal config. @Cnly what's your opinion on that change after some days of experimenting?

Cnly:
I agree that in the most ideal case we'll have a customisable option. But currently removing the heuristics greatly improved my experience in the LLVM codebase. I've also tried some other codebases and I think in those more common cases there haven't been very noticeable changes. Given that I think I'll choose for the removal of the heuristics.

theHamsta (approved):

PR #1390, opened 2021-06-13
Improve clojure highlighting

Akeboshiwind:
I've been using these for the last week or so and they seem pretty good.
Some tweaks may be needed but they're definitely a better start (imo).

## Screenshots:
(using [moonlight.nvim](https://github.com/shaunsingh/moonlight.nvim))

### `Neovim default` vs `Original tree-sittter` vs `Updated tree-sitter`
<div float="left">
    Screenshot 2021-06-13 at 23 14 52
    Screenshot 2021-06-13 at 23 14 34
    Screenshot 2021-06-13 at 23 14 34
</div>

I think the Neovim default might be coming from conjure, but I still think the tree-sitter one is cool.

### DIFF
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -111,7 +111,7 @@
 
 ; Comment
 ((sym_lit) @comment
- (#any-of? @comment "comment" "quote"))
+ (#any-of? @comment "comment"))
 
 ; Conditionals
 ((sym_lit) @conditional
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -329,7 +329,7 @@
 ;; namespaces
 (list_lit
  .
- (sym_lit) @a
- (#eq? @a "ns")
+ (sym_lit) @_include
+ (#eq? @_include "ns")
  .
  (sym_lit) @namespace)
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -53,6 +53,9 @@
 ((sym_lit) @parameter
  (#match? @parameter "^[&]"))
 
+((sym_lit) @variable.builtin
+ (#match? @variable.builtin "^[%]"))
+
 ;; TODO: General symbol highlighting
 ;((sym_lit) @symbol
 ; (#eq? @symbol @variable))
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -284,7 +284,8 @@
  .
  (sym_lit)
  .
- (str_lit)? @text
+ ;; TODO: Add @comment highlight
+ (str_lit)?
  .
  (_))
 
@@ -296,7 +297,8 @@
  .
  (sym_lit)? @function
  .
- (str_lit)? @text)
+ ;; TODO: Add @comment highlight
+ (str_lit)?)
 ;; TODO: Fix parameter highlighting
 ;;       I think there's a bug here in nvim-treesitter
 ;; TODO: Reproduce bug and file ticket
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -12,7 +12,7 @@
 
 ["'" "`"] @string.escape
 
-["~" "~@" "#" "^"] @punctuation.special
+["~" "~@" "#"] @punctuation.special
 
 ["{" "}" "[" "]" "(" ")"] @punctuation.bracket
 
@@ -236,3 +236,10 @@
   "vector" "vector-of" "vector?" "volatile!" "volatile?"
   "vreset!" "with-bindings*" "with-meta" "with-redefs-fn" "xml-seq"
   "zero?" "zipmap"))
+
+; Meta punctuation
+;; NOTE: When the above `Function definitions` query captures the
+;;       the @function it also captures the child meta_lit
+;;       We capture the meta_lit symbol (^) after so that the later
+;;       highlighting overrides the former
+"^" @punctuation.special
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -1,39 +1,18 @@
 ;; >> Litterals
 
 (dis_expr) @comment
-
 (kwd_lit) @symbol
-
 (str_lit) @string
-
 (num_lit) @number
-
 (char_lit) @character
-
 (bool_lit) @boolean
-
 (nil_lit) @constant.builtin
-
 (comment) @comment
-
 (regex_lit) @string.regex
 
-;; TODO: Quote whole quoted symbol?
-(quoting_lit
- marker: "'" @string.escape)
+["'" "`"] @string.escape
 
-(syn_quoting_lit
- marker: "`" @string.escape)
-(unquoting_lit
- marker: "~" @punctuation.special)
-(unquote_splicing_lit
- marker: "~@" @punctuation.special)
-
-(set_lit
- marker: "#" @punctuation.special)
-
-(anon_fn_lit
- marker: "#" @punctuation.special)
+["~" "~@" "#" "^"] @punctuation.special
 
 ["{" "}" "[" "]" "(" ")"] @punctuation.bracket
 
@@ -41,13 +20,7 @@
 
 ;; >> Symbols
 
-;; metadata
-;; TODO: Mark whole meta tag?
-;; TODO: If not, handle java classes?
-(meta_lit
- marker: "^" @punctuation.special)
-
-;; parameter-related
+; parameter-related
 ((sym_lit) @parameter
  (#match? @parameter "^[&]"))
 
@@ -57,7 +30,7 @@
 ;; TODO: General symbol highlighting
 ;; use @variable?
 ;((sym_lit) @symbol
-; (#eq? @symbol @variable))
+; (#not-eq? @symbol @variable))
 
 ;; dynamic variables
 ((sym_lit) @variable.builtin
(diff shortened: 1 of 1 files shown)


queries/clojure/highlights.scm
```diff
@@ -1,77 +1,333 @@
...
+  "unchecked-int" "unchecked-long" "unchecked-multiply"
+  "unchecked-multiply-int" "unchecked-negate" "unchecked-negate-int"
+  "unchecked-remainder-int" "unchecked-short" "unchecked-subtract"
+  "unchecked-subtract-int" "underive" "unquote"
+  "unquote-splicing" "unreduced" "unsigned-bit-shift-right"
+  "update" "update-in" "update-proxy" "uri?" "uuid?"
+  "val" "vals" "var-get" "var-set" "var?" "vary-meta" "vec"
+  "vector" "vector-of" "vector?" "volatile!" "volatile?"
+  "vreset!" "with-bindings*" "with-meta" "with-redefs-fn" "xml-seq"
+  "zero?" "zipmap"))
```
theHamsta, line 272: I created a plug-in for all the CL built-in functions but from my side we can also keep them here

queries/clojure/highlights.scm
```diff
@@ -1,77 +1,333 @@
...
+;; >> Context based highlighting
 
-;; other calls
+; def-likes
+; Correctly highlight docstrings
 (list_lit
  .
- (sym_lit) @function)
+ (sym_lit) @a ; Don't really want to highlight twice
+ (#match? @a "^def.*")
```
theHamsta, line 283: You have to start with underscore when you not intend to highlight the capture

queries/clojure/highlights.scm
```diff
@@ -1,77 +1,333 @@
...
+ .
+ (sym_lit)
+ .
+ (str_lit)? @text
```
theHamsta, line 287: Text?

queries/clojure/highlights.scm
```diff
@@ -1,77 +1,333 @@
...
+ ;   (sym_lit)* @parameter))])
+  
+;[((list_lit
+;   (vec_lit
+;    (sym_lit) @parameter)
+;   (_)
+;  +
+; ((vec_lit
+;   (sym_lit) @parameter)
+;  (_)))
```
theHamsta, line 317: Yes, only some of the parameters are captured. I started to parser the parameters for common lisp directly in the parser. I could be related to a bug filed at Tree-Sitter that limited the number of returned results in starred expressions

queries/clojure/highlights.scm
Akeboshiwind: Ah, TIL thanks

queries/clojure/highlights.scm
Akeboshiwind: It's a docstring.
I think it should be highlighted differently to normal strings but I wasn't sure how, maybe `@comment` would work better 🤔

theHamsta (approved):

queries/clojure/highlights.scm
Akeboshiwind: That sounds like the same issue I was having.
Only the first few parameters were being highlighted, even in very short parameter lists.

queries/clojure/highlights.scm
Akeboshiwind: I tried changing this to `@comment` but I start getting wierd bugs.
Like when I open the buffer the docstring is highlighted as a comment, but if I select it then it get's highlighted as a string when I unselect.
And `:TSHighlightCapturesUnderCursor` just shows it as a string, very strange 🤔.

I'm going to remove the capture and add a TODO comment above to re-add it.

raymond-w-ko:
I've been meaning to merge something similar but didn't have to time due to work. Glad to see my `#any-of?` predicate being used 😄  I have a general question.

In my branch here: https://github.com/raymond-w-ko/nvim-treesitter/tree/master/queries/clojure

Is it frowned upon to have loose utility files in the folder? I wrote a helper script in Clojure to do introspection of it's own variables and generate text files. The end goals is to be able to do something like `:read queries/clojure/clojure.core.functions.txt` in neovim after running the script to update the query file.

queries/clojure/highlights.scm
stsewd: Docstrings should make use of injections to highlight them as the markup they use, like markdown in rust or rst in python.

Akeboshiwind: Ah, that makes sense.

Unfortunately changing it to `@comment` seems to be a bit buggy atm.

Akeboshiwind:
It definitely would be nice to have a programatic way to generate these lists 😅, although I don't know about the standards for it.

If they're allowed then it would be great to add them to this PR.
Although if that does happen then it's worth noting that my version has lots of changes vs the one in the repo currently.
For example I split the operators like `+` into a separate group, but that could be easily handled.

queries/clojure/highlights.scm
stsewd: You need to include this in the `injections.scm` file, about the double hl, that's probably because of the order of the queries, the queries below others have more priority.

Akeboshiwind: I'm happy to add the docstrings to `injections.scm`, that's a great idea.

As for the double hl, I think it's more a bug about when the highlights get re-applied after buffer changes, here's a gif of the issue I'm having:


In the above I type roughly this: `rs:w<cr>:e<cr>`

It's on the `rs` that the docstring changes from an `@comment` to an `@string`.
And when I type `:e<cr>` it changes back.

You can also get some strange things when editing multi-line docstrings:

Typing `gg` breaks the docstring highlight:


Having the docstring page off screen breaks the highlight: 

But only when you are in insert mode inserting newlines, the `o` command works fine for some reason 🤷‍♂️

theHamsta:
@Akeboshiwind this is why I created a own plug-in  for generating the Common Lisp built-ins. You can specialize it to your language and nobody needs to understand how to run clojure files.

You could put it in scripts but it probably shouldn't edit the scm files. Maybe just a Textfile or a lua file thatis used by a custom capture. When it's generated it shouldn't be in a file that is edited by users as they will constantly manually try to do edits and it is unnecessary visual clutter programmers don't need to care for as it's automated already.

queries/clojure/highlights.scm
```diff
@@ -1,81 +1,335 @@
...
+((sym_lit) @keyword
+ (#match? @keyword "^def.*$"))
+((sym_lit) @keyword
+ (#eq? @keyword "declare"))
+((sym_lit) @keyword.function
+ (#match? @keyword.function "^(defn|defn-|fn|fn[*])$"))
+
+; Comment
+((sym_lit) @comment
+ (#any-of? @comment "comment" "quote"))
```
theHamsta, line 114: You're sure you want to have  `quote` a comment?

theHamsta:
@Akeboshiwind is this ready to merge?

Akeboshiwind:
@theHamsta Not quite sorry.
A few things came up IRL I'll try and polish it off quickly now

Akeboshiwind:
Hmm, I thought I had more changes than that 😅.

It's ready to merge now, I've been using it for a few weeks with no problems :)
(Other than the missing highlights labeled with TODOs)

stsewd:
>  (Other than the missing highlights labeled with TODOs)

You can add the new queries to
https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/clojure/injections.scm

Akeboshiwind:
Haha, looks like we're a little late.

@stsewd I tried adding this but I was getting issues:
```diff
diff --git a/queries/clojure/injections.scm b/queries/clojure/injections.scm
index 4bb7d67..b06fd66 100644
--- a/queries/clojure/injections.scm
+++ b/queries/clojure/injections.scm
@@ -1 +1,26 @@
 (comment) @comment
+
+; def-likes
+(list_lit
+ .
+ (sym_lit) @_keyword
+ (#match? @_keyword "^def*")
+ .
+ (sym_lit)
+ .
+ (str_lit)? @comment
+ .
+ (_))
+ 
+
+; Function definitions
+(list_lit
+ .
+ (sym_lit) @_keyword.function
+ (#any-of? @_keyword.function "fn" "fn*" "defn" "defn-")
+ .
+ (sym_lit)?
+ .
+ (str_lit)? @comment
+ .
+ (_))
```

If you open a file with a function that doesn't have a docstring then it fails:


The full error being:
```
treesitter/highlighter: Error executing lua: ...l/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:368: table index is nil
```

vigoux:
Damn, we need to fix mergify merging too early.

Akeboshiwind:
FYI in this case it's fine.

The injections.scm changes should probably be a separate PR seeing as there's issues with it.

stsewd:
> Haha, looks like we're a little late.
> 
> @stsewd I tried adding this but I was getting issues:
> 
> ```diff
> diff --git a/queries/clojure/injections.scm b/queries/clojure/injections.scm
> index 4bb7d67..b06fd66 100644
> --- a/queries/clojure/injections.scm
> +++ b/queries/clojure/injections.scm
> @@ -1 +1,26 @@
>  (comment) @comment
> +
> +; def-likes
> +(list_lit
> + .
> + (sym_lit) @_keyword
> + (#match? @_keyword "^def*")
> + .
> + (sym_lit)
> + .
> + (str_lit)? @comment
> + .
> + (_))
> + 
> +
> +; Function definitions
> +(list_lit
> + .
> + (sym_lit) @_keyword.function
> + (#any-of? @_keyword.function "fn" "fn*" "defn" "defn-")
> + .
> + (sym_lit)?
> + .
> + (str_lit)? @comment
> + .
> + (_))
> ```
> 
> If you open a file with a function that doesn't have a docstring then it fails:
> 
> 
> The full error being:
> 
> ```
> treesitter/highlighter: Error executing lua: ...l/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:368: table index is nil
> ```

This is because ` (str_lit)? @comment`, it should be ` (str_lit) @comment` (no `?`, as you only want to have the injection if the string exists)

PR #1393, opened 2021-06-14
Add Julia indent queries and update fold queries

savq:
The julia.vim plugin has more indentation rules for collection literals, making matrices easier to read. I don't know how that would work with tree-sitter queries so I just ignored it. This only adds one indentation level, like the indent queries for other languages. I think it's fine for now, but I'd be interested in improving it in the future.

### DIFF
--- a/queries/julia/folds.scm
+++ b/queries/julia/folds.scm
@@ -1,11 +1,17 @@
 [
- (module_definition)
- (struct_definition)
- (macro_definition)
- (function_definition)
- (compound_expression) ; begin blocks
- (let_statement)
- (if_statement)
- (for_statement)
- (while_statement)
+  (module_definition)
+  (struct_definition)
+  (macro_definition)
+  (function_definition)
+
+  (if_statement)
+  (try_statement)
+  (for_statement)
+  (while_statement)
+  (let_statement)
+  (quote_statement)
+
+  (do_clause)
+  (compound_expression)
 ] @fold
+
--- a/queries/julia/indents.scm
+++ b/queries/julia/indents.scm
@@ -1,0 +1,40 @@
+[
+  (struct_definition)
+  (macro_definition)
+  (function_definition)
+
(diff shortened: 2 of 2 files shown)


theHamsta (approved):

savq:
I forgot to add `catch_clause` and `finally_clause` to the branches, but I just tried to do it and it doesn't seem to be parsing correctly.

The Julia grammar gives `catch` a higher precedence to `finally` (unlike `else` and `elseif`) so that might be a problem. I'll see if I should open an issue/PR to fix that.

Might want to hold this PR until then..

vigoux:
Coul you mark this pr as a draft then? It will avoid merging it by thinking it's all good.

savq:
So... I went and fixed what I thought was the problem with the Julia grammar (see  tree-sitter/tree-sitter-julia#23), but that didn't actually fix the whole indentation issue (only one case).

Could this be a problem with the way `@branch` queries are handled?

vigoux:
I'd say maybe, it may be worth an issue regarding only this, to avoid holding this pr.

savq:
> it may be worth an issue regarding only this, to avoid holding this pr.

Sure. I'll open a new issue for this.

The indentation problem only happens when both `catch` and `finally` appear (which is really uncommon) so I guess it's fine for now.

theHamsta:
@savq did you resolve you indenting doubts? There is currently some problems that aren't solved for indenting. But I would still merge when your PR improves the current situation.

theHamsta:
@savq Is this ready to merge?

savq:
@theHamsta, I marked as ready because from what I understand indentation is still experimental, so having an edge case was fine as long as it was fixed later.

If it'd be better to wait until everything is working perfectly, I can mark it as draft again.

theHamsta (approved):

theHamsta:
Thank you!

PR #1402, opened 2021-06-17
One-Line Fix For Broken Spellchecking With Highlight Enabled

TheLostLambda:
Thanks for the super helpful plugin!

While setting up my new `init.lua` recently, I encountered a mysterious issue with Neovim's built-in spellchecking. It seemed to be spell checking bits of syntax that it wasn't supposed to (like function and variable names, as opposed to just strings and comments). I eventually tracked down the issue to the `highlight` feature of `nvim-treesitter` which seems to replace vim's normal `:syntax enable` to some extent.

As I now understand, Neovim's spellcheck relies on the `syntax` option to know what should be spell-checked and what shouldn't be. Unfortunately, it seems that something in `highlight.lua` prevents this from being set correctly! I've added a line to manually set this option and fix my spell checking (without breaking anything I could find?).

More visually, here is a screenshot from `master`:
image
Note that the `syntax` option is unset and the spellchecker is highlighting things like `opt.expandtab`.

Here is what it looks like with my patch:
image
The `syntax` option is now correctly set to `lua` and things like `opt.expandtab` no longer upset the spell-checker. The strings and comments in the above section, however, are still being checked as expected!

Let me know what you think! I only picked up (Neo)vim a couple of days ago, so it's possible I'm overlooking something here, but I'm happy to make any changes you think this might need!

### DIFF
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -110,6 +110,7 @@ function M.attach(bufnr, lang)
   end
 
   ts.highlighter.new(parser, {})
+  api.nvim_buf_set_option(bufnr, 'syntax', lang)
 
   local is_table = type(config.additional_vim_regex_highlighting) == 'table'
   if config.additional_vim_regex_highlighting and (not is_table or config.additional_vim_regex_highlighting[lang]) then


vigoux:
Thanks for the kind words, and welcome in the community! I am glad yo ucame right to us and tried fixing that issue! 

So, we actually chose to let this as a user option.

The rationale behind that is that tree-sitter is actually meant to replace (at least partly), the built-in syntax engine. Thus enabling both of them might be duplicating work, and we do not enable that by default.
There is an issue about exactlt that bh the way (can't search it for now, as I am in a hurry) so you might wa't to check it out.

As a summary, I'd say (and I regret to have to say that, given how friendly your are), that I am against this PR.

EDIT: I think this is actually already implemented, right below the line you changed.

TheLostLambda:
Ah! I figured this seemed like a bit too simple of a fix to have flown under the radar for so long!

I've tweaked my setup to contain:
image

And that seems to fix things!

Do you think this is the sort of thing that would be worth mentioning in the Readme? Perhaps under troubleshooting? I couldn't find any mention of that option except for in the source code (though it's likely I'm missing some in-built documentation somewhere?)

I'd be happy to pivot this PR away from code fixes and into a documentation one if you'd like!

theHamsta:
@TheLostLambda We would be really glad if you could extend our documentation so that this option is more visible. It's currently missing in `:h nvim-treesitter-highlight-mod` and the Troubleshooting section of the README.

PR #1406, opened 2021-06-17
Use `@keyword.function` consistently in bracket delimited languages

savq:
Use `@keyword.function` for `def` and `lambda`


stsewd (approved):

theHamsta:
We have `@keyword.function` until now only for languages that use keywords as delimiter for the function (`begin`, `end`) to distinguish the `end` of a function from `ends` in function.

I would prefer not to use the same capture for all the `def`, `fn, `fun` that are not paired. Didn't we close a similar PR for rust?

savq:
`fn` in Rust is highlighted with TSKeywordFunction. Same with `function` in JS. I assumed it was like that for all languages. Python was the only language I use that didn't, hence the PR.

theHamsta:
@savq  Could you do a quick check through all the languages so we either have the one option or the other for all the languages? I've seen that `func` in go would be the same issue. Then I guess we're fine to merge.

theHamsta:
But wouldn't it still be a better solution just to change Rust to `@keyword`? I always thought the rationale behind `@keyword.function` would be to just use the `@function` highlight for stuff that can be there instead of a function identifier like `function` in lua or js. But how would you highlight `def` in Python that it has a different highlight than the actual function name but also be different from an ordinary keyword?

savq:
Doing a quick grep, the languages that include `function.keyword` are

Languages with `end` delimiters:
- Fish
- Julia
- Fortran
- Lua (and teal)
- Elixir

Languages with bracket delimiters:
- Ecmascript
- Rust
- Zig


Most of the languages I use are there, so I did have a biased sample when assuming most languages used `keyword.function`.


Looking around, the languages that could be updated are:

- Bash
- Go
- Kotlin
- OCaml
- PHP
- Python
- R
- Verilog

Lisps use `function.macro` for this (and we can ignore most C-like languages).

savq:
> I'm trying to understand how you're using this highlight.

The thing with `@function` is that it highlights both definitions and calls, so using a different color for the keywords that define functions can be helpful. It's also useful when a function has multiple modifiers. In a signature like `pub const fn foo()`, you could have `fn` in different color to stand out more.

`TSKeywordFunction` will link to `Keyword` by default, so this would only have an effect if the user (or colorscheme) opts in.

I think having the option to highlight it differently can be useful, the same way `TSKeywordOperator` exists so that one can highlight those as either operators or keywords.

theHamsta:
`@function.macro` is for real macros I wouldn't touch the lisps because there are no real reserved keywords in Lisps.

I always wanted to have different highlights for calls and definitions. A pity that we rejected that in the beginning of the project. It would be really good to have https://github.com/neovim/neovim/pull/14196 to enable customization without too many official highlight groups.

Can you update this PR for the list of languages you compiled?

queries/python/highlights.scm
```diff
@@ -195,17 +195,20 @@
...
   "assert"
   "async"
   "await"
   "class"
-  "def"
   "except"
   "exec"
   "finally"
   "global"
-  "lambda"
```
theHamsta, line 208: The original intention of `@keyword.function` was to resemble the style of Vim regex highlights for `end`-based languages. They usually have the function highlight applied to the `end` block while other languages use regular `keyword` for `func` `def`
image

image

savq:
I updated the highlight definitions for the missing languages listed above (except PHP, that was added in a different PR last week).

theHamsta (approved):

queries/go/highlights.scm
```diff
@@ -88,7 +88,6 @@
   "continue"
   "default"
   "defer"
-  "func"
   "go"
```
theHamsta, line 91: `"go"` is used to start goroutines. Is it a `@keyword.function`? I don't know.

savq: "go" precedes calls, not definitions, so I don't think it is.

PR #1410, opened 2021-06-20
tree-sitter-cmake integration

uyha:
I am opening this pull request to start integrating my [tree-sitter-cmake](https://github.com/uyha/tree-sitter-cmake) parser into `nvim-treesitter`. Any recommendations for improvements are welcome.


lua/nvim-treesitter/parsers.lua
```diff
@@ -601,6 +601,13 @@ list.fortran = {
   },
 }
 
+list.cmake = {
+  install_info = {
+    url = "https://github.com/uyha/tree-sitter-cmake",
+    files = { "src/parser.c"},
+  },
+}
+
```
vigoux, line 610: Could you just add you as the maintainer on this language?
As you maintain the parser, you may want to also maintain the queries here?

queries/cmake/highlights.scm
```diff
@@ -0,0 +1,87 @@
...
+  "DEPRECATION"
+  "NOTICE"
+  "STATUS"
+  "VERBOSE"
+  "DEBUG"
+  "TRACE"
+  "CHECK_START"
+  "CHECK_PASS"
+  "CHECK_FAIL"
+] @keyword
```
vigoux, line 87: Are all of those keywords or built-in functions?

uyha: They are kind of contextual keywords, as CMake doesn't have a way to specify named arguments, they just parse it at runtime with these keywords as markers.

lua/nvim-treesitter/parsers.lua
uyha: I've added myself as the maintainer, and yes, I intend to maintain the queries here as well

theHamsta (commented):
Really looking forward for this! Keep on with your awesome work!

uyha:
I think the queries are good enough to be merged in now. I've spotted a few bugs with either the parser or the queries so I will fix it later.

theHamsta:
Does your cmake parser not understand comments yet?

queries/cmake/highlights.scm
```diff
@@ -0,0 +1,101 @@
...
+[
+ (foreach_command)
+ (endforeach_command)
+ (while_command)
+ (endwhile_command)
+] @repeat
+
+[
+ "1"
+ "ON"
```
theHamsta, line 39: wouldn't it be better to use constant highlighting for those? The vim regex highlights use bold text. You could also distinguish them in `AND` `NOT` operators, `FALSE` `TRUE` boolean

queries/cmake/highlights.scm
```diff
@@ -0,0 +1,101 @@
+[
+ (bracket_content)
+ (quoted_element)
+ (unquoted_argument)
+] @parameter
```
theHamsta, line 5: for some reason everything is highlighted as parameter. We usually highlight parameters only in a function definition.

queries/cmake/highlights.scm
```diff
@@ -0,0 +1,101 @@
...
+
+(variable) @variable
+
```
theHamsta, line 8: `(quoted_argument) @string`
?

queries/cmake/highlights.scm
```diff
@@ -0,0 +1,101 @@
...
+] @variable.builtin
+
+[
+ (normal_command)
+ (function_command)
+ (endfunction_command)
+ (macro_command)
+ (endmacro_command)
+ (message_command)
+] @function
```
theHamsta, line 21: We normally don't highlight the whole function. Just the function name.

queries/cmake/highlights.scm
```diff
@@ -0,0 +1,101 @@
...
+
+[
+ (if_command)
+ (elseif_command)
+ (else_command)
+ (endif_command)
+] @conditional
```
theHamsta, line 28: Same here. `(if) @conditional` would only highlight the `if`

uyha:
> 
> 
> Does your cmake parser not understand comments yet?

Not yet, that's something i plan to work on also

queries/cmake/highlights.scm
uyha: Yes, I missed that when reading the contributing document. I will update them accordingly.

uyha:
Back to WIP since I decided to change how the grammar is written as I understand better how `tree-sitter` works

uyha:
@theHamsta @vigoux I've finished my parser and I think the queries are also ready. Could you have a look?

theHamsta:
image

you have to wrap string with `token` so that comments can't start within a string literal

theHamsta:
I think it would be could to have string interpolation highlighted like f-strings in Python

image

theHamsta:
It would be nice when `(` `${` `}` and similar would be highlighted.

One could also highlight all CAPS as constants (not sure about that one) or give a special highlight for built-in variables

You can also get some inspiration from     `pboettch/vim-cmake-syntax`

uyha:
> 
> 
> image
> 
> you have to wrap string with `token` so that comments can't start within a string literal. At least this helps sometimes. I didn't have a look on your code.

I'll have a look at that. I didn't think about this case at all. Thanks for mentioning this.



> 
> 
> I think it would be could to have string interpolation highlighted like f-strings in Python
> 
> image

image
String interpolation highlighting works on my side, could you check your theme if capture `@variable` is highlighted?




> 
> 
> It would be nice when `(` `${` `}` and similar would be highlighted.
> 
> One could also highlight all CAPS as constants (not sure about that one) or give a special highlight for built-in variables
> 
> You can also get some inspiration from `pboettch/vim-cmake-syntax`

I thought about it but I'm not sure how it will interact with plugins that highlight pairs of brackets like https://github.com/luochen1990/rainbow

theHamsta:
`@variable` is normally not highlighted but there can be even more in a interpolation than just a variable `${foo function() 2}` the `@none` highlight will reset the highlight in the interpolation region (some one could also set green background for string and even the background will be removed with `@none` in a interpolation region.

> I thought about it but I'm not sure how it will interact with plugins that highlight pairs of brackets like https://github.com/luochen1990/rainbow.

Those groups can be deactivated in user config. You should set them nonetheless for everyone who wants them.

queries/cmake/highlights.scm
```diff
@@ -0,0 +1,40 @@
...
+ (endif)
+] @conditional
+
+[
+ (foreach)
+ (endforeach)
+ (while)
+ (endwhile)
+] @repeat
+
```
theHamsta, line 40: Do you plan to add highlights for things like `ON`, `OFF`, `TRUE`, `FALSE` and operators like `AND`, `OR`. This can also be done in a subsequent PR

uyha: yes, I want to have a minimum working version first then I will implement these improvements late.

theHamsta: But at least all character tokens like `(` `${` `}` should be included.

uyha:
> 
> 
> `@variable` is normally not highlighted but there can be even more in a interpolation than just a variable `${foo function() 2}` 
> 

But CMake doesn't have the kind of string interpolation that evaluates an expression and stringify its value, so I thought it makes more sense as capture them as `@variable`. 

>
> the `@none` highlight will reset the highlight in the interpolation region (some one could also set green background for string and even the background will be removed with `@none` in a interpolation region.
>
I am a bit confused why `@none` is mentioned, are all nodes that didn't get captured will be assign `@none`?

>
>Those groups can be deactivated in user config. You should set them nonetheless for everyone who wants them.
>
Ok, then I will make some adjustments in my parser to make this easier.

theHamsta:
>But CMake doesn't have the kind of string interpolation that evaluates an expression and stringify its value, so I thought it makes more sense as capture them as @variable.

`${...}` just means evaluate everything within the braces. The dollar braces can also be nested or there can be function calls.


> I am a bit confused why @none is mentioned, are all nodes that didn't get captured will be assign @none?

`@none` will be assigned to Vims `None` highlight group. It truely resets all other highlights. I would just write
```
(normal_var
  "${" @punctuation.special
  (variable) @none  ; or however you want to call the in-between node
  "}" @punctuation.special)

["(" ")"] @punctuation.bracket
```

> Ok, then I will make some adjustments in my parser to make this easier.

I don't understand. Your parser already has tokens for `"("` `"}` (see code above)

uyha:
>
>I don't understand. Your parser already has tokens for "(" "} (see code above), You can see them by pressing a in nvim-treesitter/playground
>
Ah yes, I just over complicate the problem for nothing. I add the queries accordingly.

theHamsta:
I just tried these
```scheme
((unquoted_argument) @boolean
 (#any-of? @boolean "ON" "OFF" "TRUE" "FALSE"))

((unquoted_argument) @keyword.operator
 (#any-of? @keyword.operator "AND" "OR" "NOT"))
```

uyha:
How about `$CACHE` and `$ENV`, should they be considered `@punctuation.special` or `@variable.builtin`?

uyha:
> 
> 
> I just tried these
> 
> ```scheme
> ((unquoted_argument) @boolean
>  (#any-of? @boolean "ON" "OFF" "TRUE" "FALSE"))
> 
> ((unquoted_argument) @keyword.operator
>  (#any-of? @keyword.operator "AND" "OR" "NOT"))
> ```

It's not quite correct since they are only booleans and operators when used inside `if`, `elseif`, and `while`

theHamsta:
> It's not quite correct since they are only booleans and operators when used inside if, elseif, and while

or any other function that uses them. You can also ignore them fore now. Later I would try to have a look how `pboettch/vim-cmake-syntax` handles them. I guess they just have a highlight for built-in identifiers which would also make sense.

> How about $CACHE and $ENV, should they be considered @punctuation.special or @variable.builtin?
How you prefer. I would do it similar to bash. They use `@punctuaction.special` for the `$` 
```
[
 "$"
] @punctuation.special

```
They also have queries for `${...}`. Would make sense to have them the same in cmake.

theHamsta:
You're right CMake does not support nested function calls. But this works on my machine with `!cmake -P %`
```
set(foo "ha")
set(bar "fo")
message(FATAL_ERROR "${${bar}o}")
```

uyha:
> 
> 
> You're right CMake does not support nested function calls. But this works on my machine with `!cmake -P %`
> 
> ```cmake
> set(foo "ha")
> set(bar "fo")
> message(FATAL_ERROR "${${bar}o}")
> ```

Yes, nested variable derefences like that are normal, but I'm not sure which capture group should be associated with it so I thought `@variable` fits best.

theHamsta:
The folding and the highlighting really work great!

theHamsta:
> Yes, nested variable derefences like that are normal, but I'm not sure which capture group should be associated with it so I thought @variable fits best.

But then `o` and `bar` would be `@variable`. TSVariable is not set by default so most users won't see it. `@none` is better because it can really delete highlights instead of just painting over them.  You can also use it in addition to `@variable` when you assume just one identifier for now.

theHamsta:
Just if you need more test cases for your parser. At the moment it has problems with parens grouping arguents like in 

`if((A OR B) AND C)`

uyha:
> 
> 
> Just if you need more test cases for your parser. At the moment it has problems with parens grouping arguents like in
> 
> `if((A OR B) AND C)`
> 
> or just `()` anywhere

Yup, another item in the TODO list.

queries/cmake/highlights.scm
uyha: Now, `ENV`, `CACHE`, `$`, `{`, `}`, `(`, and `)` are included.

theHamsta (approved):
LGTM

Julian:
I get

```
nvim-treesitter[cmake]: Error during compilation                                                                                                                                                
src/scanner.cc:77:60: error: use of undeclared identifier 'nullptr'                                                                                                                             
void *tree_sitter_cmake_external_scanner_create() { return nullptr; }                                                                                                                           
                                                           ^                                                                                                                                    
1 error generated.                                                                                                                                                                              
```

on startup after updating to this (and having `ensure_installed = "maintained"`)

uyha:
I just updated my parser. Could you try again?

Just out of curiosity, could you give the info about the compiler that you're using?

Julian:
Thanks!

I'm on macOS (11.3.1 + clang 12.0.5)

PR #1427, opened 2021-06-25
Adds support for PHP's arrow funcion

tk-shirasaka:
I added `fn` as a arrow funciton keyword for highlighting PHP code

### DIFF
--- a/queries/php/highlights.scm
+++ b/queries/php/highlights.scm
@@ -76,9 +76,17 @@
 ; Keywords
 
 [
+ "as"
+] @keyword.operator
+
+[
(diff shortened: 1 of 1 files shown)


stsewd (commented):
Thanks! I think we are going to highlight these as `@keyword.function` instead of `@keyword` https://github.com/nvim-treesitter/nvim-treesitter/pull/1406

tk-shirasaka:
@stsewd 
Thank you for reviewing this!
I've fixed the problem so can I check it again?

stsewd (approved):
Thanks!

PR #1445, opened 2021-06-28
parsers: add HCL

MichaHoffmann:
# What

A parser for HCL (https://github.com/hashicorp/hcl). It could be useful for terraform, nomad, waypoint etc. It works fairly well for me right now.  

https://github.com/MichaHoffmann/tree-sitter-hcl

I'm using the highlight queries with the `onedark` colorscheme right now and they look ok, but are probably not canonical: 

Auswahl_024

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -355,6 +355,7 @@ list.hcl = {
     branch = "main"
   },
   filetype = "hcl",
+  used_by = { "terraform", "packer", "nomad" },
 }
 
 -- FIXME(vigoux): markdown is broken for now
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -355,7 +355,6 @@ list.hcl = {
     branch = "main"
   },
   filetype = "hcl",
-  used_by = {"tf", "tfvars"}
 }
 
 -- FIXME(vigoux): markdown is broken for now
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -348,6 +348,16 @@ list.haskell = {
   }
 }
 
+list.hcl = {
+  install_info = {
+    url = "https://github.com/MichaHoffmann/tree-sitter-hcl",
+    files = {"src/parser.c", "src/scanner.cc"},
+    branch = "main"
+  },
+  filetype = "hcl",
+  used_by = {"tf", "tfvars"}
+}
+
 -- FIXME(vigoux): markdown is broken for now
 -- list.markdown = {
 --   install_info = {
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -24,21 +24,21 @@
   "]"
   "("
   ")"
-  "[*]"
-  ".*"
-] @specialChar
+] @punctuation.bracket
 
 [
   "."
+  ".*"
   ","
-] @specialChar
+  "[*]"
+] @punctuation.delimiter
 
 [
   (ellipsis)
   "\?"
   ":"
   "=>"
-] @special
+] @punctuation.special
 
 [
   "for"
@@ -55,18 +55,19 @@
   (heredoc_template)
 ] @string
 
+(template_interpolation) @string.escape
+
 (heredoc_identifier) @namespace
 (heredoc_start) @namespace
-(template_interpolation) @statement
 (numeric_lit) @number
 (bool_lit) @boolean
 (null_lit) @constant
 (comment) @comment
-(identifier) @identifier
+(identifier) @variable
 
 (block (identifier) @namespace)
 (function_call (identifier) @function)
 (function_call (function_arguments) @parameter)
-(attribute (identifier) @symbol)
+(attribute (identifier) @definition.var)
 
 (ERROR) @error
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -61,9 +61,9 @@
 
 
 [
-  (heredoc_identifier) @punctuation.delimiter
-  (heredoc_start) @punctuation.delimiter
-]
+  (heredoc_identifier)
+  (heredoc_start)
+] @punctuation.delimiter
 
 (template_interpolation) @string.escape
 
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -36,9 +36,13 @@
 [
   (ellipsis)
   "\?"
-  ":"
   "=>"
 ] @punctuation.special
+
+[
+  ":"
+  "="
+] @none
 
 [
   "for"
@@ -55,19 +59,22 @@
   (heredoc_template)
 ] @string
 
+
+[
+  (heredoc_identifier) @punctuation.delimiter
+  (heredoc_start) @punctuation.delimiter
+]
+
 (template_interpolation) @string.escape
 
-(heredoc_identifier) @namespace
-(heredoc_start) @namespace
 (numeric_lit) @number
 (bool_lit) @boolean
 (null_lit) @constant
 (comment) @comment
 (identifier) @symbol
 
-(block (identifier) @namespace)
+(block (identifier) @type)
 (function_call (identifier) @function)
-(function_call (function_arguments) @parameter)
-(attribute (identifier) @symbol)
+(attribute (identifier) @keyword)
 
 (ERROR) @error
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -68,6 +68,6 @@
 (block (identifier) @namespace)
 (function_call (identifier) @function)
 (function_call (function_arguments) @parameter)
-(attribute (identifier) @definition.var)
+(attribute (identifier) @symbol)
 
 (ERROR) @error
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -71,10 +71,10 @@
 (bool_lit) @boolean
 (null_lit) @constant
 (comment) @comment
-(identifier) @symbol
+(identifier) @variable
 
 (block (identifier) @type)
 (function_call (identifier) @function)
-(attribute (identifier) @keyword)
+(attribute (identifier) @field)
 
 (ERROR) @error
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -63,7 +63,7 @@
 (bool_lit) @boolean
 (null_lit) @constant
 (comment) @comment
-(identifier) @variable
+(identifier) @symbol
 
 (block (identifier) @namespace)
 (function_call (identifier) @function)
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -1,0 +1,72 @@
+; highlights.scm
+
+[
+  "!"
+  "\*"
+  "/"
+  "%"
+  "\+"
+  "-"
+  ">"
+  ">="
+  "<"
+  "<="
+  "=="
+  "!="
+  "&&"
+  "||"
+] @operator
+
+[
+  "{"
+  "}"
+  "["
+  "]"
+  "("
+  ")"
+  "[*]"
+  ".*"
+] @specialChar
+
+[
+  "."
+  ","
+] @specialChar
+
+[
+  (ellipsis)
+  "\?"
+  ":"
+  "=>"
+] @special
+
+[
+  "for"
+  "in"
+] @repeat
+
+[ 
+  "if"
+] @conditional
+
+[
+  (string_lit)
+  (quoted_template)
+  (heredoc_template)
+] @string
+
+(heredoc_identifier) @namespace
+(heredoc_start) @namespace
+(template_interpolation) @statement
+(numeric_lit) @number
+(bool_lit) @boolean
+(null_lit) @constant
+(comment) @comment
+(identifier) @identifier
+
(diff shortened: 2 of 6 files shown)


queries/hcl/highlights.scm
```diff
@@ -0,0 +1,72 @@
...
+[
+  "{"
+  "}"
+  "["
+  "]"
+  "("
+  ")"
+  "[*]"
+  ".*"
+] @specialChar
```
stsewd, line 29: These should be `punctuation.bracket`

stsewd (commented):
Thanks, please check https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights for valid capture groups.

queries/hcl/highlights.scm
MichaHoffmann: hey, thank you! i changed it just now. I think (hope) all capture groups are valid now.

stsewd (approved):
LGTM, you can also try adding injections for comments https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/yaml/injections.scm

theHamsta (approved):

theHamsta:
Thanks!

MichaHoffmann:
I cannot figure out the CI failures, can someone help explain them?

stsewd:
Looks like the parser doesn't compile on Windows when using cl https://github.com/nvim-treesitter/nvim-treesitter/blob/fe47dbe02bc243315ee185e9dc9c68b4e9dc70b2/.github/workflows/check-query-files-and-compilation.yml#L17-L18

Maybe you can exclude it https://github.com/nvim-treesitter/nvim-treesitter/blob/fe47dbe02bc243315ee185e9dc9c68b4e9dc70b2/.github/workflows/check-query-files-and-compilation.yml#L59 but better to figure out why isn't compiling on Windows.

lua/nvim-treesitter/parsers.lua
```diff
@@ -348,6 +348,16 @@ list.haskell = {
...
 }
 
+list.hcl = {
+  install_info = {
+    url = "https://github.com/MichaHoffmann/tree-sitter-hcl",
+    files = {"src/parser.c", "src/scanner.cc"},
+    branch = "main"
+  },
+  filetype = "hcl",
+  used_by = {"tf", "tfvars"}
```
gegoune, line 358: Should that not contain `terraform`? [hashivim/vim-terraform](https://github.com/hashivim/vim-terraform) sets `ft=terraform` for `*.tf` files.

MichaHoffmann:
> Looks like the parser doesn't compile on Windows when using cl
> 
> https://github.com/nvim-treesitter/nvim-treesitter/blob/fe47dbe02bc243315ee185e9dc9c68b4e9dc70b2/.github/workflows/check-query-files-and-compilation.yml#L17-L18
> 
> Maybe you can exclude it
> 
> https://github.com/nvim-treesitter/nvim-treesitter/blob/fe47dbe02bc243315ee185e9dc9c68b4e9dc70b2/.github/workflows/check-query-files-and-compilation.yml#L59
> but better to figure out why isn't compiling on Windows.

i checked, it did compile in windows, but the step afterwards ( check queries ) failed seemingly. At least in the Job `compile windows` 

```
Downloading...

Checking out locked revision

Compiling...

Treesitter parser for hcl has been installed
```

lua/nvim-treesitter/parsers.lua
MichaHoffmann: Is the filetype canonical? Maybe it would be better to write a ftdetect function that adds the `hcl` filetype for terraform files. What do you think?

MichaHoffmann:
looks like neovim does not build in the macos CI jobs: 
```
Undefined symbols for architecture x86_64:
  "_ts_query_cursor_set_match_limit", referenced from:
      _node_rawquery in lto.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [bin/nvim] Error 1
```

lua/nvim-treesitter/parsers.lua
gegoune: I am really not sure, but I remember that tf filetyoe is something different to default vim than terraform.

dimbleby: > hashivim/vim-terraform sets ft=terraform for *.tf files.

I do most of the maintenance on hashivim/vim-terraform.  Happy to consider tweaking the filetype we use, if that's the Right Thing To Do.  Especially happy to consider proposals made in the form of MRs...

I don't think that vim-terraform should use `hcl` - the plugin provides some genuinely terraform-specific features.  But perhaps it would make sense to use `hcl.terraform`?  

(I'm not entirely sure what all the implications of that are, but it sounds plausible and I'm open to guidance!)

>  remember that tf filetyoe is something different to default vim than terraform.

Yes, TinyFugue

theHamsta:
@MichaHoffmann  This is a home brew issue. A unreleased tree-sitter function was used in core.

lua/nvim-treesitter/parsers.lua
MichaHoffmann: `hcl.terraform` sounds right to me. I created https://github.com/hashivim/vim-terraform/pull/168 and will drop the `used_by` fields from the configuration here.

MichaHoffmann: Since other plugins set the filetype to `terraform` or `nomad` ( e.g. https://github.com/jvirtanen/vim-hcl/blob/master/ftdetect/hcl.vim ), I added `used_by: { "terraform", "nomad", "packer" }` to the parser configuration and closed the `vim-terraform` PR again. This seemed to be the most consensual configuration.

MichaHoffmann:
I think i have figured out the build errors. I was able to reproduce them using the `-pedantic` flag in gcc or clang: 

```src/scanner.cc:98:23: warning: designated initializers are a C++20 extension [-Wc++20-designator]
      Context ctx = { .type = QUOTED_TEMPLATE };
                      ^
src/scanner.cc:116:35: warning: designated initializers are a C++20 extension [-Wc++20-designator]
        context_stack.push_back({ .type = TEMPLATE_INTERPOLATION});
                                  ^
src/scanner.cc:142:33: warning: designated initializers are a C++20 extension [-Wc++20-designator]
      context_stack.push_back({ .type = HEREDOC_TEMPLATE, .heredoc_identifier = identifier });
```
I fixed them in release `0.3.1` and pointed `lockfile.json` to the new revision.

dimbleby:
I took this for a very quick spin - I mostly like it!

A couple of quibbles for your consideration

- I'm not in love with matching the first identifier on a block as `namespace`.  The HCL spec calls this a block type - maybe `type` would be better?  But perhaps this one is just an artifact of it not looking super-attractive in the colour scheme that I happen to be using

- WIBNI this respository imposed some consistency about things like heredoc starts.  I find `@punctuation.delimiter`, `@string.special`, `@constant`, and now `@namespace` all used for presumably the same concept.

There are also some, to my mind, slightly odd things in the highlighting / parsing of a block like this:
```terraform
variable "credentials" {
  description = <<EOF
Credentials to provide to the Azure provider.  Leave these empty if you are already logged in with
suitable credentials (via az login).
EOF
  type = object({
    tenant_id       = string
    subscription_id = string
    client_id       = string
    client_secret   = string
  })
}
```

image

specifically:
- Most uncomfortable is that the equals signs inside the object are being caught by some highlighting, not sure what exactly
- `object` being highlighted as a function feels ever so slightly weird, as a terraform user I'd never thought of it that way.  But perhaps in the HCL grammar that is what's going on, so fair enough
- similarly: from a terraform point of view, it would be desirable to recognise `string` as being more interesting than a regular identifier.  I guess it just isn't, in the HCL grammar.

In a typical terraform file I see a _lot_ that ends up being resolved to `identifier = identifier`, with an awful lot of the same colour everywhere (blue, in my case).  Would it be feasible / sensible to differentiate between identifiers on the left hand side of an assignment, and others?  To the devloper, this pattern feels more like "variable = value" than "identifier = identifier".

MichaHoffmann:
Hey @dimbleby,

Thanks for your comments! I have to admit i mostly selected the matches to work with my color scheme. To address some of your comments: 

* the `ìdentifier = identifier` issue can be fixed principally by changing the captures `(attribute (identifier) @symbol)` and `(identifier) @symbol` to something more sensible. ( do you have a recommendation? most of the things i tried resulted in the same highlighting in at least my colorscheme ( onedark at the moment ) )
* the `=` are caught by `(function_call (function_arguments) @parameter)` i think. its probably best to just drop that capture
* you are correct, from a syntax point of view `object` is a function there and `string` is just an identifier, although one could write a `terraform/highlights.scm` where certain keywords are highlighted differently i assume. 
* regarding `@namespace` for heredoc start and object: i basically just tried to get a first version going. what would you recommend here?

dimbleby:
I wouldn't presume to make a recommendation about heredocs, maybe one of the repository owners will chip in.

It's not your fault that there were already three different ways of marking this concept; though I suppose it would be your fault if we go to _four_ different ways ;-)

I'm no expert on how the captures do or should work, the above were more meant as user comments than any sort of advice on how the code should be written.

MichaHoffmann:
I have some ideas to address them, would it be okay for you if i do it in a subsequent MR though?

dimbleby:
By all means!  I'm just dropping by with some thoughts, feel free to ignore me completely!

MichaHoffmann:
@dimbleby No no, thank you so much for your comments. Its super cool to get some feedback! There is definitely some "works-on-my-machine" going on in my captures. I would love to get some more feedback from you in the next MR!

dimbleby:
Color scheme is https://github.com/lifepillar/vim-gruvbox8.  I'm afraid my employer would not approve of me sharing the terraform definitions that I work with.

MichaHoffmann:
I tried to fix some of your issues in the latest commit, what do you think? 

Auswahl_025

attributes now highlight differently on the left side as seen by `foo = bar.baz[*]`, the issue with the weird `=` highlighting in objects as that are function arguments is fixed. block identifiers are `@type` now and heredocs start as `punctuation.delimiter` to be uniform.

dimbleby:
Cool, thanks!

I see that `tenant_id` and `string` are still coming out the same as each other.  Are they harder to separate?

`@keyword` doesn't seem like a very natural choice for things that definitely aren't keywords.  It's hard to map concepts from very different languages into a small set of common things, but still.  

Maybe the left hand side of assignments could be `@field` and the right hand side `@variable`?  That feels to me like a reasonably faithful interpretation of what HCL is doing, what do you think?

(So far as I can see, `@variable` becomes `TSVariable`, which then does not have any highlight group associated with it by default.  But that's ok, not everything has to be coloured.  Unhighlighted is also a higlight...)

MichaHoffmann:
Mapping the left side to `@field` and right right to `@variable` seems reasonable. I'll try it out tomorrow! Yeah since `tennant_id` and `string` are object fields they are not highlighted by the `(attribute (identifier))` query, but that seems ok since the left hand side of an object element could be an arbitrary expression like a string literal or something. I'm not sure what can be done with objects at the moment.

MichaHoffmann:
Ok, i took your advice and used `@field` for attribute identifiers and `@variable` for other identifiers. it looks fine on all colorschemes i tried and also makes sense.

Furthermore i bumped the version in the `lockfile.json` to `v0.3.2` that compiles with clang on macos. I also added a CI step to my project to make sure it keeps building.

dimbleby:
Nice, I do think this is an improvement, thanks.

Current minor annoyances:

- (same as previously mentioned I think) no highlighting inside object attributes eg here my intuition expects `this` and `not` to be highlighted the same as `@field`, just like `bar` is
```terraform
foo "foo" {
  bar = {
    this = is
    not = highlighted
  }
}
```
- the highlighting of an entire interpolation as `@string.escape` feels a bit aggressive to me.  Certainly I'd want the surrounding `${}` highlighted, but I'd probably expect the contents of the interpolation just to be highlighted as whatever they are, if you see what I mean.  Just not being highlighted as `@string` is already enough to mark the difference, in my opinion.

MichaHoffmann:
Hey, 

Yes the highlighting of string interpolations is a little aggressive now, i will capture the starting and ending tokens for them in the next version explicitely in a node ( at the moment they are hidden tokens ), that will help to isolate only these tokens for the `@string.escape`.   I definitely see the point with the objects, i will think about that for the next version. Maybe a capture that highlights identifiers that are inside an object element will work here.. I will experiment with that a little!

Thanks for your input!

MichaHoffmann:
The builds finally work on all platforms. :tada: 

If nothing else is immediately important, i have no other local changes.

queries/hcl/highlights.scm
```diff
@@ -0,0 +1,80 @@
...
+  (string_lit)
+  (quoted_template)
+  (heredoc_template)
+] @string
+
+
+[
+  (heredoc_identifier) @punctuation.delimiter
+  (heredoc_start) @punctuation.delimiter
+]
```
stsewd, line 66:
```suggestion
[
  (heredoc_identifier)
  (heredoc_start)
] @punctuation.delimiter
```

stsewd:
@gegoune @dimbleby if you agree this looks good now we can merge it :+1:

queries/hcl/highlights.scm
MichaHoffmann: oh! yeah thank you!

gegoune:
@stsewd Looks good after quick test. Thanks @MichaHoffmann!

MichaHoffmann:
Thanks everyone for helping, reviewing and merging!

PR #1450, opened 2021-06-29
python(highlights): don't use type for APP

theHamsta:
Apply fix suggested in https://github.com/nvim-treesitter/nvim-treesitter/issues/1441#issuecomment-870342791

### DIFF
--- a/queries/python/highlights.scm
+++ b/queries/python/highlights.scm
@@ -9,7 +9,7 @@
 
 ;; Identifier naming conventions
 ((identifier) @type
- (#match? @type "^[A-Z]"))
+ (#match? @type "^[A-Z].*[a-z]"))
 ((identifier) @constant
  (#match? @constant "^[A-Z][A-Z_0-9]*$"))
 


queries/python/highlights.scm
```diff
@@ -9,7 +9,7 @@
 
 ;; Identifier naming conventions
 ((identifier) @type
- (#match? @type "^[A-Z]"))
+ (#match? @type "^[A-Z].*[a-z]"))
 ((identifier) @constant
  (#match? @constant "^[A-Z][A-Z_0-9]*$"))
```
stsewd, line 14: isn't this overridden by this query already?

Screenshot from 2021-06-29 14-36-21

theHamsta: It's overwritten if the highlight of constant is strong enough to overwrite type. Actually both of them are applied at the same time only Constant over type. So when Type is bold and Constant not bold Constant can not overwrite the boldness of Type just the color. I guessed that this was the issue of the use.

theHamsta: @gjeusel is there a reason why constant couldn't visually overwrite `@type` for you. Italics? Something fancy or just because it's displayed in `:TSHighlightsUnderCursors`

stsewd (approved):
it makes sense if they have extra decorations like italics and strong.

queries/python/highlights.scm
gjeusel: Something fancy, setting `hi! link pythonTSConstant white`, hence defaulting the constant highlight to the type one.

PR #1462, opened 2021-07-01
c: highlight struct and enum keywords with @type

gpanders:
In C, `struct foo` is a type. That is, given:

```c
struct foo x;
```

the variable `x` has type `struct foo`, not `foo` (and likewise for
`enum bar`). Therefore the `struct` and `enum` keywords should be
highlighted as types, not keywords.

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -3,13 +3,11 @@
 [
   "const"
   "default"
-  "enum"
   "extern"
   "inline"
   "return"
   "sizeof"
   "static"
-  "struct"
   "typedef"
   "union"
   "volatile"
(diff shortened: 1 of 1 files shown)


theHamsta:
> Therefore the struct and enum keywords should be highlighted as types, not keywords.

I would argue against that. `struct` is a reserved keyword and there is some value to visually distinguish it from identifiers.

gpanders:
On further thought I actually think I (kind of) agree.

I think it is useful to distinguish between a struct *definition*:

```c
struct foo {
    int a;
    char b;
};
```

and using a struct as a type:

```c
void do_something(struct foo *f);
```

In the later case I think `struct foo` ought to be highlighted as `@type`, but in the former case I don't think it's as clear cut, since there `struct` is being used differently.

In any case, it seems that there isn't really an obvious way to do this so I'll just add these modifications in my own query file and leave this one as-is.

theHamsta:
And the config is also shared with C++. But the argument use case should be pretty easy catch if you want to catch that.

PR #1463, opened 2021-07-01
feat(dart): add function call expression highlight

akinsho:
This is a work around since currently the upstream grammar does not support a `call_expression`. I hope to add this at some point in the future, but I'm not sure exactly when I'll have time to do it. In the short term, this allows function calls to be highlighted by using the fact that function calls are structured in a certain way, i.e. an identifier with a sibling node that has an argument part.

### DIFF
--- a/queries/dart/highlights.scm
+++ b/queries/dart/highlights.scm
@@ -3,10 +3,16 @@
 ; Methods
 ; --------------------
 (super) @function
+
 ; TODO: add method/call_expression to grammar and
 ; distinguish method call from variable access
 (function_expression_body (identifier) @function)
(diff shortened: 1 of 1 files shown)


theHamsta (approved):

akinsho:
@theHamsta I wonder if you can help, I'm seeing quite a strange situation where this and the other changes worked perfectly fine on my linux machine on the latest nightly and on my mac using the latest nightly via homebrew, I'm seeing query errors in dart. when I try these queries in the playground they work fine i.e. no errors but reverting to before I made these changes seems to fix the errors on mac.

Any idea how or why these could be behaving differently on different operating systems?

PR #1465, opened 2021-07-01
c: don't treat uppercase functions as constants

gpanders:
A macro function such as

    #define FOO(x) ...

should be highlighted with `@function.macro`, even though the name is in
all-caps (which would otherwise be treated as a constant). Similarly,
call sites of such functions, e.g.

    int y = FOO(x);

should be highlighted with `@function` instead of `@constant`.

This is as simple as moving the query definition for functions
after the query definition for constants.

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -110,15 +110,6 @@
 (number_literal) @number
 (char_literal) @character
 
-(call_expression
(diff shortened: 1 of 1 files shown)


theHamsta (approved):

theHamsta:
@vigoux

PR #1472, opened 2021-07-02
queries(hcl): improve highlighting

MichaHoffmann:
* bumped the dependency of the hcl parser to make some tokens visible
* added a query for object element keys that are identifiers to highlight them as though they were block attributes
* dont highlight the whole `quoted_template` and `heredoc_template` as strings, but only the "stringy" parts

# Before
Auswahl_026

# After
Auswahl_027

### DIFF
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -64,10 +64,12 @@
   (heredoc_start) ; END
 ] @punctuation.delimiter
 
-[
-  (template_interpolation_start) ; ${
-  (template_interpolation_end) ; }
-] @string.escape
+( template_interpolation
+  [
+    (template_interpolation_start) ; ${
+    (template_interpolation_end) ; }
+  ] @punctuation.bracket
+)
 
 (numeric_lit) @number
 (bool_lit) @boolean
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -54,18 +54,20 @@
 ] @conditional
 
 [
-  (string_lit)
-  (quoted_template)
-  (heredoc_template)
+  (quoted_template_start) ; "
+  (quoted_template_end); "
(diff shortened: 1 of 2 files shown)


queries/hcl/highlights.scm
```diff
@@ -54,18 +54,20 @@
...
-  (heredoc_start)
+  (heredoc_identifier) ; <<END
+  (heredoc_start) ; END
 ] @punctuation.delimiter
 
-(template_interpolation) @string.escape
+[
+  (template_interpolation_start) ; ${
+  (template_interpolation_end) ; }
+] @string.escape
```
theHamsta, line 70: Could this be highlighted lik `${...}`in bash?

MichaHoffmann: In the bash highlights they are `@punctuation.bracket`. I think one could write something like

```
( template_interpolation [ (template_interpolation_start) (template_interpolation_end) ] @punctuation.bracket )
```

That would feel like the bash queries. Should i do it like this?

theHamsta: I was more referring to the `@none` highlight in the beginning of the bash highlight to reset the string highlighting.

> That would feel like the bash queries. Should i do it like this?

I'd say yes. We should try to have bash, fish, cmake in sync. If you think the other solution is better you can create a PR changing it for all the languages.

MichaHoffmann: Ok, i changed it to `punctuation.bracket` and specialized the query to `template_interpolations`.

MichaHoffmann: It looks like this in my colorscheme now.

Auswahl_028

theHamsta: Maybe string.escape fits better, but I would prefer to change it for all languages at once and this is better done in a separate PR.

theHamsta (approved):

PR #1477, opened 2021-07-02
Allow unused_identifier in function definitions in Elixir

hauleth:


theHamsta (approved):

queries/elixir/highlights.scm
```diff
@@ -156,6 +156,8 @@
    [(identifier) @parameter
     (tuple (identifier) @parameter)
     (list (identifier) @parameter)
+    (unused_identifier) @comment
+    (map) @comment
```
theHamsta, line 160: Why is map a comment?

theHamsta (changes requested):
~~There are some queries without captures~~

theHamsta (approved):

theHamsta:
@nifoc could you have a look on this?

nifoc:
I think this was removed by accident (from `Definition with (some) arguments and guard(s)`):

```
(_
 (keyword_list (identifier) @parameter))
```

Otherwise LGMT ❤️

theHamsta:
@hauleth could you address the comment of @nifoc ?

hauleth:
@theHamsta yes, soonish

hauleth:
@theHamsta I just removed everything and allowed everything to be called in the parameters, as it is how Elixir works:

```elixir
def foo(bar(a: a)) do
  # …
end
```

Is completely valid code that can be spotted in real projects (when working with Erlang records).

nifoc:
Sadly this also removes all other `@parameter` highlights.

tree-sitter doesn't really support languages with pattern matching (in function heads) at the moment, which is why these queries are so *verbose* in the first place. Some patterns are explicitly listed in the query file to at least highlight the most common `@parameter`s in pattern matches.

Currently I would say it's nearly impossible to correctly highlight every possible pattern match ...
[This discussion](https://github.com/tree-sitter/tree-sitter/discussions/1096) also mentions the above issue and links to some further sources/issues.

hauleth:
@nifoc that discussion is pointless as even with such thing supporting all possible patterns as `quote` call can be used as a pattern. Also I think that *missing* feature is better than having feature that works only partially. That is why I think that just removing support for colouring `@parameter` and instead having full support for colouring function definitions is worth it.

And this is how it looks right now:

Zrzut ekranu 2021-07-13 o 22 27 56

nifoc:
Okay, now I understand.

To add a catch all query that highlights the function name, but doesn't highlight parameters (because none of the more specific queries match), I would suggest adding the following before `; Definition with (some) arguments and (optional) defaults`:

```
(call (function_identifier) @keyword.function
 (call
  function: (function_identifier) @function)
 (#lua-match? @keyword.function "^def%a*$"))
```

This will match both `foo` (and mark them as `@function`):

Bildschirmfoto 2021-07-13 um 22 41 01

nifoc:
Sorry if my previous comment caused any confusion. The solution could cause tree-sitter to run into an internal limit, which results in only some parameters being highlighted for functions with many parameters.

The following should match (and highlight) `foo` in both function definitions:
(Notice the `?` after the `(arguments ...)`)

```
; Definition with (some) arguments and (optional) defaults
(call (function_identifier) @keyword.function
 (call
  function: (function_identifier) @function
  (arguments
   [(identifier) @parameter
    (tuple (identifier) @parameter)
    (list (identifier) @parameter)
    (_
     (keyword_list (identifier) @parameter))
    (binary_op
     left: (identifier) @parameter
     operator: "\\\\")])?)
 (#lua-match? @keyword.function "^def%a*$"))

; Definition with (some) arguments and guard(s)
(call (function_identifier) @keyword.function
 (binary_op
  left:
   (call
    function: (function_identifier) @function
    (arguments
     [(identifier) @parameter
      (tuple (identifier) @parameter)
      (list (identifier) @parameter)
      (_
       (keyword_list (identifier) @parameter))])?)
  operator: "when")
 (#lua-match? @keyword.function "^def%a*$"))
```

Looks like this in the playground:

Bildschirmfoto 2021-07-13 um 23 01 23

hauleth:
However it will again not apply the `@parameter` to the passed arguments, yes? That was the core of the problem.

It seems that if you try to do almost any pattern matching in the function head, then it will fail. It seems that even the really basic pattern matching will make the parser to fail.

Zrzut ekranu 2021-07-14 o 14 57 41

nifoc:
> However it will again not apply the `@parameter` to the passed arguments, yes?

In most of your examples it will not highlight the `@parameter`.
(But with the `?` added it will at least highlight the `@function` in all cases)

Bildschirmfoto 2021-07-14 um 15 22 23

> That was the core of the problem.

Just so we're definitely talking about the same thing, can you maybe spell out what you wanted to achieve with this PR?
I was under the impression that it's mainly about the missing `@function` highlight.

theHamsta:
@nifoc I'm really not an elixir expert. Can you approve this as soon as the PR is ready? I have the feeling I'm not really helpful regarding elixir...

nifoc:
@theHamsta I think the bot merged this by accident?

(It doesn't break anything, just removes some `@parameter` highlights)

stsewd:
> @theHamsta I think the bot merged this by accident?
> 
> (It doesn't break anything, just removes some `@parameter` highlights)

it was by accident, but let's just keep adding improvements in another PR, as you said this doesn't break anything

PR #1487, opened 2021-07-04
Add keyword.return & keyword.yield

### LINKED ISSUE #1485: TSKeywordReturn feature request
When porting a colorscheme to treesitter I came across a difficulty that treesitter does not expose a highlighting group for the `return` keyword. Which makes it currently impossible to color it differently to other keywords. Though there are highlighting groups for `function` keyword and `operator` keyword [here](https://github.com/nvim-treesitter/nvim-treesitter/blob/4eb19c9e6da5c1628a5e2b4445595ecb39019efc/lua/nvim-treesitter/highlight.lua#L46-L48). The return keyword is pretty popular for C syntax like languages.

It would be useful to be able to have `TSKeyword.Return` as its own color highlighting group as it could be used in 

Inapplicable.

**Additional context**
Here is a sample of highlighting for the same theme (base16-ocean). The above one is using the default regexp based highlighting and the below one is using treesitter. As can be seen the return keyword is a different color.

Screen Shot 2021-07-04 at 17 38 02

antonk52:
Not too sure about [`teal`](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/teal/highlights.scm#L22) and [`fish`](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/fish/highlights.scm#L79-L81) since I wasn't able to find the meaning of `(WORD)` within a group of word.

I have tested this locally and it seems to work as expected.

closes #1485

@theHamsta

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,7 +137,6 @@ effect on highlighting. We will work on improving highlighting in the near futur
 @keyword.function
 @keyword.operator (for operators that are English words, e.g. `and`, `or`)
 @keyword.return
-@keyword.yield
 @operator (for symbolic operators, e.g. `+`, `*`)
 @exception
 @include keywords for including modules (e.g. import/from in Python)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -136,6 +136,8 @@ effect on highlighting. We will work on improving highlighting in the near futur
 @keyword
 @keyword.function
 @keyword.operator (for operators that are English words, e.g. `and`, `or`)
+@keyword.return
+@keyword.yield
 @operator (for symbolic operators, e.g. `+`, `*`)
 @exception
 @include keywords for including modules (e.g. import/from in Python)
--- a/queries/c_sharp/highlights.scm
+++ b/queries/c_sharp/highlights.scm
@@ -260,8 +260,6 @@
  "params"
  "operator"
  "default"
- "yield"
- "return"
  "abstract"
  "const"
  "extern"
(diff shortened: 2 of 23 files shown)


theHamsta (approved):

CONTRIBUTING.md
```diff
@@ -136,6 +136,8 @@ effect on highlighting. We will work on improving highlighting in the near futur
 @keyword
 @keyword.function
 @keyword.operator (for operators that are English words, e.g. `and`, `or`)
+@keyword.return
+@keyword.yield
```
vigoux, line 140: I would be more inclined to have those two as only one.

For me, even if those represent semantically two different things, they actually mean "return control to the caller".

Thus, I'd only have `keyword.return`.

queries/c_sharp/highlights.scm
```diff
@@ -288,3 +286,11 @@
...
 ] @keyword
 
+[
+  "return"
+] @keyword.return
+
+[
+  "yield"
+] @keyword.yield
+
```
vigoux, line 296: When capturing a single item, the recommended way is to get it directly, instead of through an alternation :
```suggestion
"return" @keyword.return

"yield" @keyword.yield
```

vigoux (commented):
Okay, so thanks for the huge amount of work you put in this, and welcome in the contributor community!

I left two comments:

- the first one is more generic, I'd rather merge those two, new groups into one
- the second is more of a styling issue, that applies to most of your additions

CONTRIBUTING.md
antonk52: This makes sense, thank you

queries/c_sharp/highlights.scm
antonk52: fixed, thank you

stsewd (approved):
Thanks!

theHamsta:
Nice addition! Thank you!

PR #1490, opened 2021-07-04
Better differentiate 'keywords' for HCL

rmagatti:
# Description
I noticed the queries could be improved for better terraform highlighting.
Maybe this should be an inherited query for terraform instead?

# Before
Screen Shot 2021-07-04 at 4 03 28 PM

# After
Screen Shot 2021-07-04 at 4 05 50 PM

### DIFF
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -68,7 +68,7 @@
   [
     (template_interpolation_start) ; ${
     (template_interpolation_end) ; }
-  ] @punctuation.bracket
+  ] @punctuation.special
 )
 
 (numeric_lit) @number
@@ -85,5 +85,13 @@
 ;
 ; highlight identifier keys as though they were block attributes
 (object_elem key: (expression (variable_expr (identifier) @field)))
+
+(variable_expr (identifier) @variable.builtin (#any-of? @variable.builtin "var" "local"))
+((identifier) @keyword (#any-of? @keyword "module" "resource" "variable" "data" "locals" "terraform" "provider" "output"))
+((identifier) @type.builtin (#any-of? @type.builtin "bool" "string" "number" "object" "tuple" "list" "map" "set" "any"))
+
+(object_elem val: (expression
+  (variable_expr
+    (identifier) @type.builtin (#any-of? @type.builtin "bool" "string" "number" "object" "tuple" "list" "map" "set" "any"))))
 
 (ERROR) @error
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -86,9 +86,10 @@
 ; highlight identifier keys as though they were block attributes
 (object_elem key: (expression (variable_expr (identifier) @field)))
 
-(variable_expr (identifier) @variable.builtin (#any-of? @variable.builtin "var" "local"))
-((identifier) @keyword (#any-of? @keyword "module" "resource" "variable" "data" "locals" "terraform" "provider" "output"))
+((identifier) @keyword (#any-of? @keyword "module" "root" "cwd" "resource" "variable" "data" "locals" "terraform" "provider" "output"))
 ((identifier) @type.builtin (#any-of? @type.builtin "bool" "string" "number" "object" "tuple" "list" "map" "set" "any"))
+(variable_expr (identifier) @variable.builtin (#any-of? @variable.builtin "var" "local" "path"))
+(get_attr (identifier) @variable.builtin (#any-of? @variable.builtin  "root" "cwd" "module"))
 
 (object_elem val: (expression
   (variable_expr


dimbleby:
The problem I see with this is that these "keywords" are not HCL keywords.  The spec is [here](https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md); and it just doesn't mention these words at all.

They do have special meaning in the terraform dialect of HCL - and I can get behind the wish to have them highlighted in terraform files.  But is that a good enough reason to make them HCL queries?

I don't know whether this plugin supports defining terraform queries as an extension of HCL queries; but that seems to me to be what we want.

If this is not supported and is difficult, it might be pragmatic not to worry about being too purist here.  Probably terraform is far the most common use of HCL, and perhaps the queries don't do very much harm in other use cases anyway.

(However this ends up being handled: suggest that "object", "tuple", "list, "map, "set" and perhaps also "any" should also be treated as types in terraform.)

rmagatti:
Yeah my first instinct here was that since these aren't really HCL keywords we'd want to maybe do something different. 

Maybe a whole new `terraform` parser that is really just the HCL one but with terraform specific queries would be a better way to go about this?

MichaHoffmann:
@rmagatti thank you, the example looks nice! In the `CONTRIBUTING.md` was a paragraph about inheriting queries where you can include a comment like 
```
; inherits: lang1,(optionallang)
```

at the start of the file. Maybe this could work here? I'm not sure but most javascript like languages have `; inherits ecma` at the beginning of their queries.

Possibly one could create queries for `terraform` without a new parser but have a `queries/terraform/highlights.scm` that inherits highlights from hcl? 

disclaimer: i have not done that yet, just speculating.

MichaHoffmann:
one could add `path.(module|root|cwd)` to keywords ( i noticed in the screenshot that only the `module` part is highlighted in `path.module` )

rmagatti:
>  In the CONTRIBUTING.md was a paragraph about inheriting queries
Yeah I read through that before, hopefully I'll have some free time to experiment making this terraform specific inherited queries instead this weekend. 🤞

rmagatti:
So one thing I don't quite understand how I'd do is having an inherited highlights query file without it being under a language parser.

In the case of typescript for example, it's still a separate parser, the highlights file then inherits from javascript's highlights file.

If anyone's got ideas on how to do this that'd be great, if this is not supported, maybe it should be somehow? I think last resort would be to add these extra queries to HCL but it's definitely preferable otherwise.

MichaHoffmann:
> thing I don't quite understand how I'd do is having an inherited highlights query file without it being under

The HCL parser also has the `used_by: "terraform"` field set. https://github.com/nvim-treesitter/nvim-treesitter/blob/f2076b267ef2d53c35a964a8752eef38d7040fd5/lua/nvim-treesitter/parsers.lua#L373

 To manage the filetype for terraform there is https://github.com/nvim-treesitter/nvim-treesitter/blob/a7c90103ceadbf0dff6dc1c67443f6c37e555da7/ftdetect/hcl.vim.

So the terraform file is also parsed by the hcl parser. The `inherits` directive is used here https://github.com/neovim/neovim/blob/0e22a40b6d328ac4aa07ed7bc40da200a48ef524/runtime/lua/vim/treesitter/query.lua#L42 to assemble the queries for the filetype. 

I think it should just work because the hcl parser understands the inherited queries and they are just added to your new queries.

rmagatti:
@MichaHoffmann that's what I thought too, but then it didn't work on my local tests. I might have just done something wrong though, I'll try a few more things and get back to you.

MichaHoffmann:
@rmagatti Oh, thats surprising! Is it ok for you if I try your branch locally too?

rmagatti:
Oh ofc, go for it

MichaHoffmann:
Okay, sorry for the late response. I played around with the `; inherits: hcl` directive and could not get it to work at all. It seemed that the queries in `queries/terraform/highlights.scm` did not even get loaded ( i put some garbage there to test ). They probably map to a parser and we would probably need a terraform parser that can then work with the queries to make the `; inherits: hcl` work. 

If that is the case i would suggest that we just go forward with your approach and incorporate @dimbleby s comments. What do you think @rmagatti ?

rmagatti:
@MichaHoffmann I've addressed @dimbleby's comments and yours from before too. This is how your addressed `path.(module|root|cwd)` comment looks like.

Let me know if you think there's a better way of going about it.

Screen Shot 2021-09-13 at 11 31 24 PM

MichaHoffmann:
LGTM, sorry for ghosting you for a month though!

theHamsta (approved):

theHamsta:
Also looks good for me! Thanks for the contribution! The next time you have to ping us when we forget something.

PR #1498, opened 2021-07-05
Folds: fix fold deduplication and improve start/stop logic

ecnerwala:
The old logic to deduplicate folds was too lax: it only checked that both the start and the end existed, not that they were part of the same fold. This broke situations like
```
struct a {
};
struct b {
};
```
because the C fold file defines a `translation-unit` fold node, which combined with `struct a`, caused `struct b`'s fold to be ignored. The new logic works correctly as long as the nodes are given in preorder (which seems to currently be the case); let me know if that assumption is bad.

Also, revamp the start/stop logic to be more explicit, and make it handle `foldnestmax` slightly better.


vigoux (approved):
Only one thing to say: it looks great.
I've been digging trying to find something to say, but it's just great !

PR #1517, opened 2021-07-06
Fix ecma template string braces highlight group

antonk52:
I believe I have discovered a bug and a potential fix for it. I noticed that the braces for template string quasis are coloured inconsistently, I have set vivid colours for these 2 highlight groups

```vim
hi TSPunctBracket guifg=Yellow
hi TSPunctSpecial guifg=Red
```

Screen Shot 2021-07-06 at 23 55 41

`:TSHighlightCapturesUnderCursor` output for the closing brace

```
* @string -> javascriptTSString -> TSString
* @none -> javascriptTSNone -> TSNone
* @punctuation.special -> javascriptTSPunctSpecial -> TSPunctSpecial
* @punctuation.bracket -> javascriptTSPunctBracket -> TSPunctBracket
```

It seems that `TSPunctBracket` is overlaying(not sure if the word suits here) `TSPunctSpecial` which should be preferred here.

By placing the template string section after the section with braces  in `queries/ecma/highlights.scm` the problem goes away and the appropriate highlight group gets applied to the closing brace within a template string. In my tests this change did not affect the highlighting of other braces such as objects, code blocks, classes, functions and methods.

Hope this helps 🙂

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -108,10 +108,6 @@
 ; Punctuation
 ;------------
 
-(template_substitution
-  "${" @punctuation.special
-  "}" @punctuation.special) @none
-
 "..." @punctuation.special
 
 ";" @punctuation.delimiter
@@ -175,6 +171,10 @@
 "]" @punctuation.bracket
 "{" @punctuation.bracket
 "}" @punctuation.bracket
+
+(template_substitution
+  "${" @punctuation.special
+  "}" @punctuation.special) @none
 
 ; Keywords
 ;----------


stsewd (approved):
Good catch!

numbcoder:
Thanks for your contribution. 
I found ruby had the same bug. Should I move the `@punctuation.special` group to after `@punctuation.bracket`?

stsewd:
> Thanks for your contribution.
> I found ruby had the same bug. Should I move the `@punctuation.special` group to after `@punctuation.bracket`?

Sure! feel free to send a PR

Joakker:
Even after updating, I seem to have the same problem still

image

`:TSHighlightCapturesUnderCursor` for the right one

image

And the left one

image

Am I still missing something?

antonk52:
@Joakker can you verify that you have updated `nvim-treesitter` plugin and have restarted Neovim and did `:TSUpdate` and the problem remains?

if so can you share you Neovim version and nvim-treesitter commit that you are experiencing the issue?

theHamsta:
It might be also relevant to how your language groups are defined. They really get painted on over the other and some times certain highlights cannot be overwritten (e.g. bold)

Joakker:
I can verify that after doing all that, I still have the same problem.
- nvim-treesitter commit: `d82e173` (current head)
- nvim version: `v0.5.0` (latest stable)

stsewd:
Can't replicate with this file

```js
const name = "Neovim";
const greet = `Hello, ${name}`;
```

What colorsheme are you using? can you replicate using other one?

Joakker:
Oh wait, I've tracked down the problem. It was in my config. Apparently the problem is in [nvim-ts-rainbow](https://github.com/p00f/nvim-ts-rainbow/blob/master/queries/square/parens.scm), since it doesn't recognize `${` as a bracket.

Thanks for your help, and sorry for the trouble

PR #1526, opened 2021-07-07
[CMake] Highlighting builtin functions and arguments/keywords for builtin functions

uyha:
Since CMake's syntax doesn't have a way differentiate arguments as a marker (e.g `STATUS`, `WARNING`, etc. in `message`, and `AND`,`OR`, etc. in `if`) and actually arguments, I think one way to make the highlight more useful is to capture CMake's builtin functions as `@function.buitlin` and their predefined mark arguments as `@keyword`.


queries/cmake/highlights.scm
```diff
@@ -19,13 +19,30 @@
...
...st|ctest_update|ctest_upload|define_property|enable_language|enable_testing|exec_program|execute_process|export|export_library_dependencies|file|find_file|find_library|find_package|find_path|find_program|fltk_wrap_ui|get_cmake_property|get_directory_property|get_filename_component|get_property|get_source_file_property|get_target_property|get_test_property|include|include_directories|include_external_msproject|include_guard|include_regular_expression|install|install_files|install_programs|install_targets|link_directories|link_libraries|list|load_cache|load_command|make_directory|mark_as_advanced|math|message|option|output_required_files|project|qt_wrap_cpp|qt_wrap_ui|remove|remove_definitions|separate_arguments|set|set_directory_properties|set_property|set_source_files_properties|set_target_properties|set_tests_properties|site_name|source_group|string|subdir_depends|subdirs|target_compile_definitions|target_compile_features|target_compile_options|target_include_directories|target_link_directories|target_link_libraries|target_link_options|target_precompile_headers|target_sources|try_compile|try_run|unset|use_mangled_mesa|utility_source|variable_requires|variable_watch|write_file)$"))
```
theHamsta, line 45: wouldn't `#any-of?` be better suited?

uyha: I tried `#any-of` but couldn't come up with a way to make it case insensitive.

PR #1543, opened 2021-07-09
[elixir] basic folds

mhanberg:
This adds support for folding in elixir files.

This adds basic folding to any do blocks and anonymous functions. I think we can probably add folding to lists and maps, but I couldn't get immediately work, so I figured I could add the bare basic support for and figure out the rest later.

cc @nifoc

### DIFF
--- a/queries/elixir/folds.scm
+++ b/queries/elixir/folds.scm
@@ -1,0 +1,9 @@
+[
+  (do_block)
+  (anonymous_function)
+  (map)
+  (struct)
+  (list)
+  (keyword_list)
+  (heredoc)
+] @fold


theHamsta (approved):

nifoc:
> This adds basic folding to any do blocks and anonymous functions.

Does this really fold `do ... end` blocks? The parser generates a `(do_block)` for these. `(block)` is apparently only used for anonymous functions.

I only quickly tested this in TS Playground, since I don't use folds myself.

mhanberg:
> > This adds basic folding to any do blocks and anonymous functions.
> 
> Does this really fold `do ... end` blocks? The parser generates a `(do_block)` for these. `(block)` is apparently only used for anonymous functions.
> 
> I only quickly tested this in TS Playground, since I don't use folds myself.

Hmm, let me check it again. I had something working but I may have goofed it up.

mhanberg:
Yep, you were right. Turns out when i added this query to my local, i hadn't created the symlink to my actual home dir, so it somehow was just working with basic blocks OOTB.

But i was able to add the other things like map and list and it all seems to work now

nifoc:
LGTM now :)

stsewd (approved):

Issue #1520, opened 2021-07-07
Question: Is it possible to embed languages, that are handling by traditional highlighting? (Vue/Pug)


zealot128:
Over at https://github.com/ikatyang/tree-sitter-vue/issues/2 we discussed on how to handle Vue embedded languages. Scss, Typescript are already working. 
There is another common embedded format, the Pug templating language (Former Jade, also similar to HAML/Slim in Ruby land).

By adding an appropriate query to queries/vue/injections.scm:

```
(template_element
	(start_tag
    	(attribute
    		((attribute_name) @lang_key (#match? @lang_key "lang"))
        	((quoted_attribute_value) @_lang (#match? @_lang "pug"))
        )
    )
    ((text) @pug)
)
```

It seems to be possible to select the correct area. I cross checked it, by using another available language (typescript/javascript) in the lang attribute, and it should work in general.

But, treesitter has no language definition for Pug yet. So my question: Is it possible to "fall back" to the Vim native syntax highlighting? 
In the other issue, one user also had reported, that by changing ``setf pug`` in the buffer, the visual selection then uses the vim highlighting.

I would be glad for any pointers, of information if it is possible! Thanks!

theHamsta:
It's not possible to mix tree-sitter and old regex highlighting at the moment. It is however not to difficult to create a parser. Apparently somebody already started https://github.com/RianFuro/tree-sitter-pug

zealot128:
@theHamsta Thank you for the quick answer. This looks promising! I will try to compile + install it manually later, as it is not available via TSInstall yet.

theHamsta:
Hi, it would be possible to use tree-sitter and vim regex as long as the vim regex will find the language region by itself. Then,  you could just set `:set syntax=on` to enable regex highlighting.

zealot128:
I've made some progress and got the parser installed.

- I've just checked out the repo, ran ``tree-sitter generate`` which only changed minor things in parser.h
- Then, I've loaded the script:

```lua
        local parser_config = require("nvim-treesitter.parsers").get_parser_configs()
    parser_config.pug = {
      install_info = {
        url = "~/OpenSource/tree-sitter-pug", -- local path or git repo
        files = {"src/parser.c", "src/scanner.cc" }
      },
      filetype = "pug", -- if filetype does not agrees with parser name
      used_by = {"vue"} -- additional filetypes that use this parser
    }
```

[...]

PR #1549, opened 2021-07-10
Feat: Added Pug parser with inital queries

zealot128:
I discussed my progress in #1520 

This is a WIP with the Pug tree-parser. If worked OK for my type of files. I am sure it does not support all of the syntax, yet, but it would be much better than a white screen.

Bildschirmfoto 2021-07-10 um 17 18 19

As you can see, the highlighting is quite ok for raw "example.pug" files.

Unfortunately, if does not work with Vue template files, yet. I've added the ``ftdetect`` rule, but it is unable to map the ``@pug`` with the pug file

```
<template lang="pug">
body: div 
  | this is unhighlighted
</template>
```

Is unhighlighted.... For test, I've changed the ``vue/injections.scm`` template rule from ``@pug`` to @typescript or @scss, and the highlighting works.... What am I missing? Where can I tell nvim how to handle the ``@pug`` marked parts as a pug file.

I've also tried using the @content + @language way described in the README without any success:

```
(template_element
 (start_tag
  (attribute
   ((attribute_name) @lang_key (#match? @lang_key "lang"))
   ((quoted_attribute_value) @language)
  )
 )
 ((text) @content)
 )
```

But it seems to be a problem with the vue-parer, which already processes the content of the script tag as html... If anybody has any ideas, I'll be glad!

### DIFF
--- a/ftdetect/pug.vim
+++ b/ftdetect/pug.vim
@@ -1,0 +1,1 @@
+au BufRead,BufNewFile *.pug		setlocal filetype=pug
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -470,6 +470,15 @@ list.glimmer = {
   used_by = { "html.handlebars" },
 }
 
+list.pug = {
+  install_info = {
+    url = "https://github.com/zealot128/tree-sitter-pug",
+    files = { "src/parser.c", "src/scanner.cc" },
+  },
+  maintainers = { "@zealot128" },
+  filetype = "pug",
+}
+
 list.vue = {
   install_info = {
     url = "https://github.com/ikatyang/tree-sitter-vue",
--- a/queries/pug/highlights.scm
+++ b/queries/pug/highlights.scm
@@ -1,0 +1,27 @@
+(comment) @comment
+(tag_name) @tag
+(
+  (tag_name) @constant.builtin
+  ; https://www.script-example.com/html-tag-liste
+  (#any-of? @constant.builtin
+   "head" "title" "base" "link" "meta" "style"
+   "body" "article" "section" "nav" "aside" "h1" "h2" "h3" "h4" "h5" "h6" "hgroup" "header" "footer" "address"
+   "p" "hr" "pre" "blockquote" "ol" "ul" "menu" "li" "dl" "dt" "dd" "figure" "figcaption" "main" "div"
+   "a" "em" "strong" "small" "s" "cite" "q" "dfn" "abbr" "ruby" "rt" "rp" "data" "time" "code" "var" "samp" "kbd" "sub" "sup" "i" "b" "u" "mark" "bdi" "bdo" "span" "br" "wbr"
+   "ins" "del"
+   "picture" "source" "img" "iframe" "embed" "object" "param" "video" "audio" "track" "map" "area"
+   "table" "caption" "colgroup" "col" "tbody" "thead" "tfoot" "tr" "td" "th "
+   "form" "label" "input" "button" "select" "datalist" "optgroup" "option" "textarea" "output" "progress" "meter" "fieldset" "legend"
+   "details" "summary" "dialog"
+   "script" "noscript" "template" "slot" "canvas")
+)
+(content) @none
+(quoted_attribute_value) @string
+(id) @constant
+(class) @constant
+(attribute_name) @symbol
+(
+  (attribute_name )  @keyword
+  (#match? @keyword "^(:|v-bind|v-|\\@)")
+) @keyword
+
--- a/queries/pug/highlights.scm
+++ b/queries/pug/highlights.scm
@@ -25,3 +25,6 @@
   (#match? @keyword "^(:|v-bind|v-|\\@)")
 ) @keyword
 
+[
+  ":"
+] @punctuation.delimiter
--- a/README.md
+++ b/README.md
@@ -206,6 +206,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [php](https://github.com/tree-sitter/tree-sitter-php) (maintained by @tk-shirasaka)
 - [x] [pioasm](https://github.com/leo60228/tree-sitter-pioasm) (maintained by @leo60228)
 - [x] [prisma](https://github.com/victorhqc/tree-sitter-prisma) (maintained by @elianiva)
+- [x] [pug](https://github.com/zealot128/tree-sitter-pug) (maintained by @zealot128)
 - [x] [python](https://github.com/tree-sitter/tree-sitter-python) (maintained by @stsewd, @theHamsta)
 - [x] [ql](https://github.com/tree-sitter/tree-sitter-ql) (maintained by @pwntester)
 - [x] [Tree-sitter query language](https://github.com/nvim-treesitter/tree-sitter-query) (maintained by @steelsojka)
--- a/queries/pug/injections.scm
+++ b/queries/pug/injections.scm
@@ -1,0 +1,7 @@
+(javascript) @javascript
+
+(
+   (attribute_name) @_attribute_name
+   (quoted_attribute_value (attribute_value )  @javascript)
+   (#match? @_attribute_name "^(:|v-bind|v-|\\@)")
+)
--- a/queries/vue/injections.scm
+++ b/queries/vue/injections.scm
@@ -25,4 +25,13 @@
     (quoted_attribute_value
       (attribute_value) @javascript)))
 
+(
+  (template_element
+    (start_tag
+      (attribute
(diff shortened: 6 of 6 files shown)


zealot128:
Finally, I've made it work with embedded pug inside vue.


I have still a PR on the tree-sitter-vue project. When that is merged, the pug highlighting within vue will also work!
https://github.com/ikatyang/tree-sitter-vue/pull/13

Bildschirmfoto 2021-07-12 um 00 06 44

lua/nvim-treesitter/parsers.lua
```diff
@@ -391,6 +391,16 @@ list.glimmer = {
...
 }
 
+list.pug = {
+  install_info = {
+    url = "https://github.com/zealot128/tree-sitter-pug",
+    files = { "src/parser.c", "src/scanner.cc" },
+  },
+  maintainers = { "@zealot128" },
+  filetype = "pug",
+  used_by = {"vue"}
```
theHamsta, line 401: Why `used_by` vue? This isn't necessary for language injection and would only have an effect if a vue parser wouldn't exist and the filetype should use the pug parser.

theHamsta (changes requested):
Appart from the `used_by` this looks good

ftdetect/pug.vim
```diff
@@ -0,0 +1 @@
+au BufRead,BufNewFile *.pug		set filetype=pug
```
vigoux, line 1: I'd rather do
```suggestion
au BufRead,BufNewFile *.pug		setlocal filetype=pug
```

queries/pug/highlights.scm
```diff
@@ -0,0 +1,28 @@
...
+   "body" "article" "section" "nav" "aside" "h1" "h2" "h3" "h4" "h5" "h6" "hgroup" "header" "footer" "address"
+   "p" "hr" "pre" "blockquote" "ol" "ul" "menu" "li" "dl" "dt" "dd" "figure" "figcaption" "main" "div"
+   "a" "em" "strong" "small" "s" "cite" "q" "dfn" "abbr" "ruby" "rt" "rp" "data" "time" "code" "var" "samp" "kbd" "sub" "sup" "i" "b" "u" "mark" "bdi" "bdo" "span" "br" "wbr"
+   "ins" "del"
+   "picture" "source" "img" "iframe" "embed" "object" "param" "video" "audio" "track" "map" "area"
+   "table" "caption" "colgroup" "col" "tbody" "thead" "tfoot" "tr" "td" "th "
+   "form" "label" "input" "button" "select" "datalist" "optgroup" "option" "textarea" "output" "progress" "meter" "fieldset" "legend"
+   "details" "summary" "dialog"
+   "script" "noscript" "template" "slot" "canvas")
+) @constant.builtin
```
vigoux, line 17: You can directly do `constant.builtin` here rather than the `tag_name` proxy.

queries/pug/highlights.scm
```diff
@@ -0,0 +1,28 @@
...
+(content) @none
+(quoted_attribute_value) @string
+(id) @constant
+(class) @constant
+; (attribute_name) @variable.builtin
+(attribute_name) @symbol
```
vigoux, line 23: Duplicate?

queries/pug/highlights.scm
```diff
@@ -0,0 +1,28 @@
...
+(
+  (attribute_name )  @_attribute_name
+  (#match? @_attribute_name "^(:|v-bind|v-|\\@)")
+) @keyword
+
```
vigoux, line 28: Same here than above.

ftdetect/pug.vim
zealot128: Fine by me, I've just peeked into the existing ftdetect rules [ftdetect/gomod.vim](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/ftdetect/gomod.vim) and [ftdetect/gmd.vim](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/ftdetect/gdscript.vim) are doing the same with ``set`` instead of ``setlocal``. Maybe they should be changed too?

queries/pug/highlights.scm
zealot128: My intention was, to make highlighting for "static" attributes different than those with Javascript interpolation.

HTML Attributes such as "v-if=" , "@click=", ":src=" are interpreted as Javascript within Vue template. That's why I used the second, more specific rule, to override the highlighting for those attribute. 
See here the "target" attribute is plain text as v-if & src are bound to Javascript values.

Bildschirmfoto 2021-07-13 um 18 56 01

queries/pug/highlights.scm
zealot128: Great! committed and tested.

ftdetect/pug.vim
zealot128: changed it to setlocal, which should make much more sense for most of these autocomands. Like i said, maybe worth changing the other 2 ftdetect scripts the same way.

lua/nvim-treesitter/parsers.lua
zealot128: removed it! as my prime goal was to get the pug highlighting within vue working I tried all avenues I could find :) In the end it is necessary to change the parser grammar within tree-sitter-vue to make it work, which I also submitted a [PR](https://github.com/ikatyang/tree-sitter-vue/pull/13). 
I removed the used_by line.

queries/pug/highlights.scm
zealot128: You mean the proxy indirection? I've moved the @keyword into the parens.

ftdetect/pug.vim
vigoux: I agree on changing those we'll do it in another pr!

theHamsta (approved):

zealot128:
About the Action failures...

I also got the same error ``./scripts/check-queries.lua:64: .../usr/share/nvim/runtime/lua/vim/treesitter/query.lua:161: query: error at position 737`` when opening any *.scm files on my local machine **after** I rebased onto origin/master.

The error is still present on my local machine, even if I check out origin/master without my changes, so I guess the error is something else. 

Only when I go BEFORE 2a0097828d35a170e43616777c06611d881324c9 then the error disappears, so I guess that commit seemed to have introduced the query problem which seems logical, as the query highlighting was touched.

theHamsta:
The error is in the vue injections. Maybe the vue parser was updated?

How can the pug parser be installed without `requires_generate_from_source` when the repo does not provide `parser.c`?

zealot128:
> The error is in the vue injections. Maybe the vue parser was updated?

The last update from tree-sitter-vue on master branch was from April, so no recent changes there. 
If I check out nvim-tressitter@master and revert the ``queries/query/highlights.scm`` back to c37e79803e21abfae960174a6c661da166c87e8b, than there is no error. So 2a0097828d35a170e43616777c06611d881324c9 introduced the error, which I receive the same when viewing any query file at all:

Bildschirmfoto 2021-07-14 um 08 55 10


> 
> How can the pug parser be installed without `requires_generate_from_source` when the repo does not provide `parser.c`?

My fork of the pug parser now includes all the bindings/*, parser.c, ... files to be able to install directly without having to run 'tree-sitter generate' before.
TSInstall pug as well as TSInstallFromGrammar pug worked on my machine.

theHamsta:
@zealot128  It seems that it has to be `text` instead of `rawtext`
```
(
  (template_element
    (start_tag
      (attribute
        (quoted_attribute_value (attribute_value) @_lang)))
    (text) @pug)
  (#match? @_lang "pug")
)
```

zealot128:
> @zealot128 It seems that it has to be `text` instead of `rawtext`
> 
> ```
> (
>   (template_element
>     (start_tag
>       (attribute
>         (quoted_attribute_value (attribute_value) @_lang)))
>     (text) @pug)
>   (#match? @_lang "pug")
> )
> ```

@theHamsta i thought so the same, but it didn't work. TreeSitterVue always parses the content of "<template>" as html so Pug would never have a Chance to handle the unmodified output via injection (VueTreeSitter does the same already for the <script> and <style> tags and provides the unmodified raw_text as target for injection)..

I opened a PR on the Vue repo which adds the "raw_text" node for the "template" tag when used in conjunction with an attribute.

https://github.com/ikatyang/tree-sitter-vue/pull/13

Is this the the reason the checks are failing here?

theHamsta:
yes, tree-sitter can check whether `raw_text` is actually allowed at that position.

zealot128:
Oh great, that means either wait until the other PR is merged, or I modify this PR first, that it does not touch Vue integration for now.

vigoux (approved):

theHamsta:
@zealot128 your injection query should now be fixed

PR #1562, opened 2021-07-13
feat: add HCL file type detection

khuedoan:
HCL parser was added in https://github.com/nvim-treesitter/nvim-treesitter/pull/1445, but neovim does not set the correct filetype for `*.hcl`, `*.tf` and `*.tfvars` by default:

image

This PR adds a `ftdetect` rule for those files:

image

### DIFF
--- a/ftdetect/hcl.vim
+++ b/ftdetect/hcl.vim
@@ -1,0 +1,1 @@
+autocmd BufRead,BufNewFile *.hcl,*.tf,*.tfvars set filetype=hcl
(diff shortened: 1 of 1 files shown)


theHamsta (approved):

gegoune:
Doesn't vim/nvim already set some filetype to *.tf files?

khuedoan:
Look like the default is TinyFugue, I just notice the [vim-terraform plugin has an `ftdetect` config](https://github.com/hashivim/vim-terraform/blob/master/ftdetect/hcl.vim), what is the correct way to introduce that file to this project? The license is ISC License (functionally equivalent to the BSD 2-Clause and MIT licenses)

gegoune:
Id be happy to have it set here. Could remove vim-terraform. 😀

theHamsta:
@khuedoan best solution is always to ask the authors

vigoux:
@Mergify refresh

PR #1566, opened 2021-07-13
Folds: remove ending marks for folds

ecnerwala:
These cause vim to become more confused when there are multiple folds ending on the same line. For instance, consider:
```
int main() {         // >2
    if (1) {         // >3
        if (0) {     // >4
    }}               // <4
}                    // <2
```
For some reason, including the ending mark on line 5 causes the level-3 fold to end on line 5 instead of line 4 (???). This should probably get fixed upstream, but this is a fine workaround for now.


vigoux (approved):
I am fine with this workaround, and the comment makes it clear.

PR #1581, opened 2021-07-18
Handle documented offset! predicate.

JoZie:
Nvim documents a *ts-predicate-offset!* predicate which is not handled by this plugin.

My `offset!` use-case is the folding of C++ code like:
``` c++
// original
int main ( int argc,
           char** argv ) {
    // some long implementation
}

// default folding
//int main ( int argc, ................

// folding with
// ((compound_statement) @fold (#offset! @fold 1 0 -1 0))
int main ( int argc,
           char** argv ) {
// ..................
)
```

### DIFF
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -156,21 +156,25 @@ function M.iter_prepared_matches(query, qnode, bufnr, start_row, end_row)
   local matches = query:iter_matches(qnode, bufnr, start_row, end_row)
 
   local function iterator()
-    local pattern, match = matches()
+    local pattern, match, metadata = matches()
     if pattern ~= nil then
       local prepared_match = {}
+      local preds = query.info.patterns[pattern]
 
       -- Extract capture names from each match
       for id, node in pairs(match) do
         local name = query.captures[id] -- name of the capture in the query
         if name ~= nil then
           local path = split(name .. ".node")
-          insert_to_path(prepared_match, path, node)
+          if preds and preds[1][1] == "offset!" and preds[1][2] == id and metadata.content then
+            insert_to_path(prepared_match, path, tsrange.TSRange.from_table(bufnr, metadata.content[1]))
+          else
+            insert_to_path(prepared_match, path, node)
+          end
         end
       end
 
       -- Add some predicates for testing
-      local preds = query.info.patterns[pattern]
       if preds then
         for _, pred in pairs(preds) do
           -- functions


theHamsta:
`iter_prepared_matches` is astonishingly slow. It would be better to update fold to use the built in iter_matches directly. `offset!` is at the moment hard-coded in core to only work with injections. It's not an ideal situation. 

For folds, I wanted to implement `@fold.outer`/`@fold.inner` as suggested by @stsewd to enable folding of the whole definition vs folding of the definition body https://github.com/nvim-treesitter/nvim-treesitter/pull/1451

theHamsta (changes requested):
But I think this is a acceptable intermediate solution. Probably it will also stay forever :smile: 

But `content` will now overwrite *all* nodes in the query not only the one it was originally targeted at.

It would probably best when `offset!` would directly mutate the tree-sitter node in core or maybe replace it with something equivalent to our `TSRange` or at least sets the updated node range per capture.

theHamsta:
@steelsojka can't we generalize `offset!` a bit better?

JoZie:
I modified the if stmt slightly. the offset should now only apply to the node speicified

theHamsta:
Ok. But wouldn't it still be better to just make `offset!` less specialized upstream?

JoZie:
> Ok. But wouldn't it still be better to just make offset! less specialized upstream?
Don't know, I'm way out of my comfort zone here. xD
I just want to tweak the range of a node make the folding work as I want. 

The optimal solution would be able to start folding form a line and column, but afaik it's just line based in vim.
That's the whole reason why i want to reduce the size of a compound_statement node (everything inside '{}') by one line. 
To me your `fold.inner` sounds just like an alias to `((compound_statement) @fold (#offset! @fold 1 0 -1 0))` at least fort C/C++.
I'm not sure what the node-type of an python body is

theHamsta:
>To me your fold.inner sounds just like an alias to ((compound_statement) @fold (#offset! @fold 1 0 -1 0))

 The purpose would be to make it configurable so you would name the new behavior `((compound_statement) @fold.inner (#offset! @fold.inner 1 0 -1 0))` while the old can still be the same.

theHamsta:
@JoZie just a heads up: your PR isn't the problem. It's more the unfinished stuff that we left in `offset!` when we can't get a good solution in short time your PR would be totally fine to merge (as long as we remember that it has to be changed here as well when `offset!` is fixed.

Would you be interested in working in a config option to fold definition bodies instead of whole definition? The inside of `{` ... `}` is needed so often that we could even have a separate predicate for it. But the preferred way would be to leverage metadata instead of collecting into tables from which only a part is used. It really has a *huge* performance impact as we saw when looking for definition with tree-sitter-

JoZie:
> Would you be interested in working in a config option to fold definition bodies instead of whole definition? The inside of `{` ... `}` is needed so often that we could even have a separate predicate for it. But the preferred way would be to leverage metadata or direct mutation in upstream instead of collecting into tables from which only a part is used. It really has a _huge_ performance 

Yes but I'd need guidance for that (where to start, what approx to do, etc.)

PR #1589, opened 2021-07-18
Mergify: tweaks around automatic merging

vigoux:


.github/mergify.yml
```diff
@@ -12,8 +12,8 @@ pull_request_rules:
     conditions:
       - and:
         - "-draft"
-        - "#check-failure=0"
```
theHamsta, line 15: this is not needed anymore?

vigoux: Nope, because of our branch protection rules

stsewd (approved):

PR #1591, opened 2021-07-18
feat(mergify): delay mergify merging by one day

vigoux:
This will avoid merging too rapidly.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -8,11 +8,29 @@ pull_request_rules:
         message: Automatically approving lockfile updates
       merge:
         method: merge
-  - name: Merge on approval
+
+  - name: Prepare for merge
     conditions:
       - and:
         - "-draft"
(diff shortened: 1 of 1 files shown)


vigoux:
Actually delay by 2 hours, to avoid blocking things unnecessarily.

vigoux:
@theHamsta what do you think about that ?

theHamsta (commented):
I guess with two approvals we can merge immediately. With only one approval I would have a longer delay. I'm fine both strategies

vigoux:
How much delay do you want?

vigoux:
So the new rule for auto-merge is the the following (with those two, add the classic "not a draft and no pending reviews").
- If there is two approvals, merge immediately
- If there is only one approval, comment now that the PR will be merged soon, and merge one day after that

theHamsta (approved):

PR #1597, opened 2021-07-19
Updated TSInstall example command.

duarteocarmo:
When getting set up, the current instructions make me think I have to write `:TSInstall {'python'}`, when the correct command is `TSInstall python`

Let me know if this makes sense 😄


theHamsta:
Maybe `<language>` would be clearer? The braces tried to say that you have to replace with your acutal language. `:TSInstall language_to_instrall`

duarteocarmo:
Just updated. I know! Maybe I’m just dumb. The best thing here even imo would be to give a literal example I think

theHamsta:
I understand that it's ambiguous

theHamsta (approved):

PR #1606, opened 2021-07-20
fix: make additional_vim_regex_highlighting actually accept a list

Conni2461:
I was debugging a telescope buffer previewer yesterday and stumped upon this "inconsistency".
The docs are talking that this could be a list of languages. https://github.com/nvim-treesitter/nvim-treesitter#modules
But you were treating it like a set.

One of both needs to be adjusted in my opinion, its your call which one :)


theHamsta (approved):

theHamsta:
Thanks!

vigoux (approved):

PR #1610, opened 2021-07-21
highlights(lua): Fix method calls

savq:
The usual `function_call` query would highlight the objects at the
beginning of a method call. The `method` query has to account for this,
and highlight the identifier as a variable again.

### DIFF
--- a/queries/lua/highlights.scm
+++ b/queries/lua/highlights.scm
@@ -132,9 +132,12 @@
 
 (property_identifier) @property
 
-(function_call (identifier) @function . (arguments))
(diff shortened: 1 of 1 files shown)


vigoux:
I think we should rather make the highlight for function calls more specific, maybe using an anchor at the start of the capture?

savq:
I didn't get anything useful when using anchors at the start or between `identifier` and `method`.
Placing all `function_call` queries in an alternation does seem to fix the need for overlapping queries.

vigoux (approved):

PR #1611, opened 2021-07-21
feat: contains

ThePrimeagen:
A couple of helpers to use with node comparison.  This is particularly
useful when using selections and node querying or scope determination.

### DIFF
--- a/lua/nvim-treesitter/ts_utils.lua
+++ b/lua/nvim-treesitter/ts_utils.lua
@@ -5,6 +5,57 @@ local utils = require "nvim-treesitter.utils"
 
 local M = {}
 
+-- determines if a contains node b.
+-- @param a the containing node
+-- @param b the node to be contained
+function M.node_contains(a, b)
+  if a == nil or b == nil then
+    return false
+  end
+
+  local start_row, start_col, end_row, end_col = b:range()
+  return M.is_in_node_range(a, start_row, start_col) and
+    M.is_in_node_range(a, end_row, end_col)
+end
+
+-- determines if a node exists within a range.  Imagine a range selection
+-- across '<,'> and an identifier.  Does the identifier exist within the
+-- selection?
+--
+-- @param a the containing node
+-- @param b the node to be contained
+M.range_contains_node = function(node, start_row, start_col, end_row, end_col)
+  local node_start_row, node_start_col, node_end_row, node_end_col = node:range()
+
+  -- There are five possible conditions
+  -- 1. node start/end row are contained exclusively within the range.
+  -- 2. The range is a single line range
+  --   - the node start/end row must equal start_row and cols have to exist
+  --     within range, inclusive
+  -- 3. The node exists solely within the first line
+  --   - node_start_col has to be inclusive with start_col, end col doesn't
+  --     matter.
+  -- 4. The node exists solely within the last line
+  --   - node_start_col doesn't matter whereas node_end_col has to be
+  --     inclusive with end_col
+  -- 5. The node starts / ends on the same rows and has to have each column
+  --    considered
+  if start_row < node_start_row and end_row > node_end_row then
+    return true
+  elseif start_row == end_row then
+    return node_start_row == start_row and node_end_row == end_row and
+    start_col <= node_start_col and end_col >= node_end_col
+  elseif start_row == node_start_row and start_row == node_end_row then
+    return start_col <= node_start_col
+  elseif end_row == node_start_row and end_row == node_end_row then
+    return end_col >= node_end_col
+  elseif start_row <= node_start_row and end_row >= node_end_row then
+    return start_col <= node_start_col and end_col >= node_end_col
+  end
+
+  return false
+end
+
 --- Gets the actual text content of a node
 -- @param node the node to get the text from
 -- @param bufnr the buffer containing the node
--- a/lua/nvim-treesitter/ts_utils.lua
+++ b/lua/nvim-treesitter/ts_utils.lua
@@ -14,42 +14,54 @@ function M.node_contains(a, b)
   end
 
   local start_row, start_col, end_row, end_col = b:range()
-  return M.is_in_node_range(a, start_row, start_col) and
-    M.is_in_node_range(a, end_row, end_col)
+  return M.is_in_node_range(a, start_row, start_col) and M.is_in_node_range(a, end_row, end_col)
 end
 
 -- determines if a node exists within a range.  Imagine a range selection
 -- across '<,'> and an identifier.  Does the identifier exist within the
 -- selection?
 --
--- @param a the containing node
--- @param b the node to be contained
+-- @param node the node to test its inclusion with the vim rarge
+-- @param start_row
+-- @param start_col
+-- @param end_row
+-- @param end_col
 M.range_contains_node = function(node, start_row, start_col, end_row, end_col)
   local node_start_row, node_start_col, node_end_row, node_end_col = node:range()
 
   -- There are five possible conditions
   -- 1. node start/end row are contained exclusively within the range.
+  if start_row < node_start_row and end_row > node_end_row then
+    return true
+  end
+
   -- 2. The range is a single line range
   --   - the node start/end row must equal start_row and cols have to exist
   --     within range, inclusive
+  if start_row == end_row then
+    return node_start_row == start_row
+      and node_end_row == end_row
+      and start_col <= node_start_col
+      and end_col >= node_end_col
+  end
+
   -- 3. The node exists solely within the first line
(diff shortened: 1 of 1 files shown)


ThePrimeagen:
I just realized there's a small bug.  I'll fix that shortly. By the way is this something that would fit within this repo?

theHamsta:
>  By the way is this something that would fit within this repo?

Yeah, absolutely. Maybe we have to align the naming with the other functions and either overload `in_node_range` (to make it work with pos,range or node) or clarify through naming that the one function is for a position, the other for a range, and another for a node `node_contains_pos`,  `node_contains_range`, `node_contains_node`.

ThePrimeagen:
If anyone can tell me what to name it, i'll rename it!  Also, I need to run the linting.  Let me know what to rename it to and I'll do that.

Also should end / start be inclusive?

ThePrimeagen:
@theHamsta I also had to update the range match algo. it turns out I wasn't necessarily thinking about it complete enough.  The algo is now "correct" as far as I can tell

ThePrimeagen:
So as I do more research I think that this may be extra work that isn't needed.

```lua
-- I need to experiment with this
LanguageTree:contains({self}, {range})               *LanguageTree:contains()*
                Determines wether This goes down the tree to recursively check childs.

                Parameters: ~
                    {range}  is contained in this language tree

                Parameters: ~
                    {range}  A range, that is a `{ start_line, start_col,
                             end_line, end_col }` table.
                    {self}
```

theHamsta:
I don't think your functions are related with `LanguageTree:contains` just checks the sub-ranges of a language.

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -5,6 +5,57 @@ local utils = require "nvim-treesitter.utils"
...
+-- @param b the node to be contained
+function M.node_contains(a, b)
+  if a == nil or b == nil then
+    return false
+  end
+
+  local start_row, start_col, end_row, end_col = b:range()
+  return M.is_in_node_range(a, start_row, start_col) and
+    M.is_in_node_range(a, end_row, end_col)
+end
```
theHamsta, line 19: I don't think this is correct since the end position is exclusive Otherwise this function would also be more elegant for the other one. Maybe have a helper function to obtain the char right before `end_row`,`end_col`? 

 It should also be possible to have only one implementation for those two functions. They are two function with completely different names doing the exact same thing. `M,get_node_range` coerces nodes and ranges to ranges.

theHamsta: But, wait in the second function you're using vim ranges. Then, the two functions are actually not the same and the one with start_row/col end_row/col is actually inclusive while the node one is not

theHamsta (changes requested):
So I think this is ok to merge after you checked whether the node-in-node function handles the last char correct (the function should be symmetrical between a and b).

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -5,6 +5,57 @@ local utils = require "nvim-treesitter.utils"
...
+
+-- determines if a node exists within a range.  Imagine a range selection
+-- across '<,'> and an identifier.  Does the identifier exist within the
+-- selection?
+--
+-- @param a the containing node
+-- @param b the node to be contained
+M.range_contains_node = function(node, start_row, start_col, end_row, end_col)
```
theHamsta, line 27: The documentation is not up-to-date

ThePrimeagen:
@theHamsta Just to be clear

1.  Update documentation
2.  What do you mean symmetrical for node_contains?  I assume this means the following cases. 
  a.  If a == b, it returns true for both
  b.  If a contains b is true, b contains a will return false     

Also, does nvim-treesitter have plenary tests?  If you want to avoid the dependency, there could always be a separate repo (though that would be super sucky)

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -5,6 +5,57 @@ local utils = require "nvim-treesitter.utils"
...
+    return true
+  elseif start_row == end_row then
+    return node_start_row == start_row and node_end_row == end_row and
+    start_col <= node_start_col and end_col >= node_end_col
+  elseif start_row == node_start_row and start_row == node_end_row then
+    return start_col <= node_start_col
+  elseif end_row == node_start_row and end_row == node_end_row then
+    return end_col >= node_end_col
+  elseif start_row <= node_start_row and end_row >= node_end_row then
+    return start_col <= node_start_col and end_col >= node_end_col
```
kyazdani42, line 53: just a hint here, you don't need to elseif after a return, a simple `if` works (and is usually easier to read :) )

theHamsta:
1)  Update documentation
2) Make sure all the functions you wrote are correct
  - I guess `node_contains` is not correct. The end point of a node is not in its own range
3) Make sure code you add doesn't add confusion
   - `is_in_node_range` sounds like it could fit for positions and nodes. Or you make it explicit :
             - `node_contains_pos`/`node_contains_node`/`visual_selection_contains_node`, or 
             - `pos_is_in_node_range`/`node_is_in_node_range`/`node_is_in_visual_selection`
             -  Maybe `node_is_in_visual_selection` could live in the `incremental_selection` module to reduce conflicts with the similarly named functions in this module
   - Vim ranges vs tree-sitter ranges always add confusion. Maybe the term `visual_selection` instead of `range` would prevent to accidentally use `node:range()` with this function
4) Make sure that your PR does not add code duplication.
   - `is_in_node_range` and `range_contains_node` do the same dangerous multicase check but slightly differently since `range_contains_node` uses Vim ranges. Make sure it is used in the visual_section module when this is the use case. Because there a similar complicated check is performed. It was incorrect in multiple ways in the beginning which only became clear in edge cases
5) nvim-treesitter has plenary (see scripts/run_tests.sh). At the moment it's not in CI because indent tests are failing. So you would make sure that only the new test you would add get run on CI or you only run the test locally

There are multiple ways how this can be solved I sure you'll find a good solution. Please understand that I'm so pedantic about this row/col/range topic but we had a lot of pain with the Vim vs treesitter ranges (with zero vs one-based with inclusive vs exclusive).

lua/nvim-treesitter/ts_utils.lua
ThePrimeagen: ah yes.  The original version of this had some elseif checks that would fall through to the bottom.  I think that is an astute point that as this has changed I no longer need the `elseif`.  And honestly, I would prefer not having them.

ThePrimeagen:
> Please understand that I'm so pedantic about this row/col/range topic but we had a lot of pain with the Vim vs treesitter ranges (with zero vs one-based with inclusive vs exclusive).

Trust me, you are not being pedantic, you are keeping things on rails.  1/2 of my time spent on playing around with tree sitter is spent thinking about the type of range / inclusive vs exclusive.  Its a real pain and I appreciate your thoroughness.  My goal isn't to "get a sweet commit and t shirt" but to make an operation I am doing 50 times all over in an upstream lib where many people can benefit from this.

> There are multiple ways how this can be solved I sure you'll find a good solution. 
You clearly think to high of my skallz

I'll address your points likely today/tomorrow

theHamsta:
> Trust me, you are not being pedantic, you are keeping things on rails. 1/2 of my time spent on playing around with tree sitter is spent thinking about the type of range / inclusive vs exclusive. Its a real pain and I appreciate your thoroughness. My goal isn't to "get a sweet commit and t shirt" but to make an operation I am doing 50 times all over in an upstream lib where many people can benefit from this.

Something that helped me is to just always use tree-sitter ranges and positions `local range = {node:range()}` as a 4 tuple `range[1], range[2] ...` instead of col/row, col/row and have them the only kind of range that we have in this repo. They are consistent with LSP ranges and positions (zero-base, exclusive). It helps to convert the Vim values to tree-sitter ones before you do any logic on them.

A useful utility function could calculate `the char before that one` which is also often needed to get from exclusive to the last char in the range

theHamsta:
There is also a function `get_node_range` which coerces 4-tuple ranges and nodes into 4-tuple ranges.

ThePrimeagen:
Telescopic Johnson and I have been batting around the idea of a meta table that just does it all for you.

theHamsta:
Like TSRange in this repo that is node&range at the same time? Just bring it to upstream nvim then it's handy for everyone ;-)

ThePrimeagen:
hasn't fallen off radar, just got caught up :)

clason:
superseded by rewrite on `main` branch (which removes `ts_utils`; sufficiently common functionality should be added upstream, otherwise just implement this locally -- we could have a "common snippets" wiki page as a compromise)

PR #1626, opened 2021-07-24
highlights(clojure): fix highlighting of #_

theHamsta:
Fix https://github.com/sogaiu/tree-sitter-clojure/issues/14

### DIFF
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -13,7 +13,10 @@
 
 ;; >> Litterals
 
-(dis_expr) @comment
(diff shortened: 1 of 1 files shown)


stsewd (approved):
This makes me think we may want priorities to be relative to the tree-sitter base (100), and make that base a setting in the highlight module.

theHamsta:
I was also thinking about relative priorities but it is sometimes also valuable to refer directly to the priority of something else outside of tree-sitter. Would have to be changed in nvim core anyways.

Issue #1111, opened 2021-03-29
Context parsing for styled components in jsx/tsx


timsofteng:
**Is your feature request related to a problem? Please describe.**
Tresitter doesn't recognize `styled` comonents in .tsx files. I'm not sure how it could be implemented unfortunately. Take a look at screenshot.
![image](https://user-images.githubusercontent.com/51912173/112873324-272d3600-90ca-11eb-9ef6-7b8b7a15a07c.png)

**Describe the solution you'd like**
Treesitter shoud parse styled component and highlight its content as css or something like that.

vigoux:
That's done by language injection, so the "only" thing to do is to :
- determine what node correspond to the `styled` thing
- use that to enable a `css` injection in the following text

elianiva:
and someone needs to make a proper css-in-js parser, unless you're OK with this broken-ish highlighting because css property isn't supposed to go outside of the curly braces and it will get worse once you introduce a bit more complex JS expression. There's a discussion about this on Zulip iirc.

![image](https://user-images.githubusercontent.com/51877647/112913280-8a978200-9123-11eb-90f4-c97ec746da73.png)

steelsojka:
We should be able to just extend the css parser with a node for `${}` interpolation. Then we can use an injection directive to use the the parent trees language... that way this would work out of the box with TS or JS.

marcelarie:
Hey!! Any news on this? I will love to know how to do this manually if not. 👍🏽

timsofteng:
I'm tracking this as well.

steelsojka:
I don't think anyone is actively working on a parser that can support this syntax. The current architecture  can support the injections but we need a parser that can support css syntax with JS template strings.

UltiRequiem:
Bump

kyazdani42:
for people wanting this, just add a query in your neovim configuration folder under `after/queries/ecma/injections.scm` with that content:
```scheme
(call_expression
  (member_expression
    (identifier) @_styled
    (#eq? @_styled "styled"))
  (template_string) @css (#offset! @css 0 1 0 -1))
```
It will not be perfect but still you'll have some highlight

CallumHoward:
> It will not be perfect but still you'll have some highlight

This works great, thanks @kyazdani42! I notice it works well for 
```typescript
const StyledH1 = styled.h1`...
``` 

but not so much for 
```typescript
const StyledComponent = styled(Component)`...
```

[...]

PR #1628, opened 2021-07-25
Styled components injection

stevenocchipinti:
Hi,

This is my first attempt at a contribution, any feedback welcome.
I've been monitoring #1111 and decided to try it myself.

I have more detailed messages in the commits, including which cases I've attempted to support.
I'm sure there are plenty of cases missing but I'm hoping this will be a good start.

If there is an easier, more generic way to handle more cases, please let me know.

### DIFF
--- a/queries/ecma/injections.scm
+++ b/queries/ecma/injections.scm
@@ -18,6 +18,7 @@
  arguments: ((template_string) @glimmer
    (#offset! @glimmer 0 1 0 -1)))
 
+; styled.div`<css>`
 (call_expression
  function: (member_expression
    object: (identifier) @_name
@@ -25,10 +26,32 @@
  arguments: ((template_string) @css
    (#offset! @css 0 1 0 -1)))
 
+; styled(Component)`<css>`
 (call_expression
  function: (call_expression
    function: (identifier) @_name
      (#eq? @_name "styled"))
+ arguments: ((template_string) @css
+   (#offset! @css 0 1 0 -1)))
+
+; styled.div.attrs({ prop: "foo" })`<css>`
+(call_expression
+ function: (call_expression
+   function: (member_expression
+    object: (member_expression
+      object: (identifier) @_name
+        (#eq? @_name "styled"))))
+ arguments: ((template_string) @css
+   (#offset! @css 0 1 0 -1)))
+
+
+; styled(Component).attrs({ prop: "foo" })`<css>`
+(call_expression
+ function: (call_expression
+   function: (member_expression
+    object: (call_expression
+      function: (identifier) @_name
+        (#eq? @_name "styled"))))
  arguments: ((template_string) @css
    (#offset! @css 0 1 0 -1)))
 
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):
LGTM, waiting for @steelsojka review because i've been away for quite a long time :)

steelsojka:
I don't this will work. The problem is styled components are pure CSS. The CSS parser can't handle javascript string tempate literals. For example:

```css
.some-class {
  margin: ${someVar}
}
```

The css parser would fail at the variable. For this to be full supported we need a custom css parser that can handle the JS template variables.

steelsojka:
If this is for partial support without any variables then this would work just fine.

stevenocchipinti:
Yea, that also was mentioned in #1111 and I had no idea how to handle that 😅 
This is just for a few basic cases.

How would a custom parser work, would it be able to reuse the existing css parser but just add support for including `${}` and then reuse the existing javascript parser for inside the `${}`?

stevenocchipinti:
Also, I've realised I can put a [`queries/ecma` directory in my config](https://github.com/stevenocchipinti/dotvim/tree/master/queries/ecma), so feel free to close this if you'd rather not have an incomplete solution merged in.

theHamsta:
@stevenocchipinti you can easily create a grammar that extends CSS. Maybe a good resource is the JSONC parser that adds comments to JSON or the CUDA parser that adds CUDA's <<<...>>> call syntax to C++.

Find out in which grammatical position variable expansions can occur add replace the original rule with a `choice` between `original` and variable expansion

theHamsta:
@steelsojka @kyazdani42 should we merge this? Or will incomplete CSS be annoying? I suppose that partial support should be ok as the CSS should be able to recover from JS injections. You are the React experts...

kyazdani42:
i'm ok for partial support, would be better than having plain js strings anyway.

sQVe:
Would it be possible to revert this?

It seems like the injection breaks very open, at least when working with `emotion`/`styled-components`, and then applies the `Error` highlight `gui` styling. This hurts readability by a large margin.

@theHamsta

theHamsta:
@sQVe any reason why  you have the TSError highlighting enabled? I don't think it's useful in any  situation. Maybe one could quickly add a css-with-template-braces grammar

sQVe:
@theHamsta No specific reason from my part. I'm just using this plugin and `gruvbox.nvim` and getting pretty bad highlighting.

Are you saying that TSError should be removed as a highlight?

sQVe:
It seems like `TSError` is linked to `Error` somewhere, which is causing the issues for me. Setting the `TSError` highlight fixes things but... I don't know - this feels like something that should be handled by `nvim-treesitter`.

theHamsta:
At least, the default from this plugin is not highlight it as it's mostly annoying when it applies to half finished text or parts of the document that the parser not fully supports. But I understand your complaint about adding an injection that almost certainly fail. I'll have a look whether it can be easily fixed (e.g. by having CSS with JS injections)

PR #1629, opened 2021-07-25
Fix teal again

theHamsta:


stsewd (approved):

theHamsta:
For indents I just removed all invalid nodes. @euclidianAce

lockfile.json
```diff
@@ -198,7 +198,7 @@
     "revision": "0e36bed171768908f331ff7dff9d956bae016efb"
   },
   "yang": {
-    "revision": "eb4ace1fdc4954d7e4f0687c687219f78a6ce64e"
+    "revision": "cc81c9a8f40e77567e66cfd83fd532817ca256b9"
```
vigoux, line 201: This version should not be changed.

vigoux (changes requested):

lockfile.json
theHamsta: But that's the point of this PR to make the automated update mergeable

vigoux: No but there's two version changes here : one for `teal` and the other one for `yang` which should not be there

theHamsta: The versions shouldn't be touched by humans. Just let the GitHub bot do it's work

vigoux: What I say is simple:
1. You said you are updating teal
2. The changes in the lockfile contain both a teal update and a yang update
3. That's not right.

theHamsta: But that was our current workflow that we fix all the breakages of an updates in seperate commits for each language. That scales much better when two or more languages are broken.

I've edited the lockfile commit: it's only teal now.

theHamsta:
@vigoux can we merge this now? When we have a blocker like this for a long time the update-parser-pr get's larger and larger every time.

theHamsta:
Or we could make teal updates manual-only like verilog. Then, we don't have problems with it.

vigoux (approved):

PR #1635, opened 2021-07-28
Add some missing ftdetect files

tssm:
I started adding Nix and Fish, which I use, but I decided to add some others that I didn't find in the Neovim runtime by looking at how support plugins do it

### DIFF
--- a/ftdetect/julia.vim
+++ b/ftdetect/julia.vim
@@ -1,0 +1,1 @@
+autocmd BufRead,BufNewFile *.jl setfiletype julia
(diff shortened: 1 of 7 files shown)


ftdetect/julia.vim
```diff
@@ -0,0 +1 @@
+autocmd BufRead,BufNewFile *.jl setfiletype=julia
```
bfredl, line 1: hmm I think it needs to be `setfiletype julia`

tssm: Right, sorry, this is why I should do this kind of things during the week. BTW, the other files use `setlocal filetype` instead. Maybe that's a better option? My understanding is `setfiletype` only works if the file type has not been set already, which seems nicer, but I'm not sure, maybe it doesn't matter

vigoux (approved):
Let's also wait for @stsewd and @theHamsta opinion on this

theHamsta:
I don't have an opinion on filetpypes :smile:

theHamsta (approved):
Looks good!

stsewd (approved):
Thanks!

PR #1636, opened 2021-07-28
feature: add godotResource queries

pierpo:
This aims at giving support to `*.tscn` and `*.tres` files for Godot Engine.

I am not sure about the filetype... I took the name given by the author of the parser, but `godotResource` as a filetype does not feel very conventional with vim. (I'd expect `gdresource`, similarly to `gdscript`)

Here's a preview:

Screenshot from 2021-07-28 19-56-25

Screenshot from 2021-07-28 19-59-12

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -545,6 +545,16 @@ list.gdscript = {
   maintainers = { "@Shatur95" },
 }
 
+list.godotResource = {
+  install_info = {
+    url = "https://github.com/PrestonKnopp/tree-sitter-godot-resource",
+    files = { "src/parser.c", "src/scanner.c" },
+    requires_generate_from_grammar = true,
+  },
+  readme_name = "Godot Resources (godotResource)",
+  maintainers = { "@pierpo" },
+}
+
 list.turtle = {
   install_info = {
     url = "https://github.com/BonaBeavis/tree-sitter-turtle",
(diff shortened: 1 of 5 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -545,6 +545,16 @@ list.gdscript = {
   maintainers = { "@Shatur95" },
 }
 
+list.godotResource = {
+  install_info = {
+    url = "https://github.com/PrestonKnopp/tree-sitter-godot-resource",
+    files = { "src/parser.c", "src/scanner.c" },
+    requires_generate_from_grammar = true,
```
theHamsta, line 552: It only requires generate from source because there has not been a recent commit. Can you ask the author whether he's planning to update to the new ABI? If the project is abandoned you could as well fork it with the source file regenerated.

theHamsta (approved):

theHamsta:
Looks really nice!

lua/nvim-treesitter/parsers.lua
pierpo: Actually I didn't even check if this was necessary. How do you know? What's the impact of having this enabled?
(sorry, getting into treesitter + nvim-treesitter is quite a run :smile: )

But yeah, I could submit a PR to the repo if necessary. Or fork, if it's dead.
I'd just need some little directions :blush:

theHamsta: When you have this option on, users need the right `tree-sitter` installed. You can see the language version of the parser.c file here: https://github.com/PrestonKnopp/tree-sitter-godot-resource/blob/c611dd49e838efe25567dece14cf6353710f6d98/src/parser.c#L8 and also the tree-sitter version in the package.json if parser.c is up.to-date

pierpo: Thanks, that was very clear!
I just submitted a PR.

Let's see if it gets merged.

pierpo:
Any comment on the filetype name? I don't really like `godotResource` but I don't know what else to put.
Do you know if there is a convention? Can I diverge from the parser's name?

theHamsta:
Yeah, you can diverge. Just set for example `ftype="gdresource"` then it would be similar to gdscript that is already in this repo

PR #1640, opened 2021-07-29
highlights(zig): highlight enum,union,error and custom number type

maxxnino:


queries/zig/highlights.scm
```diff
@@ -27,13 +27,19 @@
...
+(struct_construction
+  (type_identifier) @constructor
 )
-
 ;; other identifiers
 (type_identifier) @type
+(custom_number_type) @type.builtin
 (primitive_type) @type.builtin
 (field_identifier) @field
+(enum_identifier) @field
```
vigoux, line 40: I think we have an `enum` type that should be used here.

vigoux:
Hi and thanks for this first contribution !
I left a little comment on your changes, and started our CI to check the changes!

Welcome in the contributor community !

maxxnino:
Thank you. I will update at the weekend. Btw, I maintain my own fork of tree-sitter-zig. Fix some of the existing problems too. I know, it's not perfect yet. https://github.com/Himujjal/tree-sitter-zig/issues/8

queries/zig/highlights.scm
maxxnino: The enum_identifier is the green box in the image below, i think it should be either **constant** or **field**

enum

theHamsta (approved):

queries/zig/highlights.scm
vigoux: I'd rather have that a constant, as this is not exactly a field... Sorry to b annoying :wink:

maxxnino: I tried to find what is [highlight](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/highlight.lua) for enum type is, is it a `@type.enum`.

vigoux: Then that's not the one we are searching for, indeed.

maxxnino: > I think we have an `enum` type that should be used here.

I just want to know what is the `enum` type you mentioned above. For the sake of improve me knowledge.

vigoux: It was just a mistake from me, I thought `enum` existed, which is not true :wink:

vigoux: Could you make that a `constant` though ? As this is technically a constant ?

maxxnino: I have a day off tomorrow. I will make it

vigoux (approved):

queries/zig/highlights.scm
maxxnino: It's a constant now, i have more highlight, should i make a separate pull request or add more commit to this pull request, or combine it in one commit

vigoux: As you wish, we can make that go further already, and see for the rest afterwards ?

maxxnino: Untitled - Frame 0

vigoux (approved):

queries/zig/highlights.scm
vigoux: Great !

PR #1662, opened 2021-08-03
add keywords to scala highlights

stumash:
moving the work that was incorrectly done [here](https://github.com/tree-sitter/tree-sitter-scala/pull/29) to here.

I'd like to fill in the rest of this file, but I must admit I'm a huge noob. I see in the [java `highlights.scm` file]() for example, captures are named things like `@keyword.return` and `@function`. Where do those names come from? Are they just standard among implementors of treesitter-based color themes?


stsewd:
@stumash thanks for contributing! you can read https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md to check all valid capture groups

stumash:
Hey @stsewd, funny seeing you here 😏 . I guess you can understand why I'm so bothered about not getting query highlighting working lol [over on this PR](https://github.com/nvim-treesitter/playground/issues/5).

Thanks for the tip about the common capture groups!

In any case, I should be allowed to add only the `keyword` captures for now, right? Later, I (or anybody else) can come back and add more.

queries/scala/highlights.scm
```diff
@@ -0,0 +1,40 @@
...
+  "class"
+  "def"
+  "do"
+  "else"
+  "extends"
+  "final"
+  "finally"
+  "for"
+  "forSome"
+  "if"
```
theHamsta, line 14: But the keywords should fit the subcategories we have in contributing.md

stumash: you're right. okay I've made the changes I think you're looking for. lmk

queries/scala/highlights.scm
```diff
@@ -0,0 +1,52 @@
...
+  "else"
+  "if"
+] @conditional
+
+[
+  "for" 
+  "while"
+] @repeat
+
+"import" @includes
```
theHamsta, line 50: Can you check whether this was not `@include`?

queries/scala/highlights.scm
```diff
@@ -0,0 +1,52 @@
...
+  "override"
+  "package"
+  "private"
+  "protected"
+  "sealed"
+  "super"
+  "this"
+  "throw"
+  "trait"
+  "try"
```
theHamsta, line 30: We have a exception category for try/catch stuff

queries/scala/highlights.scm
```diff
@@ -0,0 +1,52 @@
...
+] @conditional
+
+[
+  "for" 
+  "while"
+] @repeat
+
+"import" @includes
+
+"return" @keyword.return
```
theHamsta, line 52: You can also add `yield` to that category.

queries/scala/highlights.scm
```diff
@@ -0,0 +1,52 @@
+;; keywords
+
+[
+  "abstract"
+  "case"
```
theHamsta, line 5: Case might be `conditional`

queries/scala/highlights.scm
```diff
@@ -0,0 +1,52 @@
...
+  "catch"
+  "class"
+  "def"
```
theHamsta, line 8: `def` might be `@keyword.function`

theHamsta (changes requested):

queries/scala/highlights.scm
stumash: well `yield` is really not a return statement

queries/scala/highlights.scm
stumash: what about `case object Blah extends Meh`. `case` is not conditional in that situation.

stumash:
appreciate all the review, keep em coming! 👍

queries/scala/highlights.scm
stumash: I put a special situation for `case` as a conditional at the bottom so it overrides the default situation `case class`

queries/scala/highlights.scm
stumash: right again, it was `@include` not `@includes`. fixed.

queries/scala/highlights.scm
stumash: done.

queries/scala/highlights.scm
stumash: done.

stumash:
attempting to pass these checks by merging master. prob won't work

queries/scala/highlights.scm
theHamsta: You don't have to add it if it doesn't fit. I don't know Scala.

queries/scala/highlights.scm
```diff
@@ -0,0 +1,62 @@
...
+  "override"
+  "package"
+  "private"
+  "protected"
+  "sealed"
+  "super"
+  "this"
+  "trait"
+  "type"
+  "val"
```
theHamsta, line 27: Probably, one of theses keywords doesn't exist in the Tree-Sitter grammar. the CI complains about invalid nodes.

stumash:
weird. this `./scripts/check-queries.lua` worked for me

theHamsta:
image

queries/scala/highlights.scm
stumash: ya it's just that in other languages (e.g. python or javascript) yield is use to indicate that the current function is _generator function_.

in scala, you can do
```scala
def myFunction(): List[Int] = {
  val nums = List(1,2,3)
  val moreNums = for (n <- nums) yield n + 1
  1 :: moreNums // returns List(1,2,3,4)
}
```

`yield` is more of a special word related to `for`.

stumash:
oh my god.
these keywords aren't even parsed. wow.

looks like I gotta go figure out how to do add to the grammar over at tree-sitter-scala.

I'm gonna leave this PR open until then.

theHamsta:
You can only have anonymous nodes like `"this"` when there is a verbatim `"this"` in grammar.js

stumash:
no I get that, and maybe some of the keywords I have in doublequotes can be changed into queries that select the right node, but look here: https://github.com/tree-sitter/tree-sitter-scala/issues/19

loops are literally not yet added to the grammar yet 😮 

I'm gonna go work on that instead for now

stumash:
and side note I think all the grammars _should_ have literal `"keyword"` nodes

stumash:
okay [my PR](https://github.com/tree-sitter/tree-sitter-scala/pull/30) to `tree-sitter-scala` got merged so I'm gonna get this build to re-run with another pushed commit

queries/scala/highlights.scm
stumash: done

stumash:
@theHamsta stupid question, can you tell me how to get more debug info about why this thing is failing? does it not use the master branch of the relevant `tree-sitter-<language>` projects?

theHamsta:
No, it uses the revision in lockdown.json . You would have to update it.

.metals/metals.log
```diff
@@ -0,0 +1,8043 @@
+[0m2021.08.11 16:03:37 INFO  Started: Metals version 0.10.6-M1 in workspace '/Users/smashaal/me/code/nvim-treesitter' for client Neovim 0.6.0.[0m
```
theHamsta, line 1: I don't think we need those files

theHamsta (changes requested):

theHamsta:
You can use the linter of nvim-treesitter/playground to display invalid nodes.

.metals/metals.log
stumash: 🤦 dammit I can't believe I accidentally pulled those in

theHamsta (approved):

theHamsta:
Thanks for this, also for fixing this upstream in tree-sitter-scala!

stumash:
oh no, windows failures! 🙄

stumash:
@theHamsta the CI failures have nothing to do with my PR

stsewd:
@stumash can you merge master into this branch? If that still gives problems, maybe the scala parser update broke something. We have experienced this same error in other couple of PRs when a new parser is compiled (for example https://github.com/nvim-treesitter/nvim-treesitter/pull/1651), I'm not sure about the source of the problem.

stumash:
@stsewd or @theHamsta plz approve workflows?

PR #1671, opened 2021-08-06
C#: Improve highlights + Improve folds + Add locals

Luxed:
3 main improvements in this one to further improve C# highlighting (as discussed in #1349):

1. Improved highlights:
    Most notable improvements include missing keywords, support for [LINQ query syntax](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/linq/query-syntax-and-method-syntax-in-linq), better local function support and better generics support.
 2. Improved folds:
     This was actually simplified. Instead of folding entire methods _including_ their attributes, this new way of folding will only fold their _blocks_. This makes it easier to look through code when you have methods with a lot of Attributes or a lot of arguments split into multiple lines.
3. Added locals:
    This is a very simple implementation and is most likely not very good. It works, but not perfectly, to highlight the definitions using the refactor module but hasn't been extensively tested.

I don't have any screenshots for this unfortunately but I should be able to make some if needed (although they won't show all of the changes)

### DIFF
--- a/queries/c_sharp/folds.scm
+++ b/queries/c_sharp/folds.scm
@@ -1,18 +1,15 @@
-[
- (class_declaration)
- (constructor_declaration)
- (enum_declaration)
- (interface_declaration)
- (method_declaration)
- (namespace_declaration)
- (struct_declaration)
-
- (do_statement)
- (for_each_statement)
- (for_statement)
- (if_statement)
- (switch_statement)
- (try_statement)
- (using_statement)
- (while_statement)
+body: [
+  (declaration_list)
+  (switch_body)
+  (enum_member_declaration_list)
 ] @fold
+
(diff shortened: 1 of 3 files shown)


queries/c_sharp/folds.scm
```diff
@@ -1,18 +1,15 @@
...
- (struct_declaration)
+body: [
+  (declaration_list)
+  (switch_body)
+  (enum_member_declaration_list)
+] @fold
+
+accessors: [
+  (accessor_list)
+] @fold
```
vigoux, line 9: Wo wo wo, does that really work ? Never knew you could do that...

theHamsta (approved):
Looks good to me! Thanks for this nice work!

PR #1714, opened 2021-08-18
New kotlin parser

### LINKED ISSUE #1643: Remove or mention that kotlin parser is slow.
I came here because I can't open an issue in [their repository](https://github.com/tormodatt/tree-sitter-kotlin) which is suspicious!!
I know that Kotlin has dozens of features which makes parsing it harder.
So I did some investigations and I found out that the Kotlin's parser.c is huge ([92M before compiling](https://github.com/tormodatt/tree-sitter-kotlin/blob/master/src/parser.c), 18M after compiling) , which means 
40 bigger than Bash
80 bigger than Js
It takes more than 10s seconds to open every Kotlin file even with files that contain only one line .

So I believe that it's not usable or at least slow. so I think you guys need to mention that in Readme or just remove it

Joakker:
This PR replaces the old ~~stinky~~ kotlin parser for a new and improved one, with emphasis on simplicity and performance.

Hopefully fixes #1643

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -196,10 +196,10 @@ list.java = {
 
 list.kotlin = {
   install_info = {
-    url = "https://github.com/tormodatt/tree-sitter-kotlin",
+    url = "https://github.com/Joakker/tree-sitter-kotlin",
     files = { "src/parser.c" },
   },
-  maintainers = { "@tormodatt" },
+  maintainers = { "@Joakker" },
 }
 
 list.html = {
--- a/lockfile.json
+++ b/lockfile.json
@@ -108,7 +108,7 @@
     "revision": "12ea597262125fc22fd2e91aa953ac69b19c26ca"
   },
   "kotlin": {
-    "revision": "48bfb38abd93f8d617877d7bb7f92a6bb1166285"
+    "revision": "0369bdc8a7ed42830ff718af1173129d2028769a"
   },
   "latex": {
     "revision": "2c0d03a36ee979bc697f6a9dd119174cf0ef15e0"
--- a/queries/kotlin/highlights.scm
+++ b/queries/kotlin/highlights.scm
@@ -1,191 +1,190 @@
-;;; Kotlin
-
-(simple_identifier) @variable
-; Reset some
-(navigation_suffix (simple_identifier) @none)
-(import_header (identifier (simple_identifier) @none))
-(package_header (identifier (simple_identifier) @none))
-
-
-;; Strings
-(line_string_literal) @string
-(multi_line_string_literal) @string
-; Interpolated
-(interpolated_identifier) @variable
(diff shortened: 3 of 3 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -194,13 +194,13 @@ list.java = {
...
 
-list.kotlin = {
+--[[ list.kotlin = {
   install_info = {
     url = "https://github.com/tormodatt/tree-sitter-kotlin",
     files = { "src/parser.c" },
   },
   maintainers = { "@tormodatt" },
-}
+} ]]
```
theHamsta, line 203: Is there already a link for the new kotlin parser?

Joakker: It's [here](https://github.com/Joakker/tree-sitter-kotlin), though it's definitely not ready yet

Joakker:
I don't understand what's wrong. I specified the link to the grammar so it should have installed it. @theHamsta help?

stsewd:
You probably need to change the lockfile https://github.com/nvim-treesitter/nvim-treesitter/blob/6cf0a6692d5cf81f5d2480c0435a163a7dc335d1/lockfile.json#L113-L115

Joakker:
Thanks, at least now the failing tests are not my fault

theHamsta:
There are still some failure cases
image


- `package` is recognized as an identifier
- block comments are not parsed correctly
- inline, fun  

e.g. https://github.com/fwcd/kotlin-debug-adapter/blob/eb8f4bc2f86818c84f787fa91c21f62c9ccf8b31/adapter/src/main/kotlin/org/javacs/ktda/util/Utils.kt#L14

I really recommend running your parser on large repo and try to fix all error cases `tree-sitter parser **/*.kt` (when your bash supports `**` (else use find or just *.kt)

Joakker:
Will do 👍🏽

theHamsta:
@Joakker but startup time are already better now

theHamsta:
@Joakker I see that you're working on the Kotlin parser. I think this is ready to merge! People will report issues to you.

Fix #1643.

theHamsta (approved):

Joakker:
yey-evil

tormodatt:
I tested a few files, and I'd say there are still more errors in this parser than in the old one, but the startuptime in the old one made it almost unusable. this version is much better, and if @Joakker  is motiviated to maintain this I think it can get very good! 🥇 

let we know you want me start reporting issues to your parser @Joakker , or if you have enough on your list for the time being

theHamsta:
But I suspect that there is an issue with Tree-Sitter that increased the query load time from 300ms to 3s when upgrading to 0.19.0. Maybe perf can find out what Tree-Sitter is doing while loading the old Tree-Sitter-kotlin.

Joakker:
@tormodatt feel free to report issues to me 😃

PR #1721, opened 2021-08-20
Add an update script

vigoux:
This allows to repin a language by simply doing:
```
./scripts/update-lockfile.sh [mylang]
```

### DIFF
--- a/scripts/update-lockfile.sh
+++ b/scripts/update-lockfile.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+make_ignored() {
+  if [ -n "$1" ]
+  then
+    jq keys < lockfile.json | tail --line=+2 | head --lines=-1 | tr -d "\" ," | while read lang
+    do
+      if [ "$lang" != "$1" ]
+      then
+        printf "$lang,"
+      fi
+    done
+  fi
+}
+
+TO_IGNORE=$(make_ignored $1)
+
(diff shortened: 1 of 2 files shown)


scripts/update-lockfile.sh
```diff
@@ -0,0 +1,21 @@
...
+      if [ "$lang" != "$1" ]
+      then
+        printf "$lang,"
+      fi
+    done
+  fi
+}
+
+TO_IGNORE=$(make_ignored $1)
+
```
theHamsta, line 17: You can also use jq to select the string `jq "keys | select(. != $1) | join(\",\")"` ( command composed on mobile) https://stedolan.github.io/jq/manual/#select(boolean_expression)

theHamsta: But bash should be fine also

theHamsta (approved):

scripts/update-lockfile.sh
vigoux: As you wish, I am fine with both, but I am definitely more at ease with sh than with jq :wink:

theHamsta: You could also add a new env variable for the lua script whatever works best for you.

vigoux: I am fine with keeping the old one, it is sufficient to have the ignore one, so I guess keeping it like this is good.

PR #1722, opened 2021-08-20
fix(vim): add augroup as keyword

max397574:


max397574:
not sure if this is corect @vigoux

max397574:
added wrong commit

PR #1750, opened 2021-08-24
Add json5 parser

Joakker:
This PR adds support for json5 files. It's basically the same as normal JSON, but it adds some constructions from ECMAscript.

More info [here](https://json5.org/)

```json5
{
  /* I can do comments as well */
  "hello": "world",
  look_ma: "no quotes!",
  lead_dot: .01,
  trail_dot: 1., // With trailing commas, too!
}
```

### DIFF
--- a/ftdetect/json5.vim
+++ b/ftdetect/json5.vim
@@ -1,0 +1,1 @@
+autocmd BufNewFile,BufRead *.json5 set ft=json5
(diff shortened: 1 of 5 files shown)


vigoux (approved):
Looks pretty great from here, do you think this is ready ?

Joakker:
I think it is, yeah

Issue #1136, opened 2021-04-03
Invalid indents for python


last-partizan:
**Describe the bug**

Indents not working as expected, here is example:

```python
class Foo:# Enter
▌# cursor is here
```
Expected result

```python
class Foo:# Enter
    ▌# cursor is here
```

Same with `def function`

```python
def function():# Enter
▌# cursor is here
```

But must have 4 spaces indent.

Apparently, if file has an extra newline, following current line like this:

```python
class Foo:# Enter
    ▌# cursor is here
# This empty line must exist for indent to work.
```

It works fine.

**Output of `:checkhealth nvim_treesitter`**

<details>

```
health#nvim_treesitter#check
========================================================================
## Installation
  - OK: `tree-sitter` found  0.19.4 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v15.13.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found.

## Parser/Features H L F I
  - c              ✓ ✓ ✓ ✓
  - python         ✓ ✓ ✓ ✓
  - lua            ✓ ✓ ✓ ✓
  - html           ✓ ✓ ✓ ✓
  - typescript     ✓ ✓ ✓ ✓
  - vue            ✓ . ✓ .
  - toml           ✓ ✓ ✓ ✓
  - bash           ✓ ✓ ✓ .
  - javascript     ✓ ✓ ✓ ✓

 Legend: H[ighlight], L[ocals], F[olds], I[ndents]
         +) multiple parsers found, only one will be used
         x) errors found in the query, try to run :TSUpdate {lang}
```

</details>

**Output of `nvim --version`**

<details>

```
NVIM v0.5.0-dev+1150-g16edf3af3
Build type: RelWithDebInfo
LuaJIT 2.0.5
[...]

PR #1756, opened 2021-08-26
test: Add tests for is_in_node_range

last-partizan:
~~This fixes no indents at the end of python files~~

~~(Bonus, and passes 18 tests (+2))~~

This adds tests for is_in_node_range

Refs #1136

### DIFF
--- a/tests/unit/ts_utils_spec.lua
+++ b/tests/unit/ts_utils_spec.lua
@@ -1,6 +1,6 @@
 local tsutils = require "nvim-treesitter.ts_utils"
 
-describe("is_in_node_range returns", function()
+describe("is_in_node_range", function()
   local function test_is_in_node_range(line, col)
     local node = {
       range = function()
@@ -10,31 +10,31 @@ describe("is_in_node_range returns", function()
     return tsutils.is_in_node_range(node, line, col)
   end
 
-  it("false before node start", function()
+  it("returns false before node start", function()
     assert.is_false(test_is_in_node_range(0, 0))
     assert.is_false(test_is_in_node_range(0, 1))
     assert.is_false(test_is_in_node_range(0, 2))
   end)
 
-  it("true at node start", function()
+  it("returns true at node start", function()
     assert.is_true(test_is_in_node_range(0, 3))
   end)
 
-  it("true on first line of the node", function()
+  it("returns true on first line of the node", function()
     assert.is_true(test_is_in_node_range(0, 4))
   end)
 
-  it("true between node lines", function()
+  it("returns true between node lines", function()
     assert.is_true(test_is_in_node_range(1, 2))
     assert.is_true(test_is_in_node_range(1, 20))
   end)
 
-  it("false on node end", function()
+  it("returns false on node end", function()
     -- Ranges are end-exclusive
     assert.is_false(test_is_in_node_range(2, 5))
   end)
 
-  it("false after node end", function()
+  it("returns false after node end", function()
     assert.is_false(test_is_in_node_range(2, 6))
     assert.is_false(test_is_in_node_range(3, 0))
   end)
--- a/tests/unit/ts_utils_spec.lua
+++ b/tests/unit/ts_utils_spec.lua
@@ -1,43 +1,41 @@
 local tsutils = require "nvim-treesitter.ts_utils"
 
 describe("is_in_node_range", function()
-
   function test_is_in_node_range(line, col, expected)
     local node = {
-      range = function ()
-        return unpack({0, 3, 2, 5})
-      end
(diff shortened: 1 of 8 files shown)


kyazdani42 (approved):
looks good to me, @vigoux do you remember why the base implementation is so complex ?

theHamsta:
Because actually is_in_node_range should not include end row/col because Tree-Sitter are end exclusive

theHamsta (changes requested):
Tree-Sitter ranges are end exclusive

last-partizan:
@theHamsta are you sure about that?

image

Looking at this, `[1, 8]` is last character of function definition, body block and pass statement. Why it should be excluded?

Is there examples where this could break something? for python it fixes #1136, and i haven't noticed anything suspicious using it today. But, maybe there was something suspicious, and i discarded it thinking it was broken before, i'll try to remember what it was and test if reverting fixes it.

last-partizan:
And, if ranges are end exclusive, then root cause of invalid indents at the end of python files - is that we cannot find a root.

I just checked https://tree-sitter.github.io/tree-sitter/playground, and it has top-level root `module`, and neovim does not, maybe this is causing bug?

image

last-partizan:
Turns out i missed important point, this is not simple `check point is belongs to rectangle`, with lines and cols we need to verify all those edge cases.

While I was investigating, I added some tests. And i think they should be useful, but that's my first time writing tests in lua, and there was no good example on how to write unit-test. Please, review this draft version and i'll improve them with feedback.

tests/unit/ts_utils_spec.lua
```diff
@@ -0,0 +1,34 @@
+local tsutils = require "nvim-treesitter.ts_utils"
+
+describe("Test is_in_node_range", function()
```
vigoux, line 3: This should be `is_node_in_range` alone.

tests/unit/ts_utils_spec.lua
```diff
@@ -0,0 +1,34 @@
...
+    {{0, 3, 0, 5}, 0, 1, false},
+    {{0, 3, 0, 5}, 0, 3, true},
+    {{0, 3, 0, 5}, 0, 5, false},
+  }
+  for _, param in ipairs(params) do
+    local r, line, col, expected = unpack(param)
+    it(string.format("%s, %s, %s -> %s", table_repr(r), line, col, expected), function ()
+      is_in_node_range(unpack(param))
+    end)
+  end
```
vigoux, line 33: What we (at nvim, but not here yet), tend to do, is wrap those in `it` blocks (same signature as `describe`) with the "semantic" of this test, to help when the test fails afterwards, do you think you can make that a thin?

vigoux (commented):
Pretty good for the test addition, we should enhance that for future work and add more tests as we go, following this example.

tests/unit/ts_utils_spec.lua
last-partizan: I don't understand, could you give some examples?

vigoux: Maybe do something like : 
```
it('correctly gets nodes on the edge', function()
  -- Put your test here.
end)
```

last-partizan: Oh, describe each of the calls instead of just plain params list. Yes, i'll do that.

last-partizan: Fixed :)

I actually like it this way :)

btw, is there any docs on tree-sitter node range end-exclusivity? so i can link that in test description.

tests/unit/ts_utils_spec.lua
```diff
@@ -0,0 +1,43 @@
...
+  function test_is_in_node_range(line, col, expected)
+    local node = {
+      range = function ()
+        return unpack({0, 3, 2, 5})
+      end
+    }
+    return tsutils.is_in_node_range(node, line, col)
+  end
+
+  it("false before node start", function ()
```
vigoux, line 14: Maybe add a `returns` here, this is just wording but it really sounds better afterwards.

vigoux (commented):
We are getting there, just a final comment and we'll be good to go.

tests/unit/ts_utils_spec.lua
last-partizan: Yeah, should i add it for each `it`, or can i do this just for `describe`?

vigoux: I would add it to each `it`, so that it reads as `it('returns something',...)`.

vigoux (approved):
A-we-som

theHamsta (approved):

PR #1757, opened 2021-08-27
Add support for Perl 

lcrownover:
This PR adds support for the perl language courtesy of the hard work of @ganezdragon.

https://github.com/ganezdragon/tree-sitter-perl

### DIFF
--- a/queries/perl/highlights.scm
+++ b/queries/perl/highlights.scm
@@ -13,7 +13,9 @@
 
 [
 (package_name)
-(special_variable)
+(special_scalar_variable)
+(special_array_variable)
+(special_hash_variable)
 (special_literal)
 (super)
 ] @constant
--- a/queries/perl/highlights.scm
+++ b/queries/perl/highlights.scm
@@ -129,11 +129,14 @@
 (string_double_quoted)
 (string_qq_quoted)
 (bareword)
+(transliteration_tr_or_y)
+] @string
+
+[
 (regex_pattern_qr) 
 (patter_matcher_m)
 (substitution_pattern_s)
-(transliteration_tr_or_y)
-] @string
+] @string.regex
 
 (escape_sequence) @string.escape
 
--- a/queries/perl/highlights.scm
+++ b/queries/perl/highlights.scm
@@ -26,7 +26,7 @@
   (super)
   ]
   .
-  ("::" @constant)
+  ("::" @operator)
 )
 
 (comments) @comment
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -160,6 +160,15 @@ list.ruby = {
   maintainers = { "@TravonteD" },
 }
 
+list.perl = {
+  install_info = {
+    url = "https://github.com/lcrownover/tree-sitter-perl",
+    files = { "src/parser.c", "src/scanner.cc" },
+    generate_requires_npm = true,
+  },
+  maintainers = { "@ganezdragon" },
+}
+
 list.bash = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-bash",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -162,7 +162,7 @@ list.ruby = {
 
 list.perl = {
   install_info = {
-    url = "https://github.com/lcrownover/tree-sitter-perl",
+    url = "https://github.com/ganezdragon/tree-sitter-perl",
     files = { "src/parser.c", "src/scanner.cc" },
     generate_requires_npm = true,
   },


stsewd:
Hi, in order for the parser to work you'll need to add some queries, see https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights

lcrownover:
Looks like @delphinus has got us covered.

queries/perl/highlights.scm
```diff
@@ -0,0 +1,169 @@
+; Variables
+(variable_declaration
+  .
+  (scope) @keyword)
```
stsewd, line 4: isn't this more `@namespace`, maybe? not familiar with perl.

queries/perl/highlights.scm
```diff
@@ -0,0 +1,169 @@
...
+(super)
+] @constant
+
+(
+  [
+  (package_name)
+  (super)
+  ]
+  .
+  ("::" @constant)
```
stsewd, line 27: Constant usually  means an identifier, isn't this more a `@punctuation.delimiter/@punctuation.special/@operator`

queries/perl/highlights.scm
```diff
@@ -0,0 +1,169 @@
...
+(use_constant_statement
+  constant: (identifier) @constant)
+
+[
+"require"
+] @include
+
+(method_invocation
+  .
+  (identifier) @variable)
```
stsewd, line 51: I think you can just put this query into the top

```
(identifier) @variable
```

queries/perl/highlights.scm
```diff
@@ -0,0 +1,169 @@
...
+  ("->" @operator))
+
+[
+(word_list_qw)
+(command_qx_quoted)
+(string_single_quoted)
+(string_double_quoted)
+(string_qq_quoted)
+(bareword)
+(regex_pattern_qr)
```
stsewd, line 132: We have `@string.regex` for these

lcrownover:
I'm not super familiar with perl either.

@delphinus would you mind commenting on the above reviews?

queries/perl/highlights.scm
delphinus: This is a hack to highlight `::` operator to joining a package name & a function.

* before
   - スクリーンショット 0003-08-30 12 08 09
* after
   - スクリーンショット 0003-08-30 12 08 22

I'm using `@contant` because the prefix package name is `@constant`. But as you say, it is not suitable for this, maybe.

queries/perl/highlights.scm
delphinus: スクリーンショット 0003-08-30 12 15 49

This is used for here → `my`. The grammar.js defines `my`, `our`, `local`, `state` as `$.scope` [here][]. These are not so-called “namespace”s, and “scope” is also strange, as you say. But I cannot think of better choices. cc: @ganezdragon

[here]: https://github.com/delphinus/tree-sitter-perl/blob/c03173485f9a2b0337947d35181f58e8367483e6/grammar.js#L535-L540

queries/perl/highlights.scm
delphinus: Hmm……, then the package names changed their colors from `@constant` ones to `@variable` ones.

* before
   - スクリーンショット 0003-08-30 12 29 50
* after
   - スクリーンショット 0003-08-30 12 30 03

delphinus:
@lcrownover Can I be a collaborator of https://github.com/lcrownover/nvim-treesitter ? Then I can push my changes.

lcrownover:
Thanks @delphinus , I've added you as a collaborator.

queries/perl/highlights.scm
delphinus: Fixed on d30395d

queries/perl/highlights.scm
stsewd: you need to put this at the top of the file, so the other queries are overriden.

theHamsta:
@stsewd @lcrownover is this ready to merge?

queries/perl/highlights.scm
stsewd: then this should be highlighted as an operator, we have the same for cpp https://github.com/nvim-treesitter/nvim-treesitter/blob/e31bba5d97bfbc59d94a47f40e7ad9e36a5a53ee/queries/cpp/highlights.scm#L149-L149

stsewd (approved):
just one thing, but I think is fine, not a perl dev, so @delphinus may want to confirm

theHamsta:
@lcrownover can you address @stsewd's comment?

lcrownover:
@theHamsta I wish I could but I am not well-versed in Perl. 

@delphinus or  @ganezdragon, any input on @stsewd's comment?

stsewd:
@lcrownover I think you can just change it to `@operator` or `@punctuation.delimiter` (just saw we are using this capture for other langs), but definitely isn't a constant.

theHamsta:
@lcrownover friendly ping.

theHamsta:
Thank you for the update and the effort that you put into making the perl parser and your highlights available to everyone!😊

PR #1766, opened 2021-08-29
Add tree-sitter-norg

JoeyGrajciar:


JoeyGrajciar:
@vigoux and @theHamsta currently [Neorg](https://github.com/vhyrro/neorg) provides its own query files for parser which are mapped to custom highlight groups. We would like to have query files here, but there is a question how to properly switch from custom highlight groups to treesitter ones for this kind of markdown-ish format. What do you suggest?

cc: @vhyrro

vigoux:
We will only accept highlights that are defined her (appart if we go through a full RFC-ish cycle).

We will be glad to help you guys out to determine which highlights correspond to which in your parlancebut we will not include external highlight (our CI would not even pass).

JoeyGrajciar:
> but we will not include external highlight (our CI would not even pass).

Wasn't even suggested :) These are currently latest [highlights](https://github.com/vhyrro/neorg/blob/unstable/after/queries/norg/highlights.scm)

Another question regarding CI :) Can we add language version to parser install info? We are using some C++11 features that (supposedly) `macos-latest`'s `clang` does not like. Its default language version is older `c++03` or something like that.
Also those failures `uv_dlopen: A dynamic link library (DLL) initialization routine failed.` is it some know issue?

vigoux:
The failures are unrelated, we don't even know what is the problem with those.
For the `C++` version problem, I guess I'll wait for @theHamsta opinion, as he followed a similar issue with the haskell parser,t.

vhyrro:
We're currently in the process of doing fuzzing on the parser to capture potential bugs. I don't expect there to be any real memory issues since I made all the code memory-safe (apart from the actual `return new Scanner();` stuff which should be managed by tree-sitter itself). We have already stumbled upon a few parser hangs though which is not cool lol. We hope to fix all of those soon though, it's just a few missing checks in while loops every now and again.

I know vig doesn't like custom scanners :P. Sadly it's our only way out. I've tried without one before and it simply just doesn't cut it.

---
As for the highlight situation, it's all up to you guys. I feel that names like TSType and TSAnnotation don't fit for a markdown-esque format. Would it be possible to create generic highlight groups for markup formats too? Things like `TSHeading`, `TSParagraph`, `TSUnorderedListItem`? Or do you guys have to strictly adhere to a certain set of highlight groups.

vhyrro:
Good news: after a 6-hour fuzzing session using libfuzz nothing has crashed, nothing has segfaulted, nothing has hanged! Thing seem to be boding well so far in that department.

theHamsta:
C++11 is ok. MacOS seems to have problems with C++14. The stdlib on MacOS is not versioned, so I hoped that updating to Sierra might help. If you want you can create an env variable with parsers that are allowed to fail on MacOS. At the moment Haskell is allowed to fail on MacOS.

The DLL load problems are unrelated.

JoeyGrajciar:
@theHamsta are you sure? :)
This is from macos-latest build with clang `rvalue references are a C++11 extension [-Wc++11-extensions]`
What do you think about additional parser info part called `language`. E.g.
```
language = {
  c = 11
  cpp = 14
}
```
this would apply as additional option to compiler if specified. `cpp` could be changed to `cc` to query via file extension :)

theHamsta:
@JoeyGrajciar  some of the parsers use C++11 features. For instance, it only yields a warning when you don't specify C++11 and use auto. There have been some fixes to lambdas in C++14. Not sure whether it's because of that or because it rejects all lambdas.

I don't know how the language standards will help us. If we could specify the standard, then we could just always compile with C++17. We can't add the standard to our CLI invocation of the compiler because we compile mixed projects C/C++ in a single compiler invocation. To specify standard arguments, you would have to have three invocations (one for C compilation, one for C++ and one for linking). We didn't want to have this additional complexity the last time we talked about it since it only affected MacOS clang in combination with the Haskell parser.

theHamsta:
@JoeyGrajciar maybe adding a Makefile to your project would be a solution. We don't want to add arbitrary complexity to our compilation pipeline but doing `make` would certainly be acceptable and could also be a good solution for the Haskell parser.

theHamsta:
@JoeyGrajciar  do you have further interest in adding your parsing to our repo? I think MacOS shouldn't be problem with that. We can just ignore it or just compile with a make script #1795

JoeyGrajciar:
Yes i would like to add the parser to your repo :) I will do some adjustments and you could add that makefile support to master :)

JoeyGrajciar:
Closing in favor of #1795

PR #1770, opened 2021-08-30
Add tree-sitter-dot

rydesun:
DOT is a graph description language used in software Graphviz. I have written [tree-sitter-dot](https://github.com/rydesun/tree-sitter-dot), and this PR adds support for DOT.


vigoux (approved):
Looks all good to me, the parser does not use an external scanner which is great, because it avoids segfaults ! Once CI is green, we'll be good to merge.

vigoux (approved):

stsewd (approved):

PR #1771, opened 2021-08-30
fix(lang): only use first part in multi-lang

### LINKED ISSUE #1755: Treesitter is not correctly identifying the filetype when there is a "." separating multiple filetypes
### Before reporting

- [X] I have updated my neovim version to latest _master_
- [X] I have updated my plugin to the latest version
- [X] I have run `:TSUpdate`

### Describe the bug

When the detected filetype in neovim contains a dot (e.g., `ruby.rspec`) treesitter is not correctly parsing/recognizing the first filetype (e.g., `ruby`).

### To Reproduce

Open a file and change it's filetype to something that has a dot in it.  For example, open a ruby file and change it from `filetype=ruby` to `filetype=ruby.rspec`.  Upon changing it the file will lose its ruby highlighting.

### Expected behavior

I would expect, continuing this example, when a buffer has `filetype=ruby.spec` that treesitter would first apply the `ruby` syntax highlighting.

### Output of `:checkhealth nvim_treesitter`

```markdown
health#nvim_treesitter#check
## Installation
  - OK: `tree-sitter` found  0.20.0 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v16.7.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl" }

## Parser/Features H L F I J
  - yaml           ✓ ✓ ✓ ✓ ✓ 
  - ocaml          ✓ ✓ ✓ . ✓ 
  - ocaml_interface✓ ✓ ✓ . ✓ 
  - ocamllex       ✓ . . . ✓ 
  - rst            ✓ ✓ . . ✓ 
  - fennel         ✓ ✓ . . ✓ 
  - teal           ✓ ✓ ✓ ✓ ✓ 
  - ql             ✓ ✓ . ✓ ✓ 
  - verilog        ✓ ✓ ✓ . ✓ 
  - jsdoc          ✓ . . . . 
  - sparql         ✓ ✓ ✓ ✓ ✓ 
  - javascript     ✓ ✓ ✓ ✓ ✓ 
  - gdscript       ✓ ✓ . . ✓ 
  - turtle         ✓ ✓ ✓ ✓ ✓ 
  - clojure        ✓ ✓ ✓ . ✓ 
  - devicetree     ✓ ✓ ✓ ✓ ✓ 
  - r              ✓ ✓ . . . 
  - svelte         ✓ . ✓ ✓ ✓ 
  - cpp            ✓ ✓ ✓ ✓ ✓ 
  - lua            ✓ ✓ ✓ ✓ ✓ 
  - beancount      ✓ . ✓ . . 
  - latex          ✓ . ✓ . ✓ 
  - rust           ✓ ✓ ✓ ✓ ✓ 
  - ledger         ✓ . ✓ ✓ ✓ 
  - query          ✓ ✓ ✓ ✓ ✓ 
  - python         ✓ ✓ ✓ ✓ ✓ 
  - zig            x . x x ✓ 
  - swift          . . . . . 
  - graphql        ✓ . . ✓ ✓ 
  - typescript     ✓ ✓ ✓ ✓ ✓ 
  - ruby           ✓ ✓ ✓ ✓ ✓ 
  - tsx            ✓ ✓ ✓ ✓ ✓ 
  - bash           ✓ ✓ ✓ . ✓ 
  - c              ✓ ✓ ✓ ✓ ✓ 
  - comment        ✓ . . . . 
  - supercollider  ✓ ✓ ✓ ✓ ✓ 
  - php            ✓ ✓ ✓ ✓ ✓ 
  - java           ✓ ✓ . ✓ ✓ 
  - kotlin         ✓ . . . ✓ 
  - html           ✓ ✓ ✓ ✓ ✓ 
  - bibtex         ✓ . ✓ ✓ . 
  - julia          ✓ ✓ ✓ ✓ ✓ 
  - regex          ✓ . . . . 
  - json           ✓ ✓ ✓ ✓ . 
  - toml           ✓ ✓ ✓ ✓ ✓ 
  - css            ✓ . ✓ ✓ ✓ 
  - glimmer        ✓ . . . . 
  - c_sharp        ✓ ✓ ✓ . ✓ 
  - scss           ✓ . . ✓ . 
  - erlang         . . . . . 
  - go             ✓ ✓ ✓ ✓ ✓ 
  - scala          . . . . . 
  - vue            ✓ . ✓ . ✓ 
  - jsonc          ✓ ✓ ✓ ✓ ✓ 
  - nix            ✓ ✓ ✓ . ✓ 
  - dart           ✓ ✓ . ✓ ✓ 
  - elm            . . . . . 

The following errors have been detected:
  - ERROR: zig(highlights): ...im/0.5.0/share/nvim/runtime/lua/vim/treesitter/query.lua:161: query: invalid field at position 77
  - ERROR: zig(folds): ...im/0.5.0/share/nvim/runtime/lua/vim/treesitter/query.lua:161: query: invalid node type at position 90
  - ERROR: zig(indents): ...im/0.5.0/share/nvim/runtime/lua/vim/treesitter/query.lua:161: query: invalid node type at position 200
```

```text
NVIM v0.5.0
Build type: Release
LuaJIT 2.1.0-beta3
Compilation: clang -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -DNDEBUG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/tmp/neovim-20210702-14987-rep9di/neovim-0.5.0/build/config -I/tmp/neovim-20210702-14987-rep9di/neovim-0.5.0/src -I/usr/local/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include -I/usr/local/opt/gettext/include -I/tmp/neovim-20210702-14987-rep9di/neovim-0.5.0/build/src/nvim/auto -I/tmp/neovim-20210702-14987-rep9di/neovim-0.5.0/build/include
Compiled by brew@BigSur

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "/usr/local/Cellar/neovim/0.5.0/share/nvim"

```

### Additional context

Using dots `.` in this fashion for filetypes is a supported convention of n/vim.  From `:h 'filetype'`.

```text
When a dot appears in the value then this separates two filetypenames.  Example:
        /* vim: set filetype=c.doxygen : */ ~
This will use the "c" filetype first, then the "doxygen" filetype.
This works both for filetype plugins and for syntax files.  More than one dot may appear.
```

Original question on Zulip: https://nvim-treesitter.zulipchat.com/#narrow/stream/252338-help/topic/Sub-filetype.20of.20.60ruby.2Erspec.60.20is.20not.20being.20recognized.20as.20ruby/near/250275505

vigoux:
Fixes #1755 .

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -707,7 +707,7 @@ local M = {
 }
 
 function M.ft_to_lang(ft)
-  ft = vim.split(ft, '.', true)[1]
+  ft = vim.split(ft, ".", true)[1]
   return ft_to_parsername[ft] or ft
 end
 
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -707,6 +707,7 @@ local M = {
 }
 
 function M.ft_to_lang(ft)
+  ft = vim.split(ft, '.', true)[1]
   return ft_to_parsername[ft] or ft
 end
 


thornomad:
👍 Checked out the branch and can confirm correct syntax highlighting for `ruby.rspec`.  Thanks!

vigoux:
Not sure when this will be merged, we're waiting for people to review the PR...

lua/nvim-treesitter/parsers.lua
```diff
@@ -707,6 +707,7 @@ local M = {
 }
 
 function M.ft_to_lang(ft)
+  ft = vim.split(ft, '.', true)[1]
```
stsewd, line 710: We could use a comment here that vim's filetypes can be `.` separated. And this looks like a case where we may have different filetypes per-buffer :thinking:

stsewd (approved):

lua/nvim-treesitter/parsers.lua
vigoux: I don't think we really care about the part after the `.` tbh.

My reasoning here is that vim will try to guesstimate what the "inner" languages are based on regexes, which we _know_ are less accurate than what we can do with tree-sitter. Thus we don't care about what is after this first part of the filetype.

ram02z:
I tried this branch on one of my chezmoi templates for a toml file, which is just a golang template. Highlighting isn't perfect. Also, there are errors due to the template syntax, which is understandable. However, I am content with this for the time being and it will probably be solved in the future if a golang templates parser is added.

> ft-lang-dot branch with `additional_vim_regex_highlighting = true`
image


> master branch
image

lua/nvim-treesitter/parsers.lua
stsewd: Thinking more about this, the multi lang thing expects the syntax to be combined, but with treesitter it doesn't make sense to combine things, but to use a different parser that inherits another one.

For the example in the issue it doesn't break, but for the other example it does https://github.com/nvim-treesitter/nvim-treesitter/pull/1771#issuecomment-910360775. I think explicit is better here, and ruby.spect should be added to the filetype list of parsers.lua:ruby (if the grammar is really the same as ruby).

thornomad: I am not really familiar with how treesitter works so don't have much insight on this part ... but I know from a snippets standpoint it does utilize inheritance.  That is, my `ruby` snippets are included when I am in a `ruby.rspec` filetype _in addition_ to snippets specific to `rspec` (that are not present in a `ft=ruby` file).

In terms of highlighting, I could see a similar benefit regardless of how it is achieved.

stsewd: For plugins this is great, but for tree-sitter hightlight I think it is better to have a proper parser for different filetypes instead of having two parsers hightlighting the same file.

If the `ruby.rspec` works with the ruby grammar without errors, you can just add `used_by = { "ruby.rspec" }` here

https://github.com/nvim-treesitter/nvim-treesitter/blob/06e84da7d7c0302bb89ac39f8c3a21d5c557ab5b/lua/nvim-treesitter/parsers.lua#L164-L170

vigoux: I don't really like doing it this way because of the lack of genericity.
The vim doc clearly states that dot-separated filetypes are a standard, so I don't see why we could not use that.

theHamsta: Regarding the inheritance, it would mean that `cpp.doxygen`, `java.doxygen` would mean that the doxygen regex highlights are added in addition to the language highlights. Maybe this could be useful in future to conditionally inject some document parsers.

theHamsta (approved):

theHamsta (approved):

vigoux:
Could this be merged then after @theHamsta aproval?

theHamsta:
stylua needs to be fixed. I'm gonna make this required.

theHamsta:
@vigoux could you fix the formatting? stylua wants a different kind of quotes.

PR #1783, opened 2021-09-03
Update scala queries

stevanmilic:
Add scala folds

Update scala highlights

* add highlights for imports

* add highlights for type constructors

* add highlights for constants

* add highlights for interpolation string

* add highlights for comments through injection

### DIFF
--- a/queries/scala/highlights.scm
+++ b/queries/scala/highlights.scm
@@ -7,6 +7,15 @@
 ((identifier) @variable.builtin
  (#match? @variable.builtin "^this$"))
 
+; Assume other uppercase names constants.
+; NOTE: In order to distinguish constants we highlight
+; all the identifiers that are uppercased. But this solution
+; is not suitable for all occurences e.g. it will highlight
+; an uppercased method as a constant if used with no params.
+; Introducing highlighting for those specifi cases, is probably
+; best way to resolve the issue.
+((identifier) @constant (#match? @constant "^[A-Z]"))
+
 ;; types
 
 (type_identifier) @type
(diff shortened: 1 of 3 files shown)


vigoux:
I see that scala does not yet have a maintainer. If you'd like to become one, that would greatly help us, and for you that basically means that when the scala parser has an update with breaking change, we will ping you for help to fix the queries.

stevanmilic:
Sure @vigoux, I'll be glad.

queries/scala/highlights.scm
```diff
@@ -7,6 +7,15 @@
 ((identifier) @variable.builtin
  (#match? @variable.builtin "^this$"))
 
+; Assume other uppercase names constants.
+; NOTE: In order to distinguish constants we highlight
+; all the identifiers that are uppercased. But this solution
+; is not suitable for all occurences e.g. it will highlight
+; an uppercased method as a constant if used with no params.
+; Introducing highlighting for those specifi cases, is probably
```
theHamsta, line 15: Typo

queries/scala/highlights.scm
```diff
@@ -7,6 +7,15 @@
...
+; best way to resolve the issue.
+((identifier) @constant (#match? @constant "^[A-Z]"))
```
theHamsta, line 17: So in Scala you have constants identifiers like `Pi`, `MaxVertices`? Or rather `PI` , `MAX_VERTICES`

queries/scala/highlights.scm
```diff
@@ -77,14 +106,17 @@
...
 
 [
 (symbol_literal)
 (string)
 (character_literal)
+(interpolated_string_expression)
 ] @string
 
+(interpolation "$" @punctuation.special)
+(interpolation (identifier) @variable)
```
theHamsta, line 118: Can't this be treated like interpolations in Python or Bash? The variables should get `@variable` automatically when the string highlight is removed and the interpolation query is placed on the top of the file. Maybe this does not work for Scala

queries/scala/highlights.scm
stevanmilic: `Pi` yes, check the [offical docs ](https://docs.scala-lang.org/style/naming-conventions.html#constants-values-and-variables).

queries/scala/highlights.scm
stevanmilic: Yes, I was wondering what's the best approach for interpolation. I'll try what you proposed.

stevanmilic: It's looking great with `@none` now, thanks!

theHamsta (approved):
Looks nice! Thanks for your contribution!

PR #1789, opened 2021-09-06
[CMake] Improve some bindings

uyha:


queries/cmake/highlights.scm
```diff
@@ -3,7 +3,7 @@
  (bracket_argument)
 ] @string
 
-(variable_ref) @none
+(variable_ref) @variable
```
theHamsta, line 6: The `@none` will still be important since `TSVariable` is not set by default

theHamsta (approved):

PR #1792, opened 2021-09-08
highlights(c_sharp): add global keyword

theHamsta:


theHamsta:
@Luxed is this still the correct highlight after https://github.com/tree-sitter/tree-sitter-c-sharp/pull/205 or should this be import and a capture added to the other query?

Luxed:
> 
> 
> @Luxed is this still the correct highlight after [tree-sitter/tree-sitter-c-sharp#205](https://github.com/tree-sitter/tree-sitter-c-sharp/pull/205) or should this be import and a capture added to the other query?

Good question. I'd put it with the _using_ (include) for now so it's the same highlight since it's only used in this situation.
If ever we need more control over it, we'll update it, but for now I think it's fine.

theHamsta:
Ok, then we can close this. It should still have the include highlight from the coarse highlight above.

PR #1795, opened 2021-09-09
Haskell/norg c++14 compilation

theHamsta:
I don't know yet whether this is a good idea or whether the parser themselves should just have their own `Makefile` or `npm` task but we could at least increase the required C++ standard by having there compile commands.

A better solution would of course be to have a unified compilation, release and installation mechanism in upstream tree-sitter

https://github.com/nvim-treesitter/nvim-treesitter/pull/1766 @JoeyGrajciar

### DIFF
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -93,6 +93,30 @@ function M.select_compiler_args(repo, compiler)
   end
 end
 
+function M.select_compile_command(repo, cc, compile_location)
+  if string.match(cc, "cl$") or string.match(cc, "cl.exe$") or not repo.use_makefile then
+    return {
+      cmd = cc,
+      info = "Compiling...",
+      err = "Error during compilation",
+      opts = {
+        args = vim.tbl_flatten(M.select_compiler_args(repo, cc)),
+        cwd = compile_location,
+      },
+    }
+  else
+    return {
+      cmd = "make",
+      info = "Compiling...",
+      err = "Error during compilation",
+      opts = {
+        args = { "--makefile=" .. utils.join_path(utils.get_package_path(), "scripts", "compile_parsers.makefile"), "CC=" .. cc },
+        cwd = compile_location,
+      },
+    }
+  end
+end
+
 function M.select_install_rm_cmd(cache_folder, project_name)
   if fn.has "win32" == 1 then
     local dir = cache_folder .. "\\" .. project_name
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -94,7 +94,7 @@ function M.select_compiler_args(repo, compiler)
 end
 
 function M.select_compile_command(repo, cc, compile_location)
-  if string.match(cc, "cl$") or string.match(cc, "cl.exe$") or not repo.use_makefile then
+  if string.match(cc, "cl$") or string.match(cc, "cl.exe$") or not repo.use_makefile or fn.has "win32" == 1 then
     return {
       cmd = cc,
       info = "Compiling...",
@@ -110,7 +110,11 @@ function M.select_compile_command(repo, cc, compile_location)
       info = "Compiling...",
       err = "Error during compilation",
       opts = {
-        args = { "--makefile=" .. utils.join_path(utils.get_package_path(), "scripts", "compile_parsers.makefile"), "CC=" .. cc },
+        args = {
+          "--makefile=" .. utils.join_path(utils.get_package_path(), "scripts", "compile_parsers.makefile"),
+          "CC=" .. cc,
+          "CXX_STANDARD=" .. repo.cxx_standard,
+        },
         cwd = compile_location,
       },
     }
--- a/scripts/compile_parsers.makefile
+++ b/scripts/compile_parsers.makefile
@@ -3,14 +3,16 @@
 # Stephan Seitz, 2021-09-09 21:36
 #
 CC?=cc
+CXX_STANDARD?=c++14
+C_STANDARD?=c99
 
 all: parser.so
 
 parser.o: src/parser.c
-	$(CC) -c src/parser.c -std=c99 -fPIC -I./src
+	$(CC) -c src/parser.c -std=$(C_STANDARD) -fPIC -I./src
 
 scanner.o: src/scanner.cc
-	$(CC) -c src/scanner.cc -std=c++17 -fPIC -I./src
+	$(CC) -c src/scanner.cc -std=$(CXX_STANDARD) -fPIC -I./src
 
 parser.so: parser.o scanner.o
 	$(CC) parser.o scanner.o -o parser.so -shared -Os -lstdc++
--- a/scripts/compile_parsers.makefile
+++ b/scripts/compile_parsers.makefile
@@ -1,0 +1,16 @@
+#
+# compile_parsers.makefile
+# Stephan Seitz, 2021-09-09 21:36
+#
+CC?=cc
+
+all: parser.so
+
+parser.o: src/parser.c
+	$(CC) -c src/parser.c -std=c99 -fPIC -I./src
+
+scanner.o: src/scanner.cc
+	$(CC) -c src/scanner.cc -std=c++17 -fPIC -I./src
+
+parser.so: parser.o scanner.o
+	$(CC) parser.o scanner.o -o parser.so -shared -Os -lstdc++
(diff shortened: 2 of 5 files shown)


theHamsta:
About the failures on Windows, don't you have to manually export symbols to make them accessible on Windows using MSVC. How did this work until now? Why is it failing now on GCC and MSVC?

scripts/compile_parsers.makefile
```diff
@@ -0,0 +1,16 @@
...
+#
+CC?=cc
+
+all: parser.so
+
+parser.o: src/parser.c
+	$(CC) -c src/parser.c -std=c99 -fPIC -I./src
+
+scanner.o: src/scanner.cc
+	$(CC) -c src/scanner.cc -std=c++17 -fPIC -I./src
```
theHamsta, line 13: for our current parsers, c++14 would be enough

scripts/compile_parsers.makefile
```diff
@@ -0,0 +1,16 @@
...
+
+parser.so: parser.o scanner.o
+	$(CC) parser.o scanner.o -o parser.so -shared -Os -lstdc++
```
theHamsta, line 16: Does stdc++ exist on Mac? Or will clang just do some compat magic?

JoeyGrajciar: Tried to use `-lstdc++` on my M1 and it works with Apple clang 12.0.5

theHamsta: Yes, clang just tries to behave like GCC

theHamsta:
@JoeyGrajciar any idea why this fails for windows gcc/clang https://github.com/nvim-treesitter/nvim-treesitter/runs/4373980291?check_suite_focus=true ? though it says that it has installed norg.

JoeyGrajciar:
@theHamsta yes it is weird but error is correct `norg.so` file is missing in the artifacts.

mrossinek (commented):
Hi there! Thanks for including `norg` in this PR. We moved to an organization some time ago so I'm suggesting we update these links here. I'm one of the maintainers over there who is rather involved with the TS parser too at this point so I took the liberty of adding myself to the list, too.

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -93,6 +93,34 @@ function M.select_compiler_args(repo, compiler)
   end
 end
 
+function M.select_compile_command(repo, cc, compile_location)
+  if string.match(cc, "cl$") or string.match(cc, "cl.exe$") or not repo.use_makefile then
```
JoeyGrajciar, line 97: @theHamsta This is why it is failing :) `cc` is `clang` so it will go to `make` branch it does not find it, but probably the sequence with `pushd` and `popd` returns 0

JoeyGrajciar:
```suggestion
  if string.match(cc, "cl$") or string.match(cc, "cl.exe$") or not repo.use_makefile or fn.has "win32" == 1 then
```

theHamsta:
Thanks for helping out!

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -66,12 +66,10 @@ jobs:
...
         if: matrix.os != 'windows-2022'
-        env:
-          ALLOWED_INSTALLATION_FAILURES: haskell
         run: nvim --headless -c "luafile ./scripts/check-queries.lua" -c "q"
 
       - name: Check query files (Windows)
         if: matrix.os == 'windows-2022'
         env:
-          ALLOWED_INSTALLATION_FAILURES: haskell
+          ALLOWED_INSTALLATION_FAILURES: norg
```
JoeyGrajciar, line 74: is this needed?

theHamsta: https://github.com/nvim-treesitter/nvim-treesitter/runs/4386295723?check_suite_focus=true

JoeyGrajciar: okay but that failure is different and not only `tree-sitter-norg`
```
Downloading...

Checking out locked revision

The system cannot find the path specified.

fatal: reference is not a tree: 74a721e8eed1a4a25cf495d45974ba24f315f81a

Error while checking out revision

{

  cmd = "git",

  err = "Error while checking out revision",

  info = "Checking out locked revision",

  opts = {

    args = { "checkout", "74a721e8eed1a4a25cf495d45974ba24f315f81a" },

    cwd = "C:\\Users\\runneradmin\\AppData\\Local\\nvim-data\\tree-sitter-prisma"

  }

}

Downloading...

Checking out locked revision

The system cannot find the path specified.

error: pathspec 'main' did not match any file(s) known to git

Error while checking out revision

{

  cmd = "git",

  err = "Error while checking out revision",

  info = "Checking out locked revision",

  opts = {

    args = { "checkout", "main" },

    cwd = "C:\\Users\\runneradmin\\AppData\\Local\\nvim-data\\tree-sitter-norg"

  }

}

Downloading...

Checking out locked revision

The system cannot find the path specified.

fatal: reference is not a tree: 111b07762e86efab9a918b7c721f720c37e76b0a

Error while checking out revision

{

  cmd = "git",

  err = "Error while checking out revision",

  info = "Checking out locked revision",

  opts = {

    args = { "checkout", "111b07762e86efab9a918b7c721f720c37e76b0a" },

    cwd = "C:\\Users\\runneradmin\\AppData\\Local\\nvim-data\\tree-sitter-typescript"

  }

}
```
and I think also 2-3 other failed with same reason that there was some issue with downloading

JoeyGrajciar: Also the last run for gcc it was `graphql` and `html` having issues like this

theHamsta: Master has same problem, maybe we should just fetch using curl on CI

theHamsta:
@stsewd I think this is ready to go. This allows compilation of C++14 parsers like Haskell and Norg using old compilers and the default clang on MacOS 10. The CI failures also happen on other branches

theHamsta:
Let's try curl.

theHamsta:
Maybe this trick would also work: https://github.com/emacs-tree-sitter/elisp-tree-sitter/issues/166#issuecomment-986262035 But makefiles would be a bit more flexible

mrossinek:
Just a quick note from my side: I tested this branch out and everything seems to work just as expected. This is on Linux so maybe we can get someone on Windows and Mac to try as well?

theHamsta:
@mrossinek This does not change the Windows behavior. Compilation on Windows was working before. On Mac, I got positive feedback that this would work. @kyazdani42 can we merge this? This would be maybe an alternative to the makefiles if this works: https://github.com/nvim-treesitter/nvim-treesitter/pull/2094

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -93,6 +93,34 @@ function M.select_compiler_args(repo, compiler)
...
 end
 
+function M.select_compile_command(repo, cc, compile_location)
+  if string.match(cc, "cl$") or string.match(cc, "cl.exe$") or not repo.use_makefile or fn.has "win32" == 1 then
+    return {
+      cmd = cc,
+      info = "Compiling...",
+      err = "Error during compilation",
+      opts = {
+        args = vim.tbl_flatten(M.select_compiler_args(repo, cc)),
```
vigoux, line 103: I feel like we should also use the `cxx_standard` key in the `select_compiler_args` for completeness too, do you think this is wise ?

theHamsta: If it works without C++14, I would leave it that way. I think only Haskell and Norg require C++14 at the moment which works with a recent Windows compiler or Makefiles on Linux/MacOS

theHamsta: Ah, now I remember. It is not possible when invoking a C compiler with a single invocation. This is why I was using the Makefiles. In theory, it should work when you scope the cxx_standard with `-Xc++` flag. But it didn't work when I tried it.

danymat:
Hello, what's the state of this PR ? Is it ready to merge ?

theHamsta:
I'd say it's ready.

theHamsta:
@stsewd @vigoux is there anything left to do here? I'd say we can merge.

theHamsta:
Haskell is no longer affected.  norg and foam parser would like to have C++14

FoamScience:
I'll revert back to C too (for foam parser) as C++14 portability issues are chasing me around in many places :)

theHamsta:
@clason can you check whether there's something still blocking this PR?

clason:
You rebased this PR on master?

theHamsta:
yes

clason (approved):
Hmm, I haven't followed the discussion closely, but I trust you (and CI). Let's see if cleanup is necessary ;)

PR #1797, opened 2021-09-09
Update parsers

theHamsta:
In order to merge #1796, the spec for the fennel parser needs to be updated.


theHamsta:
@otommod can you update the queries after your rewrite?

otommod:
Done, #1799

PR #1799, opened 2021-09-10
fix(fennel): update queries for the new parser

otommod:


theHamsta:
You need to update the SHA of the fennel parser in lockfilr.json

otommod:
Oh, I hadn't updated the commit hash because you had, in your PR.  I added it now.

theHamsta:
@otommod and you need also remove `scanner.c` in `parsers.lua` like I did in order to compile the fennel parser.

theHamsta (approved):

PR #1819, opened 2021-09-16
C/C++  Highlight updates

theHamsta:
@jdrouhard did I forget some things?

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -170,4 +170,6 @@
 (preproc_params
   (identifier)) @parameter
 
+(attribute_declaration) @attribute
+
 (ERROR) @error
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -127,11 +127,11 @@
 (statement_identifier) @label
 
 [
-(type_identifier)
-(primitive_type)
-(sized_type_specifier)
-(type_descriptor)
- ] @type
+ (type_identifier)
+ (primitive_type)
+ (sized_type_specifier)
+ (type_descriptor)
+] @type
 
 (declaration (type_qualifier) @type)
 (cast_expression type: (type_descriptor) @type)
@@ -169,7 +169,5 @@
 
 (preproc_params
   (identifier)) @parameter
-
-(attribute_declaration) @attribute
 
 (ERROR) @error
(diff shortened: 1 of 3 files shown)


queries/c/highlights.scm
```diff
@@ -170,4 +170,6 @@
 (preproc_params
   (identifier)) @parameter
 
+(attribute) @attribute
```
jdrouhard, line 173: This will match each attribute separately, but not the surrounding `[[` and `]]`.

If you want to match instance of `[[...]]` as a single highlight, I'd change this to `(attribute_declaration) @attribute`

jdrouhard:
Just one minor comment, but otherwise this looks good.

theHamsta:
> Just one minor comment, but otherwise this looks good.

Did I miss your comment somewhere?

jdrouhard (changes requested):
Forgot to finish the review :)

queries/c/highlights.scm
theHamsta: also changed `co_yield`/`co_return` to `@keyword.return`

theHamsta:
Why is this failing on some CIs but not on Windows and on my machine?

jdrouhard:
Is it possibly related to the fact that c is typically bundled with neovim proper?

The tree-sitter-cli version was bumped too, wonder if that has anything to do with it.

theHamsta:
Yes, this could be the problem. I think it's a bad idea to bundle parsers with Neovim. I think this can be fixed on CI by either fixing rtp or uninstalling the C parsers but I guess this will cause headache to users.

jdrouhard:
You could take the attribute highlight out of the c file for now, that should get it going.

theHamsta:
@stsewd it still works.

stsewd (approved):

theHamsta:
@jdrouhard I think there is currently also a problem when you want to build the newest tree-sitter-cpp with the newest tree-sitter-c
```
❯ tree-sitter generate
Unresolved conflict for symbol sequence:

  attribute_specifier  •  'extern'  …

Possible interpretations:

  1:  (_constructor_specifiers  attribute_specifier)  •  'extern'  …  (precedence: 0, associativity: Right)
  2:  (_declaration_modifiers  attribute_specifier)  •  'extern'  …

Possible resolutions:

  1:  Specify a higher precedence in `_declaration_modifiers` than in the other rules.
  2:  Specify a higher precedence in `_constructor_specifiers` than in the other rules.
  3:  Add a conflict for these rules: `_declaration_modifiers`, `operator_cast_definition`, `operator_cast_declaration`, `constructor_or_destructor_definition`
```
I've just set my tree-sitter-cuda to depend on the newest versions of both. So I guess I will wait until both are compatible again.

jdrouhard:
@theHamsta Yes, I'm aware of this incompatibility. The actual dependency in tree-sitter-cpp is on the most recent released tree-sitter-c which it works with. I'm waiting on a PR to be merged in tree-sitter-c before I ask for another release there so I can update tree-sitter-cpp's dependency.

If you're curious - I split out the modifiers used in `_declaration_specifiers` so I can add `virtual` to the modifiers list in cpp, which fixes parsing errors when using `virtual` in function decls/defs anywhere but the first position (like `inline virtual...`).

PR #1827, opened 2021-09-17
Kotlin queries: Handle package statements

Joakker:


theHamsta (changes requested):
:+1:  nice!

Can you make `package` `@include`

Joakker:
There you go 😃

theHamsta (approved):

PR #1836, opened 2021-09-18
Support hjson

winston0410:
Provide support for hjson, originated from https://github.com/nvim-treesitter/nvim-treesitter/pull/1657

@theHamsta It should be good this time.


theHamsta (approved):
Thank you!

PR #1838, opened 2021-09-19
Use TSInstallSync! to force update of built-in parsers

theHamsta:
This avoids CI failures like in #1819 due to https://github.com/nvim-treesitter/nvim-treesitter/pull/1819#issuecomment-921326998


theHamsta:
This did not work!

PR #1862, opened 2021-09-24
cpp: fix for breaking changes in tree-sitter-cpp

jdrouhard:
These are fixes for the breaking changes in tree-sitter/tree-sitter-cpp#133

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -33,7 +33,7 @@
     "revision": "a03f1d2d1dfbf6f8e0fdca5f9ff030228241eb57"
   },
   "cuda": {
-    "revision": "b778edf16571b9c4f5555884b9c74ef526095148"
+    "revision": "0602f5b0337422a579edcfbbf24ea30f28f93383"
   },
   "dart": {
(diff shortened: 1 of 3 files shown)


theHamsta:
Can you change the lockfile.json so that the CI passes?

theHamsta:
Sorry, I still need to update the cuda parser. I will update the SHA as soon as it is ready!

theHamsta:
Do we have to change something for the user-defined literals PR?

jdrouhard:
> 
> 
> Do we have to change something for the user-defined literals PR?

No, that one shouldn't have had any breaking changes. You can add highlights for the new identifiers if you want, but I'm not sure what groups they'd be assigned to.

jdrouhard:
CI seems to be unable to find the nightly neovim release? Did neovim stop publishing nightlies?

theHamsta:
The CI to publish the nightlies fails from time to time. It should be back again tomorrow

maxbrunsfeld:
⚡ nice work on this @jdrouhard

theHamsta (approved):

PR #1863, opened 2021-09-24
highlights(cpp): highlight literal_suffix as `@character`

theHamsta:
https://github.com/tree-sitter/tree-sitter-cpp/pull/115 introduced user-defined literals. We could think about hightlighting the suffixes. E.g. `1s`, `1ms`, "1_km`.

We could use `@character` oder `@constant.built-in`, `@attribute` or introduce a new highlight group.


stsewd (approved):
Yeah, not sure about the best capture for this. `@string.special` could also be an option, but this works on numbers too.

theHamsta:
Or just `@variable` ? I'm not very sure on this.

PR #1870, opened 2021-09-27
fix(ecma): remove `@none` from interpolation

kyazdani42:
This PR removes the `@none` override for the interpolation because it overrides every ecma based template literal, setting none with a higher priority than ecma groups which made every template string look `un`highlighted.

should fix https://github.com/nvim-treesitter/nvim-treesitter/issues/1848

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -172,16 +172,16 @@
 (ternary_expression ["?" ":"] @conditional)
 (unary_expression ["!" "~" "-" "+" "delete" "void" "typeof"]  @operator)
 
-"(" @punctuation.bracket
-")" @punctuation.bracket
-"[" @punctuation.bracket
-"]" @punctuation.bracket
-"{" @punctuation.bracket
-"}" @punctuation.bracket
+[
+  "("
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
+] @punctuation.bracket
 
-((template_substitution ["${" "}"] @punctuation.special) @none
- ; Substitutions should have a higher priority than injections.
- (#set! "priority" 105))
+((template_substitution ["${" "}"] @punctuation.special) @none)
 
 ; Keywords
 ;----------


theHamsta:
Why not put it into separate queries? The priority applies for the whole query.

kyazdani42:
Not sure what you mean by separate queries ? The priority is supposed to be set for the `${}` pattern.

kyazdani42:
And i'm not sure, but is there a way to match any element inside the template literal to higher up the priority without setting a `@none` group ?

theHamsta:
Smaller ranges should always have higher "priority" than smaller ranges. Have you tried it out whether it works. I meant a separate for `@none` so that it's not affected by the priority. Probably, the priority has been a mistake altogether and should be removed

kyazdani42:
Oh okay i understand.

Well the priority doesn't really make sense here it just breaks normal behavior IMO.

Thing is, priority was introduced to fix the fact that the injection was taking precedence over the template string substitution.
It seems though this issue should be much harder to fix than just setting a simple priority with `none`.
Not sure what road to take, but this lastly introduced query definitely breaks default behavior so we must at least revert it.

theHamsta:
For a colorscheme that does not set `TSVariable`
current master:
image

with
```
(template_substitution) @none
((template_substitution ["${" "}"] @punctuation.special)
 ; Substitutions should have a higher priority than injections.
 (#set! "priority" 105))

```


image

without the `@none`
image

For a colorscheme that sets `TSVariable`, the last two options don't make a difference 

image
image

I didn't observe a situation where the higher priority for brackets actually was important and I don't know with what the brackets could conflict.

kyazdani42:
i do not have the same outputs as you. Basically:
```scheme
(template_substitution) @none
((template_substitution ["${" "}"] @punctuation.special)
 ; Substitutions should have a higher priority than injections.
 (#set! "priority" 105))
```
This gives 

Nothing changes if i remove the `@none` part.

and
```scheme
((template_substitution ["${" "}"] @punctuation.special) @none
 ; Substitutions should have a higher priority than injections.
 (#set! "priority" 105))
```
gives


Which is wrong both for normal interpolation and no interpolation.

If i remove the query, default behavior is
 


which does not highlight the brackets. I'm not sure how to highlight the content when the interpolation is within an injection though.

theHamsta:
For the case the there should not be an injection, I would just use the query that we have right now.

I think the most elegant solution, for the `html` case would be to have a injection attribute that makes the injection range exclusive to the injected language and splits up the parent language into a before and after range and nothing in between.

theHamsta:
What is the reason that we have `@none` for text in HTML? Because that seems to prevent the JS highlighting in the `hmll` example.

I would just use `((template_substitution ["${" "}"] @punctuation.special) @none)` without any priority.

kyazdani42:
I think @stsewd added that in order to make the injected part less prior to the substitution. I'll adjust the PR to remove the priority.

theHamsta:
> order to make the injected part less prior to the substitution

I don't get that. The priority prevents all highlighting when applied to `@none`. The braces have nothing to compete with the injection region is right after the braces.

theHamsta (approved):

kyazdani42:
The important part is not about the braces here.
What happens is the HTML injection is applying highlight prior to the substitution content, which applies `@string` on the range that should be `ecma` highlighted because it's an attribute content (if i'm not mistaken).
I don't know about the `(text) @none` in html, doesn't seem to affect ecma in any way

theHamsta:
@kyazdani42 it does, it is the culprit that prevents ecma highlighting in braces.

kyazdani42:
well this is kind of unexpected behavior because the HTML parser errors out quite a lot with js interpolations.

PR #1879, opened 2021-09-30
feat(fennel): update parser and queries

otommod:


otommod:
Oh, I just noticed. Should I be using `fix` instead of `feat` in the commit message?

stsewd (approved):
> Oh, I just noticed. Should I be using fix instead of feat in the commit message?

Anything describing the changes is fine for me :)

PR #1886, opened 2021-10-01
parsers: add D parser

nawordar:
1. ~~I have problem with `:TSUpdate` and `:TSInstall d` when I use my fork. I am getting the following message:~~

    ```
    nvim-treesitter[d]: Error during compilation
    cc1: fatal error: src/parser.c: No such file or directory
    compilation terminated.
    Press ENTER or type command to continue
    ```
    ~~The only way to install is to use `:TSInstallFromGrammar d`. How can I get `:TSUpdate` to work?~~

    Solved.

2. I am not sure what to do with function names. I think I will not add highlights to them because when calling functions in D it is possible to skip parentheses like this:
    
    ```d
    void a() {
      writeln("test");
    }

    void main(string[] args) {
        a();
        a;
    }
    ```

    When declaring function or calling it with parentheses I know that it is a function name, but when I omit parentheses, I have no way of knowing if `a` is a function name or variable. If I highlighted function names where possible, it would result in inconsistent highlighting. Maybe highlighting function names only in declarations, like GitHub does, would make sense?

### TODO

- [x] Add all operators
- [x] Add all strings
  - [x] Improve `token_string` highlighting
- [x] Finish indents and folds (for now)
- [ ] Add experimental parsers table


lua/nvim-treesitter/parsers.lua
```diff
@@ -85,6 +85,15 @@ list.cuda = {
   maintainers = { "@theHamsta" },
 }
 
+list.d = {
+  install_info = {
+    url = "https://github.com/CyberShadow/tree-sitter-d",
+    files = { "src/parser.c", "src/scanner.cc" },
+    generate_requires_npm = true,
```
theHamsta, line 92: add `requires_generate_from_grammar = true,` here

theHamsta:
Will `a` always be a function call? How would you refer to the function pointer? `&a?`

nawordar:
> Will `a` always be a function call?

Not if it is a delegate assigned to a variable or a struct field. Parentheses cannot be omitted in such situation. 

> How would you refer to the function pointer? `&a`?

Yes.

`a;` is almost certainly a function, because it wouldn't make sense to have such expression with variable. More ambiguous situation would be this:

```d
struct A {
  string b() {
    return "test";
  }
}

void main() {
  A a;
  auto c = a.b; // Is `b` a field or a method?
}
```

It is a deliberate design to allow using getter functions like fields. Similar thing also applies to single parameter functions and methods:

```d
import std;

void a(string parameter) {
  writeln(parameter);
}

void main() {
  a = "test"; // Prints "test"
}
```

https://dlang.org/spec/function.html#optional-parenthesis

`tree-sitter-d` also doesn't give me a lot of information:

```d
void main() {
  import std;

  writeln();
  writeln;
}
```

is shown in a playground as:

```
module [0, 0] - [5, 1]
  decl_defs [0, 0] - [5, 1]
    func_declaration [0, 0] - [5, 1]
      fundamental_type [0, 0] - [0, 4]
      func_declarator [0, 5] - [0, 11]
        identifier [0, 5] - [0, 9]
        func_declarator_suffix [0, 9] - [0, 11]
          parameters [0, 9] - [0, 11]
      specified_function_body [0, 12] - [5, 1]
        block_statement [0, 12] - [5, 1]
          statement_list [1, 2] - [4, 10]
            import_declaration [1, 2] - [1, 13]
              import_list [1, 9] - [1, 12]
                import [1, 9] - [1, 12]
                  module_name [1, 9] - [1, 12]
                    identifier [1, 9] - [1, 12]
            expression_statement [3, 2] - [3, 12]
              expression [3, 2] - [3, 11]
                postfix_expression [3, 2] - [3, 11]
                  qualified_identifier [3, 2] - [3, 9]
                    identifier [3, 2] - [3, 9]
            expression_statement [4, 2] - [4, 10]
              expression [4, 2] - [4, 9]
                primary_expression [4, 2] - [4, 9]
                  identifier [4, 2] - [4, 9]
```

Also, are failing checks related to this PR?

theHamsta:
> Also, are failing checks related to this PR?

No, Windows builds are failing at the moment.

I think I would still map those function that have a parenthesis. Just to have a similar highlighting with C++ code for instance. Can you share screenshots that compare the two possibilities on a larger file?

nawordar:
I'm giving up with highlighting method and function calls and variables. I don't know how to do this without breaking highlighting in other places.

queries/d/highlights.scm
```diff
@@ -0,0 +1,262 @@
...
+  "while"
+] @repeat
+
+[
+  "__FILE_FULL_PATH__"
+  "__FILE__"
+  "__FUNCTION__"
+  "__LINE__"
+  "__MODULE__"
+  "__PRETTY_FUNCTION__"
```
theHamsta, line 106: Aren't these macros?

nawordar: They call these ["special keywords"](https://dlang.org/spec/expression.html#specialkeywords), but they are basically equivalent to C++ compiler macros, so I could highlight them as such, indeed

queries/d/highlights.scm
```diff
@@ -0,0 +1,265 @@
...
+  "asm"
+  "auto"
+  "body"
+  "break"
+  "class"
+  "const"
+  "continue"
+  "debug"
+  "deprecated"
+  "do"
```
theHamsta, line 124: `do` is already in repeat

theHamsta: continue probably also

theHamsta (commented):
The parser takes very long to generate but results seem to be pretty accurate. Function highlighting is a bit complicated because you gave scoped identifiers a nested instead of a flat structure.

Apart from the two repeat keywords in the keywords section, this looks nice!

queries/d/highlights.scm
nawordar: If `continue` should be in `@repeat`, should `break` be there too?

theHamsta: yes

theHamsta:
Oh, I see. Grammar.js is generated from the original grammar. This is why it takes eternities to generate parser.c

lua/nvim-treesitter/parsers.lua
```diff
@@ -85,6 +85,16 @@ list.cuda = {
...
```
theHamsta, line 92: This parser does not require `npm`. I has no JS dependencies.

theHamsta:
```
Benchmark #1: tree-sitter generate
  Time (mean ± σ):     60.662 s ±  0.825 s    [User: 59.563 s, System: 1.027 s]
  Range (min … max):   59.086 s … 61.424 s    10 runs
```
I don't know whether we want to install from a release or fork with `src/parser.c`

nawordar:
> Maybe we can also just exclude D from the parsers that get installed with `maintained`

I think it is a good idea, as this parser is still fresh and is not 100% correct yet.

lua/nvim-treesitter/parsers.lua
```diff
@@ -85,6 +85,15 @@ list.cuda = {
   maintainers = { "@theHamsta" },
 }
 
+list.d = {
+  install_info = {
+    url = "https://github.com/CyberShadow/tree-sitter-d",
+    files = { "src/parser.c", "src/scanner.cc" },
+    requires_generate_from_grammar = true,
+  },
+  maintainers = { "@CyberShadow" },
```
theHamsta, line 94: It will not be maintained when you remove this line here. You could also add a configurable table that removes certain parsers in `install.lua` from `maintained`

theHamsta (approved):

lua/nvim-treesitter/parsers.lua
nawordar: Is it actually a list of parser maintainers or maintainers of the integrations in the repo? `@CyberShadow` is a maintainer of the parser, it is not me. I just wanted to add an integration here

theHamsta: It's the maintainers of the highlights and the integration into nvim-treesitter.

README.md
```diff
@@ -140,6 +140,10 @@ For `nvim-treesitter` to support a specific feature for a specific language requ
 
 The following is a list of languages for which a parser can be installed through `:TSInstall`; a checked box means that `nvim-treesitter` also contains queries at least for the `highlight` module.
 
+Experimental parsers are parsers that are maintained, but not stable enough for
+daily use yet. They are excluded from automatic installation when
+`ensure_installed` is set to `"maintained"`.
+
```
theHamsta, line 146: :+1: can you add verilog to this category? You can add a comment why those parsers are experimental in `parsers.lua`. D parser has a generated grammar.js that takes 60s to compile and verilog is still uses using ABI language version 12 because their parser doesn't compile with 13.

scripts/update-readme.lua
```diff
@@ -19,7 +19,9 @@ for _, v in ipairs(sorted_parsers) do
     generated_text = generated_text
       .. "- [x] "
       .. link
-      .. " (maintained by "
+      .. " ("
+      .. (v.parser.experimental and "experimental, " or "")
+      .. "maintained by "
```
nawordar, line 24: @theHamsta what do you think about marking parsers as experimental in this way?

PR #1900, opened 2021-10-02
LLVM parser/highlighter added

benwilliamgraham:


lua/nvim-treesitter/parsers.lua
```diff
@@ -738,6 +738,14 @@ list.hjson = {
   maintainers = { "@winston0410" },
 }
 
+list.llvm = {
+  install_info = {
+    url = "https://github.com/benwilliamgraham/tree-sitter-llvm",
+    files = { "src/parser.c" },
```
theHamsta, line 744: branch = "main",

benwilliamgraham: Good catch!

theHamsta:
Nice!

This statement causes a parser error (on FullDebug)
```
!0 = distinct !DICompileUnit(language: DW_LANG_C99, file: !1, producer: "clang", isOptimized: false, runtimeVersion: 0, emissionKind: FullDebug, enums: !2, retainedTypes: !2, globals: !2, imports: !2)
```

Is it on purpose that the parser doesn't expose any hierarchy on the AST?

theHamsta (approved):

benwilliamgraham:
> Nice!
> 
> This statement causes a parser error (on FullDebug)
> 
> ```
> !0 = distinct !DICompileUnit(language: DW_LANG_C99, file: !1, producer: "clang", isOptimized: false, runtimeVersion: 0, emissionKind: FullDebug, enums: !2, retainedTypes: !2, globals: !2, imports: !2)
> ```

Nice catch, I'll update that now!

> Is it on purpose that the parser doesn't expose any hierarchy on the AST?
> 
> `ret` could be `@keyword.return`

Yeah, for now I am modeling the parser after LLVM's existing vim highlighter, but in the future I plan on adding more structure to facilitate folding/etc.

queries/llvm/highlights.scm
```diff
@@ -0,0 +1,12 @@
...
+(number) @number
+(comment) @comment
+(string) @string
+(label) @label
+(keyword) @keyword
+(boolean) @boolean
+(float) @number
+(constant) @identifier
+(identifier) @constant
+(punctuation) @punctuation
```
theHamsta, line 12: I think this causes the CI to fail. We only have punctuation.delimiter

benwilliamgraham: Gotcha, should be fixed now!

theHamsta (approved):

PR #1904, opened 2021-10-06
Feat: Migrate to official elixir-lang parser

connorlay:
Description TBD

### DIFF
--- a/queries/elixir/highlights.scm
+++ b/queries/elixir/highlights.scm
@@ -1,5 +1,10 @@
-(ERROR) @error
+; Punctuation delimiters
+[
+  "%"
+  ","
+] @punctuation.delimiter
 
+; Punctuation brackets
 [
   "("
   ")"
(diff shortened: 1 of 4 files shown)


queries/elixir/highlights.scm
```diff
@@ -11,144 +16,184 @@
...
+  "when"
+  "|"
+  "|>"
+  "||"
+  "|||"
+  "~>"
+  "~>>"
+  "~~~"
+  "~"
+] @operator
```
jonatanklosko, line 72: Any reason for listing them explicitly instead of matching on relevant nodes? Like [this](https://github.com/elixir-lang/tree-sitter-elixir/blob/7ae20df181b86c79d826abd5aec7a3e32e3d8438/queries/highlights.scm#L34-L51). Either way works fine, though using the nodes should be easier from maintenance perspective.

queries/elixir/highlights.scm
```diff
@@ -11,144 +16,184 @@
...
+; Unused identifiers
+((identifier) @comment (#match? @comment 
+  "^_([^_].*)?$"
 ))
-
-(call (function_identifier) @conditional (#any-of? @conditional 
+ 
+; Calling a local def function
+(call target: ((identifier) @keyword.function (#match? @keyword.function
+  "^def(.*)?$"
```
jonatanklosko, line 162: I was wondering if we should match on the prefix, in the context of `defn` and `defnp` it's useful to have it something generic like this, as anyone can add their own marcos. However, this can very well have false positives, like `default_state()`. It could be slightly better as `"^def([^_]*)?$"`, since the def macros are generally a single word. No strong opinion here though :)

queries/elixir/injections.scm
```diff
@@ -1,20 +1,14 @@
-(comment) @comment
-
-; TODO: re-add when markdown is added
-; (heredoc
-;   (heredoc_content) @markdown)
-
-((sigil
-   (sigil_start) @_start
-   (sigil_content) @regex)
- (#match? @_start "~(r|R)[/</\\\"[({|]"))
```
jonatanklosko, line 10: Any reason for dropping regexp support?

queries/elixir/highlights.scm
```diff
@@ -11,144 +16,184 @@
...
+] @character
+
+; Booleans
+(boolean) @boolean
+
+; Nil
+(nil) @constant.builtin
+
+; Calling a local function
+(call target: (identifier) @function)
```
jonatanklosko, line 141: There's an additional edge with the pipe operator, if we have `a |> b`, then `b` is a function call. However in `def a |> b`, `b` is an identifier. You may want to special case it [like this](https://github.com/elixir-lang/tree-sitter-elixir/blob/7ae20df181b86c79d826abd5aec7a3e32e3d8438/queries/highlights.scm#L140-L152).

megalithic:
@connorlay mind updating the readme to reflect the new elixir link for the official parser (from the now deprecated/abandoned parser)?

HUGE thanks for pushing elixir dev for `nvim` users ahead!

megalithic:
> @megalithic that should be updated automatically after another PR is merged.

Thanks for the heads-up @stsewd !

PR #1905, opened 2021-10-06
feat(healthcheck): support native lua healthchecks, and fixes

muniter:
After neovim/neovim#15259 lua healthchecks are called directly and are prefered over neovim autoload ones. In the case of nvim-treesitter, the discrepancy between the autoload function nvim_treesitter and the lua base module nvim-treesitter makes it neccessary to add a custom logic with a guard.

Also the mentioned PR changed the design of healthchecks to not use output capturing, therefore avoid printing and instead concatenate and call the health functions.

### DIFF
--- a/autoload/health/nvim_treesitter.vim
+++ b/autoload/health/nvim_treesitter.vim
@@ -1,3 +1,7 @@
 function! health#nvim_treesitter#check()
-  lua require 'nvim-treesitter.health'.checkhealth()
+  if has('nvim-0.6') " TODO: Remove this guard once 0.6 is stable
+    call health#report_start("info: use the Lua checkhealth :checkhealth nvim-treesitter")
+    return
+  end
+  lua require 'nvim-treesitter.health'.check()
 endfunction
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -109,17 +109,17 @@ local function query_status(lang, query_group)
   end
 end
 
-function M.checkhealth()
+function M.check()
   local error_collection = {}
   -- Installation dependency checks
   install_health()
   queries.invalidate_query_cache()
-  health_start "Parser/Features H L F I J"
   -- Parser installation checks
+  local parser_installation = {"Parser/Features H L F I J"}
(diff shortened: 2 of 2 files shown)


muniter:
@theHamsta sorry, didn't notice the failed CI. But also now that https://github.com/neovim/neovim/pull/15935 was merged the guard I had in `autoload/health/nvim_treesitter.vim` is no longer necessary, and the file Itself is also not necessary after the original PR since you have a compat branch for 0.5 but I imagine you want to keep it while people move along nightlies.

theHamsta:
@muniter I'd prefer if master works for all nvim versions. Can you update the PR?

muniter:
@theHamsta the code in this PR will work on all versions. But in a small subset of commits between both healtchecks PR linked above it will perform the healtcheck twice. One for `nvim-treesitrer` another for `nvim_tressitter` I removed the guard since it's only to avoid the duplication and for a small subset of nightlies out there. 

If you still want the guard, I can add it.

muniter:
@HearseDev you can run `:checkhealth` alone and it will show all plugins, in this case you seem to be using the wrong name and that's the error. The ex command has an autocompletion you can use.

But in this case you should run `:chechealth nvim-treesitter` and for lsp `:checkhealth vim.lsp`.

pencilcheck:
somehow still having this issue after updating

```
health#nvim_treesitter#check
  - ERROR: Failed to run healthcheck for "nvim_treesitter" plugin. Exception:
    function health#check[21]..health#nvim_treesitter#check, line 1
    Vim(lua):E5108: Error executing lua ...im/bundle/nvim-treesitter/lua/nvim-treesitter/health.lua:147: attempt to index field 'query' (a nil value)
```

Mte90:
I get:
```
  - ERROR: Failed to run healthcheck for "treesitter" plugin. Exception:
    function health#check[20]..health#treesitter#check, riga 2
    Vim(lua):E5108: Error executing lua [string ":lua"]:1: attempt to call field 'check_health' (a nil value)
    stack traceback:
    [string ":lua"]:1: in main chunk
```

PR #1910, opened 2021-10-07
Update parsers with C# fixes

theHamsta:


theHamsta:
Rebasing on #1911 let's checks pass (use v0.5.1 instead of master)

PR #1911, opened 2021-10-10
Random CI changes: try to fix Windows build

### LINKED ISSUE #1901: Will the precompiled clang and cl parsers be available?
I'm on windows and I don't use a C compiler on my machine and I use precompiled parsers in the artifacts generated from the actions workflows. But for a long time now, there aren't any successful builds with `clang` and `cl` So, is there any plan to fix those builds? Besids, they're tests too, right?

I want to be able to get parsers generated on actions, using `clang` or `cl`

I have tried the `gcc` builds, but they have problems, e.g. the `css` and `html` parsers don't support any modules (from `:checkhealth nvim_treesitter`) and the `lua` parser gives an error on every cursor move.

**Additional context**
I don't think any screenshots are necessary, but I will provide if necessary.

theHamsta:
Ideas:
- use v0.5.1 (is always available, nightly is not today)
- try out `-static-stdc++` (maybe good for Windows users that can't use parsers from CI), maybe even `-static-libgcc`
- upload artifacts before checking queries (to check what's wrong, allow download even when CI fails)
- avoid choco?

Fixes #1901


theHamsta:
grafik
Epic! The new CI takes negative time that's counting down.

theHamsta:
I'm getting strange Git GUI prompts for some of the parsers on Windows. For me, curl works in that regard better than git.

theHamsta:
@eddyekofo94 I've tried all the parser artifacts produced by gcc,clang, cl.exe. They where all working for me on Windows 10, nvim v0.5.1 64bit

theHamsta:
I can't reproduce the CI failure on my machine using the artifacts downloaded from CI.

theHamsta:
@eddyekofo94 I've downloaded the parser artifacts of clang,gcc, and cl.exe . They where all working for me on Windows 10, nvim v0.5.1 64bit

theHamsta:
I've managed to create release with binaries on each commit however the commit hash doesn't update https://github.com/theHamsta/nvim-treesitter/releases/tag/nightly-parsers

theHamsta:
@stsewd Can we try to merge this? We are blocked quite often due to nvim nightly not being available. At the moment we're not using any nightly features but nightly could be added as allowed-to-fail if not downloadable for Linux

PR #1922, opened 2021-10-13
Add basic vimscript injection in `vim.cmd`/`nvim_command`/`nvim_exec`

vhyrro:
Hi!

This PR implements injection in `vim.cmd` codeblocks to improve the readability of Neovim+Lua codebases.
Currently it works for the following scenarios:
```lua
vim.cmd [[echo "Hello World!"]]
```
And:
```lua
vim.cmd "echo 'Hello World!'"
```

Obviously you can substitute `vim.cmd` for any of the other 3 functions that we have. There is one issue that I simply cannot resolve however, and that is this:
```lua
vim.cmd [[
    echo "Hello World!"
]]
```

Your `cdef` injection is plagued by the same problem, and that is that multiline strings that are `#match?`ed don't seem to have their respective parser injected. Any idea why this may even be? When I remove the `#match?` it works however then the `#offset!` of `2` messes with single-line strings that only use one char (`"`/`'`) instead of two (`[[`). Any idea why this could even be? lol

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -222,7 +222,7 @@
     "revision": "8f6b1f357d1231c420404b5f7a368a73c25adfa2"
   },
   "vim": {
-    "revision": "d141ba6da916ebaed0ceccafa069e0d24d8932b0"
+    "revision": "1d23679256edb241ebed8da1247e340bf9e0c0ad"
   },
   "vue": {
     "revision": "91fe2754796cd8fba5f229505a23fa08f3546c06"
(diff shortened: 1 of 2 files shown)


theHamsta:
@stsewd is this ready to go?

stsewd (approved):
Tested this, it works now!

gegoune:
That doesn't seem to work for 'aliased' calls such as:

```lua
local A = vim.api
A.nvim_command 'augroup END'
```
Is there a way to get those highlighted as well?

PR #1933, opened 2021-10-19
Elixir: fix a few highlights, add ExUnit support to locals

connorlay:
Relates to https://github.com/elixir-lang/tree-sitter-elixir/pull/11 and https://github.com/elixir-lang/tree-sitter-elixir/issues/1

This PR fixes the following bugs I've encountered recently:
* Removes the highlighting of sigils as comments when used with `@moduledoc` and friends. This was matching non-documentation attribute sigils as well. I'm not sure what the fix will be, but for now I think removing this makes more sense than leaving it in.
* Adds the missing `@keyword.operator` highlight for `not in`
* Adds ExUnit test support to locals
* Updates the README to point to the new Elixir parser

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [devicetree](https://github.com/joelspadin/tree-sitter-devicetree) (maintained by @jedrzejboczar)
 - [x] [dockerfile](https://github.com/camdencheek/tree-sitter-dockerfile) (maintained by @camdencheek)
 - [x] [dot](https://github.com/rydesun/tree-sitter-dot) (maintained by @rydesun)
-- [x] [elixir](https://github.com/ananthakumaran/tree-sitter-elixir) (maintained by @nifoc)
+- [x] [elixir](https://github.com/elixir-lang/tree-sitter-elixir) (maintained by @jonatanklosko)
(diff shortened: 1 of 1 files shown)


jonatanklosko:
@connorlay in which case the sigil would work improperly? Would it treat both of these as doc?

```
@doc ~s{ok}
@nodoc ~s{ok}
```

connorlay:
> in which case would the sigil work improperly? Would it treat both of these as doc?

Yes, exactly:
image

jonatanklosko:
Interesting :thinking: So in case it helps, [the queries](https://github.com/elixir-lang/tree-sitter-elixir/blob/c3b82ff9160d4780afbba5249b9de884e379c8dd/queries/highlights.scm#L7-L21) in `elixir-lang/tree-sitter-elixir` work as expected:

connorlay:
@theHamsta Thanks for the review! I'd like to hold off on merging this until I can try the queries @jonatanklosko linked to. I'll report back with what I find.

connorlay:
@theHamsta @jonatanklosko Using the queries from tree-sitter-elixir did not make a difference for me. I wonder if this is a bug with the highlighting logic in neovim (perhaps related to the default query match limit)? Either way, I think we should merge this in and create a follow-up issue.

PR #1934, opened 2021-10-20
Fix Elixir non-doc sigils

hauleth:
It also adds support for colouring documentation metadata attributes (ex. `@doc foo: :bar`).

### DIFF
--- a/queries/elixir/highlights.scm
+++ b/queries/elixir/highlights.scm
@@ -209,7 +209,7 @@
 (unary_operator
   operator: "@"
   operand: (call
-    target: ((identifier) @_identifier (#match? @_identifier "doc$")) @comment
+    target: ((identifier) @_identifier (#any-of? @_identifier "moduledoc" "typedoc" "shortdoc" "doc")) @comment
     (arguments [
       (string)
       (boolean)
--- a/queries/elixir/highlights.scm
+++ b/queries/elixir/highlights.scm
@@ -40,7 +40,7 @@
 ; Atoms & Keywords
 [
   (atom)
-  (quoted_atom) 
+  (quoted_atom)
   (keyword)
   (quoted_keyword)
 ] @symbol
@@ -84,11 +84,11 @@
 (stab_clause operator: _ @operator)
 
 ; Local Function Calls
-(call target: (identifier) @function) 
+(call target: (identifier) @function)
 
 ; Remote Function Calls
 (call target: (dot left: [
-  (atom) @type 
+  (atom) @type
   (_)
 ] right: (identifier) @function) (arguments))
 
@@ -154,7 +154,7 @@
   "do"
   "end"
   "fn"
-  "rescue" 
+  "rescue"
   "when"
   "else"
 ] @keyword
@@ -181,38 +181,34 @@
   ])
 
 ; Module attributes
-(unary_operator 
-  operator: "@" @constant 
+(unary_operator
+  operator: "@" @constant
   operand: [
     (identifier) @constant
     (call target: (identifier) @constant)])
 
 ; Sigils
-(sigil 
+(sigil
   "~" @string.special
   ((sigil_name) @string.special) @_sigil_name
-  quoted_start: _ @string.special 
+  quoted_start: _ @string.special
   quoted_end: _ @string.special
   ((sigil_modifiers) @string.special)?
   (#not-any-of? @_sigil_name "s" "S"))
 
-(sigil 
+(sigil
   "~" @string
   ((sigil_name) @string) @_sigil_name
-  quoted_start: _ @string 
+  quoted_start: _ @string
   (quoted_content) @string
   quoted_end: _ @string
   ((sigil_modifiers) @string)?
   (#any-of? @_sigil_name "s" "S"))
 
 ; Documentation
-(unary_operator 
-  operator: "@" @comment 
-  operand: (call 
-    target: (((identifier) @comment) @_identifier)
-    (arguments [
-      (string) @comment
-      (charlist) @comment
-      (boolean) @comment
-    ]))
+(unary_operator
+  operator: "@" @comment
+  operand: (call
+    target: ((identifier) @_identifier)
+    _) @comment
   (#any-of? @_identifier "moduledoc" "typedoc" "shortdoc" "doc"))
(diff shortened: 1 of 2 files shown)


theHamsta:
@jonatanklosko

jonatanklosko:
Hey @hauleth! We stopped highlighting doc sigils in #1933, you probably need to rebase. Is the issue mentioned in https://github.com/nvim-treesitter/nvim-treesitter/pull/1933#issuecomment-947217942 fixed with this approach?

hauleth:
@jonatanklosko the problem is still there.

https://user-images.githubusercontent.com/291639/139849418-13e8632a-b386-469c-8dd4-cf1b726d6769.mov

The reason for this is that in this query:

```scheme
; Documentation
(unary_operator 
  operator: "@" @comment 
  operand: (call 
    target: (((identifier) @comment) @_identifier)
    (arguments [
      (string) @comment
      (charlist) @comment
      (boolean) @comment
    ]))
  (#any-of? @_identifier "moduledoc" "typedoc" "shortdoc" "doc"))
```

The nested match in `arguments` will be applied **always** independently whether the `#any-of?` matches or not. It sounds like bug in NeoVim, but right now the only solution I have found is to remove that query at all.

jonatanklosko:
Interesting, we also tried alternative queries that work outside nvim (https://github.com/nvim-treesitter/nvim-treesitter/pull/1933#issuecomment-947229315) and they didn't work here either.

Please resolve the conflicts with master, what's the actual change here? :D

queries/elixir/highlights.scm
```diff
@@ -181,38 +181,33 @@
...
-    target: (((identifier) @comment) @_identifier)
-    (arguments [
-      (string) @comment
-      (charlist) @comment
-      (boolean) @comment
-    ]))
+(unary_operator
+  operator: "@" @comment
+  operand: (call
+    target: (((identifier) @comment) @_identifier))
```
jonatanklosko, line 212: So even without sigils it doesn't work for strings etc?

cc @connorlay

queries/elixir/highlights.scm
```diff
@@ -181,38 +181,34 @@
...
-    (arguments [
-      (string) @comment
-      (charlist) @comment
-      (boolean) @comment
-    ]))
+(unary_operator
+  operator: "@" @comment
+  operand: (call
+    target: ((identifier) @_identifier)
+    _) @comment
```
hauleth, line 213: @jonatanklosko I have updated it. Unfortunately now it doesn't highlight strings as a comment:

Zrzut ekranu 2021-11-2 o 15 31 05

stsewd:
> The nested match in arguments will be applied always independently whether the #any-of? matches or not. It sounds like bug in NeoVim, but right now the only solution I have found is to remove that query at all.

That looks like a problem that was already fixed in neovim nightly

queries/elixir/highlights.scm
hauleth: This is what I get on master.

Zrzut ekranu 2021-11-2 o 16 08 36

And if I move that query above `(string) @string` then it is not applied at all.

Query like:

```scheme
; Documentation
(unary_operator 
  operator: "@" @comment 
  operand: (call 
    target: (((identifier) @comment) @_identifier)
    (arguments [
      (string) @comment
      (charlist) @comment
      (boolean) @comment
    ])
  (#any-of? @_identifier "moduledoc" "typedoc" "shortdoc" "doc")))
```

Nor

```scheme
; Documentation
(unary_operator 
  operator: "@" @comment 
  operand: (call 
    target: (((identifier) @comment) @_identifier)
    (arguments [
      (string) @comment
      (charlist) @comment
      (boolean) @comment
    ]
  (#any-of? @_identifier "moduledoc" "typedoc" "shortdoc" "doc"))))
```

Do not work either. It seems that there is no way to avoid matching the inner query fragments.

jonatanklosko: Wait, so what's wrong master version then? That it ignores sigils or something else I miss?

queries/elixir/highlights.scm
```diff
@@ -180,39 +180,38 @@
...
+
 ; Documentation
-(unary_operator 
-  operator: "@" @comment 
-  operand: (call 
-    target: (((identifier) @comment) @_identifier)
+(unary_operator
+  operator: "@"
+  operand: (call
+    target: ((identifier) @_identifier (#match? @_identifier "doc$")) @comment
```
jonatanklosko, line 212: Would a more specific match also work?

```suggestion
    target: ((identifier) @_identifier (#match? @_identifier "^(moduledoc|typedoc|doc)$")) @comment
```

stsewd: if you go for that, using `#any-of?` is better.

hauleth: Fixed

jonatanklosko (approved):
Thanks for figuring it out!

stsewd (approved):
Looks like this was already approved by the users that know elixir :D

queries/elixir/highlights.scm
```diff
@@ -180,39 +180,38 @@
...
   quoted_end: _ @string
   ((sigil_modifiers) @string)?
   (#any-of? @_sigil_name "s" "S"))
 
+; Module attributes
+(unary_operator
+  operator: "@"
+  operand: [
+    (identifier)
+    (call target: (identifier) @constant)]) @constant
```
stsewd, line 206: is this correct? `@constant` at the end means highlight the whole `unary_operator` node.

queries/elixir/highlights.scm
```diff
@@ -180,39 +180,38 @@
...
     (arguments [
-      (string) @comment
-      (charlist) @comment
-      (boolean) @comment
-    ]))
-  (#any-of? @_identifier "moduledoc" "typedoc" "shortdoc" "doc"))
+      (string)
+      (boolean)
+      (charlist)
+    ] @comment))) @comment
```
stsewd, line 217: same here, even `@` will be highlighted as a comment.

hauleth: That is the goal.

queries/elixir/highlights.scm
hauleth: Yeah. In reality we probably could avoid colouring the argument of the call, but in reality that changes nothing.

Issue #1935, opened 2021-10-20
[Feature]Option to build parse with zig


maxxnino:
**Describe the solution you'd like**
- Zig ship with clang compiler. So it really easy to build parse with zig using this command
```
zig c++ src/parser.c  -lc -Isrc -shared -o parser.so
```

**Additional context**
- Better file size compare to gcc 9.3 on my machine
![zig gcc](https://user-images.githubusercontent.com/34153891/138210486-ef6b85d1-6c64-48f6-aa59-7a434cbf9db6.png)
- Performance still good with ~24k line of code
![zig 2](https://user-images.githubusercontent.com/34153891/138089660-426cc576-84b5-43e4-80bf-aab0d19d1a86.png)
- All maintained parser. Don't know why some parser with really small in size, but still working
![full parser](https://user-images.githubusercontent.com/34153891/138212939-047abefd-9e49-4475-bb65-8c904b1c2514.png)

PR #1936, opened 2021-10-21
Add zig as compiler for parser

maxxnino:
From my feature request https://github.com/nvim-treesitter/nvim-treesitter/issues/1935

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -11,7 +11,7 @@ local shell = require "nvim-treesitter.shell_command_selectors"
 local M = {}
 local lockfile = {}
 
-M.compilers = { vim.fn.getenv "CC", "cc", "gcc", "clang", "cl" }
+M.compilers = { vim.fn.getenv "CC", "cc", "gcc", "clang", "cl", "zig" }
 M.prefer_git = fn.has "win32" == 1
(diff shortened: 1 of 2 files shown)


ajitid:
> @maxxnino would using the zig compiler make stuff easier for Windows users?

Can confirm, it does!

Though I'm getting this error (on Windows) when running `:checkhealth`.

image

Any poiner to why it could be happening?

maxxnino:
@ajitid It happen to me aswell, What is uv_dlsym btw
Untitled

ajitid:
hey @maxxnino, I would suggest to install gcc via msys2 and add to path [using this way](https://www.devdungeon.com/content/install-gcc-compiler-windows-msys2-cc). I reinstalled treesitter and have got no errors now.

I don't know what caused this, but running healthcheck or opening any file crashes nvim in Windows Terminal (Powershell). Running Powershell app directly works fine.

maxxnino:
@ajitid as i see some parser work, some don't. But luckly it work with my language. I already used gcc ( installed via scoop) in the past. I just don't want depend on gcc anymore.

maxxnino:
@ajitid I just found out. Update to lastest nvim-treesitter and re install the parser. fix the prolem

- Example
```
:TSInstall css cpp json
```

PR #1973, opened 2021-11-02
Add `ftdetect` setup for `glimmer` parser.

rwjblue:
Without this, `*.hbs` files are not highlighted (unless you have some other plugin that adds it for you).

Fixes https://github.com/alexlafroscia/tree-sitter-glimmer/issues/44


stsewd (approved):
thanks

PR #1990, opened 2021-11-08
Add Operators for Ruby

CKolkey:
Many operators were missing from Ruby, so, I added them. There's a list on the ruby-docs site: https://ruby-doc.org/core-2.6.2/doc/syntax/precedence_rdoc.html

Additionally, with string interpolation (`"#{xxx}"`), the closing `}` was being matched to `punctuation.bracket` instead of `punctuation.special`, so I fixed that too.

I didn't see any further details about how to contribute, so if I've overlooked anything, I'll be happy to add it.

### DIFF
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -1,8 +1,5 @@
 ; Variables
 (identifier) @variable
-(interpolation
-  "#{" @punctuation.special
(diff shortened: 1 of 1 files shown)


theHamsta:
We have a `CONTRIBUTING.md` but I guess you did everything right already.

theHamsta (approved):

PR #1994, opened 2021-11-09
Add condition function to module config

patrick96:
The function is called with the language and bufnr, if it returns false,
the module is disabled for that buffer.

This gives the user more fine-grained control over whether a module is
started.

This is an initial implementation to gather feedback. I chose to add an additional `cond` setting, but I could just as well change it to support function values for the `disabled` config. Let me know what you prefer or if you'd rather have a completely different approach.

My problem was that some modules (mainly `highlight` and [`rainbow`](https://github.com/p00f/nvim-ts-rainbow)) massively slowed down opening of large files and I wanted a way to disable some or all modules for large files. I first tried using autocmds to just disable them with `TSBufDisable` but the autocmd events fired either before the buffer was filled or after the TS modules already ran.

TODOs:

* [ ] Add to documentation

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -335,8 +335,9 @@ M.commands = {
 }
 
 -- @param mod: module (string)
--- @param ft: filetype (string)
-function M.is_enabled(mod, lang)
+-- @param lang: the language of the buffer (string)
+-- @param bufnr: the bufnr (number)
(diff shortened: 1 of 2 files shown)


theHamsta (approved):
Looks nice! Thank you

patrick96:
Wait, I didn't update the documentation yet :sweat_smile: 

I'll open another PR.

PR #2004, opened 2021-11-12
feat: get node at cursor of other win

fdschmidt93:
Currently, `ts_utils.get_node_at_cursor` does actually not work for any `winnr` other than `0`, even though the function signature suggests you can pass any valid `winnr`. This PR is a naive and minimum implementation to achieve that in a way that it should not interfere with any current other uses of `get_node_at_cursor` or `get_root_for_position`.

cc @danymat


stsewd (approved):
LGTM, thanks!

theHamsta (approved):

PR #2012, opened 2021-11-14
[fix] wrong highlight on preproc function definition

jerry901:
There was an wrong highlight issue on definitions of preproc function. The last bracket in the function definition is only getting highlighted as a bracket.

image

Actually the picture above is cpp code, but i thought the problem was based on c query, and now it works like below.

image


jerry901:
closed because there is another PR handling this problem ealrier.

PR #2016, opened 2021-11-16
R add new highlights

aspeddro:


queries/r/highlights.scm
```diff
@@ -69,23 +78,43 @@
...
  (inf)
- (nan)
- (na)
- (null)
 ] @keyword
 
+[
+  (nan)
+  (na)
+  (null)
```
theHamsta, line 98: `@variable.builtin`?

theHamsta (approved):
Nice! Thank you!

queries/r/highlights.scm
aspeddro: is `@type.builtin`

PR #2031, opened 2021-11-20
Highlight tests and add tests to CI

theHamsta:
- added highlight tests based on the tree-sitter/Sublime unit test format
- added unit tests and indent tests to CI
- marked expected failures on indent tests so we can monitor progress
     - regressions will be a test failure in future
     - fixes for indent tests that do not remove the expected failure marker will be a failure in CI


tests/indent/common.lua
```diff
@@ -84,6 +96,7 @@ function M.run_indent_test(file, runner, opts)
 
   -- load reference file
   vim.cmd(string.format("edit %s", file))
+  vim.bo.indentexpr = "nvim_treesitter#indent()"
```
theHamsta, line 99: this should not be necessary! But otherwise Rust and Lua indentexpr from ft-plugin would load!

scripts/run_tests.sh
```diff
@@ -4,15 +4,13 @@ HERE="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 cd $HERE/..
 
 run() {
-    nvim --headless --noplugin -u scripts/minimal_init.lua \
```
theHamsta, line 7: this was just an experiment. Revert!

tests/indent/common.lua
```diff
@@ -28,6 +32,9 @@ local function same_indent(state, arguments)
 end
 
 local function format_indent(arg, fmtargs)
+  if not arg or not fmtargs then
+    return
+  end
```
theHamsta, line 37: necessary if you do assertions inside the indent assertions (or have other kind of failures while checking indents)

theHamsta: can be probably removed

.github/workflows/tests.yml
```diff
@@ -0,0 +1,42 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+
+      - name: Test Dependencies
+        run: |
+          mkdir -p ~/.local/share/nvim/site/pack/plenary.nvim/start
+          cd ~/.local/share/nvim/site/pack/plenary.nvim/start
+          git clone https://github.com/nvim-lua/plenary.nvim
+          curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly --profile minimal
+          ~/.cargo/bin/cargo install --git https://github.com/theHamsta/highlight-assertions
```
theHamsta, line 27: should we move this to the org's namespace

tests/query/highlights_spec.lua
```diff
@@ -0,0 +1,96 @@
...
+        .. (row + 1)
+        .. ":"
+        .. (col + 1)
+        .. ': expected "'
+        .. assertion.expected_capture_name
+        .. '", captures: '
+        .. vim.inspect(vim.tbl_keys(captures))
+        .. '", highlights: '
+        .. vim.inspect(vim.tbl_keys(highlights))
+    )
```
theHamsta, line 85: at some point we may also want negative assertions

vigoux (approved):
I did not review it completely yet, but I approve the idea greatly, we will have to add more tests ass highlights contributions come by.

I trust the implementation.

theHamsta:
I started adding tests for Hack in #2041. Really makes it easier to add highlights.

vigoux:
I think everybody agrees on this being a major improvement for stability and regression testing.

I would not even wait for everyone's approval on this, as it does not change the core functionnalities of the plugin.

One little nitpick though is the fact that it adds an external dependency. But as that's for testing/dev only. I think it is fine.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

theHamsta:
We can use an own implementation later. I just thought it would be good to use the implementation of tree-sitter instead of emulating it.

vigoux:
I agree that it is better this way than to reinvent the wheel.

It is fine the have a dev-dependency though, the amount is still pretty low for now.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

PR #2032, opened 2021-11-20
ci: do not set any secrets for stylua

theHamsta:
I'm pretty sure no tokens are required for running stylua


theHamsta:
why is this needed?

theHamsta:
```
#### `token` (Required)

GitHub token. Required since the binary is downloaded from GitHub releases (to speed download)

```

PR #2037, opened 2021-11-21
zig: highlight zig format, type and function

maxxnino:


theHamsta (approved):

theHamsta:
I'm not 100% sure about this change. Shouldn't this just be a plain comment?

maxxnino:
It's a hint for `zig fmt` not format the following block of code. It's have a purpose, so I think it's nice to have a highlight. I just don't know what highlight group it's should belong to.

theHamsta:
I was just thinking that the same kind of annotation also exist for rustfmt, gofmt, and clang. Is this the same regex that zigfmt would use? E.g. would `//                   zig fmt: on ` also trigger zig fmt?

maxxnino:
I just tested and it doesn't trigger. I just force-pushed update it

theHamsta:
I mean it should trigger in the exact same situation as zigfmf recognizes the pattern to give visual feedback. Maybe you could look at its implemention?

maxxnino:
I checked it implementation.  You can add as many as space between `zig fmt: on` or `off` as you want, but not words. It'll work as intended. And auto trim off un-needed space after format

PR #2042, opened 2021-11-23
Add support for Hack

theHamsta:


queries/hack/highlights.scm
```diff
@@ -0,0 +1,162 @@
...
+  "float"
+  "int"
+  "string"
+  "arraykey"
+  "void"
+  "nonnull"
+  (null)
+  "mixed"
+  "dynamic"
+  "noreturn"
```
theHamsta, line 41: TODO: `@type.builtin`

theHamsta:
Based on https://github.com/nvim-treesitter/nvim-treesitter/pull/2031 to add tests. Certainly, not complete @4e554c4c could you review and provide annotated test files like the ones in this commit if something is wrong? We can also merge now and you can continue refining the queries.

Maybe it would be nice to have `$` (like for PHP) and `?` in `$foo` and `?Type` as visible tokens in the parser. Would require an upstream change

theHamsta:
Is the filetype `hack` or `hacklang`

4e554c4c:
@theHamsta I've tried installing the playground and looking at some of my job's code on this branch and it looks great!
I think I've found a couple of parser errors though. These are most likely upstream issues, but I can definitely provide some minimal examples.
I'm having a bit of trouble running the tests, however, even after installing highlight-assertions. For the hacklang tests I'm getting
```
Vim:E474: Unidentified byte: Error: /home/pounce/build/nvim-treesitter/parser/hack.so: cannot open shared object file: No such file or directory
```
Any idea why?

4e554c4c:
also https://github.com/hhvm/vim-hack/blob/main/ftdetect/hack.vim has a better ftdetect

theHamsta:
@4e554c4c did you run `:TSInstall hack`

4e554c4c:
LGTM! Before there were issues with type constants and shapes but those seem to be resolved now
test case showing this:
```
<?hh

class C extends Superclass implements Iface {
//      ^ keyword          ^ keyword
  use Trait;
  // < keyword
  const type X = shape(
  // <- keyword ^ type.builtin
    "a" => int,
  // ^ string
    "b" => string,
  //       ^ type.builtin
  );
}
```

ftdetect/hack.vim
```diff
@@ -0,0 +1,31 @@
...
+" Maintainer:   Max Wang <mwang@fb.com>
+" URL:          https://github.com/hhvm/vim-hack
+" Last Change:  April 3, 2014
+"
+" Copyright: (c) 2014, Facebook Inc.  All rights reserved.
+"
+" This source code is licensed under the MIT license found in the
+" LICENSE file found at https://github.com/hhvm/vim-hack.
+"
+au BufRead,BufNewFile *.hack,*.hackpartial
```
theHamsta, line 12: for me `BufRead,BufNewFile *.hack,*.hackpartial,*.hck` would be fine

theHamsta:
@stsewd can we merge this?

queries/hack/highlights.scm
```diff
@@ -0,0 +1,295 @@
...
+ "private"
+ "public"
+ "using"
+ "namespace"
+ "attribute"
+ "const"
+ (xhp_modifier)
+ (final_modifier)
+ "extends"
+  "insteadof"
```
vigoux, line 36: I am neatpicky but...
```suggestion
 "insteadof"
```

queries/hack/highlights.scm
```diff
@@ -0,0 +1,295 @@
...
+  "tupe"
+  (array_type)
+  "bool"
+  "float"
+  "int"
+  "string"
+  "arraykey"
+  "void"
+  "nonnull"
+  (null)
```
vigoux, line 65: I don't know hack, but is this a null type, or the constant `null` ?

tests/query/highlights/hack/attribute-type.hack
```diff
@@ -0,0 +1,15 @@
...
+//     TODO: ?operator (? not captureable at the moment)
+  ?'int' => int
+//       ^ operator
+);
+
+<<A3(1), A2(2,3,)>>
+//       ^ attribute
+type T2 = (function(T1): string);
+//    ^ type
+//          ^ function (cannot capture keyword "function" as keyword.function)
```
vigoux, line 12: Why is that ?

vigoux (commented):
I am going to trust you for most of this, just left a few comments.

queries/hack/highlights.scm
theHamsta: Fixed it to constant.builtin.

tests/query/highlights/hack/attribute-type.hack
theHamsta: (cannot capture keyword "function" as keyword.function as it does not appear in AST). This has to be fixed upstream. The current workaround is to highlight all of `function foo` as `function`

theHamsta: This should be changed to https://github.com/slackhq/tree-sitter-hack/blob/dd7c1bb8b11d290db4aec173740efdb510f0bba9/grammar.js#L455 `alias(/function(s)?/, "function")`

vigoux: Might be worth opening a PR then ?

theHamsta: https://github.com/slackhq/tree-sitter-hack/pull/36

theHamsta: image

This is very creepy: https://github.com/slackhq/tree-sitter-hack/issues/37

vigoux: Great, then I would wait for the PR to be merged, so that we directly pull the changes.

theHamsta: I don't think the upstream repo is active anymore.

4e554c4c: I spoke to some folks at slack and they said they'd look into it, it looks like slackhq/tree-sitter-hack#37 has been fixed

theHamsta:
Outstanding issue here is the merge of https://github.com/slackhq/tree-sitter-hack/pull/36 . But I don't think this should block this PR

ftdetect/hack.vim
```diff
@@ -0,0 +1,31 @@
...
+  \ setl filetype=hack |
+  \ setl syntax=hack
```
clason, line 14: I don't like that this mixes `filetype` and `syntax`. I believe the proper thing here is to make `hack` a subtype of `php`, as in `php.hack`?

ftdetect/hack.vim
```diff
@@ -0,0 +1,31 @@
...
+
+au BufRead,BufNewFile *.hhi,*.hck
+  \ setl filetype=hack |
+  \ setl syntax=php
```
clason, line 18: If it doesn't use the `hack` syntax, it doesn't belong here? Proper filetype support should go upstream (Vim).

theHamsta: Let's just remove the ftdetect.

theHamsta: hack is already on the tracking issue.

clason (changes requested):
I'm not familiar enough with Hack and PHP to comment on the queries, but I trust you there.

theHamsta:
@clason me neither, we have to trust https://github.com/nvim-treesitter/nvim-treesitter/pull/2042#issuecomment-992765190

clason (approved):

README.md
```diff
@@ -189,6 +189,7 @@ We are looking for maintainers to add more parsers and to write query files for
 - [x] [gomod](https://github.com/camdencheek/tree-sitter-go-mod) (maintained by @camdencheek)
 - [x] [gowork](https://github.com/omertuc/tree-sitter-go-work) (maintained by @omertuc)
 - [x] [graphql](https://github.com/bkegley/tree-sitter-graphql) (maintained by @bkegley)
+- [ ] [hack](https://github.com/slackhq/tree-sitter-hack)
```
clason, line 192: should this get a checkmark now, that basic highlight queries are included?

theHamsta: the checkmark depends on a maintainer being present. It's auto-generated.

clason: Ah, I see -- then we probably want to change the README (because that states the checkmark means "basically usable, has queries"; my bad!)

Since the maintainer information is already explicit, I'd prefer to use the checkmark to indicate independent information -- such as basic usability.

theHamsta: sure, sounds reasonable.

ftdetect/hack.vim
fredemmott: fwiw, this is historical: Hack used to be a superset of PHP, so in hhvm/vim-hack, we set filetype=hack to mark what tooling should be used, but we dealt with the Hack syntax by extending the PHP syntax.

It's not a good way to do things nowadays, and I'm happy to see the syntax=php isn't in the final version of this PR :)

That said: `.hhi` should be treated as a hack file, and probably should have the `hack` syntax. They're not used often - they're effectively our equivalent of header files, and used to give static analysis information about the builtin functions that are implemented in C++.

PR #2048, opened 2021-11-24
ecma: highlight private properties and methods

numToStr:
Preview:
image

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -11,6 +11,7 @@
 
 (property_identifier) @property
 (shorthand_property_identifier) @property
+(private_property_identifier) @property
 
 (variable_declarator
   name: (object_pattern
@@ -46,7 +47,7 @@
 (generator_function_declaration
   name: (identifier) @function)
 (method_definition
-  name: (property_identifier) @method)
+  name: [(property_identifier) (private_property_identifier)] @method)
 
 (pair
   key: (property_identifier) @method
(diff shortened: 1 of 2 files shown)


queries/ecma/highlights.scm
```diff
@@ -87,6 +90,9 @@
 (call_expression
   function: (member_expression
     property: (property_identifier) @method))
+(call_expression
+  function: (member_expression
+    property: (private_property_identifier) @method))
```
numToStr, line 95: I was wondering though. Is it possible to merge this into one? Like
```scheme
(call_expression
  function: (member_expression
    property: (property_identifier|private_property_identifier) @method))
```

theHamsta:
```
(call_expression
  function: (member_expression
    property: [(property_identifier) (private_property_identifier)] @method))
```

numToStr: Thanks @theHamsta. I'll update the PR shortly.

theHamsta: We just added highlight tests to this repo. If you want to you can submit your example file with highlight assertions for the change (e.g. private fields that are now highlighted)

Format would be to add a caret below pointing the highlight that you changed
```ts
class {
  #priv_field = "World";
//   ^ property
}

```

This is not a must. I'm just  Your PR is awesome already!

numToStr: Sure, why not. I'll try to add test for this PR :)

numToStr:
@theHamsta I've added the test. I think now it is good to go.

(I can't believe tests got passed in the first try 😄)

theHamsta (approved):
Nice! Thank you!

PR #2051, opened 2021-11-24
ci: add nightly again to CI

theHamsta:
Was removed because not always available. Add one nightly config to not have surprises.


kyazdani42 (approved):
good idea

PR #2052, opened 2021-11-25
zig: work around for hang neovim

maxxnino:
Because two query below cause hanging neovim. I don't know what's cause the problem yet. It had been worked really well until now.
```
(PtrTypeStart "c" @variable.builtin)
(
  (ContainerDeclType
    [
      (ErrorUnionExpr)
      "enum"
    ]
  )
  (ContainerField (IDENTIFIER) @constant)
)
```

### DIFF
--- a/queries/zig/highlights.scm
+++ b/queries/zig/highlights.scm
@@ -60,17 +60,15 @@ exception: "!" @exception
   (#eq? @variable.builtin "_")
 )
 
-(PtrTypeStart "c" @variable.builtin)
+; (PtrTypeStart "c" @variable.builtin)
 
-(
-  (ContainerDeclType
-    [
-      (ErrorUnionExpr)
-      "enum"
-    ]
-  )
-  (ContainerField (IDENTIFIER) @constant)
-)
+; (
+;   (ContainerDeclType
+;       (ErrorUnionExpr)
+;       ; "enum"
+;   )
+;   (ContainerField (IDENTIFIER) @constant)
+; )
 
 field_constant: (IDENTIFIER) @constant
 


theHamsta:
Could this be related to https://github.com/tree-sitter/tree-sitter/issues/973 I had the feeling that certain queries got slow with the bump to 0.19.0?

Can you try to reproduce this without Neovim using the tree-sitter CLI and report it to github.com/tree-sitter/tree-sitter?

maxxnino:
I'll give it a try on the weekend

maxxnino:
@theHamsta i tried `tree-sitter query highlights.scm demo.zig`. With two hanging query above, It's parse just fine, I don't know what cause the problem. Using tree sitter v0.20.1 on my arch linux virtual box
```
real    0m0.224s
user    0m0.171s
sys     0m0.037s
```

theHamsta:
@maxxnino We can merge this workaround when you promise to remember that we need to find the root cause for this problem as this can easily also affect other languages.

maxxnino:
yes, I promise

theHamsta (approved):

numToStr:
FWIW, I tried opening the following zig file and neovim died on me. Then I uninstalled the `zig` parser after that I was able to edit the file.

<details>
<summary>
demo.zig
</summary>

```zig
const std = @import("std");

const Game = enum {
    rock,
    paper,
    scissor,
};

const Union = union(enum) { Int: u8, String: []const u8 };

pub fn main() anyerror!void {
    const a: Game = .rock;
    const b = Game.paper;
    std.log.info("{} {}", .{ a, b });

    const u = Union{ .Int = 89 };
    const v = Union{ .String = &[_]u8{ 1, 2, 3 } };
    const w = Union{ .String = &[_]u8{ 'a', 'b', 'c' } };
    const x = Union{ .String = "hello" };
    std.log.info("{}", .{u});
    std.log.info("{}", .{v});
    std.log.info("{}", .{w});
    std.log.info("{s}", .{w.String});
    std.log.info("{}", .{x});

    switch (u) {
        Union.Int => |f| std.log.info("Union.Int: {d}", .{v}),
        Union.String => |f| std.log.info("Union.Int: {d}", .{v}),
    }

    const yo = switch (u) {
        Union.Int => |_| "Union::Int",
        Union.String => |_| "Union::Int",
        // else => undefined,
    };

    std.log.info("{s}", .{yo});
}
```
</details>

maxxnino:
@numToStr could you try update neovim, nvim-treesitter and reinstall parser again. If it could help.

zig

PR #2056, opened 2021-11-26
add experimental support for neos fusion

jirgn:


lockfile.json
```diff
@@ -228,7 +231,7 @@
     "revision": "111b07762e86efab9a918b7c721f720c37e76b0a"
   },
   "verilog": {
-    "revision": "8f6b1f357d1231c420404b5f7a368a73c25adfa2"
+    "revision": "6fae7414fa854b5052bee9111b200e9137797f3d"
```
theHamsta, line 234: please do not update the verilog parser

tests/query/highlights/fusion/basic.fusion
```diff
@@ -0,0 +1,61 @@
+include: SomeFile.fusion
+//<- include
+//     ^punctation.delimiter
+//       ^uri
```
theHamsta, line 4: is this really an URI? It's kind of. I'm just not very sure about that.

tests/query/highlights/fusion/basic.fusion
```diff
@@ -0,0 +1,61 @@
...
+
+namespace ns = Neos.Fusion.Space
```
theHamsta, line 6: very nice! Tests! It also helps us to see how the highlighting would look like. It looks awesome :tada: !

README.md
```diff
@@ -172,6 +172,7 @@ We are looking for maintainers to add more parsers and to write query files for
 - [x] [fennel](https://github.com/travonted/tree-sitter-fennel) (maintained by @TravonteD)
 - [x] [fish](https://github.com/ram02z/tree-sitter-fish) (maintained by @ram02z)
 - [ ] [fortran](https://github.com/stadelmanma/tree-sitter-fortran)
+- [ ] [fusion](~/Code/tree-sitter/tree-sitter-fusion)
```
theHamsta, line 175: please fix this or let the auto-update do its job.

queries/fusion/highlights.scm
```diff
@@ -0,0 +1,87 @@
...
+
+; identifiers eel
+
+(eel_object_path
+  (eel_path_identifier) @variable)
+
+(eel_object_pair
+  key: (eel_property_name) @property)
+
+(eel_method_name) @function
```
theHamsta, line 11: or `@method`? I can't tell because I don't know fusion.

theHamsta (changes requested):
Looks very good. Just mentioned some smaller details.

lockfile.json
theHamsta: https://github.com/nvim-treesitter/nvim-treesitter/runs/4335974704?check_suite_focus=true#step:10:402

queries/fusion/highlights.scm
```diff
@@ -0,0 +1,87 @@
...
+; ----------
+
+[
+ "("
+ ")"
+ "{"
+ "}"
+ "["
+ "]"
+] @punctation.bracket
```
theHamsta, line 80: I think there's type in `punctuaction` (`u` missing)

lockfile.json
jirgn: ok - i thought it is a bit harsh to do so.
I just used the ./scripts/update_lockfile.sh. I will reset all hashes but fusion

theHamsta: Thank you! And please excuse that update_lockfile.sh does not the right thing automatically. On CI, we use this script with a parameter to ignore verilog.

tests/query/highlights/fusion/basic.fusion
theHamsta: also typo here: ^punctation.delimiter

jirgn: hmm - i would say so, at least it looks like the best fit of the given highlight groups

includes in fusion can be one of three forms
1. files relative to the current file (the given case)
2. globbing patterns relative to the current file `**/*.fusion`
3. full specification with a resource schema lik `resource://SomePackage/Resources/**/*.fusion`, that also could be a globbing

tests/query/highlights/fusion/basic.fusion
jirgn: Thank u - In my opinion it also makes is a lot easier to write the grammar or queries. At least, if you do it the first time like me :)

README.md
jirgn: uups

tests/query/highlights/fusion/basic.fusion
theHamsta: Then it's fine!

queries/fusion/highlights.scm
jirgn: good catch - and a good question too.
I would go with function in this case, but you are right, this is a mismatch with the grammar definition.
I give you maybe some background, to start a discussion.
Eel Functions are always Helper Functions. There is nothing like object function definitions in this language specification.
Behind the scenes, there is an PHP Class, with an Implementation of this Function as Method. Maybe see the [Documentation](https://docs.neos.io/cms/manual/extending-neos-with-php-flow/custom-eel-helpers) for an example.

So in a way it is both a method (in the php implementation) and in my opinion 'only' function when using eel,

tests/query/highlights/fusion/basic.fusion
theHamsta: Also for me when I did the Xth time.

theHamsta:
```
(x) Invalid capture @uri in highlights for fusion.  -> @text.uri
Checking fusion locals
(x) Invalid capture @local.scope in locals for fusion.  -> @scope
(x) Invalid capture @local.definition in locals for fusion.  -> @definition.XXX where XXX is function/var/parameter
```

theHamsta (approved):

PR #2061, opened 2021-11-27
injections(yaml): inject bash on key "run"

theHamsta:
Yaml is often used for CI. This applies an heuristic that strings after a "run" key are bash scripts. Note that `((` from bash were missing (separate PR).

image

before:

image


theHamsta:
Gitlab CI would use `script`
image

kyazdani42 (approved):

PR #2067, opened 2021-11-28
highlights(cpp): add support for concepts

theHamsta:
Requires https://github.com/tree-sitter/tree-sitter-cpp/pull/138


theHamsta:
Forgot that the cuda parser needs to be updated also. There's a scheduled CI job that will probably run soon.

theHamsta:
It's only partial support for concepts. More will follow https://github.com/tree-sitter/tree-sitter-cpp/pull/147

clason (approved):
LGTM, but I barely know Stroustrup C++...

PR #2074, opened 2021-12-01
OCaml use @constant.builtin for built in constants

silenc3r:
That's what most other languages do.

### DIFF
--- a/queries/ocaml/highlights.scm
+++ b/queries/ocaml/highlights.scm
@@ -71,7 +71,9 @@
 ; Constants
 ;----------
 
-[(boolean) (unit)] @constant
+(unit) @constant.builtin
+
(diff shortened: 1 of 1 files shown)


queries/ocaml/highlights.scm
```diff
@@ -71,7 +71,7 @@
 ; Constants
 ;----------
 
-[(boolean) (unit)] @constant
+[(boolean) (unit)] @constant.builtin
```
theHamsta, line 74: This looks good but boolean should be `@boolean`

silenc3r: done

undu (approved):

theHamsta (approved):

PR #2077, opened 2021-12-01
docs: remove advice to use 0.5-compat

theHamsta:
0.5-compat branch was not actively maintained and nvim-treesitter
also runs with stable versions. We should be able to maintain
compatibility with latest stable by fall-back implementations in case
there should be a breaking change.

A lot of people were confused about what branch to use and that there
were only in frequent updates to the 0.5-branch. We can do a 0.6.0 tag
now for everyone not wanting this plugin move too fast.


kyazdani42 (approved):
indeed

PR #2083, opened 2021-12-04
rust(highlights): fix highlighting of char_literal

theHamsta:
Was `@string` instead of `@character`

### DIFF
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -151,11 +151,11 @@
 (float_literal) @float
 
 [
-  (char_literal)
   (raw_string_literal)
   (string_literal)
 ] @string
 (escape_sequence) @string.escape
+(char_literal) @character
 
 
 ;;; Keywords


.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       - name: Compile parsers Windows
         if: matrix.os == 'windows-2022'
         run: |
-          Neovim\\bin\\nvim.exe --headless -c "TSInstallSync all" -c "q"
+          Neovim\\bin\\nvim.exe --headless -c "lua require'nvim-treesitter.install'.prefer_git=false" -c "TSInstallSync all" -c "q"
```
kyazdani42, line 55: seems this is unwanted change ?

theHamsta: thanks for noting! but it was actually necessary to make CI pass. We can try with git but at the moment it's a bit luck (also with curl)

kyazdani42 (approved):

PR #2085, opened 2021-12-04
Add TSInstallAndQuit (parallel installation with post quit)

theHamsta:
Faster CI?

### DIFF
--- a/.github/workflows/check-query-files-and-compilation.yml
+++ b/.github/workflows/check-query-files-and-compilation.yml
@@ -47,12 +47,12 @@ jobs:
       - name: Compile parsers Unix like
         if: matrix.os != 'windows-2022'
         run: |
-          nvim --headless -c "TSInstallSync all" -c "q"
+          nvim --headless -c "TSInstallAndQuit all" -c "sleep 180" -c "q"
 
       - name: Compile parsers Windows
         if: matrix.os == 'windows-2022'
         run: |
-          Neovim\\bin\\nvim.exe --headless -c "TSInstallSync all" -c "q"
+          Neovim\\bin\\nvim.exe --headless -c "lua require'nvim-treesitter.install'.prefer_git=false" -c "TSInstallSync all" -c "q"
 
       - name: Post compile Windows
         if: matrix.os == 'windows-2022'
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -112,12 +112,15 @@ local function onread(handle, is_stderr)
   end
 end
 
-function M.iter_cmd(cmd_list, i, lang, success_message)
+function M.iter_cmd(cmd_list, i, message_scope, success_message, post_work_cmds)
   if i == 1 then
     started_commands = started_commands + 1
   end
   if i == #cmd_list + 1 then
     finished_commands = finished_commands + 1
+    if finished_commands == started_commands and post_work_cmds then
+      M.iter_cmd(post_work_cmds, 1, "post installation", "Finished installation!")
+    end
     return print(get_job_status() .. " " .. success_message)
   end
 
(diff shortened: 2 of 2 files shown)


.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -47,12 +47,12 @@ jobs:
...
         if: matrix.os != 'windows-2022'
         run: |
-          nvim --headless -c "TSInstallSync all" -c "q"
+          nvim --headless -c "TSInstallAndQuit all" -c "sleep 180" -c "q"
 
       - name: Compile parsers Windows
         if: matrix.os == 'windows-2022'
         run: |
-          Neovim\\bin\\nvim.exe --headless -c "TSInstallSync all" -c "q"
+          Neovim\\bin\\nvim.exe  --headless "lua require'nvim-treesitter.install'.prefer_git=false" -c "TSInstallAndQuit all" -c "sleep 180" -c "q"
```
theHamsta, line 55: also curl is faster than git clone of a whole repo

theHamsta: why doesn't Windows use curl even now?

theHamsta: was a typo: `-c` was missing

theHamsta: but installing tsx and typescript in parallel confuses Windows :cry:

theHamsta: and Windows uses too much heap space on CI :smile:

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -47,12 +47,12 @@ jobs:
       - name: Compile parsers Unix like
         if: matrix.os != 'windows-2022'
         run: |
-          nvim --headless -c "TSInstallSync all" -c "q"
+          nvim --headless -c "TSInstallAndQuit all" -c "sleep 180" -c "q"
```
kyazdani42, line 50: isn't the CI going to take too long with this sleep ?

theHamsta: No. Sleep is not executed fully, when quit happens earlier.

.github/workflows/check-query-files-and-compilation.yml
theHamsta: Our problem is that this also seems to happen right now with curl and ocaml

.github/workflows/check-query-files-and-compilation.yml
theHamsta:
```
❯ time nvim --headless -c "TSInstallAndQuit jsdoc comment" -c "sleep 180" -c "q"
[nvim-treesitter] [0/2] Downloading...
[nvim-treesitter] [0/2] Downloading...
[nvim-treesitter] [0/2] Creating temporary directory
[nvim-treesitter] [0/2] Extracting...
[nvim-treesitter] [0/2] Compiling...
[nvim-treesitter] [0/2] Creating temporary directory
[nvim-treesitter] [0/2] Extracting...
[nvim-treesitter] [0/2] Compiling...
[nvim-treesitter] [1/2] Treesitter parser for comment has been installednvim --headless -c "TSInstallAndQuit jsdoc comment" -c "sleep 180" -c "q"
 0,94s user 0,23s system 96% cpu 1,214 total
```

kyazdani42 (approved):

vigoux (approved):

stsewd (approved):

jzelinskie:
Wouldn't a better solution be copying how [packer does this](https://github.com/wbthomason/packer.nvim/blob/db3c3e3379613443d94e677a6ac3f61278e36e47/lua/packer.lua#L327-L330) by creating a "TSInstallComplete" autocommand so that the net result would be ```nvim --headless -c 'autocmd User TSInstallComplete quitall' -c 'TSInstall'```?

vigoux:
I kind of like your idea better @jzelinskie

clason:
If you `git clone`, you should also add `--depth=1` to speed up the clone (especially on older repos with a long history).

theHamsta:
@clason git is not used. We just curl the commit we need. Except on Windows. This would also change that.

You can only `--depth=1` branches while we're using raw commits. Also it is not guaranteed that git checkout will find that commit in case the commit in no longer in the history of any upstream branch.

clason:
superseded by rewrite on `main` branch (which has full async)

PR #2087, opened 2021-12-05
Apply cpp to arduino

thetic:


theHamsta:
Can you tell us which plugin sets the arduino filetype?

theHamsta (approved):

PR #2091, opened 2021-12-06
feat(fusion): update highlight queries and some tests

jirgn:
Hey,
sorry for the inconvenience. I did not expect you update the lockfile regularly. Hope this will not happen often.
I updated some parts of the grammar and this broke compatibility to nvim-treesitter queries.

Is there a way we can link the queries in the grammar repo as default?
relates to #2088

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -69,7 +69,7 @@
     "revision": "f0f2f100952a353e64e26b0fa710b4c296d7af13"
   },
   "fusion": {
-    "revision": "bee23e8728845a651a1d70211310c1247134b801"
+    "revision": "14cc1e5535a7b6469c49c21b4c877d31befa2a29"
   },
   "gdscript": {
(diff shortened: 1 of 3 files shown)


theHamsta:
Can you bump the fusion parser in the lockfile to make it work in case the pipelines fail?

The parsers are updated here. Feel free to download and commit the newest query file from your repo: https://github.com/theHamsta/nvim-treesitter/blob/529a37b2574ac8bc6316cf7f605f8b2bdcfe7a4a/.github/workflows/update-parsers-pr.yml#L44

It's also possible to lock fusion that it doesn't get updated regularly but I suppose this is not what you want.

theHamsta:
> Hey, sorry for the inconvenience.

No excuses please. We are very glad that you're contributing!

theHamsta (approved):

PR #2094, opened 2021-12-07
install: use -xc++ -std=c++14 to compile always with c++14

theHamsta:
Let's see whether this works.


theHamsta:
Nope, does not work. Maybe the order of the args would be important but maybe Makefiles are cleaner on Unix.

PR #2104, opened 2021-12-13
injections(rust): inject regex into `Regex::new`

theHamsta:


queries/rust/injections.scm
```diff
@@ -19,3 +19,16 @@
     (#eq? @_html_def "html")
 )
 
+(call_expression
+  function: (scoped_identifier
+    path: (identifier) @_regex (#eq? @_regex "Regex")
```
theHamsta, line 24: `Regex::new`

vigoux (approved):

PR #2112, opened 2021-12-14
Add OpenFOAM parser

FoamScience:
Adding support for the [OpenFOAM](https://openfoam.org/) file format using [this grammar](https://github.com/FoamScience/tree-sitter-foam) which supports:
- highlights
- indents (* basic)
- locals, injections
- folds
- text-objects

### DIFF
--- a/after/ftdetect/foam.vim
+++ b/after/ftdetect/foam.vim
@@ -1,0 +1,13 @@
+" Last Change: 2021 Dec 13
+
+function! s:foamFile(path)
+    let l:lines = getline(1, 10)
+    for line in lines
+        if match(line, 'FoamFile') >= 0
+            set filetype=foam
+        endif
+    endfor
+endfunction
+
+autocmd BufNewFile,BufRead * call s:foamFile(expand("%"))
+autocmd FileType cpp call s:foamFile(expand("%"))
--- a/after/ftdetect/foam.vim
+++ b/after/ftdetect/foam.vim
@@ -9,5 +9,5 @@ function! s:foamFile(path)
     endfor
 endfunction
 
-autocmd BufNewFile,BufRead * call s:foamFile(expand("%"))
+autocmd FileType BufNewFile,BufRead *Dict,*Properties,fvSchemes,fvSolution,*/constant/g,*/0/* call s:foamFile(expand("%"))
 autocmd FileType cpp call s:foamFile(expand("%"))
--- a/after/ftdetect/foam.vim
+++ b/after/ftdetect/foam.vim
@@ -9,5 +9,5 @@ function! s:foamFile(path)
     endfor
 endfunction
 
-autocmd FileType BufNewFile,BufRead *Dict,*Properties,fvSchemes,fvSolution,*/constant/g,*/0/* call s:foamFile(expand("%"))
+autocmd BufNewFile,BufRead *Dict,*Properties,fvSchemes,fvSolution,*/constant/g,*/0/* call s:foamFile(expand("%"))
 autocmd FileType cpp call s:foamFile(expand("%"))
--- a/after/ftdetect/foam.vim
+++ b/after/ftdetect/foam.vim
@@ -1,13 +1,26 @@
-" Last Change: 2021 Dec 13
+" Last Change: 2022 Jan 08
 
 function! s:foamFile(path)
-    let l:lines = getline(1, 10)
+    " Return if file type already set
+    if (&filetype == 'foam')
+        return
+    endif
+
+    let l:lines = getline(1, 15)
+    let l:f = -1
+    let l:o = -1
     for line in lines
-        if match(line, 'FoamFile') >= 0
-            set filetype=foam
+        if (match(line, 'FoamFile') >= 0)
+            let l:f = index(lines, line)
+        endif
+        if (match(line, 'object') >= 0)
+            let l:o = index(lines, line)
         endif
     endfor
+    if ((l:o >= 0) && (l:f) && (l:o > l:f))
+        set filetype=foam
+    endif
 endfunction
 
-autocmd BufNewFile,BufRead *Dict,*Properties,fvSchemes,fvSolution,*/constant/g,*/0/* call s:foamFile(expand("%"))
 autocmd FileType cpp call s:foamFile(expand("%"))
+autocmd BufNewFile,BufRead *Dict,*Properties,fvSchemes,fvSolution,*/constant/g,*/0/* call s:foamFile(expand("%"))
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -846,6 +846,21 @@ list.rasi = {
   maintainers = { "@Fymyte" },
 }
 
+list.foam = {
+  install_info = {
+    url = "https://github.com/FoamScience/tree-sitter-foam",
+    branch = "master",
+    files = {"src/parser.c", "src/scanner.cc"},
+    generate_requires_npm = true,
+  },
+  maintainers = { "@FoamScience" },
+  filetype = "foam",
+  used_by = {"OpenFOAM"},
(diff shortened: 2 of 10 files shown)


queries/foam/textobjects.scm
```diff
@@ -0,0 +1,8 @@
+(dict) @class.outer
+((dict_core) @class.inner)
+((key_value value: _? @_start (_)* _? @parameter.inner)
+    (#make-range! "function.inner" @_start @parameter.inner)) @function.outer
+(code (_)* @class.inner) @class.outer
+((comment) @_start ((comment)+) @_end
+ (#make-range! "comment.outer" @_start @_end))
+(comment) @comment.inner
```
theHamsta, line 8: please submit them to nvim-treesitter-textobjects

queries/foam/highlights.scm
```diff
@@ -0,0 +1,61 @@
...
+  "{"
+  "}"
+  "#{"
+  "#}"
+  ";"
+] @punctuation
+
+;; Special identifiers
+([(identifier) "on" "off" "true" "false" "yes" "no"] @attribute
+(#match? @attribute "^(uniform|non-uniform|and|or|on|off|true|false|yes|no)$")
```
theHamsta, line 60: I don't think that all those are `@attribute`. Rather `@keyword.operator` `@constant.builtin` `@boolean`. Maybe `uniform`, `non-uniform`...

after/ftdetect/foam.vim
```diff
@@ -0,0 +1,13 @@
+" Last Change: 2021 Dec 13
+
+function! s:foamFile(path)
+    let l:lines = getline(1, 10)
+    for line in lines
+        if match(line, 'FoamFile') >= 0
+            set filetype=foam
+        endif
+    endfor
+endfunction
```
theHamsta, line 10: not sure if we want to have such explicit fttype detects. I mean this will run for all users always while virtually nobody uses this format.

queries/foam/textobjects.scm
FoamScience: These made it here by mistake 😄, will push them to the other repo, thanks!

queries/foam/highlights.scm
FoamScience: I tried to stay generic here, as these queries will get used in a few different places; but if you think `@constant.builtin` is a better choice, I'll go for that; as it seems this capture will be widely supported too.

after/ftdetect/foam.vim
FoamScience: Unfortunately, this inherited from the language itself, [this workspace](https://github.com/OpenFOAM/OpenFOAM-dev/tree/master/tutorials/basic/scalarTransportFoam/pitzDaily) shows some examples of files I'm supporting (but we're not limited to those files). Virtually, there is no way to tell if it is an OpenFOAM file unless we take a peek inside.

However, I see your point. Should I limit file type detection to certain filename patterns here, and instruct users of the language to put the content-aware auto-command themselves?

Btw thanks for taking the time to review!

theHamsta: Is there a vim plugin that already handles this filetype? It would be specifically nice when READMEs or similar that talk about OpenFOAM files would not be highlighted as such

theHamsta:
` ./scripts/check-queries.lua:64: (x) Invalid capture @punctuation in highlights for foam.` -> `@punctuation.delimiter`

after/ftdetect/foam.vim
FoamScience: There are a couple of plugins I know of, but they are outdated:
[vim-OpenFoam-syntax](https://github.com/effi/vim-OpenFoam-syntax)
[VimExtensionOpenfoam](https://bitbucket.org/shor-ty/vimextensionopenfoam/src)
[vim-foam](https://github.com/lervag/vim-foam)

and they all handle the file type detection in similar ways, the most feasible is the generic content-based one unfortunately. That's OK there, but not here. That's why I'm using path patterns now; a little better than "*".

queries/foam/injections.scm
```diff
@@ -0,0 +1,9 @@
+;; Pass code blocks to Cpp highlighter
+(code (code_body) @cpp)
+
+;; Pass identifiers to Go highlighter (Cheating I know)
+;;((identifier) @lua)
+
+;; Highlight regex syntax inside literal strings
+((string_literal) @regex)
```
vigoux, line 8: Are all strings in your language regexes ??

FoamScience: Kind of, yes, we're passing all strings to POSIX regex engines internally so I thought why not highlight it as such. Non-regex words are parsed as identifiers.

FoamScience:
@vigoux @theHamsta  Plus, If you're not planning to use c++11 for compiling scanners, please add "foam" to your `ALLOWED_INSTALLATION_FAILURES` in CI workflows, because I'm requiring GCC-9+ to compile it.

theHamsta:
C++11 should work out of the box. If you need C++14, wait for #1795 to be merged.

FoamScience:
> C++11 should work out of the box. If you need C++14, wait for #1795 to be merged.

I don't think It's related to C++14, but on mac-os it's failing (probably outdated compiler?)
Here is what works for me: https://github.com/FoamScience/tree-sitter-foam/blob/master/.github/workflows/main.yml
Note the gcc-9 requirement

FoamScience:
> C++11 should work out of the box. If you need C++14, wait for #1795 to be merged.

Yep, I forgot I needed C++14 😨 Will wait for it to get merged then; thanks!

lua/nvim-treesitter/parsers.lua
```diff
@@ -802,6 +802,21 @@ list.prisma = {
...
 
+list.foam = {
+  install_info = {
+    url = "https://github.com/FoamScience/tree-sitter-foam",
+    branch = "master",
+    files = { "src/parser.c", "src/scanner.cc" },
+    generate_requires_npm = true,
+  },
+  maintainers = { "@FoamScience" },
+  filetype = "foam",
```
theHamsta, line 813: This line is actually not necessary when the parser name is also foam. But it also doesn't do any harm.

theHamsta (approved):

after/ftdetect/foam.vim
theHamsta: Can it happen that I accidentally mention the term `OpenFoam` in a normal C++ file when I'm in a OpenFoam project. Would then C++ still be highlighted properly through the injections?

FoamScience:
@theHamsta, OpenFOAM projects typically don't have standalone C++ files, just small bits inside OpenFOAM files.
But to ask your question; Currently, I'm picking the first 10 lines and see if 'FoamFile' is mentioned there. This does have a slim
chance of misinterpreting other files as OpenFOAM files though and injections wouldn't handle it well because it's called only for specific node types.

OpenFOAM files must have the following header as the first entry (after some comments):
```
FoamFile
{
    version     2.0;
    format      ascii; // binary
    class       dictionary; // or anything else
    location    "constant"; // or anything else
    // More stuff
}
```
I'm lousy at VimScript but would it be more accurate to look for "FoamFile" then "version" then "format"?

FoamScience:
This PR no longer needs to wait for the C++ stuff because my parser now uses a Pure C scanner (following the example of Haskell)
Plus, I have improved setting the file type by looking for tree keywords in a particular order.

theHamsta (approved):

theHamsta:
@FoamScience this needs a rebase. Maybe it should also be a subfiletype of cpp to fallback when someone doesn't have foam installed. Don't know whether we're handling `cpp.foam` right.

FoamScience:
@theHamsta 
I have rebased, thanks for bring it to my attention.
The issue with the filetype is that, in particular, I don't want the LSP client to trigger `clangd` when opening foam files (thinks it's C++ by default just because a first line comment).
Will a sub-type "cpp.foam" trigger it?

PR #2121, opened 2021-12-16
Highlight C enum variants as `@constant`

theHamsta:


stsewd (approved):

theHamsta:
TODO: highlight also identifiers in `case` labels (can only be constants)

vigoux (approved):

vigoux:
Is the To-do for this time or another time?

theHamsta:
Wanted to include in this PR since it was mentioned in the to-fix issue. But can also be done separately.

theHamsta:
Added the TODO and a test.

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -47,12 +47,12 @@ jobs:
...
         if: matrix.os != 'windows-2022'
         run: |
-          nvim --headless -c "TSInstallSync all" -c "q"
+          nvim --headless -c "TSInstallSync! all" -c "q"
 
       - name: Compile parsers Windows
         if: matrix.os == 'windows-2022'
         run: |
-          Neovim\\bin\\nvim.exe --headless -c "TSInstallSync all" -c "q"
+          Neovim\\bin\\nvim.exe --headless -c "TSInstallSync! all" -c "q"
```
theHamsta, line 55: necessary to test our C parser, not the built-in one.

theHamsta: another solution would be to use the "true" runtime file in tests. But the actual C parser would be better.

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -47,12 +47,12 @@ jobs:
...
         if: matrix.os != 'windows-2022'
         run: |
-          nvim --headless -c "TSInstallSync all" -c "q"
+          nvim --headless -c "TSUpdateSync all | TSInstallSync! c" -c "q"
 
       - name: Compile parsers Windows
         if: matrix.os == 'windows-2022'
         run: |
-          Neovim\\bin\\nvim.exe --headless -c "TSInstallSync all" -c "q"
+          Neovim\\bin\\nvim.exe --headless -c "TSUpdateSync all | TSInstallSync! c" -c "q"
```
theHamsta, line 55: Equivalent to `:TSUpdate all` but probably more descriptive. We should probably try to install also the built-in parsers automatically so that they fit to our queries and we could for instance highlight C attributes without conflicting with different built-in parser versions of already released nvim-versions.

theHamsta:
I changed this PR to use the parser that Neovim would select. In this instance, it would use the bundled C parser. The nvim-treesitter C parsers seems to be not installed.

clason:
As far as I know, the bundled parser is purely for the purposes of testing `vim.treesitter`. I don't think there's any fundamental opposition to switch that to another parser.

clason:
But I guess the problem is that Neovim _stable_ uses an earlier version of the same parser?

theHamsta:
@clason this was the problem earlier. E.g. we could not add `[[attribute]]` to C at the time the feature was added because most users would have the bundled parser installed.

nvim-treesitter does not install a parser when it's already installed except you would force it. This would have made the tests fail because we where explicitly searching for the parser in the plugin location. For tests, I previously tried to install our parser in addition and also test with our parser.

clason:
Is it possible for nvim-treesitter to override the bundled parser?

(Alternatively, we could put the bundled parser in a different location that is manually loaded for the tests only.)

theHamsta:
We should just make sure that we install the parser and that our runtime path is before that of the built in parser. But it is indeed a problem, that the parser and the queries get found independent from another and you might end up with a incompatible combination.

PS: tests failed because of a missing `'` in the Lua test.

clason (approved):

theHamsta:
Another problem with bundling the C parser (in case you activate tree-sitter) is that you will have regex highlighting in header files and tree-sitter highlighting in source files. The reason for this is that vim use C++ filetype for header because it can't know whether it's C or C++.

PR #2122, opened 2021-12-16
Add support for new golang `go.work` files

omertuc:
The new golang 1.18 version (currently in beta) [introduced](https://github.com/golang/go/issues/45713) a new file type
called `go.work`.

This commit adds support for the syntax of that file using the https://github.com/omertuc/tree-sitter-go-work repository

That repository is heavily based on previous work in the https://github.com/camdencheek/tree-sitter-go-mod repository, with a few
minor changes to make it work on the very similar `go.work` files.

### DIFF
--- a/queries/gowork/highlights.scm
+++ b/queries/gowork/highlights.scm
@@ -1,0 +1,14 @@
+[
+  "replace"
+  "go"
+  "use"
+] @keyword
+
+"=>" @operator
+
+(comment) @comment
+
+[
+(version)
+(go_version)
+] @string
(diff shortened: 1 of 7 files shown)


omertuc:
cc @camdencheek

camdencheek:
Looks great! I just tested it out locally, and it worked smoothly. Screenshot of the result:
Screen Shot 2021-12-16 at 15 26 21

queries/gowork/highlights.scm
```diff
@@ -0,0 +1,14 @@
+[
+  "replace"
+  "go"
+  "use"
+] @keyword
+
+"=>" @operator
+
+(comment) @comment
```
stsewd, line 9: you could also add this to the injections (for example https://github.com/nvim-treesitter/nvim-treesitter/blob/296c28be5396da069e436f58bdaa541506f99fb5/queries/lua/injections.scm#L53)

stsewd (approved):
Thanks!

queries/gowork/highlights.scm
omertuc: Okay took me a while to understand what this is supposed to do but I think I got it... so there's a "language" called "comment" and you want me to tell nvim-treesitter that any text within a `go.work`-file comment node should be treated as code in a different language, called "comment", so that for example words like TODO, FIXME, etc. would be highlighted?

omertuc: Force pushed, also added injections for gomod

theHamsta (approved):
Nice work!

PR #2129, opened 2021-12-18
highlights(markdown): separate highlight for header marker

mcchrish:


theHamsta (approved):
@stsewd other vim plugins use the same color for title and the markers. Should this be configurable?

clason:
The builtin markdown syntax does not, though: they use
```
markdownHeadingRule -> markdownRule
```

David-Else:
My markdown headers are all the wrong colour since updating to the new parser. Should I wait for a fix, or do I need to submit new highlights to my theme https://github.com/tomasiser/vim-code-dark . If so, what are they? Thanks.

theHamsta:
Could can create a PR to change the default for `@text.title`. I also don't like it, if you have a good suggestion there is always room to improve the defaults and make them more expected. @David-Else also check whether your theme also has `TSTitle` overwritten.

David-Else:
@theHamsta a change of `TSTitle` colour fixed the heading problem, thanks!

queries/markdown/highlights.scm
```diff
@@ -4,6 +4,15 @@
...
 ] @text.title
 
+[
+  (atx_h1_marker)
+  (atx_h2_marker)
+  (atx_h3_marker)
+  (atx_h4_marker)
+  (atx_h5_marker)
+  (atx_h6_marker)
+] @punctuation.delimiter
```
stsewd, line 14: on rst we are using `@punctuation.special`, but any of those work for me.

stsewd:
> Should this be configurable?

Like `@text.title.adornment`? :thinking:

stsewd (commented):
We should update the `(atx_heading) @text.title` query to be `(atx_heading (heading_content) @text.title)`

mcchrish:
Should I convert `setext_heading` to `@punctuation.special` as well?
In the default markdown syntax, it is highlighted separately from the header content.

queries/markdown/highlights.scm
```diff
@@ -1,9 +1,18 @@
 ;; From MDeiml/tree-sitter-markdown
 [
-  (atx_heading)
+  (atx_heading (heading_content))
```
stsewd, line 3: This query needs to be outside the `[..]` to make it work, otherwise it's capturing the whole atx_heading instead of just the heading_content.

stsewd:
> Should I convert setext_heading to @punctuation.special as well?

Yeah, just like the atx_heading, the content should be highlighted as title and the adornments as punctuation.special

stsewd (approved):
Thanks!

David-Else:
Hi, I can now correctly set Markdown header markers using `TSPunctSpecial`. This also sorts out lists and ordered lists. The problem is that this affects TypeScript sytax slightly, now I get `...` the wrong colour:

`was originally white, correct`
Screenshot from 2022-01-07 16-07-28
`after, it is now the colour of the markdown headings and heading markers`
Screenshot from 2022-01-07 16-07-47

What is the correct way to target these independently? Cheers,

clason:
Prefix the language: `markdownTSPunctSpecial`.

PR #2134, opened 2021-12-18
highlights(python): Improve self and cls highlight

Sh3Rm4n:
1. This function simplifies an unnecessary lua-match usage.
2. Fixes an accidental change, where "cls" was changed to "class" (fe4f320b66)
3. ~~Match functions, which have have a decorator.~~
4. ~~But exclude @staticmethod~~
5. Only highlight "cls" as @variable.builtin when the decorator is @classmethod

This might be a bit to complicated and some of this not to useful,
as this existing rule:

```scm
  (#eq? @variable.builtin "self"))
```

overwrites the queries anyway.

**EDIT**: Remove queries, which duplicate highlights of the above query.

### DIFF
--- a/queries/python/highlights.scm
+++ b/queries/python/highlights.scm
@@ -133,7 +133,7 @@
 (none) @constant.builtin
 [(true) (false)] @boolean
 ((identifier) @variable.builtin
- (#lua-match? @variable.builtin "^self$"))
+ (#eq? @variable.builtin "self"))
 
 (integer) @number
 (float) @float
@@ -275,12 +275,15 @@
       name: (identifier) @constructor)))
  (#any-of? @constructor "__new__" "__init__"))
 
-; First parameter of a method is self or cls.
+; First parameter of a classmethod is cls.
 ((class_definition
   body: (block
-          (function_definition
-            parameters: (parameters . (identifier) @variable.builtin))))
- (#any-of? @variable.builtin "self" "obj" "class"))
+          (decorated_definition
+            (decorator (identifier) @_decorator)
+            definition: (function_definition
+              parameters: (parameters . (identifier) @variable.builtin)))))
+ (#eq? @variable.builtin "cls")
+ (#eq? @_decorator "classmethod"))
 
 ;; Error
 (ERROR) @error


queries/python/highlights.scm
```diff
@@ -276,11 +276,27 @@
...
+ (#eq? @variable.builtin "cls")
+ (#eq? @_classmethod "classmethod"))
+((class_definition
+  body: (block
+    (decorated_definition
+      (decorator (identifier) @_not_staticmethod)
+        definition: (function_definition
+          parameters: (parameters . (identifier) @variable.builtin)))))
+ (#eq? @variable.builtin "self")
+ (#not-eq? @_not_staticmethod "staticmethod"))
```
theHamsta, line 294: Is this really necessary? Are there people who use the variable name self in staticmethods?

theHamsta (approved):

queries/python/highlights.scm
Sh3Rm4n: Really necessary? No. I mean, self in python is just a convention anyway. But self as an implicit argument to the method does not exist for `@staticmethod` and if 

https://github.com/nvim-treesitter/nvim-treesitter/blob/b48f9b039ba2b52aa9677012f3882056e83202b0/queries/python/highlights.scm#L135-L136

would not exist, this would be reflected by the highlight. 

But I'm fine removing that part from the query to reduce maintenance burden.

theHamsta: No, it's ok. I just wanted to ask about your motivation

queries/python/highlights.scm
```diff
@@ -276,11 +276,27 @@
...
 
 ; First parameter of a method is self or cls.
+((class_definition
+  body: (block
+    (decorated_definition
+      (decorator (identifier) @_classmethod)
+        definition: (function_definition
+          parameters: (parameters . (identifier) @variable.builtin)))))
+ (#eq? @variable.builtin "cls")
+ (#eq? @_classmethod "classmethod"))
```
stsewd, line 286:
```suggestion
((class_definition
  body: (block
    (decorated_definition
      (decorator (identifier) @_decorator)
        definition: (function_definition
          parameters: (parameters . (identifier) @variable.builtin)))))
 (#eq? @variable.builtin "cls")
 (#eq? @_decorator "classmethod"))
```

queries/python/highlights.scm
stsewd, line 294:
```suggestion
      (decorator (identifier) @_decorator)
        definition: (function_definition
          parameters: (parameters . (identifier) @variable.builtin)))))
 (#eq? @variable.builtin "self")
 (#not-eq? @_decorator "staticmethod"))
```

stsewd (commented):
I'm suggesting renaming the capture, but looks good otherwise. Also, curious if you have been dealing with code that makes use of these variables in the wrong way or just to make the queries more specific? If not, I'd prefer to keep the queries simple.

Sh3Rm4n:
> Also, curious if you have been dealing with code that makes use of these variables in the wrong way or just to make the queries more specific? If not, I'd prefer to keep the queries simple.

No, I was not dealing with code violating the convention, I was just curious how specific I could make these queries. I'll remove the `staticmethod` special case. Not sure if I should keep or remove the `classmethod` query, though.

queries/python/highlights.scm
```diff
@@ -133,7 +133,7 @@
 (none) @constant.builtin
 [(true) (false)] @boolean
 ((identifier) @variable.builtin
- (#lua-match? @variable.builtin "^self$"))
+ (#eq? @variable.builtin "self"))
```
stsewd, line 136: Looks like this query is already highlighting self everywhere, so we don't need the other queries that make reference to self

Screenshot from 2021-12-18 14-23-45

as you can see there, the capture is duplicated.

Sh3Rm4n: True, though this `self` highlight is a bit to broad, as it highlight every instance of `self`, even in free standing functions or the global scope. 

But there is no way around it because of limitations of the query system, AFAIK. So you'd suggest removing these lines?

https://github.com/nvim-treesitter/nvim-treesitter/blob/7cd2ce461b4d2e055d6d7a4063476a540f51422d/queries/python/highlights.scm#L287-L297

stsewd: yeah, since they don't really change anything.

Sh3Rm4n: Done.

stsewd (approved):
Thanks!

PR #2170, opened 2021-12-28
add rescript in ftdetect

ri7nz:


theHamsta:
See #2259

theHamsta:
Do we even have a rescript parser?

ri7nz:
@theHamsta  I'm use with this https://github.com/nkrkv/tree-sitter-rescript/commits/main

theHamsta:
Can you create a PR for vim to recognize this file type and and a PR here to support the parser?

ri7nz:
Okay 👍  Where's the repository for PR to vim ? 

For the parser, I'm new in `treesitter` stuff, could you give me some resource such a docs for me to read?

theHamsta:
@ri7nz GitHub.com/vim/vim. Just have a look at parsers.lua. It should be apparent from the other entries which meta data is necessary to install the parser. Infos about queries are in contributing.md but queries can be a separate PR.

ri7nz:
@theHamsta  I will make a PR later, thank you for insight.

PR #2174, opened 2021-12-29
improve markdown highlighting

pwntester:
This PR improves the markdown highlighting by assigning TSPunctDelimiter group to the brackets used in links and fenced blocks. It also add support for yaml frontmatter

### DIFF
--- a/queries/markdown/injections.scm
+++ b/queries/markdown/injections.scm
@@ -4,3 +4,5 @@
 
 ((html_block) @html)
 ((html_tag) @html)
+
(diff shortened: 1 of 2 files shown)


queries/markdown/injections.scm
```diff
@@ -4,3 +4,5 @@
 
 ((html_block) @html)
 ((html_tag) @html)
+
```
theHamsta, line 7: Why combined? Isn't there only one front matter?

theHamsta (approved):

PR #2175, opened 2021-12-29
feat(fusion): add indent and fold module

jirgn:
added some queries for 
* fold
* indent
* better highlight
I also removed the experimental flag, because to me it feels initially 'complete' now

### DIFF
--- a/queries/fusion/locals.scm
+++ b/queries/fusion/locals.scm
@@ -18,4 +18,4 @@
 
 (eel_parameter) @definition.parameter
 (eel_object_pair
-  key: (eel_propery_name) @definition.field)
+  key: (eel_property_name) @definition.field)
--- a/queries/fusion/locals.scm
+++ b/queries/fusion/locals.scm
@@ -1,3 +1,21 @@
+;; Fusion base
+(block) @scope
+
+(namespace_declaration
+  (alias_namespace) @definition.namespace)
+
+(property
+  (path (path_part) @definition.field))
+
+(type
+  namespace: (package_name)? @definition.namespace
+  name: (type_name) @definition.type
+)
+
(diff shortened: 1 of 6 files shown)


queries/fusion/locals.scm
```diff
@@ -1,3 +1,21 @@
+;; Fusion base
```
jirgn, line 1: These are all 'blind' definitions. I do not know how to test a resulting behavior, so i can validate if these definitions are right.
Do you have any advice or hint how to test or do any qa?

theHamsta: you can create some tests similar to https://github.com/theHamsta/nvim-treesitter/blob/2bf58527049185347308300dff5184871bd99764/tests/query/highlights_spec.lua#L26

 Theses could be general query tests where you just check here https://github.com/theHamsta/nvim-treesitter/blob/2bf58527049185347308300dff5184871bd99764/tests/query/highlights_spec.lua#L61-L69 whether assertion == capture . If you want to go more fancy you could even call `locals.find_definition(node_at_point, bufnr)` to find the definition node and check whether it matches the assertion (e.g. use `reference.1` on all the references and `definition.1` on the definition.

theHamsta (approved):

queries/fusion/locals.scm
jirgn: Thanx for the clarification.
this sounds like a good option. I will give it a try. When I have time I will make a lua crash course and give it a try.

great review as always.
all the best and a happy new year

theHamsta: Happy new year and thanks for the contribution! If you want to I can also attach this feature to the injection tests that got some how stuck in review as it does a similar task.

PR #2190, opened 2022-01-05
feat: Add detection for jsonc

yamatsum:
Add detection for jsonc

### DIFF
--- a/ftdetect/jsonc.vim
+++ b/ftdetect/jsonc.vim
@@ -1,0 +1,2 @@
+autocmd BufNewFile,BufRead *.jsonc set ft=jsonc
+autocmd BufNewFile,BufRead tsconfig.json set ft=jsonc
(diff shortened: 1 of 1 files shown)


ftdetect/jsonc.vim
```diff
@@ -0,0 +1,2 @@
+autocmd BufNewFile,BufRead *.jsonc set ft=jsonc
```
theHamsta, line 1: upstream already has this:
```
runtime/filetype.vim
889:au BufNewFile,BufRead *.jsonc                       setf jsonc
```

yamatsum: I didn't know that. I was referring to the following: https://github.com/neoclide/jsonc.vim/blob/master/ftdetect/jsonc.vim

theHamsta:
This is probably a important contribution to upstream Vim or neovim's new filetype.lua. we're trying to remove all ftdetct plugins from this repo.

PR #2193, opened 2022-01-05
Adds extra args to utils.setup_commands to add custom func args

Badhi:
As a user of the nvim-treesitter I wanted to add a command through `nvim-treesitter.utils.setup_commands` that receives the ranges like below

```
command! -range MyFun :lua require'..'.myFun(<line1>, <line2>)
```

Current implementation limits the `myFun` arguments to be only `<f-args>`.

This PR adds an extra argument to `nvim-treesitter.utils.setup_commands` - `f_args`, so that the user can add their own required argument types to their functions

### DIFF
--- a/lua/nvim-treesitter/utils.lua
+++ b/lua/nvim-treesitter/utils.lua
@@ -17,7 +17,7 @@ end
 ---
 ---@param mod string, Name of the module that resides in the heirarchy - nvim-treesitter.module
 ---@param commands table, Command list for the module
----         _ {command_name} Name of the vim user defined command, Keys:
+---         - {command_name} Name of the vim user defined command, Keys:
 ---             - {run}: (function) callback function that needs to be executed
 ---             - {f_args}: (string, default <f-args>)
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):
could you write this somehow in the documentation :) ?

theHamsta (approved):

Badhi:
I just took a look at the help doc to document the function. But now I'm wondering why there is a `ts_util` module and a `util` module both. All the `ts_util` functions are listed down under utils heading and if I list the functions in the `utils` module wouldn't that become confusing?

theHamsta:
ts_utils are all functions related to tree-sitter nodes an parsers. `utils` is just general house-keeping. You're right, they should be separated in the documentation. I started a draft to use Neovim's doc generator script to generate the module documentation directly from the source code which would prevent such mistakes.

Badhi:
In that case, I can add a comment block above the function. Will it be that enough to get included in the doc?

PR #2197, opened 2022-01-05
fix(haskell): Haskell switched to C scanner

theHamsta:


theHamsta:
Will merge this after passing of pipeline since master is broken (Haskell was allowed to fail).

PR #2198, opened 2022-01-05
Update tree-sitter-haskell

3kyro:
tree-sitter-haskell has recently moved to a new  c based scanner, making installs using `:TSinstall` fail.


3kyro:
looks like #2197 is already in place :)

PR #2199, opened 2022-01-05
Change Kotlin parser

SalBakraa:
Change the Kotlin parser used by nvim-treesitter from [Joakker/tree-sitter-kotlin](https://github.com/Joakker/tree-sitter-kotlin) to [fwcd/tree-sitter-kotlin](https://github.com/fwcd/tree-sitter-kotlin), since the latter parser appears to be more comprehensive and still in-development. Furthermore,  the parser lacked a locals and folds query so these have been added.

Comparison between the two parsers:

comparison

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [json5](https://github.com/Joakker/tree-sitter-json5) (maintained by @Joakker)
 - [x] [JSON with comments](https://gitlab.com/WhyNotHugo/tree-sitter-jsonc.git) (maintained by @WhyNotHugo)
 - [x] [julia](https://github.com/tree-sitter/tree-sitter-julia) (maintained by @mroavi, @theHamsta)
-- [x] [kotlin](https://github.com/Joakker/tree-sitter-kotlin) (maintained by @Joakker)
+- [x] [kotlin](https://github.com/fwcd/tree-sitter-kotlin) (maintained by @fwcd)
 - [x] [latex](https://github.com/latex-lsp/tree-sitter-latex) (maintained by @theHamsta by asking @clason)
 - [x] [ledger](https://github.com/cbarrete/tree-sitter-ledger) (maintained by @cbarrete)
 - [x] [llvm](https://github.com/benwilliamgraham/tree-sitter-llvm) (maintained by @benwilliamgraham)
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [json5](https://github.com/Joakker/tree-sitter-json5) (maintained by @Joakker)
 - [x] [JSON with comments](https://gitlab.com/WhyNotHugo/tree-sitter-jsonc.git) (maintained by @WhyNotHugo)
 - [x] [julia](https://github.com/tree-sitter/tree-sitter-julia) (maintained by @mroavi, @theHamsta)
-- [x] [kotlin](https://github.com/fwcd/tree-sitter-kotlin) (maintained by @fwcd)
+- [x] [kotlin](https://github.com/fwcd/tree-sitter-kotlin) (maintained by @SalBakraa)
 - [x] [latex](https://github.com/latex-lsp/tree-sitter-latex) (maintained by @theHamsta by asking @clason)
 - [x] [ledger](https://github.com/cbarrete/tree-sitter-ledger) (maintained by @cbarrete)
 - [x] [llvm](https://github.com/benwilliamgraham/tree-sitter-llvm) (maintained by @benwilliamgraham)
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -258,7 +258,7 @@ list.kotlin = {
     branch = "main",
     files = { "src/parser.c", "src/scanner.c" },
   },
-  maintainers = { "@fwcd" },
+  maintainers = { "@SalBakraa" },
 }
 
 list.html = {
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -254,10 +254,11 @@ list.java = {
 
 list.kotlin = {
   install_info = {
-    url = "https://github.com/Joakker/tree-sitter-kotlin",
-    files = { "src/parser.c" },
+    url = "https://github.com/fwcd/tree-sitter-kotlin",
+    branch = "main",
+    files = { "src/parser.c", "src/scanner.c" },
   },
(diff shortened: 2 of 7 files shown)


theHamsta:
Well, we changed the parser because fwcd and one fork of it took incredible long to build and to load. fwcd was missing some features of the fork. We had the hope that the new parser would catch up quickly.

@Joakker

theHamsta:
Wow, cool. Didn't know that fwcd was reactivated...

SalBakraa:
> Well, we changed the parser because fwcd and one fork of it took incredible long to build and to load.

Aren't the parsers compiled beforehand? All the tree-sitter repos I saw committed the generated parser code. 

> fwcd was missing some features of the fork.

But it still had more features than the one that was being used as default.

Also sorry about the query errors, I was too accustomed to neovim spitting errors when I was made a mistake in a query that I didn't think to run the `check-queries` script. sorry.

theHamsta:
Well, now it seems to be ok 3.6MB and loading also seems to be ok. We switched the old parser when its loading time increased by factor of 10 (from tree-sitter 0.18.0 to 0.19) taking several seconds to load the queries. But apparently this is fixed now, maybe also due two recent fixes.

SalBakraa:
```
(import_header
	(identifier
		(simple_identifier) @type @import)
	(import_alias
		(type_identifier) @type)?
		(#lua-match? @import "^[A-Z]"))
```

Why is `@import` here not allowed? neovim appears to process the query correctly and it only being used for the predicate.

SalBakraa:
> We switched the old parser when its loading time increased by factor of 10 (from tree-sitter 0.18.0 to 0.19) taking several seconds to load the queries. But apparently this is fixed now, maybe also due two recent fixes.

Did the loading time change because of the amount of nodes the parser generates? It seems weird that the parser would affect the speed loading the queries.

lua/nvim-treesitter/parsers.lua
```diff
@@ -254,10 +254,11 @@ list.java = {
...
 list.kotlin = {
   install_info = {
-    url = "https://github.com/Joakker/tree-sitter-kotlin",
-    files = { "src/parser.c" },
+    url = "https://github.com/fwcd/tree-sitter-kotlin",
+    branch = "main",
+    files = { "src/parser.c", "src/scanner.c" },
   },
-  maintainers = { "@Joakker" },
+  maintainers = { "@fwcd" },
```
theHamsta, line 261: you would be the maintainer. it's not about maintenance of the upstream repo but its integration into this repo

README.md
```diff
@@ -201,7 +201,7 @@ We are looking for maintainers to add more parsers and to write query files for
 - [x] [json5](https://github.com/Joakker/tree-sitter-json5) (maintained by @Joakker)
 - [x] [JSON with comments](https://gitlab.com/WhyNotHugo/tree-sitter-jsonc.git) (maintained by @WhyNotHugo)
 - [x] [julia](https://github.com/tree-sitter/tree-sitter-julia) (maintained by @mroavi, @theHamsta)
-- [x] [kotlin](https://github.com/Joakker/tree-sitter-kotlin) (maintained by @Joakker)
+- [x] [kotlin](https://github.com/fwcd/tree-sitter-kotlin) (maintained by @fwcd)
```
theHamsta, line 204: looks good to me to use the parser that was also used in Kotlin language server. I'd like to hear the opinion of @Joakker

theHamsta (approved):
Looks good to me to use parser that was also used in Kotlin languages server with your fixes. I'd like to hear the opinion of

lua/nvim-treesitter/parsers.lua
SalBakraa: Sorry about that, I misunderstood.

theHamsta:
I think the fwcd parser is far more mainstream. Do we want to add a repo key to the lockfile to not make parser installation crash on TSUpdate? At the moment of update users will still have the repo url of the old parser. We could also try to re-source parsers.lua (deleting it from package.loaded and `require`ing it again or put the parser list into separate file.

theHamsta:
The extremely slow loading problem persists on my machine. Will benchmark later again on the upstream repo. The slow load times were @SalBakraa were the original reason why we switched parsers. The tree-sitter team is actively working on improving query loading performance

SalBakraa:
> The slow load times were @SalBakraa were the original reason why we switched parsers.
I get that slow loads are an issue, but changing the parser isn't really the fix here. Unless the parser that is is being switched to fixes the issue while maintaining the same level of parsing capabilities. While @Joakker's kotlin-parser loaded much faster, it lacked support for a majority of kotlin's grammar constructs. It was very frustrating that inserting a valid kotlin statement broke the highlighting for 1 to 2 lines below what was inserted.

If the slow loads are a major issue, I believe it would be better to disable the parser and put a disclaimer that this parser loads slowly when editing kotlin files.

SalBakraa:
Also, I would love to work on the fwcd parser's grammar to improve load times, but I don't know what is causing the bottleneck.

theHamsta:
> I get that slow loads are an issue, but changing the parser isn't really the fix here. Unless the parser that is is being switched to fixes the issue while maintaining the same level of parsing capabilities.

Well, it was actually is a fix. It is much easier to start a new parser than to debug the issues of an old one that grew very huge.

>  lacked support for a majority of kotlin's grammar constructs. 

So did the main stream parser and it didn't have any commits for two years.

Some context for recent attempts to fix query performance https://github.com/tree-sitter/tree-sitter/pull/1578 https://github.com/tree-sitter/tree-sitter/issues/1510 Note, that the bump from tree-sitter 0.18 -> 0.19 caused 10x loading time increase for the Kotlin parser after which it was no longer usable. The actual issue should be fixed in tree-sitter.

SalBakraa:
> So did the main stream parser and it didn't have any commits for two years.

I apologize about that, I haven't really checked the history of the fwcd parser and just went by the fact that it was the parser linked on https://tree-sitter.github.io and it was more feature complete.

> Well, it was actually is a fix. It is much easier to start a new parser than to debug the issues of an old one that grew very huge.

While that is true, reaching parity between the parsers should have been the goal before switching.

maxbrunsfeld:
@theHamsta Could you (or anyone else who is following this issue) provide a minimal query that reproduces the query-loading performance problem for the Kotlin grammar? You could open a new issue on the Tree-sitter repo if that works for you. Thank you!

theHamsta:
The issue is actually quite old, a lot of might have changed since then: https://github.com/tree-sitter/tree-sitter/issues/973#issuecomment-810480327 I did not do a recent benchmark.

README.md
Joakker: Sorry I haven't been on. Life's been rough lately. If it's the same parser as the kotlin langserver, then it's fine by me

PR #2218, opened 2022-01-08
Fix get_node_text

MDeiml:
I noticed a small error with `get_ndoe_text` in `ts_utils.lua`. Whenever the range of the node is until the end of the file, the last line will be completely truncated.

This is because the last line returned by `nvim_buf_get_lines` is not actually the line with index `end_row` since the biggest line index is `end_row - 1`.

### DIFF
--- a/lua/nvim-treesitter/ts_utils.lua
+++ b/lua/nvim-treesitter/ts_utils.lua
@@ -21,7 +21,10 @@ function M.get_node_text(node, bufnr)
   if start_row ~= end_row then
     local lines = api.nvim_buf_get_lines(bufnr, start_row, end_row + 1, false)
     lines[1] = string.sub(lines[1], start_col + 1)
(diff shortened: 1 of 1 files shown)


theHamsta:
Is this also in core?

MDeiml:
I'm not sure I know what you mean with core?

MDeiml:
Oh, I see what you mean. I didn't know there was a `get_node_text` in neovim core. Looking at the code it does not seem to have the same problem.

theHamsta (approved):

mizlan:
This is problematic. When a node spans multiple lines but _doesn't_ end after the end of the file, it will not be truncated when it should. See https://github.com/mizlan/iswap.nvim/issues/43 for context. Because I'm unsure of what the conditional is supposed to check, I cannot make any guarantees as to what is supposed to actually occur. All I know is the condition intuitively doesn't make sense; for instance consider a node that spans two lines. I've opened a small PR for this: #2437

PR #2223, opened 2022-01-09
Reword Neovim version requirement from README

kidonng:
[Neovim 0.5](https://github.com/neovim/neovim/commit/a5ac2f45ff84a688a09479f357a9909d5b914294) is shipped with experimental tree-sitter support, this should be reflected in the README as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ For more detailed information on setting these up, see ["Advanced setup"](#advan
 
 ## Requirements
 
-- Neovim 0.5 or above
+- Neovim latest stable version or [nightly](https://github.com/neovim/neovim#install-from-source)
 - `tar` and `curl` in your path (or alternatively `git`)
 - A C compiler in your path and libstdc++ installed ([Windows users please read this!](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Windows-support)).
 
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ More examples can be found in [our gallery](https://github.com/nvim-treesitter/n
 
 **Warning: Treesitter and nvim-treesitter highlighting are an experimental feature of Neovim.
 Please consider the experience with this plug-in as experimental until Tree-Sitter support in Neovim is stable!
-We recommend using the latest version of Neovim.
+We recommend using the nightly builds of Neovim if possible.
 You can find the current roadmap [here](https://github.com/nvim-treesitter/nvim-treesitter/projects/1).
 The roadmap and all features of this plugin are open to change, and any suggestion will be highly appreciated!**
 
--- a/README.md
+++ b/README.md
@@ -36,9 +36,9 @@ The goal of `nvim-treesitter` is both to provide a simple and easy way to use th
 Traditional highlighting (left) vs Treesitter-based highlighting (right).
 More examples can be found in [our gallery](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Gallery).
 
-**Warning: Treesitter and nvim-treesitter highlighting are an experimental feature of nightly versions of Neovim.
+**Warning: Treesitter and nvim-treesitter highlighting are an experimental feature of Neovim.
 Please consider the experience with this plug-in as experimental until Tree-Sitter support in Neovim is stable!
-We recommend using the nightly builds of Neovim or the latest stable version.
+We recommend using the latest version of Neovim.
 You can find the current roadmap [here](https://github.com/nvim-treesitter/nvim-treesitter/projects/1).
 The roadmap and all features of this plugin are open to change, and any suggestion will be highly appreciated!**
 
@@ -63,7 +63,7 @@ For more detailed information on setting these up, see ["Advanced setup"](#advan
 
 ## Requirements
 
-- Neovim [nightly](https://github.com/neovim/neovim#install-from-source)
+- Neovim 0.5 or above
 - `tar` and `curl` in your path (or alternatively `git`)
 - A C compiler in your path and libstdc++ installed ([Windows users please read this!](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Windows-support)).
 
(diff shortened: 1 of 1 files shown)


clason:
The wording can be improved, but support is in fact still experimental even on nightly (0.7.0-dev)! More importantly, this plugin may still target specific features that are only introduced on master, in which case the latest stable version will _not_ work! And even if it does work, there may be improvements in the nightlies that are very desirable (like the recent tree-sitter bump that included a mitigation of a huge performance problem).

So I still strongly recommend staying with nightly if you are interested in tree-sitter.

theHamsta:
It shouldn't make any difference for end users whether they use 0.5.0 or nightly. I don't plan to do breaking changes that would affect latest stable releases. The wording that nightly is recommended seem just right to me.

theHamsta (approved):

clason:
I don't quite understand -- the PR removes the recommendation for nightly completely?

kidonng:
> More importantly, this plugin may still target specific features that are only introduced on master, in which case the latest stable version will _not_ work!

That's unfortunate, but given that v0.5 was shipped half a year ago, this plugin is probably facing more users using stable than before, so breaking changes should be considered thoroughly.

> I don't plan to do breaking changes that would affect latest stable releases.

Thank you! I bet most users would expect it that way.

> I don't quite understand -- the PR removes the recommendation for nightly completely?

Note that the warning is still there, I just updated the _requirements_. I reworded a bit to make _recommendations_ of nightly.

vigoux:
To be pretty frank, I strongly disagree with such a change.

The reason is that tree-sitter is not stable upstream, and saying that we recommend a version on neovim, as if tree-sitter was stable at this version, is really misleading.

I know that, as my involvement in the project lowered these days, my opinion is less important, but I want to say I am strongly against this.

clason:
Maybe I should add a few words to my drive-by comment. One thing is factual accuracy: currently (and in the short-term), nvim-treesitter works with a stable release (although I'd like to point out that no stable release contains the tree-sitter performance mitigation, which seems to be a major issue to some). One concern is also that this locks you out from exploiting possible nightly features that are massively beneficial (you never know when bfredl drops another huge rewrite that makes life better ;))

But another, equally important, issue is expectation management: Too often I see people installing neovim, blindly enabling all tree-sitter features, and then complaining that **Neovim** is slower, or has worse highlighting, than Vim. I believe that anything that helps people get out of that mindset and towards using tree-sitter in more aware and selective fashion is helpful to the project.

kidonng:
I don't have a word in saying how this project should evolve, but in my humble opinions the changes are fairly appropriate. Let me go through them line by line:

```diff
- **Warning: Treesitter and nvim-treesitter highlighting are an experimental feature of nightly versions of Neovim.
+ **Warning: Treesitter and nvim-treesitter highlighting are an experimental feature of Neovim.
```

There's no imperative need to emphasize "nightly versions" here, because stable versions have it as well.

```diff
- We recommend using the nightly builds of Neovim or the latest stable version.
+ We recommend using the nightly builds of Neovim if possible.
```

The recommendation is still there, and even further emphasized. "latest stable version" is implied.

```diff
## Requirements

- - Neovim [nightly](https://github.com/neovim/neovim#install-from-source)
+ - Neovim 0.5 or above
```

Do note this is a _requirement_, not a _recommendation_, saying it only requires Neovim 0.5 or above is not wrong.

```diff
#### I get `module 'vim.treesitter.query' not found`

- Make sure you have the latest nightly version of Neovim.
+ Make sure you have the latest version of Neovim.
```

If people has the latest version (whether stable or nightly), they won't get the error, hence there's no need to emphasize "nightly versions" here either.

clason:
Yes, the current changes are much less contentious.

The only issue I see is that "Neovim 0.5 or above" is _currently_ correct, but the project might want to reserve the possibility to use nightly features at any time (if they're worth it).

Basically, there is a difference of _guaranteeing_ that nvim-treesitter will work with Neovim 0.5, and not breaking that intentionally ("it should work, but you're on your own"). 

The point here (that I keep repeating) is that Neovim itself only supports exactly two versions:
1. the last stable release (currently 0.6.1);
2. the _latest_ commit on master.
I wouldn't recommend plugins to extend their support beyond what upstream promises. (In your version, the nvim-treesitter maintainers now need to make sure the plugin works on 0.5.0, 0.5.1, 0.6.0, _and_ 0.6.1!)

kidonng:
> In your version, the nvim-treesitter maintainers now need to make sure the plugin works on 0.5.0, 0.5.1, 0.6.0, _and_ 0.6.1, where before your change it was only 0.6.1!

I don't mean it that way, but perhaps it's implied so.

I changed "0.5 or above" to "latest stable version or nightly", that should be remove the last blocker right?

clason:
Yeah, I'm sure that wasn't intended; I was just explaining why I'm riding so hard on such apparently minor reformulations.

The current version seems fine to me, and an improvement.

theHamsta:
I'm sure that all are most or less on the same page. This plugin targets: Neovim nightly and the latest stable release. It will always be an adapor to how tree-sitter is indented to be used in it's current implemention. I would not make any guarantees to users using not the latest stable version but I'm sure that for users the interface to this plugin to stable versions of neovim will be mostly `
"Make highlighting work" which gives us a lot of flexibility especially since it abstracts over changes in core until they become stable enough to be used directly. Latest stable+nightly seems a god compromise to me as it gives us large base of contributors and testers.

So the requirement should be latest stable version with nightly recommended. Maybe you can refine the wording in this PR. People will always use this plugin on stable versions of neovim as long as it works. I'd prefer having those people using the same queries as nightly followers. You should always consider nvim-treesitter as an interface to a to-be-made stable API to tree-sitter but as such I think we can add latest stable as a target for this plugin as we're giving no guarantees on how our `setup` call is translated to real tree-sitter experience.

I'd also be fine to recommend nightly with a mention that stable works in most cases without any guarantees.

kidonng:
Thanks for all the inputs! The current changes are as far as I can get.

kidonng:
Thanks for all the inputs! The current changes are as far as I can get.

clason (approved):

ysangkok:
It seems like Neovim has supported this feature for a while, is is it really still recommended to use nightly? Because I still see that recommendation in the readme.

Issue #2224, opened 2022-01-09
The hpp file is detected as a foam file.


rhcher:
### Before reporting

- [X] I have updated my neovim version to latest _master_
- [X] I have updated my plugin to the latest version
- [X] I have run `:TSUpdate`
- [X] I have read the [troubleshooting section](https://github.com/nvim-treesitter/nvim-treesitter#troubleshooting)

### Describe the bug

The hpp file with some special context is detected as a foam file.

### To Reproduce

1. open hpp file I provide.
2. `:set ft?`
![image](https://user-images.githubusercontent.com/50723480/148678193-4205df33-146a-4cb6-862f-53ca7ea7ff4a.png)
3. the language server works properly, so it't not a filetype.lua issue.

### Expected behavior

The filetype is not detected as foam.

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found  0.20.2 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v17.3.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Sele
```

### Output of `nvim --version`

```text
NVIM v0.7.0-dev+858-gc7ffe406bc
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -g -Og -g -Wall -Wextra -pedantic -Wno
-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong
 -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_L
OG_LEVEL=3 -I/build/neovim-git/src/build/config -I/build/neovim-git/src/neovim-git/src -I/usr/include -I/build/neovim-git/src/build/src/nvim/auto
 -I/build/neovim-git/src/build/include
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

[...]

PR #2227, opened 2022-01-10
Force-ignore ignorecase option

FoamScience:
Addresses #2224 as suggested there.

### DIFF
--- a/after/ftdetect/foam.vim
+++ b/after/ftdetect/foam.vim
@@ -10,14 +10,14 @@ function! s:foamFile(path)
     let l:f = -1
     let l:o = -1
     for line in lines
-        if (match(line, 'FoamFile') >= 0)
+        if (match(line, '\CFoamFile') >= 0)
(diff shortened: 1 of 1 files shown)


after/ftdetect/foam.vim
```diff
@@ -10,10 +10,10 @@ function! s:foamFile(path)
...
     for line in lines
-        if (match(line, 'FoamFile') >= 0)
+        if (match(line, '\CFoamFile') >= 0)
             let l:f = index(lines, line)
         endif
-        if (match(line, 'object') >= 0)
+        if (match(line, '\Cobject') >= 0)
             let l:o = index(lines, line)
         endif
     endfor
```
theHamsta, line 19: Isn't the problem here that, any file with `object` is detected as foam file and it should be:
```
f ((l:o >= 0) && (l:f > 0) && (l:o > l:f))
```

theHamsta (approved):

PR #2234, opened 2022-01-11
Remove ftdetect for filetypes supported in Neovim core

### LINKED ISSUE #2231: Remove filetype detection
I updated my `nvim-treesitter` plugin, and suddenly some files were detected as a different filetype, this was very surprising. In my case, files that previously was detected as `mason` now got detected as `glsl`. If I disable the nvim-treesitter plugin, the files are detected as `mason` again.

Some way to disable treesitter from changing with the filetype or control how it's done

Creating my own `autocmd` doesn't seem to take affect after treesitter.
Setting the filetype with `setf` will be reverted back when reloading the file `:e`.

theHamsta:
- remove all ftdetect plugins that are officially supported nightly Neovim or in latest Vim (with PR for Neovim https://github.com/neovim/neovim/pull/17041)
- also reduce GLSL regexes to `*.glsl` fixes #2231 

This is the less radical measure compared to removing all ftdetects. We could also remove them completely.


clason:
I'd suggest opening a tracking issue for the remaining filetypes; it looks like Vim is not averse to adding filetype detection even without full syntax support, and these would be easy PRs to make for people interested in contributing (as long as Vim doesn't get spammed with them).

theHamsta:
> it looks like Vim is not averse to adding filetype detection

I mean, when you add zimbu you have to add everything :slightly_smiling_face:

kaddkaka:
The above test fail was inherited from master branch I think? So rebasing/merging and re-running the test might be succcessful. (Not sure how to do that, if I even can...)

theHamsta:
No, the fail comes from nighlty not being downloadable this day. It can just be ignored.

clason (approved):

PR #2245, opened 2022-01-13
Update lockfile and apply necessary update for Java

theHamsta:
https://github.com/tree-sitter/tree-sitter-java/pull/93 introduced `line_comment` and `block_comment`

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -105,7 +105,7 @@
     "revision": "3cb7fc28247efbcb2973b97e71c78838ad98a583"
   },
   "heex": {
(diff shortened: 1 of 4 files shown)


theHamsta:
This is blocking parser update. We don't want to lose comment highlighting on Java

clason (approved):
LGTM

theHamsta:
@clason small changes can usually be merged directly. We have "auto-merge" enabled so you don't need for the tests to finish. Merge will then be done when required tests pass
.

PR #2249, opened 2022-01-14
fix(php): change query for @constructor group

infectos:
If I understand correctly, the @constructor group should match definition of constructor method.
That way it will highlight name of the method.

### DIFF
--- a/queries/php/highlights.scm
+++ b/queries/php/highlights.scm
@@ -55,8 +55,9 @@
 ((name) @constant
  (#vim-match? @constant "^_?[A-Z][A-Z\d_]+$"))
 
-((name) @constructor
- (#lua-match? @constructor "^[A-Z]"))
+(method_declaration
+    name: (name) @constructor
+    (#eq? @constructor "__construct"))
 
 ((name) @variable.builtin
  (#eq? @variable.builtin "this"))


mikehaertl:
I can confirm that this also fixes the problem with the class name (detected as constructor before).

queries/php/highlights.scm
```diff
@@ -55,8 +55,9 @@
 ((name) @constant
  (#vim-match? @constant "^_?[A-Z][A-Z\d_]+$"))
 
-((name) @constructor
- (#lua-match? @constructor "^[A-Z]"))
+(method_declaration
+    name: (name) @constructor
+    (#eq? @constructor "__construct"))
```
theHamsta, line 60: I guess the query is trying to highlight the case `new BaseClass()` in 
```
class BaseClass {
    function __construct() {
        print "Im BaseClass-Konstruktor\n";
    }
}

class SubClass extends BaseClass {
    function __construct() {
        parent::__construct();
        print "Im SubClass-Konstruktor\n";
    }
}

class OtherSubClass extends BaseClass {
    // erbt den Konstruktor von BaseClass
}

// Im BaseClass-Konstruktor
$obj = new BaseClass();
```
you could swap the queries so that constant has higher priority or just highlight this case in combination with the `new` keyword. But I'd be also fine with this solution.

theHamsta (approved):

queries/php/highlights.scm
mikehaertl: @theHamsta The object creation could be matched by this query:
```
(object_creation_expression (qualified_name) @constructor)
```

infectos: @theHamsta 
Screenshot from 2022-01-15 08-18-24
Screenshot from 2022-01-15 08-18-40
Do we need whole creation expression or just the name?

@mikehaertl
Screenshot from 2022-01-15 08-18-52
(qualified_name) doesn't match. Do you have different results?

mikehaertl: @infectos It can be `name`  (class without path) or `qualified_name` (fully qualified name like `\some\TestClass`). An alternation expression should work:

```
(object_creation_expression
  [(name) (qualified_name)] @constructor)
```

theHamsta: If this is getting to complicated we can move the highlight of object creation as constructor to another PR or omit it entirely. I'd be fine with merging when you are.

mikehaertl: @theHamsta I think you can merge it. I'm not even sure if it's a good idea to mark object creation as `TSConstruct`.

Issue #2207, opened 2022-01-07
@constructor overrides @constant highlighting in php files


infectos:
### Before reporting

- [X] I have updated my neovim version to latest _master_
- [X] I have updated my plugin to the latest version
- [X] I have run `:TSUpdate`
- [X] I have read the [troubleshooting section](https://github.com/nvim-treesitter/nvim-treesitter#troubleshooting)

### Describe the bug

Freshly installed nvim-treesitter with php parser has `@constructor` group which sometime overrides other groups. 
For example I cannot change highlighting for constants because `@constructor` apparently has higher priority.
I am not sure it is supposed to act that way. All constants should be purple, but they are orange.
![Screenshot from 2022-01-07 05-30-21](https://user-images.githubusercontent.com/57968042/148483714-4bb5d88e-77ad-4dd4-b1d6-eb63963d8069.png)

### To Reproduce

1. Enable php parser 
2. Turn on highlighting in php files.
3. `:TSHighlightCapturesUnderCursor `on any constant.

### Expected behavior

I expect constants (declaration and calls) to be highlighted not as `@constructor`, but as `@constant` group.

### Output of `:checkhealth nvim-treesitter`

```markdown
1 nvim-treesitter: require("nvim-treesitter.health").check()
    2 ========================================================================
    3 ## Installation
    4   - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
    5   - OK: `node` found v17.1.0 (only needed for :TSInstallFromGrammar)
    6   - OK: `git` executable found.
    7   - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    8     Version: cc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
    9   - OK: Neovim was compiled with tree-sitter runtime ABI version 13 (required >=13). Parsers must be compatible with runtime ABI.
   10 
   11 ## Parser/Features H L F I J
   12   - php            ✓ ✓ ✓ ✓ ✓ 
   13   - query          ✓ ✓ ✓ ✓ ✓ 
   14   - javascript     ✓ ✓ ✓ ✓ ✓ 
   15   - ruby           ✓ ✓ ✓ ✓ ✓ 
   16 
   17   Legend: H[ighlight], L[ocals], F[olds], I[ndents], In[j]ections
   18          +) multiple parsers found, only one will be used
[...]

PR #2253, opened 2022-01-14
highlights(php): fix and add missing highlights

mikehaertl:
@theHamsta As mentioned in #2207 here are a couple of additions for PHP

 * **@type**: added class and interface names in declarations, `use` statements and static method calls
 * **@parameter** parameters in method and function declarations
 * **@const** constant declarations and accesses
 * **@namespace** namespace names
 * **@conditional** added ternary operator `? :`
 * **@string** added backtick expressions
 * **@string.escape** escape expressions like `\n` in strings
 * **@label** goto labels
 
Also added some keywords and operators and moved some things e.g. the `.` operator to **@operators**

### DIFF
--- a/queries/php/highlights.scm
+++ b/queries/php/highlights.scm
@@ -10,6 +10,23 @@
  ] @type.builtin
 (named_type (name)) @type
 (named_type (qualified_name)) @type
+(class_declaration
+  name: (name) @type)
+(base_clause
+  [(name) (qualified_name)] @type)
+(enum_declaration
+  name: (name) @type)
+(interface_declaration
+  name: (name) @type)
+(namespace_use_clause
+  [(name) (qualified_name)] @type)
+(class_interface_clause
+  [(name) (qualified_name)] @type)
+(scoped_call_expression
+  scope: [(name) (qualified_name)] @type)
+(class_constant_access_expression
+  . [(name) (qualified_name)] @type
+  (name) @constant)
 
 ; Functions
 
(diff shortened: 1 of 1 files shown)


mikehaertl:
@infectos Maybe you also want to help test this? Suggestions welcome.

queries/php/highlights.scm
```diff
@@ -173,14 +214,17 @@
 [
   "="
 
+  "."
```
theHamsta, line 217: Why is this an operator? In other languages foo.bar is punctuation. If this needs a change it should be applied to all languages

theHamsta (commented):
Looks good to me! 👍

queries/php/highlights.scm
mikehaertl: Dot is the [string concatenation operator](https://www.php.net/manual/en/language.operators.string.php):
```php
<?php
$a = 'some ' . 'text';
```

In PHP object properties are always accessed with the arrow `->` not the `.`: 
```php
<?php
$a->someProp;
$a->someMethod();
```

Static props/methods use the `::` operator:
```php
<?php
$a = SomeClass::$prop;
$a = SomeClass::someMethod();
```

mikehaertl:
@theHamsta I've become a big fan of testing lately. But the existing tests for other languages seem to be rather rudimentary. Are there plans to extend this? And does it make sense to add tests for PHP? IMO a test should cover pretty much all of the important features and is quite some work though. So in any case this would be something for later.

queries/php/highlights.scm
theHamsta: Yeah, I had the suspicion that is was concatenation but didn't whether it was also object access

theHamsta:
Feel free to add tests. The test facilities were only added very recently and after that only additions and changes received tests.

mikehaertl:
@theHamsta Ok I may consider to add tests later in another PR.

theHamsta (approved):

PR #2256, opened 2022-01-15
fixed C/C++ variablesvariables as @variable

NaincyKumariKnoldus:
Hello I have changed this variable from @constant to @variable.
Please review my PR.
Thanks!


theHamsta:
Why? We used the screaming case convention to highlight constants. If you don't want this you can just remove the query.

theHamsta:
We'll need to find a better solution for this. As I said, my favorite one would be to use Atom-style specialization syntax which would allow you to opt-out from this on a global or per-language basis.

PR #2257, opened 2022-01-15
add hocon

antosha417:
add support for hocon files

### DIFF
--- a/ftdetect/hocon.vim
+++ b/ftdetect/hocon.vim
@@ -1,0 +1,1 @@
+au BufRead,BufNewFile *.conf set filetype=hocon
--- a/ftdetect/hocon.vim
+++ b/ftdetect/hocon.vim
@@ -1,1 +1,0 @@
-au BufRead,BufNewFile *.conf set filetype=hocon
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -801,6 +801,15 @@ list.hjson = {
   maintainers = { "@winston0410" },
 }
 
+list.hocon = {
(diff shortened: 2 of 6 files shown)


theHamsta (approved):
Seems to work really nice! Thank you!

If you want you can also add example files with highlight annotation to `tests/queries/highlights` but it's not necessary. Just in case you already experimented with some example files to see whether it's working.

antosha417:
@theHamsta, I added highlights tests

antosha417:
I think I should add .conf to ftdetect for tests to pass?

theHamsta:
Can you add a ftdetect to `scripts/minimal_init.lua` https://github.com/theHamsta/nvim-treesitter/blob/dd2524171029169c410c92b028f610fa67ce2beb/scripts/minimal_init.lua#L4 ?

ftdetect/hocon.vim
```diff
@@ -0,0 +1 @@
+au BufRead,BufNewFile *.conf set filetype=hocon
```
theHamsta, line 1: please don't add this on such a generic file ending. For the test to pass, just add it to scripts/minimal_init.lua

theHamsta (changes requested):
please don't add this on such a generic file ending. For the test to pass, just add it to scripts/minimal_init.lua

ftdetect/hocon.vim
antosha417: Ok. Will do tomorrow. Thanks

antosha417: done. Now tests should pass

theHamsta (approved):

PR #2258, opened 2022-01-15
docs: add issue template for highlighting issues

theHamsta:
It would be a lot easier when highlighting issue wouldn't be reported as bugs but people would actually provide a triple of example file, parsing result and screenshot of how it currently looks like.

Also instruct reporters that examples with `ERROR`s should be reported to upstream parsers.

Open an issue at my fork to see the rendered result: https://github.com/theHamsta/nvim-treesitter/issues/new/choose

### DIFF
--- a/.github/ISSUE_TEMPLATE/highlighting_issue.yml
+++ b/.github/ISSUE_TEMPLATE/highlighting_issue.yml
@@ -8,15 +8,13 @@ body:
       value: |
         # Before reporting
         Please perform the following steps before reporting an issue.
-        - I have updated my neovim version to latest _master_
-        - I have updated my plugin to the latest version
-        - I have run `:TSUpdate`
+        - I have updated my neovim version to latest _master_.
+        - I have updated my plugin to the latest version.
+        - I have run `:TSUpdate`.
         - I have inspected the syntax tree using https://github.com/nvim-treesitter/playground and made sure
           that no `ERROR` nodes are in the syntax tree. nvim-treesitter can not guarantee correct highlighting in the
-          presence or `ERROR`s. Please report the bug directly corresponding parser's repository in case you see
-          `ERROR` in the syntax tree. You can find all repository URLs in our REAMDE.md
-        - I have used `:TSHighlightCapturesUnderCursor` from https://github.com/nvim-treesitter/playground to inspect
-          what highlight groups is nvim-treesitter is using and that no `:h syntax` highlighting is interfering.
+          presence of `ERROR`s -- in this case, please report the bug directly at corresponding parser's repository. (You can find all repository URLs in [README.md](https://github.com/nvim-treesitter/nvim-treesitter#supported-languages).)
+        - I have used `:TSHighlightCapturesUnderCursor` from https://github.com/nvim-treesitter/playground to inspect which highlight groups Neovim is using and that legacy syntax highlighting is interfering (i.e., what you are observing is actual tree-sitter highlighting).
 
   - type: textarea
     attributes:
--- a/.github/ISSUE_TEMPLATE/highlighting_issue.yml
+++ b/.github/ISSUE_TEMPLATE/highlighting_issue.yml
@@ -7,7 +7,7 @@ body:
     attributes:
       value: |
         # Before reporting
-        Please do the following steps before reporting an issue.
+        Please perform the following steps before reporting an issue.
         - I have updated my neovim version to latest _master_
         - I have updated my plugin to the latest version
         - I have run `:TSUpdate`
--- a/.github/ISSUE_TEMPLATE/highlighting_issue.yml
+++ b/.github/ISSUE_TEMPLATE/highlighting_issue.yml
@@ -26,7 +26,7 @@ body:
   - type: textarea
     attributes:
       label: Example snippet that causes the problem
-      description: Please provide an example snippet in plain text that causes the problem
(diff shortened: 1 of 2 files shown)


clason:
Very good idea. I'd take the opportunity to remove the checkboxes, though -- they (irritatingly) show up as tasks in Github and, in my opinion, don't add much to the issue (users often either ignore or check them blindly, so you can't trust them but they take a lot of space). It's better to put this as boldface instructions into the template.

theHamsta:
Updated also `bug_report.yml` to use markdown instead of checkboxes.

.github/ISSUE_TEMPLATE/highlighting_issue.yml
```diff
@@ -0,0 +1,100 @@
+name: Highlighting issue
+description: Missing or incorrect highlights or you want to change the way something is highlighted
```
clason, line 2: probably can leave out "or you want..." as that is arguably already covered by "incorrect"?

.github/ISSUE_TEMPLATE/highlighting_issue.yml
```diff
@@ -0,0 +1,100 @@
...
+labels: [ highlights ]
+
+body:
+  - type: markdown
+    attributes:
+      value: |
+        # Before reporting
+        Please do the following steps before reporting an issue.
```
clason, line 10:
```suggestion
        Please perform the following steps before reporting an issue.
```

.github/ISSUE_TEMPLATE/highlighting_issue.yml
```diff
@@ -0,0 +1,100 @@
...
+        - I have updated my neovim version to latest _master_
+        - I have updated my plugin to the latest version
+        - I have run `:TSUpdate`
+        - I have inspected the syntax tree using https://github.com/nvim-treesitter/playground and made sure
+          that no `ERROR` nodes are in the syntax tree. nvim-treesitter can not guarantee correct highlighting in the
+          presence or `ERROR`s. Please report the bug directly corresponding parser's repository in case you see
+          `ERROR` in the syntax tree. You can find all repository URLs in our REAMDE.md
+        - I have used `:TSHighlightCapturesUnderCursor` from https://github.com/nvim-treesitter/playground to inspect
+          what highlight groups is nvim-treesitter is using and that no `:h syntax` highlighting is interfering.
```
clason, line 19:
```suggestion
        - I have updated my neovim version to latest _master_.
        - I have updated my plugin to the latest version.
        - I have run `:TSUpdate`.
        - I have inspected the syntax tree using https://github.com/nvim-treesitter/playground and made sure
          that no `ERROR` nodes are in the syntax tree. nvim-treesitter can not guarantee correct highlighting in the
          presence of `ERROR`s -- in this case, please report the bug directly at corresponding parser's repository. (You can find all repository URLs in [README.md](https://github.com/nvim-treesitter/nvim-treesitter#supported-languages).)
        - I have used `:TSHighlightCapturesUnderCursor` from https://github.com/nvim-treesitter/playground to inspect which highlight groups Neovim is using and that legacy syntax highlighting is interfering (i.e., what you are observing is actual tree-sitter highlighting).
```

.github/ISSUE_TEMPLATE/highlighting_issue.yml
```diff
@@ -0,0 +1,100 @@
...
+    attributes:
+      label: Describe the highlighting problem
+      description: A clear and concise description of what should be highlighted in a different way.
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Example snippet that causes the problem
+      description: Please provide an example snippet in plain text that causes the problem
```
clason, line 31:
```suggestion
      description: Please provide an example snippet in plain text that causes the problem.
```

.github/ISSUE_TEMPLATE/highlighting_issue.yml
```diff
@@ -0,0 +1,100 @@
...
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Example screenshot
+      description: |
+        Please provide a screenshot of the current highlighting. Please also tell us the `:h colorscheme` you are using
+        and how to install it. If applicable, you can also upload a screenshot with the contents of
+        `:TSHighlightCapturesUnderCursor':
```
clason, line 71:
```suggestion
        `:TSHighlightCapturesUnderCursor'.
```

clason (approved):

.github/ISSUE_TEMPLATE/highlighting_issue.yml
theHamsta: we can also discuss personal preferences here :slightly_smiling_face: There's not always a clear indications what's the best option and sometimes some discussion is required.

clason:
One (late) thought: you probably want to disable creating blank issues; otherwise lazy people will just use that to sidestep the template...

.github/ISSUE_TEMPLATE/highlighting_issue.yml
clason: I see what you mean -- maybe "Syntax highlighting for a specific language that is missing, incorrect, or can/should be improved"?

theHamsta:
I always use blank issue :smile:. Sometimes there's no right fit, we would need a PR that would make the template list excessive.

We didn't have the problem until now. I couldn't remember of a issue without the template for a long time. The blank issue is really hidden.

Issue #2260, opened 2022-01-15
main branch vs. master branch causes problems


dfgordon:
### Before reporting

- [X] I have updated my neovim version to latest _master_
- [X] I have updated my plugin to the latest version
- [X] I have run `:TSUpdate`
- [X] I have read the [troubleshooting section](https://github.com/nvim-treesitter/nvim-treesitter#troubleshooting)

### Describe the bug

`nvim_treesitter` does not fully support the use of the `main` branch convention in a language repo.

### To Reproduce

1. Add a parser configuration to `init.vim` where the `install_info` has a `url` key that points to a valid tree-sitter language repository with a `main` branch as opposed to a `master` branch.
2. Exit and restart `nvim`.  In my configuration it automatically tries to install the language.  If that is not the case use `:TSInstall mylanguage`
3. Note the output, I got the below error

```
nvim-treesitter[applesoft]: Failed to execute the following command:
{
  cmd = "mv",
  opts = {
    args = { "tree-sitter-applesoft-tmp/tree-sitter-applesoft-master", "tree-sitter-applesoft" },
    cwd = "/home/gordon/.local/share/nvim",
    stdio = {
      [2] = <userdata 1>,
      [3] = <userdata 2>
    }
  }
}
mv: cannot stat 'tree-sitter-applesoft-tmp/tree-sitter-applesoft-master': No such file or directory
```

What appears to be happening is `nvim_treesitter` assumes there is a `master` branch (notice the name of the directory it is trying to `mv`).  If you look in the place this directory is supposed to be, it is there, except `main` is substituted for `master`.

### Expected behavior

The language should be installed without error.  I verified that the same language installs without error and produces highlights as expected if the `url` is local.

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
  - OK: `node` found v14.16.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 8.4.1 20200928 (Red Hat 8.4.1-1)
[...]

PR #2262, opened 2022-01-15
docs(readme): make custom parser example more explicit about options

theHamsta:
Use all possible options to illustrate their usage


theHamsta:
Fix #2260

clason (approved):

PR #2270, opened 2022-01-16
Add more punctuaction highlights and add some missing operators

theHamsta:
"?" is still missing:
https://github.com/tree-sitter/tree-sitter-swift/issues/38

### DIFF
--- a/queries/swift/highlights.scm
+++ b/queries/swift/highlights.scm
@@ -1,3 +1,6 @@
+[ "." ";" ":" "," "->" ] @punctuation.delimiter
+[ "\\(" "(" ")" "[" "]" "{" "}"] @punctuation.bracket ; TODO: "\\(" ")" in interpolations should be @punctuation.special
+
 ; Identifiers
 (attribute) @variable
 (simple_identifier) @variable
--- a/queries/swift/highlights.scm
+++ b/queries/swift/highlights.scm
@@ -1,4 +1,4 @@
-[ "." ";" ":" "," "->" ] @punctuation.delimiter
+[ "." ";" ":" "," ] @punctuation.delimiter
 [ "\\(" "(" ")" "[" "]" "{" "}"] @punctuation.bracket ; TODO: "\\(" ")" in interpolations should be @punctuation.special
 
 ; Identifiers
@@ -130,4 +130,5 @@
  "!="
  "!=="
  "==="
+ "->"
 ] @operator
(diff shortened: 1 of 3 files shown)


queries/swift/highlights.scm
```diff
@@ -1,3 +1,6 @@
+[ "." ";" ":" "," ] @punctuation.delimiter
+[ "\\(" "(" ")" "[" "]" "{" "}"] @punctuation.bracket
```
theHamsta, line 2: ideally "\\("` should be `@punctuation.special` in string interpolation. but it's not easy to match against

`"?"` after optional types needs a parser change: https://github.com/alex-pinkus/tree-sitter-swift/issues/100

queries/swift/highlights.scm
```diff
@@ -1,3 +1,6 @@
+[ "." ";" ":" "," "->" ] @punctuation.delimiter
```
theHamsta, line 1: `->` is apparently used to indicate return type in functions. `->` is an operator in Rust. So let's switch to that variant

tests/query/highlights/swift/string-interpolation.swift
```diff
@@ -0,0 +1,7 @@
+var one = 1
+var two = 2
+var name = "Let's do some math: \(one) + \(one) = \(one + one) = \(two)"
+//                              ^ <- punctuaction.bracket
```
clason, line 4:
```suggestion
//                              ^ <- punctuation.bracket
```
(and below)

clason (approved):

tests/query/highlights/swift/string-interpolation.swift
```diff
@@ -0,0 +1,7 @@
...
+//                                 ^ <- variable
+//                                   ^ <- punctuaction.bracket
+//                                                 ^ <- punctuaction.bracket
```
theHamsta, line 7: `CUSTOM_OPERATOR` as using in `for x in 0..<3` can not be highlighted at the moment

theHamsta: `..<` can in some cases.

theHamsta:
I don't understand why tests are failing...

theHamsta:
I will merge without the tests and try to push them in a follow PR!

alex-pinkus:
@theHamsta, any reason I shouldn't upstream these queries into https://github.com/alex-pinkus/tree-sitter-swift? I figure the consistency would be valuable, but want to check in case there's something I'm missing.

clason:
In general, queries are not 100% compatible with upstream since we use some custom/adapted capture names, so we always have to bundle ours. But if you checked them and they apply (with possible minimal changes) upstream, feel free to take them!

theHamsta:
@alex-pinkus i think in this case all changes here could also be made to the upstream queries. In general, the captures Atom expects are a bit different.

PR #2272, opened 2022-01-16
feat(lua)!: switch from our fork to MunifTanjim's

clason:
also take queries from https://github.com/MunifTanjim/nvim-treesitter-lua/tree/main/queries/lua

@MunifTanjim (it would help if you could summarize the advantages of your parser over the old one in a comment).

BREAKING CHANGE: queries are not compatible; modules will have to update

### DIFF
--- a/queries/lua/folds.scm
+++ b/queries/lua/folds.scm
@@ -1,12 +1,10 @@
 [
- (for_in_statement)
- (for_statement)
+ (do_statement)
  (while_statement)
  (repeat_statement)
  (if_statement)
- (do_statement)
+ (for_statement)
+ (function_declaration)
  (function_definition)
- (local_function)
- (function)
- (table)
+ (table_constructor)
 ] @fold
--- a/queries/lua/indents.scm
+++ b/queries/lua/indents.scm
@@ -1,20 +1,19 @@
 [
   (function_definition)
+  (function_declaration)
   (variable_declaration)
-  (local_variable_declaration)
   (field)
-  (local_function)
-  (function)
+  (do_statement)
+  (while_statement)
+  (repeat_statement)
   (if_statement)
   (for_statement)
-  (for_in_statement)
-  (repeat_statement)
   (return_statement)
-  (while_statement)
-  (table)
+  (table_constructor)
   (arguments)
-  (do_statement)
 ] @indent
+
+ @ignore
 
 [
   "end"
@@ -24,8 +23,8 @@
   "("
   ")"
   "then"
-  (else)
-  (elseif)
+  (else_statement)
+  (elseif_statement)
 ] @branch
 
 (comment) @ignore
--- a/lockfile.json
+++ b/lockfile.json
@@ -156,7 +156,7 @@
     "revision": "3b213925b9c4f42c1acfe2e10bfbb438d9c6834d"
   },
   "lua": {
-    "revision": "6f5d40190ec8a0aa8c8410699353d820f4f7d7a6"
+    "revision": "6197f42bf174b867a499e201e40515b9feef12af"
   },
   "make": {
     "revision": "a4b9187417d6be349ee5fd4b6e77b4172c6827dd"
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -151,10 +151,11 @@ list.ledger = {
 
 list.lua = {
   install_info = {
-    url = "https://github.com/nvim-treesitter/tree-sitter-lua",
-    files = { "src/parser.c", "src/scanner.cc" },
+    url = "https://github.com/MunifTanjim/tree-sitter-lua",
+    branch = "main",
+    files = { "src/parser.c", "src/scanner.c" },
   },
-  maintainers = { "@vigoux" },
+  maintainers = { "@muniftanjim" },
 }
 
 list.python = {
--- a/queries/lua/highlights.scm
+++ b/queries/lua/highlights.scm
@@ -1,36 +1,34 @@
-;;; Highlighting for lua
-
 ;;; Builtins
-(self) @variable.builtin
+
+[
+  (false)
+  (true)
+] @boolean
+
+(nil) @constant.builtin
+
+((identifier) @variable.builtin
+ (#match? @variable.builtin "self"))
 
 ;; Keywords
 
-(if_statement
-[
-  "if"
-  "then"
-  "end"
-] @conditional)
+"return" @keyword.return
 
 [
-  "else"
-  "elseif"
-  "then"
-] @conditional
+ "goto"
+ "in"
+ "local"
+] @keyword
 
-(for_statement
+(label_statement) @label
+
+(break_statement) @keyword
+
+(do_statement
 [
-  "for"
   "do"
   "end"
-] @repeat)
-
-(for_in_statement
-[
-  "for"
-  "do"
-  "end"
-] @repeat)
+] @keyword)
 
 (while_statement
 [
@@ -45,57 +43,90 @@
   "until"
 ] @repeat)
 
-(do_statement
+(if_statement
 [
+  "if"
+  "elseif"
+  "else"
+  "then"
+  "end"
+] @conditional)
+
+(elseif_statement
+[
+  "elseif"
+  "then"
+  "end"
+] @conditional)
+
+(else_statement
+[
+  "else"
+  "end"
+] @conditional)
+
+(for_statement
+[
+  "for"
   "do"
   "end"
-] @keyword)
+] @repeat)
 
+(function_declaration
 [
- "in"
- "local"
- (break_statement)
- "goto"
-] @keyword
+  "function"
+  "end"
+] @keyword.function)
 
-"return" @keyword.return
+(function_definition
+[
+  "function"
+  "end"
+] @keyword.function)
 
 ;; Operators
 
 [
- "not"
  "and"
+ "not"
  "or"
 ] @keyword.operator
 
 [
-"="
-"~="
-"=="
-"<="
-">="
-"<"
-">"
-"+"
-"-"
-"%"
-"/"
-"//"
-"*"
-"^"
-"&"
-"~"
-"|"
-">>"
-"<<"
-".."
-"#"
- ] @operator
+  "+"
+  "-"
+  "*"
+  "/"
+  "%"
+  "^"
+  "#"
+  "=="
+  "~="
+  "<="
+  ">="
+  "<"
+  ">"
+  "="
+  "&"
+  "~"
+  "|"
+  "<<"
+  ">>"
+  "//"
+  ".."
+] @operator
 
-;; Punctuation
-["," "." ":" ";"] @punctuation.delimiter
+;; Punctuations
+
+[
+  ";"
+  ":"
+  ","
+  "."
+] @punctuation.delimiter
 
 ;; Brackets
+
 [
  "("
  ")"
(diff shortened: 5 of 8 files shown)


queries/lua/folds.scm
```diff
@@ -1,12 +1,1496 @@
...
- (function)
- (table)
-] @fold
+
+
+
+
+
+
+<!DOCTYPE html>
```
theHamsta: @clason probably you wanted to download the raw document. Our users won't be to happy with HTML

clason: 🤦🏻

clason:
Test failures are due to
1. query differences
2. indentation missing from upstream queries

MunifTanjim:
Well, wow! That happened fast! 😂

> it would help if you could summarize the advantages of your parser over the old one in a comment

Let me take it from my comment on reddit.

---

### Block Comment/String

The old one didn't have separate nodes for block comment/string start/end. Since lua's block comment/string's start/end can be of arbritary length (e.g. `[=[` or `[======[`, you get the idea), having separate nodes for them helps when you're writing query that targets string content (e.g. injections queries).

It also had some bugs handling these block comments/strings.
- https://github.com/nvim-treesitter/tree-sitter-lua/issues/7

### Table access

```lua
tbl.a["b"]:c()
```

**Old parser**:

```scheme
(program
  (function_call
    (field_expression
      (identifier)
      (property_identifier))
    (string)
    (method)
    (arguments)))
```

**This parser**:

```scheme
(chunk
  (function_call
    name: (method_index_expression
      table: (bracket_index_expression
        table: (dot_index_expression
          table: (identifier)
          field: (identifier))
        field: (string))
      method: (identifier))
    arguments: (arguments)))
```

If you're using the generated tree-sitter ast to do refactoring or some complex stuff in a script, these extra information will surely help.

### Expression vs Statement

The old parser doesn't abide by some rules of lua about what is an expression and what is a statement and where can those go. For example:

**Old parser**:

```

42

---

(program
  (expression (number)))

```

Would not show any error and accept a whole line that's an expression.

**This parser**:

```

42

---

(chunk
  (ERROR (number)))

```

Would show error, because this is not valid Lua.

### Tests

This one has corpus tests for various edge cases and parses the whole `luvit/luvit` repository (more can be added) as part of testing.

MunifTanjim:
> indentation

I didn't get satisfactory result for indents queries. That's why I didn't push it. Here's a related discussion:

- https://github.com/nvim-treesitter/nvim-treesitter/discussions/1668

Here's the `indents.scm` file sitting on my disk:

```scheme
[
  (function_definition)
  (variable_declaration)
  (field)
  (do_statement)
  (while_statement)
  (repeat_statement)
  (if_statement)
  (for_statement)
  (return_statement)
  (table_constructor)
  (arguments)
  (block)
] @indent

(comment) @ignore
```

clason:
> Here's the indents.scm file sitting on my disk:

*yoink!*

clason:
One thing I worry about is that the query coverage seems to be a bit of a regression compared to what we had. It might be better to start from our queries and just adapt them where necessary?

MunifTanjim:
> One thing I worry about is that the query coverage seems to be a bit of a regression compared to what we had. It might be better to start from our queries and just adapt them where necessary?

I haven't checked the current queries in `nvim-treesitter/nvim-treesitter/queries/lua` for a while. I just checked and seems like only highlights.scm is lacking a bit. Pushing it in a few minutes.

theHamsta:
@MunifTanjim should we Auto-Import the queries from your repo (curl during parser update CI) assuming that they're made for Neovim?

theHamsta:
We should also check the queries in nvim-treesitter-textobjects before merging this

theHamsta:
Test failure comes probably from hocon or something unrelated and can be ignored

clason:
> @MunifTanjim should we Auto-Import the queries from your repo (curl during parser update CI) assuming that they're made for Neovim?

That would include moving the queries from the separate plugin to the parser repo (which should be obsolete once we make the switch).

> We should also check the queries in nvim-treesitter-textobjects before merging this

I would have made a PR after this gets merged. 

> Test failure comes probably from hocon or something unrelated and can be ignored

No, there are definitely Lua failures. It seems that indentation with this parser doesn't really work.

MunifTanjim:
> @MunifTanjim should we Auto-Import the queries from your repo (curl during parser update CI) assuming that they're made for Neovim?

If this PR is merged, there's no point in maintaining the `MunifTanjim/nvim-treesitter-lua` repo. I'll archive that repo in that case. So, nope.

clason:
> If this PR is merged, there's no point in maintaining the `MunifTanjim/nvim-treesitter-lua` repo. I'll archive that repo in that case. So, nope.

I think @theHamsta meant that you would maintain queries in the _parser_ repo, and we would pull them from there. Other parsers do similar things. This way, you could make improvements without needing to PR to this repo.

MunifTanjim:
> I think @theHamsta meant that you would maintain queries in the parser repo, and we would pull them from there. Other parsers do similar things. This way, you could make improvements without needing to PR to this repo.

Ah, I see. I can do that too. But doesn't it get confusing for the contributors? 🤔 How do they know where to update the queries?

> It seems that indentation with this parser doesn't really work.

Just curious, is it Lua in general? or this new parser? As far as I remember I couldn't get proper indentation with the current parser either. I'm just using whatever indentation comes with `vim` by default for Lua.

clason:
> Ah, I see. I can do that too. But doesn't it get confusing for the contributors? 🤔 How do they know where to update the queries?

We would tell them ;) (Similarly to how in Neovim, we tell people to report/fix bugs with legacy syntax files in Vim.) But it's not necessary; whatever is easiest to maintain.

> Just curious, is it Lua in general? or this new parser? As far as I remember I couldn't get proper indentation with the current parser either. I'm just using whatever indentation comes with vim by default for Lua.

I was under the impression that it was the new parser since the indentation tests are not marked as known failures. But I have to admit I did not check myself.

theHamsta:
Tests are failing due to Markdown. Sorry, for pushing this to master. Indent tests for Lua might also fail when the new parser is better than the old, then the expected failure need to be updated.

clason:
The highlight test for Lua is fixed now (thanks, @MunifTanjim). The expected indentation looks like correct indentation, though, while the found indentation is wrong?

queries/lua/indents.scm
```diff
@@ -1,31 +1,16 @@
...
-  "end"
-  "until"
-  "{"
-  "}"
-  "("
-  ")"
-  "then"
-  (else)
-  (elseif)
-] @branch
```
MunifTanjim, line 29: Can we add back these removed bits? That might fix the tests. This is the equivalent:

```scheme
(else_statement) @branch
(elseif_statement) @branch

[
  "end"
  "until"
  "{"
  "}"
  "("
  ")"
  "then"
] @branch
```

clason: Yep, sounds good, just did. We might iterate a bit to get the best of both worlds in all queries.

MunifTanjim:
> We would tell them ;) (Similarly to how in Neovim, we tell people to report/fix bugs with legacy syntax files in Vim.) But it's not necessary; whatever is easiest to maintain.

I think these queries are pretty stable and won't change that much in future. So I'd prefer archiving the `MunifTanjim/nvim-treesitter-lua` repo.

queries/lua/indents.scm
clason: Still some failures, but we're getting there ;)

clason:
> > We would tell them ;) (Similarly to how in Neovim, we tell people to report/fix bugs with legacy syntax files in Vim.) But it's not necessary; whatever is easiest to maintain.
> 
> I think these queries are pretty stable and won't change that much in future. So I'd prefer archiving the `MunifTanjim/nvim-treesitter-lua` repo.

OK, let's keep them here then. PRs for improving them are still appreciated :)

queries/lua/indents.scm
MunifTanjim: I'm setting up tests locally and looking into the indents test failure.

clason: OK, I think the main culprit is the `block` capture; we can't use that for indentation since it's used both where you want to and where you don't want to indent.

clason: There's only one test failure remaining (`loop.lua`), which I can't reproduce locally. In fact, the output looks weird: the mis-indented line isn't even in the test file?

clason:
@vigoux (and others) I think this is getting to the point where it'd be good if people familiar with the old parser would test-drive it for a bit and see if there are any noticeable regressions.

queries/lua/indents.scm
clason: One failure remaining (`loop.lua`); this indents correctly when the whole file (with the new line) is indented via `gg=G`, but inserting a new line adds a wrong indent. Not sure what is happening there.

clason: I've added the `break_statement` to `@ignore` -- @theHamsta is that a reasonable thing to do?

clason: But locally I then get wrongly indented newlines in `for` loops...

MunifTanjim: This is not related to `break_statement`.

So for 

```scheme
while false do
  a = 1
  b = 2
end
```

and ast

```
while_statement [93, 0] - [96, 3]
  condition: false [93, 6] - [93, 11]
  body: block [94, 2] - [95, 7]
    assignment_statement [94, 2] - [94, 7]
      variable_list [94, 2] - [94, 3]
        name: identifier [94, 2] - [94, 3]
      expression_list [94, 6] - [94, 7]
        value: number [94, 6] - [94, 7]
    assignment_statement [95, 2] - [95, 7]
      variable_list [95, 2] - [95, 3]
        name: identifier [95, 2] - [95, 3]
      expression_list [95, 6] - [95, 7]
        value: number [95, 6] - [95, 7]
```

i.e. if there are more than one statements in `body:` and cursor is on the first statement, if newline is entered,

https://github.com/nvim-treesitter/nvim-treesitter/blob/4cf688ee681bbe9a35651732b2d2281a01c40066/lua/nvim-treesitter/indent.lua#L77

is getting `body: block`, instead of getting the first statement inside `block`.

clason: OK, that looks like something that needs to be fixed in general. (If you want to take a stab at it, that would be very welcome ;))

I've marked this test as expected failure for now.

clason: @theHamsta can you see why the tests are still failing, even though I marked the tests as `XFAIL`?

MunifTanjim: > can you see why the tests are still failing, even though I marked the tests as XFAIL or added them to the expected_failures table?

I think these 2 needs to be removed from `expected_failures`.

image

clason: Oh, you mean they are failing because the expected failures _don't_ fail now? That's... not intuitive.

clason: Indeed, that was it, good catch!

vigoux (approved):
A big green tick on this one, thanks!

clason:
If @vigoux says it's good, it's good :)

Do we need some sort of "breaking changes" issue where we mention the switch?

vigoux:
Hmm... I think that the language support issue could do it?

clason:
No, that's for something different. I'm talking about something like the "Following HEAD" issue on Neovim where breaking changes are mentioned and migration paths are explained.

Like "new parser, please make sure to update queries"

theHamsta:
Yes, would be good for people to subscribe. Lanugage support would mix-up feature requests with breaking change notification.

I suposse we need to fix also nvim-treesitter-textobjects for this one (or remove the Lua queries as a quickfix.

theHamsta (approved):
Great work! Also the syntax tree is now tidied up also for Lua

clason:
> Yes, would be good for people to subscribe. Lanugage support would mix-up feature requests with breaking change notification.

Agreed; I'll open up such an issue and pin it.

> I suposse we need to fix also nvim-treesitter-textobjects for this one (or remove the Lua queries as a quickfix.

Yep, that was my next todo; luckily, @MunifTanjim's repo already contains replacement queries.

(This leaves nvim-treesitter-textsubjects, but that's third-party.)

PR #2273, opened 2022-01-16
Fix highlight query tests

theHamsta:


theHamsta:
@antosha417 your test was failing because `days` was actually highlighted as `@string` instead of `@keyword`. What is the correct highlight?

theHamsta:
This fixes the tests. Should have listened to `@MDeiml` comments about the quick fix that disabled some of the markdown parens. We didn't require branches to be up-to-date so tests where not performed on a fully rebased branch.

clason (approved):
yep, let's go with that for now!

antosha417:
I modified the grammar to parse units as keyword. So days should parse as keyword. @theHamsta you should TSupdate hocon to run tests locally.

PR #2275, opened 2022-01-16
Improve Indent module

MunifTanjim:
If line is an empty line, start from the last node on previous line.
Otherwise, start from the first node on current line.

---

**C/CPP**

- I don't fully understand the expectation for `ternary.c`. What is the definite rule there?
- Is the expectation for `preproc_func.c` file correct? Feels like it should be:
    ```c
    #define FOO(x) do { \
        x = x + 1;  \
        x = x / 2;  \
    } while (x > 0);
    ```

- We can't magically expect the `indentexpr` to trigger for this ( https://github.com/nvim-treesitter/nvim-treesitter/blob/8b1d3906199e1572c2201e494bba017d823c0c8f/tests/indent/c_spec.lua#L36 ). It is triggered when the newline is entered, not after the text is inserted.

---

**Lua**

🎉 almost 🤷‍♂️

---

**Python**

🎉 almost 🤷‍♂️

---

**Rust**

🎉🎉🎉


tests/indent/python_spec.lua
```diff
@@ -26,8 +26,8 @@ describe("indent Python:", function()
     run:new_line("basic_blocks.py", { on_line = 1, text = "wait,", indent = 4 })
     run:new_line("basic_blocks.py", { on_line = 6, text = "x += 1", indent = 4 })
     run:new_line("basic_blocks.py", { on_line = 10, text = "x += 1", indent = 8 })
-    run:new_line("basic_blocks.py", { on_line = 7, text = "x += 1", indent = 4 }, "7, after last line of a block")
-    run:new_line("basic_blocks.py", { on_line = 11, text = "x += 1", indent = 8 }, "11, after last line of a block")
+    run:new_line("basic_blocks.py", { on_line = 7, text = "x += 1", indent = 0 })
+    run:new_line("basic_blocks.py", { on_line = 11, text = "x += 1", indent = 4 })
```
MunifTanjim, line 30: Reason: `pass` is marked as `@return` node. So if the cursor is on a line with `pass`, entering newline should not keep indentation, it should dedent instead.

tests/indent/rust/macro.rs
```diff
@@ -2,10 +2,10 @@ macro_rules! foo {
     ($a:ident, $b:ident, $c:ident) => {
         struct $a;
         struct $b;
-    },
+    };
```
MunifTanjim, line 5: Reason: fix bad rust syntax.

theHamsta:
~~if the change isn't big you could use this opportunity to get rid of `queries.get_capture_matches(bufnr, capture, "indents", root, lang)` and use the native query API instead.~~ let's do that in a follow up PR. I did this already once and it wasn't too complicated but required some changes.

clason:
Could you remove the `XFAIL` annotations from (at least) the Lua indent tests and push CI again? I'd be curious to see what works now and what still fails.

MunifTanjim:
> Could you remove the XFAIL annotations from (at least) the Lua indent tests and push CI again? I'd be curious to see what works now and what still fails.

Those are still failing. I think we'll need to come up with something different from what's currently being done with `get_node_at_line`. The problem is previous line can have multiple node and they can have nested childen. For example,

image

When the cursor is on line 1, and newline is entered, we need to detect that we're inside `value: string`, instead of `assignment_statement`.

clason:
OK, that is good to know, thanks! 

It would be quite helpful if you could mention in the original PR description which tests pass now that used to fail (helps us keeping track of things).

(I intend to merge the Lua parser switch soonish; I'd just like to get a bit more feedback from other people.)

MunifTanjim:
@clason @theHamsta updated the logic for indent module. I think this is a bit better than before. Not sure about the performance tho.

lua/nvim-treesitter/indent.lua
```diff
@@ -4,50 +4,31 @@ local tsutils = require "nvim-treesitter.ts_utils"
...
     branches = get_map "@branch.node",
-    returns = get_map "@return.node",
     ignores = get_map "@ignore.node",
   }
 end, {
   -- Memoize by bufnr and lang together.
-  key = function(bufnr, _, lang)
-    return tostring(bufnr) .. "_" .. lang
+  key = function(bufnr, root, lang)
+    return tostring(bufnr) .. root:id() .. "_" .. lang
```
MunifTanjim, line 27: This is needed in case the same language is injected onto itself (happens in rust).

clason:
> Not sure about the performance tho.

Correctness before performance; up until now, tree-sitter-based indentation has never really worked, so anything is fine ;)

theHamsta:
The main performance sin is the used query API (which I try bro fix after your PR) but that has nothing to do with the indentation code itself. Without the queries Lua code show always be able to compete with vim equivalents. As long as there're no perceptible freezes everything's fine.

MunifTanjim:
> The main performance sin is the used query API.

I think I just fixed did that. 😂 I needed to access query metadata for improving indent for Python. Since the current `get_capture_matches` doesn't give access to that, I just replaced it with the native one as you suggested earlier.

theHamsta:
@MunifTanjim Really, really nice. I just saw this a minute ago. Before I was answering on phone. We really want to get rid of the API calls used before. For a lot of other application (textobjects, playground, TSHighlightsUnderCusors, locals), it can cause serious freezes. playground, TSHighlightsUnderCusors avoid it now and work fine. Also https://github.com/nvim-treesitter/nvim-treesitter/pull/1120 works instantly (at least it was before the parent cache in tree-siter was removed).

The problem with the thing before was that it was collecting all results into nested tables and then the nested table were deconstructed again to get the actual information. In most cases you'd have the same thing that you would have before but constructing all these objects and searching in them takes surprisingly long.

theHamsta:
Really nice! @kyazdani42 could you please have a look onto this since you developed the original implementation?

yioneko:
Glad to see this finally happens! I'm also working on treesitter indent for a while: https://github.com/yioneko/nvim-yati, and it has support for more languages and tests covered. I think this is a very chance to share my works for better improvements to be done in this PR and hopefully it will help a bit :)

clason:
Sure, any help in improving this is very welcome! One drawback of your plugin is that it doesn't use the standard `indents.scm` tree-sitter queries. Is there any reason for that?

kyazdani42:
I'll look into it today, thanks a lot for improving this :)

theHamsta:
@yioneko language injection is another feature that needs to be added in future.

yioneko:
@clason Currently the scm query file can be only used through `Query:iter_captures`, and captures all the nodes within given root node. For indent calculation, the query must be executed on the whole tree because the traversal finally reaches root node of the tree, which might cause performance issue (previously it is really slow, but things might change as we will switch to native query after this PR merged). 

Instead I use pure Lua code to test node on demand, that is, only capture nodes on the traversal path from bottom to up to reduce the number of nodes visited and provide more fast calculation. However I lose the power of treesitter query syntax in the meantime, but that's simply a trade-off because simple query for indent calculation is enough from my view.

yioneko:
@theHamsta oh but hope that will be much easier to add since I've already implemented it and I'm also willing to help if needed.

lua/nvim-treesitter/indent.lua
```diff
@@ -2,52 +2,68 @@ local parsers = require "nvim-treesitter.parsers"
...
 
-local function node_fmt(node)
-  if not node then
-    return nil
-  end
-  return tostring(node)
+  local a_srow = anchor:range()
+
+  local opening_delimiter = delimiter:sub(1, 1)
+  local opening_node, opening_pos = walk_sibling("backward", a_srow, #vim.fn.getline(a_srow + 1) - 1, function(n)
```
kyazdani42, line 35: so you might remove this dir parameter and remove the backward

lua/nvim-treesitter/indent.lua
```diff
@@ -2,52 +2,68 @@ local parsers = require "nvim-treesitter.parsers"
...
+    local node = root:descendant_for_range(row, col, row, col)
+    local pos = 1
+    while node and not matcher(node) do
+      if dir == "backward" then
+        node = node:prev_sibling()
+      elseif dir == "forward" then
+        node = node:next_sibling()
+      else
+        error()
+      end
```
kyazdani42, line 26: looks like this whole thing could be replaced with `node:prev_sibling()` only

lua/nvim-treesitter/indent.lua
MunifTanjim: Updated.

lua/nvim-treesitter/indent.lua
MunifTanjim: Updated.

MunifTanjim:
The captures in this PR is not good enough yet, i.e. they can't support every languages.
We can still do better and make the captures more specific and granular. Inspirations can come from:

- https://github.com/chfritz/atom-sane-indentation
- https://github.com/emacsmirror/tree-sitter-indent/blob/master/tree-sitter-indent.el
- https://github.com/yioneko/nvim-yati

clason:
Getting tree-sitter indent up to parity is a long-term project, no need to tackle everything in one go. I would prefer small, incremental, easily mergable improvement PRs unless it turns out that a total rewrite is needed.

MunifTanjim:
Yes, agree. I don't think a total rewrite is needed, the base is pretty good. Just need to add some more captures and break down some existing ones to be more granular. Also indent tests for other languages needs to be added so that we can make changes confidently without breaking things that already works. Should be done in a few followup PRs.

clason:
Great, let us know when _this_ PR is ready for review!

MunifTanjim:
It is ready. Unless you want me to add some tests for other languages.

clason:
Good, just wanted to have this explicit in case people were still waiting. Let's get this merged ASAP.

theHamsta:
We could add a tracking issue and a issue template for folding.

- people should be encouraged to provide code examples for which folding
- injection support, testing for injected languages
- special support for certain languages needed (e.g. lisp indent, some special rules for indent-significant languages like Python/Haskell? 
- benchmarks? Maybe for more critical things first but maybe benchmarks for folding are a by product

clason:
folding or indents? ;)

kyazdani42 (approved):
LGTM, this 999 number seems a bit weird to me but let's leave it for now :)

MunifTanjim:
> LGTM, this 999 number seems a bit weird to me but let's leave it for now :)

That was there just a safeguard. 😂 I got myself into infinite loop a few times while working on it. It shouldn't hit that high ever in practice.

clason:
> That was there just a safeguard. 😂 I got myself into infinite loop a few times while working on it. It shouldn't hit that high ever in practice.

Oh, you sweet summer child... There is no maximal line length that some people will easily exceed.

MunifTanjim:
> There is no maximal line length that some people will easily exceed

That's actually being used to count nodes in python arguments. I hope somebody is not passing that much arguments to a single function 🤣

clason:
Maybe add a TODO comment to that line so we can more easily revisit it later?

clason:
And merged! Thank you @MunifTanjim for tackling this! 

Onward and upward :)

PR #2280, opened 2022-01-17
Add phpdoc parser (experimental)

mikehaertl:
The parser already creates quite good results. It's definitely better than the current `comments` parser which does not really fit well with PHP.

ts_phpdoc_demo

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -244,6 +244,17 @@ list.php = {
   maintainers = { "@tk-shirasaka" },
 }
 
+list.phpdoc = {
+  install_info = {
+    url = "https://github.com/claytonrcarter/tree-sitter-phpdoc",
+    files = { "src/parser.c" },
+    -- parser.c in the repo still based on TS 0.17 due to other dependencies
+    requires_generate_from_grammar = true,
+    generate_requires_npm = true,
+  },
+  experimental = true,
+}
+
 list.java = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-java",
--- a/queries/php/injections.scm
+++ b/queries/php/injections.scm
@@ -1,3 +1,3 @@
 (text) @html
 
-(comment) @comment
+(comment) @phpdoc
--- a/queries/phpdoc/highlights.scm
+++ b/queries/phpdoc/highlights.scm
@@ -1,0 +1,42 @@
+(tag_name) @attribute
+(tag
+  (tag_name) @_tag (#eq? @_tag "@param")
+  (variable_name) @parameter
+)
+(tag
+  (tag_name) @_tag (#eq? @_tag "@property")
+  (variable_name) @property
+)
+(tag
+  (tag_name) @_tag (#eq? @_tag "@var")
+  (variable_name) @variable
+)
+(tag
+  (tag_name) @_tag (#eq? @_tag "@method")
+  (name) @method
+)
+(parameter
+  (variable_name) @parameter)
+(type_list
+  [
+    (array_type)
+    (primitive_type)
+    (named_type)
+  ] @type)
+(tag
+  (description (text) @text))
+(tag
+  [
+    (author_name)
+    (version)
+  ] @text)
+(tag
+  (email_address) @text.uri
+)
+
+[
+  "$"
+  ">"
+  "<"
+  "|"
+]@keyword


lua/nvim-treesitter/parsers.lua
```diff
@@ -244,6 +244,15 @@ list.php = {
   maintainers = { "@tk-shirasaka" },
 }
 
+list.phpdoc = {
+  install_info = {
+    url = "https://github.com/claytonrcarter/tree-sitter-phpdoc",
+    files = { "src/parser.c" },
+  },
+  used_by = { "php" },
```
mikehaertl, line 252: This line breaks PHP parsing completely. And if I try to open playground I get:

```
Error executing vim.schedule lua callback: ...local/share/nvim/runtime/lua/vim/treesitter/language.lua:33: ABI version mismatch for /home/mike/.local/share/nvim/plugge
d/nvim-treesitter/parser/phpdoc.so: supported between 13 and 13, found 12                                                                                              
stack traceback:                                                                                                                                                       
        [C]: in function '_ts_add_language'                                                                                                                            
        ...local/share/nvim/runtime/lua/vim/treesitter/language.lua:33: in function 'require_language'                                                                 
        /usr/local/share/nvim/runtime/lua/vim/treesitter.lua:38: in function '_create_parser'                                                                          
        /usr/local/share/nvim/runtime/lua/vim/treesitter.lua:93: in function 'get_parser'                                                                              
        ...d/playground/lua/nvim-treesitter-playground/internal.lua:50: in function 'get_node_at_cursor'                                                               
        ...d/playground/lua/nvim-treesitter-playground/internal.lua:413: in function 'fn'                                                                              
        ...gged/playground/lua/nvim-treesitter-playground/utils.lua:23: in function 'cb'                                                                               
        vim.lua:285: in function <vim.lua:285>
```

If I remove the line, PHP works again - but comments are still not parsed. At least I can't see anything in playground. I only see a single `comment` node. Pressing `t` for injected languages also doesn't help. 

Any ideas?

mikehaertl: Ok, created a PR with a re-generated `parser.c` from the latest TS version. That fixed the issue for me. Let's see if the PR gets merged soon. If not I'll create a fork that we can maybe include here.

https://github.com/claytonrcarter/tree-sitter-phpdoc/pull/2

theHamsta: You probably want to inject phpdoc into PHP using PHP's injections.scm. A `used_by` is not necessary.

mikehaertl: Right, then I'll drop this line. The injection file is already included in this PR.

There's another issue with the phpdoc parser though: The author can't upgrade to TS 0.19 as this would break the integration into Atom. If you have an idea how to solve this maybe you can comment there? https://github.com/claytonrcarter/tree-sitter-phpdoc/pull/2#issuecomment-1014682817

theHamsta: Add `    requires_generate_from_grammar = true,` as some parsers already have it. If the parser requires npm for generation then also `    generate_requires_npm = true,`

mikehaertl: Not sure if I understand. The parser does not require npm or generation from grammar. It includes a `parser.c` ready for compilation.

The issue is rather that the author wants to keep a version for TS 0.17 in his repo. So we will probably use 2 branches and we will reference a `revision` here (the one compatible with TS 0.2x).

mikehaertl: Oh, I think I see: If we generate from grammer it will always use the correct version of treesitter to build the `parser.c`. I'll check.

theHamsta: It will be the correct version when the user has the correct version of tree-sitter and `./node_modules/.bin` not in PATH or else it will use the old version of tree-sitter.

mikehaertl: Pushed the change. It works locally, but some tests here seem to not like it.

Could it have to do with the version constraint in the phpdoc's `package.json`?

https://github.com/claytonrcarter/tree-sitter-phpdoc/blob/0864e30e70a53f751596657f75b14dd042cd4793/package.json#L28-L29

mikehaertl: Sorry, I forgot to actually push the changes. Tests also seem fine now - problem solved. Thank you!

theHamsta (approved):

vigoux (approved):
It looks pretty okay from here, but in Zulip you say that it does not work, what does not work ?

mikehaertl:
> ... but in Zulip you say that it does not work, what does not work ?

@vigoux That's from before we solved the issues here. 

I still want to have a look at the highlights so please wait with merging. I'll remove the draft status when it's ready.

mikehaertl:
Highlight queries are complete now.

queries/phpdoc/highlights.scm
```diff
@@ -0,0 +1,29 @@
+(tag_name) @tag
+(tag
+  (tag_name) @_tag (#eq? @_tag "@param")
+  (variable_name) @parameter
+)
+(tag
+  (tag_name) @_tag (#eq? @_tag "@property")
+  (variable_name) @property
+)
+(tag
```
theHamsta, line 10: Please have a look into `CONTRIBTING.md` to see which nodes are available

mikehaertl: But `tag` is a node provided by the phpdoc parser.

https://github.com/claytonrcarter/tree-sitter-phpdoc/blob/master/grammar.js#L40

https://github.com/claytonrcarter/tree-sitter-phpdoc/blob/master/corpus/tests.txt#L30

Or what exactly is wrong here?

theHamsta: Ah, sorry I was confusing the php captures in quotes with the ones from this project :face_palm:

queries/php/injections.scm
```diff
@@ -1,3 +1,3 @@
 (text) @html
 
-(comment) @comment
+(comment) @phpdoc
```
mikehaertl, line 3: I'm not sure if this can be a problem, as this parser is marked as experimental. So what happens if a PHP user has not installed the phpdoc parser? Will this be silently ignored?

mikehaertl: Tested locally (`:TSUninstall phpdoc`). If phpdoc is not installed comment blocks are simply not parsed which is fine IMO.

queries/phpdoc/highlights.scm
```diff
@@ -0,0 +1,29 @@
+(tag_name) @tag
```
mikehaertl, line 1: I think, `@tag` is not really appropriate here. Even though in phpdoc any `@foo` construct is also called "tag" the `@tag` capture group is reserved for XML tags - which this is not. I'm just not sure what to use instead.

theHamsta: Would `@attribute` fit? It's used for C++ attributes (code annotations). The `comment` parsers uses `@text.note`: https://github.com/theHamsta/nvim-treesitter/blob/7d1fb084e2864f8f22e07e61e768561be4dd446f/queries/comment/highlights.scm#L8

queries/php/injections.scm
theHamsta: yes, injections are ignored when the parser is not installed.

queries/phpdoc/highlights.scm
mikehaertl: I've thought about `@attribute`, too. It's at least better than `@tag`. Change pushed.

I'm not sure though if phpdoc can be compared to the comment parser. The latter follows it's own standard for comments whereas phpdoc (like jsdoc) is well defined and has way more features. It would be great if we had something like `@comment.tag`, `@comment.type`, `@comment.variable` etc.

theHamsta:
@mikehaertl is this ready to merge?

mikehaertl:
@theHamsta Yes, ready to merge now. I've just pushed some more fine tunings and I think it's quite solid now. We  can still optimize later if neccessary.

PR #2286, opened 2022-01-17
highlights(php): detect constructor calls on instantiation

mikehaertl:
Required to add a highlighting for

```php
<?php
$x = new MyClass();  // (name)
$x = new app\MyClass();  // (qualified_name)
```

Also moved `@constructor` quries into the Functions section to keep things well organized.


theHamsta (approved):
Thank you!

PR #2288, opened 2022-01-18
docs: refresh CONTRIBUTING.md

vigoux:
- Mention Zulip instead of Gitter
- Use query filetype instead of scheme

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,8 +2,8 @@
 
 First of all, thank you very much for contributing to `nvim-treesitter`.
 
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):

clason:
I agree that Gitter is dead as a dodo, but https://matrix.to/#/#nvim-treesitter:matrix.org is still alive (and least visited by me much more frequently)?

theHamsta:
Sorry, for merged too quick. I'd also say that matrix should be mentioned (as general neovim/tree-sitter platform)

PR #2289, opened 2022-01-18
Prepare for ABI change

vigoux:
TODO:
- [x] Allow per parser configuration
- [x] Pass the `--abi` flag to tree-sitter when generating
- [x] Force generation from grammar for grammars that use a too new ABI

CC @clason

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -105,7 +105,9 @@ end
 
 local function is_lockfile_abi_matching(lang)
   load_lockfile()
-  return (lockfile[lang].abi or 0) <= vim.treesitter.language_version
+  local minimum_version = vim.treesitter.minimum_language_version or 13
+  local locked_version = (lockfile[lang].abi or 0)
+  return locked_version <= vim.treesitter.language_version and locked_version >= minimum_version
 end
 
 local function onread(handle, is_stderr)
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -558,9 +558,14 @@ function M.write_lockfile(verbose, skip_langs)
     if not vim.tbl_contains(skip_langs, v.name) then
       -- I'm sure this can be done in aync way with iter_cmd
       local sha = vim.split(vim.fn.systemlist("git ls-remote " .. v.parser.install_info.url)[1], "\t")[1]
-      lockfile[v.name] = { revision = sha }
-      if verbose then
-        print(v.name .. ": " .. sha)
+      local available, lang = pcall(vim.treesitter.language.inspect_language, v.name)
+      if available then
+        lockfile[v.name] = { revision = sha, abi = lang._abi_version }
+        if verbose then
+          print(string.format("%s: %s ABI %d", v.name, sha, lang._abi_version))
+        end
+      elseif lockfile[v.name] and lockfile[v.name].abi then
+        print("WARNING: could not check ABI version for", v.name)
       end
     else
       print("Skipping " .. v.name)
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -103,6 +103,11 @@ local function outdated_parsers()
   end, info.installed_parsers())
 end
 
+local function is_lockfile_abi_matching(lang)
+  load_lockfile()
+  return lockfile[lang].abi <= vim.treesitter.language_version
+end
+
 local function onread(handle, is_stderr)
   return function(err, data)
     if data then
@@ -384,8 +389,12 @@ local function install_lang(lang, ask_reinstall, cache_folder, install_folder, w
       if not string.match(yesno, "^y.*") then
         return
       end
-      generate_from_grammar = true
+      generate_from_grammar = not is_lockfile_abi_matching(lang)
     end
+  end
+
+  if not is_lockfile_abi_matching(lang) then
+    generate_from_grammar = true
   end
 
   local install_info = get_parser_install_info(lang, true)
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -105,7 +105,7 @@ end
 
 local function is_lockfile_abi_matching(lang)
   load_lockfile()
-  return lockfile[lang].abi <= vim.treesitter.language_version
+  return (lockfile[lang].abi or 0) <= vim.treesitter.language_version
 end
 
 local function onread(handle, is_stderr)
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -319,10 +319,6 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync, 
         },
       })
     end
-    local ts_args = { "generate" }
-    if vim.treesitter.language_version >= 14 then
-      table.insert(ts_args, string.format("--abi=%d", vim.treesitter.language_version))
-    end
     vim.list_extend(command_list, {
       {
         cmd = vim.fn.exepath "tree-sitter",
@@ -477,7 +473,6 @@ end
 function M.update(options)
   options = options or {}
   return function(...)
-    M.lockfile = {}
     reset_progress_counter()
     if ... and ... ~= "all" then
       local languages = vim.tbl_flatten { ... }
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -312,6 +312,10 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync, 
         },
       })
     end
+    local ts_args = { "generate" }
+    if vim.treesitter.language_version >= 14 then
+      table.insert(ts_args, string.format("--abi=%d", vim.treesitter.language_version))
+    end
     vim.list_extend(command_list, {
       {
         cmd = vim.fn.exepath "tree-sitter",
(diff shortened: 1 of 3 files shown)


theHamsta:
Maybe Tree-sitter CLI should be called with the native ABI from Neovim: `--abi`. which will make it work for future ABI updates as well as for old Neovim (when people have the new tree-sitter CLI with that argument)

theHamsta:
Originally, I thought that the CI would just determine the ABI level and write it to `lockfile.json`. Then, the one could only generate when there's a incompatible version. Although one long run, we don't want to blindly compile C files and rather have standardizes deployment aritfacts generated by a GH workflow for upstream tree-sitter, I have the feeling that not everybody would like to have the CLI and node as a requirement for tree-sitter. Generating also takes a while for some parsers. But I'm also fine with always generating.

clason:
To be clear, the motivation was for devs to override and force (easily) to install all parsers with a specified ABI. (Obviously, to test-drive the new tree-sitter commit ...)

vigoux:
Okay, so the last TODO item is not really needed in the end IMO. With the current tooling, as long as the version of the tree-sitter executable, and the version of the runtime built in neovim are the same, the installation will progressively turn the parsers to the correct ABI version.

clason:
Maybe it would be good then to add the ABI of the tree-sitter CLI to the healthcheck? (Either here for the nvim-treesitter check or upstream for the vim.treesitter check (where all the ABI information is now currently).

vigoux:
Hmmm, I don't think we can get the ABI of the tree-sitter cli easily... Maybe we can "just check" that the tree-sitter CLI supports the `--abi` flag ? (which should make the changes easier)

clason:
Ah, OK. Maybe that would be useful information to add to the `--version` flag? @maxbrunsfeld

vigoux:
As requested, the way it works now:
1. Install the parser (possibly wrong ABI for now)
2. Ask for reinstalling
3. You get a prompt: `foo has the wrong ABI: update ?`
4. It reinstalls the parser by regerating

The issue is with parsers we just install, we can't check that the ABI version just yet, and we have to wait for it to be loaded in nvim to check that.

I am going to make `TSUpdate` update the grammars with the wrong ABI too.

clason:
What happens now when tree-sitter CLI 0.20.2 is installed, which doesn't support the `--abi` flag?

clason:
No, I mean, won't the call to `tree-sitter generate --abi=...` not produce an error?

vigoux:
Now for the `TSUpdate` workflow, nothing really changes on the user's side, but when updating, we check for the ABI version, and force generation from grammar in this case.

vigoux:
> No, I mean, won't the call to `tree-sitter generate --abi=...` produce an error?

it would

clason:
That's not good :) We should guard against this.

lua/nvim-treesitter/install.lua
```diff
@@ -361,10 +367,22 @@ local function install_lang(lang, ask_reinstall, cache_folder, install_folder, w
...
+    local parser = inspect_lang(lang)
+    if parser._abi_version == vim.treesitter.language_version then
+      local yesno = fn.input(lang .. " parser already available: would you like to reinstall ? y/n: ")
+      print "\n "
+      if not string.match(yesno, "^y.*") then
+        return
+      end
+    else
+      -- Force the generation from the grammar
+      local yesno = fn.input(lang .. " has an older ABI version: would you like to reinstall ? y/n: ")
```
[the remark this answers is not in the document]
vigoux: That's a good idea. I'll try to dig into that too, but I think that the solution is a bit more difficult.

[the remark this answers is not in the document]
clason: But we want the locked ABI to be (easily) overridable for testing...

lua/nvim-treesitter/install.lua
```diff
@@ -302,13 +304,17 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync,
         },
       })
     end
+    local ts_args = { "generate" }
+    if vim.treesitter.language_version >= 14 then
```
theHamsta, line 308: I think this might also get important for older nvim as soon as parsers are generated with new ABI

vigoux: What do you mean ?

theHamsta: when there are ABI 14 generated parsers, an 13 Neovim could generate the old ABI. But I forgot that 13 is the default so it will do the right thing for now.

vigoux:
So, now that we get the parser ABI version in the lockfile, I think we can get further.
The question now is what to do when the versions do not match ? Generate from the grammar ? What if the tree-sitter CLI and the tree-sitter runtime versions are not in sync ?

maxbrunsfeld:
I'll publish new release of library and CLI later today, or tomorrow at the latest.

theHamsta:
@clason please don't merge master into PRs. Always rebase PRs.

clason:
I didn't (knowingly); I used the Github website to resolve the conflict and didn't realize that that is what it's doing (I actually expected a rebase, like you said). Now I know to steer clear of that button. Feel free to drop the commit.

lua/nvim-treesitter/install.lua
```diff
@@ -361,10 +367,22 @@ local function install_lang(lang, ask_reinstall, cache_folder, install_folder, w
       return
     end
 
-    local yesno = fn.input(lang .. " parser already available: would you like to reinstall ? y/n: ")
-    print "\n "
-    if not string.match(yesno, "^y.*") then
-      return
+    -- Now check if the ABI is not correct
+    local parser = inspect_lang(lang)
```
theHamsta, line 371: there shouldn't be a inspect_lang before install.

lua/nvim-treesitter/install.lua
```diff
@@ -361,10 +367,22 @@ local function install_lang(lang, ask_reinstall, cache_folder, install_folder, w
...
+    if parser._abi_version == vim.treesitter.language_version then
```
theHamsta, line 372: do we want an exact match or are older compatible versions also ok? I think the runtime known it's minimal compatible version.

lua/nvim-treesitter/install.lua
vigoux: Yup, that's why we need the lockfile thing now, I am going to change that

vigoux: Rewording: at this point, we know that the parser is installed, because of the `is_installed` call before hand.

lua/nvim-treesitter/install.lua
vigoux: It indeed knows it, but I am not sure this has been propagated to lua yet
If not I am going to make a PR for that.

lua/nvim-treesitter/install.lua
```diff
@@ -64,8 +64,10 @@ local function get_parser_install_info(lang, validate)
 end
 
 local function load_lockfile()
-  local filename = utils.join_path(utils.get_package_path(), "lockfile.json")
-  lockfile = vim.fn.filereadable(filename) == 1 and vim.fn.json_decode(vim.fn.readfile(filename)) or {}
+  if #lockfile == 0 then
```
theHamsta, line 67: the lockfile needs to be loaded multiple times in order to make hot-reloading on `:TSUpdate` working

lua/nvim-treesitter/install.lua
```diff
@@ -103,6 +105,11 @@ local function outdated_parsers()
   end, info.installed_parsers())
 end
 
+local function is_lockfile_abi_matching(lang)
+  load_lockfile()
+  return lockfile[lang].abi <= vim.treesitter.language_version
```
theHamsta, line 110: so the minimum supported version is currently not exposed in Neovim? We should create an issue and revisit this later when it's exposed or do it via FFI

lua/nvim-treesitter/install.lua
vigoux: Oh right

lua/nvim-treesitter/install.lua
vigoux: Just did a PR to do so, so this should land soon, and we will be able to add that to here

scripts/ci-install-macos-latest.sh
```diff
@@ -3,3 +3,5 @@ sudo ln -s $(pwd)/nvim-osx64/bin/nvim /usr/local/bin
 mkdir -p ~/.local/share/nvim/site/pack/nvim-treesitter/start
 ln -s $(pwd) ~/.local/share/nvim/site/pack/nvim-treesitter/start
 
+brew update
+brew install tree-sitter-cli@${TREE_SITTER_CLI_TAG}
```
clason, line 7: That doesn't work -- homebrew does not have multiple versions except for some special formulas that can be installed in parallel (like Python). You can only install the latest version (without `@${TAG}`).

clason: Also, the formula is just called `tree-sitter`.

clason: And you can skip `brew update`, that is implied by `brew install`.

theHamsta: there were some challenges with node on MacOS/Windows. At the moment, generate from grammar is only done on Ubuntu. When node works you can also do the same `npm` installation like Ubuntu.

clason: `brew install node` should do the trick (that's what I use locally).

lua/nvim-treesitter/install.lua
theHamsta: @vigoux Maybe I was wrong and I'm explicitly setting `lockfile = {}` on reset. Then, this would not be necessary

theHamsta:
We think that the problem that this PR tries to solve does no longer exist https://github.com/nvim-treesitter/nvim-treesitter/pull/2828#issuecomment-1101351590 . Stable Neovim supports both 13 and 14 and the next ABI change is likely to remove pre-generated files altogether from repos so we need to find a different solution. If the next ABI change still keeps generated files we remember that we already have all the code here :slightly_smiling_face:

PR #2290, opened 2022-01-18
feat(java)(highlights): add missing highlights for `@type`

captainko:
Fix:  https://github.com/tree-sitter/tree-sitter-java/issues/94
PR: https://github.com/tree-sitter/tree-sitter-java/pull/95

### DIFF
--- a/queries/java/highlights.scm
+++ b/queries/java/highlights.scm
@@ -90,6 +90,12 @@
 (constructor_declaration
   name: (identifier) @type)
 (type_identifier) @type
+((method_invocation
+  object: (identifier) @type)
+ (#lua-match? @type "^[A-Z]"))
+((method_reference
+  . (identifier) @type)
+ (#lua-match? @type "^[A-Z]"))
 
 
 


queries/java/highlights.scm
```diff
@@ -90,6 +90,12 @@
 (constructor_declaration
   name: (identifier) @type)
 (type_identifier) @type
+((method_invocation
+  object: (identifier) @type)
+ (#lua-match? @type "^[A-Z]"))
```
theHamsta, line 95: Can you help me with that? Do you want to match `Foo::bar()`? In this case you should rather include the `::` instead of doing a regex or do you want to highlight also `Foo.bar()`

theHamsta: My fault, `::` is only used in Java when you need the function for a higher-order function.

theHamsta:
Thank you!

queries/java/highlights.scm
theHamsta, line 95: Can you help me with that? Do you want to match `Foo::bar()`? In this case you should rather include the `::` instead of doing a regex or do you want to highlight also `Foo.bar()`

theHamsta (approved):
Thank you!

PR #2291, opened 2022-01-18
ci: add swift highlight tests

theHamsta:


clason:
The chore commit conflicts with the nvim-treesitter rewrite, and we'd like to rewrite the highlight tests to use Nvim (the real source of truth) anyway, so I'm closing this. The (updated?) swift tests themselves can still be added in a new PR, of course.

PR #2297, opened 2022-01-19
highlights(c): don't highlight type qualifiers in declarations as types

jdrouhard:
@theHamsta This just takes out highlighting of `type_qualifier` as a `@type` as part of a declaration. Qualifiers themselves aren't really types. The one for the cast expression was unnecessary since all `type_descriptors` are already highlighted as types.


theHamsta (approved):

theHamsta:
Thank you!

PR #2314, opened 2022-01-21
ci: bump neovim and tree-sitter cli to latest version

clason:
possibly want to wait to update to tree-sitter 0.21.0 (or whatever the upcoming version is)

also simplify tests:
* only run on push to master _or_ PR opened/synchronized
* cancel in-progress workflows for updated PRs
* only run ubuntu-latest+gcc+nightly on draft PRs
* only run ubuntu+gcc and macos+clang (plus windows x 3)

### DIFF
--- a/.github/workflows/check-query-files-and-compilation.yml
+++ b/.github/workflows/check-query-files-and-compilation.yml
@@ -9,20 +9,20 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-2022, macos-latest]
         cc: [ gcc, clang ]
-        nvim_tag: [ v0.6.0 ]
+        nvim_tag: [ v0.6.1 ]
         exclude:
           - os: macos-latest
             cc: gcc
-            nvim_tag: v0.6.0
+            nvim_tag: v0.6.1
 
         include:
           - os: windows-2022
             cc: cl
-            nvim_tag: v0.6.0
+            nvim_tag: v0.6.1
 
           - os: macos-latest
             cc: gcc-10
-            nvim_tag: v0.6.0
+            nvim_tag: v0.6.1
 
           - os: ubuntu-latest
             cc: gcc
@@ -40,7 +40,7 @@ jobs:
       - name: Install and prepare Neovim
         env:
           NVIM_TAG: ${{ matrix.nvim_tag }}
-          TREE_SITTER_CLI_TAG: v0.20.1
+          TREE_SITTER_CLI_TAG: v0.20.2
         run: |
           bash ./scripts/ci-install-${{ matrix.os }}.sh
 
--- a/.github/workflows/check-query-files-and-compilation.yml
+++ b/.github/workflows/check-query-files-and-compilation.yml
@@ -23,6 +23,10 @@ jobs:
         cc: [ gcc, clang ]
         nvim_tag: [ v0.6.1 ]
         exclude:
+          - os: ubuntu-latest
+            cc: clang
+            nvim_tag: v0.6.1
+
           - os: macos-latest
             cc: gcc
             nvim_tag: v0.6.1
@@ -30,10 +34,6 @@ jobs:
         include:
           - os: windows-2022
             cc: cl
-            nvim_tag: v0.6.1
-
-          - os: macos-latest
-            cc: gcc-10
             nvim_tag: v0.6.1
 
           - os: ubuntu-latest
--- a/.github/workflows/check-query-files-and-compilation.yml
+++ b/.github/workflows/check-query-files-and-compilation.yml
@@ -1,6 +1,18 @@
 name: Parser compilation and query file check
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - 'master'
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+    branches:
+      - 'master'
+
+# Cancel any in-progress CI runs for a PR if it is updated
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.head_ref || github.sha }}
+  cancel-in-progress: true
 
 jobs:
   check_compilation_unix_like:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,8 +28,8 @@ jobs:
 
       - name: Install and prepare Neovim
         env:
-          NVIM_TAG: v0.6.0
-          TREE_SITTER_CLI_TAG: v0.20.1
+          NVIM_TAG: v0.6.1
+          TREE_SITTER_CLI_TAG: v0.20.2
         run: |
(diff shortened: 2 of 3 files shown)


theHamsta:
Probably you can reduce to one compiler on Ubuntu/Mac

clason:
Historically, what were the outliers (platform+compiler combinations that failed even when the others passed)?

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -1,37 +1,83 @@
...
+      - uses: actions/upload-artifact@v2
+        with:
+          name: parsers-${{ matrix.os }}-${{ matrix.cc }}-x86_64
+          path: parser/*
+
+      - name: Check query files (Unix)
+        if: matrix.os != 'windows-2022'
+        run: nvim --headless -c "luafile ./scripts/check-queries.lua" -c "q"
+
+  check_compilation_all:
```
theHamsta, line 55: I'm a bit worried about this file getting too complicated by this copy and paste. I'd rather just kick out configuration and cache parser installation

clason: I think it's still fine; Neovim does the same. (If there's a better way of skipping workflows inside a matrix job, I'm all ears, though.)

clason: (I thought about simplifying the draft job by hardcoding everything, but that seemed even more error-prone.)

clason:
The biggest bottleneck is the three Windows compilers. If we can reduce that, that would be nice.

theHamsta:
@clason parallel compilation speeds up compilation on MacOS/Linux by huge factor but on Windows it will cause a crash due to OOM on the VM :smile:  when using MSVC. All in all, the CI is not a huge issue 8 minutes aren't too bad with auto-merge.

clason:
Windows clang takes 22 minutes...

clason:
Anyway, I'd like to merge this ASAP so we can build on top of it. I can remove the special case draft-only job if you wish, if that is what is blocking this?

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -1,37 +1,83 @@
 name: Parser compilation and query file check
```
MunifTanjim, line 1: Why are the artifacts uploaded in this job? Are they used for something?

clason: I don't know; I believe there might be people (or workflows) who don't want to build their own parsers?

MunifTanjim: Doing these on each commit seems like a waste 🤔 

image

clason: Agreed.

MunifTanjim: Anyway, it's not related to this PR. I was just curious.

theHamsta: It was mostly for debugging because sometimes the CI would build successfully artifacts that wouldn't work. The Windows wiki has a hint that users can download the parsers from CI. But it would probably be better to have binary releases for that.

clason: Let me know if I should remove that step.

theHamsta: >Doing these on each commit seems like a waste thinking

Well, we're not paying the servers :smile: was wondering how much data we caused by that. They seem to be persistent for a pretty long time. I was expecting that at some point we'd either hit API limits or that stuff get's quickly ejected.

theHamsta: Probably, they should be replaced by releases. I was experimenting a bit with that https://github.com/theHamsta/nvim-treesitter/releases/tag/nightly-parsers releases would also have a predictable path for download (e.g. tagged releases of our plugin could auto-download the binaries for the platform if a user opts in). But even better would be to do that outside of our org.

MunifTanjim: > Well, we're not paying the servers

I meant "waste" as in "doing unnecessary work repeatedly as it is producing and uploading the same files for each commit" 😂 I'm sure github removes old files once it hits the limit.

theHamsta: or ghcr.io ?

clason: (+1 on releases, but not sure why outside of the org would be better?)

theHamsta: Because tree-sitter 1.0 wants to offer an easy/standardized way for parsers to publish without having generated files in their repo. But sure for us it would make sense to have all PRs

clason: Oh, I see. Yes, if upstream took care of this, that'd be cake :)

theHamsta: @clason See here how releases would work https://github.com/theHamsta/nvim-treesitter/commit/a0e29464e75bc1cf146523379d2d274c6f09d4b6 . You can only do this on tags because there you can hit API limits fairly quickly :smile:

clason: Yeah, we would probably want to do a daily job or something for that.

Anyway, all that is for a different PR -- is there anything I need to change for this one? I'm just waiting for your approval ;)

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -1,37 +1,83 @@
...
 
 jobs:
-  check_compilation_unix_like:
+  check_compilation_unix_latest:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-latest ]
+        cc: [ gcc ]
+        nvim_tag: [ nightly ]
```
theHamsta, line 24: This shouldn't be nightly. Our main CI is latest stable on Ubuntu. Nightly is not always available for download and is therefore allowed to fail

theHamsta: It is also very annoying to change the required builds. You won't be able to merge this and I need to change it manually after force-merging this PR

clason: OK, so you _do_ want me to drop the fast draft CI -- you just have to say so!

theHamsta (approved):
Nice improvement! Feeling bad for having done this before...

clason:
No worries! @MunifTanjim now you can go ahead :)

theHamsta:
@clason this PR changed the required test names. All PRs will have to rebase now to be able to get merged!

clason:
I don't think it did? Or was the `0.6.0` part of the name?

(In this case, we should switch to using the `stable` tag...)

PR #2316, opened 2022-01-21
ci: cache parsers for workflows/tests

MunifTanjim:


clason:
we also need this for the `check-query-files-and-compilation` job?

MunifTanjim:
> we also need this for the check-query-files-and-compilation job?

Yep, need to add it there too. Just checking if this works first. Waiting for it to pass so that I can retrigger.

theHamsta (approved):
And ccache in addition to that on Mac and Linux?

- https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/
- https://github.com/hendrikmuhs/ccache-action

Another use for actions/cache would be `~/.cargo`, `~/.rustup` in tests.yml

.github/workflows/tests.yml
```diff
@@ -33,8 +33,14 @@ jobs:
         run: |
           bash ./scripts/ci-install-${{ matrix.os }}.sh
 
+      - name: Setup Parsers Cache
+        id: parsers-cache
+        uses: actions/cache@v2
+        with:
+          path: ./parser/
+          key: ${{ matrix.os }}-parsers-v1-${{ hashFiles('./lockfile.json', './lua/nvim-treesitter/parsers.lua', './lua/nvim-treesitter/install.lua') }}
```
theHamsta, line 41: could you add `shell_command_selectors.lua`?

theHamsta (approved):

MunifTanjim:
Oh, this is already merged! Should I open another PR for the rest of the things? Or somebody is already taking care of it? @theHamsta

theHamsta:
@MunifTanjim If you want to do it you can create another PR. The ccache stuff is probably not worth the effort but the Rust stuff certainly. I tried to compile a static binary on my repo but libmusl doesn't support dynamic loading (but could upload a build from Ubuntu 20.04 and download it in CI)

clason:
@theHamsta you didn't wait for the PR to be actually finished ;) There was quite a bit missing.

So yes, please open a second PR for the rest (triggers, other workflow) -- although it'd help to have my PR merged first.

MunifTanjim:
> it'd help to have my PR merged first

In that case, let's wait for it to get merged. I'll open another one later.

PR #2319, opened 2022-01-21
ci: use stable tag for latest nvim release

clason:
so we only need to rebase all old PRs once

@theHamsta

### DIFF
--- a/.github/workflows/update-readme.yml
+++ b/.github/workflows/update-readme.yml
@@ -14,7 +14,7 @@ jobs:
 
       - name: Prepare
         env:
-          NVIM_TAG: v0.6.1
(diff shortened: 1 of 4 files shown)


.github/workflows/update-readme.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
       - name: Prepare
         env:
-          NVIM_TAG: v0.6.1
+          NVIM_TAG: stable
```
theHamsta, line 17: there's also v0.5.1 used once in another workflow

clason: nope, not anymore ;)

theHamsta (approved):

PR #2320, opened 2022-01-21
Cache more on CI

theHamsta:


.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -56,15 +56,29 @@ jobs:
...
         if: matrix.os != 'windows-2022'
         run: |
+          export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
           nvim --headless -c "TSInstallSync all" -c "q"
 
       - name: Compile parsers Windows
         if: matrix.os == 'windows-2022'
         run: |
-          Neovim\\bin\\nvim.exe --headless -c "TSInstallSync all" -c "q"
+          Neovim\\bin\\nvim.exe --headless -c "lua require'nvim-treesitter.install'.prefer_git=false'"-c "TSInstallSync all" -c "q"
```
theHamsta, line 81: there's something wrong here....

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -56,15 +56,29 @@ jobs:
         run: |
           bash ./scripts/ci-install-${{ matrix.os }}.sh
 
+      - name: Setup Parsers Cache
+        id: parsers-cache
+        uses: actions/cache@v2
+        with:
+          path: ./parser/
+          key: ${{ matrix.os }}-parsers-v1-${{ hashFiles('./lockfile.json', './lua/nvim-treesitter/parsers.lua', './lua/nvim-treesitter/install.lua') }}
```
clason, line 64: missing `shell_command_selectors.lua`?

.github/workflows/check-query-files-and-compilation.yml
theHamsta: maybe it works without the final quote.

clason (approved):
not familiar with these actions, but looks good to me apart from the comment!

theHamsta:
We just have to make sure that curl is used on Windows.

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -56,15 +56,29 @@ jobs:
...
           bash ./scripts/ci-install-${{ matrix.os }}.sh
 
+      - name: Setup Parsers Cache
+        id: parsers-cache
+        uses: actions/cache@v2
+        with:
+          path: ./parser/
+          key: ${{ matrix.os }}-parsers-v1-${{ hashFiles('./lockfile.json', './lua/nvim-treesitter/parsers.lua', './lua/nvim-treesitter/install.lua', './lua/nvim-treesitter/shell_selectors.lua') }}
+
+      - name: ccache
```
theHamsta, line 66: not really worth it. longest step seems to be repo extraction

theHamsta: this would also be better with parallel downloading/compilation

theHamsta:
What? 
image

theHamsta:
@MunifTanjim apparently the long test times seem to come from loading the Swift parser...

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -56,15 +56,29 @@ jobs:
...
+        if: matrix.os != 'windows-2022'
+        uses: hendrikmuhs/ccache-action@v1
+        with:
+          key: ${{ matrix.os }}-${{ matrix.cc }}
```
MunifTanjim, line 70: Don't really know about what this cache will contain, but it seems like this will cache it forever? Won't it grow in size over-time?

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -56,15 +56,31 @@ jobs:
         run: |
           bash ./scripts/ci-install-${{ matrix.os }}.sh
 
+      - name: Setup Parsers Cache
+        id: parsers-cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ./parser/
+            ~/AppData/Local/nvim/pack/nvim-treesitter/start/nvim-treesitter/parser/
```
theHamsta, line 65: Windows doesn't create a symlink from cwd to the plugin folder

.github/workflows/check-query-files-and-compilation.yml
theHamsta: no, it's a 500MB ccache size

theHamsta: no really a benefit, as extracting takes longer than compiling.

.github/workflows/check-query-files-and-compilation.yml
```diff
@@ -56,15 +56,31 @@ jobs:
...
+          key: ${{ matrix.os }}-parsers-v1-${{ hashFiles('./lockfile.json', './lua/nvim-treesitter/parsers.lua', './lua/nvim-treesitter/install.lua', './lua/nvim-treesitter/shell_selectors.lua') }}
```
MunifTanjim, line 66: I think workflows shares cache key, so we have a conflict:

image

theHamsta: yes, also the cache is shared between distinct compilers. It's missing in the cache key and maybe for concurrency also cache nvim tag

.github/workflows/tests.yml
```diff
@@ -30,13 +30,19 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
 
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+
+      - uses: Swatinem/rust-cache@v1
```
MunifTanjim, line 38: Is this working as expected?

image

.github/workflows/check-query-files-and-compilation.yml
MunifTanjim: Those variables should be included in cache key.

.github/workflows/tests.yml
theHamsta: maybe because we don't have real rust repo. Let's just cache ~/.cargo even though we can't update the tool

.github/workflows/tests.yml
```diff
@@ -35,7 +35,13 @@ jobs:
           profile: minimal
           toolchain: nightly
 
-      - uses: Swatinem/rust-cache@v1
+      - name: Cache .cargo
+        id: parsers-cache
```
MunifTanjim, line 39: This id needs to be changed... might cause issues because of conflict otherwise. 🤔

MunifTanjim: Yep.
image

theHamsta:
Now it's also fast on Windows.

theHamsta:
Only Ubuntu GCC on v0.6.1 has the problem of long loading swift parser.
https://github.com/nvim-treesitter/nvim-treesitter/runs/4901528734?check_suite_focus=true

.github/workflows/tests.yml
```diff
@@ -30,13 +30,25 @@ jobs:
...
+        with:
+          profile: minimal
+          toolchain: nightly
+
+      - name: Cache .cargo
+        id: cargo-cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ./cargo/
```
MunifTanjim, line 43: One last typo 😛 `~/.cargo/`

MunifTanjim (approved):

.github/workflows/tests.yml
MunifTanjim: Not sure if this is working 🤔

image

Is `./.cargo` and `~/.cargo` same? Probably not, `./` is relative to repo.

theHamsta:
Why does this only happen on Ubuntu/GCC
```
5. 4571.39ms kotlin highlights
4. 4614.91ms elixir locals
3. 5267.51ms swift locals
2. 6508.01ms elixir highlights
1. 421931.10ms swift highlights
----------------------------------------------------------------------------------------------------
Check successful!
```

PR #2326, opened 2022-01-22
highlights(r): change groups priority

aspeddro:


theHamsta (commented):
Can you give a comment why these reorderingx were necessary?

@jimhester

aspeddro:
The reordering is so that `TSMethod`, `TSNamespace` takes precedence over `TSVariable`.

image

With the current order the highlights are not applied correctly.

image

theHamsta (approved):

PR #2327, opened 2022-01-22
docs: add missing vim table in the example

monkoose:
Improves clarity for a user that is not familiar with lua api.


theHamsta (approved):

theHamsta:
Thanks for discovering this!

PR #2328, opened 2022-01-22
feat: specify ABI version for generate on newer tree-sitter CLI

clason:
Check tree-sitter CLI version and if > 0.20.3 and generating a parser from grammar, use `--abi=vim.treesitter.language_version`.

Besides being able to opt-in to newer ABI benefits, this is a necessary workaround for an upstream bug with 0.20.3, where `parser.h` is not generated if the (optional) `--abi` flag is omitted.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -14,6 +14,7 @@ local lockfile = {}
 M.compilers = { vim.fn.getenv "CC", "cc", "gcc", "clang", "cl", "zig" }
 M.prefer_git = fn.has "win32" == 1
 M.command_extra_args = {}
+M.ts_generate_args = nil
 
 local started_commands = 0
 local finished_commands = 0
@@ -259,6 +260,15 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync, 
       )
     end
     return
+  else
+    if not M.ts_generate_args then
+      local ts_cli_version = utils.ts_cli_version()
+      if ts_cli_version and vim.split(ts_cli_version, " ")[1] > "0.20.2" then
+        M.ts_generate_args = { "generate", "--abi", vim.treesitter.language_version }
+      else
+        M.ts_generate_args = { "generate" }
+      end
+    end
   end
   if generate_from_grammar and vim.fn.executable "node" ~= 1 then
     api.nvim_err_writeln "Node JS not found: `node` is not executable!"
@@ -308,7 +318,7 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync, 
         info = "Generating source files from grammar.js...",
         err = 'Error during "tree-sitter generate"',
         opts = {
-          args = { "generate" },
+          args = M.ts_generate_args,
           cwd = compile_location,
         },
       },
--- a/lua/nvim-treesitter/utils.lua
+++ b/lua/nvim-treesitter/utils.lua
@@ -193,4 +193,13 @@ function M.to_func(a)
   return type(a) == "function" and a or M.constant(a)
 end
 
+function M.ts_cli_version()
+  if fn.executable "tree-sitter" == 1 then
+    local handle = io.popen "tree-sitter  -V"
+    local result = handle:read "*a"
+    handle:close()
+    return vim.split(result, "\n")[1]:match "[^tree%psitter ].*"
+  end
+end
+
 return M
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -5,6 +5,7 @@ local queries = require "nvim-treesitter.query"
 local info = require "nvim-treesitter.info"
 local shell = require "nvim-treesitter.shell_command_selectors"
 local install = require "nvim-treesitter.install"
+local utils = require "nvim-treesitter.utils"
 
 local health_start = vim.fn["health#report_start"]
 local health_ok = vim.fn["health#report_ok"]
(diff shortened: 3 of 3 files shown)


clason:
Hmm, that doesn't seem to work (although it did locally, at least for D).

clason:
@theHamsta My plan now is to refactor the health check into an `utils.check_ts_cli` that returns `ok, version` and use that to decide whether to set the `--abi` flag for `tree-sitter generate`. How does that sound?

clason:
@theHamsta I implemented that strategy now (with caching so we don't have to fondle the executable on every install from grammar); I think this is mergable now.

clason:
Here's what this buys us, as illustrated by CI: checking `swift` queries on 

* [Neovim stable (ABI 13)](https://github.com/nvim-treesitter/nvim-treesitter/runs/4907578908?check_suite_focus=true#step:10:91):
```
Checking swift highlights (300783.08ms)
Checking swift locals (4304.71ms)
```

* [Neovim nightly (ABI 14)](https://github.com/nvim-treesitter/nvim-treesitter/runs/4907579001?check_suite_focus=true#step:10:6):
```
Checking swift highlights (1380.49ms)
Checking swift locals (378.42ms)
```

vigoux:
Damn that's great!

lua/nvim-treesitter/install.lua
```diff
@@ -14,6 +14,7 @@ local lockfile = {}
 M.compilers = { vim.fn.getenv "CC", "cc", "gcc", "clang", "cl", "zig" }
 M.prefer_git = fn.has "win32" == 1
 M.command_extra_args = {}
+M.ts_generate_args = nil
```
vigoux, line 17: Do we need to make those public?

lua/nvim-treesitter/utils.lua
```diff
@@ -193,4 +193,13 @@ function M.to_func(a)
   return type(a) == "function" and a or M.constant(a)
 end
 
+function M.ts_cli_version()
+  if fn.executable "tree-sitter" == 1 then
+    local handle = io.popen "tree-sitter  -V"
+    local result = handle:read "*a"
+    handle:close()
+    return vim.split(result, "\n")[1]:match "[^tree%psitter ].*"
```
vigoux, line 201: Wo this regex looks weird, are you sure you wanted that?

clason: I stole this 1:1 from the healthcheck function ;)

lua/nvim-treesitter/install.lua
clason: Good question. Does it hurt? I may come in useful (this way you can override it easily, because when it's non-nil, it won't get touched further down).

lua/nvim-treesitter/utils.lua
clason: So the point here is that `tree-sitter -V` outputs something like
```
tree-sitter 0.20.3 (<commit hash if built from source>)
```
and we need to extract the number for comparison.

vigoux: But why adding the `[]` that's what I don't get

lua/nvim-treesitter/install.lua
vigoux: Huh indeed.

lua/nvim-treesitter/utils.lua
clason: 🤷🏻 

Lua regex is still a bit of a mystery to me; I just know that this works for our purpose.

clason: I should not say "regex"; Lua does _not_ have regular expressions, it just has pattern matching. In this case (quoting from the [Lua-users wiki](http://lua-users.org/wiki/PatternsTutorial):

> You can also create your own classes by wrapping a group of characters in square brackets. This will match one of the characters. If the first character inside the brackets is ^, then it will match a character not in the group.

theHamsta (approved):
LGTM, although it would probably be sufficient to call this function once. It only takes 47ms to invoke the version information so it doesn't matter.

lua/nvim-treesitter/install.lua
```diff
@@ -259,6 +260,15 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync,
       )
     end
     return
+  else
+    if not M.ts_generate_args then
+      local ts_cli_version = utils.ts_cli_version()
+      if ts_cli_version and vim.split(ts_cli_version, " ")[1] > "0.20.2" then
```
theHamsta, line 266: can you make this comparison a helper function and display in healthcheck that the user should update?

clason: > LGTM, although it would probably be sufficient to call this function once. It only takes 47ms to invoke the version information so it doesn't matter.

The function is only called once?

clason: > can you make this comparison a helper function and display in healthcheck that the user should update?

I deliberately did not, because I did not want to throw away the commit info when building from source. And the point of this check is to be able to work seamlessly even when tree-sitter CLI is not up-to-date, because I did not think we want to bump the _required_ tree-sitter version?

clason:
I'm merging this now, which does not mean I don't want to address the comments; I'm planning a follow-up PR to add a general "generate_from_source" override option, where I'm happy to add other improvements.

theHamsta:
@clason I don't think an overwrite option is necessary. `:TSInstallFromSync all` already works if you really want o do other not great experience:
- it only works for some parsers
- depends on your environment
- freezes your computer

clason:
Yeah, that's why I wanted an option to install a list of parsers in `ensure_installed` from grammar. It's also very much meant only for people wanting to test parsers on newer API, not for general consumption.

PR #2329, opened 2022-01-22
Remove ccache

theHamsta:


clason (approved):

clason:
Windows+clang still takes >20 minutes; maybe we want to remove that combination?

theHamsta:
sure, can you out-comment it? Also Mac+gcc should not give additional insights

clason:
Mac+gcc is already gone :)

PR #2330, opened 2022-01-22
ci: drop windows+clang

clason:
takes significantly longer than other combinations

probably easier now to just list the allowed combinations?


clason:
much better!

(The swift query check is worrisome, though. I'll bump the parser to see if getting rid of typescript helps.)

theHamsta (approved):

PR #2333, opened 2022-01-22
Diverse fixes for C indents and aligned_indent

### LINKED ISSUE #2369: golang: wrong indentation on new lines in certain conditions
### Describe the bug

When inserting a new line, depending on what seems to be the way to write outer function parameters (single line or multine), indentation is off.

### To Reproduce

go.mod:
```go
module test
```

main.go:

```go
package main

import "fmt"

func goodIndent(param string) {
	fmt.Println("typing o here works as expected")
}

func badIndent(
	param string, // this is the difference
) {
	fmt.Println("typing o here triggers bad indent")
}
```
If you open this file as part of a project (with `vim .`) and you add a line in the second function with `o`, indentation is off. It seems to depend on the way to write the parameters of the outer function.
disabling `indent=true` in treesitter config fixes the problem.

### Expected behavior

Indentation should be correct.

This is what is expected:
image

But this is what's happening:
image

```markdown
## Installation
  - OK: `tree-sitter` found 0.20.4 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v17.3.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
    Version: cc (GCC) 11.1.0

## Parser/Features H L F I J
  - ocaml_interface✓ ✓ ✓ . ✓ 
  - beancount      ✓ . ✓ . . 
  - python         ✓ ✓ ✓ ✓ ✓ 
  - sparql         ✓ ✓ ✓ ✓ ✓ 
  - comment        ✓ . . . . 
  - lua            ✓ ✓ ✓ ✓ ✓ 
  - ocaml          ✓ ✓ ✓ . ✓ 
  - ninja          ✓ . ✓ ✓ . 
  - nix            ✓ ✓ ✓ . ✓ 
  - glsl           ✓ ✓ ✓ ✓ ✓ 
  - json           ✓ ✓ ✓ ✓ . 
  - pug            ✓ . . . ✓ 
  - ledger         ✓ . ✓ ✓ ✓ 
  - php            ✓ ✓ ✓ ✓ ✓ 
  - clojure        ✓ ✓ ✓ . ✓ 
  - supercollider  ✓ ✓ ✓ ✓ ✓ 
  - heex           ✓ ✓ ✓ ✓ ✓ 
  - godot_resource ✓ ✓ ✓ . . 
  - typescript     ✓ ✓ ✓ ✓ ✓ 
  - fennel         ✓ ✓ . . ✓ 
  - query          ✓ ✓ ✓ ✓ ✓ 
  - cpp            ✓ ✓ ✓ ✓ ✓ 
  - vue            ✓ . ✓ ✓ ✓ 
  - latex          ✓ . ✓ . ✓ 
  - make           ✓ . . . ✓ 
  - rst            ✓ ✓ . . ✓ 
  - css            ✓ . ✓ ✓ ✓ 
  - glimmer        ✓ . . . . 
  - erlang         . . . . . 
  - gowork         ✓ . . . ✓ 
  - regex          ✓ . . . . 
  - norg           . . . . . 
  - scss           ✓ . . ✓ . 
  - svelte         ✓ . ✓ ✓ ✓ 
  - prisma         ✓ . . . . 
  - http           ✓ . . . ✓ 
  - c              ✓ ✓ ✓ ✓ ✓ 
  - teal           ✓ ✓ ✓ ✓ ✓ 
  - java           ✓ ✓ . ✓ ✓ 
  - llvm           ✓ . . . . 
  - hjson          ✓ ✓ ✓ ✓ ✓ 
  - gomod          ✓ . . . ✓ 
  - dart           ✓ ✓ . ✓ ✓ 
  - verilog        ✓ ✓ ✓ . ✓ 
  - pioasm         ✓ . . . ✓ 
  - json5          ✓ . . . ✓ 
  - bash           ✓ ✓ ✓ . ✓ 
  - vim            ✓ ✓ . . ✓ 
  - cmake          ✓ . ✓ . . 
  - yaml           ✓ ✓ ✓ ✓ ✓ 
  - zig            ✓ . ✓ ✓ ✓ 
  - bibtex         ✓ . ✓ ✓ . 
  - r              ✓ ✓ . . . 
  - devicetree     ✓ ✓ ✓ ✓ ✓ 
  - julia          ✓ ✓ ✓ ✓ ✓ 
  - pascal         ✓ ✓ ✓ ✓ ✓ 
  - gdscript       ✓ ✓ . ✓ ✓ 
  - jsdoc          ✓ . . . . 
  - turtle         ✓ ✓ ✓ ✓ ✓ 
  - ql             ✓ ✓ . ✓ ✓ 
  - toml           ✓ ✓ ✓ ✓ ✓ 
  - tsx            ✓ ✓ ✓ ✓ ✓ 
  - fusion         ✓ ✓ ✓ ✓ . 
  - html           ✓ ✓ ✓ ✓ ✓ 
  - yang           ✓ . ✓ . . 
  - hocon          ✓ . . . ✓ 
  - surface        ✓ . ✓ ✓ ✓ 
  - jsonc          ✓ ✓ ✓ ✓ ✓ 
  - dot            ✓ . . . ✓ 
  - c_sharp        ✓ ✓ ✓ . ✓ 
  - elixir         ✓ ✓ ✓ ✓ ✓ 
  - cuda           ✓ ✓ ✓ ✓ ✓ 
  - commonlisp     ✓ ✓ ✓ . . 
  - eex            ✓ . . . ✓ 
  - kotlin         ✓ ✓ ✓ . ✓ 
  - ruby           ✓ ✓ ✓ ✓ ✓ 
  - fish           ✓ ✓ ✓ ✓ ✓ 
  - ocamllex       ✓ . . . ✓ 
  - scala          ✓ . ✓ . ✓ 
  - dockerfile     ✓ . . . ✓ 
  - rust           ✓ ✓ ✓ ✓ ✓ 
  - go             ✓ ✓ ✓ ✓ ✓ 
  - perl           ✓ . . . . 
  - tlaplus        ✓ . ✓ . ✓ 
  - graphql        ✓ . . ✓ ✓ 
  - rasi           ✓ ✓ ✓ ✓ . 
  - hcl            ✓ . ✓ ✓ ✓ 

```

```text
NVIM v0.6.1
Build type: Release
LuaJIT 2.0.5
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

```

### Additional context

_No response_

theHamsta:
WIP

Fixes #2369


tests/indent/c/unfinished_comment.c
```diff
@@ -0,0 +1,2 @@
+/*
+ *
```
theHamsta, line 2: This could be fixed by
```
((comment) @aligned_indent
 (#lua-match? @aligned_indent "^/\*"))
```
but strangly this breaks the unfinished call expression in `tests/indent/c/aligned_indent.c`

lua/nvim-treesitter/indent.lua
```diff
@@ -57,6 +56,11 @@ function M.get_indent(lnum)
     return -1
   end
 
+  -- Reparse in case we got triggered by ":h indentkeys"
+  parser:parse()
```
theHamsta, line 60: while this is probably the right thing to do, it breaks the `join_lines.py` test, which is very pain full for Python devs

theHamsta:
TODO: lua parser accepts `  function ofo()` but does not indent after it

lua/nvim-treesitter/indent.lua
theHamsta: Reparse also breaks: yaml after foo: | <cr>

theHamsta: maybe we, could enable this only for some languages?

MunifTanjim:
> lua parser accepts `function ofo()` (incomplete, at the end of a file) but does not indent after it

We don't really have much control over treesitter error recovery.

Just `function ofo()` is not valid Lua, and this is what treesitter generates after error recovery:

image

Not really sure if anything can be done here.

tests/indent/lua/comment.lua
```diff
@@ -5,3 +5,9 @@
   another
   comment
 --]]
+
+  --[[
+    another
+    comment
+    a bit more indented
+  --]]
```
MunifTanjim, line 13: I don't think comment should start from arbritary indentation level. Maybe wrap it in a function if that was the intention?

```suggestion
function comment_inside_func()
  --[[
    another
    comment
    a bit more indented
  --]]
end
```

lua/nvim-treesitter/indent.lua
MunifTanjim: Doesn't treesitter reparses automatically after each character is inserted in buffer? 🤔

yioneko: No, it doesn't. Currently the reparse executes on every screen redraw: https://github.com/neovim/neovim/blob/7717f38d3f81cc739ef2b37829558d3d9eba9591/runtime/lua/vim/treesitter/highlighter.lua#L301-L305. I've encountered the exactly same problem, the reparse before calculation is definitely needed.

MunifTanjim: Just asking out of curiosity, what is a "buffer redraw" for neovim? If I hit a key on keyboard and that character appears on the buffer, is the buffer redrawn? Or does it refer to somthing else?

yioneko: The redraw refers to the update of the screen display, usually the characters sequence output by neovim to terminal. You can look at https://github.com/neovim/neovim/blob/master/src/nvim/screen.c for more detailed explanation.

tests/indent/python/branches.py
```diff
@@ -23,7 +23,7 @@ def foo(x, y):
     b)
 
 if (a and
-    b):
+        b):
```
theHamsta, line 26: actually, my python linter recommends this way were you indent the second line of the parenthesis.

CONTRIBUTING.md
```diff
@@ -276,6 +276,7 @@ the node describing the language and `@content` to describe the injection region
 @branch         ; Dedent itself when matching this node
 @ignore         ; Do not indent in this node
 @auto           ; Behaves like 'autoindent' buffer option
+@zero_indent    ; Sets this node at position 0 (no indent)
```
theHamsta, line 279: probably change to fixed-indent in future

tests/indent/lua/comment.lua
theHamsta: we need some mechanism to keep indent relative to beginning of node.

theHamsta:
Strangly, the failing test is passing on my machine.
```
                        |/**                                         |/**
              0 vs 1 => |* Function foo                              | * Function foo
              0 vs 1 => |* @param[out] x output                      | * @param[out] x output
              0 vs 1 => |* @param[in] x input                        | * @param[in] x input
              0 vs 1 => |*/                                          | */
                        |void foo(int *x, int y) {                   |void foo(int *x, int y) {
                        |    *x = y;                                 |    *x = y;
                        |}                                           |}
```

theHamsta:
```
 4 vs 4 => |    // // indendation with `enter` in insert mode is not correct    |    // indendation with `enter` in insert mode is not correct
```
With `// ` will be added automatically, but if you leave it way in the test it's not in the expectation.

rtransat:
Do you know when the PR will be merge? :wink:

clason:
When it's ready. (Note that it's still in draft, and has several conflicts.)

In general, such comments are not helpful (even with a smiley) as they put pressure on people who already do a lot of volunteer work in their spare time. I've seen more than one person quit contributing because they were fed up with constantly being nudged in this way.

rtransat:
I didn't mean to be mean. It was just a question because the fix interest me a lot but I'm not in the rush to have this feature tonight.

Thank for your reply, I'll wait for the time it takes.

yioneko:
@rtransat If you are interested, you could try my plugin https://github.com/yioneko/nvim-yati as a temporary solution. Most of the test cases of it are ported from this repo and all of them works fine there. I've also added detailed document and explanations in the code base recently, hopefully a useful reference for this PR.

theHamsta:
Can someone test the failing test? Both are running for me locally fine (also when editing). Otherwise this PR should be ready for review. We can also mark the failing tests as flaky as they seem to pass/fail depending on local setup (?) to unblock the other changes.

kyazdani42:
Sorry @theHamsta i cannot run the tests since i moved to nix :smile: I load neovim in a very bizarre way which makes the whole thing load my entire config when running the tests.

lua/nvim-treesitter/indent.lua
```diff
@@ -114,31 +132,36 @@ function M.get_indent(lnum)
...
 
     -- do not indent for nodes that starts-and-ends on same line and starts on target line (lnum)
-    if not is_processed_by_row[srow] and (q.indent[node:id()] and srow ~= erow and srow ~= lnum - 1) then
+    if
+      not is_processed_by_row[srow]
+      -- Dear stylua, please don't change the semantics of this statement!
+      -- stylua: ignore start
+      and (q.indent[node:id()] and srow ~= erow and ((srow ~= lnum - 1) or q.indent[node:id()].start_at_same_line))
+      -- stylua: ignore end
+    then
```
kyazdani42, line 141: this might be factored out at some point, it's quite noisy :smile:

lua/nvim-treesitter/indent.lua
```diff
@@ -114,31 +132,36 @@ function M.get_indent(lnum)
...
           local _, o_scol = o_delim_node:start()
-          local aligned_indent = math.max(indent, 0) + o_scol
-          if indent > 0 then
-            indent = aligned_indent
-          else
-            indent = aligned_indent + 1 -- extra space for starting delimiter
-          end
-          is_processed = true
+          o_scol = o_scol + (metadata.increment or 1)
+          return math.max(indent, 0) + o_scol
```
kyazdani42, line 164: avoid modifying the parameter maybe ?
```suggestion
          return math.max(indent, 0) + o_scol + (metadata.increment or 1)
```

kyazdani42 (commented):
looks good to me otherwise, lets fix this failing test first :)

MunifTanjim:
> Can someone test the failing test? Both are running for me locally fine (also when editing).

Those tests fail for me too 🤔 I'm running `./scripts/run_tests.sh indent`, how are you running tests locally? @theHamsta

theHamsta:
@kyazdani42 you can test the behavior by editing  the actual test files. I don't know how to fix the broken tests since I can't reproduce the failure. I suppose the CI has some different set of ftplugins loaded or different `indentkeys` `cindentkeys` set.

@MunifTanjim I'm using `./scripts/run_tests.sh indent`. Would it be possible for you to fix the tests since you can reproduce the failure? I'm just using `aligned_indent` for the C test to aligin the `*` one space relative to `/*`. This might require a re-indent after typing `*` which might not happen on CI.

MunifTanjim:
Those fails when I have `comment` parser installed. Can you try installing that parser?

When that parser is available, the indent module is working on the tree created by that parser.

theHamsta:
Ok, thanks for the hint! Then, the fix is easy. Just ignore the comment  parser (and phpdoc/jsdoc) from indentation

MunifTanjim:
> Ok, thanks for the hint! Then, the fix is easy. Just ignore the comment parser.

When `comment` parser is uninstalled, just having `(comment) @auto` is working for both `c` and `lua`. No need for `@aligned_indent` stuffs.

theHamsta:
@MunifTanjim but how can `@auto` work for C? I understood that auto keeps the indentation while you would need to indent 1 space for C and maybe as many spaces as you have leading `---` for Lua (in case you want to align with the `[[`

theHamsta:
@MunifTanjim but it seems to work to work for the tests. I just `(comment) @auto` for C/Lua comments.

lua/nvim-treesitter/indent.lua
```diff
@@ -12,21 +25,17 @@ local function get_last_node_at_line(root, lnum)
...
-local function get_matching_prev_sibling(anchor, start, matcher)
-  local start_row, start_col = start[1], start[2]
-  local node = anchor:descendant_for_range(start_row, start_col, start_row, start_col)
-  local pos = 1
-  -- TODO: reconsider this 999 limit or do something differently in future.
-  --       if anchor has more than 999 children, this would not work.
-  while pos < 999 and node and not matcher(node) do
-    node = node:prev_sibling()
-    pos = pos + 1
+local function find_delimiter(bufnr, node, delimiter)
```
theHamsta, line 28: this is actually a flaw that we don't use the whole pattern but only the capture. Because of this we need to find the child manually here which is not nice.

MunifTanjim:
> @MunifTanjim but how can @auto work for C?

`@auto` returns `-1` meaning it'll be indented using the `'autoindent'` buffer option, which just uses the same indent from the previous line.

> I understood that auto keeps the indentation while you would need to indent 1 space for C and maybe as many spaces as you have leading --- for Lua (in case you want to align with the [[

It's not alignling with `[[`. If you have this:

```lua
--[[
  a
    b
   c
]]
```

entering a newline from L2 will put the cursor below `a`, from L3 below `b`, from L4 below `c`.

MunifTanjim (approved):

theHamsta:
@kyazdani42 Is this good to go?

kyazdani42 (approved):
lgtm, lets merge this

PR #2334, opened 2022-01-22
ci: fix cache path for rust cargo

MunifTanjim:
Decreases the time for Test Dependencies to half. 

**Before**:

image

**After**:

image

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,15 +40,18 @@ jobs:
         uses: actions/cache@v2
         with:
           path: |
-            ./.cargo/
-          key: v1.0.4
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+          key: ${{ matrix.os }}-rust-cargo-v1
 
       - name: Test Dependencies
         run: |
           mkdir -p ~/.local/share/nvim/site/pack/plenary.nvim/start
           cd ~/.local/share/nvim/site/pack/plenary.nvim/start
           git clone https://github.com/nvim-lua/plenary.nvim
-          ~/.cargo/bin/cargo +nightly install --git https://github.com/theHamsta/highlight-assertions --locked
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -40,8 +40,11 @@ jobs:
         uses: actions/cache@v2
         with:
           path: |
-            ./.cargo/
-          key: v1.0.4
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+          key: ${{ matrix.os }}-rust-cargo-v1
```
theHamsta, line 47: is this really necessary? We have only one OS and `~/.cargo/bin` is sufficient to see whether it's already installed.

theHamsta: apparently, it is. I wonder where it stores which packages it installed. It can't be under `~/.rustup`

theHamsta: probably, I should just compile the tool once for ubuntu 20.04

MunifTanjim: > is this really necessary? We have only one OS 

Not really necessary here, but it's a general practice for ci caching. Don't have anything to lose by having it, but I can remove it if you want.

> ~/.cargo/bin is sufficient to see whether it's already installed.

Downloading all the dependencies of that package from crates.io takes half the time. That's the time it saves by caching those other paths.

Since the `theHamsta/highlight-assertions` doesn't have any tagged version, need to compile everytime to ensure if you update that repo, the latest code is used.

MunifTanjim: By the way, those cache paths are taken directly from here: https://github.com/actions/cache/blob/main/examples.md#rust---cargo

theHamsta: Normally it doesn't. The problem is that it sees the package as `unknown`. Normally, it wouldn't install a new version if any old version is installed which was also the intention (except with `--force`). You can also try to install a fixed tag of it but I think this won't change anything.

I just fixed tree-sitter to a tag instead of the git version that I supposed that we would need for ABI testing.

theHamsta (approved):

PR #2337, opened 2022-01-23
fix: generate viml parser from grammar

clason:
temporarily force viml parser to be generated from grammar, unblocking lockfile update PRs and CI

can be removed when general ABI compat infrastructure is in place

@vigoux

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -782,6 +782,7 @@ list.vim = {
   install_info = {
     url = "https://github.com/vigoux/tree-sitter-viml",
     files = { "src/parser.c", "src/scanner.c" },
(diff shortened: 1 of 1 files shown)


clason:
On second thought, that is probably not a good idea -- we don't want to depend on tree-sitter being installed for such an important parser. (If the Lua switch taught me anything, it's that many more people just blindly install -- possibly through another plugin -- nvim-treesitter with all maintained parsers without really thinking about it.)

I think the right fix here is to downgrade the generated parser in https://github.com/vigoux/tree-sitter-viml to ABI 13 until there is a stable Neovim release supporting ABI 14. In fact, that should be documented guidance: Parsers meant primarily for use in Neovim should target the ABI (explicitly stated) in the last stable release.

Issue #2342, opened 2022-01-23
json5


theHamsta:
JSON5 is not natively recognized by Neovim.

Support will be added by https://github.com/neovim/neovim/pull/17180

clason:
aaaand it's gone.

Issue #2259, opened 2022-01-15
Tracking issue: removing ftdetect plugins


theHamsta:
Ideally, this plugin would concentrate on its core competence of providing parser meta data and queries. ftdetects should be handled by other plugins, user configuration or in the ideal case in upstream Neovim/Vim https://github.com/nvim-treesitter/nvim-treesitter/pull/2234 , following https://github.com/nvim-treesitter/nvim-treesitter/pull/2234#issuecomment-1010319661

Right now we have the following ftdetect plugins
```
cooklang.vim
fusion.vim
gdresource.vi
gdscript.vim
glimmer.vim
glsl.vim
gowork.vim
graphql.vim
hack.vim
hcl.vim
heex.vim
hjson.vim
json5.vim
ledger.vim
nix.vim
prisma.vim
pug.vim
ql.vim
query.vim
surface.vim
teal.vim
tlaplus.vim
yang.vim
```

After #2234 being merged, we would be left with a few ftdetect plugins for small languages. All other ftdect plugins are already in upstream Neovim. This issue tracks the progress whether those languages can be removed because they've been integrated into upstream Neovim/Vim ftdetect

- [ ] rescript (https://github.com/nvim-treesitter/nvim-treesitter/pull/2170)
- [ ] re2c (https://github.com/nvim-treesitter/nvim-treesitter/pull/2159)
- [ ] .Brewfile (ruby) (#2179)
- [x] rasi (https://github.com/nvim-treesitter/nvim-treesitter/pull/2268)
- [ ] hack
- [x] fish
- [x] fusion
- [ ] gdresource
- [ ] gdscript
- [ ] glimmer
- [ ] glsl
- [ ] gowork
- [ ] graphql
- [ ] hcl
- [ ] heex
- [ ] hjson
- [x] #2342 (added to Neovim https://github.com/neovim/neovim/pull/17180 for 0.7.0)
- [ ] ledger
- [ ] nix
- [ ] prisma
- [ ] pug
- [ ] ql
- [x] query (https://github.com/neovim/neovim/pull/17905)
- [ ] surface
- [ ] teal
- [ ] tlaplus
- [ ] yang
- [ ] hocon
- [x] eex (https://github.com/nvim-treesitter/nvim-treesitter/pull/2351)

Contributors can easily create PRs like https://github.com/vim/vim/pull/9501 to make Vim (and then Neovim, which pulls in these updates regularly) support those file types natively.

[...]

PR #2343, opened 2022-01-23
chore: remove most ftdetect plugins

theHamsta:
Native support for Neovim will be added in https://github.com/neovim/neovim/pull/17180
Most other filetypes are added in https://github.com/neovim/neovim/pull/17182

Should we remove the ftplugin now or just collect them in this PR and merge on release of the Neovim stable?

Fixes https://github.com/nvim-treesitter/nvim-treesitter/issues/2342
Fixes https://github.com/nvim-treesitter/nvim-treesitter/issues/2259

### DIFF
--- a/ftdetect/fusion.vim
+++ b/ftdetect/fusion.vim
@@ -1,1 +1,0 @@
-autocmd BufRead,BufNewFile *.fusion setfiletype fusion
--- a/ftdetect/gdresource.vim
+++ b/ftdetect/gdresource.vim
@@ -1,2 +1,0 @@
-autocmd BufRead,BufNewFile *.tscn setlocal ft=gdresource
-autocmd BufRead,BufNewFile *.tres setlocal ft=gdresource
--- a/ftdetect/gdscript.vim
+++ b/ftdetect/gdscript.vim
@@ -1,1 +1,0 @@
-autocmd BufNewFile,BufRead *.gd set ft=gdscript
--- a/ftdetect/glimmer.vim
+++ b/ftdetect/glimmer.vim
@@ -1,1 +1,0 @@
-autocmd BufNewFile,BufRead *.hbs set ft=handlebars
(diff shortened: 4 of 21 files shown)


dundargoc:
You should collect them and merge when I'm done porting them to vim and neovim, which is soon.

stsewd (approved):

dundargoc:
All ftdetects are included in this pr: https://github.com/neovim/neovim/pull/17182 except for

- `query.vim` which has vimscript which I can't read.
- parts of `hcl.vim`. There's already another filetype ending with  ".ts" so Bram suggested one need to parse the file to determine what file it is. That requires vimscript which I can't read.

If anyone wanna take care of those two you're more than welcome to.

theHamsta:
Still need to decide whether to merge that immediately or when Neovim 0.7.0 gets released.

clason:
I think `query.vim` is just `.*/queries/.*%.scm` (i.e., any scheme file under `*/queries`, right?

I don't think this is relevant to Vim, so I'd just put this into our `filetype.lua`.

The `terraform` check needs someone familiar with that filetype (and the TF mud client -- good luck with that...)

Arguably, we could just make the change in filetype.lua and live with the divergence.

I'd open a separate PR for that and tag gpanders.

dundargoc:
Aha, so the `query.vim` is unique to nvim-treesitter then? Maybe we should just leave it be in `ftdetect` since it's a filetype unique to nvim-treesitter and not a "real" filetype?

> The `terraform` check needs someone familiar with that filetype (and the TF mud client -- good luck with that...)
> Arguably, we could just make the change in filetype.lua and live with the divergence.

That is a question you and @gpanders is best qualified to answer. I'd be willing to spend a little bit of time researching TF mud client files and see if I can find an easy distinguisher if it helps you keep in sync with vim's filetype detection.

> I'd open a separate PR for those two and tag gpanders.

Agree.

theHamsta:
We need to add the ftdetct the test config to let tests pass or wait for the new stable release

clason:
I think this is done now.

PR #2344, opened 2022-01-23
update TLA+ revision, add queries for PlusCal

susliko:
[PlusCal](https://learntla.com/pluscal/) is a language, written in block comments and translated into TLA+.
Recently, PlusCal support [landed](https://github.com/tlaplus-community/tree-sitter-tlaplus/releases/tag/v0.4.0) into the `tlaplus` grammar

This PR updates the `tlaplus` grammar revision and adds queries to `locals.scm` and `highlights.scm`

@ahelwer FYI

### DIFF
--- a/queries/tlaplus/highlights.scm
+++ b/queries/tlaplus/highlights.scm
@@ -225,8 +225,6 @@
 ; Reference highlighting with the same color as declarations.
 ; `constant`, `operator`, and others are custom captures defined in locals.scm
 ((identifier_ref) @constant.builtin (#is? @constant.builtin constant))
-((identifier_ref) @operator (#is? @operator operator))
+((identifier_ref) @operator (#is? @operator function))
 ((identifier_ref) @parameter (#is? @parameter parameter))
-((identifier_ref) @variable.builtin (#is? @variable.builtin builtin_variable))
-((identifier_ref) @variable.builtin (#is? @variable.builtin variable))
-((identifier_ref) @include (#is? @include sequences))
+((identifier_ref) @variable.builtin (#is? @variable.builtin var))
--- a/queries/tlaplus/highlights.scm
+++ b/queries/tlaplus/highlights.scm
@@ -205,7 +205,7 @@
 
 ; Proofs
 (proof_step_id "<" @punctuation.bracket)
-(proof_step_id (leve) @number)
+(proof_step_id (level) @number)
 (proof_step_id (name) @constant)
 (proof_step_id ">" @punctuation.bracket)
 (proof_step_ref "<" @punctuation.bracket)
--- a/queries/tlaplus/highlights.scm
+++ b/queries/tlaplus/highlights.scm
@@ -1,4 +1,7 @@
-; highlights.scm
+; ; highlights.scm
+; ; Default capture names for tree-sitter highlight found here:
+; ; https://github.com/nvim-treesitter/nvim-treesitter/blob/e473630fe0872cb0ed97cd7085e724aa58bc1c84/lua/nvim-treesitter/highlight.lua#L14-L104
+
 
 ; Keywords
 [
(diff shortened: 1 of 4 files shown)


ahelwer:
Probably want to change the `injections.scm` file too so it doesn't hijack block comments

clason:
The parser fails to build on CI -- do you need to mark it as `requires_generate_ from_grammar`?

ahelwer:
@clason the generated code is present in the grammar's repo so that shouldn't be necessary. I see it's failing on:
> Failed to execute the following command:
{
  cmd = "tar",
  err = "Error during tarball extraction.",
  info = "Extracting...",
  opts = {
    args = { "-xvzf", "tree-sitter-tlaplus.tar.gz", "-C", "tree-sitter-tlaplus-tmp" },
    cwd = "/Users/runner/.local/share/nvim"
  }
}

which seems strange.

clason:
> Also you can add yourself as a maintainer of the TLA+ grammar's integration with neovim in the top-level README.md if you want.

That is autogenerated from the contents of `parser.lua`, so the change should be made there.

lockfile.json
```diff
@@ -240,7 +240,7 @@
     "revision": "fcc5f6f4d194dede4e676834ff28a506e39e17b4"
   },
   "tlaplus": {
-    "revision": "26bbaf5d07356cacbf21defa9316988bc732c8b3"
+    "revision": "6b82b26a742964bd93d022db134a4ac3a1a15a96"
```
clason, line 243: I think this should be

```suggestion
    "revision": "ffa98bbbd80e425f07e92bf511e64c8b0dffc1c7"
```

clason: Indeed it should.

clason:
Looks like the `locals.scm` queries need to be adapted to the new parser version.

susliko:
> Looks like the `locals.scm` queries need to be adapted to the new parser version.

Could you please clarify, what adaptations should be made?

clason:
If you click on the failed workflow result, it will tell you which query failed. As I don't know anything about the parser, I cannot tell you the exact change required.

ahelwer:
Ah, we check our query files in the CI but since `locals.scm` is new it was not added to the checks. I will fix that in our repo.

queries/tlaplus/highlights.scm
```diff
@@ -57,47 +60,125 @@
...
+  "assert"
+  "await"
+  "begin"
+  "call"
+  "define"
+  "end"
+  "fair"
+  "goto"
+  "macro"
+  "or"
```
theHamsta, line 90: maybe `@keyword.operator`?

queries/tlaplus/highlights.scm
```diff
@@ -160,3 +212,21 @@
...
+(_ label: (identifier) @tag)
+(pcal_goto statement: (identifier) @tag)
+
+; Reference highlighting with the same color as declarations.
+; `constant`, `operator`, and others are custom captures defined in locals.scm
+((identifier_ref) @constant.builtin (#is? @constant.builtin constant))
+((identifier_ref) @operator (#is? @operator operator))
+((identifier_ref) @parameter (#is? @parameter parameter))
+((identifier_ref) @variable.builtin (#is? @variable.builtin builtin_variable))
+((identifier_ref) @variable.builtin (#is? @variable.builtin variable))
```
theHamsta, line 231: Please don't use `#is?` for now. It will work but our current implementation is to slow. You can out-comment them with a TODO. If TLA files tend to be rather small the performance issue might not be relevant. In that case, you can leave them here. It would be good to test edits on a few files to make sure that there are no performance problems.

susliko: I am aware of the `#is?` usage performance concerns. In practice, TLA files tend to be no longer than 1k lines. I've tested this while editing all the examples from [this](https://github.com/tlaplus/Examples/tree/master/specifications) repo and haven't noticed any lags

queries/tlaplus/highlights.scm
susliko: If you mean "or", it's not a logical operator but a part of the PlusCal `either` clause, e.g.:
```tlaplus
(* --algorithm Test
begin
  either print "a"
  or print "b"
  or print "c"
end either ;
end algorithm *)

```

clason:
Query checks are still failing:
```
(x) Invalid capture @definition.constant in locals for tlaplus.
(x) Invalid capture @definition.operator in locals for tlaplus.
(x) Invalid capture @definition.variable in locals for tlaplus.
```

susliko:
I wonder why is the set of allowed `locals` definitions limited to [this](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#locals).

For example, one might have some constant declaration statement. Highlighting the constants' references the same as their definitions is a nice thing to have. So, why there is no capture `@definition.constant`? (same applied to `@definition.operator`)

susliko:
@clason  Can we expand the list of allowed `locals` captures to include `@definition.constant` and `@definition.operator`?

clason:
I don't know, sorry. This doesn't look like an explicit list (at least I couldn't find anything) but something more hard-coded in tree-sitter.

You _can_ create subtypes, though, like `@definition.parameter.constant` (or whatever makes most sense)?

susliko:
Tree-sitter itself allows arbitrary local captures of format `@definition.X` and it works with nvim-treesitter as well.
The failing CI check gets the allowed captures while [parsing](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/scripts/check-queries.lua#L3) `CONTRIBUTING.md`. If we add missing captures there, the check should pass

clason:
Oh, I did not realize -- that is not ideal; we should not restrict working values needlessly.

@theHamsta what was the rationale for supporting only this closed list?

clason:
Maybe it would be better to decouple the documentation and the CI script: keep an explicit list of _top-level_ captures in the script (so `@definition`, `@text` etc.) and allow anything that starts with that (so `@definition.operator`, `@text.math` etc.) to match? That would make it _much_ more maintainable.

clason:
@susliko Go ahead and add the offending captures to `CONTRIBUTING.md` for now; you also want to rebase on master.

theHamsta:
@clason we do not really support sub-captures at the moment. They simply won't work for definition/references. The current local system is kind of blocked until we refactor `locals.lua` and `query.lua. @susliko If `@definition.constant`. I don't get in what context  `@definition.operator` would make sense why not use `@definition.function`? 

The system is closed because most of the added captures were typos or slightly different wording was used in the different languages. It was impossible to get unified semantic between the languages. Sub-captures as documented in upstream tree-sitter would make sense, but this behavior is currently not implement. https://github.com/neovim/neovim/pull/14196 would implement that for highlights, but other query types must also be understand that `@foo.bar` is a `@foo`.

theHamsta:
But you used operator also for `function_definition`. You can use operator when it feels appropriate for TLA+, but in general it's recommended to use the terms what other languages would understand as function/operator so that plugins feel consistent between languages. Since no other language has `definition.operator` yet it would be possible to use them. If you can find a term that is more specific to only this concept in TLA that would be better.

queries/tlaplus/highlights.scm
```diff
@@ -160,3 +212,20 @@
...
+(_ label: (identifier) @tag)
+(pcal_goto statement: (identifier) @tag)
+
+; Reference highlighting with the same color as declarations.
+; `constant`, `operator`, and others are custom captures defined in locals.scm
+((identifier_ref) @constant.builtin (#is? @constant.builtin constant))
+((identifier_ref) @operator (#is? @operator function))
+((identifier_ref) @parameter (#is? @parameter parameter))
+((identifier_ref) @variable.builtin (#is? @variable.builtin var))
+((identifier_ref) @include (#is? @include sequences))
```
theHamsta, line 231: I've looked at our current implementation, you can use subcaptures, but the naming of the path must be exactly the same at in `locals`

theHamsta (commented):
Sorry, for the inconsistencies in our locals implementation. We eventually want to make it comparable with what Atom has. We'll then also revisit the queries in TLA+ and try to use the same as in the upstream repo.

queries/tlaplus/highlights.scm
susliko: Ah, the `sequences` query is an extra, thanks for noticing.
Others (function, parameter, var, constant) are present in `locals`

susliko:
> We eventually want to make it comparable with what Atom has. We'll then also revisit the queries in TLA+ and try to use the same as in the upstream repo.

Thank you for all the efforts you're putting into this project 🙇🏻

queries/tlaplus/locals.scm
```diff
@@ -0,0 +1,39 @@
...
+(operator_definition name: (identifier) @definition.function)
+(operator_definition parameter: (identifier) @definition.parameter)
+(pcal_macro_decl parameter: (identifier) @definition.parameter)
+(pcal_proc_var_decl (identifier) @definition.parameter)
+(pcal_var_decl (identifier) @definition.var)
+(pcal_with (identifier) @definition.parameter)
+(quantifier_bound (identifier) @definition.parameter)
+(quantifier_bound (tuple_of_identifiers (identifier) @definition.parameter))
+(variable_declaration (identifier) @definition.var)
+
```
theHamsta, line 28: I think the locals are missing `@reference`

theHamsta (approved):

queries/tlaplus/locals.scm
susliko: Added `@reference`. For some reason, scope aware highlighting worked without it

PR #2347, opened 2022-01-23
feat(indent): Implement basic indent for gdscript.

rcorre:
Indent is not handled correctly when adding new lines.
It seems that functions/loops/etc. are not recognized until they have at
least one indented block.

For example, if you enter a newline after `func foo():`, the cursor will
not be indented. If you manually indent and add a line like `pass`,
e.g.:

```
func foo():
    pass
```

now any insertions above or below `pass` will be indented correctly.
This might be an issue with the grammar, as it seems to apply to highlights
as well.

The following will not be highligted

```
func foo():
```

However, the following will be:

```
func foo():
    pass
```

### DIFF
--- a/queries/gdscript/indents.scm
+++ b/queries/gdscript/indents.scm
@@ -1,0 +1,24 @@
+[
+  (if_statement)
+
+  (for_statement)
+  (while_statement)
+
+  (parenthesized_expression)
+
+  (function_definition)
+  (class_definition)
+] @indent
+
+((argument_list) @aligned_indent
+ (#set! "delimiter" "()"))
+((parameters) @aligned_indent
+ (#set! "delimiter" "()"))
+
+[
+  ")"
+  "]"
+  "}"
+  (elif_clause)
+  (else_clause)
+] @branch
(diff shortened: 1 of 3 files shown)


theHamsta (approved):

rcorre:
Thanks for the merge! Would you have any hints on how I might fix the issues mentioned in the PR body? Does it sound like an issue with the nvim-treesitter queries or the gdscript-treesitter grammar?

clason:
I am not very familiar with this, but I suspect that this is an issue with the nvim-treesitter indent module itself; it's still quite elementary and has received a lot less love than the highlighting module. @MunifTanjim has added a number of improvement recently, but a lot of work is left to do.

MunifTanjim:
The following will not be highligted

    func foo():

However, the following will be:

    func foo():
        pass

@rcorre it's a grammer issue. Without the `pass` line, it is parsed as

image

Although, you can add this in the highlight query to make it a bit nicer:

```scm
(ERROR
  (name) @function (parameters
    (identifier) @parameter)*)
```

Issue #1352, opened 2021-06-01
Embedded templates: ERB, EEX


dbernheisel:
**Language informations**

I'm looking to add support for EEX (embedded Elixir) and ERB (embedded Ruby) templates. The file extensions usually are `html.eex` and `html.erb` to indicate the base language is html; it could be any language however such as `.ex.eex` (Embedded Elixir in Elixir) or `.js.erb` (Embedded Ruby in JavaScript).

I see a parser here: https://github.com/tree-sitter/tree-sitter-embedded-template that's hardcoded for handling the HTML base language, but I don't see that it's included in nvim-treesitter.

I'll be happy to contribute PRs for this if someone could point me in the right direction.

vigoux:
Great, thanks for reporting, we'll try to work on that.

So there's some steps we'll need to take from here, and I have some questions:

1. Are you sure the parser you mentioned supports both your languages?
2. Are those filetypes already recognized by nvim, without tree-sitter?
3. Do you want us to do it, or do you want to do it with our help?

dbernheisel:
1. [tree-sitter-embedded-template](https://github.com/tree-sitter/tree-sitter-embedded-template) supports html.erb, but not html.eex. It doesn't seem like it would be a big lift to add html.eex to it since it's almost identical to erb. I could add injections to their repo to support eex, but unsure how that would end up working here for nvim-treesitter. That said, I'm not positive this is the right tool to use (it looks like it is). But unsure how to inform nvim-treesitter which injection to use and when.

2. with a minimal config (`nvim -u NORC` `NVIM v0.5.0-dev+1358-g192ea01ed`), it seems that erb is detected as filetype `eruby`, and `eelixir` is not detected. I currently use this autocmd to set filetypes: `autocmd BufNewFile,BufRead *.eex,*.leex,*.sface,*.drab setf eelixir`.

3. I'll be happy to contribute myself but need help with the bigger picture between neovim and tree-sitter for injections, and since I don't own tree-sitter-embedded-template I'm not sure if there's a plan to use it for nvim-treesitter.

---

I have a branch here that gets it working, but a problem I encounter is how to apply the injections per filetype. I thought of two solutions

1. In the list of parsers, have the same embedded-template repo be used and name them differently, html_erb and html_eex and html_ejs. However, there is an assumption on the .so file that is generated to match the name of the parser (ie embedded_template).
[...]

PR #2268, opened 2022-01-16
Add rasi ftdetect

Fymyte:
I've added an `ftdetect` file for rasi config

Signed-off-by: Fymyte <pierguill@gmail.com>

PR #2351, opened 2022-01-24
Add tree-sitter-eex parser for (L)EEx files

connorlay:
Related to https://github.com/nvim-treesitter/nvim-treesitter/issues/1352

This PR adds support for EEx and LEEx files with [tree-sitter-eex](https://github.com/connorlay/tree-sitter-eex). Unlike HEEx, EEx does not make any assumptions about what language it is injecting Elixir into. The addition of this parser to nvim-treesitter should improve the experience of using Neovim and Tree Sitter on older Phoenix applications that have not migrated to HEEx yet, as well as other usages of EEx that do not target HTML.

### DIFF
--- a/ftdetect/eex.vim
+++ b/ftdetect/eex.vim
@@ -1,2 +1,0 @@
-au BufRead,BufNewFile *.eex set filetype=eex
-au BufRead,BufNewFile *.leex set filetype=eex
--- a/ftdetect/eex.vim
+++ b/ftdetect/eex.vim
@@ -1,0 +1,2 @@
+au BufRead,BufNewFile *.eex set filetype=eex
+au BufRead,BufNewFile *.leex set filetype=eex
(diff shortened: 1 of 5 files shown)


ftdetect/eex.vim
```diff
@@ -0,0 +1,2 @@
+au BufRead,BufNewFile *.eex set filetype=eex
```
clason, line 1: We're trying to get rid of `ftdetect` rules in this plugin... #2259

connorlay: Ah ok, according to this comment I need to contribute upstream to Vim, then have that come to Neovim via a patch?

https://github.com/nvim-treesitter/nvim-treesitter/pull/2268#issuecomment-1013878699

clason: That would be ideal, yes. (Such a trivial change usually gets merged quickly.)

clason: In any case, if you drop that file from the PR, we could merge it now. (You can tell your users to put the autocommand in their `init.vim` in the meantime.)

connorlay: @clason I removed the `ftdetect` rules as requested. Is there anything else you'd like to see in this PR?

clason (approved):

PR #2354, opened 2022-01-24
doc: stress need to :TSUpdate after lockfile change

clason:
This seems to have been a problem for many people recently.


clason:
Looking at the readme, I think it's time to switch the vimscript snippets to Lua...

vigoux (approved):

PR #2356, opened 2022-01-24
refactor(installer): more modular makefile

kylo252:
- support both scanner.cc and scanner.c
- allow complete override for (`$CFLAGS`,`$CXXFLAGS`,`$LDFLAGS`)
- add `clean` target
- add `install` target
- add windows support

Tested on:
- Arch Linux
	```console
	$ make --version
	  GNU Make 4.3
	  Built for x86_64-pc-linux-gnu
	  Copyright (C) 1988-2020 Free Software Foundation, Inc.
	  License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
	  This is free software: you are free to change and redistribute it.
	  There is NO WARRANTY, to the extent permitted by law.
	```
- MSYS2/Windows11
	```console
	  $ make --version
	  GNU Make 4.3
	  Built for x86_64-pc-msys
	  Copyright (C) 1988-2020 Free Software Foundation, Inc.
	  License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
	  This is free software: you are free to change and redistribute it.
	  There is NO WARRANTY, to the extent permitted by law.
	```

### DIFF
--- a/scripts/compile_parsers.makefile
+++ b/scripts/compile_parsers.makefile
@@ -9,15 +9,16 @@ SRC_DIR      ?= ./src
 DEST_DIR     ?= ./dest
 
 ifeq ($(OS),Windows_NT)
-   MKDIR  ?= mkdir
-   RM     ?= cmd /C rmdir /Q /S
-   CP     ?= copy
-   TARGET ?= parser.dll
+   SHELL       := powershell.exe
+   .SHELLFLAGS := -NoProfile
+   RM          := Remove-Item -Force
+   CP          := Copy-Item -Recurse
+   MKDIR       := New-Item -ItemType directory
+   TARGET      := parser.dll
 else
-   MKDIR  ?= mkdir -p
-   RM     ?= rm -rf
-   CP     ?= cp
-   TARGET ?= parser.so
+   RM          := rm -rf
+   CP          := cp
+   TARGET      := parser.so
 endif
 
 ifneq ($(wildcard src/*.cc),)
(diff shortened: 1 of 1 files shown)


theHamsta:
I don't know whether it's worth to invest too much time in make. It was only added to allow the installation of `norg` on Mac and on old Linux systems where the system compiler does not use C++14 by default. If you just want to improve the current make file, that's ok but I wouldn't try to refactor the whole process to only use Make. I'd imagine it's difficult to find out whether to use nmake or GNU make on Windows and determine whether it's installed.

kylo252:
@theHamsta, my idea is to provide a robust makefile that can then be used (in later PRs) as a standalone installation method.

The main purpose of this PR is allowing more flexibility and the ability to (mostly) handle cross-platform operations.

scripts/compile_parsers.makefile
```diff
@@ -1,18 +1,51 @@
...
-parser.so: parser.o scanner.o
-	$(CC) parser.o scanner.o -o parser.so -shared -Os -lstdc++
+OBJECTS   := parser.o scanner.o
+
+all: $(TARGET)
+
+$(TARGET): $(OBJECTS)
+	$(CC) $(OBJECTS) -o $(TARGET) $(LDFLAGS)
+
+%.o: src/%.c
```
theHamsta, line 35: @kylo252 wouldn't it be better to go back to `parser.c` and `scanner.{c,cc}`. Those are the only filenames supported by upstream tree-sitter (the filenames are hard-coded into the Rust code of tree-sitter CLI. I don't know whether we want to glob in just any C file in that folder.

theHamsta (approved):

scripts/compile_parsers.makefile
kylo252: I'm not sure if it's possible without doing the glob/wildcard at some point in the recipe. The `scanner.cc` file needs to exist if we were to add it as an explicit dep for one of the targets.

Any ideas how to get around that?

PR #2358, opened 2022-01-24
Fix #2357: align PHP indents with what C does

### LINKED ISSUE #2357: Wrong indentation for new lines after a class method
- [x] The Latest commit https://github.com/nvim-treesitter/nvim-treesitter/commit/77d247b3003759da7dfd2dba8145a0ba1bbe6c8d
- [x] Updated Parser
- [x] Options set to 4 spaces for tabs

```php
# Example

class example
{
    public int $variable;
    // indentation works correctly here

    public function foo()
    {
        // indentation works correctly here
    }

        // new line starts 1 indentation to far
}
```
This always happens after the first method. It does not matter how many methods follow. Every time I want top add another method, it will add wrong indentation.

theHamsta:
@JSteitz can you test this please?

I might add more advanced stuff like aligned parenthesis in #2333

### DIFF
--- a/queries/php/indents.scm
+++ b/queries/php/indents.scm
@@ -10,14 +10,13 @@
 ] @indent
 
 [
-  "("
   ")"
-  "{"
   "}"
-  "["
   "]"
 ] @branch
 
 [
   (comment)
 ] @ignore
+
+(compound_statement "}" @indent_end)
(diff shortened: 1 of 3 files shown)


JSteitz:
Works!

clason (approved):

PR #2359, opened 2022-01-24
folds(c): fold at compound_statement

cbarrete:
Enables folding local blocks like so:

```c
void f() {
    // this can now be folded
    {
        // ...
        // ...
    }

    // this can now be folded too
    {
        // ...
        // ...
    }
}
```

### DIFF
--- a/queries/c/folds.scm
+++ b/queries/c/folds.scm
@@ -13,5 +13,6 @@
  (preproc_else)
  (preproc_ifdef)
  (initializer_list)
+ (compound_statement)
 ] @fold
 


clason (approved):

theHamsta:
@clason if possible try to give the language maintainers a chance to comment the changes. There is a reason why compound_statement wasn't before in the folds and the folds where instead defined at the logical units like loops and functions directly. Even though the one or other choice is not striktly better or worse you should give room for discussion.

clason:
We should automate this, then, using CODEOWNERS.

theHamsta:
We found that it would be quite difficult to give so many people access to this repo. We experimented with mergify but in most cases you only get a feeling yourself when the maintainer is responsive. Usually you can tag the language maintainer and just defer the merge one or two days after review if the change is not trivial (like add missing keyword) but more a taste pondering. The idea of maintainer is that they keep a bit an eye on whether changes on their files are consistent with the previous history. Most special cases, where something is done slightly different have to should have a comment. Best to lock desired behavior into tests.

clason:
To be fair, this one did look trivial... I was just not aware that the omission was by design.

(I was way too trigger happy in this specific case; I apologize for that. Still would be nice to automate this a bit, either by auto-tagging language maintainers, or curating the maintainer list and giving them restricted privileges, like the Neovim "padawan" team.)

theHamsta:
@clason Yeah, maybe we should have tried better to integrate the maintainers with restricted access. If you know of a way how to do that, that would be a good idea. Some areas of this repo already have code owners. We could extend that. I think there was no way to know that `compound_literal` means every curly brace in C. There, were also discussions before to add it so nothing bad happened.

PR #2363, opened 2022-01-25
refactor(makefile): better powershell handling

kylo252:
- add missing $(MKDIR) variable
- create a simple function to allow running `make clean` without an error if a file doesn't exist


kylo252:
@theHamsta, sorry for not double-checking that $(MKDIR) variable again on Linux 😅 

---

I'm really happy about this new `rmf` function, it's quite handy and gives us a template to re-use for other pwsh handlers as well!

theHamsta (approved):

PR #2367, opened 2022-01-25
refactor(ci): standalone lockfile

kylo252:
- `write_lockfile()` now  ~~uses plenary~~ is now async to create a job pool this leads to the file being generated in <1s
- add other meta-info to the table to make it useful as stand-alone (WIP)

---

This `write_lockfile()` isn't meant for end-users, instead it should be invoked from the CI or manually by devs. Since it may have additional deps that aren't required for using the plugin.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -1,5 +1,6 @@
 {
   "bash": {
+    "abi": 13,
     "filetype": "sh",
     "maintainers": [
       "@TravonteD"
@@ -8,6 +9,7 @@
     "url": "https://github.com/tree-sitter/tree-sitter-bash"
   },
   "beancount": {
+    "abi": 13,
     "filetype": "beancount",
     "maintainers": [
       "@polarmutex"
@@ -16,6 +18,7 @@
     "url": "https://github.com/polarmutex/tree-sitter-beancount"
   },
   "bibtex": {
+    "abi": 13,
     "filetype": "bib",
     "maintainers": [
       "@theHamsta by asking @clason"
@@ -24,6 +27,7 @@
     "url": "https://github.com/latex-lsp/tree-sitter-bibtex"
   },
   "c": {
+    "abi": 13,
     "filetype": "c",
     "maintainers": [
       "@vigoux"
@@ -32,6 +36,7 @@
     "url": "https://github.com/tree-sitter/tree-sitter-c"
   },
   "c_sharp": {
+    "abi": 13,
     "filetype": "cs",
     "maintainers": [
       "@Luxed"
@@ -40,6 +45,7 @@
     "url": "https://github.com/tree-sitter/tree-sitter-c-sharp"
   },
   "clojure": {
+    "abi": 13,
     "filetype": "clojure",
     "maintainers": [
       "@sogaiu"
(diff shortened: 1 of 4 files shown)


theHamsta:
Can you try to align with https://github.com/nvim-treesitter/nvim-treesitter/pull/2289 and try to add information on the ABI of that revision?

I though about the stand-alone use but I thought that for this there is no need why the meta information has to sit in a Lua file. We could just dump the parser table once to a JSON and then use that as a source of truth with static and lock information separated. But just generating the meta information also works. I don't know whether it's more comfortable to write Lua or JSON. JSON would be more hot-reload friendly in case a parser URL changes. Lua is easier overwriteable via user config. So probably your approach is better.

I planed to use our parser info to write a CMake script to auto-compile specified parsers in your application and maybe create a repo that creates artifacts for all platforms to auto-download.

kylo252:
Preface: I need to make something clear, this `write_lockfile()` isn't meant for end-users, instead it should be invoked from the CI or manually by devs. Since it may have additional deps that aren't required for using the plugin.

---

> Can you try to align with #2289 and try to add information on the ABI of that revision? The ABI can be determined either by the compiled parser or by a line `#define LANGUAGE_VERSION 13` that you can grep from the current revision's `src/parser.c`. Or would it be better to wait for @vigoux to finish and then rebase?
> 

I thought of a trick to handle this so we don't even need to wait for #2289, especially since I don't see it automating that step in particular.

> I though about the stand-alone use but I thought that for this there is no need why the meta information has to sit in a Lua file. We could just dump the parser table once to a JSON and then use that as a source of truth with static and lock information separated. 

That's actually my goal. It is to consolidate parser.lua with this file, since `parser.lua` isn't really used outside of invoking `TSInstall` as far as I can tell, so even having it in json format isn't really expensive, and it will be cached in memory afterwards.

> I planed to use our parser info to write a CMake script to auto-compile specified parsers in your application and maybe create a repo that creates artifacts for all platforms to auto-download.

That again is another goal, having the json standalone file allows complete autonomy for the installer without needing the neovim runtime altogether. Just specify the list of parsers and watch them get updated.

We can already put something together re-using the recipe in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/scripts/compile_parsers.makefile. Hell, it might be even easier to use something even more cross-platform and easier to install like `meson` and `ninja`. The point is the flexibility, you can even tell the user to take care of it themselves since it's more obvious now what they need to do.

vigoux:
I am going on a rant here against this change, and other changes that I feel that are falling into the same category (not necessarily yours).

TDLR: Before opening a PR, and thinking about adding a feature to fix an issue, just ask yourself if you understand the problem completely.

Long version:

This PR does not solve any issue whatsoever, and at the same time adds a dependency ? I don't get why I should add my aproval to this. Furthermore, if you think that turning to async a task that: genuinely takes less that a minute, is done mainly in CI, and only affect _at most 5 people in the world_ is solving any issue, you are probably taking the problem from an outright wrong angle.

The lockfile generation is our solution (and actually, one of the main solution), to be quite sure that most people use the same version of the parser everywhere. And having it generated as JSON is nice, because of it's universallity, and the fact that we can manipulate it outside neovim, and use it for different purposes.

Adding that `plenary` dependency for _only one damn job that may be running at some point in some dude's PC_ is a complete waste of time, at least yours, if not routers that will forward this useless amount of data for something that does not solve any issue.

I am _very strongly against this change_ that does not fix anything, and the fact that we are celebrating that addition of the word `async` in our codebase is that symptom that we are completely out of sync.

kylo252:
I feel like you may have missed the point of my PR, I don't care about the async functionality all that much, or at all. 

In the final version of this PR, `write_lockfile()` is not even needed at all. I actually just made it to make further testing easier for me, since I didn't want to copy the parsers by hand. 

I'll finish up the other tweak that I've been working and show you what I mean.

kylo252:
@theHamsta, @vigoux, I've just finished up one of the new CI features, which I believe will be helpful for #2289, so please take a look.
 
This new `scripts/abi_version.sh` can update these values without having to rely on neovim's runtime whatsoever for that. This means that it doesn't require to install any parsers, which is what I'm assuming is currently a requriement.  This means that there should't be any need for any complicated env setup for the CI, e.g. it doesn't matter if a parser (only) supports windows or not.

theHamsta:
superseded by #2828

PR #2371, opened 2022-01-25
highlights(go): highlight `package_identifier` as `@namespace`

theHamsta:
I guess `@namespace` didn't exist the last time I touched the go highlights.

@primalmotion I'm not a go user. Do you think this is appropriate?


primalmotion:
I'm not sure what treesitter means by namespace, but just like that, it sound better than variable

clason (approved):

PR #2372, opened 2022-01-26
perf: load parsers' list on demand

kylo252:
Add a new `get_parser_config()` to allow for better abstraction. This avoids hard-coding the structure of parsers-data, e.g. `install_info.url`, `maintainers`, etc.

Benefits:
- only load the list when needed
- easier stubbing for testing since there's no global list that needs to be modified anymore.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -337,7 +337,7 @@ M.commands = {
 -- @param lang: the language of the buffer (string)
 -- @param bufnr: the bufnr (number)
 function M.is_enabled(mod, lang, bufnr)
-  if not parsers.list[lang] or not parsers.has_parser(lang) then
+  if not parsers.has_parser(lang) then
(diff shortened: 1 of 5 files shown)


kylo252:
@theHamsta, the change makes `get_parser_config(s)` to be the sole interface to handling the list. This means more flexibility of how the list is stored, e.g it can now be a separate json file (lockfile), or a cached decoded version of that same file in lua. It can even be some SQL or any other arbitrary container.

referencing https://github.com/nvim-treesitter/nvim-treesitter/pull/2367#issuecomment-1021449542

kyazdani42 (commented):
this PR is breaking change, we might not want to remove the list as it would breaking many user custom configurations

theHamsta:
There are quite a lot PRs open regarding the installation of the parsers I try to combine them in some way since they all seem to conflict. Closing this since this is just a change of ABI without using it for something

PR #2376, opened 2022-01-26
Add more queries for CMake

uyha:


theHamsta:
@uyha any updates on this?

uyha:
I pause working on this. I encountered a problem when I started adding too many queries, I think. Neovim just stops responding to some queries when I add more unrelated queries. I was trying to investigate but ended up pausing it (jumping to another project).

PR #2377, opened 2022-01-27
feat: add lazy module & config setup for users

bb010g:
This makes it easier for me to setup nvim-treesitter with packer. Example:

```lua
require('packer').startup({function(use)
  use({'nvim-treesitter/nvim-treesitter',
    config = {function ()
      local treesitter_configs = require('nvim-treesitter.configs')
      treesitter_configs.setup_config({
        ensure_installed = 'maintained',
        sync_install = false,
      })
      treesitter_configs.setup_modules({
        highlight = {
          enable = true,
        },
      })
    end},
    run = {':TSUpdate'},
  })

  use({'nvim-treesitter/nvim-treesitter-refactor',
    config = {function ()
      require('nvim-treesitter.configs').setup_modules({
        refactor = {
          highlight_definitions = {
            enable = true,
            -- Set to false if you have an `updatetime` of ~100.
            clear_on_cursor_move = true,
          },
        },
      })
    end},
  })
end})
```

By keeping configuration of modules local to their `packer.use()` statements, disabling a module (e.g. via calling `packer.use()` with `disable = true`) naturally avoids setting it up. This can also facilitate overrides of existing setup that should only occur if a module is indeed used.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -17,8 +17,19 @@ local config = {
   update_strategy = "lockfile",
   parser_install_dir = nil,
 }
+-- List of keys respected in configuration (excluding "modules").
+M.config_keys = {
+  "sync_install",
+  "ensure_installed",
+  "auto_install",
+  "ignore_install",
+  "update_strategy",
+  "parser_install_dir",
+}
 -- List of modules that need to be setup on initialization.
 local queued_modules_defs = {}
+-- List of user setup actions that need to be performed on initialization.
+local queued_user_setups = {}
 -- Whether we've initialized the plugin yet.
 local is_initialized = false
 local builtin_modules = {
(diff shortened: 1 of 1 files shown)


bb010g:
This patch has been refactored for less code duplication and better extensibility. Feedback would be very much appreciated.

I'm not in love with the naming & documentation of `init_config` and `init_modules`, but don't know what would fit them better. `config.modules` is kind of a special case, at least in `nvim-treesitter.configs`, and maybe there's generally a better way to go about handling it.

I feel like the `allow_queuing` parameter shouldn't be necessary, but `setup` currently acts instantly and I didn't want to break that behavior. Would delaying the initialization performed by `setup` currently (auto-installation, ensuring installation, enabling modules) until `init` when not `is_initialized` be an okay breaking change to make? (Initialization would still run without deferral immediately when `is_initialized`.)

I'd also like confirmation that the ordering of actions in `init` makes sense.

doc/nvim-treesitter.txt
```diff
@@ -1,6 +1,7 @@
 *nvim-treesitter*  Treesitter configurations and abstraction layer for Neovim.
 
-Minimum version of neovim: nightly
+Minimum version of Neovim: latest stable
+Recommended version of Neovim: nightly
```
clason, line 4: This change looks unrelated to this PR (and unintended; nvim-treesitter is experimental and not meant to target stable).

bb010g: Ah. It looks like the README has been updated since this was written. (1c9a546fcdbec3cb8af0143409aa527d8e5ec7a9 recommended «Neovim latest stable version or [nightly](https://github.com/neovim/neovim#install-from-source)»; a189323454d1215c682c7ad7db3e6739d26339c4 changed that to «**Neovim 0.7.0 or later** (latest [nightly](https://github.com/neovim/neovim#install-from-source) recommended)»; ee2dcc1cbe28fa605bd9a26191b3d6ac90f0fa24 changed that to «**Neovim 0.7.0 or later** built with **tree-sitter 0.20.3+** (latest [nightly](https://github.com/neovim/neovim#install-from-source) recommended)».)

Would you prefer that this is updated to
```
Minimum version of Neovim: 0.7.0 or later
Recommended version of Neovim: nightly
Minimum version of tree-sitter: 0.20.3 or later
```
or that the patch is dropped from this set entirely? I'm not too attached to getting this in here; this was something I noticed while making other changes and figured would be a quick fixup that could be bundled along without trouble.

clason: Yes, please leave it the way it is on master.

bb010g: Patch dropped.

clason:
superseded by rewrite on `main` (which removes all module functionality)

PR #2378, opened 2022-01-27
feat: add `install_info.lock` & `install_info.ref`

bb010g:
Together, they allow for completely locking a revision to the currently installed revision or following the revision of a ref of the remote.

By unsetting `install_info.lock`, the revision will revert back to whatever the lockfile specifies.

This also allows nvim-treesitter to use packer-updated grammars more easily (call `packer.use()` with `cond = false` to never pollute your runtime files). (packer's local clones can keep calls to `git ls-remote` from blocking too much.)

I investigated calling `git ls-remote` in an async way for a bit as mentioned by the existing comment, but `iter_cmd` currently does not support operating on the output of a command, and I think I'd have to heavily restructure callers of `get_revision()`. In any case, this patch does introduce more synchronous blocking to existing configurations.

Example for a packer-managed local override to an existing parser:

```lua
require('packer').startup({function(use)
  use({'vigoux/tree-sitter-viml', cond = {false}})
  use({'nvim-treesitter/nvim-treesitter',
    config = {function ()
      local packer_package_root = require('packer').config.package_root
      local join_paths = require('packer.util').join_paths
      local function get_packer_opt_package_path(...)
        return join_paths(packer_package_root, 'packer', 'opt', ...)
      end

      local treesitter_parser_configs = require('nvim-treesitter.parsers').get_parser_configs()
      treesitter_parser_configs.vim.install_info.url = get_packer_opt_package_path('tree-sitter-viml')
      treesitter_parser_configs.vim.install_info.lock = 'remote'

      require('nvim-treesitter.configs').setup_config({
        ensure_installed = 'maintained',
        sync_install = false,
      })
    end},
    run = {':TSUpdate'},
  })
end})
```

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -66,17 +66,20 @@ local function load_lockfile()
   lockfile = vim.fn.filereadable(filename) == 1 and vim.fn.json_decode(vim.fn.readfile(filename)) or {}
 end
 
-local function get_revision(lang)
-  if #lockfile == 0 then
-    load_lockfile()
+local function git_ls_remote(remote_url)
+  -- I'm sure this can be done in an async way with iter_cmd
+  local lines = vim.fn.systemlist("git ls-remote " .. remote_url)
+  local refs = {}
+  for _, line in ipairs(lines) do
+    local line_values = vim.split(line, "\t")
+    local ref = table.remove(line_values, 2)
+    refs[ref] = line_values
   end
+  return refs
+end
 
-  local install_info = get_parser_install_info(lang)
-  if install_info.revision then
-    return install_info.revision
-  end
-
-  return (lockfile[lang] and lockfile[lang].revision)
+local function get_install_info_remote_sha(install_info)
+  return git_ls_remote(install_info.url)[install_info.ref or 'HEAD'][1]
 end
 
 local function get_installed_revision(lang)
@@ -84,6 +87,26 @@ local function get_installed_revision(lang)
   if vim.fn.filereadable(lang_file) == 1 then
     return vim.fn.readfile(lang_file)[1]
   end
+end
+
+local function get_revision(lang)
+  if #lockfile == 0 then
+    load_lockfile()
+  end
+
+  local install_info = get_parser_install_info(lang)
+  local lock = install_info.lock
+  if lock == true then
+    return get_installed_revision(lang)
+  end
+  if install_info.revision then
+    return install_info.revision
+  end
+  if lock == 'remote' then
+    return get_install_info_remote_sha(install_info)
+  end
+
+  return (lockfile[lang] and lockfile[lang].revision)
 end
 
 local function is_installed(lang)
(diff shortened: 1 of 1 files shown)


clason:
Hmm, I'm torn on this. For one thing, I don't like putting in workarounds specific to a package manager (or even a specific way of using said package manager); if this is really needed, it points to a more fundamental design flaw.

Specifically here, I'm also not sure what the goal is: we already lock to a specific parser revision, and usually for good reason since the parser and queries need to match. (Unless we messed up, in which case _we_ need to revert the parser revision update or fix the queries.) What you're trying (installing parsers independently) is just not supported.

Finally, the whole parser installation business is something we really want to get rid of, not improve -- right now, it's a necessary evil since upstream doesn't have a parser distribution plan yet, but as soon as they do, it's gone.

theHamsta:
I did not have the time to look onto this but before the time of nvim-treesitter, I would have used simply `cc -Isrc src/parser.c -o vim.so -O0 -shared && mkdir -p parser && mv vim.so parser` which would give the plugin manager the control to install and version control the parser.

kylo252:
All of this is superseded by https://github.com/nvim-treesitter/nvim-treesitter/pull/2367

The new script https://github.com/kylo252/nvim-treesitter/blob/acc2676668145a82fc64388e2e4b828a63588ce6/scripts/abi_version.sh offers two helpers that you can choose to utilize however you want `get_abi_version()` and `update_all_abi()` 

And all the info being in `lockfile.json` means that you can completely bypass `:TSInstall` and install the parser however you like.

bb010g:
> We should take care that TSUpdate only updates parsers managed by nvim-treesitter

With `ref = 'HEAD'` and `url` pointing at a local remote, `:TSUpdate` will never change the local remote's revision. `:TSUpdate` will recompile, if enabled, when a local remote's commit hash updates.

bb010g:
> Hmm, I'm torn on this. For one thing, I don't like putting in workarounds specific to a package manager (or even a specific way of using said package manager); if this is really needed, it points to a more fundamental design flaw.
> 
> Specifically here, I'm also not sure what the goal is: we already lock to a specific parser revision, and usually for good reason since the parser and queries need to match. (Unless we messed up, in which case _we_ need to revert the parser revision update or fix the queries.) What you're trying (installing parsers independently) is just not supported.
> 
> Finally, the whole parser installation business is something we really want to get rid of, not improve -- right now, it's a necessary evil since upstream doesn't have a parser distribution plan yet, but as soon as they do, it's gone.

My main reason for wanting to use this is for easier occasional hacking on parsers. nvim-treesitter downloads to a temporary directory, which is fine most of the time, but sometimes I want to have that directory persist. packer already provides that for me, and I just need a way to point nvim-treesitter at packer and then have it recompile when appropriate. Additionally, if I make a local commit to that packer repository (on a branch for example), `:TSUpdate` takes care of recompiling & switching over to that new version straight from nvim. `lock = true` was added because it was easy & I was in the area. I don't want to manage my own lockfile most of the time.

theHamsta:
#2828 will add URLs

PR #2379, opened 2022-01-27
chore: run parser update at least daily

clason:
This runs the update-parsers workflow every day (at 6:30 UTC, for no particular reason except before local morning) in addition to after each push to master. This ensures that we'll see parser updates even when nothing gets merged for a while.


theHamsta (approved):
@vigoux preferred the last time to align the parser update with our activity but for me it's fine to get notified

clason:
(I think our activity is a bit more sporadic now than in the beginning, too.)

PR #2381, opened 2022-01-27
highlights(r): update

aspeddro:
1. highlight `string_escape`
2. change `parameter` priority
3. update pipe `operator` https://github.com/r-lib/tree-sitter-r/pull/27

### DIFF
--- a/queries/r/highlights.scm
+++ b/queries/r/highlights.scm
@@ -11,9 +11,9 @@
 
 (comment) @comment
 
-(formal_parameters (identifier) @parameter)
-
 (identifier) @variable
+
+(formal_parameters (identifier) @parameter)
 
 ; Operators
 [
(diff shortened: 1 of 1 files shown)


queries/r/highlights.scm
```diff
@@ -48,10 +48,12 @@
   "&"
   ":"
   "~"
-  "|>"
 ] @operator)
 
-(special) @operator
```
theHamsta, line 54: Is there any difference between those two groups of operators?

theHamsta (approved):

queries/r/highlights.scm
aspeddro: `|>` has a different rule than `binary`, the rhs should be a call function.

`(special)` has the same rules than `binary` group. I believe it was put in another group because this kind of operator is not part of the language grammar.

PR #2382, opened 2022-01-27
indents(r): add indents support

aspeddro:


theHamsta:
Would you mind adding some tests for R? It can just be the local files that you used for testing.

theHamsta (approved):

tests/indent/r/func.R
```diff
@@ -0,0 +1,12 @@
+foo <- function(x) {
+  bar <- function(a, b, c) {
+    return(a + b + c)
+  }
+  return(
+    bar(
+      x,
+      1,
+      2
+    )
```
theHamsta, line 10: you need to add a test runner like for the other languages in `test/indent`

aspeddro:
1. I removed pipe tests. It was generating an indentation error. I couldn't solve.

```r
# Expected
1:10 |>
  sum()

# Found
1:10 |>
sum()
```

However indentation works when pressing Enter.

```r
1:10 |> |
```

```r
1:10 |>
  |
```



2. Add add aligned_indent for parameters

MunifTanjim:
> I removed pipe tests. It was generating an indentation error. I couldn't solve.

I'd recommending keeping that file and add it to `expected_failures` like this: https://github.com/nvim-treesitter/nvim-treesitter/blob/d91d94aa9f55631855efb46a2b4a459a4a1439a0/tests/indent/cpp_spec.lua#L16-L22

That way others will know about the error and can fix it in future.

@aspeddro

theHamsta (approved):

PR #2389, opened 2022-01-28
folds(python): add support for pattern matching

theHamsta:
Requires: https://github.com/tree-sitter/tree-sitter-python/pull/148


theHamsta:
This issue is no longer present in the current Python grammar. `_match_block`s are exposed as `block`

PR #2391, opened 2022-01-28
Vala language support

matbme:
Support for the [Vala](https://wiki.gnome.org/Projects/Vala) programming language.

I *think* I included everything that's necessary, but if not let me know! :)


ftdetect/vala.vim
```diff
@@ -0,0 +1 @@
+au BufRead,BufNewFile *.vala set filetype=vala
```
clason, line 1: We are trying to get rid of the ftdetects; these should go into upstream. (The rest can still be merged.)

matbme: Removed

queries/vala/highlights.scm
```diff
@@ -0,0 +1,387 @@
+; Pointers
+
+(address_of_identifier
+  "&" @symbol
+  (_)*
```
theHamsta, line 5: this has no effect

queries/vala/highlights.scm
```diff
@@ -0,0 +1,387 @@
...
+
+; Booleans
+
+(true) @boolean
+(false) @boolean
+
+; Operators
+
+(binary_expression
+  (_)
```
theHamsta, line 52: you don't have to name every child of an expression.

queries/vala/highlights.scm
```diff
@@ -0,0 +1,387 @@
...
+  ]
+)
+
+(function_call
+  identifier: [
+  	(identifier) @method
+    (camel_cased_identifier) @type
+    (generic_identifier (_) @type) 
+    (namespaced_identifier
+    	left: (_)
```
theHamsta, line 119: if you don't need `left` for you query, there's also no need to add it to this file

queries/vala/highlights.scm
```diff
@@ -0,0 +1,387 @@
...
+    "?" @symbol
+)
+
+"typeof" @keyword
+
+"is" @keyword
+
+(is_type_expression
+  (_)
+  "is" @keyword
```
theHamsta, line 146: but the previous, query hightlights `"is"` already always.

queries/vala/highlights.scm
```diff
@@ -0,0 +1,387 @@
...
+(string_template_variable) @variable
+
+(string_template_expression) @variable
+
+; Assignment and declaration
+
+(new_instance
+  [
+    (
+      "new" @keyword
```
theHamsta, line 236: I assume that `"new"` can only be used this way in the grammar. Wouldn't a simple list of all the keywords be more maintainable?

queries/vala/highlights.scm
```diff
@@ -0,0 +1,387 @@
...
+  ))
+  (block)
+)
+
+; Enum
+
+"enum" @keyword
+(enum_declaration
+    (camel_cased_identifier) @type
+    (_)*
```
theHamsta, line 284: This looks, all in all good, but I have the feeling that you're just repeating every rule of the grammar in the queries. They could be simplified.

queries/vala/highlights.scm
matbme: Difference

The intention is to highlight the address symbol differently from the and operator

matbme:
@theHamsta thanks for the feedback! I pushed a new commit that removes unnecessary entries in highlights.scm

queries/vala/highlights.scm
theHamsta: but `(_)*` is not needed for that. `(_)*` mean zero or more anything which does not add any information to the query.

theHamsta (approved):

matbme:
Also, Bram pushed a [patch](https://github.com/vim/vim/commit/97c554d5149c2aa4a43d689c59563e77277265d4) adding Vala filetype support. Now we just wait for Neovim to merge?

queries/vala/highlights.scm
```diff
@@ -0,0 +1,284 @@
...
+  "return" 
+  "yield"
+  "break"
+] @keyword.return
+
+[
+  (null)
+  (modifier)
+  "typeof"
+  "is"
```
theHamsta, line 37: maybe `"is"` and `typeof` are `@keyword.operator`

queries/vala/highlights.scm
```diff
@@ -0,0 +1,284 @@
...
+; Reserved keywords
+
+[
+  "return" 
+  "yield"
+  "break"
+] @keyword.return
+
+[
+  (null)
```
theHamsta, line 34: `(null)` is maybe `@constant.builtin`

queries/vala/highlights.scm
```diff
@@ -0,0 +1,284 @@
...
+; Enum
+
+(enum_declaration
+    (camel_cased_identifier) @type
+)
+
+; Loop
+
+(foreach_statement
+  loop_item: (identifier) @variable
```
theHamsta, line 240: you have a query at the end that highlights every `identifier` as `@variable`

queries/vala/highlights.scm
```diff
@@ -0,0 +1,284 @@
...
+    (identifier) @parameter
+    (camel_cased_identifier) @type
+    (uppercased_identifier) @constant
+  ]
+)
+
+; Variable
+
+(identifier) @variable
+(camel_cased_identifier) @variable
```
theHamsta, line 284: usually those two queries are at the beginning of the query files so that other queries can overwrite

clason:
> Also, Bram pushed a [patch](https://github.com/vim/vim/commit/97c554d5149c2aa4a43d689c59563e77277265d4) adding Vala filetype support. Now we just wait for Neovim to merge?

Yes. Won't have to wait long.

clason:
And it's merged.

queries/vala/highlights.scm
matbme: For some reason placing them at the top yields the opposite result. Elements inside rules like function_call and code_attribute get overwritten

matbme (commented):
@theHamsta changed the highlights you suggested. If everything else is ready, I believe it is ready to merge

theHamsta:
Nice work!

queries/vala/highlights.scm
```diff
@@ -0,0 +1,288 @@
...
+    "&&"
+    "||"
+    "&"
+    "|"
+    "^"
+    "~"
+    "|="
+    "&="
+    "^="
+    "??"
```
theHamsta, line 127: I think `"="` is missing

queries/vala/highlights.scm
```diff
@@ -0,0 +1,288 @@
...
+; Misc
+
+(number) @number
+
+[
+  "{"
+  "}"
+  "("
+  ")"
+] @punctuation.bracket
```
theHamsta, line 31: I think `"["` `"]"` is missing. Could you add those two details please? Just tried out your vala highlighting for the first time.

theHamsta: `void open (File[] files, string hint) {` array_identifier also does not get highlighted as a type. But that can be a follow up

matbme: My bad, must have removed them somewhere during simplification

matbme: > `void open (File[] files, string hint) {` array_identifier also does not get highlighted as a type. But that can be a follow up

image

I believe it was, my last commit seems to fix it.

queries/vala/highlights.scm
theHamsta: ":" could be added to the `punctuation.delimiter` section

theHamsta: ":" could be added to the `punctuation.delimiter` section

PR #2403, opened 2022-01-29
highlights(lua): don't highlight arguments as parameters

theHamsta:
This is not done in any other language. Before it would highlight the first `a` in `foo(a, a + b)` as parameter while the second `a` has the normal `variable` highlight.

### DIFF
--- a/queries/lua/highlights.scm
+++ b/queries/lua/highlights.scm
@@ -161,8 +161,6 @@
 
 ;; Functions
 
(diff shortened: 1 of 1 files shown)


theHamsta:
@MunifTanjim

stsewd (approved):

MunifTanjim (approved):
👍🏼

PR #2407, opened 2022-01-30
New make parser

alemuller:
I rewrote the makefile parser. This new version is at the branch `master` instead of `main`.

Captures

Variables
- `@variable` default variable
- `@constant` names following make's naming convention
- `@constant.macro`  special variables
- `@constant.builtins` builtin variables

Targets
- `@string` filename
- `@string.special` special targets
- `@type` patterns
- `@symbols` targets using naming convention (all, clean, etc)

I've an one issue.

The following queries are matching all `function_expansion`. They work as expected on tree-sitter-cli.
```
(function_expansion
  function: (name) @_fn (#eq? @_fn "info")
            (arguments) @text.note)

(function_expansion
  function: (name) @_fn (#eq? @_fn "warning")
            (arguments) @text.warning)

(function_expansion
  function: (name) @_fn (#eq? @_fn "error")
            (arguments) @text.danger)

```

The same query on injections seems to work correctly.

```
(function_expansion
  function: (name) @_fn (#eq? @_fn "shell")
            (arguments) @bash)
```

### DIFF
--- a/queries/make/highlights.scm
+++ b/queries/make/highlights.scm
@@ -1,120 +1,373 @@
-(comment) @comment
+;; Names
+;; =====
+(variable) @variable
+(filename) @string
+(library)  @string.special
+(pattern
+ . "%" @operator) @type
 
-(conditional
- (_ [
+(function_expansion
+  function: _ @function)
+
+(text) @text
+(quote) @string
+
+;; Keywords
+;; ========
+[
+  "include"
+  "sinclude"
+  "vpath"
+] @include
+
+[
+  "private"
+  "define"
+  "undefine"
+  "export"
+  "unexport"
+  "override"
+] @keyword
+
+[
   "ifeq"
-  "else"
   "ifneq"
   "ifdef"
   "ifndef"
- ] @conditional)
- "endif" @conditional)
+  "else"
+  "endif"
+] @conditional
 
-(rule (targets
-       (word) @function.builtin
-       (#any-of? @function.builtin
-        ".DEFAULT"
-        ".SUFFIXES"
-        ".DEFAULT"
-        ".DELETE_ON_ERROR"
-        ".EXPORT_ALL_VARIABLES"
-        ".IGNORE"
-        ".INTERMEDIATE"
-        ".LOW_RESOLUTION_TIME"
-        ".NOTPARALLEL"
-        ".ONESHELL"
-        ".PHONY"
-        ".POSIX"
-        ".PRECIOUS"
-        ".SECONDARY"
-        ".SECONDEXPANSION"
-        ".SILENT"
-        ".SUFFIXES")))
+;; Punctuation
+;; ===========
+[
+  "$"
+  "$$"
+] @punctuation.special
 
-(rule ["&:" ":" "::"] @operator)
+[
+  "("
+  ")"
+  "{"
+  "}"
+] @punctuation.bracket
 
-(export_directive "export" @keyword)
-(override_directive "override" @keyword)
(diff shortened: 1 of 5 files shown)


theHamsta:
You need to change the commit in lockfile.json to make CI work.

clason:
> You need to change the commit in lockfile.json to make CI work.

Done, so we can peek at CI.

clason:
CI fails because of a Lua query? @theHamsta

theHamsta:
No, errors are in make highlights *and* injections. Errors only happen in Lua code.

theHamsta:
I get these errors
image

Maybe the revision is wrong?

clason:
Maybe; I just bumped to the latest commit on the `master` branch.

clason:
https://github.com/alemuller/tree-sitter-make/blob/a4b9187417d6be349ee5fd4b6e77b4172c6827dd/grammar.js#L460-L466

queries/make/highlights.scm
```diff
@@ -1,120 +1,373 @@
...
+      "TARGET_MACH"))
+
+;; Targets
+;; =======
+
+;; Naming convention
+;; -----------------
+(targets
+  (filename) @clean @symbol
+   (#any-of? @clean
```
theHamsta, line 306: `@_clean`

clason:
> EDIT: aparrently it's argument not arguments argument: (_)

yep, the new branch just has `argument` (where the old one had `arguments`).

clason:
But that wasn't it, it seems.

alemuller:
Ops. I forgot to push. I was testing with the local version:

```lua
parser_config.make = {
  install_info = {
    url = "~git/tree-sitter-make",
  },
}
```

clason:
OK, fixed the queries (I changed `@string.quote` to just `@string`; we can revisit once we have Atom-style fallback captures).

I don't know why Windows `cl` fails; it query tests claim the `make` parser is not installed, but the parser compilation says otherwise.

theHamsta:
The last time I tried, I could confirm this issue: https://github.com/nvim-treesitter/nvim-treesitter/pull/2407#issuecomment-1025264037 . It's not a blocker for this PR but we should track it in an issue.

clason:
Yeah; that doesn't look like it's an error in the queries. There's been a few things where nvim-treesitter and tree-sitter cli don't produce the same result -- I had it happen with the LaTeX parser but could work around it.

theHamsta:
We also had the artifact upload for situations like this one. I'll quickly add it to see whether the compiler actually has produced something.

clason:
I think your force-push removed my fixes 😞

clason:
Yep, `make.so` seems to be generated.

theHamsta:
@clason no, we only generated one aritfact! And we don't know from which OS. I forgot to add OS/CC name in the artifacts

theHamsta:
No, make parser. So probably a silent compilation/linkage failure of msvc

clason:
Is that new? Up till today at least the parser built... I did rebase the PR on master while fixing the queries, maybe one of the makefile PRs introduced a regression?

theHamsta:
We updated the MSVC workflow recently

clason:
@alemuller what's the status of this? If you're still interested, could you rebase? (And update the injections; we use the upstream format now, same as Helix.)

clason:
@alemuller friendly ping

amaanq:
don't think he'll be back :disappointed: I can take over make soon

clason:
We'll switch to the tree-sitter-grammars fork then.

PR #2411, opened 2022-01-30
ci: only fix major version of vulnerable action

### LINKED ISSUE #2410: Workflows are referencing vulnerable actions
Hello, there! 

As part of the university research we are currently doing regarding the security of Github Actions, we noticed that one or many of the workflows that are part of this repository are referencing vulnerable versions of the third-party actions. As part of a disclosure process, we decided to open issues to notify GitHub Community. 

*Please note that there are could be some false positives in our methodology, thus not all of the open issues could be valid. If that is the case, please let us know, so that we can improve on our approach. You can contact me directly using an email: ikoishy [at] ncsu.edu* 

Thanks in advance
1. The workflow **check-query-files-and-compilation.yml** is referencing action **ilammy/msvc-dev-cmd** using references **v1.5.0**. However this reference is missing the commit [74a501b](https://github.com/ilammy/msvc-dev-cmd/commits/74a501b) which may contain fix to the vulnerability.

The vulnerability fix that is missing by actions' versions could be related to: 
(1) CVE fix 
(2) upgrade of vulnerable dependency 
(3) fix to secret leak and others. 
Please consider updating the reference to the action. 

*If you end up updating the reference, please let us know. We need the stats for the paper :-)*

clason:
fixes #2410

### DIFF
--- a/.github/workflows/check-query-files-and-compilation.yml
+++ b/.github/workflows/check-query-files-and-compilation.yml
@@ -50,7 +50,7 @@ jobs:
       CC: ${{ matrix.cc }}
     steps:
       - uses: actions/checkout@v2
-      - uses: ilammy/msvc-dev-cmd@v1.5.0
+      - uses: ilammy/msvc-dev-cmd@v1
       - uses: actions/setup-node@v2
 
       - name: Install and prepare Neovim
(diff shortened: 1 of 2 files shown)


dundargoc:
Consider upgrading the checkout version to `v2` in `lint.yml` in that case since `v1` isn't getting any patches anymore.

clason:
Good call; one workflow still had the old action.

theHamsta (approved):

PR #2412, opened 2022-01-30
ci: automatically run required workflows after parser update

dundargoc:


dundargoc:
Hmm, this might not work now that I think about it. Will need to revisit.

theHamsta:
@dundargoc you can experiment with that on your personal fork. The GH actions are always a bit complex. Sometimes you can only see what works when you try it.

PR #2417, opened 2022-01-31
feat: parser name detection in makefile

kylo252:
- retrieve parser's name from git-url if not already provided
- install target now uses the parser's name

### DIFF
--- a/scripts/compile_parsers.makefile
+++ b/scripts/compile_parsers.makefile
@@ -8,35 +8,44 @@ LDFLAGS      ?= -Os -shared
 SRC_DIR      ?= ./src
 DEST_DIR     ?= ./dest
 
+PARSER_URL   ?= $(shell git -C $(SRC_DIR) remote get-url origin)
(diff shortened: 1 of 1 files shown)


theHamsta:
We have no intention to use the Makefile to a greater extend as we're doing today. There's no intention to maintain this file and I even could with all the changes added here. If you want you may add a config option to the `setup` call to opt-in to use makefiles to cover your CentOS usecase.

kylo252:
This is a change that I've originally missed in the first place. It even showed me two separate problems that I had fixed in this PR. 

> Ultimate goal would be to don't have any install logic in this repo.

I've already pitched the idea upstream to include support for this in the cli tool directly. In the meanwhile, I'm willing to create an independent installer for nvim-treesitter, once #2367 gets merged.

kylo252:
> It even showed me two separate problems that I had fixed in this PR.

These fixes are now in #2420

PR #2422, opened 2022-01-31
ci: prefer io.write instead of print to have full control of output

dundargoc:
This should prevent different output formats on windows and unix.

### DIFF
--- a/scripts/check-queries.lua
+++ b/scripts/check-queries.lua
@@ -1,5 +1,11 @@
 -- Execute as `nvim --headless -c "luafile ./scripts/check-queries.lua"`
 
+-- Equivalent to print(), but this will ensure consistent output regardless of
+-- operating system.
+local function io_print(text)
+  io.write(text, "\n")
+end
+
 local function extract_captures()
   local lines = vim.fn.readfile "CONTRIBUTING.md"
   local captures = {}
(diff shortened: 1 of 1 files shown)


dundargoc:
https://github.com/nvim-treesitter/nvim-treesitter/pull/2421 needs to be merged first.

dundargoc:
`print()` outputs an extra newline on windows jobs. I suspect Github Actions interprets carriage return as a newline as well.

dundargoc:
Extra context: `print()` outputs an extra newline on windows jobs for some reason, making the output jarring to read.

scripts/check-queries.lua
```diff
@@ -35,7 +35,7 @@ local function do_check()
   local captures = extract_captures()
   local last_error
 
-  print "::group::Check parsers"
+  io.write "::group::Check parsers\n"
```
theHamsta, line 38: maybe add a local wrapper around that idiom?

theHamsta: but would be also fine with this solution.

dundargoc: > maybe add a local wrapper around that idiom?

Sorry, I'm still new to lua. What do you mean? Can you give an example/link?

theHamsta: write your own `print` that uses `io.write`. Like `local function io_print(text) io.write(text .. '\n') end`

theHamsta (approved):

theHamsta:
Thank you! I found those empty lines also very annoying for a long time but never found the motivation to search for the reason.

Issue #2313, opened 2022-01-21
CI: simplify tests


clason:
With the increasing number of parsers (and more coming), CI is starting to take a while...

We should pare down the number of tests:

1. reduce the combinations of platform, compiler, and Neovim version;
2. only run one combination (Ubuntu, gcc, nightly? that seems to be the fastest) on draft PRs (and push people to keep the PRs as draft until actually ready for review);

~3. avoid running all checks on new PRs twice (once for open, once for push)~ see #2314

We could also look into using labels for controlling workflows (like `ci-benchmark`, or `ci-full`, or `ci-none` -- the automated "Update README" PR for example does not need any tests).

vigoux:
This all seems like a great idea, do you have exemples about controlling CI based on labels somewhere ?

clason:
Not off the top of my head, but Homebrew uses that.

MunifTanjim:
Compiled parser can be cached based on `lockfile.json` hash and only recompile when `lockfile.json` changes. That would cut down significant amount of time (4-5 minutes on each `workflows/tests.yml` run).

clason:
Nice, we should do that then. Do you know how? 😄

MunifTanjim:
I'll open a PR soon. 😄

clason:
I'll merge my PR shortly, once tree-sitter releases the new version. (I can also merge it now if that helps.)

theHamsta:
@MunifTanjim the relevant files are lockfile.json and `parsers.lua`. The compilation should only happen when those files change or we only compile on one config when those files have not changed. Could also do ccache caching on Ubunut/MacOS (I planned also caching for compiling the Rust dependency for the tests)

theHamsta:
@clason it would be interesting not to fail when people have the old tree-sitter CLI when we do the ABI upgrade.

clason:
Yes, that should be possible if we do the version check (only specify ABI if tree-sitter CLI is new enough).

clason:
then approve and merge plz :]

theHamsta:
@MunifTanjim there are workflows like this which let you test whether specific files changed similar to Gitlabs if changed option https://github.com/tj-actions/changed-files
 
and ... I forgot about `install.lua`. That should also trigger a recompilatoin

clason:
Another thing we could think about is running the lockfile update PR more or less regularly (say, once daily) instead of every push to master -- which could be either too often or too seldom, depending on our activity...

[...]

PR #2425, opened 2022-01-31
ci: skip swift check on neovim stable

dundargoc:
This is meant as a temporary workaround until neovim 0.7 is released.

More context: https://github.com/nvim-treesitter/nvim-treesitter/issues/2313#issuecomment-1025258357

### DIFF
--- a/.github/workflows/check-query-files-and-compilation.yml
+++ b/.github/workflows/check-query-files-and-compilation.yml
@@ -84,7 +84,14 @@ jobs:
         run: cp -r ~/AppData/Local/nvim/pack/nvim-treesitter/start/nvim-treesitter/parser/* parser
         shell: bash
 
+       # NOTE: this is a temporary workaround to skip swift tests on ubuntu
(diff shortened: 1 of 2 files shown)


clason (approved):

clason:
Screenshot 2022-01-31 at 22 52 33

I like those numbers... :D

dundargoc:
I'll be honest, I'm not a great fan of this workaround because it's a bit ugly. But given that this is a temporary workaround and because I've given good context on what to remove and when I still think this could be worth it (especially considering the CI times are reduced by more than 5-7 times).

clason:
Yeah, it's not great, but think of all the trees this is saving!

dundargoc:
:evergreen_tree: :deciduous_tree: :christmas_tree: :tanabata_tree: :palm_tree:

theHamsta (approved):
yeah, I guess that's ok for a temporal thingy.

PR #2436, opened 2022-02-02
chore(latex): update parser

clason:
This is a significant refactor, but with the patient help of @pfoerster, I got the queries back in pretty much the same shape they were in before. Some now possible improvements are left on the table:

* the `\section` etc. commands are now part of the grammar and can (and should) be captured as such
* same for new commands, references, glossary entries etc.
* injections for `pycode`, `minted`, etc. environments (as a future PR).

(Although there might be difficulties with those since they are defined via anonymous nodes.)

### DIFF
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -4,10 +4,10 @@
  (inline_formula)
 ] @text.math
 
-;; This highlights the whole environment like vimtex does
-((environment
+((generic_environment
   (begin
-   name: (word) @_env)) @text.math
+   name: (curly_group_text
+           (text) @_env))) @text.math
    (#any-of? @_env
       "displaymath" "displaymath*"
       "equation" "equation*"
@@ -20,8 +20,27 @@
       "gather" "gather*"
       "flalign" "flalign*"))
 
+;; File inclusion commands
 [
-  (generic_command_name)
+  "\\documentclass"
+  "\\usepackage"
+  "\\input"
+  "\\include"
+  "\\subfile"
+  "\\subfileinclude"
+  "\\subfileinclude"
+  "\\includegraphics"
+  "\\addbibresource"
+  "\\bibliography"
+  "\\includesvg"
+  "\\includeinkscape"
+  "\\usepgflibrary"
+  "\\usetikzlibrary"
+] @include
+
+;; Definitions
+[
+  (command_name)
   "\\newcommand"
   "\\renewcommand"
   "\\DeclareRobustCommand"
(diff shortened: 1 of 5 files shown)


seblj:
I am really looking forward to a better parser for latex, so I wanted to try this PR to play with some injections. I got this error when I opened a latex file:
image

Note: I uninstalled the parser before installing it again, and I am on the latest commit to master on neovim

clason:
Yes, you jumped the gun. This needs quite a bit more work.

(The parser wasn't really the biggest problem, LaTeX just needs a lot more and more precise queries.)

seblj:
> Yes, you jumped the gun. This needs quite a bit more work.
> 
> (The parser wasn't really the biggest problem, LaTeX just needs a lot more and more precise queries.)

Okay! Just wanted to comment before this would eventually get merged since it was out of draft, and therefore not sure if you had the same issue

clason:
Don't worry, CI won't let us merge with such errors.

clason:
Also, don't expect any wonders; the goal of this PR is simply to update to the breaking changes and get a working set of queries with no major regressions. To get _better_ highlighting from this will require follow-up work (from anybody interested in improving highlighting).

seblj:
> Also, don't expect any wonders; the goal of this PR is simply to update to the breaking changes and get a working set of queries with no major regressions. To get _better_ highlighting from this will require follow-up work (from anybody interested in improving highlighting).

Yeah I know. I am mostly waiting for the updated parser as it fixes some bugs where it reported error in the tree, when it was valid latex. This made for example spellchecking with `spellsitter.nvim` not so enjoyable. By looking at the tree when checking out this PR, it seems to work like this is fixed in the parser. The highlighting for latex with treesitter has never really bothered me, so if the highlighting is pretty much the same as before this PR, it would be a big improvement in my eyes because of the updated parser

clason:
I think this is ready for testing now. @theHamsta I'd appreciate a review.

queries/latex/highlights.scm
```diff
@@ -413,7 +392,8 @@
       "flalign" "flalign*"))
 
 (end
- name: (_) @text.environment.name
+ name: (curly_group_text
+         (text) @text.environment.name)
   (#not-any-of? @text.environment.name
```
theHamsta, line 397: do we have this highlight group?

theHamsta (approved):
I didn't have a look on what they changed exactly. But when you just changed the highlights to equivalent nodes then everything should be fine

queries/latex/highlights.scm
clason: Looks like we do :)

clason:
> I didn't have a look on what they changed exactly. But when you just changed the highlights to equivalent nodes then everything should be fine

Mostly. There's also some refactor (like having `curly_group_text` etc. which you need to unwrap to get rid of the braces), and they make use of anonymous nodes for commands (which nvim-treesitter can't seem to match? that's an issue I'd like to discuss with you offline when you have a bit of time).

(I also did some moving around and removal of duplicate captures (references) to make it easier for me to find my way around.)

seblj:
Hmm weird, I still get the same issue as I commented about earlier. But I couldn't reproduce in a docker container with minimal config, so there is probably just something that I have messed up then :)

clason:
Yeah, make sure you don't have stale queries or parsers lying around in your rtp.

seblj:
It was this one that caused the issues: https://github.com/nvim-treesitter/nvim-treesitter-textobjects which probably has stale queries. So I just disabled it for latex for now

clason:
Thanks for the heads-up; I'll make a follow-up PR to text-objects (which I personally don't use). The (often silent) reliance of other plugins on specific parser versions is a problem we've recently seen with the Lua parser switch and that needs to be addressed at some point.

Let me know if the highlights look good, then I'll merge.

seblj:
> Thanks for the heads-up; I'll make a follow-up PR to text-objects (which I personally don't use). The (often silent) reliance of other plugins on specific parser versions is a problem we've recently seen with the Lua parser switch and that needs to be addressed at some point.
Awesome, no problem!

> Let me know if the highlights look good, then I'll merge.

It looks good for me!

PR #2441, opened 2022-02-03
feat(vim): highlight default parameters

reegnz:
Support to tree-sitter-viml was added with https://github.com/vigoux/tree-sitter-viml/pull/57

### DIFF
--- a/queries/vim/highlights.scm
+++ b/queries/vim/highlights.scm
@@ -36,6 +36,7 @@
 (function_declaration name: (_) @function)
 (call_expression function: (identifier) @function)
 (function_declaration parameters: (parameters (identifier) @parameter))
+(function_declaration parameters: (parameters (default_parameter (identifier) @parameter)))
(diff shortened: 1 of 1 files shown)


queries/vim/highlights.scm
```diff
@@ -36,6 +36,7 @@
 (function_declaration name: (_) @function)
 (call_expression function: (identifier) @function)
 (function_declaration parameters: (parameters (identifier) @parameter))
+(function_declaration parameters: (parameters (default_parameter (identifier) @parameter)))
```
theHamsta, line 39: wouldn't 

```
(default_parameter (identifier) @parameter)
```
be sufficient

reegnz: Probably would be. Let me change that.

reegnz:
I accidentally deleted the branch, so I'll open a new PR for this.

PR #2448, opened 2022-02-04
ci: make Update README PR skip CI

clason:
no need to run workflows on this one.

### DIFF
--- a/.github/workflows/update-readme.yml
+++ b/.github/workflows/update-readme.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:
-          commit-message: Update README
+          commit-message: Update README [skip ci]
           title: Update README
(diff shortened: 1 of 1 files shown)


.github/workflows/update-readme.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
         with:
-          commit-message: Update README
+          commit-message: Update README [skip ci]
```
theHamsta, line 38: did you test whether this works? We had this message there before but it was not working at that time. We can just try out

theHamsta (approved):

.github/workflows/update-readme.yml
clason: Hmm, yeah, you're right, this gets stripped on commit. Probably a security measure. Never mind then!

PR #2458, opened 2022-02-05
ci/introduce nvim variable

dundargoc:
- ci: introduce platform-independent environment variable NVIM
- ci: combine "Check query files" steps into a single step


dundargoc:
This requires setting all steps to run bash (well, specifically the windows ones but this is easier) but that's honestly a positive on its own since mixing shell syntax between steps can get confusing in my opinion.

dundargoc:
It's not possible to combine the steps "Compile parsers Unix like" and "Compile parsers Windows" since they're not identical. I tried removing the `prefer_git=false` part in windows and but couldn't recreate the problem with caching (the first run was slower which I assume is due to the caching but all subsequent runs were fast). Could I get a bit more background on why we can't use git on windows or what step I'm missing?

theHamsta:
`prefer_git=false` is already the default for Unix

theHamsta (approved):

dundargoc:
> `prefer_git=false` is already the default for Unix

Oh? Then it's possible to combine those steps as well then.

PR #2461, opened 2022-02-05
Add test for #2086 and test entering a new line when in insert mode

theHamsta:
This shows that #2086 is resolved but shows some Problems with Python were we don't re-parser on new-line ~~and one test for Lua~~.

### DIFF
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -5,7 +5,6 @@ local tsutils = require "nvim-treesitter.ts_utils"
 local M = {}
 
 M.avoid_force_reparsing = {
-  python = true,
   yaml = true,
 }
 
--- a/queries/python/indents.scm
+++ b/queries/python/indents.scm
@@ -26,7 +26,6 @@
 
   (concatenated_string)
 ] @indent
-
   
 (if_statement
   condition: (parenthesized_expression) @aligned_indent
--- a/tests/indent/common.lua
+++ b/tests/indent/common.lua
@@ -140,6 +140,17 @@ function M.indent_new_line(file, spec, opts, xfail)
   table.insert(before, spec.on_line + 1, indent .. spec.text)
 
   compare_indent(before, after, xfail)
+
+  before, after = M.run_indent_test(file, function()
+    -- move to the line and input the new one
+    vim.cmd(string.format("normal! %dG$", spec.on_line))
+    vim.cmd(string.format(vim.api.nvim_replace_termcodes("normal! a<cr>%s", true, true, true), spec.text))
+  end, opts)
+
+  indent = type(spec.indent) == "string" and spec.indent or string.rep(" ", spec.indent)
+  table.insert(before, spec.on_line + 1, indent .. spec.text)
+
+  compare_indent(before, after, xfail)
 end
 
 local Runner = {}
--- a/tests/indent/c/issue-2086.c
+++ b/tests/indent/c/issue-2086.c
@@ -1,0 +1,3 @@
+{
+    statement;
+    statement;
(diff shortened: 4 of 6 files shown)


clason (approved):

lua/nvim-treesitter/indent.lua
```diff
@@ -5,7 +5,6 @@ local tsutils = require "nvim-treesitter.ts_utils"
 local M = {}
 
 M.avoid_force_reparsing = {
-  python = true,
```
theHamsta, line 8: this will make indentation after `\` stop to work

queries/python/indents.scm
```diff
@@ -27,6 +27,8 @@
   (concatenated_string)
 ] @indent
 
+; TODO: this predicate is not executed often enough to prevent accidental matches
+;((expression_statement) @indent (#lua-match? "\\\\\\n" @indent))
```
theHamsta, line 31: this fixes the indentation after `\`, but seems to match unrelated statements which seems to be a bug in the Nvim's query implementation @MunifTanjim do you see a reason why this shouldn't work? The actual Lua query for `find` is `\\\n` so a verbatim `\` followed by a `\n`. All `\` have to be escaped in the query file to not enter escape sequences in the string in the query file

#2407 #2376  also reported that `match` accidentally match on unrelated queries.

lua/nvim-treesitter/indent.lua
theHamsta: before you would have strange indentation on starting a new line using `<cr>` in insert mode. E.g. indenting a aligned_indent though the parens are already closed.

tests/indent/common.lua
```diff
@@ -140,6 +140,17 @@ function M.indent_new_line(file, spec, opts, xfail)
   table.insert(before, spec.on_line + 1, indent .. spec.text)
 
   compare_indent(before, after, xfail)
+
+  before, after = M.run_indent_test(file, function()
+    -- move to the line and input the new one
+    vim.cmd(string.format("normal! %dG$", spec.on_line))
+    vim.cmd(string.format(vim.api.nvim_replace_termcodes("normal! a<cr>%s", true, true, true), spec.text))
```
theHamsta, line 147: also test starting a new line with `<cr>` in insert mode vs the already test `o` in normal mode

clason (approved):

queries/python/indents.scm
MunifTanjim: I don't think it's about the query. Need to check which initial node is detected by the indent module here: https://github.com/nvim-treesitter/nvim-treesitter/blob/58a4897e6d5f21a9178ae0f802d47af26d67a219/lua/nvim-treesitter/indent.lua#L103-L107

theHamsta: But adding this query that does not match in the file at all shouldn't change highlighting behavior, right?

MunifTanjim: Oh I see... this query is incorrect I think, `((expression_statement) @indent (#lua-match? "\\\\\\n" @indent))` is causing runtime error (it does so in playground). It should be: `((expression_statement) @indent (#lua-match? @indent "\\\\\\n"))`

kyazdani42 (approved):

queries/python/indents.scm
theHamsta: Oops, thanks for noting this

MunifTanjim: btw, for this

```python
a = 2 \
    + 2
```

when you enter a newline after L1


```python
a = 2 \

    + 2
```

And parse the tree, the `expression_statement` in L1 just becomes `a = 2` (the `\\n` is not part of it anymore). So that query will be useless.

theHamsta: But the capture won't fix the problem anyway...

theHamsta:
Let's merge this for the additional tests, as we have two other PRs depending on those tests. On of them (#2475), trying to fix a similar problem like the one with Python continuation lines.

PR #2462, opened 2022-02-05
Add Gleam parser and queries

connorlay:
Relates to: https://github.com/J3RN/tree-sitter-gleam/issues/1 and https://github.com/J3RN/tree-sitter-gleam/issues/12

cc @lpil @J3RN

This PR adds support for the [Gleam](https://gleam.run/) language via [tree-sitter-gleam](https://github.com/J3RN/tree-sitter-gleam). I have not measured it, but the highlighting performance is excellent on my machine. The queries are based on those in the parser's repository, adapted to use nvim-treesitter's additional capture groups.

image
An example web server taken from the Gleam homepage. Color theme is [Melange](https://github.com/savq/melange).


queries/gleam/highlights.scm
```diff
@@ -0,0 +1,135 @@
...
+(argument label: (label) @symbol ":" @symbol)
+(type_constructor_argument label: (label) @symbol ":" @symbol)
+(record_update_argument label: (label) @symbol ":" @symbol)
+(function_parameter name: (identifier) @symbol ":" @symbol)
+
+; Types
+((type_identifier) @type (#not-any-of? @type "True" "False" "Nil" "Error" "Ok"))
+
+; Built-in Types
+((type_identifier) @constant (#any-of? @constant "Nil" "Error" "Ok"))
```
theHamsta, line 117: maybe event `@constant.builtin`

queries/gleam/highlights.scm
```diff
@@ -0,0 +1,167 @@
...
+(public_external_function name: (identifier) @function)
+(external_function_body (string) @namespace . (string) @function)
+
+; Pipe Operator
+(binary_expression operator: "|>" right: (identifier) @function)
+
+; Parser Errors
+(ERROR) @error
+
+; TODO does nvim-treesitter support #is-not? local matches for highlights?
```
connorlay, line 166: @theHamsta Is it possible to define highlights that take into account the scope of a node? Consider the following:
```gleam
pub fn foo() {
  fizz.buzz()
}

pub fn foo(bar) {
  bar.baz()
}
```
`fizz` should be highlighted as `@namespace` since it is not defined anywhere in the function parameters or body. However, `bar` should be highlighted as `@variable` because it is defined within the function parameter list.

Thanks 😄

theHamsta: yes, this is possible using `(#is? @variable parameter)`. But our `#is?` implementation is super slow. It's waiting for something like https://github.com/nvim-treesitter/nvim-treesitter/pull/1120 but something that is preferable something faster that `:parent()`

connorlay: Good to know, I'll add a note about this in the highlights.scm and we can revisit it later. Thanks!

queries/gleam/indents.scm
```diff
@@ -0,0 +1,21 @@
...
+  (public_type_definition)
+  (tuple)
+  (type_definition)
+] @indent
+
+[
+  ")"
+  "]"
+  "}"
+] @indent_end @branch
```
theHamsta, line 21: Again, tests would be helpful

queries/gleam/highlights.scm
```diff
@@ -0,0 +1,158 @@
+; Keywords
+[
+  "as"
+  "const"
+  "external"
+  "let"
+  "opaque"
+  "pub"
+  "todo"
+  "try"
```
theHamsta, line 10: Is try also more in direction of `@exception` or more like "launch" or general control flow?

theHamsta (approved):

connorlay:
Let's hold off on merging this until I have some tests in place. Thanks!

theHamsta (approved):

queries/gleam/highlights.scm
lpil: It's flow control, similar to this Go code:

```go
if err != nil {
        return err
}
```

lua/nvim-treesitter/parsers.lua
```diff
@@ -326,6 +326,15 @@ list.elixir = {
   maintainers = { "@jonatanklosko" },
 }
 
+list.gleam = {
+  install_info = {
+    url = "https://github.com/J3RN/tree-sitter-gleam",
+    files = { "src/parser.c" },
+    branch = "main",
+  },
+  maintainers = { "@J3RN" },
```
theHamsta, line 335: You would be the maintainer for nvim-treesitter for this parser.

connorlay: For the maintainers section, is it intended to be the maintainer of the queries and tests in nvim-treesitter, or the maintainer of the parser itself? For example, I am the maintainer for tree-sitter-heex and the repository lives under my namespace, but I did not write tree-sitter-gleam.

lua/nvim-treesitter/parsers.lua
```diff
@@ -323,7 +323,16 @@ list.elixir = {
...
+  maintainers = { "@jonatanklosko", "@connorlay" },
+}
+
+list.gleam = {
+  install_info = {
+    url = "https://github.com/J3RN/tree-sitter-gleam",
+    files = { "src/parser.c" },
+    branch = "main",
+  },
+  maintainers = { "@J3RN", "@connorlay" },
```
theHamsta, line 335: @connorlay does J3RN also contribute to nvim-treesitter? Otherwise, he should be removed from the maintainers as I and user otherwise will unnecessarily ping him when there are nvim-treesitter issues

theHamsta: otherwise this PR is fine

connorlay: No J3RN is the maintainer of the tree-sitter-gleam parser itself, but not the nvim-treesitter integration.

connorlay:
@theHamsta I updated the maintainers list as you requested. I think this is ready to merge 😄

theHamsta:
Thank you a lot for your help!

PR #2464, opened 2022-02-05
feat(highlights/julia): Add missing highlights

savq:
- Add highlights for:
    + command strings
    + abstract type definitions
    + module `end` delimiters

- Distinguish `:` in quotes and ranges (it's already highlighted
  differently in ternary expressions).

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -9,6 +9,7 @@
   (triple_string)
   (string)
 ] @string
+(command_string) @string.special
 
 (string
   prefix: (identifier) @constant.builtin)
@@ -83,6 +84,8 @@
 (typed_expression
   (parameterized_identifier) @type .)
 
+(abstract_definition
+  name: (identifier) @type)
 (struct_definition
   name: (identifier) @type)
 
@@ -129,10 +132,11 @@
 ] @comment
 
 [
+  "abstract"
   "const"
   "macro"
-  "struct"
   "primitive"
+  "struct"
   "type"
 ] @keyword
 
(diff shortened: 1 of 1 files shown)


queries/julia/highlights.scm
```diff
@@ -168,17 +172,19 @@
 (export_statement
   ["export"] @include)
 
-[
-  "using"
-  "module"
-  "import"
-] @include
+(module_definition
+  ["module" "end"] @namespace)
```
theHamsta, line 176: namespace is usually for namespace identifiers

savq: Ok. would
```
(module_definition
  "module" @include
  name: (identifier) @namespace
  "end" @include)
```
make more sense in this case?

theHamsta: for the identifer yes, but `include` is usually only used for keywords importing a module from another file

theHamsta: > Highlighting profits from a more flexible approach

No, users will just get unexpected highlights. If more flexibility is need, usually a new capture for the new thing is better as some people get really frustrated when they can't change a "wrong" highlight (`@keyword.module` with default `@keyword` highlight which could also apply for Ruby modules). From the experience with this repo, you really need strong reasons to give a distinct highlight to keywords. `module`/`end` might be conceivable for default `include` though

savq: > include is usually only used for keywords importing a module from another file

You're right. I was comparing with the rust queries for reference, but I forgot `mod` does both (declaration or import). Although I also agree with @clason we can be more lenient with some aspects of the highlighting.

Also, on second thought I won't write the `@namespace` query; simply because in practice, module namespacing is used very little in Julia and there aren't any other `@namespace` queries that relate to it.

theHamsta: Yes, it's a similar concept. It would be no problem to use include here especially when Rust already uses `mod` in the same way. Just have a look whether it fits into general look and feel.

theHamsta:
@clason is this ready? You are our Julia expert :slightly_smiling_face:

clason (approved):

clason:
> @clason is this ready? You are our Julia expert 🙂

Yes, looks good, but you are the highlight names expert ;)

Issue #2463, opened 2022-02-05
:TSUpdate encounters compiler error when linking Scrt1.o from glibc 2.33 with ld 2.36.1


jokeyrhyme:
### Describe the bug

Howdie, thanks for sharing your work!

Recently, when running `:TSUpdate`, I get the following error message:

```
nvim-treesitter[norg]: Error during compilation
/usr/bin/ld: /usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../lib/Scrt1.o: in function `_start':
(.text+0x24): undefined reference to `main'
collect2: error: ld returned 1 exit status
make: *** [HOME/.local/share/nvim/site/pack/packer/start/nvim-treesitter/scripts/compile_parsers.makefile:38: parser.so] Error 1
Press ENTER or type command to continue
```

### To Reproduce

1. `nvim`
2. `:TSUpdate`
3. observe compiler error from nvim-treesitter plugin

### Expected behavior

No error

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.4 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v17.3.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 11.1.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 13 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - scss           ✓ . . ✓ . 
  - pascal         ✓ ✓ ✓ ✓ ✓ 
  - erlang         . . . . . 
  - javascript     ✓ ✓ ✓ ✓ ✓ 
  - elixir         ✓ ✓ ✓ ✓ ✓ 
  - bibtex         ✓ . ✓ ✓ . 
  - comment        ✓ . . . . 
  - surface        ✓ . ✓ ✓ ✓ 
  - jsdoc          ✓ . . . . 
  - c              ✓ ✓ ✓ ✓ ✓ 
  - eex            ✓ . . . ✓ 
  - sparql         ✓ ✓ ✓ ✓ ✓ 
  - cmake          ✓ . ✓ . . 
[...]

PR #2465, opened 2022-02-05
fix: -Os should be used during compilation phase in makefile

theHamsta:
- also add `CXX_STANDARD` again as it's the only variable used from the Lua side

There might be more errors in the Makefile https://github.com/nvim-treesitter/nvim-treesitter/issues/2463#issue-1125016147 let's see what the problem was there. I would prefer to return back to the basic Makefile we had in the beginning (which already had mentioned error of using the optimization flag during linkage phase)

 https://github.com/nvim-treesitter/nvim-treesitter/issues/2463#issue-1125016147  was related to a `LDFLAGS` which was set in the user's environment and not by our Lua code.

### DIFF
--- a/scripts/compile_parsers.makefile
+++ b/scripts/compile_parsers.makefile
@@ -1,56 +1,13 @@
-#
-# compile_parsers.makefile
-#
+CC?=cc
+CXX_STANDARD?=c++14
 
-CXX_STANDARD ?= c++14
-CFLAGS       ?= -Os -std=c99 -fPIC
-CXXFLAGS     ?= -Os -std=$(CXX_STANDARD) -fPIC
-LDFLAGS      ?= -shared
-SRC_DIR      ?= ./src
-DEST_DIR     ?= ./dest
+all: parser.so
 
-ifeq ($(OS),Windows_NT)
-   SHELL       := powershell.exe
-   .SHELLFLAGS := -NoProfile -command
-   CP          := Copy-Item -Recurse -ErrorAction SilentlyContinue
-   MKDIR       := New-Item -ItemType directory -ErrorAction SilentlyContinue
-   TARGET      := parser.dll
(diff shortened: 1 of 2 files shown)


kyazdani42:
why would the code size optimization fix a linker issue :o ?

theHamsta:
I only wanted to fix the `-Os` problem first, but the found that the other issue was indeed related to makefile variables set from the outside enviroment. I think it would be good to have the Makefile as simple as possible to avoid any more maintenance issues.

Best would be to have no alternative installation mechanism. But getting c++14 for norg, was really challenging any other way.

theHamsta:
`-shared` should have been in LDFLAGS since LDFLAGS are sometimes understood as LD_EXTRAFLAGS it would be possible to set all variables in the makefile in Lua but I think that the general Makefile will find application out-side of nvim-treesitter while we only want to compile norg on Apple.

theHamsta:
Fixes #2494

scripts/compile_parsers.makefile
```diff
@@ -22,25 +18,21 @@ else
    rmf         = rm -rf $(1)
 endif
 
-ifneq ($(wildcard $(SRC_DIR)/*.cc),)
+ifneq ($(wildcard src/*.cc),)
```
kyazdani42, line 21: is this change needed ?

kyazdani42: actually all the removed substitutions ?

kyazdani42 (approved):

scripts/compile_parsers.makefile
theHamsta: Acutally, src was hard-coded in the targets anyways. `$(SRC_DIR)` might just be a risk that someone actually has this defined. but I changed it now to use `$(SRC_DIR)` which might be a bad idea

PR #2470, opened 2022-02-06
ci: rename workflows and jobs to be more consistent

dundargoc:
Feel free to give other suggestions. This isn't a clear improvement but after the fifth time of mixing together the different workflows I thought it might still be worth a shot.


theHamsta (approved):
Sounds to me like an improvement!

dundargoc:
What's the status on in?

PR #2472, opened 2022-02-06
ci/apt related fixes

dundargoc:
- ci: remove unnecessary apt calls
- ci: use apt-get instead of apt to avoid warnings from apt


dundargoc:
Shaves off 3x10s per pull request/push.

theHamsta (approved):

PR #2473, opened 2022-02-06
chore: expose start and stop highlighter

### LINKED ISSUE #1944: Pause highlighting
Currently highlighting for (big or in big projects) typescript / javascript files are pretty slow. makes typing very sluggish.

either a command / function to pause queries. so it can be used to pause highlight updates on InsertEnter and continue on InsertLeave.

currently im using `au InsertEnter *ts :TSBufDisable highlight` `au InsertLeave *ts TSBufEnable`.
but this cause all TS highlighting to be undone on disable.

**Additional context**
n/a

kyazdani42:
Decompose highlighter module in small functions to allow exporting a
start and stop functions without the syntax change.
Also fix linter issues in configs.lua

Should fix #1944

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -71,8 +71,8 @@ end
 -- @param bufnr buffer number, defaults to current buffer
 -- @param lang language, defaults to current language
 local function enable_module(mod, bufnr, lang)
-  local bufnr = bufnr or api.nvim_get_current_buf()
-  local lang = lang or parsers.get_buf_lang(bufnr)
+  bufnr = bufnr or api.nvim_get_current_buf()
+  lang = lang or parsers.get_buf_lang(bufnr)
   M.attach_module(mod, bufnr, lang)
 end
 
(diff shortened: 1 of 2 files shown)


clason:
Can you explain why `bufnr` and `lang` need to be global? Won't that lead to problems with multiple open buffers?

theHamsta:
@clason they aren't global since they are local function parameters. I prefer the shadowing style since this prevents accidental globals. But our CI catches those globals anyway.

theHamsta (approved):
This looks good although we should probably also make the attach/detach of the module more reliable. I also don't know whether users would know that this `stop` function would not only stop highlighting but also all other tree-sitter functionality

kyazdani42:
@clason the linter is not happy when redefining variables. That's why i removed the local which does not make them global.

@theHamsta the stop function will be used through `nvim-treesitter.highlight.stop` so i didn't specify `highlighter` because it is already semantically correct i guess :)

clason:
I'm also not enthusiastic about modifying arguments ;) The standard pattern is what was there before (copy to a local variable or default). But if you're sure it's fine in this specific situation...

kyazdani42:
The LSP complains, i wouldn't have changed that otherwise. I've double checked though:
```lua
function v(b) b = b or "a" end; v(); print(b)
-- ni
```l

clason:
From my knowledge of language servers, I wouldn't take their word for gospel ;)

clason (approved):

PR #2474, opened 2022-02-06
refacto: expose a function to set custom capture for the higlighter

kyazdani42:
BREAKING: remove custom_captures from highlight config.

This allows plugin authors to extend the map. It also avoids settings user configuration specific values during the highlighter initialization (SOC).
Not sure how much value this brings, and might potentially break a lot of people configurations. This is questionable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,10 +273,6 @@ lua <<EOF
 require'nvim-treesitter.configs'.setup {
   highlight = {
     enable = true,
-    custom_captures = {
-      -- Highlight the @foo.bar capture group with the "Identifier" highlight group.
-      ["foo.bar"] = "Identifier",
-    },
     -- Setting this to true will run `:h syntax` and tree-sitter at the same time.
     -- Set this to `true` if you depend on 'syntax' being enabled (like for indentation).
     -- Using this option may slow down your editor, and you may see some duplicate highlights.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -273,10 +273,6 @@ lua <<EOF
 require'nvim-treesitter.configs'.setup {
   highlight = {
     enable = true,
-    custom_captures = {
-      -- Highlight the @foo.bar capture group with the "Identifier" highlight group.
-      ["foo.bar"] = "Identifier",
-    },
```
theHamsta, line 279: How does this change anything in the user experience? Wouldn't it by better to leave it in the config (since it uses the value at the right moment)? Especially since I think that changing the `hl_map` to late often does not have the desired outcome. We also remove that setting and tell users to use `hl_map` directly

kyazdani42: well this was a bit what i thought. I don't really like the map being set during the highlighter initialization (it might run a 1000 times during a user session which is not really expected).
I just put back the default custom_captures = {} in the config because it broke telescope. 
The reason why is to avoid plugins doing funky stuff like this:
https://github.com/nvim-telescope/telescope.nvim/blob/f262e7d56d37625613c5de0df5a933cccacf13c5/lua/telescope/previewers/utils.lua#L97-L116

kyazdani42: and yes i guess telling user to configure their hl_map directly would make more sense. but it's not documented in the neovim side yet.

lua/nvim-treesitter/configs.lua
```diff
@@ -22,8 +22,9 @@ local is_initialized = false
 local builtin_modules = {
   highlight = {
     module_path = "nvim-treesitter.highlight",
-    enable = false,
+    -- @deprecated: use `highlight.set_custom_captures` intead
```
theHamsta, line 25: Typo

theHamsta (approved):

PR #2471, opened 2022-02-06
refacto: deprecate used_by in parsers.lua

kyazdani42:
TODO: should i create a public function to modify this list instead of providing the user with the table ?

- remove print_warning function from utils.lua (unused)
- cleanup some functions in parsers.lua (shadowing parameters and wrong
  bufnr used).
- log a deprecation notice when using used_by in a parser definition
- default the filetype_to_parsername table to the list of filetypes
  previously in the used_by keys
- update the README to indicate that change

fixes #2373

PR #2477, opened 2022-02-06
fix: readme typo, removing deprecated docs

carlosala:
The title is descriptive, thanks for the work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,10 +361,11 @@ EOF
 ```
 
 If you wish to set a specific parser for a filetype, you should extend the `filetype_to_parsername` table:
+
 ```vim
 lua <<EOF
-local ft_to_parser = require"nvim-treesitter.parser".filetype_to_parsername
-filetype_to_parsername.someft = "python" -- the someft filetype will use the python parser and queries.
+local ft_to_parser = require"nvim-treesitter.parsers".filetype_to_parsername
+ft_to_parser.someft = "python" -- the someft filetype will use the python parser and queries.
 EOF
 ```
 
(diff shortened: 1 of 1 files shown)


clason:
Are you sure that this is the change you intended? This reverts _intentional_ changes.

carlosala:
These are the docs I expect to have after #2471. `used_by` is deprecated, so removing its docs would be nice. And the other two lines are typos.

clason:
No, please check the diff again. I don't think you intended this: you're removing

* a rename to a doc snippet to match the corresponding rename in the code
* a new(!) doc snippet to show how you can use parsers with other filetypes now that `used_by` is removed.

carlosala:
I think that I don't understand something.
  - I removed a snippet explaining how to use `used_by`, which now does nothing, just reports an error.
  - I fixed new snippet, because `nvim-treesitter.parser` does not exist and is not using the local variable defined the previous line.

kyazdani42 (approved):
@clason that was my fault, i introduced typos indeed, this PR is good.
@carlosala can you also update the doc file please ?

carlosala:
Sure!

clason:
> @clason that was my fault, i introduced typos indeed, this PR is good.

Are you sure? I agree on 1., I confused that with a different snippet in the docs (sorry!), but the `ft_to_parser.someft` doesn't look correct -- you renamed that to `filetype_to_parsername` in that PR, didn't you?

clason:
Sorry, I indeed failed to read this correctly. I apologize!

carlosala:
@kyazdani42 I don't see where should I change something in the doc file, I don't find anything related to this. Could you please give me some hint?

theHamsta (approved):

kyazdani42:
@carlosala cannot find the reference to the doc i wrote earlier :confused: lets merge this. thanks for the patch :)

PR #2479, opened 2022-02-06
test: restructure indent tests

MunifTanjim:
Often times I add a bunch of `print` statements in `indent.lua` module to track exactly what happens in each steps. And I want to run indent on a specific line to debug it.

But running `./scritps/run_tests.sh indent` runs indent on all the languages. It's impossible to find debug logs for a specific line. So I need to go into all the indent `*_spec.lua` files and change `describe` to `pending` to skip those. As tests for more languages are being added, doing that is so fraustrating.

With this PR, it will be possible to run `./scripts/run_tests.sh indent/lua` to run indent tests for a specific language without modifying `*_spec.lua` files for all the other languages.


theHamsta:
It was posy before to run `indent/lua_spec.lua`

MunifTanjim:
> It was posy before to run indent/lua_spec.lua

🤦🏼 Why didn't I think for writing the whole filename!!! 😂

theHamsta:
Me neither, just discovered that today. Would still love to have the tests restructured to make it easier to work with them.

MunifTanjim:
Reopening to continue the discussion.

> (also have the functions in indent.lua directly testable with less global-state vim functions)

I'm not sure what you meant by this 🤔

MunifTanjim:
From https://github.com/nvim-treesitter/nvim-treesitter/pull/2480#pullrequestreview-874089671

> Maybe we should just have reduce the boiler plate in the indent tests. Can we describe the tests only with the example files without the _spec ones?

Unless we're making a tool for tree-sitter indent test, the `_spec.lua` files are needed for running tests with plenary.nvim. But we can sure reduce the boilerplates and make writing tests easier and less verbose.

theHamsta:
But we also have only one _spec file for all highlight tests with test assertions directly in the source files. I'm not sure whether this would be an improvement since you can run the test for multiple languages easier when you have different _spec files.

MunifTanjim:
If I add debug `print` calls inside `indent.lua` file:
 
- Running `./scripts/run_tests.sh indent/javascript_spec.lua` doesn't output anything from those `print` statements 🤷🏼‍♂️ 
- Running `./scripts/run_tests.sh indent/` outputs those `print` statements. But running selective tests selectively for a single language is a massive pain.

theHamsta:
I never see saw print statements in any of the tests :'-(

MunifTanjim:
> I never see saw print statements in any of the tests :'-(

Just looked into plenary source... removing this condition fixes the issue for me: https://github.com/nvim-lua/plenary.nvim/blob/2a278c8a12a399e25b78a43ebcd4f3996cd4e4b6/lua/plenary/test_harness.lua#L113

Unblocks me for now. Not sure why these conditions are added in plenary. Couldn't find much documentation.

MunifTanjim:
Closing since the PR is too outdated.

PR #2480, opened 2022-02-06
test(indent): fix typos

MunifTanjim:


theHamsta (approved):
Maybe we should just have reduce the boiler plate in the indent tests. Can we describe the tests only with the example files without the _spec ones?

PR #2485, opened 2022-02-07
feat(vim): highlight default parameters

reegnz:
Support to tree-sitter-viml was added with https://github.com/vigoux/tree-sitter-viml/pull/57


reegnz:
Follow up of #2441

PR #2500, opened 2022-02-09
highlights(cpp): Highlight method with nested qualified_identifier

Sh3Rm4n:
so that methods like these are correctly highlighted

```cpp

class A {
    class B {
       void foo();
    };
};

void A::B::foo() {
//         ^^^
}
```

This only increases the nesting level by one.
AFAIK, arbitrary nesting is difficult to do with current queries.

But this nesting is a pretty common case


theHamsta (approved):

theHamsta:
Thank you!

PR #2506, opened 2022-02-09
chore: remove lua heredocs in documentation and readme

### LINKED ISSUE #2503: Should README default to Lua?
At this stage, should we move from `init.vim` with `lua <<EOF` to `init.lua` and give the config examples the `lua` highlight?

kyazdani42:
fixes #2503

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ To make sure a parser is at the latest compatible version (as specified in `nvim
 
 Each module provides a distinct tree-sitter-based feature such as [highlighting](#highlight), [indentation](#indentation), or [folding](#folding); see [`:h nvim-treesitter-modules`](doc/nvim-treesitter.txt) or ["Available modules"](#available-modules) below for a list of modules and their options.
 
+Following examples assume that you are configuring neovim with lua. If you are using vimscript, see `:help lua-heredoc`.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -98,10 +98,9 @@ To make sure a parser is at the latest compatible version (as specified in `nvim
 
 Each module provides a distinct tree-sitter-based feature such as [highlighting](#highlight), [indentation](#indentation), or [folding](#folding); see [`:h nvim-treesitter-modules`](doc/nvim-treesitter.txt) or ["Available modules"](#available-modules) below for a list of modules and their options.
 
-All modules are disabled by default and need to be activated explicitly in your `init.vim`, e.g., via
+All modules are disabled by default and need to be activated explicitly in your `init.lua`, e.g., via
```
theHamsta, line 101: I'd still say that init.vim is more common

theHamsta (approved):

README.md
kyazdani42: it is, but the following code is in lua so better to be explicit here. Should i merge ?

theHamsta: would you mind referencing `:h lua-heredoc` for someone not clear on what to do with the Lua code (e.g. *"(see `:h lua-heredoc` if you're using Vimscript for init.vim)"*)

PR #2513, opened 2022-02-11
Improve Vala syntax highlight

matbme:
Tweaks to Vala's highlights.scm which adds coloring support for preprocessor and verbatim. I also fixed some highlighting inconsistencies.


queries/vala/highlights.scm
```diff
@@ -28,13 +28,10 @@
   "}"
   "("
   ")"
-  "["
-  "]"
 ] @punctuation.bracket
 
 [
   ";"
-  ":"
```
theHamsta, line 37: Same here?

queries/vala/highlights.scm
```diff
@@ -128,7 +129,6 @@
     "&="
     "^="
     "??"
-    "="
   ] @operator
```
theHamsta, line 132: Same here?

matbme:
My bad, I forgot to add these changes to the highlights.scm file in my repository, adding them now.

clason (approved):

PR #2541, opened 2022-02-18
Support gjs and gts

NullVoxPopuli:
More information on these syntaxes here:
https://github.com/emberjs/rfcs/pull/779


clason:
Thank you, but we are trying to get rid of the filetype detection -- this belongs upstream, not in this plugin. Please open a PR for vim/vim (which owns the filetype detection code); then it will be included in Neovim as part of the ongoing patch porting efforts.

PR #2548, opened 2022-02-19
ci(tests): download highlight-assertions for faster CI

theHamsta:
This avoids any Rust build dependencies.


clason (approved):

theHamsta:
Test dependencies: 1s

I hope that rust-build actions will soon enable builds with GNU ABI. Then, I would not need to setup a Docker config or cargo xcross manually.

PR #2549, opened 2022-02-19
Add embedded template to support ERB files

npezza93:


queries/embedded_template/injections.scm
```diff
@@ -0,0 +1,2 @@
+(content) @html @combined
+(code) @ruby @combined
```
theHamsta, line 2: I think we haven't a complete implementation of `@combined` in nvim atm.

theHamsta (approved):

PR #2552, opened 2022-02-19
highlights(php): highlight use declaration name as type

wilmanbarrios:


queries/php/highlights.scm
```diff
@@ -29,6 +29,8 @@
   (name) @constant)
 (trait_declaration
   name: (name) @type)
+(use_declaration
+    (name) @type)
```
theHamsta, line 33: is this always a type or could you also `use` a function? Same for your other PR but there namespaced expressions where already always highlighted as type

wilmanbarrios: >is this always a type or could you also use a function?

Is always a type. Functions are used at the top:

```php
<?php

namespace App\Models;

use function Foo\Barr\bazz;

class User extends Authenticatable 
{ 
    //
}
```
>Same for your other PR but there namespaced expressions where already always highlighted as type

The majority of cases are always highlighted as type. But this case as an example is not highlighted as a type anymore, with this query, gets fixed.

theHamsta (approved):

PR #2553, opened 2022-02-19
highlights(php): named arguments as constructor

wilmanbarrios:


queries/php/highlights.scm
```diff
@@ -62,6 +62,9 @@
 (object_creation_expression
   [(name) (qualified_name)] @constructor)
 
+(argument
+    (name) @constructor)
```
theHamsta, line 66: Any reason for this change? This is not done for Python named arguments.

wilmanbarrios: I don't know much about syntax highlighting in Python but, looking VSCode way of highlighting named arguments in various themes I thought would be cool to have this highlight

VSCode- Community Material Theme Palenight
image

Neovim - Palenight colorscheme _**without** this PR_
image

Neovim - Palenight colorscheme _**with** this PR_
image

theHamsta: but why not use `@parameter` for this highlight. All other usages of `@constructor` are for the type being constructed (alias the constructor call)

theHamsta: Hmm, I guess since this can only be used on constructors...

theHamsta: Python uses `@parameter` but it can also be used in normal function calls and normally `@parameter` is only used at the callee site... let me think a bit about that...

wilmanbarrios: >Hmm, I guess since this can only be used on constructors...

Same here

theHamsta: @kyazdani42 @stsewd what do you think about which highlight is best for this?

kyazdani42: i don't have a strong opinion about this, but i think `@parameter` makes more sense when it's a named function parameter. Using constructor is not really appropriate in this scenario though.

theHamsta: would agree with that. @wilmanbarrios can you change to `@parameter`?

wilmanbarrios: Yes sure, later today I will make de change :+1:

wilmanbarrios:
@theHamsta all done :ok_hand:

wilmanbarrios:
I messed up something in the rebase I will make another PR.

PR #2557, opened 2022-02-20
CSS styled-components highlighting fix

### LINKED ISSUE #1111: Context parsing for styled components in jsx/tsx

haras-unicorn:
Fixes #1111 at least temporarily.

Two new directives have been added:
 - `inject_without_children!`: excludes ranges of all children from a nodes content
 - `inject_without_named_children!`: excludes ranges of named children from a nodes content

I was instructed by @theHamsta to create `inject_without_children!` that excludes ranges of only named children, but that was not enough for this implementation, because `offset!` also adds its own unnamed children to a node, so those had to be excluded as well. In the particular case of ECMA template strings, this would include backticks which the CSS parser could not handle.

See the discussion at #1111 for more details.


Hrle97:
Here is a good showcase:

showcase

I tested this a bit further and noticed some edge cases where this is not working properly, but I don't know how to fix it:



It works for 99% of use cases in a project I have for work, though.

theHamsta:
Can you please post examples in plain text? What is the result of the parse tree? Is it due to the nested  backticks? Does the Javascript parser use `scanner.cc` for parsing the interpolation templates?

Please also add tests for this commit to show that the behavior is correct and also stays correct in case we will have future modifications.

@clason what path do you prefer? Emulate the tree-sitter upstream behavior where it's the default to exclude children of the parent language in the injections ("combined") in Nvim core or just have our own capture here? Is `inject_without children!` a clear name or is there some better alternative? `inject_without children_of_parent_language!`, `#inject-nested!`. `@injection.combined` like upstream tree-sitter (which is confusing)

theHamsta:
I thought that highlight tests would give us more insights on whether the CSS injection works correctly. But I understand that this is difficult to test at the moment.

Hrle97:
> Please also add tests for this commit to show that the behavior is correct and also stays correct in case we will have future modifications. Probably you would just need to add some tests files like the ones that we already have in the test folder. You can atm only have assertions in comments in the top-level language. So you would have to squeeze the CSS template strings into the same line and have JS comments with assertions below it.

I added a test file, but to test properly some template strings have to be multiline - is this okay?

Here's an example:

```ts
const StyledComponentWithProp = styled(Component).attrs({ prop: true })<{
  isFixed?: boolean;
  margin: number | string;
}>`
  width: 100%;
/* ^ css */
  margin: ${({ margin }) => margin};
  ${({ isFixed }) => isFixed && css`position: fixed;`}
/*                                  ^ css */
`;
```

> Can you please post examples in plain text? What is the result of the parse tree?

I'm sorry but I don't know how to generate the parse tree. Could you provide an example? This is all really new to me, so my knowledge is lacking. 

> Is it due to the nested backticks? Does the Javascript parser use scanner.cc for parsing the interpolation templates?

It is not due to the nested backticks and the JS parser parses the template substitutions well. When you use `#offset!` it adds the backticks as child nodes which would solve all our problems if Neovim's implementation wasn't including children by default.

In example:

```ts
const StyledDiv = styled.div<{ isMobile?: boolean }>`
  width: 100%;
  margin: ${({ isMobile }) => (isMobile ? "2px" : "4px")};
`;
```

using the injection query:

```
(call_expression
 function: (member_expression
   object: (identifier) @_name
     (#eq? @_name "styled"))
 arguments: ((template_string) @css
   (#offset! @css 0 1 0 -1)
   (#inject_without_children! @css)))
```

the first child is an unnamed node of type '\`' (the first backtick), the second child is a node of type "template_substitution" (the ${...} thing), and the third child is an unnamed node of type '\`' (the second backtick). Then, the `#inject_without_children!` directive removes all ranges of all children (named and unnamed) from the template string content effectively splitting it into two subranges between the first backtick and the template substitution and between the template substitution and the second backtick. 

If we were to use the `#inject_without_named_children!` directive it would only remove the template substitution child from the content of the template string because the backtick children nodes created by `#offset!` are not named. This creates two subranges from the start of the template string (including the first backtick) up to the template substitution and from the end of the template substitution up to the end of the template string (including the last backtick) making the CSS parser panic.

Hrle97:
> I thought that highlight tests would give us more insights on whether the CSS injection works correctly. But I understand that this is difficult to test at the moment.

I can try to add a highlights test if you want, but its getting late and I will have limited time over the week to work on this, so I can't guarantee I will be able to come up with anything soon.

Hrle97:
> @clason what path do you prefer? Emulate the tree-sitter upstream behavior where it's the default to exclude children of the parent language in the injections ("combined") in Nvim core or just have our own capture here? Is inject_without children! a clear name or is there some better alternative? inject_without children_of_parent_language!, #inject-nested!. @injection.combined like upstream tree-sitter (which is confusing). How should we treat @combined in core which is not the @injection.combined of upstream tree-sitter. I guess upstream always has only one root of injected language without distinction of multiple roots/"@combined" but would need to look into implementation to confirm that.

I know my input wasn't asked here, but I feel the need to say this. My intuition tells me to emulate upstream behaviour because I believe it is best to follow standards, unless there is some kind of bottleneck going on in Neovim of which I'm not aware.

theHamsta:
@Hrle97  as I said. I think it is not needed as it's really difficult to test. Thanks a lot for your improvement and reviving that old PR!

theHamsta:
> I know my input wasn't asked here, but I feel the need to say this. My intuition tells me to emulate upstream behaviour because I believe it is best to follow standards, unless there is some kind of bottleneck going on in Neovim of which I'm not aware.

But then we wouldn't merge your PR and try to bring this behavior to upstream Neovim right away. This is why we should wait until we know how we proceed in the core implementation

clason:
> @clason what path do you prefer? Emulate the tree-sitter upstream behavior where it's the default to exclude children of the parent language in the injections ("combined") in Nvim core or just have our own capture here? Is inject_without children! a clear name or is there some better alternative? inject_without children_of_parent_language!, #inject-nested!. @injection.combined like upstream tree-sitter (which is confusing). How should we treat @combined in core which is not the @injection.combined of upstream tree-sitter. I guess upstream always has only one root of injected language without distinction of multiple roots/"@combined" but would need to look into implementation to confirm that.

I haven't had the time yet to think through the ramifications of this specific situation, but my gut feeling is that matching upstream behavior is always better, all else being equal (which it almost never is) -- not least because in most cases that informs parser behavior. There's also a fairly low bar to making breaking changes in core if they lead to actual improvements (after all, that's one of the main reasons why tree-sitter is still considered experimental).

Maybe we can ask Max for his views on this and go from there? (It's not inconceivable that _upstream_ changes behavior if that turns out to be an improvement in practice...)

theHamsta:
I also would almost always try to go for implementing the upstream behavior. The only doubt is that we implemented something different of what we thought what the upstream behavior would be.

Best to open a Github discussion with a short summary of the context for all people involved.

Hrle97:
> But then we wouldn't merge your PR and try to bring this behavior to upstream Neovim right away. This is why we should wait until we know how we proceed in the core implementation

That is true, hence why I said this is probably just a temporary solution.

clason:
The injection logic has been changed significantly. If the query changes are still applicable, please open a new PR for them from latest `master`.

lucario387:
This is obsolete as tree-sitter-styled is added and should fully support this use case

PR #2561, opened 2022-02-21
Various fixes to queries/perl/*

leonerd:
* Recognise many more keywords
* Assign highlight groups to control keywords, not entire structures
* Add some folding rules

Verymuch a work-in-progress as many bigger fixes will require improvements in `tree-sitter-perl`, but this is a good start.

### DIFF
--- a/queries/perl/highlights.scm
+++ b/queries/perl/highlights.scm
@@ -1,84 +1,115 @@
-; Variables
-(variable_declaration
-  .
-  (scope) @keyword)
+; Misc keywords
 [
-(single_var_declaration)
-(scalar_variable)
-(array_variable)
-(hash_variable)
-(hash_variable)
+  "my" "our" "local"
+  "next" "last" "redo"
+  "goto"
+  "package"
+;  "do"
+;  "eval"
+] @keyword
+
+; Keywords for including
+[ "use" "no" "require" ] @include
+
+; Keywords that mark conditional statements
+[ "if" "elsif" "unless" "else" ] @conditional
+
+; Keywords that mark repeating loops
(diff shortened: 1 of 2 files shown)


clason (approved):

queries/perl/highlights.scm
```diff
@@ -1,85 +1,116 @@
...
+[ "sub" ] @keyword.function
+
+; Lots of builtin functions, except tree-sitter-perl doesn't emit most of
+;   these yet
+;[
+;  "print" "printf" "sprintf" "say"
+;  "push" "pop" "shift" "unshift" "splice"
+;  "exists" "delete" "keys" "values"
+;  "each"
+;] @function.builtin
```
theHamsta, line 38: Isn't it possible to recognize them with `#any-of?`? I'd imagine that they are at least present in the tree as `identifier`s in call position.

leonerd: Uh quite possibly? :) I have never seen/heard-of `#any-of` before so perhaps a reference link would be useful?

clason: You can look at other queries (Python, cmake). The documentation is `:h lua-treesitter-predicates`.

leonerd: Ah, I see. That'd add quite a lot. To go through the *full* set of core functions, would probably add about 200 names here. I'm happy to do that but perhaps in a separate commit, to keep this one moving.
What do people think?

theHamsta: sounds good. I feel that the distinction built-in function vs regular one is not essential

PR #2563, opened 2022-02-22
feat(vlang): initial support

tami5:
Add support for vlang filetypes.

- [ ] Highlight `C` as builtin variable. This is FFI in vlang land,
  where C act like extern and access c functions. The vlang parser differentiate between C function calls and normal function callas as well as c and vlang arguments but I believe  highlighting C as builtin variable is sufficient indicator for now. I  tried to use `offset!` but failed. Any suggestions?
- [x] Set up parser url. the vlang parser is located within [vls] repo. Is installing from nested repo supported? `tree_sitter_v/src/parser.c`?
- [ ] correctly parse `string_interpolation`
- [ ] keyword shouldn't be highlighted when they are arguments or they are initialized/defined in scope?

[vls]: https://github.com/vlang/vls/tree/master/tree_sitter_v

cc @elianiva @theHamsta

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1068,6 +1068,17 @@ list.proto = {
   filetype = "proto",
 }
 
+list.v = {
+  install_info = {
+    url = "~/sources/vls/tree_sitter_v",
+    files = { "src/parser.c", "src/scanner.c" },
+    generate_requires_npm = false,
+    requires_generate_from_grammar = false,
+  },
+  filetype = "vlang",
+  maintainers = { "@tami5" },
+}
+
 local M = {
   list = list,
   filetype_to_parsername = filetype_to_parsername,
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1070,8 +1070,9 @@ list.proto = {
 
 list.v = {
   install_info = {
-    url = "~/sources/vls/tree_sitter_v",
+    url = "https://github.com/vlang/vls",
     files = { "src/parser.c", "src/scanner.c" },
+    location = "tree-sitter-v/tree_sitter_v",
     generate_requires_npm = false,
     requires_generate_from_grammar = false,
   },
--- a/queries/v/folds.scm
+++ b/queries/v/folds.scm
@@ -1,8 +1,6 @@
-;; WARNING: Doesn't seems to work ..
 [(function_declaration)
  (const_declaration)
  (type_declaration)
- (var_declaration)
  (import_declaration)
  (if_expression)
  (struct_declaration)] @fold
--- a/queries/v/folds.scm
+++ b/queries/v/folds.scm
@@ -1,0 +1,8 @@
+;; WARNING: Doesn't seems to work ..
+[(function_declaration)
+ (const_declaration)
+ (type_declaration)
+ (var_declaration)
+ (import_declaration)
+ (if_expression)
+ (struct_declaration)] @fold
--- a/queries/v/highlights.scm
+++ b/queries/v/highlights.scm
@@ -22,7 +22,7 @@
   (#not-has-parent? @_parent call_expression special_call_expression)))
 
 ((identifier) @variable.builtin
- (#match? @variable.builtin "^err$"))
+ (#match? @variable.builtin "^(err|macos|linux|windows)$"))
 
 ;; C: TODO: fixme make `C`.exten highlighted as variable.builtin
 ; ((binded_identifier) @content
--- a/queries/v/highlights.scm
+++ b/queries/v/highlights.scm
@@ -3,8 +3,17 @@
 
 ;; Identifiers -------------------
 (import_path) @namespace
-[(module_identifier) ] @variable.builtin
-[(identifier)] @variable
+(module_identifier) @variable.builtin
+(identifier) @variable
+(interpreted_string_literal) @string
+(string_interpolation) @none
+
+; TODO: Have differnt highlight to make then standout + highlight }{$ as special
+; ((string_interpolation
+;   (identifier) @constant
+;   "$" @punctuation.special
+;   "${" @punctuation.special
+;   "}" @punctuation.special))
 
 [(type_identifier) (array_type) (pointer_type)] @type
 
@@ -371,8 +380,6 @@
 ;; Literals
 
 (int_literal) @number
-
-(interpreted_string_literal) @string
 
 (rune_literal) @string
 
--- a/queries/v/highlights.scm
+++ b/queries/v/highlights.scm
@@ -1,0 +1,387 @@
+;; reference https://github.com/vlang/vls
+;; rev: c3e9874fa6c3b38beaa50d53aff4967403e251a4
+
+;; Identifiers -------------------
+(import_path) @namespace
+[(module_identifier) ] @variable.builtin
+[(identifier)] @variable
+
+[(type_identifier) (array_type) (pointer_type)] @type
+
+(field_identifier) @property
+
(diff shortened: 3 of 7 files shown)


theHamsta:
Installing from sub location is supported please have a look at the TSX/typescript parser which share a repo with different subfolders

tami5:
> Installing from sub location is supported please have a look at the TSX/typescript parser which share a repo with different subfolders

😍 thanks for the patch

lua/nvim-treesitter/parsers.lua
```diff
@@ -938,6 +938,18 @@ list.lalrpop = {
   maintainers = { "@traxys" },
 }
 
+list.v = {
+  install_info = {
+    url = "https://github.com/vlang/vls",
+    files = { "src/parser.c", "src/scanner.c" },
+    location = "tree-sitter-v/tree_sitter_v",
```
theHamsta, line 945: yeah, this is quite hacky. Even I forgot that we have to do this hack since it will copy the parser to a `tree-sitter-{lang}` repo :man_facepalming:

theHamsta: we could auto-add that `tree-sitter-{lang}/` prefix since its not obvious that it is necessary

queries/v/indents.scm
```diff
@@ -0,0 +1,18 @@
+[(import_declaration
+  (const_declaration)
+  (var_declaration)
+  (type_declaration)
```
theHamsta, line 4: image

 `var_declaration` apparently does not exist as a node

queries/v/folds.scm
```diff
@@ -0,0 +1,8 @@
+;; WARNING: Doesn't seems to work ..
+[(function_declaration)
+ (const_declaration)
+ (type_declaration)
+ (var_declaration)
+ (import_declaration)
```
theHamsta, line 6: image

queries/v/indents.scm
tami5: Uh most overlooked it. Thanks

queries/v/locals.scm
```diff
@@ -0,0 +1,29 @@
...
+  name: (binded_identifier
+          name: (identifier) @definition.function)))
+
+(const_declaration (const_spec (identifier) @definition.var))
+
+(identifier) @reference
+(type_identifier) @reference
+
+((call_expression function: (identifier) @reference)
+ (set! reference.kind "call"))
```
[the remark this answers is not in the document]
tami5: done. btw, here a list of files where it is still mentioned.

```
queries/php/locals.scm  
queries/php/locals.scm  
queries/php/locals.scm  
queries/php/locals.scm  
queries/php/locals.scm  
queries/php/locals.scm  
queries/php/locals.scm  
queries/php/locals.scm  
queries/php/locals.scm  
queries/php/locals.scm  
queries/php/locals.scm  
queries/php/locals.scm  
queries/php/locals.scm  
queries/c/locals.scm    
queries/c/locals.scm    
queries/cpp/locals.scm  
queries/rust/locals.scm 
queries/rust/locals.scm 
queries/go/locals.scm
queries/go/locals.scm
queries/go/locals.scm
queries/go/locals.scm
queries/go/locals.scm
queries/ocaml/locals.scm
queries/ocaml/locals.scm
queries/ocaml/locals.scm
queries/ocaml/locals.scm
queries/ocaml/locals.scm
queries/ocaml/locals.scm
queries/ocaml/locals.scm
queries/ocaml/locals.scm
queries/ocaml/locals.scm
```

lua/nvim-treesitter/parsers.lua
tami5: Oh yah that's odd thing to set see. `tree-sitter-{lang}/` make sense to have it pre-added.

Maybe in the future location would be the name of the directory and clone would check for the location value and moved it, instead of the whole repo, even then it is still hacky 😆.

tami5:
So currently I'm working on correctly highlighting variables inside `interpreted_string_literal `

```clojure
(
  (interpreted_string_literal _ @_node) @string
  (#not-has-type? @_node string_interpolation))
```

hover this just overwrite it
```clojure
(interpreted_string_literal) @string
```

🤔

theHamsta:
This will work similarly to how it's done for Python interpolations https://github.com/tami5/nvim-treesitter/blob/lang/vlang/initial_support/queries/python/highlights.scm#L8 or bash https://github.com/nvim-treesitter/nvim-treesitter/blob/f735498a645e1a2aca7a0cfdaa2d7f8cec543846/queries/bash/highlights.scm#L1-L4

tami5:
> > So currently I'm working on correctly highlighting variables inside `interpreted_string_literal `
> > ```clojure
> > (
> >   (interpreted_string_literal _ @_node) @string
> >   (#not-has-type? @_node string_interpolation))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > hover this just overwrite it
> > ```clojure
> > (interpreted_string_literal) @string
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > thinking
> 
> This will work similarly to how it's done for Python interpolations [tami5/nvim-treesitter@`lang`/vlang/initial_support/queries/python/highlights.scm#L8](https://github.com/tami5/nvim-treesitter/blob/lang/vlang/initial_support/queries/python/highlights.scm#L8) or bash
> 
> https://github.com/nvim-treesitter/nvim-treesitter/blob/f735498a645e1a2aca7a0cfdaa2d7f8cec543846/queries/bash/highlights.scm#L1-L4

Oh wow, thanks @theHamsta

queries/v/highlights.scm
```diff
@@ -3,8 +3,17 @@
...
+(identifier) @variable
+(interpreted_string_literal) @string
+(string_interpolation) @none
+
+; TODO: Have differnt highlight to make then standout + highlight }{$ as special
+; ((string_interpolation
+;   (identifier) @constant
+;   "$" @punctuation.special
+;   "${" @punctuation.special
+;   "}" @punctuation.special))
```
tami5, line 16: Okay it worked as you've suggested @theHamsta, however, couldn't quite figure out making identifiers highlight  differently

this is tree with two identifiers that start with `$`

```clojure
(interpreted_string_literal 
  (string_interpolation (identifier))
  (string_interpolation (identifier))
```

theHamsta: I think this requires a change to the parser. Can you try to use a plain `"$"` here without the immediate https://github.com/vlang/vls/blob/e4ae12f117e7241a61db6a0c11e184d6ca0fe6b3/tree_sitter_v/grammar.js#L1585 or use ` alias(token.immediate("$"), "$")`. The braces seem to be handled by the external parser. In case you're using, nvim-treesitter/playground press `a` to see all anonymous tokens that are visible to the query engine.

theHamsta:
I would suggest the following changes: https://github.com/theHamsta/nvim-treesitter/commit/b12af177ae2702104b6ee15837c67c94aecd0437 You put a comment regard `"import" @include` I can't judge how this should be fixed since I didn't have a look at the namespaced identifiers.

theHamsta:
@tami5 any plans to continue with this one?

tami5:
> @tami5 any plans to continue with this one?

Thanks for the heads up. I totally forgot about it. I'm going to finish up tomorrow insa

tami5:
@theHamsta I believe the PR is mergable for the purpose of initial support, other pending stuff I will open issues to track them.

theHamsta:
Thank you! In general you should avoid merging master and instead do rebasing.

theHamsta (approved):

PR #2577, opened 2022-02-25
RegExp, GraphQL, Markdown in HTML; HTML and CSS in JS

### LINKED ISSUE #2569: Embedded Regexp in HTML <input pattern="">
[HTML `<input>` elements can accept a `pattern` attribute containing a regexp](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern) 

nvim-treesitter should inject `pattern` (from regexp) into the attribute text content
Screen Shot 2022-02-23 at 23 53 56

manually parsing?

**Additional context**

https://github.com/tree-sitter/tree-sitter-html/issues/37#issuecomment-1049494680

bennypowers:
- [x] HTML: add regexp for input patterns (closes #2569)
- [x] HTML: add `<script type="application/graphql">`
- [x] HTML: add `<script type="application/json">`
- [x] HTML: add `<script type="text/markdown">`
- [x] HTML: `<script async defer>` parses as JS
- [x] HTML: `<script>` still parses as js
- [x] JS: add `html` as HTML embed for tagged template literals
- [x] JS: add `css` as CSS embed for tagged template literals

Screen Shot 2022-02-24 at 20 49 30

## Notes and Questions to Reviewers:
- I haven't yet figured out how to express "a script_element with no attribute child", i.e. `<script>` so plain scripts are broken
- ~~I'm not sure why markdown isn't working~~ PEBCAK - installed the parser and it's fine
- I listed input twice, ~~once for the self-closing variant, another for the start_tag variant.~~ once for the quoted attribute value and once for unquoted (this is legal HTML) Is it possible to alternate (`OR`, i.e. `||` in js syntax) on the node type?

### DIFF
--- a/queries/ecma/injections.scm
+++ b/queries/ecma/injections.scm
@@ -2,38 +2,25 @@
   (#match? @_jsdoc_comment "^/\\*\\*[^\\*].*\\*/")) @jsdoc)
 
 (comment) @comment
+(regex_pattern) @regex
 
 (call_expression
  function: ((identifier) @language)
  arguments: ((template_string) @content
    (#offset! @content 0 1 0 -1)))
 
+; TODO: map tag names to languages
 (call_expression
  function: ((identifier) @_name
    (#eq? @_name "gql"))
  arguments: ((template_string) @graphql
    (#offset! @graphql 0 1 0 -1)))
 
-; html`<input value="hello"/>`
-(call_expression
- function: ((identifier) @_name
-   (#eq? @_name "html"))
- arguments: ((template_string) @html
-   (#offset! @html 0 1 0 -1)))
-
 (call_expression
  function: ((identifier) @_name
    (#eq? @_name "hbs"))
  arguments: ((template_string) @glimmer
    (#offset! @glimmer 0 1 0 -1)))
-
-; css`:host { display: none; }`
-(call_expression
- function: (member_expression
-   object: (identifier) @_name
-     (#eq? @_name "css"))
- arguments: ((template_string) @css
-   (#offset! @css 0 1 0 -1)))
 
 ; styled.div`<css>`
 (call_expression
@@ -61,7 +48,6 @@
  arguments: ((template_string) @css
    (#offset! @css 0 1 0 -1)))
 
-
 ; styled(Component).attrs({ prop: "foo" })`<css>`
 (call_expression
  function: (call_expression
@@ -71,8 +57,6 @@
         (#eq? @_name "styled"))))
  arguments: ((template_string) @css
    (#offset! @css 0 1 0 -1)))
-
-(regex_pattern) @regex
 
 ((comment) @_gql_comment
   (#eq? @_gql_comment "/* GraphQL */")
--- a/queries/ecma/injections.scm
+++ b/queries/ecma/injections.scm
@@ -14,11 +14,26 @@
  arguments: ((template_string) @graphql
    (#offset! @graphql 0 1 0 -1)))
 
+; html`<input value="hello"/>`
+(call_expression
+ function: ((identifier) @_name
+   (#eq? @_name "html"))
+ arguments: ((template_string) @html
+   (#offset! @html 0 1 0 -1)))
+
 (call_expression
  function: ((identifier) @_name
    (#eq? @_name "hbs"))
  arguments: ((template_string) @glimmer
    (#offset! @glimmer 0 1 0 -1)))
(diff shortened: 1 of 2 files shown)


queries/html/injections.scm
```diff
@@ -6,7 +6,50 @@
...
+          (attribute_name) @_attr
+          (quoted_attribute_value (attribute_value) @regex)))
+       (#eq? @_attr "pattern")))
+
+(script_element
+   (start_tag
+      ((attribute
+           (attribute_name) @_attr (#eq? @_attr "type")
+           (quoted_attribute_value (attribute_value) @_type))
+        (#any-of? @_type "module" "application/javascript")))
```
omertuc, line 32: Might want to use a more complete list https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#textjavascript

queries/html/injections.scm
```diff
@@ -6,7 +6,50 @@
...
+          (quoted_attribute_value (attribute_value) @regex)))
+       (#eq? @_attr "pattern")))
+
+(element
+   (start_tag
+      (tag_name) @_tagname (#eq? @_tagname "input")
+      ((attribute
+          (attribute_name) @_attr
+          (quoted_attribute_value (attribute_value) @regex)))
+       (#eq? @_attr "pattern")))
```
omertuc, line 25:
```suggestion
(element
   (_
      (tag_name) @_tagname (#eq? @_tagname "input")
      ((attribute
          (attribute_name) @_attr
          (quoted_attribute_value (attribute_value) @regex)))
       (#eq? @_attr "pattern")))
```

omertuc:
>I'm not sure why markdown isn't working

Markdown is working for me. Did you make sure you have markdown installed? Injections that are not installed are ignored silently

queries/html/injections.scm
bennypowers: done thanks

queries/html/injections.scm
bennypowers: great suggestion. I tried the same for unquoted attrs, but i think that's a genuine case for alternations, which appear to not be supported in these files?

bennypowers:
Thanks for the suggestions, Omer

I applied as many as i could, but there are still issues, namely that negations are apparently not supported, so that `<script>console.log()</script>` will not get highlighted
Screen Shot 2022-02-25 at 7 42 23

bennypowers:
I think this is ready. Thanks to @stsewd in https://github.com/nvim-treesitter/nvim-treesitter/discussions/2582#discussioncomment-2258096 I matched the plain `<script>` tag. Do note that with that line added, embedded languages in template literals no longer work though

Screen Shot 2022-02-26 at 19 00 56

kyazdani42 (approved):

queries/ecma/injections.scm
```diff
@@ -12,12 +12,27 @@
  arguments: ((template_string) @graphql
    (#offset! @graphql 0 1 0 -1)))
 
+; html`<input value="hello"/>`
```
theHamsta, line 15: couldn't we enable this rule for any language?

bennypowers: gql tag maps to graphql language, for instance

theHamsta: You don't have to remove all of them. We eventually will solve the problem somehow. For instance I have a pull request that uses filetype.lua. we could also vendor this file or use plenaries file type library. to resolve the aliases at some point. The same problem exists for markdown.

theHamsta: @bennypowers any updates on that remark?

bennypowers: How's that?

bennypowers: @theHamsta can you please clarify, have latest changes addressed your comment?

theHamsta: Do the patterns cause any performance problems? Wouldn't it be better to create predicate that resolves the languages from `"application/XXX"`?

amaanq:
@bennypowers is this ok? Regarding this comment:
> Do the patterns cause any performance problems? Wouldn't it be better to create predicate that resolves the languages from "application/XXX"?

I'd like to merge then

bennypowers:
Closing in favour of smaller PRs:
- [`<input pattern="...">`](https://github.com/nvim-treesitter/nvim-treesitter/pull/4521)

PR #2594, opened 2022-02-28
parsers: Add todotxt parser

arnarg:
Adds a parser for the todo.txt syntax along with highlighting. Syntax can be found here: https://github.com/todotxt/todo.txt#todotxt-format-rules

As this syntax doesn't necessary map well to highlight names that exist in colorschemes available I mapped them to highlight names that look good (e.g. makes sense that a done task is greyed out so `@comment` made sense to me).

With my coloscheme ([gruvbox.nvim](https://github.com/ellisonleao/gruvbox.nvim/)) it looks like this:
Screenshot from 2022-02-28 14-10-44

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -411,6 +411,17 @@ list.c_sharp = {
   maintainers = { "@Luxed" },
 }
 
+list.todotxt = {
+  install_info = {
+    url = "https://github.com/arnarg/tree-sitter-todotxt.git",
+    files = { "src/parser.c" },
+    branch = "main",
+  },
+  filetype = "todotxt",
+  maintainers = { "@arnarg" },
+  experimental = true,
+}
+
 list.typescript = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-typescript",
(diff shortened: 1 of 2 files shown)


arnarg:
I got a failing lint check that I have fixed now but a workflow with parser compilation failed with `Error: parser for ocaml_interface is not installed` which I would think is unrelated to my changes.

lua/nvim-treesitter/parsers.lua
```diff
@@ -411,6 +411,17 @@ list.c_sharp = {
   maintainers = { "@Luxed" },
 }
 
+list.todotxt = {
+  install_info = {
+    url = "https://github.com/arnarg/tree-sitter-todotxt.git",
+    files = { "src/parser.c" },
+    branch = "main",
+  },
+  filetype = "todotxt",
```
kyazdani42, line 420: is this a legit filetype ? @theHamsta i think we don't accept these kind of parsers right ? Not sure about that. Otherwise it would be good to go.

theHamsta: We have no filetype detects but that would be fine for since it doesn't interfere with anything. We could also omit the filetype since the language of the parser is `todotxt` already.

theHamsta (approved):

kyazdani42 (approved):

PR #2597, opened 2022-03-01
highlights(cpp): Add operator cast highlight

Sh3Rm4n:
See https://en.cppreference.com/w/cpp/language/cast_operator for reference.

Before 

grafik

After

grafik

### DIFF
--- a/queries/cpp/highlights.scm
+++ b/queries/cpp/highlights.scm
@@ -61,6 +61,7 @@
  (#lua-match? @constructor "^[A-Z]"))
 
 (operator_name) @function
+"operator" @function
 "static_assert" @function.builtin
 
 (call_expression
@@ -71,7 +72,7 @@
               name: (qualified_identifier
                       name: (identifier) @function)))
 (call_expression
-  function: 
+  function:
       (qualified_identifier
         name: (qualified_identifier
               name: (qualified_identifier


queries/cpp/highlights.scm
```diff
@@ -63,6 +63,10 @@
 (operator_name) @function
 "static_assert" @function.builtin
 
+(function_definition
+    declarator: (operator_cast
+      ("operator" @function _)))
+
```
theHamsta, line 69: shouldn't we just always highlight `"operator"` as `@keyword.operator` with the rest of the query not being needed?

Sh3Rm4n: Hm. Not sure. Both of these approaches would make the highlighting consistent between all operator uses, but `operator` is technically not a keyword, AFAIK. 

Previously,  all operator overload functions were highlighted correctly as functions / members, as this query would also do this for this special case.

But thinking about it, to be correct this query must be just:

```suggestion
(operator_cast ("operator" @function _))
```

In the end I'm fine with `@keyword`.  `@keyword.operator` might be confusing, because the "real" operator is following `"operator"` directly, e.g. : 

```cpp
T operator->()
//^^^^^^^^^^ All the same color now
```

theHamsta: I see... Before all uses of operator were highlighted as "function" anyways. So we could go with your original function or just `"operator" @function`, right?

PR #2598, opened 2022-03-01
Add slint parser configuration and queries

jrmoulton:
Is there anything else that needs to be changed to have this merged?

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -461,6 +461,16 @@ list.supercollider = {
   filetype = "supercollider",
 }
 
+list.slint = {
+  install_info = {
+    url = "https://github.com/jrmoulton/tree-sitter-slint",
+    files = { "src/parser.c" },
+    branch = "main",
+  },
+  maintainers = { "@jrmoulton" },
+  filetype = "slint",
+}
+
 list.haskell = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-haskell",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -469,6 +469,7 @@ list.slint = {
   },
   maintainers = { "@jrmoulton" },
   filetype = "slint",
+  experimental = true,
 }
 
 list.haskell = {
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -468,7 +468,6 @@ list.slint = {
     branch = "main",
   },
   maintainers = { "@jrmoulton" },
-  filetype = "slint",
   experimental = true,
 }
 
--- a/queries/slint/highlights.scm
+++ b/queries/slint/highlights.scm
@@ -18,7 +18,10 @@
 (value) @number
 (string) @string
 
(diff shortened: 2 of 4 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -461,6 +461,17 @@ list.supercollider = {
...
 }
 
+list.slint = {
+  install_info = {
+    url = "https://github.com/jrmoulton/tree-sitter-slint",
+    files = { "src/parser.c" },
+    branch = "main",
+  },
+  maintainers = { "@jrmoulton" },
+  filetype = "slint",
```
clason, line 471: I don't think you need that if your parser has the same name as the filetype.

clason:
You could also add the parser (with the matching revision) to the `lockfile.json` -- this way users don't have to wait for the automated update (which is currently blocked by another parser).

jrmoulton:
Those should be done now

queries/slint/highlights.scm
```diff
@@ -0,0 +1,97 @@
...
+(comment) @comment
+
+(value) @number
+(string) @string
+
+[
+"struct"
+"property"
+"callback"
+"import"
```
theHamsta, line 25: `@include`?

queries/slint/highlights.scm
```diff
@@ -0,0 +1,97 @@
...
+"from"
+"root"
+"parent"
+"this"
+"for"
```
theHamsta, line 30: `@conditional` `@repeat`?

jrmoulton: Thanks for the reviews. I appreciate it

queries/slint/highlights.scm
```diff
@@ -18,19 +18,29 @@
...
+
 [
 "struct"
 "property"
 "callback"
-"import"
-"from"
 "root"
 "parent"
 "this"
```
theHamsta, line 42: Is `this` a variable.builtin/constant.builtin? I don't know what the semantics of `in` is
 In some languages it is a keyword.operator

jrmoulton: Oh you're right, they should be variable.builtin. And after looking at other languages queries it seems like `in` should be `@repeat`

jrmoulton:
If there are any other changes that should be made to the highlights I can get them done

jrmoulton:
Perfect. Thanks for all the reviews!

queries/slint/highlights.scm
```diff
@@ -11,7 +11,7 @@
 (function_identifier) @function
 
 (reference_identifier) @keyword
-(visibility_modifier) @keyword
+(visibility_modifier) @include
```
theHamsta, line 14: I assume that `visibility_modifier` is something like public/private. This is usually not used in such a way in other languages. Or is the visibility_modifier directly used on import?

jrmoulton: Gotcha. For now it just references `export` but it's used in a similar way to public/private. In that case I'll change it back

theHamsta: Ok, then you were probably right. `export` should be `@include`. Sorry for the confusion.

theHamsta (approved):

jrmoulton:
Thanks!

PR #2616, opened 2022-03-08
fix: prefer gmake over make

theHamsta:
The problem with #2613 was that `if vim.fn.executable "gmake" then` should be `if vim.fn.executable "gmake" == 1 then` and probably `gmake` was missing on that CI environment.


clason (approved):

theHamsta:
Let's hope that this doesn't cause the same problem :see_no_evil:

PR #2620, opened 2022-03-11
Fix fennel function scope

harrygallagher4:
The current `locals.scm` for fennel binds a function's name only inside of the function itself. Setting the scope to "parent" is consistent with fennel's scoping.


theHamsta (approved):

theHamsta:
Goog catch! Thank you!

PR #2627, opened 2022-03-12
ci: ignore solidity for lockfile update

clason:
parser needs to be built from non-standard branch, which fails CI

@YongJieYongJie

PR #2629, opened 2022-03-12
Follow install info branch

theHamsta:
Alternative to #2627

We should still think of following just upstream with solidtiy


theHamsta:
Seems to do the right thing locally

clason (approved):

PR #2667, opened 2022-03-22
fix(nix): update highlight queries to parser change

clason:


clason:
We _really_ need better error messages for these checks that tell us _which_ part(s) of the query failed.

Issue #2665, opened 2022-03-22
Allow changes in the working directory when compiling the tree-sitter language


Dekker1:
**Is your feature request related to a problem? Please describe.**

I'm working on a language that includes a tree-sitter parser. Although it is great that this project can immediately provide language support in NVIM, it is a bit frustrating to install the tree-sitter language. Because the tree-sitter grammar is located in a subdirectory of the repository, using a configuration that I would expect to work gives a compilation error:

Configuration:
```
parser_config.mylang = {
	install_info = {
		url = "https://github.com/[repo_url]",
		branch = "develop",
		files = {"[subdir]/src/parser.c"}
	},
	filetype = "mylang",
	used_by = {"mylang"}
}
```

Error:
```
nvim-treesitter[mylang]: Error during compilation                                                                                                  
[subdir]/src/parser.c:1:10: error: 'tree_sitter/parser.h' file not found with <angled> include; use "quotes" instead                
#include <tree_sitter/parser.h>                                                                                                                      
         ^~~~~~~~~~~~~~~~~~~~~~                                                                                                                      
         "tree_sitter/parser.h"
```

It would be great if there was some way to provide either the include path or the subdirectory in which the tree-sitter structure is located.

**Describe the solution you'd like**

An additional field in `install_info` that allows the user to specify a subdirectory of the repository in which the tree-sitter structure is located. The compilation should then be executed from this directory.

**Describe alternatives you've considered**

- Additional include paths could be made explicit. (This would require additional arguments to the internal compilation function).
- The correct include path might be derived from the `files` path assuming a standard tree-sitter structure. (This makes an additional assumption and might require some path logic).
- The subdirectory might be provided in some form in the `url` field, like with Python `pip`'s installing from URL. (Would require more complex URL schemas).

**Additional context**

I understand that this is maybe not the most common scenario and that most tree-sitter grammars are seperate from a language. However, for a developing language that uses the tree-sitter grammar internally, using only one repository that includes the grammar to keep it in sync seems logical to me.

theHamsta:
For local development, you can just set it to the subfolder where your tree-sitter directory with standard tree-sitter layout is located. We have a work-around for this for tsx/typescript and ocaml parsers. See https://github.com/nvim-treesitter/nvim-treesitter/pull/2563#discussion_r812208864 as the vlang PR faced the same problem. We're open for PR that make this setting more expected/discoverable.

Dekker1:
[...]

PR #2670, opened 2022-03-23
edit(readme): note about location data in parser install_info

Dekker1:
This PR adds a note about a currently undocumented feature that solves #2665

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,7 @@ parser_config.zimbu = {
     files = {"src/parser.c"},
     -- optional entries:
     branch = "main", -- default branch in case of git repo if different from master
+    location = "", -- location of the tree-sitter parser in the (larger) repository
     generate_requires_npm = false, -- if stand-alone parser without npm dependencies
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -360,6 +360,7 @@ parser_config.zimbu = {
     files = {"src/parser.c"},
     -- optional entries:
     branch = "main", -- default branch in case of git repo if different from master
+    location = "", -- location of the tree-sitter parser in the (larger) repository
```
theHamsta, line 363: I think the default value is `nil`. Empty string is truthy in Lua

Dekker1: Happy to fix the `nil` issue, and make it a bit more clear that this is for "non-standard" locations.
However, I do not understand the comment about the required start using `tree-sitter-{lang}`. I current am using `location = "misc/tree-sitter-{lang}` as my location and that seems to work perfectly fine. Does it have to start with `tree-sitter-{lang}`, or maybe end with `tree-sitter-{lang}`?

clason:
included in the rewrite on `main`.

PR #2674, opened 2022-03-24
Add capture groups for remaining builtin highlight groups

gpanders:
Some builtin highlight groups (see `:h group-name`) do not yet have
associated capture groups, so add them.

Note that there is a related [PR in neovim][1] to add these groups to the
*default* `hl_map`, but this PR does not depend on that one in any way.

[1]: https://github.com/neovim/neovim/pull/17835

### DIFF
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -490,6 +490,10 @@ Boolean literals: `True` and `False` in Python.
 `TSCharacter`
 Character literals: `'a'` in C.
 
+						       *hl-TSCharacterSpecial*
+`TSCharacterSpecial`
+Special characters.
+
 								*hl-TSComment*
 `TSComment`
 Line comments and block comments.
@@ -514,6 +518,14 @@ Constants defined by macros: `NULL` in C.
 							    *hl-TSConstructor*
 `TSConstructor`
 Constructor calls and definitions: `{}` in Lua, and Java constructors.
+
+								  *hl-TSDebug*
+`TSDebug`
+Debugging statements.
+
+								 *hl-TSDefine*
+`TSDefine`
+Preprocessor #define statements.
 
 								  *hl-TSError*
 `TSError`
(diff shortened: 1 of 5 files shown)


clason:
You also need to add the captures to https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md, since that is the source of truth for the CI (which enforces capture names used by bundled queries).

clason:
Sorry you got hit by `luacheck` throwing a hissy fit about unrelated code (maybe an update?)... Probably just needs a set of parens?

clason:
Thanks! (To make up for it, we don't enforce conventional commits ;))

LGTM, but I defer to @theHamsta on this.

clason (approved):

clason:
Guess no objections :)

doc/nvim-treesitter.txt
```diff
@@ -723,6 +748,14 @@ Type (and class) definitions and annotations.
 `TSTypeBuiltin`
 Built-in types: `i32` in Rust.
 
+							  *hl-TSTypeQualifier*
+`TSTypeQualifier`
+Qualifiers on types, e.g. `const` or `volatile` in C or `mut` in Rust.
```
theHamsta, line 753: actually `@const` is cStorageClass. Where is the difference to storageclass?

gpanders: storageclass refers to how a value is actually stored at the machine level. This is not relevant in many languages, but can be a critical distinction in systems languages. Examples are `static` or `register` in C, `comptime` or `linksection` in Zig, etc. `const` should not be considered storage class because it has no bearing on where a variable is stored, it only modifies the type.

theHamsta: But I saw `const` being used in `synatx/c.vim` as `cStorageClass`. I couldn't find the other one used in the syntax file.

PR #2679, opened 2022-03-26
feat: add vimdoc parser and queries

clason:
Using @bfredl's shiny new `conceal` support on master. The captures are open for bike shedding.

I've add injection support for Lua in codeblocks, which (of course) tries to highlight all code as Lua, even if it's Vimscript (or just plain text). I don't think it's possible to be any smarter about it; it's up for debate whether this is an improvement over no highlighting at all. (I think... maybe?) 

I'm marking this as experimental since there seem to be some issues with the scanner (things like `"*"` and `:lua` after a backtick). @vigoux

### DIFF
--- a/queries/help/highlights.scm
+++ b/queries/help/highlights.scm
@@ -1,0 +1,14 @@
+(headline) @text.title
+(column_heading) @text.title
+(tag
+   "*" @conceal (#set! conceal "")
+   name: (_) @label)
+(option
+   "'" @conceal (#set! conceal "")
+   name: (_) @text.literal)
+(hotlink
+   "|" @conceal (#set! conceal "")
+   destination: (_) @text.uri)
+(backtick
+   "`" @conceal (#set! conceal "")
+   content: (_) @string)
--- a/lockfile.json
+++ b/lockfile.json
@@ -299,6 +299,9 @@
   "vim": {
     "revision": "c9d70082af14988842eb071c6c0b07e8d1d993ac"
   },
+  "help": {
+    "revision": "5e4851189b3d005275263cfc7f5a4a6ea2dbdf9b"
+  },
   "vue": {
     "revision": "91fe2754796cd8fba5f229505a23fa08f3546c06"
   },
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -827,6 +827,16 @@ list.vim = {
   maintainers = { "@vigoux" },
 }
 
+list.help = {
+  install_info = {
+    url = "https://github.com/vigoux/tree-sitter-vimdoc",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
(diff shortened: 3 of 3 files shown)


theHamsta (approved):

clason:
Do take a look at the captures, though -- I'm sure they can be improved. 

(Also, what's up with that Windows runner?)

queries/help/highlights.scm
```diff
@@ -0,0 +1,14 @@
+(tag
+   "*" @_conceal (#set! conceal "")
+   name: (_) @tag)
```
theHamsta, line 3: I think `@tag` means XML tag for our plugin can you check please?

clason: I wouldn't know, that's why I'm asking :)

clason: Suggestions very welcome! (as in, just tell me what to use, I have almost no personal opinion on this.)

clason: (`@symbol` is no good, though -- it links to `Identifier`, which for programming languages is usually the "default" text color)

clason: How about `@keyword`?

clason: Yeah, `@label` could work, as these are jump targets, just like `goto` labels.

clason:
Regarding injections: Is there _any_ type of automatic syntax detection possible with tree-sitter?

dundargoc:
Closes https://github.com/neovim/neovim/issues/17855

queries/help/injections.scm
```diff
@@ -0,0 +1 @@
+(code_block) @lua
```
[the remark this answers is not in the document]
clason: Yes, that's the main question on this PR.

I used Lua because I knew @dundargoc would primarily look at that ;)

Also, the audience for nvim-treesitter is probably more interested in Lua than Vimscript, and I wanted to see if mis-highlighting the latter (or Python, or pseudocode, or...) would be worse than no highlighting at all.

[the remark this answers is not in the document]
vigoux: I think that mishighlighting is worst that now highlithing bacause it is misleading.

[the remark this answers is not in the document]
theHamsta: we could also experiment with some kind of autodetect. Maybe when in doubt and first parser yields errors maybe the next parser can parse sucessfully.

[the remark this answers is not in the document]
clason: I'd start with some heuristics like vim uses for filetype disambiguation (look at the first line(s), see if there's something like `local` or `let`)?

clason:
Language injections could be generalized if https://github.com/nvim-treesitter/nvim-treesitter/pull/1240/files#diff-81f58ecd0f4978b575136c86619adc8dbc9a0c9e87f87702ddf22615d0ac6c4bR93-R102 was done.

PR #2681, opened 2022-03-26
Add elvish

ckafi:
Parser for the language/shell [Elvish](https://elv.sh/).
Includes `highlight` queries

### DIFF
--- a/after/ftplugin/elvish.vim
+++ b/after/ftplugin/elvish.vim
@@ -1,1 +1,0 @@
-setlocal commentstring=#\ %s
--- a/after/ftplugin/elvish.vim
+++ b/after/ftplugin/elvish.vim
@@ -1,0 +1,1 @@
+setlocal commentstring=#\ %s
--- a/queries/elvish/highlights.scm
+++ b/queries/elvish/highlights.scm
@@ -1,0 +1,79 @@
+(comment) @comment
+
+(if "if" @conditional)
+(if (elif "elif" @conditional))
+(if (else "else" @conditional))
+
+(while "while" @repeat)
+(while (else "else" @repeat))
+(for "for" @repeat)
+(for (else "else" @repeat))
+
+(try "try" @exception)
+(try (catch "catch" @exception))
+(try (else "else" @exception))
+(try (finally "finally" @exception))
+
+(import "use" @include)
+(import (bareword) @string.special)
+
+(wildcard ["*" "**" "?"] @string.special)
+
+(command argument: (bareword) @parameter)
+(command head: (identifier) @function)
+((command head: (identifier) @keyword.return)
+ (#eq? @keyword.return "return"))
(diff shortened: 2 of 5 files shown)


kyazdani42 (approved):
@theHamsta or @clason could you have a look at this please ?

clason:
Please remove the ftdetect and ftplugin changes; nvim-treesitter only handles parser and query installation. All changes to legacy syntax files _must_ go through https://github.com/vim/vim.

after/ftplugin/elvish.vim
```diff
@@ -0,0 +1 @@
+setlocal commentstring=#\ %s
```
theHamsta, line 1: We want to avoid ftplugins. Please try to create a PR at vim/vim ti support this natively

queries/elvish/highlights.scm
```diff
@@ -0,0 +1,79 @@
+(comment) @comment
+
+(if "if" @conditional)
+(if (elif "elif" @conditional))
+(if (else "else" @conditional))
```
theHamsta, line 5: Why not just `["if" "elif" "else"]`?

theHamsta: Also for all the other keiwords

queries/elvish/highlights.scm
```diff
@@ -0,0 +1,79 @@
...
+(variable_assignment
+  "set" @keyword
+  (lhs (identifier) @variable))
+
+(temporary_assignment
+  "tmp" @keyword
+  (lhs (identifier) @variable))
+
+(variable_deletion
+  "del" @keyword
```
theHamsta, line 59: Maybe `@keyword.operator`

ckafi:
@clason Can do, but you might want to update your README:
> If Neovim does not detect your language's filetype by default, you can add a short Vimscript file to nvim-treesitter's ftdetect runtime directory.

queries/elvish/highlights.scm
ckafi: My understanding is that `@keyword.operator` is for keywords used in expressions, like `and` / `or`. The `del` keyword begins a statement.

clason:
Well, that wasn't really meant to apply for _Pull Requests_, just for local config. But I replaced it nevertheless.

queries/elvish/highlights.scm
ckafi: You're right, I misunderstood how anonymous nodes work. I'm going to update this, but `else` should be in nested queries, because `if`/`while`/`for`/`try` can all have an `else` branch.

ckafi:
Understood. Also your `ftdetect` and `after/ftplugin` contain a lot of files that do exactly the same thing for different languages.

clason:
Yeah, we're trying to get rid of them.

theHamsta (approved):

PR #2723, opened 2022-04-02
Add highlighting query for Scss `@for` loops

omertuc:
Add highlighting query for Scss `@for` loops

The Scss language only had highlighting queries for `@each` loops, this
commit adds support for `@for` loops as well.

Also moved all loop-related keywords (including newly added ones) to
`@repeat` rather than `@keyword`

Before:
image

After:
image

### DIFF
--- a/queries/scss/highlights.scm
+++ b/queries/scss/highlights.scm
@@ -3,10 +3,17 @@
 [
   "@mixin"
   "@media"
-  "@while"
-  "@each"
   "@include"
 ] @keyword
+
+[
+  "@while"
+  "@each"
+  "@for"
+  "from"
+  "through"
+  "in"
+] @repeat
 
 (single_line_comment) @comment
 (function_name) @function
@@ -28,6 +35,9 @@
 (each_statement (key) @parameter)
 (each_statement (value) @parameter)
 (each_statement (variable_value) @parameter)
+
+(for_statement (variable) @parameter)
+(for_statement (_ (variable_value) @parameter))
 
 (argument) @parameter
 (arguments (variable_value) @parameter)


queries/scss/highlights.scm
```diff
@@ -6,6 +6,10 @@
   "@while"
   "@each"
   "@include"
+  "@for"
+  "from"
+  "through"
+  "in"
 ] @keyword
```
theHamsta, line 13: we could take this chance to make all loop related keywords `@repeat`

queries/scss/highlights.scm
```diff
@@ -29,6 +33,9 @@
 (each_statement (value) @parameter)
 (each_statement (variable_value) @parameter)
 
+(for_statement (variable) @parameter)
+(for_statement (_ (variable_value) @parameter))
+
 (argument) @parameter
 (arguments (variable_value) @parameter)
```
theHamsta, line 40: not sure why parameter is used for loops and arguments in this file. might have it's reasons

queries/scss/highlights.scm
omertuc: Done

queries/scss/highlights.scm
omertuc: Not familiar enough with the conventions to have any opinion about this. Anything you want me to change?

theHamsta (approved):

queries/scss/highlights.scm
theHamsta: Usually `@parameter` only applies to the parameters in function declarations. Not sure how they are used here in scss. Overall your PR looks good.

kyazdani42 (approved):
i think we can merge this @theHamsta

PR #2730, opened 2022-04-04
Fix Java indentation for method chaining

dwoznicki:
Solves an issue where Java indent was dedented one too far while method chaining.

### Issue

```java
public class Testo {
    void foo() {
        new StringBuilder()
            .append("a") // add newline here
        |<-- cursor ends up here
    }
}
```

### Expected behavior

```java
public class Testo {
    void foo() {
        new StringBuilder()
            .append("a") // add newline here
            |<-- cursor should be here
    }
}
```

### DIFF
--- a/queries/java/indents.scm
+++ b/queries/java/indents.scm
@@ -11,6 +11,8 @@
   (formal_parameters)
 ] @indent
 
+(expression_statement (method_invocation) @indent)
+
 [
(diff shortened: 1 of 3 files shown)


theHamsta (approved):

theHamsta:
This might be an issue for more languages

PR #2738, opened 2022-04-04
Highlight Python exception keywords

fratajczak:
Python keywords are used as the example in `:h TSException`:
```
Exception related keywords: try, except, finally in Python.
```
But they are actually only highlighted as `TSKeyword` as of now, not `TSException`. 

This PR adds `try`, `except`, `finally` and `raise` to the python exception highlights, as well as `else` when used in a try clause, i.e
```python
try:
    print(123)
except:
    raise(IOError)
else:
    print(456)
finally:
    cleanup()
```


queries/python/highlights.scm
```diff
@@ -237,6 +237,17 @@
 
 ["for" "while" "break" "continue"] @repeat
 
+[
+  "try"
+  "except"
+  "raise"
+  "finally"
+] @exception
```
theHamsta, line 245: can you remove them  from `@keyword`?

fratajczak: [4f8d071](https://github.com/nvim-treesitter/nvim-treesitter/pull/2738/commits/4f8d07182a982e06fcef900cf8e1e78412c8507a) sure

theHamsta (approved):

theHamsta:
Thank you!

PR #2749, opened 2022-04-07
feat: add astro parser

### LINKED ISSUE #1763: Support for Astro components in `.astro` files
First off, I just wanted to say a *HUGE* thank you to all of the maintainers of this project. Adding treesitter into my Neovim setup has elevated my daily development experience *so* much. I very much appreciate all of the hard work that has gone into this plugin and ecosystem to bring Treesitter support to Neovim!

I'm opening this issue to try and get some feedback/guidance on how to go about adding language support for [Astro components in `.astro` files](https://docs.astro.build/core-concepts/astro-components). Astro components really are basically just HTML files other than two main differences:
1. Like `jsx`, in Astro components you can use curly braces to inject Javascript into your component. The behavior as far as syntax highlighting goes is identical to `jsx`.
2. At the top of each Astro component, there will be YAML-styled frontmatter fences that define an area where any Javascript can be injected into the component. Essentially, instead of using script tags like in normal HTML:
```javascript
<script>
// javascript goes here
// in HTML normally
</script>
```
Astro frontmatter fences for writing Javascript in the component would look like:
```javascript
---
// javascript goes here
// in Astro components
---

// all HTML comes down here below
<html lang="en-US">
// more html
</html>
```

After doing some preliminary reading on Treesitter, parsers, and queries, my impression is that a custom parser for Astro is likely unnecessary, as the file is essentially just Javascript if between the `---` frontmatter fences, and `jsx` if below the set of `---` frontmatter fences. I'm just a bit confused about how to get started on supporting this. I'd really appreciate if if I could have someone point me in the direction of what I might try, any docs I should read for more info, etc. Unfortunately at the moment, Astro only officially supports custom highlighting for VS Code (as with all web dev stuff 😢), even though I was able to write a custom config to get their VS Code LSP extension working nicely with `nvim-lspconfig`. 

Thank you for your help!

virchau13:
This adds a parser for the Astro web framework. Fixes #1763.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1005,6 +1005,15 @@ list.elvish = {
   maintainers = { "@ckafi" },
 }
 
+list.astro = {
+  install_info = {
+    url = "https://github.com/virchau13/tree-sitter-astro",
+    branch = "master",
+    files = { "src/parser.c", "src/scanner.cc" },
+  },
+  maintainers = { "@virchau13" },
+}
+
 local M = {
   list = list,
   filetype_to_parsername = filetype_to_parsername,
--- a/queries/astro/folds.scm
+++ b/queries/astro/folds.scm
@@ -1,0 +1,1 @@
+; inherits: html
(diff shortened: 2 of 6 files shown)


theHamsta (approved):

virchau13:
Should be fixed (it was the trailing comma, right?)

PR #2752, opened 2022-04-07
Add scheme support

6cdh:
I added Scheme language support.

The query files can also get improved highlights automatically.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -44,6 +44,7 @@ local list = setmetatable({}, {
 list.scheme = {
   install_info = {
     url = "https://github.com/6cdh/tree-sitter-scheme",
+    branch = "main",
     files = { "src/parser.c" },
   },
   maintainers = { "@6cdh" },
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -41,6 +41,14 @@ local list = setmetatable({}, {
   end,
 })
 
+list.scheme = {
+  install_info = {
+    url = "https://github.com/6cdh/tree-sitter-scheme",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@6cdh" },
+}
+
 list.javascript = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-javascript",
--- a/queries/scheme/highlights.scm
+++ b/queries/scheme/highlights.scm
@@ -6,7 +6,7 @@
 (character) @character
 (boolean) @boolean
 (string) @string
-[(comment)] @comment
+(comment) @comment
 
 ;; highlight for datum comment
 ;; copied from ../clojure/highlights.scm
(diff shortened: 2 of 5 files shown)


queries/scheme/highlights.scm
```diff
@@ -0,0 +1,181 @@
...
+;; special forms
+
+(list
+ "["
+ (symbol)+ @variable
+ "]")
+
+(list
+ .
+ (symbol) @f
```
theHamsta, line 41: use `@_f` to make CI happy

queries/scheme/highlights.scm
```diff
@@ -0,0 +1,181 @@
...
+  "number->string" "string->number"
+  ;; boolean
+  "boolean?" "not" "boolean=?"
+  ;; pair
+  "pair?" "cons" 
+  "car" "cdr" 
+  "caar" "cadr" "cdar" "cddr" 
+  "caaar" "caadr" "cadar" "caddr" "cdaar" "cdadr" "cddar" "cdddr"
+  "caaaar" "caaadr" "caadar" "caaddr" "cadaar" "cadadr" "caddar" "cadddr"
+  "cdaaar" "cdaadr" "cdadar" "cdaddr" "cddaar" "cddadr" "cdddar" "cddddr"
```
theHamsta, line 134: This is why I love lisp 😄

lua/nvim-treesitter/parsers.lua
```diff
@@ -41,6 +41,14 @@ local list = setmetatable({}, {
   end,
 })
 
+list.scheme = {
+  install_info = {
+    url = "https://github.com/6cdh/tree-sitter-scheme",
+    files = { "src/parser.c" },
```
theHamsta, line 47: add `branch = "main",` here

queries/scheme/highlights.scm
```diff
@@ -0,0 +1,181 @@
...
+
+(number) @number
+(character) @character
+(boolean) @boolean
+(string) @string
+(comment) @comment
+
+;; highlight for datum comment
+;; copied from ../clojure/highlights.scm
+([(comment) (directive)] @comment
```
theHamsta, line 13: is the priority really necessary? For clojure it was only needed because it had s-expressions that could be comments and the parser would parse data within that special comment-sexpr.

queries/scheme/highlights.scm
```diff
@@ -0,0 +1,181 @@
...
+;; operators
+
+((symbol) @operator
+ (#any-of? @operator
+  "+" "-" "*" "/" "=" "<=" ">=" "<" ">"))
+
+;; keyword
+
+((symbol) @keyword
+ (#any-of? @keyword
```
theHamsta, line 66: I think the other lisps use `@function.macro` for those but keyword would also not be wrong

theHamsta (approved):

queries/scheme/highlights.scm
6cdh: Yes. In Scheme R6RS, there is datum comment `#;` which comment the next datum:

```scheme
(list #;(1 2) 2)
; => (2)
```

where datum means a subset of Scheme object: number, string, list, vector, etc.
I have to add priority to make the whole datum comment highlight as normal comment.
I've tried putting the rule at the end of the file, but it doesn't work.

queries/scheme/highlights.scm
6cdh: They are called syntactic keywords.
That means when we see it, we will know how things in this form are evaluated that different with the function arguments evaluation.
These forms may be special forms (can't implement in macro), or can be implemented in macro. But I think keyword is a appropriate name.

PR #2759, opened 2022-04-08
Add more filetypes for treesitter-markdown

mehalter:
There are several markdown filetypes that are common that all support the same syntax. This adds `md`, `rmd`, and `pandoc` to all be matched to the `markdown` treesitter parser.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -15,6 +15,8 @@ local filetype_to_parsername = {
   cls = "latex",
   sty = "latex",
   OpenFOAM = "foam",
+  pandoc = "markdown",
+  rmd = "markdown",
 }
 
 local list = setmetatable({}, {


clason (approved):

lua/nvim-treesitter/parsers.lua
```diff
@@ -15,6 +15,9 @@ local filetype_to_parsername = {
   cls = "latex",
   sty = "latex",
   OpenFOAM = "foam",
+  pandoc = "markdown",
+  md = "markdown",
```
clason, line 19: That doesn't seem right: `md` is not a filetype for (Neo)Vim -- files with this *extension* have the *filetype* `markdown`.

And can you point to where the `pandoc` filetype is used in the wild?

mehalter: Ah thanks for  pointing the neovim `md -> markdown` conversion out. I have come from long time vim usage and this is not necessarily true. The `pandoc` filetype comes into play if the user is using the [`vim-pandoc`](https://github.com/vim-pandoc/vim-pandoc) plugin which requires sets the filetype to `pandoc` to use some of their specific features such as completion/mappings/TOC functionality/etc.

Let me know what you think of the `pandoc` filetype, but I will go ahead and remove the `md` one though!

clason: > Ah thanks for pointing the neovim md -> markdown conversion out.

No, that's in Vim, too -- we take the filetype stuff straight from upstream.

> The pandoc filetype comes into play if the user is using the [vim-pandoc](https://github.com/vim-pandoc/vim-pandoc) plugin which requires sets the filetype to pandoc to use some of their specific features such as completion/mappings/TOC functionality/etc.

That's fine with me, then.

mehalter: Cool, thanks. My commit is updated to just include the `rmd` and `pandoc` addition.

CaeChao: It seems to me that this will overwrite  `autocmd FileType vimwiki set syntax=pandoc` with `set syntax=on` on a Vimwiki markdown file, which causes the pandoc syntax disabled on Vimwiki.

clason: You should use `vim.filetype.add()` to overwrite the builtin filetype detection.

PR #2768, opened 2022-04-10
Add c injection for cgo comments

kraftwerk28:
Adds C highlighting for [cgo](https://pkg.go.dev/cmd/cgo) ([example of code](https://github.com/kraftwerk28/gost/blob/master/blocks/rxkbcommon/rxkbcommon.go#L3-L34)). Though it only works for multiline comment `/* */`, because single line comment doesn't "merge" into a single TS node.


theHamsta (approved):

kevinhwang91:
Bad perf.

PR #2771, opened 2022-04-11
Add WGSL parser

szebniok:
I've added support for my [WebGPU Shading Language](https://www.w3.org/TR/WGSL/) parser.
I've created the `highlights` and `folds` queries.

### DIFF
--- a/ftdetect/wgsl.vim
+++ b/ftdetect/wgsl.vim
@@ -1,0 +1,1 @@
+au BufRead,BufNewFile *.wgsl set filetype=wgsl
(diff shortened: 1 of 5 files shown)


ftdetect/wgsl.vim
```diff
@@ -0,0 +1 @@
+au BufRead,BufNewFile *.wgsl set filetype=wgsl
```
clason, line 1: Please remove that; nvim-treesitter no longer acts as a filetype plugin. You should make a PR adding this filetype to Vim, which will then be ported to Neovim. This way, other people will profit from this, too.

theHamsta (approved):
Looks good to me! Thank you for contributing!

clason:
@theHamsta Can we squash merge such PRs? It's good to keep commits atomic to facilitate bisecting and make the commit history easier to grok.

PR #2791, opened 2022-04-15
julia: update queries to parser changes

clason:


queries/julia/highlights.scm
```diff
@@ -5,13 +5,13 @@
 ((identifier) @constant
   (#match? @constant "^[A-Z][A-Z_]{2}[A-Z_]*$"))
 
-[
-  (triple_string)
-  (string)
-] @string
-(command_string) @string.special
+(character_literal) @string
```
theHamsta, line 8: I think characters have a separate highlight

clason: Just tell me what to use and I'll change it.

theHamsta: I think it's just `@character`. Can't check on mobile

theHamsta (approved):

PR #2800, opened 2022-04-15
chore: remove compat module for vim.ui

theHamsta:


clason (approved):

clason:
Sorry, I missed that. Will hold off on merging as this is technically breaking (meaning should have a `chore!: ...` commit message).

(I'd give it a few days for distros to catch up, then start with a 0.7 bump to README (and check?), then merge these chore PRs.)

PR #2801, opened 2022-04-15
chore: deprecate ts_utils.get_node_text

theHamsta:
Could also just silently do `M.get_node_text = vim.treesitter.query.get_node_text`. Then, downstream users don't have to switch for each function that gets moved from ts_utils to core

### DIFF
--- a/lua/nvim-treesitter/ts_utils.lua
+++ b/lua/nvim-treesitter/ts_utils.lua
@@ -2,35 +2,21 @@ local api = vim.api
 
 local parsers = require "nvim-treesitter.parsers"
 local utils = require "nvim-treesitter.utils"
+local ts_query = vim.treesitter.query
 
 local M = {}
 
 --- Gets the actual text content of a node
+-- @deprecated Use vim.treesitter.query.get_node_text
 -- @param node the node to get the text from
 -- @param bufnr the buffer containing the node
 -- @return list of lines of text of the node
 function M.get_node_text(node, bufnr)
-  local bufnr = bufnr or api.nvim_get_current_buf()
-  if not node then
-    return {}
-  end
-
-  -- We have to remember that end_col is end-exclusive
-  local start_row, start_col, end_row, end_col = M.get_node_range(node)
-
-  if start_row ~= end_row then
-    local lines = api.nvim_buf_get_lines(bufnr, start_row, end_row + 1, false)
-    lines[1] = string.sub(lines[1], start_col + 1)
-    -- end_row might be just after the last line. In this case the last line is not truncated.
-    if #lines == end_row - start_row + 1 then
-      lines[#lines] = string.sub(lines[#lines], 1, end_col)
-    end
-    return lines
-  else
-    local line = api.nvim_buf_get_lines(bufnr, start_row, start_row + 1, false)[1]
-    -- If line is nil then the line is empty
-    return line and { string.sub(line, start_col + 1, end_col) } or {}
-  end
+  vim.notify_once(
+    "nvim-treesitter.ts_utils.get_node_text is deprecated: use vim.treesitter.query.get_node_text",
+    vim.log.levels.WARN
+  )
+  vim.treesitter.query.get_node_text(node, bufnr)
 end
 
 --- Determines whether a node is the parent of another
@@ -326,8 +312,8 @@ function M.swap_nodes(node_or_range1, node_or_range2, bufnr, cursor_to_second)
   local range1 = M.node_to_lsp_range(node_or_range1)
   local range2 = M.node_to_lsp_range(node_or_range2)
 
-  local text1 = M.get_node_text(node_or_range1)
-  local text2 = M.get_node_text(node_or_range2)
+  local text1 = ts_query.get_node_text(node_or_range1)
+  local text2 = ts_query.get_node_text(node_or_range2)
 
   local edit1 = { range = range1, newText = table.concat(text2, "\n") }
   local edit2 = { range = range2, newText = table.concat(text1, "\n") }
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -341,11 +341,6 @@ get_node_at_cursor(winnr)~
 `winnr` will be 0 if nil.
 Returns the node under the cursor.
 
-						      *ts_utils.get_node_text*
-get_node_text(node, bufnr)~
-
-Returns the text content of a `node`.
-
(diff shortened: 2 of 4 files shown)


clason:
I think I wouldn't bother with this now as 0.7 is just released and is now the minimal supported version.

As things get upstreamed over the 0.8 dev cycle, this is the right approach though. (Going forward, we should be much more explicit about what we consider exposed API and private internal functions, and not worry about messing with the later. ("The beatings will continue until morale improves."))

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -2,35 +2,21 @@ local api = vim.api
...
-      lines[#lines] = string.sub(lines[#lines], 1, end_col)
-    end
-    return lines
-  else
-    local line = api.nvim_buf_get_lines(bufnr, start_row, start_row + 1, false)[1]
-    -- If line is nil then the line is empty
-    return line and { string.sub(line, start_col + 1, end_col) } or {}
-  end
+  require("nvim-treesitter.utils").notify(
+    "nvim-treesitter.utils is deprecated: use vim.treesitter.query.get_node_text",
```
clason, line 16:
```suggestion
    "nvim-treesitter.ts_utils is deprecated: use vim.treesitter.query.get_node_text",
```

I don't think there was any decision of upstreaming  `ts_utils` _in its entirety_ so far?

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -2,35 +2,21 @@ local api = vim.api
...
-    if #lines == end_row - start_row + 1 then
-      lines[#lines] = string.sub(lines[#lines], 1, end_col)
-    end
-    return lines
-  else
-    local line = api.nvim_buf_get_lines(bufnr, start_row, start_row + 1, false)[1]
-    -- If line is nil then the line is empty
-    return line and { string.sub(line, start_col + 1, end_col) } or {}
-  end
+  require("nvim-treesitter.utils").notify(
```
clason, line 15: Use  `vim.notify`, since we're deprecating.

theHamsta: this wrapper is just adding `nvim-treesitter` as the source for vim.notify

lua/nvim-treesitter/ts_utils.lua
theHamsta: no, just this function

lua/nvim-treesitter/ts_utils.lua
clason: And I'm saying "no wrapper needed" :)

clason: Sorry, should have been clearer: We want to use `notify_once` for deprecation messages to avoid spamming them continuously.

lua/nvim-treesitter/ts_utils.lua
clason: Yes, so why write "nvim-treesitter.ts_utils`" (implying the whole module) "is deprecated"?

lua/nvim-treesitter/ts_utils.lua
theHamsta: Is it even safe to display messages in contexts where `get_node_text` is used (e.g. highlighting queries, at least not in our predicates)

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -2,35 +2,21 @@ local api = vim.api
...
-      lines[#lines] = string.sub(lines[#lines], 1, end_col)
-    end
-    return lines
-  else
-    local line = api.nvim_buf_get_lines(bufnr, start_row, start_row + 1, false)[1]
-    -- If line is nil then the line is empty
-    return line and { string.sub(line, start_col + 1, end_col) } or {}
-  end
+  vim.notify_once(
+    "nvim-treesitter.utils.get_node_text is deprecated: use vim.treesitter.query.get_node_text",
```
clason, line 16:
```suggestion
    "nvim-treesitter.ts_utils.get_node_text is deprecated: use vim.treesitter.query.get_node_text",
```

lua/nvim-treesitter/ts_utils.lua
clason: Not sure; `notify_once` is meant for exactly that purpose (avoid spamming in hot loops), but there may still be a performance hit.

Since right now were ripping the bandaids off anyway, I'd just hard-deprecate it and not worry about it; there will be a bunch of breaking commits at the same time, anyway.

clason:
While you're at it, could also remove https://github.com/nvim-treesitter/nvim-treesitter/blob/4542402e34568eb607059e3ff6a3594aaca850fd/lua/nvim-treesitter.lua#L1-L3

clason:
Alternatively, if you want to be nice to plugins (which I would vote not to be ;)), you could turn this into a wrapper like
```lua
M.get_node_text = ts_query.get_node_text or function(node, bufnr)
   vim.notify_once(
     "nvim-treesitter.ts_utils.get_node_text is deprecated: use vim.treesitter.query.get_node_text",
     vim.log.levels.WARN
   )
   vim.treesitter.query.get_node_text(node, bufnr)
end
```
This would put the notify in the hot path only for plugins that should be updated anyway.

clason (approved):

clason:
LGTM!

clason:
@theHamsta It looks like the upstream function is actually _not_ the same function, so this has broken a number of upstream plugins.

(Sorry, I should have verified this myself!)

bennypowers:
ok so what are plugin authors supposed to do now?

clason:
as it says on the tin: use `vim.treesitter.query.get_node_text()`

bennypowers:
which is not the same function, right?

clason:
Is it not?

bennypowers:
it's not the same function

what i'm **SURE YOU MEAN TO SAY** is that "while deprecating, we made an undocumented breaking change to the plugin signature, requiring you to pass in the bufnr or `0`, have fun updating each single call site"

clason:
What _I'm_ sure you mean to say is "I think you made a mistake in your wrapper, it should actually be ... Here, let me make a PR with the fix."

max397574:
the difference is that the new one just returns a string afaict so you have to use string.split(..., "\n")

wookayin:
It's worth mentioning that part of this PR (c3848e713a8272e524a7eabe9eb0897cf2d6932e) were later reverted by #2833.

PR #2806, opened 2022-04-16
chore: bump minimal Nvim version to 0.7 and check

clason:
Checks minimal version in `:checkhealth nvim-treesitter`

Also recommend nightlies

First in a series of PRs to upgrade to Nvim 0.7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ For more detailed information on setting these up, see ["Advanced setup"](#advan
 
 ## Requirements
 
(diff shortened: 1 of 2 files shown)


clason:
OK, now we run into the situation I have always dreaded: stylua and luacheck fighting each other...

clason:
I'll just drop the `ok` -- people don't need to be reminded _here_  what their Neovim version is.

theHamsta (approved):

clason:
Merging now as the change is not breaking. I'll mention it in the following HEAD issue anyway.

PR #2810, opened 2022-04-16
fix(go): reduce indent after },)

akinsho:
This PR relates to #2369 and hopefully fixes it.

This fix is basically a direct implementation of the comment that @theHamsta made in https://github.com/nvim-treesitter/nvim-treesitter/issues/2369#issuecomment-1034229102.

Locally this seems to mean that when indent after a `}` the indent is reset, so for example
```go
func example() {
} <CURSOR>
	-><CURSOR>
```

becomes

```go
func example() {
}<CURSOR>
<CURSOR>
```

I added in the `}` as I was seeing this behaviour there as well as in the example above.

### DIFF
--- a/queries/go/indents.scm
+++ b/queries/go/indents.scm
@@ -18,6 +18,11 @@
   "}"
 ] @branch
 
+[
+ "}"
+ ")"
+] @indent_end
+
 (parameter_list ")" @branch)
 
 (comment) @ignore
--- a/tests/indent/go/issue-2369-newline.go
+++ b/tests/indent/go/issue-2369-newline.go
@@ -1,0 +1,8 @@
+// https://github.com/nvim-treesitter/nvim-treesitter/issues/2369
+package main
+
+import "fmt"
+
+func new_line() {
+	fmt.Println("Hello, World!")
+}
--- a/tests/indent/go_spec.lua
+++ b/tests/indent/go_spec.lua
@@ -18,4 +18,8 @@ describe("indent Go:", function()
   describe("new line:", function()
     run:new_line("issue-2369.go", { on_line = 13, text = "// some comment", indent = 1 })
   end)
+
+  describe("new line after )/}:", function()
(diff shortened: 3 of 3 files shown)


theHamsta (approved):

theHamsta:
Is this already tested in our indent tests?

akinsho:
@theHamsta I'm not familiar with how things are being tested for indents, I can see that there is a test under indents for issue #2369, but it seems to be testing a different issue with indents than this one is fixing 🤷🏿‍♂️

theHamsta:
You would just add your example file into the go folder and add an entry here to start a new line at line 5.

https://github.com/theHamsta/nvim-treesitter/blob/789961cadccde4ed081c12c17da2add43b10d447/tests/indent/go_spec.lua#L19

theHamsta:
Thanks!

akinsho:
@theHamsta while we are at it 😅 I notice that when typing something like a top level multiple var declaration the indentation isn't reduced for the closing paren e.g.

```go
var (
  thing = 1
  thingTwo = 2
  ) // <-- This is incorrect

// It should be
var (
  thing = 1
  thingTwo = 2
) // <-- This paren should be at 0 instead of indented
```

I have no idea how to fix it though, since each declaration after the opening paren should be indented which happens correctly, just that if the character type on the line after `thingTwo` is a `)` then the indent should be re-adjusted. Is this possible with treesitter indent?

PS: I can open a new issue for this if it's something that can be tackled?

theHamsta:
Pretty strange, especially since it does correctly within a `block`. I guess one would have to debug indent.lua to see what's going on
```
package main

var (
	thing    = 1
	thingTwo = 2
) // <-- This paren should be at 0 instead of indented

var (
	thing    = 1
	thingTwo = 2
)

func main() {
	// It should be
	var (
		thing    = 1
		thingTwo = 2
	)
}
```

akinsho:
Ah, right that's a bit more than I have time to dig into atm, but as you say, weird it works in block scope but not otherwise.

PR #2813, opened 2022-04-16
refactor!: update to Neovim 0.7 APIs

clason:
* set highlight groups via nvim_set_hl
* define autocommands via nvim_create_autocmd
* port plugin/nvim-treesitter.vim to Lua
* port healthcheck to Lua

Missing:

* user commands
* keymappings for `incremental_selection`

as these currently rely on complex string munging, which isn't possible in Lua. The mappings require a different (and in my view better) approach for specifying mappings in module setup, see https://github.com/nvim-treesitter/playground/pull/75#issuecomment-1100647721. The user commands require someone familiar with the current code to extricate and convert. (Personally, I would suggest just defining them explicitly one-by-one in `plugin/*.lua` without the complicated generator.)


clason:
Not quite sure why the indentation tests fail...

theHamsta:
I'm not sure whether we should use the opportunity to revert https://github.com/nvim-treesitter/nvim-treesitter/pull/2674 before merging this PR (while it's still possible w/o conflicts). Our repo is in a inconsistent state since. We could however leave the those highlights in the floating state in which they are currently are and try to add them one by one.

clason:
That is a separate issue that needs a broader discussion, including core. I don't want to tie this PR to that in any way.

kyazdani42:
looks good to me globally, not sure why the indentation tests fail though.

clason:
Me neither 😞 

It could be a bug (or subtle difference) the new autocommand API. But not sure why that would happen on CI but not locally.

clason:
Not related to these changes, it seems, as I get exactly the same test failures locally on master.

clason:
Reverting the changes to the module autocommands still triggers the failures, so this has to be due the bump to Neovim 0.7 itself (not sure why this didn't affect other PRs, then).

Would appreciate if someone familiar with indentation could look into this.

milanglacier:
Could you please add descriptions (i.e `desc` key) to those autocmd and keymaps created by nvim_create_autocmd or vim.keymap.set?

Without `desc` key, `:Telescope autocommands` or `:Telescope keymaps`, or `:nmap <CR>`, or `:au filetype` (suppose CR is the map of incremental selection trigger key) will only show a `<Lua function>` as the rhs which is not informative at all. (It is hard to track the usage of keymaps/ the usage of autocmd very easily)

clason:
Still don't understand why the indent tests are failing; would appreciate it if someone more familiar with the indent code could take a look at it.

kyazdani42:
@clason i'll take a look at this this afternoon

plugin/nvim-treesitter.lua
```diff
@@ -0,0 +1,118 @@
...
+end
+vim.g.loaded_nvim_treesitter = true
+
+-- setup modules
+require("nvim-treesitter").setup()
+
+local api = vim.api
+
+-- define autocommands
+local augroup = api.nvim_create_augroup("NvimTreesitter", {})
```
kyazdani42, line 14: augroup duplicated with the one in configs, this will fail loading treesitter for buffers :)

kyazdani42:
@clason i think switching nvim-treesitter.vim to lua somehow breaks indentation. Might be caused by loading order i believe. i've tried wrapping the removed vim file in a vim.cmd in the nvim-treesitter.lua file to make sure that wasn't the lua code, and same thing happens.

plugin/nvim-treesitter.lua
kyazdani42: should we share an augroup for nvim-treesitter globally or use one per autocmd group ? I've fixed this by change `NvimTreesitter` to `NvimTreesitterConfigs` in `configs.lua`

max397574:
you could also test to put the lua code into `lua << EOF` in the vim file to make sure that's the problem

kyazdani42:
the lua code is working fine, even in a heredoc. It's changing the file from vim to lua that is somehow breaking the indents.

plugin/nvim-treesitter.lua
clason: That is a good question. I tried to 1:1 replicate the current structure (but missed a `clear = false` ;)).

But, yes, I think a global augroup would be cleaner. (I forgot that that `create_autocmd` also takes a string for the `group`, so the constant `create_augroup` isn't actually needed.)

clason:
> @clason i think switching nvim-treesitter.vim to lua somehow breaks indentation. Might be caused by loading order i believe. i've tried wrapping the removed vim file in a vim.cmd in the nvim-treesitter.lua file to make sure that wasn't the lua code, and same thing happens.

Yes, that was my suspicion as well, but I don't know _why_ that should be the case. What part of the indentation module relies on loading order?

(I should add that I tested this locally and couldn't reproduce it manually, so this seems to be specific to the indent _tests_?)

plugin/nvim-treesitter.lua
kyazdani42: i think this would be a good solution overall :)

kyazdani42:
thing is, running the tests, the plugin file is never loaded.

clason:
Yeah...

kyazdani42:
found the culprit... `vim.cmd [[runtime! plugin/nvim-treesitter.vim]]`

max397574:
Perhaps I'm completely wrong with what I write here
but
https://github.com/nvim-treesitter/nvim-treesitter/blob/master/scripts/minimal_init.lua#L3

clason:
> found the culprit... `vim.cmd [[runtime! plugin/nvim-treesitter.vim]]` (in minimal_init.lua)

Oh, is that sourced by the tests?

kyazdani42:
seems so, changing to lua fixes the tests

clason:
Never would have thought to look there... Thanks!

I'll refactor the augroups and ping you again for a last review. Do you have any thoughts on the user commands and keymappings? I didn't touch those, although they should be changed as well. (In a followup PR, probably.)

lua/nvim-treesitter/configs.lua
```diff
@@ -150,7 +156,7 @@ local function disable_mod_conf_autocmd(mod)
   end
   -- TODO(kyazdani): detach the correct autocmd... doesn't work when using %s, cmd.
```
clason, line 157: @kyazdani42 does that comment still apply?

kyazdani42:
I think these are safe to refactor. The implementation i've "stolen" long ago from nvim-lsp but i think it has changed quite a bit since then, but we've never rewritten it.
Also, users shouldn't use our internal api to define commands, those who do are doomed to have breaking configs :smile:

lua/nvim-treesitter/configs.lua
kyazdani42: don't think so, it's quite old

clason:
> I think these are safe to refactor. The implementation i've "stolen" long ago from nvim-lsp but i think it has changed quite a bit since then, but we've never rewritten it.
> Also, users shouldn't use our internal api to define commands, those who do are doomed to have breaking configs 😄

OK, will do the commands in a followup PR. The keymappings require a more serious refactor that is above my paygrade.

kyazdani42:
Got a huge stacktrace error when opening nvim with the latest changes, seems like the augroup is not yet created when used in configs.lua

max397574:
> OK, will do the commands in a followup PR. The keymappings require a more serious refactor that is above my paygrade.

are there keymaps other than the one for incremental selection?

clason:
> Got a huge stacktrace error when opening nvim with the latest changes, seems like the augroup is not yet created when used in configs.lua

Hmm, probably depends on how you `require` nvim-treesitter?

Then I'll have to think about the best way to handle those... (The nvim-treesitter design is... byzantine, at least to me.)

Should every module have its own augroup, then?

max397574:
wouldn't it make sense for each module to have it's own augroup so if you disable one module it can clear it's autocommands separately from the other ones?

clason:
@kyazdani42 can you try the latest commit? I think that should also remove your TODO.

kyazdani42 (approved):
indeed it removes the TODO ! well done :)
works well for me.

PR #2818, opened 2022-04-17
vala: use grammar from @vala-lang

Prince781:
Use the grammar from https://github.com/vala-lang/tree-sitter-vala

This grammar seems to work better on more inputs and is maintained more frequently.

### DIFF
--- a/queries/vala/highlights.scm
+++ b/queries/vala/highlights.scm
@@ -1,297 +1,245 @@
-; Identifiers
+; highlights.scm
 
-((identifier) @constant (#match? @constant "^[A-Z][A-Z\\d_]+$"))
+; higlight comments and symbols
+(comment) @comment
+(symbol) @symbol
+(member_access_expression (_) (identifier) @symbol)
 
-(namespaced_identifier
-  left: [
-    ; Lowercased names in lhs typically are variables, while camel cased are namespaces
-    ; ((identifier) @namespace (#match? @namespace "^[A-Z]+[a-z]+$"))
-    ((identifier) @variable (#match? @variable "^[a-z]"))
-    (_)
-  ]
-  right: [
-    ; Lowercased are variables, camel cased are types
-    ; ((identifier) @parameter (#match? @parameter "^[a-z]"))
-    ((identifier) @type (#match? @type "^[A-Z]+[a-z]+$"))
-    (_)
-  ]
+; highlight constants
+(
+  (member_access_expression (identifier) @constant)
+  (#match? @constant "^[A-Z][A-Z_0-9]*$")
 )
 
-((identifier) @constructor (#match? @constructor "^[A-Z]*[a-z]+"))
+(
+  (member_access_expression (member_access_expression) @include (identifier) @constant)
+  (#match? @constant "^[A-Z][A-Z_0-9]*$")
+)
 
-; Pointers
+; highlight types and probable types
+(type (symbol (_)? @namespace (identifier) @type))
+(
+  (member_access_expression . (identifier) @type)
+  (#match? @type "^[A-Z][A-Za-z_0-9]{2,}$")
+)
 
-(address_of_identifier "&" @symbol)
-(pointer_type "*" @symbol)
-(indirection_identifier "*" @symbol)
+; highlight creation methods in object creation expressions
+(
+  (object_creation_expression (type (symbol (symbol (symbol)? @include (identifier) @type) (identifier) @constructor)))
+  (#match? @constructor "^[a-z][a-z_0-9]*$")
+)
 
-; Misc
+(unqualified_type (symbol . (identifier) @type))
+(unqualified_type (symbol (symbol) @namespace (identifier) @type))
 
-(number) @number
-
+(attribute) @attribute
+(namespace_declaration (symbol) @namespace)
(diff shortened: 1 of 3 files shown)


clason:
Be aware that changing parsers for a supported language with incompatible queries is a massively breaking change for users. This should not be done lightly.

Prince781:
@clason sure, but the current parser for Vala doesn't seem to [work](https://github.com/vala-lang/tree-sitter-vala/issues/1#issuecomment-1083987163), so I don't think this will affect users much.

Prince781:
@clason Do I need generating from grammar if I already provide `src/parser.c`?

clason:
Sorry, that issue is a bit vague to conclude that. The old vala parser is tested by and passes our CI on Neovim 0.7, so something else must be going on there.

clason:
> @clason Do I need generating from grammar if I already provide `src/parser.c`?

No.

clason:
@matbme

clason:
The highlights are using an invalid capture name; the only currently allowed captures in nvim-treesitter are those listed in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights

Prince781:
@clason sorry, should be fixed

clason:
I still will not merge this without a (sufficiently clear) explanation of why your parser is significantly and objectively better than the old one.

theHamsta:
The old one didn't really work great (couldn't parse many constructs) so I don't think we break to many users' experiences. Can't tell about the new one I guess one just needs to try it out.

clason:
Would still like to see _some_ before/after example (and would like to hear from the previous maintainer), just on principle...

(This will become less of an issue when `maintained` is removed, but while this is still around in the wild, it will affect unsuspecting users.)

theHamsta:
You could also quickly fix the bug about changing parsers by including the parser URL in the lockfile.

clason:
> You could also quickly fix the bug about changing parsers by including the parser URL in the lockfile.

That would be good in any case, but here the queries seem to be incompatible, which is the actual breaking change I am worried about.

theHamsta:
@matbme

clason:
Already pinged them yesterday. I'd give them a few more days (it is a holiday for many people, after all), but no response would be sufficient reason to switch.

matbme:
I was aware of the [duplication of effort](https://github.com/matbme/tree-sitter-vala/issues/2), but it's interesting to see this new parser in a group with other similar projects. It's true I didn't have a lot of time lately to improve the parser, so I could agree on a change.

Looking at the proposed parser, I see that some things such as some preproc statements and verbatim are still not implemented, which I could gladly help add if that's ok with @Prince781.

theHamsta:
We can still wait with the change. @Prince781, @matbme it would be good to see a complete parser with maybe combined efforts.

A good indication is that a parser can parse large repos of the language without errors https://github.com/theHamsta/tree-sitter-commonlisp/blob/c7e814975ab0d0d04333d1f32391c41180c58919/.github/workflows/ci.yml#L21-L31

clason:
I'm converting this to draft; feel free to reopen (and rebase) when you have combined efforts on whichever parser you decide on!

(By that time, we should have the infrastructure in place to handle switching parser URLs transparently to the user.)

matbme:
@clason I think It's time to transition Vala support to the new repo. I don't have the time to fix everything in the current one and @Prince781's implementation is more flexible.

clason:
@matbme If you approve this change, that's fine with me. But please squash your commits and mark the change as breaking (by using the commit message `feat(vala)!: switch to vala-lang parser`).

@theHamsta should take a look at the new queries.

I would appreciate it if you could look at other downstream plugins (especially `tree-sitter-textobjects`) and check if they have vala queries that need to be adapted (and help them to do so).

queries/vala/highlights.scm
```diff
@@ -1,297 +1,246 @@
...
+ "set"
+ "signal"
+ "struct"
+ "switch"
+ "throw"
+ "throws"
+ "try"
+ "virtual"
+ "with"
+] @keyword
```
theHamsta, line 123: I think some of those keywords fall under `@exception` or `@repeat`/`@conditional`

Prince781: Hmmm...it looks like there were some duplicate entries for some of these keywords.

queries/vala/highlights.scm
```diff
@@ -1,297 +1,245 @@
...
+(enum_declaration (symbol) @type)
+(enum_value (identifier) @constant)
+(errordomain_declaration (symbol) @type)
+(errorcode (identifier) @constant)
+(constant_declaration (identifier) @constant)
+(method_call_expression (member_access_expression (identifier) @function))
+; highlight macros
+(
+ (method_call_expression (member_access_expression (identifier) @function.macro))
+ (#match? @function.macro "^assert[A-Za-z_0-9]*$" "error" "info" "debug" "print" "warning" "warning_once")
```
theHamsta, line 54: didn't know that our `#match?` would look at more than one argument

theHamsta (approved):
This looks good. Are we ready to merge?

Also this is the first PR that makes `storageclass` and `type.qualifier` canoncial

Prince781:
Yes, I think it's ready to merge.

PR #2824, opened 2022-04-18
Use branch with-generated-files of alex-pinkus/tree-sitter-swift

d12bb:
Fixes #2823

PR #2826, opened 2022-04-18
feat(swift): switch to branch with generated files

### LINKED ISSUE #2823: Use `with-generated-files` branch of alex-pinkus/tree-sitter-swift
Currently, the parsers for Swift are generated on installation with quite an overhead (see [the original pr](https://github.com/nvim-treesitter/nvim-treesitter/pull/2265#issue-1105008932)). Now that the repo has a branch with the generated grammar included ([alex-pinkus/tree-sitter-swift#149](https://github.com/alex-pinkus/tree-sitter-swift/issues/149#issuecomment-1099849713)), this should be used instead.

clason:
rebase of #2824

closes #2823

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -282,7 +282,7 @@
     "revision": "cfa3f0ac82cd680b4efaa090a0f78035ad43a7c6"
   },
   "swift": {
-    "revision": "b219297b69837a5d55f1235236c6f77b65f8b0ab"
+    "revision": "9607603b9bd7a174b06567367ae3949c7a19f011"
   },
   "teal": {
     "revision": "fcc5f6f4d194dede4e676834ff28a506e39e17b4"
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -416,9 +416,8 @@ list.ocamllex = {
 list.swift = {
   install_info = {
     url = "https://github.com/alex-pinkus/tree-sitter-swift",
+    branch = "with-generated-files",
     files = { "src/parser.c", "src/scanner.c" },
-    requires_generate_from_grammar = true,
-    generate_requires_npm = true,
   },
 }
 


theHamsta (approved):

alex-pinkus:
Just want to warn you that this will lead to query time regressions, since I don’t specify ABI 14 when generating.

clason:
That would have been very good to know before 😞 

And the time difference is significant, as I recall... Would it be too much to ask to (also) generate with ABI 14 (on a second branch)? 😬

alex-pinkus:
Honestly, I don’t like the idea of adding yet another generated files branch. This is exactly why I don’t like checking in generated files (well, one of the reasons). It seems to me that until ABI 14 is the default, it’d be best to keep regenerating from source.

That said, I think `generate_requires_npm` can be false now, so it should be a bit cheaper.

clason:
Totally understandable. Let's hope tree-sitter will make parser generation/distribution more painless in the next version...

I'll leave it here and wait until someone complains, then likely revert (modulo the `generate_requires_npm`, which is already very helpful in itself).

alex-pinkus:
Well, someone has complained -- see https://github.com/alex-pinkus/tree-sitter-swift/issues/197. Do you think we should go back to mainline? I might have time to make that change this weekend.

clason:
> Have you explored any options around using a bona fide package manager instead?

Well, in a very real sense, `nvim-treesitter` _is_ the package manager for tree-sitter parsers... The issue is exactly that the Neovim ecosystem is very diverse and needs to support Linux, BSD, macOS (both Intel and M1), Windows (our eternal nemesis), and NixOS. The situation is not helped by the fact that the vast majority of parsers do _not_ tag or release, and that queries often depend on specific parser commits. (As nvim-treesitter is currently -- and deliberately -- the "bleeding edge" for Neovim treesitter development, this is not necessary a bad thing -- _for now_.)

Ideally (and this is what we are kinda holding our breath for), parser distribution should be tackled by upstream, either through a standard cross-platform GH action for parser developers, or a wasm target that tree-sitter can handle, or something else entirely. (The fact that Zed is macOS-only dampens that hope, though...)

> Tactically, I'd like to see neovim users benefit from the ABI 14 improvements, but I'm not going to make with-generated-files use ABI 14 because other editors haven't upgraded yet. This makes me want to switch back to mainline. I'll take a crack at that this weekend if I have a chance!

Yes, I think that is indeed the most reasonable solution. Thank you for following up on this!

PR #2828, opened 2022-04-18
fix(install): add url to lockfile

### LINKED ISSUE #2827: Include parser URL in lockfile
Currently, the lockfile only contains the commit hash of a parser. This is fine for the most part, but if we switch parsers for a language in `parsers.lua`, this will not be picked up by `:TSUpdate`, leading to a broken installation.

Instead, the lockfile should contain the full URL of the parser (repo _and_ branch), and that should be checked (and kept updated via the CI action).

theHamsta:
Fixes #2827

### DIFF
--- a/scripts/update-lockfile.sh
+++ b/scripts/update-lockfile.sh
@@ -1,20 +1,11 @@
 #!/usr/bin/env bash
 
-make_ignored() {
-  if [ -n "$1" ]
-  then
-    while read -r lang; do
-      if [ "$lang" != "$1" ]
-      then
-        printf "%s," "$lang"
-      fi
-    done < <(jq 'keys|@sh' -c lockfile.json)
-  fi
-}
+if [ -n "$1" ]
+then
+  TO_IGNORE=$(jq -r "keys | map(if . != \"$1\" then . else \"\" end) | @csv" -c lockfile.json)
+fi
 
-TO_IGNORE=$(make_ignored $1)
-
-SKIP_LOCKFILE_UPDATE_FOR_LANGS="$TO_IGNORE" nvim --headless -c "luafile ./scripts/write-lockfile.lua" -c "q"
+SKIP_LOCKFILE_UPDATE_FOR_LANGS="$TO_IGNORE,$SKIP_LOCKFILE_UPDATE_FOR_LANGS" nvim --headless -c "luafile ./scripts/write-lockfile.lua" -c "q"
 # Pretty print
 cp lockfile.json /tmp/lockfile.json
 cat /tmp/lockfile.json | jq --sort-keys > lockfile.json
--- a/.github/workflows/update-lockfile.yml
+++ b/.github/workflows/update-lockfile.yml
@@ -32,10 +32,7 @@ jobs:
         env:
           SKIP_LOCKFILE_UPDATE_FOR_LANGS: verilog,gleam,nix
         run: |
-          ./nvim.appimage --headless -c "luafile ./scripts/write-lockfile.lua" -c "q"
-          # Pretty print
-          cp lockfile.json /tmp/lockfile.json
-          cat /tmp/lockfile.json | /tmp/jq --sort-keys > lockfile.json
+          ./scripts/update-lockfile.sh
 
       - name: Commit changes
         run: |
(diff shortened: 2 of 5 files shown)


theHamsta:
could also include the `files` with the same code as now. This would allow to also install correctly when a scanner gets added or file gets renamed.

clason:
That would basically mean moving (most of) the info from the `parsers.lua` to the lockfile? I would be in favor of that. (As long as there is only a single source of truth.)

theHamsta:
Look at the code. The lua file stays the source of truth and dumps its info into the lockfile (except revision is updated). I suppose that's more convenient than editing JSON. The JSON would stay a generated artifact. This would also make @kylo252 since the JSON would then be enough for any build script to reproduce our setup.

clason:
Agreed. I prefer this way as well.

theHamsta:
After that we should merge #2289. 

We need to decide how to get the `ABI`. Either like #2289 via `parser._abi_version` or
```
      abi_version=$(gh api "repos/${repo}/contents/src" --jq '.[] | select(.name="parser.c").download_url' \
        | xargs curl -LSs \
        | grep -oP "define LANGUAGE_VERSION \K(\d+)")
```
from https://github.com/nvim-treesitter/nvim-treesitter/pull/2367

#2289 would require to install all parsers in the update-lockfile PR to do `parser._abi_version` while #2367 executes a complex bash command (gh, jq, xargs should exist on CI).

clason:
I prefer the simpler version that leverages the existing tree-sitter infrastructure.

theHamsta:
I've discovered a bug in `./scripts/update-parser.sh` which made the CI just fail because it ignored SKIP_LOCKFILE_UPDATE_FOR_LANGS. Should I add the fix in this PR

theHamsta:
@clason I'm not sure which one is simpler. I almost tend that the bash version is simpler, but the other would also be no problem when we keep the following order:
- update lockfile.json as usual (every run)
- install the new parsers (if change in lockfile.json)
- update lockfile.json this time to get the ABI

clason:
Step 3 is only needed once, and could be done manually, right?

clason:
This doesn't seem to include the branch info -- but I guess that is implicitly taken care of by the commit hash?

theHamsta:
> Step 3 is only needed once, and could be done manually, right?

Nobody would do that manually. ABI could change in every commit.

> This doesn't seem to include the branch info -- but I guess that is implicitly taken care of by the commit hash?

Branch info is implied by commit. I had the complete install_info before there, but then changed it to only what is necessary. Since the branch info is used to get the new revision and I don't want the lockfile to influence its own update.

clason:
> Nobody would do that manually. ABI could change in every commit.

Then I misunderstood

> update lockfile.json this time to get the ABI

You meant "to **set** the ABI"?

theHamsta:
@clason We could however ignore the ABI completely for now. If we require the latest Neovim, everyone would have the same runtime which is 13&14. The next ABI update of tree-sitter is likely to remove the generated files altogether. So we could ignore the problem as we  need to find a different solution anyway.



> >   update lockfile.json this time to get the ABI

> You meant "to set the ABI"?

**get** the ABI from the parser to **set** it in the lockfile. The art is to know the ABI for something that is not installed yet. We could however just always install the newest revision of all parsers, but this would imply to install all parsers on every commit.

clason:
>  If we require the latest Neovim,

We already do, so we're fine for now.

> The next ABI update of tree-sitter is likely to remove the generated files altogether. So we could ignore the problem as we need to find a different solution anyway.

Agreed. Let's wait what upstream does.

scripts/update-lockfile.sh
```diff
@@ -12,9 +12,7 @@ make_ignored() {
   fi
 }
 
-TO_IGNORE=$(make_ignored $1)
-
```
theHamsta, line 16: this is still wrong

theHamsta: I doubt that this worked at any time before... Even the original master does not work with `update-lockfile python" Will implement it in lua

theHamsta: should work now

clason:
superseded by rewrite on `main`

PR #2831, opened 2022-04-18
fix(statusline): adjust to the new API (query)

ranjithshegde:
Just a quick fix before the refactor recommended by @clason

### DIFF
--- a/lua/nvim-treesitter.lua
+++ b/lua/nvim-treesitter.lua
@@ -4,6 +4,7 @@ local info = require "nvim-treesitter.info"
 local configs = require "nvim-treesitter.configs"
 local parsers = require "nvim-treesitter.parsers"
 local ts_query = vim.treesitter.query
+local ts_utils = require "nvim-treesitter.ts_utils"
 
 -- Registers all query predicates
 require "nvim-treesitter.query_predicates"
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter.lua
```diff
@@ -48,15 +49,16 @@ function M.statusline(opts)
...
+  --   options = { indicator_size = opts }
+  -- end
+  local bufnr = options.bufnr or 0
   local indicator_size = options.indicator_size or 100
   local type_patterns = options.type_patterns or { "class", "function", "method" }
   local transform_fn = options.transform_fn or transform_line
   local separator = options.separator or " -> "
 
-  local current_node = ts_query.get_node_at_cursor()
+  local current_node = ts_utils.get_node_at_cursor()
```
clason, line 61: That function is deprecated; please use the upstream version!

clason (approved):

lua/nvim-treesitter.lua
clason, line 61: oops...

clason (approved):

lua/nvim-treesitter.lua
ranjithshegde: I could find the `get_node_text()` upstream but I just cant find the `get_node_at_cursor() `upstream. I am on the latest build of nvim

clason: Yeah, that was my bad. I got a bit carried away there. (This _should_ be upstreamed, but clearly hasn't yet.)

PR #2832, opened 2022-04-18
fixup: return upstream text

clason:
another fixup... @theHamsta 

probably needs more, as the upstream function looks different (also accounts for string parsers)


vsedov:
I think it also might be wise to have more test cases, to prevent this from happening again in the future. Maybe?

clason:
Can't test downstream plugins, unfortunately.

PR #2834, opened 2022-04-18
fixup: adapt to upstream function

clason:
more fixes to adapt to upstream `get_node_text` returning a string rather than a table.

### DIFF
--- a/lua/nvim-treesitter/ts_utils.lua
+++ b/lua/nvim-treesitter/ts_utils.lua
@@ -347,8 +347,8 @@ function M.swap_nodes(node_or_range1, node_or_range2, bufnr, cursor_to_second)
   local text1 = ts_query.get_node_text(node_or_range1)
   local text2 = ts_query.get_node_text(node_or_range2)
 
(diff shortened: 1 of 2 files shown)


lua/nvim-treesitter/ts_utils.lua
```diff
@@ -347,8 +347,8 @@ function M.swap_nodes(node_or_range1, node_or_range2, bufnr, cursor_to_second)
   local text1 = ts_query.get_node_text(node_or_range1)
   local text2 = ts_query.get_node_text(node_or_range2)
 
-  local edit1 = { range = range1, newText = table.concat(text2, "\n") }
-  local edit2 = { range = range2, newText = table.concat(text1, "\n") }
+  local edit1 = { range = range1, newText = text2 }
+  local edit2 = { range = range2, newText = text1 }
```
fsouza, line 351: this is not enough. text1 and text2 are used as table in multiple places in this function, I'm working on a patch but let me know if you'd rather have me hold off

Issue #2825, opened 2022-04-18
Better control over conceals and highlighting


clason:
Currently, conceals are set unconditionally, which is a regression over legacy syntax files which usually gate the more controversial ones (like JSON quotes) behind a global configuration option.

We should also add a way of enabling/disabling individual conceals; at the same time, we should _not_ replicate the zoo of language-specific `g:foo`globals. One possible idea is to use the now supported atom-style fallbacks for the conceal capture (`@conceal.fancy`, `@conceal.quotes`, or somesuch) to define levels of conceals which can then be selected by a predicate (which would have to be configurable; no idea yet how to do this with queries).

An alternative would be to define the conceals in separate queries that are optionally loaded, but that would likely have performance implications (and require additional query handling functionality).

clason:
Come to think of it, this is an even larger issue: a not small number of Vim syntax files use global options to control highlighting (at the syntax level). To achieve parity, we need some way of influencing the queries from the Neovim side.

theHamsta:
> Come to think of it, this is an even larger issue: a not small number of Vim syntax files use global variables to control highlighting (at the syntax level). To achieve parity, we need some way of influencing the queries from the Neovim side.

I don't see a need for that. A better way to select the queries (e.g. to filter out queries). We can already do conditional queries based on predicates. I used them in external repos when we wanted to make certain queries configurable.

Please don't add them to the highlights.

clason:
I'm sorry, I don't get this. Are you saying we should _not_ make highlights configurable (in some way)?

theHamsta:
> Are you saying we should not make highlights configurable (in some way)?

yes. It's far more easy to just let users make the changes they need for their queries than to explain and maintain switches. When captures are specific enough every capture in the queries can be turned to a no-op.

I would love to give more options to edit queries in a general way but I wouldn't add some predefined choices (on/off) into the queries.

clason:
I think we're talking past each other here (any maybe should take it to the nvim-treesitter chat): I do _not_ want to add a bunch of configuration options; I thought I made this clear.

On the other hand, we _should_ make it easy for users to configure the queries; if that's already possible and just needs to be documented, so much the better. Can you give an example how that would look?

theHamsta:
also just rather leave away something that could be controversial.

clason:
Yeah, but then other people will request it (witness JSON quotes conceal). You can't make both happy without offering choice.

theHamsta:
[...]

Issue #2663, opened 2022-03-21
Tracking Issue: Tree-sitter-based Concealing


theHamsta:
As https://github.com/nvim-treesitter/nvim-treesitter/issues/2660 mentioned it is now possible to use the conceal feature for tree-sitter highlights using https://github.com/neovim/neovim/pull/17776. This issues is for tracking and discussing for which languages conceal would make sense and how we would structure the queries for that. Please post here when you would like to have some conceal features for a certain language.

The following languages already have PRs:

 - [ ] markdown #2662
 - [ ] JSON (https://github.com/nvim-treesitter/nvim-treesitter/pull/2664, see #645)

TODOs:

- [ ] RST
- [ ] LaTeX (needs discussion because of huge number of single-character replacements)

gegoune:
Would be nice to be able to conceal `"`s in `json` files. Related https://github.com/nvim-treesitter/nvim-treesitter/issues/645.

clason:
Also, a fundamental question: Do we want to put these into the `highlights.scm` queries (following the legacy syntax precedent), or create a new `conceal.scm` query type so people who don't want conceal can just ignore them?

theHamsta:
@clason I also had this question. We could logically merge `conceal.scm` into `highlight.scm` but keep them logically separated on filesystem. If we assume that most queries are as small as for markdown it would be tolerable to have them in the highlights.scm.

clason:
I don't know what the best option is, either. For Markdown I agree, but I was mostly thinking of LaTeX.

We could also think of handling LaTeX as a special case? I agree that we should not needlessly increase the number of modules and options.

theHamsta:
For LaTeX, we probably need a check on the conceallevel for more fancy conceals. Could be handled by predicates. Are there other reasons why LaTeX is special? I think most of the needed tokens (e.g. `$` for math mode are already in the token tree.

clason:
No, just the fancy conceals (meaning, you have a replacement instead of just hiding a capture). But there are [a lot of them](https://github.com/nvim-treesitter/nvim-treesitter/issues/645)...

megalithic:
I think this is what i'm most excited about.. `- [ ]` rendering as `- ` and `- [x]` as `- ` basically a checkbox that is checked, or any number of options for that (as also seen in the various org-mode plugins).

j-hui:
[...]

PR #2835, opened 2022-04-18
highlights(markdown): conceal check boxes

theHamsta:
Demo for @megalithic (https://github.com/nvim-treesitter/nvim-treesitter/issues/2663#issuecomment-1101757822)

not sure whether one would only count `- [ ]` as checkbox or every `[ ]`. There is also ☑ which is sometimes rendered in the same style and sometimes as a green tick.

Non-ascii conceals should be somehow guarded as mentioned in #2825

### DIFF
--- a/queries/markdown/highlights.scm
+++ b/queries/markdown/highlights.scm
@@ -88,3 +88,6 @@
 (inline_link 
   "["  @conceal
   (#set! conceal ""))
+
+((shortcut_link) @conceal (#set! conceal "") (eq? @conceal "[ ]"))
+((shortcut_link) @conceal (#set! conceal "☒") (eq? @conceal "[x]"))


megalithic:
🙏 thank you so much @theHamsta !

MDeiml:
This is probably broken now (or will break soon), because the grammar now actually emits `(task_list_marker_checked)` and `(task_list_marker_unchecked)` nodes. So the queries would change to
```scm
((task_list_marker_checked) @conceal (#set! conceal ""))
((task_list_marker_unchecked) @conceal (#set! conceal "☒"))
```
(Also imo [patched fonts](https://www.nerdfonts.com/) contain nicer checkmarks. I also conceal the list items themeselves like this
```scm
(list_item [
  (list_marker_plus)
  (list_marker_minus)
  (list_marker_star)
  (list_marker_dot)
  (list_marker_parenthesis)
] @conceal [
    (task_list_marker_checked)
    (task_list_marker_unchecked)
](#set! conceal ""))
```
)

theHamsta:
@megalithic do you want to create a PR? I'm not sure if we want that at all since support for those unicode characters might be very limited depending on font

akinsho:
@theHamsta I presume(d) something like this is way too opinionated to be merged. Whilst I like and use this effect, I can imagine you will get a lot of issues from users who do not like it who want it to be turned off. Might be a good use case for a README

PR #2838, opened 2022-04-19
fix(ts_utils): fix swap_nodes after get_node_text change

### LINKED ISSUE #2849: Statusline is broken
### Describe the bug

There are three issues with this API
1) it returns `\n` because of recent changes. That will break my statusline plugin as the statusline height is 1. It can not display multiple lines
2) Too much info to be useful
3) get_node_text from `ts_query` may be broken.

### To Reproduce

Take following code as an example:
```lua
-- Trim spaces and opening brackets from end
local transform_line = function(line)
  return line:gsub("%s*[%[%(%{]*%s*$", "")
end
```
When I put cursor on `return` and run
```lua
lua print(require'nvim-treesitter').statusline())
```
I got
```lua
 function(line)\n\treturn line:gsub("%s*[%[%(%{]*%s*$", "")\nend
```
This is too much info for a status to display.

### Expected behavior

It used to return `function(line)`

The root cause is using 
```lua
vim.treesitter.query.get_node_text()
```

https://github.com/nvim-treesitter/nvim-treesitter/blob/bd2f5d770d0a98b1d89bb95dab7ca396b01ce1fe/lua/nvim-treesitter.lua#L37

I think it used to be ts_utils.get_node_text()

```markdown
## Installation
  - OK: `tree-sitter` found 0.20.6 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v16.14.2 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
    Version: cc (GCC) 11.2.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - yaml           ✓ ✓ ✓ ✓ ✓ 
  - hcl            ✓ . ✓ ✓ ✓ 
  - dart           ✓ ✓ . ✓ ✓ 
  - json           ✓ ✓ ✓ ✓ . 
  - html           ✓ ✓ ✓ ✓ ✓ 
  - css            ✓ . ✓ ✓ ✓ 
  - javascript     ✓ ✓ ✓ ✓ ✓ 
  - java           x ✓ . ✓ ✓ 
  - vue            ✓ . ✓ ✓ ✓ 
  - typescript     ✓ ✓ ✓ ✓ ✓ 
  - python         ✓ ✓ ✓ ✓ ✓ 
  - jsdoc          ✓ . . . . 
  - rust           ✓ ✓ ✓ ✓ ✓ 
  - c              ✓ ✓ ✓ ✓ ✓ 
  - lua            ✓ ✓ ✓ ✓ ✓ 
  - toml           ✓ ✓ ✓ ✓ ✓ 
  - go             ✓ ✓ ✓ ✓ ✓ 
  - tsx            ✓ ✓ ✓ ✓ ✓ 
  - cpp            ✓ ✓ ✓ ✓ ✓ 
  - jsonc          ✓ ✓ ✓ ✓ ✓ 

## The following errors have been detected:
  - ERROR: java(highlights): /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:172: query: invalid node type at position 2028
    java(highlights) is concatenated from the following files:
    | [ERROR]:"/home/ray/.local/share/nvim/site/pack/packer/opt/nvim-treesitter/queries/java/highlights.scm", failed to load: /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:172: query: invalid node type at position 2028
```

```text
NVIM v0.7.0-dev+1400-g233014f92
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by ray

Features: +acl +iconv +tui
See ":help feature-compile"

```

### Additional context

_No response_

fsouza:
After some discussion, it looks like the easiest thing to do for now is
to keep a private copy of get_node_text (just to skip the deprecation
message) and invoke that, until core provides an equivalent function
that can return the node content in a table representing the node
"lines".

Also fixes the statusline.

Fix #2849.

### DIFF
--- a/lua/nvim-treesitter/ts_utils.lua
+++ b/lua/nvim-treesitter/ts_utils.lua
@@ -2,20 +2,10 @@ local api = vim.api
 
 local parsers = require "nvim-treesitter.parsers"
 local utils = require "nvim-treesitter.utils"
-local ts_query = vim.treesitter.query
 
 local M = {}
 
---- Gets the actual text content of a node
--- @deprecated Use vim.treesitter.query.get_node_text
--- @param node the node to get the text from
--- @param bufnr the buffer containing the node
--- @return list of lines of text of the node
-function M.get_node_text(node, bufnr)
-  vim.notify_once(
-    "nvim-treesitter.ts_utils.get_node_text is deprecated: use vim.treesitter.query.get_node_text",
-    vim.log.levels.WARN
-  )
+local function get_node_text(node, bufnr)
   local bufnr = bufnr or api.nvim_get_current_buf()
   if not node then
     return {}
@@ -37,6 +27,37 @@ function M.get_node_text(node, bufnr)
     -- If line is nil then the line is empty
     return line and { string.sub(line, start_col + 1, end_col) } or {}
   end
+end
+
+---@private
+function M._get_line_for_node(node, type_patterns, transform_fn, bufnr)
+  local node_type = node:type()
+  local is_valid = false
+  for _, rgx in ipairs(type_patterns) do
+    if node_type:find(rgx) then
+      is_valid = true
+      break
+    end
+  end
+  if not is_valid then
+    return ""
+  end
+  local line = transform_fn(vim.trim(get_node_text(node, bufnr)[1] or ""))
+  -- Escape % to avoid statusline to evaluate content as expression
+  return line:gsub("%%", "%%%%")
+end
+
+--- Gets the actual text content of a node
+-- @deprecated Use vim.treesitter.query.get_node_text
+-- @param node the node to get the text from
+-- @param bufnr the buffer containing the node
+-- @return list of lines of text of the node
+function M.get_node_text(node, bufnr)
+  vim.notify_once(
+    "nvim-treesitter.ts_utils.get_node_text is deprecated: use vim.treesitter.query.get_node_text",
+    vim.log.levels.WARN
+  )
+  return get_node_text(node, bufnr)
 end
 
 --- Determines whether a node is the parent of another
@@ -344,8 +365,8 @@ function M.swap_nodes(node_or_range1, node_or_range2, bufnr, cursor_to_second)
   local range1 = M.node_to_lsp_range(node_or_range1)
   local range2 = M.node_to_lsp_range(node_or_range2)
 
-  local text1 = ts_query.get_node_text(node_or_range1)
-  local text2 = ts_query.get_node_text(node_or_range2)
+  local text1 = get_node_text(node_or_range1, bufnr)
+  local text2 = get_node_text(node_or_range2, bufnr)
 
   local edit1 = { range = range1, newText = table.concat(text2, "\n") }
   local edit2 = { range = range2, newText = table.concat(text1, "\n") }
--- a/lua/nvim-treesitter.lua
+++ b/lua/nvim-treesitter.lua
@@ -3,7 +3,6 @@ local utils = require "nvim-treesitter.utils"
 local info = require "nvim-treesitter.info"
 local configs = require "nvim-treesitter.configs"
 local parsers = require "nvim-treesitter.parsers"
-local ts_query = vim.treesitter.query
 local ts_utils = require "nvim-treesitter.ts_utils"
 
 -- Registers all query predicates
@@ -20,23 +19,6 @@ end
 
 function M.define_modules(...)
   configs.define_modules(...)
-end
-
-local get_line_for_node = function(node, type_patterns, transform_fn, bufnr)
-  local node_type = node:type()
-  local is_valid = false
-  for _, rgx in ipairs(type_patterns) do
-    if node_type:find(rgx) then
-      is_valid = true
-      break
-    end
-  end
-  if not is_valid then
-    return ""
-  end
-  local line = transform_fn(vim.trim(ts_query.get_node_text(node, bufnr) or ""))
-  -- Escape % to avoid statusline to evaluate content as expression
-  return line:gsub("%%", "%%%%")
 end
 
 -- Trim spaces and opening brackets from end
@@ -67,7 +49,7 @@ function M.statusline(opts)
   local expr = current_node
 
   while expr do
-    local line = get_line_for_node(expr, type_patterns, transform_fn, bufnr)
+    local line = ts_utils._get_line_for_node(expr, type_patterns, transform_fn, bufnr)
     if line ~= "" and not vim.tbl_contains(lines, line) then
       table.insert(lines, 1, line)
     end


lua/nvim-treesitter/ts_utils.lua
```diff
@@ -360,25 +360,28 @@ function M.swap_nodes(node_or_range1, node_or_range2, bufnr, cursor_to_second)
       range1["end"].line < range2.start.line
       or (range1["end"].line == range2.start.line and range1["end"].character < range2.start.character)
     then
-      line_delta = #text2 - #text1
+      line_delta = (range2["end"].line - range2["start"].line + 1) - (range1["end"].line - range1["start"].line + 1)
     end
 
     if range1["end"].line == range2.start.line and range1["end"].character < range2.start.character then
+      local lines1 = vim.split(text1, "\n")
+      local lines2 = vim.split(text2, "\n")
```
fsouza, line 368: this isn't the fastest, could try to figure out something faster (like counting the characters from the end of the string to a `\n`), but this is very much of an edge case and I don't expect people to be swapping very large nodes, so the extra memory usage should be ok? (famous last words)

fsouza: Pattern matching would also work, something like:

```lua
local function length_last_line(str)
  if str == "" then
    return 0
  end
  local first, last = string.find(str, "\n?[^\n]+$")
  return last - first
end

-- ...

local length_last_line1 = length_last_line(text1)
local length_last_line2 = length_last_line(text2)
```

Though that isn't super readable 🤷

theHamsta: Wasn't aware that get_node_text wasn't no longer working on lines. Seems a bit odd for me to join the lines just to split them afterwards which occurs in many occasions of get_node_text. Maybe we should just revert it. Can't you get the #lines from the node range?

theHamsta: You should get the length of the last line by looking at the start/end cols of the node range

fsouza: > You should get the length of the last line by looking at the start/end cols of the node range

OMG that's right. Sorry I'll fix that later today.

fsouza: Fixed it. I feel like it'd be cool to have some tests for this, but I assume that would require mocking get_node_text. Let me know if you have any thoughts on this.

theHamsta: You can have a look on the tests that were there when we removed `get_node_text?

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -337,18 +337,26 @@ function M.memoize_by_buf_tick(fn, options)
   end
 end
 
+local function length_last_line(range)
+  if range.start.line == range["end"].line then
+    return range["end"].character - range.start.character + 1
+  else
+    return range["end"].character + 1
+  end
+end
```
theHamsta, line 346: I'm not sure whether this is correct. When end col = 0 then get_node_text would not include this last line also the + 1 shouldn't be necessary. As `get_vim_range` shows it is apparently not possible to get the length of the last line without accessing the actual text. Probably you pattern matching approach would be better (or return to ye old get_node_text)

fsouza: My thought with +1 is that ranges are zero-based and end-inclusve? So, for example, if range.end.character is 12, it means it includes characters 0, 1, 2, ..., 12, i.e. 13 characters.

I see what you mean though, range.end.character may be "out of bounds"? (accessing the 0th element in an empty array is an out of bound errors, but it seems that that's expected to work here?)

Are you saying that for a range like `{start = {line = 1, col = 0}, end = {line = 10, col = 0}}`, get_node_text will only return 9 lines? In that case this won't work either: https://github.com/nvim-treesitter/nvim-treesitter/pull/2838/files#diff-ee62282252e6dfba5eebd5b1a5a1ff9df64f01392723b3044965d5c7469e9915R371

fsouza: but yeah it sounds like the best solution for now is to keep nvim-treesitter's version of get_node_text that returns a table with the lines instead of using vim.treesitter.query.get_node_text?

theHamsta: @clason are you fine with reverting? Most plugin can do the switch easily while for some application the join/split is just a needless round trip. For swap it's not critical since swap is slow due to the queries finding the swap objects

clason:
> @clason are you fine with reverting? Most plugin can do the switch easily while for some application the join/split is just a needless round trip. For swap it's not critical since swap is slow due to the queries finding the swap objects

I've already re-added the old implementation so plugins are no longer broken.

However, I don't like having two almost-but-not-entirely identical functions _with the same name_ here and upstream. I would prefer if we would either

1. add an option table with `table=true` (default `false`) to the upstream function, or
2. rename the function here into something that makes the difference clearer.

To make absolutely sure: the only difference (beside the `bufnr` argument, which is a _good_ change in any case) is the question of returning a table vs a concatenated string?

theHamsta:
The `bufnr` was there before but it was optional. I think the return value is the only difference.

clason:
Then my (strong) preference would be option 1.

fsouza:
> Then my (strong) preference would be option 1.

That would make nvim-treesitter dependent on nightly though. Is that ok?

clason:
I think that's something that can be backported to 0.7.1 in this case, so it wouldn't be too bad. We still have the old code, anyway, so we can refactor this into a private method that we can call here without deprecation warnings until the release.

fsouza:
@clason @theHamsta I just pushed an updated commit making use of a private copy of get_node_text that won't print the deprecation warning. I also rewrote the commit message to describe the new change. Let me know what yall think of that.

fsouza:
Sorry, fixed the linter violation 🙈

fsouza:
@clason @theHamsta friendly ping :D

clason (approved):

clason:
While you're at it, could you also change the use in https://github.com/nvim-treesitter/nvim-treesitter/blob/bd2f5d770d0a98b1d89bb95dab7ca396b01ce1fe/lua/nvim-treesitter.lua#L37
back to the internal function? Should be 
```lua
get_node_text(node)[1]
```
with the internal function. (Probably best to move this into `ts_util` and expose it as a private function.)

This would then close #2849

clason:
Yes, and marked `---@private` for good measure.

fsouza:
> Yes, and marked `---@private` for good measure.

Cool, should be good to go now!

fsouza:
@clason do you think we can merge or should we wait for @theHamsta?

clason:
I'd like to wait for @theHamsta just to make sure. I'd prefer fixing it once ;)

AckslD:
Sorry if I might be misunderstanding something but one other issue seems to be that `swap_nodes` takes a node or range which is used by `treesitter-textobjects` [here](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/master/lua/nvim-treesitter/textobjects/swap.lua#L19). When this gets passed to `get_node_text` it breaks since a if it's a range calling the method `:start` clearly doesn't work.

fsouza:
> Sorry if I might be misunderstanding something but one other issue seems to be that `swap_nodes` takes a node or range where the latter is used by `treesitter-textobjects` [here](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/master/lua/nvim-treesitter/textobjects/swap.lua#L19). When this gets passed to `get_node_text` it breaks since a if it's a range calling the method `:start` clearly doesn't work.

As far as I can tell this PR is reverting the behavior back to what it used to be. Is nvim-treesitter-textobjects also failing if you use this PR?

AckslD:
@fsouza It does indeed work with this branch :+1: Just looking at the code I couldn't figure out how ranges are handled in `swap_nodes`

fsouza:
> @fsouza It does indeed work with this branch 👍 Just looking at the code I couldn't figure out how ranges are handled in `swap_nodes`

it's handled in get_node_range (which is called from get_node_text): https://github.com/nvim-treesitter/nvim-treesitter/blob/4a53674d5b90ab959322d381121c16072b00ac4b/lua/nvim-treesitter/ts_utils.lua#L301-L307

theHamsta (approved):

PR #2842, opened 2022-04-19
Add org parser

TerseTears:
This simply adds the [tree-sitter-org](https://github.com/milisims/tree-sitter-org) parser to the list of parsers. Highlights, folds, etc. are handled by the separate [orgmode](https://github.com/nvim-orgmode/orgmode) plugin for now it seems. Regardless, this fixes the issue of having to separately add the parser to `nvim-treesitter` for the purposes of simply accessing treesitter information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [ocaml](https://github.com/tree-sitter/tree-sitter-ocaml) (maintained by @undu)
 - [x] [ocaml_interface](https://github.com/tree-sitter/tree-sitter-ocaml) (maintained by @undu)
 - [x] [ocamllex](https://github.com/atom-ocaml/tree-sitter-ocamllex) (maintained by @undu)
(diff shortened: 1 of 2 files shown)


theHamsta:
I think links to additional language extensions next to the parsers would be cool. Like nvim-orgmode, neorg, or nvim-treesitter-commonlisp would be cool.

theHamsta (approved):

TerseTears:
Thank you for the review and approval. I just noticed there's a workflow for updating the languages list in the README. Would I still be able to add info on related plugins then? And would this be okay:

```markdown 
- [ ] [org](https://github.com/milisims/tree-sitter-org) See also [nvim-orgmode](https://github.com/nvim-orgmode/orgmode), [tree-sitter-commonlisp](https://github.com/theHamsta/tree-sitter-commonlisp), [neorg](https://github.com/nvim-neorg/neorg)
```

theHamsta:
Not, if you modify the workflow. You could add a "related plugins" table to the parser info and let the workflow add it to the README

PR #2851, opened 2022-04-21
fix(ts_utils): fix wrong usage of get_node_text

AckslD:
I think this was not right?


clason:
Indeed and thanks, but duplicate of https://github.com/nvim-treesitter/nvim-treesitter/pull/2838.

PR #2721, opened 2022-04-02
indents(hcl): no indentation after brace

theHamsta:
Fixes #2590

PR #2855, opened 2022-04-23
hcl: fix indentation queries

MichaHoffmann:
This PR continues the work done by @theHamsta in #2721 by fixing the queries to make the tests pass.


kyazdani42:
looks like the hcl tests all fail :/

MichaHoffmann:
> looks like the hcl tests all fail :/

my bad, i messed up the queries, should be better now.

MichaHoffmann:
i tried to add a test for function calls, but that failed and i have to adapt the parser for that to work i fear ( maybe a breaking change )

MichaHoffmann:
I added a failing test for the problem that if you an inline tuple like 
```
test {
  x = [1, 2]
y = [3, 4]
}
```
then the next attribute will be less indented. I think thta is because we decrease the indentation level since we see the "tuple_end" but it was not more indented before since we have no multiline tuple. ( it works fine with multiline tuples )

MichaHoffmann:
Is there a way to only do the "@dedent" on multiline tuples?

theHamsta:
@MichaHoffmann you can write query that only captures multiple tuples

MichaHoffmann:
> @MichaHoffmann you can write query that only captures multiple tuples

I dont think i understand completely yet. Do you have an example?

theHamsta:
All tests pass for me when I remove `@dedent`. Can you give an example when `@dedent` is needed?

MichaHoffmann:
> All tests pass for me when I remove `@dedent`. Can you give an example when `@dedent` is needed?
> 
> I don't understand which situation you want to handle but if it involves multiple nodes being present you can just name all of the in the query and only give some of the the capture. The other nodes are then just there to only apply the rule when they are present in the neightbor hood (`(parent (child) @foo (child) @foo)` will only apply foo when there are at least two of them.

Example: https://github.com/nvim-treesitter/nvim-treesitter/runs/6142952479?check_suite_focus=true#step:8:123

Not sure what is going on with that test though.

MichaHoffmann:
From the test failures it seems like there is a problem with anything that comes after a block

Something like this:
```
test {
  x = "foo"
}
  y = "bar"
```

Is there a way to write a query to dedent attributes that come after such a block? I tried but could not make it work

theHamsta:
Me neither, tests seem to pass but without the dedent it indents the following line. I think this has probably something to do with indent.lua. `@dedent` just shouldn't be necessary but for some reason it still thinks that you are still inside the `block`. This might have something to do with how the parser set the range. Maybe we should have a look on how whitespaces at the end of a range get treated.

MichaHoffmann:
I had a look at the parse tree using the nvim-treesitter playground plugin. I could not find a meaningful difference in the tree generated for tuples, objects or blocks. Here a test with blocks, objects and tuples:

```
body [0, 0] - [15, 1]                                             │  1 test {
  block [0, 0] - [13, 1]                                          │  2   x = [
    identifier [0, 0] - [0, 4]                                    │  3     1,
    block_start [0, 5] - [0, 6]                                   │  4     2,
    body [1, 2] - [12, 3]                                         │  5     3,
      attribute [1, 2] - [5, 3]                                   │  6   ]
        identifier [1, 2] - [1, 3]                                │  7   y = {
        expression [1, 6] - [5, 3]                                │  8     a: 1,
          collection_value [1, 6] - [5, 3]                        │  9     b: 2,
            tuple [1, 6] - [5, 3]                                 │ 10   }
              tuple_start [1, 6] - [1, 7]                         │ 11   nest {
              expression [2, 4] - [2, 5]                          │ 12     c = 4
                literal_value [2, 4] - [2, 5]                     │ 13   }
                  numeric_lit [2, 4] - [2, 5]                     │ 14 }
              expression [3, 4] - [3, 5]                          │ 15 rest {
                literal_value [3, 4] - [3, 5]                     │ 16 }
                  numeric_lit [3, 4] - [3, 5]                     │~
              expression [4, 4] - [4, 5]                          │~
                literal_value [4, 4] - [4, 5]                     │~
                  numeric_lit [4, 4] - [4, 5]                     │~
              tuple_end [5, 2] - [5, 3]                           │~
      attribute [6, 2] - [9, 3]                                   │~
        identifier [6, 2] - [6, 3]                                │~
        expression [6, 6] - [9, 3]                                │~
          collection_value [6, 6] - [9, 3]                        │~
            object [6, 6] - [9, 3]                                │~
              object_start [6, 6] - [6, 7]                        │~
              object_elem [7, 4] - [7, 8]                         │~
                key: expression [7, 4] - [7, 5]                   │~
                  variable_expr [7, 4] - [7, 5]                   │~
                    identifier [7, 4] - [7, 5]                    │~
                val: expression [7, 7] - [7, 8]                   │~
                  literal_value [7, 7] - [7, 8]                   │~
                    numeric_lit [7, 7] - [7, 8]                   │~
              object_elem [8, 4] - [8, 8]                         │~
                key: expression [8, 4] - [8, 5]                   │~
                  variable_expr [8, 4] - [8, 5]                   │~
                    identifier [8, 4] - [8, 5]                    │~
                val: expression [8, 7] - [8, 8]                   │~
                  literal_value [8, 7] - [8, 8]                   │~
                    numeric_lit [8, 7] - [8, 8]                   │~
              object_end [9, 2] - [9, 3]                          │~
      block [10, 2] - [12, 3]                                     │~
        identifier [10, 2] - [10, 6]                              │~
        block_start [10, 7] - [10, 8]                             │~
        body [11, 4] - [11, 9]                                    │~
          attribute [11, 4] - [11, 9]                             │~
            identifier [11, 4] - [11, 5]                          │~
            expression [11, 8] - [11, 9]                          │~
              literal_value [11, 8] - [11, 9]                     │~
                numeric_lit [11, 8] - [11, 9]                     │~
        block_end [12, 2] - [12, 3]                               │~
    block_end [13, 0] - [13, 1]                                   │~
  block [14, 0] - [15, 1]                                         │~
    identifier [14, 0] - [14, 4]                                  │~
    block_start [14, 5] - [14, 6]                                 │~
    block_end [15, 0] - [15, 1]                                   │~
~                                                                 │~
```

MichaHoffmann:
It works if i mess with the "is_empty_line" logic in indent.lua. If `node` is always initialized as `node = get_first_node_at_line(root, lnum)` then all tests pass.

MichaHoffmann:
Ok the last commit fixes the tests finally.

theHamsta:
I also saw that `get_last_node_at_line` and first `get_node_at_line` are very brittle. Thanks for fixing the tests. Do you have a suggestion on how we could fix that in `indent.lua`? Maybe check all the nodes that end at that point?

theHamsta (approved):

MichaHoffmann:
> I also saw that `get_last_node_at_line` and first `get_node_at_line` are very brittle. Thanks for fixing the tests. Do you have a suggestion on how we could fix that in `indent.lua`? Maybe check all the nodes that end at that point?
> 
> Is this ready to merge?

Yeah i think so, let me rebase really quick though. I'll add a few more tests and queries for function arguments in a subsequent PR ( that might need changes in the parser ).

`get_last_node_at_line` feels like it should iterate through the children/siblings somehow, but im not sure tbh, i'll try to think of something.

PR #2865, opened 2022-04-26
feat: add m68k parser and queries

grahambates:
Adding my parser for Motorola 68000 assembly.

### DIFF
--- a/queries/m68k/highlights.scm
+++ b/queries/m68k/highlights.scm
@@ -1,0 +1,73 @@
+(symbol) @variable
+
+(label name: (symbol) @constant)
+
+[
+  (instruction_mnemonic)
+  (directive_mnemonic)
+] @function.builtin
+
+(include (directive_mnemonic) @include)
+(include_bin (directive_mnemonic) @include)
+(include_dir (directive_mnemonic) @include)
+
+
+(size) @attribute
+
+(macro_definition name: (symbol) @function.macro)
+(macro_call name: (symbol) @function.macro)
+
+[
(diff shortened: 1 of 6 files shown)


queries/m68k/highlights.scm
```diff
@@ -0,0 +1,73 @@
...
+  (address_register)
+  (data_register)
+  (float_register)
+  (named_register)
+] @keyword
+
+(repeat (control_mnemonic) @repeat)
+(conditional (control_mnemonic) @conditional)
+
+(interpolated (macro_arg) @embedded)
```
theHamsta, line 50: there is no `@embedded`. You probably want to use `@none`

theHamsta (commented):
Seems to look ok, apart from that little comment

queries/m68k/highlights.scm
grahambates: I've removed it. The previous query mapping `macro_arg` to `variable.builtin` is fine in this context too.

theHamsta: You probably meant to use `injections.scm`. See how injections work for C. It is not in `highlight.scm` for our repo

grahambates: Thanks for that. I had a look and I don't think it's really suitable here as it's a different use case to C's macro args. I'm happy to explain in more detail, but in short I think it makes sense to leave all macro args highlighted as variables. I did however add injections for comments and I've now added locals queries too, which I've tried out with `nvim-treesitter-refactor`.

Issue #2307, opened 2022-01-20
Support for protobuf


sainad2222:
Any plans for adding support for Google's protocol buffer(proto3)?

References: [proto.vim](https://github.com/protocolbuffers/protobuf/blob/3f5fc4df1de8e12b2235c3006593e22d6993c3f5/editors/proto.vim)

clason:
Sure, PR welcome! I've added it to the tracking issue.

theHamsta:
I didn't find tree-sitter-protobuf too useful. It failed to parse even simple documents.

sainad2222:
@theHamsta I'm sorry, which one you are talking about? [This?](https://github.com/mitchellh/tree-sitter-proto)

hiberabyss:
seems no indents for protobuf.

PR #2866, opened 2022-04-26
feat(proto): add parser

fsouza:
Related to #2307.

First time doing this, so let me know if I did anything wrong. I did notice that the parser has some [queries predefined](https://github.com/mitchellh/tree-sitter-proto/tree/main/queries), so I wonder if we could use those?

### DIFF
--- a/queries/proto/highlights.scm
+++ b/queries/proto/highlights.scm
@@ -1,0 +1,62 @@
+[
+  "syntax"
+  "package"
+  "option"
+  "import"
+  "service"
+  "rpc"
+  "returns"
+  "message"
+  "enum"
+  "oneof"
+  "repeated"
+  "reserved"
+  "to"
+] @keyword
+
+[
+  (key_type)
+  (type)
+  (message_name)
+  (enum_name)
+  (service_name)
+  (rpc_name)
+  (message_or_enum_type)
+] @type
+
+(enum_field
+  (identifier) @constant)
+
+[
+  (string)
+  "\"proto3\""
+] @string
+
+(int_lit) @number
+
+(float_lit) @float
+
+[
+  (true)
+  (false)
+] @boolean
+
+(comment) @comment
+
+[
+  "("
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
+  "<"
+  ">"
+]  @punctuation.bracket
+
+[
+ ";"
+ ","
+] @punctuation.delimiter
+
+"=" @operator
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1050,6 +1050,16 @@ list.m68k = {
   filetype = "asm68k",
 }
 
+list.proto = {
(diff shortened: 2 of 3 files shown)


theHamsta (approved):
Nice! Finally a proto parser that seems to work! Thanks also for creating the highlights.

Just a few comments

queries/proto/highlights.scm
```diff
@@ -0,0 +1,48 @@
...
+  "returns"
+  "message"
+  "enum"
+  "oneof"
+  "repeated"
+  "reserved"
+  "to"
+] @keyword
+
+[
```
theHamsta, line 17: `"\"proto3\""` could be `@string`, `=` and `;` could be highlighted.

Enum identifiers could be highlighted as `@constant` but not a must

queries/proto/highlights.scm
```diff
@@ -0,0 +1,48 @@
...
+  (enum_name)
+  (service_name)
+  (rpc_name)
+] @type
+
+(string) @string
+
+[
+  (int_lit)
+  (float_lit)
```
theHamsta, line 30: there is a separate highlight for `@float`

queries/proto/highlights.scm
```diff
@@ -0,0 +1,48 @@
...
+] @number
+
+[
+  (true)
+  (false)
+] @constant.builtin
```
theHamsta, line 36: `@boolean`?

queries/proto/highlights.scm
fsouza: Pushed all suggestions except for `=`. Do you think we should use operator for that?

theHamsta: I think `@operator` should be good for that

queries/proto/highlights.scm
```diff
@@ -0,0 +1,55 @@
...
+
+(comment) @comment
+
+[
+  "("
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
```
theHamsta, line 51: `"<"  `">"` could belong to that category

queries/proto/highlights.scm
```diff
@@ -0,0 +1,55 @@
...
+]  @punctuation.bracket
+
+
+";" @punctuation.delimiter
```
theHamsta, line 55: also `","` (use in `map<foo,bar>`

theHamsta: I'm unsure about `map` itself. Leave it like it is or use `@type`?

theHamsta: `(message_or_enum_type) @type`

fsouza:
@theHamsta I think I got it all, let me know if I missed anything or if there's any new feedback!

theHamsta:
@fsouza No, it's great. Thank you a lot for this contribution. I've been waiting a long time for that. Previous proto-parsers I found couldn't reliably parse most files.

theHamsta (approved):

PR #2874, opened 2022-04-28
Reformat "parserinfo" as table with related plugins column

TerseTears:
This modifies the automatically generated "parserinfo" section of the README to display the supported languages as a table with an additional "related plugins" column.

The result can be seen on the branch, where I've also added related plugins for "org".
https://github.com/TerseTears/nvim-treesitter/tree/readme#supported-languages

Let me know what you think and if any fixes are required or if I should add other related plugins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,6 +272,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [zig](https://github.com/maxxnino/tree-sitter-zig) (maintained by @maxxnino)
 <!--parserinfo-->
 
+For related information on the supported languages, including related plugins, see [this wiki page](https://github.com/nvim-treesitter/nvim-treesitter/wiki/Supported-Languages-Information).
 
 # Available modules
 


clason:
Hmm, that's not a bad idea, in general, but currently I think this is quite a bit less readable than the current list.

I am also not convinced that the parser table is the right place for this kind of ephemeral information; I'd rather add a second table just for related plugins or (better) a Wiki page that is (statically) linked from the README section.

The situation is different if a plugin is the source of truth for our queries (i.e., if it's a "language plugin" that officially maintains the queries, and we just pull them in periodically, like Vim does with some of its syntax files).

clason:
You _can_ change this PR to add the link to the wiki to the Readme, though ;)

TerseTears:
Great Idea. I added a related page to the Wiki. I will update the branch accordingly then.

clason:
Thanks for the work on the Wiki! A couple of comments:

1. I think it's better to leave out parsers without information -- otherwise the actual information is a bit harder to find.
2. My impression was that "related plugins" should be related to _nvim-treesitter_ -- plugins either extending (e.g., by adding or replacing queries) or making use of (for some language-specific feature) the included parser. (For example, VimTeX is a great plugin, but it has zero connection to tree-sitter -- quite the opposite! they conflict to a significant extent -- and therefore does not belong in this list.)

To put it bluntly: I don't think we want a general "language database for Neovim" here, as that has a much too wide scope and will be too hard to maintain.

TerseTears:
Yes, I agree. I updated the wiki page to include only what's relevant to tree-sitter.

TerseTears:
Yes, I looked at the source and it only uses tree-sitter for highlight groups. I will remove that as well then. Thank you.

theHamsta (approved):

PR #2878, opened 2022-04-29
highlights(ocaml): highlight units as such

undu:
Units were shadowed by the punctuation highlights in all situations.
Restrict the context where parentheses are highlighted as punctuation so
() can be highlighted as constants.

Before:
before

After
after
:

### DIFF
--- a/queries/ocaml/highlights.scm
+++ b/queries/ocaml/highlights.scm
@@ -71,7 +71,8 @@
 ; Constants
 ;----------
 
-(unit) @constant.builtin
+; Don't let normal parens take priority over this
+((unit) @constant.builtin (#set! "priority" 105))
 
 (boolean) @boolean
 


queries/ocaml/highlights.scm
```diff
@@ -124,7 +124,10 @@
 
 "%" @punctuation.special
 
-["(" ")" "[" "]" "{" "}" "[|" "|]" "[<" "[>"] @punctuation.bracket
+; try not to override highlighting for the unit constant
+(parenthesized_expression ["(" ")"] @punctuation.bracket)
```
theHamsta, line 128: if that works it's ok. If not there is also the posibility to use `(#set! priority 101)`

theHamsta (approved):

queries/ocaml/highlights.scm
undu: Thanks, I'll try It as soon as I get some time at the computer

undu: You're right, current approach leaves some parens unhilighted, trying your approach

undu: This works, thanks!

theHamsta:
Thank you!

PR #2901, opened 2022-05-04
fix: Use ColorScheme event to define highlights

4513ECHO:
Before this, highlights were cleared when colorscheme was changed to one which is not support nvim-treesitter. This will fix the problem.

### DIFF
--- a/plugin/nvim-treesitter.lua
+++ b/plugin/nvim-treesitter.lua
@@ -116,6 +116,13 @@ local highlights = {
   TSTagAttribute = { link = "TSProperty", default = true },
 }
 
-for k, v in pairs(highlights) do
-  api.nvim_set_hl(0, k, v)
-end
+api.nvim_create_autocmd("ColorScheme", {
+  pattern = "*",
+  group = augroup,
+  callback = function()
+    for k, v in pairs(highlights) do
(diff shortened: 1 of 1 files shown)


clason:
Thank you, but I'm not sure that is the right fix. It would be better to first understand _why_ the groups are cleared. Probably the links need to be set with a bang (but they weren't before, so they should have been cleared also before the refactor to 0.7 APIs -- can you test that?)

4513ECHO:
Yesterday (perhaps after the refactor) I encountered cleared highlights, but it occured rarely.
Most colorschemes execute `hi clear` I have no idea why highlights were not cleared after colorschemes which does not support treesitter execute `hi clear` in before the refactor. also I have no idea why highlights are claered.
Is this a bug of `nvim_set_hl` or treesitter?

clason:
Unclear, which is why I want to understand this first before adding code as a workaround.

Would you mind trying to investigate further and getting reproducible steps I can test myself? Also please do test with a commit _before_ the refactor so we can be _sure_ it's related to the new API.

It would be better to open an issue with your findings and discuss there first. I'm closing this PR for now so it doesn't get accidentally merged; we can reopen it if it turns out to be the correct fix!

PR #2910, opened 2022-05-06
fix: duplicated autocommand for set indentexpr

ifplusor:
When open a new buffer, it will be add a new autocommand.

image

### DIFF
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -192,13 +192,32 @@ end
 
 local indent_funcs = {}
 
+local function autocmd_is_defined(event, pattern, command_str)
+  autocommands = vim.api.nvim_get_autocmds({
+    event = event,
+    pattern = pattern,
+  })
+  for _, autocmd in ipairs(autocommands) do
+    if autocmd.command == command_str then
+      return true
+    end
+  end
+  return false
+end
+
 function M.attach(bufnr)
   indent_funcs[bufnr] = vim.bo.indentexpr
   vim.bo.indentexpr = "nvim_treesitter#indent()"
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/indent.lua
```diff
@@ -192,13 +192,23 @@ end
 
 local indent_funcs = {}
 
+local function autocmd_is_defined(event, pattern, command_str)
+  return vim.api.nvim_exec(string.format("au %s %s", event, pattern), true):find(command_str) ~= nil
```
clason, line 196: We definitely want to avoid string manipulation and calling Vimscript functions here. If necessary (which I don't think it is; see my other comment), use `nvim_get_autocommands`.

clason (changes requested):

lua/nvim-treesitter/indent.lua
clason: I still think the other approach is much better, though.

ifplusor:
> I think a better fix (more in line with standard practices) would be to add `buffer = bufnr` (and `group = "NvimTreesitter"`) to this autocommand. That should prevent multiple autocommands when attaching to multiple buffers.
> 
> Even better would be to create a new `NvimTreesitterIndent` autogroup (at the top of the file), use that for the attach, and delete the autocommands for that group and `bufnr` in the `detach` function. (Look at the other modules to see how that should look.)

You're right. My initial idea was to make as few changes as possible.

clason:
> You're right. My initial idea was to make as few changes as possible.

I understand (and so did I, when I converted that autocommand -- and look where we are now ;)) But we should take this opportunity to do it right.

clason:
superseded by the rewrite on `main`

PR #2926, opened 2022-05-10
Add method to definitions in ECMA locals

neanias:
In Telescope, function declarations were being exposed in the selector,
but not method definitions which are also a form of function
declaration.

### DIFF
--- a/queries/ecma/locals.scm
+++ b/queries/ecma/locals.scm
@@ -26,6 +26,10 @@
   ((identifier) @definition.var)
    (#set! definition.var.scope parent))
 
+(method_definition
+  ((property_identifier) @definition.var)
+   (#set! definition.var.scope parent))
+
 ; References
 ;------------
 
--- a/queries/ecma/locals.scm
+++ b/queries/ecma/locals.scm
@@ -23,11 +23,11 @@
   (identifier) @definition.import)
 
 (function_declaration
-  ((identifier) @definition.var)
+  ((identifier) @definition.function)
    (#set! definition.var.scope parent))
 
 (method_definition
-  ((property_identifier) @definition.var)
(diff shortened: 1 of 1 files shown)


queries/ecma/locals.scm
```diff
@@ -26,6 +26,10 @@
   ((identifier) @definition.var)
    (#set! definition.var.scope parent))
 
+(method_definition
+  ((property_identifier) @definition.var)
+   (#set! definition.var.scope parent))
```
theHamsta, line 31: isn't this a for method parameters`

neanias: Which bit, sorry?

kyazdani42: `definition.var` and `definition.var.scope` ?

theHamsta: What is property_identifier when used inside a method

neanias: `property_identifier` seems to be the `identifier` equivalent for `method_definition`. I didn't find an associated `identifier` in the Tree-Sitter playground for methods, only functions. I wasn't clear on the difference/use of `definition.var.scope` from the documentation so added it to match the `function_declaration` query

theHamsta: but what would `property_identifier` be in a code example?

neanias: Yes, it would be the function name. I suspected that they should both be `@definition.function` as well, but wasn't sure.

```javascript
class Foo {
  bar() {
//^^^ property_identifier
  }
}
```

theHamsta: can you change it also for function?

theHamsta (approved):

PR #2934, opened 2022-05-13
Add the rnoweb filetype

bamonroe:
Rnoweb is a form of R markdown but entirely in Latex. I asked for the Latex grammar to be extended to accommodate the injections for R, but it was recommended to me to make a separate grammar. I have been using this grammar for a little while in neovim, and with my latest fixes, I think it's in a state that would be useful to others. Please let me know if there are any additional edit's you'd like me to include.


queries/rnoweb/injections.scm
```diff
@@ -0,0 +1,7 @@
+(latex) @latex @combined
+(rchunk
+	(renv_content) @r @combined
```
theHamsta, line 3: `@combined` is not yet supported by nvim (or it has a different understanding of combined)

theHamsta:
When I try to fuzz the scanner, it is trapped in an endless loop. I guess here a check for `lexer->lookahead == '\0'` is required https://github.com/bamonroe/tree-sitter-rnoweb/blob/8355efd90004c529719604d45ce5089fe26f51d5/src/scanner.c#L132-L137

Please always fuzz your parser when you have a handwritten scanner: instructions here https://github.com/ikatyang/tree-sitter-markdown/issues/14#issue-725850717

queries/rnoweb/injections.scm
bamonroe: I'm having a hard time finding in the documentation where I saw the `@combined` flag, but if i exclude it from the (latex) query, then there are errors throughout the document where (rchunk) and (rinline) nodes break latex environments. It appears to be doing something useful to me. Can you point me to where in the neovim code `@combined` is defined?

bamonroe:
> When I try to fuzz the scanner, it is trapped in an endless loop. I guess here a check for `lexer->lookahead == '\0'` is required https://github.com/bamonroe/tree-sitter-rnoweb/blob/8355efd90004c529719604d45ce5089fe26f51d5/src/scanner.c#L132-L137
> 
> Please always fuzz your parser when you have a handwritten scanner: instructions here [ikatyang/tree-sitter-markdown#14 (comment)](https://github.com/ikatyang/tree-sitter-markdown/issues/14#issue-725850717)

Thanks for this, I was unaware of this step in the process. I'll let you know when I've got it sorted out.

queries/rnoweb/injections.scm
theHamsta: I think it should be ok. It's just not behaving the way tree-sitter does.  It is defined or better used in languagetree.lua in neovim

bamonroe:
I've pushed the recommended changes to the parser, which can be seen [here](https://github.com/bamonroe/tree-sitter-rnoweb/commit/a429697b1ce2b9cfe49812e92144103c3c0e0c00).

I use Arch Linux and had to modify your instructions somewhat. The relevant libfuzzer path is as follows:
```
export LIB_FUZZER_PATH=/usr/lib/clang/13.0.1/lib/linux/libclang_rt.fuzzer-x86_64.a
```
Note the dependency on the current version of clang; future viewers of this comment should check their version.

I also had to modify the `build-fuzzers` script. Line 18 of the script sets `default_fuzz_flags="-fsanitize=fuzzer,address,undefined"` which works fine for `clang`, but this is followed by a call to `make`. In the `Makefile`, it appears `gcc` is used as `${CC}`, which doesn't support the `-fsanitize=fuzzer` flag causing `make` to fail with an error:
```
cc: error: unrecognized argument to '-fsanitize=' option: 'fuzzer'
```

To fix this, I added:
```
export CC
export CXX
export LINK
```
Just above the `make` command in the `build-fuzzers` script. This set `${CC}` to `clang`, which does support the fuzzer flag.

I have been running the fuzzer overnight, and it is still progressing, in that output is still ticking away, but I'm unfamiliar with the fuzz process and don't know if this long execution time is indicative of some kind of error (no error is indicated in the output thus far). @theHamsta Is this the usual amount of time for this process?

theHamsta:
@bamonroe it will run infinite time and just try out different inputs in a clever way to cover all parts of your code base. Not crashing is a good sign

bamonroe:
I'm not sure what do about the `Test queries / Parser compilation (windows-2022, cl, stable) (pull_request)` failed check. Looking at the logs, it appears the parser is correctly installed in the `Compile parsers` step, but isn't found in the `Check query files` step. Is this a pipeline problem?

bamonroe:
> @bamonroe it will run infinite time and just try out different inputs in a clever way to cover all parts of your code base. Not crashing is a good sign

OK great, then I guess it's working as intended :) Let me know if there are any other checks you'd like me to do.

bamonroe:
I'm not sure what to do here. I've been trying to get a Windows VM running to compile the code manually, but I'm totally unfamiliar with development on Windows and am struggling to get the various build-tools in order. @theHamsta do you have any ideas as to why it's failing that check? It appears to be passing the Windows GCC test just fine.

theHamsta:
We had this sometimes for certain parsers even when the would compile totally fine with Windows outside the CI. Can you add your parser to `ALLOWED_INSTALLATION_FAILURES` env variable for Windows?

bamonroe:
@theHamsta could you take a look and see if I've done this correctly? I just copied what appeared to the logic of the NVIM environment variable, I'm otherwise unfamiliar with github actions.

bamonroe:
Looks like all the checks have passed, is there anything else you'd like me to do to get this merged?

theHamsta:
@bamonroe can you please rebase to have a clean commit history.

bamonroe:
All set, no additional changes were made

theHamsta (approved):

bamonroe:
It looks like a check failed, but it's for jsondoc, which must be introduced by some other commit.

theHamsta:
json with comments and fusion are Gitlab hosted. Maybe they changed the URL paths there or maybe ALLOWED_INSTALLATION_FAILURES="" somehow influences the test

bamonroe:
I don't think it's the ALLOWED_INSTALLATION_FAILURES="rnoweb", I had included that when the previous checks were done and they all passed. The only difference between this check and the last was the rebase. I'll rebase on the master branch again in a couple days to see if some other commit fixes this issue.

theHamsta:
but why doesn't master then fail. I can check the Gitlab links this evening. Maybe they've changed something in their API

bamonroe:
@theHamsta any luck? Have you had a chance to look?

bamonroe:
Amazing! Thanks for all the help @theHamsta

Issue #2920, opened 2022-05-08
:TSToggle or TSDisable highlight enable default neovim highlights.


zorgick:
**Is your feature request related to a problem? Please describe.**

When `:TSToggle` or `:TSDisable` commands are used with **highlight** parameter, 'syntax' option is set to 'ON', enabling default neovim syntax highlighting. This disturbs me a lot, when I really need to switch off any highlighting at all (for fugitive diffs for example). So I always have to execute `:syntax off`, after disabling treesitter highlight.

**Describe the solution you'd like**
Can this behavior be disabled, so that all highlighting is switched off at once?

theHamsta:
At the moment this is not possible. This has probably be managed by language tree https://github.com/theHamsta/neovim/blob/e709f83889ad515359a30cd41a719bafc7852241/runtime/lua/vim/treesitter/highlighter.lua#L184 The language tree could try to restore the state `syntax` had before it got attached to the buffer.

The usecase to not syntax is probably at bit rare. Maybe it's not too much to require a user command that does `:TSDiable | set syntax off`

zorgick:
I've already tried to switch off all syntax with <BAR> commands, like `:TSDiable | set syntax off`, but it seems, that default syntax remains. So anyway I need to call to consecutive commands anyway. Do you use fugitive in your neovim setup? If so how do you manage highlighting with this plugin to see diffs clearly ?

zorgick:
Thank you! Checked it `:TSDiable | set syntax off`, everything works now

PR #2937, opened 2022-05-14
fix(commands): add -bar to allow other commands

kyazdani42:
addresses #2920


clason (approved):

theHamsta (approved):

theHamsta:
I must admit that I have absolutely no clue what bar is doing but I guess it's necessary to `|`. I don't want to know why this has to be explicit in vim.

clason:
Because how vim parses command arguments; the bar might be part of the arguments for the command.

theHamsta:
So we can no longer install the language `|`..

PR #2943, opened 2022-05-15
feat: add "fold_one_line_after" to allow folding without first line

theHamsta:
This allows to keep the current behavior of Python folding as an opt-in also for other languages.

grafik
grafik

multi-line function signature (still only keeping the first line with function name)
grafik
 

Addresses problems with https://github.com/nvim-treesitter/nvim-treesitter/pull/1451

This also changes our code to have "config" only modules. That can not be attached to buffers. This could also be used to expose the options that we have for `install.lua` via our setup call. I'm not sure whether this is the right way. The option could also be placed directly at `fold.lua` but I think it would be easier to document (same as with install.lua) when we would just show the options in the README in the setup call like options for other modules.

This is still kind of a hack, ideally Vim could just have an option to display the folds in a different way

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -48,6 +48,14 @@ local builtin_modules = {
     enable = false,
     is_supported = queries.has_indents,
   },
+  fold = {
+    module_path = "nvim-treesitter.fold",
+    do_not_attach = true,
+    fold_one_line_after = false,
+    is_supported = function(lang)
+      return queries.has_folds(lang)
+    end,
+  },
 }
 
 local attached_buffers_by_module = caching.create_buffer_cache()
@@ -56,7 +64,7 @@ local attached_buffers_by_module = caching.create_buffer_cache()
 local function resolve_module(mod_name)
   local config_mod = M.get_module(mod_name)
 
-  if not config_mod then
+  if not config_mod or config_mod.do_not_attach then
     return
   end
 
@@ -88,7 +96,9 @@ local function enable_module(mod, bufnr, lang)
     end
   end
 
-  M.attach_module(mod, bufnr, lang)
+  if not mod.do_not_attach then
+    M.attach_module(mod, bufnr, lang)
+  end
 end
 
 -- Enables autocomands for the module.
@@ -142,7 +152,9 @@ local function disable_module(mod, bufnr)
   if module.enabled_buffers then
     module.enabled_buffers[bufnr] = false
   end
-  M.detach_module(mod, bufnr)
+  if not mod.do_not_attach then
+    M.detach_module(mod, bufnr)
+  end
 end
 
 -- Disables autocomands for the module.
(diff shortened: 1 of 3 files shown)


theHamsta:
TODO: docs after we decided whether this option should be attach to module or lua file. Should this be a "fold style" option instead of boolean i.e. do we expect more variants in future?

lua/nvim-treesitter/configs.lua
```diff
@@ -48,6 +48,14 @@ local builtin_modules = {
     enable = false,
     is_supported = queries.has_indents,
   },
+  fold = {
+    module_path = "nvim-treesitter.fold",
+    can_attach = false,
```
kyazdani42, line 53: this can lead to issues when checking, could you reverse the logic ? Something like `do_not_attach` ?

kyazdani42 (commented):
i think its good otherwise, but forcing ~= false check can be error prone i believe since if we forget it, it will falsify every module.

theHamsta:
@kyazdani42 any alternative ideas on how to avoid adding more logic to non-attachable modules? Maybe just put the settings directly to the `fold.lua`

lua/nvim-treesitter/configs.lua
```diff
@@ -48,6 +48,14 @@ local builtin_modules = {
     enable = false,
     is_supported = queries.has_indents,
   },
+  fold = {
+    module_path = "nvim-treesitter.fold",
+    do_not_attach = true,
+    fold_one_line_after = false,
```
theHamsta, line 54: probably this should also allow a list/map of languages. So `true` or `{"python" =true, "cpp" = true}`

kyazdani42: i think maybe taking either a list or a boolean it would be better indeed.

kyazdani42 (approved):

stsewd:
So, excluding the first like isn't exactly the same as the previous behavior, the previous behavior was to fold only the body of the function/method, this won't work in functions like

```python
def foo(
   one,
   two,
):
    print("foo")
    print("foo")
    print("foo")
    print("bar")

```

theHamsta:
Well it would work, it would still show you the function signature which I would assume to have close to the folds. Is there any difference to `ident` folding then (without the fix to include on line before the indentation increase)? So that we would need to re-implement it?

In each case, I shouldn't use a bool as we want to add more behaviors later.

Hubro:
It would be great if there was another selector that only folds *inside* the matched block. For example, the Rust parser parses everything from `{` to `}` in a function declaration as `block`. This includes the opening and closing brackets.

Take this example:

image

Current behavior if I use `@fold` on `block`:

image

The behavior I would prefer:

image

Perhaps a name like `@fold_inset` would make sense for this?

theHamsta:
We thought about `@fold_inner`. Any reason why you would like to have the last `brace` not included into the fold given that solo `}` doesn't give much information. Or is this more about being symmetrical for the `{` that is present?

A problem with `@fold_inner` vs the one-line-after/one-line-before approach is that would probably fold like that
```
void main 
{
   ---
}
```
instead of 
```
```
void main 
{---
}
```

or 
```
void main 
{---
```

Hubro:
@theHamsta I guess it's mostly because I prefer the way it looks. Also, folds in Vim have a tendency to break while editing. This just looks like a broken fold to me:

image

And it would probably prompt me to press `zx` and `zX` a few times in an attempt to fix it. Seeing the opening `{` and closing `}` gives me confidence that the fold is still "correct".

clason:
superseded by upstream `vim.treesitter.foldexpr`

PR #2951, opened 2022-05-20
Add tree-sitter-rec for RecUtils

thmsmlr:
I recently put together a tree sitter implementation for [RecUtils](https://www.gnu.org/software/recutils/), I figure we could add it here for others to use

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1080,6 +1080,16 @@ list.v = {
   maintainers = { "@tami5" },
 }
 
+list.rec = {
+  install_info = {
+    url = "https://github.com/thmsmlr/tree-sitter-rec",
+    files = { "src/parser.c" },
+    branch = "main",
+  },
+  maintainers = { "@thmsmlr" },
+  filetype = "rec",
+}
+
 local M = {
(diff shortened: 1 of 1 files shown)


kyazdani42 (commented):
Would you mind adding a few queries for it ? Or will you do that later ?

lua/nvim-treesitter/parsers.lua
```diff
@@ -1068,6 +1068,15 @@ list.proto = {
   filetype = "proto",
 }
 
+list.rec = {
+  install_info = {
+    url = "https://github.com/thmsmlr/tree-sitter-rec",
+    files = { "src/parser.c" },
```
theHamsta, line 1074: you have to add `branch = "main"` to make it compile

theHamsta:
The grammar is very simple (https://github.com/thmsmlr/tree-sitter-rec/blob/main/grammar.js). Is this already supported by Vim's regex highlighting. With those few nodes, I suppose there would only be additional value when we have queries for that language

theHamsta:
Are there any plans to add queries or can we close?

clason:
> Are there any plans to add queries or can we close?

guess not... 😞

PR #2987, opened 2022-06-04
injections(ecma): add injection for commented graphql template strings

jacksonludwig:
Sometimes we want to highlight graphql strings without using a template literal like `gql`. In vscode, you can tell the editor to highlight the string as graphql by adding a comment like this:

```typescript
const query = /* GraphQL */ 
`
  type Thing {
    name: String!
  }
`;
```

This injection adds that capability.

Credit goes to @lukas-reineke. He gave me this injection snippet on reddit, but I thought it may as well be added to the repo.

### DIFF
--- a/queries/ecma/injections.scm
+++ b/queries/ecma/injections.scm
@@ -57,12 +57,6 @@
 
 (regex_pattern) @regex
 
-(variable_declarator
-  (comment) @comment
-  (#eq? @comment "/* GraphQL */")
-  (template_string) @graphql)
-
-(export_statement
-  (comment) @comment
-  (#eq? @comment "/* GraphQL */")
+((comment) @_gql_comment
+  (#eq? @_gql_comment "/* GraphQL */")
   (template_string) @graphql)
--- a/queries/ecma/injections.scm
+++ b/queries/ecma/injections.scm
@@ -60,3 +60,6 @@
 ((comment) @_gql_comment
   (#eq? @_gql_comment "/* GraphQL */")
   (template_string) @graphql)
+
+(((template_string) @_template_string
+ (#match? @_template_string "^`#graphql")) @graphql)
--- a/queries/ecma/injections.scm
+++ b/queries/ecma/injections.scm
@@ -56,3 +56,8 @@
    (#offset! @css 0 1 0 -1)))
 
 (regex_pattern) @regex
+
+(variable_declarator
+  (comment) @gql_comment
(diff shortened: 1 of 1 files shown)


theHamsta (approved):

jacksonludwig:
@theHamsta Sorry, but I do not know much about writing parsers. Should the capture be named just `@comment` or something like that? It seems to work for me but I see the CI does not like `@gql_comment` 😅

lukas-reineke:
> Should the capture be named just @comment or something like that? 

`@comment` has special meaning, to stay with the convention of the other ones it should probably be `@_name`

I think to add it here, you should also make it a bit more robust. I wrote the query for your specific example.
Like, does the `GraphQL` comment need to be case sensitive? Might be better to allow lowercase as well?
And does this have to be after a variable declaration only? Maybe there are other places where this should work as well?

I don't do much JavaScript anymore, I can't answer these questions.

lukas-reineke:
> And does this have to be after a variable declaration only?

I see you already also added `export_statement`. I think it should probably just work for every comment + template string pair when the comment matches.
You can update the query by just removing the first `variable_declarator` (but keep the `()`)

jacksonludwig:
@lukas-reineke thanks for your reply and sorry to bother you with this. Removing the `variable_declarator` so it applies to more cases makes sense.

In regards to your `@_name` tip, I am not totally clear. It seems that when I use a name like `@gql_comment`, the tests fail with `./scripts/check-queries.lua:84: (x) Invalid capture @gql_comment in injections for [tsx,js,ts,jsx].` Is there something I have to modify so the tests know the capture name is valid?

Thanks again.

lukas-reineke:
To check that the comment matches `GraphQL`, we need to capture it.
What you name this capture doesn't really matter as long as it is not used somewhere else. So in my example I used `gql_comment`.  
If you just put this in your config that would be fine, but here we don't want to do that, because someone else might use that name.  
Convention for these helper captures is they need to start with `_`. That is what the test is checking.  
You could use `@_gql_comment`, or anything else that starts with `_`. The other injection queries use `@_name`, so I suggested that.

jacksonludwig:
I understand now, thanks for the explanation.

jacksonludwig:
I've also added the following case:
```typescript
const myQuery = `#graphql
  query {
    something
  }
`;
```

kyazdani42 (approved):

PR #3025, opened 2022-06-15
parsers: add HLSL

theHamsta:
WIP I have little experience with HLSL. So I'm exploring the features for while...


theHamsta:
I think this is ready. Just fixed a major problem with the parser. It now parses the shaders of https://github.com/EmbarkStudios/kajiya fine except for macro constructs (which currently also fail for C/C++)

theHamsta:
I might add legacy effect syntax or OBS syntax at some point

stsewd (approved):

Issue #2959, opened 2022-05-23
Add support for custom parser install locations


dncnmcdougall:
nvim-treesitter provides excellent convenience for managing the installing of both treesitter parsers and modules that use them.
I want to use the parsers in projects outside of nvim, but the default install path is not ideal (and changes based on you r package manager.)

It would be nice to have a configuration option which is a path that the parsers will be installed to. 
Something like:
```lua
require'nvim-treesitter.configs'.setup {
  -- A list of parser names, or "all"
  ensure_installed = { "c", "lua", "rust" },

  -- Install parsers synchronously (only applied to `ensure_installed`)
  sync_install = false,

  -- List of parsers to ignore installing (for "all")
  ignore_install = { "javascript" },


  -- Path in which to install parsers. Default: "${plugin_dir}/parser"
  parser_install_path= "~/treesitter/parsers"

-- ...
}
```

I have not seriously thought about the details of the option. Like what it should be called, and how much power it needs.
Should it be able to expand environment variables? Expand the home "~", etc.
Also should it be set to a default value, or should it default to ```lua nil```, which means install in the plugin dir?

dncnmcdougall:
I am also happy to try and implement a solution for this issue, but would value some starting guidance.

kyazdani42:
I think this might be a problem because neovims `get_parser` needs the parsers to be installed in the runtimepath i believe.
@theHamsta do you think it would still be feasible ? If so, i'm not sure it adds to much value to be honest.
@dncnmcdougall in the meantime, would a symlink do the trick ?

dncnmcdougall:
Hi @kyazdani42, thanks for the reply.
Presumably nvim-treesitter adds the current parsers directory to the runtime path? Is there a reason it could not add an arbitrary directory instead?

AS to symlinks, I had not thought of that. In the mean time I hard coded the parser directory into the script I was writing. This feels similar, and works, but is a little hard to automate when moving to a new computer or reinstalling a system.

kyazdani42:
I'll be waiting on @clason and @theHamsta suggestion about this then

dncnmcdougall:
Thanks. I am happy to do the work on this, I think it will be a relatively gentle intro into the workings of all the bits and pieces.

clason:
[...]

PR #3031, opened 2022-06-17
 Add support for custom parser install locations #2959 

dncnmcdougall:
Moves get_parser_install_dir into configs out of utils and adds a config option to satisfy  [#2959 ](https://github.com/nvim-treesitter/nvim-treesitter/issues/2959)

Previously `get_parser_install_dir` and `get_parser_info_dir` were both in
utils. This is because they only took into the package dir and the site
dir.
Now there is a configuration option, `parser_install_dir` that takes
precedence over the package and site directories. This is parsed in with
``` lua
configs.setup(
{
    parser_install_dir = "some/path",
})
```

As such the `get_parser_install_dir` needs to take this option in. One
was is to leave it in usils and parse it in:
```lua
utils.get_parser_install_dir("parser", configs.get_parser_install_dir_option())
```
But that will always require access to configs, or will work
inconsistently. So it was easier to move the whole setting into configs:
```lua
configs.get_parser_install_dir("parser")
```

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -544,10 +544,11 @@ function M.get_parser_install_dir(folder_name)
 
   if config.parser_install_dir then
     local parser_dir = utils.join_path(config.parser_install_dir, folder_name)
-    return utils.create_or_resue_writable_dir(parser_dir, 
-            utils.join_space("Could not create parser dir '",dir,"': "), 
-            utils.join_space("Parser dir '",dir,"' should be read/write.")
-        )
+    return utils.create_or_resue_writable_dir(
+      parser_dir,
+      utils.join_space("Could not create parser dir '", parser_dir, "': "),
+      utils.join_space("Parser dir '", parser_dir, "' should be read/write.")
+    )
   end
 
   local package_path = utils.get_package_path()
@@ -561,7 +562,11 @@ function M.get_parser_install_dir(folder_name)
   local site_dir = utils.get_site_dir()
   local parser_dir = utils.join_path(site_dir, folder_name)
 
-  return utils.create_or_resue_writable_dir(parser_dir, nil, utils.join_space("Invalid rights,", package_path, "or", parser_dir, "should be read/write")))
+  return utils.create_or_resue_writable_dir(
+    parser_dir,
+    nil,
+    utils.join_space("Invalid rights,", package_path, "or", parser_dir, "should be read/write")
+  )
 end
 
 function M.get_parser_info_dir(parser_install_dir)
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -15,7 +15,7 @@ local config = {
   ensure_installed = {},
   ignore_install = {},
   update_strategy = "lockfile",
-  parser_install_dir = nil
+  parser_install_dir = nil,
 }
 -- List of modules that need to be setup on initialization.
 local queued_modules_defs = {}
@@ -384,9 +384,8 @@ function M.setup(user_data)
   config.parser_install_dir = user_data.parser_install_dir or nil
   if config.parser_install_dir then
     config.parser_install_dir = vim.fn.expand(config.parser_install_dir, ":p")
-    vim.cmd("set runtimepath+="..config.parser_install_dir)
+    vim.cmd("set runtimepath+=" .. config.parser_install_dir)
   end
-  
 
   local ensure_installed = user_data.ensure_installed or {}
   if #ensure_installed > 0 then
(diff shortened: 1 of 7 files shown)


clason:
Can't the code be further refactored to directly access this option, same as `ensure_installed` and friends?

I am strongly in favor of removing the current machinery for working around a nixity in favor of having to explicitly specify the directory.

clason:
Also something to consider: _not_ add the specified directory to `rtp` automatically, and instead document that the specified directory should be in `rtp` (e.g., under `.local/site/nvim`) or added to it manually. Would again simplify things and avoid the danger of weird edge cases.

dncnmcdougall:
I don't follow what you mean by "to directly access this option, same as `ensure_installed`".
As far as I can tell `ensure_installed` only exists in the `setup` function. It is not kept in memory or used outside of `setup`. 
This is different to `parser_indtall_dir`, which is needed for explicit installing, updating and removing. 
Interestingly it is not necessary outside of `setup` for checking what is installed, as that directly queries the runtimepath.

One way to _simplify_ the code would be to try the package dir, or nix dir in `setup` if `parser_install_dir` is `nil`. That would make the `get_parser_install_dir` simpler.

Are these inline with what you were thinking?

dncnmcdougall:
> Also something to consider: _not_ add the specified directory to `rtp` automatically, and instead document that the specified directory should be in `rtp` (e.g., under `.local/site/nvim`) or added to it manually. Would again simplify things and avoid the danger of weird edge cases.

But would that not reduce the user ergonomics? 
I want to add this option so that the parsers are in a custom directory. If this plugin does not add that directory to the runtime path I would add it myself (after being confused why the plugin installs them like I expect, but does not find them afterwards. I just fought with this during development.)
If they are not in the `rtp` they are not installed.

But that does raise the question: what happens if you specify a custom directory inside the `rtp`? Can cim handel that?

clason:
> As far as I can tell ensure_installed only exists in the setup function. It is not kept in memory or used outside of setup.

Yes, it is accessed in `setup`, then passed around explicitly through the `nvim-treesitter.install` functions. (Not necessarily the best pattern, mind you, but it does the job.) We could do the same for the `parser_install_dir`.

> One way to simplify the code would be to try the package dir, or nix dir in setup if parser_install_dir is nil. That would make the get_parser_install_dir simpler.

I would even drop the nix dir. Either we can use the package dir (or the site dir; I don't care which, but there should only be _one_ default directory), or a user has to specify a different one that they know works for them.

clason:
> But that does raise the question: what happens if you specify a custom directory inside the rtp? Can cim handel that?

Of course, that's what `rtp` is for ;) It is a ergonomic regression for _certain_ (not all!) custom directories, at the cost of explicitness. (Basically, if something goes wrong, it is _much_ harder to undo that addition. So even if those are fewer cases, the cost-benefit analysis still may pan out. Or may not; that's why I want to discuss this here!)

dncnmcdougall:
OK so there are three things to discuss:
1. Should we simplify code by making the check for directory existence happen in `setup` rather than in `configs.get_parser_install_dir`
2. Simplify code by reducing the number of default directories.
3. Make the directory an explicit option into `install` and `uninstall`, which affects `ensure_installed`, `update` and a bunch of commands.
4. Decide whether to automatically add the custom path to the `runtimepath`.

For 1. I don't mind making that change.

For 2. (changing the number of default directories) I'd prefer not to change the current behaviour, as that becomes a much larger discussion as it affects a lot more people.

For 3. The way I would do this is by including the dir in the options used to construct the install and update functions (and introduce this pattern of options for `uninstall`.) This ends up being a meaningless change as that would just call the `configs.get_parser_install_dir`
An alternative is to parse it in at call time (rather than construction time.) This is difficult as all the functions are designed to take a varargs list of parser names. So we'd need to change the api quite a bit so that the first argument is the directory. 
This alternative raises the question of how it is passed in for the commands `:TSUpdate` and co.
I personally think that hiding the path behind `configs.get_parser_install_dir` is a simple way to do this. And I think that a user would expect the commands `:TSUpdate` and co to respect the option set in `setup`

For 4. the question of  `runtimepath`, I suppose I am easy either way (provided there is sufficient documentation), but would prefer automatically adding the custom path to the runtime pathas that feels to me to be inside the scope of managing the installation of parsers.

clason:
Thank you for this analysis! 

1. is moot based on your response to 3.
2. is a decision the maintainers need to make. I _really_ believe this would be a positive change even if technically breaking, given how many issues we had that ended up being related to outdated parsers silently remaining installed in this "fallback" directory. It is not unrelated to this PR, since this change _allows_ removing the fallback.
3. It sounds like the current approach is indeed the simplest one, so let's stick with that 👍 
4. My worry is about what happens when the directory is _already_ in runtimepath. We also need to think about precedence (_where_ should it be in `runtimepath`? before or after other directories that potentially include parsers?) We also shouldn't use `vim.cmd`; we can and should use `vim.opt.runtimepath:append()` or  `prepend`.

lua/nvim-treesitter/utils.lua
```diff
@@ -65,25 +86,25 @@ function M.setup_commands(mod, commands)
   end
 end
 
-function M.get_path_sep()
-  return fn.has "win32" == 1 and "\\" or "/"
-end
+function M.create_or_resue_writable_dir(dir)
```
theHamsta, line 89: I think the name of the is function is a bit unfortunate. It reads like a general function, but the error message still point to the usage for creating the parser dir. You could either stay with this function and give it  a new name that's exclusive for creating the parser dir or generalize the error messages either by making the strings more general, provide the error messages as arguments or return error objects and let the caller create the error.

Apart from this, this change looks good to me.

lua/nvim-treesitter/configs.lua
```diff
@@ -528,6 +534,42 @@ function M.init()
   end
 end
 
+-- If parser_install_dir is not nil is used or created.
+-- If parser_install_dir is nil try the package dir of the nvim-treesitter
+-- plugin first, followed by the "site" dir from "runtimepath". "site" dir will
+-- be created if it doesn't exist. Using only the package dir won't work when
+-- the plugin is installed with Nix, since the "/nix/store" is read-only.
```
theHamsta, line 541: unrelated to this PR: did somebody plan to create a Nix package for the parsers. We could provide the install info as JSON and installing via a nix flake would be the most natural thing to do for Nix users. Especially when they can profit from binary caching.

lua/nvim-treesitter/utils.lua
dncnmcdougall: Thank you, I missed that. I have changed the function to have error messages parsed in, with sensible defaults.

theHamsta (approved):
I think this is ready. @clason what do you think?

I tested that everything works as expected.

lua/nvim-treesitter/configs.lua
clason: I think Nix should take care of Nix users. I am _very_ opposed to catering to a single distribution that insists to be different (for whatever good reason they have, but that is on them, not us).

clason:
I'm not completely adverse to including the `rtp` modification here; just gave my personal preference.

But let's merge it as-is, and follow up in a future PR.

👍

lua/nvim-treesitter/configs.lua
theHamsta: But this is what I was saying. That Nix users should install it the nix way their own nix flake. I was asking whether there are any known efforts.

theHamsta:
@dncnmcdougall do you want to create a PR for @clason 's suggestion?

dncnmcdougall:
Hi @theHamsta and @clason, thank you for your comments and review.

I am probably happy to make another PR for the suggestions, but I am not quite clear what they are?
Are they the comments to dealing with installed packages in a more predictable way?
As suggested by only having the explicit directory? That is: `parser_install_dir` must always be set? 

More predictable:
1. Communicating when and where things are installed.
2. Communicating when and why an install or uninstall failed.
3. Making the install location more predictable.

Have I understood the comments/requests at all?

PR #3035, opened 2022-06-19
feat(swift): update parser and queries to 0.3.0

clason:
I see that our queries are almost-but-not-completely the same as the upstream ones. Is the divergence intentional, or should/can I just pull in the latest upstream version?

@alex-pinkus @theHamsta

### DIFF
--- a/queries/swift/highlights.scm
+++ b/queries/swift/highlights.scm
@@ -21,7 +21,7 @@
 (function_declaration (simple_identifier) @method)
 (function_declaration ["init" @constructor])
 (throws) @keyword
-(async) @keyword
+"async" @keyword
 (where_keyword) @keyword
 (parameter external_name: (simple_identifier) @parameter)
 (parameter name: (simple_identifier) @parameter)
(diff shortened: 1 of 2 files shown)


clason:
...which I just did, as a separate commit to be squashed so you can look at the non-necessary changes.

alex-pinkus:
I think those are from pull requests that happened upstream but never got incorporated here. E.g. https://github.com/alex-pinkus/tree-sitter-swift/pull/135 — should be safe to pull in here

clason:
Thanks! Having a versioned upstream that provides usable queries makes updates easy. 

(It would be good to document this somewhere -- which parsers use upstream queries and which don't, so we don't duplicate efforts and can prevent divergence by pointing PRs to them upstream.)

queries/swift/highlights.scm
```diff
@@ -3,7 +3,6 @@
 
 ; Identifiers
 (attribute) @variable
-(simple_identifier) @variable
```
theHamsta, line 6: Any reason to remove this?

theHamsta (approved):

queries/swift/highlights.scm
clason: To reflect upstream changes; I believe it's replaced by 
```
(pattern bound_identifier: (simple_identifier)) @variable
```
further down.

alex-pinkus: In https://github.com/alex-pinkus/tree-sitter-swift/pull/135, the PR author complained that it was “too broad, it overlaps with many other patterns.” I’m not sure why that was causing a problem but it seemed fine to me to narrow it.

Issue #2212, opened 2022-01-07
Injected markdown highlights indented docstrings


bockel:
### Before reporting

- [X] I have updated my neovim version to latest _master_
- [X] I have updated my plugin to the latest version
- [X] I have run `:TSUpdate`
- [X] I have read the [troubleshooting section](https://github.com/nvim-treesitter/nvim-treesitter#troubleshooting)

### Describe the bug

When injecting markdown into languages with indented docstrings such as python, the leading indentation of the docstring is interpreted as markdown.

For example:

```python
class Example:
    """
    Misinterpreted as markdown `indented_code_block`
    """
```

While

```python
class Example:
    """Correctly interpreted as markdown `codespan`"""
```

### To Reproduce

1. Set `after/queries/python/injections.scm` with the following

    ; Module docstring
    ((module . (expression_statement (string) @markdown))
      (#offset! @markdown 0 3 0 -3))

2. Open a new python file with the content

```python
class Example:
    """
    Misinterpreted as markdown **bold** and `indented_code_block`
    """
```

3. View the `:TSPlaygroundToggle`

```
class_definition [60, 0] - [382, 87]
  name: identifier [60, 6] - [60, 10]
  body: block [61, 4] - [382, 87]
    expression_statement [61, 4] - [63, 7]
      string [61, 4] - [63, 7]
        indented_code_block [62, 0] - [63, 4]
```

### Expected behavior

You should see the following TS syntax:

```
class_definition [60, 0] - [382, 87]
  name: identifier [60, 6] - [60, 10]
  body: block [61, 4] - [382, 87]
    expression_statement [61, 4] - [63, 7]
      string [61, 4] - [63, 7]
        paragraph [62, 0] - [63, 0]
          strong_emphasis [62, 31] - [62, 39]
          code_span [62, 44] - [62, 65]
```

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found  0.20.2 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v17.3.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: Apple clang version 13.0.0 (clang-1300.0.29.30)
[...]

PR #3048, opened 2022-06-22
Switch to split markdown parser

MDeiml:
My new split markdown parser works quite well now (better than the non-split one). This PR changes the config to point at the new branch because I want to keep the old branch for now as other projects are using it.

It also adds some other changes that are needed for the two parsers to work:
1. An `exclude_children!` directive, that might also be very useful for something like #2212 
2. Splitting markdown queries into block and inline ones and add the injection for inline into block grammar
3. Add an `include_dir` option to parser configs (needed, because the two grammars don't live in the repos root directory)

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -192,7 +192,10 @@
     "revision": "a4b9187417d6be349ee5fd4b6e77b4172c6827dd"
   },
   "markdown": {
-    "revision": "be3e08acfd85bd87d85f41fde74fdcec25f76dbe"
+    "revision": "acb097808683eea7c6d6d469644a275b9f6a64f5"
+  },
+  "markdown_inline": {
+    "revision": "acb097808683eea7c6d6d469644a275b9f6a64f5"
   },
   "ninja": {
     "revision": "0a95cfdc0745b6ae82f60d3a339b37f19b7b9267"
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -528,8 +528,19 @@ list.hcl = {
 list.markdown = {
   install_info = {
     url = "https://github.com/MDeiml/tree-sitter-markdown",
+    location = "tree-sitter-markdown",
     files = { "src/parser.c", "src/scanner.cc" },
-    branch = "main",
+    branch = "split_parser",
+  },
+  experimental = true,
+}
+
+list.markdown_inline = {
+  install_info = {
+    url = "https://github.com/MDeiml/tree-sitter-markdown",
+    location = "tree-sitter-markdown-inline",
+    files = { "src/parser.c", "src/scanner.cc" },
+    branch = "split_parser",
   },
   experimental = true,
 }
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -532,6 +532,7 @@ list.markdown = {
     files = { "src/parser.c", "src/scanner.cc" },
     branch = "split_parser",
   },
+  maintainers = { "@MDeiml" },
   experimental = true,
 }
 
@@ -542,6 +543,7 @@ list.markdown_inline = {
     files = { "src/parser.c", "src/scanner.cc" },
     branch = "split_parser",
   },
+  maintainers = { "@MDeiml" },
   experimental = true,
 }
 
--- a/queries/markdown_inline/highlights.scm
+++ b/queries/markdown_inline/highlights.scm
@@ -1,0 +1,67 @@
+;; From MDeiml/tree-sitter-markdown
+[
+  (code_span)
+  (link_title)
+] @text.literal
+
+[
+  (emphasis_delimiter)
+  (code_span_delimiter)
+] @punctuation.delimiter
+
+(emphasis) @text.emphasis
+
+(strong_emphasis) @text.strong
+
+[
+  (link_destination)
+  (uri_autolink)
+] @text.uri
+
+[
+  (link_label)
+  (link_text)
+  (image_description)
+] @text.reference
+
+[
+  (backslash_escape)
+  (hard_line_break)
+] @string.escape
+
+; "(" not part of query because of
+; https://github.com/nvim-treesitter/nvim-treesitter/issues/2206
+; TODO: Find better fix for this
+(image ["!" "[" "]" "("] @punctuation.delimiter)
+(inline_link ["[" "]" "("] @punctuation.delimiter)
+(shortcut_link ["[" "]"] @punctuation.delimiter)
+
+([
+  (code_span_delimiter)
+  (emphasis_delimiter)
+] @conceal
+(#set! conceal ""))
+
+; Conceal inline links
+(inline_link
+  [
+    "["
+    "]"
+    "("
+   (link_destination)
+    ")"
+] @conceal
+(#set! conceal ""))
+
+
+; Conceal image links
+(image
+  [
+    "!"
+    "["
+    "]"
+    "("
+   (link_destination)
+    ")"
+] @conceal
+  (#set! conceal ""))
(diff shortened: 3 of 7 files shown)


clason:
Also, please don't forget to mark this as a breaking change (`feat(markdown)!: switch to split parser`) as this will invalidate downstream queries.

MDeiml:
I also noticed a bug (which probably should be solved separately from this PR): A node from an injected language used for highlighting can have a range that might include bytes that lie outside of the included range. E.g. with this
```
> *Foo
> Bar*
```
The emphasis spans over the second `>`, but that one is not part of the included range for the inline content. It is still highlighted as emphasis.

We would need to calculate the intersection between included range and node range for every node used in an injected language and use that instead.

theHamsta:
@MDeiml wow, awesome that we now have an `exclude_children!` directive. We can of course merge it. But I'm wondering whether we should just make it always the default as this is the way upstream tree-sitter behaves (except "injection.combined" is set https://tree-sitter.github.io/tree-sitter/syntax-highlighting#language-injection).

MDeiml:
I also think excluding children is the better default, but making this the default would probably need changes (and some discussion) in neovim itself. Maybe it's better to first test this directive here and later make a PR to neovim?

lua/nvim-treesitter/parsers.lua
```diff
@@ -528,9 +528,22 @@ list.hcl = {
 list.markdown = {
   install_info = {
     url = "https://github.com/MDeiml/tree-sitter-markdown",
-    files = { "src/parser.c", "src/scanner.cc" },
-    branch = "main",
+    include_dir = "tree-sitter-markdown/src",
+    files = { "tree-sitter-markdown/src/parser.c", "tree-sitter-markdown/src/scanner.cc" },
```
theHamsta, line 532: `v`, `tsx`, `typescript`, `ocaml` `ocaml_interface` solve this using the `location` property. Can you either port the old ones to the `include_dir` or use `location`?

clason: (with preference for the latter, if feasible ;))

lockfile.json
```diff
@@ -191,9 +191,6 @@
   "make": {
     "revision": "a4b9187417d6be349ee5fd4b6e77b4172c6827dd"
   },
-  "markdown": {
-    "revision": "ab15701d8f3f68aeb74e30573b7d669a6ef2a7ed"
-  },
```
clason, line 196: Don't need to remove this; the "Update Lockfile" PR will take care of that (I think), and then you don't have to worry about conflicts or constantly rebasing this file.

theHamsta: but the SHA has to be changed in case the new queries are incompatible with the old parser

clason: ...which they are

but wouldn't this only be a problem for the short window between merging this and the subsequent Update Lockfile PR?

MDeiml: > but the SHA has to be changed in case the new queries are incompatible with the old parser

Should I add this back then? Or change it to the new has?

clason: Ideally, yes, always adapt the lockfile to parser changes -- saves PR churn, if nothing else.

lua/nvim-treesitter/parsers.lua
```diff
@@ -528,9 +528,22 @@ list.hcl = {
...
+    branch = "split_parser",
+  },
+  maintainers = { "@MDeiml" },
+  experimental = true,
+}
+
+list.markdown_inline = {
+  install_info = {
+    url = "https://github.com/MDeiml/tree-sitter-markdown",
+    location = "tree-sitter-markdown-inline",
```
theHamsta, line 542:
```suggestion
    location = "tree-sitter-markdown/tree-sitter-markdown-inline",
```

lua/nvim-treesitter/parsers.lua
```diff
@@ -528,9 +528,22 @@ list.hcl = {
 list.markdown = {
   install_info = {
     url = "https://github.com/MDeiml/tree-sitter-markdown",
+    location = "tree-sitter-markdown",
```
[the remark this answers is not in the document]
theHamsta: yeah, `location` is a bit wonky

lua/nvim-treesitter/parsers.lua
theHamsta: we wanted to trim the unnecessary prefix from `location` anyway. It's just confusing...

queries/markdown_inline/highlights.scm
```diff
@@ -0,0 +1,74 @@
...
+(image "!" @punctuation.delimiter)
+(image "[" @punctuation.delimiter)
+(image "]" @punctuation.delimiter)
+(image "(" @punctuation.delimiter)
+; (image ")" @punctuation.delimiter)
+
+(inline_link "[" @punctuation.delimiter)
+(inline_link "]" @punctuation.delimiter)
+(inline_link "(" @punctuation.delimiter)
+; (inline_link ")" @punctuation.delimiter)
```
kyazdani42, line 41: why is this commented ?
also i think you can do:
```
(inline_link [ "[" "]" "(" ")" ] @punctuation.delimiter)
```

kyazdani42 (approved):
looks good to me otherwise :)

lua/nvim-treesitter/parsers.lua
theHamsta: @MDeiml tried to fix this here https://github.com/nvim-treesitter/nvim-treesitter/pull/3067

lua/nvim-treesitter/parsers.lua
```diff
@@ -528,9 +528,22 @@ list.hcl = {
...
+    branch = "split_parser",
+  },
+  maintainers = { "@MDeiml" },
+  experimental = true,
+}
+
+list.markdown_inline = {
+  install_info = {
+    url = "https://github.com/MDeiml/tree-sitter-markdown",
+    location = "tree-sitter-markdown_inline/tree-sitter-markdown-inline",
```
theHamsta, line 542: @MDeiml we removed the weird prefix here https://github.com/nvim-treesitter/nvim-treesitter/pull/3048 can you also remove it here? I think it was very intuitive why it had to be `tree-sitter-markdown_inline`

lockfile.json
```diff
@@ -192,7 +192,10 @@
     "revision": "a4b9187417d6be349ee5fd4b6e77b4172c6827dd"
   },
   "markdown": {
-    "revision": "be3e08acfd85bd87d85f41fde74fdcec25f76dbe"
+    "revision": "acb097808683eea7c6d6d469644a275b9f6a64f5",
+  },
+  "markdown_inline": {
+    "revision": "acb097808683eea7c6d6d469644a275b9f6a64f5",
```
theHamsta, line 198: JSON does not allow this comma. Same with `markdown`

theHamsta:
I'm having similar performance issues as with the old parser. Was the intention of the split to improve performance?

queries/markdown_inline/highlights.scm
MDeiml: I just took this from the old markdown queries without changing it. I think it was commented because there was a bug where tree-sitter crashed and it was determined that commenting this out would solve that, but I'm not sure it's still relevant.

I'm gonna change to your more concise version though.

MDeiml:
It's a bit faster, but still slow. I think parsing time after edits is still linear with file size (should be linear with size of edit), because all inline ranges with more complicated elements like emphasis or code spans still get reparsed. This could be fixed, by detecting which inline ranges actually got edited and then just reusing the old trees for all the other ones. But that is something that tree-sitter itself should be able to detect, so maybe I'm gonna create a PR in for that in tree-sitter.

clason:
Just on the off-chance: will https://github.com/tree-sitter/tree-sitter/pull/1783 be of any help?

MDeiml:
I don't think so. There is no "syntactically wrong markdown", so error recovery is not relevant.

Issue #3056, opened 2022-06-23
[Haskell] Outermost function when using $ operator isn't parsed as a function


pyjamafish:
### Describe the highlighting problem

In Haskell, when I have multiple functions chained together using the $ operator, the outermost function doesn't get highlighted as a `TSFunction`.

### Example snippet that causes the problem

```haskell
exactMatches :: Code -> Code -> Int
exactMatches actualPegs guessPegs =
    length $ filter foundMatch $ zip actualPegs guessPegs where
        foundMatch :: (Peg, Peg) -> Bool
        foundMatch (actual, guess) = actual == guess
```

### Tree-sitter parsing result

![image](https://user-images.githubusercontent.com/29507110/175365484-0521a577-a935-407c-ba84-d274ed36e52e.png)

### Example screenshot

![image](https://user-images.githubusercontent.com/29507110/175365578-14396254-e33e-4fbb-b860-013fa1a7fc96.png)

colorscheme: [onedark.nvim](https://github.com/navarasu/onedark.nvim), installed with Packer

### Expected behavior

The outer function should be highlighted as a `TSFunction`, just like how it is when I use parentheses instead of $.
![image](https://user-images.githubusercontent.com/29507110/175366214-ac287b78-6fb6-42d8-9275-e0d3dbd8e4cc.png)

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
  - WARNING: `node` executable not found (only needed for :TSInstallFromGrammar, not required for :TSInstall)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: gcc (GCC) 11.2.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - dockerfile     ✓ . . . ✓ 
  - typescript     ✓ ✓ ✓ ✓ ✓ 
  - dot            ✓ . . . ✓ 
  - hjson          ✓ ✓ ✓ ✓ ✓ 
  - rust           ✓ ✓ ✓ ✓ ✓ 
  - scala          ✓ . ✓ . ✓ 
[...]

PR #3066, opened 2022-06-25
fix(haskell): highlight exp_name as function in infix operations

kyazdani42:
fixes https://github.com/nvim-treesitter/nvim-treesitter/issues/3056, although i'm not very confident this is the right change

### DIFF
--- a/queries/haskell/highlights.scm
+++ b/queries/haskell/highlights.scm
@@ -116,6 +116,7 @@
 
 (exp_infix (variable) @operator)  ; consider infix functions as operators
 
+(exp_infix (exp_name) @function)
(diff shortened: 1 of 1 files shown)


theHamsta (approved):

theHamsta:
I'm not very confident to judge about this either do you want to ask for a third opinion or just merge?

clason:
YOLO

clason:
(Haskell is listed as "unmaintained", so it's clearly a best effort by rank amateurs. People will be sure to let us know if we got it wrong...)

kyazdani42:
Okay then :)

PR #3067, opened 2022-06-25
Chore: remove prefix from parser "location" property

theHamsta:
This was very confusing when using the location property as it
always had to start with "tree-sitter-<lang>" for no real reason (just
because our installer would rename the download to that name).


kyazdani42 (approved):
indeed

PR #3078, opened 2022-06-28
fix(beancount): support folding org headers

polarmutex:
change folding queries to support folding org headers for beancount files


clason:
@polarmutex Thank you! I overlooked this PR since the description didn't indicate it was related to the parser bump! (Hint, hint ;))

Also, you always need to bump the parser (update the lockfile) and change the queries/parser settings in parallel, otherwise CI will block the updates (and things need to be messily sorted out manually).

PR #3096, opened 2022-06-30
highlights(c): highlight member-access as operator

guijan:
2022-06-30-145853_826x219_scrot

Here's how a C standard draft refers to `.` in various places. I've
replaced the irrelevant parts with (...):

https://www.iso-9899.info/n1570.html#6.6p9
>(...) member-access . and -> operators (...)

https://www.iso-9899.info/n1570.html#6.4.6p1
>punctuator: one of (...) . (...)

https://www.iso-9899.info/n1570.html#6.4.6p2
>A punctuator (...) may specify an operation to be performed (...)
>in which case it is known as an operator (...)

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -63,6 +63,7 @@
   ">>"
 
   "->"
+  "."
 
   "<"
   "<="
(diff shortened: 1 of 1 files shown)


vigoux (approved):
I agree that is makes more sense to have `->` and `.` be highlighted the same.
Thanks for the changes.

PR #3107, opened 2022-07-02
highlights(c): highlight `,` operator as operator

guijan:
Test program:
2022-07-02-164250_761x430_scrot

Comma punctuator highlight group:
2022-07-02-161805_955x140_scrot

Comma operator highlight group:
2022-07-02-161413_954x194_scrot

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -46,6 +46,12 @@
 
 "#include" @include
 
+[ ";" ":" "," ] @punctuation.delimiter
+
+"..." @punctuation.special
+
+[ "(" ")" "[" "]" "{" "}"] @punctuation.bracket
+
 [
   "="
 
@@ -90,19 +96,16 @@
   "++"
 ] @operator
 
+;; Make sure the comma operator is given a highlight group after the comma
+;; punctuator so the operator is highlighted properly.
+(comma_expression [ "," ] @operator)
+
 [
  (true)
  (false)
 ] @boolean
 
-[ ";" ":" "," ] @punctuation.delimiter
-
-"..." @punctuation.special
-
 (conditional_expression [ "?" ":" ] @conditional)
-
-
-[ "(" ")" "[" "]" "{" "}"] @punctuation.bracket
 
 (string_literal) @string
 (system_lib_string) @string


clason:
I've _never_ seen `,` classified as an operator, and this makes zero sense to me -- even in the example you show. Can you link to some part of the standards that would justify this?

theHamsta:
Well, in theory you can overload `operator,` but I would prefer not to have it for syntax highlighting where we have the punctuation capture precisely for `,` `;` `.`. Also I guess`,` is on an operator in some contexts and the overloading of `,` is deprecated in c++20 because to allow proper `[x,y,z]`.

guijan:
> I've _never_ seen `,` classified as an operator, and this makes zero sense to me -- even in the example you show. Can you link to some part of the standards that would justify this?

https://www.iso-9899.info/n1570.html#6.5.17
It's an operator that is not often used, but an operator. Note that `,` is only an operator inside an expression, the comma that separates the parameter or argument lists of a function isn't an operator. This change only places the comma operator in the @operator group, the comma punctuator remains the same.

> Well, in theory you can overload `operator,` but I would prefer not to have it for syntax highlighting where we have the punctuation capture precisely for `,` `;` `.`. Also I guess`,` is on an operator in some contexts and the overloading of `,` is deprecated in c++20 because to allow proper `[x,y,z]` indexing in future.

This is the C highlighter, not the C++ highlighter. The 2 languages are full of those subtle differences.

> I think we can remove the highlighting of operator `,` in case that the second screenshot shows the current highlighting. 

All 3 screenshots in the original post show the change made by this PR.

>Would be good if you have followed the highlight-bug template as we would know what the current parsing result is and what is current behavior and what expected behavior

I'll open an issue through the highlight-bug template soon then, I'm cloning the latest neovim right now.

theHamsta:
> https://www.iso-9899.info/n1570.html#6.5.17
It's an operator that is not often used, but an operator

How something is called is not as relevant as what user would expect or how the concept is used in comparison to other languages.


> This is the C highlighter, not the C++ highlighter. The 2 languages are full of those subtle differences.

The parsers for C and C++ share most of the code. Also the highlights you wrote will be applied for both languages.

 My error that this a PR and not an issue. It's easy to oversee that on my phone.

For me, this change would be ok. But I'm quite neutral about it. Would like to see @vigoux 's review.

theHamsta (approved):

alemuller:
Nice catch. Some nitpicking:

```scm
;; Make sure the comma operator is given a highlight group after the comma
;; punctuator so the operator is highlighted properly.
```

In C, the coma is always a *punctuator* (this a term used to classify tokens), regardless if it is used as operator or as a delimiter/separator.

Also, the scheme defines *captures*, not *highlight group*.

By the way, someone probably should add `field("operator",",")` in `comma_expression` to the tree-sitter-c grammar.js to match others operators.

guijan:
> In C, the coma is always a _punctuator_ (this a term used to classify tokens) regardless if it is used as operator or as a delimiter/separator.
> 
> Also, the scheme defines _captures_, not _highlight group_.

Would this be a more appropriate comment?
```scheme
;; Make sure the comma punctuator as an operator is captured after the comma 
;; punctuator capture so the operator is highlighted properly.
```

kyazdani42 (approved):
I don't feel strongly about this change, lets merge this, we should not be picky about these kind of changes anyway.

Issue #3120, opened 2022-07-05
Bad JS indenting


ignamartinoli:
### Describe the bug

After writing this code

```javasctipt
class Maybe {
	constructor(value) {
		this.value = value;
	}

	bind = function(func) {
		value = func(this.value);

		return Maybe(value);
	}
}
```

if I hit `<Enter>` at the bottom, the cursor is placed a level of indentation further like this

```javascript
		return Maybe(value);
	}
}
	|
```

### To Reproduce

1. Do the mentioned above action

### Expected behavior

The cursor on the first character of the new line

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
  - OK: `node` found v18.4.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (Gentoo 12.1.1_p20220625 p8) 12.1.1 20220625
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - erlang         . . . . . 
  - fennel         ✓ ✓ . . ✓ 
  - fortran        ✓ . ✓ ✓ . 
  - haskell        ✓ . . . ✓ 
  - javascript     ✓ ✓ ✓ ✓ ✓ 
  - cmake          ✓ . ✓ . . 
  - vue            ✓ . ✓ ✓ ✓ 
  - kotlin         ✓ ✓ ✓ . ✓ 
  - cuda           ✓ ✓ ✓ ✓ ✓ 
  - dot            ✓ . . . ✓ 
  - json           ✓ ✓ ✓ ✓ . 
  - vim            ✓ ✓ ✓ . ✓ 
  - surface        ✓ . ✓ ✓ ✓ 
  - cpp            ✓ ✓ ✓ ✓ ✓ 
  - ruby           ✓ ✓ ✓ ✓ ✓ 
  - rust           ✓ ✓ ✓ ✓ ✓ 
  - solidity       ✓ . . . . 
  - llvm           ✓ . . . . 
  - gomod          ✓ . . . ✓ 
[...]

PR #3121, opened 2022-07-05
fix(ecma): indent end for bracket

kyazdani42:
should fix https://github.com/nvim-treesitter/nvim-treesitter/issues/3120


clason (approved):

kyazdani42:
@theHamsta test added.

Issue #3136, opened 2022-07-08
cmdline opens and closes when running incremental selections when cmdheight=0


s1n7ax:
### Describe the bug

When running init_selection incremental selection while cmdheight=0 it opens up the cmdline and closes to run a command. I looks like neovim glitching. Possible solution is to use `<cmd>` instead `:` for keybinds.

```vim
nmap ,s <cmd>init_selection<cr>
```

https://user-images.githubusercontent.com/18459807/177929414-59b8be52-ecc6-4b0b-8353-c09340be3545.mp4

### To Reproduce

1. Open a .lua file
2. Press keybind for init_selection

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.6 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v18.4.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 12.1.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - rego           ✓ . . . ✓ 
  - hcl            ✓ . ✓ ✓ ✓ 
  - javascript     ✓ ✓ ✓ ✓ ✓ 
  - godot_resource ✓ ✓ ✓ . . 
  - haskell        ✓ . . . ✓ 
  - scheme         ✓ . ✓ . ✓ 
  - slint          ✓ . . ✓ . 
  - supercollider  ✓ ✓ ✓ ✓ ✓ 
  - sparql         ✓ ✓ ✓ ✓ ✓ 
  - scala          ✓ . ✓ . ✓ 
  - tsx            ✓ ✓ ✓ ✓ ✓ 
  - query          ✓ ✓ ✓ ✓ ✓ 
  - typescript     ✓ ✓ ✓ ✓ ✓ 
  - jsdoc          ✓ . . . . 
  - todotxt        ✓ . . . . 
  - comment        ✓ . . . . 
  - c_sharp        ✓ ✓ ✓ . ✓ 
  - v              ✓ ✓ ✓ ✓ ✓ 
  - phpdoc         ✓ . . . . 
[...]

PR #3138, opened 2022-07-08
fix(selection): use <cmd> instead of : for mapping

kyazdani42:
fixes https://github.com/nvim-treesitter/nvim-treesitter/issues/3136

### DIFF
--- a/lua/nvim-treesitter/incremental_selection.lua
+++ b/lua/nvim-treesitter/incremental_selection.lua
@@ -129,7 +129,7 @@ function M.attach(bufnr)
     else
       mode = "x"
     end
-    local cmd = string.format(":lua require'nvim-treesitter.incremental_selection'.%s()<CR>", funcname)
+    local cmd = string.format("<cmd>lua require'nvim-treesitter.incremental_selection'.%s()<CR>", funcname)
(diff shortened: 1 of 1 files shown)


kyazdani42:
@s1n7ax can you check this works reliably for you ?

clason (approved):

s1n7ax:
@[kyazdani42](https://github.com/kyazdani42) I thought it would but no. It works for `init_selection` but it breaks all the other `*_incremental` selections.  The reason is `visual_selection_range` coming from Neovim is invalid when it's ran with `<cmd>`. I have tried vim.keymap and passing lua closure which didn't work either.

```lua
local function range_matches(node)
  local csrow, cscol, cerow, cecol = visual_selection_range()
  local srow, scol, erow, ecol = ts_utils.get_vim_range { node:range() }
  print(csrow, cscol, cerow, cecol)
  print(srow, scol, erow, ecol)
  return srow == csrow and scol == cscol and erow == cerow and ecol == cecol
end
```

I just realized this is related to a bug I created last year. https://github.com/neovim/neovim/issues/15144

kyazdani42:
let's close this then. I think cmdheight=0 came with a lot of undesired UX issues.

PR #3139, opened 2022-07-08
fix: add more indentations for ruby

zhengpd:
For `unless`:

```ruby
unless cond
  stmt
end
```

For `assignment`:

```ruby
ret =
  if cond
    stmt
  else
    stmt2
  end
```


kyazdani42:
could you add some tests for it please ? we try to have as much tests as we can to avoid regressions because we might change the algorithms bit by bit :)

zhengpd:
@kyazdani42 I added some tests. Maybe some testing instructions should be added to contributing guide.

kyazdani42 (approved):
thanks :) we'll add some notes to contributing.md file, forgot we didn't specify for tests there.

PR #3144, opened 2022-07-08
fix: avoid installing parsers multiple times when using auto_install

smjonas:
This fixes the following issue I came across when using `auto_install = true`.
To reproduce:
1. `:TSUninstall bash`
2. `:e ~/test.sh` (parser will be installed)
3. `:e ~/test.sh` (user will be prompted whether parser should be installed again which is unexpected)

Also this includes a small refactor, because specifying `ask_reinstall` is not necessary.


theHamsta (approved):

theHamsta:
Thank you!

PR #3149, opened 2022-07-09
fix(valahighlight): fix the break block and improve the highlights

Decodetalkers:
image
Finally

### DIFF
--- a/queries/vala/folds.scm
+++ b/queries/vala/folds.scm
@@ -1,0 +1,7 @@
+[
+  (namespace)
+	(function_definition)
+	(class_declaration)
+	(interface_declaration)
+] @fold
+
--- a/queries/vala/highlights.scm
+++ b/queries/vala/highlights.scm
@@ -307,9 +307,11 @@
 
 ;;; namespace 
 (namespaced_identifier
-	left: (identifier) @namespace
-	right: (identifier) @type
-	)
+  left: (identifier) @namespace
+)
+(namespaced_identifier
+  right: (identifier) @type
+)
 (primitive_type) @type
 ;;; define variable
 ;;(generic_identifier
--- a/queries/vala/highlights.scm
+++ b/queries/vala/highlights.scm
@@ -66,10 +66,6 @@
 ] @keyword.operator
 
 (this) @variable.builtin
-;;; move enum to namespace
-[
-  "enum"
-] @namespace
 
 ;;; move new and delet to operator
 [
@@ -78,6 +74,7 @@
 ] @keyword.operator
 ;;; keyword
 [
+  "enum"
   (modifier)
   "namespace"
   "var"
--- a/queries/vala/highlights.scm
+++ b/queries/vala/highlights.scm
@@ -57,7 +57,6 @@
   "yield"
   "break"
 ] @keyword.return
-(primitive_type)
 
 (null) @constant.builtin
 
@@ -202,7 +201,6 @@
 
 ; Types
 
-(primitive_type) @type
 
 (nullable_type
     (_) @type
@@ -317,7 +315,7 @@
 (generic_identifier
 	(identifier) @type )
 (declaration 
-	type_name: (identifier) @primitive_type
+	type_name: (identifier) @type
 	(identifier) @variable)
 
 ;;; inline etc
@@ -334,3 +332,4 @@
 	(identifier) @variable
 	(identifier) @constant)
 (array_identifier (identifier) @type)
+type_name: (identifier) @type
--- a/queries/vala/highlights.scm
+++ b/queries/vala/highlights.scm
@@ -2,21 +2,25 @@
 
 ((identifier) @constant (#match? @constant "^[A-Z][A-Z\\d_]+$"))
 
-(namespaced_identifier
-  left: [
-    ; Lowercased names in lhs typically are variables, while camel cased are namespaces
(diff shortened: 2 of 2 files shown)


Decodetalkers:
If meet highlight problem , it is the problem of treesitter.. not mime.. :) the treesitter still have a lot of problem, I just fix and improve the highlight...

theHamsta:
> meet highlight problem , it is the problem of treesitter.. not mime.. :) the treesitter still have a lot of problem, I just fix and improve the highlight...

I don't understand. Did you discover anything specific?

theHamsta:
@matbme

Decodetalkers:
> 

I just test a snippet, and with playground I found a lot of error, so the highlight works very bad.. but I have test the repo upstream , it works well

queries/vala/folds.scm
```diff
@@ -0,0 +1,7 @@
+[
+  (namespace)
+	(function_definition)
+	(class_declaration)
+	(interface_declaration)
+] @fold
+
```
theHamsta, line 7: Nice, that this got added!

Decodetalkers:
image
As you see, There is a lot of error

queries/vala/highlights.scm
```diff
@@ -54,46 +58,56 @@
...
 [
   "typeof"
   "is"
 ] @keyword.operator
 
+(this) @variable.builtin
+;;; move enum to namespace
+[
+  "enum"
+] @namespace
```
theHamsta, line 72: I don't think this is `@namespace`

Decodetalkers:
```vala


namespace Lsp {
    [CCode (default_value = "LSP_TEXT_DOCUMENT_SYNC_KIND_Unset")]
    enum TextDocumentSyncKind {
        Unset = -1,

        None = 0,

        Full = 1,

        Incremental = 2
    }
    class DocumentSymbol : Object, Json.Serializable {
        private Vala.SourceReference? _source_reference;
        public string name { get; set; }
        public string? detail { get; set; }
        public SymbolKind kind { get; set; }
        public bool deprecated { get; set; }
        private Range? _initial_range;
        public Range range {
            owned get {
                if (_initial_range == null)
                    _initial_range = new Range.from_sourceref (children.first ()._source_reference);
                
                return children.fold<Range> ((child, current_range) => current_range.union (child.range), _initial_range);
            }
        }
        public Range selectionRange { get; set; }
        public Gee.List<DocumentSymbol> children { get; private set; default = new Gee.LinkedList<DocumentSymbol> (); }
        public string? parent_name;

        private DocumentSymbol () {}

        /**
         * @param type the data type containing this symbol, if there was one (not available for Namespaces, for example)
         * @param sym the symbol
         */

        public new void Json.Serializable.set_property (ParamSpec pspec, Value value) {
            base.set_property (pspec.get_name (), value);
        }

        public new Value Json.Serializable.get_property (ParamSpec pspec) {
            Value val = Value (pspec.value_type);
            base.get_property (pspec.get_name (), ref val);
            return val;
        }

        public unowned ParamSpec? find_property (string name) {
            return this.get_class ().find_property (name);
        }

        public Json.Node serialize_property (string property_name, Value value, ParamSpec pspec) {
            if (property_name != "children")
                return default_serialize_property (property_name, value, pspec);
            var node = new Json.Node (Json.NodeType.ARRAY);
            node.init_array (new Json.Array ());
            var array = node.get_array ();
            foreach (var child in children)
                array.add_element (Json.gobject_serialize (child));
            return node;
        }

        public bool deserialize_property (string property_name, out Value value, ParamSpec pspec, Json.Node property_node) {
            error ("deserialization not supported");
        }
    }
}
```
snippet is here, you can test yourself, but with the repo upstream there is no error,have you synchronize with https://github.com/vala-lang/tree-sitter-vala ?

queries/vala/highlights.scm
```diff
@@ -54,46 +58,56 @@
...
+;;; conditional
 [
   "if"
   "else"
   "switch"
   "case"
   "default"
+  "throws"
+	"try"
+  "catch"
```
theHamsta, line 110: There is `@exception` for those three

Decodetalkers: > There is `@exception` for those three

fixed

queries/vala/highlights.scm
Decodetalkers: > I don't think this is `@namespace`

What is your suggestion, I think it should be another color..or just put it in keyword?

Decodetalkers: > I don't think this is `@namespace`

now it has returned to keyword.

Decodetalkers:
I see, there is something error with 
https://github.com/matbme/tree-sitter-vala, and you use this repo..

matbme:
Hmm, AFAIK this is all fixed in [https://github.com/vala-lang/tree-sitter-vala](). Maybe it's time to bring back #2818.

Decodetalkers:
> Hmm, AFAIK this is all fixed in https://github.com/vala-lang/tree-sitter-vala. Maybe it's time to bring back #2818.

I wonder if the key words will be the same, either all scheme file will be rewrite..

clason:
Yeah, that PR _will_ be merged, so it's best to park any work on Vala queries until then.

clason:
It's merged now, so you'll have to rebase (but check the new highlight first whether they already address your problems).

I'm taking the liberty as marking this PR as draft while you do that!

PR #3152, opened 2022-07-09
add type def highlight, add field declaration highlight and remove co…

vcraescu:
* add `@type.definition` highlight
* add `field_declaration` to be highlighted as `@field`
* removed constant highlight based on uppercase chars
In golang constants, identifiers are not supposed to be uppercase nor include "_" and the current query will incorrectly highlight any uppercased field (which is quite common in golang because all the abbrev must be uppercased: ID, HTML, URL etc.)

ID field is incorrectly highlighted as a constant (italic):

image

### DIFF
--- a/queries/go/highlights.scm
+++ b/queries/go/highlights.scm
@@ -5,7 +5,6 @@
 ; Identifiers
 
 (type_identifier) @type
-(type_spec name: (type_identifier) @type.definition)
 (field_identifier) @property
 (identifier) @variable
 (package_identifier) @namespace
--- a/queries/go/highlights.scm
+++ b/queries/go/highlights.scm
@@ -5,6 +5,7 @@
 ; Identifiers
 
 (type_identifier) @type
+(type_spec name: (type_identifier) @type.definition)
 (field_identifier) @property
 (identifier) @variable
 (package_identifier) @namespace
@@ -14,9 +15,6 @@
 
 ((identifier) @constant
  (#eq? @constant "_"))
-
-((identifier) @constant
- (#vim-match? @constant "^[A-Z][A-Z\\d_]+$"))
 
 (const_spec
   name: (identifier) @constant)
@@ -201,6 +199,7 @@
 
 (keyed_element
   . (literal_element (identifier) @field))
+(field_declaration name: (field_identifier) @field)
 
 (comment) @comment
 


queries/go/highlights.scm
```diff
@@ -5,6 +5,7 @@
 ; Identifiers
 
 (type_identifier) @type
```
theHamsta, line 7: This capture is not used at the moment (despite being documented) and I don't think it makes sense to add it. There is a plan to bring back opt-in semantic highlighting but I think we shouldn't mix it with regular highlighting. Especially when we decide to have `@type.definition` then there should be probably something analogous for `@xxx.definition` for other type of definition.

In the past, we had the policy that newly introduced highlights should have their own dedicated PRs to discuss them better and also add them for multiple languages to see and discuss whether they make sense. This also makes it easier to create a tracking issue to list the remaining languages that are still missing that new highlight group

queries/go/highlights.scm
```diff
@@ -15,9 +16,6 @@
 ((identifier) @constant
  (#eq? @constant "_"))
 
-((identifier) @constant
- (#vim-match? @constant "^[A-Z][A-Z\\d_]+$"))
```
theHamsta, line 19: Can you describe your motivation to remove this? Where does it produce bad highlighting for you? This help others to understand your changes better

theHamsta:
Apart from type.definition, this looks great to me! Thanks for discovering the mentioned issues!

queries/go/highlights.scm
vcraescu: I posted a screenshot of what it does and it is completely wrong for golang. There is no such rule to uppercase the constants in golang, it is just a normal identifier and you cannot highlight them without using some kind of scope.

queries/go/highlights.scm
theHamsta: So I'd prefer plain `@type` for now and have a discussion for `@type.definition` in a separate PR. If this highlight is important it should probably be added to the parsers in `tree-sitter/tree-sitter-Xxx` as they are the most popular languages

vcraescu: I supposed you mean you would like to be consistent and also have something like `method.definition` ? 
I removed `@type.definition` from the MR.

vcraescu: `@type.definition` is important just to be able to match how the other IDE work (VSCode/JetBrains). Both are able to discriminate between usage and declaration. I just looked at https://github.com/nvim-treesitter/nvim-treesitter/blob/e2b2d2357b227b1d2548b6bfb3506b3a93b9e636/lua/nvim-treesitter/highlight.lua#L113 and saw `@type.definition` and I thought it's perfectly valid to add it 😁

vcraescu: I created another PR for that: https://github.com/nvim-treesitter/nvim-treesitter/pull/3154

theHamsta (approved):

PR #3154, opened 2022-07-09
add golang type definition highlight

vcraescu:
Continuing the discussion started at https://github.com/nvim-treesitter/nvim-treesitter/pull/3152
regarding `@type.definition` highlight if we should it or not.

### DIFF
--- a/queries/go/highlights.scm
+++ b/queries/go/highlights.scm
@@ -5,6 +5,7 @@
 ; Identifiers
 
 (type_identifier) @type
+(type_spec name: (type_identifier) @type.definition)
 (field_identifier) @property
 (identifier) @variable
 (package_identifier) @namespace


theHamsta:
This is actually the kind of discussion I prefer. For some context. A few highlight groups where added in e2b2d2357. But just there documentation which has the disadvantage that usually nobody cares about having them properly integrated into our repo and generating inconsistent documentation.

This was by the way how the highlight groups were originally created. But we realized that this causes some highlight groups that were ambiguous and inconsistently used in the queries. So we removed most of the highlight groups that weren't needed. After that we tried to avoid introducing new highlight groups.

Some new highlight groups were still added afterwards for some good reason. But we tried to add them for all the languages to be sure that they make sense also from the naming.

I think at the end of this PR we should either
- remove `@type.definition` from documentation
- add `@type.definition` for the most important languages and create a tracking issue for all the languages to check whether it would sense to add them (with checkbox todo Vs done)

vcraescu:
Isn't possible to have default highlights and then for each language have a specific highlight for that language?

So when you define a highlight for `@type.definition`, we will have 2 highlights: `TSTypeDefinition` and `TSGoTypeDefinition` / `TSJavascriptDefinition` etc. That way we can define a default colour scheme for all the languages but we could also overwrite the colour scheme for a particular language. 
Jetbrains products work like this, they do have `Language Defaults` which can be overwritten if you want to, otherwise, the colours are inherited from the `Language Defaults`. 

It's impossible to define a colour scheme that looks good for all the languages as long as you cannot control how all the language's highlights are defined.

clason:
> Isn't possible to have default highlights and then for each language have a specific highlight for that language?

That is exactly how it already works: a capture `@foo` from a parser for `lang` is mapped to `langTSFoo`, which is then mapped to `TSFoo` (which is then mapped to one of the standard highlight groups like  `Function`). A colorscheme can assign colors to each of these directly, including `highlight langTSFoo guifg=red` etc.

vcraescu:
@clason Thanks a lot for the clarification!

clason:
Note that this was recently changed on Neovim 0.8: you can now map captures directly to highlight groups, with the (implicit) language tacked on at the end, e.g., `hi link @type.definition.typescript Error` (with fallback to `@type.definition` and then to `@type` if the more specialized groups don't exist).

theHamsta:
@clason should we think of this as specialization or try to not forget the issue and add `typedef ...  X`/`using X =` for C, `type  X ...` in Rust

clason (approved):

PR #3177, opened 2022-07-14
Move class keyword from highlighting group

baco:
Usually the python `class` keyword has the same highlighting color as the other keywords in the function group. It has more sense to move the keyword from one group to the other.

### DIFF
--- a/queries/python/highlights.scm
+++ b/queries/python/highlights.scm
@@ -202,6 +202,7 @@
 ] @keyword.operator
 
 [
+  "class"
   "def"
   "lambda"
 ] @keyword.function
@@ -210,7 +211,6 @@
   "assert"
   "async"
   "await"
-  "class"
   "exec"
   "global"
   "nonlocal"


theHamsta:
> Usually the python class keyword has the same highlighting color as the other keywords in the function group. It has more sense to move the keyword from one group to the other.

What do you mean by "usually"? What has `class` to do functions? `keyword.function` was originally thought for keyword pairs starting and ending functions (e.g. ` function` ... `end` in lua)

baco:
> > Usually the python class keyword has the same highlighting color as the other keywords in the function group. It has more sense to move the keyword from one group to the other.
> 
> What do you mean by "usually"? What has `class` to do functions? `keyword.function` was originally thought for keyword pairs starting and ending functions (e.g. ` function` ... `end` in lua)

I mean that in most themes, editors, etc., the `class` keyword, specifically for Python (hence the PR over just highlighting, and that language only), is highlighted with the same color and format as the `def` or `lambda` keyword, used to denote function declarations.

It has not the same functionality, but that's the categorization it's given among editors. Using, generally, a different color/format than other lesser operators like `async`, `await`, `with`, `pass` or even `print`.

theHamsta:
class/struct have the keyword highlight in this repo. In theory, there is a Vim highlight group called `Structure` that is thought for `enum`/`struct`/`class`. But we removed it because the name was a bit confusing (people would think it would be for the identifier of a `struct`. Maybe it would be better to understand as `@keyword.struct` or `TSKeywordStruct`.

We would be happy if you have a suggestion on how to solve this better, but we shouldn't change this for one language only but then for all languages consistently.

You can add `TSFunction` highlight to `class` in Python also in your config files.

baco:
> You can add `TSFunction` highlight to `class` in Python also in your config files.

Ok, fair enough. Meanwhile. How?

theHamsta:
In a Python file `:TSQueryEditUserAfter highlights`, add the query you wanted to add here in the new file. You can use priority if the highlight still doesn't get overwritten `("class" @keyword.function (#set! "priority" 105))`.

theHamsta:
@clason @kyazdani42 @vigoux do we need `@keyword.struct`/`TSKeywordStruct` which would correspond to the builtin highlight `Structure`. We had `Structure` once but we removed it since the name was confusing and there was no strong requirement to have it.

kyazdani42:
in java, csharp and js, we highlight "class" as keyword. I'm not sure we want to change it for python.

baco:
take for instance this JS highlighting taken from VSCodium:
image
look how the keywords `function` and `class` are put in the same highlighting group; aside from other highlighting groups (hence the different color) like the keywords `return` or `if` for flow-control.

With the current classification of `class`, for Python, as a any other keyword, the `class` keyword tends to receive different coloring than function declaration (with this colorscheme would be purple instead of blue).

kyazdani42:
We don't aim to implement vscode highlighting. A class to me has a different semantic/meaning than a function, and so it should be kept separated.
Although i wouldn't be against adding a keyword.class or something like that since we have a keyword.function

baco:
It is kind of arbitrary the line in the sand there. A class, to you, “has a different semantic/meaning than a function, and so it should be kept separated”, but rather has to be grouped with a bunch of other common keywords, just because “in java, csharp and js, we highlight 'class' as keyword”.
It also has a different semantic/meaning there. It doesn't control flow like an "if", a "switch" or a "for", it rather defines types; but seems not enough to tag it `@keyword.struct`/`TSKeywordStruct` as @theHamsta proposed. Why?

kyazdani42:
I don't have a very string opinion tbh, that was just a suggestion.
I think going for keyword structure could also work, but since this is not a struct and the keyword is pretty common in many languages, we could give it its own hl group.

PR #3180, opened 2022-07-14
injections(julia): add Markdown injections again

theHamsta:
image


theHamsta:
This should probably apply only to strings just before functions

clason (approved):

theHamsta:
Restricted this to top-level strings. Is this found also on struct/module level?

clason:
Looks to have been already added!

PR #3182, opened 2022-07-15
fix: disable elm parser in markdown to avoid segfault

kyazdani42:
the Elm parser segfaults when injected in markdown. Disable it for now.
See https://github.com/elm-tooling/tree-sitter-elm/issues/124

@theHamsta @clason i'm not sure how to debug this, i haven't managed to make the elm parser segfault when not injected.

### DIFF
--- a/queries/markdown/injections.scm
+++ b/queries/markdown/injections.scm
@@ -1,6 +1,7 @@
 (fenced_code_block
   (info_string
     (language) @language)
+  (#not-match? @language "elm")
   (code_fence_content) @content (#exclude_children! @content))
 
 ((html_block) @html)


theHamsta (approved):

kyazdani42:
@theHamsta when i run the fuzzer, i first get:
```
lib/src/query.c:927:11: runtime error: index -1 out of bounds for type 'AnalysisStateEntry[8]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior lib/src/query.c:927:11 in
```
i'm not sure about other errors, i've left it running for 10minutes and nothing came up. But this index out of bounds could be the reason injections makes the parser crash.

theHamsta:
@kyazdani42 try to run the parser without checking queries. You have to modify the source code of the parser main for that

kyazdani42:
@theHamsta Not sure i understand this correctly. How does modifying the elm parser source code allows to run without checking queries ?

PR #3195, opened 2022-07-18
fix(ci): ubuntu appimage fails

kyazdani42:
Cannot find the nightly release on github neovim release page anymore.
I suppose we can use stable now.


clason:
That's a regression from today (a PR broke the Windows release workflow) that'll get fixed soon. I wouldn't worry about it here (and it would be good to have a nightly canary on CI!)

kyazdani42:
I see :) i'm closing this then

PR #3226, opened 2022-07-27
highlights(php): highlight more string types and escapes

megabug:
`encapsed_string`s are double-quoted strings. Highlight their non-escaped portions like normal single-quoted `string`s.

`nowdoc`s are non-parsed `heredoc`s. Highlight them like normal single-quoted `string`s.

`escape_sequence`s can appear in `heredoc`s and `shell_command_expression`s, not just `encapsed_string`s. Highlight these too.

Before:

After:

### DIFF
--- a/queries/php/highlights.scm
+++ b/queries/php/highlights.scm
@@ -112,10 +112,12 @@
 
 [
  (string)
+ (encapsed_string)
  (heredoc)
+ (nowdoc)
  (shell_command_expression) ; backtick operator: `ls -la`
  ] @string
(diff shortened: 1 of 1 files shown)


theHamsta (approved):

theHamsta:
Do you think that the EOS should be highlighted differently?

megabug:
Sure. Latest patch highlights the bodies of heredocs/nowdocs, not the entire string:

Screen Shot 2022-07-27 at 18 32 52

PR #3233, opened 2022-07-28
feat(ts_utils): allow starting selection after last character

AckslD:
Currently the visual selection cannot start after the final character of a line. This is because the start position is set before entering the visual selection in which case the cursor cannot be after the last character. If one instead first starts visual selection this is possible and one just needs to switch side with `o` and go to the last.

### DIFF
--- a/lua/nvim-treesitter/ts_utils.lua
+++ b/lua/nvim-treesitter/ts_utils.lua
@@ -259,14 +259,14 @@ function M.update_selection(buf, node, selection_mode)
   selection_mode = selection_mode or "charwise"
   local start_row, start_col, end_row, end_col = M.get_vim_range({ M.get_node_range(node) }, buf)
 
-  vim.fn.setpos(".", { buf, start_row, start_col, 0 })
-
   -- Start visual selection in appropriate mode
   local v_table = { charwise = "v", linewise = "V", blockwise = "<C-v>" }
   ---- Call to `nvim_replace_termcodes()` is needed for sending appropriate
   ---- command to enter blockwise mode
   local mode_string = vim.api.nvim_replace_termcodes(v_table[selection_mode] or selection_mode, true, true, true)
   vim.cmd("normal! " .. mode_string)
+  vim.fn.setpos(".", { buf, start_row, start_col, 0 })
+  vim.cmd "normal! o"
   vim.fn.setpos(".", { buf, end_row, end_col, 0 })
 end
 
--- a/tests/unit/ts_utils_spec.lua
+++ b/tests/unit/ts_utils_spec.lua
@@ -39,3 +39,70 @@ describe("is_in_node_range", function()
     assert.is_false(test_is_in_node_range(3, 0))
   end)
 end)
+
+describe("update_selection", function()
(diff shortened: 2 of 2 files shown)


kyazdani42 (approved):

kyazdani42:
Could you show me some steps to reproduce the issue you mentioned ? I cannot see any difference before and after this change.

AckslD:
Hi @kyazdani42, lets say you have the buffer:
```
foo

bar
```
and you call
```
:lua require('nvim-treesitter.ts_utils').update_selection(0, {0, 3, 2, 1}, v)
```
Previously this would also select the `o` of `foo` even though the start-column is after this character `3`.

Does that make sense?

kyazdani42:
looks ok to me. First column is 1, not 0.
So `3` will match the last character of the first line and `1` matches the 1st of the last line, inclusive.

AckslD:
I always get the indexing of columns wrong 🙃 but in any case, this allows one to select the column after the last character.

lua/nvim-treesitter/ts_utils.lua
```diff
@@ -259,14 +259,14 @@ function M.update_selection(buf, node, selection_mode)
...
-  vim.fn.setpos(".", { buf, start_row, start_col, 0 })
-
   -- Start visual selection in appropriate mode
   local v_table = { charwise = "v", linewise = "V", blockwise = "<C-v>" }
   ---- Call to `nvim_replace_termcodes()` is needed for sending appropriate
   ---- command to enter blockwise mode
   local mode_string = vim.api.nvim_replace_termcodes(v_table[selection_mode] or selection_mode, true, true, true)
   vim.cmd("normal! " .. mode_string)
+  vim.fn.setpos(".", { buf, start_row, start_col, 0 })
+  vim.cmd "normal! o"
```
theHamsta, line 269: and it's possible to use `o` here and all other selection mode like `V` and `<C-v>` keep working? I don't know very much about this area of Vim and would trust you. Could we add a test? We would need tests anyway when upstreaming this function to Neovim

Maybe suggest your change also here: https://github.com/neovim/neovim/pull/18232/

AckslD: @theHamsta I added some small tests (charwise, linewise and blockwise). There are not many functional tests so wasn't fully sure about the desired format. Also, is there some description somewhere for how to run the tests locally? I tried just running `scripts/run_tests.sh` but I need to setup eg plenary locally etc. So I haven't actually executed the tests that I added yet, but hopefully they will pass.

theHamsta (approved):

theHamsta (approved):

lua/nvim-treesitter/ts_utils.lua
AckslD: Woops, the new tests didn't actually run. I didn't put them in a file ending with `_spec`. Moved them to `ts_utils_spec.lua` now instead.

kyazdani42:
@AckslD seems linting is incorrect :)

AckslD:
I managed to run tests locally now :) so should be passing I think

PR #3235, opened 2022-07-29
feat(highlight): re-apply default highlights on colorscheme changes

jdrouhard:
If the plugin is loaded after a colorscheme is set that defines any of
these highlight groups, the default won't be applied. Subsequent
"highlight clear" commands (common when switching colorschemes) will
then clear any of those highlights, but these defaults never have
another opportunity to be initialized.

Effectively, if you load neovim with a colorscheme that has definitions
for some of these highlight groups, then load treesitter, then switch
colorschemes, many of these default links will be absent resulting in
colors that do not appear the same as if that colorscheme had been
used at startup.

Hooking the ColorScheme event with an autocmd that just reapplies these
defaults gives every colorscheme switch the opportunity to get the
defaults for non-explicitly-defined groups.

### DIFF
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -156,4 +156,92 @@ function M.set_custom_captures(captures)
   end
 end
 
+function M.set_default_hlgroups()
+  local highlights = {
+    TSNone = { default = true },
+    TSPunctDelimiter = { link = "Delimiter", default = true },
+    TSPunctBracket = { link = "Delimiter", default = true },
+    TSPunctSpecial = { link = "Delimiter", default = true },
+
+    TSConstant = { link = "Constant", default = true },
+    TSConstBuiltin = { link = "Special", default = true },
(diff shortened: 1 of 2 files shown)


kyazdani42 (approved):

lua/nvim-treesitter/highlight.lua
```diff
@@ -156,4 +156,92 @@ function M.set_custom_captures(captures)
   end
 end
 
+function M.set_default_highlights()
```
vigoux, line 159: Rather

```suggestion
function M.set_default_hlgroups()
```

vigoux (changes requested):

lua/nvim-treesitter/highlight.lua
jdrouhard: Done.

vigoux (approved):

theHamsta:
Nice! Thank you!

PR #3250, opened 2022-08-01
No fallback install location

### LINKED ISSUE #2777: Treesitter not removing the parsers when running `TSUninstall all`
### Describe the bug

I ran `TSInstall all` by accident so that installed all parsers. I wanted to get rid of all of them so I ran `TSUninstall all` and I restarted neovim. When opened neovim again, `TS` started to install everything back even tho I uninstalled everything. When running `TSInstallInfo` it was showing all of the parsers being installed. The only workaround was to delete everything with `treesitter` in their name from `local/share/nvim/...` 

### To Reproduce

1. Run `:TSInstall all`
2. Restart Neovim
3. Run `:TSUninstall all`
4. Restart Neovim

### Expected behavior

When opening Neovim again (after you ran `:TSUninstall all`), it will start to install all of the parsers back ignoring what you have in `ensure_installed = {}`. 

_If you run `:TSInstallInfo` after running `:TSUninstall all` you will still see all of the parsers being installed._

```markdown
## Installation
  - OK: `tree-sitter` found 0.20.1 (062421dece3315bd6f228ad6d468cba083d0a2d5) (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v15.14.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
    Version: Apple clang version 13.0.0 (clang-1300.0.29.30)
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.
```

```text
NVIM v0.7.0-dev+1292-ge9b53f3fb
Build type: Release
LuaJIT 2.1.0-beta3

_ps: I was able to reproduce this behaviour on the latest Neovim stable version as well_
```

### Additional context

- I am on `macOS Big Sur`
- I use `Plug` for installing the plugins:
```vim
Plug 'nvim-treesitter/nvim-treesitter', {'do': ':TSUpdate'}
``` 
- The plugins are installed in `/config/nvim/plugged/` and I have `Read & Write` permissions for this folder
- My treesitter config:
```lua
  ensure_installed = {
         "yaml",
         "markdown",
         "comment",
         "css",
         "html",
         "javascript",
         "json",
         "lua",
         "php",
         "phpdoc",
         "regex",
         "scss",
         "vim",
         "vue",
     },
  highlight = {
    enable = true,
    disable = { "html", "bash" },
  },
  indent = {
    enable = true,
  },
}

```

theHamsta:
- don't allow to silently fallback to alternative install location
- remove argument of function that wasn't used anywhere (neither in function nor in call sites)
- display warning when parser is still installed after `:TSUninstall`. nvim-treesitter will only uninstall from configured install dir
- warn when uninstalling a parser in ensure_uninstalled

grafik

Fixes #2777

Please test before merging! Will also test a few days. I think this is not urgent but helpful for issues in future.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -489,6 +489,16 @@ function M.uninstall(...)
         return api.nvim_err_writeln(err)
       end
 
+      if vim.tbl_contains(configs.get_ensure_installed_parsers(), lang) then
+        vim.notify(
+          "Uninstalling "
+            .. lang
+            .. '. But the parser is still configured in "ensure_installed" setting of nvim-treesitter.'
+            .. " Please consider updating your config!",
+          vim.log.levels.ERROR
+        )
+      end
+
       local parser_lib = utils.join_path(install_dir, lang) .. ".so"
       local all_parsers = vim.api.nvim_get_runtime_file("parser/" .. lang .. ".so", true)
       if vim.fn.filereadable(parser_lib) == 1 then
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -490,11 +490,36 @@ function M.uninstall(...)
       end
 
       local parser_lib = utils.join_path(install_dir, lang) .. ".so"
-
-      local command_list = {
-        shell.select_rm_file_cmd(parser_lib, "Uninstalling parser for " .. lang),
-      }
-      M.iter_cmd(command_list, 1, lang, "Treesitter parser for " .. lang .. " has been uninstalled")
+      local all_parsers = vim.api.nvim_get_runtime_file("parser/" .. lang .. ".so", true)
+      if vim.fn.filereadable(parser_lib) == 1 then
+        local command_list = {
+          shell.select_rm_file_cmd(parser_lib, "Uninstalling parser for " .. lang),
+          {
+            cmd = function()
+              local all_parsers_after_deletion = vim.api.nvim_get_runtime_file("parser/" .. lang .. ".so", true)
+              if #all_parsers_after_deletion > 0 then
+                vim.notify(
+                  "Tried to uninstall parser for "
+                    .. lang
+                    .. "! But the parser is still installed (not by nvim-treesitter)."
+                    .. " Please delete the following files manually: "
+                    .. table.concat(all_parsers_after_deletion, ", "),
+                  vim.log.levels.ERROR
+                )
+              end
+            end,
+          },
+        }
+        M.iter_cmd(command_list, 1, lang, "Treesitter parser for " .. lang .. " has been uninstalled")
+      elseif #all_parsers > 0 then
+        vim.notify(
+          "Parser for "
+            .. lang
+            .. " is installed! But not by nvim-treesitter! Please manually remove the following files: "
+            .. table.concat(all_parsers, ", "),
+          vim.log.levels.ERROR
+        )
+      end
     end
   end
 end
--- a/README.md
+++ b/README.md
@@ -115,6 +115,9 @@ require'nvim-treesitter.configs'.setup {
   -- List of parsers to ignore installing (for "all")
   ignore_install = { "javascript" },
 
+  ---- If you need to change the installation directory of the parsers (see -> Advanced Setup)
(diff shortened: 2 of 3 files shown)


lua/nvim-treesitter/install.lua
```diff
@@ -490,11 +490,36 @@ function M.uninstall(...)
       end
 
       local parser_lib = utils.join_path(install_dir, lang) .. ".so"
-
-      local command_list = {
-        shell.select_rm_file_cmd(parser_lib, "Uninstalling parser for " .. lang),
-      }
-      M.iter_cmd(command_list, 1, lang, "Treesitter parser for " .. lang .. " has been uninstalled")
+      local all_parsers = vim.api.nvim_get_runtime_file("parser/" .. lang .. ".so", true)
+      if vim.fn.filereadable(parser_lib) == 1 then
```
theHamsta, line 494: we could also try to uninstall parser from all locations but I think `:TSUninstall` is rather thought for managing parsers installed nvim-treesitter and not for instance the built-in ones.

lua/nvim-treesitter/install.lua
```diff
@@ -490,11 +490,36 @@ function M.uninstall(...)
...
+                    .. table.concat(all_parsers_after_deletion, ", ")
+                    .. ")"
+                )
+              end
+            end,
+          },
+        }
+        M.iter_cmd(command_list, 1, lang, "Treesitter parser for " .. lang .. " has been uninstalled")
+      elseif #all_parsers > 0 then
+        error(
```
theHamsta, line 515: Is `error` the right thing to do here? Or is printing a hint enough?

theHamsta: Much better with vim.notify!

theHamsta:
If also added a warning when uninstalling a parser in `ensure_uninstalled`. This should fix #2777

theHamsta:
Should we uninstall parsers when the install_dir is not in runtimepath? Or is this rather a configuration error?

lua/nvim-treesitter/install.lua
kyazdani42: I agree the installer should be managed for nvim-treesitter and not custom installs. Actually, we also could move the installer in a separate plugin, and provide a proper API to get the nvim-treesitter queries versions.
Let me know if this seems like a good idea, although it's still nice to have this in nvim-treesitter.

clason: Yes. Core will ship with some parsers soon(tm), and these should never be touched. Nvim-treesitter should only touch its own install location (but inform about others).

clason:
> Should we uninstall parsers when the install_dir is not in runtimepath? Or is this rather a configuration error?

Configuration error. Touching a user's filesystem should be done extremely conservatively (i.e., only when we're confident that we are doing the right thing).

theHamsta:
> >   Should we uninstall parsers when the install_dir is not in runtimepath? Or is this rather a configuration error?

> Configuration error. Touching a user's filesystem should be done extremely conservatively (i.e., only when we're confident that we are doing the right thing).

But the current detection of the parser via "loadable" status is very buggy. nvim-treesitter will think that parsers that have already been loaded are still installed.

clason:
> But the current detection of the parser via "loadable" status is very buggy. nvim-treesitter will think that parsers that have already been loaded are still installed.

I don't see how this contradicts (or even relates) to my statement? 

In any case, it's clear that we need more introspection for treesitter parsers (and queries), and we need that in core.

theHamsta:
I just feel that maybe we should just look for files in when uninstalling. At the moment, we first check whether we can load a parser instead of checking for a file in our install location.

theHamsta:
@clason should we merge this? This is also part of nvim-treesitter being a better citizen and installing parsers in random locations.

clason:
Yes, go ahead!

teto:
I come from https://github.com/NixOS/nixpkgs/issues/189838 . Why would the parsers need to be RW ? that breaks nvim-treesitter with nix as packages are installed read-only (for reproducibility and security reasons). Same would happen for someone that installs a grammar via apt for instance.

clason:
Because you want to update them?

theHamsta:
Nix can configure a location for the parsers or handle the installation of the parsers via nix (instead of using the installation routine of this plugin).

PR #3251, opened 2022-08-01
chore: show in TSInstallInfo when parser is still loaded

theHamsta:
should we be more precise in `:TSInstallInfo` when parser is still active after uninstallation?

### DIFF
--- a/lua/nvim-treesitter/info.lua
+++ b/lua/nvim-treesitter/info.lua
@@ -14,11 +14,13 @@ local function install_info()
 
   local parser_list = parsers.available_parsers()
   table.sort(parser_list)
-  for _, ft in pairs(parser_list) do
-    local is_installed = #api.nvim_get_runtime_file("parser/" .. ft .. ".so", false) > 0
-    api.nvim_out_write(ft .. string.rep(" ", max_len - #ft + 1))
+  for _, lang in pairs(parser_list) do
+    local is_installed = #api.nvim_get_runtime_file("parser/" .. lang .. ".so", false) > 0
+    api.nvim_out_write(lang .. string.rep(" ", max_len - #lang + 1))
     if is_installed then
       api.nvim_out_write "[✓] installed\n"
+    elseif pcall(vim.treesitter.inspect_lang, lang) then
+      api.nvim_out_write "[✗] not installed (but still loaded. Restart Neovim!)\n"
     else
       api.nvim_out_write "[✗] not installed\n"
     end


lua/nvim-treesitter/info.lua
```diff
@@ -14,11 +14,13 @@ local function install_info()
 
   local parser_list = parsers.available_parsers()
   table.sort(parser_list)
-  for _, ft in pairs(parser_list) do
-    local is_installed = #api.nvim_get_runtime_file("parser/" .. ft .. ".so", false) > 0
-    api.nvim_out_write(ft .. string.rep(" ", max_len - #ft + 1))
+  for _, lang in pairs(parser_list) do
```
theHamsta, line 17: renamed `ft` here, because `lang` is more precise

lua/nvim-treesitter/info.lua
```diff
@@ -14,11 +14,13 @@ local function install_info()
...
+    local is_installed = #api.nvim_get_runtime_file("parser/" .. lang .. ".so", false) > 0
+    api.nvim_out_write(lang .. string.rep(" ", max_len - #lang + 1))
     if is_installed then
       api.nvim_out_write "[✓] installed\n"
+    elseif vim._ts_has_language(lang) then
```
theHamsta, line 22: oh no, we're using a private function here!

kyazdani42: so what should we do about that :smile: ?

clason: use `vim.treesitter.require_language(lang)` instead?

theHamsta: I would prefer the proper function than triggering the error on a missing language in `inspect_language`. Would still be possible to `pcall` and check for error or nil return but that would also be a bit relying on the internal behavior of that. If it just returned

theHamsta: changed it to `pcall(vim.treesitter.instect_language, lang)`.

kyazdani42 (approved):

lua/nvim-treesitter/info.lua
theHamsta: I dont' know whether `pcall(vim.treesitter.instect_language, lang)` was a good idea. It has the side effect that it will load all the parsers. `_has_language` will only check whether the parser is loaded.

clason: Well, that would be a good argument to make `_has_language` (or a better version of that) public?

theHamsta: Maybe as `is_loaded`... But should be fine as is for now

PR #3253, opened 2022-08-01
chore: trim space in check-health

### LINKED ISSUE #3248: Stray space in check health implementation
How `:checkhealth` looks like when `vim.opt.list=true` and shows leading spaces.

bild

Culprit: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/health.lua#L126

Suggestion:
```diff
- out = out .. status .. " "
+ out = out .. status
```

theHamsta:
Fixes #3248

### DIFF
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -128,7 +128,7 @@ function M.check()
           table.insert(error_collection, { parser_name, query_group, err })
         end
       end
-      table.insert(parser_installation, out)
+      table.insert(parser_installation, vim.fn.trim(out, " ", 2))
     end
   end
   local legend = [[


lua/nvim-treesitter/health.lua
```diff
@@ -128,7 +128,7 @@ function M.check()
           table.insert(error_collection, { parser_name, query_group, err })
         end
       end
-      table.insert(parser_installation, out)
+      table.insert(parser_installation, vim.trim(out, " ", 2))
```
kyazdani42, line 131: why the additional parameters ? I dont see them in the documentation

theHamsta: ooops, meant to use `vim.fn.trim` which is right-trim only

kyazdani42 (approved):

Issue #3255, opened 2022-08-02
node_incremental for incremental selection jumping around


Jaco-Minnaar:
### Describe the bug

When I hit the key mapped to `node_incremental`, I'm put back in Normal mode, the cursor jumps to the end of the file, and a new line is inserted somewhere in the file.

### To Reproduce

1. Hit key mapped to `init_selection` (this works as expected).
2. Hit key mapped to `node_incremental`.
3. See described behavior.

### Expected behavior

The visual selection to increment to the next node.

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.6 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v16.15.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 12.1.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - go             ✓ ✓ ✓ ✓ ✓
  - json           ✓ ✓ ✓ ✓ .
  - vim            ✓ ✓ ✓ . ✓
  - lua            ✓ ✓ ✓ ✓ ✓
  - zig            ✓ . ✓ ✓ ✓
  - c              ✓ ✓ ✓ ✓ ✓
  - bash           ✓ ✓ ✓ . ✓
  - cpp            ✓ ✓ ✓ ✓ ✓
  - c_sharp        ✓ ✓ ✓ . ✓
  - css            ✓ . ✓ ✓ ✓
  - dockerfile     ✓ . . . ✓
  - glsl           ✓ ✓ ✓ ✓ ✓
  - php            ✓ ✓ ✓ ✓ ✓
  - python         ✓ ✓ ✓ ✓ ✓
  - scss           ✓ . . ✓ .
  - tsx            ✓ ✓ ✓ ✓ ✓
  - typescript     ✓ ✓ ✓ ✓ ✓
  - yaml           ✓ ✓ ✓ ✓ ✓
  - html           ✓ ✓ ✓ ✓ ✓
  - rust           ✓ ✓ ✓ ✓ ✓

[...]

PR #3256, opened 2022-08-02
Revert changes that break incremental_selection visual mode keymaps

argshook:
[This change](https://github.com/nvim-treesitter/nvim-treesitter/commit/22824614163c54ef5d366b15c8685e24d33763db) introduced an issue which rendered incremental selection keymaps unusable.
[This issue](https://github.com/nvim-treesitter/nvim-treesitter/issues/3255) describes behaviour in more detail.

This commit reverts the part of the code which introduced the issue

### DIFF
--- a/lua/nvim-treesitter/incremental_selection.lua
+++ b/lua/nvim-treesitter/incremental_selection.lua
@@ -136,12 +136,8 @@ function M.attach(bufnr)
     else
       mode = "x"
     end
-    vim.keymap.set(
-      mode,
-      mapping,
-      M[funcname],
-      { buffer = bufnr, silent = true, noremap = true, desc = FUNCTION_DESCRIPTIONS[funcname] }
-    )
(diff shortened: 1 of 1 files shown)


kyazdani42 (commented):
why though ? Shouldn't we try to fix the issue properly instead of going back to what it was ? I think vim.keymap.set uses 'remap = false' instead of 'noremap = true'

kyazdani42 (approved):

kyazdani42:
closing in favor of #3257

argshook:
> ~why though ? Shouldn't we try to fix the issue properly instead of going back to what it was ? I think vim.keymap.set uses 'remap = false' instead of 'noremap = true'~ EDIT: haven't seen the discussion. Indeed we should go back then.

in my book i'd first aim to revert the change that introduces the issue, if the proper solution takes time to release. If there's a better solution already, i'm all for it.

kyazdani42:
yes but since there is already a better solution :)

PR #3268, opened 2022-08-04
go folds.scm composit_literal

ray-x:
image
Fold to: (I used a modified version of fold.lua that keep last `}`
image


theHamsta (approved):
Thank you!

PR #3269, opened 2022-08-04
fix(zig): Do not indent on newline after block

phynalle:
An additional indentaion on newline is appended.

This PR fixes a bug that an additional indentaion on newline is appended.

Current Behaviour is 
```
}<cursor>

to

}
<indent><cursor>
```

But expected behaviour should be below
```
}<cursor>

to

}
<cursor>
```

### DIFF
--- a/queries/zig/indents.scm
+++ b/queries/zig/indents.scm
@@ -5,6 +5,8 @@
   (InitList)
 ] @indent
 
+(Block "}" @indent_end)
+
 [
   "("
   ")"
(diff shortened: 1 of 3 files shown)


theHamsta (changes requested):
Can you add a test please? Copy `tests/indent/go_spec.lua` to`tests/indent/zig_spec.lua` and put a test file in a zig folder.

phynalle:
I have added it

theHamsta (approved):
Thank you :heart: !

phynalle:
oops I missed a commit which enables expandtab

PR #3277, opened 2022-08-06
ts_utils nil check for get_node_text

ray-x:
This error only happened when the pulgin tried to get_node_text while the current buffer was being edited/changed by other plugins(e.g. Formatter).
I mapped CursorHoleI to Treesitter Status Line and have seen the lines being empty from time to time.


theHamsta (approved):
Thank you!

PR #3279, opened 2022-08-06
add details section for supported languages

chapeupreto:
This commit adds a details section for the supported languages so it is easier to read the README file.


theHamsta:
Rendered preview of the README available here: https://github.com/chapeupreto/nvim-treesitter/tree/patch-1

I usually only go for the languages to get the parser links and the maintainers but I guess I'm a minority in that regard. I would be ok with that change although quite neutral. Would love to hear other voices.

clason:
> I usually only go for the languages to get the parser links and the maintainers but I guess I'm a minority in that regard.

Me, too -- that's the single most useful part of the README for me...

Issue #3288, opened 2022-08-10
[GO] indentation problems with `switch` `select` and `struct` keywords


geefuoco:
### Describe the bug

Indentation of curly brackets is incorrect when using certain keywords

```go
func incorrect(word string) {
  switch word {

} // <---
  select word {

} // <---

  arr := []struct {

} // <---
}

```

### To Reproduce

1. Create go file
2. Use `switch`, `select`, or `[]struct` inside of a function

### Expected behavior

```go
func correct(word string) {
  switch word {

  } // <---
  select {
  
  } // <---

  arr := []struct {

  } // <---
}
```

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
  - OK: `node` found v16.13.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - scheme         ✓ . ✓ . ✓
  - hjson          ✓ ✓ ✓ ✓ ✓
  - jsonc          ✓ ✓ ✓ ✓ ✓
  - fusion         ✓ ✓ ✓ ✓ .
  - elm            ✓ . . . ✓
  - ledger         ✓ . ✓ ✓ ✓
  - yang           ✓ . ✓ ✓ .
  - v              ✓ ✓ ✓ ✓ ✓
  - ninja          ✓ . ✓ ✓ .
  - nix            ✓ ✓ ✓ . ✓
  - gomod          ✓ . . . ✓
  - go             ✓ ✓ ✓ ✓ ✓
  - gowork         ✓ . . . ✓
  - json5          ✓ . . . ✓
  - proto          ✓ . ✓ . .
  - graphql        ✓ . . ✓ ✓
  - m68k           ✓ ✓ ✓ . ✓
  - fennel         ✓ ✓ . . ✓
  - ruby           ✓ ✓ ✓ ✓ ✓
[...]

PR #3289, opened 2022-08-10
indents(go): don't indent on all composite_literal

theHamsta:
I don't feel to confident about this change. I don't use go and our test coverage is not exhaustive.


theHamsta:
Works for issue author: https://github.com/nvim-treesitter/nvim-treesitter/issues/3288#issuecomment-1211293265

PR #3300, opened 2022-08-15
highlights(swift): add keywords "actor", "nonisolated"

theHamsta:
Introduced in https://github.com/alex-pinkus/tree-sitter-swift/commit/07bdffb5e6f34fb1e578f31b653479d7c4d6306c

@alex-pinkus

### DIFF
--- a/queries/swift/highlights.scm
+++ b/queries/swift/highlights.scm
@@ -33,6 +33,8 @@
   "typealias"
   "struct"
   "class"
+  "actor"
+  "nonisolated"
   "enum"
   "protocol"
   "extension"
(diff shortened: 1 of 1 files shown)


theHamsta:
Should
```
  "override"
  "convenience"
  "required"
````
get any special highlights?

alex-pinkus:
From glancing at other grammars, I don't see anything more specific than `@keyword` that seems like others use that would apply to those. They're pretty ordinary function modifiers (`convenience` and `required` only apply to initializers, but they don't for instance _replace_ the keyword `func`). Is there anything I'm missing that would imply otherwise? Certainly open to learning more -- I know a lot more about Swift than I do about highlight queries.

Issue #2282, opened 2022-01-17
Tracking issue: language support requests


clason:
This is a tracking issue collecting requests for new languages to add to nvim-treesitter. If you would like to add a language to the list, comment on this issue with the following information:

* name of the language
* link to an existing tree-sitter parser (if available)
* link to a repository containing queries (if available)
* language grammar description, vim syntax plugin etc. if none of the above exist.

*Note: having a language on this list does not mean a maintainer will do the work for adding it; rather this list is meant as guidance for community contributions.*

---

# Languages needing a PR

(Parsers and simple queries exist already and just need to be included in our list and tested.)

* [x] [Proto](https://github.com/mitchellh/tree-sitter-proto) (#2307)
* [x] [Solidity](https://github.com/JoranHonig/tree-sitter-solidity) (#1168)
* [x] [SQL](https://github.com/nvim-treesitter/nvim-treesitter/pull/1403) (#2511)
* [x] [org](https://github.com/milisims/tree-sitter-org) (#2841)
* [ ] [TJ's Neovim-specific Lua parser](https://github.com/tjdevries/tree-sitter-lua) (#543)
* [ ] [ReScript](https://github.com/nkrkv/nvim-treesitter-rescript/tree/main/queries/rescript) 
* [ ] [gotmpl](https://github.com/ngalaiko/tree-sitter-go-template/tree/master/queries) 

# Languages needing queries

(A parser exists, but queries for highlights, indent, injections, folds, and/or locals are still missing. Some guidance for writing these can be found in [CONTRIBUTING.md](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#parser-configurations) and the [tree-sitter documentation](https://tree-sitter.github.io/tree-sitter/using-parsers#pattern-matching-with-queries).)

* [x] [Agda](https://github.com/tree-sitter/tree-sitter-agda) (#1431)
* [x] [Objective-C](https://github.com/merico-dev/tree-sitter-objc) (#1824)
* [x] [Vimdoc](https://github.com/vigoux/tree-sitter-vimdoc) (#2679)

# Languages needing parsers

(Not even a parser exists yet; if you are familiar with the language and tree-sitter, consider contributing one. You can also help by searching for new parsers in case someone has created by now.)

* [x] luadoc (comments with emmyLua annotations; similar to `jsdoc`/`phpdoc`)
* [x] htmldjango/jinja2
[...]

PR #3303, opened 2022-08-15
parsers: add tree-sitter-gitignore

theHamsta:
Not feeling too confident about the captures yet. Any feedback welcome!

@andy2mrqz 

Addresses https://github.com/nvim-treesitter/nvim-treesitter/issues/2282#issuecomment-1215780654

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1146,6 +1146,16 @@ list.tiger = {
   maintainers = { "@ambroisie" },
 }
 
+list.gitignore = {
+  install_info = {
+    url = "https://github.com/shunsambongi/tree-sitter-gitignore",
+    files = { "src/parser.c" },
+    branch = "main",
+    requires_generate_from_grammar = true,
+  },
+  maintainers = { "@theHamsta" },
+}
+
 local M = {
   list = list,
   filetype_to_parsername = filetype_to_parsername,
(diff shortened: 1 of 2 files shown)


theHamsta:
Please note that `gitignore` is not a known filetype to vim. You need to add something like
```
augroup filetypedetect
    au! BufRead,BufNewFile .gitignore set filetype=gitignore
augroup end
```
to your init.vim. Support for the filetype should be contributed to github.com/vim/vim

shunsambongi:
@theHamsta I was planning on contributing this parser eventually, but I had become a little busy, so I appreciate this PR. I don't know if it would help you at all, but I was playing around with some highlight queries in https://github.com/shunsambongi/gitignore.nvim/blob/0b8108524dbc0d15a34b6e1ca5bbd75b04a5ddfb/queries/gitignore/highlights.scm

shunsambongi:
> Please note that `gitignore` is not a known filetype to vim. You need to add something like
> 
> ```
> augroup filetypedetect
>     au! BufRead,BufNewFile .gitignore set filetype=gitignore
> augroup end
> ```
> 
> to your init.vim. Support for the filetype should be contributed to github.com/vim/vim

I also just wanted to point out that there are a bunch of tools that use ignore files that can be highlighted using the parser being added in this PR. For example, [sharkdp/fd](https://github.com/sharkdp/fd)  (.fdignore), [BurntSushi/ripgrep](https://github.com/BurntSushi/ripgrep) (.rgignore), [prettier](https://prettier.io/docs/en/ignore.html) (.prettierignore), etc. So you would just need to modify the autocmd to support these files.

theHamsta:
Commited your changes

stsewd (approved):

PR #3309, opened 2022-08-16
rename ocaml interface filetype

aspeddro:
Syncs with the filetypes detected by LSP.

https://github.com/neovim/nvim-lspconfig/blob/da7461b596d70fa47b50bf3a7acfaef94c47727d/lua/lspconfig/server_configurations/ocamllsp.lua#L19

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -429,7 +429,7 @@ list.ocaml_interface = {
     location = "interface",
   },
   maintainers = { "@undu" },
-  filetype = "ocamlinterface",
+  filetype = "ocaml.interface",
 }
 
 list.ocamllex = {
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -18,6 +18,7 @@ local filetype_to_parsername = {
   pandoc = "markdown",
   rmd = "markdown",
   cs = "c_sharp",
+  ["ocaml.interface"] = "ocaml_interface"
 }
 
 local list = setmetatable({}, {
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -18,7 +18,7 @@ local filetype_to_parsername = {
   pandoc = "markdown",
   rmd = "markdown",
   cs = "c_sharp",
-  ["ocaml.interface"] = "ocamlinterface",
+  ["ocaml.interface"] = "ocaml_interface",
 }
 
 local list = setmetatable({}, {
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -18,7 +18,7 @@ local filetype_to_parsername = {
   pandoc = "markdown",
   rmd = "markdown",
   cs = "c_sharp",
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -429,7 +429,7 @@ list.ocaml_interface = {
     location = "interface",
   },
   maintainers = { "@undu" },
-  filetype = "ocamlinterface",
+  filetype = "ocaml.interface",
```
theHamsta, line 432: Can you add the old name to `filetype_to_parsername` table?

lua/nvim-treesitter/parsers.lua
```diff
@@ -18,6 +18,7 @@ local filetype_to_parsername = {
   pandoc = "markdown",
   rmd = "markdown",
   cs = "c_sharp",
+  ["ocaml.interface"] = "ocamlinterface",
```
theHamsta, line 21: but the filetype is `ocamlinterface` and parser name is `ocaml_interface`

lua/nvim-treesitter/parsers.lua
```diff
@@ -18,6 +18,7 @@ local filetype_to_parsername = {
   pandoc = "markdown",
   rmd = "markdown",
   cs = "c_sharp",
+  ["ocaml.interface"] = "ocaml_interface",
```
theHamsta, line 21: you mean `["ocamlinterface"] = "ocaml_interface"`

theHamsta: `["ocam.linterface"] = "ocaml_interface" was already mentioned below. `ocaml_interface` is the name in the parser C binary

aspeddro: The filetype detected by LSP is `ocaml.interface`

```lua
filetypes = { 'ocaml', 'ocaml.menhir', 'ocaml.interface', 'ocaml.ocamllex', 'reason', 'dune' }
```

theHamsta: but the filetype that was here before was `ocamlinterface`

undu: When I made the chanes I checked that Vim didn't allow for filetypes to have that dot, hence the type in LSP is different from the vim filetype. I'm not sure what's the current situation.

theHamsta: Vim will fallback to `ocaml` with `ocaml.interface` has not traditional highlighting. I think this behavior was also implemented for nvim-treesitter though we typically recognize the string as a whole for most parsers.

Should we only keep the new filetype? Or will there be users with `ocamlinterface` configured?

undu: Ale is configured to use ocamlinterface, and do the conversion: https://github.com/dense-analysis/ale/blob/a02a4f2811f810877f3c3859cca963f9578ff94a/autoload/ale/handlers/ocamllsp.vim#L7

I opened several PRs to follow that convention, detailed here: https://github.com/ocaml/vim-ocaml/pull/61

IMO changing lspconfig to follow that convention is easier than doing the reverse

aspeddro: Closing because the LSP setup does the conversion ocamlinterface => 'ocaml.interface'

PR #3325, opened 2022-08-19
Update JS for upstream "Optional chain" change

theHamsta:
Follow upstream change in https://github.com/tree-sitter/tree-sitter-javascript/pull/225

Let's see whether TS already has this change

### DIFF
--- a/queries/typescript/highlights.scm
+++ b/queries/typescript/highlights.scm
@@ -26,7 +26,7 @@
       ((import_specifier
           name: (identifier) @type)))))
 
-; punctuation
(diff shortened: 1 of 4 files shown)


queries/typescript/highlights.scm
```diff
@@ -44,6 +44,8 @@
 (pair
   ":" @punctuation.delimiter)
 
+"?." @punctuation.delimiter
```
theHamsta, line 47: typescript does have this change yet

clason (approved):

PR #3328, opened 2022-08-20
injections(lua): highlight string as `query` if starts with `;; query`

numToStr:
Before and After

image

### DIFF
--- a/queries/lua/injections.scm
+++ b/queries/lua/injections.scm
@@ -17,6 +17,6 @@
   (#eq? @_vimcmd_identifier "vim.treesitter.query.set_query"))
 
 ;; highlight string as query if starts with `;; query`
-((string ("string_content") @query) (#lua-match? @query "^%s*;;%s?query"))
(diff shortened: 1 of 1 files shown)


savq:
Should this allow any number of leading semi-colons? ~Scheme~ the query language only requires one semi-colon for comments. Is there any reason in particular 2 should be preferred for this?

numToStr:
> Is there any reason in particular 2 should be preferred for this?

Nope no particular reason.

> Should this allow any number of leading semi-colons? the query language only requires one semi-colon for comments.

Yeah, I could change this to accept multiple leading semi colons.

numToStr:
@savq I changed it to accepts one or more leading `;`

theHamsta (approved):
We had some performance problems with regexes. But strings are probably not that frequent and Lua regexes generally fast (I hope this not for usage for `set_query`. I think we failed at some point when you want to write longer queries in Lua vs dedicated files. We need to improve the query-runtimepath experience )

PR #3333, opened 2022-08-21
ci: add workflow_dispatch to update actions

clason:
Allows more easily triggering these workflows manually (e.g., in case of an important upstream fix).


theHamsta:
So this will add a button "run this action manually"?

theHamsta (approved):

theHamsta:
@clason still awake :laughing: ?

clason:
Yes, exactly:
Screenshot 2022-08-23 at 09 08 45

PR #3345, opened 2022-08-23
feat(meson): add support for meson

Decodetalkers:
I write a treesitter for meson build, and I have tried it myself, I think the treesitter works good

### DIFF
--- a/queries/meson/highlights.scm
+++ b/queries/meson/highlights.scm
@@ -1,0 +1,72 @@
+(comment) @comment
+(number) @number 
+(bool) @boolean
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+  "["
+  "]"
+]
+@punctuation.bracket
+
+[
+  "and"
+  "not"
+  "or"
+  "in"
+] @keyword.operator
+
+[
+  "="
+  "=="
+  "!="
+  "+"
+  "/"
+  "/="
+  "+="
+  "-="
+  ">"
+  ">="
+] @operator
+
+[
+  "if"
+  "elif"
+  "else"
+  "endif"
+] @conditional
+
+[
+  "foreach"
+  "endforeach"
+  (keyword_break)
+  (keyword_continue)
+] @repeat
+
+;;; format
+(string) @string
+["@"] @punctuation.special
+
+(experession_statement
+  object: (identifier) @variable)
+
+(normal_command
+  command: (identifier) @function)
+
+(list
+  variable: (identifier) @variable)
+
+(formatunit
+  variable: (identifier) @variable)
+
+(variableunit
+  value: (identifier) @variable)
+(pair
+  key: (identifier) @property)
+(pair
+  value: (identifier) @variable)
+
+(escape_sequence) @string.escape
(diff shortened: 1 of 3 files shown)


queries/meson/highlights.scm
```diff
@@ -0,0 +1,55 @@
...
+@punctuation.bracket
+
+[
+  "="
+  "=="
+  "!="
+  "and"
+  "+"
+  "+="
+  "not"
```
theHamsta, line 22: `@keyword.operator`

queries/meson/highlights.scm
```diff
@@ -0,0 +1,55 @@
...
+[
+  "foreach"
+  "endforeach"
+  (keyword_break)
+  (keyword_continue)
+] @repeat
+
+;;; format
+(string) @string
+["@"] @keyword
```
theHamsta, line 41: keyword?

queries/meson/highlights.scm
```diff
@@ -0,0 +1,55 @@
...
+
+(experession_statement
+	object: (identifier) @variable)
+
+(normal_command
+	command: (identifier) @function)
+
+variable: (identifier) @variable
```
theHamsta, line 49: I think here is something missing

queries/meson/highlights.scm
Decodetalkers: > keyword?

So what is better?, it is this one
image

a format keyword

theHamsta: But that's not a keyword in the sense of a reserved language keyword. Maybe you could use `@punctuation.special` like bash and python use for string interpolation?

queries/meson/highlights.scm
theHamsta: Same for `"and"`

queries/meson/highlights.scm
Decodetalkers: > But that's not a keyword in the sense of a reserved language keyword. Maybe you could use `@punctuation.special` like bash and python use for string interpolation?

Thanks

queries/meson/highlights.scm
Decodetalkers: > Same for `"and"`

What is same for and mean?

Decodetalkers: friendly ping @@

theHamsta: `@keyword.operator` for `"not"` and `"and"`. `@keyword` for all signs

theHamsta: does it have `"or"`?

Decodetalkers: > does it have `"or"`?

Thanks, it does have, I have forgotten it

theHamsta:
Thank you!

theHamsta (approved):

theHamsta:
You may still add  the standard `(comment) @comment` injection that many languages have.

Decodetalkers:
> You may still add  the standard `(comment) @comment` injection that many languages have.

I will make a PR soon

PR #3346, opened 2022-08-23
feat(vim): update to latest highlights

vigoux:


queries/vim/highlights.scm
```diff
@@ -134,6 +145,13 @@
   "clear"
 ] @keyword)
 
+;; Command command
+
+(command_attribute name: _ @property)
+(command_attribute
+  val: (behavior
+    (identifier)? @function) @constant)
```
vigoux, line 153: Just notived that this is a bit weird, do you think we can keep it as is ?

clason (approved):

queries/vim/highlights.scm
vigoux: Let's keep it like this for now, I'll work on making that nicer later.

Issue #3350, opened 2022-08-23
Improve information about javascript files with flow annotations


buschco:
**Is your feature request related to a problem? Please describe.**
I spend like one day trying to find out how to highlight my javascript files with flow type annotations:

- [tree-sitter-typescript](https://github.com/tree-sitter/tree-sitter-typescript)'s tsx parser can handle flow typed files -> I asked them to add this info to their readme (https://github.com/tree-sitter/tree-sitter-typescript/pull/219)
- you need to tell `nvim-treesitter` to use the tsx parser for flow typed files like so:

```
au BufRead * if getline(1) =~ '// @flow' | setlocal ft=javascriptreact | endif

lua << EOF
local ft_to_parser = require"nvim-treesitter.parsers".filetype_to_parsername
ft_to_parser.javascriptreact = "tsx"
EOF
```
(I use `javascriptreact` here because [coc-flow only supports `javascript` or `javascriptreact`](https://github.com/amiralies/coc-flow/blob/master/package.json#L22). `javascriptflow` would be better imo).

As described in the README [`vim.filetype.add()`](https://neovim.io/doc/user/lua.html#vim.filetype.add()) could be an good alternative to the `au BufRead ...` part but I think nvim is not quite ready yet as described [here](https://github.com/neovim/neovim/issues/19909).

**Describe the solution you'd like**
This PR linking to this issue, so others can highlight their flow typed files with less 'effort'

**Describe alternatives you've considered**
Add this info to the [tree-sitter-typescript](https://github.com/tree-sitter/tree-sitter-typescript) README but its too nvim-treesitter related imo.

PR #3351, opened 2022-08-23
Improve information about javascript files with flow annotations #3350

buschco:


theHamsta:
The README list is updated automatically

PR #3358, opened 2022-08-24
injections(jsx): highlight `<style jsx>` as css

numToStr:
~Although this is only valid for jsx/tsx, I am putting this in `ecma` so both can inherit the injections.~ https://github.com/nvim-treesitter/nvim-treesitter/pull/3358#issuecomment-1225340609

Created `jsx/injections.scm` with the injection query which is (and should be) then inherited by `javascript` and `tsx` injections.

For some reason, `javascript` was already importing the `jsx` injections, thus no changes required.

---

Before and After

image

### DIFF
--- a/queries/jsx/injections.scm
+++ b/queries/jsx/injections.scm
@@ -1,0 +1,8 @@
+; Styled Jsx <style jsx>
+(jsx_element
+  (jsx_opening_element
+    (identifier) @_name (#eq? @_name "style")
+    (jsx_attribute) @_attr (#eq? @_attr "jsx"))
+  (jsx_expression (template_string) @css
+    (#offset! @css 0 1 0 -1))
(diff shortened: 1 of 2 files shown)


numToStr:
(CI failure looks unrelated to this PR)

numToStr:
It seems putting this in ecma won't work after all, as `typescript` parser is throwing error for this injection so I might have to put this into `javascript` and `tsx` parser individually. Thoughts on how to handle this better?

theHamsta:
> For some reason, javascript was already importing the jsx injections, thus no changes required.

Yes, javascript already contains the jsx grammar, only typescript is strangely splitted into to grammars with and without jsx.

theHamsta (approved):
Thank you

PR #3365, opened 2022-08-25
fix(hl_map): adapt to upstream changes on master

clason:
https://github.com/neovim/neovim/pull/19931 adds direct support for highlighting capture groups (with fallback), obviating the need for `vim.treesitter.highlighter.hl_map`. Instead of

`hl_map["@keyword"] = "TSKeyword"`

users can simply

`hi link TSKeyword @keyword`

Check for the existence of `hl_map` and either use the former or the latter.

@bfredl

### DIFF
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -6,116 +6,133 @@ local configs = require "nvim-treesitter.configs"
 
 local M = {}
 
-local hlmap = vim.treesitter.highlighter.hl_map
-
 -- nvim-treesitter Highlight Group Mappings
 -- Note: Some highlight groups may not be applied upstream, some may be experimental
+-- TODO(clason): deprecated and will be removed for Nvim 0.8
 
-hlmap["annotation"] = "TSAnnotation"
+local default_map = {
+  ["annotation"] = "TSAnnotation",
 
-hlmap["attribute"] = "TSAttribute"
+  ["attribute"] = "TSAttribute",
 
-hlmap["boolean"] = "TSBoolean"
+  ["boolean"] = "TSBoolean",
 
-hlmap["character"] = "TSCharacter"
-hlmap["character.special"] = "TSCharacterSpecial"
+  ["character"] = "TSCharacter",
+  ["character.special"] = "TSCharacterSpecial",
 
-hlmap["comment"] = "TSComment"
+  ["comment"] = "TSComment",
 
-hlmap["conditional"] = "TSConditional"
+  ["conditional"] = "TSConditional",
 
-hlmap["constant"] = "TSConstant"
-hlmap["constant.builtin"] = "TSConstBuiltin"
-hlmap["constant.macro"] = "TSConstMacro"
+  ["constant"] = "TSConstant",
+  ["constant.builtin"] = "TSConstBuiltin",
+  ["constant.macro"] = "TSConstMacro",
 
-hlmap["constructor"] = "TSConstructor"
+  ["constructor"] = "TSConstructor",
 
-hlmap["debug"] = "TSDebug"
-hlmap["define"] = "TSDefine"
+  ["debug"] = "TSDebug",
+  ["define"] = "TSDefine",
 
-hlmap["error"] = "TSError"
-hlmap["exception"] = "TSException"
+  ["error"] = "TSError",
+  ["exception"] = "TSException",
 
-hlmap["field"] = "TSField"
+  ["field"] = "TSField",
 
-hlmap["float"] = "TSFloat"
+  ["float"] = "TSFloat",
 
-hlmap["function"] = "TSFunction"
-hlmap["function.call"] = "TSFunctionCall"
-hlmap["function.builtin"] = "TSFuncBuiltin"
-hlmap["function.macro"] = "TSFuncMacro"
+  ["function"] = "TSFunction",
+  ["function.call"] = "TSFunctionCall",
+  ["function.builtin"] = "TSFuncBuiltin",
+  ["function.macro"] = "TSFuncMacro",
 
-hlmap["include"] = "TSInclude"
+  ["include"] = "TSInclude",
 
-hlmap["keyword"] = "TSKeyword"
-hlmap["keyword.function"] = "TSKeywordFunction"
-hlmap["keyword.operator"] = "TSKeywordOperator"
-hlmap["keyword.return"] = "TSKeywordReturn"
+  ["keyword"] = "TSKeyword",
+  ["keyword.function"] = "TSKeywordFunction",
+  ["keyword.operator"] = "TSKeywordOperator",
+  ["keyword.return"] = "TSKeywordReturn",
 
-hlmap["label"] = "TSLabel"
+  ["label"] = "TSLabel",
 
-hlmap["method"] = "TSMethod"
-hlmap["method.call"] = "TSMethodCall"
+  ["method"] = "TSMethod",
+  ["method.call"] = "TSMethodCall",
 
-hlmap["namespace"] = "TSNamespace"
+  ["namespace"] = "TSNamespace",
 
-hlmap["none"] = "TSNone"
-hlmap["number"] = "TSNumber"
+  ["none"] = "TSNone",
+  ["number"] = "TSNumber",
 
-hlmap["operator"] = "TSOperator"
+  ["operator"] = "TSOperator",
 
-hlmap["parameter"] = "TSParameter"
-hlmap["parameter.reference"] = "TSParameterReference"
+  ["parameter"] = "TSParameter",
+  ["parameter.reference"] = "TSParameterReference",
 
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/highlight.lua
```diff
@@ -6,116 +6,229 @@ local configs = require "nvim-treesitter.configs"
...
+    ["@text.environment"] = "TSEnvironment",
+    ["@text.environment.name"] = "TSEnvironmentName",
+
+    ["@text.note"] = "TSNote",
+    ["@text.warning"] = "TSWarning",
+    ["@text.danger"] = "TSDanger",
+
+    ["@todo"] = "TSTodo",
+
+    ["@type"] = "TSType",
```
bfredl, line 220: should the 0.8 branch really do this for captures that now will be handled in core, like `@type`, `@string` and so on? (while nvim-treesitter still handling its own groups, like `@text.emphasis` , of course)

clason: There is no 0.8 branch...

But yes, this is purely a compatibility shim; after 0.8 is released with core highlighting, we'll have to revisit the whole thing and simplify drastically.

(There may still be a good reason to keep the now well-established `TSFoo` highlight groups around, for the sake of color schemes -- better to have all if we want to keep some.)

bfredl: Ok, so at a later point (after 0.8), this entire section could be made opt-in rather than opt-out (or removed entirely, provided there is another solution for the special groups that are regardless needed).

clason: Indeed; that is a strategic decision on how far the `TSFoo` groups are considered "branding".

lua/nvim-treesitter/highlight.lua
```diff
@@ -6,116 +6,229 @@ local configs = require "nvim-treesitter.configs"
...
-hlmap["keyword.return"] = "TSKeywordReturn"
+  hlmap["type"] = "TSType"
+  hlmap["type.builtin"] = "TSTypeBuiltin"
+  hlmap["type.qualifier"] = "TSTypeQualifier"
+  hlmap["type.definition"] = "TSTypeDefinition"
 
-hlmap["label"] = "TSLabel"
+  hlmap["variable"] = "TSVariable"
+  hlmap["variable.builtin"] = "TSVariableBuiltin"
+else
```
theHamsta, line 120: can't this be done in one branch only and the if/else only what `hl_map` is and what should be done with it. You could also just define our own source-of-truth table and the either modify `hl_map` with it or set the mapping. Would also be fine to invert the mapping for post-hl_map.

theHamsta: you can add the `@` by string concat. would also be fine to for me to make setting `TS...` mapping opt-in now for post hl_map nvims

clason: I'm sorry, but I don't understand any of this.

`hl_map` in `vim.treesitter` will go away, never to return. Until 0.8 is released, we still need to support it -- unless we want to got master only?

I do _not_ want to be clever about something that will be removed within weeks.

theHamsta: I'd would have just kept one mapping from `@X` -> `Y` or `X` -> `Y`. Having  `@X` -> `Y` and `X` -> `Y` is redundant.

I'd rather remove it right away and provide a "but I need TS..." mappings fallback. We can leave the fallback in our code but there is no need to make anyone on a nightly commit stability guarantees. Better let them add call a compat function and give them some time to migrate. Then, we don't need another PR in two weeks

clason: OK, if you want to break 0.7, be my guest.

theHamsta: but 0.7 has `hl_map` so your branch decision should work. I would just break nightly till two weeks before 0.8

clason: What we do when 0.8 is released is a separate question (and depends on what 0.8 ships with, which is not clear yet).

As I wrote earlier, _this_ PR is only concerned with keeping exactly the same (previous) behavior on 0.7 and master (after the Neovim PR lands).

If nvim-treesitter decides to drop `TS...` highlight groups like a hot potato, that is fine; ripping out the both branches (and `custom_captures` support) is a trivial PR.

theHamsta: > If nvim-treesitter decides to drop TS... highlight groups like a hot potato, that is fine; ripping out the both branches (and custom_captures support) is a trivial PR.

but it shouldn't. We should just stop intrusively without being asked changing mapping when nvim now uses `@foo` captures and it's meant to be used that way in future. We need to provide `TS...` also in future if the user needs it as not every color scheme will update. I don't see a urgent need to adapt to a nightly in case that's not settled yet. But for me it seems to be pretty sure that `@foo` will be supported and that it has already a mapping set.

If you think it is more convenient for users you can also just manually call the opt-in function for now. Nvim core could go the `@type -> TSType -> Type` route if that should be supported in future for all setups. I think the intention of upstreaming is that it should make less of a difference whether users use nvim-treesitter or just pure core.

clason: > Nvim core could go the @type -> TSType -> Type route if that should be supported in future for all setups.

But it **won't**; that's the whole point. You either map the capture directly, or rely on the default links to the standard Vim (not nvim-treesitter!) highlight groups.

And that _is_ settled and will happen within days (including removal of `hl_map`), at which point users on master will suddenly lose half their highlighting.

This PR seemed to me the simplest way to prevent that from happening; if you have better and more complex plans, go ahead and push to this PR (or make your own). Or not, and deal with the fallout.

theHamsta: staying with `TS...` would be also an option but then I wouldn't change it for 0.8. Being nightly is precisely to let users experience how 0.8 will behave.

theHamsta: > And that is settled and will happen within days (including removal of hl_map), at which point users on master will suddenly lose half their highlighting.

they'll still have what upstream nvim offers and I don't get how letting the breaking change happen now vs when master release changes anything. Again, I'm open whether we go opt-in or opt-out but we can make the decision also right now.

clason: The whole point of this PR is that it's _not_ a breaking change (for nvim-treesitter)! 

And we need both branches to keep support for 0.7, so I don't see how the timing in relation to when the Neovim PR is merged matters here?

theHamsta: Could we still just use one table?

clason: Oh, _that_ is your point here? Sure, if you prefer. I just didn't to spend much time on something that will be removed in few weeks anyway when 0.8 is released and we drop support for 0.7.

(I also don't want to hold this PR up by bike-shedding on the option name.)

theHamsta: > on something that will be removed in few weeks anyway when 0.8 is released and we drop support for 0.7.

ok, so you would then remove the one half of the duplication...I would have left in a way that it worked with both and never bother again.

I thought you wanted to remove both halfs. My argument was just that I would see our plan for 0.8 in the new half. But we can post-pone that for another PR. Again for a version change, it would be ok for me to drop `TS...` in order to push the new `@...` way while providing people a way to opt into `TS...` if they need it. We can also do this transition post 0.8

theHamsta (approved):
Should be ok when you remove the one half of the duplication after 0.8 release.

lua/nvim-treesitter/highlight.lua
clason: Agreed. (And this _needs_ to be done post-0.8 since we don't know what will be in 0.8 until then...)

I also think that long-term, we should deprecate the custom `TS...` highlight groups in favor of "just map captures directly". That needs to be a soft deprecation, though, since `@foo.bar` highlight groups trigger a warning in 0.7, so colorschemes won't be able to (easily) include it right away.

So this needs some thought and a clearly documented transition guide for users and downstream plugin/colorscheme authors.

theHamsta: > I also think that long-term, we should deprecate the custom TS... highlight groups in favor of "just map captures directly". That needs to be a soft deprecation, though, since @foo.bar highlight groups trigger a warning in 0.7, so colorschemes won't be able to (easily) include it right away.

wouldn't providing a function to bring them back be soft enough? This is what I was proposing. It requires a manual action, so people can think whether they can adopt the new approach right away or whether they need the old highlights maybe even for a longer time. Sure, that's a bit more pushy than doing this step when some colorschemes already switched.

clason: Yes, of course, _after_ 0.8 is released. We should also give users and especially plugin/colorscheme authors some time to adapt before we open the floodgates for the inevitable "your colorscheme broke!!!eleven!" complaints.

PR #3377, opened 2022-08-27
feat(agda): better highlight

Decodetalkers:
Add highlight for bool

add highlight for function

Log: better highlight

### DIFF
--- a/queries/agda/highlights.scm
+++ b/queries/agda/highlights.scm
@@ -20,15 +20,16 @@
 ; Set
 (SetN) @type.builtin
 
+(expr . (atom) @function)
+
+((atom) @boolean
+  (#any-of? @boolean "true" "false" "True" "False"))
 
 ;; Imports and Module Declarations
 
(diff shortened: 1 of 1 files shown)


Decodetalkers:
before
image

after
image

queries/agda/highlights.scm
```diff
@@ -20,16 +20,17 @@
 ; Set
 (SetN) @type.builtin
 
+(expr . (atom) @function)
+
+((atom) @boolean
+  (#vim-match? @boolean "(true|false|True|False)$"))
```
theHamsta, line 26: `(#any-of? @boolean "true" "false"  "True" "False")`

Decodetalkers: > `(#any-of? @boolean "true" "false" "True" "False")`

Like this?
image

theHamsta:
Thank you!

theHamsta (approved):

PR #3380, opened 2022-08-27
chore(meson): add injections for meson

Decodetalkers:
add comment as injections as other languages like bash

Log: add injections for meson


theHamsta (approved):
Thank you!

PR #3390, opened 2022-08-27
Haskell injection: support inline json

guibou:
Support for inline json using the quasiquoter `aesonQQ`: https://github.com/sol/aeson-qq#readme

### DIFF
--- a/queries/haskell/injections.scm
+++ b/queries/haskell/injections.scm
@@ -56,3 +56,13 @@
  (#eq? @_name "hsx")
  ((quasiquote_body) @html)
 )
+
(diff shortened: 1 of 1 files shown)


theHamsta (approved):
Thank you!

theHamsta:
The tests are not happy :cry:

guibou:
> The tests are not happy

Yes, because I changed the name to match what's already in the file at the latest second and even when proof-reading the 3 lines diff 10 times, I did not spot the mistake. In short, I'm stupid, lazy, ... ;(

Sorry about that. Should be fixed now (hopefully I had not introduced another simple mistake)

PR #3392, opened 2022-08-29
Fix broken SQL Keyword Highlights

matthias-Q:
With the [commit](https://github.com/DerekStride/tree-sitter-sql/commit/aa9934d5d2b8dbe25b17f29afcfcb3d26340f325) some keywords have been split up to account for possible multiple spaces in between (e.g. `group..by` instead of `group.by`, spaced denoted by dots here). 
This removed the keyword `keyword_group_by` and replaced it with `keyword_group` and `keyword_by`. This also affects `order by`.

Additionally, window functions are now parse-able, thus some related keywords have been added.


theHamsta (approved):
Thank you!

theHamsta:
I think you need to modify the SHA for the SQL parser I. Lockfile.json

matthias-Q:
I have updated the lockfile, but I think you have to approve it again. @theHamsta

PR #3393, opened 2022-08-29
Add postresql-simple sql to haskell injections

MrcJkb:
This PR adds a haskell injection for [postgresql-simple](https://hackage.haskell.org/package/postgresql-simple) quasiquotes.

### DIFF
--- a/queries/haskell/injections.scm
+++ b/queries/haskell/injections.scm
@@ -66,3 +66,13 @@
   ((quasiquote_body) @json)
 )
 
+
+;; -----------------------------------------------------------------------------
+;; SQL
+
+; postgresql-simple
+(quasiquote
+  (quoter) @_name 
+  (#eq? @_name "sql")
+  ((quasiquote_body) @sql)
+)


theHamsta (approved):
Thank you!

queries/haskell/injections.scm
```diff
@@ -56,3 +56,13 @@
...
  ((quasiquote_body) @html)
 )
+
+;; -----------------------------------------------------------------------------
+;; SQL
+
+; postgresql-simple
+(quasiquote
+  (quoter) @_name 
+  (#any-of? @_name "s" "sql")
```
theHamsta, line 66: isn't `s` a bit ambiguous? It could also be something else

MrcJkb: Yes, `s` could conflict with a quasiquoter from another library. For some reaon I've seen `s` used with that library (maybe in older versions?), although `sql` is the quasiquoter that removes whitespace.

Another option could be to match on the `query*` and `execute*` functions:

```
;; qualified postgresqul-simple functions
(exp_apply
  (exp_name
    (qualified_variable
      (((variable) @_func_name)(#any-of? @_func_name "query" "query_" "queryWith" "queryWith_" "execute" "execute_" "executeMany"))
    )
  )
  (exp_name) ;; the Connection argument
  (quasiquote
    ((quasiquote_body) @sql)
  )
)
;; unqualified postgresqul-simple functions
(exp_apply
  (exp_name
    (((variable) @_func_name)(#any-of? @_func_name "query" "query_" "queryWith" "queryWith_" "execute" "execute_" "executeMany"))
  )
  (exp_name) ;; the Connection argument
  (quasiquote
    ((quasiquote_body) @sql)
  )
)
```

MrcJkb: I have removed the `s` match.

theHamsta: I have too few experience with the Haskell ecosystem to judge whether those injections would be unambiguous. Maybe leave them for your personal config?

Issue #3330, opened 2022-08-20
Enable html parser for htmldjango filetype


ambadyanands:
**Is your feature request related to a problem? Please describe.**
Currently there's no TreeSitter support for `htmldjango` files by default. I couldn't find any dedicated parser for [django template language](https://docs.djangoproject.com/en/dev/ref/templates/language/) either.

**Describe the solution you'd like**
Enable HTML parser for `htmldjango` files. This will also help to use plugins like [nvim-ts-autotag](https://github.com/windwp/nvim-ts-autotag) that depends on TreeSitter for working with HTML tags.

**Describe alternatives you've considered**
Currently I use the following to enable HTML parser for `htmldjango` files:
`require"nvim-treesitter.parsers".filetype_to_parsername.htmldjango = 'html'`

It'll be great if this can be enabled by default since the above solution may not seem obvious for all (I would haven't figured this out myself without the help of this [suggestion here](https://github.com/windwp/nvim-ts-autotag/issues/48#issuecomment-1220647167)). 

**Additional context**
Going through issue https://github.com/nvim-treesitter/nvim-treesitter/issues/74, I came to know that [`C++` extends `C` grammar](https://github.com/nvim-treesitter/nvim-treesitter/pull/63). Maybe the same can be applied here too? i.e. to extend `html` for `htmldjango` files.

Thanks.

theHamsta:
Adding `htmldjango` #2282

theHamsta:
> Going through issue https://github.com/nvim-treesitter/nvim-treesitter/issues/74, I came to know that https://github.com/nvim-treesitter/nvim-treesitter/pull/63. Maybe the same can be applied here too? i.e. to extend html for htmldjango files.

This would be one approach. The problem could be that `{{ something }}` may occur in almost every position of the HTML grammar. One might add it as an extra much like comments in `C`, but another approach is to follow the strategy of ERB parser which is currently not fully supported by Neovim. Then, you would only parse the templates and look for `{{ foo }}` and inject HMTL in everything else except your child nodes (which are `{{ foo }}`).

ambadyanands:
@theHamsta Thanks, I understand what you mentioned last seems to be the best approach. My suggestion, on the other hand, was only a temporary workaround until there's proper support since I think it's better than no support at all.

theHamsta:
Feel free to open a PR!

PR #3402, opened 2022-08-30
parsers: use html parser for django template language

ambadyanands:
Temporarily enable HTML parser for `htmldjango` filetype as suggested in https://github.com/nvim-treesitter/nvim-treesitter/issues/3330 until there's a proper parser for [Django Template Language](https://docs.djangoproject.com/en/dev/ref/templates/language/).


theHamsta:
Thank you!

PR #3405, opened 2022-08-31
feat: add tree-sitter-sxhkdrc

RaafatTurki:
This adds my recently created [sxhkdrc grammar](https://github.com/RaafatTurki/tree-sitter-sxhkdrc) which is the configuration format for the [simple x hotkey daemon](https://github.com/baskerville/sxhkd)

I've also added highlighting queries to make it useful from the get go:
image

At 1st I've had my highlight queries within the grammar repo itself and `tree-sitter highlight` worked using it but nvim wouldn't use it (`checkhealth nvim-treesitter` says my grammar has no highlighting support), Only after moving my queries into this plugin did nvim start highlighting things. I'm relatively new to the nvim-treesitter scene, Is there a way to relocate my queries back into my repo and have nvim-treesitter use them (that way I can update the queries without PRing here) or is that a no no?

*before*
image
*after*
image

Making those queries was a bit unusual as this isn't a programming/markup language so let me know if I'm linking a query against a really odd highlight group.

Also currently I must set the local `filetype` with something like `au BufEnter sxhkdrc,*.sxhkdrc set ft=sxhkdrc` as nvim doesn't really recognize the `sxhkdrc` filetype, how do other parsers overcome such issue?

I plan on adding folding support and unit tests in the future
LMK what you think!

### DIFF
--- a/queries/sxhkdrc/injections.scm
+++ b/queries/sxhkdrc/injections.scm
@@ -1,0 +1,3 @@
+(command) @bash
+
+(comment) @comment
--- a/queries/sxhkdrc/injections.scm
+++ b/queries/sxhkdrc/injections.scm
@@ -1,3 +1,1 @@
-(command) @bash
-
 (comment) @comment
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1178,7 +1178,7 @@ list.tiger = {
 
 list.sxhkdrc = {
   install_info = {
-    url = "~/sectors/treesitter/tree-sitter-sxhkdrc",
+    url = "https://github.com/RaafatTurki/tree-sitter-sxhkdrc",
     files = {"src/parser.c"},
     branch = "master",
     generate_requires_npm = false,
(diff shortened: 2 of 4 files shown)


RaafatTurki:
Adding folds and injections was much easier than I thought however each come with it's own little issue
folds: currently they count the post whitespace as a part of the fold
*expanded*
image
*collapsed*
image

injections: sxhkdrc commands are almost bash code but not exactly bash hence some issue arise when injecting bash, I think I'll revert the bash injection
image

theHamsta:
> [simple x hotkey daemon](https://github.com/baskerville/sxhkd)

now it gets really specific :smile: I think you need to run `stylua`

theHamsta (approved):

queries/sxhkdrc/injections.scm
```diff
@@ -1,3 +1 @@
-(command) @bash
-
```
theHamsta: Was this on purpose? Just asking since you removed it in the stylua commit

RaafatTurki: Ah, removing it actually had nothing to do with stylua, forgot to split that into its own commit.
Yes it was on purpose, bash injections don't play that nice with sxhkdrc commands

PR #3409, opened 2022-08-31
fix(highlight): ensure links are rebuilt on colorscheme change

akinsho:
This PR fixes an issue where, for users on nvim nightly, the links for the captures are not reset when a user reloads or changes their colorscheme. This was previously done for the default highlight groups, but this logic did not include creating the links now need on 0.8. Not sure if this code is all destined for a fire at some point, but it's useful for people who are doing this in the short term.

### DIFF
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -136,6 +136,8 @@ local function link_all_captures()
   end
 end
 
+link_all_captures()
+
 local function should_enable_vim_regex(config, lang)
   local additional_hl = config.additional_vim_regex_highlighting
   local is_table = type(additional_hl) == "table"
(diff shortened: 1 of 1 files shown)


theHamsta (approved):
Thank you! Can you fix the lint's?

lua/nvim-treesitter/highlight.lua
```diff
@@ -177,6 +179,9 @@ function M.set_custom_captures(captures)
 end
 
 function M.set_default_hlgroups()
+  if not ts.highlighter.hl_map then
+    link_all_captures()
```
theHamsta, line 183: I think this has to run at least once

akinsho: @theHamsta it gets called in the plugin file

https://github.com/nvim-treesitter/nvim-treesitter/blob/549723ff7ba495dbda580fffebb358fbfbda443e/plugin/nvim-treesitter.lua#L44

so it will get called when all the other highlights get set

akinsho:
The linting all seems fine, I think the issue is that the tests are expecting pre 0.8 highlights, are these expected to match nightly? also seems not to have been flagged earlier

PR #3411, opened 2022-09-01
[Ruby] Add global variable highlighting

CKolkey:
Adds global variable highlights to ruby.

### DIFF
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -1,5 +1,6 @@
 ; Variables
 (identifier) @variable
+(global_variable) @variable.global
 
 ; Keywords
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -165,6 +165,7 @@ effect on highlighting. We will work on improving highlighting in the near futur
 ```
 @variable
 @variable.builtin
+@variable.global
 ```
 
 #### Text


queries/ruby/highlights.scm
```diff
@@ -1,5 +1,6 @@
 ; Variables
 (identifier) @variable
+(global_variable) @variable.global
```
theHamsta, line 3: you would need to add that to `CONTRIBUTING.md` or adapt the CI script to only check for prefixes (here `@variable`

clason: Now that fallbacks are in Neovim, we can probably stick to top-level captures now?

CKolkey: Thanks, I've added it.

theHamsta: I would still keep some standardized prefixes, but be more liberal with some per-suffixes when it could be interesting to distinguish one feature of a language. We don't need to set highlights as `TS*` will get deprecated. I wonder whether `@variable.global` is general enough to add to more languages (probably not needed). The check coul

theHamsta (approved):

queries/ruby/highlights.scm
theHamsta: @clason was your suggestion to not use captures like the proposed here?

clason: Sorry, I just realized the comment was ambiguous. By all means, use specialized captures in queries -- my suggestion was exactly as you said, to relax the CI check to only consider the top-level captures (so anything `@variable.foo` passes as long as `@variable` is allowed).

CKolkey:
Should `["variable.global"] = "TSVariableGlobal"` in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/highlight.lua#L13

I see the deprecation notice though, perhaps there's another way (or this shouldn't be done at all)

queries/ruby/highlights.scm
theHamsta: @clason I just wanted to relax the capture check to only check the prefix. For testing, I added a suffix it got accepted, I wanted to introduce a typo but also the typo for a known capture got accepted. Should we still document the specializations somewhere to ensure that the capture was at least intentional? We also introduced the check because there were often synonyms or typos that were intended to be another capture. Would also be fine to allow all suffixes

PR #3415, opened 2022-09-01
Add support for menhir (OCaml parser generator)

Kerl13:
Hi!

This patch adds support for [menhir](http://gallium.inria.fr/~fpottier/menhir/) files (menhir is a parser generator for [OCaml](https://ocaml.org/)). The parser has been used a little outside of nvim already, and is reasonably tested. I hope it can be a nice addition to nvim-treesitter :slightly_smiling_face: 

Let me know if I did something wrong in the PR, I'll be happy to fix it

cheers!

### DIFF
--- a/queries/menhir/highlights.scm
+++ b/queries/menhir/highlights.scm
@@ -1,0 +1,29 @@
+["%parameter" "%token" "%type" "%start" "%attribute" "%left" "%right" "%nonassoc" "%public" "%inline" "%prec"] @keyword
+["%on_error_reduce"] @exception
+
+["let"] @keyword.function
+
+[(equality_symbol) ":" "|" ";" ","] @punctuation.delimiter
+
+["=" "~" "_"] @operator
+(modifier) @operator
+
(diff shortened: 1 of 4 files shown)


queries/menhir/highlights.scm
```diff
@@ -0,0 +1,29 @@
+["%parameter" "%token" "%type" "%start" "%attribute" "%left" "%right" "%nonassoc" "%public" "%inline" "%prec"] @keyword
+["%on_error_reduce"] @exception
+
+["let"] @keyword.function
```
theHamsta, line 4: does let start a function or just a grammar rule?

theHamsta (approved):
Looks good to me :+1:

queries/menhir/highlights.scm
Kerl13: It's just a grammar rule.
There are no functions in menhir (well, you can have parameterised grammar rules which kinda look like functions but in the end it's a grammar rule).
Do you think `@keyword` would be more appropriate?

theHamsta: it should be ok

PR #3417, opened 2022-09-01
feat: add luau

williamboman:
Heavily derived from the Lua queries, with modifications.


williamboman:
Need to give this a bit more time. Indentation rules seems a bit off.

PR #3422, opened 2022-09-03
feat(qmljs): boolean

Decodetalkers:
boolean is not in the right color, so fix it

Log: fix the color of boolean for qmljs


theHamsta (approved):

theHamsta:
Thank you!

PR #3429, opened 2022-09-04
injections(make): exclude children for bash injection

theHamsta:


theHamsta:
Ah, I know why this isn't working. Neovim still creates a new bash parsed region for each of the bash injection instead of having just one bash parser with multiple injected ranges.

clason:
superseded by switch to upstream injections

PR #3434, opened 2022-09-05
highlights(swift): fix for_statement variable declaration

alex-pinkus:
Incorporates a fix from upstream, caught due to the emacs tree-sitter plugin using an older version of tree-sitter that does stricter validation.﻿


alex-pinkus:
See: https://github.com/alex-pinkus/tree-sitter-swift/pull/221

clason (approved):

PR #3442, opened 2022-09-06
feat: Fold sublists instead of lists in markdown

ixru:
I think it makes more sense to fold sublists instead as the items relate to each other.
Could keep both, but I find it unexpected and noisy.


theHamsta:
@MDeiml

MDeiml:
I don't have any strong opinion on this, but I don't see any reason to disable folding for top level lists. Sublists can be folded either way.

ixru:
I don't know how to do that, but if it is possible to discern between toplevel and sub, I would like that too.
Removed it because it makes it harder to interact with the intended fold.
image
Here the inner list also gets a fold, which gets in the way as your intent would probably be to close the outer one.

MDeiml:
Ah, I understand your intend now. I agree that that would be better. You probably want to fold `list_item` though, not sublists. I'd suggest something like.

```
(section (list))
(list_item)
```

ixru:
Yes good idea with the section. I don't understand what you mean with the second one though.
I'm winging it, but I imagine sublists are what I want, that would be fold number 2 and 3 in the expectation above. 
Likely I misunderstand something as the query doesn't yield what I expect, that being (list_item (list)) capturing one line below as shown in the bug.

ixru:
I think it is correct now, it won't be any more broken than it already is.

I tried looking at the external scanner, but it was too complex for me. 
https://github.com/MDeiml/tree-sitter-markdown/blob/e375ba95ff9a12418f9b9e7c190f549d08b5380a/tree-sitter-markdown/src/scanner.cc#L419
If BLOCK_CONTINUATION is meant for paragraph continuation, I think maybe this needs to check for STATE_WAS_SOFT_LINE_BREAK.
Or maybe it has to do with the extra_indentation added when pushing to open_blocks.
Idk if it it helps, you probably have a better grasp of it.

ixru:
I'm not sure, I found what caused the bug, but I'm not sure how block_continuation is supposed to work.

This is a bit confusing as I believe there are two errors. Looking at the image above, on line 13 the indentation is split, some belongs to the paragraph above, some to the list_marker_star. I think the split is caused by [this](https://github.com/MDeiml/tree-sitter-markdown/blob/e375ba95ff9a12418f9b9e7c190f549d08b5380a/tree-sitter-markdown/src/scanner.cc#L111) (which I think should be block + 2), but the confusion is in which of them is supposed to own the indentation. Neither the paragraph owning it, nor including it in the list_marker_star seems correct. I would expect it to belong to the list_item, as a node before the list_marker_star. At line 16 you see the same, the paragraph above owning indentation even though it is not a continuation of the paragraph. This causes the folding to get messed up, as the line is counted as part of the marked list_item at line 13. You can see the same thing happening for all paragraphs in the list in the playground.

MDeiml:
Hm I see. So all the folds are one line to big because the indentation of the next line belongs to the last list item.

I'll see if there is something that can be done, but it's not because of the external parser, rather which node "takes ownership" of the "block continuation" (in this case indentation)

ixru:
Yes exactly.  There's no sense in what I said about the scanner, it was just a last ditch effort. If you can't fix it, please dump your thoughts on what you think is wrong so that I have something to pull on.

MDeiml:
It's just that the beginning of the next line is counted as part of the previous list item. On itself I think that makes some sense, I'm not sure if this is easy or hard to change. You would need to somehow make the block continuation part of the next node.

A far easier fix would be to use the `offset` directive in the fold query to remove the last line, I think it should be sth like adding `offset! 0 0 0 -1` after the capture name, but you can look at the nvim documentation for more info.

amaanq:
Status of this @ixru? Is it still useful to add?

ixru:
Sorry I seem to have forgotten. It works better than what is now, but they are both still wrong because of the above issue. Merge or close as you wish.

amaanq:
try using `(#offset! @fold 0 0 0 -1)` as suggested - if you can get that to work better or as-wanted I'll merge, otherwise close

ixru:
Offset cannot work as not all sub lists contain whitespace of the proceeding list. I tried some other things too with very inconsistent results, hard to work with. It needs to be fixed in parser I think.

MDeiml:
I'm open. Can you give an example on what the parser should output in your opinion? Nodes that (mostly) range from beginning of the line until including the newline?

MDeiml:
Also maybe just open an issue for this in the parser repo

ixru:
I will do that

clason:
Setting to draft until the upstream issue is resolved.

ribru17:
Upstream issue is now resolved (in nightly), but we should probably hold off on using it until 0.11 is the minimum supported version by nvim-treesitter

clason:
Didn't we make the directive backwards compatible?

ribru17:
It is backwards compatible in that the query won't break, but on 0.10 folding the inner list in

```markdown
- list
  - inner list
    - 1
    - 2
  - second item
- list
```

produces

```
- list
  - inner list
- list
```

so the non-nightly behavior is a bit wonky but not really the end of the world I suppose

clason:
Hmm. That's not great. We can make these changes on `main` (and Nvim) at least.

helins:
Is there anything blocking this from being merged?

helins:
Just some clarifications because this issue is important for me + I might have another problem related to `block_continuation`s in a similar way.

Mainly addressed to @ribru17  I suppose.

A) Your fix rely solely on trimming whitespace and is not related to the `block_continuation` discussed earlier, is that correct?

B) You describe [suboptimal behavior](https://github.com/nvim-treesitter/nvim-treesitter/pull/3442#issuecomment-2525367574) on 0.10. Isn't it the current behavior? Actually, if I were to fold on `second item`, I would still get the same "bad result", which is even less intuitive.


Thanks!

ribru17:
> A) Your fix rely solely on trimming whitespace and is not related to the `block_continuation` discussed earlier, is that correct?

Correct :+1: 

> B) You describe [suboptimal behavior](https://github.com/nvim-treesitter/nvim-treesitter/pull/3442#issuecomment-2525367574) on 0.10. Isn't it the current behavior? Actually, if I were to fold on `second item`, I would still get the same "bad result", which is even less intuitive.

Interesting, yeah it looks that way. I had thought that there was no list folding at all, but maybe I removed it in my personal config and forgot. Looks like merging an updated fix (which would work perfectly for 0.11) would also be a slight improvement on 0.10, even if not a complete fix. I can make a new PR if that's alright with @clason

clason:
We're getting close to a freeze on the master branch anyway, with more parsers daily switching to ABI 15.

ribru17:
Does that mean we should go ahead with this, or does it mean hold off for now?

clason:
Depends on the nature of the fix. If it's not a regression for 0.10, I don't mind.

ribru17:
I don't think it is, after what @helins pointed out (and I checked myself just in case). I'll put up a PR real quick

ribru17:
#7702

PR #3447, opened 2022-09-07
feat(viml): bump parser and queries

clason:


clason:
OK, this is a problem. Looks like the (now) bundled VimL parser on nightly takes precedence over the one installed by nvim-treesitter (at least on CI), which leads to problems with the query.

We need to make sure that nvim-treesitter disables or overrides the parsers from core.

@theHamsta

queries/vim/highlights.scm
```diff
@@ -151,6 +152,8 @@
 
 ;; Command command
 
+(command) @string
```
theHamsta, line 155: An alternative could be `@function.call` but I understand the wish to distinguish function calls from command invocations in viml

theHamsta (approved):

theHamsta:
This was always a problem. Probably core could handle this by switching back to the override instead of merging the queries. Or have some kind allow list/block list for queries. Or have some versioning for parsers and queries to find the right queries to the right parser or we just give up maintaining the languages that got merged to core. People would then contribute parser and query updates directly to core.

Would probably prefer allow/block patterns onto runtime paths most, since this allows users also to block certain nvim-treesitter parsers/queries and we could still have some experimentation phase before stuff gets merged into core.

clason:
Why did you merge this? This was a breaking change on master!

PR #3448, opened 2022-09-07
feat: add jsonnet

nawordar:
Neovim does not have a built in Jsonnet file type detection, so it only works with google/vim-jsonnet extension or with ftdetect file added (https://github.com/google/vim-jsonnet/blob/master/ftdetect/jsonnet.vim). Is it fine, or maybe that ftdetect file should be added to Neovim first? I'm not sure if they would even accept a PR that would only add file type detection without other features.


clason:
That is fine, you can always add the filetype detection manually.

If you want to include it out of the box, the PR needs to go to vim/vim since we port the filetype detection from there. (How easy that is depends on whether it conflicts with one of the million supported extensions; then you need to add a check for the contents to decide.)

clason:
#20119

theHamsta (approved):
Thank you!

PR #3456, opened 2022-09-08
feat: add tree-sitter-gitattributes

ObserverOfTime:
Vim doesn't have a `gitattributes` filetype so it will have to be added by the user.

```lua
-- for example:
vim.filetype.add {
    filename = {
        [".gitattributes"] = "gitattributes",
    },
    pattern = {
        [".*/git/attributes"] = "gitattributes",
        [".*/info/attributes"] = "gitattributes",
    }
}
```

---

I hadn't written a parser before this one, so I marked it as experimental just in case.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1205,6 +1205,17 @@ list.gitignore = {
   maintainers = { "@theHamsta" },
 }
 
+list.gitattributes = {
+  install_info = {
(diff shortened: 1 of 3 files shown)


clason:
You can also make a PR to Vim to add that filetype detection.

ObserverOfTime:
I could do that. I already have a basic Vim syntax file for it too.

theHamsta (approved):

PR #3445, opened 2022-09-07
highlights(sql): change table_expression to table_reference

matthias-Q:
The node name `table_expression` has been refactored and renamed into `table_reference` in [here](https://github.com/DerekStride/tree-sitter-sql/commit/b032eaa5122e70cd4b012e2272eb357e1971ddc7).
Therefore, the queries for highlighting do not work anymore. They break with the following message:

```
ERROR 2022-09-07T07:56:07.092 2000835 decor_provider_invoke:37: error in provider treesitter/highlighter:line: Error executing lua: /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:172: query: error at position 257
stack traceback:
        [C]: in function '_ts_parse_query'
        /usr/local/share/nvim/runtime/lua/vim/treesitter/query.lua:172: in function 'get_query'
        ...al/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:122: in function 'new'
        ...al/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:247: in function 'get_query'
        ...al/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:265: in function 'fn'
        ...l/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:194: in function 'for_each_tree'
        ...al/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:255: in function 'on_line_impl'
        ...al/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:303: in function <...al/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:299>
```

I have changed the query accordingly. However it fails to match on the second query:
```
(table_reference
  name: (identifier) @type
  table_alias: (identifier) @variable)
```

I am still quite fresh in the tree-sitter world. Could you take another look?.

PR #3458, opened 2022-09-08
Update SQL parser queries

DerekStride:
@clason This PR should fix the errors with the SQL queries in https://github.com/nvim-treesitter/nvim-treesitter/pull/3446.

I've set it up to merge with [`nvim-treesitter:update-lockfile-pr`](https://github.com/nvim-treesitter/nvim-treesitter/tree/update-lockfile-pr) instead of `master` so you can merge it directly into your PR.

### DIFF
--- a/queries/sql/highlights.scm
+++ b/queries/sql/highlights.scm
@@ -6,11 +6,10 @@
   name: (identifier) @function.call
   parameter: [(field)]? @parameter)
 
-(table_expression
+(table_reference
   name: (identifier) @type)
 
-(table_expression
-  name: (identifier) @type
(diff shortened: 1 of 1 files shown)


clason:
Thanks!

But it'd be easier for me the other way around: include the parser lockfile update in the query PR.

(This would keep breaking updates isolated, making pinpointing issues easier.)

matthias-Q:
This does exactly the same as https://github.com/nvim-treesitter/nvim-treesitter/pull/3445

clason:
That had an unrelated viml bump ;)

PR #3477, opened 2022-09-10
build: add a rockspec

teto:
so that users can install the plugin via luarocks but even better: 
so that other plugins can reference this one and add it as a dependency.

### DIFF
--- a/contrib/nvim-treesitter-scm-1.rockspec
+++ b/contrib/nvim-treesitter-scm-1.rockspec
@@ -1,0 +1,32 @@
+local MODREV, SPECREV = "scm", "-1"
+rockspec_format = "3.0"
+package = "nvim-treesitter"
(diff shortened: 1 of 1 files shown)


theHamsta:
Aren't rockspecs for proper stand-alone packages? Which plugin manager uses luarocks for dependency management. I think packer supports rockspecs but you can also use `requires` to specify requirements for plugins. Can you show how one would one use it?

theHamsta (approved):

teto:
if you configure luarocks to do a flat install, ie. prevent luarocks to create the nested folders. In which case you can handle your plugins with luarocks natively see https://teto.github.io/posts/2022-06-22-neovim-plugin-luarocks-2.html (goes into more details, and how I leverage that with nix).

PR #3483, opened 2022-09-11
parsers: add blueprint parser

GabMus:
You can find the parser [in this repo](https://gitlab.com/gabmus/tree-sitter-blueprint).

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1224,6 +1224,15 @@ list.gitattributes = {
   experimental = true,
 }
 
+list.blueprint = {
+  install_info = {
+    url = "https://gitlab.com/gabmus/tree-sitter-blueprint.git",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@gabmus" },
+  experimental = true,
+}
+
 local M = {
   list = list,
   filetype_to_parsername = filetype_to_parsername,
--- a/queries/blueprint/highlights.scm
+++ b/queries/blueprint/highlights.scm
@@ -1,0 +1,57 @@
+(object_id) @variable
+
+(string) @string
+(escape_sequence) @string.escape
+
+(comment) @comment
+
+(constant) @constant.builtin
+
+(boolean) @boolean
+
+(using) @include
+
+(template) @keyword
+
+(decorator) @attribute
+
+(property_definition (property_name) @property)
+
+(object) @type
+
+(signal_binding (signal_name) @function.builtin)
+(signal_binding (function (identifier)) @function)
+(signal_binding "swapped" @keyword)
+
+(styles_list "styles" @function.macro)
+(layout_definition "layout" @function.macro)
+
+(gettext_string "_" @function.builtin)
+
+(menu_definition "menu" @keyword)
+(menu_section "section" @keyword)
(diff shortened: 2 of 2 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -1224,6 +1225,17 @@ list.gitattributes = {
   experimental = true,
 }
 
+list.blueprint = {
+  install_info = {
+    url = "https://gitlab.com/gabmus/tree-sitter-blueprint",
+    files = { "src/parser.c" },
+    branch = "master",
+    requires_generate_from_grammar = false,
```
clason, line 1233: these two lines are not needed

scripts/minimal_init.lua
```diff
@@ -4,6 +4,7 @@ vim.cmd [[runtime! plugin/nvim-treesitter.lua]]
 
 vim.cmd [[au BufRead,BufNewFile *.conf set filetype=hocon]]
 vim.cmd [[au BufRead,BufNewFile *.gleam set filetype=gleam]]
+vim.cmd [[au BufRead,BufNewFile *.blp set filetype=blueprint]]
```
clason, line 7: This is not needed, the script is obsolete.

clason (changes requested):

GabMus:
@clason I applied those changes

lua/nvim-treesitter/parsers.lua
```diff
@@ -18,6 +18,7 @@ local filetype_to_parsername = {
   pandoc = "markdown",
   rmd = "markdown",
   cs = "c_sharp",
+  blp = "blueprint",
```
clason, line 21: That's not how this is meant to be used: It maps _filetypes_ to parser names, not _extensions_.

You should add the mapping of `*.blp` to filetype `blueprint` in your config, then the `blueprint` parser will just work out of the box.


(The existing entries are misleading and partially wrong.)

GabMus: Fixed, thanks for the review

queries/blueprint/highlights.scm
```diff
@@ -0,0 +1,57 @@
...
+
+(boolean) @boolean
+
+(using) @include
+
+(template) @keyword
+
+(decorator) @attribute
+
+(property_definition (property_name) @parameter)
```
theHamsta, line 18: I don't know the exact context here. But might `@property` be more appropriate here?

GabMus: Sure, that's more appropriate. Just changed it.

theHamsta (approved):

lua/nvim-treesitter/parsers.lua
```diff
@@ -1224,6 +1224,15 @@ list.gitattributes = {
   experimental = true,
 }
 
+list.blueprint = {
+  install_info = {
+    url = "https://gitlab.com/gabmus/tree-sitter-blueprint",
```
theHamsta, line 1229: for other Gitlab parsers we had sometimes problems without the `.git` suffix during version upgrade. Let's see whether that's still the case

GabMus: I can add the suffix if you think it helps

clason (approved):

PR #3494, opened 2022-09-12
highlights(elvish): Use `@function.builtin` for operator-like functions

ckafi:
Elvish uses prefix notation even for mathematical operators (i.e. `<= 1 2 3 4`),  so these should be considered functions.


theHamsta (approved):
Looks good to me thank you

PR #3500, opened 2022-09-13
highlights(sql): include new keywords

matthias-Q:
* added keywords for pattern matching
* remove duplicate keywords
* added hashing functions as `@function.call`


matthias-Q:
I don't understand why the Actions fail... I updated the lockfile to the latest upstream commit sha.

theHamsta:
grafik

matthias-Q:
Ah ok, my mistake. It should be fixed in the last commit.

clason (approved):

PR #3504, opened 2022-09-14
fix(highlight): respect g:skip_ts_default_groups on ColorScheme autocmd

nullchilly:
From my understanding you need to enable `skip_ts_default_groups` in order to set [custom colorscheme](https://github.com/catppuccin/nvim/blob/main/lua/catppuccin/groups/integrations/treesitter.lua)

Without it after `:colorscheme catppuccin` the highlight captures will link to the deprecated TS* highlights

I can simply merge the old TS* highlight groups but that's rather ugly

image

image

### DIFF
--- a/lua/nvim-treesitter/highlight.lua
+++ b/lua/nvim-treesitter/highlight.lua
@@ -183,7 +183,7 @@ function M.set_custom_captures(captures)
 end
 
 function M.set_default_hlgroups()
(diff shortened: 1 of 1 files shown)


clason (approved):

clason:
All this will go away soon anyway once 0.8 is released.

PR #3506, opened 2022-09-15
parsers: add nickel parser

jlesquembre:
Add nickel parser: https://github.com/nickel-lang/tree-sitter-nickel

Nickel lang: https://nickel-lang.org/

The highlights.scm file is from: https://github.com/nickel-lang/tree-sitter-nickel/blob/main/queries/highlights.scm (I asked the author and he is ok with upstreaming it)

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1211,6 +1211,16 @@ list.gitignore = {
   maintainers = { "@theHamsta" },
 }
 
+list.nickel = {
+  install_info = {
+    url = "https://github.com/nickel-lang/tree-sitter-nickel",
+    files = { "src/parser.c", "src/scanner.cc" },
+    branch = "main",
+    generate_requires_npm = false,
+    requires_generate_from_grammar = false,
+  },
+}
+
 list.gitattributes = {
   install_info = {
     url = "https://github.com/ObserverOfTime/tree-sitter-gitattributes",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1216,8 +1216,6 @@ list.nickel = {
     url = "https://github.com/nickel-lang/tree-sitter-nickel",
     files = { "src/parser.c", "src/scanner.cc" },
     branch = "main",
-    generate_requires_npm = false,
-    requires_generate_from_grammar = false,
   },
 }
 
--- a/queries/nickel/highlights.scm
+++ b/queries/nickel/highlights.scm
@@ -1,4 +1,4 @@
-(comment) @comment
+(comment) @comment @spell
 
 [
   "if"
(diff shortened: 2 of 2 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -1213,6 +1213,16 @@ list.gitignore = {
   maintainers = { "@theHamsta" },
 }
 
+list.nickel = {
+  install_info = {
+    url = "https://github.com/nickel-lang/tree-sitter-nickel",
+    files = { "src/parser.c", "src/scanner.cc" },
+    branch = "main",
+    generate_requires_npm = false,
+    requires_generate_from_grammar = false,
```
clason, line 1222:
```suggestion
```
No need to specify the defaults.

queries/nickel/highlights.scm
```diff
@@ -0,0 +1,64 @@
+(comment) @comment
```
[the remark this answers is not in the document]
jlesquembre: I don't think so, but I'm not sure, I just took it from the grammar repo: https://github.com/nickel-lang/tree-sitter-nickel

jlesquembre:
@clason thanks for the review, I think this should fix the error: https://github.com/nvim-treesitter/nvim-treesitter/pull/3506/commits/4304885945742f55306f6db532262ecf7f5949aa

queries/nickel/highlights.scm
```diff
@@ -0,0 +1,64 @@
+(comment) @comment @spell
+
+[
+  "if"
+  "then"
+  "else"
+  "forall"
```
theHamsta, line 7: `@repeat`?

queries/nickel/highlights.scm
```diff
@@ -0,0 +1,64 @@
+(comment) @comment @spell
+
+[
+  "if"
+  "then"
+  "else"
```
theHamsta, line 6: `@conditional` for if/else/then

theHamsta: you already have the conditional highlight. I don't think keyword is necessary then

queries/nickel/highlights.scm
```diff
@@ -0,0 +1,64 @@
...
+(interpolation_end) @punctuation.bracket
+
+(record_field) @field
+
+(builtin) @function.builtin
+
+(fun_expr pats:
+  (pattern id:
+    (ident) @parameter
+  )+
```
theHamsta, line 59: the `+` will have no effect here, but it also does not harm

queries/nickel/highlights.scm
jlesquembre: Not sure, `forall` is used to define polymorphic types: https://nickel-lang.org/user-manual/typing/#type-polymorphism
It's not really a loop

queries/nickel/highlights.scm
jlesquembre: Probably not, removed

theHamsta:
Thank you!

Issue #3323, opened 2022-08-19
Statusline indicator, parsing argument error 


ggerico:
### Describe the bug

`:echo nvim_treesitter#statusline(90)`

E5108: Error executing lua ...start/nvim-treesitter/lua/nvim-treesitter/statusline.lua:19: attempt to index local '
options' (a number value)

**here is the fix**   → `echo nvim_treesitter#statusline("90")`

### To Reproduce

`:echo nvim_treesitter#statusline(90)`

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.6 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v14.15.1 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: Apple clang version 13.1.6 (clang-1316.0.21.2.5)
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - lua            ✓ ✓ ✓ ✓ ✓
  - rust           ✓ ✓ ✓ ✓ ✓

  Legend: H[ighlight], L[ocals], F[olds], I[ndents], In[j]ections
         +) multiple parsers found, only one will be used
         x) errors found in the query, try to run :TSUpdate {lang}
```

### Output of `nvim --version`

```text
NVIM v0.8.0-dev-3720-g0fdf59ac9
Build type: Release/youbike/map/
LuaJIT 2.1.0-beta3
Compiled by germain@MacBookPro-2.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/Cellar/neovim/HEAD-0fdf59a_1/share/nvim"
```

MaMrEzO:
This is because of commented lines, which handle `opts` and check it to be a number: 

https://github.com/nvim-treesitter/nvim-treesitter/blob/088dfbc54abfefeab8d3f1efc8be1ea48888b320/lua/nvim-treesitter/statusline.lua#L16-L18

which seems the `README` relies on it; however, you can call it like this:
```vim 
[...]

PR #3512, opened 2022-09-16
fix: #3323, handle `opts` as number in statusline

MaMrEzO:


lua/nvim-treesitter/statusline.lua
```diff
@@ -13,9 +13,9 @@ function M.statusline(opts)
     return
   end
   local options = opts or {}
-  -- if type(opts) == "number" then
-  --   options = { indicator_size = opts }
-  -- end
```
Antikyth, line 18: Why was this already there, but commented out? Are there any unwanted consequences that this may re-add that were the reason for it being commented out originally?

MaMrEzO: As documented [here](https://github.com/nvim-treesitter/nvim-treesitter#statusline-indicator), the `statusline` can accept number as opts, and the commit lead to comment it: [fix(statusline): adjust to the new API (query)](https://github.com/nvim-treesitter/nvim-treesitter/commit/2490a8bbcaca5c84b86ea82661854e1cfb865571#diff-73d8fd4bc707f46576768511e62a4a2ddc0bb32f1e676b4b208e9d61e193d8acL51)

theHamsta (approved):

lua/nvim-treesitter/statusline.lua
theHamsta: I'm not aware of any risk of enabling this again. Should we just try it?

PR #3531, opened 2022-09-22
Expand locals to include properties and methods

neanias:
The previous locals were lacking a few locals related to JS classes. This expands the locals to include properties defined on classes (e.g. `this.foo = "bar"` or `static #targets`) as both vars & references, as well as private methods on classes (e.g. `#bar(x) { x }` and `#bar(y)`).

### DIFF
--- a/queries/javascript/locals.scm
+++ b/queries/javascript/locals.scm
@@ -1,5 +1,20 @@
 ; inherits: ecma,jsx
 
+; Both properties are matched here.
+;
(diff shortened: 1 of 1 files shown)


neanias:
Is there a way to re-run the actions? I can't see a re-run button anywhere

neanias:
I'm not clear why these tests are failing as they all pass on my machine

theHamsta (approved):

PR #3537, opened 2022-09-23
Add query to highlight comment marker `NOTE`

### LINKED ISSUE #3536: Add comment highlight for the `NOTE` marker
(redirect from original issue https://github.com/stsewd/tree-sitter-comment/issues/13)

It is custom in source code files to add special markers in comments, most known probably `TODO`. Then it's easy to to `grep(1)` for this later.

The current known comment markers are defined here:
https://github.com/nvim-treesitter/nvim-treesitter/blob/9b13872f19aa97916ba0ae92bac6fb9e2c1dd06d/queries/comment/highlights.scm#L10-L20

What I quickly noticed when I started to use treesitter for comments, is that the `NOTE` marker that I frequently use was not included. I [found a way](https://github.com/stsewd/tree-sitter-comment/issues/13#issuecomment-1243443143) to add this my self however I wonder if this could possibly be included by default.

Well, I claim `NOTE` is common but this could of course likely be biased by my own usage...
Thus I tried to get some numbers from GitHub's new Code Search:
https://cs.github.com/?scopeName=All+repos&scope=&q=%2F%28%3F-i%29+NOTE+%2F

`/(?-i) NOTE /` matched to 725k files, compared to
* `FIXME`: 500k files
* `HACK`:  122k files
* `TODO` : 5.3M files
* `WARNING`: 350k files
* `BUG`: 137k files

So `NOTE` seems to be at least as common as some already recognized comment markers. These numbers should not be exact, since this simple regex don't even try to match the context of a comment. However it seems to say that it's in the same ballpark at least.

Could we add `NOTE` to the default recognized comment markers?

erikw:
This is a common comment marker, and nvim-tree sitter would provide more
value to developers by highlighting this by default! Please see #3536
for further motivation and discussion.

Fixes #3536

### DIFF
--- a/queries/comment/highlights.scm
+++ b/queries/comment/highlights.scm
@@ -7,11 +7,17 @@
 
 (tag (name) @text.note (user)? @constant)
 
+((tag ((name) @text.note))
+ (#any-of? @text.note "NOTE"))
+
+("text" @text.note
+ (#any-of? @text.note "NOTE"))
+
 ((tag ((name) @text.warning))
- (#any-of? @text.warning "TODO" "HACK" "WARNING" "NOTE"))
+ (#any-of? @text.warning "TODO" "HACK" "WARNING"))
 
 ("text" @text.warning
- (#any-of? @text.warning "TODO" "HACK" "WARNING" "NOTE"))
+ (#any-of? @text.warning "TODO" "HACK" "WARNING"))
 
 ((tag ((name) @text.danger))
  (#any-of? @text.danger "FIXME" "XXX" "BUG"))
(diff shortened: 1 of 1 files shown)


clason:
We have a dedicated `@text.note`; it would probably be better to use that.

clason:
By which I mean, please use that ;)

erikw:
> We have a dedicated `@text.note`; it would probably be better to use that.

✅ Done

The font is a bit thinner with this, but that's fine for me

Screen Shot 2022-09-23 at 19 48 08

clason:
You can link the `@text.note` to whatever highlight group you prefer (by default it's `TSNote`, but you can just `hi link TSNote TSWarning` if you prefer that).

clason (approved):

PR #3555, opened 2022-09-28
update vimdoc parser and queries

clason:


vigoux (approved):

clason:
Hmm, query check is failing -- @vigoux can you see what's wrong here?

vigoux:
Seems not related to this change ? `typescript` something something...
Rerunning to be sure.

melkster:
Not sure if this is the right place but the `help` change seems to have broken the syntax of helpdoc labels (are they called that?), making them always left-aligned when concealed.

hmm

clason:
Yes, that is known. The parser isn't really meant primarily for syntax highlighting yet; the focus is more on doc generation. It's clearly something to be fixed, though.

melkster:
Ok!

To anyone reading this, you can disable the syntax highlighting of help files by doing this:

```lua
  highlight = {
    enable = true,
    disable = { 'help' },
  },
})
```

PR #3562, opened 2022-09-30
highlights: many new and updated captures

ObserverOfTime:


queries/query/highlights.scm
```diff
@@ -29,3 +29,6 @@
 
 ((program . (comment) @include)
  (#match? @include "^;\ +inherits\ *:"))
+
+((program . (comment) @preproc)
```
theHamsta, line 33: We don't currently use `@preproc` even though documentation states otherwise. If this capture is needed i would prefer to introduce it in a separate PR and for all languages that would need it (or track progress of porting to this capture in a PR)

ObserverOfTime: There are two other usages (one of which is mine):
https://github.com/nvim-treesitter/nvim-treesitter/blob/aebc6cf6bd4675ac86629f516d612ad5288f7868/queries/gitattributes/highlights.scm#L40
https://github.com/nvim-treesitter/nvim-treesitter/blob/aebc6cf6bd4675ac86629f516d612ad5288f7868/queries/vala/highlights.scm#L82-L87

ObserverOfTime: I dropped the commit and resubmitted it with other languages in #3567.

theHamsta (approved):

clason:
VimL highlights are owned by upstream; changes should be contributed to that repo so we don't diverge from core.

PR #3567, opened 2022-10-01
highlights: use @preproc where appropriate

ObserverOfTime:
Use the `@preproc` capture to highlight certain preprocessor directives and shebangs.
It could also be used to highlight (some or all) at-rules in CSS if that's desirable. 
There are definitely more languages which could use it that I'm not familiar with.

### DIFF
--- a/queries/agda/highlights.scm
+++ b/queries/agda/highlights.scm
@@ -33,7 +33,7 @@
 
 ;; Pragmas and comments
 
-(pragma) @constant.macro
+(pragma) @preproc
 
 (comment) @comment
 
--- a/queries/bash/highlights.scm
+++ b/queries/bash/highlights.scm
@@ -129,3 +129,6 @@
   value: (word) @parameter)
 
 (regex) @string.regex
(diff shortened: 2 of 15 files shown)


lewis6991 (approved):
LGTM, will let someone else merge.

theHamsta:
I think this is ready to merge. @ObserverOfTime I still think it would be useful to document the rationale that you used to decide which highlights should be `@preproc`. Maybe this can be a follow-up?

ObserverOfTime:
> I still think it would be useful to document the rationale that you used to decide which highlights should be `@preproc`. 

Would this suffice?

```diff
--- i/CONTRIBUTING.md
+++ w/CONTRIBUTING.md
@@ -96,3 +96,3 @@
 @none to disable completely the highlight
-@preproc
+@preproc generic preprocessor directives & shebangs
 @punctuation.delimiter for `;` `.` `,`
```

theHamsta:
That would be a good start!

Issue #2860, opened 2022-04-24
Scrolling becomes very slow for long list of "define" items (c/c++)


Frefreak:
### Describe the bug

I notice when highlight is enabled, when scrolling in files with a large list of define (one each line) the UI becomes very slow. Disable `highlight` makes it smooth again. Is this expected or if there's any workaround for this?

### To Reproduce

1. open `/usr/include/bfd.h` (`binutils` package might include this file)
2. scrolls with ctrl-d or ctrl-f from the beginning. Everything should be fine until about 36% of the file which is inside a long list of "define" items beginning with "#define BFD_RELOC_SPARC_DISP64 BFD_RELOC_64_PCREL"

### Expected behavior

everything is smooth and fast

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.6 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v17.9.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 11.2.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - rasi           ✓ ✓ ✓ ✓ . 
  - cuda           ✓ ✓ ✓ ✓ ✓ 
  - comment        ✓ . . . . 
  - jsdoc          ✓ . . . . 
  - jsonc          ✓ ✓ ✓ ✓ ✓ 
  - regex          ✓ . . . . 
  - dockerfile     ✓ . . . ✓ 
  - yaml           ✓ ✓ ✓ ✓ ✓ 
  - gdscript       ✓ ✓ . ✓ ✓ 
  - yang           ✓ . ✓ ✓ . 
  - fusion         ✓ ✓ ✓ ✓ . 
  - ninja          ✓ . ✓ ✓ . 
  - ledger         ✓ . ✓ ✓ ✓ 
  - nix            ✓ ✓ ✓ . ✓ 
  - turtle         ✓ ✓ ✓ ✓ ✓ 
  - dart           ✓ ✓ . ✓ ✓ 
[...]

Issue #1292, opened 2021-05-12
Horrible input lag on 2MB C header file.


markus-oberhumer:
**Describe the bug**

Horrible input lag on 2MB C header file, up to several seconds.

I'm using the latest official nvim nightly build and have removed all plugins except nvim-treesitter.

```
  require'nvim-treesitter.configs'.setup {
    ensure_installed = {'c', 'cpp',},
    highlight = {
      enable = true,
    },
    indent = {
      enable = false,
    },
    incremental_selection = {
      enable = false,
      keymaps = {
        node_incremental = 'grn',
        init_selection = 'gnn',
        scope_incremental = 'grc',
        node_decremental = 'grm',
      },
    },
  }
```

**To Reproduce**

Try to open this linux kernel source code file in nvim with C + CPP treesitter enabled:

```nvim linux.git/drivers/gpu/drm/amd/include/asic_reg/bif/bif_5_1_sh_mask.h```

and then start moving the cursor around.

https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/amd/include/asic_reg/bif/bif_5_1_sh_mask.h

**Output of `:checkhealth nvim_treesitter`**

<details>
<code>

health#nvim_treesitter#check
========================================================================
## Installation
  - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required
  for :TSInstall)
  - OK: `node` found v14.16.1 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl" }
  - OK: Neovim was compiled with tree-sitter runtime ABI version 13 (required >=13). Parsers must be compatible with
  runtime ABI.

## Parser/Features H L F I
  - c              ✓ ✓ ✓ ✓
  - cpp            ✓ ✓ ✓ ✓

 Legend: H[ighlight], L[ocals], F[olds], I[ndents]
         +) multiple parsers found, only one will be used
         x) errors found in the query, try to run :TSUpdate {lang}

</code>
</details>

**Output of `nvim --version`**

<details>
<code>

NVIM v0.5.0-dev+1318-g61aefaf29
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
[...]

PR #3570, opened 2022-10-01
docs: add example of how to use function to disable a module

jedrzejboczar:
I found out that `disable` can be a function be looking through the source code, but it may be helpful to include this in the readme. 

In the readme I added an example of what I actually used this for - disabling highlight on large files. This is a nice workaround for https://github.com/nvim-treesitter/nvim-treesitter/issues/2996,  https://github.com/nvim-treesitter/nvim-treesitter/issues/2860, https://github.com/nvim-treesitter/nvim-treesitter/issues/1292, etc. This solution is also quite performant as from my tests `fs_stat` takes <100 us.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,14 @@ require'nvim-treesitter.configs'.setup {
     -- the name of the parser)
     -- list of language that will be disabled
     disable = { "c", "rust" },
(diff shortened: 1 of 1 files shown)


theHamsta (approved):

theHamsta:
Thank you!

PR #3574, opened 2022-10-02
Add ebnf parser

RubixDev:
~~The parser is part of [this repo](https://github.com/RubixDev/ebnf) under [crates/tree-sitter-ebnf](https://github.com/RubixDev/ebnf/tree/main/crates/tree-sitter-ebnf)~~

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -378,7 +378,7 @@
     "revision": "8bd2056818b21860e3d756b5a58c4f6e05fb744e"
   },
   "tsx": {
-    "revision": "0ab9d99867435a7667c5548a6617a6bf73dbd830"
+    "revision": "0ae382803abce0807e90f498105c713b9233e0b2"
   },
   "turtle": {
     "revision": "085437f5cb117703b7f520dd92161140a684f092"
@@ -387,7 +387,7 @@
     "revision": "035f549ec8c043e734f04341d7ccdc669bb2ba91"
   },
   "typescript": {
-    "revision": "0ab9d99867435a7667c5548a6617a6bf73dbd830"
+    "revision": "0ae382803abce0807e90f498105c713b9233e0b2"
   },
   "v": {
     "revision": "66b92a89ef1e149300df79c0b2a934ad959c8eec"
--- a/queries/typescript/highlights.scm
+++ b/queries/typescript/highlights.scm
@@ -14,6 +14,7 @@
 "type"
 "readonly"
 "override"
+"satisfies"
 ] @keyword
 
 ; types


RubixDev:
I moved the parser to [a separate repo](https://github.com/RubixDev/tree-sitter-ebnf), as it seems installation from monorepos is not supported.

theHamsta:
> I moved the parser to [a separate repo](https://github.com/RubixDev/tree-sitter-ebnf), as it seems installation from monorepos is not supported

It's supported. See vlang or tsx/typescript

queries/ebnf/highlights.scm
```diff
@@ -0,0 +1,47 @@
...
+ (#match? @grammar.nonterminal.upper "^[A-Z][A-Z0-9_]+$"))
+
+((identifier) @grammar.nonterminal.lower
+ (#match? @grammar.nonterminal.lower "^[a-z][a-z0-9_]+$"))
+
+;;; Punctuation ;;;;
+[
+ ";"
+ ","
+] @punctuation.delimeter
```
theHamsta, line 30: this is a typo. Should be `@punctuation.delimiter`. CI fails because of that

queries/ebnf/highlights.scm
```diff
@@ -0,0 +1,47 @@
+;;;; Simple tokens ;;;;
+(terminal) @string @grammar.terminal
```
theHamsta, line 2: CI fails because `@grammar` is a undocumented new highlight type, There are a few possibilities to fix this:

- Is `@grammar` needed for other languages? Then `@grammar` or better `@grammar.punctuation` should be documented in `CONTRIBUTING.md` and added to all languages where necessary
- Is `@grammar` similar to a existing highlight group in `CONTRIBUTING.md`? Then, it should be changed to `@XXX.grammar.nonterminal` where `XXXX` is a documented capture. This way users will have a good default experience when their color scheme does not support `@grammar`

queries/ebnf/highlights.scm
RubixDev: Fixed it, thanks

RubixDev:
> > I moved the parser to [a separate repo](https://github.com/RubixDev/tree-sitter-ebnf), as it seems installation from monorepos is not supported
> 
> It's supported. See vlang or tsx/typescript

Alright, thanks for the info, will move it back into the monorepo then

queries/ebnf/highlights.scm
theHamsta: Yes, that would be ok. But I think your old `@grammar` categories were also ok, since they can be surely extended to other languages

theHamsta:
Your EBNF files read nicely. Have you tried generating `grammar.json` nodes with it? It might be a good alternative to JS, an alternative frontend for tree-sitter CLI

lua/nvim-treesitter/parsers.lua
```diff
@@ -1243,6 +1243,17 @@ list.twig = {
   filetype = "twig",
 }
 
+list.ebnf = {
+  install_info = {
+    url = "https://github.com/ebnf.git",
```
theHamsta, line 1248: you meant to put `https://github.com/RubixDev/ebnf` here

queries/ebnf/highlights.scm
```diff
@@ -0,0 +1,47 @@
...
+(terminal) @string.grammar
+
+(special_sequence) @string.special.grammar
+
+(integer) @number
+
+(comment) @comment.block
+
+;;;; Identifiers ;;;;
+(identifier) @variable.grammar
```
theHamsta, line 11: `@symbol` (or `@symbol.grammar` would also be an option)

lua/nvim-treesitter/parsers.lua
RubixDev: Thanks, I thought I just copied it from my personal config...

queries/ebnf/highlights.scm
RubixDev: I agree, changed it to `@symbol.grammar` now

theHamsta:
Ooops, closed by mistake. New PR: https://github.com/nvim-treesitter/nvim-treesitter/pull/3860

queries/ebnf/highlights.scm
ObserverOfTime: The `@grammar` captures could also be used for ANTLR/Yacc/Bison/JFlex/CUP/etc. but none of them have tree-sitter grammars (that I know of) yet and I think existing captures can adequately describe all their tokens anyway.

Highlights can currently be customised with `@string.ebnf` and so on without needing a custom suffix.

PR #3592, opened 2022-10-06
Add regex injections for php

gbprod:
This commit introduce regex language injection for php preg_* functions.


queries/php/injections.scm
```diff
@@ -1,3 +1,19 @@
...
+  (#any-of? @_preg_func_identifier "preg_match" "preg_replace" "preg_filter"
+   "preg_match_all" "preg_replace_callback_array" "preg_replace_callback"
+   "preg_split"))
+
+((function_call_expression
+  function: (_) @_preg_func_identifier
+  arguments: (arguments (argument (encapsed_string (string_value) @regex)))+)
+  (#any-of? @_preg_func_identifier "preg_match" "preg_replace" "preg_filter"
+   "preg_match_all" "preg_replace_callback_array" "preg_replace_callback"
+   "preg_split"))
```
ObserverOfTime, line 19: These can be combined:
```scm
((function_call_expression
  function: (_) @_preg_func_identifier
  arguments: (arguments (argument (_ (string_value) @regex)))+)
  (#any-of? @_preg_func_identifier "preg_match" "preg_replace" "preg_filter"
   "preg_match_all" "preg_replace_callback_array" "preg_replace_callback"
   "preg_split"))
```

ObserverOfTime (changes requested):

queries/php/injections.scm
gbprod: :+1:

ObserverOfTime:
> I don't know if this repo is the place for those kind of injections, maybe we can have a wiki for that ?

There are a few other languages with somewhat complex regex injections (e.g. Kotlin).

Documenting the Lua pattern injections was suggested in https://github.com/nvim-treesitter/nvim-treesitter/pull/3578#issuecomment-1265678590.
Maybe we could do the same for regex in similar cases?

gbprod:
> > I don't know if this repo is the place for those kind of injections, maybe we can have a wiki for that ?
> 
> There are a few other languages with somewhat complex regex injections (e.g. Kotlin).
> 
> Documenting the Lua pattern injections was suggested in [#3578 (comment)](https://github.com/nvim-treesitter/nvim-treesitter/pull/3578#issuecomment-1265678590). Maybe we could do the same for regex in similar cases?

Maybe we can put this pull request aside until this discussion is over ;)

queries/php/injections.scm
```diff
@@ -1,3 +1,10 @@
 (text) @html
 
 (comment) @phpdoc
+
+;; regex
+
+((function_call_expression
+  function: (_) @_preg_func_identifier
+  arguments: (arguments . (argument (_ (string_value) @regex))))
+    (#match? @_preg_func_identifier "^preg_"))
```
theHamsta, line 10: `#lua-match?` is generally preferred because it is faster. `#match?` uses the slower vim regex which supports more features but is a bit slower

theHamsta (approved):

queries/php/injections.scm
gbprod: Updated ! Thanks

gbprod:
Do you think we can merge this ?

PR #3593, opened 2022-10-06
highlights(r): add spell

maximilian-muecke:


queries/r/highlights.scm
```diff
@@ -10,7 +10,7 @@
 (string) @string
 (string (escape_sequence) @string.escape)
 
-(comment) @comment
+(comment) @comment @spell
```
theHamsta, line 13: Maybe also `@string`? Or do strings contain mostly non-text data

maximilian-muecke: They contain mainly text data, but I wouldn't recommend it, since strings are used for subsetting and thus camel case as well as other common variable names will have wrong spelling. 

image

theHamsta (approved):

PR #3595, opened 2022-10-06
feat(php): add queries for bash injections

gbprod:
This commit allows to inject bash syntax into relevant function arguments and shell expression.

### DIFF
--- a/queries/php/injections.scm
+++ b/queries/php/injections.scm
@@ -1,3 +1,14 @@
 (text) @html
 
 (comment) @phpdoc
+
+;; bash
+
+((function_call_expression
+  function: (_) @_shell_func_identifier
+  arguments: (arguments . (argument (_ (string_value) @bash))))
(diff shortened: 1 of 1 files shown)


queries/php/injections.scm
```diff
@@ -1,3 +1,14 @@
...
+
+;; bash
+
+((function_call_expression
+  function: (_) @_shell_func_identifier
+  arguments: (arguments . (argument (_ (string_value) @bash))))
+  (#any-of? @_shell_func_identifier "shell_exec" "escapeshellarg" 
+   "escapeshellcmd" "exec" "passthru" "proc_open" "shell_exec" "system"))
+
+((expression_statement (shell_command_expression (string_value) @bash)))
```
theHamsta, line 13: I guess `string_value` is without the quotes, right?

theHamsta (approved):

queries/php/injections.scm
gbprod: Yes!

PR #3600, opened 2022-10-07
fix(php): add indent for enums

gbprod:
Indent for enum body is not correct:

```php
enum Test 
{
|
```
instead of:

```php
enum Test 
{
    |
```


theHamsta (approved):

theHamsta:
Thank you!

PR #3601, opened 2022-10-07
fix: don't overwrite highlights when defining

vigoux:
This is just being a good citizen for people that already define highlights :wink:


clason:
Actually, we should remove all that since Nvim 0.8 is released with the new capture groups...

vigoux:
Well maybe we should keep that but hide it behind a `has` call?

clason:
I don't think so. We only support the latest release version (0.8) and the latest commit on master.

clason (approved):

clason:
Shouldn't block this simple fix, though.

PR #3609, opened 2022-10-10
Refactor Julia highlights

savq:
See https://github.com/tree-sitter/tree-sitter-julia/pull/54

Remove a bunch of patterns with `argument_list`. No longer necessary
with `short_function_definition`.

Other minor changes including:

- Add boolean literals See
  https://github.com/tree-sitter/tree-sitter-julia/pull/44
- Add highlights for `for_binding` and `subtype_clause` operators

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -1,87 +1,25 @@
+;;; Identifiers
+
 (identifier) @variable
 
-(operator) @operator
-(range_expression ":" @operator)
-(pair_expression "=>" @operator)
-
-;; In case you want type highlighting based on Julia naming conventions (this might collide with mathematical notation)
-;((identifier) @type ; exception: mark `A_foo` sort of identifiers as variables
-  ;(match? @type "^[A-Z][^_]"))
+; ;; If you want type highlighting based on Julia naming conventions (this might collide with mathematical notation)
+; ((identifier) @type
+;   (match? @type "^[A-Z][^_]"))  ; exception: Highlight `A_foo` sort of identifiers as variables
 
 (macro_identifier) @function.macro
-(macro_identifier (identifier) @function.macro) ; for any one using the variable highlight
+(macro_identifier
+  (identifier) @function.macro) ; for any one using the variable highlight
+
 (macro_definition
-  name: (identifier) @function.macro
-  ["macro" "end" @keyword])
+  name: (identifier) @function.macro)
+
+(quote_expression ":" (identifier)) @symbol
+
+
+;;; Fields and indexes
 
 (field_expression
-  (identifier)
   (identifier) @field .)
-
-(function_definition
-  name: (identifier) @function)
-(call_expression
-  (identifier) @function.call)
-(call_expression
-  (field_expression (identifier) @method.call .))
-(broadcast_call_expression
-  (identifier) @function.call)
-(broadcast_call_expression
-  (field_expression (identifier) @method.call .))
-(parameter_list
-  (identifier) @parameter)
-(parameter_list
-  (optional_parameter .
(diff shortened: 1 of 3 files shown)


theHamsta (approved):
Nice! Thank you!

 To test the queries on CI you would need to update lockfile.json in this PR. We can then wait for the upstream PR.

clason:
Looking forward to the upstream PR getting merged ;)

queries/julia/highlights.scm
```diff
@@ -1,35 +1,64 @@
...
+(identifier) @variable
 
-;; In case you want type highlighting based on Julia naming conventions (this might collide with mathematical notation)
-;((identifier) @type ; exception: mark `A_foo` sort of identifiers as variables
-  ;(match? @type "^[A-Z][^_]"))
+; ;; If you want type highlighting based on Julia naming conventions (this might collide with mathematical notation)
+; ((identifier) @type
+;   (match? @type "^[A-Z][^_]"))  ; exception: Highlight `A_foo` sort of identifiers as variables
 ((identifier) @constant
   (#match? @constant "^[A-Z][A-Z_]{2}[A-Z_]*$"))
```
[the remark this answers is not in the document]
clason: If there's a problem (that is not created by this PR), why not raise an issue -- or, better yet, a PR -- about it, with a clear description of what is "inconsistent"?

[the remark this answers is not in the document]
fredrikekre: Yea you are right, was just trying out this and https://github.com/tree-sitter/tree-sitter-julia/pull/54 and happen to notice. PR here: #3624.

savq:
tree-sitter-julia was updated and I updated the lockfile here.

I don't know what might be causing the query tests to fail. Everything seems to work OK for me. Any ideas?

clason:
Did you rebase on master?

clason:
I think you need to update the other queries as well, especially `locals.scm`.

queries/julia/highlights.scm
```diff
@@ -1,106 +1,113 @@
...
 
 (field_expression
-  (identifier)
   (identifier) @field .)
 
+(subscript_expression
+  (_)
+  (range_expression
+    (identifier) @constant.builtin .)
+  (#eq? @constant.builtin "end"))
```
theHamsta, line 28: `end` is not really a constant. It has no value and is rather a keyword. If the range expression is similar to a loop it could be `@repeat`

clason (approved):

queries/julia/highlights.scm
savq: When indexing an array with a range, `end` can be used as a shorthand for the last index. The capture that was being used before was `@number`, and I updated it to `@constant.builtin` in https://github.com/nvim-treesitter/nvim-treesitter/pull/2977

I don't mind the capture that's used here, as long as it isn't `@keyword` because I think the `end` keyword that's used to close blocks is conceptually different from the index `end`.

savq:
I updated locals.scm. Everything is working fine now :)

queries/julia/highlights.scm
clason: If this is specific to the `end` in `foo[1:end]` and doesn't match `do...end` (please confirm), I -- and likely @theHamsta -- fully agree.

clason:
Thanks -- keep them coming!

queries/julia/highlights.scm
savq: > If this is specific to the end in foo[1:end] and doesn't match do...end

Yep. This one is just for indices.

I made a separate rule for `do … end` in line 173 to highlight it as `@function.keyword` since it's just another way to define a lambda.

PR #3620, opened 2022-10-11
Add indents for php anonymous functions

gbprod:


gbprod:
Does not work... I'll fix it...

Issue #3621, opened 2022-10-11
undefined symbol in `*.so` (known working parser)


ExpandingMan:
### Describe the bug

I have a custom parser installed with the following
```lua
local TSParsers = require("nvim-treesitter.parsers")
local parser_config = TSParsers.get_parser_configs()
parser_config.julia = {
    install_info = {
        url = "https://github.com/savq/tree-sitter-julia",
        files = {"src/parser.c"},
        branch = "dev", 
    },
    filetype = "jl",
}
```
This was just working last night on a similar machine (so a very recent commit may be the culprit).

Whenever I load a `*.jl` file I get the following errors:
```
               Error detected while processing BufReadPost Autocommands for "*":
Error executing lua callback: /tmp/.mount_nvimtiyi5g/usr/share/nvim/runtime/filetype.lua:22: Error executing lua: /tmp/.mount_n
vimtiyi5g/usr/share/nvim/runtime/filetype.lua:23: Vim(append):Error executing lua callback: Failed to load parser: uv_dlopen: /
home/expandingman/.local/share/nvim/site/pack/packer/start/nvim-treesitter/parser/julia.so: undefined symbol: tree_sitter_julia
_external_scanner_create
stack traceback:
        [C]: in function '_ts_add_language'
        ...g/usr/share/nvim/runtime/lua/vim/treesitter/language.lua:36: in function 'require_language'
        ...nvimtiyi5g/usr/share/nvim/runtime/lua/vim/treesitter.lua:38: in function '_create_parser'
        ...nvimtiyi5g/usr/share/nvim/runtime/lua/vim/treesitter.lua:96: in function 'get_parser'
        .../start/nvim-treesitter/lua/nvim-treesitter/highlight.lua:161: in function 'start'
        .../start/nvim-treesitter/lua/nvim-treesitter/highlight.lua:167: in function 'attach'
        ...er/start/nvim-treesitter/lua/nvim-treesitter/configs.lua:477: in function 'attach_module'
        ...er/start/nvim-treesitter/lua/nvim-treesitter/configs.lua:500: in function 'reattach_module'
        ...er/start/nvim-treesitter/lua/nvim-treesitter/configs.lua:108: in function <...er/start/nvim-treesitter/lua/nvim-tree
sitter/configs.lua:107>
        [C]: in function 'nvim_cmd'
[...]

PR #3625, opened 2022-10-11
docs: add small note in README clarifying custom parser installation

ExpandingMan:
I got myself very confused over this, see #3621.  This is a small update to the readme that will hopefully prevent others from running into this.


clason:
Good idea in principle, but:
1. I don't think we need to have that information twice (the comment is enough);
2. some parsers require a  `scanner.cc` instead.

amaanq (approved):

amaanq:
Thanks!

PR #3639, opened 2022-10-14
highlights(help): sync queries

seblj:
I just copied the queries over from https://github.com/neovim/tree-sitter-vimdoc.

I am wondering if there is something that could be done with the queries for the languages that comes bundled with neovim. Since the queries from this repo overrides the queries that comes bundled with neovim, the queries now need to be copied over to this repo and also over to `neovim`

I can't think of a good solution to this myself.

Just thought I could ask here while opening this PR in case anyone has a smart solution to this


clason:
>  Since the queries from this repo overrides the queries that comes bundled with neovim, the queries now need to be copied over to this repo and also over to neovim

This is not an issue since the parsers in this repo and in a user's Neovim installation may be different versions, and queries need to match the parser version. This could happen if a user has
1. a stable version (instead of nightly) or
2. a nightly build that comes without parsers (e.g., homebrew with `--HEAD`).

So users may still rely on this plugin to get (newer versions of) bundled parsers and then need the matching queries.

seblj:
> This is not an issue since the parsers in this repo and in a user's Neovim installation may be different versions, and queries need to match the parser version. This could happen if a user has
> 
> 1. a stable version (instead of nightly) or
> 2. a nightly build that comes without parsers (e.g., homebrew with `--HEAD`).
> 
> So users may still rely on this plugin to get (newer versions of) bundled parsers and then need the matching queries.

That's what I figured

seblj:
BTW: I see that the tests are failing with an invalid node at some position, but I don't understand why since I saw the help parser got updated in this repo as well earlier today

theHamsta:
we had this problem earlier with the C parser, that the wrong parser got loaded.

I'm not sure whether we should maintain the help queries here at all, since people can also contribute changes to nightly nvim, tough might also have it here in case people on stable nvim want the latest nightly update of this parser. So having it here for completeness might make sense

seblj:
> I'm not sure whether we should maintain the help queries here at all, since people can also contribute changes to nightly nvim, tough might also have it here in case people on stable nvim want the latest nightly update of this parser. So having it here for completeness might make sense

I was thinking the same thing, but I also think it would affect people more with removing queries from here and say that they should use nightly to receive parser updates. For me it's totally okay by doing it like this, at least for the moment, because it's pretty easy to contribute to update the queries if you notice they are missing

clason:
> say that they should use nightly to receive parser updates. 

Note that not all nightly builds bundle parsers.

On the other hand, it's perfectly OK to "park" this one parser and wait with updates until things stabilize; I've marked it as no longer experimental when switching, but it's not yet ready to be really used for highlighting purposes.

seblj:
Anything else that needs to be done for this to be merged? @clason 

It makes the experience with using the parser quite a lot better IMO.

clason:
> Anything else that needs to be done for this to be merged? @clason

Someone needs to figure out why bundled parsers interfere with our parsers, at least in CI. (Core should give plugins precedence for parsers and queries, so this doesn't happen locally.)

clason:
Included in #3754, thanks!

Issue #3647, opened 2022-10-15
Rust `mut` must be type qualifier, not keyword


Trard:
### Describe the highlighting problem

Rust `mut` must be `TSTypeQualifier`. Currently, this is `TSKeyword`.

### Example snippet that causes the problem

```rust
fn main() {
    let mut string = String::from("Hello");
    
    string.push_str(" World!");
}
```

### Tree-sitter parsing result

```
function_item [0, 0] - [4, 1]
  "fn" [0, 0] - [0, 2]
  name: identifier [0, 3] - [0, 7]
  parameters: parameters [0, 7] - [0, 9]
    "(" [0, 7] - [0, 8]
    ")" [0, 8] - [0, 9]
  body: block [0, 10] - [4, 1]
    "{" [0, 10] - [0, 11]
    let_declaration [1, 4] - [1, 43]
      "let" [1, 4] - [1, 7]
      mutable_specifier [1, 8] - [1, 11]
      pattern: identifier [1, 12] - [1, 18]
      "=" [1, 19] - [1, 20]
      value: call_expression [1, 21] - [1, 42]
        function: scoped_identifier [1, 21] - [1, 33]
          path: identifier [1, 21] - [1, 27]
          "::" [1, 27] - [1, 29]
          name: identifier [1, 29] - [1, 33]
        arguments: arguments [1, 33] - [1, 42]
          "(" [1, 33] - [1, 34]
          string_literal [1, 34] - [1, 41]
            """ [1, 34] - [1, 35]
            """ [1, 40] - [1, 41]
          ")" [1, 41] - [1, 42]
      ";" [1, 42] - [1, 43]
    expression_statement [3, 4] - [3, 31]
      call_expression [3, 4] - [3, 30]
        function: field_expression [3, 4] - [3, 19]
          value: identifier [3, 4] - [3, 10]
          "." [3, 10] - [3, 11]
          field: field_identifier [3, 11] - [3, 19]
        arguments: arguments [3, 19] - [3, 30]
          "(" [3, 19] - [3, 20]
          string_literal [3, 20] - [3, 29]
            """ [3, 20] - [3, 21]
            """ [3, 28] - [3, 29]
          ")" [3, 29] - [3, 30]
      ";" [3, 30] - [3, 31]
    "}" [4, 0] - [4, 1]
```

### Example screenshot

[...]

PR #3514, opened 2022-09-16
highlights(rust): add new capture for lifetimes

Antikyth:
This PR adds a highlight capture for lifetimes in Rust: `@label.lifetime-only-in-rust`. Currently, lifetimes are simply picked up as `@label`; the highlighting of this capture should default to `@label`, so existing hightlighting still works.

I was not sure if a highlight group needed to be added alongside this. _The current state of this PR does not include a new highlight group for lifetimes_, but [I have a branch prepared](https://github.com/Antikyth/nvim-treesitter/tree/lifetimes-highlight-group) which does add an accompanying highlight group: `TSLabelRustLifetime`.

I think that maybe the fallback to `@label` might not be automatic, and that that highlight group might actually be required. I added it in a separate branch though, as I'm not certain. I'd like some feedback on that.

Issue #3262, opened 2022-08-03
Track languages with missing `TSMethodCall` and/or `TSFunctionCall`


lfenzo:
#3222 reintroduces the `TSMethodCall` and `TSFunctionCall`highlight groups allowing the distinction between the definition of functions/methods and their calls throughout the highlighted code. Although some languages had their `queries/*/highlights.scm` modified in #3222, most of them still need to be ported in order to support the new hl groups (see discussion in #3210 for instructions on how to adapt a language query).

This issue tracks all the languages which either require changes in their `TSFunction` and `TSMethod` groups or don't even have them defined in their `highlights.scm` file. The languages that currently require attention are:
- [ ] astro
- [ ] beancount
- [ ] bibtex
- [ ] clojure
- [ ] cmake
- [ ] comment
- [ ] commonlisp
- [ ] cooklang
- [ ] c_sharp
- [ ] css
- [ ] cuda
- [ ] d
- [ ] dart
- [ ] devicetree
- [ ] dockerfile
- [ ] dot
- [ ] ecma
- [ ] eex
- [ ] elm
- [ ] elvish
- [ ] embedded_template
- [ ] fennel
- [ ] fish
- [ ] foam
- [ ] fortran
- [ ] fusion
- [ ] gdscript
- [ ] gleam
- [ ] glimmer
- [ ] glsl
- [ ] godot_resource
- [ ] gomod
- [ ] gowork
- [ ] graphql
- [ ] hack
- [ ] haskell
- [ ] hcl
- [ ] heex
- [ ] help
- [ ] hjson
- [ ] hocon
- [ ] html
- [ ] html_tags
- [ ] http
- [ ] javascript
- [ ] jsdoc
- [ ] json
- [ ] json5
- [ ] jsonc
- [ ] jsx
- [ ] lalrpop
- [ ] languages.txt
- [ ] latex
- [ ] ledger
- [ ] llvm
- [ ] m68k
- [ ] make
- [ ] markdown
- [ ] markdown_inline
- [ ] ninja
- [ ] nix
- [ ] ocaml
- [ ] ocaml_interface
- [ ] ocamllex
- [ ] pascal
- [ ] phpdoc
- [ ] pioasm
- [ ] prisma
- [ ] proto
- [ ] pug
- [ ] ql
- [ ] qmljs
- [ ] query
- [ ] r
- [ ] rasi
- [ ] regex
- [ ] rego
- [ ] rnoweb
- [ ] rst
[...]

PR #3648, opened 2022-10-15
`@type.qualifier` and `@storageclass` in rust

Trard:
fixes https://github.com/nvim-treesitter/nvim-treesitter/issues/3647

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -156,15 +156,16 @@ effect on highlighting. We will work on improving highlighting in the near futur
 #### Types
 
 ```scheme
-@type            ; type or class definitions and annotations
-@type.builtin    ; built-in types
-@type.definition ; type definitions (e.g. `typedef` in C)
-@type.qualifier  ; type qualifiers (e.g. `const`)
+@type                  ; type or class definitions and annotations
+@type.builtin          ; built-in types
+@type.definition       ; type definitions (e.g. `typedef` in C)
+@type.qualifier        ; type qualifiers (e.g. `const`)
 
-@storageclass    ; visibility/life-time/etc. modifiers (e.g. `static`)
-@attribute       ; attribute annotations (e.g. Python decorators)
-@field           ; object and struct fields
-@property        ; similar to `@field`
+@storageclass          ; visibility/life-time modifiers
+@storageclass.lifetime ; life-time modifiers (e.g. `static`)
+@attribute             ; attribute annotations (e.g. Python decorators)
+@field                 ; object and struct fields
+@property              ; similar to `@field`
 ```
 
 #### Identifiers
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -25,7 +25,6 @@
 
 (self) @variable.builtin
 
-(lifetime   ["'" (identifier)] @label)
 (loop_label ["'" (identifier)] @label)
 
 
@@ -172,7 +171,6 @@
 [
   "async"
   "await"
-  "const"
   "default"
   "dyn"
   "enum"
@@ -182,16 +180,25 @@
   "match"
   "move"
   "pub"
-  "ref"
-  "static"
   "struct"
   "trait"
   "type"
   "union"
   "unsafe"
   "where"
-  (mutable_specifier)
 ] @keyword
+
+[
+ "ref"
+ (mutable_specifier)
+] @type.qualifier
+
+[
+ "const"
+ "static"
+] @storageclass
+
+(lifetime ["'" (identifier)] @storageclass.lifetime)
 
 "fn" @keyword.function
 [


theHamsta:
`@type.qualifier` is only used for vala/blueprint. There was a PR by @gpanders  in the past that added multiple captures from upstream Neovim to the documentation. Those captures are mostly not used yet (e.g. `@debug`, @type.definition`, `@type.qualifier`, `@storageclass`).


It is a major effort to add a new capture and I usually ask to add them for all languages or track the progress for language support so that we keep consistent highlighting. This is also to answer the question on how the new capture should be documented and how it is different from existing and to-be-added ones.

Examples of how we followed this strategy are
- https://github.com/nvim-treesitter/nvim-treesitter/pull/3567
- https://github.com/nvim-treesitter/nvim-treesitter/issues/3262

Another example of adding an isolated capture for only one language is
- https://github.com/nvim-treesitter/nvim-treesitter/pull/3154

Trard:
> This is also to answer the question on how the new capture should be documented and how it is different from existing and to-be-added ones.

@type.qualifier is a keyword that qualifies usage of a variable.

> It is a major effort to add a new capture and I usually ask to add them for all languages or track the progress for language support so that we keep consistent highlighting.

So we need to create a track?

theHamsta:
> `@type.qualifier` is a keyword that qualifies usage of a variable.

but there is also `@storageclass`, probably we can just live with `@type.qualifier` only. Our documentation now gives examples for when to use `@storageclass`. I think it would help us a lot to answer what should be `@type.qualifier` in the C/C++ family of languages, or we can just tackle `const`/`mut` for now and create an issue were we describe for what else we want to use or just complete Rust for now for `@type.qualifier` only or `@type.qualifier`/`@storageclass` and create an issue with ticks for all other languages so that we need to migrate.

`@type.qualifier`/`@storageclass` might get complicated and interpreted differently by people. Is `pub` `public` a qualifier on the type or does it "affect on how a variable" is stored? The documentation does not give of us much without references in major languages. When the distinction is too complicated some of the highlight might not be necessary. @gpanders  might want to chime in on how to introduce a new capture. Ideally, I would like to see PRs that introduce the capture for multiple languages to see whether everyone has the same understanding of the capture.

Trard:
> I think it would help us a lot to answer what should be @type.qualifier in the C/C++ family of languages

I think `@type.qualifier`/`@storageclass` means really different things.

`@storageclass` affects how a variable is stored in memory.
For example:
-  C++: extern, static, thread_local, (Probably mutable, because [reference says](https://en.cppreference.com/w/cpp/language/cv#:~:text=The%20C%2B%2B%20language%20grammar%20treats%20mutable%20as%20a%20storage%2Dclass%2Dspecifier%2C%20rather%20than%20a%20type%20qualifier%2C%20but%20it%20does%20not%20affect%20storage%20class%20or%20linkage.)
**"The C++ language grammar treats mutable as a [storage-class-specifier](https://en.cppreference.com/w/cpp/language/storage_duration), rather than a type qualifier, but it does not affect storage class or linkage."**)
- Rust: static
- Java: static

`@type.qualifier` affects how a variable is used.
For example:
- C++: const, volatile
- Rust: const, mut, ref
- JavaScript: const
- Java: final

Pub, public, private, protected, etc aren't `@storageclass` because they don't affect how a variable is stored in memory. Also, they aren't `@type.qualifier` because they don't affect **how** a variable is used. They affect **where** something (like rust `mod`, `struct`, `field of struct`, etc) can be used.

I think it would be nice to create another track `@storageclass`/`@type.qualifier` issue. And leave this PR only for rust.

theHamsta:
> I think it would be nice to create another track @storageclass/@type.qualifier issue. And leave this PR only for rust.

Sounds good. Could you then add `@type.qualifier`/`@storage` to Rust.

I'm now also seeing that the name  `@type.qualifier` is a bit unfortunate as it doesn't follow the scoping subcategory scheme. It's not a subcategory of `type` and would rather be a `@qualifier.type` or `@keyword.qualifier`. But I think we just leave it the way it is as it's now also document in upstream Neovim. Color schemes will then need  `@type.qualifier as fallback to `@type` will be unexpected

clason:
> document in upstream Neovim. 

That can be changed.

theHamsta:
@clason just for vala/blueprint (blueprint is used for vala or other languages using GTK).

clason:
(I meant "Neovim repo" -- I don't think those captures are documented upstream at all.)

gpanders:
>EDIT In fact, grepping across the repo for @type.qualifier doesn't yield anything?

The only reference was removed in https://github.com/neovim/neovim/commit/030b422d1e9517ed1b1c70fd8002b74881c80650. But it was never documented (outside of the source code).

FWIW I agree with @Trard's explanation on the difference between "type qualifier" and "storage class" 100%. The distinction does not make sense in all languages of course, but in those that it does it's a valid distinction.

>It's not a subcategory of type and would rather be a @qualifier.type or @keyword.qualifier.

We can change this -- as @clason mentioned it's not documented upstream so it's fair game :)

gpanders:
@Trard In Rust doesn’t `const` mean “evaluated at compile time”? Do `const` variables or functions actually exist in the final binary, or do they only ever exist in memory during compilation?

If that is indeed the case (I am not an expert on Rust `const` semantics) I would argue that `const` in Rust is a `@storageclass`, not a type qualifier.

Trard:
> @Trard In Rust doesn’t `const` mean “evaluated at compile time”? Do `const` variables or functions actually exist in the final binary, or do they only ever exist in memory during compilation?

You are right. Rust consts are essentially inlined(copied) wherever they are used. If it possible they actually don't exist in binary.

ObserverOfTime:
I believe `@storageclass` should also be used for lifetime (replacing #3514).

Trard:
> I believe `@storageclass` should also be used for lifetime (replacing #3514).

`@label` has no relation with lifetimes. Lifetimes affect how (long) a variable is stored in memory, so I added they to `@storageclass`.

ObserverOfTime:
`@storageclass.lifetime` won't work.

Trard:
> `@storageclass.lifetime` won't work.

Why?
image
image

Theme:
image

ObserverOfTime:
Because the CI won't allow it. You'll have to rebase and add it to the "Non-standard" section in `CONTRIBUTING.md`.
But, I think that section should be removed and captures that are *too* specific should only be used in local overrides.

theHamsta:
> > I believe @storageclass should also be used for lifetime (replacing https://github.com/nvim-treesitter/nvim-treesitter/pull/3514).

> @label has no relation with lifetimes. Lifetimes affect how (long) a variable is stored in memory, so I added they to @storageclass.

I would try to be careful when going this specific with highlights. It is fantastic when you can go this specific with the distinction but you also have to consider how it affects users that just want very basic distinction between reserved keywords and identifiers (think of basic vim syntax files that only highlight keywords). I don't see a problem with using `@storageclass` for lifetimes, but it can cause such an accidental switch from the keywords to identifiers. Something similar happened with `@structure` which seems to be thought as keyword category but was suggested in PRs for highlight identifiers for `struct` because this seems what the name appear to suggest. No change required for this PR, but something to keep in mind when porting other languages to avoid random flips in highlighting for users.
 
`:h group-name` gives explanations and example for hl-groups, but even the C syntax files don't use the highlight groups for the examples given. Some other strange examples include `Debug` (what is meant with that?), `Statement` (like how "any statement", syntax files seem to use it as "any keyword" or ), `Identifier` (syntax files seem to use this very rarely highlight any identifiers, but color schemes seem to like to use it for a nice color)   

@ObserverOfTime CI should allow now arbitrary captures as long as they start with a know prefix.

theHamsta (approved):

Trard:
> @ObserverOfTime CI should allow now arbitrary captures as long as they start with a know prefix.

Should lifetimes be `@storageclass` or `@storageclass.lifetime`?

theHamsta:
I would like `@storageclass.lifetime`. You can fix the CI, in case it is complaining. @clason @gpanders is this ready to merge?

ObserverOfTime:
I don't think it's worth adding a new capture that's only used once in a single language. Would we be able to use it in another language? Maybe if we renamed it to something more generic?

ObserverOfTime:
> I'm now also seeing that the name `@type.qualifier` is a bit unfortunate as it doesn't follow the scoping subcategory scheme. It's not a subcategory of `type` and would rather be a `@qualifier.type` or `@keyword.qualifier`.

I had an idea for this:

```scheme
@modifier         ; various type/variable modifiers (e.g. `mut` / `final`)
@modifier.access  ; modifiers that affect access (e.g. `public` / `private`)
@modifier.inherit ; modifiers related to inheritance (e.g. `override`)
```

We could also rename `@storageclass` to `@modifier.storage` but that would have to be changed upstream too.

clason:
> We could also rename @storageclass to @modifier.storage but that would have to be changed upstream too.

Not a problem.

Trard:
> I would like `@storageclass.lifetime`. You can fix the CI, in case it is complaining.

CI is complaining. How do I fix it?

clason:
I don't know why it's complaining (`@storageclass` is on the allowed list in `CONTRIBUTING.md`).

But I strongly prefer the suggestion from @ObserverOfTime about using `@modifier`. We should aim for fewer and more generic top-level captures.

ObserverOfTime:
> I don't know why it's complaining (`@storageclass` is on the allowed list in `CONTRIBUTING.md`).

But `@storageclass.lifetime` isn't, we've been over this. :)

> Hmm, `StorageClass` is a standard legacy syntax group, though.

Which is why I'm reluctant to change `@storageclass`.
Sure, `@type.definition` maps to `Typedef`, but that's because `@type` also maps to a legacy group.
That won't be the case for `@modifier` so having just `@modifier.storage` upstream seems weird to me.
Plus, every instance of `@storageclass` here, upstream, and in themes would have to be changed.

clason:
> But @storageclass.lifetime isn't, we've been over this. :)

Yes, but the CI should only check top-level captures, we've talked about this ;)

ObserverOfTime:
> Yes, but the CI should only check top-level captures, we've talked about this ;)

I don't know what it *should* do, I know what it *does*.  ¯\\\_(ツ)\_/¯

clason:
https://github.com/nvim-treesitter/nvim-treesitter/pull/3479

clason:
> That won't be the case for `@modifier`

Yes, but as I wrote, that's easy to change. We have a chance here to make a more rational set of highlight/syntax groups, and not the legacy "this made sense for C, so we're gonna use it for ALL THE LANGUAGES forever". (Which we can still link to legacy groups, however it makes sense.)

clason:
> > I would like `@storageclass.lifetime`. You can fix the CI, in case it is complaining.
> 
> CI is complaining. How do I fix it?

To make CI pass, you can add `@storageclass.lifetime` to the list in `CONTRIBUTING.md`.

Issue #3572, opened 2022-10-02
Update to Nvim 0.8 highlighting changes


clason:
Now that Nvim 0.8 is released, we should yeet all traces of the old `TSFoo` highlight groups -- including the tests. 

In particular the one that is failing now (and thus blocking all PRs) that the stable version doesn't set these by default anymore -- @theHamsta

theHamsta:
Would be nice to have a PR for that. I will return from vacation on oct 9

theHamsta:
The highlight tests are already able to test for queries only. One would just have to remove the translation to highlight groups in the tests

clason:
Tests should pass again with my PR that you just merged; but the tests explicitly check for `TSFoo` highlights, so we can't remove those groups until we (search and) replace those in the corpus.

theHamsta:
We should also tag a v0.8.0 release of nvim-treesitter. Maybe @vigoux could do that with his admin privileges?

clason:
I'd rather tag a v.0.7.0 to indicate the last version compatible with Nvim 0.7?

theHamsta:
The point is rather to have a snapshot that will work with whatever release the user choses. In the past, we have snapshotted more or less with nvim releases. We could wait for 0.8.0 until we remove the other highlight groups

clason:
Neovim 0.8 got released last Friday...

clason:
A separate issue, lest I forget, is to remove the upstreamed functions and instead call their Neovim version. (But I'd rather wait for some refactoring on those to be done first.)

This also applies to the highlighting module that can just be replaced with `vim.treesitter.start()`.

clason:
Closed in #3656

PR #3654, opened 2022-10-16
docs: document all the captures in CONTRIBUTING.md

ObserverOfTime:
I used `scheme` as the language marker so that GitHub can highlight the comments properly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,150 +89,175 @@ effect on highlighting. We will work on improving highlighting in the near futur
 
 #### Misc
 
-```
-@comment
-@debug
-@error for error `ERROR` nodes.
-@none to disable completely the highlight
-@preproc
-@punctuation.delimiter for `;` `.` `,`
-@punctuation.bracket for `()` or `{}`
-@punctuation.special for symbols with special meaning like `{}` in string interpolation.
+```scheme
+@comment  ; line and block comments
+@error    ; syntax/parser errors
+@none     ; completely disable the highlight
+@preproc  ; various preprocessor directives & shebangs
+@define   ; preprocessor definition directives
+@operator ; symbolic operators (e.g. `+` / `*`)
 ```
 
-#### Constants
+#### Punctuation
 
+```scheme
+@punctuation.delimiter ; delimiters (e.g. `;` / `.` / `,`)
+@punctuation.bracket   ; brackets (e.g. `()` / `{}` / `[]`)
+@punctuation.special   ; special symbols (e.g. `{}` in string interpolation)
 ```
-@constant
-@constant.builtin
-@constant.macro
-@string
-@string.regex
-@string.escape
-@string.special
-@character
-@character.special
-@number
-@boolean
-@float
+
+#### Literals
+
+```scheme
+@string            ; string literals
+@string.regex      ; regular expressions
+@string.escape     ; escape sequences
+@string.special    ; other special strings (e.g. dates)
+
+@character         ; character literals
+@character.special ; special characters (e.g. wildcards)
+
+@boolean           ; boolean literals
+@number            ; numeric literals
+@float             ; floating-point number literals
 ```
 
 #### Functions
 
-```
-@function
-@function.builtin
-@function.call
-@function.macro
-@parameter
+```scheme
+@function         ; function definitions
+@function.builtin ; built-in functions
+@function.call    ; function calls
+@function.macro   ; preprocessor macros
 
-@method
-@method.call
-@field
-@property
+@method           ; method definitions
+@method.call      ; method calls
 
-@constructor
+@constructor      ; constructor calls and definitions
+@parameter        ; parameters of a function
 ```
 
 #### Keywords
 
-```
-@conditional (e.g. `if`, `else`)
-@repeat (e.g. `for`, `while`)
-@label for C/Lua-like labels
-@keyword
-@keyword.function (keyword to define a function, e.g. `func` in Go, `def` in Python)
-@keyword.operator (for operators that are English words, e.g. `and`, `or`)
-@keyword.return
-@operator (for symbolic operators, e.g. `+`, `*`)
-@exception (e.g. `throw`, `catch`)
-@include keywords for including modules (e.g. import/from in Python)
-@storageclass
+```scheme
+@keyword          ; various keywords
+@keyword.function ; keywords that define a function (e.g. `func` in Go, `def` in Python)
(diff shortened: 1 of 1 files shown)


clason:
Gotta adapt the CI script that scrapes these then, I'm afraid...

ObserverOfTime:
> Gotta adapt the CI script that scrapes these then, I'm afraid...

I'll look into that.

clason:
It's just the `@spell` and `@conceal` captures, I believe -- the script can't find them inside backticks.

clason:
Guess I was wrong; it only reports the first error. It looks like it can't find any captures...

ObserverOfTime:
> Invalid capture @variable.global in highlights for ruby.

I removed this capture since it doesn't actually have any highlighting defined.
I guess the CI doesn't fall back to checking the parent for non-standard captures.

clason:
> I guess the CI doesn't fall back to checking the parent for non-standard captures.

It should; but maybe it's not working properly? @theHamsta

ObserverOfTime:
I'll add a block with non-standard captures for now and see if that appeases the CI.

ObserverOfTime:
Yep, that worked.

clason:
Looks good to me!

clason (approved):

CONTRIBUTING.md
```diff
@@ -89,150 +89,176 @@ effect on highlighting. We will work on improving highlighting in the near futur
...
-@debug
-@error for error `ERROR` nodes.
-@none to disable completely the highlight
-@preproc
-@punctuation.delimiter for `;` `.` `,`
-@punctuation.bracket for `()` or `{}`
-@punctuation.special for symbols with special meaning like `{}` in string interpolation.
+```scheme
+@comment  ; line and block comments
+@debug    ; debugging statements [NOTE: unused]
```
theHamsta, line 94: let's remove this, when nobody can give an example what is meant by debugging statements

theHamsta:
Can we remove all documentation of `TS*` in `doc/`? It would be good to have just one source of truth for the definition of the captures and `TS*` is deprecated anyway. Captures are mostly documented in upstream Neovim.

clason:
> Can we remove all documentation of `TS*` in `doc/`? It would be good to have just one source of truth for the definition of the captures and `TS*` is deprecated anyway. 

Yes, the `TS*` groups should be removed now that Nvim 0.8 is released (see https://github.com/nvim-treesitter/nvim-treesitter/issues/3572)

> Captures are mostly documented in upstream Neovim.

No, they are not. Neovim does and will not _prescribe_ any specific captures (except that any `@foo.bar.<lang>` capture will automatically be mapped to a highlight group of the same name).

clason:
> Yes, the `TS*` groups should be removed now that Nvim 0.8 is released (see #3572)

Done: https://github.com/nvim-treesitter/nvim-treesitter/pull/3656

CONTRIBUTING.md
clason: @theHamsta is the capture czar, but if it's not used by any current query, I'd suggest leaving it out for now; if somebody ends up needing it, they can easily add it (and provide a good description).

CONTRIBUTING.md
```diff
@@ -89,150 +89,176 @@ effect on highlighting. We will work on improving highlighting in the near futur
...
+@text.math             ; math environments (e.g. `$ ... $` in LaTeX)
+@text.environment      ; text environments of markup languages
+@text.environment.name ; text indicating the type of an environment
+@text.reference        ; text references, footnotes, citations, etc.
+
+@text.note             ; info notes
+@text.warning          ; warning notes
+@text.danger           ; danger/error notes
+
+@todo                  ; todo notes [NOTE: unused]
```
clason, line 206: I'd remove that one, too -- it stands out like a sore thumb in this list.

CONTRIBUTING.md
```diff
@@ -89,150 +89,176 @@ effect on highlighting. We will work on improving highlighting in the near futur
...
-@preproc
-@punctuation.delimiter for `;` `.` `,`
-@punctuation.bracket for `()` or `{}`
-@punctuation.special for symbols with special meaning like `{}` in string interpolation.
+```scheme
+@comment  ; line and block comments
+@error    ; syntax/parser errors
+@none     ; completely disable the highlight
+@preproc  ; various preprocessor directives & shebangs
+@define   ; preprocessor definition directives [NOTE: unused]
```
clason, line 97: Why did you add it if it's not used? (Honestly curious.)

clason (commented):
For the remaining unused captures that match the builtin groups, I'm in principle OK with keeping them (but would drop the "unused" note).

Or better yet, actually use them :) (@gpanders)

ObserverOfTime:
I plan to submit another PR soon that will make some queries more compliant with the documentation and use most if not all of the unused captures. I can drop the note but I'd rather not remove them entirely for the time being.

CONTRIBUTING.md
ObserverOfTime: Maybe we could rename it to `@text.todo`?

CONTRIBUTING.md
ObserverOfTime: I wanted to document every capture that had a `TS*` equivalent.
I didn't add `@macro` and `@structure` since they didn't have one.

clason: Ah, that makes sense. But we don't look back; `TS*` is dead and gone now ;)

CONTRIBUTING.md
clason: Yes, that would be good (and maybe use it for `@comment` highlight in your follow-up PR).

CONTRIBUTING.md
```diff
@@ -89,150 +89,175 @@ effect on highlighting. We will work on improving highlighting in the near futur
...
 
-#### Variables
+```scheme
+@type            ; type or class definitions and annotations
+@type.builtin    ; built-in types
+@type.definition ; type definitions (e.g. `typedef` in C)
+@type.qualifier  ; type qualifiers (e.g. `const`)
 
+@storageclass    ; visibility/life-time/etc. modifiers (e.g. `static`)
+@attribute       ; attribute annotations (e.g. Python decorators)
```
Diaoul, line 165: > I plan to submit another PR soon that will make some queries more compliant with the documentation and use most if not all of the unused captures. I can drop the note but I'd rather not remove them entirely for the time being.

I imagine this is the reason why this is not the case just yet for Python decorators :hugs:

PR #3656, opened 2022-10-16
feat!: remove obsolete `TS*` highlight groups

clason:
Treesitter highlight groups are handled by upstream Nvim since 0.8.

BREAKING CHANGE: assign highlights to captures directly (e.g., `hi link @keyword Keyword`).

Drops support for Nvim 0.7


vigoux (approved):

theHamsta (approved):

clason:
Closes #3572

lua/nvim-treesitter/highlight.lua
```diff
@@ -175,104 +41,4 @@ function M.detach(bufnr)
   enable_syntax(bufnr)
 end
 
--- TODO(clason): remove obsolete function after bump to 0.8
-function M.set_custom_captures(captures)
-  for capture, hlgroup in pairs(captures) do
```
theHamsta, line 180: Removing this probably requires updating the readme

clason: Ah, missed that. I'll make a fixup PR.

numToStr:
I am just wondering, now that the `TS*` hl groups are removed, how can we define language specific hl groups like `rustTSType`?

clason:
`@type.rust`

shadmansaleh:
Do we have nvim-0.7 compat tag? I think we should have them when we drop support for nvim versions. I can see there's a 0.5 compat branch but don't see one for 0.6 or 0.7. Was a decision already made not to have them after 0.5 ?

clason:
Yes, the decision has been made to only actively support 
1. the latest Nvim release
2. the latest commit on `master`

While nvim-treesitter (just like treesitter support in general) is still considered experimental, it makes no sense to make stronger compatibility guarantees.

oxalica:
Just breaks the highlighting after a plugin update. :disappointed: 
I really think there should be a deprecation-warning time window before the removal.

PR #3662, opened 2022-10-17
feat: meson ternaryoperator and buildin vairable

Decodetalkers:
? is the operator of ternaryoperator , so add ? to operator

buildin variable

### DIFF
--- a/queries/meson/highlights.scm
+++ b/queries/meson/highlights.scm
@@ -37,7 +37,9 @@
   ">"
   ">="
 ] @operator
-
+[
+  "?"
+] @conditional.ternary
 [
   "if"
   "elif"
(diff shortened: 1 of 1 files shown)


queries/meson/highlights.scm
```diff
@@ -36,6 +36,7 @@
   "-="
   ">"
   ">="
+	"?"
 ] @operator
```
theHamsta, line 40: There was a discussion in the past on whether ternary should be `@operator` or `@conditional`. The `@conditional` side won (see C/C#/Java). I also preferred `@operator` back then. But we should keep consistency.

 We now query fallbacks. So now a `@conditional.ternary` would be possible if applied to consistently across languages

Decodetalkers: So when will this keyword be available?

theHamsta: Strange, the check was intended to allow everything that starts with `@conditional.Xxxx` let me check what's wrong

Decodetalkers:
@theHamsta Since this pr have not merged, so I add something else, for the variable buildin vaiable, but I think keyword is not right, can you review it? 
Now it is what it is  looks like 
image

tristan957:
Meson keywords look correct.

PR #3678, opened 2022-10-17
feat(html): lit attr binding interpolation

bennypowers:
Fixes https://www.reddit.com/r/neovim/comments/y62tr4/javascript_lit_html_syntax_highlighting_ts/

Before
Screenshot from 2022-10-17 21-12-57

After
Screenshot from 2022-10-17 21-13-08

### DIFF
--- a/queries/html_tags/injections.scm
+++ b/queries/html_tags/injections.scm
@@ -4,4 +4,14 @@
 ((script_element
   (raw_text) @javascript))
 
+; lit-html style template interpolation
+((attribute
+  (quoted_attribute_value (attribute_value) @javascript))
+  (#match? @javascript "\\$\\{")
+  (#offset! @javascript 0 2 0 -1))
+((attribute
+  (attribute_value) @javascript)
+  (#match? @javascript "\\$\\{")
+  (#offset! @javascript 0 2 0 -2))
+
 (comment) @comment


bennypowers:
If we implement your suggestion @thekaganugur we get this:

image

There are two things to note here:
1. the `${` doesn't get parsed up as a template literal part, rather as an expression `$` followed by a block `{}`
2. in the case of arrow functions, I'm not sure how much can be done here, because the HTML parser should probably correctly identify the `>` in the arrow function as closing the tag. Surrounding the part in quotes though solves that problem.

thekaganugur:
@bennypowers 

> the ${ doesn't get parsed up as a template literal part, rather as an expression $ followed by a block {}

This is more than fine thanks.

About the arrow function, do you know where should I open an issue about that? Maybe https://github.com/tree-sitter/tree-sitter-html ?

bennypowers:
they might be willing to make a special case, but strictly speaking html shouldn't parse that as you intent

lucario387:
Isn't the solution here is to inject jsx into that html blocks, instead of injecting html, and add some javascripts to it?

lucario387:
From what I see
- You are injecting jsx-valid strings as html, then have to add extra injections for html to make it work like jsx
- Jsx can be written as valid html, and can be parsed into that by the current jsx parser
- Javascript is actually javascript + jsx (tyvm ReactJS), so injecting javascript will also mean valid jsx elements

bennypowers:
> Javascript is actually javascript + jsx

What does that mean, exactly? JSX is distinct language from JavaScript, and from HTML. I like apples, and sometimes I eat apple pie, but brother - apple pie and apples are two different things.

Not all HTML parses correctly in a jsx program (particularly blank space)

As hard as it may be to believe, this is not about react or jsx, is about HTML and JavaScript

lucario387:
> What does that mean, exactly? JSX is distinct language from JavaScript, and from HTML. I like apples, and sometimes I eat apple pie, but brother - apple pie and apples are two different things.

yes, in general cases, javascript and jsx are separate entites, but as in current nvim-treesitter, javascript is a combination of queries from javascript and jsx, to easen the checks for if it's a react project or not
https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/javascript/injections.scm

lucario387:
Still, I don't think it's nice to add this as HTML :D. Would be nice extra colorings tho

amaanq:
I like this

amaanq (approved):

ObserverOfTime (approved):

amaanq:
@bennypowers Should this go under html_tags instead? If so, add it there and we can merge

renxzen:
Is there a way to fix this when quotes are not used in every attribute value?

image

seems like the problem is that it's not recognizing ${*} as a valid attribute value, unless it has quotes around it

image

PR #3683, opened 2022-10-18
highlights(rust): namespace/type confusion; highlight `for<'a>` and bracketed types better

cormacrelf:
#### 1. `@namespace` -> `@type` fix

```rust
trait Trait { type Assoc; }
fn thing<T: Trait>(arg: T::Assoc) {}
```

Previously, T in `T::Assoc` was highlighted as `@namespace`. Now it's `@type`.

#### 2. `for_lifetimes` highlights

Highlights the for as a keyword, and the angle brackets as brackets.

```rust
type T = &'static dyn for<'a> Fn(&'a str) -> i32;
```

#### 3. `<T as Trait>` brackets (`bracketed_type`)

Highlights the angle brackets as brackets.

```rust
type A = <B as Trait>::Assoc;
<B as Trait>::method(5);
```

#### Before:

Screen Shot 2022-10-18 at 12 17 55 pm

#### After:

Screen Shot 2022-10-18 at 12 17 05 pm

### DIFF
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -268,6 +268,8 @@
 (closure_parameters "|"    @punctuation.bracket)
 (type_arguments  ["<" ">"] @punctuation.bracket)
 (type_parameters ["<" ">"] @punctuation.bracket)
+(bracketed_type ["<" ">"] @punctuation.bracket)
+(for_lifetimes ["<" ">"] @punctuation.bracket)
 
 ["," "." ":" "::" ";"] @punctuation.delimiter
 
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -72,6 +72,9 @@
 (scoped_type_identifier
   path: (identifier) @namespace)
 (scoped_type_identifier
+  path: (identifier) @type
+  (#lua-match? @type "^[A-Z]"))
+(scoped_type_identifier
  (scoped_identifier
   name: (identifier) @namespace))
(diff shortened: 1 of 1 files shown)


clason:
Could you rebase on the latest parser change? I don't think it will affect this, but I want to make sure the tests are run with the current version.

cormacrelf:
Added one more tweak, highlighting self/crate/super in `pub(self)` and all scoped identifiers e.g. `use super::Thing`. I chose `@keyword`.

queries/rust/highlights.scm
```diff
@@ -71,6 +71,9 @@
   name: (identifier) @namespace))
 (scoped_type_identifier
   path: (identifier) @namespace)
+(scoped_type_identifier
+  path: (identifier) @type
+  (#lua-match? @type "^[A-Z]"))
```
clason, line 76: These sort of "ALL CAPS is different" has historically not been very popular; too likely to trigger false positives in some contexts.

cormacrelf: Scroll up and down -- I'm just fixing an existing scheme of differentiation by capitalisation that forgot a case. For Rust it has not been a problem because the compiler has enabled-by-default warnings on types that aren't camel-case, and modules or functions that aren't snake-case.

clason: As long as it's deliberate and targeted for the language, that's fine by me.

clason (approved):

PR #3684, opened 2022-10-18
feat(diff): add diff parser

gbprod:
This commit introduce diff parser.

I'm not sure but I introduce new highlights (`@diff.add` and `@diff.delete`) as I can't find any highlight matching to this usage.

Parser comes from https://github.com/the-mikedavis/tree-sitter-diff

Thanks for your awesome work !

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -203,6 +203,9 @@ Mainly for markup languages.
 @text.note             ; info notes
 @text.warning          ; warning notes
 @text.danger           ; danger/error notes
+
+@text.diff.add         ; added text (for diff files)
+@text.diff.delete      ; deleted text (for diff files)
 ```
 
 #### Tags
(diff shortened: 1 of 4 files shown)


clason:
Maintainers are people maintaining the parser support _in this repo_ (i.e., fixing queries if a parser update breaks them), not the parser itself. So you shouldn't list other people there unless you've verified that they agree to provide support here and react to pings.

gbprod:
Ok, I wasn't aware of that! 
So, I think I should the maintainer here.

Thanks for your feedback!

clason:
> I'm not sure but I introduce new highlights (@diff.add and @diff.delete) as I can't find any highlight matching to this usage.

I'm wary of adding new top-level groups that are this specific. Maybe `@text.diff.add` would be better (and not require changes to `CONTRIBUTING.md`)?

@theHamsta

ObserverOfTime:
> Maybe `@text.diff.add` would be better (and not require changes to `CONTRIBUTING.md`)?

The new groups will have to be added either way as proven in https://github.com/nvim-treesitter/nvim-treesitter/pull/3654#issuecomment-1279934539.

ObserverOfTime:
I'm against using this parser for the `diff` filetype because it doesn't support every diff format commonly used (see https://github.com/the-mikedavis/tree-sitter-diff/issues/1#issuecomment-1246994096). It only supports the unified format, as used by `git-diff`.

I propose renaming the filetype to `gitdiff` and using it only as a [git-commit](https://github.com/the-mikedavis/tree-sitter-git-commit) (https://github.com/nvim-treesitter/nvim-treesitter/issues/2282#issuecomment-1272294543) injection.

theHamsta (approved):
Looks good to me! Waiting just a bit for more comments. If @clason or @kyazdani42 think this is save to merge please go ahead

PR #3685, opened 2022-10-18
highlights(rust): panic!() and assert!() as @exception

cormacrelf:
This is what rust.vim used to do with the rustPanic and rustAssert highlight groups. If I'm remembering right, this is slightly better as it also catches `debug_assert!()`.

Edit: for clarity `println!` still highlights as `@function.macro`.

Screen Shot 2022-10-18 at 1 30 46 pm

### DIFF
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -273,3 +273,8 @@
 (inner_attribute_item ["!" "#"] @punctuation.special)
 (macro_invocation "!" @function.macro)
 (empty_type "!" @type.builtin)
+
+(macro_invocation macro: (identifier) @_ident @exception "!" @exception
+ (#eq? @_ident "panic"))
+(macro_invocation macro: (identifier) @_ident @exception "!" @exception
+ (#contains? @_ident "assert"))


theHamsta:
@vigoux

theHamsta:
Not sure if `@exception` should be reserved to keywords only. Did have a problem with `@macro` until now. But not a strong opinion, `@exception` would fit also.

queries/rust/highlights.scm
```diff
@@ -273,3 +273,8 @@
 (inner_attribute_item ["!" "#"] @punctuation.special)
 (macro_invocation "!" @function.macro)
 (empty_type "!" @type.builtin)
+
+(macro_invocation macro: (identifier) @_ident @exception "!" @exception
+ (#eq? @_ident "panic"))
+(macro_invocation macro: (identifier) @_ident @exception "!" @exception
+ (#contains? @_ident "assert"))
```
MunifTanjim, line 280: I don't think `assert` should be highlighted with `@exception`. `assert` will only fail in the case where expectation isn't met. Just as `assert` can result in exception, any function can do the same... doesn't mean everything should be highlighted with `@exception`.

cormacrelf: The same argument would apply to any language that has a `throw` keyword without the type system tracking function throwability, yet we still [highlight the `throw` as `@exception` in C++](https://github.com/nvim-treesitter/nvim-treesitter/blob/9ada5f70f98d51e9e3e76018e783b39fd1cd28f7/queries/cpp/highlights.scm#L119-L124). The `assert!` macro is sugar for `if _thing_ { panic!(...); }`. There is a meaningful difference between "calling a function (any function may throw!)" and "calling panic!() or assert!()" which is essentially the origin point of any panic.

I believe, like the original rust.vim authors, that highlighting it as an exception-throwing mechanism is useful because it becomes prominent, as all "very local possible panic" points should. I am tempted to argue the same for "unwrap" and "expect", but those are a little more user-defined, as opposed to being macros in prelude. Nevertheless, scanning code for "danger" sections is made markedly easier by doing this and I really missed it from the old rust.vim.

clason:
Could you rebase on the latest parser change?

cormacrelf:
Done.

clason (approved):

PR #3688, opened 2022-10-18
ci: publish parsers as prerelease

theHamsta:
This could be the basis for a commend `:TSUpdateBinary`

Comparison between different release mechanism:

- GH packages
        :-1:  requires supported package manager: e.g. npm, ruby gems, vcpkg, container images (how does brew use GH packages? Via Ruby gems? Is it a packages manager supported by GH backend?)
        :+1:  arbitrary versions available
- Create prerelease with individual parsers as release artifacts:
        :-1: :   works for one OS, multiple OSes hit Github API (solvable with custom GH API key?)
        :-1:  only latest available
- Create prerelease with all parsers bundled as artifacts:
        :+1:  works for multiple OSes hit Github API, file size of bundles ~7-~15MB (~100MB uncompressed, pick only those needed for TSUpdateBinary, cl.exe output bigger than by other compilers) -> https://github.com/theHamsta/nvim-treesitter/releases
        :-1:  only latest available

For all :    
 :-1: only have parsers our CI currently compilers. For some configs, we're missing parsers that require npm (at least I think that not all parsers are not always installed except for Ubuntu).

Binaries will be available as `https://github.com/nvim-treesitter/nvim-treesitter/releases/download/latest-parsers-{{ matrix.os }}-{{ matrix.cc }}/parsers-{{ matrix.os }}-{{ matrix.cc }.tar.gz`

### DIFF
--- a/.github/workflows/test-queries.yml
+++ b/.github/workflows/test-queries.yml
@@ -84,3 +84,16 @@ jobs:
 
       - name: Check query files
         run: $NVIM --headless -c "luafile ./scripts/check-queries.lua" -c "q"
+
+      - name: Compress parsers
+        run: tar czf parsers-${{ matrix.os }}-${{ matrix.cc }}.tar.gz parser
+
+      - uses: "marvinpinto/action-automatic-releases@latest"
+        if: ${{ github.ref == 'refs/heads/master' }}
+        with:
+           repo_token: "${{ secrets.GITHUB_TOKEN }}"
+           automatic_release_tag: "latest-parsers-${{ matrix.os }}-${{ matrix.cc }}"
+           prerelease: true
+           title: "Parser artifacts (${{ matrix.os }}-${{ matrix.cc }})"
+           files: |
+             parsers-${{ matrix.os }}-${{ matrix.cc }}.tar.gz


clason:
I think this could/should be part of transitioning to a release schedule for nvim-treesitter (so binary parsers are part of a release, while people that want to follow HEAD to have the latest and greatest can be expected to still compile themselves).

ObserverOfTime:
Couldn't all platforms be in the same release?

theHamsta:
> I think this could/should be part of transitioning to a release schedule for nvim-treesitter (so binary parsers are part of a release, while people that want to follow HEAD to have the latest and greatest can be expected to still compile themselves).

> Couldn't all platforms be in the same release?

Answers to both questions, this isn't really about releases, but I'm rather abusing releases to store binary artifacts at fixed URLs. The workflow I'm currently using isn't meant for binary artifacts and using the same release tag/title will cause the different workflow to overwrite each other. There are other workflows specialized in generating the artifacts for releases (given the release, they will generate the artifacts on the `"release"` trigger).

Was thinking of having the plugin+artifacts in on tar. I didn't find resources on how well plugin managers support that pattern and how to handle updates in that case. What I actually wanted was something like GH packages were you can have files associated with a semver that can judge whether something newer exists. Something like GH packages could live independent of this repo. It could also maintain some security verification, e.g.:
- `scanner.cc` must have a checksum in the repo (i.e. at least once reviewed)
- `grammar.json` is extracted via tree-sitter-cli

Derived artifacts would be only allowed to be generated from `scanner.cc` and `grammar.json`. Are binary storages for nix a solution to this?
 

Releases would maybe have the advantage that we could do a security check on the currently distributed parsers. Right now any external parser could just slip in arbitrary binaries. Until now I failed at tagging the repo. Until now I thought that I at least tagged on a Neovim release but apparently not https://github.com/nvim-treesitter/nvim-treesitter/tags (Remembering now that we had branches for that)??? It doesn't seem to be very typical be plugins and I don't know who would maintain that. @clason do you want admin rights for this tasks?

About security, wasm binaries would also be useful. I think it is possible to implement the `TSLanguage` C structure also with a WASM interface that reads wasm parser binaries or to JIT compile a language description. Since `TSLanguage` is more or less just a vtable, one could implement something like this without changes to Neovim and still use the speed of compiled for trusted sources.

clason:
My thoughts:

1. We need to think really hard about how we can ensure maintaining nvim-treesitter long-term.
2. Tagging commits before breaking changes (even retroactively for 0.8) would be a feasible compromise for now (even if I'm not ready to make stability guarantees just yet; we're right in the middle of a very productive refactoring/upstreaming period).
3. Binary parser distribution is really something that upstream (tree-sitter) needs to take care of.
4. I wouldn't touch Nix with a ten-foot pole.
5. Security for Neovim plugins is a pipe dream; there's just no way we (as in "Neovim") could provide anything that would stand up to serious security requirements.

theHamsta:
> I wouldn't touch Nix with a ten-foot pole.

There is nix as a developer tool and nix as CI build environment. Was thinking of the later in the end it will just produce binaries. They have a lot of tooling around 

> Binary parser distribution is really something that upstream (tree-sitter) needs to take care of.

Agreed. But I feel that there is no urgency in upstream to find a solution for end users as tree-sitter is only consumed by dev while wrapped in language extension for end users. So it would be good to think of a solution to contribute upstream or propose as a standalone project.

> Security for Neovim plugins is a pipe dream; there's just no way we (as in "Neovim") could provide anything that would stand up to serious security requirements.

It is normal that each plugin needs to be trusted, but we're a bit extreme by forwarding 131 external dependencies. If we could verify all of them with a single by a single workflow it would be a benefit for all of us. For Vim, tree-sitter is out-of-question for being a binary distribution.

While working on this PR, parsers are really effectively compressible. You can see that with the tars produced by this PR and also by Helix which stays small even when distributing lots of TS parsers in their release. Would be maybe worth to directly support loading `.so.tar.gz`

clason:
I don't think it's reasonable to try to do upstream's job here -- especially since upstream has actually started working on that.

Issue #3642, opened 2022-10-14
Documentation doesn't mention how to unfold everything at startup


jhaand:
### Describe the bug

When enabling folding with Treesitter, all the folds remain closed when opening the file. I would like to configure nvim to open all the folds on startup. I can't find any documentation on how enable this behaviour. 

### To Reproduce

Observed:\
Convigured nvim with treesitter as follows
```-- Treesitter folding
vim.wo.foldmethod = 'expr'
vim.wo.foldexpr = 'nvim_treesitter#foldexpr()' ```

Open a supported filetype. 

### Expected behavior

Now all the folds are closed when opening a file. I would rather have them opened. Now I have to start every file with a `zR`. 

I would expect a proper method for achieving this described in the documentation. 

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
  - OK: `node` found v18.7.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (Debian 12.2.0-1) 12.2.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - lua            ✓ ✓ ✓ ✓ ✓
  - bash           ✓ ✓ ✓ . ✓
  - rust           ✓ ✓ ✓ ✓ ✓
  - toml           ✓ ✓ ✓ ✓ ✓
  - help           ✓ . . . .
  - vim            ✓ ✓ ✓ . ✓
  - cpp            ✓ ✓ ✓ ✓ ✓

  Legend: H[ighlight], L[ocals], F[olds], I[ndents], In[j]ections
         +) multiple parsers found, only one will be used
         x) errors found in the query, try to run :TSUpdate {lang}
```

### Output of `nvim --version`

```text
NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by team+vim@tracker.debian.org

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
[...]

PR #3697, opened 2022-10-21
Disable folding at startup.

jhaand:
As discussed per issue https://github.com/nvim-treesitter/nvim-treesitter/issues/3642 .

This PR adds a line to README.md for the configuration. It's now possible to enable folding via nvim-treesitter but disable folding everything at startup. 
Tested with vim -d to see if folding during diff editing still works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,6 +387,7 @@ Tree-sitter based folding. *(Technically not a module because it's per windows a
 ```vim
 set foldmethod=expr
 set foldexpr=nvim_treesitter#foldexpr()
+set foldenable=false                     -- Disable folding at startup. 
 ```
 
 This will respect your `foldminlines` and `foldnestmax` settings.


theHamsta:
Thank you! @clason is this the recommended vim way? I don't know anything about these options :confused:

theHamsta (approved):

clason:
I don't use fold, but I believe the vimmy way is to set `foldlevelstart` to a very large number (e.g., `99`).

jhaand:
> I don't use fold, but I believe the vimmy way is to set `foldlevelstart` to a very large number (e.g., `99`).

I got a reply on Twitter that the foldenable method seems more clear and is the approved way. `foldlevelstart` sounds less clear to me. That's why I went with this solution.  

link: https://twitter.com/jhaand/status/1580851696418242560

jhaand:
@clason What's needed to close this PR?

theHamsta:
Thank you!

jhaand:
Thank you for finishing this.

PR #3693, opened 2022-10-20
highlights 2: electric boogaloo

ObserverOfTime:
This is a continuation of #3562 (and #3567).

The plan is to fix some inconsistencies and use many of the lesser-used captures.

I will be updating this PR until it's complete ~~(hopefully by the end of the month)~~.
Then, I will tag all the language maintainers so that changes can be upstreamed if needed.

---

- Fixes #2339 (except `<?hh`)
- Fixes #2713
- Fixes #3236
- Fixes #3431
- Fixes #3460
- Resolves #3675
- Closes #1429
- Closes #2133
- Closes #2136
- Closes #3478
- Closes #3662

---

#### Progress
* [x] c
* [x] c_sharp
* [x] comment
* [x] cpp
* [x] css
* [x] cuda
* [x] d
* [x] dart
* [x] ecma
* [x] elixir
* [x] elm
* [x] elvish
* [x] fortran
* [x] gdscript
* [x] gitattributes
* [x] gleam
* [x] glsl
* [x] go (#3154)
* [x] graphql
* [x] hack
* [x] haskell
* [x] java
* [x] julia
* [x] kotlin
* [x] lalrpop
* [x] llvm
* [x] make
* [x] meson
* [x] ocaml
* [x] pascal
* [x] php
* [x] pioasm
* [x] python
* [x] ql
* [x] qmljs
* [x] rasi
* [x] ruby
* [x] rust (#3648)
* [x] scala
* [x] scss
* [x] solidity
* [x] swift
* [x] sxhkdrc
* [x] tiger
* [x] toml
* [x] twig
* [x] typescript
* [x] vala
* [x] verilog
* [x] vim (vigoux/tree-sitter-viml#130)
* [x] wgsl
* [x] yang
* [x] zig

PR #3705, opened 2022-10-23
highlights(comment) Use @text.todo as default tag highlight

rafikdraoui:
None of the other highlight names in the query file are linked to a default highlight. By using `@text.todo` as the base highlight, we can ensure that all comment tags will fall back to be highlighted as the standard `Todo` highlight, which is more likely to be defined in color schemes than `@text.note`.

Color schemes can still define `@text.note`, `@text.warning` and `@text.danger` for more fine-grained highlighting of specific tag names.

### DIFF
--- a/queries/comment/highlights.scm
+++ b/queries/comment/highlights.scm
@@ -7,7 +7,7 @@
 
 ":" @punctuation.delimiter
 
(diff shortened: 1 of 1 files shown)


clason:
Duplicate of #3693

(But thanks!)

PR #3714, opened 2022-10-25
fix(ocaml): change query order for Variables

rbjorklin:
By moving the Variables query higher it gives the @variable capture group a lower precedence when highlighting.
This makes it possible to distinguish between @function and @variable.

Fixes tree-sitter/tree-sitter-ocaml#61.

### DIFF
--- a/queries/ocaml/highlights.scm
+++ b/queries/ocaml/highlights.scm
@@ -18,6 +18,13 @@
 
 [(constructor_name) (tag)] @constructor
 
+; Variables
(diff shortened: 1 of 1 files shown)


rbjorklin:
Pinging @undu since he's the maintainer of the OCaml queries.

undu:
Thanks for contributing, I'll test this later today

undu (approved):

clason (approved):

PR #3721, opened 2022-10-26
highlights(rust): handle non-builtin attributes

LoipesMas:
Tree-sitter-rust changed attribute parsing a bit and non-builtin attributes have different identifier.
Added them to highlight as macros.
Not sure if anything else needs to be changed as well

Related issue: https://github.com/tree-sitter/tree-sitter-rust/issues/134


clason:
I think that has been made obsolete by the latest upstream changes (which removed a lot of special parsing of attributes).

Please take a look at the latest version of parser and queries and if there is still something that needs to be done, open a new PR!

PR #3726, opened 2022-10-27
adding vhs (.tape) support

caarlos0:
[VHS](https://github.com/charmbracelet/vhs) is a tool that generates terminal gifs from code, like asciicinema, but with actual gifs as result, and instead of recoding it, you generate them from code.

`.tape` is the extension of those files.

First time doing this, not sure if I did everything right... 

- TS: https://github.com/charmbracelet/tree-sitter-vhs
- "compiler": https://github.com/charmbracelet/vhs

cc/ @maaslalani


theHamsta (approved):

caarlos0:
error seems to be unrelated:

```
Error: parser for zig is not installed
```

anyway, merged master, and updated to latest tree-sitter-vhs as well

PR #3733, opened 2022-10-28
improve PHP variables and $this highlights

webns:
This pull request changes the way of highlighting "$" sign and variable name  in PHP. In PHP all variables starts from "$" and I see no good reason why they should have different color as they do for now.

Current highlight
image

After PR
image

Maybe I'm missing something and some reasons exist for having different color?

### DIFF
--- a/queries/php/highlights.scm
+++ b/queries/php/highlights.scm
@@ -96,8 +96,8 @@
 
 (const_declaration (const_element (name) @constant))
 
-((name) @variable.builtin
- (#eq? @variable.builtin "this"))
+((variable_name) @variable.builtin
+ (#eq? @variable.builtin "$this"))
 
 ; Namespace
 (namespace_definition
@@ -142,7 +142,6 @@
 ] @keyword.function
 
 [
- "$"
  "abstract"
  "break"
  "class"


theHamsta:
@tk-shirasaka

tk-shirasaka:
Sorry for the late confirmation.
Looks good! Thanks for the update.

tk-shirasaka (commented):
Sorry for the late confirmation.
Looks good! Thanks for the update.

tk-shirasaka (approved):

theHamsta (approved):

gbprod:
Hey!
I really LOVE to have "$" a different color from variable name... so I don't like this change :(
I'm trying to override highlights to have this back but I can't manage to do it.
Can you help me ?

(IMHO, as most of IDE colors "$" differently from variable name, we should rollback this)

webns:
@gbprod Hi! Thanks for your opinion, maybe we should discuss this small change more closely and decide what is better. If everyone likes having "$" in a different color I don't mind to revert the changes.

I'm used to phpstorm's highlights where "$" is the same color as the variable name, so it looks weird to me when they are different.

For those who want to have different color, you could add this to your `.config/nvim/after/queries/php/highlights.scm ` file

```
;; extends

((variable_name) @variable.builtin
 (#eq? @variable.builtin "this"))

[
 "$"
 ] @keyword
```

gbprod:
@webns Thanks for your feedback. I've checked more IDE (I only use Neovim and sometimes vscode) and there seems to be no consensus. Some have a different color (Vim, Neovim, VsCode, Emacs), others no (PhpStorm, SublimeText). As I use (Neo)Vim for a long time, I've get used to have a different color for `$`.
 
I think both behaviors should be possible as it's a matter of personal taste.
 
As I can manage to have my blue `$` back with your query, it's ok for me :)

PR #3736, opened 2022-10-29
feat: markdown table highlights and inline

folke:
This PR adds extra markdown support:
- [x] table pipe characters are mapped to `@punctuation.special`
- [x] table headers are mapped to `@text.title`
- [x] table cells will be parsed by `markdown_inline`

image

The only thing I'm not sure of, is that the `markdown_inline` for table cells with `conceallevel=3` would now mess existing table formatting without conceal.
image

### DIFF
--- a/queries/markdown/highlights.scm
+++ b/queries/markdown/highlights.scm
@@ -19,6 +19,14 @@
   (fenced_code_block)
 ] @text.literal
 
+(pipe_table_header (pipe_table_cell) @text.title)
+[
+ (pipe_table_row)
+ (pipe_table_delimiter_row)
+ (pipe_table_header)
+] "|" @punctuation.special
+(pipe_table_delimiter_cell) @punctuation.special
+
 [
   (fenced_code_block_delimiter)
 ] @punctuation.delimiter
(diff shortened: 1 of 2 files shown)


clason:
@MDeiml

queries/markdown/highlights.scm
```diff
@@ -19,6 +19,14 @@
   (fenced_code_block)
 ] @text.literal
 
+(pipe_table_header (pipe_table_cell) @text.title)
+[
+ (pipe_table_row)
+ (pipe_table_delimiter_row)
+ (pipe_table_header)
+] "|" @punctuation.special
```
MDeiml, line 27: I think it might highlight ever "|" not just the ones in tables? Not sure though about this specific query syntax.

MDeiml (commented):
Looks good to me otherwise

queries/markdown/highlights.scm
folke: I thought so first to, but it seems to work as expected when adding a `\|` in a cell. (with or without the inline)

folke: It works, because the only other way a `|` character can be in the text that should not be a row separator it has to be `\|`, but those are properly parsed by the markdown parser. That's why it works as intended

MDeiml: If you checked it I'm happy :)

MDeiml (approved):

clason (approved):

PR #3737, opened 2022-10-29
markdown: fixup for #3736

folke:
I wan't able to make this work. I think it's not possible right now, because of `\` in `markdown` and `backslash_escape` in `markdown_inline`.

@MDeiml you were right about my other PR though.

This fixes the `|` outside tables.

### DIFF
--- a/queries/markdown/highlights.scm
+++ b/queries/markdown/highlights.scm
@@ -20,11 +20,10 @@
 ] @text.literal
 
 (pipe_table_header (pipe_table_cell) @text.title)
-[
- (pipe_table_row)
- (pipe_table_delimiter_row)
- (pipe_table_header)
-] "|" @punctuation.special
+
+(pipe_table_header "|" @punctuation.special)
+(pipe_table_row "|" @punctuation.special)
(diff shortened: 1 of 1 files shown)


clason:
@MDeiml 

(I don't think we need to use conventional commits here: just `<lang>: change` is clearer.)

MDeiml:
This change would conceal all back slashes, not only those part of escapes

folke:
> 

Yes you're right. I couldn't figure out to make this work.

Something like the below looked promising, but does not work for some reason:

```
((backslash_escape) @conceal
  (#offset! @conceal 0 0 0 -1))
```

MDeiml:
Sounds good though. This should work universally

folke:
@MDeiml just to clarify, I changed this PR from the "concealing backslash", to fixup the table pipe characters. Probbably should have closed this one and opened a new one instead 😄 

This PR makes sure only pipe characters in a table are highlighted to fix y previous PR.

theHamsta (approved):

MDeiml (approved):

PR #3741, opened 2022-10-30
highlights(r): update

aspeddro:


clason:
Test failure

tests/query/highlights/r/test.r
```diff
@@ -0,0 +1,47 @@
...
+b <- list(name = "r", version = R.version$major)
+#          ^ @parameter
+#                 ^ @string
+#                                        ^ @operator
+#                                           ^ @field
+
+Lang$new(name = "r")$print()
+#     ^ @method.call
+
+for(i in 1:10) {
```
theHamsta, line 32: maybe even also `@repeat`

theHamsta: is `in` always `@repeat` or just in this loop context?

theHamsta (approved):

PR #3745, opened 2022-10-30
fix(lockfile): remove typo entry

figsoda:


clason:
Thank you, but that file is autogenerated. Better to find out _why_ there's a duplicate entry and fix that.

figsoda:
I don't think this particular entry is autogenerated, it was introduced in https://github.com/nvim-treesitter/nvim-treesitter/pull/2549 and there are no other references to the word `embdedded_template` in the repository

clason:
Sure, let's try it and see what happens.

clason (approved):

PR #3752, opened 2022-10-31
fish: update highlight queries

taekwombo:
I'm dropping this in case upcoming [fish breaking changes](https://github.com/ram02z/tree-sitter-fish/pull/16) are merged.

### DIFF
--- a/queries/fish/highlights.scm
+++ b/queries/fish/highlights.scm
@@ -7,14 +7,21 @@
  "||"
  "|"
  "&"
- "="
- "!="
  ".."
  "!"
  (direction)
  (stream_redirect)
- (test_option)
 ] @operator
+
+
+; match operators of "test" command
+(command
+  name: (word) @function (#match? @function "^test$")
+  argument: (word) @operator (#match? @operator "^(!?=|-[a-zA-Z]+)$"))
+
+; match operators of "["
+(test_command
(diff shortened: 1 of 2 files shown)


clason:
Thanks! It would be helpful if you updated the parser while you're at it (just replace the commit id in the `lockfile.json`) -- this will prevent the upcoming automatic update from breaking and blocking other updates.

clason:
Ah, I see that your parser PR isn't merged yet. I'll mark this as draft to indicate it shouldn't be merged yet. It'd be best if you added the parser bump yourself here when the parser PR is merged.

queries/fish/highlights.scm
```diff
@@ -7,15 +7,22 @@
...
 
+
+; match operators of "test" command
+(command
+  name: (word) @function (#match? @function "^test$")
+  argument: (word) @operator (#match? @operator "^(!?=|-[a-zA-Z]+)$"))
+
+; match operators of "["
+(test_command
+  argument: (word) @operator (#match? @operator "^(!?=|-[a-zA-Z]+)$"))
```
theHamsta, line 24: `\\=` works

theHamsta:
Merged with https://github.com/nvim-treesitter/nvim-treesitter/pull/3851

PR #3754, opened 2022-11-01
test: remove bundled parsers in CI

clason:
Remove bundled parsers in CI to make sure we test queries against the parser version specified in the lockfile.

Also bump `help` and `lua` queries (which were blocked by this issue) while I'm at it.

### DIFF
--- a/queries/help/highlights.scm
+++ b/queries/help/highlights.scm
@@ -2,6 +2,8 @@
 (h2) @text.title
 (h3) @text.title
 (column_heading) @text.title
+(column_heading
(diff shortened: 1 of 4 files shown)


clason:
I give up; I don't have an Ubuntu machine to test. What we need to do is to change the installation script to extract the appimage (like we do on macOS) and then remove the bundled `usr/lib/nvim/parser` dir.

@theHamsta 

(Even better would be to try to find out why CI doesn't prioritize the plugin's parsers like Neovim does with a "proper" installation and fix that.)

PR #3768, opened 2022-11-05
hocon: update grammar

antosha417:


antosha417:
I see now that we have github-actions for this

PR #3779, opened 2022-11-06
feat(t32): initial support

xasc:
TRACE32 is a debugger for embedded software development that comes with a scripting language for automation. Here is a request to integrate the existing parser.

### DIFF
--- a/queries/t32/highlights.scm
+++ b/queries/t32/highlights.scm
@@ -1,0 +1,141 @@
+[
+  "="
+  "^^"
+  "||"
+  "&&"
+  "+"
+  "-"
+  "*"
+  "/"
+  "%"
+  "|"
+  "^"
+  "=="
+  "!="
+  ">"
+  ">="
+  "<="
+  "<"
+  "<<"
+  ">>"
+  ".."
+  "--"
+  "++"
+  "+"
+  "-"
+  "~"
+  "!"
+  "&"
+  "->"
+  "*"
+] @operator
+
+[
+ "("
+ ")"
+ "{"
+ "}"
+ "["
+ "]"
+] @punctuation.bracket
+
+[
+  ","
+  "."
+  ";"
+] @punctuation.delimiter
+
+[
+  (access_class)
+  (address)
+  (bitmask)
+  (file_handle)
+  (frequency)
+  (integer)
+  (percentage)
+  (time)
+] @number
+
+(float) @float
+
+(string) @string
+
+(path) @string.special
+
+(symbol) @symbol
+
+(character) @character
+
+; Commands
+(command_expression
+  command: (identifier) @keyword)
+(macro_definition
+  command: (identifier) @keyword)
+
+; Returns
+(
+  (command_expression
+    command: (identifier) @keyword.return)
+  (#match? @keyword.return "^[eE][nN][dD]([dD][oO])?$")
+)
+(
+  (command_expression
+    command: (identifier) @keyword.return)
+  (#match? @keyword.return "^[rR][eE][tT][uU][rR][nN]$")
+)
+
+; Subroutine calls
+(subroutine_call_expression
+  command: (identifier) @keyword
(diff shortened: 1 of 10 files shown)


queries/t32/highlights.scm
```diff
@@ -0,0 +1,134 @@
...
+  (time)
+] @number
+
+(float) @float
+
+(string) @string
+
+(path) @string.special
+
+(symbol) @string.special
```
theHamsta, line 65: Perhaps`@symbol` applies here. But I don't know the scripting language

theHamsta (approved):

queries/t32/highlights.scm
xasc: Thank you for your feedback. `(symbol)` in this context refers to a debug symbol that is part of a binary for an embedded target. Most of the time debug symbols are acting as a shorthand name for a specific memory address or address range, so symbol probably makes sense here.

On the other hand, paths are very similar to a string without any delimiters. The PRACTICE parse uses the command context to determine whether the next argument is supposed to be a file path and accepts both a string and a path.

xasc:
The latest set of changes should resolve the incorrect queries for locals. I had to introduce new node types, so the change scope got larger than what I had initially hoped for.

clason:
Tests are failing due to wrong capture mismatch with the parser.

xasc:
Thank you very much for the hint. The last grammar update was missing the updated parser sources. Both parts should be in sync now.

clason:
Unfortunately, that didn't do the trick...

clason:
(Also, just a note for future PRs -- not just to this repo: Never make a PR straight from your fork's master, but always open a feature branch and make the PR from there. Otherwise you trigger Github's paranoia and some things will require additional steps.)

xasc:
Looks like as if an older version of the grammar was getting pulled. I guess this was caused by the outdated entry in _lockfile.json_.

clason:
> Looks like as if an older version of the grammar was getting pulled. I guess this was caused by the outdated entry in _lockfile.json_.

Yes, an entry in the `lockfile.json` will mean that this exact commit is used. (It's still good practice to add the new parser with the correct hash to the initial PR to make it self-contained; otherwise we'll have to run a few automated PRs afterwards.)

xasc:
Thank you for the clarification. The grammar is now pinned to the latest compatible version.

clason:
Note that this will _not_ lock the parser for ever; there's an automated CI script that runs (at least) daily and updates the commit hashes in the lockfile to the latest HEAD. If you don't want that, you need to pin the parser (to a tag) in the `parsers.lua` entry.

clason:
@theHamsta any idea why the Windows-CL check keeps failing?

xasc:
Doesn't seem like the MSVC CI job is currently catching build errors:
```
2022-12-16T21:50:59.9826050Z cmd:  pushd C:\Users\runneradmin\AppData\Local\nvim-data\tree-sitter-t32 & gcc  -o parser.so -I./src src/parser.c src/scanner.c -shared -Os -lstdc++ & popd
2022-12-16T21:50:59.9826485Z 
2022-12-16T21:51:01.2720860Z Compiling...
2022-12-16T21:51:01.2721231Z 
2022-12-16T21:51:01.2722399Z ret:  
2022-12-16T21:51:01.2722641Z 
2022-12-16T21:51:01.2722928Z cmd.err:  Error during compilation
2022-12-16T21:51:01.2723186Z 
2022-12-16T21:51:01.2723399Z vim.v.shell_error:  0
2022-12-16T21:51:01.2723637Z 
....
2022-12-16T21:51:01.3629270Z 
2022-12-16T21:51:01.4276964Z Treesitter parser for t32 has been installed
```

The compile stepsis passed without error, but then the test step will fail, because the parser is not available.

theHamsta:
Please don't merge master into your branch! Always rebase.

xasc:
All right. Thank you very much.

PR #3791, opened 2022-11-10
Switch Erlang parser source as upstream is abandoned

filmor:


clason:
Looks like the new fork is not compatible with our queries. Can you update them?

Please be aware that this means the switch is a breaking change for downstream plugins, so it would be good to summarize the needed changes in the PR description.

filmor:
Why do you think that it's incompatible? It is itself a (squashed) fork of the previous implementation, which in the meantime only got a single grammar-related merge. The failure was probably due to me not changing the lockfile.

clason:
No, as you can see, the bundled queries do not work with the new parser. The "single grammar-related merge" must have changed something.

clason:
The grammar looks to be completely rewritten; since Facebook decided not to fork the parser but start the repo from scratch (boo!), there's nothing to it but to go through the queries one at a time to check if they need to be adapted.

Are you willing to do that?

filmor:
Yeah, I have already started but got interrupted :)

clason:
No hurry, just wanted to check what the plan is!

clason (approved):

filmor:
It's not quite ready, yet, The old queries are quite lacking and some things have changed structurally. I'll work on this a bit more :)

clason:
Note: you're only allowed to use the captures listed under https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights in queries (unless you mark them as private with a leading underscore).

filmor:
> Note: you're only allowed to use the captures listed under https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights in queries (unless you mark them as private with a leading underscore).

Ooh, thanks for that. I was looking at https://neovim.io/doc/user/treesitter.html#treesitter-highlight which I think doesn't match exactly.

clason:
>  I was looking at https://neovim.io/doc/user/treesitter.html#treesitter-highlight which I think doesn't match exactly.

No, this something that is specific to nvim-treesitter (an attempt to introduce some uniformity in the 100+ queries we have to maintain).

queries/erlang/highlights.scm
```diff
@@ -26,79 +62,68 @@
...
+;; Macros
+(macro_call_expr
+  "?" @macro
+  name: (_) @macro
+) @macro
+
+;; Preprocessor
+(pp_define) @define
+[(pp_include) (pp_include_lib)] @include
+(_preprocessor_directive) @preprocessor
```
clason, line 80:
```suggestion
(_preprocessor_directive) @preproc
```

queries/erlang/highlights.scm
```diff
@@ -26,79 +62,68 @@
...
 ] @exception
+((atom) @exception (#any-of? @exception "error" "throw" "exit"))
+
+((atom) @boolean (#any-of? @boolean "true" "false"))
+
+;; Macros
+(macro_call_expr
+  "?" @macro
+  name: (_) @macro
+) @macro
```
clason, line 75:
```suggestion
) @constant.macro
```
Not quite clear what you're trying to achieve here, though -- isn't that the same as simply
```
(macro_call_expr) @macro
```

queries/erlang/highlights.scm
```diff
@@ -26,79 +62,68 @@
...
-	;;;TODO
-	"$"
- ] @operator
+
+;; Records
+(record_field_expr _ @structure) @structure
+(record_field_name _ @structure) @structure
+(record_name name: _ @structure) @structure
+(record_decl name: _ @structure)
+(record_field name: _ @structure ty: _ @type)
```
clason, line 98:
```suggestion
(record_field_expr _ @type) @type
(record_field_name _ @type) @type
(record_name name: _ @type) @type
(record_decl name: _ @type)
(record_field name: _ @type ty: _ @type)
```
You could try `@type.structure`; CI should only check top-level captures but that has not worked before.

And again: why not capture the whole `(record_field_expr)` as `@type`?

clason (changes requested):

queries/erlang/highlights.scm
filmor: I'm still learning a bit. I think one core issue I ran into is that I tried to capture `(atom) @constant`, which seems to take priority over pretty much everything else. Also, I don't think that `@type` fits here for everything, `@field` might be more appropriate. What I want to achieve is

    A#some_record.some_field
    |- @variable
     |__________|- @type
                 |- @punctuation
                  |________|- @field

clason: > I'm still learning a bit. I think one core issue I ran into is that I tried to capture (atom) @constant, which seems to take priority over pretty much everything else.

Yes, but the full capture should still have priority, so the `@constant` is being "painted over" with `@type`. (Order matters here.)

> Also, I don't think that @type fits here for everything, @field might be more appropriate. What I want to achieve is

Ah, that's different; I was just referring to using the same capture inside, which should be unnecessary. `@type` is used for `structs`, which as far as I can tell is what `records` are analogues of, so they should apply to `some_record`. `some_field` is indeed different, and `@field` sounds correct for that.

queries/erlang/highlights.scm
```diff
@@ -26,79 +62,68 @@
...
+  "try"
 ] @exception
+((atom) @exception (#any-of? @exception "error" "throw" "exit"))
+
+((atom) @boolean (#any-of? @boolean "true" "false"))
+
+;; Macros
+(macro_call_expr
+  "?" @macro
+  name: (_) @macro
```
clason, line 74: these will still fail.

queries/erlang/highlights.scm
```diff
@@ -26,79 +62,68 @@
...
-  name: (atom) @function)
-;;;lambda
-(lambda_clause
-  arguments:
-    (pattern) @variable
-)
+(var) @variable
+(atom) @constant
+(integer) @number
+(float) @float
```
clason, line 129: Move these up to the top, so the more specific captures override them.

filmor: That doesn't seem to be enough, I need to also change their priority. It looks to me a bit like the highlighting prioritises more specific matches.

clason: Then I'd recommend leaving out the `(atom) @constant` capture, or at least make it much more specific (if that is possible). Otherwise it's easier to leave the other queries as they are for now (but of course changing the capture name as discussed).

lua/nvim-treesitter/parsers.lua
```diff
@@ -412,7 +412,7 @@ list.scss = {
 
 list.erlang = {
   install_info = {
-    url = "https://github.com/AbstractMachinesLab/tree-sitter-erlang",
+    url = "https://github.com/WhatsApp/tree-sitter-erlang",
     files = { "src/parser.c" },
     branch = "main",
   },
```
ObserverOfTime, line 418: You should probably change the maintainer to yourself.

clason:
@filmor friendly ping -- the `@structure` captures need to be fixed (see https://github.com/nvim-treesitter/nvim-treesitter/pull/3791#discussion_r1027125304), then we can merge this.

queries/erlang/highlights.scm
clason: No, it's not a capture upstream, it's a highlight group -- that's different. (And that will likely be changed to `@struct`, which is what LSP semantic tokens use.)

filmor:
Just fyi, I'll work a more on the highlightling, it's currently a little worse than it was before (though, the new grammar at least handles all my files correctly).

clason:
That's fine, thanks for tackling this necessary change!

PR #3778, opened 2022-11-06
highlights(ecma): distinguish between "default" in export and switch

Trard:
`default` keyword is used in [switch statements](https://262.ecma-international.org/12.0/#prod-DefaultClause) and [export declarations](https://262.ecma-international.org/12.0/#prod-ExportDeclaration).

In switch statements it must be highlighted as `@conditional`.
In export declarations it must be highlighted as `@keyword`.

PR #3792, opened 2022-11-10
highlights(ecma): remove `switch` from `@keyword`

Trard:
`switch` must be removed from `@keyword` because it's already highlighted as `@conditional`
 
 ---
 
> In your screen shot it looks like `switch` is not highlighted as `@conditional`. Maybe that could be adapted as well...

It's `@keyword` and `@conditional` at same time.
image

I can create another PR to remove it from keywords

_Originally posted by @Trard in https://github.com/nvim-treesitter/nvim-treesitter/issues/3778#issuecomment-1310383801_

 ---

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -239,7 +239,6 @@
 "let"
 "set"
 "static"
-"switch"
 "target"
 "typeof"
 "var"


vigoux (approved):
It seems to be more consistent with the other languages, thanks for the contribution.

PR #3798, opened 2022-11-12
fix(astro): make the script element use typescript instead of javascript

virchau13:
The Astro parser should use the typescript parser for the `<script>` element instead of the javascript parser as defined by `; inherits: html`. This overrides it.

Note: Is this the correct way to do things? Does this completely override the HTML `javascript` injection or will this break at some point?

### DIFF
--- a/queries/astro/injections.scm
+++ b/queries/astro/injections.scm
@@ -5,3 +5,6 @@
 
 ((interpolation
     (raw_text) @tsx))
+
+((script_element
+    (raw_text) @typescript))


virchau13:
Hey this is kind of important for Astro users to work properly with Typescript, is there anything blocking the merge of this PR? I'll be happy to submit code PRs if necessary

clason (approved):

clason:
As you are the astro maintainer, I trust you on this. If it breaks, I'll ping you :)

PR #3804, opened 2022-11-14
fix(installer): ignore globally installed parsers

kylo252:
### Problem

`is_installed` is picking up parsers shipped by neovim v0.8 and these parsers have unknown version
and compatability.

_Note: this applies to any parser found in the rtp_

https://github.com/nvim-treesitter/nvim-treesitter/blob/2072692aaa4b6da7c354e66c2caf4b0a8f736858/lua/nvim-treesitter/install.lua#L94-L98

### Solution

only consider a parser installed if ~~its revision is available~~ it's available in the `parsers_install_dir`.

_Edit: the revision file can still remain after uninstalling a parser, so it's not reliable heuristic!_

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -91,10 +91,18 @@ local function get_installed_revision(lang)
   end
 end
 
+---Checks if parser is installed with nvim-treesitter
 ---@param lang string
 ---@return boolean
 local function is_installed(lang)
-  return #api.nvim_get_runtime_file("parser/" .. lang .. ".so", false) > 0
+  local matched_parsers = vim.api.nvim_get_runtime_file("parser/" .. lang .. ".so", true) or {}
+  for _, path in ipairs(matched_parsers) do
+    local install_dir = configs.get_parser_install_dir()
+    if vim.startswith(path, install_dir) then
+      return true
+    end
+  end
+  return false
 end
 
 ---@param lang string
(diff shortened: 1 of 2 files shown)


theHamsta (approved):

lua/nvim-treesitter/install.lua
```diff
@@ -91,10 +91,18 @@ local function get_installed_revision(lang)
...
 local function is_installed(lang)
-  return #api.nvim_get_runtime_file("parser/" .. lang .. ".so", false) > 0
+  local matched_parsers = vim.api.nvim_get_runtime_file("parser/" .. lang .. ".so", true) or {}
+  for _, path in ipairs(matched_parsers) do
+    local install_dir = configs.get_parser_install_dir()
+    if vim.startswith(path, install_dir) then
+      return true
+    end
+  end
+  return false
```
theHamsta, line 105: so `ensure_installed = "c"` will not be satisfied with the built-in parser. I suppose this is ok

theHamsta:
Thank you!

@clason @ObserverOfTime are you ok with this change or should the function be renamed to make it's meaning cleared (e.g. `is_installed_by_nvim_treesitter`). Feel free to merge if PR is ok.

folke:
Added a new PR that also checks for installed when updating.
Otherwise nvim-treesitter always installs the Neovim included parsers when doing `TSUpdate`

https://github.com/nvim-treesitter/nvim-treesitter/pull/3862

PR #3805, opened 2022-11-14
Updated rego scm

FallenAngel97:
Hello!

Based on @shaded-enmity work, I copied everything here from  https://github.com/FallenAngel97/tree-sitter-rego/pull/2

Suggestions are welcome

### DIFF
--- a/queries/rego/highlights.scm
+++ b/queries/rego/highlights.scm
@@ -1,16 +1,64 @@
 ; highlights.scm
-"import" @include
-"package" @include
+[
+  (import) 
+  (package)
+] @module
 
+[
+  (with)
+  (as)
+  (every)
+  (some)
+  (in)
+  (not)
+  (if)
+  (contains)
+  (else)
+  (default)
+  "null"
+] @keyword
 
-(reserved_keywords) @keyword
+[
+  "true"
+  "false"
+] @boolean
+
+[
+  (assignment_operator)
+  (bool_operator)
+  (arith_operator)
+  (bin_operator)
+] @operator
+
+[
+  (string)
+  (raw_string)
+] @string
+
+(term (ref (var))) @variable
+
 (comment) @comment
-(rego_block rego_rule_name: (identifier) @function)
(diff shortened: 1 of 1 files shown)


queries/rego/highlights.scm
```diff
@@ -1,16 +1,64 @@
 ; highlights.scm
-"import" @include
-"package" @include
+[
+  (import) 
+  (package)
+] @module
```
krzykli, line 5: I think `@include` fits this one better

queries/rego/highlights.scm
```diff
@@ -1,16 +1,64 @@
...
+[
+  (open_paren)
+  (close_paren)
+  (open_bracket)
+  (close_bracket)
+  (open_curly)
+  (close_curly)
+] @punctuation.bracket
+
+(rule (rule_head (var) @attribute))
```
krzykli, line 59: In my tests `@method` highlight looked a bit better than `@attribute` for rule heads

queries/rego/highlights.scm
FallenAngel97: done

queries/rego/highlights.scm
FallenAngel97: done

clason:
You need to update the parser commit in the `lockfile.json` in this PR to make CI pass.

FallenAngel97:
Thanks, @clason !
Updated

clason:
No, that's not the error on CI: it's complaining about 
```
 Invalid capture @output-var in highlights for rego.
```

You may only use the captures that are explicitly listed on https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md.

FallenAngel97:
Thank you. It seems, now it's a valid one

krzykli (approved):
I think this is good for the time being.
There might be some tweaks in the near future but getting the mostly working grammar updated is a priority.

clason (approved):

PR #3818, opened 2022-11-18
update R queries

kevinushey:
Related to recent updates to the R tree-sitter grammar: https://github.com/nvim-treesitter/nvim-treesitter/pull/3817

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -288,7 +288,7 @@
     "revision": "0695cd0760532de7b54f23c667d459b5d1332b44"
   },
   "r": {
-    "revision": "80efda55672d1293aa738f956c7ae384ecdc31b4"
+    "revision": "3bc2af26b7c077f57fff69fccef19579896fa6fe"
   },
   "racket": {
     "revision": "09cb27a06415bce529a26774a842f5a80d50d362"
--- a/queries/r/highlights.scm
+++ b/queries/r/highlights.scm
@@ -59,7 +59,12 @@
   (NA) (NA_integer_) (NA_real_) (NA_complex_) (NA_character_)
 ] @constant.builtin
 
+; Namespaces
+("::"  lhs: (identifier)) @namespace
+(":::" lhs: (identifier)) @namespace
+
 ; Calls
+(call (identifier) @function)
 
 ; Error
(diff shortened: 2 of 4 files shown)


clason:
Indentation test for R fail now. Could you take a look at it?

kevinushey:
I see this when I try to run tests locally:

```
$ ./scripts/run_tests.sh
E492: Not an editor command: PlenaryBustedDirectory tests/ { minimal_init = './scripts/minimal_init.lua' }
```

What's the right way for me to get `plenary.vim` installed for local testing?

clason:
I can't help more than pointing to the test scripts (I don't think it matters as long as `plenary.nvim` is in runtimepath).

I personally just watch CI and then manually execute whatever the failing test does.

kevinushey:
Looks like it gets installed on CI here:

https://github.com/kevinushey/nvim-treesitter/blob/6bf1950ada77bd411163ca071a46bb3552c3e1fa/.github/workflows/tests.yml#L33-L39

so I guess I have to run that locally first. Will give it a shot...

clason:
You can use the trick from here in your local copy of the minimal init.lua (just replace packer by plenary):
https://github.com/nvim-lua/kickstart.nvim/blob/c8a140577832ed4f958f4964400b2591abea3825/init.lua#L1-L8

kevinushey:
I can't figure out how to get neovim to use the local indentation query files when running tests. :-/

clason:
just copy them to your local `.config/nvim/queries` directory

kevinushey:
Doesn't work for me.

https://neovim.io/doc/user/treesitter.html#treesitter-query states that it looks for queries using `runtimepath`, but even with me making sure `"."` comes first in `runtimepath` the queries bundled in the repository aren't used.

I don't have the bandwidth to debug or contribute further so I'm just going to roll back the changes in the R grammar.

clason:
Yeah, we really need a `make test ` to make it easier to test non-highlighting queries. I'll try to sort the indents out this weekend.

clason:
Looks like you indeed reverted the grammar improvements, so this is moot for now. Feel free to reopen when you re-add (some of) them!

PR #3839, opened 2022-11-21
feat: add smali

ObserverOfTime:
I forked the grammar to fix an issue but I don't plan on maintaining the fork.
If the author ever merges my patch, I will switch to the original repository.

### DIFF
--- a/queries/smali/highlights.scm
+++ b/queries/smali/highlights.scm
@@ -1,0 +1,113 @@
+; Types
+
+(class_identifier) @type
+
+(primitive_type) @type.builtin
+
+(array_type
+  "[" @punctuation.bracket)
+
+; Methods
+
+(method_declaration
+  (method_identifier) @method)
+
+(statement
+  (opcode) @_invoke
+  (full_method_identifier
+    (method_identifier) @method.call)
+  (#lua-match? @_invoke "^invoke"))
+
+(method_identifier
+  "<init>(" @constructor
+  ")" @constructor)
+
+(method_identifier
+  "<clinit>(" @constructor
+  ")" @constructor)
+
+; Fields
+
+(field_identifier) @field
+
+; Parameters
+
+(parameter) @parameter
+
+; Variables
+
+(variable) @variable
+
+; Labels
+
+(label) @label
+
+; Operators
+
+(opcode) @keyword.operator
+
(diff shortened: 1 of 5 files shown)


clason:
What's the upstream then?

ObserverOfTime:
https://git.sr.ht/~yotam/tree-sitter-smali

clason:
And is the issue so bad that we can't use the original? Many parsers have bugs and pending PRs, and yet we use them. (Switching forks is not painless, so I prefer not to do it if it's not permanent and a significant improvement.)

ObserverOfTime:
The original doesn't support character literals.

Guess I'll just remind the author to merge my patch.

clason:
That would be preferable, yes. Let's see what happens.

queries/smali/highlights.scm
```diff
@@ -0,0 +1,110 @@
...
+(array_type
+  "[" @punctuation.bracket)
+
+; Methods
+
+(method_declaration
+  (method_identifier) @function)
+
+((statement) @function.call
+ (#lua-match? @function.call "^invoke"))
```
theHamsta, line 16: isn't `invoke` a keyword? Could be `@keyword.function` although `@keyword.function` was only used until now for keywords introducing a function definition. `@keyword` would probably more general

ObserverOfTime: This doesn't apply to `invoke-xyz` but to the function it calls.
I guess I could specifically target the function node though.

theHamsta:
@ObserverOfTime is this still a draft? Are you planning to continue working on this?

ObserverOfTime:
I'm waiting in case the author decides to merge my patch.

amaanq:
The author's last commit was 11mo ago...so it could be a while if not never. I'd love to be able to have smali support as soon as I can :)

theHamsta:
@amaanq do you want to open a new PR? I will close this one then you can create a new one.

amaanq:
> @amaanq do you want to open a new PR? I will close this one then you can create a new one. Do you plan to maintain a fork?

Sure, I'd like to maintain a fork of the upstream repo actually...although I am not very familiar with making treesitter parsers and scheme. Hopefully it isn't too bad 😆 I'll open a new PR when ready.

PR #3848, opened 2022-11-22
chore: print OS version on health report

theHamsta:
Often we get issue report without knowing which version of Windows or Linux or MacOS they are running.

E.g.
```

Installation ~
- OK `tree-sitter` found 0.20.7 (870fb8772f3e47b5aef4c7000a78d61f3aa2b005) (parser generator, only needed for :TSInstallFromGrammar)
- OK `node` found v17.9.1 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `/usr/lib/ccache/clang-16` executable found. Selected from { "/usr/lib/ccache/clang-16", "cc", "gcc", "clang", "cl", "zig" }
  Version: Ubuntu clang version 16.0.0 (++20221123042319+910204cfbdf3-1~exp1~20221123042440.568)

OS Info:
{
  machine = "x86_64",
  release = "5.15.0-47-generic",
  sysname = "Linux",
  version = "#51-Ubuntu SMP Thu Aug 11 07:51:15 UTC 2022"
} ~
```

### DIFF
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -94,6 +94,8 @@ local function install_health()
       )
     end
   end
+
+  health.report_start("OS Info:\n" .. vim.inspect(vim.loop.os_uname()))
 end
 
 local function query_status(lang, query_group)


lua/nvim-treesitter/health.lua
```diff
@@ -94,6 +94,9 @@ local function install_health()
       )
     end
   end
+
+  local systeminfo = vim.fn.has "win32" == 1 and 'systeminfo | findstr /B /C:"OS Name" /C:"OS Version"' or "uname -a"
```
theHamsta, line 98: this command takes a while to run on Windows. `cmd /c ver` is faster but will show `Windows 10.0.22000` for Windows 11 (we probably just want to know that it is Windows, maybe build version)

theHamsta: better to use `vim.loop.os_uname()` https://github.com/neovim/neovim/pull/21157

PR #3852, opened 2022-11-24
highlights(typescript): Add satisfies keyword

Shraymonks:
Support the new [satisfies operator](https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#the-satisfies-operator) added in TypeScript 4.9

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -375,7 +375,7 @@
     "revision": "8bd2056818b21860e3d756b5a58c4f6e05fb744e"
   },
   "tsx": {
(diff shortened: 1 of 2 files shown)


clason:
@steelsojka

clason:
Does the parser support this operator? It doesn't look like it.

Shraymonks:
I have a PR out to add support for it: https://github.com/tree-sitter/tree-sitter-typescript/pull/228. I guess that needs to be merged first and the lockfile should be updated.

clason:
Yes, of course, can't use a commit before it's merged ;)

PR #3861, opened 2022-11-24
highlights(typescript): Fix as highlight

Shraymonks:
`as` is currently always highlighted as `include`:
https://github.com/nvim-treesitter/nvim-treesitter/blob/1b3f93dcb90e1c282fe9259e5a15f34d504fd5a1/queries/ecma/highlights.scm#L213-L217

Adds a query for highlighting as `keyword` in `as_expression` type assertions.

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -213,8 +213,12 @@
 [
 "import"
 "from"
-"as"
 ] @include
+
+(export_specifier "as" @include)
+(import_specifier "as" @include)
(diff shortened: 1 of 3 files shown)


theHamsta:
Don't you want to make the `@include` query for `"as"` more specific? It would still keep the `@include` highlight in a `as_expression` just with `@keyword` in addition

theHamsta (approved):

theHamsta:
Thank you!

PR #3862, opened 2022-11-24
fix(installer): don't update parsers that are not installed

folke:
When installing nvim-treesitter without any config and then
running `TSUpdate`, the global parsers will be installed.

This PR fixes that

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -115,7 +115,7 @@ end
 ---@return table
 local function outdated_parsers()
   return vim.tbl_filter(function(lang)
-    return needs_update(lang)
+    return is_installed(lang) and needs_update(lang)
   end, info.installed_parsers())
 end
 


clason (approved):

folke:
Migt be better to move the check up to info.installed_parsers, but didnt want to make too many changes.
This at least fixes updates

liskin:
But why? nvim-treesitter will happily use those global parsers, and if their revision doesn't match what its queries expect, then it leads to failures like https://github.com/nvim-treesitter/nvim-treesitter/issues/5536#issuecomment-1803645369.
Either a) nvim-treesitter should only use its own parsers, or b) `:TSUpdate` should update all parsers that it would potentially end up using. Perhaps a) was the case when this change was made, but it certainly isn't the case now (that would make [the second bullet point here](https://github.com/nvim-treesitter/nvim-treesitter/tree/8996612bfbebe1657e1bb55a95201c96cab945c6#i-get-query-error-invalid-node-type-at-position) irrelevant).

If b) is the way to go (and I believe it should be), can we please revert this?

liskin:
> The whole functionality has already been rewritten on the main branch.

Is that going to be merged any time soon? Doesn't seem to be compatible with stable neovim releases (https://github.com/neovim/neovim/commit/e3e6fadfd82861471c32fdcabe00bbef3de84563 isn't in neovim 0.9) so presumably it'll be a while. Therefore I believe we need a fix for the master branch now.

PR #3866, opened 2022-11-24
ci: fix check for valid capture

theHamsta:
Arguments to vim.startswith where swapped. Prefix has to go second.

:man_facepalming: :man_facepalming:


theHamsta:
This blocks a few PRs

theHamsta:
When #3860 passes it should be fine

theHamsta:
@ObserverOfTime  Just for info we wanted to just check that the prefix of the capture is documented. Languages can subscope arbitrarily from that

ObserverOfTime (approved):

PR #3881, opened 2022-11-27
injections: add more regex injections

ObserverOfTime:
I added a missing node to the `string.regex` highlights for Perl but didn't add the injections because the patterns might include expressions which should be interpolated and highlighted by the grammar (though they currently aren't).

### DIFF
--- a/queries/julia/injections.scm
+++ b/queries/julia/injections.scm
@@ -6,3 +6,8 @@
   (line_comment)
   (block_comment)
 ] @comment
+
+((prefixed_string_literal
+   prefix: (identifier) @_prefix) @regex
+ (#eq? @_prefix "r")
+ (#offset! @regex 0 2 0 -1))
--- a/queries/julia/injections.scm
+++ b/queries/julia/injections.scm
@@ -1,6 +1,6 @@
-; TODO: re-add when markdown is added.
-; ((triple_string) @markdown
-;   (#offset! @markdown 0 3 0 -3))
+((string_literal) @markdown
+ (#match? @markdown "^\"\"\"")
+ (#offset! @markdown 0 3 0 -3))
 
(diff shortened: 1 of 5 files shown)


queries/julia/injections.scm
```diff
@@ -1,8 +1,13 @@
-; TODO: re-add when markdown is added.
-; ((triple_string) @markdown
-;   (#offset! @markdown 0 3 0 -3))
+((string_literal) @markdown
+ (#match? @markdown "^\"\"\"")
+ (#offset! @markdown 0 3 0 -3))
```
theHamsta, line 3: This is unrelated to regexes

ObserverOfTime: True, but I noticed the todo and decided to fix it. I'll split the commit.

theHamsta: It should probably work fine now

theHamsta (approved):

queries/julia/injections.scm
```diff
@@ -1,8 +1,13 @@
...
 
 [
   (line_comment)
   (block_comment)
 ] @comment
+
+((prefixed_string_literal
```
theHamsta, line 10: Should regex really be injected into all raw-strings?

ObserverOfTime: `typeof(r"foo")` is `Regex` according to the [manual](https://docs.julialang.org/en/v1/manual/strings/#man-regex-literals).

clason: @theHamsta look further below: it only injects into strings prefixed with `r`

theHamsta: Sorry, for the confusion. I thought it would be similar to Python r-strings

PR #3882, opened 2022-11-27
feat: c# required keyword

iabdelkareem:
With C# 11 that was recently released the [`required`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/required) keyword was introduced to the language. The keyword is to be used on properties like the example below.

This PR will add it to the keywords list to fix the broked highlighting

```csharp
public class Person
{
    public required string FirstName { get; init; }
    public required string LastName { get; init; }
}
```

### DIFF
--- a/queries/c_sharp/highlights.scm
+++ b/queries/c_sharp/highlights.scm
@@ -364,6 +364,7 @@
  "readonly"
  "static"
  "volatile"
+ "required"
 ] @storageclass
 
 [


theHamsta (approved):

theHamsta:
@Luxed

queries/c_sharp/highlights.scm
```diff
@@ -364,6 +364,7 @@
  "readonly"
  "static"
  "volatile"
+ "required"
```
ObserverOfTime, line 367: I think this should be `@type.qualifier`.

queries/c_sharp/highlights.scm
```diff
@@ -364,6 +364,7 @@
...
 ] @storageclass
```
ObserverOfTime, line 368: I think this should be `@type.qualifier`.

theHamsta: this is why I left this PR open. For me there is no obvious distinction between those two highlight groups.

iabdelkareem: In this case, do you guys think we should move `const` & `readonly` from `@storageclass` to `@type.qualifier` as well ?

Luxed: I agree with both of you. `@type.qualifier` sounds better than `@storageclass`

Luxed: I looked into the C# documentation because I didn't yet know of this keyword.
Turns out they classify it the same as the `static` keyword. So making it a `@storageclass` makes the most sense to be consistent.
So I am changing my mind. `@storageclass` makes the most sense, So let's keep it like that for now 😛.
Here are the links for reference: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/required and https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/static.

PR #3874, opened 2022-11-25
Highlight pointer asterisks as @type fix #3841

mfc73:
Fixes #3841

PR #3883, opened 2022-11-27
C, Cpp, highlight little improvements

aaronrancsik:
- make `primitive_type`  `@type.builtin` 
- make `auto` `@type.builtin`

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -130,7 +130,6 @@
 
 [
  (type_identifier)
- (primitive_type)
  (sized_type_specifier)
  (type_descriptor)
 ] @type
@@ -141,6 +140,8 @@
 
 (type_definition
   declarator: (type_identifier) @type.definition)
+
+(primitive_type) @type.builtin
 
 ((identifier) @constant
  (#lua-match? @constant "^[A-Z][A-Z0-9_]+$"))
--- a/queries/cpp/highlights.scm
+++ b/queries/cpp/highlights.scm
@@ -5,6 +5,7 @@
 
 (parameter_declaration
   declarator: (reference_declarator) @parameter)
+
 ; function(Foo ...foo)
 (variadic_parameter_declaration
   declarator: (variadic_declarator
@@ -38,6 +39,8 @@
 
 (alias_declaration
   name: (type_identifier) @type.definition)
+
+(auto) @type.builtin
 
 (namespace_identifier) @namespace
 ((namespace_identifier) @type
@@ -143,7 +146,6 @@
  "co_await"
  "concept"
  "requires"
- (auto)
 ] @keyword
 
 [
--- a/tests/query/highlights/cpp/test.cpp
+++ b/tests/query/highlights/cpp/test.cpp
@@ -5,8 +5,8 @@
 
 auto main( int argc, char** argv ) -> int
       //       ^ parameter
-      //    ^ type
-      //    ^ type
+      //    ^ type.builtin
+      //    ^ type.builtin
       //                  ^ operator
 {
     std::cout << "Hello world!" << std::endl;
--- a/tests/query/highlights/cpp/test.cpp
+++ b/tests/query/highlights/cpp/test.cpp
@@ -4,6 +4,7 @@
 //       ^ string
 
 auto main( int argc, char** argv ) -> int
+// ^ type.builtin
       //       ^ parameter
       //    ^ type.builtin
       //    ^ type.builtin


clason:
Honestly? "More like VS Code" is not a compelling argument here. Can you explain how these changes are actually better?

queries/c/highlights.scm
```diff
@@ -142,6 +141,10 @@
 (type_definition
   declarator: (type_identifier) @type.definition)
 
+(primitive_type) @type.builtin
```
aaronrancsik, line 144: Primitive types not just simple 'type', with 'builtin' hl the color schemes can differentiate that see attached image blue return types in function.

ObserverOfTime:
I agree with `primitive_type` and `auto` being `type.builtin` but disagree with the rest.

queries/c/highlights.scm
```diff
@@ -130,7 +130,6 @@
 
 [
  (type_identifier)
- (primitive_type)
```
aaronrancsik, line 133: primitive type has it's own hl group so no need to be here

queries/c/highlights.scm
```diff
@@ -177,7 +180,11 @@
   declarator: (identifier) @parameter)
 
 (parameter_declaration
-  declarator: (pointer_declarator) @parameter)
+  declarator: (pointer_declarator) @pointer.declarator (#set! "priority" 101))
```
aaronrancsik, line 183: make the c pointer declarator `*` (star) character a new hl group to make it possible to color it (blue in example image)
Without priority the `operator` hl group would override it.

queries/c/highlights.scm
```diff
@@ -177,7 +180,11 @@
...
+
```
aaronrancsik, line 184: make the pointer parameter(name)  color with `@parameter` hl group priority to overcome the pointer declarator..

queries/cpp/highlights.scm
```diff
@@ -4,7 +4,12 @@
  (#match? @field "(^_|^m_|_$)"))
 
 (parameter_declaration
-  declarator: (reference_declarator) @parameter)
+  declarator: (reference_declarator) @reference.declarator (#set! "priority" 101)) ; priority to override operator
```
aaronrancsik, line 7: same as with c pointer to cpp reference.

queries/cpp/highlights.scm
```diff
@@ -39,6 +44,8 @@
 (alias_declaration
   name: (type_identifier) @type.definition)
 
+(auto) @type.builtin
```
aaronrancsik, line 47: auto is a builtin language tool to command compiler to decide the type so it's I'm open to create it's own hl group like @auto because it's a bit complicated to make it's own color in color scheme.

queries/cpp/highlights.scm
```diff
@@ -70,6 +77,12 @@
 "operator" @function
 "static_assert" @function.builtin
 
+[
+ "<<"
```
aaronrancsik, line 81: In cpp these operators often overwritten and used like functions.

queries/c/highlights.scm
```diff
@@ -142,6 +141,10 @@
 (type_definition
   declarator: (type_identifier) @type.definition)
 
+(primitive_type) @type.builtin
+
+((type_qualifier) @type.builtin (#set! "priority" 101))
```
aaronrancsik, line 146: It's unnecessary I will remove it.

aaronrancsik:
@clason @ObserverOfTime  I've tried to explain it as @clason asked.

ObserverOfTime:
My statement stands.

aaronrancsik:
> My statement stands.

I see, thank for the feedback. Le me ask, without new hl groups  how can a color scheme select the pointer decelerators `*` for example?

clason:
You can always add your own queries to your config. The question here is about the best default experience, which this is going a bit far for.

aaronrancsik:
I see, thanks!
Honestly I didn't know that I can modify these with config :facepalm: (I've kept my fork for a year now)
So should I modify this PR to only the  auto and primitive type changes?

queries/c/highlights.scm
theHamsta: there's a separate PR to highlight it as `@type`   #3874 . If you're introducing something new it should be a sub-scope of an existing category (e.g. `@operator.pointerdeclaration`)

queries/cpp/highlights.scm
theHamsta: again it is not worth to create a new highlight category out of the nothing. The subscoping mechanism is thought to have as few groups as possible on the top-level and go more specific in the subscope. In each case, the reference declarator and pointer declarator should go into the same category. Such a change should probably be then applied to Rust and go also. So it might be worth to discuss this topic in a separate PR. In each case it should be in sync with #3874

queries/cpp/highlights.scm
theHamsta: So what? What has this to do with highlighting? I would expect that all operator symbols would just have the `@operator` highlighting. It shouldn't matter how they are used in that specific instance.

theHamsta:
You would need to update the C++ test files for your changes (`@type` -> `@type.builtin` where appropriate)

ObserverOfTime (approved):

theHamsta (approved):

PR #3884, opened 2022-11-27
fix: is_installed correctly detects installed parsers

stevearc:
`is_installed` sometimes returns false for parsers that are already installed. This can happen when the detected install_dir is an absolute path and the runtime `.so` files are relative paths, or vice-versa.

I've changed the logic to always convert the paths to absolute paths before doing the prefix comparison. I've also added a guard for the case where the install_dir is nil.


theHamsta (approved):
Thank you for fixing this!

PR #3889, opened 2022-11-28
injections(go): inject regex into some regexp functions

theHamsta:


queries/go/injections.scm
```diff
@@ -1 +1,14 @@
...
+(call_expression
+  (selector_expression) @_function (#any-of? @_function
+                                    "regexp.Match"
+                                    "regexp.MatchReader"
+                                    "regexp.MatchString"
+                                    "regexp.Compile"
+                                    "regexp.CompilePOSIX"
+                                    "regexp.MustCompile"
+                                    "regexp.MustCompilePOSIX"
+                                    "regexp.Match")
```
ObserverOfTime, line 12: `regexp.Match` is duplicated.

ObserverOfTime (changes requested):

ObserverOfTime (approved):

PR #3842, opened 2022-11-21
improve highlights and injections for Nix

figsoda:
- highlight parameters
- add injections for code blocks (phases, runCommand*, writers) based on https://github.com/cstrahan/tree-sitter-nix/pull/31 by @nrdxp
- add injections for `builtins.match` (regex)

PR #3890, opened 2022-11-29
improve nix highlights and injections

figsoda:
follow up to #3842, now that I'm more familiar with this

### DIFF
--- a/queries/nix/highlights.scm
+++ b/queries/nix/highlights.scm
@@ -73,114 +73,31 @@
 ; basic identifiers
 (variable_expression) @variable
 
-; builtin functions
-; modified from `nix __dump-builtins | jq keys`
-((identifier) @function.builtin (#any-of? @function.builtin
-  "add"
-  "all"
-  "any"
-  "attrNames"
-  "attrValues"
-  "baseNameOf"
-  "bitAnd"
-  "bitOr"
-  "bitXor"
-  "break"
-  "catAttrs"
-  "ceil"
-  "compareVersions"
-  "concatLists"
-  "concatMap"
-  "concatStringsSep"
-  "deepSeq"
-  "dirOf"
-  "div"
-  "elem"
-  "elemAt"
-  "fetchClosure"
-  "fetchGit"
-  "fetchTarball"
-  "fetchurl"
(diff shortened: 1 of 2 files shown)


queries/nix/highlights.scm
```diff
@@ -62,33 +65,122 @@
...
-; import
-((identifier) @_i (#eq? @_i "import")) @include
-
-; null
-((identifier) @_i (#eq? @_i "import")) @constant.builtin
-
-; these are technically functions but they act more like keywords (abort and throw are control flow, derivation is a core language construct)
-((identifier) @_i (#match? @_i "^(abort|derivation|throw)$")) @keyword
+; modified from `nix __dump-builtins | jq keys`
+((identifier) @function.builtin (#any-of? @function.builtin
```
figsoda, line 78: if this is too much I can just keep the ones that are not directly available without `builtins`, which is like 1/4 of those

rummik: Technically the remaining ones are available with a `__` prefix iirc.  ex: `builtins.readDir` is accessible via `__readDir`

figsoda: I added all the builtins/primops I was able to find

queries/nix/highlights.scm
```diff
@@ -1,22 +1,26 @@
...
+  "in"
+  "inherit"
+  "let"
+  "rec"
+  "with"
 ] @keyword
 
+; these are technically functions but they act more like keywords (abort and throw are control flow, derivation is a core language construct)
+((identifier) @keyword
+  (#any-of? @keyword "abort" "derivation" "throw"))
```
theHamsta, line 13: abort/throw `@expection`?

theHamsta (commented):
This looks overall good! Please see my small comment.

queries/nix/highlights.scm
figsoda: fixed it

figsoda:
@theHamsta do you mind giving this another review? Thanks!

theHamsta:
Thank you!

theHamsta (approved):

PR #3901, opened 2022-11-30
feat(twig): display twig errors

gbprod:


clason:
I'm not so sure we want this -- this displays _all_ parsing errors, including while typing!

(Neovim core doesn't highlight `@error` captures by default for this very reason, but some colorschemes foolishly do.)

gbprod:
Ok! Thanks for your feedback!
I close

clason:
(although other languages have those as well, so it might be OK -- @theHamsta ?)

gbprod:
I think gitcommit have this...

PR #3909, opened 2022-12-02
feat: added highlights for markdown checkboxes

folke:
See example below. Before they did not have highlights.

image

### DIFF
--- a/queries/markdown/highlights.scm
+++ b/queries/markdown/highlights.scm
@@ -49,6 +49,10 @@
   (thematic_break)
 ] @punctuation.special
 
+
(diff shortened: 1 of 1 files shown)


clason (approved):

clason:
Be aware that this falls back to `@text.todo` if the colorscheme does not define `@text.todo.checked` etc., which might look weird.

folke:
totally, but then they just will need to update their clorscheme then and have nice looking checkboxes ;)

clason:
> totally, but then they just will need to update their clorscheme then and have nice looking checkboxes ;)

Oh you sweet summer child...

PR #3917, opened 2022-12-03
feat(hcl,terraform): split terraform from hcl, add specialized queries for tf

MichaHoffmann:
This enables us to have different queries for
terraform and general hcl. It solve the situation
where other dialects of hcl (nomad, packer, etc)
might use a terraform keyword and get highlighted
out of place.


MichaHoffmann:
Changes in highlighting make template interpolations a bit easier to see; also a few false positives are fixed.

Before:
image

After:
image

theHamsta (approved):
You chose to fork terraform from hcl in your parser with a fork of the hcl grammar. Would it make sense to also switch to your hcl parser? Wouldn't it be easier to just overwrite rules of the upstream parser and by that automatically receiving updates from upstream or do you prefer to maintain all possible dialects in one repo?

MichaHoffmann:
> You chose to fork terraform from hcl in your parser with a fork of the hcl grammar. Would it make sense to also switch to your hcl parser? Wouldn't it be easier to just overwrite rules of the upstream parser and by that automatically receiving updates from upstream or do you prefer to maintain all possible dialects in one repo?

Hey, im sorry i dont completely understand, can you explain again please? To explain my thinking: I tried to organize it in a way that `hcl` and `terraform` parsers are using the same grammar, just tagged by a different name, so any fixes to the grammar will get applied to both. I just chose to add the terraform dialect as a seperate library since i think downstream clients have the expectation that the language name matches the exported symbol in the generated library ( i.e. `TSLanguage *tree_sitter_hcl` for hcl ). Is there a better, more canonical way to organize this?

theHamsta:
There are different possibilities of how to maintain different dialects of a language and keep them in sync. You chose the way that tree-sitter-typescript uses which manages both typescript and tsx in one repo. There is also the possibility to add the original grammar to your `package.json` and extend it: https://github.com/theHamsta/tree-sitter-glsl/blob/master/grammar.js#L3. You can then overwrite individual rules of that upstream grammar and possibly re-use the previous rule https://github.com/theHamsta/tree-sitter-glsl/blob/master/grammar.js (referred by `original` here). The consequences of this approach:
- TSInstallFromGrammar for your grammar will require `npm`
- you don't have to copy/paste changes of the upstream grammar. Just do a `npm update`. You can use a git dependency in your grammar to fit better to our rolling release strategy but could also rely on versions publish to npm (if any)
- you might have to switch between two repos to have all the context to edit the grammar (this might be easier with the `makeGrammar` approach)

MichaHoffmann:
Ah, thats pretty cool, i didnt think of that, thanks! II would say, given that both languages are exactly the same, the `makeGrammar` approach makes it a bit easier for me though!

theHamsta:
> Ah, thats pretty cool, i didnt think of that, thanks! I would say, given that both languages are exactly the same, the makeGrammar approach makes it a bit easier for me though!

Why so? If there are (almost) no changes to the original grammar, then the grammar.js of the derived grammar can be quiet short. Here only one extra is added to grammar https://gitlab.com/WhyNotHugo/tree-sitter-jsonc/-/blob/main/grammar.js (i.e. comment to JSON without comments. You would only need to add rules that are different from the original.

MichaHoffmann:
> are (almost) no changes to the original grammar, then the grammar.js of the derived grammar can be quiet short. Here only one extra is added to grammar https://gitlab.com/WhyNotHugo/tree-sitter-jsonc/-/blob/main/grammar.js (i.e. comment to JSON

Its pretty nice to be able to just import the external scanner and wrap its methods, though that could also be done using a subrepo probably. If its easier for `nvim-treesitter` i can also use a dedicated repository of course!

theHamsta:
It should not matter for nvim-treesitter as long as both grammars don't diverge with regard to the shared queries. This has been a pain point with JS/JSX/TS/TSX which pull in breaking changes at different times and it difficult to tell which dialect supports which variant of a query

MichaHoffmann:
> It should not matter for nvim-treesitter as long as both grammars don't diverge with regard to the shared queries. This has been a pain point with JS/JSX/TS/TSX which pull in breaking changes at different times and it difficult to tell which dialect supports which variant of a query

Ah in this case its not an issue; i intended terraform to use exactely the same parser as hcl always, its just so terraform can use other queries than general hcl.

theHamsta:
@MichaHoffmann wouldn't in that case a software solutions better? I think in upstream Neovim it was planned (and partially implemented) to have different names for grammar name, parser file name and query name. In this, casa we would just need to configure that Terraform ft uses the terraform queries but the HCl parser.

clason:
That did happen: https://github.com/neovim/neovim/pull/19496

MichaHoffmann:
Ah i wasnt aware that this is possible!  Thanks for notifying me, that would indeed be better, do we have examples of that already?

MichaHoffmann:
Had a brief look at neovim and couldnt find a "#filetype?" predicate and no example of having different parser name and query names.

MichaHoffmann:
Should we go ahead with this PR and once one of those is implemented i earmark terraform queries to be adapted with either of those and remove dialects again from my upstream parser?

theHamsta:
> Had a brief look at neovim and couldnt find a "#filetype?" predicate and no example of having different parser name and query names.

This would be something for a PR to add.

PR #3922, opened 2022-12-04
feat(rust): injections comment in rust

Decodetalkers:
This will highlight code in commend line
image

### DIFF
--- a/queries/rust/injections.scm
+++ b/queries/rust/injections.scm
@@ -32,3 +32,24 @@
     name: (identifier) @_new (#eq? @_new "new"))
   arguments: (arguments
     (raw_string_literal) @regex))
+
+
+; comment begin with /// ```
+(
+ (line_comment) @_first 
+ (_) @rust
+ (line_comment) @_last 
+ (#match? @_first "^/// ```$") 
+ (#match? @_last "^/// ```$")
+ (#offset! @rust 0 4 0 4)
+)
+
+; comment begin with //! ```
+(
+ (line_comment) @_first 
+ (_) @rust
+ (line_comment) @_last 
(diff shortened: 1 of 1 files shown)


theHamsta:
It would be nicer to inject markdown and properly exclude the leading `//!` from the injection ranges but our current code seems to not be able to due this due to bugs in the implementation

ObserverOfTime:
Injecting markdown would not inject rust in unlabelled code blocks.

amaanq:
rust-analyzer does this (albeit at a much higher resource cost than tree-sitter), so I'll mark this as draft until we can exclude the leading //[!/] in every line

valpackett:
> Injecting markdown would not inject rust in unlabelled code blocks.

What would be really great to solve this is an ability to query **into** injected languages' parse trees, so *Rust* injection queries could drill down into Markdown and specify that fenced code blocks without any language name should have Rust injected back into them.

I have a similar use case for this feature in a different context currently, looks like this:

```hcl
code {
  lang = "yaml"
  content = <<-EOT
    this:
    - is
    - yaml: true
      because: "another variable outside of the heredoc says yaml"
      but: {{ templated | with(some: 'template_language') | that_is(always_used_for_strings) }}
  EOT
}
```

and it's sad that I can't write a query in the *host* language that captures the `"yaml"` as an `injection.language` to be used for `(text)` nodes **inside** of the parsed **templating** language.

clason:
Isn't that what `injection.parent` is for?

clason:
> It would be nicer to inject markdown and properly exclude the leading `//!` from the injection ranges but our current code seems to not be able to due this due to bugs in the implementation

This ☝️ is the way to go here (using `injection.parent`).

PR #3925, opened 2022-12-04
Fix Compiling Issue by Adding the Option -std=c99

wenijinew:


theHamsta:
A C standard can only be specified when compiling to a single language and does not work in mixed language (C and C++). Adding a language standard would complicate our build system by invoking three commands for parser, scanner and linking. There is a option to use makefile for nvim-treesitter. Our makefile specifies the required standards for C and C++.

I don't remember how this option was called (I'm on my phone right now). You can find out by grepping the source code for `makefile`

wenijinew:
@theHamsta 

Do you mean this file?
nvim-treesitter/scripts/compile_parsers.makefile

How shall I use it?

wenijinew:
I checked the code and found only list.norg has use_makefile = true.

PR #3929, opened 2022-12-05
highlights(typescript): add `module` keyword

zweimach:
Add highlights for the `module` keyword that is used inside declaration files (`.d.ts`).

### DIFF
--- a/queries/typescript/highlights.scm
+++ b/queries/typescript/highlights.scm
@@ -11,6 +11,7 @@
   "namespace"
   "override"
   "satisfies"
+  "module"
 ] @keyword
 
 (as_expression "as" @keyword)


queries/typescript/highlights.scm
```diff
@@ -11,6 +11,7 @@
   "namespace"
   "override"
   "satisfies"
+  "module"
```
theHamsta, line 14: i guess that's ok here. Some `module` keywords from other languages might be  `@include`

zweimach: I agree, but for TypeScript, `module` is synonymous with the `namespace` keyword, with a little difference when used with the `declare` keyword.

theHamsta (approved):

PR #3951, opened 2022-12-08
feat(cmake): add highlighting for builtin commands and some highlighting for special parameters of commonly used commands

uyha:


queries/cmake/highlights.scm
```diff
@@ -50,36 +59,67 @@
...
+)
+
+(normal_command
+  (identifier) @_function
   . (argument) @variable
-  (#match? @function.builtin "\\c^(set)$")
+  (#match? @_function "\\c^set$")
```
theHamsta, line 96: is this query only there to highlight `@variable`

queries/cmake/highlights.scm
```diff
@@ -50,36 +59,67 @@
   (argument)* @parameter
 )
 
+(block_def
+  (block_command
+    (block) @function.builtin
+    (argument (unquoted_argument) @constant)
+    (#any-of? @constant "SCOPE_FOR" "POLICIES" "VARIABLES" "PROPAGATE")
```
theHamsta, line 66: are those really constants? What's their value?

uyha: These are a bit weird since CMake uses these "constants" to name parameters. Let's say if we have some like this in Lua
```lua
block{
SCOPE_FOR = {"POLICIES", "VARIABLES"},
PROPAGATE = {"var1", "var2"}
}
```
CMake would have something like this
```cmake
block(SCOPE_FOR POLICIES VARIABLES PROPAGATE var1 var2)
endblock()
```
So they are constants, but not really, however I can't find a more appropriate binding for these

theHamsta (approved):

theHamsta:
Thank you!

PR #3953, opened 2022-12-09
feat(php): improve indents for match and switch

gbprod:
Indent for `match` and `switch` are broken, this will allows to provide this default indents:

```php
<?php

match($e) {
    true => "ok",
    false => "ko",
}

switch($e) {
    case true:
        echo 'ok';
        break;
    case false:
        echo 'ko';
        break;
}
```

### DIFF
--- a/queries/php/indents.scm
+++ b/queries/php/indents.scm
@@ -7,6 +7,9 @@
   (arguments)
   (formal_parameters)
   (enum_declaration_list)
+  (switch_block)
+  (match_block)
+  (case_statement)
   "["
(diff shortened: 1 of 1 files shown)


gbprod:
For match indents, this depends on https://github.com/tree-sitter/tree-sitter-php/pull/159

clason:
Converting to draft then until that PR is merged.

gbprod:
> Converting to draft then until that PR is merged.

It's merged! 🙂

clason (approved):

PR #3959, opened 2022-12-11
lua: Use @method.call

tomtomjhj:


queries/lua/highlights.scm
```diff
@@ -172,7 +172,7 @@
 (function_call name: (dot_index_expression field: (identifier) @function.call))
 (function_declaration name: (dot_index_expression field: (identifier) @function))
 
-(method_index_expression method: (identifier) @method)
+(method_index_expression method: (identifier) @method.call)
```
theHamsta, line 175: those two languages are not related it would have been better to have two PRs for things like that.

From my point the Lua part looks ok. I'm not sure about LaTeX. Can ` command_name` only occur in command invokation or also in command definition (probably not `command_name` there. @clason

theHamsta (approved):

queries/lua/highlights.scm
clason: It's also used in definitions, so this needs to be made more targeted (distinguish based on `field`).

clason (changes requested):
LaTeX change for `command_name` is too broad; this will also apply the capture to definitions.

tomtomjhj:
Thanks for the feedback. Removed the commit for L<sup>A</sup>T<sub>E</sub>X. I'll try other approaches later.

theHamsta (approved):

clason (approved):

clason:
Speaking of which: we should probably sync with the new upstream queries: https://github.com/MunifTanjim/tree-sitter-lua/tree/main/queries

(After that, I'll update the bundled core queries.)

PR #3971, opened 2022-12-14
feat: add Ada

briot:
The grammar was built from the official Ada grammar available in the ISO standard, and from an existing similar grammar used for the Emacs ada-mode.
The resulting grammar is able to parse all source files in the GNAT runtime (the kind of "official" compiler for Ada), as well as the 3 million+ lines of code at my company), without ERROR, though of course I could not check all resulting trees.
The testsuite is reasonably complete at this stage.

### DIFF
--- a/queries/ada/highlights.scm
+++ b/queries/ada/highlights.scm
@@ -1,0 +1,179 @@
+;; highlight queries.
+;; See the syntax at https://tree-sitter.github.io/tree-sitter/using-parsers#pattern-matching-with-queries
+;; See also https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#parser-configurations
+;; for a list of recommended @ tags, though not all of them have matching
+;; highlights in neovim.
+
+[
+   "abort"
+   "abs"
+   "abstract"
+   "accept"
+   "access"
+   "all"
+   "array"
+   "at"
+   "begin"
+   "declare"
+   "delay"
+   "delta"
+   "digits"
+   "do"
+   "end"
+   "entry"
+   "exit"
+   "generic"
+   "interface"
+   "is"
+   "limited"
+   "null"
+   "of"
+   "others"
+   "out"
+   "pragma"
+   "private"
+   "range"
+   "synchronized"
+   "tagged"
+   "task"
+   "terminate"
+   "until"
+   "when"
+] @keyword
+[
+   "aliased"
+   "constant"
+   "renames"
+] @storageclass
+[
+   "mod"
+   "new"
+   "protected"
+   "record"
+   "subtype"
+   "type"
+] @keyword.type
+[
+   "with"
+   "use"
+] @include
+[
+   "body"
+   "function"
+   "overriding"
+   "procedure"
+   "package"
+   "separate"
+] @keyword.function
+[
+   "and"
+   "in"
+   "not"
+   "or"
+   "xor"
+] @keyword.operator
+[
+   "while"
+   "loop"
+   "for"
+   "parallel"
+   "reverse"
+   "some"
+] @repeat
+[
+   "return"
+] @keyword.return
+[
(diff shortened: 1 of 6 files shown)


queries/ada/highlights.scm
```diff
@@ -0,0 +1,174 @@
...
+   "delta"
+   "digits"
+   "interface"
+   "limited"
+   "of"
+   "private"
+   "range"
+   "synchronized"
+   "tagged"
+] @StorageClass
```
theHamsta, line 47: `@storageclass`

queries/ada/highlights.scm
```diff
@@ -0,0 +1,174 @@
...
+[
+   "mod"
+   "new"
+   "protected"
+   "record"
+   "subtype"
+   "type"
+] @type.definition
```
theHamsta, line 55: I don't think `@type.definition` applies to any of those keywords. `@type.definition` is not a good name. It is thought for the identifier of a typdeffed definition

queries/ada/highlights.scm
```diff
@@ -0,0 +1,174 @@
...
+
+(subprogram_declaration "is" @keyword.function "abstract"  @keyword.function)
+(aspect_specification "with" @keyword.function)
+
+(full_type_declaration "is" @type.definition)
+(subtype_declaration "is" @type.definition)
+(record_definition "end" @type.definition)
+(full_type_declaration (_ "access" @type.definition))
+(array_type_definition "array" @type.definition "of" @type.definition)
+(access_to_object_definition "access" @type.definition)
```
theHamsta, line 141: same here

queries/ada/textobjects.scm
```diff
@@ -0,0 +1,22 @@
+;; Support for high-level text objects selections.
```
theHamsta, line 1: textobjects live in a separate repo github.com/nvim-treesitter/nvim-treesitter-textobjects

queries/ada/highlights.scm
```diff
@@ -0,0 +1,176 @@
...
+   "xor"
+] @keyword.operator
+[
+   "while"
+   "loop"
+   "for"
+   "parallel"
+   "reverse"
+   "some"
+] @keyword.repeat
```
theHamsta, line 82: Just `@repeat`

queries/ada/highlights.scm
```diff
@@ -0,0 +1,176 @@
...
+   "renames"
+] @storageclass
+[
+   "mod"
+   "new"
+   "protected"
+   "record"
+   "subtype"
+   "type"
+] @keyword.type
```
theHamsta, line 55: this is a new capture. but as a subcapture of course allowed

queries/ada/textobjects.scm
theHamsta: ping

queries/ada/highlights.scm
briot: Let's keep it for now, and we can change it if/when treesitter proposes a standard approach for this one. In a strongly typed language like Ada, it is kind of nice to highlight type definition statements in a different color (in my daily neovim usage, this is the main difference between the regexp-based highlighting and the treesitter one).

queries/ada/textobjects.scm
briot: Oups, I thought I had removed it previously, sorry

queries/ada/highlights.scm
theHamsta: We once had `@structure` which was specifically for highlighting the `struct` keyword in C. It has a built-in highlight but I don't think it got added again.

briot: Perhaps I should use @keyword.typedef to make it clearer that it is to define a type, and not an actual type (though a keyword would not be a type anyway)

theHamsta (approved):

PR #3973, opened 2022-12-14
feat: allow to set keymappings to `false`

### LINKED ISSUE #3954: Optional keymaps for module *incremental-selection*
Atm there is no way to disable keymaps for *incremental-selection*, neither individually nor globally. My use case requires calling init, increment and decrement as part of a larger function.

Looking at the code this seems easy to achieve. A first test with an empty string as indicator works as expected. I can create a PR if you want, and change the approach if necessary.

Have a nice day!

theHamsta:
Fixes #3954


theHamsta:
@notEvil does this work for you? I have not tested it yet.

notEvil:
Yes it does :) Thanks!

ObserverOfTime:
Wouldn't setting all keymaps to `false` and/or individual keymaps to `nil` make more sense?

theHamsta:
@ObserverOfTime doesn't work at the moment since I think our implementation recursively merges the defaults with the user provided values and `nil` would delete the entry which will cause it to be replaced by the default (didn't test whether this truely happens)

ObserverOfTime (approved):

PR #3978, opened 2022-12-15
feat: add rockspec

MrcJkb:
nvim-treesitter is a common dependency across Neovim plugins. Using luarocks may alleviate the need for users to specify their plugins' dependencies in their plugin manager.
(e.g., vim-plug or packer).

See also:
https://teto.github.io/posts/2021-09-17-neovim-plugin-luarocks.html

NOTE:
If this is merged, a package on www.luarocks.org will need to be added, which can be done from only one account.
See luarocks/luarocks-site#173.

### DIFF
--- a/nvim_treesitter-scm-1.rockspec
+++ b/nvim_treesitter-scm-1.rockspec
@@ -1,0 +1,38 @@
+local _MODREV, _SPECREV = 'scm', '-1'
+
+rockspec_format = "3.0"
+package = 'nvim_treesitter'
+version = _MODREV .. _SPECREV
+
+description = {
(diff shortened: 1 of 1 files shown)


theHamsta:
Again? I thought we already had a rockspec https://github.com/nvim-treesitter/nvim-treesitter/tree/master/contrib At least this one is less hidden

MrcJkb:
Oh, I did not see that one. Then we can probably close this.

theHamsta:
I don't know. I don't know how rockspecs work. It would maybe be more discoverable in the top-level directory. Maybe I'll have a look at loarocks

PR #3981, opened 2022-12-17
injections (vue): fix js/css injection conditions

### LINKED ISSUE #3980: Vue <script> tags are being parsed by both javascript and typescript parsers even when specifying the lang as "typescript"
### Describe the bug

As of now, when writing a `.vue` file, the raw text inside `script` tags are being parsed by both javascript and typescript parsers, leading to folding/indenting issues

### To Reproduce

1. Install `javascript`, `typescript`, `html`, `vue` parsers from `nvim-treesitter`
2. Create a simple vue file as below
```vue
<template>OK</template>
<script lang="ts">
function foo(bar: number): number {
  return bar;
}
</script>
```

### Expected behavior

The parsed AST should be something like the following (through overriding default injections)
image

Actual AST generated by `nvim-treesitter` vue injections
image

```markdown

Installation ~
- OK `node` found v16.15.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
  Version: cc (GCC) 12.2.0

OS Info:
{
  machine = "x86_64",
  release = "6.0.12-zen1-1-zen",
  sysname = "Linux",
  version = "#1 ZEN SMP PREEMPT_DYNAMIC Thu, 08 Dec 2022 11:03:40 +0000"
} ~

Parser/Features H L F I J
  - todotxt        ✓ . . . .
  - diff           ✓ . . . .
  - scss           ✓ . . ✓ .
  - c              ✓ ✓ ✓ ✓ ✓
  - vim            ✓ ✓ ✓ . ✓
  - css            ✓ . ✓ ✓ ✓
  - make           ✓ . ✓ . ✓
  - lua            ✓ ✓ ✓ ✓ ✓
  - vue            ✓ . ✓ ✓ ✓
  - cpp            ✓ ✓ ✓ ✓ ✓
  - git_rebase     ✓ . . . ✓
  - toml           ✓ ✓ ✓ ✓ ✓
  - html           ✓ ✓ ✓ ✓ ✓
  - typescript     ✓ ✓ ✓ ✓ ✓
  - tsx            ✓ ✓ ✓ ✓ ✓
  - ninja          ✓ . ✓ ✓ .
  - python         ✓ ✓ ✓ ✓ ✓
  - query          ✓ ✓ ✓ ✓ ✓
  - help           ✓ . . . ✓
  - gitcommit      ✓ . . . ✓
  - markdown       ✓ . ✓ . ✓
  - gitattributes  ✓ . . . ✓
  - gitignore      ✓ . . . .
  - jsdoc          ✓ . . . .
  - sxhkdrc        ✓ . ✓ . ✓
  - comment        ✓ . . . .
  - jsonc          ✓ ✓ ✓ ✓ ✓
  - rasi           ✓ ✓ ✓ ✓ .
  - java           ✓ ✓ . ✓ ✓
  - regex          ✓ . . . .
  - json           ✓ ✓ ✓ ✓ .
  - dockerfile     ✓ . . . ✓
  - javascript     ✓ ✓ ✓ ✓ ✓
  - markdown_inline✓ . . . ✓
  - yaml           ✓ ✓ ✓ ✓ ✓
  - sql            ✓ . . . ✓
  - bash           ✓ ✓ ✓ . ✓

```

```text
NVIM v0.9.0-dev-516+gb55ccb432
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by lucario387@spacegliders

Features: +acl +iconv +tui
See ":help feature-compile"

```

### Additional context

I think the problem arises due to [inheriting `html_tags`](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/html_tags/injections.scm) at [L1 of vue `injections.scm`](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/vue/injections.scm#L1)

Removing the line makes the parser acts correctly on my computer (like the expected behavior picture)

Edit no.1: The line above still holds, but then default javascript inside <script> and <style> css highlights becomes broken, as javascript is no longer injected
So maybe adding predicates to `html_tags` to inject javascript only when not specifying `lang` for `<script>` tags, and similarly to `<style>` tags?

lucario387:
My attempt to fix #3980 

Changes:
- Will no longer inherits `html_tags`
- Will only injects javascript if `lang="js"` is specified, or if there is no `lang` attributes
- Remove `typescript` as a validate predicate for typescript injection as there is no mention of `lang="typescript"` being a valid attribute on [vuejs website](https://vuejs.org/)

| Before AST | After AST |
| -- | --|
| image | image | 

Any helps would be appreciated

### DIFF
--- a/queries/vue/injections.scm
+++ b/queries/vue/injections.scm
@@ -1,4 +1,35 @@
-; inherits: html_tags
+(
+  (style_element
+    (start_tag) @_no_lang
+    (raw_text) @css)
+  (#not-contains? @_no_lang "lang=")
+) 
+
+(
+  (style_element
+    (start_tag
+      (attribute
+        (quoted_attribute_value (attribute_value) @_lang)))
+    (raw_text) @css)
+  (#eq? @_lang "css")
+)
+
+; if start_tag does not specify `lang="..."` then set it to javascript
+(
+ (script_element
+    (start_tag) @_no_lang 
+  (raw_text) @javascript)
+ (#not-contains? @_no_lang "lang=")
+)
+
+(
+  (script_element
+    (start_tag
+      (attribute
+        (quoted_attribute_value (attribute_value) @_lang)))
+    (raw_text) @javascript)
+  (#eq? @_lang "js")
+)
 
 (
   (style_element
@@ -15,7 +46,7 @@
       (attribute
         (quoted_attribute_value (attribute_value) @_lang)))
     (raw_text) @typescript)
-  (#any-of? @_lang "ts" "typescript")
+  (#eq? @_lang "ts")
 )
 
 ((interpolation


queries/vue/injections.scm
```diff
@@ -1,4 +1,35 @@
...
+ (#not-contains? @_no_lang "lang=")
+)
+
+(
+  (script_element
+    (start_tag
+      (attribute
+        (quoted_attribute_value (attribute_value) @_lang)))
+    (raw_text) @javascript)
+  (#any-of? @_lang "js" "javascript")
```
gegoune, line 31: Does it require case for `ts` as well or will it just work for `typescript`?

queries/vue/injections.scm
```diff
@@ -1,4 +1,35 @@
...
+(
+  (style_element
+    (start_tag
+      (attribute
+        (quoted_attribute_value (attribute_value) @_lang)))
+    (raw_text) @css)
+  (#eq? @_lang "css")
+)
+
+; if start_tag does not specify `lang="..."` then set it javascript
```
gegoune, line 17:
```suggestion
; if start_tag does not specify `lang="..."` then set it to javascript
```

queries/vue/injections.scm
lucario387: > Does it require case for `ts` as well or will it just work for `typescript`?
This is based on the `ts` case below that, and it works well from what I experienced, so I do it for javascript as well

lucario387: Actually, when setting `lang` attribute to either `javascript` or `typescript`, volar will responds back with an error

@gegoune should i remove both it and typescript then

image

gegoune: `lang` should be `ts` or `js`, not full names.

My original comment was a mistake by not looking at entire file but only your changes - apologies for confusing you.

lucario387: > `lang` should be `ts` or `js`, not full names.
> 
> My original comment was a mistake by not looking at entire file but only your changes - apologies for confusing you.

Just removed it, can you have another check to see if there's issues?

lucario387:
@ObserverOfTime  im asking for a help, is there any way to optimize the injection queries of this PR?

I don't know how to check whether an attribute exists or not.

Unsure of how to do it so im brute forcing it to default , specifically the `@_no_lang` part

ObserverOfTime:
> I don't know how to check whether an attribute exists or not.

I don't know if that's possible either. @theHamsta might know more.

lucario387:
@theHamsta sorry for interfering, but can you have a look at the injections queries of this PR?
I am not sure of whether I should:
- remove `html_tags` and adding custom cases to make it works like when having `html_tags` injections,
  - if yes, can you suggest of a way you think is nice to capture a tag when it doesn't have a specified attribute? I can only think of using regex with whitespaces like `lang \s*=` 
- If not, should i use something like `#set!` instead, like how Helix done their vue injections

Ty for helping

theHamsta:
@lucario387 if it is to complicated you can create your own little helper predicate that extracts the language via a regex. Like `#extract-lang!( @_no_lang "lang\\s*=\\s*(\w+)")`.

Is this PR good enough for a intermediate solution? For me it looks like already like an improvement.

theHamsta (approved):

lucario387:
> Like #extract-lang!( @_no_lang "lang\\s*=\\s*(\w+)"). This is better that having 10 different queries for each special case that all need to run.

I will learn more about predicates/directives and do a separate PR for it later. 
For a temporary solution, I think this is fine

theHamsta:
We have our custom predicates here:

https://github.com/theHamsta/nvim-treesitter/blob/627abd02e7faf2f7a55e2891e297ff9044a70645/lua/nvim-treesitter/query_predicates.lua#L130-L104

You have access to the nodes hit by the capture. To set the language for injection, you predicate would need to write the language metadata: `metadata.language = ....`.  `match[pred[2]]` would refer to the node referred to by the second argument to the predicate (e.g. `#extract-lang!( @_no_lang "lang\s*=\s*(\w+)").` the `@_no_lang` node . `pred[3]` could simply be a string, here "lang\s*=\s*(\w+)". When the `language` metadata is set, you can just use `@content` to describe where to inject the language that you just computed.

Issue #3987, opened 2022-12-17
java wrong highlight of: '!', spread param, global field


JosefLitos:
### Describe the highlighting problem

the unary negation operator is not highlighted as an operator

### Example snippet that causes the problem

```java
!false;
```

### Tree-sitter parsing result

```
unary_expression [53, 2] - [53, 8]
    operator: "!" [53, 2] - [53, 3]
    operand: false [53, 3] - [53, 8]
```

### Example screenshot

![image](https://user-images.githubusercontent.com/54900518/208253782-ebd27043-bb46-4bca-9d3a-e67c88750e8b.png)

### Expected behavior

![image](https://user-images.githubusercontent.com/54900518/208253794-53e9e83a-c208-4219-958b-f85147e1ae6d.png)
the structure looks identical, is is only needed to add "!" to the operator list

### Output of `:checkhealth nvim-treesitter`

nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.7 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v16.18.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 12.2.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## OS Info:
{
  machine = "x86_64",
  release = "6.0.12-arch1-1",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Thu, 08 Dec 2022 11:03:38 +0000"
}

## Parser/Features H L F I J
  - json           ✓ ✓ ✓ ✓ .
  - vim            ✓ ✓ ✓ . ✓
  - css            ✓ . ✓ ✓ ✓
  - yaml           ✓ ✓ ✓ ✓ ✓
  - bibtex         ✓ . ✓ ✓ .
  - lua            ✓ ✓ ✓ ✓ ✓
  - html           ✓ ✓ ✓ ✓ ✓
  - latex          ✓ . ✓ . ✓
  - java           ✓ ✓ . ✓ ✓
  - python         ✓ ✓ ✓ ✓ ✓
  - rust           ✓ ✓ ✓ ✓ ✓
[...]

PR #3990, opened 2022-12-17
highlights(java): fixed '!'+param+global highlight

JosefLitos:
Java highlighting is currently missing highlighting for '!' operator. #3987

### DIFF
--- a/queries/java/highlights.scm
+++ b/queries/java/highlights.scm
@@ -20,7 +20,8 @@
   name: (identifier) @parameter)
 
 (spread_parameter
- (variable_declarator) @parameter) ; int... foo
+ (variable_declarator
+   name: (identifier) @parameter)) ; int... foo
 
 ;; Lambda parameter
 (inferred_parameters (identifier) @parameter) ; (x,y) -> ...
@@ -50,6 +51,7 @@
 "&&"
 "|"
 "||"
+"!"
 "!="
 "=="
 "*"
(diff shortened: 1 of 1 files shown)


JosefLitos:
overwriting the highlight by adding another one after it isn't the cleanest solution, but certainly the simplest. If there is a better way to do it, please tell me, but I couldn't think of one.

ObserverOfTime:
You should squash the last two commits.

JosefLitos:
Not sure about the commit message, but squashed now.

ObserverOfTime:
Not the first commit too, but whatever.

ObserverOfTime (approved):

JosefLitos:
> Not the first commit too, but whatever.

I seem to have misunderstood. Did you mean to only squash the commits but not change the first commit's message?
I thought it wouldn't describe the commit anymore. Or was it something else?
I am quite new to github in general, so I would like to learn from my mistakes (when they've already happened).

ObserverOfTime:
The third commit was fixing an issue in the second commit so they could be squashed.
The first commit was fine on its own and didn't need to be squashed with the others.

JosefLitos:
Oh, you meant the 2 commits as with each other, not literally execute 'squash' on both of them with `git rebase -i`. My bad.
Thanks for the explanation.

ObserverOfTime:
Thank you for the contribution.

Issue #2206, opened 2022-01-06
Extremely slow startup with markdown


Fymyte:
### Before reporting

- [X] I have updated my neovim version to latest _master_
- [X] I have updated my plugin to the latest version
- [X] I have run `:TSUpdate`
- [X] I have read the [troubleshooting section](https://github.com/nvim-treesitter/nvim-treesitter#troubleshooting)

### Describe the bug

I experience an extremely long startup time with markdown files using the latest version of the plugin.
I am running the latest version of the markdown parser.
Uninstalling the parser fix the issue.
Rollback before commit
[8ada8faf2fd5a74cc73090ec856fa88f34cd364b](https://github.com/nvim-treesitter/nvim-treesitter/commit/142c017ecb7db8b9c6b735a83657141c5bf628eb)
fixes the issue too so I suppose the problem comes from this commit.

### To Reproduce

update to latest version and try to open a markdown file

### Expected behavior

Almost instant startup for 10 line files

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found  0.20.1 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v17.3.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 11.1.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 13 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - ocaml_interface✓ ✓ ✓ . ✓ 
  - beancount      ✓ . ✓ . . 
  - python         ✓ ✓ ✓ ✓ ✓ 
  - sparql         ✓ ✓ ✓ ✓ ✓ 
  - comment        ✓ . . . . 
  - lua            ✓ ✓ ✓ ✓ ✓ 
  - ocaml          ✓ ✓ ✓ . ✓ 
  - dot            ✓ . . . ✓ 
  - ql             ✓ ✓ . ✓ ✓ 
  - glsl           ✓ ✓ ✓ ✓ ✓ 
  - json           ✓ ✓ ✓ ✓ . 
  - pug            ✓ . . . ✓ 
  - ledger         ✓ . ✓ ✓ ✓ 
[...]

PR #3991, opened 2022-12-17
highlights(markdown_inline): added ')' to links

JosefLitos:


ObserverOfTime:
As per the comment, this was intentionally omitted due to #2206.
@Fymyte can you check whether the issue reoccurs with this PR?

JosefLitos:
Oh, I didn't know about this issue, though, I remember having the same problem and not using TreeSitter for markdown for the same reason. Though that doesn't seem to be my case anymore.

JosefLitos:
Also, I don't understand the parser nor the language generally, but wouldn't this be fixable by capturing the links in a simillar way to:
```query
(reference
   ref: ([
      (link_label)
      (link_text)
      (image_description)
   ] @text.reference)
   link: ([
      (link_destination)
      (uri_autolink)
   ] @text.uri)
      @punctuation.delimiter)
```

theHamsta:
@MDeiml

theHamsta:
> Oh, I didn't know about this issue, though, I remember having the same problem and not using TreeSitter for markdown for the same reason. Though that doesn't seem to be my case anymore.

@MDeiml can you confirm that this query is safe now?

MDeiml:
Oh, sry, missed that first ping. I couldn't find out what caused the problem so I don't know if it is fixed. I can test it tomorrow.

Fymyte:
Sorry for the late reply. I don't see any performance issue with this patch on my side.

theHamsta (approved):

JosefLitos:
I am sorry, I didn't remove the `;TODO` comment and link to the solved issue. Could someone in the future make sure it is removed as well?

Arrow-x:
This breaks Vimwiki!

theHamsta:
> This breaks Vimwiki!

How? Bad performance?

Arrow-x:
without the commit:
image
with:
image

clason:
What _doesn't_ break vimwiki?!

clason:
(Spoiler: Many legacy plugins rely on legacy syntax; they are not compatible with treesitter, period.)

ecly:
@Arrow-x / @clason any recommendations/workarounds on how to still use this with Vimwiki?

Arrow-x:
my bad I think it is not this commit's fault but the one that comes after it

PR #4004, opened 2022-12-21
fix(health): align the list of parsers when doing checkhealth

dundargoc:
Some parser names are too long for the given space, which shifts the
features so they aren't in line.


clason:
While you're addressing papercuts: can you remove the trailing whitespaces after the last column, too? :)

dundargoc:
That is no trailing whitespace; it's a concealed `~`. I think this needs to be fixed upstreams.

clason (approved):

PR #4015, opened 2022-12-22
fix(ts_utils): Remove side effects from `update_selection`

phgz:
At the moment, the regular `vim.keymap.set` does not work with nvim-treesitter-textobjects mappings because of side effects interference in `update_selection` function (I suspect `vim.cmd "normal! o"`).

Also related to https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/342.

This patch aims at having no side effects when updating selection.

I've only tested it manually and it seems to be working fine. Someone could help me with more robust tests?


theHamsta (approved):
This looks good! It would be good if a few more people could test it. Maybe it resolves even a few of the problems with incremental selection.

`vim.keymap.set` in nvim-treesitter would work with this but not for line visual mode like `Vaf`

PR #4021, opened 2022-12-23
injections(vue): Update injection queries

lucario387:
Follow up of #3981 
Update Vue injection queries with more queries to properly capture cases where it has no `lang` attribute

https://user-images.githubusercontent.com/29790821/209374232-63c1fa60-10ea-4a6b-8246-9d17a3df8079.mp4
It will still correctly capture even if we insert multiple spaces and newlines
https://user-images.githubusercontent.com/29790821/209374644-e178b5c5-8b49-407c-b2ec-1b609eacd173.mp4

### DIFF
--- a/queries/vue/injections.scm
+++ b/queries/vue/injections.scm
@@ -1,52 +1,80 @@
 (
   (style_element
-    (start_tag) @_no_lang
+    (start_tag) @_no_attribute
     (raw_text) @css)
-  (#not-contains? @_no_lang "lang=")
+  (#match? @_no_attribute "<\\s*style\\s*>")
 ) 
 
 (
   (style_element
     (start_tag
       (attribute
-        (quoted_attribute_value (attribute_value) @_lang)))
+        (attribute_name) @_no_lang))
     (raw_text) @css)
-  (#eq? @_lang "css")
-)
-
-; if start_tag does not specify `lang="..."` then set it to javascript
-(
- (script_element
-    (start_tag) @_no_lang 
-  (raw_text) @javascript)
- (#not-contains? @_no_lang "lang=")
-)
-
-(
-  (script_element
-    (start_tag
-      (attribute
-        (quoted_attribute_value (attribute_value) @_lang)))
-    (raw_text) @javascript)
-  (#eq? @_lang "js")
-)
+  (#not-eq? @_no_lang "lang")
+) 
 
 (
   (style_element
     (start_tag
       (attribute
-        (quoted_attribute_value (attribute_value) @_lang)))
-    (raw_text) @scss)
-  (#any-of? @_lang "scss" "postcss" "less")
+        (attribute_name) @_lang
+        (quoted_attribute_value (attribute_value) @_css)))
+    (raw_text) @css)
+  (#eq? @_lang "lang")
+  (#eq? @_css "css")
+)
+
+; If script tag does not have any extra attributes, set it to javascript
+(
+  (script_element
+    (start_tag) @_no_attribute
+    (raw_text) @javascript)
+  (#match? @_no_attribute "<\\s*script\\s*>")
+) 
+
+; if start_tag does not specify `lang="..."` then set it to javascript
+(
+ (script_element
+    (start_tag
+      (attribute
+        (attribute_name) @_no_lang))
+    (raw_text) @javascript)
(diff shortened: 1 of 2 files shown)


lucario387:
For performance concerns, at least for vue, 
- `<script>` tag has a maximum of 7 attributes + 1 tagname, making the loop inside my custom predicate runs 15 times,
- `<style>` tag has a maximum of 5 attributes + 1 tagname, so the total amount is 11 times
I am not sure whether traversing like that is the fastest way or not, but it is the most accurate capture that I've found working
image

lucario387:
For js/ts injections, it is nigh-impossible for vues' files to capture script elements using `@language` and `@content` as the parser is named javascript and typescript, while the name in `lang` of vue script elements can only be `js` or `ts`

For css/scss, it is possible, but then will be missing `less`, `postcss` and `stylus` injections as there is no parser for those languages, so not really optimal too

theHamsta:
Ah, now I understand. The vue parser already parses the style attribute. I thought you would need to parse the attribute itself. Then no custom predicate should be needed. It is simpler and more efficient when this. I thought you wanted to parse `raw_text` with `"style=css"`.

> For js/ts injections, it is nigh-impossible as of now for vues' files to capture script elements using @language and @content as the parser is named javascript and typescript, while the name in lang of vue script elements can only be js or ts

> For css/scss, it is possible, but then will be missing less, postcss and stylus injections as there is no parser for those languages, so not really optimal too

A missing language will not cause an error. It will automatically work once the language is added.

Yes, this is a known problem (see https://github.com/neovim/neovim/pull/18317, https://github.com/neovim/neovim/pull/17127)

It is strange that it does not extend the HTML parser.

lucario387:
> Ah, now I understand. The vue parser already parses the style attribute. I thought you would need to parse the attribute itself. Then no custom predicate should be needed. It is simpler and more efficient when this. I thought you wanted to parse `raw_text` with `"style=css"`.

Updated to `not-eq?` checks

lucario387:
ah but using `not-eq?` makes it so I cannot capture the tag at all
image

theHamsta:
@lucario387 maybe I don't understand the full problem until now. When it's nicer with the custom predicate we can of course use it. But we had bad experience with predicates that were traversing the node tree. Let me have a look at it when I'm less tired. I don't use Vue and I'm a bit confused about what the queries need to do exactly

lucario387:
> @lucario387 maybe I don't understand the full problem until now. When it's nicer with the custom predicate we can of course use it. But we had bad experience with predicates that were traversing the node tree. Let me have a look at it when I'm less tired. I don't use Vue and I'm a bit confused about what the queries need to do exactly

Ah, sorry for interrupting then,
Essentially, what my custom predicate does is to do an `eq?` check on the node pred[n] with pred[3], with its parent must be start_node, pred[4], pred[5], ..., pred[n-1]
So it cannot go through the entire tree and has to match case-by-case.

I thought that using `not-eq?` with the queries will have similar results to manual traversal through the tree but apparently it's not

lucario387:
@theHamsta nvm, I managed to make it work now
it works exactly like my custom predicate of before

theHamsta (approved):

theHamsta:
@lucario387 this looks great now! Can you check whether some of your ideas should also be applied to HTML? The grammar structure seems to be quite similar and HTML and vue should stay in sync in that case.

lucario387:
> @lucario387 this looks great now! Can you check whether some of your ideas should also be applied to HTML? The grammar structure seems to be quite similar and HTML and vue should stay in sync in that case.

HTML syntax marks `lang` attribute for `style` and `script` tags as deprecated and is also warned against usage, so I don't think it should be moved there.

If any, I think the script `no_lang` part can be moved to `html_tags`, then I will add extra cases for script types there (As there are cases where the parts inside `script` can be typescript/vue/etc,...

Even then, I will wait on neovim@neovim#18317 for the syntax sync, it'll be much easier to manage then

lucario387:
It certainly can be applied, but 
- I haven't done any research into possible MIME types of script and style tags for HTML, 
- I haven't tried out many cases for HTML typescript injections, so may lead to regressions to captures for HTML/Vue as a whole if I were to inherits HTML injections

theHamsta:
Thank you!

PR #4025, opened 2022-12-24
highlights(wgsl): highlight "->" for return types as `@punctuation.delimiter`

theHamsta:


theHamsta:
@szebniok

queries/wgsl/highlights.scm
```diff
@@ -50,7 +50,7 @@
 
 "return" @keyword.return
 
-[ "," "." ":" ";" ] @punctuation.delimiter
+[ "," "." ":" ";" "->" ] @punctuation.delimiter
```
theHamsta, line 53: in Rust `->` is `@operator`. It can be used for match expressions and function return types. In WGSL, it's only return types. In don't know whether you can count that as an operator, it's rather syntactical element of function definitions. But maybe `@operator` would be better for consistency.

ObserverOfTime (approved):

queries/wgsl/highlights.scm
szebniok: @theHamsta I'm sorry, I was unavailable during last week. The only language that I can think of where the `->` is used just for type definitions is [Python](https://docs.python.org/3/reference/grammar.html). And the [current highlights for it](https://github.com/nvim-treesitter/nvim-treesitter/blob/8c71c6c5edfa447aaa867858e2e913340ea964b7/queries/python/highlights.scm#L224) define it as an `@operator`

theHamsta: This use is most similar to Rust where wgsl borrowed most of its syntax. It not really a `@operator` as it does not really work on two operands or have an type of return type. But `@operator` could be used to be consistent with Rust. `@punctuation.delimiter` or `@punctuation.special` could be used since this is purely syntactical separation between function arguments and return type. `@operator` could be seen as transferring args to the return type. Feel free to open PR if you feel that the current highlight is wrong.

PR #4028, opened 2022-12-25
Injections(vue): Fix vue injection regressions

lucario387:
Change L5 and L33 to denote start of node's text so that `match?` can work properly (the `<` and `>` was recognized as something else, unsure what)
So explicity escape it to guarantees it won't be read as something else

CC @theHamsta again today, sorry for the regression :bow:

### DIFF
--- a/queries/vue/injections.scm
+++ b/queries/vue/injections.scm
@@ -3,6 +3,7 @@
     (start_tag) @_no_attribute
     (raw_text) @css)
   (#match? @_no_attribute "^\\<\\s*style\\s*\\>$")
+  ; unsure why, but without escaping &lt; and &gt; the query breaks
 ) 
 
 (
--- a/queries/vue/injections.scm
+++ b/queries/vue/injections.scm
@@ -2,7 +2,7 @@
   (style_element
     (start_tag) @_no_attribute
     (raw_text) @css)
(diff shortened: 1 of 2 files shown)


lucario387:
Sorry, will reopen again later today, I need to fix the regression I create (again)

queries/vue/injections.scm
```diff
@@ -2,7 +2,7 @@
   (style_element
     (start_tag) @_no_attribute
     (raw_text) @css)
-  (#match? @_no_attribute "<\\s*style\\s*>")
+  (#match? @_no_attribute "\\<\\s*style\\s*\\>")
```
lucario387, line 5: Somehow, without explicitly escaping `<` and `>`, the entire capture works incorrectly, unsure why that's the case

queries/vue/injections.scm
```diff
@@ -2,7 +2,7 @@
   (style_element
     (start_tag) @_no_attribute
     (raw_text) @css)
-  (#match? @_no_attribute "<\\s*style\\s*>")
+  (#match? @_no_attribute "^\\<\\s*style\\s*\\>")
```
lucario387, line 5: Unsure why, but when the `<` character was not explicitly escaped, the whole match returns true no matter the content inside

theHamsta (approved):

PR #4031, opened 2022-12-25
Update Slint highlight queries for grammar rewrite

jrmoulton:


jrmoulton:
I did a fairly major rewrite of the Slint grammar and these are the updated highlight queries

clason:
You need to wait until the updated parser is merged -- or, even better, you can update the parser by changing the [commit hash in `lockfile.json`](https://github.com/nvim-treesitter/nvim-treesitter/blob/a2d7e78b0714a0dc066416100b7398d3f0941c23/lockfile.json#L344-L346) in this PR.

jrmoulton:
Awesome. I've just done that

clason:
The tests are still failing; I think the injections need to be updated as well!

theHamsta:
@jrmoulton the problem are the indent queries.

clason:
Can we make this clearer in the test output? Right now, it's really confusing...

theHamsta:
It should now print language and query type.

clason:
Thanks, but wouldn't it be better to do this on the original error message?

theHamsta (approved):

jrmoulton:
Thanks for the help!

PR #4016, opened 2022-12-22
Update Julia highlights

savq:
- Replace `scoped_identifier` with `field_expression` in function names
- Replace true and false with `boolean_literal`

See https://github.com/tree-sitter/tree-sitter-julia/pull/85

PR #4035, opened 2022-12-26
Update Julia highlights

savq:
- Update patterns for type definitions:
  The grammar was updated to parse `abstract type` and `primitive type` as a single token.
- Update `@symbol` capture to highlight quoted operators as symbols.
- Add pattern to highlight adjoint operator.
- Update patterns and captures for:
  - "keyword" operators (`in`, `isa`, and `where`)
  - type operators (`<:`, `>:`)
  - Parametrized types.

---

- See https://github.com/tree-sitter/tree-sitter-julia/pull/87
- There's also a few changes that I forgot to add to https://github.com/nvim-treesitter/nvim-treesitter/pull/4016

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -204,7 +204,7 @@
     "revision": "505f5bd90053ae895aa3d6f2bac8071dd9abd8b2"
   },
   "julia": {
-    "revision": "f254ff9c52e994f629a60821662917d2c6c0e8eb"
+    "revision": "1cd300bda52e680872053cd55a228c1809cb0c3a"
   },
   "kotlin": {
     "revision": "b953dbdd05257fcb2b64bc4d9c1578fac12e3c28"
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -13,7 +13,7 @@
 (macro_definition
   name: (identifier) @function.macro)
 
-(quote_expression ":" (identifier)) @symbol
+(quote_expression ":" [(identifier) (operator)]) @symbol
 
 (field_expression
   (identifier) @field .)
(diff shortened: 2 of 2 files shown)


savq:
I also have a question related to `@type.qualifier`. Is this capture meant for variable declarations or for type definitions?

`abstract type` and `primitive type` are now parsed as a single token, so `abstract` cannot be highlighted differently than `type` but, that's what you would expect, no? You can't have `type` without `abstract`. I saw this was added in a large refactoring (#3693), so maybe `@type.qualifier` didn't fit this use case anyways.

The two remaining keywords `const` and `mutable` have two different purposes: `const` is for variable declarations and `mutable` is for type definitions. I don't know which one should really be highlighted as `@type.qualifier`. Maybe I'm not getting the point of this highlight group?

theHamsta:
> I also have a question related to `@type.qualifier`. Is this capture meant for variable declarations or for type definitions?

It is primarily used for variable declarations, describing modifying the type like `mut`/`const`/`lowp`. It could also be used in type declarations.

> `abstract type` and `primitive type` are now parsed as a single token, so `abstract` cannot be highlighted differently than `type` but, that's what you would expect, no? You can't have `type` without `abstract`. I saw this was added in a large refactoring (#3693), so maybe `@type.qualifier` didn't fit this use case anyways.

I think if you go with `@keyword` it would be fine. `@type.qualifier` was `@keyword` before it existed. So it is used when you declare a type like `abstract type Foo {}` where `Foo` declares an interface?

> 
> The two remaining keywords `const` and `mutable` have two different purposes: `const` is for variable declarations and `mutable` is for type definitions. I don't know which one should really be highlighted as `@type.qualifier`. Maybe I'm not getting the point of this highlight group?

For me you could use `@type.qualifier` in both situations. It would be like `mutable` and `const` in C++ where you can put `mutable` only in the type definition. For me there is no real difference between `@type.qualifier` and `@storage`.

theHamsta:
I guess you need to update `lockfile.json` to the Julia revision you are targeting https://github.com/tree-sitter/tree-sitter-julia/pull/87

clason:
> I guess you need to update `lockfile.json` to the Julia revision you are targeting [tree-sitter/tree-sitter-julia#87](https://github.com/tree-sitter/tree-sitter-julia/pull/87)

Yes, this needs to wait until the linked parser PR is merged (which is why _this_ PR is still a draft).

theHamsta:
wh

> > I guess you need to update `lockfile.json` to the Julia revision you are targeting [tree-sitter/tree-sitter-julia#87](https://github.com/tree-sitter/tree-sitter-julia/pull/87)
> 
> Yes, this needs to wait until the linked parser PR is merged (which is why _this_ PR is still a draft).

why not test already with the PR?

savq:
> For me you could use @type.qualifier in both situations.

Ok. I'll just leave it like that for now. I can link it to keyword in my colorscheme anyways.

> why not test already with the PR?

I usually squash the PRs in tree-sitter-julia, so the hash in the lockfile corresponds to the  squashed commit.

clason (approved):

PR #4040, opened 2022-12-26
fix: require_language is needed after installation to update Vim's language table

### LINKED ISSUE #4026: auto_install requires restart or reinstallation for highlighting
### Describe the bug

With `auto_install` enabled, opening a filetype for a language for the first time successfully installs the parser for that language. Although it is installed, it does not highlight the current buffer or other buffers of that filetype. 

### To Reproduce

1. Create a treesitter config such as the following (currently using lazy.nvim, but reproduced with packer.nvim). In particular, have `auto_install` set to `true` and highlights enabled:
```lua
{
	build = ":TSUpdate",
	config = function()
		require("nvim-treesitter.configs").setup({
			auto_install = true,
			highlight = {
				enable = true,
			},
		})
	end,
}
```
2. Open a filetype that has not been opened before. Optionally, run `:TSUninstall` first to ensure the parser is not installed.
3. Observe a message such as `[nvim-treesitter] [1/1] Treesitter parser for go has been installed` confirming the parser has been installed, but that highlights are not being applied.
4. Either open a new neovim instance to see highlights applied, or run `TSInstall!` in the current instance to see highlights applied.

### Expected behavior

After the parser is installed automatically, highlights should be visible immediately without having to reinstall the parser or restart neovim.

```markdown

Installation ~
- OK `tree-sitter` found 0.20.7 (parser generator, only needed for :TSInstallFromGrammar)
- OK `node` found v19.1.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.

OS Info:
{
  machine = "arm64",
  release = "22.1.0",
  sysname = "Darwin",
  version = "Darwin Kernel Version 22.1.0: Sun Oct  9 20:15:52 PDT 2022; root:xnu-8792.41.9~2/RELEASE_ARM64_T8112"
} ~

```

```text
NVIM v0.9.0-dev-553+g9be24110f-dirty
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by rob@Roberts-MacBook-Air.local

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/HEAD-9be2411/share/nvim"

```

### Additional context

When running `:checkhealth nvim-treesitter`, a parser that has been auto-installed and but is not yet highlighting does not appear in the `Parser/Features` table. After reinstalling or starting a new neovim instance, it does appear.

This bug only seems to occur with auto_install enabled, and does not occur after installing manually with `auto_install` disabled.

theHamsta:
else `vim._ts_has_lang` will return false.

Fixes #4026

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -387,6 +387,7 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync, 
       cmd = function()
         for _, buf in ipairs(vim.api.nvim_list_bufs()) do
           if parsers.get_buf_lang(buf) == lang then
+            vim._ts_remove_language(lang)
             vim.treesitter.language.require_language(lang)
             for _, mod in ipairs(require("nvim-treesitter.configs").available_modules()) do
               require("nvim-treesitter.configs").reattach_module(mod, buf)
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -387,6 +387,7 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync, 
       cmd = function()
         for _, buf in ipairs(vim.api.nvim_list_bufs()) do
           if parsers.get_buf_lang(buf) == lang then
+            vim.treesitter.language.require_language(lang)
             for _, mod in ipairs(require("nvim-treesitter.configs").available_modules()) do
               require("nvim-treesitter.configs").reattach_module(mod, buf)
             end


clason (approved):

lua/nvim-treesitter/install.lua
```diff
@@ -387,6 +387,7 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync,
       cmd = function()
         for _, buf in ipairs(vim.api.nvim_list_bufs()) do
           if parsers.get_buf_lang(buf) == lang then
+            vim.treesitter.language.require_language(lang)
```
theHamsta, line 390: when we do a `vim._ts_remove_language(lang)` even hot-switching parsers works. Like when you had an incompatible parser installed and then install the new one. It would work in the same session. Neovim will leak the module handles anyways :smile: 

Normally, you would need to keep track of lib here until the last parser gets garbage collected until you can `dlclose` the lib.
https://github.com/neovim/neovim/blob/6f25623e793e058c78481b7fb7a42eeff642694b/src/nvim/lua/treesitter.c#L221-L163

theHamsta: So, when you see those weird red errors on a incompatible parser, you can to TSInstall or autoinstall and it will get fixed in the same session. We're reattaching to all buffers, so the new parser gets used. If anyone keeps around a reference to an old parser, it is still legal to use it as the lib-handle just got leaked.

clason (approved):

PR #4047, opened 2022-12-27
feat: also reload parsers after uninstallation

theHamsta:


lua/nvim-treesitter/install.lua
```diff
@@ -42,6 +42,26 @@ local function get_job_status()
...
 end
 
+---@param lang string
+---@return function
+local function reattach_if_possible_fn(lang)
+  return function()
+    for _, buf in ipairs(vim.api.nvim_list_bufs()) do
+      if parsers.get_buf_lang(buf) == lang then
+        vim._ts_remove_language(lang)
+        local ok = pcall(vim.treesitter.language.require_language, lang)
```
theHamsta, line 52: TODO: this could silence an error during installation. During uninstallation, failure is expected unless there's another parser.

theHamsta: arguably, we should only require the `lang` if it was required before the `remove`

clason:
Not quite sure what this is supposed to achieve?

theHamsta:
@clason to keep the parser dict up to date. When you're uninstalling a parser right now, the parser will still be used until the end of the session. With this PR you would either unload the parser and disable tree-sitter or go on with an alternative parser not installed by nvim-treesitter. It mainly tries to be symmetric with `:TSInstall`/`:TSUpdate` that also now hot-reload the parsers.

clason (approved):

PR #4052, opened 2022-12-28
parsers: Add FHIR Shorthand (FSH)

mgramigna:
Language Info: https://build.fhir.org/ig/HL7/fhir-shorthand/overview.html
TS Grammar: https://github.com/mgramigna/tree-sitter-fsh

Developed the grammar alongside some developers of the language itself. It seems to work/be correct! Let me know if I did this contribution right.. first time contributing to `nvim-treesitter`. Thanks :)

### DIFF
--- a/queries/fsh/highlights.scm
+++ b/queries/fsh/highlights.scm
@@ -16,7 +16,7 @@
   "->"
 ] @punctuation.special
 
-; Entity Keywords
+; Entities
 [
   "Profile"
   "Alias"
@@ -29,7 +29,7 @@
   "Logical"
   "Resource"
   "RuleSet"
-] @keyword
+] @type.definition
 
 ; Metadata Keywords
 [
@@ -63,13 +63,13 @@
   "where"
   "system"
   "exactly"
-] @keyword
+] @keyword.operator
 
 ; Types
 [
   "Reference"
   "Canonical"
-] @type
+] @type.builtin
 
 
 (sd_metadata (parent (name))) @type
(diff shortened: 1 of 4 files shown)


ObserverOfTime (commented):
Can some of the keywords use more specific captures?
(e.g. `@keyword.operator`, `@function`, `@type.qualifier`)

mgramigna:
@ObserverOfTime Thanks for the review! I read through [the available highlight captures](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights) some more, and updated some captures to use `@type.definition`, `@keyword.operator`, and `@type.builtin`, as those were more appropriate based on the description.

queries/fsh/highlights.scm
```diff
@@ -29,7 +29,7 @@
   "Logical"
   "Resource"
   "RuleSet"
-] @keyword
+] @type.definition
```
ObserverOfTime, line 32: `@type.definition` is only used for type aliases.

ObserverOfTime (changes requested):

queries/fsh/highlights.scm
mgramigna: Ah okay, maybe type.builtin makes more sense for this one

ObserverOfTime: After a quick look at the docs, keeping it as `@keyword` is probably the best option.

mgramigna: Got it

theHamsta (approved):

PR #4055, opened 2022-12-29
feat: add smithy

nickrodgers42:
PR to address https://github.com/nvim-treesitter/nvim-treesitter/issues/2282#issuecomment-1347352190

Add parser and queries for [smithy](https://smithy.io/2.0/index.html) from [indoorvivants/tree-sitter-smithy](https://github.com/indoorvivants/tree-sitter-smithy)

### DIFF
--- a/queries/smithy/highlights.scm
+++ b/queries/smithy/highlights.scm
@@ -1,0 +1,116 @@
+
+; Keywords
+
+[
+  "namespace"
+  "service"
+  "structure"
+  "operation"
+  "list"
+  "map"
+  "union"
+  "resource"
+  "set"
+  "metadata"
+  "apply"
+  "for"
+  "with"
+  "enum"
+  "intEnum"
+] @keyword
+
+(simple_type_name) @type.builtin
+"use" @include
+
+[
+  ":"
+  "."
+] @punctuation.delimiter
+
+
+[
+  "("
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
+  "#"
+] @punctuation.bracket
+
+(node_keywords) @boolean
+(number) @number
+(quoted_text) @string
+[
+  (comment)
+  (documentation_comment)
+] @comment
+[
+  ("@")
+] @operator
+
+(use_statement
+  (absolute_root_shape_id
+    (namespace) @namespace
+    (identifier) @type
+  ))
+
+(trait
+  (shape_id) @attribute
+)
+
+(node_string_value
+  (shape_id) @type
+)
+
+; (trait_body) @local.scope
+
+(shape_member_kvp
+  (identifier) @variable
+  (shape_id) @type
+)
+(shape_member_elided
+  (identifier) @variable
+)
+(structure_shape_statement
+  (identifier) @type.definition
+)
+(operation_shape_statement
+  (identifier) @type.definition
+)
+(list_shape_statement
+  (identifier) @type.definition
+)
+(simple_shape_statement
+  (identifier) @type.definition
+)
+(resource_shape_statement
+  (identifier) @type.definition
+)
+(union_shape_statement
+  (identifier) @type.definition
+)
+(map_shape_statement
+  (identifier) @type.definition
+)
+(service_shape_statement
+  (identifier) @type.definition
+)
+(set_shape_statement
+  (identifier) @type.definition
+)
+(enum_shape_statement
+  (identifier) @type.definition
+)
+(operation_shape_statement
+  (identifier) @type.definition
+)
+(operation_error
+  (identifier) @type
+)
+(structure_resource
+  (shape_id) @type
+)
+(mixins
+  (shape_id) @type
+)
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -581,6 +581,18 @@ list.scala = {
   maintainers = { "@stevanmilic" },
 }
 
+list.smithy = {
+  install_info = {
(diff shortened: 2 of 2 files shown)


clason:
Did you ask @keynmol whether they actually are prepared to maintain the queries here (react in a timely manner to issues and PRs reported here, in particular updating the queries to breaking parser changes)?

nickrodgers42:
I have not formally asked @keynmol, but from this [Issue comment](https://github.com/indoorvivants/tree-sitter-smithy/issues/4#issuecomment-1349551836) on their repo I was assuming as much

clason:
Yeah, please ask people first before "volunteering" them!

keynmol:
@nickrodgers42 thanks for doing this, in this case your assumption was correct even though I was quite vague.

> whether they actually are prepared to maintain the queries here

I'm _reluctantly_ prepared. Queries being maintained outside of the grammar have always felt weird to me - is there a tracking issue in nvim-treesitter to allow copying queries from the grammar repo?

I'd love to work on that issue (recently I wrote 30 lines of lua so you can see I'm a bit of an expert) to reduce the maintenance burden on me and other grammar authors.

clason:
> I'm reluctantly prepared. Queries being maintained outside of the grammar have always felt weird to me - is there a tracking issue in nvim-treesitter to allow copying queries from the grammar repo?

Ideally, queries should be maintained in the grammar repo, but they need to be collected here for the sake of integration testing and to make sure that there is some consistency in the used captures (this is actually enforced through our CI). This means that highlight queries are not completely editor-agnostic, unfortunately, so we can't just copy the queries blindly and automatically from upstream but have to do so manually (which we do). That's the big blocker here rather than the workflow code.

But if your parser repo does contain queries (that at least roughly match the ones here), that makes it much easier to deal with breaking parser updates, so the maintenance burden here should be fairly low).

keynmol:
> but they need to be collected here for the sake of integration testing and to make sure that there is some consistency in the used captures

Can you expand on this - do you mean your integration tests ensure that highlight queries don't use random non-existent highlight groups?

What about pinning the grammar repos at the version where they last worked and running integration tests that actually clone the grammar repo at a particular revision and run integration tests from there?

Maintaining a single git revision in some file is easier than syncing files IMO, and better for the users who don't need to copy queries manually if they install a 3rd party Smithy grammar. To be clear - this is the usecase I'm most interested in helping to solve.

The grammar in question was born out of a [blog post](https://blog.indoorvivants.com/2022-05-12-smithy-neovim#creating-syntax-grammar) and the limitation specified in the docs:

> Note that neither :TSInstall nor :TSInstallFromGrammar copy query files from the grammar repository. If you want your installed grammar to be useful, you must manually add query files to your local nvim-treesitter installation

clason:
> Can you expand on this - do you mean your integration tests ensure that highlight queries don't use random non-existent highlight groups?

Yes, exactly. (Although highlight groups can be defined freely, so it's more about a consistent set of groups that colorschemes can rely on.)

> What about pinning the grammar repos at the version where they last worked and running integration tests that actually clone the grammar repo at a particular revision and run integration tests from there?

That's exactly what the `lockfile.json` is for. We only merge an update after all CI passes. 

> Maintaining a single git revision in some file is easier than syncing files IMO, and better for the users who don't need to copy queries manually if they install a 3rd party Smithy grammar. To be clear - this is the usecase I'm most interested in helping to solve.

But there's no guarantee that a grammar repo will contain queries (most in fact don't). The main usecase of this plugin is for the bundled grammars -- using it to install non-bundled languages is not really a priority. It's more meant to allow for local development; ideally working grammars should be contributed here!

clason:
And to be perfectly honest: this is only one of a number of glaring issues around parser development and distribution, which I'm hesitant to spent much resources on since they really need to (and hopefully will) be addressed by the tree-sitter project itself.

queries/smithy/highlights.scm
```diff
@@ -0,0 +1,116 @@
...
+
+(shape_member_kvp
+  (identifier) @variable
+  (shape_id) @type
+)
+(shape_member_elided
+  (identifier) @variable
+)
+(structure_shape_statement
+  (identifier) @type.definition
```
theHamsta, line 76: I think `@type.definition` is a bit a confusing name. It is meant for type identifiers defined by `typedef` type aliases. Is this the same semantics here?

theHamsta (commented):
This looks overall awesome! Thank you! Just read the small comment I wrote.

queries/smithy/highlights.scm
```diff
@@ -0,0 +1,116 @@
...
+
+(trait
+  (shape_id) @attribute
+)
+
+(node_string_value
+  (shape_id) @type
+)
+
+; (trait_body) @local.scope
```
theHamsta, line 66: is this meant as a note for a possible future `locals.scm`

clason:
@nickrodgers42 ping

amaanq:
I'm willing to pick this up/maintain it

theHamsta:
the open conversations should still be addressed.

queries/smithy/highlights.scm
amaanq: they were incorrect, in my PR they've been addressed as types now

queries/smithy/highlights.scm
amaanq: not sure, I'm not very familiar with the locals.scm myself but I am pretty sure this was not meant to be `@local.scope`, it is not a locally scoped variable.

PR #4057, opened 2022-12-29
highlights(haskell): lambda and function with type signature

zweimach:
Add highlights for lambda bindings (`lambdaOne` below) and prevents capturing bindings as functions if signatures have different names (the last three bindings in the example below).  
  
| Before                                                                                                         | After                                                                                                          |
| -------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------- |
| image | image |

### DIFF
--- a/queries/haskell/highlights.scm
+++ b/queries/haskell/highlights.scm
@@ -111,9 +111,12 @@
 (function
   name: (variable) @function
   patterns: (patterns))
-((signature (fun)) . (function (variable) @function))
-((signature (context (fun))) . (function (variable) @function))
-((signature (forall (context (fun)))) . (function (variable) @function))
+(function
+  name: (variable) @function
+  rhs: (exp_lambda))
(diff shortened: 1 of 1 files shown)


theHamsta (approved):

theHamsta:
> if signatures have different names (the last three bindings in the example below).

This would be an error right? Is it necessary to validate that?

zweimach:
Did you mean an error from the tree-sitter parser?
There's none from what I know.
Errors on the gutter in the example came from Haskell Language Server.

theHamsta:
> Did you mean an error from the tree-sitter parser?
> There's none from what I know.
> Errors on the gutter in the example came from Haskell Language Server.

No, errors as ok incorrekt Haskell.

zweimach:
> This would be an error right? Is it necessary to validate that?

Yes, it should be an incorrect Haskell code (type signatures without accompanying binding).
This may not be necessary for highlighting, but it is a small sanity check.

Issue #4056, opened 2022-12-29
Error during compilation 


felipesbarbosa:
### Tressiter compiling error

When compiling tressiter it returned this error:

```
nvim-treesitter[c]: Error during compilation
/usr/bin/ld: can't find -lstdc++: File or directory may not exist
collect2: error: ld returned 1 exit status 
```

The real message may differ a little because it's being displayed in Portuguese for me and i've traslated it.

But /usr/bin/ld does exist and seems like it's properly installed

My _tressiter.lua_ config file i've used is:

```
require'nvim-treesitter.configs'.setup {
  -- A list of parser names, or "all"
  ensure_installed = { "help", "javascript", "python", "typescript", "c", "lua", "rust" },

  -- Install parsers synchronously (only applied to `ensure_installed`)
  sync_install = false,

  -- Automatically install missing parsers when entering buffer
  -- Recommendation: set to false if you don't have `tree-sitter` CLI installed locally
  auto_install = true,

  highlight = {
    -- `false` will disable the whole extension
    enable = true,

    -- Setting this to true will run `:h syntax` and tree-sitter at the same time.
    -- Set this to `true` if you depend on 'syntax' being enabled (like for indentation).
    -- Using this option may slow down your editor, and you may see some duplicate highlights.
    -- Instead of true it can also be a list of languages
   additional_vim_regex_highlighting = false,
  },
}
```
I've also tryed installing libstdc++-devel instead of libstdc++ but received the same error.

### To Reproduce

1. On fedora 37
2. Copy the config file referenced on the issue into a tressiter.lua config file
3. Open the file in neovim
4. With packer installed run a _:so_ command

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
  - WARNING: `node` executable not found (only needed for :TSInstallFromGrammar, not required for :TSInstall)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 12.2.1 20221121 (Red Hat 12.2.1-4)
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## OS Info:
{
[...]

PR #4059, opened 2022-12-29
install: allow pure C parsers to be compiled without `-lstdc++`

theHamsta:
E.g. installing C [here](https://github.com/nvim-treesitter/nvim-treesitter/issues/4056) does not require libstdc++.

### DIFF
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -99,7 +99,8 @@ function M.select_compiler_args(repo, compiler)
     }
     if
       #vim.tbl_filter(function(file)
-        return file:match "%.cc$" or file:match "%cpp$" or file:match "%.cxx$"
+        local ext = vim.fn.fnamemodify(file, ":e")
+        return ext == "cc" or ext == "cpp" or ext == "cxx"
       end, repo.files) > 0
     then
       table.insert(args, "-lstdc++")
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -96,8 +96,14 @@ function M.select_compiler_args(repo, compiler)
       repo.files,
       "-shared",
       "-Os",
-      "-lstdc++",
     }
+    if
+      #vim.tbl_filter(function(file)
+        return file:match "%.cc$" or file:match "%cpp$" or file:match "%.cxx$"
+      end, repo.files) > 0
+    then
+      table.insert(args, "-lstdc++")
+    end
     if fn.has "win32" == 0 then
       table.insert(args, "-fPIC")
     end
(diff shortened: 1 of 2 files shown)


lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -96,8 +96,14 @@ function M.select_compiler_args(repo, compiler)
...
       "-Os",
-      "-lstdc++",
     }
+    if
+      #vim.tbl_filter(function(file)
+        return file:match "%.cc$" or file:match "%cpp$" or file:match "%.cxx$"
+      end, repo.files) > 0
+    then
+      table.insert(args, "-lstdc++")
+    end
```
theHamsta, line 106: I didn't apply the same change for zig, as I assumed that to be much rarer.

theHamsta:
Is there a way to invalidate our caches to actually compile the parsers? I thought we included the SHA of install.lua in the hash? Maybe because it was a change in `select_shell_command.lua`?

lua/nvim-treesitter/shell_command_selectors.lua
clason: not on Windows, it seems...

theHamsta: For MSVC there is no distinction between C and C++

clason: I was specifically talking about  `zig`, which I've seen used on Windows. (Hilarity ensued.)

But I don't think we should support that, really: `zig c++` is _not_ ready for productive use yet.

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -96,8 +96,14 @@ function M.select_compiler_args(repo, compiler)
       repo.files,
       "-shared",
       "-Os",
-      "-lstdc++",
     }
+    if
+      #vim.tbl_filter(function(file)
+        return file:match "%.cc$" or file:match "%.cpp$" or file:match "%.cxx$"
```
ObserverOfTime, line 102: This should be more performant but it probably doesn't matter much.
```suggestion
        local ext = vim.fn.fnamemodify(file, ":e")
        return ext == "cc" or ext == "cpp" or ext == "cxx"
```

lua/nvim-treesitter/shell_command_selectors.lua
theHamsta: No idea how this works, this is why I didn't change that. If anybody can tell me how I should change that zig line, I'm happy. We would probably need to test zig when we actually want to support it

lua/nvim-treesitter/shell_command_selectors.lua
theHamsta: I wouldn't do speculations on performance when we use a vim-script function that requires conversion. But it reads clearer I guess (apart from magic happening with `:e`)

ObserverOfTime: > I wouldn't do speculations on performance when we use a vim-script function that requires conversion.

We could use `vim.split` instead. The point was to avoid running 3 pattern matches.

ObserverOfTime (approved):

PR #4067, opened 2022-12-31
ci: trim final `"]` when parsers get added in update-lockfile.yml

theHamsta:
See https://github.com/nvim-treesitter/nvim-treesitter/pull/4066

The diff will look different when a whole parser gets added:
```
@ ["gleam","revision"]
- "cfcbca3f8f734773878e00d7bfcedea98eb10be2"
+ "3eb2e1783f3bf6f85c16cdd150e2f256b2f6844e"
@ ["nix","revision"]
- "6b71a810c0acd49b980c50fc79092561f7cee307"
+ "1b69cf1fa92366eefbe6863c184e5d2ece5f187d"
@ ["wgsl_bevy"]
+ {"revision":"c81dc770310795caea5e00996505deba024ec698"}
```

### DIFF
--- a/.github/workflows/update-lockfile.yml
+++ b/.github/workflows/update-lockfile.yml
@@ -45,7 +45,7 @@ jobs:
           git config user.name "GitHub"
           git config user.email "noreply@github.com"
           git add lockfile.json
-          UPDATED_PARSERS=$(echo $(/tmp/jd -set /tmp/old_lockfile.json lockfile.json | grep @ | sed 's/","revision"\]//' |  sed 's/@ \["//') | sed 's/ /, /g')
+          UPDATED_PARSERS=$(/tmp/jd -set /tmp/old_lockfile.json lockfile.json | sed -n '/@/s/","revision"]//;s/"]//;s/@ \["//p' | sed 'N;s/\n/, /')
           echo "UPDATED_PARSERS=$UPDATED_PARSERS" >> $GITHUB_ENV
           git commit -m "Update parsers: $UPDATED_PARSERS" || echo 'No commit necessary!'
           git clean -xf


.github/workflows/update-lockfile.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           git config user.name "GitHub"
           git config user.email "noreply@github.com"
           git add lockfile.json
-          UPDATED_PARSERS=$(echo $(/tmp/jd -set /tmp/old_lockfile.json lockfile.json | grep @ | sed 's/","revision"\]//' |  sed 's/@ \["//') | sed 's/ /, /g')
+          UPDATED_PARSERS=$(echo $(/tmp/jd -set /tmp/old_lockfile.json lockfile.json | grep @ | sed 's/","revision"\]//' | sed 's/"\]//' |  sed 's/@ \["//') | sed 's/ /, /g')
```
ObserverOfTime, line 48: This can be simplified (and maybe even further but I'm not a `sed` expert):
```suggestion
          UPDATED_PARSERS=$(/tmp/jd -set /tmp/old_lockfile.json lockfile.json | sed -n '/@/s/","revision"]//;s/"]//;s/@ \["//p' | sed 'N;s/\n/, /g')
```

theHamsta: Is this really simpler? I've tried to avoid any complex commands and rather go with many simple that I was sure that they won't cause any errors on strange input. I don't understand what you're doing here since I don't know about any `sed` flags.

If your suggestion works I can blindly accept it. Did you test it?

ObserverOfTime: > Is this really simpler?

It's simpler in the sense that it runs 3 commands instead of 7.
You can ignore it if you find it complicated.

> I don't understand what you're doing here

The first `sed` matches lines that contain `@` and runs the substitutions on those.
The `-n` argument and `p` flag make it so only the matching lines are printed.

The second `sed` uses the `N` command to read the next line so that the `\n` substitution works.

> If your suggestion works I can blindly accept it. Did you test it?

I did test it locally and it works but I don't know if it has any edge cases.

ObserverOfTime (approved):

Issue #4034, opened 2022-12-26
Unify HTML-esque parsers injections


lucario387:
**Is your feature request related to a problem? Please describe.**
Right now, svelte, vue, html are basically rolling their own versions of injections for scripts and styles tags, they could be unified into one, preferably through `html_tags`, then each language can add extra queries for their own use cases

**Describe the solution you'd like**
- Have injection queries that inherits javascript on very specific instance, and should be generalized
 
- Have it inject based on MIME types
- (Upstream Neovim `@language` detection ?)

**Describe alternatives you've considered**
Alternatives I've tried while updating Vue injections were:
- Injecting Javascript, and similarly css, when its a script tag without attributes (`<script>(raw_text)</script>`)
- MIME type detection (`text/javascript`, `text/babel`, ...)  
- Use `#set!` to set a language, but I think it works as parsing the entire structure again and not overriding the previous `@language`

**Additional context**
Was discussing a bit of it on the [Vue injection PR](https://github.com/nvim-treesitter/nvim-treesitter/issues/4021#issuecomment-1364678785), but I don't know exactly how to progress so I create this issue

- Right now, one cannot just inherit `html_tags` as means for unification, because of a too generalized query that captures all script tags as javascript (it can contain jsx and typescript too)
- For general HTML, If both cases in the alternatives above don't occur, I don't really know exactly how to capture those tags without the contents being captured twice (once as javascript, another one as javascript/jsx/typescript/jsx, and css the same time with scss) 
  - I believe svelte script tags is still being captured both as javascript and typescript, can someone confirm it?
-  For vue/svelte, there is an extra attribute `lang` for preprocessors, it can be used to detect language but HTML5 doesn't have the attribute.

PR #4073, opened 2023-01-01
injections(html, vue, svelte): Update injections

lucario387:
- A partial implementation for #4034 
- Update html_tags injections file so that vue/svelte parsers can inherit it without having a script/style tag being captured twice
- Update vue, svelte injections
- Add test file for html, svelte

Preview:

| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/queries/html/injections.scm
+++ b/queries/html/injections.scm
@@ -1,12 +1,1 @@
-((style_element
-  (raw_text) @css))
-
-((attribute
-   (attribute_name) @_attr
-   (quoted_attribute_value (attribute_value) @css))
- (#eq? @_attr "style"))
-
-((script_element
-  (raw_text) @javascript))
-
-(comment) @comment
+; inherits html_tags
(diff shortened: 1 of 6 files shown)


theHamsta:
@steelsojka, @WhyNotHugo @elianiva

theHamsta (approved):

lucario387:
Very much unsure of what is the syntax of svelte/preprocessors of svelte. The only thing I can see from the docs is using `lang` as a markup for preprocessor, but I don't know what other custom preprocessors are used. So I won't fix it unless someone can guide me/lmk of the general use-case in real world

For vue, I'm using it for daily stuffs so I can be pretty sure of its real life usages

theHamsta:
@lucario387 svelte is an extension of the HTML grammar, so everything fine for HTML should be also fine for svelte.

lucario387:
@theHamsta Just used a very basic svelte file for test file, taken from images of prior svelte injections PR

Issue #3986, opened 2022-12-17
Unexpected indentation after closing tag in tsx/jsx


wyattlefevre:
### Describe the bug

In tsx/jsx files, treesitter appears to be indenting incorrectly on lines directly beneath closing tags where the closing tag is own its own line.

### To Reproduce

Create a simple tsx or jsx file and create a new line beneath a closing tag that is on its own line.

### Expected behavior

Indentation should be in line with closing tag. Instead, the cursor is indented one too many times.

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.7 (b268e412ad4848380166af153300464e5a1cf83f) (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v18.12.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: Apple clang version 14.0.0 (clang-1400.0.29.202)
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## OS Info:
{
  machine = "x86_64",
  release = "22.1.0",
  sysname = "Darwin",
  version = "Darwin Kernel Version 22.1.0: Sun Oct  9 20:14:54 PDT 2022; root:xnu-8792.41.9~2/RELEASE_X86_64"
}

## Parser/Features H L F I J
  - lua            ✓ ✓ ✓ ✓ ✓
  - pascal         ✓ ✓ ✓ ✓ ✓
  - gitattributes  ✓ . . . ✓
  - phpdoc         ✓ . . . .
  - git_rebase     ✓ . . . ✓
  - comment        ✓ . . . .
  - gitcommit      ✓ . . . ✓
  - jsdoc          ✓ . . . .
  - blueprint      ✓ . . . .
  - sparql         ✓ ✓ ✓ ✓ ✓
  - sql            ✓ . . . ✓
  - gdscript       ✓ ✓ . ✓ ✓
  - bash           ✓ ✓ ✓ . ✓
  - arduino        ✓ ✓ ✓ ✓ ✓
  - html           ✓ ✓ ✓ ✓ ✓
  - godot_resource ✓ ✓ ✓ . .
[...]

PR #4076, opened 2023-01-02
fix(html): fix incorrect indents for elements/texts after end tag

lucario387:
- HTML was only indenting child elements of element captures, but dedents all end tags (including script/style elements)
Change it so it now only dedents end tag of non-script/style elements
- Indent for elements was specified, but no ending mark for indents means directly following text/element will be incorrectly indented
  - It will still indent correctly if you add one space after the ending tag (not nice)
  - So adding `indent_end` to properly mark end of an element
Before:
```html
<body>
  <script>
</script>
  <div>
  </div>
    | cursor will be here if you click enter on the line above
</body>
```
After: 
```html
<body>
  <script>
  </script>
  <div>
  </div>
  | cursor will be here if you click enter on the live above
</body>
```

### DIFF
--- a/queries/html_tags/indents.scm
+++ b/queries/html_tags/indents.scm
@@ -1,9 +1,32 @@
 [
-  (element)
+  (
+    (element
+      (start_tag 
+        (tag_name) @_not_special)
+    )
+    (#not-any-of? @_not_special "meta" "link")
+  )
+  (element (self_closing_tag))
 ] @indent
 
+; These tags are usually written one-lined and doesnt use self-closing tags so special-cased them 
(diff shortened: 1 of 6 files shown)


theHamsta (changes requested):
Can you please add your example as a test case?

lucario387:
@theHamsta moving this to draft to also tackle #3986 

I will push tests to cover for both cases

lucario387:
@theHamsta 

> Can you please add your example as a test case?

Just added tests, can you have a check again?

theHamsta (approved):

PR #4082, opened 2023-01-03
Switch to a better Meson grammar

EleanorNB:
https://github.com/staysail/tree-sitter-meson is a more accurate and complete Meson grammar than the current one, by the admission of [the creator of the current one](https://github.com/Decodetalkers/tree-sitter-meson/issues/4#issuecomment-1301503254).

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -82,7 +82,7 @@ list.agda = {
 
 list.meson = {
   install_info = {
-    url = "https://github.com/Decodetalkers/tree-sitter-meson",
+    url = "https://github.com/staysail/tree-sitter-meson",
     branch = "master",
     files = { "src/parser.c" },
   },


ObserverOfTime:
Did @Decodetalkers agree to keep maintaining the new grammar here?
If not, would you be willing to maintain it yourself?

EleanorNB:
@ObserverOfTime Hmm, no I don't have a commitment from @staysail to maintain it. I'll follow that up. (You really don't want me maintaining anything.)

Decodetalkers:
> Did @Decodetalkers agree to keep maintaining the new grammar here? If not, would you be willing to maintain it yourself?

He/She need to add the sheme file first.. but I have tried the grammer, their still have a lot of bugs, you can try it with helix, I think the gammer have much bugs..So ..

My grammer will still maintain it

Decodetalkers:
> @ObserverOfTime Hmm, no I don't have a commitment from @staysail to maintain it. I'll follow that up. (You really don't want me maintaining anything.)

Try use helix to view the meson.build under wlroots. and neovim.. I think my grammer can handle the highlight better, but I need others' option..

EleanorNB:
Oh, I see the issue. Sorry, didn't read it all through. I'll try again later.

PR #4095, opened 2023-01-04
ci: properly fix lockfile action

ObserverOfTime:
No more `sed`.


clason:
Yep, that works, thank you!

clason (approved):

PR #4101, opened 2023-01-05
folds(rust): add array_expression

p00f:


p00f:
This is needed to fold "vertical" arrays like this:
image

theHamsta (approved):

PR #4103, opened 2023-01-05
highlights(haskell): function and variable bindings

zweimach:
This PR changes signature names to have the same highlights as the bindings (see the example below).
That's because the actual type is after the `::` operator, and it also keeps the highlighting in line with Elm's, which has a similar syntax.

| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/queries/haskell/highlights.scm
+++ b/queries/haskell/highlights.scm
@@ -106,16 +106,19 @@
 
 (variable) @variable
 (pat_wildcard) @variable
+(signature name: (variable) @variable)
 
-(signature name: (variable) @type)
 (function
   name: (variable) @function
   patterns: (patterns))
 (function
   name: (variable) @function
   rhs: (exp_lambda))
+((signature (variable) @function (fun)) . (function (variable)))
 ((signature (variable) @_type (fun)) . (function (variable) @function) (#eq? @function @_type))
+((signature (variable) @function (context (fun))) . (function (variable)))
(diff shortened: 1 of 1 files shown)


theHamsta:
Wouldn't it make sense to distinguish the both. The `@type` highlight on the signature kind of makes sense... But I'm not sure.

theHamsta:
@guibou @lf- @obiwanjacobi

lf-:
Hi! I'm lightly against this change since I quite liked being able to see at a glance which line is a type declaration by it all being a different colour. That said, this isn't a very strong preference, and consistency is probably good. 

It's probably most correct to do what you're doing here: the highlighting of other languages puts variable names in the variable-name capture group.

theHamsta:
However this is resolved it would be maybe good to stay in sync with Elm since there might be people using both. I don't use Haskell but my first impression was that the before looked better and made a distinction between signature and definition

lf-:
> However this is resolved it would be maybe good to stay in sync with Elm since there might be people using both. I don't use Haskell but my first impression was that the before looked better and made a distinction between signature and definition

agreed absolutely on both points

guibou:
I personally have no argument in favor or defavor of both styles, so I'll favor the uniformity wit elm, as suggested.

ObserverOfTime (approved):
I don't know much about Haskell but this PR makes sense according to the [docs](https://wiki.haskell.org/Type_signature).

PR #4107, opened 2023-01-06
allow negative assertion in injection tests

lucario387:
Use new negative assertion in tests to make sure typescript/scss injection are not also injected as javascript/css

### DIFF
--- a/tests/query/injections/html/test-html-injections.html
+++ b/tests/query/injections/html/test-html-injections.html
@@ -5,38 +5,20 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <link href="css/style.css" rel="stylesheet">
-    <style>
-    footer{
-/*    ^ css
-*/
-    }
-    </style>
-    <style title="Test Style without type attribute">
-    footer{
-/*    ^ css
-*/
-    }
-    </style>
-    <style type="text/css" title="test style with defined type attribute">
-    footer{
-/*    ^ css
-*/
-    }
-    </style>
+    <style> footer{ } </style>
+    <!--      ^ css  -->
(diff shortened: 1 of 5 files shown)


tests/query/injections/svelte/test-svelte-injections.svelte
```diff
@@ -5,6 +5,7 @@
 <script lang="ts">
   const foo: number = 1
 //            ^ typescript
+//            ^ !javascript
```
theHamsta, line 8: can you test whether the test suite for injections supports negative assertions?

lucario387: Yeah looks like it doesn't support negative assertion just yet, will add this to injection spec then

tests/query/injections/html/test-html-injections.html
```diff
@@ -5,38 +5,20 @@
...
-    </style>
-    <style type="text/css" title="test style with defined type attribute">
-    footer{
-/*    ^ css
-*/
-    }
-    </style>
+    <style> footer{ } </style>
+    <!--      ^ css  -->
+    <style title="Test Style without type attribute"> footer{ } </style>
```
lucario387, line 10: Due to misunderstanding, I thought before that `highlight-assertions` can use comments from injected parsers too
It was not capturing things inside the script and style tags because i use css/js-style comments
Fixed to make sure they are captured now, and they all still passed :D

lucario387:
Convert one test to negative assertion for testing, and the test does return that the range should not have that lang injected
image

lucario387:
https://github.com/nvim-treesitter/nvim-treesitter/actions/runs/3860709482/jobs/6581204891
It fails, as expected

theHamsta (approved):

PR #4108, opened 2023-01-06
Update indents for Ruby

CKolkey:
Add `rescue` and `ensure` as a `@branch` captures for ruby

Standard formatting for `rescue` and `ensure` are as follows:

```ruby
def foo
  # ...
rescue StandardError
  # ...
ensure
  # ...
end
```
https://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/ensure.html


lucario387:
think you should add tests

CKolkey:
Sure... I get this when I try to run `scripts/run_tests.sh` (which I presume is the way to run the tests)

```
E492: Not an editor command: PlenaryBustedDirectory tests/ { minimal_init = './scripts/minimal_init.lua' }
```

CKolkey:
Alright, figured out I could just run `:lua require('plenary.test_harness').test_directory(vim.fn.expand("%:p"))` from in the spec file with my normal config. Seems like the script included expects `plenary` to be source-able, but doesn't ship it.

theHamsta (approved):
Thank you!

PR #4109, opened 2023-01-06
ci: add release workflow for luarocks upload

mrcjkb:
This PR introduces a release workflow that uploads a package to luarocks when a tag is pushed.

It also adds some modifications to the spec so that luarocks doesn't fail on test files that have deliberate syntax errors.
See also: https://github.com/nvim-neotest/neotest/pull/185#issuecomment-1363010085

I have tested the `Luarocks Upload` step with my fork. But I cannot test the `Test Install` step, because Iuarocks won't find the version uploaded to my account.
The Makefile may need some tweaking. Maybe @teto (who came up with the fix) has some input...

For the release workflow to work, someone with a Luarocks account will have to add their API key to this repo's secrets.

github-add-luarocks-api-key


nvim-treesitter-scm-1.rockspec
```diff
@@ -0,0 +1,48 @@
...
+      INST_LUADIR='$(LUADIR)',
+      INST_CONFDIR='$(CONFDIR)',
+    },
+    copy_directories = {
+      'autoload',
+      'plugin',
+      'queries'
+    }
+  }
+end
```
MrcJkb, line 48: I'm not sure if we need the build type to be `'make'` for `'scm'` versions.

teto: we need a Makefile because the default luarocks builder will try to build all the .c files it sees, except this fails because in the test folder, there are faulty .c files precisely to test treesitter on failure.

Makefile
```diff
@@ -0,0 +1,17 @@
+# https://github.com/luarocks/luarocks/wiki/Creating-a-Makefile-that-plays-nice-with-LuaRocks
+build: 
+	echo "do nothing"
+
+install:
+	echo $(INST_PREFIX)
```
teto, line 6: the echo could probably be removed as it was for debug.

teto:
> I have tested the Luarocks Upload step with my fork. But I cannot test the Test Install step, because Iuarocks won't find the version uploaded to my account.

I've removed my package https://luarocks.org/modules/teto/nvim-treesitter from the root manifest so @MrcJkb you can squat the root manifest and check  the install step. I hope nvim-treesitter maintainers can setup a key so we can play with this :)

MrcJkb:
> 

Thanks.
The rock was found, but the build has failed: https://github.com/MrcJkb/nvim-treesitter-fork/actions/runs/3869722833/jobs/6596044247

theHamsta:
Sorry for not responding to this for so long time! I was quite busy in the past and was quite hesitant to adopt this without fully understanding luarocks and how Github secrets work.

Am I right in the assumption that when I set a repo secret only triggeres that would have access to the secret anyways are allowed to access a secret via an action?

I would love to test this first but I don't know whether I will have time for this in the next days. If you @teto want to manage this for the perspective of our repo I can add you as one of the admins, so you can manage the secrets and tag releases.

mrcjkb:
No problem 😄 

> Am I right in the assumption that when I set a repo secret only triggeres that would have access to the secret anyways are allowed to access a secret via an action?

Anyone who can push tags will be able to trigger this workflow. But nobody (not then the repo owner) can see the secret once it has been set. It will also be masked in logs, so for example, calling `echo "$LUAROCKS_API_KEY"` will print `***`.

I've been working on a [github action for this workflow](https://github.com/mrcjkb/luarocks-tag-release) that tests the install both before and after publishing to Luarocks.
I'm hoping to transfer it to the `nvim-lua` organisation of they're interested. It might be a good idea to wait and use that workflow when it's ready.

teto:
@theHamsta thanks for the trust.I've created a "neovim" account already https://luarocks.org/modules/neovim that is used to push lspconfig and I can create a per-project api key.  I am definitely interested in making it work with nvim-treesitter and handling the administrative side. I wouldn't interfere with the rest of the repo (dont have the time anyway :'( ).

Also it's better to wait for the github action as fixes can be propagated more easily. It should definitely be handled to the nvim-lua community (if you agree), so that you dont have to maintain it on your own ! Let me know when you want to make the move (shoot me an email or pm on matrix).

theHamsta:
> 



> No problem 😄
> 
> > Am I right in the assumption that when I set a repo secret only triggeres that would have access to the secret anyways are allowed to access a secret via an action?
> 
> Anyone who can push tags will be able to trigger this workflow. But nobody (not then the repo owner) can see the secret once it has been set. It will also be masked in logs, so for example, calling `echo "$LUAROCKS_API_KEY"` will print `***`.
> 
> I've been working on a [github action for this workflow](https://github.com/mrcjkb/luarocks-tag-release) that tests the install both before and after publishing to Luarocks. I'm hoping to transfer it to the `nvim-lua` organisation of they're interested. It might be a good idea to wait and use that workflow when it's ready.

But how can you prevent anyone changing the workflow and do something else with the key. I'd assume that only triggered that would have access to the repo secrets would allow to trigger the secret

mrcjkb:
@teto  I'd be more than happy to transfer it over. I've already asked Conni2461 if they would be interested and [he said he will poke tj about it](https://github.com/nvim-lua/plenary.nvim/pull/435#issuecomment-1381636090).
I still have a small change I want to make (which I'm working on right now). At the moment, it only supports generating release rockspecs from a simple builtin template. I would like users to be able to use their own template, which would give them a lot more flexibility.

@theHamsta I personally don't have any workflows that are triggered by tags yet. But I have a bunch that use secrets to push to cachix. Whenever someone opens a PR that would trigger one of those actions, I have to approve the build manually for it to run. The same builds run automatically if I am the author of the PR. I guess it would be the same for this workflow.

theHamsta:
@teto I added you to our project. Could you manage luarocks releases? I try to create tags whenever there is a Neovim release to give users some orientation with compatibility. master is a moving release supporting current `stable` and `nightly` tag of Neovim. Each named tag of `nvim-treesitter` should support the corresponding Neovim tag.

theHamsta (approved):

teto:
I've just added the secrets I will merge and create a release later on. This may need some finer tuning later on (I know of some potential issue but merging this will make it easier)

PR #4110, opened 2023-01-06
fix(utils): swap_nodes calculates correct char_delta

mizlan:
The char_delta is not calculated correctly right now when there are two treesitter nodes being swapped, one directly following the other. This is rare but can happen for example when attempting to swap "print" and "(1)" in "print(1)". In this case an incorrect char_delta is calculated because of a bug in range comparison.

### DIFF
--- a/tests/unit/ts_utils_spec.lua
+++ b/tests/unit/ts_utils_spec.lua
@@ -110,13 +110,14 @@ end)
 describe("swap_nodes", function()
   local function swap(case)
     vim.api.nvim_buf_set_lines(0, 0, -1, false, case.lines)
-    local a = vim.treesitter.get_node_at_pos(0, case.a[1], case.a[2], {})
-    local b = vim.treesitter.get_node_at_pos(0, case.b[1], case.b[2], {})
+    local a = vim.treesitter.get_node_at_pos(0, case.a[1], case.a[2], { lang = case.lang })
+    local b = vim.treesitter.get_node_at_pos(0, case.b[1], case.b[2], { lang = case.lang })
     tsutils.swap_nodes(a, b, 0, true)
   end
 
   it("works on adjacent nodes", function()
     swap {
+      lang = "python",
       lines = { "print(1)" },
       a = { 0, 0 },
       b = { 0, 5 },
@@ -131,6 +132,7 @@ describe("swap_nodes", function()
 
   it("moves cursor correctly with multiline nodes", function()
     swap {
+      lang = "lua",
       lines = { "x = { [[", "]], [[", ".....]]}" },
       a = { 0, 6 },
       b = { 1, 4 },
--- a/tests/unit/ts_utils_spec.lua
+++ b/tests/unit/ts_utils_spec.lua
@@ -125,13 +125,13 @@ describe("swap_nodes", function()
     }
 
     it("swaps text", function() end)
-    assert.equal(vim.api.nvim_buf_get_lines(0, 0, -1, false), { "(1)print" })
+    assert.same(vim.api.nvim_buf_get_lines(0, 0, -1, false), { "(1)print" })
 
     it("moves the cursor", function() end)
-    assert.equal(vim.api.nvim_win_get_cursor(0), { 1, 3 })
+    assert.same(vim.api.nvim_win_get_cursor(0), { 1, 3 })
   end)
 
-  it("moves cursor correctly with multiline nodes", function()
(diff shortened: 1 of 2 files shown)


mizlan:
This bug was found when I was trying to debug https://github.com/mizlan/iswap.nvim/issues/64

theHamsta:
Can you add a test that would detect the previously wrong code?

mizlan:
I am unsure why `vim.treesitter.get_node_at_position` seems to be `nil`

mizlan:
Ah it was renamed from position to pos.

tests/unit/ts_utils_spec.lua
```diff
@@ -106,3 +106,35 @@ describe("update_selection", function()
...
+  end
+
+  it("moves cursor correctly with adjacent nodes", function()
+    assert.equal(
+      swap_and_return_cursor {
+        lines = { "print(1)" },
+        a = { 0, 0 },
+        b = { 0, 5 },
+      },
+      { 1, 3 }
```
theHamsta, line 126: ok, cursor position is interesting. Wouldn't it be interesting to assert on the swapped text also?

mizlan: The bug specifically is only involved with the location of the cursor after the movement, but yes it would be smart to add the text content to the assert as well

mizlan:
Oops did not note API change again

mizlan:
I am beyond my depth as to why my tests are erroring in the log:
```
Failed to load parser: uv_dlopen: /home/runner/.local/share/nvim/site/pack/nvim-treesitter/start/nvim-treesitter/parser/.: cannot read file data: Is a directory
            
            stack traceback:
            	...nux64/share/nvim/runtime/lua/vim/treesitter/language.lua:36: in function 'require_language'
            	...r/nvim-linux64/share/nvim/runtime/lua/vim/treesitter.lua:38: in function '_create_parser'
            	...r/nvim-linux64/share/nvim/runtime/lua/vim/treesitter.lua:96: in function 'get_parser'
            	...r/nvim-linux64/share/nvim/runtime/lua/vim/treesitter.lua:287: in function 'get_node_at_pos'
            	...-treesitter/nvim-treesitter/tests/unit/ts_utils_spec.lua:113: in function 'swap'
            	...-treesitter/nvim-treesitter/tests/unit/ts_utils_spec.lua:119: in function <...-treesitter/nvim-treesitter/tests/unit/ts_utils_spec.lua:118>```

I would appreciate someone could take a look at why this error is occurring. I can run the equivalent function locally so I am wondering if it has to do with the treesitter parsers not being installed?

**Review by theHamsta — APPROVED** [review]

**Comment by theHamsta** [comment]
> 



> I am beyond my depth as to why my tests are erroring in the log:
> 
> ```
> Failed to load parser: uv_dlopen: /home/runner/.local/share/nvim/site/pack/nvim-treesitter/start/nvim-treesitter/parser/.: cannot read file data: Is a directory
>             
>             stack traceback:
>             	...nux64/share/nvim/runtime/lua/vim/treesitter/language.lua:36: in function 'require_language'
>             	...r/nvim-linux64/share/nvim/runtime/lua/vim/treesitter.lua:38: in function '_create_parser'
>             	...r/nvim-linux64/share/nvim/runtime/lua/vim/treesitter.lua:96: in function 'get_parser'
>             	...r/nvim-linux64/share/nvim/runtime/lua/vim/treesitter.lua:287: in function 'get_node_at_pos'
>             	...-treesitter/nvim-treesitter/tests/unit/ts_utils_spec.lua:113: in function 'swap'
>             	...-treesitter/nvim-treesitter/tests/unit/ts_utils_spec.lua:119: in function <...-treesitter/nvim-treesitter/tests/unit/ts_utils_spec.lua:118>
> ```
> 
> I would appreciate someone could take a look at why this error is occurring. I can run the equivalent function locally so I am wondering if it has to do with the treesitter parsers not being installed?

It tries to say that it does not know the filetype of the buffer.

File: tests/unit/ts_utils_spec.lua [hunk 2]
```diff
@@ -106,3 +106,42 @@ describe("update_selection", function()
     )
   end)
 end)
+
+describe("swap_nodes", function()
+  local function swap(case)
```
theHamsta, line 111: just set `ft` on the buffer. `lang` shouldn't be necessary

mizlan:
I finally got it working! can be merged :]

PR #4111, opened 2023-01-06
doc: Move capture/highlight group descriptions

savq:
Move capture/highlight group descriptions from CONTRIBUTING to the docs.

---

I didn't move the descriptions for the other captures (locals, indents, etc). Should I?


theHamsta:
Please use a separate file for the captures. `CONTRIBUTING.md` is used to determine valid captures for CI. It would probably make sense to have all the captures in one place.

clason:
As @theHamsta mentioned, the purpose of this list is _not_ (primarily) to document the available highlight groups but to specify the valid captures (these are not the same, even though they have identical names!); this list is used by our CI action so it needs to remain where it is, in its current form.

We could _also_ document the highlight groups in our help file, but that is a) a long list and b) not complete, since queries can use subcaptures of the allowed captures.

savq:
Yeah, I didn't know that the contributing file was used for that. I had to update the list of highlight groups in my colorscheme, and I thought I might as well update the docs here.

I like @theHamsta's idea of having a separate file that only has the list of highlights because using the contributing file for that is pretty weird. It makes sense to do that when the list of captures is more "stable".

> queries can use subcaptures of the allowed captures.

Are subcaptures the ones that use periods? (`@foo.bar`) I thought the periods were a naming convention. If it is meaningful, does that mean neovim could have something like [Helix's longest matching key](https://docs.helix-editor.com/themes.html#syntax-highlighting)?

clason:
> Are subcaptures the ones that use periods? (@foo.bar) I thought the periods were a naming convention. 

Yes, exactly. Neovim (core!) has an implicit fallback mechanism where the longest defined highlight group is used for a capture: so first `@type.definition.julia` is tried; if that is not defined (e.g., via `:highlight @type.definition.julia`), then `@type.definition` is tried, and finally `@type` (which is linked by default to the standard group `Type`).

theHamsta:
I thought that there could be a separate help file with code blocks much like CONTRIBUTING.md (tags could be grouped) for anyone preferring to look up our groups in vimdoc (even though I prefer the markdown).

PR #4113, opened 2023-01-07
Fix scala

theHamsta:
No, idea why "=>" is no longer in the parse trees. It is still in grammar, but maybe it gets parsed differently https://github.com/tree-sitter/tree-sitter-scala/blob/master/grammar.js#L790 Did find the energy to bisect for the exact reason.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -30,10 +30,10 @@
     "revision": "7175a6dd5fc1cee660dce6fe23f6043d75af424a"
   },
   "c_sharp": {
-    "revision": "1bd772f69b0db577122533514a239b184650adf2"
+    "revision": "f076a8efe30a0b8c249eeb61383f6d914376114c"
   },
   "clojure": {
-    "revision": "8c23e0ec078af461ccad43fffbbfc204aa6bc238"
+    "revision": "50468d3dc38884caa682800343d9a1d0fda46c9b"
   },
   "cmake": {
     "revision": "a32265307aa2d31941056d69e8b6633e61750b2f"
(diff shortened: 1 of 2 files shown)


theHamsta:
@stevanmilic any ideas?

ckipp01:
Hey @theHamsta I can also try to chase this down on the actual grammar side if you'd like. We've made a ton of changes to it over the past couple weeks.

I'm not super familiar with the inner workings of `nvim-treesitter`, but can you expand a bit on what you mean by 

> "=>" is no longer in the parse trees

Since this is a `fix` I'm assuming something is broken, but I don't see any failed tests or reports. How can I reproduce or see what you're fixing?

theHamsta:
> Hey @theHamsta I can also try to chase this down on the actual grammar side if you'd like. We've made a ton of changes to it over the past couple weeks.
> 
> I'm not super familiar with the inner workings of `nvim-treesitter`, but can you expand a bit on what you mean by
> 
> > "=>" is no longer in the parse trees
> 
> Since this is a `fix` I'm assuming something is broken, but I don't see any failed tests or reports. How can I reproduce or see what you're fixing?

We can no longer use "=>" in queries and `"=>"` no longer appears as a symbol in the grammar https://github.com/neovim/neovim/blob/f7ad46e69ed407579517694ccf3dc1beffe7acdc/src/nvim/lua/treesitter.c#L252-L262 (see result using `:lua print(vim.inspect(vim.treesitter.inspect_language("scala")))` nor does it appear as anonymous node parsed AST (e.g. using `:lua vim.treesitter.show_tree()` on a scala file (press "a" to show anonymous nodes.

theHamsta:
grafik

as an example
```scala

import scala.util.Random

val x: Int = Random.nextInt(10)

x match
  case 0 => "zero"
  case 1 => "one"
  case 2 => "two"
  case _ => "other"

```

ckipp01:
Alright, so this should be fixed as of https://github.com/tree-sitter/tree-sitter-scala/pull/92. Just need to refresh the lock file commit for Scala and it should be good to go.

theHamsta:
Ok when the old files queries are still good we can let the update workflow handle this. Thank you!

ckipp01:
Just a note, we commit changes and regenerate in different prs to not conflict since we have a multiple prs open at a given time lately, so the fix is merged, but the generation pr isn't yet. So https://github.com/tree-sitter/tree-sitter-scala/pull/94 will actually have the commit needed.

theHamsta:
Should we require to always generate the scala parser from source? I guess not since you're regenerating from time to time.

ckipp01:
> Should we require to always generate the scala parser from source? I guess not since you're regenerating from time to time.

No, not necessary. Normally we wouldn't be doing this separate, but we have a few people doing regular updates to it at the moment because we just finally got access to the grammar. In a couple weeks this flow will change and it will always be generated on pr.

PR #4114, opened 2023-01-07
Remove update_selection side-effects

coinator:
With [4bcd3a3](https://github.com/nvim-treesitter/nvim-treesitter/commit/4bcd3a3cc31b9628a48ff126938262292b1ed988) some issues have come up since `feedkeys` is being used to pass `gv`. 

With the current implementation if gv is remapped, e.g. `map gv <NOP>` then `update_selection` wouldn't work properly. That's why I suggest passing instead `:normal! gv` , as it would limit side-effects.


theHamsta:
@kiyoon @michelesr does this impact textobjects? Feedkeeys is really not what we want here but might this accidentally cause a mode switch?

theHamsta (approved):

kiyoon:
@theHamsta I like this change, and in fact this is how I would implement (I use `vim.cmd[[normal! gv]])`.

I can tell you that the 10,000 consistency tests have passed in Python, and operator pending mode works as well

PR #4116, opened 2023-01-08
fix(latex): highlights for math_environment

mathjiajia:
`math_environment` delimiter is not `math`


clason:
This is intentional (in similarity to VimTeX's highlighting); the math environments are part of the math context, same as the dollar signs for inline math. They should be distinguished from non-math environments.

mathjiajia:
If you think so.

anyway I will use local queries to `extend` this

PR #4120, opened 2023-01-08
injections(cpp): add injections for raw string literals

jdrouhard:
the delimiter describes the language to use for the injection content

```cpp
auto python_code = R"python(
  import os

  if __name__ == '__main__':
    print(os.name)
)python";
```

### DIFF
--- a/queries/cpp/injections.scm
+++ b/queries/cpp/injections.scm
@@ -1,3 +1,7 @@
 (preproc_arg) @cpp
 
 (comment) @comment
+
+(raw_string_literal
+  delimiter: (raw_string_delimiter) @language
+  (raw_string_content) @content)


ObserverOfTime:
Out of curiosity, what's the use case for this?

jdrouhard:
Delimiter itself can be anything, but conventionally you can use it to denote the embedded language (if there is one).

Uses could be:

- regex strings (`R"regex([Rr]eg(ular)?\s*[Ee]x(pression)?)regex"`)
- raw JSON or YAML or TOML for config
- HTML/CSS for web server stuff

There are so many ways to use this, as it's pretty flexible and not very intrusive. Delimiters are completely optional.

jdrouhard:
@sam-mccall can comment further. This is a relatively new idea, but we should at least sync the injections between this plugin and the upstream tree-sitter-cpp repo. See https://github.com/tree-sitter/tree-sitter-cpp/pull/181 for the initial PR that adds this capability. @sam-mccall is one of the main devs of clangd, so he may be able to comment on this paradigm's use in llvm, at least.

Now that this is possible in the parser, I'll definitely be using it for any sort of language-injection I need in my raw string literals.

sam-mccall:
clang-format has supported it since 2017: https://github.com/llvm/llvm-project/commit/9ad83fe7f64dbb7f2bebafa8e9b7637f98de8c8d. It's been conventional in Google's internal codebase for longer than that (particularly text protobufs embedded in tests is common), and you can see likely Google-influenced examples in e.g. [absl](https://cs.github.com/?scopeName=All+repos&scope=&q=repo%3Aabseil%2Fabseil-cpp+%2FR%22%5Cw%2F) as well as in llvm/llvm-project, which is where I spend most of my time.

It seems a fairly obvious extrapolation from markdown's ```cpp blocks, and the injection facility in tree-sitter seems to anticipate this sort of usage. But I actually don't know how common this usage is in C++ at large.

ObserverOfTime:
It does appear to be used in LLVM and Google repos though the delimiter is often arbitrary (e.g. `EOF`).
Note also that `pb` won't be injected as protobuf since the parser is named `proto` and aliases are not supported.

theHamsta:
It shouldn't matter when the delimiter is arbitrary. We're allowed to fail with language resolution. It would just be a bit sad if someone would end up with an unintendend language for all the one letter abbreviations, but I don't think that the risk for this is very high.

theHamsta (approved):

theHamsta:
Though I'd admit that it is generally a bit unusual to have embedded files in C++

jdrouhard:
Not that unusual to have regex strings or JSON in test code :)

ObserverOfTime:
Seems to be mostly used for tests and sql queries at the moment.

theHamsta:
Wouldn't you rather use `R"JSON( ....)JSON"` with upper case?

PR #4128, opened 2023-01-09
folds(fennel): add `let`, `fn`, `let_clause`, `quoted_list`, `local`,…

p00f:
… `global` folds


p00f:
@clason I think this is all

clason (approved):

p00f:
Should i put all of them on different lines?

p00f:
Thanks

PR #4132, opened 2023-01-09
Update to Ruby Indents for Square Brackets

CKolkey:
I don't think the opening bracket should be included in `@branch`. I also moved `'end'` so it would be in the same order as in the previous group. That might be over the top.

### DIFF
--- a/queries/ruby/indents.scm
+++ b/queries/ruby/indents.scm
@@ -29,16 +29,15 @@
 ] @indent_end
 
 [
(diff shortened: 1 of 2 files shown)


theHamsta (approved):

theHamsta:
@TravonteD

TravonteD (approved):
Im good with this change, as it works with most ruby style guides

PR #4136, opened 2023-01-10
feat: let filetype qml use the highlight of qmljs

Decodetalkers:
Someone like qml to mark qmljs, so let 'qml'
use qmljs

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -96,7 +96,7 @@ list.qmljs = {
     branch = "master",
     files = { "src/parser.c", "src/scanner.c" },
   },
-  filetype = "qmljs",
+  filetype = "qml",
   maintainers = { "@Decodetalkers" },
 }
 


ObserverOfTime:
Can't you just set the filetype to `qml` instead?
https://github.com/nvim-treesitter/nvim-treesitter/blob/5b1299868302184f2e15819a719df698ad1aa251/lua/nvim-treesitter/parsers.lua#L99

Decodetalkers:
> Can't you just set the filetype to `qml` instead?
> https://github.com/nvim-treesitter/nvim-treesitter/blob/5b1299868302184f2e15819a719df698ad1aa251/lua/nvim-treesitter/parsers.lua#L99

Upstream use qmljs as it's name

ObserverOfTime:
> Upstream use qmljs as it's name

That doesn't matter for the filetype.

Decodetalkers:
> > Upstream use qmljs as it's name
> 
> That doesn't matter for the filetype.

I see

ObserverOfTime:
Let's limit this PR to `qml` for now. `svg` should be added after a proper `xml` parser is supported.

ObserverOfTime (approved):

PR #4153, opened 2023-01-12
refactor(scala): update the Scala highlight queries

ckipp01:
There has been a lot of changes recently to the Scala grammar and we've been merging in a handful of things before syncing the actual generated parser. We are currently at a stage where we're going to sync everything, but there are some breaking changes that are included in https://github.com/tree-sitter/tree-sitter-scala/pull/135, so when that commit gets synced here stuff will break. We've taken the queries that existed here and have been building off of them in the actual upstream repo. This will hopefully make the transition smooth and updates the queries to ensure that there is no breakage. It's recommended to wait until that pr is merged, synced here, and then merge this in around the same time.

Co-authored-by: eugene yokota <eed3si9n@gmail.com>
Co-authored-by: Anton Sviridov <keynmol@gmail.com>

CC @stevanmilic since you did all the initial work on these.

### DIFF
--- a/queries/scala/highlights.scm
+++ b/queries/scala/highlights.scm
@@ -1,28 +1,9 @@
 ; CREDITS @stumash (stuart.mashaal@gmail.com)
 
-;; variables
-
-(identifier) @variable
-
-((identifier) @variable.builtin
- (#lua-match? @variable.builtin "^this$"))
-
-(interpolation) @none
-
-; Assume other uppercase names constants.
-; NOTE: In order to distinguish constants we highlight
-; all the identifiers that are uppercased. But this solution
-; is not suitable for all occurrences e.g. it will highlight
-; an uppercased method as a constant if used with no params.
-; Introducing highlighting for those specific cases, is probably
-; best way to resolve the issue.
-((identifier) @constant (#lua-match? @constant "^[A-Z]"))
-
-;; types
-
-(type_identifier) @type
-
 (class_definition
+  name: (identifier) @type)
+
+(enum_definition
   name: (identifier) @type)
 
 (object_definition
@@ -30,26 +11,52 @@
 
 (trait_definition
   name: (identifier) @type)
+
+(full_enum_case
+  name: (identifier) @type)
+
+(simple_enum_case
+  name: (identifier) @type)
+
+;; variables
+
+(class_parameter 
+  name: (identifier) @parameter)
+
+
+(interpolation) @none
+
+;; types
 
 (type_definition
   name: (type_identifier) @type.definition)
 
+(type_identifier) @type
+
+
+;; val/var definitions/declarations
+
+(val_definition
+  pattern: (identifier) @variable)
+
+(var_definition
+  pattern: (identifier) @variable)
+
+(val_declaration
+  name: (identifier) @variable)
+
+(var_declaration
+  name: (identifier) @variable)
+
 ; method definition
 
-(class_definition
-  body: (template_body
-    (function_definition
-      name: (identifier) @method)))
-(object_definition
-  body: (template_body
-    (function_definition
-      name: (identifier) @method)))
-(trait_definition
-  body: (template_body
-    (function_definition
-      name: (identifier) @method)))
+(function_declaration
+      name: (identifier) @method)
 
-; imports
+(function_definition
+      name: (identifier) @method)
+
+; imports/exports
 
 (import_declaration
   path: (identifier) @namespace)
@@ -59,13 +66,24 @@
   path: (identifier) @type) (#lua-match? @type "^[A-Z]"))
 ((stable_identifier (identifier) @type) (#lua-match? @type "^[A-Z]"))
 
-((import_selectors (identifier) @type) (#lua-match? @type "^[A-Z]"))
+(export_declaration
+  path: (identifier) @namespace)
+((stable_identifier (identifier) @namespace))
+
+((export_declaration
+  path: (identifier) @type) (#lua-match? @type "^[A-Z]"))
+((stable_identifier (identifier) @type) (#lua-match? @type "^[A-Z]"))
+
+((namespace_selectors (identifier) @type) (#lua-match? @type "^[A-Z]"))
 
 ; method invocation
 
 
 (call_expression
   function: (identifier) @function.call)
+
+(call_expression
+  function: (operator_identifier) @function.call)
 
 (call_expression
   function: (field_expression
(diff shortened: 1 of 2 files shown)


clason:
@ckipp01 You need to include the parser update to `lockfile.json` into your PR so the queries can be tested against the correct (new) parser revision.

queries/scala/highlights.scm
```diff
@@ -1,55 +1,62 @@
...
+  name: (identifier) @type.definition)
+
+(trait_definition
+  name: (identifier) @type.definition)
+
+(full_enum_case
+  name: (identifier) @type.definition)
+
+(simple_enum_case
+  name: (identifier) @type.definition)
```
theHamsta, line 19: no, they are all just `@type`. `@type.definition` is just a confusing capture that are meant for type aliases only (C's typedef)

queries/scala/highlights.scm
```diff
@@ -1,55 +1,62 @@
...
-; is not suitable for all occurrences e.g. it will highlight
-; an uppercased method as a constant if used with no params.
-; Introducing highlighting for those specific cases, is probably
-; best way to resolve the issue.
-((identifier) @constant (#lua-match? @constant "^[A-Z]"))
-
 ;; types
 
+(type_definition
+  name: (type_identifier) @type.definition)
```
theHamsta, line 32: is this a type alias?

keynmol: It is, yes, like

```
type MyStuff = String
```

So considering your previous comment, closer to C's typedef, right?

queries/scala/highlights.scm
ckipp01: Gotcha, I updated them then.

theHamsta (approved):

queries/scala/highlights.scm
```diff
@@ -135,18 +154,29 @@
...
+  "implicit"
+  "extension"
+  "with"
 ] @keyword
 
 [
   "abstract"
   "final"
-  "implicit"
+  "using"
```
ObserverOfTime, line 167: Shouldn't this be the same as `given`?

queries/scala/highlights.scm
```diff
@@ -135,18 +154,29 @@
...
-  "protected"
   "sealed"
 ] @type.qualifier
 
+(inline_modifier) @storageclass
+
+[
+  "private"
+  "protected"
+] @storageclass
```
ObserverOfTime, line 177: Access modifiers shouldn't be `@storageclass` unless they affect storage in memory.

queries/scala/highlights.scm
```diff
@@ -110,20 +122,27 @@
...
 ] @string
 
 (interpolation "$" @punctuation.special)
 
 ;; keywords
 
+(opaque_modifier) @type.qualifier
+(infix_modifier) @keyword
+(transparent_modifier) @type.qualifier
+(open_modifier) @type.qualifier
```
ObserverOfTime, line 138: These could all be `@type.qualifier` despite `infix` not necessarily applying to types.
(The plan is to rename it to `@modifier` or something eventually so it makes more sense.)

ObserverOfTime (changes requested):

queries/scala/highlights.scm
ObserverOfTime: Should we consider renaming `@type.definition` to `@type.alias` to make it less confusing?

queries/scala/highlights.scm
ckipp01: Actually yea I believe so. @keynmol was this an oversight by chance?

keynmol: Definitely not intentional

queries/scala/highlights.scm
ckipp01: Is there sort of a definitive guide stating this somewhere? From the contributing docs it says:

> @storageclass          ; visibility/life-time modifiers

And when this was added in https://github.com/tree-sitter/tree-sitter-scala/pull/138 we also saw similar verbiage in the sublime docs.

queries/scala/highlights.scm
ckipp01: Alrighty, updated.

queries/scala/highlights.scm
theHamsta: yeah, that would be a lot better. Probably, even better when upstream to Neovim because this is how we got that name.

clason: Just a reminder: all upstream Neovim did was to add some default links to suitable-looking "standard groups"; these are not "blessed" in any way. Standardization of capture names is something only nvim-treesitter is interested in.

queries/scala/highlights.scm
ckipp01: Alrighty, I've reverted these back.

queries/scala/highlights.scm
ckipp01: Just to clarify, so these _are_ ok right now? Or you want `infix` changed? I sort of like infix being a keyword here at the moment, since like you mentioned, it's not really applying to `type` at all.

queries/scala/highlights.scm
eed3si9n: Note that in the TextMate intro blog post https://macromates.com/blog/2005/introduction-to-scopes/ Allan Odgaard wrote

> The tags are implicit in TextMate, using the language grammar. So if we apply this example grammar:

```
scopeName = 'c_source';
patterns = (
    { name = 'storage'; match = 'char'; },
    { name = 'modifier'; match = 'const'; },
    { name = 'operator'; match = '[=*]'; },
    { name = 'string'; begin = '"'; end = '"';
        patterns = (
            { name = 'constant'; match = '\\\\.'; },
        );
    }
);
```

> To this line of code:

```c
char const* str = "Hello world\n";
```

So in TextMate terminology, "storage" refers not exactly just to memory storage, but what in statically-typed programming language community would call "type" if not "field". [Sublime](https://www.sublimetext.com/docs/scope_naming.html#storage) therefore uses `storage.modifier` to describe things like `static`, `inline`, `const`, `public` and `private`. They are modifiers that can be applied to fields.

Do we have a similar concept here?

stevanmilic (approved):
great work 👏🏻

queries/scala/highlights.scm
ObserverOfTime: Yes, they're fine.

queries/scala/highlights.scm
ObserverOfTime: > Do we have a similar concept here?

Not yet.

ObserverOfTime (approved):

PR #4160, opened 2023-01-14
ci: stop running update-lockfile after every commit

clason:
This is wasteful and noisy for unrelated commits, while the intended use
(making sure that new parsers have a lockfile entry) can be covered by
either making sure this is part of the initial PR (ideally) or running
the action manually.

### DIFF
--- a/.github/workflows/update-lockfile.yml
+++ b/.github/workflows/update-lockfile.yml
@@ -3,9 +3,6 @@ name: Update lockfile
 on:
   schedule:
     - cron: '30 6 * * *'
-  push:
-    branches:
-      - master
   workflow_dispatch:
 
 jobs:


theHamsta:
Do you prefer cron over the after commit? The original intention of this after commit was to only have updates when we were working on the repo anyways. But cron is of course sufficient.

theHamsta (approved):

clason:
I strongly prefer cron; because right now we're basically _constantly_ working on the repo -- so we're getting constant PRs or force-pushes (which I constantly get notified about).

An alternative would be to add some throttling (e.g., no more than one run per hour or so), but that strikes me as needless complication.

theHamsta:
What I meant is that when I'm online I would rather commit and merge the update right away than waiting for cron to trigger a few hours later.

clason:
You can always trigger the workflow manually if you know that a parser update is _needed_.

clason:
You are also welcome to adjust the cron job to run when it's more convenient for you (or have it run twice a day).

PR #4165, opened 2023-01-15
indents(python): Tweak node's range for indentation

lucario387:
The If/for/while/try/function/class/with statements 
aren't supposed to be created with one line, so use `offset!` to tweak end_row to `end_row + 1`

Regression?: Now the nodes listed above cannot dedent right away anymore, the user must press backspace/dedent once on their own to tell the parser to end the indent range

Example of the change:
```python
def foo():|<CR>

# Before

def foo():
|

# After

def foo():
    |
```

### DIFF
--- a/queries/python/indents.scm
+++ b/queries/python/indents.scm
@@ -3,11 +3,6 @@
   (dictionary)
   (set)
 
-  (if_statement)
-  (for_statement)
-  (while_statement)
-  (with_statement)
-  (try_statement)
   (import_from_statement)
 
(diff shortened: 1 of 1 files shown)


theHamsta:
I don't know whether the indent module currently respects the range metadata

lucario387:
> I don't know whether the indent module currently respects the range metadata

It doesn't, I was mistakening vim syntax-based indenting with this, so I closed it to reduce confusion

PR #4167, opened 2023-01-16
fix(dart): add `@indent_end` to  }

SzelamC:
Thx for replying for fixing the indentation issue for dart

### DIFF
--- a/queries/dart/indents.scm
+++ b/queries/dart/indents.scm
@@ -21,6 +21,10 @@
   "]"
 ] @branch
 
+[
+ "}"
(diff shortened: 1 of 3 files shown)


theHamsta:
Can you add the failing sample you provided as a test case? I don't know whether there is already a dart_spec.lua. If not you see how indents/c_spec.lua works.

theHamsta:
Thank you!

amaanq (approved):

amaanq:
Thanks!

Issue #4170, opened 2023-01-16
Crashes Neovim when using Treesitter and the Scala grammar


ckipp01:
### Describe the bug

I originally created this over in https://github.com/nvim-telescope/telescope.nvim/issues/2328 but was able to minimize it further thanks to @gvolpe. A preface is that we've made a ton of changes to [tree-sitter-scala](https://github.com/tree-sitter/tree-sitter-scala) recently, so we very well could have done something wrong there causing this, but we can only reproduce when in Neovim so were're hoping to just gain some more insight into what could be causing this to explode when used with Neovim.

The issues is that Neovim just seems to crash when `tree-sitter-scala` is loaded at times. I mainly have hit on this when using telescope previews, but you can also minimize it to happen in a markdown block as well. I haven't hit on this when just opening a Scala file and editing it which is odd.

### To Reproduce

1. Create a `test.md`
2. Inside of that file include:

````md
# Test

```scala
object Main
```
````

3. You can use a minimal config like this:
```lua
require("packer").startup(function(use)
  use({ "wbthomason/packer.nvim" })
  use({
    "nvim-treesitter/nvim-treesitter",
    run = ":TSUpdate",
    config = function()
      require("nvim-treesitter.configs").setup({
        ensure_installed = { "scala" },
        highlight = {
          enable = true
        },
      })
    end
  })
end)
```

4. Open the file with `nvim -u ~/.config/nvim/crash.lua test.md`
5. Try to add `extends Foo` to the `object Main` line.
6. See the crash

### Expected behavior

I'd expect no crash here.

### Output of `:checkhealth nvim-treesitter`

<details>
<summary>healthcheck results</summary>
<br>

```
==============================================================================
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- OK `tree-sitter` found 0.20.7 (parser generator, only needed for :TSInstallFromGrammar)
- OK `node` found v18.13.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: Apple clang version 14.0.0 (clang-1400.0.29.202)
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "22.2.0",
  sysname = "Darwin",
[...]

PR #4171, opened 2023-01-17
chore: sync latest Scala parser

ckipp01:
I just missed the cron to generate this by like an hour, so I'm just
manually shooting this in. This should help in the crashes reported in
https://github.com/nvim-treesitter/nvim-treesitter/issues/4170 since it
includes https://github.com/tree-sitter/tree-sitter-scala/pull/160.


clason:
No need; the workflow can be triggered manually (which I just did) -- separate PRs are just necessary if the parser update requires query adaptations.

PR #4175, opened 2023-01-18
chore: update Scala parser and add in new query

ckipp01:
Note this can supersede https://github.com/nvim-treesitter/nvim-treesitter/pull/4174 since I also wanted to sync the queries.

### DIFF
--- a/queries/scala/highlights.scm
+++ b/queries/scala/highlights.scm
@@ -23,7 +23,6 @@
 (class_parameter 
   name: (identifier) @parameter)
 
-
 (interpolation) @none
 
 ;; types
@@ -32,7 +31,6 @@
   name: (type_identifier) @type.definition)
 
 (type_identifier) @type
-
 
 ;; val/var definitions/declarations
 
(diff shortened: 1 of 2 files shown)


queries/scala/highlights.scm
```diff
@@ -96,6 +93,9 @@
 (generic_function
   function: (identifier) @function.call)
 
+(interpolated_string_expression
+  interpolator: (identifier) @function.call)
```
theHamsta, line 97: Sting Interpolation is not a function call in other languages

ckipp01: Yea, so I went back and forth about this. The main reason I chose this is that under the hood in Scala this is actually represented as a `def`. For example you can see this when hovering on the actual interpolator in Scala:

Screenshot 2023-01-18 at 09 25 49

It's a macro call, but I thought `function.call` was the most fitting here. Would you suggest something else?

theHamsta: a it's the interpolator not the dollar style variable interpolation!

theHamsta (approved):

clason:
don't merge master 😭

theHamsta:
Wanted to squash the commit of the webui into the real one

PR #4183, opened 2023-01-20
feat: add htmldjango

ObserverOfTime:
It's not perfect but it works.

Closes #3402.


theHamsta (approved):

theHamsta:
Would be nice when we could at some point get our injections en-par with upstream tree-sitter so that we don't need an uni-parser for template and programming language.

ObserverOfTime:
> Would be nice when we could at some point get our injections en-par with upstream tree-sitter so that we don't need an uni-parser for template and programming language.

Can you explain the differences? I'm not aware of how injections work upstream.

theHamsta:
Upstream excludes child nodes per default from the injection. We only have an directive that doesn't seem to work correctly in all cases (e.g. injections of second order). We also have `@combined` and not-`@combined` which doesn't exist for upstream injections. Also we don't have incremental parsing for injections yet.

E.g. there is a ERB embedded template parser that is agnostic which language it templates. I don't think we can make it work with current nvim correctly while it is intended to work without special directives. The idea is that you have both languages in the same document and wherever the template language has nodes, they get automatically excluded from the templated language.

PR #4186, opened 2023-01-20
fix: detach behaviour on buffer reload

LiadOz:


lua/nvim-treesitter/caching.lua
```diff
@@ -14,32 +14,41 @@ function M.create_buffer_cache()
...
+          loaded_buffers[tostring(bufnr)] = nil
           return true
         end,
+        on_reload = function() end, -- this is needed to prevent on_detach being called on buffer reload
       })
     end
 
-    items[type_name][bufnr] = value
+    items[tostring(bufnr)][type_name] = value
+    cache.has(type_name, bufnr)
```
theHamsta, line 35: what does this line do?

theHamsta:
Would it be complicated to write a test that detected the old erroneous behavior? I have a bit of fear that your consideration might get lost if the code would get refactored at some point.

theHamsta (approved):

lua/nvim-treesitter/caching.lua
LiadOz: It was left there by mistake, I'll remove it

LiadOz:
> Would it be complicated to write a test that detected the old erroneous behavior? I have a bit of fear that your consideration might get lost if the code would get refactored at some point.

I'll take a look

lua/nvim-treesitter/caching.lua
LiadOz: removed

theHamsta:
> > Would it be complicated to write a test that detected the old erroneous behavior? I have a bit of fear that your consideration might get lost if the code would get refactored at some point.
> 
> I didn't find a way to programmatically cause a buffer to reload, it seems that the only way to do so is to have a loaded file change from outside of neovim, I think it will be too much to simulate in a unit test

Isn't `:edit` without argument doing this `:h reload`?

PR #4188, opened 2023-01-20
docs: fix comment alignment

ObserverOfTime:
Also, `@conditional.ternary` should not be in the "keywords" section, while `@conditional` should.

Possible solutions:

- Make a new group for conditionals (not worth it)
- Replace `@conditional.ternary` with `@operator` or `@operator.ternary`
- Just don't mention `@conditional.ternary` in the docs (this has my vote)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -139,32 +139,33 @@ effect on highlighting. We will work on improving highlighting in the near futur
 #### Keywords
 
 ```scheme
-@keyword          ; various keywords
-@keyword.function ; keywords that define a function (e.g. `func` in Go, `def` in Python)
-@keyword.operator ; operators that are English words (e.g. `and` / `or`)
-@keyword.return   ; keywords like `return` and `yield`
+@keyword             ; various keywords
+@keyword.function    ; keywords that define a function (e.g. `func` in Go, `def` in Python)
+@keyword.operator    ; operators that are English words (e.g. `and` / `or`)
+@keyword.return      ; keywords like `return` and `yield`
 
-@conditional      ; keywords related to conditionals (e.g. `if` / `else`)
-@conditional.ternary ; Ternary operator: condition ? 1 : 2
-@repeat           ; keywords related to loops (e.g. `for` / `while`)
-@debug            ; keywords related to debugging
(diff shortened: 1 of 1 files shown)


theHamsta:
Well, the original issue was whether `"?"  ":"` should be `@operator` or `@conditional`. `@conditional` won in the ballot while there were some people including me who preferred `@operator`. I resurrected the possibility for free choice while the fallback to `@conditional` would stay.

I think it would be good if it would be in contributing just that new ternaries are marked as such, but in theory I'm open to each of the three suggested solutions.

CONTRIBUTING.md
```diff
@@ -139,32 +139,33 @@ effect on highlighting. We will work on improving highlighting in the near futur
...
-@label            ; GOTO and other labels (e.g. `label:` in C)
-@include          ; keywords for including modules (e.g. `import` / `from` in Python)
-@exception        ; keywords related to exceptions (e.g. `throw` / `catch`)
+@keyword             ; various keywords
+@keyword.function    ; keywords that define a function (e.g. `func` in Go, `def` in Python)
+@keyword.operator    ; operators that are English words (e.g. `and` / `or`)
+@keyword.return      ; keywords like `return` and `yield`
+
+@conditional         ; keywords related to conditionals (e.g. `if` / `else`)
+@conditional.ternary ; ternary operator (e.g. `?` / `:`)
```
theHamsta, line 148: to me that looks fine as separate category :+1:

theHamsta (approved):

CONTRIBUTING.md
ObserverOfTime: Do we really want a separate category for 2 captures?

We have several special cases that aren't documented. This can be one of them.

theHamsta: also fine for me. Then, we need to remind people using ternaries. Don't you think there are people/color schemes wanting to remap this?

ObserverOfTime: If they wanted to remap it, it'd probably be to `Operator`.

theHamsta: Yes :-)

ObserverOfTime: Well, let's leave this decision for another time…

PR #4189, opened 2023-01-20
Add support for tree-sitter-ini

theHamsta:
Works with classic one-line key-value pairs. Not with Python-style inis with multiline values (highlighting is still ok with used queries)

Addresses https://github.com/nvim-treesitter/nvim-treesitter/issues/2282#issuecomment-1398856301

@dsully

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -18,6 +18,7 @@ local filetype_to_parsername = {
   cs = "c_sharp",
   tape = "vhs",
   dosini = "ini",
+  confini = "ini",
 }
 
 ---@class InstallInfo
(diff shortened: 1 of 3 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -17,6 +17,7 @@ local filetype_to_parsername = {
   rmd = "markdown",
   cs = "c_sharp",
   tape = "vhs",
+  dosini = "ini",
```
ObserverOfTime, line 20: `confini` is also a thing.

ObserverOfTime (approved):

dsully:
Thanks! Any plans to add Python-style ?

theHamsta:
@dsully you would have to ask at https://github.com/justinmk/tree-sitter-ini

PR #4196, opened 2023-01-22
Partially revert json highlight updates

lucario387:
This PR partially reverts da6dc21 as currently it breaks highlighting if the value of a pair is not a string (can also be boolean, number, array and object)

@ObserverOfTime is it ok to revert this?


ObserverOfTime (approved):
Right, that was my bad.

PR #4197, opened 2023-01-22
Bump thrift

amaanq:
I refactored the grammar a bit, which is reflected in the queries.

### DIFF
--- a/queries/thrift/highlights.scm
+++ b/queries/thrift/highlights.scm
@@ -36,18 +36,21 @@
 ; Variables
 
 (identifier) @variable
-(annotation_identifier) @variable
+(enum_identifier) @variable
 
 ; Constants
 
 (const_identifier) @constant
-(enum_identifier) @constant
 (enum_member) @constant
 
 ; Types
 
 (field_type) @type
 (type_identifier) @type
+
+; Attributes
+
+(annotation_identifier) @variable
 
 ; Operators
 
--- a/queries/thrift/highlights.scm
+++ b/queries/thrift/highlights.scm
@@ -27,7 +27,7 @@
 
 ; Fields
 
-(field) @field
+(field_identifier) @field
 
 ; Parameters
 
@@ -36,6 +36,18 @@
 ; Variables
 
 (identifier) @variable
+(annotation_identifier) @variable
+
+; Constants
+
+(const_identifier) @constant
+(enum_identifier) @constant
+(enum_member) @constant
+
+; Types
+
+(field_identifier) @type
+(type_identifier) @type
 
 ; Operators
 
@@ -110,6 +122,7 @@
 (boolean) @boolean
 
 (typedef_definition) @type.definition
+(namespace_definition) @type.definition
 
 [
   "const"
@@ -128,6 +141,8 @@
 ["(" ")"] @punctuation.bracket
 
 ["[" "]"] @punctuation.bracket
+
+["<" ">"] @puncutation.bracket
 
 [
   ";"
--- a/queries/thrift/highlights.scm
+++ b/queries/thrift/highlights.scm
@@ -50,7 +50,7 @@
 
 ; Attributes
 
-(annotation_identifier) @variable
+(annotation_identifier) @attribute
 
 ; Operators
 
--- a/queries/thrift/highlights.scm
+++ b/queries/thrift/highlights.scm
@@ -31,7 +31,7 @@
 
 ; Parameters
 
-(parameter) @parameter
+(param_identifier) @parameter
 
 ; Variables
 
@@ -87,7 +87,8 @@
   "void"
 ] @keyword
 
-; Deprecated
+; Deprecated Keywords
+
 [
   "async"
   "byte"
--- a/queries/thrift/highlights.scm
+++ b/queries/thrift/highlights.scm
@@ -36,7 +36,6 @@
 ; Variables
 
 (identifier) @variable
-(enum_identifier) @variable
 
 ; Constants
 
@@ -45,6 +44,7 @@
 
 ; Types
 
+(enum_identifier) @type
 (field_type) @type
 (type_identifier) @type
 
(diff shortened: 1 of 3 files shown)


amaanq:
Ok, think I've ironed out everything now

queries/thrift/highlights.scm
```diff
@@ -110,6 +123,7 @@
 (boolean) @boolean
 
 (typedef_definition) @type.definition
+(namespace_definition) @type.definition
```
ObserverOfTime, line 126: What makes this different from namespace?

queries/thrift/highlights.scm
```diff
@@ -27,15 +27,27 @@
...
 
 ; Parameters
 
-(parameter) @parameter
+(param_identifier) @parameter
 
 ; Variables
 
 (identifier) @variable
+(annotation_identifier) @variable
```
ObserverOfTime, line 39: Maybe `@attribute`?

queries/thrift/highlights.scm
```diff
@@ -27,15 +27,27 @@
...
+
+; Constants
+
+(const_identifier) @constant
+(enum_identifier) @constant
+(enum_member) @constant
```
ObserverOfTime, line 45: Shouldn't `enum_identifier` be `@type` instead?

queries/thrift/highlights.scm
amaanq: thrift namespaces aren't normal namespaces per se, it's more like
`namespace lang foo`

so you can export a namespace to different languages with different names, e.g.

```thrift
namespace c_glib TTest
namespace cpp thrift.test
namespace delphi Thrift.Test
namespace go thrifttest
namespace java thrift.test
namespace js ThriftTest
namespace lua ThriftTest
namespace netstd ThriftTest
namespace perl ThriftTest
namespace php ThriftTest
namespace py ThriftTest
namespace py.twisted ThriftTest
namespace rb Thrift.Test
namespace st ThriftTest
namespace xsd test (uri = 'http://thrift.apache.org/ns/ThriftTest')
```

queries/thrift/highlights.scm
amaanq:
```thrift
struct A {
    1: required string faststr,
    2: required string string(pilota.rust_type = "string"),
}
```
pilota.rust_type is the annotation_identifier, I'm indifferent but if you think that's better off as an attribute then sure, but I think it's better for you to first see an example

ObserverOfTime: Shouldn't that be `@parameter` then?

queries/thrift/highlights.scm
amaanq: it's possible you're confusing the ident as enum Foo { ... } where Foo is said ident, that's not the case 

I labeled enum_member differently because they can be used when assigning consts


```thrift
enum Numberz {
  NAN,
  ONE = 1,
  TWO = 2,
  ...
}

const Numberz myNumberz = Numberz.ONE;
```

NAN, ONE, TWO are the enum_identifiers

in Numberz.ONE, ONE is the "enum_member", should've been named accessor maybe? or just enum_identifier too

queries/thrift/highlights.scm
amaanq: hmm, not really since it's not a function call, it's more "annotating" to the thrift compiler how certain fields should be dealt with in certain languages/scenarios

ObserverOfTime: So `@attribute` seems better after all.

amaanq: Thought it over, yes I agree. Gonna make a couple more changes, like fixing how enum_identifier and enum_member were handled, and rename this to attribute

queries/thrift/highlights.scm
```diff
@@ -36,19 +36,22 @@
...
 (enum_member) @constant
 
 ; Types
 
 (field_type) @type
 (type_identifier) @type
 
+; Attributes
+
+(annotation_identifier) @variable
```
ObserverOfTime, line 53: But you didn't change it. :eyes:

queries/thrift/highlights.scm
```diff
@@ -36,19 +36,22 @@
 ; Variables
 
 (identifier) @variable
-(annotation_identifier) @variable
+(enum_identifier) @variable
 
 ; Constants
 
 (const_identifier) @constant
-(enum_identifier) @constant
```
ObserverOfTime: I think this was fine as `@constant`.

queries/thrift/highlights.scm
amaanq: Whoops..fixing

queries/thrift/highlights.scm
amaanq: I redefined how they are parsed [here](https://github.com/amaanq/tree-sitter-thrift/commit/f4af461e2e40343e5b2fe628d41e8afe69a15944)

ObserverOfTime: So there is no `enum_identifier` now?

amaanq: enum Foo - Foo is now an enum_identifier instead of a generic type_identifier

```thrift
enum Numberz {
    One,
}

const Numberz myNumberz = Numberz.One;
```

Both One's are now enum_members which is more sound to me

ObserverOfTime: So…shouldn't it be `@type`?

amaanq: yep you're right, didn't think about that

amaanq: fixed

ObserverOfTime (approved):

ObserverOfTime:
Will be merged when the conflict is resolved.

clason:
and please rebase on master (don't merge master into the PR!)

amaanq:
I t hink I screwed up one of the old lockfiles..shouldn't matter anyways but my bad!

clason:
looks good!

amaanq:
Thanks a ton @ObserverOfTime!

PR #4199, opened 2023-01-22
injections(python): inject rst in docstrings

ObserverOfTime:
Includes an injection for the special comments supported by [Sphinx](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autofunction).

> For module data members and class attributes, documentation can either be put into a comment with special formatting (using a `#:` to start the comment instead of just `#`), or in a docstring after the definition.

---

Ideally, we would have a special parser that extends `rst` but this is good enough for now.
Here's how it looks on the different docstring styles (with the gruvbox theme):

**Sphinx** (decent)

image

**Google** (acceptable)

image

**NumPy** (deficient)

image

*Not sure why the `PEP 484` reference isn't even highlighted in this case.*

### DIFF
--- a/queries/python/injections.scm
+++ b/queries/python/injections.scm
@@ -7,5 +7,14 @@
   (#lua-match? @_string "^r.*")
   (#set! injection.language "regex"))
 
+((expression_statement
+   (string (string_content) @injection.content))
+ (#set! injection.language "rst"))
+
+((comment) @injection.content
+ (#lua-match? @injection.content "^#: ")
+ (#offset! @injection.content 0 3 0 0)
+ (#set! injection.language "rst"))
+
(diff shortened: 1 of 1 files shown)


theHamsta:
This injection was there before but got removed later as RST wasn't used in all code bases to add them there should be an option to configure them. There should be open issue or PRs with different approachea to fix the Python docstring configuration problem but none of them was merged yet if I remember correctly.

ObserverOfTime:
Then let's leave this as a draft for now.

XVilka:
Are there any plans to revive this PR?

clason:
If the injection is not universally welcome, it's out of scope here and should be added in a custom config or plugin.

PR #4203, opened 2023-01-22
feat: add capnp

amaanq:
This PR adds a capnp parser+queries

Parser: https://github.com/amaanq/tree-sitter-capnp
Reference: https://capnproto.org/language.html

cc: @marziply since they first brought it up in #2282 

There are some oddities with this language in general that I'd like some advice on as to how to define the queries for. Cap'n Proto fields have a numeric identifier as a type of "version" for that field, e.g. 

```capnp
struct Test {
  voidField    @0  : Void;
  boolField    @1  : Bool;
}
```

So voidField has a version of 0, boolField 1. you can also optionally add a ! to indicate that the number is an inline (immediate) value, as opposed to a reference to a value stored elsewhere in the data structure, for more efficient encoding and decoding of data. 

At first I thought of just using operator, but that wouldn't work and doesn't fit well. I also thought of string.special, but it's not a string at all so that also doesn't fit well. I'm not sure of the use case of `@symbol`, whether if it's just for a special char or if it even applies here. That and the import vs $import are also confusing, I put them both under @include but I wonder if I should just omit $ and instead declare it as `@punctuation.special`.

Thanks!
-Amaan

### DIFF
--- a/queries/capnp/highlights.scm
+++ b/queries/capnp/highlights.scm
@@ -1,0 +1,118 @@
+; Preproc (?)
+
+(unique_id) @preproc
+
+; Includes
+
+[
+  "import"
+  "$import"
+] @include
+
+(import_path) @string
+
+; Types
+
+(primitive_type) @type.builtin
+
+[
+  "annotation"
+  "enum"
+  "group"
+  "interface"
+  "List"
+  "struct"
+  "union"
+] @type.builtin
+
+; Methods
+
+(method_identifier) @method
+
+; Fields
+
+(field_identifier) @field
+
+; Parameters
+
+(param_identifier) @parameter
+(return_identifier) @parameter
+
+; Variables
+
+(identifier) @variable
+
+; Constants
+
+(const_identifier) @constant
+(enum_member) @constant
+
+; Types
+
+(enum_identifier) @type
+(extend_type) @type
+(field_type) @type
+(generic_identifier) @type
+(type_identifier) @type
+
+; Attributes
+
+(attribute) @attribute
+(annotation_identifier) @attribute
+
+; Operators
+
+[
+ ; @ ! -
+  "="
+] @operator
+
+; Keywords
+
+[
+  "const"
+  "extends"
+  "import"
+  "$import"
+  "namespace"
+  "using"
+] @keyword
+
+; Literals
+
+(string_literal) @string
+
+(data_string) @string.special
+(namespace) @string.special
+
+(number) @number
+
+(float) @float
+
+(boolean) @boolean
+
+; Misc
+
+[
+  "const"
+] @type.qualifier
+
+[
+  "*"
+] @punctuation.special
+
+["{" "}"] @punctuation.bracket
+
+["(" ")"] @punctuation.bracket
+
+["[" "]"] @punctuation.bracket
+
+[
+  ","
+  ";"
+  "->"
+] @punctuation.delimiter
+
+; Comments
+
+(comment) @comment @spell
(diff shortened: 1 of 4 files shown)


queries/capnp/highlights.scm
```diff
@@ -0,0 +1,118 @@
...
+; Keywords
+
+[
+  "const"
+  "extends"
+  "import"
+  "$import"
+  "namespace"
+  "using"
+] @keyword
```
theHamsta, line 79: Isn't `include` already `@include`

theHamsta:
I think the version attributes could be `@attribute`

queries/capnp/highlights.scm
amaanq: Not sure what you mean? Do you mean the "import" and "$import" being a duplicate kind of?

theHamsta: I wanted to say that "include" shouldn't be `@keyword` when it's already `@include`

amaanq:
> I think the version attributes could be `@attribute`

Hmm, not the biggest fan of that but that would at least fit without looking awkward, attributes imo feel like an optional note/hint that there's something different about whatever the attribute is covering. The `@number!` fields are always present

queries/capnp/highlights.scm
amaanq: fixed

theHamsta (approved):
This looks great to merge! I'm waiting to see whether someone has a good idea for the version tag/attribute

queries/capnp/highlights.scm
```diff
@@ -0,0 +1,116 @@
...
+
+[
+ ; @ ! -
+  "="
+] @operator
+
+; Keywords
+
+[
+  "const"
```
ObserverOfTime, line 73: Duplicated in `@type.qualifier`.

ObserverOfTime (commented):
> I'm waiting to see whether someone has a good idea for the version tag/attribute

How about `@label` or `@namespace`?

---

The `T` in the following example should be `@type.definition` if possible:

```cpp
using T = Foo.Bar;
```

queries/capnp/highlights.scm
```diff
@@ -0,0 +1,116 @@
...
+  "namespace"
+  "using"
+] @keyword
+
+; Literals
+
+(string_literal) @string
+
+(data_string) @string.special
+(namespace) @string.special
```
ObserverOfTime, line 84: Why not `@namespace`?

amaanq: Cap'n Proto namespace declarations are like the following

```capnp
$Cxx.namespace("capnproto_test::capnp::test");

$import "/capnp/c++.capnp".namespace("capnp");
```

I could make the highlight a namespace, but that'd look awkward as it's quoted

queries/capnp/highlights.scm
amaanq: > > I'm waiting to see whether someone has a good idea for the version tag/attribute
> 
> How about `@label`?

Yeah, I like label, thanks!

amaanq: fixed const

amaanq:
Could a using directive be `@type.definition`?

i.e.
```capnp
using OuterNestedEnum = TestNestedTypes.NestedEnum;
````

OuterNestedEnum could be `@type.definition`

amaanq:
Need to fix one bug, draft until then

amaanq:
Good now

ObserverOfTime (approved):

amaanq:
If the namespace query looks good to you @ObserverOfTime, it's ready to merge :)

PR #4205, opened 2023-01-23
feat(thrift): refactor queries for better highlights

amaanq:
Fixes bugs where certain ones went past where they should've ended


amaanq:
Initial issue stemmed from me being pretty new to highlights, but after some time with this and capnp I think I have a more solid understanding now, thanks :)

theHamsta (approved):

PR #4206, opened 2023-01-23
highlights(php): declare directives and `:`

zweimach:
Is `@preproc` the right capture for this?

| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/queries/php/highlights.scm
+++ b/queries/php/highlights.scm
@@ -110,9 +110,7 @@
 (conditional_expression) @conditional
 
 ; Directives
-(declare_directive "strict_types" @preproc)
-(declare_directive "ticks" @preproc)
-(declare_directive "encoding" @preproc)
+(declare_directive ["strict_types" "ticks" "encoding"] @preproc)
 
 ; Basic tokens
 
--- a/queries/php/highlights.scm
+++ b/queries/php/highlights.scm
@@ -110,7 +110,7 @@
 (conditional_expression) @conditional
 
 ; Directives
-(declare_directive ["strict_types" "ticks" "encoding"] @preproc)
+(declare_directive ["strict_types" "ticks" "encoding"] @parameter)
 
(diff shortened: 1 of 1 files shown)


queries/php/highlights.scm
```diff
@@ -108,6 +108,12 @@
 
 ; Conditions ( ? : )
 (conditional_expression) @conditional
+
+; Directives
+(declare_directive "strict_types" @preproc)
+(declare_directive "ticks" @preproc)
+(declare_directive "encoding" @preproc)
```
lucario387, line 115:
```suggestion
(declare_directive ["strict_types" "ticks" "encoding"] @preproc)
```

lucario387: And are there other keywords for declare?
If there is then maybe you should do something else

zweimach: I didn't know this syntax, thanks!

zweimach: There are only these three directives in [the parser](https://github.com/tree-sitter/tree-sitter-php/blob/973694ffcdeebca245b7ecf0d7c4cadd4f41b3c9/grammar.js#L566).

gbprod:
IMHO, `@parameter` would be more suitable (like for named arguments) as it's not really a preprocessor directive.
What do you think ?

zweimach:
> IMHO, `@parameter` would be more suitable (like for named arguments) as it's not really a preprocessor directive.

Yes, it could be.

I picked `@preproc` because IMO directives are similar to `#pragma` in C/C++.
What do you think @tk-shirasaka?

tk-shirasaka:
I agree with `@parameter`.

I think that if the syntax is expanded, it is easy to be interpreted differently by each person.
(I also thought `ini_set` was semantically close).

theHamsta (approved):

PR #4212, opened 2023-01-23
map quarto ft to markdown parser

jmbuhr:
Adds the [quarto](https://quarto.org/) filetype to use the markdown parser. When the filetype was very new we had just changed the filetype from quarto to markdown, but now it is time to make use of it :)

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -15,6 +15,7 @@ local filetype_to_parsername = {
   OpenFOAM = "foam",
   pandoc = "markdown",
   rmd = "markdown",
+  quarto = "markdown",
   cs = "c_sharp",
   tape = "vhs",
   dosini = "ini",


theHamsta:
I looks like it would not be very difficult to generate a derived parser from markdown that captures the peculiarities of Quatro (the data in the front matter), curly braces around the language in magic code blocks. Interested in creating such a parser? Or is Quatro just markdown?

theHamsta (approved):

jmbuhr:
Quarto is based on pandoc and pandoc's markdown, with some minor things added that shouldn't bother plain pandoc markdown, so deriving a new parser is definitely the way to go and would help not just quarto users! So far there is only a regular regex-bases syntax file for this: https://github.com/vim-pandoc/vim-pandoc-syntax and TS is the future.

I most likely will not have much time in the coming months, so if somebody does this first I wouldn't be mad ;) 
I have no experience writing parsers, but if no one does so I will take a stab at it once I find the time (today I got stuck at renaming instances of "markdown" to "quarto" and then not being able to install the parser)

theHamsta:
It won't be necessary to write a new parser. Tree-sitter let's you extend rules of an existing parser. So you just can add stuff that quarto has in addition to the normal parser. E.g. JSON with comments references the original JSON parser, but extends its extras (rules that are allowed in any position of the grammar) https://gitlab.com/WhyNotHugo/tree-sitter-jsonc/-/blob/main/grammar.js

So you could just add a few additional rules to the existing markdown parser. A problem with this extension mechanism is that it does not work for `src/scanner.cc`. So you would need to copy that file and rename `tree_sitter_markdown_*` to `tree_sitter_<new grammar name>_*`

jmbuhr:
Thanks that's a great starting point! Does the same principal of extension work if I then add queries to nvim-treesiter?

theHamsta:
Extending of queries is not a feature of tree-sitter. But this syntax works in Neovim and Helix

https://github.com/nvim-treesitter/nvim-treesitter/blob/7bef1d53302f1087a07f83a4321582d835f44d4f/queries/cpp/highlights.scm#L1

https://github.com/nvim-treesitter/nvim-treesitter/blob/7bef1d53302f1087a07f83a4321582d835f44d4f/queries/jsonc/highlights.scm#L1

PR #4214, opened 2023-01-23
refactor(capnp, thrift): improve grammar & queries to stable for the foreseeable future

amaanq:
Apologies for the rapid flurry of PR's, I hadn't been able to get local highlights to reflect their change on demand after a parser/query update. Finally figured that out today and I've gotten a lot more improvements done as a result.

Thanks!
-Amaan

### DIFF
--- a/queries/capnp/highlights.scm
+++ b/queries/capnp/highlights.scm
@@ -62,6 +62,8 @@
 (local_const) @constant
 (enum_member) @constant
 
+(void) @constant.builtin
+
 ; Types
 
 (enum_identifier) @type
@@ -88,7 +90,6 @@
   "extends"
   "namespace"
   "using"
-  (void)
   (annotation_target)
 ] @keyword
 
--- a/queries/capnp/highlights.scm
+++ b/queries/capnp/highlights.scm
@@ -1,6 +1,7 @@
 ; Preproc (?)
 
 (unique_id) @preproc
+(lang_attrs) @preproc
 
 ; Includes
 
@@ -21,10 +22,9 @@
   "enum"
   "group"
   "interface"
-  "List"
   "struct"
   "union"
-] @type.builtin
+] @keyword
 
 ; Typedefs
 
@@ -36,11 +36,16 @@
 
 ; Methods
 
+(annotation_definition_identifier) @method
 (method_identifier) @method
 
 ; Fields
 
 (field_identifier) @field
+
+; Properties
+
+(property) @property
 
 ; Parameters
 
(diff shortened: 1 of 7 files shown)


queries/capnp/highlights.scm
```diff
@@ -82,11 +88,14 @@
   "extends"
   "namespace"
   "using"
+  (void)
```
theHamsta, line 91: `@type.builtin` or `@keyword`? I don't know `void` in the context of capnp

queries/capnp/highlights.scm
```diff
@@ -119,6 +129,8 @@
   "->"
 ] @punctuation.delimiter
 
+(data_hex) @symbol
```
theHamsta, line 132: data doesn't sound like `@symbol`. `@symbol` is rather something like a named constant in the field of functional languages

queries/thrift/highlights.scm
```diff
@@ -47,22 +47,24 @@
...
 (exception_identifier) @type
 (exception_param_type) @type
 (field_type) @type
+(param_type) @type
 (type_identifier) @type
 
 ; Attributes
 
 (annotation_identifier) @attribute
+(uri_def) @attribute
```
theHamsta, line 60: is uri here like an annotation to something?

queries/capnp/highlights.scm
amaanq: Capnp datas are unique, e.g.
```capnp
  dataField      @13 : Data    = 0x"62 61 72"; # "bar"
```

this is how you would write raw binary types, like b"" in python, so this *specific* leading 0x imo should be a special kind of highlight. what do you think?

queries/thrift/highlights.scm
amaanq: it's used in special namespace defs, like so
```thrift
namespace xsd test (uri = 'http://thrift.apache.org/ns/OneWayTest')
```

queries/capnp/highlights.scm
amaanq: it's actually @constant.builtin, just fixed it in latest commit!
Void is the type, void the value :)

queries/capnp/highlights.scm
theHamsta: @number ?

amaanq: hmm, @number *just* for the 0x prefix is weird..but if thats convention sure, let me know

amaanq: @string.special works for me, let me update

amaanq: nevermind, i have that defined for the inner data string, it'd be weird to have them both be that

PR #4220, opened 2023-01-25
Update syntax highlighting for hcl

nocode99:
cc @MichaHoffmann 

First off - appreciate all the work from everyone going into this plugin! I was recently messing around with LunarVim and liked the syntax-highlighting it was providing for terraform. Through the help of my colleague @pappasam - he helped me understand how syntax highlighting works. From there, I was able to determine that LunarVim was shipping an old version of the tree-sitter parsers.

I'm hoping we can bring back this particular change because it's visually much more pleasing. I'm still learning tree-sitter parser syntax so not sure if this is the best approach and also know this is subjective. Screenshots with before/after.

## BEFORE

image

## AFTER

image

### DIFF
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -93,7 +93,7 @@
 
 ; var.foo, data.bar
 ;
-; first element in get_attr is a keyword or a reference to a keyword
+; first element in get_attr is a variable.builtin or a reference to a variable.builtin
 (expression (variable_expr (identifier) @variable.builtin) (get_attr (identifier) @field))
 
 (ERROR) @error
--- a/queries/hcl/highlights.scm
+++ b/queries/hcl/highlights.scm
@@ -50,7 +50,7 @@
   "in"
 ] @repeat
 
-[ 
+[
   "if"
   "else"
   "endif"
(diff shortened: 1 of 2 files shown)


queries/hcl/highlights.scm
```diff
@@ -94,6 +94,6 @@
 ; var.foo, data.bar
 ;
 ; first element in get_attr is a keyword or a reference to a keyword
-(expression (variable_expr (identifier) @keyword) (get_attr (identifier) @field))
```
theHamsta, line 97: if this change is necessary, then the comment explaining the keyword needs to be updated.

nocode99: Ah yes - thanks for catching that. Fixed.

nocode99:
It does seem we would also need to adjust the `terraform` highlights for conditionals. There's still one I need to figure out:

image

MichaHoffmann:
> It does seem we would also need to adjust the `terraform` highlights for conditionals. There's still one I need to figure out:
> 
> image

I wonder if its fixed if we remove the outmost `expression` wrapper; both should be `variable_expr`

MichaHoffmann (approved):
lgtm

theHamsta (approved):

Issue #3058, opened 2022-06-23
Basic Highlight Overriding


barrett-ruth:
### Describe the highlighting problem

All I want to do is override a highlight group. I have tried many times and cannot figure it out.

Lua keyword `local` is highlighted according to TSKeyword using `:TSHighlightCapturesUnderCursor`, the correct behavior.

I would simply like to un-italicize the word `local` while keeping all other TSKeywords the same.

I have tried with `set_query`, but that overrides all the other `highlights` queries. I have also used `set_custom_captures`, but this is overridden by `TSKeyword`. The only way I could fix the issue is removing "local" from the lua queries file in the source code itself.

### Example snippet that causes the problem

```lua
local hi = 'hello'
```

### Tree-sitter parsing result

```
local_declaration: variable_declaration [0, 0] - [0, 18]
  "local" [0, 0] - [0, 5]
  assignment_statement [0, 6] - [0, 18]
    variable_list [0, 6] - [0, 8]
      name: identifier [0, 6] - [0, 8]
    "=" [0, 9] - [0, 10]
    expression_list [0, 11] - [0, 18]
      value: string [0, 11] - [0, 18]
        start: "string_start" [0, 11] - [0, 12]
        content: "string_content" [0, 12] - [0, 17]
        end: "string_end" [0, 17] - [0, 18]
```

### Example screenshot

Device doesn't have screenshotting set up.
sainnhe/gruvbox-material

(don't think this is relevant)

### Expected behavior

Change the highlighting of keyword `local`

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
  - OK: `node` found v18.3.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 12.1.0
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - c              ✓ ✓ ✓ ✓ ✓ 
  - html           ✓ ✓ ✓ ✓ ✓ 
  - query          ✓ ✓ ✓ ✓ ✓ 
[...]

PR #4227, opened 2023-01-26
feat: add kdl

amaanq:
https://github.com/kdl-org/kdl/blob/main/SPEC.md

Grammar/Parser: https://github.com/amaanq/tree-sitter-kdl

~~Not sure why, but whoever requested it in #2282 seems to have deleted their comment, which is unfortunate.~~
cc: @MultisampledNight since they requested it first in #2282
(dunno why I couldn't find the comment before, oops)

Only thing missing is having the special slashdash comments actually be commented out. Still working on that as this was a pain to begin with.

I would also like someone to go over my `@number/@float` queries, float is embedded into number so I'm not sure if I separated them properly.

Thanks!
-Amaan

### DIFF
--- a/queries/kdl/folds.scm
+++ b/queries/kdl/folds.scm
@@ -1,3 +1,3 @@
 ; Folds
 
-(node children: (node_children) @fold)
+(node_children) @fold
--- a/queries/kdl/folds.scm
+++ b/queries/kdl/folds.scm
@@ -1,0 +1,3 @@
+; Folds
+
+(node children: (node_children) @fold)
--- a/README.md
+++ b/README.md
@@ -245,6 +245,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [JSON with comments](https://gitlab.com/WhyNotHugo/tree-sitter-jsonc.git) (maintained by @WhyNotHugo)
 - [x] [jsonnet](https://github.com/sourcegraph/tree-sitter-jsonnet) (maintained by @nawordar)
 - [x] [julia](https://github.com/tree-sitter/tree-sitter-julia) (maintained by @theHamsta)
+- [x] [kdl](https://github.com/amaanq/tree-sitter-kdl) (experimental, maintained by @amaanq)
 - [x] [kotlin](https://github.com/fwcd/tree-sitter-kotlin) (maintained by @SalBakraa)
 - [x] [lalrpop](https://github.com/traxys/tree-sitter-lalrpop) (maintained by @traxys)
 - [x] [latex](https://github.com/latex-lsp/tree-sitter-latex) (maintained by @theHamsta, @clason)
--- a/lockfile.json
+++ b/lockfile.json
@@ -218,6 +218,9 @@
   "julia": {
     "revision": "9d368185be7e8139f2eda93d8b0acc2a54031718"
   },
+  "kdl": {
+    "revision": "c168689e3fea6c3cb99bfb7c930932b276fcce0a"
+  },
   "kotlin": {
     "revision": "e4637037a5fe6f25fe66c305669faa0855f35692"
   },
--- a/lockfile.json
+++ b/lockfile.json
@@ -219,7 +219,7 @@
     "revision": "9d368185be7e8139f2eda93d8b0acc2a54031718"
   },
   "kdl": {
-    "revision": "cb23b8dbe15005efa083df53fe07506b9421896c"
+    "revision": "d79b87b0a4814d000268e1f8a1993070f8ab6584"
   },
   "kotlin": {
     "revision": "e4637037a5fe6f25fe66c305669faa0855f35692"
--- a/lockfile.json
+++ b/lockfile.json
@@ -219,7 +219,7 @@
     "revision": "9d368185be7e8139f2eda93d8b0acc2a54031718"
   },
   "kdl": {
-    "revision": "50cf459513ef4377592a55c421f3fc40ce9b58a0"
+    "revision": "cb23b8dbe15005efa083df53fe07506b9421896c"
   },
   "kotlin": {
     "revision": "e4637037a5fe6f25fe66c305669faa0855f35692"
--- a/lockfile.json
+++ b/lockfile.json
@@ -219,7 +219,7 @@
     "revision": "9d368185be7e8139f2eda93d8b0acc2a54031718"
   },
   "kdl": {
-    "revision": "c168689e3fea6c3cb99bfb7c930932b276fcce0a"
+    "revision": "b59c704ca4ea65b78ddf52bf91f53f43622e58fa"
   },
   "kotlin": {
     "revision": "e4637037a5fe6f25fe66c305669faa0855f35692"
--- a/lockfile.json
+++ b/lockfile.json
@@ -219,7 +219,7 @@
     "revision": "9d368185be7e8139f2eda93d8b0acc2a54031718"
   },
   "kdl": {
-    "revision": "b59c704ca4ea65b78ddf52bf91f53f43622e58fa"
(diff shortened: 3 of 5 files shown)


amaanq:
Force pushed since I added readme

amaanq:
Ok, I'm working out comments and might've run into a bug. Draft until fixed

amaanq:
fixed the bug, now it's just comments that are the issue. here's several types of comments with the slash dash (/-) in kdl

image

the first one should have the whole block commented out, i.e. till the }. In fact, showing captures under cursor does show comment is there, but other highlights are overriding it, how can we either a) set conditionals on those to not highlight on comment, or b) override them with comment

the second one should have only /-key=1 commented, but "foo" got captured too since props or values after a /- can be commented, how can we get it to only pick 1 or the other? and same issue with =1 where other highlights take over

third one same as second

the last one should have everything from { } (inclusive) commented, captures does show its a comment, but again other hls take over

Thanks to anyone that can help me out!

MultisampledNight:
<details>
<summary>(fwiw, here is an equivalent text version of that file)</summary>

```kdl
// entire node stuff commented out
/-this "stuff" a=1 {
  node
  in
  a
  node
}

// property is commented out
also /-a=1 "stuff" {
  node
  in
  a
  node
}

// name is commented out
but /-"stuff" a=1 {
  node
  in
  a
  node
}

// children are commented out
actually "stuff" a=1 /- {
  node
  in
  a
  node
}
```

<3

</summary>

MultisampledNight:
Found that anonymous nodes might be interesting. The first case works slightly better with this query:

```
(node (node_slash_dash) (_)? @comment) @comment
```

(question mark only added for purity, has no effect here)

unknown

MultisampledNight:
...wow, okay, nevermind. That gives really inconsistent results I don't quite grasp.

MultisampledNight:
Seems like it's possible to configure priorities for queries, found that through https://github.com/nvim-treesitter/nvim-treesitter/issues/3058:

```
(node (node_slash_dash) (set! "priority" 105)) @comment
```

unknown

theHamsta:
You seem to be using a scanner.c. Even though, you're not doing something complicated there. You could use https://github.com/vigoux/tree-sitter-fuzz-action in your repo

amaanq:
> You seem to be using a scanner.c. Even though, you're not doing something complicated there. You could use https://github.com/vigoux/tree-sitter-fuzz-action in your repo

I had an issue detecting EOFs without just checking for 0 beforehand in the scanner. This helped alleviate a bug where tree-sitter would parse indefinitely at the end.

theHamsta (approved):

amaanq:
> Seems like it's possible to configure priorities for queries, found that through https://github.com/nvim-treesitter/nvim-treesitter/issues/3058 (after looking a bit more around, seems like that's been added in https://github.com/neovim/neovim/pull/15114 and is documented in `treesitter.txt` under `treesitter-highlight-priority`):
> 
> ```
> (node (node_slash_dash) (#set! "priority" 105)) @comment
> ```
> 
> unknown
> 

have you had success with this? I'll play around with it soon so I can update the PR asap

MultisampledNight:
Partially. It worked fine for the whole node and children slashdash comments, but not for the individual props/args (I guess `value` needs some further refinement since `prop` also contains one).

This is what I've got:

```diff
diff --git a/queries/kdl/highlights.scm b/queries/kdl/highlights.scm
index e83f4731..79dc53e8 100644
--- a/queries/kdl/highlights.scm
+++ b/queries/kdl/highlights.scm
@@ -50,7 +50,7 @@
   (multi_line_comment)
 ] @comment
 
-(node (node_slash_dash)) @comment
+(node (node_slash_dash) (#set! "priority" 105)) @comment
 (node (node_prop_or_arg_slash_dash) @comment (prop) @comment)
 (node (node_prop_or_arg_slash_dash) @comment (value) @comment)
-(node_children (node_children_slash_dash)) @comment
+(node_children (node_children_slash_dash) (#set! "priority" 105)) @comment

```

amaanq:
Ok, I worked around KDL's IDL a bit and got it to work.. I guess rules are meant to be bent a bit

Here's sample output on my end now that I'll update v shortly
image

amaanq:
unsure about folds since I've always been confused on whether or not say a fold marked @fold

foo {
  .....
}

should include or exclude the {} itself

Otherwise this is good to squash n' merge now @theHamsta

MultisampledNight:
@amaanq I found `:help treesitter-highlight-priority`, which says (shortened):

> Treesitter uses `nvim_buf_set_extmark()` to set highlights with a default priority of 100. This enables plugins to set a highlighting priority lower or higher than tree-sitter.

As such, I believe higher means more important, lower means less important, and all other rules had a priority equal to the default of 100.t

amaanq:
> @amaanq I found `:help treesitter-highlight-priority`, which says (shortened):
> 
> > Treesitter uses `nvim_buf_set_extmark()` to set highlights with a default priority of 100. This enables plugins to set a highlighting priority lower or higher than tree-sitter.
> 
> As such, I believe higher means more important, lower means less important, and all other rules had a priority equal to the default of 100.

Thanks! What do you think of the folds? I don't use folds personally so it'd be nice to hear from you

MultisampledNight:
I don't use them either, personally. But I'll set them up and test.

amaanq:
more of a TODO, but I'm thinking (maybe for the future I'm still relatively new to this) that we have uncommented stuff in a capture without comments, since anything commented should not have the normal highlight to begin with, then commented stuff is a pure comment capture, and then add injections to comments (I think that's their purpose)

MultisampledNight:
It works okay, apart from a bit of weirdness I don't think you can do something about. Take this for example:

```kdl
but /-"stuff" a=1 {
  in
  a
  node {
    nestification {
      wow "a"
    }
  }
}
```

- `za` on `node` closes the children of `but`.
- `za` on `nestification`/`wow` closes the children of `nestification`, but leaves over a closing `}`.
- `za` on the closing `}` of `nestification` closes the children of `nestification`, with the `}` this time.
- Doesn't seem like I can close `node` only.

amaanq:
Thanks to @ahlinc [here](https://github.com/amaanq/tree-sitter-kdl/pull/2), improvements have been made but as a result there's still a bug or two to be fixed when using kdl's examples instead of their tests. Namely this fails:

```kdl
// Nodes can be separated into multiple lines
title \
  "Some title"


// Files must be utf8 encoded!
smile "😁"

// Instead of anonymous nodes, nodes and properties can be wrapped
// in "" for arbitrary node names.
"!@#$@$%Q#$%~@!40" "1.2.3" "!!!!!"=true

// The following is a legal bare identifier:
foo123~!@#$%^&*.:'|?+ "weeee"

// And you can also use unicode!
ノード　お名前="☜(ﾟヮﾟ☜)"

// kdl specifically allows properties and values to be
// interspersed with each other, much like CLI commands.
foo bar=true "baz" quux=false 1 2 3
```

Draft until fixed

amaanq:
Good to merge now, did a lot of thorough testing

queries/kdl/folds.scm
```diff
@@ -0,0 +1,3 @@
+; Folds
+
+(node children: (node_children) @fold)
```
theHamsta, line 3: Just `(node_children) @fold` would e equivalent?

theHamsta (approved):

queries/kdl/folds.scm
amaanq: yep you're right, done

amaanq:
@theHamsta do you think this is good to merge now? Or just waiting for someone elses input?

theHamsta:
Thank you!

amaanq:
Thank you too! Glad to be a contributor :smile:

PR #4237, opened 2023-01-28
highlights(php): fix queries related to namespaces

zweimach:
This PR could have been a discussion thread, but I wanted to show the code.

This PR refines the queries related to namespaces and scoping.
Mainly for consistency with other languages' namespace-related highlights.

| Before | After |
| ------ | ----- |
| image | image |
| image | image |
| image | image |

Comparison with other languages:

| Language | Screenshot |
| -------- | ---------- |
| PHP (after this PR) | image |
| Rust | image |
| C++ | image |
| Elm | image |

While at it, I also fixed queries related to `@constant`, `@function.call`, and captures for the `as` keyword:
| Before | After |
| ------ | ----- |
| image | image |
| image | image |
| image | image |

I will tag everyone involved in my previous PR: @tk-shirasaka @Lucario387 @gbprod @theHamsta.
What do you think?

### DIFF
--- a/queries/php/highlights.scm
+++ b/queries/php/highlights.scm
@@ -32,7 +32,7 @@
 (namespace_use_clause
   [(name) @type
    (qualified_name (name) @type)])
-(namespace_aliasing_clause (name)) @type.definition
+(namespace_aliasing_clause (name) @type.definition)
 (class_interface_clause
   [(name) @type
    (qualified_name (name) @type)])
--- a/queries/php/highlights.scm
+++ b/queries/php/highlights.scm
@@ -5,9 +5,9 @@
 ; Constants
 
 ((name) @constant
- (#vim-match? @constant "^_?[A-Z][A-Z\d_]*$"))
+ (#lua-match? @constant "^_?[A-Z][A-Z%d_]*$"))
 ((name) @constant.builtin
- (#vim-match? @constant.builtin "^__[A-Z][A-Z\d_]+__$"))
+ (#lua-match? @constant.builtin "^__[A-Z][A-Z%d_]+__$"))
 
 (const_declaration (const_element (name) @constant))
 
--- a/queries/php/highlights.scm
+++ b/queries/php/highlights.scm
@@ -8,25 +8,31 @@
  (primitive_type)
  (cast_type)
  ] @type.builtin
-(named_type (name)) @type
-(named_type (qualified_name)) @type
+(named_type
+  [(name) @type
+   (qualified_name (name) @type)])
 (class_declaration
   name: (name) @type)
 (base_clause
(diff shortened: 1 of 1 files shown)


queries/php/highlights.scm
```diff
@@ -2,39 +2,55 @@
 
 (variable_name) @variable
 
+; Constants
+
+((name) @constant
+ (#vim-match? @constant "^_?[A-Z][A-Z\d_]*$"))
+((name) @constant.builtin
+ (#vim-match? @constant.builtin "^__[A-Z][A-Z\d_]+__$"))
```
theHamsta, line 10: looks like this could also be handled by the faster lua match?

theHamsta (approved):

queries/php/highlights.scm
```diff
@@ -2,39 +2,55 @@
...
 (enum_declaration
   name: (name) @type)
 (interface_declaration
   name: (name) @type)
 (namespace_use_clause
-  [(name) (qualified_name)] @type)
-(namespace_aliasing_clause (name)) @type.definition
+  [(name) @type
+   (qualified_name (name) @type)])
+(namespace_aliasing_clause (name) @type.definition)
```
theHamsta, line 35: wasn't aware that we actually use this weird capture

queries/php/highlights.scm
zweimach: Thanks, I have added your suggestion.

gbprod:
Not sure it's interesting here but I really like the way vscode hightlight php use : 

2023-01-30-155147_439x51_scrot

That's why I add in my custom highlights this code to have the same colors: 

```scheme
(namespace_name_as_prefix) @namespace
```

2023-01-30-155416_527x300_scrot

2023-01-30-155430_468x199_scrot

Do you think it could be the default behavior ?

lucario387:
> Do you think it could be the default behavior ?

This make it clearer to the end user, I really like it, just like how semantic tokens in Jdtls allow me to somewhat know whether I import classes, or annotations or other node types (of course its not necessary, but syntax sugar is beautiful/looks better no?)

image

tk-shirasaka:
I think it is generally good, but I was curious about the handling of aliases.
In other languages, aliases seem to be assigned the same highlight as the origin, so I thought it would be better to follow this practice.

| Language  | Screenshot | Screenshot (use alias) |
|:------------:|:-------------:|:------------------------:|
| PHP           | image | image |
| Python      | image | image |
| Typescript | image | image |

zweimach:
I actually agree with @tk-shirasaka.
@theHamsta What do you think? Does `@type.definition` includes import renaming too?
But that is out of the scope of this PR. It should be in a separate PR.

theHamsta (approved):

PR #4239, opened 2023-01-28
chore: update thrift queries

amaanq:
I recently added support for a large extension of thrift, this PR addresses the query updates for that

### DIFF
--- a/queries/thrift/highlights.scm
+++ b/queries/thrift/highlights.scm
@@ -1,24 +1,27 @@
-; Namespace
+ ; Namespace
 
 (namespace_definition) @namespace
 
 ; Includes
 
-[
   "include"
+[
   "cpp_include"
 ] @include
 
 (include_path) @string
+(package_path) @string
 
 ; Types
 
-(primitive_type) @type.builtin
+(primitive) @type.builtin
 
 [
   "list"
   "map"
   "set"
+  "sink"
+  "stream"
   "void"
 ] @type.builtin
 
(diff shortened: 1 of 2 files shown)


amaanq:
always forgetting lockfile change... :slightly_smiling_face:

amaanq:
amended for v0.3.0 being published

queries/thrift/highlights.scm
```diff
@@ -57,6 +60,9 @@
 ; Attributes
 
 (annotation_identifier) @attribute
+(exception_modifier) @attribute
+(field_modifier) @attribute
+(function_modifier) @attribute
```
theHamsta, line 65: are modifiers annotations? or `@type.qualifier`?

theHamsta (approved):

queries/thrift/highlights.scm
amaanq: not entirely sure what `@type.qualifier` is, but they are keywords before a function like async, before a field like required, optional

theHamsta: yes, then not `@attribute`. Attribute is for annotations that are not actually necessary in the code but can be added for additional meta info the compiler/refection or alike (originally for C++ attributes) or those Go serialization tags. `@type.qualifier` and `@storageclass` are typically used for stuff like `const`, `async`, `volatile` or just `@keyword`.

amaanq: gotcha, thanks! fixed now

theHamsta (approved):

PR #4243, opened 2023-01-30
feat(ecma): add nospell to unrecognized keywords

saurabhchardereal:
Since these are not English words per se.

Screenshot-30-01-2023-03-01-54

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -274,3 +274,12 @@
   "default" @keyword)
 (switch_default
   "default" @conditional)
+
+; Spell
+;----------
+[
+  "const"
+  "instanceof"
+  "typeof"
+  "async"
+] @nospell


theHamsta (approved):

lewis6991:
Only comments and strings are spell checked, so this change should be unnecessary.

saurabhchardereal:
> Only comments and strings are spell checked, so this change should be unnecessary.

Weird. I checked whether my [treesitter](https://github.com/saurabhchardereal/dotfiles/blob/main/config/nvim/lua/plugins/nvim-treesitter.lua) config is responsible for this or not but I can't seem to find what's causing it to highlight. Is there a way to debug which queries are getting applied?

lewis6991:
If you are running 0.9dev, try `:Inspect` with the cursor over the misspelled words.

saurabhchardereal:
> If you are running 0.9dev, try `:Inspect` with the cursor over the misspelled words.

This is what it showed for `const` inside comment and `const` keyword:

lewis6991:
Can you make sure your `spelloptions` contains `noplainbuffer`. This should be set automatically, but there could be something messing with this (likely your config).

saurabhchardereal:
Ok. So, after few hours of debugging I traced it back to [emmet-vim](https://github.com/mattn/emmet-vim) which is odd since it doesn't play with any queries at all. Maybe it was somehow conflicting with spell options? I dunno since I'm not really good at vimscript. I guess it's finally time I should drop that plugin. 

Thanks for the quick response though. 👍

PR #4246, opened 2023-01-30
fix(ql): Add moduleMember to indent file

pwntester:


queries/ql/indents.scm
```diff
@@ -1,4 +1,5 @@
 [
+  (moduleMember)
```
theHamsta, line 2: could you add a test for this?

pwntester: I did, thanks!

pwntester: Actually, thanks to the tests I realize that the indentation is not yet ok. Will fix and add more tests

pwntester: Ok, seems like its working now! BTW, I tried running the `scripts/run_tests.sh` script but it seems to be missing the `plugin/plenary.vim` file. Is that correct?

theHamsta (approved):

tests/indent/ql/qlpack.yml
```diff
@@ -0,0 +1,7 @@
+---
+library: false
+name: test/test
+version: 0.0.1
+dependencies:
+  codeql/go-all: '*'
```
theHamsta, line 6: what does this file do?

pwntester: That file tells the CodeQL interpreter where to get the libraries so it does not throw any errors on the `import go` statement. But I can simplify it and remove this file and the import line

theHamsta (approved):

Issue #4258, opened 2023-02-02
incremental_selection error if some empty lines in the end of file


haoxianhan:
### Describe the highlighting problem

if some empty lines in the end of file, would throw error messages when incremental_selection to the whole file.

https://user-images.githubusercontent.com/3611408/216228818-649a1a2c-56b3-429b-ba11-bef3f8d51ed9.mp4


### Example snippet that causes the problem

``` bash
#!/bin/sh

hello() {
    echo "hello"
}

main

# next some lines are empty


```

### Tree-sitter parsing result

```
comment [0, 0] - [0, 9]
function_definition [2, 0] - [4, 1]
  name: word [2, 0] - [2, 5]
  body: compound_statement [2, 8] - [4, 1]
    command [3, 4] - [3, 16]
      name: command_name [3, 4] - [3, 8]
        word [3, 4] - [3, 8]
      argument: string [3, 9] - [3, 16]
command [6, 0] - [6, 4]
  name: command_name [6, 0] - [6, 4]
    word [6, 0] - [6, 4]
```

### Example screenshot

![image](https://user-images.githubusercontent.com/3611408/216227597-91c7e77b-cc6e-4b17-8a5e-6e3342676ce3.png)


### Expected behavior

_No response_

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.7 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v17.9.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: cc (GCC) 12.2.1 20230111
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## OS Info:
{
  machine = "x86_64",
  release = "6.1.8-arch1-1",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Tue, 24 Jan 2023 21:07:04 +0000"
}

## Parser/Features         H L F I J
  - yaml                ✓ ✓ ✓ ✓ ✓
  - bash                ✓ ✓ ✓ . ✓
  - erlang              ✓ . ✓ . .
[...]

PR #4260, opened 2023-02-02
fix: incremental_selection error

haoxianhan:
if some empty lines in the end of file,
would throw error messages when incremental_selection to the whole file.

fix issue: #4258


kiyoon:
@theHamsta This is good. This also fixes https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/377.

This is caused when we change `vim.fn.setpos()` to `vim.api.nvim_win_set_cursor()` in #4228.  
I noticed that the former is (1, 1)-indexed and the latter is (1,0)-indexed.  

In `vim.fn.setpos()`, passing (row, 0) and (row, 1) is the same without error. It goes to the beginning of the line. However, in `vim.api.nvim_win_set_cursor()`, it is more strict and it will raise an error. It's good that we prevent this confusion from earlier on before the `update_selection` function.

theHamsta (approved):

PR #4264, opened 2023-02-02
chore: organize languages in parsers.lua

amaanq:
Title


amaanq:
cc: @ObserverOfTime

clason:
Probably not worth it; let's just remind people when contributing new languages.

PR #4266, opened 2023-02-03
feat(julia/highlights): Add `as` keyword

savq:


clason (approved):

amaanq:
Would it not be `@keyword`?

savq:
`as` can only be used inside `import` statements, and all the other import related keywords use `@include` so I think it's fine.

clason:
This is similar to `then` being captured as `@conditional` in Lua's `if foo then bar end`.

amaanq:
> `as` can only be used inside `import` statements, and all the other import related keywords use `@include` so I think it's fine.

gotcha, just wasn't sure if it was only used here

theHamsta (approved):

PR #4270, opened 2023-02-03
feat: add gosum

amaanq:
https://go.dev/ref/mod#go-sum-files

### DIFF
--- a/queries/gosum/highlights.scm
+++ b/queries/gosum/highlights.scm
@@ -1,0 +1,33 @@
+[
+  "alpha"
+  "beta"
+  "dev"
+  "pre"
+  "rc"
+  "+incompatible"
+] @keyword
+
+[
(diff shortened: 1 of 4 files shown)


queries/gosum/highlights.scm
```diff
@@ -0,0 +1,35 @@
...
+[
+ (number)
+ (number_with_decimal)
+ (hex_number)
+] @number
+
+(checksum
+  (module_path)
+  (version)
+  "go.mod" @string)
```
ObserverOfTime, line 28: Wouldn't this be enough?

```suggestion
(checksum
  "go.mod" @string
```

amaanq: yeah, fixed

ObserverOfTime (approved):

PR #4274, opened 2023-02-04
feat: add RON

amaanq:
This PR adds support for RON (Rusty Object Notation) https://github.com/ron-rs/ron

cc: @xrelkd as they first requested it in #2282

This is my first time adding queries that aren't highlights, so I'd really like a maintainer to have a look over folds, indents, and injections just for a sanity check.

I also noticed a lot of parsers have an injection for comments where it's just `@comment`, it'd be nice if someone could explain the reason/why behind that :)

Lastly, I removed redundant branch = "master" from my parsers, I also noticed quite a few parsers have this and have a redundant filetype field as well (which is meant for when the parser name doesn't equal the filetype name right?). If it's alright, I'll fix up other filetype/branch redundancies as well.

### DIFF
--- a/queries/ron/indents.scm
+++ b/queries/ron/indents.scm
@@ -1,0 +1,12 @@
+[
+  (array)
+  (map)
+  (tuple)
+  (struct)
+] @indent
+
+[ "{" "}" ] @branch
+
+[ "(" ")" ] @branch
+
+[ "[" "]" ] @branch
--- a/README.md
+++ b/README.md
@@ -288,6 +288,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [regex](https://github.com/tree-sitter/tree-sitter-regex) (maintained by @theHamsta)
 - [x] [rego](https://github.com/FallenAngel97/tree-sitter-rego) (maintained by @FallenAngel97)
 - [x] [rnoweb](https://github.com/bamonroe/tree-sitter-rnoweb) (maintained by @bamonroe)
+- [x] [ron](https://github.com/amaanq/tree-sitter-ron) (maintained by @amaanq)
 - [x] [rst](https://github.com/stsewd/tree-sitter-rst) (maintained by @stsewd)
 - [x] [ruby](https://github.com/tree-sitter/tree-sitter-ruby) (maintained by @TravonteD)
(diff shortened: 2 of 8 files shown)


theHamsta:
> I also noticed a lot of parsers have an injection for comments where it's just @comment, it'd be nice if someone could explain the reason/why behind that :)

Vim regex highlighting, highlighted certain things like `TODO`, `WARNING`, `@foo`, or license key words. This does no longer with tree-sitter. So stsewd created a parser for that https://github.com/stsewd/tree-sitter-comment. I personally don't use it.

theHamsta (approved):
LGTM! Thank you! I wanted to add this long time ago myself.

queries/ron/indents.scm
```diff
@@ -0,0 +1,12 @@
...
+  (map)
+  (tuple)
+  (struct)
+] @indent
+
+[ "{" "}" ] @branch
+
+[ "(" ")" ] @branch
+
+[ "[" "]" ] @branch
```
theHamsta, line 12: should we add a simple test file for testing indents?

amaanq: sure, but I'm not very familiar with testing indents. can you link an example?

theHamsta: It should be sufficient to add a file like this https://github.com/nvim-treesitter/nvim-treesitter/blob/588cdb6c0652356dacc7d29bb4bb2b273df2df86/tests/indent/css_spec.lua#L1-L22 and put a example `ron` file in a subfolder. The `new line:` section is optional and allows you to test how newly inserted lines in a file should be indented.

theHamsta: @amaanq this is not strictly necessary. We can also merge as is.

amaanq: > For this config block, you probably need to add `ft = "ron"` as Neovim doesn't recognize this file type natively. https://github.com/nvim-treesitter/nvim-treesitter/blob/588cdb6c0652356dacc7d29bb4bb2b273df2df86/tests/indent/css_spec.lua#L7

Yeah I've submitted a PR to vim to add that, should be merged soon :)

amaanq: > @amaanq this is not strictly necessary. We can also merge as is.

Yeah that's fine, I'll add a test shortly either way!

PR #4282, opened 2023-02-06
feat: improve Thrift queries

amaanq:
This PR improves Thrift queries.

* Set branch to main (if it's not master you should do this right?)
* Added indents, injections, and locals, improved folds

One semi-issue I had was getting the indent to go to the function identifier level after parameters if there's an exception (in thrift, a comma signifies the end of a function definition in a service so this would happen in the absence of a comma, note the screenshot has the comma after the exception parameters signifying the function end)

image

Screenshot shows my cursor position after hitting enter after a parameters list. It should ideally be aligned where throws is (under the function name), but I could not figure that out. If anyone reviewing this has a solution for that, that'd be a great addition to the indents.

Thanks!

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -417,7 +417,7 @@
     "revision": "0ff887f2a60a147452d52db060de6b42f42f1441"
   },
   "thrift": {
-    "revision": "634a73fd2c80e169f302917ba665c07ec0b6ff7b"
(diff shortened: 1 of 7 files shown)


amaanq:
ping @theHamsta

theHamsta (approved):

amaanq:
Thank you!

PR #4292, opened 2023-02-09
astro: add injections for scss

virchau13:
This adds injections for SCSS, which resolves part of https://github.com/virchau13/tree-sitter-astro/issues/9.


ObserverOfTime (approved):

theHamsta:
Thank you!

PR #4302, opened 2023-02-12
feat: add snakemake

osthomas:
Hi!

Snakemake is a workflow management system based on Python: https://snakemake.readthedocs.io/en/stable/
The grammar can be found here, but it is not complete: https://snakemake.readthedocs.io/en/stable/snakefiles/writing_snakefiles.html#grammar
I took a stab at implementing a parser for it based on the Python parser because I wanted consistent highlighting between Python and snakemake: https://github.com/osthomas/tree-sitter-snakemake

I've been using it for a bit and would be happy to see it included in nvim-treesitter.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1499,10 +1499,7 @@ list.smali = {
 list.snakemake = {
   install_info = {
     url = "https://github.com/osthomas/tree-sitter-snakemake",
-    branch = "main",
-    location = "tree-sitter-snakemake",
     files = { "src/parser.c", "src/scanner.c" },
-    requires_generate_from_grammar = false,
   },
   maintainer = { "@osthomas" },
   experimental = true,
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1496,6 +1496,18 @@ list.smali = {
   maintainers = { "@amaanq" },
 }
 
+list.snakemake = {
+  install_info = {
+    url = "https://github.com/osthomas/tree-sitter-snakemake",
+    branch = "main",
+    location = "tree-sitter-snakemake",
+    files = { "src/parser.c", "src/scanner.c" },
+    requires_generate_from_grammar = false,
+  },
+  maintainer = { "@osthomas" },
+  experimental = true,
+}
+
 list.smithy = {
   install_info = {
     url = "https://github.com/indoorvivants/tree-sitter-smithy",
--- a/queries/snakemake/highlights.scm
+++ b/queries/snakemake/highlights.scm
@@ -59,6 +59,6 @@
 
 ;; directive labels in block context (eg. within 'run:')
 ((identifier) @label
+  (#any-of? @label "input" "log" "output" "params" "resources" "threads" "wildcards")
   (#has-ancestor? @label "directive")
-  (#has-ancestor? @label "block")
-  (#any-of? @label "input" "log" "output" "params" "resources" "threads" "wildcards"))
+  (#has-ancestor? @label "block"))
--- a/queries/snakemake/highlights.scm
+++ b/queries/snakemake/highlights.scm
@@ -1,0 +1,64 @@
+; inherits: python
+
+;; Compound directives
+["rule" "checkpoint" "module"] @keyword
+
+;; Top level directives (eg. configfile, include)
+(module (directive name: _ @keyword))
+
+;; Subordinate directives (eg. input, output)
+(_ body: (_ (directive name: _ @label)))
+
+;; rule/module/checkpoint names
+(rule_definition name: (identifier) @type)
+(module_definition name: (identifier) @type)
+(checkpoint_definition name: (identifier) @type)
+
+
+;; Rule imports
+(rule_import
+  "use" @include
+  "rule" @include
+  "from" @include
+  "as"? @include
+  "with"? @include
+)
+
+
+;; Rule inheritance
+(rule_inheritance
+  "use" @keyword
+  "rule" @keyword
+  "with" @keyword
+)
+
+
+;; Wildcard names
+(wildcard (identifier) @variable)
+
+
+;; builtin variables
+((identifier) @variable.builtin
+  (#any-of? @variable.builtin "checkpoints" "config" "gather" "rules" "scatter" "workflow"))
+
+
+;; References to directive labels in wildcard interpolations
+
+;; the #any-of? queries are moved above the #has-ancestor? queries to
+;; short-circuit the potentially expensive tree traversal, if possible
+;; see:
+;; https://github.com/nvim-treesitter/nvim-treesitter/pull/4302#issuecomment-1685789790
+
+;; directive labels in wildcard context
+((wildcard (identifier) @label)
+  (#any-of? @label "input" "log" "output" "params" "resources" "threads" "wildcards"))
+((wildcard (attribute object: (identifier) @label))
+  (#any-of? @label "input" "log" "output" "params" "resources" "threads" "wildcards"))
+((wildcard (subscript value: (identifier) @label))
+  (#any-of? @label "input" "log" "output" "params" "resources" "threads" "wildcards"))
+
+;; directive labels in block context (eg. within 'run:')
+((identifier) @label
+  (#has-ancestor? @label "directive")
+  (#has-ancestor? @label "block")
+  (#any-of? @label "input" "log" "output" "params" "resources" "threads" "wildcards"))
--- a/queries/snakemake/injections.scm
+++ b/queries/snakemake/injections.scm
@@ -1,0 +1,3 @@
+; inherits: python
+
+(wildcard (constraint) @regex)
--- a/queries/snakemake/injections.scm
+++ b/queries/snakemake/injections.scm
@@ -1,3 +1,4 @@
 ; inherits: python
 
-(wildcard (constraint) @regex)
+(wildcard (constraint) @injection.content
+  (#set! injection.language "regex"))
--- a/lockfile.json
+++ b/lockfile.json
@@ -509,6 +509,9 @@
   "smithy": {
     "revision": "cf8c7eb9faf7c7049839585eac19c94af231e6a0"
   },
+  "snakemake": {
+    "revision": "65a6c3b4671877821082164da0a310851b211953"
+  },
   "solidity": {
     "revision": "168020304759ad5d8b4a88a541a699134e3730c5"
   },
--- a/queries/snakemake/folds.scm
+++ b/queries/snakemake/folds.scm
@@ -1,0 +1,6 @@
+; inherits: python
+[
+  (rule_definition)
+  (module_definition)
+  (checkpoint_definition)
+] @fold
--- a/queries/snakemake/locals.scm
+++ b/queries/snakemake/locals.scm
@@ -1,0 +1,1 @@
+; inherits: python


queries/snakemake/highlights.scm
```diff
@@ -0,0 +1,61 @@
...
+  "from" @include
+)
+
+
+;; Rule inheritance
+(rule_inheritance
+  "use" @keyword
+  "rule" @keyword
+  "from"? @include
+)
```
ObserverOfTime, line 23: Why are these different?

queries/snakemake/injections.scm
```diff
@@ -0,0 +1,29 @@
+; inherits: python
+
+(wildcard (constraint) @regex)
+
+;; Injection of bash in shell directives is disabled by default. To enable,
```
ObserverOfTime, line 5: Why is it disabled?

ObserverOfTime (changes requested):
Don't just use `@keyword` everywhere.

queries/snakemake/highlights.scm
osthomas: Thank you for taking a look!

You're right, this is not entirely consistent. The difference is that a rule import takes a rule from another workflow file, whereas an inheritance takes a rule from the same file, but modifies some of its parameters. However, it's also possible to do a modifying rule import, which imports a rule and immediately modifies it. The `use rule (from)` syntax is used for both.

I should probably use `@type` for rule names though and adjust the `as` for imports.

smk_py

queries/snakemake/injections.scm
osthomas: Typically, you'll use wildcards such as `{input}` or `{output}` in the shell directives, which would no longer be highlighted with the injection. If you tend to write complicated scripts in the shell directive, it may be desirable to have proper syntax highlighting in there, but most of the time you're probably better of creating a proper `.sh` file and calling that. I left it in there to show how it's done in case someone wants it, but it could also be removed entirely.

osthomas:
> Don't just use `@keyword` everywhere.

I'm not sure what else to use. Workflow rules are defined with a set of directives (input, output, params, threads, etc. There's quite a few). These come from a specified pool and cannot be freely chosen.

queries/snakemake/highlights.scm
ObserverOfTime: > However, it's also possible to do a modifying rule import, which imports a rule and immediately modifies it. 

That's still an import. :)

queries/snakemake/injections.scm
ObserverOfTime: We have something similar in `yaml`. Just enable it.

ObserverOfTime:
> I'm not sure what else to use.

Maybe `@constant.builtin`?

queries/snakemake/highlights.scm
osthomas: > > However, it's also possible to do a modifying rule import, which imports a rule and immediately modifies it.
> 
> That's still an import. :)

Fair enough. I've updated the parser and the queries to reflect that.
Not entirely sure what to do with the final optional `with` for a modifying import. I've left it as `@include` for now,  what do you think?
smk

queries/snakemake/injections.scm
osthomas: I would favor to leave it disabled by default because `{}` has syntactical meaning in bash which is different from its meaning of wildcard interpolation in snakemake and because it will reduce the consistency between the directives in as they're displayed in the rule block vs the shell block.

ObserverOfTime: Then I vote for removing them altogether.

ObserverOfTime (changes requested):

queries/snakemake/injections.scm
ObserverOfTime: Then I vote for removing them altogether.

osthomas: [Done](https://github.com/osthomas/nvim-treesitter/commit/321c1e45174ed35663c8a3cef7a92c5c29cbd512#diff-aa5c2fc80b8fec60ad306efda25b19808e8640ff0e37a0cf1bb4beccd490ed3bL5)

queries/snakemake/highlights.scm
```diff
@@ -0,0 +1,61 @@
+; inherits: python
+
+;; Compound directives
+["rule" "checkpoint" "module"] @keyword
+
+;; Simple directives
+(directive name: _ @keyword)
```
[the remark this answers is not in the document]
osthomas: I don't think directives should be `@field`s because not all of them can be referenced during interpolation. For example, you can define a `message:` directive in a rule, but you cannot refer to this with `{message}` in the shell block. 
Also, directives themselves can have fields that can be referred to individually (`input.a`, `input.b`). You could even assign a dictionary to `input.a`, and you would later refer to its keys as `input.a.key1`, `input.a.key2`, ...
Any wildcard names are also available as fields (`wildcards.f`), which is why I chose `@field` for them.

In my opinion, it is important to highlight the elements in a wildcard interpolation (`input.a`) the same way as their definition `input: a = "..."`

Concerning changing directive names from `@keyword` to `@variable.builtin` -- there is also the case of the `run` directive, which is an alternative to `shell` that runs python code instead of a shell command. This is syntactically close to `if x: do()`, in which `if` is a keyword. There are also free-standing directives which can appear outside of a block, for example `configfile: "config.yaml"`, which I think is captured better by `@keyword`. More generally, the word-colon sequence is reserved for keywords in base Python. I think the breadth of functions fulfilled by these directives warrants their `@keyword` status.

Screenshot_20230212_192628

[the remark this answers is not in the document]
ObserverOfTime: OK

ObserverOfTime:
CI error seems unrelated.

osthomas:
I just updated tree-sitter-snakemake to include the latest upstream changes from tree-sitter-python.

Anything else I should do here?

ObserverOfTime (approved):

osthomas:
Pinging @theHamsta because I think this may have been buried.

queries/snakemake/highlights.scm
```diff
@@ -0,0 +1,63 @@
...
+  "with"? @include
+)
+
+
+;; Rule inheritance
+(rule_inheritance
+  "use" @keyword
+  "rule" @keyword
+  "with"? @keyword
+)
```
amaanq, line 25:
```suggestion
[
  "use"
  "rule"
  "with"
] @include
```

queries/snakemake/highlights.scm
```diff
@@ -0,0 +1,63 @@
...
+      "config"
+      "input"
+      "log"
+      "output"
+      "params"
+      "resources"
+      "threads"
+      "wildcards"
+   )
+)
```
amaanq, line 63: This seems unnecessarily costly, a code snippet w/ the corresponding ts tree would be nice to find better ways to add captures for it

amaanq (changes requested):

queries/snakemake/highlights.scm
osthomas: Thank you for your comments! These are, however, not equivalent. The rule import imports a rule from an external file, the rule inheritance modifies a rule defined in the current file with alternative parameters. 



But actually, the `?` in `"with"?` for inheritance can go.

```suggestion
;; Rule imports
(rule_import
  "use" @include
  "rule" @include
  "from" @include
  "as"? @include
  "with"? @include
)


;; Rule inheritance
(rule_inheritance
  "use" @keyword
  "rule" @keyword
  "with" @keyword
)
```

queries/snakemake/highlights.scm
osthomas: Here's an example. These keywords are context-specific and only have "special snakemake meaning" in particular contexts.
tree

queries/snakemake/highlights.scm
amaanq: I would argue they almost do the same thing, you're using something defined somewhere else, one just changed parameters, I don't think its relative location is a candidate for having it marked as `@keyword` vs `@include`, however the actual "rule" keyword itself could be `@keyword` vs use, with, as, from

osthomas: My argument here is that rule inheritance `use rule with` is basically equivalent to class inheritance, whereas `use rule from` is an import, they just happen to share the same syntax.
Do you still think it should be changed? `as` and `with` would still need to be scoped to avoid clashes with Python's context manager `with x as y`

amaanq: From snakemake's loose ebnf-like syntax spec: 
```
userule = "use" "rule" (identifier | "*") "from" identifier ["as" identifier] ["with" ":" norunparams]
```

Scope them, sure, as to avoid conflict w/ regular Python keywords but these should absolutely be consistent then, and I'd go so far as to say having them be two separate rules in the grammar is incorrect but that's out of the scope of this issue :)

queries/snakemake/highlights.scm
```diff
@@ -0,0 +1,63 @@
+; inherits: python
+
+;; Compound directives
+["rule" "checkpoint" "module"] @keyword
+
+;; Simple directives
+(directive name: _ @keyword)
```
amaanq, line 7:
```suggestion
(directive name: _ @field)

(rule_definition name: (identifier) @type)
```

queries/snakemake/highlights.scm
```diff
@@ -0,0 +1,63 @@
...
+(rule_inheritance
+  "use" @keyword
+  "rule" @keyword
+  "with"? @keyword
+)
+
+
+;; Wildcard names
+;; as @field for consistency with attributes in wildcard interpolations
+(wildcard (identifier) @field)
```
amaanq, line 30:
```suggestion
```

Field doesn't make much sense

queries/snakemake/highlights.scm
amaanq, line 63:
```suggestion
((wildcard (identifier) @keyword)
  (#any-of? @keyword "config" "input" "log" "output" "params" "resources" "threads" "wildcards"))

((wildcard (attribute object: (identifier) @keyword))
  (#any-of? @keyword "config" "input" "log" "output" "params" "resources" "threads" "wildcards"))
```

queries/snakemake/injections.scm
```diff
@@ -0,0 +1,3 @@
+; inherits: python
+
+(wildcard (constraint) @regex)
```
amaanq, line 3:
```suggestion
(wildcard (constraint) @regex)

((directive 
  name: _ @_direc 
  (directive_parameters (string) @bash))
  (#eq? @_direc "shell")
  (#offset! @bash 0 1 0 -1))
```

Could be iffy with wildcards but it's ok for me

amaanq (changes requested):

amaanq:
I took some time to go over the parser & queries more in-depth, check out my feedback now @osthomas

queries/snakemake/highlights.scm
osthomas: Naturally, I disagree :)
The snippet you show doesn't mark the `"from" identifier` part as optional, so it appears to refer to the import, not inheritance, and conceptually, I think they differ.

From a highlighting perspective, if I do

```python
class A:
    def greet(self):
        print("Hi")
    def leave(self):
        print("Bye")

class B(A):
    def greet(self):
        print("Hello")
```

then I don't expect the inheritance to be highlighted as an import. Rule inheritance is the same thing.

If I failed to persuade you now, I'll concede. Just wanted to make my reasoning clear :)

queries/snakemake/highlights.scm
osthomas: Agreed on rule names.

For `@keyword` vs `@field`, see my arguments here: https://github.com/nvim-treesitter/nvim-treesitter/pull/4302#discussion_r1103862281

queries/snakemake/highlights.scm
osthomas: My main concern here is consistency of highlighting between the definition and interpolation of wildcards. The parameters to directives are `keyword_argument`s, in which `identifier` is highlighted as `@parameter` --> `(wildcard (identifier) @parameter)`?
Related: https://github.com/nvim-treesitter/nvim-treesitter/pull/4302#discussion_r1103862281

So to summarize this `@field` and the other `@keyword` issue, I want to achieve 3 things here:

1. Capture the hierarchical structure between directives and their parameters (`keyword` > `parameter/field`). Making the directives `field` loses that.
2. Maintain consistency between definitions and accessions. Dropping `field` (/`parameter`?) from wildcard identifiers loses that.
3. Avoid touching base Python as much as possible. Switching from `keyword` to `field` for directives and dropping `field` from wildcard identifiers maintains hierarchy, but loses consistency, unless `(keyword_argument (identifier)` is demoted from `parameter`.

queries/snakemake/highlights.scm
osthomas: Thanks, this simplifies it for the `shell` blocks. Unfortunately, it doesn't catch the `run` blocks, which can contain arbitrary Python code and in which these wildcards are expanded in a non-string context (eg. no `wildcard` scope). That's why I used `has-ancestor` there.


```suggestion
((wildcard (identifier) @keyword)
  (#any-of? @keyword "config" "input" "log" "output" "params" "resources" "threads" "wildcards"))
((wildcard (attribute object: (identifier) @keyword))
  (#any-of? @keyword "config" "input" "log" "output" "params" "resources" "threads" "wildcards"))
  
(
  (identifier) @keyword
  (#has-ancestor? @keyword "directive")
  (#has-ancestor? @keyword "block")
  (#any-of? @keyword
      "config"
      "input"
      "log"
      "output"
      "params"
      "resources"
      "threads"
      "wildcards"
   )
)
```

queries/snakemake/injections.scm
osthomas: I had a bash injection in the queries before, but ended up removing it and I no longer favor it. From my personal experience writing snakemake, it's much more useful to have good highlighting of wildcards than of bash, because you're mostly just calling external scripts or short commands with parameters from the rule. More complicated logic is typically outsourced into dedicated files.

queries/snakemake/highlights.scm
amaanq: I missed the from part being optional....now I can agree with you here.

queries/snakemake/highlights.scm
amaanq: I still don't think it should be keyword, what about attribute, property, label, type?

queries/snakemake/highlights.scm
amaanq: Can I get a code sample

clason: Note: `#has-ancestor?` does a full tree traversal and is pretty expensive; use sparingly, if at all. (`#has-parent?` is somewhat cheaper but still not free.)

queries/snakemake/highlights.scm
osthomas: Okay, I think one of these sounds good for the "subordinate" directives (eg. `input`, `output`, which belong to `rule`). Ill think about which one of your suggestions to use. I'm leaning towards `attribute` or `label` currently.

For the "freestanding" directives (eg. `configfile`), which appear at the top level (same level as `rule`), `keyword` still feels appropriate to me. Would you agree to a separation of these?

queries/snakemake/highlights.scm
osthomas: See lower part in the screenshot here: https://github.com/nvim-treesitter/nvim-treesitter/pull/4302#discussion_r1188321045

Here is another example with more nesting (hence `#has-ancestor?` instead of `#has-parent?`). 
```
# Do something pointless with text files.
rule a:
    input:
        a = "a.txt",
        b = "b.txt"
    output: "out.txt"
    params:
        x_times = 5
    run:
        with open(output[0], "w") as outf:
            for i in range(params.x_times):
                with open(input.a, "r") as in_a, open(input.b, "r") as in_b:
                    outf.write(f"\n -- Round {i + 1} / {params.x_times}\n")
                    for line_a in in_a:
                        for line_b in in_b:
                            outf.write(line_b + line_a[::-1])
```
run_example

queries/snakemake/highlights.scm
osthomas: Updated queries to cut down on keyword use: https://github.com/nvim-treesitter/nvim-treesitter/pull/4302/commits/3b4275af2f11875b75de6432195d150acdbe48e7

Please re-open if you disagree.

queries/snakemake/highlights.scm
osthomas: Unfortunately, I still don't see how this can be solved another way.

osthomas:
I'm not sure what caused the previous CI failures.
Maybe the `; inherits: python` in the queries causes issues in the CI environment? The current Python queries are required, without them there's indeed an error.

amaanq:
I recently updated the python grammar quite a bit w/ some bug fixes and 3.12 stuff, maybe you need to update your grammar to use the latest release

osthomas:
> 



> I recently updated the python grammar quite a bit w/ some bug fixes and 3.12 stuff, maybe you need to update your grammar to use the latest release

It's up to date as of today. I also included the scanner rewrite in C, so that should be future-proof for now, I hope.

queries/snakemake/highlights.scm
osthomas: Updated queries to cut down on keyword use: https://github.com/nvim-treesitter/nvim-treesitter/pull/4302/commits/3b4275af2f11875b75de6432195d150acdbe48e7

Please re-open if you disagree.

amaanq:
Awesome, let me revisit this - it's been a while

queries/snakemake/highlights.scm
```diff
@@ -0,0 +1,60 @@
...
+;; Rule inheritance
+(rule_inheritance
+  "use" @keyword
+  "rule" @keyword
+  "with" @keyword
+)
+
+
+;; Wildcard names
+(wildcard (identifier) @parameter)
```
amaanq, line 37:
```suggestion
(wildcard (identifier) @variable)
```

queries/snakemake/highlights.scm
```diff
@@ -0,0 +1,60 @@
...
+      "resources"
+      "threads"
+      "wildcards"
+   )
+)
+
+((wildcard (identifier) @attribute)
+  (#any-of? @attribute "config" "input" "log" "output" "params" "resources" "threads" "wildcards"))
+((wildcard (attribute object: (identifier) @attribute))
+  (#any-of? @attribute "config" "input" "log" "output" "params" "resources" "threads" "wildcards"))
```
amaanq, line 60:
```suggestion
((wildcard (identifier) @variable.builtin)
  (#any-of? @variable.builtin "config" "input" "log" "output" "params" "resources" "threads" "wildcards"))
((wildcard (attribute object: (identifier) @variable.builtin))
  (#any-of? @variable.builtin "config" "input" "log" "output" "params" "resources" "threads" "wildcards"))
```

amaanq (changes requested):

queries/snakemake/highlights.scm
amaanq: Same for the has-ancestor one, just seeing if that's still needed atm

amaanq:
Okay, so for the has-ancestor issue, I see a couple solutions

1) keep it as is, but move the any-of check first, I'm assuming (would need confirmation), that predicates are executed in order of appearance, and so an any-of check will be faster than the has-ancestors, so checking for any-of first for variables would be better.

2) remove it completely, even I don't like this

3) Remove the has-ancestor checks, assume people will use snakemake in a sane manner and not use these special identifiers outside of rules :) this gives us the best of both worlds, but with a slight potential for false positives

I'm fine w/ any option @osthomas

queries/snakemake/highlights.scm
osthomas: 1. I would like to keep the special wildcard identifiers consistent with their counterparts in the directive block and therefore prefer `@attribute` here, if that's possible.

2. But as for `@variable.builtin`: thinking about it, this should probably exist for a select subset of identifiers, one listed example being `config` because this is a global (builtin) variable in the snakemake context. There are a few more of these.
3. I realized `(subscript value: (identifier) @...)` needs to be added here because you can also do `input[0]` in addition to `input.named_something`.

OK on 1/2/3?

osthomas:
> Okay, so for the has-ancestor issue, I have a few solutions
> 
>     1. keep it as is, but move the any-of check first, I'm assuming (would need confirmation), that predicates are executed in order of appearance, and so an any-of check will be faster than the has-ancestors, so checking for any-of first for variables would be better.
> 
>     2. remove it completely, even I don't like this
> 
>     3. Remove the has-ancestor checks, keep any-of and just run that on every identifier, and assume people will use snakemake in a sane manner and not use these special identifiers outside of rules :) this gives us the best of both worlds, but with a slight potential for false positives
> 
> 
> I'm fine w/ any option @osthomas

I prefer option 1, but I don't know how to test that assumption.

queries/snakemake/highlights.scm
amaanq: Ah tbh, for 1) I'm not sure attribute even fits there for the directive names, maybe field or label would be slightly better (I like label more personally)

2) Sure, whichever apply make them `@variable.builtin`

3) sure, add that as well.

osthomas:
> Awesome, let me revisit this - it's been a while
> 
> I also don't recommend using git submodules, you should add the latest stable release in your package.json as a dependency, and move the contents of the nested snakemake folder into the main dir, maybe the docs aren't crystal clear about how to set up an inherited grammar, but there's examples like cpp to learn from

Sorry, I had missed your edit.
When setting up the project I indeed struggled with how to best tackle that. I ended up adapting the approach from https://github.com/theHamsta/tree-sitter-cuda, which uses a submodule, but I'm open to adapt that at some point. I'll have a look at the cpp parser.

amaanq:
Sure, I think he uses a submodule because parsers back then weren't releasing new versions regularly when changes/fixes were made, but I do ensure they stay relatively updated w.r.t. releases

osthomas:
I got rid of the submodule now.

amaanq:
Please don't update the lockfile yourself, remove those changes except for your entry for snakemake. And also, in the last query with has-ancestor, any-of should be first like I mentioned

lua/nvim-treesitter/parsers.lua
```diff
@@ -1480,6 +1480,18 @@ list.smali = {
   maintainers = { "@amaanq" },
 }
 
+list.snakemake = {
+  install_info = {
+    url = "https://github.com/osthomas/tree-sitter-snakemake",
+    branch = "main",
+    location = "tree-sitter-snakemake",
```
clason, line 1487:
```suggestion
```

This is only needed if the grammar is in a subdirectory of the repo (which is not the case here).

lua/nvim-treesitter/parsers.lua
```diff
@@ -1480,6 +1480,18 @@ list.smali = {
...
+    files = { "src/parser.c", "src/scanner.c" },
+    requires_generate_from_grammar = false,
```
clason, line 1489:
```suggestion
```
No need to specifiy the default value.

lua/nvim-treesitter/parsers.lua
```diff
@@ -1480,6 +1480,18 @@ list.smali = {
   maintainers = { "@amaanq" },
 }
 
+list.snakemake = {
+  install_info = {
+    url = "https://github.com/osthomas/tree-sitter-snakemake",
+    branch = "main",
```
clason, line 1486:
```suggestion
```

No need to specify the default branch. (It's enough to have a lockfile entry then.)

queries/snakemake/injections.scm
clason, line 3: Neovim uses the upstream format for injections now (see CONTRIBUTING.md). You need to rebase on (not merge!) the latest `master`.

clason (changes requested):

amaanq (approved):

amaanq:
@clason needs your approval

PR #4306, opened 2023-02-12
chore(fish): update test command highlight query

ram02z:
Updated `highlights.scm` to reflect breaking changes from https://github.com/ram02z/tree-sitter-fish/pull/19

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -108,7 +108,7 @@
     "revision": "517195970428aacca60891b050aa53eabf4ba78d"
   },
   "fish": {
-    "revision": "6675b56266b3f615fb112205b6b83a79315309c4"
+    "revision": "f9176908c9eb2e11eb684d79e1d00f3b29bd65c9"
   },
   "foam": {
     "revision": "c238f4af9a5723a212cf1a4c9b31dd5c1d5270a2"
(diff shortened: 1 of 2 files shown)


clason:
Please always include the parser update in a breaking query change. (If the parser update has not been merged yet and you are opening this PR in advance, mark it as draft.)

ram02z:
Strange, I merged https://github.com/ram02z/tree-sitter-fish/pull/19 before opening this PR.

clason:
Yes, but you need to update the `lockfile.json` to point to the new revision; this doesn't happen magically and instantly, you know ;)

ObserverOfTime:
The lockfile was updated but the node still exists in one other capture.

clason:
> The lockfile was updated but the node still exists in one other capture.

No it was not; the PR was only merged 3 hours ago.

ram02z:
> Yes, but you need to update the `lockfile.json` to point to the new revision; this doesn't happen magically and instantly, you know ;)

I do know that, thats why I updated the `lockfile.json` with the new revision before I opened the PR ;)

> The lockfile was updated but the node still exists in one other capture.

Thanks! I forgot one of the removed nodes from `highlights.scm`. This should be fixed now.

ObserverOfTime (approved):

clason:
Sorry about that, then! I should have looked more closely. (We had a number of  PRs, so I was a bit on autopilot.)

ram02z:
> Sorry about that, then! I should have looked more closely. (We had a number of  PRs, so I was a bit on autopilot.)

No worries :)

PR #4310, opened 2023-02-12
Update Zig highlights

EleanorNB:
Recently I rearranged the keywords. This PR brings the highlights query file up to date with that.

### DIFF
--- a/queries/zig/highlights.scm
+++ b/queries/zig/highlights.scm
@@ -2,7 +2,7 @@
   (container_doc_comment)
   (doc_comment)
   (line_comment)
-] @comment
+] @comment @spell
 
 [
   variable: (IDENTIFIER)
@@ -82,7 +82,7 @@ field_constant: (IDENTIFIER) @constant
 [
   (LINESTRING)
   (STRINGLITERALSINGLE)
-] @string
+] @string @spell
 
 (CHAR_LITERAL) @character
 (EscapeSequence) @string.escape
--- a/queries/zig/highlights.scm
+++ b/queries/zig/highlights.scm
@@ -2,11 +2,7 @@
   (container_doc_comment)
   (doc_comment)
   (line_comment)
-] @comment @spell
-
-((line_comment) @text.note
-  (#match? @text.note "^// *zig fmt: (on|off) *$")
-)
+] @comment
 
 [
   variable: (IDENTIFIER)
@@ -27,7 +23,7 @@ parameter: (IDENTIFIER) @parameter
     field_access: (IDENTIFIER)
     parameter: (IDENTIFIER)
   ] @type
-  (#match? @type "^[A-Z]")
+  (#match? @type "^[A-Z]([a-z]+[A-Za-z0-9]*)*$")
 )
 ;; assume camelCase is a function
 (
@@ -36,7 +32,7 @@ parameter: (IDENTIFIER) @parameter
     field_access: (IDENTIFIER)
     parameter: (IDENTIFIER)
   ] @function
-  (#match? @function "^[a-z]+[A-Z]+")
+  (#match? @function "^[a-z]+([A-Z][a-z0-9]*)+$")
 )
 
 ;; assume all CAPS_1 is a constant
@@ -51,7 +47,7 @@ parameter: (IDENTIFIER) @parameter
 [
   function_call: (IDENTIFIER)
   function: (IDENTIFIER)
-] @function.call
+] @function
 
 exception: "!" @exception
 
@@ -60,15 +56,17 @@ exception: "!" @exception
   (#eq? @variable.builtin "_")
 )
 
-; (PtrTypeStart "c" @variable.builtin)
+(PtrTypeStart "c" @variable.builtin)
 
-; (
-;   (ContainerDeclType
-;       (ErrorUnionExpr)
-;       ; "enum"
-;   )
-;   (ContainerField (IDENTIFIER) @constant)
-; )
+(
+  (ContainerDeclType
+    [
+      (ErrorUnionExpr)
+      "enum"
+    ]
+  )
+  (ContainerField (IDENTIFIER) @constant)
+)
 
 field_constant: (IDENTIFIER) @constant
 
@@ -84,44 +82,46 @@ field_constant: (IDENTIFIER) @constant
 [
   (LINESTRING)
   (STRINGLITERALSINGLE)
-] @string @spell
+] @string
 
 (CHAR_LITERAL) @character
 (EscapeSequence) @string.escape
 (FormatSequence) @string.special
 
-[
-  "allowzero"
-  "volatile"
-  "threadlocal"
-  "inline"
-  "noinline"
-  "noalias"
-] @type.qualifier
-
-[
-  "anytype"
-  "anyframe"
-  (BuildinTypeExpr)
-] @type.builtin
-
 (BreakLabel (IDENTIFIER) @label)
 (BlockLabel (IDENTIFIER) @label)
 
 [
-  "true"
-  "false"
-] @boolean
+  "asm"
+  "defer"
+  "errdefer"
+  "nosuspend"
+  "test"
+] @keyword
 
 [
-  "undefined"
-  "unreachable"
-  "null"
-] @constant.builtin
+  "fn"
+] @keyword.function
 
 [
-  "else"
+  "and"
+  "or"
+  "orelse"
+] @keyword.operator
+
+[
+  "return"
+  "break"
+  "continue"
+  "async"
+  "await"
+  "suspend"
+  "resume"
+] @keyword.return
+
+[
   "if"
+  "else"
   "switch"
 ] @conditional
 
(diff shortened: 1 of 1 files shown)


queries/zig/highlights.scm
```diff
@@ -84,44 +82,46 @@ field_constant: (IDENTIFIER) @constant
...
-  "null"
-] @constant.builtin
+  "return"
+  "break"
+  "continue"
+  "async"
+  "await"
+  "suspend"
+  "resume"
+] @keyword.return
```
theHamsta, line 120: I think other languages use `@repeat` for break/continue when used for loops. Not sure for `aync`/`await`

queries/zig/highlights.scm
```diff
@@ -2,11 +2,7 @@
   (container_doc_comment)
   (doc_comment)
   (line_comment)
-] @comment @spell
```
theHamsta, line 5: so spelling will be removed with this PR? Intention is to let nvim spell on `@string`/`@comment`?

clason: Nvim doesn't bundle zig parser or queries, so if it's dropped here, there will be no spelling.

theHamsta: I thought nvim had a fallback to `@string`/`@comment` if the query doesn't provide any `@spell`

EleanorNB: I don't see how that bit could have been removed as I didn't touch it, but it's fixed now.

queries/zig/highlights.scm
EleanorNB: Moved those. Still unsure where to put coroutine keywords. I'd really like a `@keyword.coroutine`.

theHamsta: If you think a new category would make sense and not only to this language we could create a new PR adding `@keyword.coroutine` to all languages with async/await

theHamsta: Also `keyword.return` kind of makes sense for async/await/co_yield. We just haven't found a consistent way across languages

EleanorNB: Ok, moved coroutine keywords back into `@keyword.return` for now.

The other languages I know of with coroutines are Go (one keyword, `go`) and Lua (no keywords). With just those two it might not be worth it. Do you know any others?

amaanq: > Ok, moved coroutine keywords back into `@keyword.return` for now.
> 
> The other languages I know of with coroutines are Go (one keyword, `go`) and Lua (no keywords). With just those two it might not be worth it. Do you know any others?

Python: async/await
Rust: async/await

theHamsta: c++ now has also `co_await`/`co_yield`/`co_return`. `co_yield`/`co_return` are a bit a mixture between `return` and `coroutine`

amaanq: doing a quick query results in these languages:
```sh
rg -g "*.scm" -i "async|await"                                                   ─╯
queries/c_sharp/highlights.scm
340: "async"
341: "await"

queries/tlaplus/highlights.scm
84:  "await"

queries/racket/highlights.scm
74:  "#%app" "#%datum" "#%declare" "#%expression" "#%module-begin" "#%plain-app" "#%plain-lambda" "#%plain-module-begin" "#%printing-module-begin" "#%provide" "#%require" "#%stratified-body" "#%top" "#%top-interaction" "#%variable-reference" "->" "->*" "->*m" "->d" "->dm" "->i" "->m" "..." ":do-in" "==" "=>" "_" "absent" "abstract" "all-defined-out" "all-from-out" "and" "any" "augment" "augment*" "augment-final" "augment-final*" "augride" "augride*" "begin" "begin-for-syntax" "begin0" "case" "case->" "case->m" "case-lambda" "class" "class*" "class-field-accessor" "class-field-mutator" "class/c" "class/derived" "combine-in" "combine-out" "command-line" "compound-unit" "compound-unit/infer" "cond" "cons/dc" "contract" "contract-out" "contract-pos/neg-doubling" "contract-struct" "contracted" "current-contract-region" "define" "define-compound-unit" "define-compound-unit/infer" "define-contract-struct" "define-custom-hash-types" "define-custom-set-types" "define-for-syntax" "define-local-member-name" "define-logger" "define-match-expander" "define-member-name" "define-module-boundary-contract" "define-namespace-anchor" "define-opt/c" "define-sequence-syntax" "define-serializable-class" "define-serializable-class*" "define-signature" "define-signature-form" "define-splicing-for-clause-syntax" "define-struct" "define-struct/contract" "define-struct/derived" "define-syntax" "define-syntax-rule" "define-syntaxes" "define-unit" "define-unit-binding" "define-unit-from-context" "define-unit/contract" "define-unit/new-import-export" "define-unit/s" "define-values" "define-values-for-export" "define-values-for-syntax" "define-values/invoke-unit" "define-values/invoke-unit/infer" "define/augment" "define/augment-final" "define/augride" "define/contract" "define/final-prop" "define/match" "define/overment" "define/override" "define/override-final" "define/private" "define/public" "define/public-final" "define/pubment" "define/subexpression-pos-prop" "define/subexpression-pos-prop/name" "delay" "delay/idle" "delay/name" "delay/strict" "delay/sync" "delay/thread" "do" "else" "except" "except-in" "except-out" "export" "extends" "failure-cont" "field" "field-bound?" "file" "flat-murec-contract" "flat-rec-contract" "for" "for*" "for*/and" "for*/async""for*/first" "for*/fold" "for*/fold/derived" "for*/foldr" "for*/foldr/derived" "for*/hash" "for*/hasheq" "for*/hasheqv" "for*/last" "for*/list" "for*/lists" "for*/mutable-set" "for*/mutable-seteq" "for*/mutable-seteqv" "for*/or" "for*/product" "for*/set" "for*/seteq" "for*/seteqv" "for*/stream" "for*/sum" "for*/vector" "for*/weak-set" "for*/weak-seteq" "for*/weak-seteqv" "for-label" "for-meta" "for-space" "for-syntax" "for-template" "for/and" "for/async" "for/first" "for/fold" "for/fold/derived" "for/foldr" "for/foldr/derived" "for/hash" "for/hasheq" "for/hasheqv" "for/last" "for/list" "for/lists" "for/mutable-set" "for/mutable-seteq" "for/mutable-seteqv" "for/or" "for/product" "for/set" "for/seteq" "for/seteqv" "for/stream" "for/sum" "for/vector" "for/weak-set" "for/weak-seteq" "for/weak-seteqv" "gen:custom-write" "gen:dict" "gen:equal+hash" "gen:set" "gen:stream" "generic" "get-field" "hash/dc" "if" "implies" "import" "include" "include-at/relative-to" "include-at/relative-to/reader" "include/reader" "inherit" "inherit-field" "inherit/inner" "inherit/super" "init" "init-depend" "init-field" "init-rest" "inner" "inspect" "instantiate" "interface" "interface*" "invariant-assertion" "invoke-unit" "invoke-unit/infer" "lambda" "lazy" "let" "let*" "let*-values" "let-syntax" "let-syntaxes" "let-values" "let/cc" "let/ec" "letrec" "letrec-syntax" "letrec-syntaxes" "letrec-syntaxes+values" "letrec-values" "lib" "link" "local" "local-require" "log-debug" "log-error" "log-fatal" "log-info" "log-warning" "match" "match*" "match*/derived" "match-define" "match-define-values" "match-lambda" "match-lambda*" "match-lambda**" "match-let" "match-let*" "match-let*-values" "match-let-values" "match-letrec" "match-letrec-values" "match/derived" "match/values" "member-name-key" "mixin" "module" "module*" "module+" "nand" "new" "nor" "object-contract" "object/c" "only" "only-in" "only-meta-in" "only-space-in" "open" "opt/c" "or" "overment" "overment*" "override" "override*" "override-final" "override-final*" "parameterize" "parameterize*" "parameterize-break" "parametric->/c" "place" "place*" "place/context" "planet" "prefix" "prefix-in" "prefix-out" "private" "private*" "prompt-tag/c" "prop:dict/contract" "protect-out" "provide" "provide-signature-elements" "provide/contract" "public" "public*" "public-final" "public-final*" "pubment" "pubment*" "quasiquote" "quasisyntax" "quasisyntax/loc" "quote" "quote-syntax" "quote-syntax/prune" "recontract-out" "recursive-contract" "relative-in" "rename" "rename-in" "rename-inner" "rename-out" "rename-super" "require" "send" "send*" "send+" "send-generic" "send/apply" "send/keyword-apply" "set!" "set!-values" "set-field!" "shared" "stream" "stream*" "stream-cons" "stream-lazy" "struct" "struct*" "struct-copy" "struct-field-index" "struct-guard/c" "struct-out" "struct/c" "struct/contract" "struct/ctc" "struct/dc" "struct/derived" "submod" "super" "super-instantiate" "super-make-object" "super-new" "syntax" "syntax-case" "syntax-case*" "syntax-id-rules" "syntax-rules" "syntax/loc" "tag" "this" "this%" "thunk" "thunk*" "time" "unconstrained-domain->" "unit" "unit-from-context" "unit/c" "unit/new-import-export" "unit/s" "unless" "unquote" "unquote-splicing" "unsyntax" "unsyntax-splicing" "values/drop" "when" "with-continuation-mark" "with-contract" "with-contract-continuation-mark" "with-handlers" "with-handlers*" "with-method" "with-syntax" "~?" "~@" "λ"

queries/svelte/folds.scm
8: (await_statement)

queries/svelte/indents.scm
5:  (await_statement)
15:  (await_end_expr)

queries/clojure/highlights.scm
178:  "associative?" "atom" "await" "await-for" "await1"

queries/ecma/highlights.scm
238:"async"
239:"await"

queries/dart/highlights.scm
205:    "await"
223:    "async"
224:    "async*"

queries/vala/highlights.scm
93: "async"

queries/thrift/highlights.scm
102:  "async"

queries/swift/highlights.scm
25:"async" @keyword

queries/rust/highlights.scm
172:  "async"
173:  "await"

queries/cpp/highlights.scm
146: "co_await"

queries/hack/highlights.scm
20: "async"
21: "await"

queries/zig/highlights.scm
152:  "async"
154:  "await"

queries/python/highlights.scm
40:              "ReferenceError" "RuntimeError" "StopIteration" "StopAsyncIteration" "SyntaxError" "IndentationError" "TabError"
245:  "async"
246:  "await"
```

Sorry for the semi-spam :)

EleanorNB: Ok, well that's enough to justify a new category then. What's the process for that? Do I just go through all of these queries and put the words in a new category, or do we need additional tooling or documentation changes?

EleanorNB: (And if we're going to do this, let's wait until after this PR is merged to avoid conflicts.)

theHamsta: The CI checks for prefixes in CONTRIBUTING.md. You would add the new category to CONTRIBUTING.md and change all the highlights where necessary. If it's to much for one PR you should open an issue with a list which languages are still missing.

EleanorNB: Ok, I'll get on that once this is merged.

amaanq: I can help too, seems like a cool addition

theHamsta (approved):

queries/zig/highlights.scm
EleanorNB: #4334 is open.

PR #4323, opened 2023-02-14
SQL highlights update

matthias-Q:
Add recently added keywords
restructures usage of nodes a bit. Most of them where just used as `@keyword`
I have grouped some keywords as `@attributes` and `@conditionals`

### DIFF
--- a/queries/sql/highlights.scm
+++ b/queries/sql/highlights.scm
@@ -11,7 +11,7 @@
 (keyword_float) @function.call
 
 (count
-  name: (identifier) @function.call
+  name: (keyword_count) @function.call
   parameter: [(field)]? @parameter)
 
 (table_reference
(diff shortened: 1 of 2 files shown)


matthias-Q:
@theHamsta do you need a before after screenshot?

queries/sql/highlights.scm
```diff
@@ -27,13 +27,61 @@
...
+    parameter: [(literal)]?)))
+
 (literal) @string
 (comment) @comment
 (marginalia) @comment
 
 ((literal) @number
  (#lua-match? @number "^%d+$"))
 
+(parameter) @variable
```
theHamsta, line 42: Can you tell us why `@parameter` doesn't apply here?

theHamsta (approved):

theHamsta:
@derekstride

Thank you! Screenshots are typically not needed. Queries should just be logical and consistent with other languages

queries/sql/highlights.scm
matthias-Q: Yes, you are right. I will change this to a `@parameter` this evening.

DerekStride (approved):
Thanks @matthias-Q!

matthias-Q:
I have changed `@variable` to `@parameter` and added `@float` which was merged upstream yesterday.

clason:
> @theHamsta what is the usual workflow here? For these kinds of PR, do we need to ping the maintainer of the grammar?

Yes, that is the correct etiquette.

matthias-Q:
@DerekStride after your last approval, I have added `@float` query using a regex. Can you check that?

queries/sql/highlights.scm
```diff
@@ -27,13 +27,64 @@
...
+
 (literal) @string
 (comment) @comment
 (marginalia) @comment
 
 ((literal) @number
  (#lua-match? @number "^%d+$"))
 
+((literal) @float
+ (#lua-match? @float "^[-]?%d*.?%d*$"))
```
DerekStride, line 43:
```suggestion
 (#lua-match? @float "^[-+]?%d*\.%d*$"))
```

The `.` must be present so we need to remove the `?` and I think make it `\.` so it matches a literal `.` character and not any character.

Something like `+1.1` might be valid too so adding the `+` to the start of the regex would be good.

matthias-Q: The `+` is not supported by the grammar at the moment. I will leave this out and make the other changes.

clason:
Please rebase and squash your commits.

PR #4342, opened 2023-02-18
feat(fortran): Update queries

navdeeprana:
From https://github.com/stadelmanma/tree-sitter-fortran/pull/78

### DIFF
--- a/queries/fortran/highlights.scm
+++ b/queries/fortran/highlights.scm
@@ -11,10 +11,6 @@
  "device"
  "dimension"
  "endtype"
- "global"
- "grid_global"
- "host"
- "import"
  "in"
  "inout"
  "intent"
@@ -27,6 +23,8 @@
 
 [
  "contains"
+ "import"
+ "include"
  "private"
  "public"
  ] @include
@@ -67,8 +65,10 @@
  "exit"
  "extends"
  "format"
+ "global"
  "goto"
- "include"
+ "grid_global"
+ "host"
  "interface"
  "intrinsic"
  "non_intrinsic"
@@ -190,7 +190,7 @@
   (name) @function)
 
 (subroutine_call
-  (identifier) @function)
+  (identifier) @function.call)
 
 (keyword_argument
   name: (identifier) @keyword)
--- a/queries/fortran/highlights.scm
+++ b/queries/fortran/highlights.scm
@@ -5,147 +5,149 @@
 (comment) @comment
 
 [
-  (intrinsic_type)
-  "pointer"
-] @type
+ (intrinsic_type)
+ "allocatable"
+ "attributes"
+ "device"
+ "dimension"
+ "endtype"
+ "global"
+ "grid_global"
+ "host"
+ "import"
+ "in"
+ "inout"
+ "intent"
+ "optional"
+ "out"
+ "pointer"
+ "type"
+ "value"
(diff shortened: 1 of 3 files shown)


theHamsta (approved):

queries/fortran/highlights.scm
```diff
@@ -5,147 +5,149 @@
...
+ (intrinsic_type)
+ "allocatable"
+ "attributes"
+ "device"
+ "dimension"
+ "endtype"
+ "global"
+ "grid_global"
+ "host"
+ "import"
```
theHamsta, line 17: I don't think all of these are `@type`

queries/fortran/highlights.scm
```diff
@@ -157,26 +159,38 @@
...
 
 (end_subroutine_statement
   (name) @function)
 
+(end_module_procedure_statement
+  (name) @function)
+
 (subroutine_call
-  (identifier) @function.call)
+  (identifier) @function)
```
theHamsta, line 194: Why is this removed?

theHamsta (changes requested):
Added some small comments

ObserverOfTime:
The previous groups were better.

queries/fortran/highlights.scm
navdeeprana: A mistake. I will fix it.

queries/fortran/highlights.scm
navdeeprana: `import` goes in `@keyword`, I think. The rest are subroutine / function attributes [[1]](https://docs.nvidia.com/hpc-sdk/compilers/cuda-fortran-prog-guide/#cfpg-sub-func-qualifiers). For example
```fortran
attributes(host) function foo(x) result(y)
    real, intent(in) :: x
    real :: y
    y = x + 1
end function foo
```
 Should I move them to `@attribute` or `@keyword`?

theHamsta: Isn't `import` `@include`?

navdeeprana: moved `import` and `include` to `@include`. `global` etc to `@keyword`.

theHamsta (approved):

queries/fortran/highlights.scm
```diff
@@ -5,147 +5,149 @@
...
+ "endtype"
+ "in"
+ "inout"
+ "intent"
+ "optional"
+ "out"
+ "pointer"
+ "type"
+ "value"
+ ] @type
```
theHamsta, line 22: I don't think type applies to any of the keywords

theHamsta (changes requested):
@ObserverOfTime

queries/fortran/highlights.scm
```diff
@@ -5,147 +5,149 @@
...
-] @type.definition
+ (none)
+ "implicit"
+ ] @attribute
 
 [
-  "in"
-  "inout"
-  "out"
-] @storageclass
```
theHamsta, line 35: they should stay `@storageclass`

ObserverOfTime (commented):
Some keywords should be `@type.qualifier` (i.e. modifiers). I'll let you decide which.

navdeeprana:
Thanks @amaanq for the better version in PR #4389. I guess this can be closed now?

PR #4343, opened 2023-02-18
feat: add starlark

amaanq:
cc: @folliehiyuki since he first requested it in #2282

This PR adds support for Starlark: https://github.com/bazelbuild/starlark
Grammar: https://github.com/amaanq/tree-sitter-starlark

It is mainly derived from Python with some minor differences. The two differences I'd like looked over by a maintainer are:

- Imports being replaced by a load() call, where individual strings are a module and as such I'd like to have `@conceal` work for the strings in there, but I could not get that to work and my testing would have the actual "load" keyword concealed.
- Classes being replaced by a struct call, e.g. `a = struct(x=123, y=False, z="string")`, so x/y/z are fields. Only issue is when matching I identified the struct in struct() as `@_func` to capture it later, but it seems like this capture is still visible outside of simply matching. No idea how to remove that.

Thanks,
Amaan

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,6 +304,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [solidity](https://github.com/YongJieYongJie/tree-sitter-solidity) (maintained by @YongJieYongJie)
 - [x] [sparql](https://github.com/BonaBeavis/tree-sitter-sparql) (maintained by @BonaBeavis)
 - [x] [sql](https://github.com/derekstride/tree-sitter-sql) (maintained by @derekstride)
+- [x] [starlark](https://github.com/amaanq/tree-sitter-starlark) (maintained by @amaanq)
 - [x] [supercollider](https://github.com/madskjeldgaard/tree-sitter-supercollider) (maintained by @madskjeldgaard)
 - [x] [surface](https://github.com/connorlay/tree-sitter-surface) (maintained by @connorlay)
 - [x] [svelte](https://github.com/Himujjal/tree-sitter-svelte) (maintained by @elianiva)
--- a/lockfile.json
+++ b/lockfile.json
@@ -392,6 +392,9 @@
   "sql": {
     "revision": "0d7a121b2a08fb37109f7be1cc6654443cad661f"
   },
+  "starlark": {
+    "revision": "2e94347f71147c268a4c5c0ce188be8baf2fce5a"
+  },
   "supercollider": {
     "revision": "90c6d9f777d2b8c4ce497c48b5f270a44bcf3ea0"
   },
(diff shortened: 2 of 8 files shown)


ObserverOfTime (changes requested):
Use `; inherits: python` instead of duplicating all the python queries.

amaanq:
> Use `; inherits: python` instead of duplicating all the python queries.

There's a bunch to delete since they're not a part of Starlark's grammar, like try/catch/raise/finally; is it possible to inherit python and remove those too? If so, how?

ObserverOfTime:
You could move the relevant queries to a separate directory (like `html` / `html_tags`) and inherit from that in both but I don't think it's worth it in this case.

Can you at least explicitly separate the starlark queries from the python ones with a comment (e.g. `;; starlark queries`)?

amaanq:
> You could move the relevant queries to a separate directory (like `html` / `html_tags`) and inherit from that in both but I don't think it's worth it in this case.
> 
> Can you at least explicitly separate the starlark queries from the python ones with a comment (e.g. `;; starlark queries`)?

Yeah, now highlights and locals have a section for this , injections inherits from Python, and folds/indents only contain removed items

ObserverOfTime (approved):

FollieHiyuki:
Sorry for the late reply. I just tried this branch today on some of my simple Bazel build files. Looks good to me.

Thanks for the work!

amaanq:
> 1 last comment though. Does it need to be applied only to `bzl` files? Other programs use Starlark too (e.g. [ascode](https://github.com/mcuadros/ascode), [isopod](https://github.com/cruise-automation/isopod))

I didn't know about starlark having its own extension too...that has some bugs that I'm fixing rn.
I see two solutions, recognize both .bzl and .star for the filetypes that tree-sitter-starlark can parse, or set bzl files to starlark filetypes in vim (the latter is easier but feels more incorrect, I'm not sure if we can set two filetypes for one grammar)

amaanq:
Related: https://github.com/vim/vim/pull/12049

amaanq:
This is ready to merge FYI, ft detection was merged as well :)

theHamsta (approved):

theHamsta:
Thank you!

FollieHiyuki:
I saw the PR got merged, but I'm still putting my final comment here.

Yes, recognizing `*.bzl` as Starlark filetype is the correct way in my opinion. Having the filetype detection for `*.bzl` and `*.star` would be enough for most cases. The users can always add their application-specific extensions to `filetype.lua`.

amaanq:
While I did get Starlark ft detection merged upstream, I think it's ok to leave it as is (and keep .bzl as Bazel files...oh well)

PR #4351, opened 2023-02-20
feat(smali): improve queries

amaanq:
cc: @ObserverOfTime as he helped me with Smali initially when I was new :) would def like an eyeball or two over the query changes. The grammar was about half-rewritten as it wasn't complete nor correct in its broader statements and organization

Also note that I am a contributor to the sourcehut repo since @yotamN and I agreed to co-maintain the parser over there, no need to worry about that going stale or out of date

### DIFF
--- a/queries/smali/highlights.scm
+++ b/queries/smali/highlights.scm
@@ -4,52 +4,97 @@
 
 (primitive_type) @type.builtin
 
-(array_type
-  "[" @punctuation.bracket)
+((class_identifier) @type.builtin
+  (#vim-match? @type.builtin "^L(android|com/android|dalvik|java)/"))
 
 ; Methods
 
-(method_declaration
-  (method_identifier) @method)
+(method_definition
+  (method_signature (method_identifier) @method))
 
-(statement
+(expression
   (opcode) @_invoke
-  (full_method_identifier
-    (method_identifier) @method.call)
+  (value
+	(body
+	  (full_method_signature
+        (method_signature (method_identifier) @method.call))))
   (#lua-match? @_invoke "^invoke"))
 
-(method_identifier
-  "<init>(" @constructor
-  ")" @constructor)
+(method_handle
+  (full_method_signature
+	(method_signature (method_identifier) @method.call)))
 
-(method_identifier
-  "<clinit>(" @constructor
-  ")" @constructor)
+(call_site) @method.call
+
+(custom_invoke
+  (method_signature (method_identifier) @method.call))
+
+(annotation_value
+  (body
+    (method_signature (method_identifier) @method.call)))
+
+(annotation_value
+  (body
+    (full_method_signature
+      (method_signature (method_identifier) @method.call))))
+
+(field_definition
+  (value
+	(body
+		(method_signature (method_identifier) @method.call))))
+
+(field_definition
+  (value
+	(body
+	  (full_method_signature
+		(method_signature (method_identifier) @method.call)))))
+
+((method_signature
+  (method_identifier) @constructor)
+  (#any-of? @constructor "<init>" "<clinit>"))
 
 ; Fields
 
 (field_identifier) @field
 
-; Parameters
-
-(parameter) @parameter
+(annotation_key) @field
 
 ; Variables
 
-(variable) @variable
+(variable) @variable.builtin
+
+(local_directive
+  (identifier) @variable)
+
+; Parameters
+
+(parameter) @parameter.builtin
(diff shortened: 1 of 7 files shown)


queries/smali/highlights.scm
```diff
@@ -4,52 +4,106 @@
...
 
-(variable) @variable
+(variable) @variable.builtin
+
+(local_directive
+  (identifier) @variable)
+
+; Parameters
+
+(parameter) @parameter ; more like @parameter.builtin but that doesn't exist in nvim-treesitter
```
ObserverOfTime, line 77: You can still use `@parameter.builtin` if you want.

queries/smali/highlights.scm
```diff
@@ -4,52 +4,106 @@
...
-(array_type
-  "[" @punctuation.bracket)
+((class_identifier) @type.builtin
+  (#lua-match? @type.builtin "^Landroid/"))
+((class_identifier) @type.builtin
+  (#lua-match? @type.builtin "^Lcom/android/"))
+((class_identifier) @type.builtin
+  (#lua-match? @type.builtin "^Ldalvik/"))
+((class_identifier) @type.builtin
+  (#lua-match? @type.builtin "^Ljava/"))
```
ObserverOfTime, line 14: These should be combined.

queries/smali/highlights.scm
```diff
@@ -4,52 +4,106 @@
...
+  (#lua-match? @conditional "^cmp"))
 
-"=" @operator
+((opcode) @exception
+  (#lua-match? @exception "^throw"))
+
+[
+  "="
+  ".."
+  "@"
```
ObserverOfTime, line 105: Why is `@` an operator?

ObserverOfTime (changes requested):

queries/smali/highlights.scm
amaanq: that's not in the list in CONTRIBUTING.md though, is that fine then?

queries/smali/highlights.scm
amaanq: how so? I tried just stacking each lua-match expression in one but then none would work

queries/smali/highlights.scm
amaanq:
```smali
    invoke-custom {v0, v1}, call_site_0("doSomething", (LCustom;Ljava/lang/String;)Ljava/lang/String;, "just testing")@LBootstrapLinker;->backwardsLink(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/String;)Ljava/lang/invoke/CallSite;

    invoke-custom {v0, v1}, call_site_2("doSomething", (LCustom;Ljava/lang/String;)Ljava/lang/String;, invoke-static@Lnonsense;->somemethod()V)@LBootstrapLinker;->normalLink(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;Ljava/lang/invoke/MethodHandle;)Ljava/lang/invoke/CallSite;
```

very niche statements in Smali

queries/smali/highlights.scm
ObserverOfTime: This should work.
```clojure
((class_identifier) @type.builtin
  (#vim-match? @type.builtin "^L(android|com/android|dalvik|java)/"))
```

queries/smali/highlights.scm
clason: Yes. Only top-level captures are enforced; subcaptures can be arbitrary.

ObserverOfTime: Yes, it just won't be highlighted any differently by default.

queries/smali/highlights.scm
ObserverOfTime: Looks more like `@punctuation.delimiter` to me.

queries/smali/highlights.scm
clason: (careful: we just care about capture name to _allow_ consistent highlighting -- _how_ or even _whether_ something is highlighted completely depends on the colorscheme. we should be careful to always make this distinction.)

queries/smali/highlights.scm
amaanq: mmm yeah, its to separate targets and their types so that makes more sense to me too then

ObserverOfTime (approved):

PR #4360, opened 2023-02-22
feat(kotlin): improve highlight queries

JosefLitos:
it would be nice, if the parser could distinguish arguments and giving arguments by specified name of the parameter. The first one is an argument, but the latter is just the name of the parameter and therefore it should be recognized as a parameter rather than a `value_argument->simple_identifier`

### DIFF
--- a/queries/kotlin/highlights.scm
+++ b/queries/kotlin/highlights.scm
@@ -48,7 +48,7 @@
 (type_identifier) @type
 
 ; '?' operator, replacement for Java @Nullable
-(nullable_type) @attribute
+(nullable_type) @punctuation.special
 
 (type_alias
 	(type_identifier) @type.definition)
--- a/queries/kotlin/highlights.scm
+++ b/queries/kotlin/highlights.scm
@@ -408,7 +408,7 @@
 ; NOTE: `interpolated_identifier`s can be highlighted in any way
 (line_string_literal
 	"$" @punctuation.special
-	(interpolated_identifier) @variable)
+	(interpolated_identifier) @none @variable)
 (line_string_literal
 	"${" @punctuation.special
 	(interpolated_expression) @none
@@ -416,7 +416,7 @@
 
 (multi_line_string_literal
     "$" @punctuation.special
-    (interpolated_identifier) @variable)
+    (interpolated_identifier) @none @variable)
 (multi_line_string_literal
 	"${" @punctuation.special
 	(interpolated_expression) @none
--- a/queries/kotlin/highlights.scm
+++ b/queries/kotlin/highlights.scm
@@ -47,6 +47,9 @@
 
 (type_identifier) @type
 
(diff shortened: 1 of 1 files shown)


queries/kotlin/highlights.scm
```diff
@@ -47,6 +47,9 @@
 
 (type_identifier) @type
 
+; '?' operator, replacement for Java @Nullable
+(nullable_type) @attribute
```
ObserverOfTime, line 51:
```suggestion
(nullable_type) @punctuation.special
```

ObserverOfTime (changes requested):

queries/kotlin/highlights.scm
```diff
@@ -405,15 +408,15 @@
 ; NOTE: `interpolated_identifier`s can be highlighted in any way
 (line_string_literal
 	"$" @punctuation.special
-	(interpolated_identifier) @none)
+	(interpolated_identifier) @variable)
```
theHamsta, line 411: We need to look why `@none` no longer works. Actually this is supposed to be `@none` and `@variable`. If `@variable` has no highlight set then the string is supposed to be highlighted as `NONE` (reset the string highlighting for the interpolated range.

So
```
(interpolated_identifier) @none @variable)
```

ObserverOfTime (approved):

amaanq:
This should be partially reverted, nullable captures the entirety of `String?` thus making all of that `@punctuation.special`, but you'd want only `?` captured. However, nullable_type is a sequence of a type then a `_quest` where _quest is just the question mark. This prevents capturing the ? literally as a string or as a node, which imo is a horrible design choice as _quest is only used here. I would say put nullable_type under type as well for now, and then submit a PR/issue to rectify this.

I also wonder if a predicate can just capture the ? as well, that'd work too

JosefLitos:
@amaanq I completely agree with you. Unfortunately I am kind of guessing how the query syntax works because I still haven't managed to properly figure it out (if you could suggest me something to learn how the `#eq` or `#lua-match` work, it would be very helpful).
It is also somewhat weird that compared to `C++` parser, which has 2MB in size, `Kotlin` has just around 3MB, yet it is noticeably slower and jumping to a `file.kt` causes a second long lag, which isn't ideal.

ObserverOfTime:
> nullable_type captures the entirety of `String?`

Right, I forgot about that. That's why it wasn't highlighted before.

> Not a big fan of the Kotlin grammar as a whole

Yeah, it's a bit of a mess (as is the language server).

amaanq:
cpp is a bit heavy too, but I think the main issue comes from the size of the large state counts; in cpp its 6500/1800 but I believe cpp is more complex than kotlin in terms of lexical parsing

A simple regex match would likely work (like #vim-match), but I think the bigger problem is the way the grammar is structured. I'll submit a PR in a bit if I don't forget.

PR #4365, opened 2023-02-23
fix: use correct command separators for powershell

akriese:
`Windows powershell` and `Powershell` use different command separators (";") than `cmd.exe` ("&"). Before the fix, the plugin would simply not compile the treesitter parsers at all when running `nvim` from any powershell.

As there are also powershell users outside of Windows, one might want to think about implementing an extra case for those guys.

### DIFF
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -279,7 +279,7 @@ end
 
 function M.make_directory_change_for_command(dir, command)
   if fn.has "win32" == 1 then
-    return string.format("pushd %s & %s & popd", cmdpath(dir), command)
+    return string.format([[cmd /C "pushd %s & %s & popd"]], dir, command)
   else
     return string.format("cd %s;\n %s", dir, command)
   end


lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -279,7 +279,11 @@ end
 
 function M.make_directory_change_for_command(dir, command)
   if fn.has "win32" == 1 then
-    return string.format("pushd %s & %s & popd", cmdpath(dir), command)
+    if vim.o.shell == "cmd.exe" then
```
theHamsta, line 282: is this a reliable way to detect that we're executing in powershell? Could shell be `C:\Windows\system32\cmd.exe`?

theHamsta: I think it would be simpler to just execute every command in `iter_cmd_sync` either in `cmd.exe` or `powershell.exe`

theHamsta: Similar to https://github.com/nvim-treesitter/nvim-treesitter/blob/598b878a2b44e7e7a981acc5fc99bf8273cd1d0f/lua/nvim-treesitter/shell_command_selectors.lua#L20-L22

akriese: > is this a reliable way to detect that we're executing in powershell? Could shell be `C:\Windows\system32\cmd.exe` or just `cmd`?

I was wondering the same thing. In `:h shell` it says, that `cmd.exe` is default on Windows, but I guess one could specify a full path which would mess this solution up.

akriese: > I think it would be simpler to just execute every command in `iter_cmd_sync` either in `cmd.exe` or `powershell.exe`

I'll try that out!

theHamsta: maybe you could also regex on the shell like `cmd(.exe)?$`

akriese: I guess, even using regex might lead to open edge cases. I just implemented your suggestion of using `cmd /C` :)

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -279,7 +279,7 @@ end
 
 function M.make_directory_change_for_command(dir, command)
   if fn.has "win32" == 1 then
-    return string.format("pushd %s & %s & popd", cmdpath(dir), command)
+    return string.format([[cmd /C "pushd %s & %s & popd"]], dir, command)
```
theHamsta, line 282: but will this also change the cwd for powershell. My idea was actually to execute everything `iter_cmd_sync` with cmd

PR #4374, opened 2023-02-24
feat(dart): add folds

### LINKED ISSUE #3844: Dart folds query
I've noticed that Dart folding capabilities are limited since we currently don't have a `queries/dart/folds.scm` file and are falling back to `queries/dart/locals.scm` `@scope` captures. Could we add such file to improve folding or is it intentionally missing?

amaanq:
Closes #3844

This PR adds distinct folds to dart.

Side note: I am not a fan of the grammar in its current state (much like Kotlin), but rewriting that would be a huge task - I'll save that for another day :)

Thanks,
Amaan

### DIFF
--- a/queries/dart/folds.scm
+++ b/queries/dart/folds.scm
@@ -1,0 +1,15 @@
+[
+  (class_definition)
+  (enum_declaration)
+  (extension_declaration)
+
+  (arguments)
+  (function_body)
+
+  (block)
+  (switch_block)
+
+  (list_literal)
+  (set_or_map_literal)
+  (string_literal)
+] @fold


theHamsta (approved):

theHamsta:
@akinsho

akinsho:
@amaanq do you mean you want to create a new grammar for dart rather than use Usernobody's? It's not seeing a great deal of love atm due to time constraints and tbh it was never designed with editor highlighting in mind and was for some completely different purpose when originally built. In any event I think they'd be open to a big refactor. Although could also just be nice to have something geared to usage in editors 🤷🏿‍♂️

PR #4381, opened 2023-02-24
docs: restore correct highlight groups

### LINKED ISSUE #4106: Restore documentation for tree-sitter highlight groups
In 42ab95d5e11f247c6f0c8f5181b02e816caa4a4f, the old `TS*` highlight groups were removed. However, instead of updating the documentation to use the newer `@*` highlight groups, the docs were also removed.

Currently the description for these highlight groups is in a [section of the contributing file](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights). It might be better to move this back to the documentation. It might also be better to have a more compact layout than the previous one. Like in [`:help highlight-groups`](https://neovim.io/doc/user/syntax.html#highlight-groups).

If this seems alright, I'll open the PR.

amaanq:
Closes #4106 

Related: #3656 (and cc @clason since it was his PR)

Note that I didn't add the `@definition*` groups as iirc those are not used for highlighting since it's slow

### DIFF
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -472,6 +472,335 @@ Note: This is highly experimental, and folding can break on some types of
       In any case, feel free to open an issue with the reproducing steps.
 
 ==============================================================================
+HIGHLIGHTS                                        *nvim-treesitter-highlights*
+
+The following is a list of highlights groups, the syntactic elements they
+apply to, and some examples.
+
+							      *hl-@attribute*
+`@attribute`
+Annotations that can be attached to the code to denote some kind of meta
+information. e.g. C++/Dart attributes, Python decorators.
+
+								*hl-@boolean*
+`@boolean`
+Boolean literals: `True` and `False` in Python.
+
+							      *hl-@character*
+`@character`
+Character literals: `'a'` in C.
+
+						       *hl-@character.special*
+`@character.special`
+Special characters.
+
+								*hl-@conceal*
+`@conceal`
+Concealing: e.g. quotes in JSON.
+
+								*hl-@comment*
+`@comment`
+Line comments and block comments.
+
+							    *hl-@conditional*
+`@conditional`
+Keywords related to conditionals: `if`, `when`, `cond`, etc.
+
+							    *hl-@conditional.ternary*
+`@conditional.ternary`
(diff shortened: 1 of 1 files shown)


clason:
> Note that I didn't add the @definition* groups as iirc those are not used for highlighting since it's slow

They're only used in locals, not highlights.

My concern is that we are duplicating information from CONTRIBUTING.md, which is the actual source of truth (because it's used in our CI). Is there some way we can make sure they stay in sync? Or make CI use the docs instead?

clason:
I think this list needs some explanation, too: That these are the top-level captures used in _our_ highlight queries, but that subcaptures may be used individually which fall back to the top-level captures (no long text, just a mention with a pointer to `:h treesitter-highlight-groups`).

(I'm personally strongly against conflating captures and highlights, but I may be in the minority here.)

clason:
I'm not against it, mind you, just reiterating my concerns to explain why I removed them in the first place. I agree that syncing is not ideal, but it should not be too hard to make the CI script scrape the tags from `treesitter.txt` instead of `CONTRIBUTING.md`.

And the rest is just a matter of putting a bit of introductory text there.

theHamsta:
I'm not sure whether we should globally document the highlight groups instead of our usage (= name the tags differently in case Neovim wants to document their default linked captures) of them or not document them at all. The reason why the documentation was removed was in the end that those are no longer defined by nvim-treesitter.

For the other PR that proposed this there was the idea to keep the more compact style of CONTRIBUTING.md with code blocks and to have it in a separate do file to be referenced by contributing and the CI script. I would like to have the documentation only once. We could also do like nvim-lspconfig and generate help tags that point into markdown.

theHamsta:
In general, it would be great if we would have less duplication of documentation in this repo. At the moment there are some duplication of documentation in vim-help docs:
- API docs (could be generated from EmmyLua doc comments, I saw a GH action one time ago that could do that or the Neovim Python script)
- General overview: duplicated in README.md and Vim-docs

clason:
I'd like to park this until
1. we have decided on the stable set of highlight groups 
2. we have reworked CI

for the Nvim-treesitter 1.0 release. I agree that we should focus our documentation in general (full -- ideally generated -- documentation in treesitter.txt, simple "getting started" only in README.md).

PR #4382, opened 2023-02-24
Keyword coroutine

### LINKED ISSUE #4334: New keyword group for keywords related to coroutines
Recently I rearranged the keywords in the Zig highlights query. Everything had a logical place except for `async`/`await`/`suspend`/`resume`, which had to be awkwardly jammed in in other places.

We should have a new group, `@keyword.coroutine`, for keywords related to the running of coroutines: `async`/`await`/`suspend`/`resume` in Zig, `async`/`await` in Python, `go` in Go etc.

The alternative is to jam them all in `@keyword.return` but that just doesn't feel right

## Language checklist
This list was populated by a search through queries for `async|await`. It is therefore to be considered incomplete and possibly inaccurate.
- [ ] c_sharp
- [ ] tlaplus
- [ ] racket
- [ ] svelte
- [ ] clojure
- [ ] ecma
- [ ] dart
- [ ] vala
- [ ] thrift
- [ ] swift
- [ ] rust
- [ ] cpp
- [ ] hack
- [ ] zig
- [ ] python

Do we change them upstream first or are our copies allowed to desynchronise?

amaanq:
Closes #4334 

This PR introduces a new capture `@keyword.coroutine`

I am not an expert in even half the languages I modified, but I did take the time to read the docs where applicable and I applied my best judgement.

cc: @EleanorNB as it was your suggestion

Might be good to cc some maintainers too as this is a breaking change


amaanq:
Also, in some languages like c++, something like "co_yield" or "co_return" could be `@keyword.coroutine.return`, would that be a fine change to make in those scenarios where both could apply?

theHamsta (approved):

theHamsta:
I think we can leave this open for a while I'm sure people will have opinions on this change.

ObserverOfTime:
Won't this change make "return" the only keyword that will ever be in `@keyword.return`? :thinking:

amaanq:
> Won't this change make "return" the only keyword that will ever be in `@keyword.return`? :thinking: 

if I were to move coroutine return keywords like co_yield/co_return then yes, so maybe I won't touch those 🙂

ObserverOfTime:
I think it'd be fine to move them and remove the `@keyword.return` group.

amaanq:
> I think it'd be fine to move them and remove the `@keyword.return` group.

would returns just fall under `@keyword` then?

ObserverOfTime:
Wait, nevermind. I saw the vala change and forgot that "yield" isn't always (or even usually) a coroutine keyword.

amaanq:
> Wait, nevermind. I saw the vala change and forgot that "yield" isn't always (or even usually) a coroutine keyword.

I only read the docs here: https://wiki.gnome.org/Projects/Vala/AsyncSamples

I'm not familiar with Vala but to me it seemed like it was only used in an async manner

ObserverOfTime:
> I'm not familiar with Vala but to me it seemed like it was only used in an async manner.

Yeah, it's basically "await".

amaanq:
> > I'm not familiar with Vala but to me it seemed like it was only used in an async manner.
> 
> Yeah, it's basically "await".

So...doesn't that fit `@keyword.coroutine` then?

ObserverOfTime:
Yes, it's fine. I just got a little confused.

amaanq:
Another thing that might be nice to add is `@keyword.export`, but that might be even more niche than coroutine.

ObserverOfTime (approved):

amaanq:
Changes I'm not 100% certain about, so pinging maintainers

Racket: @6cdh
Svelte: @elianiva
Tlaplus: @ahelwer, @susliko

I'd appreciate your feedbacks on this

queries/svelte/highlights.scm
```diff
@@ -8,6 +8,17 @@
   (as)
 ] @keyword
 
+((special_block_keyword) @keyword.coroutine
+  (#match? @keyword.coroutine "^(await|then)$"))
+
+(then) @keyword.coroutine
```
elianiva, line 14: I don't think we should include .then since it's just a function call, not a special syntax

amaanq: Which then (match or rule)? Or both?

elianiva: both, it should be the same colour as a regular function
nothing special to it IMO

amaanq: Sure, thanks for your input!

queries/racket/highlights.scm
```diff
@@ -71,9 +71,15 @@
...
...t-out" "provide" "provide-signature-elements" "provide/contract" "public" "public*" "public-final" "public-final*" "pubment" "pubment*" "quasiquote" "quasisyntax" "quasisyntax/loc" "quote" "quote-syntax" "quote-syntax/prune" "recontract-out" "recursive-contract" "relative-in" "rename" "rename-in" "rename-inner" "rename-out" "rename-super" "require" "send" "send*" "send+" "send-generic" "send/apply" "send/keyword-apply" "set!" "set!-values" "set-field!" "shared" "stream" "stream*" "stream-cons" "stream-lazy" "struct" "struct*" "struct-copy" "struct-field-index" "struct-guard/c" "struct-out" "struct/c" "struct/contract" "struct/ctc" "struct/dc" "struct/derived" "submod" "super" "super-instantiate" "super-make-object" "super-new" "syntax" "syntax-case" "syntax-case*" "syntax-id-rules" "syntax-rules" "syntax/loc" "tag" "this" "this%" "thunk" "thunk*" "time" "unconstrained-domain->" "unit" "unit-from-context" "unit/c" "unit/new-import-export" "unit/s" "unless" "unquote" "unquote-splicing" "unsyntax" "unsyntax-splicing" "values/drop" "when" "with-continuation-mark" "with-contract" "with-contract-continuation-mark" "with-handlers" "with-handlers*" "with-method" "with-syntax" "~?" "~@" "λ"
```
6cdh, line 74: Don't change this. It's generated by a script that mentioned in the comment. Just add the query after it so it will override queries before it.

amaanq: Gotcha, thanks!

amaanq:
Clojure: (`@sogaiu` said to reach out to someone else) @dannyfreeman, while you do use Emacs I'd like your opinion on this since you were the last person to open a Clojure-related issue

queries/racket/highlights.scm
```diff
@@ -74,6 +74,12 @@
...
    ))
 
+(list
+  .
+  (symbol) @keyword.coroutine
+  (#any-of? @keyword.coroutine
+  "for/async" "for*/async"))
```
6cdh, line 81: I'm sorry to comment this again. But `for/async`, `for*/async` are not coroutine equivalent in Racket.
There are no coroutines in Racket. `engine` like coroutine but it can be preempted by a external
trigger. Another way is to use continuation.
I think we should cancel racket support here.

ahelwer:
For TLA+ the `await` keyword doesn't really have anything to do with coroutines, it just means "wait until this condition is satisfied"

queries/racket/highlights.scm
amaanq: No don't be sorry, I appreciate the feedback! I'll remove the changes to racket now

amaanq: Btw, is this relevant? https://docs.racket-lang.org/reference/async-channel.html

amaanq:
> > Clojure: (`@sogaiu` said to reach out to someone else) @dannyfreeman, while you do use Emacs I'd like your opinion on this since you were the last person to open a Clojure-related issue
> 
> The await symbols seem fine. I think some of the clojure core.async primitives could also fall under this category, but they are a (popular) library, not a part of core clojure. You would need to change
> 
> https://github.com/nvim-treesitter/nvim-treesitter/pull/4382/files#diff-dafc3f7b515f6050f09fb4565dcbf4c1fdc8ea5ec30c36add4c50fee7341ee0cR115
> 
> from
> 
> ```
> ((sym_lit) @keyword.coroutine
> ```
> 
> to
> 
> ```
> ((sym_name) @keyword.coroutine
> ```
> 
> Since the core.async names could be namespaced
> 
> and add the clojure core.async primitives
> 
> `<! <!! >! >!!` and perhaps add some of the other functions and macros offered by core async that operate on channels.

I know almost nothing about Clojure so I'm making assumptions here, but aren't sym_names not keywords? That'd be more of a function/method call than keyword. And `<! <!! >! >!!` seem like operators but I could be wrong

ahelwer:
@amaanq yeah sure that works for TLA+ then

queries/racket/highlights.scm
6cdh: No. They are different with coroutine.
`async-channel` is for synchronize between threads.
`future` is for parallel, not concurrency.

People can implements coroutines in continuation, and give its name whatever they want.
It should not exceed 30 lines of code. That's why Racket say it has coroutine support.

amaanq: ty!

amaanq:
@theHamsta this seems ready now

clason:
>  I do not use neovim

The hope is that these queries can be useful for all editors that use tree-sitter for highlighting -- even Emacs ;)

theHamsta:
Ready to merge??

amaanq:
Godot was updated recently, but otherwise yes this is ready

amaanq:
Godot just got updated, otherwise yes!

theHamsta (approved):

PR #4389, opened 2023-02-26
feat(fortran): update highlight queries

amaanq:
Better alternative to #4342, also fills in the missing blanks

### DIFF
--- a/queries/fortran/highlights.scm
+++ b/queries/fortran/highlights.scm
@@ -1,43 +1,40 @@
-(identifier) @variable
-(string_literal) @string
-(number_literal) @number
-(boolean_literal) @boolean
-(comment) @comment
+; Preprocs
+
+(preproc_file_line) @preproc
+
+; Namespaces
+
+(program_statement
+  (name) @namespace)
+
+(end_program_statement
+  (name) @namespace)
+
+(module_statement
+  (name) @namespace)
+
+(end_module_statement
+  (name) @namespace)
+
+(submodule_statement
+  (name) @namespace)
+
+(end_submodule_statement
+  (name) @namespace)
+
+; Includes
 
 [
-  (intrinsic_type)
-  "pointer"
-] @type
-
-[
-  "allocatable"
-  "attributes"
-  "device"
-  "dimension"
-  "global"
-  "grid_global"
-  "host"
-  "optional"
-  "private"
-  "public"
-  "value"
-] @type.qualifier
-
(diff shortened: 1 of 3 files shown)


queries/fortran/highlights.scm
```diff
@@ -1,114 +1,272 @@
...
-  "import"
-] @include
+; Labels
 
 [
-  (none)
-  "implicit"
-  "intent"
-] @attribute
+  "goto"
```
theHamsta, line 159: I'm not sure if that is what `@label` is supposed to be. Does C this? Actually I would expect  `@label` for what comes after `@goto`

amaanq: Oops, I mistook what `@label` was for - fixed

amaanq:
FYI I can pick up fortran as a maintainer, spent a good amount of time researching all these keywords :slightly_smiling_face:

queries/fortran/highlights.scm
```diff
@@ -1,114 +1,274 @@
...
+] @type.builtin
+
+; Typedefs
+
+[
+  "classis"
+  "endtype"
+  "type"
+  "typeis"
+] @type.definition
```
ObserverOfTime, line 108: These should be `@keyword`. The ever-misleading `@type.definition` capture is for type aliases.

ObserverOfTime (changes requested):

ObserverOfTime (approved):

queries/fortran/highlights.scm
```diff
@@ -117,69 +272,66 @@
...
-(module_statement
-  (name) @namespace)
+(string_literal) @string
 
-(function_statement
-  (name) @function)
+(assumed_size) @character.special
+(data_value "*" @character.special)
+(format_identifier "*" @character.special)
+(unit_identifier "*" @character.special)
```
theHamsta, line 320: `@character` is for char literals

amaanq: those all are just asterisks which are special

theHamsta: But they are not character literals. CharSpecial is another of those confusing categories. I could not think of a single accusation where this could be needed.

clason: Think `\n` or `\t`.

amaanq: Okay I'll just leave it under `@operator` then

PR #4391, opened 2023-02-27
cmake: simplify highlight queries

uyha:
Currently, the highlight queries try to match the commands with their keywords according to the CMake document. This approach performs poorly with the current state of nvim's tree-sitter. So I simplify it for now to just assume that unquoted arguments are keywords for function calls. Since this is the convention for CMake so I hope that it works well. This should also improve the startup time when opening CMake files.


amaanq:
I don't get the point of this, are you experiencing major editor slowdown/lag with the current queries? Do you have a reproducible file to test on?

uyha:
well, first is this [issue](https://github.com/nvim-treesitter/nvim-treesitter/issues/1956) (pun intended) and then the startup time when opening a CMake is noticable also as indicated [here](https://github.com/uyha/tree-sitter-cmake/issues/11).

uyha:
> I still think this removes a lot of functionality. Some of the bigger matches/any-ofs could be removed, sure, but try incrementally testing with smaller ones..like:
> 
> (normal_command
>   (identifier) @_function
>   (#match? @_function "\\c^unset$")
>   . (argument)
>   (argument) @storageclass
>   (#any-of? @storageclass "CACHE" "PARENT_SCOPE")
> )
Oh, I totally forgot that I have those queries there also, thanks for catching that.

> And for any [A-Z]* keywords (all caps), what do you think about default highlighting it as @constant? (Before the matches due to query precedence)
I'm not sure i understand this correctly, do you mean to move the query to the beginning so it can be overridden by following queries?

amaanq:
> I'm not sure i understand this correctly, do you mean to move the query to the beginning so it can be overridden by following queries?

Yes, it'd be worthwhile have a general [A-Z] constant match at the beginning (or at least before all the in-depth matches)

For example, there seems to be this:
```scheme
(normal_command
  (identifier)
  (argument (unquoted_argument)) @constant
  (#match? @constant "^[A-Z][A-Z_]+$")
)
```
Which is pretty much what I was talking about, except for the actual identifier instead of argument as well.

uyha:
> Which is pretty much what I was talking about, except for the actual identifier instead of argument as well.

Oh no, `argument` is what needs to be there, not `identifier`. For example, given `message(STATUS Hello)`, the `identifier` in this case refers to `message`, while `argument` will refer to `STATUS` and `Hello`. What should be highlighted as a `@constant` is `STATUS` not `message`.

amaanq:
The first issue is definitely weird, 

> 

Oh, maybe I misinterpreted that query wrong.  Say we have a function/rule, any CAPS identifiers inside that function/rule (or at the top level) should be `@constant`

uyha:
Maybe the name is confusing but `identifier` does not bind to the arguments in the function call. It only binds to the name of the function, so there is no `identifier` inside that function.

uyha:
but they are in the grammar already

queries/cmake/highlights.scm
```diff
@@ -1,3 +1,9 @@
+(normal_command
+  (identifier)
+  (argument (unquoted_argument)) @constant
+  (#match? @constant "^[A-Z][A-Z_]+$")
```
theHamsta, line 4: Does this work also with lua-match?

uyha: I am not sure what you mean

theHamsta (approved):

queries/cmake/highlights.scm
theHamsta: `lua-match?` uses the Lua implementation fro regular expressions which is more efficient than vim's `match?`. `match?` supportsore features than Lua patterns

Sorry! I forgot about this comment. The PR looks great apart from that minor considation

uyha: oh, I have no idea about `lua-match?`, I will try it out.

ObserverOfTime:
> but they are in the grammar already

I meant `NOT`, `AND`, `OR`, etc.

PR #4393, opened 2023-02-27
feat: add prql

matthias-Q:
Adds a parser highlights and injections for PRQL. I am willing to maintain this parser.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1061,6 +1061,16 @@ list.proto = {
   filetype = "proto",
 }
 
+list.prql = {
+  install_info = {
+    url = "https://github.com/PRQL/tree-sitter-prql",
+    branch = "main",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@matthias-Q" },
+  filetype = "prql",
+}
+
 list.pug = {
   install_info = {
     url = "https://github.com/zealot128/tree-sitter-pug",
--- a/queries/prql/highlights.scm
+++ b/queries/prql/highlights.scm
@@ -1,0 +1,140 @@
+[
+  (keyword_from)
+  (keyword_filter)
+  (keyword_derive)
+  (keyword_group)
+  (keyword_aggregate)
+  (keyword_sort)
+  (keyword_take)
+  (keyword_window)
+  (keyword_join)
+  (keyword_select)
+  (keyword_switch)
+  (keyword_inner)
+  (keyword_right)
+  (keyword_full)
+  (keyword_in)
+  (keyword_rolling)
+  (keyword_rows)
+  (keyword_expanding)
+  (keyword_let)
+  (keyword_prql)
+  (keyword_switch)
+  (keyword_append)
+  (keyword_remove)
+  (keyword_intersect)
+] @keyword
+
+(literal) @string
+
+(assignment
+  alias: (field) @field)
+
+alias: (identifier) @field
+
+(f_string) @string.special
+(s_string) @string.special
+
+(comment) @comment
+
+(keyword_func) @keyword.function
+
+(function_call
+  (identifier) @function.call)
+
+[
+  "+"
+  "-"
+  "*"
+  "/"
+  "="
+  "=="
+  "<"
+  "<="
+  "!="
+  ">="
+  ">"
+  "->"
+  (bang)
+] @operator
+
+[
+  "("
+  ")"
+  "["
+  "]"
+] @punctuation.bracket
+
+[
+  ","
+  "."
+  (pipe)
+] @punctuation.delimiter
+
+(literal
+  (integer) @number)
+
+(integer) @number
+
+(literal
+  (decimal_number) @float)
+
+(decimal_number) @float
+
+[
+  (keyword_min)
+  (keyword_max)
+  (keyword_count)
+  (keyword_count_distinct)
+  (keyword_average)
+  (keyword_avg)
+  (keyword_sum)
+  (keyword_stddev)
+  (keyword_count)
+] @function.call
+
+[
+ (keyword_side)
+ (keyword_version)
+ (keyword_target)
+ (keyword_null)
+ (keyword_format)
+] @attribute
+
+(target) @function.builtin
+
+ [
+  (date)
+  (time)
+  (timestamp)
+] @string.special
+
+[
+  (keyword_left)
+  (keyword_inner)
+  (keyword_right)
+  (keyword_full)
+  (keyword_csv)
+  (keyword_json)
+] @method
+
+[
+  (keyword_true)
+  (keyword_false)
+] @boolean
+
+[
+ (keyword_and)
+ (keyword_or)
+] @keyword.operator
+
+(function_definition
+  (keyword_func)
+  name: (identifier) @function)
+
+(parameter
+  (identifier) @parameter)
+
+(variable
+  (keyword_let)
+  name: (identifier) @constant)
--- a/queries/prql/highlights.scm
+++ b/queries/prql/highlights.scm
@@ -111,7 +111,7 @@ alias: (identifier) @field
   (keyword_full)
   (keyword_csv)
   (keyword_json)
-] @method
+] @method.call
 
 [
   (keyword_true)
--- a/queries/prql/highlights.scm
+++ b/queries/prql/highlights.scm
@@ -53,7 +53,6 @@ alias: (identifier) @field
   "!="
   ">="
   ">"
-  "->"
   (bang)
 ] @operator
 
@@ -68,6 +67,7 @@ alias: (identifier) @field
   ","
   "."
   (pipe)
+  "->"
 ] @punctuation.delimiter
 
 (integer) @number
--- a/queries/prql/highlights.scm
+++ b/queries/prql/highlights.scm
@@ -70,13 +70,7 @@ alias: (identifier) @field
   (pipe)
 ] @punctuation.delimiter
 
-(literal
-  (integer) @number)
-
 (integer) @number
-
-(literal
-  (decimal_number) @float)
 
 (decimal_number) @float
 
--- a/queries/prql/highlights.scm
+++ b/queries/prql/highlights.scm
@@ -97,7 +97,6 @@ alias: (identifier) @field
  (keyword_side)
  (keyword_version)
  (keyword_target)
- (keyword_null)
  (keyword_format)
 ] @attribute
 
@@ -138,3 +137,5 @@ alias: (identifier) @field
 (variable
   (keyword_let)
   name: (identifier) @constant)
+
+ (keyword_null) @constant.builtin
--- a/queries/prql/highlights.scm
+++ b/queries/prql/highlights.scm
@@ -10,32 +10,31 @@
   (keyword_join)
   (keyword_select)
   (keyword_switch)
-  (keyword_inner)
-  (keyword_right)
-  (keyword_full)
-  (keyword_in)
+  (keyword_append)
+  (keyword_remove)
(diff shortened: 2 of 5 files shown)


amaanq:
ci error is unrelated

queries/prql/highlights.scm
```diff
@@ -0,0 +1,140 @@
...
+(function_definition
+  (keyword_func)
+  name: (identifier) @function)
+
+(parameter
+  (identifier) @parameter)
+
+(variable
+  (keyword_let)
+  name: (identifier) @constant)
```
ObserverOfTime, line 140: Not `@variable`?

queries/prql/highlights.scm
```diff
@@ -0,0 +1,140 @@
...
+] @string.special
+
+[
+  (keyword_left)
+  (keyword_inner)
+  (keyword_right)
+  (keyword_full)
+  (keyword_csv)
+  (keyword_json)
+] @method
```
ObserverOfTime, line 119: I don't think `@method` applies here.

queries/prql/highlights.scm
```diff
@@ -0,0 +1,140 @@
...
+  (keyword_sum)
+  (keyword_stddev)
+  (keyword_count)
+] @function.call
+
+[
+ (keyword_side)
+ (keyword_version)
+ (keyword_target)
+ (keyword_null)
```
ObserverOfTime, line 100: `(keyword_null)` should probably be `@constant.builtin`.

queries/prql/highlights.scm
```diff
@@ -0,0 +1,140 @@
...
+  (keyword_remove)
+  (keyword_intersect)
+] @keyword
+
+(literal) @string
+
+(assignment
+  alias: (field) @field)
+
+alias: (identifier) @field
```
ObserverOfTime, line 33: For clarity, since this confused me a little.
```suggestion
(alias: (identifier) @field)
```

queries/prql/highlights.scm
matthias-Q: Yeah, I know that `@method` is meant for things like class methods, but this language is fairly simplistic. In this case it refers to the join method or the "parse that file" method.

queries/prql/highlights.scm
matthias-Q: If I add the parenthesis, the highlights will break

matthias-Q: No, same error

amaanq (changes requested):

queries/prql/highlights.scm
```diff
@@ -0,0 +1,140 @@
...
+  (keyword_min)
+  (keyword_max)
+  (keyword_count)
+  (keyword_count_distinct)
+  (keyword_average)
+  (keyword_avg)
+  (keyword_sum)
+  (keyword_stddev)
+  (keyword_count)
+] @function.call
```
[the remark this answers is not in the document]
matthias-Q: I don't know. There are so few functions/builtin functions that I do not want to distinguish between these by color.

queries/prql/highlights.scm
```diff
@@ -0,0 +1,140 @@
...
+  (keyword_expanding)
+  (keyword_let)
+  (keyword_prql)
+  (keyword_switch)
+  (keyword_append)
+  (keyword_remove)
+  (keyword_intersect)
+] @keyword
+
+(literal) @string
```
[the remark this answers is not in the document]
matthias-Q: Since PRQL is close to SQL, I was reusing the node definition for `$.literal` from over there.

lua/nvim-treesitter/parsers.lua
```diff
@@ -1061,6 +1061,16 @@ list.proto = {
   filetype = "proto",
 }
 
+list.prql = {
+  install_info = {
+    url = "https://github.com/PRQL/tree-sitter-prql",
+    branch = "main",
```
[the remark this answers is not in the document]
matthias-Q: Isn't explicit not better than implicit? Besides, alot of other parsers do seem to specify the default branch regardless

lua/nvim-treesitter/parsers.lua
```diff
@@ -1061,6 +1061,16 @@ list.proto = {
...
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@matthias-Q" },
+  filetype = "prql",
```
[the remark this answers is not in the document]
matthias-Q: same as previous comment. I think being explicit here is better.

queries/prql/highlights.scm
```diff
@@ -0,0 +1,140 @@
...
+
+(assignment
+  alias: (field) @field)
+
+alias: (identifier) @field
+
+(f_string) @string.special
+(s_string) @string.special
```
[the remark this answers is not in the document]
matthias-Q: Parsing interpolated strings is not supported upstream as of now. I can create an issue and try work on that in the next days.
I would convert these special strings then to normal strings for now.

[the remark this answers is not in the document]
matthias-Q: I have created https://github.com/PRQL/tree-sitter-prql/issues/1 upstream to parse interpolated strings in the future.

queries/prql/highlights.scm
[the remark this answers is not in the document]
amaanq: Gotcha, might be good to rename _literal_string to string then

queries/prql/highlights.scm
```diff
@@ -0,0 +1,140 @@
...
+  (keyword_rolling)
+  (keyword_rows)
+  (keyword_expanding)
+  (keyword_let)
+  (keyword_prql)
+  (keyword_switch)
+  (keyword_append)
+  (keyword_remove)
+  (keyword_intersect)
+] @keyword
```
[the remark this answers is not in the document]
matthias-Q: What others are missing? I just double checked, but as far as I can see, all of the keywords that are currently supported are in the highlights.

queries/prql/highlights.scm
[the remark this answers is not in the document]
amaanq: I think that writing your queries based on your colorschemes highlighting is not the right way to go about this. If that's a concern of yours, you can link the highlight groups for that specific filetype in your config.

Please be as correct as possible with the queries :)

lua/nvim-treesitter/parsers.lua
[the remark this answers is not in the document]
amaanq: Yes, it was a far-back todo of mine to clean that up FYI

lua/nvim-treesitter/parsers.lua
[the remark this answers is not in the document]
amaanq: explicit in the prior one makes a *bit* more sense, definitely not here.

queries/prql/highlights.scm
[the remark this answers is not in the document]
amaanq: I can help improve the grammar in a couple days (or earlier if time permits)

queries/prql/highlights.scm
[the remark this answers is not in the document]
amaanq: Was just that one, oops

queries/prql/highlights.scm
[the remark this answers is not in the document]
matthias-Q: No, that was not the concern. The "problem" with languages like PRQL or SQL is that you have a lot of keyworks that are kind of function call and they follow each other. Thus everything is in the same color and therefore useless. The contribution guideline also states that since languanges are vastly different one could be a bit more free in choosing the "marker" like `@function.call`

queries/prql/highlights.scm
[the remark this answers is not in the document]
matthias-Q: Yes, for sure. But this PR is not the right place to discuss the structure of the parser itself. The grammar is use elsewhere, and this PR is focused on add the queries to nvim-treesitter.
I'll create another issue in the upstream repo.

queries/prql/highlights.scm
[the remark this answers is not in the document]
ObserverOfTime: I believe `@function.call` is fine here.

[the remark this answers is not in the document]
amaanq: >  The "problem" with languages like PRQL or SQL is that you have a lot of keyworks that are kind of function call and they follow each other.

Thanks for the clarification, this is ok then

queries/prql/highlights.scm
```diff
@@ -0,0 +1,147 @@
...
+[
+  ","
+  "."
+  (pipe)
+] @punctuation.delimiter
+
+(literal
+  (integer) @number)
+
+(integer) @number
```
amaanq, line 76: redundant

queries/prql/highlights.scm
```diff
@@ -0,0 +1,147 @@
...
+
+(literal
+  (decimal_number) @float)
+
+(decimal_number) @float
```
amaanq, line 81: also redundant

queries/prql/highlights.scm
```diff
@@ -0,0 +1,147 @@
...
+  "*"
+  "/"
+  "="
+  "=="
+  "<"
+  "<="
+  "!="
+  ">="
+  ">"
+  "->"
```
amaanq, line 56: `->` could be `@punctuation.delimiter`, just commenting to double check

matthias-Q: I think `@operator` fits quite well.

amaanq: I read the PRQL book a bit, `->`doesn't operate on anything, but more delimits function parameters and the body, and switch statement cases. So I think it fits better in `@punctuation.delimiter`

matthias-Q: Yes, you are probably right. I changed it.

amaanq (approved):

amaanq:
Ty! LGTM

queries/prql/highlights.scm
amaanq: Yeah, I think C/C++ are the only languages I can think of off the top of my head where -> is actually an operator, oftentimes it's a delimiter :)

queries/prql/highlights.scm
ObserverOfTime: I'm still not convinced we should use `@method` here. What does SQL use?

matthias-Q: Consider this example:
```
from_text format:csv """
a,b,c
1,2,3
"""

from invoices
join side:left employees [==id]
join side:inner employees [==id]
join side:right employees [==id]
```
Keywords are `from_text`, `from` and `join` while `csv` or the join type are always instruction on how to perform the previous action (e.g. from_text or join)
Therefore I find the label `@method` more fitting than `keyword`

SQL is a bad example. I did a cleanup of these queries last week because almost everything was a keyword, hence some statements consists of uniformly colored words.

amaanq: Maybe `@method.call` then? Aren't you `calling` json on the table/data?

matthias-Q: Yes, that could be an alternative. I will change it later this day.

amaanq: @ObserverOfTime

amaanq:
> @amaanq should I rebase & squash commits, or will you do it when merging?

I'll squash when it's ready

lua/nvim-treesitter/parsers.lua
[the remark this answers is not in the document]
theHamsta: It's needed if it's not master

theHamsta (approved):

ObserverOfTime (approved):

amaanq:
Squashed & Merged, thanks for maintaining!

PR #4395, opened 2023-02-28
feat: add passwd

amaanq:
Pretty much just /etc/passwd in Linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [ocamllex](https://github.com/atom-ocaml/tree-sitter-ocamllex) (maintained by @undu)
 - [ ] [org](https://github.com/milisims/tree-sitter-org)
 - [x] [pascal](https://github.com/Isopod/tree-sitter-pascal.git) (maintained by @Isopod)
+- [x] [passwd](https://github.com/ath3/tree-sitter-passwd) (maintained by @amaanq)
 - [x] [perl](https://github.com/ganezdragon/tree-sitter-perl) (maintained by @lcrownover)
 - [x] [php](https://github.com/tree-sitter/tree-sitter-php) (maintained by @tk-shirasaka)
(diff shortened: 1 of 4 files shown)


ObserverOfTime (changes requested):
It might be worth differentiating `home` and `shell`.
Perhaps `@constant` and `@string.special` respectively?

amaanq:
> It might be worth differentiating `home` and `shell`. Perhaps `@constant` and `@string.special` respectively?

I agree, how about `@variable` for home? It's not really a constant imo

ObserverOfTime:
I mean, it shouldn't change so...

amaanq:
> I mean, it shouldn't change so...

Yeah that's true, alright

ObserverOfTime:
Does `@text.uri` overwrite the other highlights or does it just add an underline? I can't check at the moment.

amaanq:
> Does `@text.uri` overwrite the other highlights or does it just add an underline? I can't check at the moment.

Underline, see:
image

ObserverOfTime:
I see that overwrites them. 🤔

amaanq:
> I see that overwrites them. 🤔

I didn't replace `home` with `@constant` in that screenshot yet, the blue is `@string.special`

ObserverOfTime:
It seems that `@text.uri` has to be specified first, otherwise it overwrites the rest.

amaanq:
> It seems that `@text.uri` has to be specified first, otherwise it overwrites the rest.

In what way? I don't notice that in my theme at least, but I'll change the order either way

ObserverOfTime:
> In what way? I don't notice that in my theme at least, but I'll change the order either way

I'm assuming your theme only underlines `@text.uri`. It's coloured in mine.

ObserverOfTime (approved):

amaanq:
> > In what way? I don't notice that in my theme at least, but I'll change the order either way
> 
> I'm assuming your theme only underlines `@text.uri`. It's coloured in mine.

ahhh thanks, didn't think about that

PR #4396, opened 2023-02-28
fix(eex): map eelixir filetype to the eex parser

mhanberg:
The filetype for the `eex` is actually called `eelixir`.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -354,7 +354,7 @@ list.eex = {
     files = { "src/parser.c" },
     branch = "main",
   },
-  filetype = "eex",
+  filetype = "eelixir",
   maintainers = { "@connorlay" },
 }
 


amaanq:
Shouldn't this go under the specific parser?

e.g.
```lua
list.pug = {
  install_info = {
    url = "https://github.com/zealot128/tree-sitter-pug",
    files = { "src/parser.c", "src/scanner.cc" },
  },
  maintainers = { "@zealot128" },
  filetype = "pug",
  experimental = true,
}
```

So put filetype="eelixir" accordingly.. maybe

mhanberg:
Ahh, I see. the map at the beginning is really only for parsers that have multiple filetypes.

I'll make that change.

amaanq:
Thanks! Will merge once CI finishes

amaanq (approved):

amaanq:
I noticed you have a parser for monkey, do you have any interest in adding/maintaining that here? Would make it easier for ppl to install :)

PR #4397, opened 2023-02-28
feat: add matlab

amaanq:
This PR adds Matlab support...

But I'm not a fan of the grammar in its current state (when I do rewrite it and if the PR is merged I'll add locals & indents as I'd rather not spend time on that now)


ObserverOfTime:
Does it also parse `scilab` and `octave` files properly? If so, they should be added to the aliases.

amaanq:
No, while very similar to Octave, there's a few minute differences that wouldn't make it work well with this. In fact, now that you brought it to my attention, octave could easily be extended from Matlab.

Some notes I pulled online:
* Octave supports C-style autoincrement and assignment operators.
* MATLAB differentiates between character strings and character arrays, while Octave does not.
* Octave supports both printf and fprintf as a command for printing to the screen. MATLAB requires fprintf.
* MATLAB does not allow whitespace before the transpose operator but Octave does (it is just an operator like others).
* MATLAB always requires ... for line continuation.
* Octave supports `z = y = x + 3`
   MATLAB require `y = x + 3`, `z = y`
   
   etc...
   
   Scilab is even more different iirc

ObserverOfTime (approved):

PR #4398, opened 2023-02-28
feat(starlark): improve assertion parsing

amaanq:
Title

### DIFF
--- a/queries/starlark/highlights.scm
+++ b/queries/starlark/highlights.scm
@@ -218,7 +218,6 @@
 ] @keyword.function
 
 [
-  "assert"
   "async"
   "await"
   "exec"
@@ -255,6 +254,11 @@
 (ERROR) @error
 
 ;; Starlark-specific
+
+;; Assertion calls
+(assert_keyword) @keyword
+
+(assert_builtin) @function.builtin
 
 ;; Struct definitions
 ((call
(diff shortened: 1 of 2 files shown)


queries/starlark/highlights.scm
```diff
@@ -256,6 +255,11 @@
 
 ;; Starlark-specific
 
+;; Assertion calls
+(assert_keyword) @keyword
+
+(assert_builtin) @function.builtin
```
ObserverOfTime, line 261: What's the difference between these two?
Can you use `@exception` for either or both?

amaanq:
```
assert a = b
assert_eq(a = b)
assert.eq(a = b)
```

Are assert functions normally captured under `@exception`?

ObserverOfTime: Probably not the functions, but the first `assert` should be `@exception`.

amaanq: Python has it as `@keyword`

amaanq: Python has it as `@keyword`

ObserverOfTime: We might be using it inconsistently.
Let's keep it consistent with python here.

ObserverOfTime (approved):

queries/starlark/highlights.scm
amaanq: Agreed, especially since Starlark is just a dialect of Python

PR #4400, opened 2023-02-28
feat: add PO

amaanq:
https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html

### DIFF
--- a/queries/po/highlights.scm
+++ b/queries/po/highlights.scm
@@ -1,0 +1,33 @@
+; Keywords
+
+[
+  "msgctxt"
+  "msgid"
+  "msgid_plural"
+  "msgstr"
+  "msgstr_plural"
+] @keyword
+
+; Punctuation
+
+[ "[" "]" ] @punctuation.bracket
+
+; Literals
+
+(string) @string
+
+(escape_sequence) @string.escape
+
+(number) @number
+
+; Comments
+
+(comment) @comment @spell
+
+(comment (reference (text) @string.special.path))
+
+(comment (flag (text) @preproc))
+
+; Errors
+
+(ERROR) @error
--- a/README.md
+++ b/README.md
@@ -280,6 +280,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [php](https://github.com/tree-sitter/tree-sitter-php) (maintained by @tk-shirasaka)
 - [x] [phpdoc](https://github.com/claytonrcarter/tree-sitter-phpdoc) (experimental, maintained by @mikehaertl)
 - [x] [pioasm](https://github.com/leo60228/tree-sitter-pioasm) (maintained by @leo60228)
+- [x] [po](https://github.com/erasin/tree-sitter-po) (maintained by @amaanq)
 - [x] [Path of Exile item filter](https://github.com/ObserverOfTime/tree-sitter-poe-filter) (experimental, maintained by @ObserverOfTime)
 - [x] [prisma](https://github.com/victorhqc/tree-sitter-prisma) (maintained by @elianiva)
 - [x] [proto](https://github.com/mitchellh/tree-sitter-proto) (maintained by @fsouza)
--- a/lockfile.json
+++ b/lockfile.json
@@ -320,6 +320,9 @@
   "pioasm": {
     "revision": "924aadaf5dea2a6074d72027b064f939acf32e20"
   },
+  "po": {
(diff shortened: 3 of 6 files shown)


queries/po/highlights.scm
```diff
@@ -0,0 +1,33 @@
...
+
+(escape_sequence) @string.escape
+
+(number) @number
+
+; Comments
+
+(comment) @comment @spell
+
+(comment (reference (text) @string.special.path))
```
ObserverOfTime, line 27: I'm not keen on captures with three parts to be honest.

ObserverOfTime, line 27: I'm not keen on captures with three parts to be honest.

```suggestion
(comment (reference (text) @text.uri))
```

queries/po/highlights.scm
```diff
@@ -0,0 +1,33 @@
...
+
+(comment (flag (text) @label))
```
ObserverOfTime, line 29:
```suggestion
(comment (flag (text) @preproc))
```

queries/po/highlights.scm
```diff
@@ -0,0 +1,33 @@
...
+; Punctuation
+
+[ "[" "]" ] @punctuation.bracket
+
+; Literals
+
+(string) @string
+
+(escape_sequence) @string.escape
+
```
ObserverOfTime, line 20: Missing the highlight for format strings.

```suggestion

(string_fragment) @string.special

```

ObserverOfTime (changes requested):

queries/po/highlights.scm
clason: I don't think that's avoidable, though (especially not with our one-size-fits-none `@text` capture)...

And remember that there's builtin fallbacks, so you don't have to define a separate highlight group for it if you don't want to. There's a reason we only enforce top-level groups in CI.

clason: *Ceterum censeo*: We should replace `@text` with Helix' `@markup` hierarchy: https://docs.helix-editor.com/themes.html#syntax-highlighting

amaanq: It references a specific location in a file normally, e.g.
```
#: gnulib-lib/getopt.c:345
```

queries/po/highlights.scm
amaanq: Yeah I agree

queries/po/highlights.scm
amaanq: Do you mean like %s, %n?

queries/po/highlights.scm
ObserverOfTime: I know.

queries/po/highlights.scm
ObserverOfTime: Yep.

queries/po/highlights.scm
amaanq: > I know.

I think `@string.special` makes more sense due to the part after the file path doesn't really fit under `@text.uri`, but I'm open to adding both

queries/po/highlights.scm
amaanq: > Yep.

string_fragment captures any string content that isn't an escape sequence, so this doesn't fit. That being said, I could update the grammar to parse those as well, or wait until your format string injections is merged and inject here then.

ObserverOfTime: > string_fragment captures any string content that isn't an escape sequence, so this doesn't fit.

Oh, I didn't realise that. Can you try to capture them with `#match?` then?

queries/po/highlights.scm
ObserverOfTime: Well, it doesn't really fit `@string.special` either, but it's fine for now.

Maybe one day we'll rework all the highlight groups (including `@text` -> `@markup`).

amaanq: Mmm yeah I agree, let's rework this when that happens

queries/po/highlights.scm
amaanq: Mmm, sure but string_fragment is a token, how would I match inside that? This seems better suited for your injection (the printf format one)

The rule:

`string_fragment: _ => token.immediate(prec(1, /[^"\\]+/)),`

ObserverOfTime: I thought it matched words. Nevermind then.

ObserverOfTime (approved):

PR #4403, opened 2023-02-28
ci: fix luarocks tag release process

teto:
Now tests the rockspec during PRs


amaanq:
Does this publish the package to luarocks? Might be a better idea to rename the CI action to "Publish"

amaanq:
Ty!

.github/workflows/release.yml
```diff
@@ -1,17 +1,19 @@
 name: "release"
-on:
-  push:
-    tags:
-      - '*'
+on: push
```
mrcjkb, line 2:
```suggestion
on:
  push:
    tags: # Will upload to luarocks.org
      - "*"
  pull_request: # Will test a local install without uploading to luarocks.org
  
```

teto:
should be good now

.github/workflows/release.yml
```diff
@@ -1,17 +1,23 @@
 name: "release"
 on:
   push:
-    tags:
-      - '*'
+    tags: # Will upload to luarocks.org
+      - "*"
+  pull_request: # Will test a local install without uploading to luarocks.org
```
clason, line 6: Is that really necessary?

.github/workflows/release.yml
```diff
@@ -1,17 +1,23 @@
...
           LUAROCKS_API_KEY: ${{ secrets.LUAROCKS_API_KEY }}
         with:
+          name: nvim-treesitter
+          version: 0.1
+          license: apache 2
           detailed_description: |
             The goal of nvim-treesitter is both to provide a simple and easy way to use the interface for tree-sitter in Neovim
             and to provide some basic functionality such as highlighting based on it.
-          build_type: "make"
+          template: nvim-treesitter-luarocks.template
```
clason, line 23: this file should not live on the top level

.github/workflows/release.yml
teto: this allows to check the rockspec doesn't get broken. Some lua packages relies on `luarocks test` to run tests.
What I could do is here is run the action only when the github workflow or template is changed.

.github/workflows/release.yml
teto: sure. Should I create a "contrib" folder and move it there or to "scripts" ?

clason: I'd prefer if the luarocks files live in `contrib`, yes -- makes their status clear.

.github/workflows/release.yml
clason: Yes, that would be better. Let's start with template only.

clason (approved):

.github/workflows/release.yml
```diff
@@ -1,17 +1,26 @@
...
     runs-on: ubuntu-latest
     steps:
-      - uses: nvim-neorocks/luarocks-tag-release@v2.2.0
+      - uses: actions/checkout@v3
+      - uses: nvim-neorocks/luarocks-tag-release@v5
         env:
           LUAROCKS_API_KEY: ${{ secrets.LUAROCKS_API_KEY }}
         with:
+          name: nvim-treesitter
+          version: 0.1
```
mrcjkb, line 21: Sorry, I'm a bit late to the party.
Just noticed that the version is fixed here.
Is this left over from testing?

.github/workflows/release.yml
```diff
@@ -1,17 +1,26 @@
...
+          license: apache 2
```
mrcjkb, line 22: The action can fetch the license info from the repo metadata, so this technically isn't needed.

PR #4405, opened 2023-02-28
Update gdscript queries

PrestonKnopp:
Updates highlights, locals, and indents and adds folds. Also add myself as maintainer.

### DIFF
--- a/queries/gdscript/highlights.scm
+++ b/queries/gdscript/highlights.scm
@@ -78,8 +78,8 @@
 ((identifier) @variable.builtin
   (#any-of? @variable.builtin "self" "super"))
 
-((attribute (identifier) (attribute_call (identifier) @constructor))
- (#eq? @constructor "new"))
+(attribute_call (identifier) @keyword.operator
+ (#eq? @keyword.operator "new"))
 
 ;; Match Pattern
 
--- a/queries/gdscript/highlights.scm
+++ b/queries/gdscript/highlights.scm
@@ -5,9 +5,7 @@
 (name) @variable
 (type) @type
 (comment) @comment @spell
-(get_node) @string @symbol    ; fallback to string
-(string_name) @string @symbol
-(node_path) @string @symbol
+(string_name) @string
 (string) @string
 (float) @float
 (integer) @number
@@ -21,6 +19,11 @@
 (breakpoint_statement) @debug
 (inferred_type) @operator
 [(true) (false)] @boolean
+
+[
+  (get_node)
+  (node_path)
+] @text.uri
 
 (class_name_statement
   (name) @type) @keyword
--- a/queries/gdscript/highlights.scm
+++ b/queries/gdscript/highlights.scm
@@ -178,7 +178,7 @@
 
 ;; Bulitin Annotations
 
-((annotation (identifier) @attribute)
+((annotation "@" @attribute (identifier) @attribute)
  (#any-of? @attribute
   ; @GDScript
   "export" "export_category" "export_color_no_alpha" "export_dir"
--- a/queries/gdscript/highlights.scm
+++ b/queries/gdscript/highlights.scm
@@ -1,18 +1,24 @@
 ;; Basic
+(ERROR) @error
+
 (identifier) @variable
 (name) @variable
 (type) @type
-(comment) @comment
-(get_node) @string
+(comment) @comment @spell
+(get_node) @string @symbol    ; fallback to string
+(string_name) @string @symbol
+(node_path) @string @symbol
 (string) @string
 (float) @float
 (integer) @number
 (null) @constant
 (setter) @function
 (getter) @function
+(set_body "set" @keyword.function)
+(get_body "get" @keyword.function)
 (static_keyword) @type.qualifier
 (tool_statement) @keyword
-(breakpoint_statement) @keyword
+(breakpoint_statement) @debug
 (inferred_type) @operator
 [(true) (false)] @boolean
 
@@ -23,16 +29,13 @@
   "const" @type.qualifier
   (name) @constant)
 
-((identifier) @variable.builtin
-  (#eq? @variable.builtin "self"))
+(expression_statement (string) @comment @spell)
 
 ;; Identifier naming conventions
 ((identifier) @type
   (#lua-match? @type "^[A-Z]"))
 ((identifier) @constant
   (#lua-match? @constant "^[A-Z][A-Z_0-9]*$"))
-((identifier) @type
-  (#lua-match? @type "^[A-Z][A-Z_0-9]*$") . (_))
 
 ;; Functions
 (constructor_definition) @constructor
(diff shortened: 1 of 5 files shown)


queries/gdscript/highlights.scm
```diff
@@ -1,18 +1,24 @@
...
+
 (identifier) @variable
 (name) @variable
 (type) @type
-(comment) @comment
-(get_node) @string
+(comment) @comment @spell
+(get_node) @string @symbol    ; fallback to string
+(string_name) @string @symbol
+(node_path) @string @symbol
```
amaanq, line 10: Why are these `@symbol`? For node_path, if it's a location/uri-like string, maybe `@text.uri`?

queries/gdscript/highlights.scm
```diff
@@ -23,16 +29,13 @@
   "const" @type.qualifier
   (name) @constant)
 
-((identifier) @variable.builtin
-  (#eq? @variable.builtin "self"))
+(expression_statement (string) @comment @spell)
```
amaanq, line 32: Are strings here actually comments?

queries/gdscript/highlights.scm
```diff
@@ -123,4 +155,186 @@
   "puppetsync"
 ] @keyword
 
-"return" @keyword.return
+"func" @keyword.function
+
+[
+  "await"
```
amaanq, line 161: Just a note for #4382, will move this to `@keyword.coroutine` once this is merged

queries/gdscript/highlights.scm
```diff
@@ -123,4 +155,186 @@
...
+  "return"
+] @keyword.return
+
+(call (identifier) @keyword.return
+      (#eq? @keyword.return "yield"))
```
amaanq, line 166: Would yield also apply to `@keyword.coroutine`?

queries/gdscript/highlights.scm
```diff
@@ -123,4 +155,186 @@
...
+  ; @GDScript
+  "export" "export_category" "export_color_no_alpha" "export_dir"
+  "export_enum" "export_exp_easing" "export_file" "export_flags"
+  "export_flags_2d_navigation" "export_flags_2d_physics"
+  "export_flags_2d_render" "export_flags_3d_navigation"
+  "export_flags_3d_physics" "export_flags_3d_render" "export_global_dir"
+  "export_global_file" "export_group" "export_multiline" "export_node_path"
+  "export_placeholder" "export_range" "export_subgroup" "icon" "onready"
+  "rpc" "tool" "warning_ignore"
+ ))
```
amaanq, line 189: Are all annotation identifiers not `@attribute`? Is the annotation itself not entirely `@attribute` as well? I'll play around with these changes shortly..

queries/gdscript/indents.scm
```diff
@@ -1,24 +1,77 @@
...
   (else_clause)
 ] @branch
+
+[
+ (string)
+ (comment)
+ (array)
+ (dictionary)
+ (parenthesized_expression)
+] @auto
```
amaanq, line 28:
```suggestion
[
 (string)
 (comment)
 (array)
 (dictionary)
 (parenthesized_expression)
 (ERROR)
] @auto
```

amaanq (changes requested):

queries/gdscript/highlights.scm
PrestonKnopp: Ah yes, `@text.uri` fits better than `@symbol`.

`get_node` is sugar for getting a ref to a node in godot's "scene tree" (`$get/this/node`). It's symbol like I would say. What would you suggest?

And now that I see it `string_name` should actually just be `@string`.

queries/gdscript/highlights.scm
PrestonKnopp: Yes, strings as standalone statements are valid comments. Here's the PR to verify: https://github.com/godotengine/godot/pull/72971

queries/gdscript/highlights.scm
PrestonKnopp: Yes it would apply

queries/gdscript/highlights.scm
PrestonKnopp: I was thinking about that. There's only a handful of annotations, you can't define your own, and they are hard coded in Godot's Gdscript parser. Maybe the highlighting can show when you have a typo at glance?

> Is the annotation itself not entirely @attribute as well?

I followed queries from java: https://github.com/nvim-treesitter/nvim-treesitter/blob/3fe80dbecdf1fa88cf81a4d7b30ab0714f8c443a/queries/java/highlights.scm#L37-L40

and python: https://github.com/nvim-treesitter/nvim-treesitter/blob/3fe80dbecdf1fa88cf81a4d7b30ab0714f8c443a/queries/python/highlights.scm#L83-L87

... which I see now that python does "@" with a priority.

I'll mark the whole annotation as `@attribute`.

queries/gdscript/highlights.scm
PrestonKnopp: Great suggestion, highlight support for `@text.uri` looks really good on both `get_node` and `node_path`.

queries/gdscript/highlights.scm
PrestonKnopp: I have subscribed to #4382.

queries/gdscript/highlights.scm
amaanq: No, no tool should really be using that to determine links/paths to open, and I don't know of any. I'll leave this one up to you though, since the `$` beforehand is a bit funky but I'm not familiar with godot at all.

queries/gdscript/highlights.scm
amaanq: While it's good to gain understanding and learn the standard from other queries, it's best not to think of it as a "one size fits all" when there are rules with similar names/functionality :)

queries/gdscript/highlights.scm
PrestonKnopp: Okay, I'll keep the change to `@text.uri`.

queries/gdscript/highlights.scm
PrestonKnopp: That's good to know that the queries should be tailored specifically for the language. I added `@attribute` to the `"@"` and kept the any-of check.

clason:
Please don't merge master into a PR branch; it's better to rebase to keep history linear.

(Otherwise we'll just squash all your commits down to one.)

PrestonKnopp:
My bad, I didn't know the sync would merge rather than rebase.

queries/gdscript/highlights.scm
```diff
@@ -41,16 +47,46 @@
...
+  (#any-of? @variable.builtin "self" "super"))
+
+((attribute (identifier) (attribute_call (identifier) @constructor))
+ (#eq? @constructor "new"))
+
+;; Match Pattern
+
+[
+  (underscore)
+  (pattern_open_ending)
```
amaanq, line 88: what is pattern_open_ending? and is underscore used in a pinctuation-like manner?

queries/gdscript/highlights.scm
```diff
@@ -1,38 +1,44 @@
 ;; Basic
+(ERROR) @error
+
 (identifier) @variable
 (name) @variable
 (type) @type
-(comment) @comment
-(get_node) @string
+(comment) @comment @spell
+(string_name) @string
```
amaanq, line 8: string_name is a confusing name, can you provide more context/info as to what it is?

amaanq (changes requested):

queries/gdscript/highlights.scm
PrestonKnopp: Good catch, underscore should be renamed to `pattern_wildcard` in the grammar. Here's example usage of the two:

```gdscript
match some_value:
    ["open ended array pattern", ..]: pass
    ["match this", _, "and this"]: pass
    _: print("Match anything here")
```

queries/gdscript/highlights.scm
PrestonKnopp: Sure, it's syntax sugar for Godot's concept of interned strings. They look like:

```gdscript
&"interned string"
```

queries/gdscript/highlights.scm
```diff
@@ -41,16 +47,46 @@
...
+
+(enumerator left: (identifier) @constant)
+
+;; Special Builtins
+
+((identifier) @variable.builtin
+  (#any-of? @variable.builtin "self" "super"))
+
+((attribute (identifier) (attribute_call (identifier) @constructor))
+ (#eq? @constructor "new"))
```
theHamsta, line 82: constructor is not for the keyword new but for the type identifier to be constructed. C++ uses `@keyword.operator` for `new`.

theHamsta (approved):

queries/gdscript/highlights.scm
PrestonKnopp: Done. Also realized (attribute (identifier)) still matches all attribute calls so removed that nesting.

queries/gdscript/highlights.scm
amaanq: Go has `_` under `@constant`, and `..` under `@operator`, does that fit better here? Seems to be used similarly

PrestonKnopp: Yeah, those sound good to me.

PrestonKnopp: Done.

amaanq:
Thanks!

amaanq (approved):

PR #4413, opened 2023-03-01
feat: add `@string.documentation` and `@comment.documentation`

amaanq:
Not familiar with Julia, do give me info on more locations where it could apply @clason

### DIFF
--- a/queries/gleam/highlights.scm
+++ b/queries/gleam/highlights.scm
@@ -83,10 +83,13 @@
 
 ; Comments
 [
+  (comment)
+] @comment @spell
+
+[
   (module_comment)
   (statement_comment)
-  (comment)
-] @comment
+] @comment.documentation @spell
 
 ; Unused Identifiers
 [
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -226,8 +226,18 @@
 (prefixed_command_literal
   prefix: (identifier) @function.macro) @string.special
 
+((string_literal) @string.documentation
+  . [
+      (module_definition)
+      (abstract_definition)
+      (struct_definition)
+      (function_definition)
+      (assignment)
+      (const_declaration)
+    ])
+
 [
   (line_comment)
   (block_comment)
-] @comment
+] @comment @spell
 
--- a/queries/python/highlights.scm
+++ b/queries/python/highlights.scm
@@ -176,7 +176,18 @@
 (escape_sequence) @string.escape
 
 ; doc-strings
-(expression_statement (string) @spell)
+
+(module . (expression_statement (string) @string.documentation @spell))
+
+(class_definition
+  body:
+    (block
+      . (expression_statement (string) @string.documentation @spell)))
+
+(function_definition
+  body:
+    (block
+      . (expression_statement (string) @string.documentation @spell)))
 
 ; Tokens
 
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -148,6 +148,18 @@
   (block_comment)
 ] @comment @spell
 
+((line_comment) @comment.documentation
+  (#lua-match? @comment.documentation "^///[^/]"))
+((line_comment) @comment.documentation
+  (#lua-match? @comment.documentation "^///$"))
+((line_comment) @comment.documentation
+  (#lua-match? @comment.documentation "^//!"))
+
+((block_comment) @comment.documentation
+  (#lua-match? @comment.documentation "^/[*][*][^*].*[*]/$"))
+((block_comment) @comment.documentation
+  (#lua-match? @comment.documentation "^/[*][!]"))
+
 (boolean_literal) @boolean
 (integer_literal) @number
 (float_literal) @float
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,12 +89,13 @@ effect on highlighting. We will work on improving highlighting in the near futur
 #### Misc
 
 ```scheme
-@comment  ; line and block comments
-@error    ; syntax/parser errors
-@none     ; completely disable the highlight
-@preproc  ; various preprocessor directives & shebangs
-@define   ; preprocessor definition directives
-@operator ; symbolic operators (e.g. `+` / `*`)
+@comment               ; line and block comments
+@comment.documentation ; comments documenting code
+@error                 ; syntax/parser errors
+@none                  ; completely disable the highlight
+@preproc               ; various preprocessor directives & shebangs
+@define                ; preprocessor definition directives
+@operator              ; symbolic operators (e.g. `+` / `*`)
 ```
 
 #### Punctuation
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -108,17 +108,18 @@ effect on highlighting. We will work on improving highlighting in the near futur
 #### Literals
 
 ```scheme
-@string            ; string literals
-@string.regex      ; regular expressions
-@string.escape     ; escape sequences
-@string.special    ; other special strings (e.g. dates)
+@string               ; string literals
+@string.documentation ; string documenting code (e.g. Python docstrings)
+@string.regex         ; regular expressions
+@string.escape        ; escape sequences
+@string.special       ; other special strings (e.g. dates)
 
-@character         ; character literals
-@character.special ; special characters (e.g. wildcards)
+@character            ; character literals
+@character.special    ; special characters (e.g. wildcards)
 
-@boolean           ; boolean literals
-@number            ; numeric literals
-@float             ; floating-point number literals
+@boolean              ; boolean literals
+@number               ; numeric literals
+@float                ; floating-point number literals
 ```
 
 #### Functions
--- a/queries/ada/highlights.scm
+++ b/queries/ada/highlights.scm
@@ -94,11 +94,9 @@
 [
    "exception"
    "raise"
-]  @exception
-(comment)         @comment
-(comment)         @spell       ;; spell-check comments
-(string_literal)  @string
-(string_literal)  @spell       ;; spell-check strings
+] @exception
+(comment) @comment @spell
+(string_literal) @string @spell
 (character_literal) @string
 (numeric_literal) @number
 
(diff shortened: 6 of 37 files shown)


clason:
@savq are you OK with the changes to the Julia query?

queries/julia/highlights.scm
```diff
@@ -226,8 +226,11 @@
 (prefixed_command_literal
   prefix: (identifier) @function.macro) @string.special
 
+((string_literal) @string.documentation
+  (function_definition))
+
```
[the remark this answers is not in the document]
amaanq: How? That'd need to apply `@string` to all strings, and then check if theres a subsequent function definition to see if `@string.documentation` should be applied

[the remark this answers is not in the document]
amaanq: Or maybe I misunderstood, do you mean move this capture right after the `@string` one?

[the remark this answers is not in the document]
clason: No, I think `@string_literal` is the docstring (as opposed to `@string`)?

Otherwise this is fine, but you'd need to add more constructs -- just replicate the markdown injections.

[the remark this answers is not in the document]
amaanq: Great, thanks

queries/python/highlights.scm
```diff
@@ -176,7 +176,7 @@
 (escape_sequence) @string.escape
 
 ; doc-strings
-(expression_statement (string) @spell)
+(expression_statement (string) @string.documentation @spell)
```
ObserverOfTime, line 179: Does this capture all the docstrings?

ObserverOfTime (commented):
Can we make it `@string.docs` so it's not too long? :)

amaanq:
> Can we make it `@string.docs` so it's not too long? :)

My initial thought was to have it be `@string.doc`, decisions decisions

savq:
> @savq are you OK with the changes to the Julia query?

Yep. LGTM.

> My initial thought was to have it be @string.doc

I agree. I´d also prefer a shorter name.

queries/python/highlights.scm
amaanq: image

Yes, but maybe it captures too much? Like it'll capture random floating strings..although tbf they don't do anything code-wise so I guess that's fine to assume it's a docstring

lucario387: Shouldn't it only capture the `"""` right after a class/function definition?
Example of the tree with docstring
image
Something like
```scheme
(function_definition
  body: 
    (block
      . (expression_statement) @string.doc))
```
 would be better imo

clason: Yes, and similar for classes and whatnot. Basically flip the Julia queries ;)

clason:
No need to mark this as breaking, since the fallback to `@string` means that users won't notice any difference until they start using the new capture.

clason:
> Can we make it @string.docs so it's not too long? :)

I kinda prefer the longer form, as it's more explicit -- it's not like we have to pay for every character.

But no strong feelings either way -- I care much more about consistency, so we should use short forms for all captures in this case; and if we don't, we should use the long form here as well.

As another data point, Helix uses `@comment.documentation` for, e.g., Rust (which we should add in analogy as well), and all else being equal, I would prefer towards greater alignment so we can share more queries.

(I should also warn that Neovim core will at some point in the near future make its own decision about capture names and default links, and there's no guarantee that we will come to the same conclusions as nvim-treesitter.)

ObserverOfTime:
> (I should also warn that Neovim core will at some point in the near future make its own decision about capture names and default links, and there's no guarantee that we will come to the same conclusions as nvim-treesitter.)

Can we all work towards standardising the captures across all editors instead? :)

clason:
> Can we all work towards standardising the captures across all editors instead? :)

That is my hope, but for that to happen everyone needs to compromise somewhat. (If there were some sort of "official" captures, Neovim would certainly use them.)

amaanq:
> > (I should also warn that Neovim core will at some point in the near future make its own decision about capture names and default links, and there's no guarantee that we will come to the same conclusions as nvim-treesitter.)
> 
> Can we all work towards standardising the captures across all editors instead? :)

Yeah...this would be really nice, I'll start a discussion at helix's core repo soon

clason:
> Yeah...this would be really nice, I'll start a discussion at helix's core repo soon

I think I'll open an issue at tree-sitter directly to reach as many people as possible (including Zed). Helix is probably our best target for coordination, indeed.

> Idk where the Emacs people are at

To the best of my knowledge, that's still at https://github.com/emacs-tree-sitter/tree-sitter-langs

ObserverOfTime:
Note that GitHub also uses tree-sitter for some languages.

amaanq:
> I should also warn that Neovim core will at some point in the near future make its own decision about capture names and default links, and there's no guarantee that we will come to the same conclusions as nvim-treesitter.

Well that's a shame, I think unifying query convention even for neovim core would be great

If Helix uses `@comment.documentation` for Rust & others, then `@string.documentation` fits better after all

clason:
> Well that's a shame, I think standardizing query convention even for neovim core would be great

Well, yes, we would standardize -- just no guarantee that we would standardize on what nvim-treesitter decided... (We wouldn't deviate just for the sake of it, of course.)

> If Helix uses @comment.documentation for Rust & others, then @string.documentation fits better after all

Although it's more important to be consistent throughout our queries, I think?

clason (approved):

amaanq:
> Although it's more important to be consistent throughout our queries, I think?

Yeah, but this is our only (so far) capture related to docs

theHamsta:
> > I should also warn that Neovim core will at some point in the near future make its own decision about capture names and default links, and there's no guarantee that we will come to the same conclusions as nvim-treesitter.
> 
> Well that's a shame, I think standardizing query convention even for neovim core would be great
> 
> If Helix uses `@comment.documentation` for Rust & others, then `@string.documentation` fits better after all

It think `@comment.documentation` + `@string.documentation` makes sense, assuming that most people and color schemes could just stay on plain `@comment`/`@string` and only give emphasis if they wish to.

clason:
No, I meant generally (long capture names vs. abbreviated)

amaanq:
> No, I meant generally (long capture names vs. abbreviated)

Ahh gotcha

> It think @comment.documentation + @string.documentation makes sense, assuming that most people and color schemes could just stay on plain @comment/@string and only give emphasis if they wish to.

Would adding `@comment.documentation` be a welcome PR as well? :)

clason:
> Would adding @comment.documentation be a welcome PR as well? :)

Yes, can be in this PR, too!

clason:
No hurry! There's bound to be a number of such languages (R, Java, Go, Dart, ...) -- if it's too many, maybe a separate PR is indeed better.

queries/rust/highlights.scm
```diff
@@ -148,6 +148,12 @@
   (block_comment)
 ] @comment @spell
 
+((line_comment) @comment.documentation
+  (#match? @comment.documentation "^//[/!][^/].*"))
+
+((block_comment) @comment.documentation
+  (#match? @comment.documentation "^/\\*[\\*!][^\\*].*\\*/"))
+
```
amaanq, line 156: This doesn't match empty doc comments, like just `//!` or `///`, would like advice on how to fix that

amaanq:
Doc comments aren't done, just thought I'd get what I had done pushed for people to look at

lucario387:
Should(n't) all of these queries ignore the comment markers, if possible?

clason:
No, since we don't have separate captures for `@comment.delimiter` (or `marker`).

amaanq:
For C++, this is on the official site:

Although it is not part of the C++ standard, /** and */ are often used to indicate documentation blocks; this is legal because the second asterisk is simply treated as part of the comment. C-style comments cannot be nested.

Do we want to add support here?

amaanq:
Covered as much as I can....whew

queries/gleam/highlights.scm
```diff
@@ -82,11 +82,14 @@
...
+[
+  (comment)
+] @comment @spell
+
 [
   (module_comment)
   (statement_comment)
-  (comment)
-] @comment
+] @comment.documentatiion @spell
```
yochem, line 92:
```suggestion
] @comment.documentation @spell
```

clason:
I didn't expect these to rely so much on regex matching, which can be expensive...

Is it possible to replace at least some of them with `lua-match`ing patterns, which are cheaper?

amaanq:
> I didn't expect these to rely so much on regex matching, which can be expensive...
> 
> Is it possible to replace at least some of them with `lua-match`ing patterns, which are cheaper?

A lot rely on the fact that it starts with /** or ///, and for the former ends in */, can this be matched with Lua patterns?

queries/gleam/highlights.scm
amaanq: Good eyes! Thanks

yochem:
Yes, lua-matching can do that, see `:h lua-patterns`. But if it is just checking if it starts with `/*` and ends with `*/`, wouldn't it be even better to write a new predicate, e.g. `#starts-with?` using `vim.treesitter.query.add_predicate()` that uses `vim.startswith()` maybe?

yochem:
True, I was thinking that would be cheaper because it doesn't have to scan the whole string, but only start and end. But maybe lua does that when using a pattern like `x.*y`.

amaanq:
Lua patterns was annoying with asterisks, but for the vim matches here's an alternative I've come up with now

```scm
((block_comment) @comment.documentation
  (#lua-match? @comment.documentation "^/[*][*][^*]+[*]/$"))

((line_comment) @comment.documentation
  (#lua-match? @comment.documentation "^///[^/]"))
  ```

**Comment by amaanq** [comment]
I've moved my matches to lua-match, some things I'd like to note for anyone questioning them:

For triple slash doc comments, `///` and `/// whatever else goes here` is valid, but `////` is not, hence the double query to check for not another immediate fourth / but to match exactly `///` as well.

For multiline doc comments, `/** */` is allowed as long as there's either a newline or space or really anything between the opening 2nd asterisk and closing asterisk, so simply `/***/` won't match.

File: queries/cpp/highlights.scm [hunk 5]
```diff
@@ -189,3 +189,6 @@
   ["<" ">"] @punctuation.bracket)
 
 (literal_suffix) @operator
+
+((comment) @comment.documentation
+  (#lua-match? @comment.documentation "^/[*][*][^*].*[*]/$"))
```
ObserverOfTime, line 194: This should be in C.

amaanq: It doesn't seem to be an "official" standard whereas c++ does consider it

ObserverOfTime (approved):

PR #4332, opened 2023-02-17
fix: keymap may not exist

tomtomjhj:

PR #4415, opened 2023-03-02
fix: remove wrong/pointless tostring

tomtomjhj:
This also fixes the bug #4332


amaanq:
When writing the code docs, I noticed that inconsistency but presumed it was meant to be like that and was ok...thanks for confirming my suspicions

amaanq (approved):

amaanq:
Really been waiting for a fix for this...

PR #4419, opened 2023-03-02
Revert "injections(html): update injection queries"

lucario387:
Reverts nvim-treesitter/nvim-treesitter#4177

- The queries were faulty and the tests was not able to note it
- Update tests to properly check for faulty cases

### DIFF
--- a/queries/html_tags/injections.scm
+++ b/queries/html_tags/injections.scm
@@ -1,9 +1,10 @@
 ; <style>...</style>
 (
   (style_element
-    (start_tag
-      (tag_name) .) 
+    (start_tag) @_no_attribute
     (raw_text) @css)
+  (#match? @_no_attribute "^\\<\\s*style\\s*\\>$")
+  ; unsure why, but without escaping &lt; and &gt; the query breaks
 ) 
 
 ; <style blocking> ...</style>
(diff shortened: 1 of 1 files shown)


lucario387:
Draft for now

queries/html_tags/injections.scm
```diff
@@ -1,9 +1,10 @@
 ; <style>...</style>
 (
   (style_element
-    (start_tag
-      (tag_name) .) 
+    (start_tag) @_no_attribute
     (raw_text) @css)
+  (#match? @_no_attribute "^\\<\\s*style\\s*\\>$")
```
theHamsta, line 6: you could try lua patterns (with lua-match?). Regular match uses Vim regexes which have their own quirks.

lucario387: I've done it locally.

I'm also trying locally if the queries before were faulty due to nightly or is it because of false queries, bisecting rn

lucario387:
It seems that even reverting to this commit doesn't fix the issue, reverting even further it might be

lucario387:
I may need some help here @theHamsta 
For this codeblock
image


| Neovim's Playground | nvim-treesitter/playground |
| ------------------- | -------------------------- |
| image |

lucario387:
Should(n't) all of these queries ignore the comment markers, if possible?

clason:
Looks like Neovim doesn't show the `css` injections. Are you sure you're on the latest commit of Neovim?

lucario387:
> Looks like Neovim doesn't show the css injections. Are you sure you're on the latest commit of Neovim?
yes
```
➜ nvim -v
NVIM v0.9.0-dev-1094+ga0292b4e5
Build type: Release
LuaJIT 2.1.0-beta3

```

PR #4426, opened 2023-03-02
feat: add "exist?" predicate

lucario387:
The intention for using `not-any-of?` in #4073 was that it should inject javascript/css *only if* `lang` and `type` attribute were not found, and if it exists, use them to deduct the language

#4419: Failed

With this PR, I added a new predicate `exists?` to implement the desired behavior that should have been in #4073 

TODO: 
- [ ] Figure out why <script defer> failed in Vue files, but not HTML files


lucario387:
For some `<style scoped>`, they are considered as raw text (1), but for some, css/js is properly injected (2)
image

image

Very flaky but not sure why

PR #4429, opened 2023-03-02
highlights(sql): add missing keywords

matthias-Q:
* added missing keywords
* moved some instructional keywords to the group `@attribute` e.g.:
  * `collate`, `concurrently` or `engine`

There are some statements, which can definitely improved for example:
```sql
alter table my_table
modify column if exists val6 int;

select
    max(c) over (partition by a,b
                order by a asc nulls first
                rows between 3 preceding
                and 3 following exclude current row)
from
    tab;
```
Here, some lines are just `@keywords` with no distinction.

@DerekStride can you take a look?

### DIFF
--- a/queries/sql/highlights.scm
+++ b/queries/sql/highlights.scm
@@ -2,13 +2,16 @@
   name: (identifier) @function.call
   parameter: [(field)]? @parameter)
 
-(keyword_gist) @function.call
-(keyword_btree) @function.call
-(keyword_hash) @function.call
-(keyword_spgist) @function.call
-(keyword_gin) @function.call
-(keyword_brin) @function.call
-(keyword_float) @function.call
+[
+  (keyword_gist)
+  (keyword_btree)
+  (keyword_hash)
+  (keyword_spgist)
+  (keyword_gin)
+  (keyword_brin)
+  (keyword_float)
+  (keyword_array)
+] @function.call
 
 (count
   name: (keyword_count) @function.call
@@ -33,7 +36,7 @@
     parameter: [(literal)]?)))
 
 (literal) @string
-(comment) @comment
+(comment) @comment @spell
 (marginalia) @comment
 
 ((literal) @number
@@ -53,7 +56,6 @@
  (keyword_asc)
  (keyword_desc)
  (keyword_external)
- (keyword_stored)
  (keyword_parquet)
  (keyword_csv)
  (keyword_rcfile)
@@ -61,21 +63,27 @@
  (keyword_orc)
  (keyword_avro)
  (keyword_jsonfile)
+ (keyword_sequencefile)
  (keyword_temp)
- (keyword_cached)
  (keyword_uncached)
  (keyword_unlogged)
  (keyword_terminated)
  (keyword_escaped)
- (keyword_partitioned)
- (keyword_location)
- (keyword_lines)
  (keyword_unsigned)
  (keyword_nulls)
  (keyword_last)
  (keyword_materialized)
  (keyword_temp)
  (keyword_temporary)
+ (keyword_delimited)
+ (keyword_replication)
+ (keyword_auto_increment)
+ (keyword_default)
+ (keyword_collate)
+ (keyword_concurrently)
+ (keyword_engine)
+ (keyword_character)
+  (keyword_concurrently)
 ] @attribute
 
 [
@@ -91,7 +99,6 @@
   (keyword_where)
   (keyword_index)
   (keyword_join)
-  (keyword_on)
   (keyword_primary)
   (keyword_delete)
   (keyword_create)
@@ -107,10 +114,9 @@
   (keyword_outer)
   (keyword_inner)
   (keyword_order)
+  (keyword_partition)
   (keyword_group)
-  (keyword_by)
   (keyword_having)
-  (keyword_as)
   (keyword_limit)
   (keyword_offset)
   (keyword_table)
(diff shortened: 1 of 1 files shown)


queries/sql/highlights.scm
```diff
@@ -214,6 +255,7 @@
   "!="
   ">="
   ">"
+  "<>"
```
amaanq, line 258: Is it *just* used like <> to operate on something? Looks like brackets..but I'm probably wrong

queries/sql/highlights.scm
```diff
@@ -199,6 +238,8 @@
   (keyword_and)
   (keyword_or)
   (keyword_not)
+  (keyword_by)
+  (keyword_on)
 ] @keyword.operator
```
amaanq, line 243: Are these all casts (those with primitive-like names)? Like turn a float to int with int(2.5)?

queries/sql/highlights.scm
```diff
@@ -160,18 +163,54 @@
...
+  (keyword_sort)
+  (keyword_compute)
+  (keyword_stats)
+  (keyword_comment)
+  (keyword_partitioned)
+  (keyword_location)
+  (keyword_cached)
+  (keyword_lines)
+  (keyword_stored)
 ] @keyword
```
amaanq, line 196: It feels like some of these could be `@type.qualifier`, e.g. in C you can have `int* restrict b;` where restrict would be `@type.qualifier`

queries/sql/highlights.scm
matthias-Q: "not equal to" oeprator

queries/sql/highlights.scm
matthias-Q: That sounds like a good idea. I will test that out. Sometimes it is hard to find a reasonable pattern in SQL keywords

queries/sql/highlights.scm
matthias-Q: I am not sure if I understand that question.
`keyword_on` is mostly used in join statements
`select a.col1, b.col1 from a left join b on a.id=b.id`

`keyword_by` is used in conjunction with `keyword_order`, `keyword_group` or `keyword_partition`. But in some dialects also with `partitioned` or `sort`...

queries/sql/highlights.scm
amaanq: Anything that modifies how a variable/item behaves at compile time or runtime fits under `@type.qualifier`. Some more examples could be abstract, const, static, dynamic etc..

queries/sql/highlights.scm
```diff
@@ -160,18 +163,54 @@
...
 
+(keyword_with) @repeat
+
+(keyword_as) @label
```
DerekStride, line 200: What's the reasoning for using repeat & label here?

queries/sql/highlights.scm
DerekStride: > Sometimes it is hard to find a reasonable pattern in SQL keywords

I agree, SQL has so many different keywords it's hard to sort through them all.

queries/sql/highlights.scm
matthias-Q: This was an attempt to diversify keywords a bit.

So `with` was probably not a good idea, since it just occurs once at the start of CTEs and then only in different locations (e.g. in hive create tables...) So I would revert that to a normal `keyword`

`keyword_as` as a `@label` since it creates an alias and this can be seen as a `label`. However, this is probably not necessary and purely cosmetic.

I think I will revert both changes.

amaanq: labels are for literal identifiers that are marked as a label, such as `goto Foo` where Foo is a label

matthias-Q: I removed these two queries

queries/sql/highlights.scm
matthias-Q: I added these as `@type.qualifier` and some others as `@attribute` From my perspective they are now grouped in a reasonable manner

matthias-Q:
I went through the tests that we have in the repo and checked on some of these queries there. It might be that more keywords in `@keyword` could be moved into either `@attribute` or `@type.qualifier`. But I guess it is a good clean up for now.

queries/sql/highlights.scm
amaanq: cached, location, partitioned, and maybe computer might seem like they fit in `@storageclass`, this is usually for qualifiers or attributes that affect lifetime or memory of whatever

matthias-Q: These keywords all belong to hive specific statements:
https://impala.apache.org/docs/build/html/topics/impala_create_table.html

I would rather argue that they should be moved back to be a normal `@keyword` as they start each line of these create table statements

Some of the `@attributes` are maybe better of as a `@storageclass`
e.g. `keyword_materialized`, `keyword_temp, `keyword_logged` and the file formats

amaanq: can't speak on materialized or logged, but temp does seem to fit, and if the other 2 are related then that as well

ObserverOfTime (approved):

queries/sql/highlights.scm
```diff
@@ -52,31 +56,40 @@
...
+ (keyword_unlogged)
+ (keyword_external)
+ (keyword_parquet)
+ (keyword_csv)
+ (keyword_rcfile)
+ (keyword_textfile)
+ (keyword_orc)
+ (keyword_avro)
+ (keyword_jsonfile)
+ (keyword_sequencefile)
```
amaanq, line 91: Why are these storageclass? This is more to do with data manipulation

ObserverOfTime: Storage format. Makes sense to me.

amaanq: A little odd but sure..

PR #4430, opened 2023-03-03
fix(html): fix faulty css/js injections

lucario387:
- This partially reverts the queries added in #4073 
- Add tests for expected behavior

The intention for the queries were: Only if `lang` and `type` is not an attribute, then inject css/js. At the time, I did not understand `any-of?` so they were faulty.

For the reasonings of not noticing this, it was that I didn't code in Vue for the past 2 months (lol)
Revert back to regex-based matching to ensure correct behavior

### DIFF
--- a/queries/html_tags/injections.scm
+++ b/queries/html_tags/injections.scm
@@ -1,22 +1,12 @@
 ; <style>...</style>
-(
-  (style_element
-    (start_tag
-      (tag_name) .) 
-    (raw_text) @css)
-) 
-
 ; <style blocking> ...</style>
 ; Add "lang" to predicate check so that vue/svelte can inherit this
 ; without having this element being captured twice
 (
   (style_element
-    (start_tag
-      (attribute
-        (attribute_name) @_no_set_type))
-    (raw_text) @css)
-  (#not-any-of? @_no_set_type "type" "lang")
-) 
+    (start_tag) @_no_type_lang 
+      (#not-match? @_no_type_lang "\\s(lang|type)\\s*\\=")
+    (raw_text) @css))
 
 (
   (style_element
@@ -30,22 +20,12 @@
 )
 
 ; <script>...</script>
-(
-  (script_element
-    (start_tag
-      (tag_name) .) 
-    (raw_text) @javascript)
-) 
-
 ; <script defer>...</script>
 (
   (script_element
-    (start_tag
-      (attribute
-        (attribute_name) @_no_set_type))
-    (raw_text) @javascript)
-  (#not-any-of? @_no_set_type "type" "lang")
-) 
+    (start_tag) @_no_type_lang 
+      (#not-match? @_no_type_lang "\\s(lang|type)\\s*\\=")
+    (raw_text) @javascript))
 
 (
   (script_element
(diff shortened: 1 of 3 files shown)


queries/html_tags/injections.scm
```diff
@@ -1,22 +1,12 @@
...
   (style_element
-    (start_tag
-      (attribute
-        (attribute_name) @_no_set_type))
-    (raw_text) @css)
-  (#not-any-of? @_no_set_type "type" "lang")
-) 
+    (start_tag) @_no_type_lang
+    (raw_text) @css
+    (#not-vim-match? @_no_type_lang "\\s(lang|type)\\s*\\=")))
```
amaanq, line 9: Is #not-vim-match? needed? What are you exactly trying to capture that #not-any-of? "type" "lang" wasn't catching?

lucario387: Yes, sadly
`#not-any-of?` doesn't mean `only if these ... doesn't exist`, it just means, `if not any of these ... exists`

amaanq: Could a #lua-match? **not** matching those work? Would be faster..

lucario387: It will never work with lua-match, simply because lua-match doesn't support (...|...) syntax :)

lucario387:
Is queries grouping here fine, or should i remove the `(start_tag (tag_name) .)` one @ObserverOfTime ?

From inspecting the tree, both of these are injected for `<script></script>`, so I think removing one will suffice, but there are certain cases that the regex one is false (like 1% or less of cases)

queries/html_tags/injections.scm
```diff
@@ -1,22 +1,12 @@
 ; <style>...</style>
-(
-  (style_element
-    (start_tag
-      (tag_name) .) 
-    (raw_text) @css)
-)
```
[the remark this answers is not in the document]
ObserverOfTime: If it works it's fine. ¯\\_(ツ)\_/¯

queries/html_tags/injections.scm
```diff
@@ -55,7 +41,7 @@
         (quoted_attribute_value (attribute_value) @_javascript)))
     (raw_text) @javascript)
   (#eq? @_type "type")
-  (#eq? @_javascript "text/javascript")
+  (#any-of? @_javascript "text/javascript" "module")
```
ObserverOfTime, line 44: Can you add a test for "module" too?

lucario387: I dont think it's necessary as it's equal to text/javascript

queries/html_tags/injections.scm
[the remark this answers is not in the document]
lucario387: K then i will remove it :) 

Don't think 1% or less of use-case is worth the extra imo

lucario387:
Seems to fail due to flaky test or smth else as injection tests are still true

amaanq:
> Seems to fail due to flaky test or smth else as injection tests are still true
> 
> image

Yeah it happens ever so often, I manually re-ran it and it's fine now

lucario387:
okay squashed and added a test for module :)

ObserverOfTime (approved):

PR #4433, opened 2023-03-03
feat: add lua patterns

amaanq:
Alternative to #3578 

I sort of dislike having the classes be `@keyword`, and the grammar isn't all that great...(I can fork it if anyone thinks the queries would be better handled that way since the current one uses inline/hidden rules annoyingly)

### DIFF
--- a/queries/luap/highlights.scm
+++ b/queries/luap/highlights.scm
@@ -1,0 +1,35 @@
+"." @character
+
+[
+  (anchor_begin)
+  (anchor_end)
+] @string.escape
+
+[
+  "[" "]"
+  "(" ")"
+] @punctuation.bracket
+
+[
+  (zero_or_more)
+  (shortest_zero_or_more)
+  (one_or_more)
+  (zero_or_one)
+] @operator
+
+(range
+  from: (character) @constant
+  "-" @punctuation.delimiter
+  to: (character) @constant)
+
+(set
+  (character) @constant)
+
+(class) @keyword
+
+(negated_set
+  "^" @operator
+  (character) @constant)
+
+(balanced_match
+  (character) @parameter)
--- a/README.md
+++ b/README.md
@@ -260,6 +260,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [ledger](https://github.com/cbarrete/tree-sitter-ledger) (maintained by @cbarrete)
 - [x] [llvm](https://github.com/benwilliamgraham/tree-sitter-llvm) (maintained by @benwilliamgraham)
 - [x] [lua](https://github.com/MunifTanjim/tree-sitter-lua) (maintained by @muniftanjim)
+- [x] [lua patterns](https://github.com/vhyrro/tree-sitter-luap) (maintained by @amaanq)
 - [x] [m68k](https://github.com/grahambates/tree-sitter-m68k) (maintained by @grahambates)
 - [x] [make](https://github.com/alemuller/tree-sitter-make) (maintained by @lewis6991)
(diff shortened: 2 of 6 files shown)


clason:
maybe mark it as experimental for now, then? 

(*ceterum censeo* we _really_ need to start categorizing parsers and taking that seriously)

amaanq:
> maybe mark it as experimental for now, then?
> 
> (_ceterum censeo_ we _really_ need to start categorizing parsers and taking that seriously)

Categorize them in what way? I do try to help out with parsers that aren't the best as much as possible..but I'm only 1 person :laughing:

clason:
> Categorize them in what way? 

In tiers, like "core", "community maintained", "experimental", "broken" -- and then handle these differently in our CI/documentation (e.g., don't auto-update experimental or broken languages, and don't include them in the build tests etc.)

Basically, keep the list of maintainers updated, and drop maintainers (and parser tiers) aggressively.

This would also help with expectation management, so people could only install "core" languages and be confident that they work well and don't break too often.

> I do try to help out with parsers that aren't the best as much as possible..but I'm only 1 person 😆

Yes, precisely, and we need to make sure your efforts are not wasted ;)

amaanq:
> In tiers, like "core", "community maintained", "experimental", "broken" -- and then handle these differently in our CI/documentation (e.g., don't auto-update experimental or broken languages, and don't include them in the build tests etc.)

Yeah that is a **really** good idea, and could help with the CI breaking ever so often..would "core" be those at tree-sitter? And say a self-made parser (by me or anyone else but NOT at tree-sitter) be "community maintained"?

amaanq:
Here's some screenshots anyways, one thing I kept in mind was keeping consistency with Regex highlights:

image

image

image

clason:
> Yeah that is a **really** good idea, and could help with the CI breaking ever so often..would "core" be those at tree-sitter? And say a self-made parser (by me or anyone else but NOT at tree-sitter) be "community maintained"?

No, those would be parsers that nvim-treesitter core maintainers (like you ;)) care about enough to keep running (or that are stable enough that no maintenance is necessary).

"community" would be those maintainers we always need to ping manually (which could be automated as well).

(TL;DR: it's about maintaining the _queries_ here, not the parsers upstream.)

amaanq:
Man that windows ci is taking forever...typical

Feel free to merge whenever so that your neovim PR can be updated accordingly :)  (as long as class being captured as `@keyword` is ok to you)

amaanq:
Text inside brackets weren't highlighted as `@constant` (e.g. A and Z in A-Z), classes inside brackets weren't highlight as `@keyword` because for whatever reason the inner class node is hidden (_class) and is used there, and I added fixes to only inject inside the string content and inject in both parameters for gsub/gmatch (both can have luap but before it was only injecting into the first), and maybe some other things I forgot

ObserverOfTime (approved):

amaanq:
Switched to my fork, queries look nicer and capture better as well

amaanq:
cc @max397574 iirc you expressed interest in this

queries/luap/highlights.scm
```diff
@@ -0,0 +1,35 @@
...
+  "-" @punctuation.delimiter
+  to: (character) @constant)
+
+(set
+  (character) @constant)
+
+(class) @keyword
+
+(negated_set
+  "^" @string.regex
```
ObserverOfTime, line 31: I'd make this one `@operator` too.
```suggestion
  "^" @operator
```

ObserverOfTime (changes requested):

ObserverOfTime (approved):

PR #4435, opened 2023-03-03
fix(ecma): properly capture constants in const declarations

### LINKED ISSUE #4434: distinction between const and let variables in javascript and typescript files

The OneDarkPro colorscheme in VSCode highlights variables declared via `const` and `let` differently for JavaScript and TypeScript files. I wanted to have a similar highlighting method for my neovim theme.

From what I can see, tree-sitter sees both variable as a `@variable.typescript` type. So, it can't make a distinction between a `const` and `let` variable.

Unlike JavaScript, in Rust files, the parser is aware of this context and can highlight them accordingly. 

#### JavaScript Snippet

```typescript
let someLetVariable = "let variable";

const someConstVariable = "const variable";
```

#### Rust Snippet

```rust
fn main() {
    const PI: f64 = 3.14;
    let name = 18;

    println!("{PI}, {name}");
}
```

#### InspectTree Output JavaScript

```
(lexical_declaration); [1:1-38:0]
 (variable_declarator); [1:5-37:0]
  name: (identifier); [1:5-20:0]
  value: (string); [1:23-37:0]
   (string_fragment); [1:24-36:0]
(lexical_declaration); [3:1-44:2]
 (variable_declarator); [3:7-43:2]
  name: (identifier); [3:7-24:2]
  value: (string); [3:27-43:2]
   (string_fragment); [3:28-42:2]
```

#### InspectTree Output Rust

```
(function_item); [1:1-2:5]
 name: (identifier); [1:4-8:0]
 parameters: (parameters); [1:8-10:0]
 body: (block); [1:11-2:5]
  (const_item); [2:5-26:1]
   name: (identifier); [2:11-13:1]
   type: (primitive_type); [2:15-18:1]
   value: (float_literal); [2:21-25:1]
  (let_declaration); [3:5-19:2]
   pattern: (identifier); [3:9-13:2]
   value: (integer_literal); [3:16-18:2]
  (expression_statement); [5:5-30:4]
   (macro_invocation); [5:5-29:4]
    macro: (identifier); [5:5-12:4]
    (token_tree); [5:13-29:4]
     (expression_statement); [5:13-29:4]
      (parenthesized_expression); [5:13-29:4]
       (string_literal); [5:14-28:4]
     (string_literal); [5:14-28:4]
```

#### Neovim Highlight JavaScript

image

#### Neovim Highlight Rust

image

#### VSCode Highlight Colors

image

#### Output of `:checkhealth nvim-treesitter`

```text

Installation ~
- OK `node` found v19.4.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.

Parser/Features H L F I J
  - c              ✓ ✓ ✓ ✓ ✓
  - astro          ✓ ✓ ✓ ✓ ✓
  - yaml           ✓ ✓ ✓ ✓ ✓
  - lua            ✓ ✓ ✓ ✓ ✓
  - json           ✓ ✓ ✓ ✓ .
  - vim            ✓ ✓ ✓ . ✓
  - bash           ✓ ✓ ✓ . ✓
  - toml           ✓ ✓ ✓ ✓ ✓
  - python         ✓ ✓ ✓ ✓ ✓
  - rust           ✓ ✓ ✓ ✓ ✓
  - tsx            ✓ ✓ ✓ ✓ ✓
  - typescript     ✓ ✓ ✓ ✓ ✓
  - help           ✓ . . . x
  - c_sharp        ✓ ✓ ✓ . ✓
  - javascript     ✓ ✓ ✓ ✓ ✓
  - css            ✓ . ✓ ✓ ✓
  - java           ✓ ✓ . ✓ ✓
  - svelte         ✓ . ✓ ✓ ✓

- ERROR help(injections): .../nightly/share/nvim/runtime/lua/vim/treesitter/query.lua:261: query: invalid node type at position 14 for language help
  help(injections) is concatenated from the following files:
  | [ERROR]:"/Users/jannatinnaim/.asdf/installs/neovim/nightly/share/nvim/runtime/queries/help/injections.scm", failed to load: .../nightly/share/nvim/runtime/lua/vim/treesitter/query.lua:261: query: invalid node type at position 14 for language help
```

```text
NVIM v0.9.0-dev-1100+g1dd2424b1
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /Applications/Xcode_13.2.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fno-common -fdiagnostics-color=always -fstack-protector-strong -DINCLUDE_GENERATED_DECLARATIONS -DMIN_LOG_LEVEL=3 -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_TS_HAS_SET_MATCH_LIMIT -DNVIM_TS_HAS_SET_ALLOCATOR -DNVIM_UNIBI_HAS_VAR_FROM -I/Users/runner/work/neovim/neovim/.deps/usr/include/luajit-2.1 -I/Library/Frameworks/Mono.framework/Headers -I/Users/runner/work/neovim/neovim/.deps/usr/include -I/Users/runner/work/neovim/neovim/build/src/nvim/auto -I/Users/runner/work/neovim/neovim/build/include -I/Users/runner/work/neovim/neovim/build/cmake.config -I/Users/runner/work/neovim/neovim/src -I/Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include -I/Users/runner/work/neovim/neovim/.deps/usr/include -I/Users/runner/work/neovim/neovim/.deps/usr/include -I/Users/runner/work/neovim/neovim/.deps/usr/include -I/Users/runner/work/neovim/neovim/.deps/usr/include -I/Users/runner/work/neovim/neovim/.deps/usr/include -I/Users/runner/work/neovim/neovim/.deps/usr/include

```

### Additional context

NOTE: It didn't highlight `PI` in the rust file a different color just because it was all caps. It works the same even if it's lower cased. Just thought of putting it here after seeing how GitHub's parser highlights `PI`.

amaanq:
Fixes #4434

I also changed the const identifier to allow single letter const identifiers as well

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -24,10 +24,15 @@
  (#lua-match? @type "^[A-Z]"))
 
 ((identifier) @constant
- (#lua-match? @constant "^[A-Z_][A-Z%d_]+$"))
+ (#lua-match? @constant "^_*[A-Z][A-Z%d_]*$"))
 
 ((shorthand_property_identifier) @constant
- (#lua-match? @constant "^[A-Z_][A-Z%d_]+$"))
+ (#lua-match? @constant "^_*[A-Z][A-Z%d_]*$"))
+
+(lexical_declaration
+  "const"
+  . (variable_declarator
+      . name: (identifier) @constant))
 
 ((identifier) @variable.builtin
  (#vim-match? @variable.builtin "^(arguments|module|console|window|document)$"))
--- a/tests/query/highlights/ecma/const.js
+++ b/tests/query/highlights/ecma/const.js
@@ -1,0 +1,13 @@
+_FOO = 4
+// <- @constant
+__A__ = 2
+// <- @constant
+_ = 2
+// <- @variable
+A_B_C = 4
+// <- @constant
+_1 = 1
+// <- @variable
+
+const A = 2
+//    ^ @constant


lucario387:
"_" will be highlighted as a const, even though it is used to fill parameters only

queries/ecma/highlights.scm
```diff
@@ -24,10 +24,15 @@
  (#lua-match? @type "^[A-Z]"))
 
 ((identifier) @constant
- (#lua-match? @constant "^[A-Z_][A-Z%d_]+$"))
+ (#lua-match? @constant "^[A-Z_][A-Z%d_]*$"))
```
lucario387, line 27: `_` will be highlighted as a const, even though it is used to fill parameters only

amaanq: check now

lucario387: and now `__A` is broken.

it was working fine before

amaanq: `const A = 2`, A wouldn't match

I can fix that, oops

amaanq: Here's the fix..
image

lucario387: Think this new pattern still fails with `_1`, while the old one doesn't (but should `_1` be accepted)?

amaanq: Is _1 a legal identifier in js/ts?

lucario387: (also, think you should just add them into a testfile in ecma folder, just for the double check)

lucario387: > Is _1 a legal identifier in js/ts? And if it is, idk, that doesn't seem like a constant to me :)

It's a valid identifier, not an usual one

amaanq: lmk what you think

amaanq: > > Is _1 a legal identifier in js/ts? And if it is, idk, that doesn't seem like a constant to me :)
> 
> It's a valid identifier, not an usual one

Yeah I realized it's an identifier, sure, I mean to ask is it convention for such an identifier to be a constant? e.g. normally SCREAMING_SNAKE_CASE is for constants, but _1 doesn't really fit that

lucario387: tests look fine.

i dont think i should have the call on how `_1` should be, will let you and others decide

jNaimXIII: You should take a look at this.

https://github.com/olimorris/onedarkpro.nvim/pull/164#issuecomment-1453912862

lucario387: @jNaimXIII I did mention this yesterday
https://github.com/nvim-treesitter/nvim-treesitter/issues/4434#issuecomment-1453559966

You just need to add a `@readonly` highlight group and i think it's fine

jNaimXIII: > tests look fine.
> 
> i dont think i should have the call on how `_1` should be, will let you and others decide

In JS we usually use the `_` to denote throwaway parameters or, to introduce a variable with the same name in nested scope without loosing reference to the outer scope one.

image

So, if that's the case, the name of the variable shouldn't matter. Anything declared with a `const` should be highlighted the same as a constant.

lucario387: i think this is better to be delegated to semantic tokens highlighting in nvim 0.9 tbh

for this to highlight properly with treesitter is too costly

lucario387: The `const (smth)` highlighted as const part* is to be delegated to semantic tokens highlighting in 0.9, to be more clear

jNaimXIII: So, not something tree-sitter should be doing on its own is what you're saying. Well then, when Neovim does ship this feature, themes will need to update their highlighting for this to take effect, right?

lucario387: `@readonly` is not a token linked to a default hlgroup by default in nvim 0.9, so theme maintainers need to add it in yes

amaanq: > @jNaimXIII I did mention this yesterday [#4434 (comment)](https://github.com/nvim-treesitter/nvim-treesitter/issues/4434#issuecomment-1453559966)
> 
> You just need to add a `@readonly` highlight group and i think it's fine

Why? It's fine to distinguish constants that are declared as constants and then leave their usage whatever it may be (in most cases, `@variable`). This is partially what using local definitions (locals.scm) fixes, except that it uses an inefficient and outdated method. 

In most languages, you'd have enum member/const declarations be highlighted as constant, but then their usage isn't exactly clear and relies on either `@variable` or regex/lua matching. So for the time being, this distinction here is fine until we get some better semantic token-like highlighting by default

jNaimXIII: Just for context, I proposed this highlighting changed based on a VSCode theme called OneDarkPro. This theme highlights JS / TS code in this manner so, if you want you can refer to this.

https://github.com/Binaryify/OneDark-Pro

jNaimXIII: > Just for context, I proposed this highlighting changed based on a VSCode theme called OneDarkPro. This theme highlights JS / TS code in this manner so, if you want you can refer to this.
> 
> https://github.com/Binaryify/OneDark-Pro

As I mentioned here, https://github.com/olimorris/onedarkpro.nvim/pull/164#issuecomment-1454005720 the theme is just marking every `constant` value with that color it seems. Then again, here comes that **semantic token** part which I don't understand much about.

amaanq: It's not really about the theme, but more what is captured under `@constant` queries and const variable definition identifiers weren't, but in this PR they are

lucario387: im pretty sure this is the behavior you want @jNaimXIII ?
image

lucario387: But the question still remains, does highlighting only the declaration to a different color while follow usage of the variable stays the same color make it more or less confusing

lucario387: @jNaimXIII semantic tokens is LSP stuffs, not treesitter stuffs, lets just keep it short

jNaimXIII: Hence the issue mentioned here. https://github.com/olimorris/onedarkpro.nvim/pull/164#issuecomment-1453912862

amaanq: > Then again, here comes that semantic token part which I don't understand much about.

Given a definition marked say, `@definition.constant`, all subsequent identifiers that are the same as the definition identifier ini the same **scope** will be highlighted as `@constant`, because it is being referenced there. So, this relies on a good and accurate `locals.scm` file, where the proper rules are put under `@scope`, `@reference`, and `@definition.*`. It is not implemented the best in terms of performance, hence why it's disabled by default iirc.

lucario387: it will be very costly if only done from queries @jNaimXIII to have the behavior you expected.

it's better to wait for 0.9 and this can be much easier supported by theme maintainers

jNaimXIII: Welp, playing the waiting game it is then. Thanks for discussing this though.

amaanq: > it will be very costly if only done from queries @jNaimXIII to have the behavior you expected.
> 
> it's better to wait for 0.9 and this can be much easier supported by theme maintainers 
> 

Do you mean this PR or using locals.scm is costly?

lucario387: > Do you mean this PR or using locals.scm is costly?

I mean by the amount of queries, and the possible tree range need to be parsed for each one you may need to do to have close to the same behavior he expected

what if the variable is from a different file, is a object key, ....
there is a lot to be done

lucario387: example is if someone do
```ts
const x = {foo: "bar"} as const;
```
can you really highlight `x.foo` with `foo` being readonly with treesitter with a short and performant query? I doubt it but if it's possible then nice.
not to mention nested ones :)

jNaimXIII: > 

Now that you mention it, the highlighting isn't consistent on VSCode either.

image

image

amaanq: > what if the variable is from a different file, is a object key, ....

Different file is not tree-sitters job so this shouldn't even be considered, tree-sitter is not an LSP.
By object key, can you provide a code example so I can understand what you mean better?

amaanq: > example is if someone do
> 
> ```ts
> const x = {foo: "bar"} as const;
> ```
> 
> can you really highlight `x.foo` with `foo` being readonly with treesitter with a short and performant query? I doubt it but if it's possible then nice. not to mention nested ones :)

I don't know js/ts all that well, but is this saying the object is a constant object?

lucario387: > I don't know js/ts all that well, but is this saying the object is a constant object?

yes

jNaimXIII: You can use this as reference I guess.

image

amaanq: How I'd highlight it from tree-sitter's POV is as follows:

x <- `@constant`
. <- `@punctuation.delimiter`
foo <- `@field`

foo is still a field, whether it's constant or not doesn't mean much in terms of highlighting (at least imo, maybe it means a lot to js/ts devs), that'd just be awkward imo; But to reiterate, this is more an LSPs job anyways

lucario387: well yes, that was my question since the start of this discussion
will (partial) semantic tokens highlighting using treesitter make reading the code more or less confusing

amaanq: > You can use this as reference I guess.
> 
> image

Random note: I'm surprised `@variable.builtin` isn't highlighted differently (like console)

So, in the current state of things, `const object` would have object be highlighted as `@constant`, but subsequent references not. locals.scm adds subsequent scoped references to be highlighted `@constant` too, but at a performance cost

amaanq: > will (partial) semantic tokens highlighting using treesitter make reading the code more or less confusing

imo no, but this is language-dependent

jNaimXIII: > well yes, that was my question since the start of this discussion will (partial) semantic tokens highlighting using treesitter make reading the code more or less confusing

Take a look at the following code. It won't matter if it's partially highlighted as long as you can differ between a `const` and `let` / `parameter`. You can instantly see that `firebase` and `settings` are constants whereas `doc` is function param.

image

ObserverOfTime (approved):

jNaimXIII:
So, what exactly did this PR fix? Will it only highlight the line of declaration or the references as well?

lucario387:
Only declaration

lucario387:
Only declaration

amaanq:
...and ensure single letter capitalized identifiers are captured as `@constant` too

PR #4437, opened 2023-03-03
v: inject regex

yochem:
Inject regex for the methods in the vlib module `re`:

- [`re.regex_base`](https://modules.vlang.io/regex.html#regex_base)
- [`re.regex_opt`](https://modules.vlang.io/regex.html#regex_opt)
- [`re.RE.compile_opt`](https://modules.vlang.io/regex.html#RE.compile_opt)

Some notes:

- It only checks if the name (field of the call expression) is one of regex_base, regex_opt and compile_opt, which I think will be enough of a check, because these are fairly unusual names. But I could also check if the module name is `re`.
- Pattern has to be of node type `raw_string_literal` (thus an `r''` string), but could also include 'normal' strings if wanted
- Noticed raw strings were not highlighted, so also added the highlight for them.

### DIFF
--- a/queries/comment/highlights.scm
+++ b/queries/comment/highlights.scm
@@ -1,39 +1,39 @@
 (_) @spell
 
 ((tag
-  (name) @text.todo
+  (name) @text.todo @nospell
   ("(" @punctuation.bracket (user) @constant ")" @punctuation.bracket)?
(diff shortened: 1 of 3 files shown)


yochem:
Sorry, something went wrong with the branch, I'll open a new PR.

PR #4440, opened 2023-03-04
fix(go): Indent on type switch case properly

evanphx:
The type switch construct emits a `type_case` node rather than an `expression_case`, so we need to explicitly ident on it.

### DIFF
--- a/queries/go/indents.scm
+++ b/queries/go/indents.scm
@@ -7,6 +7,7 @@
   (literal_value)
   (expression_case)
   (communication_case)
+  (type_case)
   (default_case)
   (block)
   (call_expression)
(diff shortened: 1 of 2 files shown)


amaanq:
Good catch!

amaanq:
Can you rename the commit to what I have in the title?

evanphx:
Sure, do you mean the commit message?

amaanq:
yes

evanphx:
Looked at the other commits, I gotcha. It's renamed now.

amaanq (approved):

evanphx:
Thanks for the quick turnaround on this PR!

PR #4455, opened 2023-03-06
indents(vue): Update dedent logic for <template>

lucario387:
"</" was previously used as a workaround to pass tests, but doesn't really work well in actual usage

"</" usage was also replicating usage for JSX template fragment, but in actuality it should be similar to normal HTML elements instead of that

Changing dedent logic to `(end_tag)` instead of relying on anchors

@amaanq have a look

### DIFF
--- a/queries/vue/indents.scm
+++ b/queries/vue/indents.scm
@@ -4,4 +4,4 @@
   (template_element)
 ] @indent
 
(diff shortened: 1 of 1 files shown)


amaanq:
That reads much nicer, thank you!

amaanq (approved):

PR #4468, opened 2023-03-09
highlights(prql): update to prql 0.6.0

matthias-Q:
Updates grammer for prql version 0.6.x

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -345,7 +345,7 @@
     "revision": "42d82fa18f8afe59b5fc0b16c207ee4f84cb185f"
   },
   "prql": {
-    "revision": "5f6c4e4a90633b19e2077c1d37248989789d64be"
+    "revision": "a5a1fdb45826f7f0d94e258ee5ada2e973708191"
   },
   "pug": {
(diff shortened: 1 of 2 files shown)


amaanq:
can switch/case fit under `@conditional` and loop under `@repeat`?

matthias-Q:
I would like to keep them under `@keyword`. Broadly speaking, every line, or transform, in PRQL starts with one of the major keywords (`from`, `derive`, `aggregate`...) and I feel `loop` and `case` fall under the same class.

The logic behind `loop` is not yet implemented, since the documentation is not finished yet. But I opted to add the keyword already.

ObserverOfTime:
> can switch/case fit under `@conditional` and loop under `@repeat`?

I also agree with this. SQL already uses `@conditional`.

matthias-Q:
Yes, the latest rev is a5a1fdb 
I have updated the lockfile in this PR as well.

amaanq:
I'll merge for now so CI isn't breaking anymore but please do update it appropriately, thank you :slightly_smiling_face:

amaanq (approved):

PR #4472, opened 2023-03-09
Partially Revert "fix(ecma): properly capture constants in const declarations"

### LINKED ISSUE #4466: recent ECMA / Javascript constant highlighting is inconsistent

The following PR ( https://github.com/nvim-treesitter/nvim-treesitter/commit/c553e6c5608ba54e210e9131bf9ba5e36ef18e57 ) introduces changes that assign variables declared as const to the @constant highlight group.  I would personally like to see this reverted, but I'm curious how other JS devs feel about it. 

The largest issue I have with this is the inconsistency it introduces.  With this patch const variables are highlighted a different color than other variables in their declaration, but when const variables are read they remain the standard variable highlighting. Besides the inconsistency, it doesn't seem helpful for the intended purpose of reminding the developer. 

It also adds ambiguity with capitalized constants, which have historically had the @constant group, but are generally used differently. Neither C or C++ do this for const declarations, so it makes me wonder why we feel it would be helpful for JS?  

I realize enabling semantic highlighting somewhat resolves this, and I'm able to override this rule in my own queries. It just doesn't seem like a good default. Is it just me? 

```javascript
const data = await res.json();
const users = data.Users;
```

```
lexical_declaration [151, 6] - [151, 40]
  variable_declarator [151, 12] - [151, 39]
    name: identifier [151, 12] - [151, 17]
```

### Example screenshot

image

### Expected behavior

I realize this is subjective, but it would be ideal if const variable declarations were highlighted as @variable and not @constant.  @constant should remain enabled for capitalized constants, etc.  

Alternatively, allow semantics to handle highlighting readonly variables, etc. 

```markdown

Installation ~
- OK `node` found v16.16.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.

OS Info:
{
  machine = "arm64",
  release = "22.3.0",
  sysname = "Darwin",
  version = "Darwin Kernel Version 22.3.0: Thu Jan  5 20:48:54 PST 2023; root:xnu-8792.81.2~2/RELEASE_ARM64_T6000"
} ~

  - c_sharp             ✓ ✓ ✓ . ✓
  - cmake               ✓ . ✓ . .
  - diff                ✓ . . . .
  - embedded_template   ✓ . . . ✓
  - fennel              ✓ ✓ ✓ . ✓
  - fish                ✓ ✓ ✓ ✓ ✓
  - graphql             ✓ . . ✓ ✓
  - help                ✓ . . . ✓
  - html                ✓ ✓ ✓ ✓ ✓
  - http                ✓ . . . ✓
  - java                ✓ ✓ . ✓ ✓
  - jsdoc               ✓ . . . .
  - json5               ✓ . . . ✓
  - jsonc               ✓ ✓ ✓ ✓ ✓
  - jsruby              ✓ . . . ✓
  - make                ✓ . ✓ . ✓
  - perl                ✓ . ✓ . ✓
  - php                 ✓ ✓ ✓ ✓ ✓
  - prisma              ✓ . . . .
  - regex               ✓ . . . .
  - ruby                ✓ ✓ ✓ ✓ ✓
  - scss                ✓ . ✓ ✓ .
  - sql                 ✓ . . . ✓
  - svelte              ✓ . ✓ ✓ ✓
  - tsx                 ✓ ✓ ✓ ✓ ✓
  - ymlruby             ✓ . . . ✓

```

```text
NVIM v0.9.0-dev-1147+ge389b1890
```

### Additional context

_No response_

amaanq:
This partially reverts commit c553e6c5608ba54e210e9131bf9ba5e36ef18e57.

Closes #4466

I was a bit naive in understanding how js consts work (I only use it for writing tree-sitter grammars..oops)

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -29,11 +29,6 @@
 ((shorthand_property_identifier) @constant
  (#lua-match? @constant "^_*[A-Z][A-Z%d_]*$"))
 
-(lexical_declaration
-  "const"
-  . (variable_declarator
-      . name: (identifier) @constant))
-
 ((identifier) @variable.builtin
  (#vim-match? @variable.builtin "^(arguments|module|console|window|document)$"))
 


ObserverOfTime (approved):
I don't mind this either way.

amaanq:
> I don't mind this either way.

Neither do I, but it seems like JS devs do, and const in JS isn't the same as Go or Rust, it really just means immutable/readonly vs a compile time/global constant

PR #4475, opened 2023-03-10
feat: improve yuck highlighting

Philipp-M:
Improves yuck highlighting, see https://github.com/helix-editor/helix/pull/6242 how it looks like in helix

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -516,7 +516,7 @@
     "revision": "2c0e6be8dd4dcb961c345fa35c309ad4f5bd3502"
   },
   "yuck": {
-    "revision": "fefe8e32ac5306b261af6af52571e13edf7b420c"
+    "revision": "e3d91a3c65decdea467adebe4127b8366fa47919"
   },
   "zig": {
     "revision": "f3bc9ffe9ca10f52dee01999b5b6ce9a4074b0ac"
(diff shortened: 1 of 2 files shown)


amaanq:
Thanks! FYI: Helix will be moving to last wins like how we have it, as Max has also agreed it's better this way... :)

Relevant gist to our chats to help figure out where to coordinate our efforts if you're interested: https://gist.github.com/amaanq/ed0f3f80ea5307fd206b1d3b88b3220d

Philipp-M:
> Why not just do
> 
> ```scheme
> (string) @string
> ```
> 
> ?

It's more explicit, it only matches what really is a string and nothing else. But if you want to have a fallback to string (e.g. because of non highlighted interpolation operators if the scope/id is not configured otherwise), I can do that, though I prefer only applying highlights when it makes sense.

amaanq:
Hmm it's fine, I remember this made interpolations more defined which I liked, sorry for forgetting :sweat_smile: 

But generally, languages with interpolations *do* highlight the entirety of a string anyways

amaanq (approved):

Philipp-M:
> But generally, languages with interpolations do highlight the entirety of a string anyways

Interesting, well it is completely highlighted, with the defined ids/scopes at least, but I understand that a fallback might be useful (though, since all the inner parts of the interpolation are highlighted, it looks weird IMHO when e.g. `@punctuation.special` is not defined as color, I'd rather leave it white/not highlighted, so as is)

amaanq:
I'm w/ you on that, but not everyone might be :wink:

PR #4480, opened 2023-03-10
highlights(typescript): `infer`, template literal type & punctuations

zweimach:
| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/queries/typescript/highlights.scm
+++ b/queries/typescript/highlights.scm
@@ -12,6 +12,7 @@
   "override"
   "satisfies"
   "module"
+  "infer"
 ] @keyword
 
 (as_expression "as" @keyword)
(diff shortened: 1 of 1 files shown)


queries/typescript/highlights.scm
```diff
@@ -55,11 +58,21 @@
 (pair
   ":" @punctuation.delimiter)
 
+(index_signature
+  ":" @punctuation.delimiter)
+
+(opting_type_annotation
+  "?:" @punctuation.delimiter)
```
amaanq, line 65: is this not an operator?

zweimach: No, it is actually the same notation as `function (optionalParam?: string)`.
I tried to make the `?` sigil a `punctuation.special`, but the grammar interprets `?:` as one symbol.

amaanq (approved):

PR #4481, opened 2023-03-10
highlights(prql): change `loop` and `case` highlight

matthias-Q:
As discussed in #4468
`loop` will be `@repeat` and
`case` will be `@conditional`


theHamsta (approved):
Thank you!

Issue #3316, opened 2022-08-18
Generating parser from grammar doesn't work on Windows with shell other than `cmd`


binyomen:
### Describe the bug

The Win32 section of `make_directory_change_for_command` uses syntax that assumes the current shell is `cmd`. If the current shell isn't `cmd`, then it may behave differently. In my case with the shell set to `pwsh`, trying to install the gitignore parser prints out "Generating source files from grammar.js..." and then nothing else. The `tree-sitter` CLI no longer seems to be running, and looking in the cache directory the parser files have not been generated.

### To Reproduce

You need to be on Windows.

Set your shell to `pwsh` with:

```lua
    vim.opt.shell = 'pwsh'
    vim.opt.shellcmdflag = '-NoLogo -NoProfile -ExecutionPolicy RemoteSigned -Command [Console]::InputEncoding=[Console]::OutputEncoding=[System.Text.Encoding]::UTF8;'
    vim.opt.shellredir = '2>&1 | Out-File -Encoding UTF8 %s; exit $LastExitCode'
    vim.opt.shellpipe = '2>&1 | Out-File -Encoding UTF8 %s; exit $LastExitCode'
    vim.opt.shellquote = ''
    vim.opt.shellxquote = ''

    vim.opt.shelltemp = false
```

Also make sure `gitignore` is included in your `ensure_installed` list under `require('nvim-treesitter.configs').setup`.

When starting `nvim`, "Generating source files from grammar.js..." will be printed but the installation process won't continue past there.

### Expected behavior

The gitignore parser to be successfully installed.

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.6 (ccd6bf554d922596ce905730d98a77af368bba5c) (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v18.7.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `clang` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: clang version 14.0.6
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - html           ✓ ✓ ✓ ✓ ✓
  - json           ✓ ✓ ✓ ✓ .
[...]

Issue #1965, opened 2021-11-01
Shellslash should be considered during installation


zgpio:

theHamsta:
Can you please give use some details what the this would change or even create a PR for this? Is is about using vim on Windows with a command shell other that cmd.exe or powershell? How would setting this option affect the installation on cmd.exe?

PR #4487, opened 2023-03-11
fix(windows): fix invalid powershell separator during install

adam-wolski:
Fixes nvim installation on vim configuration that use powershell.

Fixes https://github.com/nvim-treesitter/nvim-treesitter/issues/3316 
Maybe fixes https://github.com/nvim-treesitter/nvim-treesitter/issues/1965

There's a similar pull request here https://github.com/nvim-treesitter/nvim-treesitter/pull/4365 but [has potential problems](https://github.com/nvim-treesitter/nvim-treesitter/pull/4365#issuecomment-1445416964)

### DIFF
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -301,7 +301,11 @@ end
 ---@return string command
 function M.make_directory_change_for_command(dir, command)
   if fn.has "win32" == 1 then
-    return string.format("pushd %s & %s & popd", cmdpath(dir), command)
+    if vim.o.shell == "powershell" or vim.o.shell == "pwsh" then
+      return string.format("pushd %s ; %s ; popd", cmdpath(dir), command)
+    else
+      return string.format("pushd %s & %s & popd", cmdpath(dir), command)
+    end
   else
     return string.format("cd %s;\n %s", dir, command)
   end
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -301,10 +301,10 @@ end
 ---@return string command
 function M.make_directory_change_for_command(dir, command)
   if fn.has "win32" == 1 then
-    if vim.o.shell == "powershell" or vim.o.shell == "pwsh" then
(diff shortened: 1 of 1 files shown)


theHamsta:
This is kind of a duplicate for https://github.com/nvim-treesitter/nvim-treesitter/pull/4365. Main problem is that shell == `powershell` or `shell==cmd` is not reliable. shell ==cmd is better as it is the default. But `powershell` can be `powershell` for Windows `powershell` or `pwsh` if user installed or a complete path to `pwsh` if it hasn't been added to path. Preferably, `cwd` would be implemented by just using always either `cmd` or `powershell` or some other trick to ensure the working directory of the current command.

The original form of https://github.com/nvim-treesitter/nvim-treesitter/pull/4365 was `if shell == "cmd"` which I think we could merge as it is at least more reliable than `if shell = "powershell"` (I personally use `pwsh`).

adam-wolski:
> This is kind of a duplicate for #4365. Main problem is that shell == `powershell` or `shell==cmd` is not reliable. shell ==cmd is better as it is the default. But `powershell` can be `powershell` for Windows `powershell` or `pwsh` if user installed or a complete path to `pwsh` if it hasn't been added to path. Preferably, `cwd` would be implemented by just using always either `cmd` or `powershell` or some other trick to ensure the working directory of the current command.
> 
> The original form of #4365 was `if shell == "cmd"` which I think we could merge as it is at least more reliable than `if shell = "powershell"` (I personally use `pwsh`).
> 
> EDIT: you have `pwsh` in your check. Could we change that to that the last path segment contains either of the two? So it would also work with `pwsh.exe` or `C:\foo\path\to\pwsh.exe`

We could check for `shell==cmd` and if not then we do powershell way.

theHamsta (approved):
Ok, not perfect. Maybe we can merge this as a quick fix. @amaanq  what do you think is `vim.o.shell = "cmd"` reliable enough or will there be people with shell == `"cmd.exe"`

amaanq:
I don't use windows, but maybe just checking if it starts with cmd would work?

adam-wolski:
I've changed it to finding `cmd` in the shell variable.

amaanq:
The plights of windows..thanks for your contribution!

PR #4488, opened 2023-03-11
highlights(cmake): match digits in unquoted arguments

matze:
Before
Bildschirmfoto vom 2023-03-11 16-02-26
After
Bildschirmfoto vom 2023-03-11 16-02-00

CC @uyha.

### DIFF
--- a/queries/cmake/highlights.scm
+++ b/queries/cmake/highlights.scm
@@ -1,7 +1,7 @@
 (normal_command
   (identifier)
   (argument (unquoted_argument)) @constant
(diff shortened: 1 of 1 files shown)


uyha:
yeah, i missed that

uyha:
while you're at it, can you also add the `@` symbol, it is used in the [configure_file](https://cmake.org/cmake/help/latest/command/configure_file.html) command

matze:
Yeah, good point, added it.

amaanq (approved):

PR #4490, opened 2023-03-11
chore: clean up parsers.lua

amaanq:
:broom:

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -53,6 +53,8 @@ end
 ---@field generate_requires_npm boolean|nil
 ---@field requires_generate_from_grammar boolean|nil
 ---@field location string|nil
+---@field use_makefile boolean|nil
+---@field cxx_standard string|nil
 
 ---@class ParserInfo
 ---@field install_info InstallInfo
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -36,12 +36,9 @@ for ft, lang in pairs {
   systemverilog = "verilog",
   cls = "latex",
   sty = "latex",
-  OpenFOAM = "foam",
   pandoc = "markdown",
   rmd = "markdown",
   quarto = "markdown",
-  cs = "c_sharp",
-  tape = "vhs",
   dosini = "ini",
   confini = "ini",
 } do
@@ -75,21 +72,16 @@ local list = setmetatable({}, {
 list.ada = {
   install_info = {
     url = "https://github.com/briot/tree-sitter-ada",
-    branch = "master",
     files = { "src/parser.c" },
   },
-  filetype = "ada",
   maintainers = { "@briot" },
 }
 
 list.agda = {
   install_info = {
     url = "https://github.com/AusCyberman/tree-sitter-agda",
-    branch = "master",
     files = { "src/parser.c", "src/scanner.cc" },
-    generate_requires_npm = true,
   },
-  filetype = "agda",
   maintainers = { "@Decodetalkers" },
 }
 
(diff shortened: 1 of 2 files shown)


clason (approved):

amaanq:
ocamllex is on abi 11, that's a first

lua/nvim-treesitter/parsers.lua
```diff
@@ -320,17 +307,15 @@ list.dhall = {
 list.diff = {
   install_info = {
     url = "https://github.com/the-mikedavis/tree-sitter-diff",
-    branch = "main",
```
theHamsta, line 323: deleting branch `main` might cause problems when no lockfile is used because we will then assume `master`. We're not using `git`, so we need to decide for one branch. We can anything commitish via curl but I don't think there is anything like default-branch when no lockfile is used. Will this work with `branch` deleted? https://github.com/nvim-treesitter/nvim-treesitter/blob/350e7e5e31a5265fbccdf96b00a413c32583c2cc/lua/nvim-treesitter/install.lua#L691-L700 I think yes, it looks like `git ls-remote` gives the default branch first (though this might be just the first branch created. So it would have been incorrect even before)

Since I think nobody used nvim-treesitter with no lockfile we can do this clean up. Can we then remove the option to run without lockfile?

theHamsta (approved):

lua/nvim-treesitter/parsers.lua
amaanq: > Can we then remove the option to run without lockfile?

I didn't know that was a thing...but yes as I wouldn't see why anyone wouldn't use the lockfile

amaanq:
Really not understanding the 2 failing cis

Mac fails randomly with a parser
Windows cl fails with t32 twice now

amaanq:
Enabled debug output and I see why now:

Mac compilation fails for norg due to me removing the cxx_standard and use_makefile fields, didn't know those were actually used so I'll add code docs for them as well

clason:
and you thought all these fields were just added for giggles ;)

lua/nvim-treesitter/parsers.lua
theHamsta: grep for `get_update_strategy`. In theory, you could always install from parser's master.

theHamsta:
Yes, mac uses a very old stdlib by default and I guess norg just want C++14

amaanq:
Now why is windows failing w/ t32..

lua/nvim-treesitter/parsers.lua
ObserverOfTime: > I didn't know that was a thing...but yes as I wouldn't see why anyone wouldn't use the lockfile

I don't use the lockfile when adding new languages. :)

lua/nvim-treesitter/parsers.lua
```diff
@@ -1180,8 +1114,8 @@ list.query = {
     url = "https://github.com/nvim-treesitter/tree-sitter-query",
     files = { "src/parser.c" },
   },
-  readme_name = "Tree-sitter query language",
   maintainers = { "@steelsojka" },
+  readme_name = "Tree-Sitter query language",
```
ObserverOfTime, line 1118: I'm late but it's actually capitalised as "Tree-sitter".

lua/nvim-treesitter/parsers.lua
amaanq: when adding locally or when adding them in a PR?

lua/nvim-treesitter/parsers.lua
clason: yeah, noticed that on the update README PR as well

lua/nvim-treesitter/parsers.lua
ObserverOfTime: Locally, before submitting the PR.

amaanq: Hmm me too, so what's the issue? or does branch need to be specified in that case as well if it's not master?

ObserverOfTime: I remember having to specify it.

PR #4491, opened 2023-03-11
refactor!: remove setting the update strategy

amaanq:
@theHamsta


theHamsta (approved):
I doesn't look like a lot of code. Do you think anyone is aware of this probably not :sweat_smile:

amaanq:
Should I remove the breaking change indication then?

clason:
always better to err on the side of caution

PR #4495, opened 2023-03-12
feat(thrift): update queries based on upstream changes

amaanq:
Went from 1400/1400 - 920/3 state count and as a result the binary went from 1.1mb to 125kb in size :)

### DIFF
--- a/queries/thrift/highlights.scm
+++ b/queries/thrift/highlights.scm
@@ -1,6 +1,7 @@
- ; Namespace
+; Variables
 
-(namespace_definition) @namespace
+((identifier) @variable
+  (#set! "priority" 95))
 
 ; Includes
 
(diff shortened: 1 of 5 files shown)


clason:
Thrifty!

amaanq:
Might be time I start improving some bloated-to-death commonly used parsers (Verilog, D, C#... :vomiting_face:)

clason:
vimdoc/viml could use some love as well (and are much closer to our heart) ;)

amaanq:
viml is a nightmare..still don't fully understand that "language" :laughing: 

vimdoc seems ok? It's 97kb but smaller would be better

I'll take a look at both though, would definitely benefit the most users

clason:
Yeah, size is not really the issue for these, but edge cases in a horribly underspecified language...

amaanq:
> Yeah, size is not really the issue for these, but edge cases in a horribly underspecified language...

Any examples I can have as reference? :)

clason:
The whole issue tracker ;)

(But this is getting off-topic here; I shouldn't even have brought it up.)

queries/thrift/highlights.scm
```diff
@@ -9,96 +10,118 @@
...
 
-(function_identifier) @function
+(function_definition
+  (parameters
+    (parameter (identifier) @parameter)))
 
-; Fields
+(throws
+  (parameters
+    (parameter (identifier) @parameter.exception)))
```
ObserverOfTime, line 30: Why not just `@parameter`?

amaanq: mmm, in this case I'd find it useful for anyone using this to easily distinguish exception parameters, here's an example

```thrift
  map<binary,list<ColumnOrSuperColumn>> multiget_slice(1:required list<binary> keys,
                                                       2:required ColumnParent column_parent,
                                                       3:required SlicePredicate predicate,
                                                       4:required ConsistencyLevel consistency_level=ConsistencyLevel.ONE)
                                        throws (1:InvalidRequestException ire, 2:UnavailableException ue, 3:TimedOutException te),
```

the parameters after throws are exceptions

ObserverOfTime: I think `InvalidRequestException` should be `@exception` and `ire` should be `@parameter` and so on.

amaanq: > I think InvalidRequestException should be @exception

It's a type though, no? I guess it'd be nice to distinguish from regular types though

amaanq: e.g. in Python

image

Exceptions are still types.. :shrug:

ObserverOfTime (approved):

PR #4496, opened 2023-03-12
Fix injection based on "call_expression" for "ecma"

Serranya:
Also add a comment what the query does.

Before:
before

After:
after


amaanq (approved):

amaanq:
Not sure why you closed this

Serranya:
@amaanq I didn't know that deleting my fork would close the PR.

PR #4505, opened 2023-03-14
python: fix indents

georgeharker:
Python will fail to indent after a for statement

eg

```
for i in range(5):<CR>
*
Cursor incorrectly here
```

will not indent.  Same for if statements.

```
if True:
*
Cursor incorrectly here
```

This is due to the line which validates if the node starts and ends on the same line - the implications of changing this for all languages are unclear.  However, for python an additional tag "indent_immediate" is set which allows these statements to be indented passing them through to the remainder of the indentation logic below.

### DIFF
--- a/.github/workflows/test-queries.yml
+++ b/.github/workflows/test-queries.yml
@@ -24,8 +24,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-2022, macos-latest]
-        cc: [ gcc, clang ]
-        nvim_tag: [ stable ]
+        cc: [gcc, clang]
+        nvim_tag: [stable]
         exclude:
           - os: ubuntu-latest
             cc: clang
@@ -59,10 +59,12 @@ jobs:
       - uses: ilammy/msvc-dev-cmd@v1
       - uses: actions/setup-node@v3
 
+      - name: Install tree-sitter CLI
+        run: npm i -g tree-sitter-cli
+
       - name: Install and prepare Neovim
         env:
           NVIM_TAG: ${{ matrix.nvim_tag }}
-          TREE_SITTER_CLI_TAG: v0.20.6
         run: |
           bash ./scripts/ci-install-${{ matrix.os }}.sh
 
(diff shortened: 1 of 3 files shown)


theHamsta:
Amazing! This seems to fix a lot of things. Could you add few tests for the improvements? It's just putting test files here: https://github.com/nvim-treesitter/nvim-treesitter/blob/350e7e5e31a5265fbccdf96b00a413c32583c2cc/tests/indent/python_spec.lua#L4 and specifying where to add a new line: https://github.com/nvim-treesitter/nvim-treesitter/blob/350e7e5e31a5265fbccdf96b00a413c32583c2cc/tests/indent/python_spec.lua#L19-L30

theHamsta:
This could potentially also apply to other indent-based languages like toml.

georgeharker:
How do I run the tests so I can iterate locally?  Did the various searches and couldn't find info.

georgeharker:
still not sure how to run locally, but added some tests, and fixed up a bare try: which is error status, but should indent for ease

amaanq:
check ci logs, 1 test failed

georgeharker:
Have some fixes for the way aligned_indent worked, so that it explicitly checks if the last line needs additional indent (controlled by a final_line_indent metadata setting

this handles

```
if (blah
        and needs_more_indent):
    body
````

and

```
while (blah
       and doesn't_need_more_indent):
    body
````

**Comment by georgeharker** [comment]
Will rebase later today

File: queries/python/indents.scm [hunk 1]
```diff
@@ -21,24 +16,45 @@
...
+  condition: (parenthesized_expression) @aligned_indent
+  (#set! "delimiter" "()")
+  (#set! "final_line_indent" 1) ; parenthesized_expression already indented
 )
+
 ((ERROR "(" . (_)) @aligned_indent
  (#set! "delimiter" "()"))
 ((argument_list) @aligned_indent
  (#set! "delimiter" "()"))
-((argument_list) @aligned_indent
```
theHamsta, line 38: Can you explain why this was a problem?

theHamsta:
Hmm, again this. Rebasing would close the PR...

georgeharker:
Yep. I'll do that later today for sure.

PR #4509, opened 2023-03-15
docs: add note in README on compilation errors with old compilers

bonderado:
Motivation:
Several parsers that need to be compiled using the c99 standard do not specify it, because it is the default in recent versions of the compiler (e.g.: awk, latex, vim), however, this is not true at my workplace, where an older compiler is installed to enforce environment stability, thus resulting in compilation errors.

My solution:
With this pull request, I would like to add support for cxx_standard in  select_compiler_args, so if the compiler defaults to a standard that is earlier than what is needed, in init.lua one can specify it like this:

parser_config['awk'].install_info.cxx_standard = 'c99'
parser_config['latex'].install_info.cxx_standard = 'c99'
parser_config['vim'].install_info.cxx_standard = 'c99'

and have the compilation succeded.

I find this a better alternative to bugging individual parsers.

I hope you find this useful.
Thank you for your work on nvim-treesitter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -694,3 +694,17 @@ require'nvim-treesitter.configs'.setup {
     --
 }
 ```
+
+#### I experience compilation errors when installing certain parsers
+
+Some parsers do not specify a standard for their compilation and rely on the compiler having a sufficient default.
+Check that your compiler is a recent version and update if necessary.
+If you cannot update the compiler, but the one you have supports the minimum standard for the failing parser, you can force compilation via Makefile by adding in your Lua config:
+
+```lua
+local parser_config = require "nvim-treesitter.parsers".get_parser_configs()
+-- For parsers written in c99 or earlier the following line is sufficient.
+parser_config['parser'].install_info.use_makefile = true
+-- For parsers with parts written in c++ you may need to set the following.
+ parser_config['parser'].install_info.cxx_standard = 'c++11'
+```


theHamsta:
This does not work because GCC doesn't allow to set standards  for mixed language compilations.

bonderado:
Just for clarification.

This is my first experience with lua and  I do not know if this is a corner case due to the environment that is mandated  at work, where I have:
gcc version 4.8.5 202150623 (Red Hat 4.8.5-44) (GCC)
that means that I cannot compile the norg parser that requires c++14, but other than that, if I do not have my modification and try to install everything, I get errors mentioning features only available in c99 for:
awk, comment, dart, foam, haskell, javascript, julia, kotlin,  lalrpop, latex, markdown, nix, qmljs, rnoweb, rst, scala, svelte, t32, tsx, typescript, vim, yuck.

I also get errors related to c++11 for:
astro, hack, nickel, pug, racket, tlaplus.
I verified that  in:
function M.select_compile_command(repo, cc, compile_location)
adding a " print(cc , repo.use_makefile, make)  "
outputs:   cc nil gmake, so the first branch of that conditional is followed, where no standard is explicitly set and the default one is not good.

With my change and setting the standard appropriately, all parsers (except norg) compile, but this could be something that only I need.

Thank you again for taking the time for this

theHamsta:
Here are the defaults of the makefile when no cxx_standard is set https://github.com/nvim-treesitter/nvim-treesitter/blob/master/scripts/compile_parsers.makefile

bonderado:
Thank you again, your suggestions do work.

Since the fix to the compilation errors can be all in the Lua configuration of user sthat need it (me and not many others, I guess),  I decided to remove my code and add a note in the troubleshooting section of the README.
If it is clear and valuable enough you can merge it, otherwise it will remain available via the closed pull request. 
I can make small adjustments if necessary.

All the best

theHamsta:
It would maybe be good to have a global option to always use Makefiles globally instead of needing to set it individually for every parser config.

README.md
```diff
@@ -694,3 +694,17 @@ require'nvim-treesitter.configs'.setup {
     --
 }
 ```
+
+#### I experience compilation errors when installing certain parsers
+
+Some parsers do not specify a standard for their compilation and rely on the compiler having a sufficient default.
```
theHamsta, line 700: It is not about the configs not specifying the config. Even if the parsers would state `c99`. We would be able to set it without the makefile.

theHamsta (approved):

README.md
```diff
@@ -694,3 +694,17 @@ require'nvim-treesitter.configs'.setup {
...
+Check that your compiler is a recent version and update if necessary.
+If you cannot update the compiler, but the one you have supports the minimum standard for the failing parser, you can force compilation via Makefile by adding in your Lua config:
+
+```lua
+local parser_config = require "nvim-treesitter.parsers".get_parser_configs()
+-- For parsers written in c99 or earlier the following line is sufficient.
+parser_config['parser'].install_info.use_makefile = true
+-- For parsers with parts written in c++ you may need to set the following.
+ parser_config['parser'].install_info.cxx_standard = 'c++11'
```
theHamsta, line 709: when you have this problem, wouldn't you want to use this for all parsers? This line is also from downgrading from c++14 which is the default for the makefile. 

`-- For parsers with parts written in c++ you may need to set the following if your compiler does not support c++14 yet (which is the default for makefile compilation)` 

also leading whitespace.

theHamsta:
Tree-sitter uses `-xc -std=c99`. I remember having some problems with those per-language options. Tree-sitter only sets it when there is a `scanner.c` (so no C++).

clason:
We are moving away from handling compilation details in favor of relying on `tree-sitter build` (and WASM parsers later on). So this is obsolete with 1.0.

PR #4511, opened 2023-03-16
fix(python): graceful indentation on hanging paren, hanging function def, list, set, dict

georgeharker:
These fixes are something I noticed about how the error states were handled and interactive typing.  

```
(
    a,
        b,
        *used to incorrectly indent to here
```

now produces
```
(
    a,
    b,
```

```
def f(a,
    b,
    *used to incorrectly indent to here
```

now produces

```
def f(a,
      b,
```


georgeharker:
with some small grammar tweaks this can be made to work similarly for {dict | set} [list] - but that would mean making them use align_indent, and I want to think about that a bit before I make that change.

In essence the real difference for those would be:

```
# now
[a,
    b

# vs
[a,
 b
```

amaanq:
I'm not against that fwiw

georgeharker:
The more I think about it the more consistent I think it would be. 
I'll want to write more tests for those cars so I'll have to wait until tomorrow.

lucario387:
Another request, can you also support it so `align_indent` only applies for nodes inside the tree that is before the ending delimiter?

axieax:
Not sure if this is entirely related to this issue, but I'm getting indentation issues with new lines after opening parentheses (e.g. lists, dictionaries).


https://user-images.githubusercontent.com/62098008/225645203-3a575ae2-858d-422c-aadc-24f30d91eec5.mp4

georgeharker:
> Another request, can you also support it so `align_indent` only applies for nodes inside the tree that is before the ending delimiter?

Like you did with @end_indent? Yes absolutely.

georgeharker:
Right now the issue I'm having with any generalization is that the delims get confused between the error cases - in other words the node's delim set for say

```
(ERROR "(" . (_)) @aligned_indent
 (#set! "delimiters" "()"))

```

Seem to be overwritten by

```
(ERROR "{" . (_)) @aligned_indent
 (#set! "delimiters" "{}"))

```

Despite the latter query not actually matching the current line.... this is confusing.

georgeharker:
> Please try to no create PR from your master branch.

Sorry, will figure out how to switch the branch for the pr and fix.

lucario387:
@georgeharker If you are closing this PR, may I copy all the changes to my PR #4516, and we move on from that?

georgeharker:
I just saw your PR, thanks for catching those. I was planning to iterate further on list, set and dict to address issues and have them respect aligned_indent. But I could happily do that in a new PR and I'd welcome starting from what you have, which seem like great fixes. So yes, please copy the fixes I have and I'll rebase any further fixes ontop

georgeharker:
> Also, how will it handle this case?
> 
> https://github.com/nvim-treesitter/nvim-treesitter/pull/4516#discussion_r1138843296

Bit quite sure. The ERROR cases are colliding and we get the wrong delims. I'm trying to figure out why. It's really weird.

georgeharker:
If there are two new rules
```
(ERROR "[" . (_)) @aligned_indent
 (#set! "delimiters" "[]"))
(ERROR "{" . (_)) @aligned_indent
 (#set! "delimiters" "{}"))
```

Then only the latter runs, sort of regardless of the string once an ERROR is in place, ie:

```
[1,2,3,<CR>
```

produces a tree like ```(ERROR "[" (integer) "," (integer) "," (integer) "," ```

*But* the query that matches it will be the one with a "{" - it's like ERROR eats everything after?

This makes it really hard to make aligned indent work and I do not understand sufficiently why the latter query is matched for the above text (It is though, debug prints say so).

lucario387:
> If there are two new rules
> 
> ```
> 
> (ERROR "[" . (_)) @aligned_indent
> 
>  (#set! "delimiters" "[]"))
> 
> (ERROR "{" . (_)) @aligned_indent
> 
>  (#set! "delimiters" "{}"))
> 
> ```
> 
> 
> 
> Then only the latter runs, sort of regardless of the string once an ERROR is in place, ie:
> 
> 
> 
> ```
> 
> [1,2,3,<CR>
> 
> ```
> 
> 
> 
> produces a tree like ```(ERROR "[" (integer) "," (integer) "," (integer) "," ```
> 
> 
> 
> *But* the query that matches it will be the one with a "{" - it's like ERROR eats everything after?
> 
> 
> 
> This makes it really hard to make aligned indent work and I do not understand sufficiently why the latter query is matched for the above text (It is though, debug prints say so).

Hmm, i'll try to investigate the tree more later today

georgeharker:
For example dumping the nodes with the following:

```

(ERROR "[" . (_)) @aligned_indent

 (#set! "Thisnode" "blah")

 (#set! "delimiters" "[]"))

(ERROR "{" . (_)) @aligned_indent

 (#set! "Thatnode" "oops")

 (#set! "delimiters" "{}"))

```

Shows that only the latter runs (thisnode isn't set, meaning it's not that both run and one overwrites). 

My bet is that ERROR is somehow terminal deep inside tree-sitter, it's an odd one for sure.

georgeharker:
Looks like we actually get all back as captures inside the query runner iterations - I'm guessing something about the way they get evaluated makes it look like only the last ran. 

But basically ```ERROR -anything here- ``` will alias. Not sure where that leaves things for matching incomplete portions of the grammar.

lucario387:
> Looks like we actually get all back as captures inside the query runner iterations - I'm guessing something about the way they get evaluated makes it look like only the last ran.
> 
> But basically `ERROR -anything here- ` will alias. Not sure where that leaves things for matching incomplete portions of the grammar.

https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/c/indents.scm#L12-L14
Does this also get matched?

georgeharker:
Yes!!! It does and is disambiguated between other ERROR cases. 

Here's what's going on 

```
((ERROR "X" . (_)) @aligned_indent 
(#set! "delimiters" "XY"))
```

Will alias for all X. 

```
(ERROR "X" @aligned_indent (#set! "delimiters" "XY") . (_))
```

Will not. But of course sets the indent on a child node. So if we (only in the error case) fish the first aligned_indent out of the children nodes if the parent doesn't have it set, we have a solution.

amaanq:
This is really interesting...I don't normally account for such cases in the indents.scm I write for my grammars (most aren't indent based anyways..), but this is good to consider

georgeharker:
closing in favor of #4516

PR #4513, opened 2023-03-16
queries(lua): add more keywords related to variable/function builtins, debug, and coroutines

amaanq:
Up for debate, especially the metamethods. I'm thinking a new group like `@function.special` for those in Lua and Python

### DIFF
--- a/queries/lua/highlights.scm
+++ b/queries/lua/highlights.scm
@@ -129,7 +129,10 @@
 (identifier) @variable
 
 ((identifier) @variable.builtin
- (#any-of? @variable.builtin "self" "string" "table" "vim"))
+ (#any-of? @variable.builtin "_G" "_VERSION" "debug" "io" "jit" "math" "os" "package" "self" "string" "table" "utf8"))
+
+((identifier) @keyword.coroutine
+  (#eq? @keyword.coroutine "coroutine"))
 
 (variable_list
    attribute: (attribute
@@ -180,8 +183,11 @@
     ;; built-in functions in Lua 5.1
     "assert" "collectgarbage" "dofile" "error" "getfenv" "getmetatable" "ipairs"
     "load" "loadfile" "loadstring" "module" "next" "pairs" "pcall" "print"
-    "rawequal" "rawget" "rawset" "require" "select" "setfenv" "setmetatable"
-    "tonumber" "tostring" "type" "unpack" "xpcall"))
+    "rawequal" "rawget" "rawlen" "rawset" "require" "select" "setfenv" "setmetatable"
+    "tonumber" "tostring" "type" "unpack" "xpcall"
+    "__add" "__band" "__bnot" "__bor" "__bxor" "__call" "__concat" "__div" "__eq" "__gc"
+    "__idiv" "__index" "__le" "__len" "__lt" "__metatable" "__mod" "__mul" "__name" "__newindex"
+    "__pairs" "__pow" "__shl" "__shr" "__sub" "__tostring" "__unm"))
 
 ;; Others
 


queries/lua/highlights.scm
```diff
@@ -129,7 +129,13 @@
 (identifier) @variable
 
 ((identifier) @variable.builtin
- (#any-of? @variable.builtin "self" "string" "table" "vim"))
+ (#any-of? @variable.builtin "_G" "_VERSION" "io" "math" "os" "package" "self" "string" "table" "utf8" "vim"))
```
clason, line 132: We deliberately keep this parser and query generic (not specifically targeting for Neovim dev, unlike TJ's parser). So this should only be variables actually built in to the language.

amaanq: These are generic, see the index here - https://www.lua.org/manual/5.3/

The only exception is "vim", which is fine imo :)

clason: Yes, I was specifically talking about `vim`, of course. The rest is fine (and `vim` is not, in my view).

clason: (remember that we are talking about making it easier to share queries -- and this would be a move in the wrong direction).

Also, it's very easy to add that to an `; extends` query locally.

amaanq: Makes sense...that conversation has been a bit stale lately though :confused: seems like we need Helix to implement logic to use last wins, then we can standardize query captures/predicates...well that's getting off topic. I'll remove vim now

queries/lua/highlights.scm
```diff
@@ -129,7 +129,13 @@
 (identifier) @variable
 
 ((identifier) @variable.builtin
- (#any-of? @variable.builtin "self" "string" "table" "vim"))
+ (#any-of? @variable.builtin "_G" "_VERSION" "io" "math" "os" "package" "self" "string" "table" "utf8"))
+
+((identifier) @debug
+  (#eq? @debug "debug"))
```
ObserverOfTime, line 135: Are we even using `@debug` anywhere?
It might be worth delegating this to a separate PR for more languages.

amaanq: godot uses it, that's about it :laughing:

I don't think checking for debug identifiers to set as a debug capture works for every language, and a lot don't support "debugging" built in to the language

theHamsta: It is not clear what `debug` is there for. It is just a relic because there is also a Vim highlight group with that name (where you would also not know what it is for)

amaanq: Maybe we should remove it then..?

theHamsta: I would really like this getting removed. It's one of the documented built-ins of Neovim though. It was added in as a group of 6 captures that were added at the same time for Neovim and nvim-treesitter with only getting slowly adopeted by nvim-treesitter. Usually, you would only add a new captures they get introduced for a few languages were we see the need for it.

amaanq: Ok, I'll remove it from here and submit another PR removing it as it's pretty useless and hard to understand its purpose.

amaanq:
Can `jit` be added too? :thinking:

queries/lua/highlights.scm
theHamsta: If it can be consistently added to multiple languages it's ok. It would just be good to not of captures that are only used by one or two languages and aren't 100% where they would apply to new languages. Having a capture already in some languages helps to get some context how to use something and to evaluate whether it is actually needed.

amaanq: I don't really see this being able to be added to most languages, just a few exceptions like godot and maybe debug here..

amaanq: Even in vim in the runtime/syntax dir:
```sh
rg -i "hi" | rg "Debug" | tr \: \\n | rg ".vim" | uniq | sort
```
gives a handful of languages (vim has support for 683)
```
8th.vim
aptconf.vim
cl.vim
cmod.vim
d.vim
eiffel.vim
elm.vim
euphoria3.vim
euphoria4.vim
forth.vim
freebasic.vim
fstab.vim
groovy.vim
haskell.vim
help.vim
hollywood.vim
initex.vim
java.vim
javascript.vim
krl.vim
maple.vim
mel.vim
nasm.vim
obse.vim
pike.vim
plaintex.vim
pov.vim
progress.vim
rego.vim
remind.vim
reva.vim
sass.vim
scilab.vim
```

clason: Vim's syntax files are a _mess_, with highlight groups added without rhyme or reason (apart from whatever made sense to the maintainer at the time). Let's not propagate old mistakes without thinking.

amaanq: So...we should remove debug then

ObserverOfTime (approved):

PR #4514, opened 2023-03-16
feat(python): add "return" @indent_end

lucario387:
`return ...` is usually the last line of a
function/statement, so mark some possible patterns of `return` with
`@indent_end` to dedent the line after it

@amaanq should this be added?

### DIFF
--- a/queries/python/indents.scm
+++ b/queries/python/indents.scm
@@ -72,6 +72,15 @@
 (list_pattern "]" @indent_end)
 
 
+(return_statement
+  [
+    (_) @indent_end 
+    (_ (_) @indent_end .)
+    (attribute 
+      attribute: (_) @indent_end)
+    "return" @indent_end
+  ] .)
+
 [
   ")"
   "]"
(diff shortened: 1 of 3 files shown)


amaanq:
I guess even in an early return it should dedent the statement (if/else/whatever) that it's in..this seems ok to me. Yield doesn't apply right? I haven't touched python in a while but iirc you can put statements after a yield statement without "having" to dedent (unlike returns where it'd be unreachable code)

amaanq (approved):

lucario387:
> I guess even in an early return it should dedent the statement (if/else/whatever) that it's in..this seems ok to me. Yield doesn't apply right? I haven't touched python in a while but iirc you can put statements after a yield statement without "having" to dedent (unlike returns where it'd be unreachable code)

I don't think yield applies, there is nothing stopping one from adding multiple yields at once, but not the same for return as the following ones are unreachable

theHamsta:
We should probably also check if `match` is handled correctly.

lucario387:
@theHamsta some 5 minute research with adding the queries in

```python
match parameter:
    case "1":
        return True # Entering here will lead to the cursor at indent 4
    case True:
        return (1, 2, 3) # Also indent 4
    case False:
        return # Here it is indent 0
```
If you think it's fine, I will open another PR for this

PR #4515, opened 2023-03-16
feat: add LLVM TableGen

amaanq:
https://llvm.org/docs/TableGen/ProgRef.html

Any LLVM fans out there...? :)

### DIFF
--- a/queries/tablegen/highlights.scm
+++ b/queries/tablegen/highlights.scm
@@ -1,0 +1,156 @@
+; Preprocs
+
+(preprocessor_directive) @preproc
+
+; Includes
+
+"include" @include
+
+; Keywords
+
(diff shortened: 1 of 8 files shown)


queries/tablegen/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+
+[
+  "foreach"
+] @repeat
+
+; Variables
+
+(identifier) @variable
+
+(var) @tag ; need something more suitable, but nothing fits as "correctly" as @tag, maybe @variable.builtin
```
ObserverOfTime, line 45: `@variable.builtin` sounds fine.

ObserverOfTime (approved):

PR #4516, opened 2023-03-16
fix(indent): fix align affecting external node

lucario387:
Problem: 
- `@aligned_indent` was affecting lines outside of its range (if the ending delimiter exists) even though it shouldn't be doing that
- My new queries leads to regressions (see https://github.com/nvim-treesitter/nvim-treesitter/issues/1136#issuecomment-1471684547)

Fixes:
- Add in checks in `indent.lua`
- Revert queries that lead to regressions
- Add extra return patterns for proper dedent to account for the missing `@indent_end` marks 

Todo:
- [x] Add aligned_indent logic
- [x] Revert queries
- [x] Add return patterns
- [x] Add regression tests

Please help tests as I don't think i have found all the edge cases

cc @amaanq , @theHamsta , @lkhphuc, @georgeharker


lucario387:
One problem with this query of mine (that is also same with match statement if it were to get added into nvim-treesitter)

The last return always dedent to indent 0, idk why

```python
def a():
    if a:
        return # Pressing Enter here will have a new line with indent 4
    if b:
        return 1 # But here it's indent 0
```

Is this ok?

tests/indent/python/return_dedent.py
```diff
@@ -14,3 +14,26 @@ def a():
     return (
         1, 2, 3
     )
+
+def a():
+    return b(
+        1, 2, 3
```
theHamsta, line 20: how would 
```
b(1,
2,
3)
````
be indented?

Comment (lucario387) [reply tests/indent/python/return_dedent.py:20]: Interesting, pressing enter on that line leads to
```python
b(1,
2,
3)
  |
```

lucario387: I think @georgeharker said that he's planning to change the behavior to support this kind of style as well, so I am not sure whether I should try to implement it here

theHamsta: No, I meant what would happen when you indent the whole expression or when you start a new line after
````
b(1,
```
?

lucario387:
Cherry-picked #4511 from @georgeharker.

georgeharker:
The failures are caused by a remaining return statement followed by an if - roughly indent.lua:244

lucario387:
image
test failed because of gleam

georgeharker:
This looks great!

amaanq:
I'm very appreciative of both your efforts, Python indents have been not-so-great for so long :) now all that's left is maybe the bracket bug..

The changes look good to me but I'm not good with Lua code in general :slightly_smiling_face:

amaanq (approved):

lucario387:
idk, there are so many bracket bugs that I am kinda in fear to fix it

And if the bracket bug is supposed to mean `I forgot to add @indent_end to end of a block`, then that's probably something to be added in the future

georgeharker:
I have a fix;) let's get this one in and I'll follow separately with that. Your question about the c for loop was helpful.

lucario387:
> I have a fix;) let's get this one in and I'll follow separately with that. Your question about the c for loop was helpful.

That was more so my "why is it working" query just to maintain old for loop behavior while I was refactoring C indents
Alas no one seem to notice a difference

georgeharker:
I finally got moving list, dict, set to aligned indent to work - there's a lot of gotchas, which I'll go back and update on that PR.

#4524

is that PR - I'm happy to do whichever makes sense, this PR is key to that other working (it includes it, for that reason, but the intent isn't to assume that should supersede this).

lucario387:
> I finally got moving list, dict, set to aligned indent to work - there's a lot of gotchas, which I'll go back and update on that PR.
> 
> 
> 
> #4524
> 
> 
> 
> is that PR - I'm happy to do whichever makes sense, this PR is key to that other working (it includes it, for that reason, but the intent isn't to assume that should supersede this).

There will be lots of merge conflicts if this and your PR were to be merged as one.

There is still some differences between this and your #4524, so can you manually copy and paste the diffs? It's in the dedent return queries and the tests

lucario387:
If this is merged, then #4524 will need to rebase, but that's probably not that much of an issue :')

I only change the behavior for the line right after `@aligned_indent`, so the scope of change is much smaller

lucario387:
Superseded by #4524

PR #4518, opened 2023-03-16
Use pre-generated parser files for sql grammar

DerekStride:
## What

As of https://github.com/derekstride/tree-sitter-sql/pull/100 we're publishing the `parser.c` source code artifact on the `gh-pages` branch, which should make it easier to install.

I tried running `scripts/update-lockfile.sh sql` to bump the lockfile to use the revision on the new branch but that didn't seem to work. I'd commit the new revision manually but I'm concerned it's an issue with the script. I'll investigate.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1277,7 +1277,7 @@ list.sql = {
   install_info = {
     url = "https://github.com/derekstride/tree-sitter-sql",
     files = { "src/parser.c" },
-    requires_generate_from_grammar = true,
+    branch = "gh-pages",
   },
   maintainers = { "@derekstride" },
 }
--- a/lockfile.json
+++ b/lockfile.json
@@ -429,7 +429,7 @@
     "revision": "05f949d3c1c15e3261473a244d3ce87777374dec"
   },
   "sql": {
-    "revision": "d6963a2426972a6ffa08ff92a5dba7922d68c1f5"
+    "revision": "4cb5b36d70687bfe4687c68483b4dacde309ae6f"
   },
   "squirrel": {
     "revision": "3fefc6b9bb2b4de1b1c461783f675918cd957546"


clason:
Not an issue, you just need to edit parsers.lua to remove the generate_from_grammar key.

clason:
But note that we always compile from source; there's no binary installation feature yet (as we support many more platforms than GHA does, so the fallback compilation is needed in any case).

lua/nvim-treesitter/parsers.lua
```diff
@@ -1277,7 +1277,7 @@ list.sql = {
   install_info = {
     url = "https://github.com/derekstride/tree-sitter-sql",
     files = { "src/parser.c" },
-    requires_generate_from_grammar = true,
+    branch = "gh-pages",
```
DerekStride, line 1280: > Not an issue, you just need to edit parsers.lua to remove the generate_from_grammar key.

Is this all that's needed then?

DerekStride:
> But note that we always compile from source; there's no binary installation feature yet

We're not publishing binaries (yet?) anyways.

clason:
Ah, sorry, then I misinterpreted the description.

lua/nvim-treesitter/parsers.lua
clason: Should be, if the lockfile hash points to the head of that branch

clason: Which you need to do manually in this case.

DerekStride:
> Ah, sorry, then I misinterpreted the description.

I updated it to make it more clear

lua/nvim-treesitter/parsers.lua
DerekStride: I updated the lockfile to point to the revision from the `gh-pages` branch.

clason:
Flakey test gonna flake.

theHamsta (approved):

PR #4520, opened 2023-03-17
fix(ecma): template literal injections

bennypowers:
The change in https://github.com/nvim-treesitter/nvim-treesitter/commit/e3ebc8ec5d586162f3c408417621daa59ba8ea62 broke lit-html highlighting. 

Here is the new, broken injection:

Screenshot from 2023-03-17 09-35-28

Here it is with the parens, which causes a runtime error
Screenshot from 2023-03-17 09-35-34
I'd now need to wrap the template literal in parens, which is not the same kind of function call, as demonstrated in this screenshot:
Screenshot from 2023-03-17 09-53-54

This also assigns the HTML parser for SVG tagged literals, since svg uses the HTML parser
```js
svg`<svg>...</svg>`
```

### DIFF
--- a/queries/ecma/injections.scm
+++ b/queries/ecma/injections.scm
@@ -11,6 +11,19 @@
      (#offset! @content 0 1 0 -1))
 )
 
+; html`...`, sql`...` etc
+(call_expression 
+ function: ((identifier) @language)
+ arguments: ((template_string) @content
+     (#offset! @content 0 1 0 -1))
+)
+
+; svg`...`, which uses the html parser
+(call_expression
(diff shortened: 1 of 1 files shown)


bennypowers:
Yes, I originally used that alternation, but committed the separate queries because I don't understand the perf implications. Either is fine for me. I will commit them tomorrow when I get back home, or you are welcome to update the PR.

Thank you for the review

bennypowers:
image

amaanq:
Awesome thanks!

amaanq (approved):

PR #4523, opened 2023-03-17
fix(v): add `vlang` as filetype

tobealive:
This PR extends highlighting for V files to file types set to `vlang`.

Currently, highlighting requires `ft=v`. However, until the recent merge to also include `ft=v` in nvim-lspconfig, `vls` required `ft=vlang`, which will disable tree-sitter highlighting for V. 

Since the recommendations to set the file type for V are just becoming more consistent and both are valid, this PR makes it work interchangeably.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1480,6 +1480,7 @@ list.v = {
     files = { "src/parser.c", "src/scanner.c" },
     location = "tree_sitter_v",
   },
+  filetype = "vlang",
   maintainers = { "@kkharji" },
 }
 


theHamsta:
This repo does not handle filetypes. Please make a PR to github.com/vim/vim directly !

theHamsta:
Does Vim support `vlang` filetype detection? It it always `vlang` or should support both `v` and `vlang`?

clason:
Note that we also have a `verilog` parser, so there shouldn't be confusion between these two languages.

ObserverOfTime (approved):
[v-vim](https://github.com/ollykel/v-vim) also sets it to `vlang`.

amaanq:
I prefer to have fts not end in "lang"...but the ambiguity with verilog and v should be solved in vim/neovim with an appropriate pr to peek at the file contents and determine through there (some other languages have this)

theHamsta:
@tobealive Thank you for this contribution! And please excuse the delay in review.

Issue #2293, opened 2022-01-18
Notice of Breaking Changes


clason:
This issue is used to announce breaking changes (and possibly other critical info). Users are strongly encouraged to subscribe to this issue to get notified of such changes.

Every breaking change is added as a new comment.

clason:
https://github.com/nvim-treesitter/nvim-treesitter/pull/2272 switched our Lua parser to one with better coverage. Old queries are not compatible with the new parser.

To make sure you update to the new one, do `:TSUninstall lua` followed by `:TSInstall lua`. Third-party modules providing Lua queries need to adapt to the new parser.

ahmedelgabri:
> #2272 switched our Lua parser to one with better coverage. Old queries are not compatible with the new parser.
> 
> To make sure you update to the new one, do `:TSUninstall lua` followed by `:TSInstall lua`. Third-party modules providing Lua queries need to adapt to the new parser.

After following this I'm getting this error

```
Error detected while processing FileType Autocommands for "*":
E5108: Error executing lua ...ed-0.6.1/share/nvim/runtime/lua/vim/treesitter/query.lua:161: query: invalid node type at position 1
stack traceback:
        [C]: in function '_ts_parse_query'
        ...ed-0.6.1/share/nvim/runtime/lua/vim/treesitter/query.lua:161: in function 'get_query'
        ...cker/start/nvim-treesitter/lua/nvim-treesitter/query.lua:91: in function 'get_query'
        ...r-textobjects/lua/nvim-treesitter/textobjects/select.lua:57: in function 'attach'
        ...er/start/nvim-treesitter/lua/nvim-treesitter/configs.lua:458: in function 'attach_module'
        ...er/start/nvim-treesitter/lua/nvim-treesitter/configs.lua:481: in function 'reattach_module'
        [string ":lua"]:1: in main chunk
Press ENTER or type command to continue
```

treatybreaker:
Getting the same output here as well after doing a full reinstall.

```
Error detected while processing function 3[3]..4[71]..matchup#delim#get_current[1]..<SNR>86_get_delim_multi[3]..matchup#ts_engine#get_delim[3]..<SNR>74_forward:
line    1:
E5108: Error executing lua ...im/0.6.1/share/nvim/runtime/lua/vim/treesitter/query.lua:161: query: invalid node type at position 1
stack traceback:
        [C]: in function '_ts_parse_query'
[...]

PR #4524, opened 2023-03-17
refactor(indent)!: rework aligned indentation, refactor indent captures to be aligned with upstream styling, fix python indenting

georgeharker:
Further to #4511 (closed) and #4516 which is not yet in,

This includes #4516 - and the intent is not to usurp that work, because it is super important for this to function.  So can clean up the history on this in whichever way makes sense, or base ontop of that once it is in.

List, set, dict and tuple all allow aligned_indent

eg

```
l = [1,2,3,4,<CR>
     | cursor here
d = {1:2,3:4,<CR>
     | cursor here
s = {1,2,3,4,<CR>
     | cursor here
```

List, set, dict and tuple all use aligned_indent which permits both hanging and aligned styles.

The final closing delim can be optionally specified to dedent (python says both are fine, allow queries to specify which to use).

Multiple opening delims on one line and multiple closing on a line are collapsed so as not to over indent

The final line of aligned_indent blocks which must in some cases be treated specially to avoid clashing with the next line is treated the same regardless of whether the aligned_indent node actually uses aligned indentation or just normal indentation.

Finally, try: on it’s own will indent when typing live but make no guaranteeds about whole-file formatting.

Part of the implementation for this is noted in the history of #4511 - which is that 

```
((ERROR "X" . (_)) @aligned_indent 
(#set! "delimiters" "XY"))
```

Will alias for all X. 

```
(ERROR "X" @aligned_indent (#set! "delimiters" "XY") . (_))
```

Will not. But of course sets the indent on a child node. Which means we need to have a way to fish things out of the children for error states.

Finally both

```
l = [
    1
    ]
````
and
```
l = [
    1
]
````
are legal.
the ```dedent_bare_closing_delim``` setting controls which to select.  Selected it on for the latter style as it matches existing python test specs - but both are legal.

### CODE REVIEW

**Comment by lucario387** [comment]
There is still some diff between this and my PR, but it's okay ig

**Comment by georgeharker** [comment]
I have a failure in Yang also. 

Basically I noticed that aligned indent needed to (optionally) handle the last line specially even if the block it's working with is not align indented. 

Lmk how best to handle merges / what your thoughts are.

**Comment by lucario387** [comment]
> I have a failure in Yang also. 
> 
> Basically I noticed that aligned indent needed to (optionally) handle the last line specially even if the block it's working with is not align indented. 
> 
> Lmk how best to handle merges / what your thoughts are.

If this still works fine, then I think either:
- This get merged -> you rebasing the new changes to your PR branch
- This wont get merged -> you should copy paste some of the diffs between this and your pr branch

**Comment by lucario387** [comment]
ok ...

**Comment by georgeharker** [comment]
> > I have a failure in Yang also.
> > Basically I noticed that aligned indent needed to (optionally) handle the last line specially even if the block it's working with is not align indented.
> > Lmk how best to handle merges / what your thoughts are.
> 
> If this still works fine, then I think either:
> 
> * This get merged -> you rebasing the new changes to your PR branch
> * This wont get merged -> you should copy paste some of the diffs between this and your pr branch

So I had missed a 

```
                indent = math.max(indent - indent_size, 0)
```

when we get outside the range of the indent node.

I had also neglected to forcibly return rather than allow other nodes to further increment when the position was set to be absolute - but that's the only time we should forcibly return the indent.

I / think/ I now have all your fixes also.

**Comment by lucario387** [comment]
@amaanq @theHamsta does this suffice?

**Comment by lucario387** [comment]
This looks a bit too messy of a commit history and will probably require a squash :)

**Comment by amaanq** [comment]
I usually squash prs with 3+ commits anyways :)

File: lua/nvim-treesitter/indent.lua [hunk 1]
```diff
@@ -243,14 +242,17 @@ function M.get_indent(lnum)
           end
         else
           -- aligned indent
-          if (c_delim_node and c_is_last_in_line and
-              c_srow and o_srow ~= c_srow and c_srow < lnum - 1) then
+          if c_delim_node and c_is_last_in_line and c_srow and o_srow ~= c_srow and c_srow < lnum - 1 then
```
lucario387, line 245: pretty sure c_delim_node here isn't necessary cause if c_srow can only exist if c_delim_node is not nil, so checking for `c_srow` is enough

lewis6991:
This is a pretty complicated change that seems is only specific to python and adds captures that aren't documented.

Not sure about this myself.

georgeharker:
I agree it's fairly complex. I'm happy to document the captures. I do think that it applies to aligned indent anywhere that it is used. Though a lot is to make python work better (which it doesn't right now).

lewis6991:
The long term goal is that all editors use the same captures in a relatively standardized way. Adding new captures for fairly specific cases is a tough sell.

If you could document this better and demonstrate its use for 1 or 2 more languages, then that improves the proposition.

lucario387:
> I agree it's fairly complex. I'm happy to document the captures. I do think that it applies to aligned indent anywhere that it is used. Though a lot is to make python work better (which it doesn't right now).

If it's to be documented
- `immediate_indent` seems to be similar in functionality to Helix's `@extend`, so this can become a capture instead of setting metadata like this
- `final_line_indent`, `immediate_indent`, `dedent_bare...` can become captures instead of setting using directives like this (I think it's interchangeable if the usage is like how you are doing currently)
  - If they were to be captures instead, maybe we can write the queries differently to make it look "cleaner" ? not sure, bunch of `#set!` for some of them doesn't look as clean as ` [ (a) (b) (c) ] @capture` to me
- Can this be extended for usage in yaml indent? :)

lewis6991:
Actually in general, upstream tend to use directives instead of captures. It's more verbose but for the best I think since it makes it easier to use queries together.

E.g. the new injection format will switch to directives.

lucario387:
Hmm, understood

If that's the case then my thoughts for the added directives from this PR is:
- `immediate_indent` directive is (very) useful and is similar to helix `@extend`, and they also use it for python, gdscript and ponylang, so this should be kept
- `final_line_indent` and `dedent_bare...`: not sure, maybe changing the behavior of `immediate_indent` to account for some cases that these directives are used for is easier

lewis6991:
If you prefer to use captures then something like `@indent.var` should be used.

georgeharker:
Aligned indent has a couple of options 

Ie 

```
L = [
    3
    ]

L = [
    3
]

And it's hanging variants 

L = [3
    ]

```

Which you get for the non aligned version by default is controlled by a setting now. But could be a different capture. 

If theres consensus on what the right presentation is I'm happy to implement that. 

Other than that the general aligned indent improvements seem important at least for python. I'll look for other examples where It would be useful.

queries/python/indents.scm
```diff
@@ -20,6 +16,19 @@
   (concatenated_string)
 ] @indent
 
+((list) @aligned_indent
+ (#set! "delimiter" "[]")
+ (#set! "dedent_bare_closing_delim" 1)
```
lewis6991, line 21: You should be able to exclude `1` .

queries/python/indents.scm
```diff
@@ -20,6 +16,19 @@
...
+)
+((dictionary) @aligned_indent
+ (#set! "delimiter" "{}")
```
lewis6991, line 24: A note for future. Each delimiter should be provided separately (as separate args) and not as one string.

lewis6991 (commented):
I've only done a superficial review but it looks fine to me, excluding the pending documentation we have discussed.

queries/python/indents.scm
lucario387: @georgeharker I think this could be separated into setting opening and closing delimiter rn, and could be done in this PR. The change would only require switching 5LOC

georgeharker:
Sounds like the tasks are:

- [ ] Make ```delimiter``` directive two directives - ```open_delimiter``` and ```close_delimiter```
- [ ] Rename ```dedent_bare_closing_delim``` and ```decent_lone_close_delimiter ```
- [ ] Rename ```final_line_indent``` to ```avoid_last_matching_next```
- [ ] Document ```immediate_indent```
- [ ] Document ```open_delimiter``` and ```close_delimiter```
- [ ] Document ```decent_lone_close_delimiter```
- [ ] Document ```avoid_last_matching_next```

queries/python/indents.scm
georgeharker: Tried but that doesn't work - I'll have another shot at that, but I'm not sure how that's supposed to work as it's hard to distinguish between not set and set without setting something that evaluates in a truthy way

queries/capnp/indents.scm
```diff
@@ -14,10 +14,12 @@
...
 
 ((struct_shorthand (property)) @aligned_indent
-  (#set! "delimiter" "()"))
+  (#set! "open_delimiter" "(")
+  (#set! "close_delimiter" ")"))
 
 ((const_list (const_value)) @aligned_indent
-  (#set! "delimiter" "[]"))
+  (#set! "open_delimiter" "(")
+  (#set! "close_delimiter" ")"))
```
lucario387, line 22: It's wrong here

queries/yuck/indents.scm
```diff
@@ -9,7 +9,8 @@
 
 ; TODO: can't get this to work, goal is to indent at the property ":" prefix
 ; ((list (identifier) (property)) @aligned_indent
-;   (#set! "delimiter" ":"))
+;   (#set! "open_delimiter" "(")
+;   (#set! "close_delimiter" ")"))
```
lucario387, line 13: Here too
Maybe this works now? Not sure

lucario387:
What does `dedent_lone_close_modifier` do?
Can it be substitute with `@branch`, or is the usage similar to `@branch` ? If yes then I will try to figure out something

lua/nvim-treesitter/indent.lua
```diff
@@ -36,13 +36,18 @@ end
...
+---@return boolean|nil only_delim_at_start
 local function find_delimiter(bufnr, node, delimiter)
   for child, _ in node:iter_children() do
     if child:type() == delimiter then
       local linenr = child:start()
       local line = vim.api.nvim_buf_get_lines(bufnr, linenr, linenr + 1, false)[1]
       local end_char = { child:end_() }
-      return child, #line == end_char[2]
+      local trimmed_before_delim
+      trimmed_before_delim, _ = line:sub(1, end_char[2] - 1):gsub("[%s%" .. delimiter .. "]*", "")
```
lucario387, line 47: I think this has to change, if `delimiter` is `s` for example, `"%s%" .. delimiter` will become `%s%s`

queries/python/indents.scm
georgeharker: agreed, pushed that change.

queries/capnp/indents.scm
georgeharker: thanks

queries/yuck/indents.scm
georgeharker: Certainly copy pasta.  I'll address that first and can look at whether it now works separately (I didn't change that comment, so it may or may not)

georgeharker:
> What does `dedent_lone_close_modifier` do?
> Can it be substitute with `@branch`, or is the usage similar to `@branch` ? If yes then I will try to figure out something

This controls the behavior for aligned indents' final closing delimiter

With:

```
l = [1
]
```

without

```
l = [1
       ]
```

This is admittedly a bit niche - and frankly whilst both are legal python I'm fine with the latter being our choice.  In fact the tests pass either way - so I'm going to remove this - it adds complexity and probably isn't preferred.

georgeharker:
I've added documentation, and removed the ```@end_indent``` that are no longer needed - they're only needed for ```@indent``` blocks.  I /think/ most of the remaining generator expressions etc probably should be ```@aligned_indent```.  I haven't done that in this PR yet.

doc/nvim-treesitter.txt
```diff
@@ -222,7 +222,112 @@ Supported options:
...
+    a,
+    b,
+    c
+  )
+<
+To specify the delimiters to use `open_delimiter` and `close_delimiter`
+should be used.  Eg.
+>
+ ((argument_list) @aligned_indent
+  (#set! "open_delimiter" "(")
```
lewis6991, line 296: When I said to separate these I had this in mind:

```
(#set! indent.delimeters "(" ")")
```

lua/nvim-treesitter/indent.lua
georgeharker: Yep, agreed, a proper escape here is more sensible, done.

doc/nvim-treesitter.txt
georgeharker: oh, ok.. I will rework then

georgeharker: Can I understand a bit more. 

Does 

```(#set! "Param" "(" ")")```

Work right now? A cursory reading of the query.lua implementation indicates it does not and that there can be only one value to set. 

Secondly, the location to set is unquoted and has a period in the path. Does that work right now? 

Is the ask that it would be 

```(#set! aligned_indent.delimiters "(" ")")```

Or

```(#set! indent.delimiters "(" ")")```

If the latter I don't quite understand as this is relevant to aligned_indent only.

File: lua/nvim-treesitter/indent.lua (hunk 1 above)
Comment (georgeharker) [reply lua/nvim-treesitter/indent.lua]: done.

File: doc/nvim-treesitter.txt (hunk 7 above)
Comment (georgeharker) [reply doc/nvim-treesitter.txt:296]: Yeah, ```runtime/lua/vim/treesitter/query.lua``` (see https://github.com/neovim/neovim/blob/master/runtime/lua/vim/treesitter/query.lua#L422)  has the set directive parser either accept ```(#set! @capture "key" "value")``` or ```(#set! "key" "value")``` as documented in the help.  Options are to revert back to having a single ```delimiters``` with a string assumed to be two chars long, or keep as-is.  I'm not sure how to do anything more than that without additional upstream changes.

Comment (lucario387) [reply doc/nvim-treesitter.txt:296]: I'd keep both set opening and closing delimiter (for now), as it works and is more easily extendable for the near future.
Besides, once `#set!` supports setting multiple values, it would be easy to change the code to accomodate for the changes.

File: queries/python/indents.scm (hunk 3 above)
Comment (georgeharker) [reply queries/python/indents.scm:23]: I am leaving wholesale changes to the way delimiters are specified out of this PR - it makes it larger and should be done with a clearer understanding of how that should match the way injections work.

File: doc/nvim-treesitter.txt (hunk 7 above)
Comment (georgeharker) [reply doc/nvim-treesitter.txt:296]: I just reverted that chunk of changes in how delimiters are specifed - so keeps the diff small.  It isn't the point of this PR, and I would if needed be happy to apply that as a follow on.  I did this before I saw your comment annoyingly.  But maybe it's still the right thing to do and apply separating them out as a separate PR.

File: queries/python/indents.scm (hunk 3 above)
Comment (georgeharker) [reply queries/python/indents.scm:23]: If the sentiment is strong to have them split, it's easy to do so.  But I can do so as a clear separate commit to view just those diffs, with the others squashed.

File: doc/nvim-treesitter.txt (hunk 7 above)
Comment (georgeharker) [reply doc/nvim-treesitter.txt:296]: If the sentiment is strong to have them split, it's easy to do so. But I can do so as a clear separate commit to view just those diffs, with the others squashed.

Comment (lucario387) [reply doc/nvim-treesitter.txt:296]: Cause we changing the `#set!` to different delimiters is easier than asking upstream to support `#set!` changing to list of values, it's a really big breaking change (I believe).

Comment (georgeharker) [reply doc/nvim-treesitter.txt:296]: ok, I brought that back and the docs still reflect it.  So it is 

```
(#set! "open_delimiter" "{")
(#set! "close_delimiter" "}")
```

Just that diff is a separate commit so it's clear which is doing what.

**Comment by georgeharker** [comment]
Ok to avoid the heavily indented replies:

Everything is rebased now. There's two commits in the current PR

One fixes the implementation of aligned indent and is relevant to any language which uses it including C. 

The other reworks the way delimiters are specified to split the open and close to two set directives which sounds like it will play nicer with future plans. Though the actual specification may again change back to one directive with two values when upstream changes occur. 

Docs reflect the combined state. 

It remains the case that some of the comprehensions might benefit from being aligned indent rather than indent. That should probably follow as a separate PR. That would just involve the python indent scm file. 

Please let me know if folks want more doing here. I'm going to leave it alone for a bit, but it would be lovely to get it squared away if everyone is in agreement.

**Comment by lucario387** [comment]
cc-ing @theHamsta for reviews :D

File: doc/nvim-treesitter.txt (hunk 7 above)
Comment (lewis6991) [reply doc/nvim-treesitter.txt:296]: Ok to keep them as separate directives for now, but please try and scope all captures and set keys into `indent.*`. And no, I don't believe quotes should not be necessary.

```
(#set! indent.delimiter.open "{")
(#set! indent.delimiter.close "}")
```

The `set` directives are completely implemented by us and `nvim-treesitter` can override the implementation if it wishes (until upstream can update).

Our queries aren't being shared right now, so what's important at this stage is to design a good system that's coherent with other query schemes.

File: doc/nvim-treesitter.txt [hunk 8]
```diff
@@ -222,7 +222,112 @@ Supported options:
...
+ ((if_statement) @indent
+   (#set! "immediate_indent" 1))
+<
+
+Will allow:
+>
+  if True:<CR>
+      # Auto indent to here
+
+`@indent_end`				*nvim-treesitter-indentation-indent_end*
```
lewis6991, line 257: Can we rename this to `@indent.end`?

doc/nvim-treesitter.txt
```diff
@@ -222,7 +222,112 @@ Supported options:
...
+An `@indent_end` capture is used to specify that the indented region ends and
+any text subsequent to the capture should be dedented.  
+
+`@branch`				*nvim-treesitter-indentation-branch*
```
lewis6991, line 261: `@indent.branch` ?

doc/nvim-treesitter.txt
```diff
@@ -222,7 +222,112 @@ Supported options:
...
+An `@branch` capture is used to specify that a dedented region starts
+at the line including the captured nodes.
+
+`@dedent`				*nvim-treesitter-indentation-dedent*
+A `@dedent` capture specifies dedenting starting on the next line.
+>
+`@aligned_indent`			*nvim-treesitter-indentation-aligned_indent*
```
lewis6991, line 268: `@indent.aligned`?

doc/nvim-treesitter.txt
```diff
@@ -222,7 +222,112 @@ Supported options:
       enable = true
     },
   }
+
+`@indent`				    *nvim-treesitter-indentation-queries*
+Queries can use the following captures: `@indent` and `@dedent`,
+`@branch`, `@indent_end` or `@aligned_indent`. An `@ignore` capture tells
+treesitter to ignore indentation and a `@zero_indent` capture sets
```
lewis6991, line 229: `@zero_indent` -> `@indent.reset`?

doc/nvim-treesitter.txt
```diff
@@ -222,7 +222,112 @@ Supported options:
...
+the indentation to 0.
+
+`@indent`				    *nvim-treesitter-indentation-indent*
```
lewis6991, line 232: `@indent.start`/`@indent.begin`?

lewis6991:
Sorry if my comments are pushing this PR out of the intended scope. However, the overall situation for indents has been bit of a no man's land, so I'd really like to start shaping this into something more coherent. Sooner rather than later, since not many languages have queries yet.

georgeharker:
Ok, cool, I can do that.

Right now, without changes in neovim where the set directive is defined, this will create keys like ```"indent.begin"``` so the code will go looking for ```metadata["indent.begin"]```, eventually I would imagine it would be good to have that go into a key ```"indent"``` on the node which would be hard to do in this change as it also needs a neovim patch.  So how about I redo the names and we can deal with the directive side after changes in neovim?

lucario387:
Imo, let's change what we can first, the new directives for python can be change to follow the syntax of upstream. It's only used in python as of now, so changing this should not be breaking to anyone
For the capture, it's best to keep it as it is for this PR, as you'd need to fix all the indents.scm files inside nvim-treesitter, which is a lot

lewis6991:
> For the capture, it's best to keep it as it is for this PR, as you'd need to fix all the indents.scm files inside nvim-treesitter, which is a lot

84 files. That is a lot.

I'll let @georgeharker decide what to do. I'd be happy to get this sorted ASAP.

georgeharker:
The changes for 

> Imo, let's change what we can first, the new directives for python can be change to follow the syntax of upstream. It's only used in python as of now, so changing this should not be breaking to anyone
> For the capture, it's best to keep it as it is for this PR, as you'd need to fix all the indents.scm files inside nvim-treesitter, which is a lot

I think I agree on the captures - looking at the implementation, we will have to alter the way that capture selection works too - (we'd need to do something like the ```insert_to_path``` during ```get_indents``` to decode into sensible names for the rest of the indent code).  However, similarly for the directives, the additional "." in the name does make the code less clean as you have to go and grab say ```q.aligned_indent[node:id()]["indent.delimiter"]``` Rather than ```q.aligned_indent[node:id()].indent.delimiter``` as the latter won't work rn.  My understanding is that the set directive we run is in neovim's ```runtime/treesitter``` - not the one in ```nvim-treesitter/query.lua```.

I'm totally happy to work on that and follow with a PR that does a more full rework of the scheme, I am wondering if it should be in this PR or a separate pair.

My proposal for a follow on PRs would be:

neovim:
* allow ```#set!``` to create paths for the name.  Something like ```insert_to_path```, such that ```((if_statement) @indent (#set! indent.immediate 1))``` would create ```indent.immediate``` rather than a key called ```"indent.imediate"```.
* Optionally allow ```#set!``` to set multiple values

nvim-treesitter:
* Have ```get_indents``` map the capture name to a local name to use in the code. For example, ```"indent.begin"``` would become available as ```q.begin[node:id()]```. 
* Define the query captures to be
  * indent.begin (currently indent)
  * indent.end (currently indent_end)
  * indent.branch (currently branch)
  * indent.align (currently align_indent)
  * indent.zero (currently zero_indent)
  * indent.ignore (currently ignore)
  * indent.dedent (currently dedent)
  * indent.auto (currently auto)
* Figure out if the directives should go under a path related to the capture ie when setting delimiter should you do ```(#!set indent.align.delimeter)``` which makes sense, but you'd also have to do ```(#!set indent.begin.increment)```
  * If so Define the set directives as
    * indent.<X>.delimeter (or ident.open_delimeter and ident.close_delimeter)
    * indent.<X>.avoid_last_matching_next
    * indent.<X>.increment
    * indent.<X>.immediate
    * indent.<X>.start_at_same_line (evidently used by cpp)
  * If not, there are questions about how to implement the set directive to have the set property appear on the relevant capture query.  But we would define the set directives as
    * indent.delimeter (or ident.open_delimeter and ident.close_delimeter)
    * indent.avoid_last_matching_next
    * indent.increment
    * indent.immediate
    * indent.start_at_same_line (evidently used by cpp)
* Change the docs
* Change all scm files

I do think there's some unknowns there and it seems like reaching consensus on the edits would be great.

Totally happy to help with this.  

Should we try to do so in a separate PR / open an issue to discuss the design then do a PR for neovim and a PR for the above?

lewis6991:
> allow #set! to create paths for the name. Something like insert_to_path, such that ((if_statement) @indent (#set! indent.immediate 1)) would create indent.immediate rather than a key called "indent.imediate".

This is really just an implementation detail which hasn't got a string enough proposition *right now*, and may end up obfuscating things too much. Keep this on the back burner and we can potentially revisit once everything else is implemented.

doc/nvim-treesitter.txt
clason: `@indent.zero` then

georgeharker:
Okay,

I have a commit pushed which has indent.X syntax for all the directives, and uses indent.property for the set directives.  All ```indents.scm`` files have been updated, all tests  pass locally.

The docs have been updated.

Captures are:

* indent.auto
* indent.begin
* indent.end
* indent.dedent
* indent.branch
* indent.ignore
* indent.align
* indent.zero

Properties are:

* indent.immediate
* indent.start_at_same_line
* indent.open_delimiter
* indent.close_delimiter
* indent.increment
* indent.avoid_last_matching_next

doc/nvim-treesitter.txt
georgeharker: I went for indent.align. But can alter if needed.

queries/bibtex/indents.scm
```diff
@@ -1,10 +1,10 @@
 [
  (entry)
-] @indent
+] @indent.begin.begin
```
lucario387, line 3: wrong capture here

queries/bibtex/indents.scm
```diff
@@ -1,10 +1,10 @@
...
 
 [
  "{"
  "}"
-] @branch
+] @indent.begin.branch
```
lucario387, line 8: .

queries/bibtex/indents.scm
```diff
@@ -1,10 +1,10 @@
...
 
-(comment) @ignore
+(comment) @indent.begin.ignore
```
lucario387, line 10: here too (im just going to use dots to notes)

queries/chatito/indents.scm
```diff
@@ -2,6 +2,6 @@
  (intent_def)
  (slot_def)
  (alias_def)
-] @indent
+] @indent.begin.begin
```
lucario387, line 5: .

queries/chatito/indents.scm
```diff
@@ -2,6 +2,6 @@
...
 
-(ERROR "]") @indent
+(ERROR "]") @indent.begin.begin
```
lucario387, line 7: .

queries/python/indents.scm
```diff
@@ -18,67 +14,94 @@
...
-(tuple_pattern ")" @indent_end)
-(list_pattern "]" @indent_end)
+(parenthesized_expression ")" @indent.end)
+(generator_expression ")" @indent.end)
+(list_comprehension "]" @indent.end)
+(set_comprehension "}" @indent.end)
+(dictionary_comprehension "}" @indent.end)
+
+(tuple_pattern ")" @indent.end)
+(list_pattern "]" @indent.end)
```
lucario387, line 87: shouldn't this be removed if you are using `@indent.align`

queries/python/indents.scm
```diff
@@ -18,67 +14,94 @@
...
+ (#set! indent.avoid_last_matching_next 1))
+((tuple) @indent.align
+ (#set! indent.open_delimiter "(")
+ (#set! indent.close_delimiter ")"))
 
-(list "]" @indent_end)
-(dictionary "}" @indent_end)
-(set "}" @indent_end)
+(ERROR "[" @indent.align (#set! indent.open_delimiter "[") (#set! indent.close_delimiter "]") . (_)) 
+(list "]" @indent.end)
```
lucario387, line 74: This should not be needed if you are using `@indent.align`, or has something changed again?

queries/python/indents.scm
```diff
@@ -18,67 +14,94 @@
...
 
-(parenthesized_expression ")" @indent_end)
-(generator_expression ")" @indent_end)
-(list_comprehension "]" @indent_end)
-(set_comprehension "}" @indent_end)
-(dictionary_comprehension "}" @indent_end)
+(ERROR "{" @indent.align (#set! indent.open_delimiter "{") (#set! indent.close_delimiter "}") . (_)) 
+(dictionary "}" @indent.end)
+(set "}" @indent.end)
```
lucario387, line 78: Also these 2

queries/query/indents.scm
```diff
@@ -1,8 +1,8 @@
 [
   (list)
-] @indent
+] @indent.begin.begin
```
lucario387, line 3: also here

queries/query/indents.scm
```diff
@@ -1,8 +1,8 @@
...
 
 [
   "["
   "]"
-] @branch
+] @indent.branch.begin
```
lucario387, line 8: this too

queries/smali/indents.scm
```diff
@@ -9,7 +9,7 @@
   (sparse_switch_directive)
   (subannotation_directive)
   (list)
-] @indent
+] @indent.begin.begin
```
lucario387, line 12: this too

lucario387:
Uhh, I think the `CONTRIBUTING.md` also needs update with these changes, that's supposed to be where the doc lies iirc

Also this needs to be marked as breaking change now

queries/bibtex/indents.scm
georgeharker: Thanks for catching these.  Did a quick additional check for "indent.begin.".  I thought i'd cleaned these up, but obv. missed a few.

queries/python/indents.scm
georgeharker: Yes, you are right.  Thanks

georgeharker:
> Uhh, I think the `CONTRIBUTING.md` also needs update with these changes, that's supposed to be where the doc lies iirc
> 
> Also this needs to be marked as breaking change now

Im not sure how to denote a breaking change - but I've updated CONTRIBUTING.md (as well as built in vim help - which is more detailed).

lucario387:
> 

This will most likely be squashed, again, so reword the first commit of the chain to 
`refactor(indent)!: Fix aligned_indent for python`

CONTRIBUTING.md
```diff
@@ -315,14 +315,14 @@ the node describing the language and `@content` to describe the injection region
...
-@indent_end     ; marks the end of indented block
-@aligned_indent ; behaves like python aligned/hanging indent
-@dedent         ; dedent children when matching this node
-@branch         ; dedent itself when matching this node
-@ignore         ; do not indent in this node
-@auto           ; behaves like 'autoindent' buffer option
-@zero_indent    ; sets this node at position 0 (no indent)
+@indent.begin       ; indent children when matching this node
+@indent.end         ; marks the end of indented block
+@indent.alig        ; behaves like python aligned/hanging indent
```
lucario387, line 320: Typo

georgeharker: thanks.

georgeharker:
> This will most likely be squashed, again, so reword the first commit of the chain to
> `refactor(indent)!: Fix aligned_indent for python`

Great - have done that.

lucario387:
> Great - have done that.

Come to think of it, im sorry for the initial suggestion. This scope is kind of much much bigger than what I suggested, so can you reword again + change the title of this PR? Refactoring the entire indent captures is big of its own

georgeharker:
> > Great - have done that.
> 
> Come to think of it, im sorry for the initial suggestion. This scope is kind of much much bigger than what I suggested, so can you reword again + change the title of this PR? Refactoring the entire indent captures is big of its own

No worries - have reworded that commit to be much more descriptive, changed the title of the PR

doc/nvim-treesitter.txt
```diff
@@ -222,7 +222,112 @@ Supported options:
       enable = true
     },
   }
+
+`@indent`				    *nvim-treesitter-indentation-queries*
+Queries can use the following captures: `@indent.begin` and `@indent.dedent`,
+`@indent.branch`, `@indent.end` or `@indent.alig`. An `@indent.ignore` capture tells
```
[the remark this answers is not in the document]
georgeharker: thanks.

georgeharker:
@theHamsta what do you think?

georgeharker:
@amaanq @theHamsta - there's a three of commits here now it's squashed.  If the latter commit which reworks things to be compatible naming wise with injections etc is problematic, I can split it out to a separate PR.  What are your thoughts on this going in?

lewis6991 (approved):

amaanq:
Seems fine to me, but now I have to rewrite the indents query in all my upstream parsers :neutral_face:

amaanq (approved):

georgeharker:
> @georgeharker Bass got merged, can you rebase and update indents for it

Done!

doc/nvim-treesitter.txt
```diff
@@ -222,7 +222,112 @@ Supported options:
...
+>
+  if True:<CR>
+      # Auto indent to here
+
+`@indent.end`				*nvim-treesitter-indentation-indent.end*
+An `@indent.end` capture is used to specify that the indented region ends and
+any text subsequent to the capture should be dedented.  
+
+`@indent.branch`		    *nvim-treesitter-indentation-indent.branch*
+An `@branch` capture is used to specify that a dedented region starts
```
clason, line 262:
```suggestion
An `@indent.branch` capture is used to specify that a dedented region starts
```

And similarly below

georgeharker: thanks

georgeharker:
Is anything else needed here?

lucario387:
I'm just waiting for this PR to try some new indent queries :)
Won't try to make this more and more, would rather do it myself later.

amaanq:
No, hamsta's just busy. I'll merge soon then :shrug:

lucario387:
> No, hamsta's just busy. I'll merge soon then shrug

An entry into #2293 as well right? :shy:

amaanq:
Here we go :grin:

amaanq:
Thank you for your hard work :)

amaanq:
@georgeharker I'm not sure if this is expected, and I guess I should've tested this out sooner (I don't really use python much so I wasn't really bothered to), but I have automatic ending pairs placed when I create a list/tuple/dict, when newlining inside that here's what happens:

Right before enter:
image

After:
image

Ideally the bracket would be aligned how it should be (below the opening bracket, so one indent back), but for most people's use cases I'm sure it's fine..

georgeharker:
That's super interesting.

I do not have auto ending pairs on.

So if I type

Screenshot 2023-03-24 at 10 47 13 AM

and then a CR I get

Screenshot 2023-03-24 at 10 48 21 AM


I tried also with ```noexpandtab``` set, my ts=4 and sw=4 anyway.  I get the same behavior as above which i think is correct. 

How do you enable the auto inserts of the closing parens?  Maybe I can have a play and figure out what's wrong.

georgeharker:
Update: if I have a different var name that's a different lenght and ```noexpandtab``` is set, it does seem to only have indented the close by 4.. hmm.

Screenshot 2023-03-24 at 10 53 44 AM

Let me see if I can understand what the deal is.

Screenshot 2023-03-24 at 10 55 54 AM

The first maybe should have ] dedented by one shift.  

Screenshot 2023-03-24 at 10 58 30 AM

I wonder if this was a late edit on the ```@indent.end``` being removed.

amaanq:
It seems to happen randomly, variable names being a potential trigger. I tried a different name and got this:

image

Would be nice if you could investigate :slightly_smiling_face:

georgeharker:
I'm having trouble matching what you're seeing  - how do I enable auto pairs?

georgeharker:
ignore the above - something has changed after a sync and it's not doing tree sitter indents - which means either something changed or I messed up.

georgeharker:
So repeating the above now with tree sitter actually on (Ran TSInstall all) ...

Screenshot 2023-03-24 at 12 04 27 PM

does

Screenshot 2023-03-24 at 12 05 56 PM

on return

Screenshot 2023-03-24 at 12 06 49 PM

does

Screenshot 2023-03-24 at 12 07 17 PM


Both of which I'm ok with.

Screenshot 2023-03-24 at 12 07 52 PM

The parser here has an issue here with the error state

Screenshot 2023-03-24 at 12 07 58 PM

When it should be more like this

Screenshot 2023-03-24 at 12 09 17 PM

But that's a separate issue and I'm not sure how to resolve that - it's not new to the changes, it just doesn't parse in a way the queries are likely to match it.

amaanq:
Hmm I see, thanks for the clarification :slightly_smiling_face: I guess I just presumed it felt awkward because my variable was just `a`, when in reality it was technically fine

georgeharker:
Screenshot 2023-03-24 at 12 18 44 PM

goes to

Screenshot 2023-03-24 at 12 18 51 PM

Which is so different from what you're seeing.  I'll try again with ```noexpandtab```.

Screenshot 2023-03-24 at 12 21 22 PM

does

Screenshot 2023-03-24 at 12 20 40 PM

which doesn't seem right

Screenshot 2023-03-24 at 12 21 51 PM

does

Screenshot 2023-03-24 at 12 21 58 PM

and after a entering contents and doing ```gg=G``


Screenshot 2023-03-24 at 12 24 19 PM

which again is not correct.

```:set expand tab```
```gg=G```

gets back 
Screenshot 2023-03-24 at 12 25 26 PM

So something is 'off' with noexpandtab.

**Comment by georgeharker** [comment]
I do think there's an issue - but in the interaction with expandtab.  And I think there's an issue with the parsing of error states which means it doesn't correctly indent 

```
def f():
    l = [<CR>
```

but the queries can't match it as the parse tree is pretty messed up at that point, and that too I think is a separate issue.  I think I'm still not seeing the same closing bracket behavior as you, it's not indented unless it's the aligned style (ie there's something on the opening delimiter line:

Screenshot 2023-03-24 at 12 30 10 PM

does

Screenshot 2023-03-24 at 12 30 54 PM

But I can't repro this one:

**Comment by amaanq** [comment]
It seems to happen with no items, even with my minimal config. See:

With 1, pressing enter after 1 (good):
image

Without 1, pressing enter inside the brackets (bad):
image

**Comment by georgeharker** [comment]
> It seems to happen with no items, even with my minimal config. See:
> 
> With 1, pressing enter after 1 (good): image
> 
> Without 1, pressing enter inside the brackets (bad): image
> 
> It would be nice to have the end bracket aligned "correctly", but I guess that would depend on the contents inside - I like to create the brackets first then put items in.

I think what you're seeing here is that in the first case it uses aligned mode, and in the second it doesn't align because it assumes (because there's nothing on the line after the [ ) that it's hanging indent style. I can't seem to repro the exact indents you're seeing, what are your ```ts, sw, and expandtab``` settings. For hanging mode it should have the ] at the same indent as mylist.

**Comment by axieax** [comment]
> It seems to happen randomly, variable names being a potential trigger. I tried a different name and got this (which, arguably, feels more right but that's because of the variable name):
> 
> image
> 
> Would be nice if you could investigate slightly_smiling_face

I just wanted to add that I've been experiencing the same problem with the Python TS parser. Sorry don't meant to revive this PR, but perhaps another issue can be created from this? Even with autopairs off, the same indent seems to occur every time I try to initialize a new multi-line list, dictionary or set etc. I've been tweaking around my settings for quite a bit, but am still unsure about how to resolve this issue.

I'm currently using the following options:
```
ts: 4
sw: 4
expandtab
```

**Comment by smartinellimarco** [comment]
In my case I found that Lazy.nvim is not correctly finding the correct module to setup when using the 'opts' key in the plugin configuration. This causes treesitter not being configured properly. You can double check your settings are correctly loaded by running 'TSConfigInfo'. The solution is doing a setup call using the 'config' key.
Example ('lua/plugins/treesitter.lua'):
This works.
```lua
return {
  {
    'nvim-treesitter/nvim-treesitter',

    -- Do not use 'opts' key since treesitter setup 
    -- module is not found correctly
    config = function()
        ensure_installed = 'all',
        highlight = {
          enable = true,
        },
        indent = {
          enable = true
        },
        incremental_selection = {
          enable = true,
          keymaps = {
            mnit_selection = "gnn",
            node_incnemental = "grn",
            scope_incremental = "grc",
            node_decremental = "grm",
          }
        }
      })
    end
  }
}
```
This does not
```lua
return {
  {
    'nvim-treesitter/nvim-treesitter',
    opts = {
      ensure_installed = 'all',
      highlight = {
        enable = true,
      },
      indent = {
        enable = true
      },
      incremental_selection = {
        enable = true,
        keymaps = {
          mnit_selection = "gnn",
          node_incnemental = "grn",
          scope_incremental = "grc",
          node_decremental = "grm",
        }
      }
    }
  }
}
```

lucario387:
I don't see why that's related to this PR

PR #4529, opened 2023-03-19
feat: add Luadoc

amaanq:
Mix of EmmyLua and what I noticed commonly seen in lua repos (and what lua_ls recognized)

### DIFF
--- a/queries/luadoc/highlights.scm
+++ b/queries/luadoc/highlights.scm
@@ -1,0 +1,133 @@
+; Keywords
+
+[
+  "@class"
+  "@type"
+  "@param"
+  "@alias"
+  "@field"
+  "@generic"
+  "@vararg"
+  "@diagnostic"
+  "@deprecated"
+  "@cast"
+  "@overload"
+  "@enum"
+  "@language"
+  "@see"
+  "extends"
+  (diagnostic_identifier)
+] @keyword
+
+[
+  "@async"
+] @keyword.coroutine
+
+(language_injection "@language" (identifier) @parameter)
+
+(function_type ["fun" "function"] @keyword.function)
+
+[
+  "@return"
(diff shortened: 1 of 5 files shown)


queries/luadoc/highlights.scm
```diff
@@ -0,0 +1,133 @@
...
+  "@see"
+  "extends"
+  (diagnostic_identifier)
+] @keyword
+
+[
+  "@async"
+] @keyword.coroutine
+
+(language_injection "@language" (identifier) @keyword)
```
ObserverOfTime, line 26: Shouldn't the identifier be something else?

ObserverOfTime (changes requested):

queries/luadoc/highlights.scm
amaanq: I couldn't think of a better fit, but yeah I agree. Here's an example

```lua
---@language JSON
local jsonText = [[{
    "name":"Emmy"
}]]
```

ObserverOfTime: Maybe `@constant`?

amaanq: Imo that doesn't feel exactly right, but is ok (another one could be `@type` but that also feels not exactly right), I'll just go w/ that

ObserverOfTime: Could just use `@parameter`.

amaanq: Yeah, that would fit better and be more consistent in highlighting with other annotations, that didn't cross my mind - thanks

amaanq: Forgot to push - fixed now

ObserverOfTime (approved):

PR #4534, opened 2023-03-20
comment(highlights): reduce user mention priority

ObserverOfTime:
This prevents it from overwriting luadoc (or other) highlights.


theHamsta (approved):

amaanq:
I've always gotten irked by this but forget it right away..thanks!

Issue #4554, opened 2023-03-24
Python indenting almost perfect but with some corner cases


bellini666:
After seeing [this commit](https://github.com/nvim-treesitter/nvim-treesitter/commit/d1333dd) I tried using indenting from treesiter on python again. Almost everything seems to be working fine now, except for some corner cases that I'll describe:

1) `raise`, `continue`, `break`, etc not returning to the previous indentation like `return` does.

For example, return works like this (which is the correct behaviour):

```python
if foo:
    return| <- Press enter here
| <- the cursor goes to here
```

But the following with `raise` does the following:

```python
if foo:
    raise Exception| <- Press enter here
    | <- the cursor goes to here
```

In that example, `raise` should behave the same as the return. The same goes for `continue`/`break` inside loops, and probably `pass` as well.

or

```python
def foo():
    try:| <- Press enter here
| <- the cursor goes to here
```

2) `match` indentation not working:

```python
match foo:| <- Press enter here
| <- the cursor goes to here
```

The correct behaviour would be:

```python
match foo:| <- Press enter here
    | <- the cursor goes to here
```

Also for:

```python
match foo:
    case bar:| <- Press enter here
| <- the cursor goes to here
```

The correct behaviour would be:

```python
match foo:
    case bar:| <- Press enter here
        | <- the cursor goes to here
```

3) `return` in the latest method in a class should go the the class indentation.

Currently what happens is:

```python
class Foo:
    def foo(self):
        return bar| <- Press enter here
| <- the cursor goes to here
```

The correct behaviour would be:

```python
class Foo:
    def foo(self):
        return bar| <- Press enter here
    | <- the cursor goes to here
```

Note that if I add another method defined after `foo`, the correct behaviour will happen

4) Similar to `1`, but when the return in a method is the await value of a call (yes, that specific)

```python
class Foo:
    async def foo(self):
        return await bar()| <- Press enter here
        | <- the cursor goes to here
```

The correct behaviour would be:

```python
class Foo:
    async def foo(self):
        return await bar()| <- Press enter here
    | <- the cursor goes to here
```

Note that the correct behaviour would happen for `return bar()`, `return bar`, `return await bar`, the issue only happens when it does `return await bar()`

---

There can be other examples, but those are the ones that I found after doing some tests with the latest treesiter version.

amaanq:
@georgeharker

amaanq:
FWIW

[...]

PR #4556, opened 2023-03-25
feat(python): add more edge-case indent pattern

lucario387:
- [x] match-case indents: https://docs.python.org/3/whatsnew/3.10.html
- [x] break-continue should dedent
~raise Exception dedent~ moved to discussion in #4557 
~fix return dedent~ moved to discussion in #4557

### DIFF
--- a/queries/python/indents.scm
+++ b/queries/python/indents.scm
@@ -44,6 +44,10 @@
  (#set! indent.immediate 1))
 ((with_statement) @indent.begin
  (#set! indent.immediate 1))
+((match_statement) @indent.begin
+ (#set! indent.immediate 1))
+((case_clause) @indent.begin
+ (#set! indent.immediate 1))
 
 (if_statement
   condition: (parenthesized_expression) @indent.align
--- a/queries/python/indents.scm
+++ b/queries/python/indents.scm
@@ -78,6 +78,11 @@
 
 (ERROR "{" @indent.align (#set! indent.open_delimiter "{") (#set! indent.close_delimiter "}") . (_)) 
 
+[
+  (break_statement)
+  (continue_statement)
+] @indent.dedent
+
 (parenthesized_expression ")" @indent.end)
 (generator_expression ")" @indent.end)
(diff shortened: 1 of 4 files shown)


queries/python/indents.scm
georgeharker, line 1: Agreed.

georgeharker (approved):
This looks good and is an improvement - this should go it, and we should try and figure out the details of whether it's node start or end that things are supposed to attach to separately to this.

queries/python/indents.scm
lucario387: I'm kindof trying to do return and raise Exceptions indent too, but yeah at the moment this is fine

lucario387:
@amaanq This resolves the 1st and 2nd request of #4554, the 3rd and 4th is much more than I expected so I will move it to a different PR

amaanq:
Hmm I would've thought Exceptions would be easy, is this fine to merge as is?

lucario387:
> Hmm I would've thought Exceptions would be easy, is this fine to merge as is?

Exceptions have
`raise Exception`
and 
`raise Exception(...)`
I won't push a fix just for the first case. The second one is align indented so needs a fix/change with that in mind

lucario387:
> Hmm I would've thought Exceptions would be easy, is this fine to merge as is?

Well it's a better than before state, and doesn't break anything else
With exceptions I expect it to be a little bit more breaking than usual so I will move it to a different PR to let this be accepted

amaanq:
I don't often see Exceptions with some code/message in it spanning multiple lines, but I could always be wrong.

lucario387:
> I don't often see Exceptions with some code/message in it spanning multiple lines, but I could always be wrong.

I also never use multiline Exceptions, but just to be safe, it'll be out of the scope of this PR

amaanq (approved):

georgeharker:
> I don't often see Exceptions with some code/message in it spanning multiple lines, but I could always be wrong.

I don't often either, but have done when the exception grabs a whole load of context to take along with it.

PR #4557, opened 2023-03-25
@indent.end should end at the node's end row, @indent.branch should end at the end of its capture

georgeharker:
The impetus for these changes are from looking at #4554

One of interesting cases there was

```
def foo(x):
    return bar(1,
               2,
               3)<CR>
```

Which failed to dedent properly.  It became clear when looking at this that the dedent wasn't occurring because an ```argument_list``` had already been processed which was considered to cover the return line - and the @indent.end was considered unnecessary because that line had already been processed.

To fix that, we can track processing for node ends separately to starts.

We can also simplify the return capture treatment with this approach (and add exceptions, break, continue etc). 

This also dedents correctly inside nested indentations.

What is troublesome is that the branch and dedents are still tracked at node start - if I alter them completely to operate on node end, which seems intuitive, cascading if-else in c gets messed up and doesn't dedent correctly - so I need to  think about that and would welcome some other's input on it too.  It is not totally clear why this would need to be the case to me.  I could sort of buy @indent.branch  needing to be tracked at node start.  And maybe what's needed is clarity in the docs about which, it's possible that the c indents just bake in assumptions about where the branch and dedent apply which could be adjusted.  

So I am marking this as draft as I imagine it will need some thinking.

Cascading ifs in c may actually not be correct even before this change.

For example

```
int f() {
    if (a && b && c &&
            d > 5) { // should be 4 less

    } else if (a == 4 and b ==6 &&
            4 == j) { // should be 3 more

    }
}
```

### DIFF
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -42,7 +42,7 @@ local function find_delimiter(bufnr, node, delimiter)
       local linenr = child:start()
       local line = vim.api.nvim_buf_get_lines(bufnr, linenr, linenr + 1, false)[1]
       local end_char = { child:end_() }
-      local trimmed_after_delim
+      local trimmed_after_delim, trimmed_before_delim
       local escaped_delimiter = delimiter:gsub("[%-%.%+%[%]%(%)%$%^%%%?%*]", "%%%1")
       trimmed_after_delim, _ = line:sub(end_char[2] + 1):gsub("[%s" .. escaped_delimiter .. "]*", "")
       trimmed_before_delim, _ = line:sub(1, end_char[2] - 1):gsub("[%s" .. escaped_delimiter .. "]*", "")
@@ -182,31 +182,37 @@ function M.get_indent(lnum)
     local is_processed = false
 
     local did_end = false
-    if not is_processed_by_row[erow]
-       and (q.indent["end"][node:id()] and q.indent["end"][node:id()]["indent.after"] and erow < lnum - 1)
+    if
+      not is_processed_by_row[erow]
+      and (q.indent["end"][node:id()] and q.indent["end"][node:id()]["indent.after"] and erow < lnum - 1)
     then
-       --indent = math.max(indent - indent_size, 0)
-       indent = indent - indent_size
-       is_processed = true
-       did_end = true
+      --indent = math.max(indent - indent_size, 0)
+      indent = indent - indent_size
+      is_processed = true
+      did_end = true
     end
 
-    if not is_processed_by_row[srow]
-       and (q.indent["end"][node:id()] and not q.indent["end"][node:id()]["indent.after"] and erow <= lnum - 1)
+    if
+      not is_processed_by_row[srow]
+      and (q.indent["end"][node:id()] and not q.indent["end"][node:id()]["indent.after"] and erow <= lnum - 1)
     then
-       --indent = math.max(indent - indent_size, 0)
-       indent = indent - indent_size
-       is_processed = true
-       did_end = true
+      --indent = math.max(indent - indent_size, 0)
+      indent = indent - indent_size
+      is_processed = true
+      did_end = true
     end
 
-    if not is_processed_by_row[srow] and not did_end
-       and ((q.indent.branch[node:id()] and srow <= lnum - 1 and erow >= lnum - 1) or
-            (q.indent.dedent[node:id()] and srow < lnum - 1 and erow >= lnum - 1))
+    if
+      not is_processed_by_row[srow]
+      and not did_end
+      and (
+        (q.indent.branch[node:id()] and srow <= lnum - 1 and erow >= lnum - 1)
+        or (q.indent.dedent[node:id()] and srow < lnum - 1 and erow >= lnum - 1)
+      )
     then
-       --indent = math.max(indent - indent_size, 0)
-       indent = indent - indent_size
-       is_processed = true
+      --indent = math.max(indent - indent_size, 0)
+      indent = indent - indent_size
+      is_processed = true
     end
     -- process any remaining indent anyway, often branch blocks will be reindented in
     -- chained conditionals
@@ -247,12 +253,11 @@ function M.get_indent(lnum)
     -- do not indent for nodes that starts-and-ends on same line and starts on target line (lnum)
     if should_process and q.indent.align[node:id()] and (srow ~= erow or is_in_err) and (srow ~= lnum - 1) then
       local metadata = q.indent.align[node:id()]
-      local o_delim_node, o_is_last_in_line, o_is_first_in_line  ---@type TSNode|nil, boolean|nil, boolean|nil
+      local o_delim_node, o_is_last_in_line ---@type TSNode|nil, boolean|nil
       local c_delim_node, c_is_last_in_line, c_is_first_in_line ---@type TSNode|nil, boolean|nil, boolean|nil
       local indent_is_absolute = false
       if metadata["indent.open_delimiter"] then
-        o_delim_node, o_is_last_in_line, o_is_first_in_line = 
-          find_delimiter(bufnr, node, metadata["indent.open_delimiter"])
+        o_delim_node, o_is_last_in_line, _ = find_delimiter(bufnr, node, metadata["indent.open_delimiter"])
       else
         o_delim_node = node
       end
@@ -277,16 +282,14 @@ function M.get_indent(lnum)
             if c_is_last_in_line then
               -- If current line is outside the range of a node marked with `@aligned_indent`
               -- Then its indent level shouldn't be affected by `@aligned_indent` node
-              if c_srow and (c_srow < lnum - 1)
-              then
+              if c_srow and (c_srow < lnum - 1) then
                 indent = math.max(indent - indent_size, 0)
               end
             end
             if c_is_first_in_line then
               -- If current line is outside the range of a node marked with `@aligned_indent`
               -- Then its indent level shouldn't be affected by `@aligned_indent` node
-              if c_srow and (c_srow <= lnum - 1 and metadata["indent.dedent_hanging_closing"])
-              then
+              if c_srow and (c_srow <= lnum - 1 and metadata["indent.dedent_hanging_closing"]) then
                 indent = math.max(indent - indent_size, 0)
               end
             end
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -182,18 +182,14 @@ function M.get_indent(lnum)
     local is_start_processed = false
     local is_end_processed = false
 
-    if
-      not is_end_processed_by_row[erow]
-      and q.indent["end"][node:id()] and erow < lnum - 1
-    then
+    if not is_end_processed_by_row[erow] and q.indent["end"][node:id()] and erow < lnum - 1 then
       indent = indent - indent_size
       is_end_processed = true
     end
 
     if
       not is_start_processed_by_row[srow]
-      and ((q.indent.branch[node:id()] and srow == lnum - 1) or
-           (q.indent.dedent[node:id()] and srow ~= lnum - 1))
+      and ((q.indent.branch[node:id()] and srow == lnum - 1) or (q.indent.dedent[node:id()] and srow ~= lnum - 1))
     then
       indent = indent - indent_size
       is_start_processed = true
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -145,7 +145,8 @@ function M.get_indent(lnum)
   end
 
   -- tracks to ensure multiple indent levels are not applied for same line
-  local is_processed_by_row = {}
+  local is_start_processed_by_row = {}
+  local is_end_processed_by_row = {}
 
   if q.indent.zero[node:id()] then
     return 0
@@ -155,6 +156,7 @@ function M.get_indent(lnum)
     -- do 'autoindent' if not marked as @indent
     if
       not q.indent.begin[node:id()]
+      and not q.indent["end"][node:id()]
       and not q.indent.align[node:id()]
       and q.indent.auto[node:id()]
       and node:start() < lnum - 1
@@ -177,18 +179,28 @@ function M.get_indent(lnum)
 
     local srow, _, erow = node:range()
 
-    local is_processed = false
+    local is_start_processed = false
+    local is_end_processed = false
 
     if
-      not is_processed_by_row[srow]
-      and ((q.indent.branch[node:id()] and srow == lnum - 1) or (q.indent.dedent[node:id()] and srow ~= lnum - 1))
+      not is_end_processed_by_row[erow]
+      and q.indent["end"][node:id()] and erow < lnum - 1
     then
       indent = indent - indent_size
-      is_processed = true
+      is_end_processed = true
+    end
+
+    if
+      not is_start_processed_by_row[srow]
+      and ((q.indent.branch[node:id()] and srow == lnum - 1) or
+           (q.indent.dedent[node:id()] and srow ~= lnum - 1))
+    then
+      indent = indent - indent_size
+      is_start_processed = true
     end
 
     -- do not indent for nodes that starts-and-ends on same line and starts on target line (lnum)
-    local should_process = not is_processed_by_row[srow]
+    local should_process = not is_start_processed_by_row[srow]
     local is_in_err = false
     if should_process then
       local parent = node:parent()
@@ -203,7 +215,7 @@ function M.get_indent(lnum)
       )
     then
       indent = indent + indent_size
-      is_processed = true
+      is_start_processed = true
     end
 
     if is_in_err and not q.indent.align[node:id()] then
@@ -287,7 +299,7 @@ function M.get_indent(lnum)
             indent = indent
           end
         end
-        is_processed = true
+        is_start_processed = true
         if indent_is_absolute then
           -- don't allow further indenting by parent nodes, this is an absolute position
           return indent
@@ -295,7 +307,8 @@ function M.get_indent(lnum)
       end
     end
 
-    is_processed_by_row[srow] = is_processed_by_row[srow] or is_processed
+    is_start_processed_by_row[srow] = is_start_processed_by_row[srow] or is_start_processed
+    is_end_processed_by_row[erow] = is_end_processed_by_row[erow] or is_end_processed
 
     node = node:parent()
   end
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -181,15 +181,15 @@ function M.get_indent(lnum)
 
     local is_processed = false
 
+    local did_end = false
     if not is_processed_by_row[erow]
        and (q.indent["end"][node:id()] and q.indent["end"][node:id()]["indent.after"] and erow < lnum - 1)
     then
        --indent = math.max(indent - indent_size, 0)
        indent = indent - indent_size
        is_processed = true
+       did_end = true
     end
-    -- do not process indent, dedent if there's already an end
-    is_processed_by_row[srow] = is_processed_by_row[srow] or is_processed
 
     if not is_processed_by_row[srow]
        and (q.indent["end"][node:id()] and not q.indent["end"][node:id()]["indent.after"] and erow <= lnum - 1)
@@ -197,11 +197,10 @@ function M.get_indent(lnum)
        --indent = math.max(indent - indent_size, 0)
        indent = indent - indent_size
        is_processed = true
+       did_end = true
     end
-    -- do not process indent, dedent if there's already an end
-    is_processed_by_row[srow] = is_processed_by_row[srow] or is_processed
 
-    if not is_processed_by_row[srow]
+    if not is_processed_by_row[srow] and not did_end
        and ((q.indent.branch[node:id()] and srow <= lnum - 1 and erow >= lnum - 1) or
             (q.indent.dedent[node:id()] and srow < lnum - 1 and erow >= lnum - 1))
     then
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -45,7 +45,8 @@ local function find_delimiter(bufnr, node, delimiter)
       local trimmed_after_delim
       local escaped_delimiter = delimiter:gsub("[%-%.%+%[%]%(%)%$%^%%%?%*]", "%%%1")
       trimmed_after_delim, _ = line:sub(end_char[2] + 1):gsub("[%s" .. escaped_delimiter .. "]*", "")
-      return child, #trimmed_after_delim == 0
+      trimmed_before_delim, _ = line:sub(1, end_char[2] - 1):gsub("[%s" .. escaped_delimiter .. "]*", "")
+      return child, #trimmed_after_delim == 0, #trimmed_before_delim == 0
     end
   end
 end
@@ -145,8 +146,7 @@ function M.get_indent(lnum)
   end
 
   -- tracks to ensure multiple indent levels are not applied for same line
-  local is_start_processed_by_row = {}
-  local is_end_processed_by_row = {}
+  local is_processed_by_row = {}
 
   if q.indent.zero[node:id()] then
     return 0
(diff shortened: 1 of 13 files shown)


queries/python/indents.scm
```diff
@@ -84,22 +88,10 @@
...
-    (attribute 
-      attribute: (_) @indent.end)
-    (call
-      arguments: (_ ")" @indent.end))
-    "return" @indent.end
-  ] .)
+((return_statement) @indent.end)
+((raise_statement) @indent.end)
+((break_statement) @indent.end)
+((continue_statement) @indent.end)
```
lucario387, line 94:
```suggestion
[
  (return_statement)
  (raise_statement)
  (break_statement)
  (continue_statement)
] @indent.end
```

georgeharker: Good point, thanks.

lucario387:
I'll keep trying on #4556 with a `indents.scm` changes only for #4554, but this might be more preferred

lucario387:
> Cascading ifs in c may actually not be correct even before this change.

Parenthesized expressions were never captured with `@indent.align` though, at least from the codeblock you used, it seemed you expect an it to align

georgeharker:
I think I would.

I think I lack some clarity on whether indent.dedent should apply after the tagged block whereas @indent.branch should apply at the start - I probably just need a bit of time to think (more sleep). 

There's something there which is troubling me in terms of cascading ifs with multi line conditions and it'd be great to get that squared away. 

You should definitely do the scm oriented fixes tho

queries/c/indents.scm
```diff
@@ -9,6 +9,12 @@
   (expression_statement)
 ] @indent.begin
 
+(
+ (parenthesized_expression) @indent.align
+ (#set! indent.open_delimiter "(")
+ (#set! indent.close_delimiter ")")
+)
+
```
lucario387, line 17: Is this the norm for C-style languages?

georgeharker: I feel like C has definitive camps and I'd hate to be picking one vs the other.

see
https://codingart.readthedocs.io/en/latest/c/Formatting.html

it's not uncommon to want the multiline conditional statement to be aligned within the parenthesis - which with sw=4,ts=4 and just ```if ( ``` probably doesn't matter but would for a cascaded if with ```else if (``` where the normal indent would just be 4.

Aligned doesn't one vs the other that it just say it'll align if you start on the same line...

Open to feedback on that.

georgeharker: When you get into C stylistics, there's a lot of different opinions

https://softwareengineering.stackexchange.com/questions/3450/most-readable-way-to-format-long-if-conditions

But most of them have some form of alignment with the open paren.

amaanq: There's different styles/preferences for sure with C. I'll check out how this works in a few

lucario387: > 

Through the first look, this seems to be a question of
- Are they indenting it to indent + 4
- Or are they indenting it so it aligns with the opening parenthesis

georgeharker:
The more I think about it, the more convinced i am that both ```@indent.branch``` and ```@indent.dedent``` begin at the start of the node.  Whereas ```@indent.end``` begins at the end of the capture.  I would love some other thoughts on that though.

queries/c/indents.scm
georgeharker: I'm going to remove it from this PR - it's something to think about, but it wasn't the point, so it should go separately.

amaanq: Just FYI, not only does this depend on user preference, but imo it depends on where the parenthesized expression is too. Maybe an easy way to have user-configurable options/styles would be nice

lucario387:
instead of this, can we ... create a new one? :D 
I know looking at Helix may not be nice but they have `@extend.prevent-once`, which is used only for return/break statements in ... python and gdscript, so I don't think it's bad to add something similar to it here

georgeharker:
That's certainly a less potentially risky idea.  

Please bear with me, i'm still mid quite a bit of scratching my head on this.

What I think I found odd is that @indent.end doesn't actually dedent, and from a documentation standpoint it would be great to have clarity for folks on when a given indent rule applies (start of the matched capture, or end).

- ```@indent.begin``` is sort of clear, the lines after the start of the matched node are indented (unless indent.immediate is set, in which case it's the entire node)
- ```@indent.branch``` says the start line is dedented
- ```@indent.dedent``` says the lines after the start of the matched node are dedented, until the end of the matched node
- ```@indent.end``` says...? As currently implemented it has an effect but isn't actually explicitly dedenting.

Right now in a language like ruby, eg

```
def somefun
  assert f is not true
  <type end>
```

the fact that end dedents isn't due to it's being captured by ```@indent.end``` it's because it's also captured by ```@indent.branch``` which performs the actual dedent.  You can see this by commenting out the "end" in the ```indent.branch``` line for ruby.  In fact ```@indent.end``` most of the time does nothing at all, unless the next line is blank, in which case it grabs the first node of the current line rather than the last of the previous line - but given we know the line is blank, that basically just skips up the node tree, just undoing the special case treatment of blank lines grabbing the previous line's ending node.

I think this is why in "C" we have to tag "}" as ```@indent.branch```, because basically the tag as ```@indent.end``` doesn't really do much.


So to take some examples, multiline conditions I think things get interesting, for a hypothetical grammar (ignoring the cascading part of C)

```
if (condone &&
    condtwo) {
    statements1;
} elseif (condthree &&
          condfour) {
    statements2;
}
```

If should indent the block - ie the compound statement that ends on ```elseif``` line.  If it were not a compound statement, it would begin after the if condition (ie on the statement_expr in the consequence), but it's not the if itself.  However, here it is ```{``` that should be ```@indent.begin```. The ```}``` seems like it would be an ```@indent.end``` (though the definition of where that applies is wooly to me).  ```elseif``` seems like it would be a ```@indent.branch```, whilst it's condition might be ```indent.align````.  In fact, here, ```elseif``` being ```@indent.branch``` wouldn't matter, as the ```}``` is already an ```@indent.end``` if that did something meaningful.

The thing about the above is that because the ```@indent.branch``` and ```@indent.end``` are single line it sort of doesn't matter if it's the start or end of the node.

Hanging returns are another interesting case, avoiding python for a moment, what would we expect from a return statement in a C-like language?

```
int foo(int a, int b, int c, int e) {
    return somefun(a, b, c,
                   e);<CR>
```

the above maybe conflates an assumption about aligned indent on parameters, so how about

```
int foo(int a, int b, int c, int e) {
    return somefun(a, b, c,
        e);<CR>
```

We might assume ```@indent.begin``` is on the compound statement block's ```{``` of the function.  Maybe we want return in c-like languages to dedent, in which case it would be rational to tag ```@indent.end``` on say a return_expression if there were such a thing - but where does it occur?  Probably after the return_expression.

I like the idea you had of a new capture, so by way of exploring what that might look like:

- ```@indent.begin``` the lines after the start line of the matched node are indented This will be true through the entire matched node. (unless ```indent.immediate``` is set, in which case it's the entire node, including the start line - mostly used to ensure indents after errors).  
- ```@indent.branch``` says the start line _only_ is dedented
- ```@indent.end``` says ? something like the line after the start of the captured node is dedented
- ```@indent.dedent``` says the lines after the start of the matched node are dedented, until the end of the matched node
- ```@indent.end_after``` says the line after the end of the matched node is dedented

There's some asymmetry there.  ```@indent.end``` still has dubious meaning, and why isn't there an ```@indent.begin_after``` (it would end up with similar treatment to ```@indent.end``` right now and get applied to the list after the end, only on blank lines) ?  It also remains necessary to tag ```}``` as ```@indent.branch``` to get it to dedent automatically.

Alternatively,

- ```@indent.begin``` the lines after the start line of the matched node are indented This will be true through the entire matched node. (unless ```indent.immediate``` is set, in which case it's the entire node, including the start line - mostly used to ensure indents after errors).  
- ```@indent.branch``` says the start line through to the end of the node is dedented - (this through to end of node change makes zero breaking diffs on current queries but makes things symmetrical).  It would apply if a dedent has not already been applied - dedents do not stack ontop of branches, just like indents do not for the same lines
- ```@indent.end``` says the lines starting at the matched node is dedented. This will be true through to the end of the matched node.  (unless ```@indent.after``` is set which will dedent the line after the end, used for returns).  An end precludes subsequent indent or dedent on the same line.
- ```@indent.dedent``` says the lines after the start of the matched node are dedented, until the end of the matched node.  Dedent is stacked together and does not preclude subsequent indent.
- ```@indent.align``` treats two styles in one go, if the line is a sole open delimiter, it will use hanging style.  If it is not, it will choose align style.  The ```@indent.open_delimiter``` and ```@indent.close_delimiter``` should be set.  Additionally, setting ```indent.dedent_hanging_closing``` will cause a lone ending delimiter (first in line) to dedent for the hanging style, and ```indent.avoid_last_matching_next``` will cause the final line to indent one more to avoid clashing with the next (Eg. for python).


Maybe this is clearer, maybe it is not - I would welcome feedback.

It definitely explicitly handles ```@indent.end```.  It seems to remove the need to use lots of ```@indent.branch``` to get the auto dedents to work.  It seems to be easy to make the modifications to the ```indents.scm``` files for various languages, and I've done so for C, go, css, python, ruby already - obviously all would need doing.

I'll update the PR which is not too horrific, and in many ways simpler than the original one.

This is not something I would want to rush in, it really does require some thought

georgeharker:
It wasn't super hard to fix the grammars that exist - assuming tests are doing a decent job.

I think I could now write an accurate statement about each of the captures, when to use it and the range over which it applies / their options. - it would be pretty close to above, I just would want to verify the statements.

There's some subtly about what precludes what - end precludes dedent / branch (so they don't stack).  But doesn't actually preclude an indent, unless it was an indent from a node above.  Though multiple indents are still disallowed - that was true before too, as was a dedent being overridden by an indent, and their interaction with parent nodes on the same line.

I guess my goal is to have explainable, an defensible meanings for all the indent captures... oh and also make returns and raise statements work in python.

georgeharker:
@amaanq @lucario387 @theHamsta for input / thoughts.

lua/nvim-treesitter/indent.lua
```diff
@@ -251,7 +282,14 @@ function M.get_indent(lnum)
             if c_is_last_in_line then
               -- If current line is outside the range of a node marked with `@aligned_indent`
               -- Then its indent level shouldn't be affected by `@aligned_indent` node
-              if c_srow and c_srow < lnum - 1 then
+              if c_srow and (c_srow < lnum - 1) then
+                indent = math.max(indent - indent_size, 0)
```
lucario387, line 286: I am still unsure of whether we should have math.max() here
My thoughts then were to have this as a "reset" button for going out of range, and not "reset back to at least 0" button. It often times go below 0 and then will get added back up to the correct value

lucario387:
> It wasn't super hard to fix the grammars that exist - assuming tests are doing a decent job.

Sadly, it's probably not. We have 80+ indent query files, with only ~25-30 of them have testing. And even those aren't capturing all the typical cases either. If it's not breaking for those test cases then probably it's fine,

lua/nvim-treesitter/indent.lua
georgeharker: Agreed. It's an odd one and the ordering is strange because we deal with decent first but iirc it can go negative on some lines before being reindented. 

I need to look at some examples with the debug statements I have locally on. Thanks for pointing this out.

lua/nvim-treesitter/indent.lua
```diff
@@ -251,7 +282,14 @@ function M.get_indent(lnum)
...
+              end
+            end
+            if c_is_first_in_line then
+              -- If current line is outside the range of a node marked with `@aligned_indent`
+              -- Then its indent level shouldn't be affected by `@aligned_indent` node
+              if c_srow and (c_srow <= lnum - 1 and metadata["indent.dedent_hanging_closing"]) then
                 indent = math.max(indent - indent_size, 0)
```
lucario387, line 293: This one too, but im not sure. WIth your changes it seems like the behavior for this case is different for some cases out there. Of course it won't affect anyone and anything because it's not widely used yet, but may also need to reconsider

georgeharker:
Ahh. Ok. That's a good point. 

I would have to take a pass on checking those then. Mostly for where ```@indent.branch``` is used.

lua/nvim-treesitter/indent.lua
lucario387: Second look and if the line contains a single ending delimiter, it may get dedented twice

lucario387:
The indent.branch changes seems nice though :). The c-style chaining if-else may have an easier way out than bunch of edge cases like now

lua/nvim-treesitter/indent.lua
georgeharker: agreed on care needed - the other changes removed some ```@indent.dedents``` which in python were hiding the inherent operation of the ```@align.indent```

Which is actually

```
b = [
    1,
    2,
    3,
    ]
````

if as in python

```
b = [
    1,
    2,
    3,
]
```

is desirable, the new option does that.  The ```@indent.branch``` did that before but also made this complex, and ```@indent.end``` didn't actually do that.

I think you are correct, but we already indented by one just above.  So the indent + dedent gets us to 0, and the next catches the dedent.

However, it reads oddly, should be revisited. 

TBH, I'm pretty surprised by some of the python indents - particularly in the nested collections.  However, PEP8 checkers and other linters for python say it's correct and minor changes report it's no longer correct.

I'll revisit this.

**Comment by georgeharker** [comment]
@lucario387 thanks for your comments. I'll take some time to think them over and won't mark them as resolved until further discussion

**Comment by lucario387** [comment]
Of course, this is experimental anyways :)
I'm also trying a different approach, with #4558 .
Taking the inspiration from `@extend.prevent-once` of Helix, it's to make the node follow its parent node's indent level. I'll probably prefer this approach over mine as it has more granular control.

**Comment by georgeharker** [comment]
Sounds good. It would be great to understand what you discover there- especially with @indent.end and the need for @indent.branch

It's definitely ideal situation would be that the documented behaviors are sufficient to write queries without having to mod the code.  There's a subjective element there as to whether the docs are sufficient and whether the captures do the right things and / or provide a sufficient palette of primitives to easily write those queries. Honestly the branch thing had me scratching my head for a while. And whilst it seems ok for it to be used to fix particular peculiarities of certain languages it feels like right now end and branch take similar roles and you have to use both often.

**Comment by amaanq** [comment]
I really should add indent tests for my languages 😆

**Comment by lucario387** [comment]
If you are changing the behavior of @indent.end, we should also consider also changing its behavior before the indent calculation
https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/indent.lua#L129-L137

Probably find the last character after trimmed and removing commentstrings is better, because if whitespaces/comments exist then @indent.end will not be captured at all
(Also why c indents were working quite well, until you have a trailing space :))

**Comment by georgeharker** [comment]
Is it reasonable to assume the comment query is called comment? Or is there any other way to determine comment chars?

**Comment by amaanq** [comment]
Do you mean the capture? All comments will be under the `@comment` capture.

**Comment by georgeharker** [comment]
> Do you mean the capture? All comments will be under the `@comment` capture.

There's some subtly here - there's two times an @end might be found, on return as a previous line and on the actual ending line.  We're talking here about the first that ensures dedent after.  

It's possible to do as you suggested - but in theory a comment capture can imply indenting to comments... so I want to be careful that the indent code can see comment nodes.

**Comment by georgeharker** [comment]
Come to think of it, because we would now capture ends on the actual line end, there's a few cases

When typing

```
def foo() {
    <insert }>
```

which will capture the end on that line and dedent, correctly.. immediately } is typed we get

```
def foo() {
}
```

Then

```
def foo() {
    <insert } /*foo*/>
```

Which will dedent on the insertion of } just the same and then type /*foo*/

When that happens the next line is blank and we're into the code looking for the previous line last node - which will be comment.  But.. it doesn't matter, we already dedented.

That's not true for whole file indent.

Then there's 

```
def foo() {
    <insert }> /*foo*/
```

This still dedents in live typing


Previously this was complex because } would have dedented due to @indent.branch, which would have left the next line indented, and we would have gone to find the last node and the end handling was implicit.

So, IIUC the remaining concern would be something like whole file indent on


```
int foo() {
}
   /*foo*/
   int a() {
   }
```

whereby we would want to find the blank line previously, ensure we went one before that and found the end - which the comment interferes with.

however it actually does work to produce

```

int foo() {
}
/*foo*/
int a() {
}
```
as does the whole file indent on

```

int foo() {
}
   /*foo*/

   int a() {
   }
```

**Comment by georgeharker** [comment]
You're correct

```
def foo():
    if True:
        return somecomplexfn(a,b,c,
                             d,e,f) # return<CR>
*cursor here
```

vs

```
def foo():
    if True:
        return somecomplexfn(a,b,c,
                             d,e,f)<CR>
    *cursor here
```

**Comment by lucario387** [comment]
> You're correct
> 
> 
> 
> ```
> 
> def foo():
> 
>     if True:
> 
>         return somecomplexfn(a,b,c,
> 
>                              d,e,f) # return<CR>
> 
> *cursor here
> 
> ```
> 
> 
> 
> vs
> 
> 
> 
> ```
> 
> def foo():
> 
>     if True:
> 
>         return somecomplexfn(a,b,c,
> 
>                              d,e,f)<CR>
> 
>     *cursor here
> 
> ```

this has been a recurring issue with the indent module that was probably neglected? thanks to semi colon.

As this can have cascading effects to all other indent captures, it'd be nice to figure out what the starting node should be for each indent calculation

**Comment by georgeharker** [comment]
Right then. Rolls up sleeves

**Comment by lucario387** [comment]
> > You're correct
> > 
> > 
> > 
> > ```
> > 
> > def foo():
> > 
> >     if True:
> > 
> >         return somecomplexfn(a,b,c,
> > 
> >                              d,e,f) # return<CR>
> > 
> > *cursor here
> > 
> > ```
> > 
> > 
> > 
> > vs
> > 
> > 
> > 
> > ```
> > 
> > def foo():
> > 
> >     if True:
> > 
> >         return somecomplexfn(a,b,c,
> > 
> >                              d,e,f)<CR>
> > 
> >     *cursor here
> > 
> > ```
> 
> this has been a recurring issue with the indent module that was probably neglected? thanks to semi colon. I also kind of look over it and never care about it until like a few weeks ago
> 
> As this can have cascading effects to all other indent captures, it'd be nice to figure out what the starting node should be for each indent calculation

**Comment by lucario387** [comment]
Through purely anecdotal usage of this branch
- I like the new `indent.end`, but have a feeling that this is becoming more and more similar to `indent.dedent`. Maybe soft deprecate `indent.dedent`?
- `indent.begin` is not exactly behaving like how the docs is mentioning it
  - Docs: Indent children of this node
  - Reality: Indent lines after this node
  - Realistically, if this were to only indent children nodes, then `indent.dedent` use case will only be to dedent cascading indents, (i.e if-else chain)
- `indent.branch` -> `indent.dedent-once`, maybe? To match the exact use case of it, mostly for interactive typing
- No comment about `indent.{align,zero,ignore,auto}` for now

**Comment by lucario387** [comment]
I don't really know where to put this line of thoughts so I will put it here :)

Designs for indent logic, not related to this branch at all:
- `indent.begin` is not exactly behaving like how the docs is mentioning it
  - Docs: Indent children of this node
  - Reality: Indent lines after this node
  - `indent.immediate` can easily fill in the gap so it behaves like the old `indent.begin`.
    - This can probably be a separate PR right away, but will need to test for 80+ languages tho :sigh:
  - Realistically, if this were to only indent children nodes, then `indent.dedent` use case will only be to dedent cascading indents, (i.e if-else chain)
- `indent.branch` -> `indent.dedent-once`, maybe? To match the exact use case of it, mostly for interactive typing
- How do we define handlers for `is_in_err`? 
  - As long as a node has a sibling containing error (prev or next siblings don't matter), then it will affect the entire tree. Most recent example: #4525 
  - Should the queries be narrowed to account for this edge case, or how should we change the behavior without having performance issues :). 
- How do we find the start node for indent calc?
  - `@indent.end` is used to short out the calculation in master now. Which is nice, but what would be the cost for re-running the entire queries :)
  - Current logic is affected by whitespaces/trailing comments

**Comment by georgeharker** [comment]
> * `indent.begin` is not exactly behaving like how the docs is mentioning it
>   
>   * Docs: Indent children of this node
>   * Reality: Indent lines after this node
>   * `indent.immediate` can easily fill in the gap so it behaves like the old `indent.begin`.
>     
>     * This can probably be a separate PR right away, but will need to test for 80+ languages tho :sigh:
>   * Realistically, if this were to only indent children nodes, then `indent.dedent` use case will only be to dedent cascading indents, (i.e if-else chain)

I think that's my fault.  I didn't change the behavior of ```@indent.begin``` vs when it was ```@indent``` in previous changes.  Nor should this PR change what ```@indent.begin``` does.  So I think when I wrote the docs, I just made an incorrect statement - which certainly should be fixed.  It's the ```CONTRIBUTING.md``` doc that's wrong.  The help text is correct.

I will add a task to fix that.

**Comment by georgeharker** [comment]
> How do we define handlers for `is_in_err`?

This one is tough - because the error state parse trees can be quite messy / not matching expectations.

**Comment by georgeharker** [comment]
I guess the most accurate statement of ```@indent.begin``` is


```@indent.begin``` the lines after the start line of the matched node are indented This will be true through the entire matched node. (unless ``` indent.immediate l``` is set, in which case it's the entire node, including the start line - mostly used to ensure indents after errors).

But I think that's an accurate statement before this change. And notwithstanding the indent.immediate option accurate before too.

**Comment by georgeharker** [comment]
Ah.... ok, I finally get it.

There's a few possible modalities here

either indents attach to nodes over their range.

Or they attach at a start or end of a captured node and apply thereafter.

The original and current HEAD implementation of ```@indent.begin``` is open ended.  It starts at the next line and continues forever. (Or handles some special cases if indent.immediate is set - which prevents it not opening the indent if the start and end are the same).

Similarly HAD and original implementation of ```@indent.branch``` and ```@indent.dedent``` are open ended.

I can see pros and cons of these for the insert newline type uses,, but it seems to make things harder.

With that realization in mind i'll try and gain some clarity on options for what ```@indent.end``` means and bounded variants of these captures.

**Comment by lucario387** [comment]
The insert newline can be replicable if we remove indent.end, make indent.begin child only, and have it to be enforced to newlines with `indent.immediate`.

**Comment by lucario387** [comment]
> With that realization in mind i'll try and gain some clarity on options for what `@indent.end` means and bounded variants of these captures.

No need to rush, it can also be a separate PR from this, and then rebased into here :D

**Comment by georgeharker** [comment]
Totally. I think I have a decent understanding. I certainly don't want to push the mess I have right now, but I think I can see a way to get somewhere.

**Comment by georgeharker** [comment]
I also likely have working code to deal with comments etc when inserting new lines. That's actually a bit more hairy than it first looked.

**Comment by lucario387** [comment]
> I also likely have working code to deal with comments etc when inserting new lines. That's actually a bit more hairy than it first looked.

Is it though, my version is like this
```lua
  if is_empty_line then
    local prevlnum = vim.fn.prevnonblank(lnum)
    --- Todo: How to handle tabs
    local indent = vim.fn.indent(prevlnum)
    -- It should not be affected by trailing whitespace
    local prevline = vim.trim(vim.fn.getline(lnum))
    node = get_last_node_at_line(root, prevlnum)
    if node:type():match "comment" then
      -- Should ignore same line comment
      -- but not comments that spans the whole line
      -- And re-trim the trailing spaces
      local start_node = get_first_node_at_line(root, prevlnum)
      if start_node:id() ~= node:id() then
        prevline = vim.trim(prevline:sub(1, node:start() + 1))
        local col = indent + #prevline
        node = root:descendant_for_range(prevlnum - 1, col - 1, prevlnum - 1, col - 1)
      end
    end

    -- If captured node is @indent.end
    -- This means the previous tree has ended and should not be accounted for indent calculation
    -- if q.indent["end"][node:id()] then
    --   node = get_first_node_at_line(root, lnum)
    -- end
  else
    node = get_first_node_at_line(root, lnum)
  end
```

georgeharker:
Something like that. I'm skipping possibly multiple comments because some languages have comments that are per line. Sigh.

lucario387:
I don't think it's nice to skip multiple ones though
That just means skipping the entire comment tree, don't think it's nice (What if you are in a comment parser injected area)

georgeharker:
I'm also not sure we should be going for start of line - because of interplay with possible range bounded indents - which I was playing with.

georgeharker:
> I don't think it's nice to skip multiple ones though
> That just means skipping the entire comment tree, don't think it's nice (What if you are in a comment parser injected area)

I think it makes a difference if you're inside a comment vs if there are multiple between you and the previous node.

For example in C

```
for (;;) {
    while (x>4) {
        // stuff
    } // until done
    // this is a continued comment<CR>
}
```

it definitely wouldn't be good to skip out of inside a comment.

I've actually made some decent progress on possibilities for node-range-bound captures (ie. the range of a capture indent wise is bounded by the nodes it attaches to, not everything after).

I still need a bit of time to consider some other languages, but I think there's some really nice properties of these.  It's not clear if they work for all languages or not.  So it might be best to implement the current captures too (possibly renamed).

It seems like all captures have a form like this

```
StartLine
/\
 | Body
\/
EndLine
|  After
\/
````

Where Body is start < line < end (and may be empty)

I think it's possible to express all indent / dedent captures in this form and have them be modified by ```#set!``` to alter which of the above they include or exclude, and have sensible defaults for ```indent.begin```, ```indent.end```, ```indent.dedent``` and ```indent.branch```.

One of the nice properties of having the ranges not be everything-after-start is that the cascading ifs get much simpler.  I have a branch with a sketch up on my GitHub but it /really/ needs some cleanup before I rebase and push it to a PR,  and many languages need mods - so I'd want to go implement the current indents as that feels way safer, and allows migration over time.

Finally, ```indent.immediate``` is an artifact of wanting not to have indents occur on lines which open /close same line.  This gets interesting with nodes which are already scoped only to the start line, when a newline is inserted... however, the need goes away for it entirely if the node we find when looking backwards after a "blank" line is allowed to expand to the current line and the above structure helps with this (the "after" portion is easily automatically allowed in specific circumstances).

I might have something in a few days which would be worth discussion.

lucario387:
That is, I feel like, a lot of calculation just to get the starting node for indentation

From what you described, it seems like we will need to reach sibling nodes of the current node, and then possibly more magic manipulation too, just to get what you want.

Tbh, my implementation is also not that fast (one line :sigh:)

Will be checking out your implementation some day next week :)

georgeharker:
I agree. It does feel like a lot. Thankfully it's only on a new line with a blank not for every line. It's a tough one.

lucario387:
#4592 is merged, does this need a rebase? :thinking:

lucario387:
After some testing, again :)

I like this idea more now, and come to realization that separating indent.{branch,dedent,end} from indent.{begin,align} is a nice concept, but will require more thoughtful implementation (and queries rewrite again)

As of now. All dedent calc is applied before indent is taken into consideration, but if a line is already indented by a child's node, then that line can no longer be dedented. There is definitely a middle ground here to improve upon. And controlling whether each line is applied at most one dedent from a node at `srow` seems okay-ish

lucario387:
> I made some decent progress on this but am still slammed work wise - i'll see if I can find some time to upload what I had.

It's fine :D. No need to rush.

clason:
Since the master branch is locked and we are considering migrating indentation to Zed's approach on the main branch, I'm closing this. I do appreciate the effort and look forward to future contributions!

PR #4558, opened 2023-03-25
feat(python): add return/raise dedents

lucario387:
(random plan that I come up with, instead of changing `@indent.end` behavior, I will add a directive to allow control over it, either that or add some extra queries to make sure it works)

- [ ] Add the new directive/capture 
  - It will most likely follow Helix's [`@extend.prevent-once`](https://docs.helix-editor.com/guides/indent.html#capture-types)
- [ ] add raise exceptions dedents
- [ ] fix return dedents
- [x] add tests for both cases

### DIFF
--- a/queries/python/indents.scm
+++ b/queries/python/indents.scm
@@ -81,34 +81,65 @@
 [
   (break_statement)
   (continue_statement)
+  (raise_statement)
+  (return_statement)
 ] @indent.dedent
 
-(parenthesized_expression ")" @indent.end)
-(generator_expression ")" @indent.end)
-(list_comprehension "]" @indent.end)
-(set_comprehension "}" @indent.end)
-(dictionary_comprehension "}" @indent.end)
+(return_statement 
+  [
+    ; indent.align
+    (list "]" @indent.dedent)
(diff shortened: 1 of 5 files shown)


queries/python/indents.scm
```diff
@@ -81,35 +81,66 @@
...
+  ]
+)
+
+(raise_statement
+  (call
+    arguments: (argument_list ")" @indent.dedent)))
+
+(parenthesized_expression 
+  ")" @indent.end
+  (#not-has-ancestor? @indent.end return_statement raise_statement))
```
theHamsta, line 126: I'd prefer copy&paste to `has-ancestor?` for perf reasons.

lucario387: > I'd prefer copy&paste to `has-ancestor?` for perf reasons.

I don't exactly know how to make `has-ancestor?` works because there are so many possible other ancestors, as compared to 2 outliers
Can I use `not-has-parents?` instead? I have laid out which cases needs it but have yet to push it

amaanq: parents would be fine, way less expensive

PR #4563, opened 2023-03-26
feat(cpp): simplify namespace query rules now that parser is a bit more consistent

jdrouhard:
Most recent parser version [0b6d0eb9abdf7cea31961cd903eeed5bbd0aae74](https://github.com/tree-sitter/tree-sitter-cpp/commit/0b6d0eb9abdf7cea31961cd903eeed5bbd0aae74) improves the node tree related to namespaces. This deprecates the need for extra rules to try to get @namespace applied to the right nodes.

also, `namespace_identifier` can't name a constant since it only names things guaranteed to be namespaces, or type/enums when in a qualified identifier as part of scope resolution.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -63,7 +63,7 @@
     "revision": "eedb93bf9e22e82ed6a67e6c57fd78731b44f591"
   },
   "cpp": {
-    "revision": "03fa93db133d6048a77d4de154a7b17ea8b9d076"
+    "revision": "0b6d0eb9abdf7cea31961cd903eeed5bbd0aae74"
   },
   "css": {
     "revision": "769203d0f9abe1a9a691ac2b9fe4bb4397a73c51"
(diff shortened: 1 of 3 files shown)


amaanq:
update the lockfile to point to that revision

jdrouhard:
> update the lockfile to point to that revision

Should say that this query will work just fine with the current version, it just won't look quite as nice until that other commit. Required was maybe too forceful. I don't mind waiting on the github action to update it in the normal course of things.

amaanq:
Kotlin is breaking the lockfile update but I'll probably hotfix that shortly, up to you

amaanq (approved):

amaanq:
query test is failing

jdrouhard:
Cuda and hlsl need to be updated as well, I guess?

amaanq:
> Cuda and hlsl need to be updated as well, I guess?

not familiar with these, but probably...

@theHamsta

theHamsta:
I have a cron workflow updating cuda and half for CPP updates. I've already updated cuda.  Hlsl only has a 24h cadence so I will trigger manually now. Should be fine to also update them in a few minutes to their newest version

amaanq:
Arduino needs an update as well

@ObserverOfTime

PR #4500, opened 2023-03-13
Add support for ponylang

mfelsche:
This add support for ponylang by adding this parser: https://github.com/mfelsche/tree-sitter-ponylang to the supported parsers, setting the revision, adding myself as the maintainer etc.

The queries seem legit. Highlighting works well. All queries have been tested in the tree-sitter playground and with nvim-treesitter as good as possible.

PR #4565, opened 2023-03-26
feat: add Pony

amaanq:
Supersedes #4500

cc @mfelsche


queries/pony/highlights.scm
```diff
@@ -0,0 +1,289 @@
...
+; Conditionals
+
+[
+  "if"
+  "ifdef"
+  "iftype"
+  "then"
+  "else"
+  "elseif"
+  "until"
```
ObserverOfTime, line 67: Should "until" be `@repeat`? What are we using in other languages that have it?

queries/pony/highlights.scm
```diff
@@ -0,0 +1,289 @@
...
+  "as"
+  "consume"
+  "recover"
+  "object"
+  "where"
+] @keyword
+
+[
+  "fun"
+  "be"
```
ObserverOfTime, line 30: "be" should be @keyword.coroutine.

queries/pony/highlights.scm
```diff
@@ -0,0 +1,289 @@
...
+
+; Fields
+
+(field name: (identifier) @field)
+
+(member_expression "." (identifier) @field)
+
+; Constructors
+
+(constructor "new" @constructor (identifier) @constructor)
```
ObserverOfTime, line 122: "new" should be @keyword.operator.

ObserverOfTime (changes requested):

queries/pony/highlights.scm
amaanq: Seems like `@repeat` in most languages

ObserverOfTime (approved):

PR #4566, opened 2023-03-26
feat: add FIRRTL

amaanq:
https://github.com/chipsalliance/firrtl


queries/firrtl/highlights.scm
```diff
@@ -0,0 +1,191 @@
...
+
+  "infer"
+  "read"
+  "write"
+  "rdwr"
+] @keyword
+
+[
+  "defname"
+] @keyword.variable
```
ObserverOfTime, line 46: Move this with the other keywords.

amaanq:
Hmm, gh isn't showing the update I made but I pushed it already, seems to be semi-down rn

PR #4567, opened 2023-03-27
feat(scala): add in local variables queries

ckipp01:
This syncs the queries that were updated in https://github.com/tree-sitter/tree-sitter-scala/pull/196.

PR #4568, opened 2023-03-27
docs: add a small note about local queries

ckipp01:
This just adds a small note in the CONTRIBUTING section about local
queries and how they differ from those listed upstream.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -240,6 +240,12 @@ Used for XML-like tags.
 
 ### Locals
 
+Note: pay specific attention to the captures here as they are a bit different to
+those listed in the upstream [Local Variables
+docs](https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables).
+Some of these docs didn't exist when `nvim-treesitter` was created and the
+upstream captures are more limiting than what we have here.
+
 ```scheme
 @definition            ; various definitions
 @definition.constant   ; constants


theHamsta (approved):

CONTRIBUTING.md
```diff
@@ -240,6 +240,12 @@ Used for XML-like tags.
 
 ### Locals
 
+Note: pay specific attention to the captures here as they are a bit different to
+those listed in the upstream [Local Variables
+docs](https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables).
+Some of these docs didn't exist when `nvim-treesitter` was created and the
+upstream captures are more limiting than what we have here.
+
```
theHamsta, line 248: Is it relevant/necessary to give reason why they are different? Maybe it gives additional context, but I don't know whether upstream is more limiting

ckipp01: Ah, well this comes from [this comment](https://github.com/nvim-treesitter/nvim-treesitter/pull/4567#issuecomment-1484910842). I wasn't sure why they were different and this was the answer I got. I interpreted 

> and two, the captures they "have" are sometimes not enough for us

meaning that the upstream ones were more limited. Is that not correct? Or did I misunderstand this?

ObserverOfTime: Upstream doesn't have subcaptures within definitions. But the plan is to eventually be as consistent with upstream as possible. Once a standard for highlight captures has been decided, maybe it can be extended to locals.

ckipp01: > Upstream doesn't have subcaptures within definitions. But our plan is to eventually be as consistent with upstream as possible.

I think this is a great plan.

PR #4578, opened 2023-03-28
feat(spell): support lisp like languages

kawas44:
Hello,

Here is a small PR to add spell feature to commonlisp, scheme, racket, clojure and fennel.

Regards,
Kawas


ObserverOfTime (changes requested):
You can add `@spell` next to the other capture.

kawas44:
> You can add `@spell` next to the other capture.

Ok, it is now done more cleanly except for `scheme` where we should not have `@spell` on sexp-comment or byte-str.

ObserverOfTime (approved):

PR #4580, opened 2023-03-29
Fix highlighting for Python

flamarique:
Plugin `vscode.nvim` is based on `nvim-treesitter.nvim`. Currently, there is a bug in the highlighting which should be addressed.

The following code snippet looks like this in VSCode:
Screenshot 2023-03-29 at 13 24 32

whereas it looks as follows when using `vscode.nvim` - `nvim-treesitter.nvim`:
Screenshot 2023-03-29 at 13 24 15

- keyword "class" should be coloured like "def" or "lambda"
- it should distinguish between constructor function definition from constructor call (this is the case in VSCode)

### DIFF
--- a/queries/python/highlights.scm
+++ b/queries/python/highlights.scm
@@ -247,11 +247,11 @@
 [
   "def"
   "lambda"
+  "class"
 ] @keyword.function
 
 [
   "assert"
-  "class"
   "exec"
   "global"
   "nonlocal"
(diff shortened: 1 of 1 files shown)


amaanq:
Class is not a function keyword, and we don't distinguish constructor declarations vs calls. Take this issue/PR up with your theme not us (you can change this yourself in your after/queries/python runtime path as well).

PR #4585, opened 2023-03-30
fix(css): Capture `class_selector` as `@type`

anthony-S93:
Before:
2023-03-30-155728_436x177_scrot

After: 
2023-03-30-155609_459x215_scrot


queries/css/highlights.scm
```diff
@@ -19,6 +19,8 @@
  (universal_selector)
  ] @type
 
+((class_selector) @type (#set! "priority" 110))
```
ObserverOfTime, line 22: Does it not work without the priority?

lucario387:
https://github.com/nvim-treesitter/nvim-treesitter/blob/a21c7e0a05a50c3ebf1389bc37b8e8048be04485/queries/css/highlights.scm#L49-L52

This have higher priority as it's placed later. You should narrow down the query or change that instead of setting priority

queries/css/highlights.scm
MunifTanjim: can we do it for `class_name` and `id_name` instead?

Remove `class_name` and `id_name` from

```
[
 (class_name)
 (id_name)
 (property_name)
 (feature_name)
 (attribute_name)
 ] @property
```

and add

```
[
 (class_name)
 (id_name)
] @type
```

MunifTanjim: While we're at it. Maybe also do:

```
(class_selector "." @punctuation.special)
(id_selector "#" @punctuation.special)
```

lucario387: id isn't really type tho, more of a specifier/variable

MunifTanjim:
```css
#id-name,
.class-name {
}
```

Yeah, they're not really `@type`... same for `class_name`.

But I'm not sure if any of the existing captures fits them.

They're not `@variable` either, css has variable and they're a different thing.

They are specifier. But is there a capture group for that?

amaanq: maybe `@attribute`, idk css

MunifTanjim: attribute too are different thing 😂 (updated the example in previous comment)

PR #4586, opened 2023-03-30
fix: set filetype for markdown_inline parser

seblj:
## Problem:

`markdown_inline` doesn't have a connected filetype, which means the newly merged feature `vim.treesitter.language.get_filetypes` won't work as expected for markdown if the parser is used.

## Solution

Set `filetype = "markdown"` on the `markdown_inline` parser for it to correctly refer back to markdown as the filetype for the parser

## Context

https://github.com/neovim/neovim/pull/22643#issuecomment-1490040341

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -894,6 +894,7 @@ list.markdown_inline = {
     files = { "src/parser.c", "src/scanner.c" },
   },
   maintainers = { "@MDeiml" },
+  filetype = "markdown",
   readme_name = "markdown_inline (needed for full highlighting)",
   experimental = true,
 }


clason:
I don't think this will work like that; won't this also start the `markdown_inline` parser for `markdown` file (which is not wanted)?

seblj:
Woops, yes you are right... We need another way to only register the filetype for a parser then

clason:
Probably best handled by the commenting plugin, though -- treat `markdown_inline` as a synonym for `markdown`.

seblj:
Hmm, not sure I agree since that would mean all commenting plugins (and other plugins where parser->ft would be nice) need to have a special case for `markdown_inline` when it would be really that it would "just work"...

But sure, it isn't too much trouble for the plugin authors so it works

clason:
Right now, `markdown` _is_ special. The calculation changes with more such split parsers, or other examples where one-way mappings are necessary.

In any case, thank you for raising this point; it's something I'll need to think about.

clason:
In fact, I think _not_ registering these parsers for a filetype is in fact the correct thing to do so commenting plugins can walk up the tree and find the correct commentstring for the injecting parser (in this case, `markdown`) -- otherwise you could never uncomment with the `comment` parser installed.

clason:
You would iterate based on filetype -- these "partial" parsers should not be associated to a filetype, so you'd keep walking up the tree until you get a language that has a filetype associated. (This would be an argument for either registering all parsers even if `lang`== `filetype`, or "unregistering" partial parsers, which means we'd need to flip the table.)

seblj:
> You would iterate based on filetype -- these "partial" parsers should not be associated to a filetype, so you'd keep walking up the tree until you get a language that has a filetype associated. (This would be an argument for either registering all parsers even if `lang`== `filetype`, or "unregistering" partial parsers, which means we'd need to flip the table.)

Yes! Either one of those methods would probably work

clason:
So nothing to be done, in fact :)

PR #4587, opened 2023-03-30
fix(ruby): @conditional highlights

treywood:
These highlight queries caused some odd highlighting for me when I have nested modules inside of a conditional:

Screenshot 2023-03-30 at 3 17 40 PM

In the above screenshot I do not expect the `::` between constants to be red, but they're getting the `@conditional.ruby` highlight.

The `(then)` node in `(if (then))` is actually the entire block guarded by the conditional, not just the "then" keyword, which I think was the intended target of these queries.

### DIFF
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -44,16 +44,8 @@
  "if"
  "unless"
  "when"
+ "then"
  ] @conditional
-
-(if
-  "end" @conditional)
-(if
-  (then) @conditional)
-(unless
-  (then) @conditional)
-(elsif
-  (then) @conditional)
 
 [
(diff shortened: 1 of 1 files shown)


amaanq (approved):

queries/ruby/highlights.scm
```diff
@@ -44,17 +44,9 @@
  "if"
  "unless"
  "when"
+ "then"
  ] @conditional
 
-(if
-  "end" @conditional)
```
ObserverOfTime, line 50: Shouldn't this be kept?

amaanq:
Was gonna ask about that but I forgot to actually send the comment...

PR #4589, opened 2023-03-31
Polishing JS highlighting

veged:
The main idea behind this PR is to introduce more detailed groups for some punctuation and few more things.


queries/ecma/highlights.scm
```diff
@@ -129,20 +165,22 @@
 (template_string) @string
 (escape_sequence) @string.escape
 (regex_pattern) @string.regex
-(regex "/" @punctuation.bracket) ; Regex delimiters
+(regex "/" @punctuation.bracket.regex)
+(regex_flags) @string.regex.flags
+((string_fragment) @string.special
+ (#eq? @string.special "use strict"))
```
ObserverOfTime, line 171: This could be `@preproc`.

queries/ecma/highlights.scm
```diff
@@ -30,24 +30,71 @@
...
-  name: (identifier) @function)
-(method_definition
-  name: [(property_identifier) (private_property_identifier)] @method)
+(function name: (identifier) @function)
+(function_declaration name: (identifier) @function)
+(generator_function name: (identifier) @function)
+(generator_function_declaration name: (identifier) @function)
+(method_definition name: [(property_identifier) (private_property_identifier)] @method)
+(method_definition name: (property_identifier) @method.builtin
+  (#eq? @method.builtin "constructor"))
```
ObserverOfTime, line 97: This is fine.

queries/ecma/highlights.scm
```diff
@@ -30,24 +30,71 @@
...
+  "eval"
+  "isFinite"
+  "isNaN"
+  "parseFloat"
+  "parseInt"
+  "decodeURI"
+  "decodeURIComponent"
+  "encodeURI"
+  "encodeURIComponent"
+  "require"))
```
ObserverOfTime, line 86: This is good.

queries/ecma/highlights.scm
```diff
@@ -30,24 +30,71 @@
...
+  "ArrayBuffer"
+  "DataView"
+  "Error"
+  "EvalError"
+  "InternalError"
+  "RangeError"
+  "ReferenceError"
+  "SyntaxError"
+  "TypeError"
+  "URIError"))
```
ObserverOfTime, line 73: These should be `@type.builtin`.

ObserverOfTime (changes requested):
I'm against the changes in this PR except for the ones I commented on.

veged:
> I'm against the changes in this PR except for the ones I commented on.

actually the most valuable (from my point of view) is the bunch of `@punctuation.*` declarations which allows to highlight differently same punctuation symbols in different contexts (i.e. curly brackets in statements and in object literals) — is it something wrong with that idea?

clason:
I don't think we officially want to have third-layer captures (which become four-layer with the implicit language fallback). The goal for _this project_ is to define consistent captures that can be targeted by colorschemes; too detailed language-specific captures are out of scope and better suited to a language-specific plugin.

We are also trying to align more with upstream and Helix, so this works counter to that goal (unless those also have similar punctuation captures?)

veged:
> I don't think we officially want to have third-layer captures (which become four-layer with the implicit language fallback). The goal for _this project_ is to define consistent captures that can be targeted by colorschemes; too detailed language-specific captures are out of scope and better suited to a language-specific plugin.

that's the matter of performance or just complexity in semantic?

> We are also trying to align more with upstream and Helix, so this works counter to that goal (unless those also have similar punctuation captures?)

could you please provide some repos? I'll check it...

queries/ecma/highlights.scm
lucario387: this should be separated into `@variable.builtin` for arguments, module, windows, console, document and `type.builtin` for everything else i think

queries/ecma/highlights.scm
```diff
@@ -196,91 +234,99 @@
...
+(function body: (statement_block ["{" "}"] @punctuation.bracket.function))
+(arrow_function parameters: (formal_parameters ["(" ")"] @punctuation.bracket.function))
+(arrow_function parameters: (formal_parameters "," @punctuation.delimiter.function))
+(arrow_function "=>" @punctuation.special.function)
+(arrow_function body: (statement_block ["{" "}"] @punctuation.bracket.function))
+(generator_function_declaration "*" @punctuation.special.function)
+(method_definition "*" @punctuation.special.function)
+(method_definition parameters: (formal_parameters ["(" ")"] @punctuation.bracket.function))
+(method_definition parameters: (formal_parameters "," @punctuation.delimiter.function))
+(method_definition body: (statement_block ["{" "}"] @punctuation.bracket.function))
```
lucario387, line 293: I personally think this section is a bit too much and seems like what you want to configure on your own. Probably better to let users add them instead, or only some of these should be added

queries/ecma/highlights.scm
```diff
@@ -196,91 +234,99 @@
...
-[
-  "function"
-] @keyword.function
+"debugger" @keyword @keyword.debugger
 
-[
-  "new"
-  "delete"
-] @keyword.operator
+"with" @keyword @keyword.with
```
lucario387, line 320: Is there any specific reason why you would want both of these captures for one node?

queries/ecma/highlights.scm
```diff
@@ -196,91 +234,99 @@
...
+
+["return" "yield" "export"] @keyword @keyword.return
+(export_statement "default" @keyword @keyword.return)
+
+(break_statement "break" @keyword @keyword.break)
+(continue_statement "continue" @keyword @keyword.continue)
+
+"function" @keyword @keyword.function
+
+["throw" "try" "catch" "finally"] @keyword @exception
```
lucario387, line 332: Same with this

queries/ecma/highlights.scm
veged: yep — I think it's useful to have ability for different highlights for different "types" of keywords

lucario387: As in, why are they captured with `@keyword` and `@keyword.x`. It will fall back to `@keyword` if `@keyword.x` doesn't exist, so no need to capture twice

veged: oh... that's nice! I didn't know about such "magic" ;-) will fix accordingly

queries/ecma/highlights.scm
lucario387: Hmm, on second look, `Promise` and `Intl` aren't exactly types though?
Probably more separation is necessary

ObserverOfTime: > Hmm, on second look, `Promise` and `Intl` aren't exactly types though?

They are classes.

lucario387: Intl is almost never used as types in typescript im pretty sure. Promise im okay with it as type

lucario387: For consistencies, I agree.

In general usage, probably not. I will just override locally, that's probably better

ObserverOfTime: Maybe `@namespace.builtin` for `Intl`?

lucario387: Definitely more correct than `@type.builtin` to me

queries/ecma/highlights.scm
veged: @lucario387 

> As in, why are they captured with `@keyword` and `@keyword.x`. It will fall back to `@keyword` if `@keyword.x` doesn't exist, so no need to capture twice

actually there is a case for double declaration `@x` and `@x.y` — if I have some highlights for both of them and wish to combine them accordingly — i.e. make all `@keyword` bold and have specific colours for each `@keyword.TYPE` (now it's applies only most precise highlight declaration, i.e. colours from `@keyword.TYPE` without anything from common `@keyword`)

lucario387: Then they can be applied from userspace instead. No need to flush the main queries with unnecessary separations like such

PR #4607, opened 2023-04-04
update sql parser

treywood:
https://github.com/DerekStride/tree-sitter-sql/pull/132

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -438,7 +438,7 @@
     "revision": "05f949d3c1c15e3261473a244d3ce87777374dec"
   },
   "sql": {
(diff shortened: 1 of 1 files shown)


nerdrew (approved):

clason:
Thanks, but we have an automated script that does this once a day; a dedicated PR is only necessary if there are breaking changes that require updating queries as well.

Or is this urgent for some reason?

treywood:
> Or is this urgent for some reason?

Nope it's not urgent, I just didn't know that this was automated. feel free to close!

clason:
Next run is in ~12h then :)

PR #4618, opened 2023-04-08
Add Python else elif dedent

lucario387:
not guarantee to work, just testing the waters 

- If this doesn't work so well, maybe changing the indent algo instead
- Does not support multiline elif statement

### DIFF
--- a/queries/python/indents.scm
+++ b/queries/python/indents.scm
@@ -83,6 +83,14 @@
   (continue_statement)
 ] @indent.dedent
 
+(ERROR
+  (_) @indent.dedent ":" .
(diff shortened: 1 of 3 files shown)


lucario387:
This is mostly just a QOL for when in error nodes

 For example:
```python
if True:
  ...
  else:
```
It will change to 
```python
if True:
  ...
else:
```

ObserverOfTime (approved):

theHamsta (changes requested):
Can we have some tests for the changes?

lucario387:
> Can we have some tests for the changes?

Just added one.

theHamsta (approved):

PR #4621, opened 2023-04-08
 fix(rust): highlight SNAKE_CASE scoped identifier name as @constant 

tomtomjhj:
<details>
<summary>outdated</summary>

fix(rust): don't highlight scoped constant as @type
* This rule is redundant because the `@type` rule at line 8 already does the job and is not overriden by other rule.
* This rule falsely overrides the rule for `@constant`, which resulted in CONSTANT_ITEM in Self::CONSTANT_ITEM being highlighted as `@type`.

</details>

### DIFF
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -82,6 +82,9 @@
 ((scoped_identifier
     name: (identifier) @type)
  (#lua-match? @type "^[A-Z]"))
+((scoped_identifier
+    name: (identifier) @constant)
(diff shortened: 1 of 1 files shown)


theHamsta:
Wouldn't that change apply to more rules in the `; Assume that uppercase names in paths are types` section?

tomtomjhj:
Oh I guess I misunderstood the grammar structure. I restored the rule and added a rule for constant.

theHamsta (approved):

PR #4622, opened 2023-04-08
feat(markdown_inline): add @nospell where appropriate

tomtomjhj:


theHamsta (approved):

theHamsta:
Thank you!

 I assume we need to set `@nospell` also for code because we can't be sure the code will have an injection to set spell/nospell on its own.

@MDeiml

MDeiml:
Yes, fenced and indented code blocks should probably have nospell

tomtomjhj:
If `@nospell` is added to `[ (link_title) (indented_code_block) (fenced_code_block) ] @text.literal` in `markdown`, comments in the code block are not spell checked, despite that the comment has `@spell.injected_lang` which is applied after `@nospell.markdown`.

PR #4623, opened 2023-04-08
fix(dart): try statement wouldn't indent

RobertBrunhage:
## Before

```dart
void test() {
  try {
  | <--- cursor is here
  } catch(e) {}
}
```

## After

```dart
void test() {
  try {
    | <--- cursor is here
  } catch(e) {}
}
```

### DIFF
--- a/queries/dart/indents.scm
+++ b/queries/dart/indents.scm
@@ -10,6 +10,7 @@
   (list_literal)
   (return_statement)
   (arguments)
+  (try_statement)
 ] @indent.begin
 
 [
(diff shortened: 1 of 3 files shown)


amaanq (approved):

theHamsta:
Can you add your example as a test please?

theHamsta (approved):

RobertBrunhage:
@theHamsta or @amaanq while working on this I also found that this was a problem (writing the try part and expanding the `{}` would result in an incorrect indentation and "error" in the `:InspectTree`. See screenshot below.

image

Not too sure what can be done to fix that (was there before my fix also)

PR #4625, opened 2023-04-08
refactor(html_inject): Use lua-match instead of match

lucario387:
Use equivalent pattern to `match? (a|b)` using lua-match instead.


amaanq:
is the preceding whitespace check necessary?

lucario387:
> is the preceding whitespace check necessary?

Just to be safe, in case some hash keys contains the string

amaanq (approved):

PR #4626, opened 2023-04-08
feat(cpp): improve function.call to support any level of nesting for qualified identifiers

jdrouhard:
Recent commit added support for distinguishing function.call but only supported up to 3 levels of nesting and each were hard coded.

This adds a query that uses `has-ancestor` to support any qualified function call if it detects a nesting level greater than 3 for namespace/type qualifiers.

### DIFF
--- a/queries/cpp/highlights.scm
+++ b/queries/cpp/highlights.scm
@@ -49,58 +49,93 @@
 (destructor_name
   (identifier) @method)
 
+; functions
 (function_declarator
-      declarator: (qualified_identifier
-        name: (identifier) @function))
+  (qualified_identifier
+    (identifier) @function))
 (function_declarator
-      declarator: (qualified_identifier
-        name: (qualified_identifier
-          name: (identifier) @function)))
+  (qualified_identifier
+    (qualified_identifier
+      (identifier) @function)))
 (function_declarator
-      declarator: (template_function
-        name: (identifier) @function))
+  (qualified_identifier
+    (qualified_identifier
+      (qualified_identifier
+        (identifier) @function))))
+((qualified_identifier
+  (qualified_identifier
+    (qualified_identifier
+      (qualified_identifier
+        (identifier) @function)))) @_parent
+  (#has-ancestor? @_parent function_declarator))
+
 (function_declarator
-      declarator: (template_method
-        name: (field_identifier) @method))
-((function_declarator
-      declarator: (qualified_identifier
-        name: (identifier) @constructor))
- (#lua-match? @constructor "^[A-Z]"))
+  (template_function
+    (identifier) @function))
 
 (operator_name) @function
 "operator" @function
 "static_assert" @function.builtin
 
 (call_expression
-  function: (qualified_identifier
-              name: (identifier) @function.call))
+  (qualified_identifier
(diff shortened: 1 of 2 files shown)


queries/cpp/highlights.scm
```diff
@@ -71,32 +71,13 @@
...
-                      name: (identifier) @function.call)))
-(call_expression
-  function:
-      (qualified_identifier
-        name: (qualified_identifier
-              name: (template_function
-                      name: (identifier) @function.call))))
+((qualified_identifier
+     (identifier) @function.call) @_parent
+ (#has-ancestor? @_parent call_expression))
```
theHamsta, line 76: Oh, sorry I forgot about this issue. Maybe we can revert that commit?

I would avoid has-ancestor. Getting the parent of a node is very costly and it will run on almost every node. The hardcoded pattern is definitely better. Maybe we can run has-ancestor only at the third or fourth level which will definitely be rarer

jdrouhard: Sounds good. Updated.

jdrouhard: Also refactored the query file a bit to group related queries together, and cleaned up a bit of the indenting and labelling of nodes. If you prefer the old way (explicit labels), I can go back, but I don't think they're terribly useful for cpp.

queries/cpp/highlights.scm
```diff
@@ -49,58 +49,93 @@
...
+  (qualified_identifier
+    (qualified_identifier
+      (qualified_identifier
+        (identifier) @function))))
+((qualified_identifier
+  (qualified_identifier
+    (qualified_identifier
+      (qualified_identifier
+        (identifier) @function)))) @_parent
+  (#has-ancestor? @_parent function_declarator))
```
lucario387, line 70: Hmm, doesn't this also run into the issue of needing a `has-ancestor?` to actually highlight properly?
There really is no beautiful way to avoid it then :')

jdrouhard: It'll only use `has-ancestor` if there's a node that matches 4 nested scopes for qualified identifiers.

jdrouhard:
Ping @theHamsta

lucario387 (approved):

queries/cpp/highlights.scm
theHamsta: Let's really hope that it is rare. If there are occurrences of `has-ancestors?` the predicate could be triggered pretty often.

theHamsta: If this distinction between `@function` and `@function.call` gets too complex we might want to revert to what it was before the distinction.

theHamsta:
@lucario387 you should now also be able to merge PRs. Are you using Matrix so that we could invite you to our nvim-treesitter org channel? Just ping us in the Matrix channel linked in the readme of this repo

lucario387:
I don't really know who and when to ping, is it okay to ping just you in the neovim treesitter channel @theHamsta?

theHamsta:
> I don't really know who and when to ping, is it okay to ping just you in the neovim treesitter channel @theHamsta?

yes, this is what I meant. I don't know your username on matrix.

PR #4646, opened 2023-04-12
refactor(indent): separate tracker for indent and dedent

lucario387:
Taking the inspiration from #4557, and from recents issues, this will test whether it is plausible to separate indent tracker with dedent tracker or not

Todo:
- [ ] Checking logic
- [ ] Re write queries (if needed)
  - [ ] C/C++ if-else: Multiline if else were incorrect (#4557)
  - [ ] Python else/elif/return/break dedent (#4554)
  - [ ] Dart indent
  - [ ] Lua func/table extra indent fix
  - [ ] Ecma method chaining with new logic
  - [ ] And some more :)
- [ ] Add tests, and remove XFAIL for some tests if possible

### DIFF
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -163,7 +163,8 @@ function M.get_indent(lnum)
   end
 
   -- tracks to ensure multiple indent levels are not applied for same line
-  local is_processed_by_row = {}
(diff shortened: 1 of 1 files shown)


lucario387:
The fact that it only fails ... 3 tests is concerning i feel

PR #4658, opened 2023-04-15
injections(nix): dynamic language injection via comments

figsoda:
continuation of #3902 since https://github.com/neovim/neovim/pull/21548 is in stable

requires https://github.com/neovim/neovim/pull/23015 to work correctly since `#gsub!` is bugged in neovim 0.9.0

this allows doing things like this to highlight code blocks of arbitrary languages with comments

```nix
/* lua */ ''
  print("Hello, world")
''
```

### DIFF
--- a/queries/nix/injections.scm
+++ b/queries/nix/injections.scm
@@ -1,6 +1,15 @@
 ((comment) @injection.content
  (#set! injection.language "comment"))
 
+((comment) @injection.language
+  . ; this is to make sure only adjacent comments are accounted for the injections
+  [
+    (string_expression (string_fragment) @injection.content)
+    (indented_string_expression (string_fragment) @injection.content)
+  ]
+  (#gsub! @injection.language "/%*%s*([%w%p]+)%s*%*/" "%1")
+  (#set! injection.combined))
+
 (apply_expression
   function: (_) @_func
   argument: [


amaanq:
It seems like the referenced bug fix got back ported, if so, anyone installing from that point onward will have the fix, right? If you could confirm that to be the case I can merge now..

amaanq (approved):

figsoda:
the backport isn't in a stable release yet, tested both on the release-0.9 branch and by applying the patch to 0.9.0 and it works as expected

figsoda:
Unfortunately this makes highlighting large files extremely slow, do you have any suggestions on how I can fix that?

amaanq:
> Unfortunately this makes highlighting large files extremely slow, do you have any suggestions on how I can fix that?

Wait for 0.10.0, Lewis should be adding a lot of performance improvements for that (incl. a potential async implementation of ts highlighting/injections)

FWIW, js also suffers with this with the comment to check if its eq to GraphQL then inject into the following template string.

lucario387:
Doesn't that also raise the base requirements to 0.9, even while having all of these compat modules for 0.8?

figsoda:
It does, I didn't know we are keeping nvim-treesitter compatible with 0.8

figsoda:
@lucario387 switched to `@injections.*` and rebased onto main since I'm hoping 0.10 to fix the performance issues. Untested since I don't have anything set up yet, and the main branch doesn't work with my neovim 0.9 or my config

runtime/queries/nix/injections.scm
```diff
@@ -1,6 +1,14 @@
 ((comment) @injection.content
  (#set! injection.language "comment"))
 
+((comment) @injection.language
+  [
+    (string_expression (string_fragment) @injection.content)
+    (indented_string_expression (string_fragment) @injection.content)
+  ]
+  (#gsub! @language "/%*%s*([%w%p]+)%s*%*/" "%1")
```
lucario387, line 9: `@language` -> `@injection.language`

clason:
@lucario387 can you take a look at (and test) this change?

lucario387:
> @lucario387 can you take a look at (and test) this change? 

I had taken a look and test used it before, and encountered some latency when I have multiple of these comment-injected content blocks, which is why I suggested the changes of using anchor. 

It's not as bad as before, but irl usage, probably not that bad? I don't use Nix so I cannot tell the usage rate of such config style

rnhmjoj:
I'm not sure what you did to this branch, the last readable diff is https://github.com/nvim-treesitter/nvim-treesitter/commit/3219a7ecc88d678c7d552ade4326d432c602bb5f.
Anyway, another useful dynamic injection you can now do is
```scheme
; Highlight writeText(Dir) "dir/fname.lang" '''blablabla'
(apply_expression
  function:
    (apply_expression
      function: (_) @_func
      argument: (string_expression (string_fragment) @language))
  argument: [(string_expression (string_fragment) @content)
             (indented_string_expression (string_fragment) @content)]
  (#match? @_func "(^|\\.)writeText(Dir)?$")
  (#gsub! @language ".*%.(.*)" "%1"))
@combined
```
@figsoda can you add it?

MobSenpai:
Very useful feature, if it's added.

lucario387 (approved):
Not too performant, but from the 20k LoC test file that @figsoda sent on matrix, the parse time doesn't go up by that much.

figsoda:
thanks for the rebases and testing, i forget about this PR :facepalm:

PR #4522, opened 2023-03-17
injections(html): various <script type>

bennypowers:
Replaces the `<script type>` parts from #2577 

image

<details><summary>test file</summary>

```html
<script>
  html`<p></p>`
  html(`<p></p>`)
  svg`<p></p>`
  svg(`<p></p>`)
</script>

<script type="module">
  html`<p></p>`
  html(`<p></p>`)
  svg`<p></p>`
  svg(`<p></p>`)
</script>

<input pattern="[0-9]+">
<input pattern=[0-9]+ type="tel">

<script type="application/json">{ "true": false }</script>
<script type="importmap">{ "true": false }</script>
<script type="text/markdown">## Hello *World*!</script>
<script type="application/graphql">query OK { dokey }</script>
<script type="javascript">function foo () {}</script>
<script type="module">function foo() {}</script>
<script type="application/ecmascript">function foo() {}</script>
<script type="text/ecmascript">function foo() {}</script>
<script type="text/javascript">function foo() {}</script>
```

</details>

PR #4659, opened 2023-04-15
feat(markdown): configured aliases for fenced code block languages

### LINKED ISSUE #2131: `markdown_fenced_languages` not used in Markdown
### Before reporting

- [X] I have updated my neovim version to latest _master_
- [X] I have updated my plugin to the latest version
- [X] I have run `:TSUpdate`
- [X] I have read the [troubleshooting section](https://github.com/nvim-treesitter/nvim-treesitter#troubleshooting)

### Describe the bug

With:
```vim
let g:markdown_fenced_languages = [
  \ 'bash=sh', 'javascript', 'js=javascript', 'json=javascript', 'typescript',
  \ 'ts=typescript', 'php', 'html', 'css', 'rust', 'sql']
```

` ```ts `

should enable code highlighting the same as:

` ```typescript `

but `ts` has no highlighting, as if it were just:

` ``` `

I can now delete the whole `markdown_fenced_languages` section. I always thought it was a waste of space, but it would be nice to alias `ts=typescript` by default as it seems common.

### To Reproduce

` ```ts `

should enable code highlighting the same as:

` ```typescript `

### Expected behavior

_No response_

```markdown
## Installation
  - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
  - OK: `node` found v16.8.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
    Version: cc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-4)

## Parser/Features H L F I J
  - markdown       ✓ . . . ✓ 
  - css            ✓ . ✓ ✓ ✓ 
  - javascript     ✓ ✓ ✓ ✓ ✓ 
  - typescript     ✓ ✓ ✓ ✓ ✓ 
  - c              ✓ ✓ ✓ ✓ ✓ 
  - json           ✓ ✓ ✓ ✓ . 
  - vim            ✓ ✓ . . ✓ 
  - jsdoc          ✓ . . . . 
  - rust           ✓ ✓ ✓ ✓ ✓ 
  - lua            ✓ ✓ ✓ ✓ ✓ 
  - bash           ✓ ✓ ✓ . ✓ 
  - jsonc          ✓ ✓ ✓ ✓ ✓ 
  - html           ✓ ✓ ✓ ✓ ✓ 
  - tsx            ✓ ✓ ✓ ✓ ✓ 

```

```text
NVIM v0.6.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc-11 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/runner/work/neovim/neovim/build/config -I/home/runner/work/neovim/neovim/src -I/home/runner/work/neovim/neovim/.deps/usr/include -I/usr/include -I/home/runner/work/neovim/neovim/build/src/nvim/auto -I/home/runner/work/neovim/neovim/build/include
Compiled by runner@fv-az65-618
```

### Additional context

_No response_

bennypowers:
Closes #2131  

This PR, in the same vein as #4522, hard codes configuration for certain language aliases, namely `sh => bash`, `js => javascript` and `ts => typescript`

before:

Screenshot from 2023-04-15 20-51-06

after:
Screenshot from 2023-04-15 20-51-12

### DIFF
--- a/lua/nvim-treesitter/query_predicates.lua
+++ b/lua/nvim-treesitter/query_predicates.lua
@@ -1,5 +1,30 @@
 local query = require "vim.treesitter.query"
 
+local html_script_type_languages = {
+  ["importmap"] = "json",
+  ["module"] = "javascript",
+  ["application/ecmascript"] = "javascript",
+  ["text/ecmascript"] = "javascript",
+}
+
+local injection_language_aliases = {
+  ex = "elixir",
+  fc = "func",
+  fir = "firrtl",
+  ha = "hare",
+  hs = "haskell",
+  js = "javascript",
+  kt = "kotlin",
+  ml = "ocaml",
+  pl = "perl",
+  py = "python",
+  ts = "typescript",
+  rs = "rust",
+  sh = "bash",
+  ungram = "ungrammar",
+  uxn = "uxntal",
+}
+
 local function error(str)
   vim.api.nvim_err_writeln(str)
 end
@@ -119,19 +144,17 @@ query.add_predicate("has-type?", function(match, _pattern, _bufnr, pred)
   return vim.tbl_contains(types, node:type())
 end)
 
-local html_script_type_languages = {
-  ["importmap"] = "json",
-  ["module"] = "javascript",
-  ["application/ecmascript"] = "javascript",
-  ["text/ecmascript"] = "javascript",
-}
-
----@param match string
----@param metadata table
+---@param match (TSNode|nil)[]
+---@param _ string
+---@param bufnr integer
+---@param pred string[]
 ---@return boolean|nil
-query.add_directive("set-lang-from-mimetype!", function(match, pattern, bufnr, predicate, metadata)
-  local capture_id = predicate[2]
+query.add_directive("set-lang-from-mimetype!", function(match, _, bufnr, pred, metadata)
+  local capture_id = pred[2]
   local node = match[capture_id]
+  if not node then
+    return
+  end
   local type_attr_value = vim.treesitter.get_node_text(node, bufnr)
   local configured = html_script_type_languages[type_attr_value]
   if configured then
@@ -139,6 +162,26 @@ query.add_directive("set-lang-from-mimetype!", function(match, pattern, bufnr, p
   else
     local parts = vim.split(type_attr_value, "/", {})
     metadata.language = parts[#parts]
+  end
+end)
+
+---@param match (TSNode|nil)[]
+---@param _ string
+---@param bufnr integer
+---@param pred string[]
+---@return boolean|nil
+query.add_directive("set-lang-from-alias!", function(match, _, bufnr, pred, metadata)
+  local capture_id = pred[2]
+  local node = match[capture_id]
+  if not node then
+    return
+  end
+  local injection_alias = vim.treesitter.get_node_text(node, bufnr)
+  local configured = injection_language_aliases[injection_alias]
+  if configured then
+    metadata.language = configured
+  else
+    metadata.language = injection_alias
   end
 end)
 
--- a/lua/nvim-treesitter/query_predicates.lua
+++ b/lua/nvim-treesitter/query_predicates.lua
@@ -15,7 +15,7 @@ local non_filetype_match_injection_language_aliases = {
   ts = "typescript",
 }
 
-local function get_language_name_from_fenced_code_block_tag(injection_alias)
+local function get_parser_from_markdown_info_string(injection_alias)
   local match = vim.filetype.match { filename = "a." .. injection_alias }
   return match or non_filetype_match_injection_language_aliases[injection_alias] or injection_alias
 end
@@ -165,14 +165,14 @@ end)
 ---@param bufnr integer
 ---@param pred string[]
 ---@return boolean|nil
-query.add_directive("set-lang-from-alias!", function(match, _, bufnr, pred, metadata)
+query.add_directive("set-lang-from-info-string!", function(match, _, bufnr, pred, metadata)
   local capture_id = pred[2]
   local node = match[capture_id]
   if not node then
     return
   end
   local injection_alias = vim.treesitter.get_node_text(node, bufnr)
-  metadata.language = get_language_name_from_fenced_code_block_tag(injection_alias)
+  metadata.language = get_parser_from_markdown_info_string(injection_alias)
 end)
 
 -- Just avoid some annoying warnings for this directive
--- a/lua/nvim-treesitter/query_predicates.lua
+++ b/lua/nvim-treesitter/query_predicates.lua
@@ -7,23 +7,18 @@ local html_script_type_languages = {
   ["text/ecmascript"] = "javascript",
 }
 
-local injection_language_aliases = {
+local non_filetype_match_injection_language_aliases = {
   ex = "elixir",
-  fc = "func",
-  fir = "firrtl",
-  ha = "hare",
-  hs = "haskell",
-  js = "javascript",
-  kt = "kotlin",
-  ml = "ocaml",
   pl = "perl",
-  py = "python",
-  ts = "typescript",
-  rs = "rust",
   sh = "bash",
-  ungram = "ungrammar",
   uxn = "uxntal",
+  ts = "typescript",
 }
+
+local function get_language_name_from_fenced_code_block_tag(injection_alias)
+  local match = vim.filetype.match { filename = "a." .. injection_alias }
+  return match or non_filetype_match_injection_language_aliases[injection_alias] or injection_alias
+end
 
 local function error(str)
   vim.api.nvim_err_writeln(str)
@@ -177,12 +172,7 @@ query.add_directive("set-lang-from-alias!", function(match, _, bufnr, pred, meta
     return
   end
   local injection_alias = vim.treesitter.get_node_text(node, bufnr)
-  local configured = injection_language_aliases[injection_alias]
-  if configured then
-    metadata.language = configured
-  else
-    metadata.language = injection_alias
-  end
+  metadata.language = get_language_name_from_fenced_code_block_tag(injection_alias)
 end)
 
 -- Just avoid some annoying warnings for this directive
--- a/queries/markdown/injections.scm
+++ b/queries/markdown/injections.scm
@@ -4,7 +4,7 @@
   (#not-match? @_lang "elm") ; prevent segfault when using elm parser
   (code_fence_content)
     @content
-    (#set-lang-from-info-string @_lang)
+    (#set-lang-from-info-string! @_lang)
     (#exclude_children! @content))
 
 ((html_block) @html)
--- a/queries/markdown/injections.scm
+++ b/queries/markdown/injections.scm
@@ -4,7 +4,7 @@
   (#not-match? @_lang "elm") ; prevent segfault when using elm parser
   (code_fence_content)
     @content
-    (#set-lang-from-alias! @_lang)
+    (#set-lang-from-info-string @_lang)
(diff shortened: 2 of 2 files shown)


amaanq:
This is something I've wanted before. The solution I had in mind was to use a grammars package.json, as the injection regex field contains this info on how to match it in markdown or any other similar style. Instead, could we rename the table to injection_language_tags? and I'll add as many as I can find from my grammars.

bennypowers:
Yeah that sounds good I'm not married to the local config in the file, but this did work for us in the past (#4522)

Perhaps we could merge this commit so that users are unblocked, and then go back and refactor the precise location of the config table.

In the mean time, I renamed the var as you suggested and moved the config to the top of the file.

amaanq:
Lmk your thoughts now, I fixed up the unnecessary commits & added missing fts

queries/markdown/injections.scm
```diff
@@ -1,8 +1,12 @@
...
   (info_string
-    (language) @language)
-  (#not-match? @language "elm")
-  (code_fence_content) @content (#exclude_children! @content))
+    (language) @_lang)
+  (#not-match? @_lang "elm") ; prevent segfault when using elm parser
+  (code_fence_content)
+    @content
+    (#set-lang-from-markdown-fence-tag! @_lang)
+    ; (#exclude_children! @content))
```
bennypowers, line 8: should either replace or remove

bennypowers:
yeah @amaanq looks good habibi

amaanq (approved):

lua/nvim-treesitter/query_predicates.lua
```diff
@@ -1,5 +1,30 @@
...
+  kt = "kotlin",
+  ml = "ocaml",
+  pl = "perl",
+  py = "python",
+  ts = "typescript",
+  rs = "rust",
+  sh = "bash",
+  ungram = "ungrammar",
+  uxn = "uxntal",
+}
```
theHamsta, line 26: shouldn't these be user configurable?

Also the interpretation of injection alias should be upstreamed at some time and maybe even resolve that abreviations by default. It's fine to have them as a experiment here. But when Neovim is trying to import our markdown queries they will need to support that directive in some way or have a more general solution

amaanq: I was thinking this could be a temporary helpful bandaid-like patch until Neovim implements something better.

bennypowers: I believe @amaanq 's suggestion was in the right direction, and that they should be configurable as well

So ideally, priority would be

0. Hard coded aliases
1. aliases derived from config in parser package
2. User aliases

I also think that each of those could probably be a subsequent PR

theHamsta: yeah, this can also happen in a follow-up PR. I was just thinking of giving the user a chance to manipulate/extend the hardcoded table.

LuisA92:
+1 for user configuration. I've been using [MyST](https://myst-parser.readthedocs.io/en/latest/syntax/code_and_apis.html) flavored markdown to write scientific documents. Some code blocks begin with the syntax ` ```{code-cell} <language>`. Currently, treesitter recognizes `{code-cell}` as the language.

lucario387:
Is there anyway to access vim.filetype table in 0.8 (and 0.9)? That way mappings like this aren't necessary I'd assume

amaanq:
> Is there anyway to access vim.filetype table in 0.8 (and 0.9)? That way mappings like this aren't necessary I'd assume

No, it's a local table, so not accessible w/ vim.filetype

ObserverOfTime:
This is also useful for rst.

theHamsta:
> Is there anyway to access vim.filetype table in 0.8 (and 0.9)? That way mappings like this aren't necessary I'd assume

Plenary also has a list of extensions https://github.com/nvim-lua/plenary.nvim/blob/master/data/plenary/filetypes/base.lua. It would be nice to hook  use/abuse as it already has a mechanism to register for users extensions. But extensions are not exactly the same as aliases (file extensions are only subset of possible aliases for a language). Maybe it should be an additional option lang.register instead? I don't have a clear opinion on that yet.

Is there already an Neovim upstream issue on this topic? It would be good to answer whether we should go with an experimental solution for some injections ahead and later upstream it or whether there could be an upstream solution from the beginning

lucario387:
`vim.filetype.match` is probably the closest we can get to extension->filetype mapping. And from this we can map it out again

From what I can see, we already have somewhat of a map from  filetype->parser in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/parsers.lua#L28. So adding more should not be much of an issue (i.e. if `injection.language` doesn't match to a parser name right away, do a `vim.filetype.match` to get ft -> getting parser name again)

bennypowers:
> https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/parsers.lua#L28.

Looking through that map it seems like it's not quite complete yet, and wouldn't serve here without some additional work, which I maintain is outside the scope of this PR. I will check out the `vim.filetype.match` suggestion though, which might be in scope here 

Perhaps after this merges, a subsequent PR (with regression tests!) should establish a central place for things like mimetype or alias to filetype. Perhaps this should be it's own module that exposes some `get_language_by_extention_or_mimetype`, or whatever.

theHamsta:
> > https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/parsers.lua#L28.
> 
> Looking through that map it seems like it's not quite complete yet, and wouldn't serve here without some additional work, which I maintain is outside the scope of this PR. I will check out the `vim.filetype.match` suggestion though, which might be in scope here
> 
> Perhaps after this merges, a subsequent PR (with regression tests!) should establish a central place for things like mimetype or alias to filetype. Perhaps this should be it's own module that exposes some `get_language_by_extention_or_mimetype`, or whatever.
> 
> edit: as you can see, vim.filetype.match gives incomplete results:
> 
> ```lua
> {
>   ex = {
>     aliased_to = "elixir"
>   },
>   fc = {
>     aliased_to = "func",
>     ft_match = "func"
>   },
>   fir = {
>     aliased_to = "firrtl",
>     ft_match = "firrtl"
>   },
>   ha = {
>     aliased_to = "hare",
>     ft_match = "hare"
>   },
>   hs = {
>     aliased_to = "haskell",
>     ft_match = "haskell"
>   },
>   js = {
>     aliased_to = "javascript",
>     ft_match = "javascript"
>   },
>   kt = {
>     aliased_to = "kotlin",
>     ft_match = "kotlin"
>   },
>   ml = {
>     aliased_to = "ocaml",
>     ft_match = "ocaml"
>   },
>   pl = {
>     aliased_to = "perl"
>   },
>   py = {
>     aliased_to = "python",
>     ft_match = "python"
>   },
>   rs = {
>     aliased_to = "rust",
>     ft_match = "rust"
>   },
>   sh = {
>     aliased_to = "bash"
>   },
> ```

When we merge this we will get a PR for every desired change in the initial list you provided.

With `vim.filetype.match(filename = "foo."..extension)` we would cover most of your list. Mimetypes if needed can be their own file if `filetype.match` doesn't give any result.

bennypowers:
@theHamsta is this more to your liking?

lua/nvim-treesitter/query_predicates.lua
```diff
@@ -1,5 +1,25 @@
...
+
+local non_filetype_match_injection_language_aliases = {
+  ex = "elixir",
+  pl = "perl",
+  sh = "bash",
+  uxn = "uxntal",
+  ts = "typescript",
+}
+
+local function get_language_name_from_fenced_code_block_tag(injection_alias)
```
lucario387, line 18: can this be named something else, like `get_parser_from_extension` or such?

bennypowers: it's not an extension, though. GFM spec calls it an ["info string"](https://github.github.com/gfm/#info-string)

bennypowers: how's that?

lucario387: Agreed, but it can be changed into other names, because usage of this can be, for example, in vue/svelte `lang="..."` attributes, where having this directive will be useful. From the viewpoint of using it in otherplaces, it's easier to follow if the functions related to this directive is named to reflect its usage

lucario387 (approved):

queries/markdown/injections.scm
```diff
@@ -1,8 +1,11 @@
 (fenced_code_block
   (info_string
-    (language) @language)
-  (#not-match? @language "elm")
-  (code_fence_content) @content (#exclude_children! @content))
+    (language) @_lang)
+  (#not-match? @_lang "elm") ; prevent segfault when using elm parser
```
lucario387, line 4: Sorry for not mentioning, but is there any other parser that causes segfault here? Aside from this, I agree with the change

amaanq: Don't think so, Nim might if we add it

lua/nvim-treesitter/query_predicates.lua
```diff
@@ -119,19 +139,17 @@ query.add_predicate("has-type?", function(match, _pattern, _bufnr, pred)
...
----@param match string
----@param metadata table
+---@param match (TSNode|nil)[]
+---@param _ string
+---@param bufnr integer
+---@param pred string[]
 ---@return boolean|nil
-query.add_directive("set-lang-from-mimetype!", function(match, pattern, bufnr, predicate, metadata)
-  local capture_id = predicate[2]
+query.add_directive("set-lang-from-mimetype!", function(match, _, bufnr, pred, metadata)
```
theHamsta, line 147: Maybe set language from alias?

theHamsta (approved):
Would be ok for me as a first draft though this needs to be configurable (not also by extensions at some time)

bennypowers:
agreed. Looking just now through lua/nvim-treesitter/configs.lua, it's not obvious to me where this config would live, so i think it's best to open a new issue and gain consensus before attempting to make this configurable

lucario387 (approved):

theHamsta:
Should we document this new directive somewhere? I think even the other predicates/directives of nvim-treesitter are missing documentation. Can also be left for follow-up

lucario387:
There will need to have a follow up PR for other directives, and that's about it

WhyNotHugo:
Thanks a lot for this patch!

AckslD:
Hope it's okay to ask here: is there a follow-up issue/PR for user configuration that I can subscribe to?

clason:
`:h vim.treesitter.language.register`

lua/nvim-treesitter/query_predicates.lua
mmbosschaert: Has the feature to allow users to manipulate or extend the hardcoded table been implemented yet?

PR #4660, opened 2023-04-15
perf(ecma): remove GraphQL comment injection

amaanq:


queries/ecma/injections.scm
```diff
@@ -78,10 +78,6 @@
 
 (regex_pattern) @regex
 
-((comment) @_gql_comment
-  (#eq? @_gql_comment "/* GraphQL */")
-  (template_string) @graphql)
-
 (((template_string) @_template_string
  (#match? @_template_string "^`#graphql")) @graphql)
```
lucario387, line 82: I prefer commenting it out more than removing it
Also, should the gql section right below it also be commented out?

amaanq: Yeah done (since we can readd it once 0.10 is out w/ ts improvements)

Also fixed a bug where the beginning ` was included in the range of gql injection, & used lua-match

lucario387 (approved):

PR #4661, opened 2023-04-16
fixed deprecated call query.get_query in lua/nvim-treesitter/query.lua

kyerpotts:
Deprecated call was resulting in error messages being displayed when sumneko_lua lsp is configured

### DIFF
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -105,7 +105,7 @@ do
   ---@param query_name string
   function M.get_query(lang, query_name)
     if cache[lang][query_name] == nil then
-      cache[lang][query_name] = ts.get_query(lang, query_name)
+      cache[lang][query_name] = ts.get(lang, query_name)
     end
 
     return cache[lang][query_name]


theHamsta (approved):

lua/nvim-treesitter/query.lua
```diff
@@ -105,7 +105,7 @@ do
   ---@param query_name string
   function M.get_query(lang, query_name)
     if cache[lang][query_name] == nil then
-      cache[lang][query_name] = ts.get_query(lang, query_name)
+      cache[lang][query_name] = ts.get(lang, query_name)
```
lucario387, line 108: if the file is correct, `ts` is `nvim-treesitter.compat`, and `ts.get` is not a thing, only `get_query` exists though, no?

lucario387, line 108: I believe this is not the cause for the deprecation notice but somewhere else. Can you have a look at where the problem might be?

lucario387 (changes requested):

guonaihong:
I need this feature。
image

lucario387:
> I need this feature。 image

This PR doesn't solve that issue for you, it will be because of some plugin that you use. Please check which plugin is causing that, not just reporting here

guonaihong:
image

I am pretty sure it is this project. I upgraded to nvim 9.0 and got these errors. I modified the local query.lua and the errors decreased.

guonaihong:
@lucario387 If I manually edit this line of code, the warning will disappear.

guonaihong:
image

lucario387:
> image

From the image, yours is `tsq.get_query`, while this PR is `ts.get_query`. Either you are running on an older version of nvim-treesitter, or something else.

lua/nvim-treesitter/query.lua
kyerpotts: That's correct, however if you look at the fully qualified call in

PR #4664, opened 2023-04-16
fix(ruby): bare_symbol should be `@symbol`

serhii-budnik:
Right now symbol is placed in the same category as heredoc which is not quite right. I as a user expect to see array of symbols highlighted in the same color as `simple_symbol`

|before|after|
|------|-----|
|Screenshot 2023-04-16 at 9 03 43 PM|Screenshot 2023-04-16 at 9 06 25 PM|

more examples
Screenshot 2023-04-16 at 9 12 01 PM

I don't see that Ruby has any tests right now but if I need to add smth let me know, appreciate any help on this

P.S.
here is an example from tree-sitter-ruby repo
https://github.com/tree-sitter/tree-sitter-ruby/blob/master/queries/highlights.scm#L104-L109

### DIFF
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -150,12 +150,12 @@
  ] @string
 
 [
- (bare_symbol)
  (heredoc_beginning)
  (heredoc_end)
  ] @constant
 
 [
+ (bare_symbol)
  (simple_symbol)
  (delimited_symbol)
  (hash_key_symbol)


amaanq:
Can you reword the commit to "fix(ruby): bare_symbol should be `@symbol`" or something similar

serhii-budnik:
commit message has been updated

amaanq (approved):

amaanq:
Thanks!

PR #4670, opened 2023-04-18
refactor(smali): refactor queries from upstream

amaanq:
:musical_note:

### DIFF
--- a/queries/smali/highlights.scm
+++ b/queries/smali/highlights.scm
@@ -5,7 +5,7 @@
 (primitive_type) @type.builtin
 
 ((class_identifier) @type.builtin
-  (#vim-match? @type.builtin "^L(android|com/android|dalvik|java)/"))
+  (#match? @type.builtin "^L(android|com/android|dalvik|java|kotlinx)/"))
 
 ; Methods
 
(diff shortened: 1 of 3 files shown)


queries/smali/highlights.scm
```diff
@@ -5,7 +5,7 @@
 (primitive_type) @type.builtin
 
 ((class_identifier) @type.builtin
-  (#vim-match? @type.builtin "^L(android|com/android|dalvik|java)/"))
+  (#match? @type.builtin "^L(android|com/android|dalvik|java|kotlinx)/"))
```
lucario387, line 8: can't this be any-of?

amaanq, line 8: no, ".super Ljava/lang/Object;"

queries/smali/highlights.scm
```diff
@@ -183,9 +177,6 @@
 
 (comment) @comment @spell
 
-(class_definition
-  (comment) @comment.documentation)
-
```
lucario387, line 188: So..., what is the alternative, or was it unnecessary from the beginning?

amaanq: Should not have been removed, oops

lucario387 (approved):
Seeing as you are the maintainer of smali too, I don't see why I shouldn't trust your changes :)

PR #4675, opened 2023-04-19
chore(injections)!: Move to upstream syntax

lucario387:
Since 0.9, `@lang` syntax is still available fallback but will soon be deprecated.
Because of that, new syntax should be adopted once 0.9 becomes the baseline requirements for nvim-treesitter

Todo:
- [ ] Update queries
  - [x] Odin
  - [x] Usd 
- [ ] Rebase + Update if necessary

Blocker:
- neovim/neovim#23409
- Some parsers, notably can be seen in HTML/Vue/JS will not work properly sometimes without `include-children`. The `<script></script>` range can be considered as jsx elements when there's no non-whitespace characters in the middle of it
  - For those cases, the injected region was zero-byte, leading to a top-level tree unexpectedly
-  ...

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -284,17 +284,29 @@ query.
 ### Injections
 
 Some captures are related to language injection (like markdown code blocks). They are used in `injections.scm`.
-You can directly use the name of the language that you want to inject (e.g. `@html` to inject html).
 
-If you want to dynamically detect the language (e.g. for Markdown blocks) use the `@language` to capture
-the node describing the language and `@content` to describe the injection region.
+If you want to dynamically detect the language (e.g. for Markdown blocks) use the `@injection.language` to capture
+the node describing the language and `@injection.content` to describe the injection region.
+
+For regions that don't have a corresponding `@injection.language`, you need to manually set the language 
+through `(#set injection.language "lang_name")`
 
 ```scheme
-@{lang}   ; e.g. @html to describe a html region
+@injection.language ; dynamic detection of the injection language (i.e. the text of the captured node describes the language)
+@injection.content  ; region for the dynamically detected language
+```
+To combine all matches of a pattern as one single block of content, add `(#set! injection.combined)` to such pattern
 
-@language ; dynamic detection of the injection language (i.e. the text of the captured node describes the language)
-@content  ; region for the dynamically detected language
-@combined ; combine all matches of a pattern as one single block of content
+For example, to inject javascript into HTML's `<script>` tag
+
+```html
+<script>someJsCode();</script>
+```
+
+```query
+(script_element
+  (raw_text) @injection.content
+  (#set! injection.language "javascript")) ; set the parser language for @injection.content region to javascript
 ```
 
 ### Indents
--- a/plugin/query_predicates.lua
+++ b/plugin/query_predicates.lua
@@ -38,10 +38,10 @@ query.add_directive('set-lang-from-mimetype!', function(match, _, bufnr, pred, m
   local type_attr_value = vim.treesitter.get_node_text(node, bufnr)
   local configured = mimetype_aliases[type_attr_value]
   if configured then
-    metadata.language = configured
+    metadata['injection.language'] = configured
   else
     local parts = vim.split(type_attr_value, '/', {})
-    metadata.language = parts[#parts]
+    metadata['injection.language'] = parts[#parts]
   end
 end)
 
@@ -67,7 +67,7 @@ query.add_directive('set-lang-from-info-string!', function(match, _, bufnr, pred
 
   local injection_alias = vim.treesitter.get_node_text(node, bufnr)
   local filetype = vim.filetype.match({ filename = 'a.' .. injection_alias })
-  metadata.language = filetype or injection_aliases[injection_alias] or injection_alias
+  metadata['injection.language'] = filetype or injection_aliases[injection_alias] or injection_alias
 end)
 
 query.add_directive('downcase!', function(match, _, bufnr, pred, metadata)
@@ -95,40 +95,6 @@ query.add_directive('downcase!', function(match, _, bufnr, pred, metadata)
   else
     metadata[key] = string.lower(text)
   end
-end)
-
----@param match (TSNode|nil)[]
----@param pred string[]
----@param metadata table
----@return boolean|nil
----TODO(clason): remove when switching to upstream injections
-query.add_directive('exclude_children!', function(match, _, _, pred, metadata)
-  local capture_id = pred[2]
-  local node = match[capture_id]
-  if not node then
-    return
-  end
-
-  local start_row, start_col, end_row, end_col = node:range()
-  local ranges = {}
-  for i = 0, node:named_child_count() - 1 do
-    local child = node:named_child(i) ---@type TSNode
-    local child_start_row, child_start_col, child_end_row, child_end_col = child:range()
-    if child_start_row > start_row or child_start_col > start_col then
-      table.insert(ranges, {
-        start_row,
-        start_col,
-        child_start_row,
-        child_start_col,
-      })
-    end
-    start_row = child_end_row
-    start_col = child_end_col
-  end
-  if end_row > start_row or end_col > start_col then
-    table.insert(ranges, { start_row, start_col, end_row, end_col })
-  end
-  metadata.content = ranges
 end)
 
 -- Trim blank lines from end of the region
--- a/runtime/queries/arduino/injections.scm
+++ b/runtime/queries/arduino/injections.scm
@@ -1,3 +1,5 @@
-(preproc_arg) @arduino
+((preproc_arg) @injection.content
+ (#set! injection.language "arduino"))
 
-(comment) @comment
+((comment) @injection.content 
+ (#set! injection.language "comment"))
--- a/runtime/queries/astro/injections.scm
+++ b/runtime/queries/astro/injections.scm
@@ -1,19 +1,23 @@
-; inherits: html
+; inherits: html_tags
 
-((frontmatter
-    (raw_text) @typescript))
+(frontmatter
+  (raw_text) @injection.content
+  (#set! injection.language "typescript"))
 
-((interpolation
-    (raw_text) @tsx))
+(interpolation
+  (raw_text) @injection.content
+  (#set! injection.language "tsx"))
 
-((script_element
-    (raw_text) @typescript))
+(script_element
+  (raw_text) @injection.content
+  (#set! injection.language "typescript"))
 
-((style_element
-   (start_tag
-     (attribute
-       (attribute_name) @_lang_attr
-       (quoted_attribute_value (attribute_value) @_lang_value)))
-   (raw_text) @scss)
+(style_element
+ (start_tag
+   (attribute
+     (attribute_name) @_lang_attr
+     (quoted_attribute_value (attribute_value) @_lang_value)))
+ (raw_text) @injection.content
  (#eq? @_lang_attr "lang")
- (#eq? @_lang_value "scss"))
+ (#eq? @_lang_value "scss")
+ (#set! injection.language "scss"))
--- a/runtime/queries/awk/injections.scm
+++ b/runtime/queries/awk/injections.scm
@@ -1,2 +1,5 @@
-(comment) @comment
-(regex) @regex
+((comment) @injection.content 
+ (#set! injection.language "comment"))
+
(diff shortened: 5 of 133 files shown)


clason:
We should also update https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#injections then.

queries/ecma/injections.scm
```diff
@@ -73,31 +82,33 @@
...
-   right: (string) @html
-            (#offset! @html 0 1 0 -1))
+          (#any-of? @_prop "outerHTML" "innerHTML"))
+  right: 
+    [
+      (template_string)
+      (string)
+    ] @injection.content
+    (#offset! @injection.content 0 1 0 -1)
+    (#set! injection.language "html"))
```
lucario387, line 114: Don't see why we can't merge these into one, so I did it in this PR as well

lucario387:
That's about it for now I think :thinking:

lucario387:
I'm just a bit in fear that somehow Nix is gonna cause issues, so cc @figsoda if you can help test the nix files :D

queries/haskell/injections.scm
```diff
@@ -2,11 +2,11 @@
...
- ((quasiquote_body) @content)
+ ((quoter) @injection.language)
+ ((quasiquote_body) @injection.content)
 )
 
-(comment) @comment
+((comment) @injection.content (#set! injection.language "comment"))
 
 ;; -----------------------------------------------------------------------------
 ;; shakespeare library
```
ObserverOfTime, line 12: You should also update the comment for consistency.

queries/html_tags/injections.scm
```diff
@@ -2,40 +2,39 @@
...
 (script_element
    (start_tag
       ((attribute
            (attribute_name) @_attr (#eq? @_attr "type")
            (quoted_attribute_value (attribute_value) @_type))))
-   (raw_text) @content (#set-lang-from-mimetype! @_type))
+   (raw_text) @injection.content (#set-lang-from-mimetype! @_type))
 
 ; <a style="/* css */">
 ((attribute
```
ObserverOfTime, line 40: These still use the old syntax.

queries/jq/injections.scm
```diff
@@ -28,4 +31,6 @@
              "sub"
              "gsub"))
   (args . (args
-    (query (string) @regex))))
+    (query 
+      (string) @regex
```
ObserverOfTime, line 35:
```suggestion
      (string) @injection.content
```

queries/jq/injections.scm
```diff
@@ -12,7 +12,10 @@
              "splits"
              "sub"
              "gsub"))
-  (args . (query (string) @regex)))
+  (args . 
+    (query 
+      (string) @regex
```
ObserverOfTime, line 17:
```suggestion
      (string) @injection.content
```

queries/pascal/injections.scm
```diff
@@ -1,4 +1,4 @@
-(comment) @comment
+((comment) @injection.content (#set! injection.language "comment"))
 ; There is no parser for assembly language yet. Add an injection here when we
 ; have a parser.
 ; (asmBody) @asm
```
ObserverOfTime, line 4:
```suggestion
; ((asmBody) @injection.content (#set! injection.language "asm"))
```

queries/scala/injections.scm
```diff
@@ -1 +1 @@
-(comment) @comment
+(comment) @injection.content
```
ObserverOfTime, line 1:
```suggestion
((comment) @injection.content (#set! injection.language "comment"))
```

queries/smali/injections.scm
```diff
@@ -1 +1 @@
-(comment) @comment
+(comment) @injection.content
```
ObserverOfTime, line 1:
```suggestion
((comment) @injection.content (#set! injection.language "comment"))
```

queries/sparql/injections.scm
```diff
@@ -1 +1 @@
-(comment) @comment
+(comment) @injection.content
```
ObserverOfTime, line 1:
```suggestion
((comment) @injection.content (#set! injection.language "comment"))
```

queries/sql/injections.scm
```diff
@@ -1,3 +1,3 @@
-(comment) @comment
-(marginalia) @comment
+(comment) @injection.content
+(marginalia) @injection.content
```
ObserverOfTime, line 2:
```suggestion
([(comment) (marginalia)] @injection.content
 (#set! injection.language "comment"))
```

queries/vue/injections.scm
```diff
@@ -5,62 +5,68 @@
...
-        (quoted_attribute_value (attribute_value) @_css)))
+        (quoted_attribute_value (attribute_value) @injection.language)))
+    (raw_text) @injection.content)
     (#eq? @_lang "lang")
-    (#eq? @_css "css")
-    (raw_text) @css))
+    (#any-of? @injection.language "css" "scss"))
 
-; TODO: When nvim-treesitter have postcss and less parser, use @language and @content instead
+; TODO: When nvim-treesitter have postcss and less parser, use @injection.language and @injection.content instead
```
ObserverOfTime, line 13:
```suggestion
; TODO: When nvim-treesitter has postcss and less parsers, use @injection.language and @injection.content instead
```

ObserverOfTime (changes requested):

queries/haskell/injections.scm
lucario387: Agreed
Was just :s -ing and did not look at the style that much.
I will update in following PRs

ObserverOfTime: I meant this one. It's not in the diff.
https://github.com/nvim-treesitter/nvim-treesitter/blob/8e0e12f3dc32539928da01051d6ebb2423b5cfd2/queries/haskell/injections.scm#L15-L20

lucario387:
> Just a heads up, you should look into amending to commits and fixing up commits for the future grin very useful for small changes to not become 12 commits

I will squash them, don't worry :)
Keeping it as it is so I can checkout to previous commit easier

lucario387:
Squash'd

ObserverOfTime (approved):

lucario387:
I just added the updated health check as well. This is probably going to be the only breaking change for nvim-treesitter once 0.9 is the base.

Will continue to rebase this in case injections are updated by other users

queries/lua/injections.scm
```diff
@@ -3,57 +3,65 @@
...
   (#any-of? @_vimcmd_identifier "vim.treesitter.query.set" "vim.treesitter.query.parse_query" "vim.treesitter.query.parse"))
 
-; vim.rcprequest(123, "nvim_exec_lua", "return vim.api.nvim_buf_get_lines(0, 0, -1, false)", false)
 ((function_call
   name: (_) @_vimcmd_identifier
-  arguments: (arguments . (_) . (string) @_method . (string content: _ @lua)))
+  arguments: (arguments . (_) . (string) @_method . (string content: _ @injection.content)))
   (#any-of? @_vimcmd_identifier "vim.rpcrequest" "vim.rpcnotify")
-  (#any-of? @_method "\"nvim_exec_lua\"" "'nvim_exec_lua'" "[[nvim_exec_lua]]"))
+  (#any-of? @_method "\"nvim_exec_lua\"" "'nvim_exec_lua'" "[[nvim_exec_lua]]")
```
ObserverOfTime, line 28: Can you apply my suggestion in https://github.com/nvim-treesitter/nvim-treesitter/pull/4703#discussion_r1175679536 while you're at it?

lucario387: Applied.

amaanq (approved):

lucario387:
@clason I believe since 0.9, `@combined` now has to be set through `(#set! injection.combined)` instead of `@injection.combined` ? Through testing php files it was not recognizing `@injection.combined` anymore.

clason:
Yes, that's correct, see `:h treesitter-language-injections` (which could be worded clearer to make explicit that these are directives).

lucario387:
@clason I've just removed `exclude_children!` through this PR, as noted by the `TODO`, and the PR is pointing to `main` instead of `master` now.

clason:
> https://github.com/neovim/neovim/issues/23409

Is not a blocker for this PR, as the `include-children` directive is required for proper injections in those cases anyway.

lucario387:
Well yes, by itself, neovim/neovim#23409 is not a blocker, but it shows an issue with not adding `include-children` to previously included injections. I can add back `include-children` to the fts that I know is not working properly (html/js/vue/astro/tsx...), but as the main branch won't become the default one for the foreseeable future, testing it will not be the priority of anyone. 

If the time comes for it tobe the default, and bugs show up, finding where to add/remove `include-children` will be more of a chore than this one imho.

clason:
Yes, we _need_ to add `include-children` where necessary. I hope some people will start testing this branch sooner rather than later, so this does become relevant ;)

I'll leave it up to you to decide which approach is the easiest, as you are the person most likely to do it :)

lucario387:
Anyhow, @clason I've just pushed changes for the queries that I've encountered issues with when it doesn't have `include-children`, those being html-families parsers/queries. Aside from that, I haven't tested the majority of them :')

clason:
You can check the injections in core (vimdoc needs them, I believe), also in the markdown PR -- just in case you haven't.

lucario387:
@clason For markdown, it's a bit complicated.

Looking at helix, all of their markdown queries has `include-unnamed-children`, even ones that we marked as `exclude-children`. So what should the way out be here?

For lua injections, the node doesn't exist before the content (unlike in html, where the reverse happens), so I couldn't perform a zero-byte injection there, so free of bugs :confetti_ball:

lucario387:
> I'll leave it up to you to decide which approach is the easiest, as you are the person most likely to do it :)

I'm fine with merging this, and then I'll try to find a way to call for testers, or some example codes to test it out.

lucario387:
So from the diffs, it suggests that `include-unnamed-children` will be in nightly sooner or later. I'll add it into this PR then @clason

clason:
Yes, that PR -- which is parked, pending folke's PR for actually using it -- will need to upstream a number of predicates and directives.

clason:
Just to be clear: In the queries, not in the implementation?

lucario387:
Yes, only in the queries.

clason:
Looking at the code, I think our `include-children` behaves like `include-unnamed-children` anyway (@lewis6991 ?), so let's try just with the former for now.

lucario387:
The only difference here is `markdown_inline` injection in `markdown`, where we explicitly excluded it before, while Helix used `include-unnamed`. I'll keep it without the `include` directive, to be consistent with ourselves first

clason:
> The only difference here is `markdown_inline` injection in `markdown`, where we explicitly excluded it before, while Helix used `include-unnamed`. I'll keep it without the `include` directive, to be consistent with ourselves first

Not sure you can do that, as the logic has changed in the "new injections"?

lucario387:
Okay without `include-children`, markdown breaks instantly with this error message:
```
 Vim(append):Error executing lua callback: ...l/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:260: Ranges 
can only be made from 6 element long tables or nodes.
```

So `include..` is necessary

lucario387:
I don't think so, as it's in nvim and not here.

With all injections in markdown marked with `(#set! include_children)`, but only

```scheme
([
  (inline)
  (pipe_table_cell)
 ] @injection.content
 (#set! injection.language "markdown_inline"))
```
not having it (before migrating to new syntax), it works fine. Even with it included, it still works ok, but it was excluded due to heavy performance issues before

clason:
So you have added these to _some_ queries, but have not checked all of them (understandably so)? How should we proceed then?

lucario387:
> So you have added these to some queries, but have not checked all of them (understandably so)? How should we proceed then?

I will spend the next hour adding it `include-children` to all queries, instead

Then I'll create another PR to slowly remove them

lucario387:
@clason Ok, just added back include-children to all but `markdown`, `vim`, `c`, and `gitcommit`. 
You can double check and then merge this now

clason:
Do we need that for the comment injections? Leaving those out would simplify the work quite a bit.

lucario387:
I'm not sure, tbh.

Most of the errors I met were with zero-byte injections, and there can't be zero-byte injections for `comment`, which makes .. sense ig

PR #4685, opened 2023-04-21
feat(c): highlight parenthesized function pointer decl identifiers as `@function`

amaanq:
Motivation - function pointers in struct fields

Before:
image

After:
image

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -130,7 +130,6 @@
 
 [
  (type_identifier)
- (sized_type_specifier)
  (type_descriptor)
 ] @type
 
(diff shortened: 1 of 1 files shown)


queries/c/highlights.scm
```diff
@@ -146,6 +145,8 @@
 
 (primitive_type) @type.builtin
 
+(sized_type_specifier _ @type.builtin type: _)
```
lucario387, line 148: What does this correspond to in C code? `int32_t` and the such or is it for the function pointers highlight

amaanq: unsigned/signed

lucario387: ah okay then

lucario387:
Consider me as approved of this PR but I don't know how it should be managed. I will let hamsta make the decision

theHamsta (approved):
I would think that the highlights are 100% needed but given that the function pointers are highlighted in other occasions as function it would make sense

PR #4686, opened 2023-04-21
perf: remove match where possible

amaanq:
Bound to have made a silly mistake somewhere :grin:

### DIFF
--- a/queries/dockerfile/highlights.scm
+++ b/queries/dockerfile/highlights.scm
@@ -55,9 +55,3 @@
 
 (expose_instruction
   (expose_port) @number)
-
-((stopsignal_instruction) @number
-  (#match? @number "[0-9][0-9]?$"))
-
-((stopsignal_instruction) @constant.builtin
-  (#match? @constant.builtin "SIG(ABRT|HUP|INT|KILL|QUIT|STOP|TERM|TSTP)$"))
--- a/queries/smali/highlights.scm
+++ b/queries/smali/highlights.scm
@@ -82,7 +82,10 @@
   (#lua-match? @keyword.return "^return"))
 
 ((opcode) @conditional
-  (#match? @conditional "^(if|cmp)"))
+  (#lua-match? @conditional "^if"))
+
+((opcode) @conditional
+  (#lua-match? @conditional "^cmp"))
 
 ((opcode) @exception
   (#lua-match? @exception "^throw"))
--- a/queries/starlark/highlights.scm
+++ b/queries/starlark/highlights.scm
@@ -29,7 +29,7 @@
 
 ((attribute
     attribute: (identifier) @field)
- (#match? @field "^([A-Z])@!.*$"))
+ (#lua-match? @field "^%u@!.*$"))
 
 ((identifier) @type.builtin
  (#any-of? @type.builtin
@@ -148,7 +148,7 @@
 (comment) @comment @spell
 
 ((module . (comment) @preproc)
-  (#match? @preproc "^#!/"))
+  (#lua-match? @preproc "^#!/"))
 
 (string) @string
 [
--- a/queries/t32/highlights.scm
+++ b/queries/t32/highlights.scm
@@ -81,7 +81,7 @@
 (
   (command_expression
     command: (identifier) @keyword.return)
-  (#match? @keyword.return "^[rR][eE][tT][uU][rR][nN]$")
+  (#lua-match? @keyword.return "^[rR][eE][tT][uU][rR][nN]$")
 )
 
 ; Subroutine calls
@@ -97,14 +97,14 @@
   (identifier) @constant)
 (
  (argument_list (identifier) @constant.builtin)
- (#match? @constant.builtin "^[%/][a-zA-Z][a-zA-Z0-9.]*$")
+ (#lua-match? @constant.builtin "^[%%/][%l%u][%l%u%d.]*$")
 )
 
 (
   (command_expression
     command: (identifier) @keyword
     arguments: (argument_list . (identifier) @label))
-  (#match? @keyword "^[gG][oO][tT][oO]$")
+  (#lua-match? @keyword "^[gG][oO][tT][oO]$")
 )
 (labeled_expression
   label: (identifier) @label)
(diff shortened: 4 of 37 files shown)


clason (approved):

queries/t32/highlights.scm
```diff
@@ -97,14 +97,14 @@
...
- (#match? @constant.builtin "^[%/][a-zA-Z][a-zA-Z0-9.]*$")
+ (#lua-match? @constant.builtin "^[%%/][%l%u][%l%u%d.]*$")
 )
 
 (
   (command_expression
     command: (identifier) @keyword
     arguments: (argument_list . (identifier) @label))
-  (#match? @keyword "^[gG][oO][tT][oO]$")
+  (#lua-match? @keyword "^[gG][oO][tT][oO]$")
```
ObserverOfTime, line 107:
```suggestion
  (#any-of? @keyword "goto" "GOTO")
```

ObserverOfTime (changes requested):

queries/t32/highlights.scm
clason: Won't match `Goto`.

queries/dockerfile/highlights.scm
```diff
@@ -60,4 +60,4 @@
   (#match? @number "[0-9][0-9]?$"))
 
 ((stopsignal_instruction) @constant.builtin
-  (#match? @constant.builtin "SIG(ABRT|HUP|INT|KILL|QUIT|STOP|TERM|TSTP)$"))
```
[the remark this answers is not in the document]
amaanq: Hmm, what do you mean? That the rule can never be captured to begin w/?

queries/smali/highlights.scm
```diff
@@ -82,7 +82,10 @@
   (#lua-match? @keyword.return "^return"))
 
 ((opcode) @conditional
-  (#match? @conditional "^(if|cmp)"))
+  (#lua-match? @conditional "^cmp"))
```
lucario387, line 85: Missing if here

queries/starlark/highlights.scm
```diff
@@ -29,7 +29,7 @@
 
 ((attribute
     attribute: (identifier) @field)
- (#match? @field "^([A-Z])@!.*$"))
+ (#lua-match? @field "^%d@!.*$"))
```
lucario387, line 32: is this %u or %d

queries/t32/highlights.scm
```diff
@@ -76,12 +76,12 @@
 (
   (command_expression
     command: (identifier) @keyword.return)
-  (#match? @keyword.return "^[eE][nN][dD]([dD][oO])?$")
+  (#any-of? @keyword.return "end" "END" "enddo" "ENDDO")
 )
 (
   (command_expression
     command: (identifier) @keyword.return)
-  (#match? @keyword.return "^[rR][eE][tT][uU][rR][nN]$")
```
lucario387, line 84: Do languages like this allow for things like `rEtUrN` or not?

ObserverOfTime: Presumably. But would anyone write it like that?

queries/dockerfile/highlights.scm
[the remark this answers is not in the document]
lucario387: Probably some case like this? 
https://github.com/nvim-treesitter/nvim-treesitter/pull/4430/files

Where the signal is something else under the node and not it exactly

queries/t32/highlights.scm
clason: If the grammar allows it, it should be highlighted correctly.

queries/dockerfile/highlights.scm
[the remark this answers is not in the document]
ObserverOfTime: I think when I last checked the signal didn't have its own node and only the `STOPSIGNAL` keyword did.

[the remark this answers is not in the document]
ObserverOfTime: I think when I last checked the signal didn't have its own node and only the `STOPSIGNAL` keyword did.

queries/t32/highlights.scm
```diff
@@ -97,14 +97,14 @@
...
- (#match? @constant.builtin "^[%/][a-zA-Z][a-zA-Z0-9.]*$")
+ (#lua-match? @constant.builtin "^[%%/][%l%u][%l%u%d.]*$")
 )
 
 (
   (command_expression
     command: (identifier) @keyword
     arguments: (argument_list . (identifier) @label))
-  (#match? @keyword "^[gG][oO][tT][oO]$")
+  (#lua-match? @keyword "[gG][oO][tT][oO]]")
```
ObserverOfTime, line 107:
```suggestion
  (#lua-match? @keyword "^[gG][oO][tT][oO]$")
```

queries/dockerfile/highlights.scm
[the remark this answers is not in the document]
ObserverOfTime: `. _` won't work either when there is no node.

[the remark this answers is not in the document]
ObserverOfTime: `. _` won't work either when there is no node.

queries/smali/highlights.scm
```diff
@@ -82,7 +82,13 @@
...
 
 ((opcode) @conditional
-  (#match? @conditional "^(if|cmp)"))
+  (#lua-match? @conditional "^if"))
+
+((opcode) @conditional
+  (#lua-match? @conditional "^cmp"))
+
+((opcode) @exception
+  (#lua-match? @exception "^throw"))
```
ObserverOfTime, line 91: Duplicate?
```suggestion
```

queries/dockerfile/highlights.scm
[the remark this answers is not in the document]
amaanq: Hmm does it not? I was thinking _ would capture anything after "STOPSIGNAL" here: https://github.com/camdencheek/tree-sitter-dockerfile/blob/8ee3a0f7587b2bd8c45c8cb7d28bd414604aec62/grammar.js#L155

queries/t32/highlights.scm
```diff
@@ -97,14 +97,14 @@
...
- (#match? @constant.builtin "^[%/][a-zA-Z][a-zA-Z0-9.]*$")
+ (#lua-match? @constant.builtin "^[%%/][%l%u][%l%u%d.]*$")
 )
 
 (
   (command_expression
     command: (identifier) @keyword
     arguments: (argument_list . (identifier) @label))
-  (#match? @keyword "^[gG][oO][tT][oO]$")
+  (#lua-match? @keyword "[gG][oO][tT][oO]")
```
ObserverOfTime, line 107: It will also match `foogotobar` without `^ $`.

amaanq: whoops

queries/dockerfile/highlights.scm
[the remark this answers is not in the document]
ObserverOfTime: > If not then I'll just remove it, the grammar should be updated to allow users to capture that node

Do the same for the capture above it.

queries/smali/highlights.scm
amaanq: aha, I copied & pasted w/o replacing it w/ if

amaanq:
Can't merge even though I resolved your suggested changes @ObserverOfTime, mind checking that out?

ObserverOfTime (approved):

PR #4692, opened 2023-04-22
fix(python,starlark): update field match pattern

### LINKED ISSUE #4691: Python: Missing `@field` highlight names after the latest update.
### Describe the bug

Pretty self-explanatory from the title. 

2023-04-22-074650_658x244_scrot

### To Reproduce

1. Create a python class.
2. Assign an attribute to `self`

### Expected behavior

`@field` highlight is assigned for python attributes.

```markdown

Installation ~
- OK `node` found v19.9.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
  Version: cc (GCC) 12.2.1 20230201

OS Info:
{
  machine = "x86_64",
  release = "6.2.12-arch1-1",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Thu, 20 Apr 2023 16:11:55 +0000"
} ~

  - awk                 ✓ . . . ✓
  - gitignore           ✓ . . . .
  - haskell             ✓ . ✓ . ✓
  - html                ✓ ✓ ✓ ✓ ✓
  - htmldjango          ✓ . ✓ ✓ ✓
  - java                ✓ ✓ . ✓ ✓
  - sql                 ✓ . . . ✓
  - swift               ✓ ✓ . . .

```

```text
NVIM v0.9.0
Build type: Release
LuaJIT 2.1.0-beta3

```

### Additional context

_No response_

amaanq:
Closes #4691 

@theHamsta I don't know what the original intention was, it works w/ match? but not lua-match? weirdly.


theHamsta:
I don't exactly remember, I think the idea was that fields shouldn't match constants.

amaanq:
I think lowercase makes sense, all caps are constants and PascalCase is (presumably) for types

lucario387:
It does raise a question. How many captures are broken because of #4686? Do we need regression tests for this (I think yes)

amaanq:
I think just this one, it was a funky one to begin w

lucario387:
Does this match snake_case or not?

lucario387:
This pattern does not match `self._a_b`, while the pattern before #4686 does match it

queries/python/highlights.scm
```diff
@@ -29,7 +29,7 @@
 
 ((attribute
     attribute: (identifier) @field)
- (#lua-match? @field "^%u@!.*$"))
+ (#lua-match? @field "^%l.*$"))
```
lucario387, line 32: This pattern does not match self._a_b, while the pattern before https://github.com/nvim-treesitter/nvim-treesitter/pull/4686 does match it

lucario387 (changes requested):
This pattern does not match self._a_b, while the pattern before https://github.com/nvim-treesitter/nvim-treesitter/pull/4686 does match it

queries/python/highlights.scm
amaanq: Good point, one sec

lucario387: Also, regression tests :). I know it's a chore but probably for the better in this case

amaanq: It's not something to really be touched again (only reason was for perf improvements), but if you'd like to add on tests feel free to

lucario387 (approved):

lucario387:
This still misses `self._A` from before #4686 but maybe it's fine, maybe it's not

amaanq:
> This still misses `self._A` from before #4686 but maybe it's fine, maybe it's not

I was debating over adding this, but such a thing isn't always a field, tut tut

lucario387:
The old pattern only doesn't match if it's `UPPER_SNAKE_CASE`. Imo not having some highlights for it seems worse (will be `@variable` if not match`, which feels wrong to me)

amaanq:
Fixed & added some tests

queries/python/highlights.scm
```diff
@@ -324,14 +324,14 @@
...
 ((class_definition
   body: (block
           (expression_statement
             (assignment
               left: (_
                      (identifier) @field)))))
- (#lua-match? @field "^%u@!.*$"))
+ (#lua-match? @field "^%l.*$"))
```
lucario387, line 334: Is there any reason why these aren't using the same patterns as the attributes?

amaanq: I haven't touched Python in so long, isn't @ for decorators and other similar stuff? I don't see how it could appear here

lucario387: I'm okay with both, but is it normal for python to declare private fields right inside block and not through `__init__`?

amaanq: Idk, not my language :^)

lucario387: Neither do I :^)

hamsta will do the call then

PR #4704, opened 2023-04-24
Update Smali

amaanq:


amaanq:
Smali tests worked fine for me locally but fail here, but the opposite is true for many tests that work here but fail for me locally :laughing: not a fun time

lucario387:
Seems like there's some indent queries changes but you forgot to push it?

amaanq:
> Seems like there's some indent queries changes but you forgot to push it?

No there aren't any

lucario387:
Missing `@indent.branch` for `.end *` stuffs.
`@indent.end` is for lines after a node
`@indent.branch` is for the line containing that node

tests/indent/smali_spec.lua
```diff
@@ -0,0 +1,25 @@
...
+    })
+  end)
+
+  describe("new line:", function()
+    run:new_line("field.smali", { on_line = 7, text = 'value1 = "test"', indent = 1 })
+    run:new_line("field.smali", { on_line = 10, text = "value2 = Lsome/enum;", indent = 2 })
+    run:new_line("array_and_switch.smali", { on_line = 43, text = "1 2 3 4 5 6 200", indent = 2 })
+    run:new_line("array_and_switch.smali", { on_line = 48, text = "Label10:", indent = 2 })
+    run:new_line("method.smali", { on_line = 7, text = ".registers 10", indent = 1 })
+    run:new_line("parameter.smali", { on_line = 20, text = ".annotation runtime Lsome/annotation;", indent = 2 })
```
lucario387, line 22: Running tests, this line is breaking but im not sure if the intended indentation here is 2 or 3

amaanq: it should be 2

lucario387: Curious why, because the line above it is `.annotation ....` too, which means it should be indented one more to accomodate for that

amaanq:
> Missing `@indent.branch` for `.end *` stuffs. `@indent.end` is for lines after a node `@indent.branch` is for the line containing that node
> 
> Should I push it? The change is only adding `@indent.branch`

Ahhh, sure yeah go ahead

lucario387:
This is the only one that fails. Maybe extra queries are necessary
image

amaanq:
I have no clue as to why, the logic should just capture the method and parameter node as 1 indent each

lucario387:
> I have no clue as to why, the logic should just capture the method and parameter node as 1 indent each

Not really, they aren't on the same line so iterating over them makes each of them add 1 to indent count

lucario387:
And you also have https://github.com/nvim-treesitter/nvim-treesitter/blob/12a21c94a8e125ac9b9a9c980bb10e87bc2beb2a/queries/smali/indents.scm#L2

Which adds one more

lucario387:
Ah okay understanding why, but wait a bit then

amaanq:
> And you also have
> 
> https://github.com/nvim-treesitter/nvim-treesitter/blob/12a21c94a8e125ac9b9a9c980bb10e87bc2beb2a/queries/smali/indents.scm#L2
> 
> Which adds one more

Inside annotation directives it should be indented one, yeah

lucario387:
> > And you also have
> > https://github.com/nvim-treesitter/nvim-treesitter/blob/12a21c94a8e125ac9b9a9c980bb10e87bc2beb2a/queries/smali/indents.scm#L2
> > 
> > Which adds one more
> 
> Inside annotation directives it should be indented one, yeah

So should it be 3 or 2 here, because it's inside annotation directives, parameters and methods, which is 3

amaanq:
> > > And you also have
> > > https://github.com/nvim-treesitter/nvim-treesitter/blob/12a21c94a8e125ac9b9a9c980bb10e87bc2beb2a/queries/smali/indents.scm#L2
> > > 
> > > Which adds one more
> > 
> > 
> > Inside annotation directives it should be indented one, yeah
> 
> So should it be 3 or 2 here, because it's inside annotation directives, parameters and methods, which is 3

Line 20 should be 2 as it's not inside the annotation directive

lucario387:
> Line 20 should be 2 as it's not inside the annotation directive

The `line 20` part is actually to `press "o" at line 20", then insert text`, so the actual line that the text is inserting is line 21

amaanq:
> > Line 20 should be 2 as it's not inside the annotation directive
> 
> The `line 20` part is actually to `press "o" at line 20", then insert text`, so the actual line that the text is inserting is line 21

Wow...I thought it meant the indentation level of that line (pressing o on 19), so yes 3 is right

amaanq:
Funny enough w/ 3 it fails locally (as do most tests...)

image

lucario387 (approved):

ObserverOfTime (approved):

lucario387:
Is it okay to merge now? :) 

Looks fine to me

PR #4707, opened 2023-04-25
injections(lua): tweak query to capture string content

MunifTanjim:
- Make it similar to other queries in the file
- No change needed once https://github.com/MunifTanjim/tree-sitter-lua/pull/20#discussion_r1176509444 is merged


lucario387:
I think this is safe to merge?

lucario387 (approved):

PR #4711, opened 2023-04-26
fix(lua): injection eq should be any-of

amaanq:


queries/lua/injections.scm
```diff
@@ -40,7 +40,7 @@
 (function_call
   (dot_index_expression
     field: (identifier) @_method
-    (#eq? @_method "gmatch" "gsub"))
+    (#any-of? @_method "gmatch" "gsub"))
```
lucario387, line 43: Can we merge this with the pattern right above it? the `"find "match"` one. They are the same from what I can see

lucario387: Ah no, dismiss my comments

lucario387 (approved):

PR #4721, opened 2023-04-27
feat: add luau

amaanq:
cc @leaxoy, @williamboman you might be interested as well :)

### DIFF
--- a/queries/luau/highlights.scm
+++ b/queries/luau/highlights.scm
@@ -1,0 +1,252 @@
+; Preproc
+
+(hash_bang_line) @preproc
+
+;; Keywords
+
+"return" @keyword.return
+
+[
+ "local"
+ "type"
+ "export"
+] @keyword
+
+(do_statement
+[
+  "do"
+  "end"
+] @keyword)
+
+(while_statement
+[
+  "while"
+  "do"
+  "end"
+] @repeat)
+
+(repeat_statement
+[
+  "repeat"
+  "until"
+] @repeat)
+
+[
+  (break_statement)
+  (continue_statement)
+] @repeat
+
+(if_statement
+[
+  "if"
+  "elseif"
+  "else"
+  "then"
+  "end"
+] @conditional)
+
+(elseif_statement
+[
+  "elseif"
+  "then"
+  "end"
+] @conditional)
+
+(else_statement
+[
+  "else"
+  "end"
+] @conditional)
+
+(for_statement
+[
+  "for"
+  "do"
+  "end"
+] @repeat)
+
+(function_declaration
+[
+  "function"
+  "end"
+] @keyword.function)
+
+(function_definition
+[
+  "function"
+  "end"
+] @keyword.function)
+
+;; Operators
+
+[
+  "and"
+  "not"
+  "or"
+  "in"
+  "typeof"
+] @keyword.operator
+
+[
+  "+"
+  "-"
+  "*"
+  "/"
+  "%"
+  "^"
+  "#"
+  "=="
+  "~="
+  "<="
+  ">="
+  "<"
+  ">"
+  "="
+  "&"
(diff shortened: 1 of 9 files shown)


amaanq:
It extends off the Lua one, and correctness

queries/luau/indents.scm
```diff
@@ -0,0 +1,5 @@
+; inherits: lua
+
+[
+  (object_type)
```
lucario387, line 4: Is there any other case where types should be indented? jfc

amaanq: maybe array types but I severely doubt anyone putting that on a newline

{
  [number]: T
}

lucario387: Hmm, okay then

lucario387 (approved):
Lets wait for either leaxoy or williamboman to have a look before merging. Other than that I'm okay with the folds and indent changes

ObserverOfTime (commented):
Why are the highlights, injections, and locals not inheriting from lua?

amaanq:
> Why are the highlights, injections, and locals not inheriting from lua?

no labels, gotos, parameters are different, etc

queries/luau/highlights.scm
```diff
@@ -0,0 +1,246 @@
...
+
+; Constants
+
+((identifier) @constant
+ (#lua-match? @constant "^[A-Z][A-Z_0-9]+$"))
+
+; Builtins
+
+((identifier) @variable.builtin
+ (#any-of? @variable.builtin "_G" "_VERSION" "debug" "io" "jit" "math" "os" "package" "self" "string" "table" "utf8"))
```
ObserverOfTime, line 147: Most of these are closer to namespaces rather than variables.
Only "self" is a variable and "_VERSION" is more of a constant.

amaanq: I took it from lua (granted it's not the "same" language though)

xactlyblue:
Excited to see this implemented, will most certainly be extremely useful.

queries/luau/highlights.scm
ObserverOfTime: They are the same in lua.

amaanq: yeah I can agree w changing _VERSION, but do you think the others besides self should be `@namespace` instead? (for lua too)

ObserverOfTime: Maybe?

ObserverOfTime (changes requested):

ObserverOfTime (approved):

leaxoy:
LGTM! Thanks!

xactlyblue:
Awesome, thanks!

xactlyblue:
I believe newer versions of Luau support custom type definitions, is that planned to be added in the future? An example of this could be:

```luau
export type Repository = {
    name: string,
    createdBy: string
}
```

I haven't used the new type definitions feature too much so forgive me if I made a mistake 😅.

amaanq:
Everything that was documented is supported, at least try the parser itself before asking, and if it isn't a bug should be filed at the parser not here then

xactlyblue:
I apologize, I guess it could've just been a setup mistake on my end but it gave me normal Lua highlighting (the file ended in .luau) and since my file doesn't use much of the luau-specific features aside from custom type definitions I just assumed it was not yet implemented since this is a relatively new feature.

amaanq:
Ensure ft is set to luau, here's what it looks like for me:

image

xactlyblue:
You're completely right, I apologize! I just tried forcing the buffer to be a luau file-type with `:set filetype=luau` and it works perfectly now. Totally my fault, sorry! Thanks for the help!

amaanq:
won't need it soon - https://github.com/vim/vim/pull/12317

PR #4722, opened 2023-04-27
highlights(comment): fix nospell precedence

MunifTanjim:
**Before**

image
image

**After**

image
image

### DIFF
--- a/queries/comment/highlights.scm
+++ b/queries/comment/highlights.scm
@@ -1,39 +1,39 @@
 (_) @spell
 
 ((tag
-  (name) @text.todo @nospell
+  (name) @nospell @text.todo
   ("(" @punctuation.bracket (user) @constant ")" @punctuation.bracket)?
   ":" @punctuation.delimiter)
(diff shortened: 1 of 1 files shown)


queries/comment/highlights.scm
MunifTanjim, line 1: There are a lots of palces where `@spell` or `@nospell` is applied after some other hl_group.

Is there a way to decrease priority of `@spell` and `@nospell` instead of fixing all the highlight queries?

clason: That should be fixed in Neovim; `@spell` queries should never affect highlighting, so this is a bug.

MunifTanjim: Any pointers on how to approach the fix?

clason: Hmm, no, sorry. We had other issues with treesitter and spelling -- best open an issue a d tag Lewis.

MunifTanjim:
Will open a PR on Neovim.

PR #4725, opened 2023-04-28
feat: add ocaml indents

lucario387:
Testing with code from:

- https://www.cs.umd.edu/class/spring2016/cmsc330/lectures/examples.pdf
- https://v2.ocaml.org/learn/taste.html
- https://v2.ocaml.org/manual

Edit: Oh my, there's a LOT of ERROR nodes, which isn't handled

### DIFF
--- a/queries/ocaml/indents.scm
+++ b/queries/ocaml/indents.scm
@@ -1,0 +1,69 @@
+[ 
+  (let_binding)                 ; let = ...
+  (type_binding)
+
+  (external)
+  (record_declaration)          ; { foo: bar ... }
+  (structure)                   ; struct ... end
+  (signature)                   ; sig ... end
+  (value_specification)         ; val x: bar
+
+  (do_clause)                   ; do ... done
+  (match_case)
+
+  (field_expression)
+  (application_expression)
+  (parenthesized_expression)    ; ( ... )
+  (record_expression)           ; { foo = bar ... }
+  (list_expression)             ; [...;...]
+  (try_expression)              ; try ... with
+] @indent.begin
+
+; handle if-else if-else case
+(if_expression
+  condition: (_) @indent.begin)
+(then_clause) @indent.begin
+((else_clause 
+   (_) @_not_if) @indent.begin
+ (#not-has-type? @_not_if if_expression))
+
+((parameter) @indent.begin
+ (#set! indent.start_at_same_line))
+
+(_ (ERROR) @indent.begin
+ (#set! indent.start_at_same_line))
+
+(ERROR "|" @indent.begin
+ (#set! indent.start_at_same_line))
+
+
+(try_expression
+  "with" @indent.branch
+  [
+    (match_case) @indent.dedent
+    (match_case 
+      [
+        (parenthesized_expression) 
+        (list_expression)
+      ] @indent.dedent)
+  ]
+)
+
+[
+  "}"
+] @indent.branch @indent.end
+
+
+(list_expression "]" @indent.branch)
+(parenthesized_expression ")" @indent.branch)
+
+";;" @indent.end
+
+(do_clause "done" @indent.end @indent.branch) 
+(structure "end" @indent.end @indent.branch)
+(signature "end" @indent.end @indent.branch)
+
+[
+  (string)
+  (comment) 
+] @indent.auto
--- a/queries/ocaml_interface/indents.scm
+++ b/queries/ocaml_interface/indents.scm
@@ -1,0 +1,1 @@
+; inherits: ocaml


amaanq:
neat!

lucario387:
Would be nice if someone who uses ocaml regularly can have a look at this :thinking: 

I'm not sure what should be the codestyle we should follow

lucario387:
@undu Sorry for pinging, but can you have a look at this PR?.

undu:
Thanks for opening the PR!

I'll try to review this this long weekend, I will probably not have time to do it before Monday, however

lucario387:
> Thanks for opening the PR!
> 
> I'll try to review this this long weekend, I will probably not have time to do it before Monday, however

Tyvm for supporting :D. I will be gladly accepting any suggestions from you as I believe it's better to follow the maintainer's intentions here, I am not familiar enough with ocaml to know the formatting convention

undu:
I'm trying it on some files, the results are not quite there yet:
- Expressions in the if and else branches are not being indented
```ocaml
if yes then
yes
else if maybe then
maybe
else
no
in
```
- Nested expressions sometimes do not add an indentation level
```ocaml
let _, value =
List.fold_left
(fun ...
 )
[]
[]
```
- When a record type is defined with one line per field, the fields are not indented
```ocaml
{
record with
field
}
```
- When a record field has a definition in a different line, it's not indented:
```ocaml
{
field=
value
}
```
- An indentation level is added when a list separator is found, not sure how it's done in other languages, but in ocaml code, those separators maintain the same level of indentation as the start of list `[`
```ocaml
[
  (a, tuple)
  ; (another, tuple)
  ]
```
- When the parameters in a function definition overflow into a new line normal indentation (2 spaces) is used. I think it'd be better to use 4, something similar like in Python.
```ocaml
let function has a lot of parameters that it
  overflows =
  let yes = yes in
  yes
```
- In one case (end of gencert.lib) at some point the indentation disappears when closing a function inside a parameter value:
```ocaml
  Option.fold
    ~none:(Ok yes)
    ~some:(fun no ->
        do something no >>= fun _ ->
Ok no
)
>>= fun continue ->
continue without
|> indentation
```
- Taking the start of function as the starting point for indentation can sometimes make the code squashed to the right IMO, the formatter I use uses a fixed, 8-space indentation, is something like that possible with this? A normal indentation might be too low.
```ocaml
value
|> Process_with_a_function_in_a_contrived_way (fun param ->
                                               let body = f param in
                                               body
                                               )
```
All of these can be seen in the following files, which are processed by another formatter so it's easy to see the difference using git diff:
- https://github.com/xapi-project/xen-api/blob/master/ocaml/gencert/lib.ml
- https://github.com/xapi-project/xen-api/blob/master/ocaml/tests/binpack_test.ml
- https://github.com/xapi-project/xen-api/blob/master/ocaml/tests/test_features.ml

Note that I don't expect the indentation levels to exactly match between formatters. Ocaml is a whitespace-insensitive language so any indentation is OK as long as it doesn't lead to confusion.

Please feel free to use more files in the repository to test the indentation, I've seen that trying the longer files seems to freeze neovim, so I wouldn't recommend using them

lucario387:
> I'm trying it on some files, the results are not quite there yet:
> 
> - Expressions in the if and else branches are not being indented
> 
> ```ocaml
> 
> if yes then
> 
> yes
> 
> else if maybe then
> 
> maybe
> 
> else
> 
> no
> 
> in
> 
> ```
> 
> - Nested expressions sometimes do not add an indentation level
> 
> ```ocaml
> 
> let _, value =
> 
> List.fold_left
> 
> (fun ...
> 
>  )
> 
> []
> 
> []
> 
> ```
> 
> - When a record type is defined with one line per field, the fields are not indented
> 
> ```ocaml
> 
> {
> 
> record with
> 
> field
> 
> }
> 
> ```
> 
> - When a record field has a definition in a different line, it's not indented:
> 
> ```ocaml
> 
> {
> 
> field=
> 
> value
> 
> }
> 
> ```
> 
> - An indentation level is added when a list separator is found, not sure how it's done in other languages, but in ocaml code, those separators maintain the same level of indentation as the start of list `[`
> 
> ```ocaml
> 
> [
> 
>   (a, tuple)
> 
>   ; (another, tuple)
> 
>   ]
> 
> ```
> 
> - When the parameters in a function definition overflow into a new line normal indentation (2 spaces) is used. I think it'd be better to use 4, something similar like in Python.
> 
> ```ocaml
> 
> let function has a lot of parameters that it
> 
>   overflows =
> 
>   let yes = yes in
> 
>   yes
> 
> ```
> 
> - In one case (end of gencert.lib) at some point the indentation disappears when closing a function inside a parameter value:
> 
> ```ocaml
> 
>   Option.fold
> 
>     ~none:(Ok yes)
> 
>     ~some:(fun no ->
> 
>         do something no >>= fun _ ->
> 
> Ok no
> 
> )
> 
> >>= fun continue ->
> 
> continue without
> 
> |> indentation
> 
> ```
> 
> - Taking the start of function as the starting point for indentation can sometimes make the code squashed to the right IMO, the formatter I use uses a fixed, 8-space indentation, is something like that possible with this? A normal indentation might be too low.
> 
> ```ocaml
> 
> value
> 
> |> Process_with_a_function_in_a_contrived_way (fun param ->
> 
>                                                let body = f param in
> 
>                                                body
> 
>                                                )
> 
> ```
> 
> All of these can be seen in the following files, which are processed by another formatter so it's easy to see the difference using git diff:
> 
> - https://github.com/xapi-project/xen-api/blob/master/ocaml/gencert/lib.ml
> 
> - https://github.com/xapi-project/xen-api/blob/master/ocaml/tests/binpack_test.ml
> 
> - https://github.com/xapi-project/xen-api/blob/master/ocaml/tests/test_features.ml
> 
> 
> 
> Note that I don't expect the indentation levels to exactly match between formatters. Ocaml is a whitespace-insensitive language so any indentation is OK as long as it doesn't lead to confusion.
> 
> 
> 
> Please feel free to use more files in the repository to test the indentation, I've seen that trying the longer files seems to freeze neovim, so I wouldn't recommend using them

Thank you very much for the extremely useful information. 

I'll try to apply the information given to match the info you want (except for indent 4 on very long let_binding, as it requires a indent logic rewrite)

lucario387:
@undu for the diffs between this and the last commit you tried:

- Expressions in the if and else branches are not being indented
  - I was not marking them for indenting, have added
- Nested expressions sometimes do not add indentation level
  - I think I've dealt with this, please let me know if I have missed some
- Record type not indented, Multi-line record field not indented
  - Similar to if-else, I was not marking this. Should be resolved now
- `]` not dedented.
  - Should be properly marked now
- `gencert.lib` fun not marked
  - I was not expecting `fun ...` to have a different node to `function ...` :D. Should be properly marked now
- indenting based on start of func:
  - This is reverted? to basing on normal indentation. Setting to some fixed level is not yet supported.

Unresolved:
- Different indents for arguments: I will try, but this seems impossible as of now

undu:
Found some others:
- Childs of while block are not being indented, the `done` that closes the block is being dedented too much:
```ocaml
  while true do
  let "_ = compute yes in
  ()
done
```

- when definig a value in an interface, the parameters are not indented if they start in a new line:
```ocaml
module Something : sig
  val a_func  :
  param:string
  -> Path.t
  -> unit
```

- Type definitions inside module signatures are not indented:
```ocaml
module Module : sig
type t = private string
end
```

- Comments documenting a value in an interface, which appear at the following line are dedented instead of keeping the same indentation level, and force the `end` to close the module signature to be indented instead of dedented:
```ocaml
module Module : sig
  val value : string
(** [value] is very important *)
  end
```

- When a type definition of a record, the first field is not indented at all, the following fields are indented 9 spaces:
```ocaml
type runstateinfo = {
state: int32
         ; missed_changes: int32
         ; state_entry_time: int64
         ; time0: int64
         ; time1: int64
         ; time2: int64
         ; time3: int64
         ; time4: int64
         ; time5: int64
}
```

- This would be nice if it was changed, but not really necessary, sometimes when adding parentheses there's no need to add an indent level, for example:
```ocaml
try
  ()
with _ -> (
    (Command.run cli) ;
    Command.run cli_2
  )
```

examples found in:
- https://github.com/xapi-project/xen-api/blob/master/ocaml/xsh/xsh.ml
- https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/lib/xenops_sandbox.ml
- https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/lib/xenops_sandbox.mli
- https://github.com/xapi-project/xen-api/blob/master/ocaml/xenopsd/xc/xenctrlext.mli

lucario387:
@undu try no.4 :smile:

- `done` dedented too much
  - I marked the wrong node as dedenting, it should be better now
- Parameters aren't indented
  - I cannot figure out how to make it reactively indent, but it should work with `=` motions now
- module not indented properly
  - Again, I cannot guarantee on reactively indenting, but `=` motions should work
- Type definition not indented properly
  - I believe this has been fixed as of this commit, or at least it's fixed with `=` motions
- `try ... with _ -> (...)`, the section inside `()` should not be indented once more
  - Done.

lucario387:
Also, as a note, this is only adding support for `.ml` files, I do know that `.mli` is for interfaces only but haven't looked at what should be put inside a `.mli` file, aside from modules signatures

undu:
Note that ocaml tree-sitter grammar includes the ocamlinterface one, since module interfaces can be defined in normal ocaml modules.

I still see the issues with the while loop in xsh.ml and the record type definition issue with xenctrlext.mli

lucario387:
> Note that ocaml tree-sitter grammar includes the ocamlinterface one, since module interfaces can be defined in normal ocaml modules.
> 
> I still see the issues with the while loop in xsh.ml and the record type definition issue with xenctrlext.mli

```ml
module Xenforeignmemory : sig
  type handle

  type mapping =
    (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t

  type prot = {read: bool; write: bool; exec: bool}

  val acquire : unit -> handle

  val release : handle -> unit

  val with_mapping :
    handle
    -> domid
    -> prot
    -> Int64.t list
    -> ?on_unmap_failure:(string -> unit)
    -> (mapping -> unit)
    -> unit
  (** [with_mapping handle domid prot pages ~on_unmap_failure f] maps the
      [pages] in the domain [domid] to the current domain with the flags
      [prot] and allows to use this memory as a bigarray inside the scope of
      [f]. [on_memory_failure] may be defined to capture an unmap failure
      after [f] goes out of scope. This is useful for logging, for example. *)
end

```

This is how it looks like as of the latest commit on my end, after pressing `gg=G`, for the xenctrl file.

For xsh.ml files, the while loop looks as below
```ml
    while true do
      let r =
        (if can_read a' then [ain] else [])
        @ if can_read b' then [Unixfd.(!bin)] else []
      in
      let w =
        (if can_write a' then [bout] else [])
        @ if can_write b' then [aout] else []
      in
      let r, w, _ = Unix.select r w [] (-1.0) in
      (* Do the writing before the reading *)
      List.iter
        (fun fd -> if aout = fd then write_from b' a' else write_from a' b')
        w ;
      List.iter (fun fd -> if ain = fd then read_into a' else read_into b') r
    done
```
I'm not sure what is wrong about this, so it'd be nice to have more info

undu:
I didn't update the branch properly, sorry about that, now it looks good!

undu (approved):

lucario387:
Tyvm for the support :D

I did not expect ocaml queries to have these many cases.

amaanq (approved):

nullchilly (approved):

PR #4729, opened 2023-04-28
fix(ecma): enable highlight for @keyword.export

yavorski:
Enable highlight for `export` keyword in `javascript` filetype

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -341,6 +341,10 @@
   "finally"
 ] @exception
 
+[
+  "export"
+] @keyword.export
+
 (export_statement
   "default" @keyword)
 (switch_default


amaanq:
It's already under keyword, no point in this

yavorski:
Hi, noticed that setting highlight for `@keyword.export` works correctly for `typescript` filetype, but it does not for `javascript`.
In `js` and `mjs` the `export` keyword is recognized just as `keyword` which is limitating for having consistent highlight for `import` and `export` highlights in the same files. This pull request fix this.

amaanq:
It's `@keyword` in typescript too

yavorski:
@amaanq yes it is, but setting the highlight for `@keyword.export` works in `typescript` but it does not in `javascript`.

amaanq:
I have no clue as to what you're saying.

yavorski:
@amaanq  This will enable you to set different color for the `export` keyword which is different from just the default `keyword` color. This is already possible and working for `.ts` files, but it is **not** working/possible for `.js` and `.mjs` files.

amaanq:
Yeah no such distinction exists here, you might be talking about an LSPs highlights, either way you can extend queries yourself in your own config.

yavorski:
This distinction already exists for `typescript` language and I am not sure why it does **not** exists for `javascript`.
You can clearly see from the picture with `TSHighlightCapturesUnderCursor` that `export` is defined for `ts` but for not `js`.

ts-js-export

amaanq:
probably from your theme, last time I'm saying this but you (and plugins) can extend queries, that's what's happening here, a theme or plugin of yours is doing this, NOT us

yavorski:
Kewl, The issue was indeed in `after/queries` extended query. Thanks for pointing it out!

PR #4736, opened 2023-04-30
feat: add py-script injections for html

lucario387:
Too niche, maybe?

I put this in html instead of `html_tags` as there can definitely be some cases where people name their components `py-script`, but this can't happen in html.

Infos are from

- https://docs.pyscript.net/latest/tutorials/index.html
- https://github.com/pyscript/pyscript/

### DIFF
--- a/queries/html/injections.scm
+++ b/queries/html/injections.scm
@@ -1,1 +1,24 @@
 ; inherits html_tags
+
+(element
+  (start_tag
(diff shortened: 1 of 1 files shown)


amaanq:
eqs are cheap, and nothing is too niche (we have cooklang, cpon, and uxntal :laughing:)

amaanq (approved):

PR #4741, opened 2023-05-01
feat(go): highlight constructors

amaanq:
Priority is to beat the lsp, removed _ as that's not a constant but more a discard, could be `@comment` like Gleam but `@constant` is just wrong

### DIFF
--- a/queries/go/highlights.scm
+++ b/queries/go/highlights.scm
@@ -15,9 +15,6 @@
 
 (label_name) @label
 
-((identifier) @constant
- (#eq? @constant "_"))
-
 (const_spec
   name: (identifier) @constant)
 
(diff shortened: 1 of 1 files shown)


lucario387:
Im not sure if that's nice or not

queries/go/highlights.scm
```diff
@@ -41,6 +38,16 @@
 (method_spec 
   name: (field_identifier) @method) 
 
+; Constructors
+
+((call_expression (identifier) @constructor)
+  (#lua-match? @constructor "^[nN]ew.+$")
+  (#set! "priority" 150))
```
lucario387, line 45: Wait why is `.+` necessary here? Shouldn't constructor be `new` and `New` only?

ObserverOfTime:
I'm not sure about this. These are still just methods and not really constructors. It might be better to capture `Foo` in `Foo{bar}` as the constructor instead.

amaanq:
It is the **norm** in go - from a quick fuzzy search in golang/go

image

ObserverOfTime:
> It is the **norm** in go - from a quick fuzzy search in golang/go

Is `Make...` also used as often?

ObserverOfTime (approved):

lucario387 (approved):

PR #4742, opened 2023-05-01
ci: tighten run requirements, apply formatting

amaanq:
* For example, a typo fix doesn't need every action to run
* Added a \n above the parsers list in the README as formatting specs indicate it should be there, and this will alleviate people using autoformatters having to remove that

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -12,9 +12,9 @@ pull_request_rules:
   - name: Prepare for merge
     conditions:
       - and:
-        - "-draft"
-        - "#approved-reviews-by=1"
-        - "#review-requested=0"
+          - "-draft"
+          - "#approved-reviews-by=1"
+          - "#review-requested=0"
     actions:
       comment:
         message: |
@@ -25,13 +25,13 @@ pull_request_rules:
   - name: Merge on approval
     conditions:
       - and:
-        - or:
-          - "#approved-reviews-by>=2"
-          - and:
-            - "#approved-reviews-by=1"
-            - "updated-at>=1 day ago"
-        - "-draft"
-        - "#review-requested=0"
+          - or:
+              - "#approved-reviews-by>=2"
+              - and:
+                  - "#approved-reviews-by=1"
+                  - "updated-at>=1 day ago"
+          - "-draft"
+          - "#review-requested=0"
     actions:
       merge:
         method: rebase
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,7 +2,15 @@ name: Linting and style checking
 
 on:
   push:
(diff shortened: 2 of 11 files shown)


.github/workflows/lint.yml
```diff
@@ -2,7 +2,11 @@ name: Linting and style checking
 
 on:
   push:
+    paths:
+      - "lua/**"
```
lucario387, line 6: Does this mean this doesn't lint tests files anymore?

.github/mergify.yml
```diff
@@ -12,9 +12,9 @@ pull_request_rules:
   - name: Prepare for merge
     conditions:
       - and:
-        - "-draft"
-        - "#approved-reviews-by=1"
-        - "#review-requested=0"
+          - "-draft"
+          - "#approved-reviews-by=1"
+          - "#review-requested=0"
```
lucario387, line 17: Seems like formatting error to me. Shouldn't all `- ...` be indented more here?

amaanq: No? But I think it comes from prettier's formatting

lucario387: Ok then :+1:

amaanq:
Only issue is CI still "requires" tests even if they're "not needed" depending on the files changes - not sure how I'd fix that

lucario387:
> Only issue is CI still "requires" tests even if they're "not needed" depending on the files changes - not sure how I'd fix that

Probably git diff from within the workflow and check for the diffs -> from that get the changes?

.github/mergify.yml
```diff
@@ -25,13 +25,13 @@ pull_request_rules:
...
+          - or:
+              - "#approved-reviews-by>=2"
+              - and:
+                  - "#approved-reviews-by=1"
+                  - "updated-at>=1 day ago"
+          - "-draft"
+          - "#review-requested=0"
     actions:
       merge:
         method: rebase
```
theHamsta, line 37: We disabled mergify. How about removing this file?

.github/workflows/test-queries.yml
```diff
@@ -4,9 +4,15 @@ on:
   push:
     branches:
       - "master"
+    paths:
+      - "lockfile.json"
+      - "queries/**"
```
theHamsta, line 9: We already have a cache for regenerating the parsers, changes in the build/install logic should also trigger building and tests.

It's very tricky to discover bugs when a important file of the cache was forgotten. Not sure if the paths could be inverted to only apply to your documentation usecase

theHamsta: Same holds for changes to parser.lua which add a new parser. Would try to use the same pattern as for the parser build cache plus queries

PR #4751, opened 2023-05-03
fix: remove unnecessary offset from html injection

lucario387:
Don't see why this is necessary when the insides can be captured directly


lucario387:
Duplicate of #4752

amaanq:
didn't notice that, good catch

PR #4753, opened 2023-05-03
fix: add @combined to markdown html

lucario387:
| Before  | After   |
| ------- | ------- |
| image | image |

### DIFF
--- a/queries/markdown/injections.scm
+++ b/queries/markdown/injections.scm
@@ -7,7 +7,7 @@
     (#set-lang-from-info-string! @_lang)
     (#exclude_children! @content))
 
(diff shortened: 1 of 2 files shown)


clason:
@MDeiml

lucario387:
> This worked for me. my plugin splitjoin.nvim, which previously didn't work at all in markdown html, only worked partially under this configuration, but that could be a plugin thing

Probably because of `exclude_children`, I'd assume, as Helix still uses include-unnamed-children

MDeiml:
Sorry for not responding, looks good

> @MDeiml

PR #4759, opened 2023-05-04
ref(rust): highlight include macros as `@include`

Iron-E:
e.g. `include_str!`

They are currently highlighted as `@macro`, which is accurate, but the `@include` group feels more precise (and similar to C, where `#include` is highlighted with this group)


amaanq:
I disagree w/ this simply because include can be used in a variety of macros, user-made or other builtins, that have the word include, and it's best not to tread down the path of segmenting as many macros as possible into "better-fitting" captures

PR #4760, opened 2023-05-04
ref(c_sharp): highlight `<`, `>` as bracket only in type param list

Iron-E:
This allows `x > 4` to show `>` as `@operator`, whereas it is currently `@punctuation.bracket`

### DIFF
--- a/queries/c_sharp/highlights.scm
+++ b/queries/c_sharp/highlights.scm
@@ -305,9 +305,9 @@
  "}"
  "("
  ")"
- "<"
- ">"
 ] @punctuation.bracket
+
+(type_argument_list ["<" ">"] @punctuation.bracket)
 
 [
  (this_expression)


queries/c_sharp/highlights.scm
```diff
@@ -305,10 +305,10 @@
  "}"
  "("
  ")"
- "<"
- ">"
```
amaanq, line 309: Restore these, > and < are also operators

amaanq (changes requested):

queries/c_sharp/highlights.scm
Iron-E: That's the `@punctuation.bracket` list; `<` and `>` are still in the `@operator` list (see line 270 and 272). These characters being in this list unconditionally is the problem

amaanq: Whoops, yeah

amaanq (approved):

Iron-E:
Thanks for the reviews!

PR #4761, opened 2023-05-04
ref(java): highlight `Foo<Bar>` as `@punctuation.delimiter`

Iron-E:
It is currently highlighted as `@operator`

### DIFF
--- a/queries/java/highlights.scm
+++ b/queries/java/highlights.scm
@@ -258,6 +258,9 @@
 
 [ "(" ")" ] @punctuation.bracket
 
+(type_arguments [ "<" ">" ] @punctuation.bracket)
+(type_parameters [ "<" ">" ] @punctuation.bracket)
+
 ; Exceptions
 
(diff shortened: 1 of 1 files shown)


queries/java/highlights.scm
```diff
@@ -258,6 +258,9 @@
 
 [ "(" ")" ] @punctuation.bracket
 
+(type_arguments . "<" @punctuation.bracket ">" @punctuation.bracket .)
```
lucario387, line 261: Is there any reason why it isn't `[ "<" ">" ] @punctuation.bracket` here? I'm pretty sure there isn't a necessity to anchor start and end node to this

Iron-E: I was trying to protect it from any future syntax change, but that's probably overly careful and/or invalid. I'll use the simpler version you suggested :+1:

amaanq (approved):

PR #4765, opened 2023-05-06
feat(erlang): add keyword and operator suppored by otp25

haoxianhan:
update some new keyword highlights in Erlang otp25.

### DIFF
--- a/queries/erlang/folds.scm
+++ b/queries/erlang/folds.scm
@@ -2,6 +2,7 @@
   (fun_decl)
   (anonymous_fun)
   (case_expr)
+  (maybe_expr)
   (map_expr)
   (export_attribute)
   (export_type_attribute)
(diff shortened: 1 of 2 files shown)


amaanq:
Is there a reason the conditionals have a 95 priority?

amaanq (approved):

haoxianhan:
I have no idea too. The `highlights.scm` here is quite different from the [source](https://github.com/WhatsApp/tree-sitter-erlang/blob/main/queries/highlights.scm). I need to take some time to understand. :)

amaanq:
That's fine, our precedence is last wins. I'll add a suggestion to remove the priority, just apply it

amaanq:
Couldn't since gh doesn't allow you to select above n lines of untouched code for suggestions...pushed it myself

Issue #4668, opened 2023-04-17
SQL file lose syntax


mr-scrpt:
### Describe the highlighting problem

For some reason, syntax highlighting is lost on the first line - I checked in other editors, such as VSCode and Sublime, and there is no loss of highlighting, the entire file is correctly highlighted until the very end. I launched the playground and I see that it highlights an error for me, but I don't understand what exactly the error is.

### Example snippet that causes the problem

```
CREATE TABLE "Role" (
  "roleId" bigint generated always as identity,
  "name" varchar NOT NULL
);

ALTER TABLE "Role" ADD CONSTRAINT "pkRole" PRIMARY KEY ("roleId");
CREATE UNIQUE INDEX "akRoleName" ON "Role" ("name");
CREATE TABLE "Account" (
  "accountId" bigint generated always as identity,
  "login" varchar(64) NOT NULL,
  "password" varchar NOT NULL
);

ALTER TABLE "Account" ADD CONSTRAINT "pkAccount" PRIMARY KEY ("accountId");
CREATE UNIQUE INDEX "akAccountLogin" ON "Account" ("login");

CREATE TABLE "AccountRole" (
  "accountId" bigint NOT NULL,
  "roleId" bigint NOT NULL
);

ALTER TABLE "AccountRole" ADD CONSTRAINT "pkAccountRole" PRIMARY KEY ("accountId", "roleId");
ALTER TABLE "AccountRole" ADD CONSTRAINT "fkAccountRoleAccount" FOREIGN KEY ("accountId") REFERENCES "Account" ("accountId") ON DELETE CASCADE;
ALTER TABLE "AccountRole" ADD CONSTRAINT "fkAccountRoleRole" FOREIGN KEY ("roleId") REFERENCES "Role" ("roleId") ON DELETE CASCADE;
CREATE TABLE "Area" (
  "areaId" bigint generated always as identity,
  "name" varchar NOT NULL,
  "ownerId" bigint NOT NULL
);

ALTER TABLE "Area" ADD CONSTRAINT "pkArea" PRIMARY KEY ("areaId");
CREATE UNIQUE INDEX "akAreaName" ON "Area" ("name");
ALTER TABLE "Area" ADD CONSTRAINT "fkAreaOwner" FOREIGN KEY ("ownerId") REFERENCES "Account" ("accountId");

CREATE TABLE "AreaAccount" (
  "areaId" bigint NOT NULL,
  "accountId" bigint NOT NULL
);

[...]

PR #4771, opened 2023-05-08
feat(highlights): add new keywords for sql parser

DerekStride:
## What

refs: https://github.com/nvim-treesitter/nvim-treesitter/issues/4668, https://github.com/DerekStride/tree-sitter-sql/issues/137, https://github.com/DerekStride/tree-sitter-sql/pull/138

Adding highlight queries for keywords added in https://github.com/DerekStride/tree-sitter-sql/pull/138.


amaanq (approved):

amaanq:
keyword_ naming convention makes it very hard to tell what is what :shrug:

DerekStride:
> keyword_ naming convention makes it very hard to tell what is what 🤷

Keywords in SQL are case-insensitive, we make the parser handle the variations with specific named nodes, e.g. `select`, `Select`, `SELECT`, `SeLeCt` are all valid (in practice only the first 3 are typically seen) and all result in a single `keyword_select` named node.

Without that convention the `highlights.scm` file would have to include all the variations.

amaanq:
> > `keyword_` naming convention makes it very hard to tell what is what shrug
> 
> Keywords in SQL are case-insensitive, we make the parser handle the variations with specific named nodes, e.g. `select`, `Select`, `SELECT`, `SeLeCt` are all valid (in practice only the first 3 are typically seen) and all result in a single `keyword_select` named node.
> 
> Without that convention the `highlights.scm` file would have to include all the variations.

a good convention would be to just alias it, keyword_select -> "SELECT" (most common variation)

ObserverOfTime (approved):

PR #4772, opened 2023-05-08
fix(typescript): remove duplicate capture

amaanq:
It's in ecma already (added a missing operator too now)

### DIFF
--- a/queries/typescript/highlights.scm
+++ b/queries/typescript/highlights.scm
@@ -55,9 +55,6 @@
 (type_annotation
   ":" @punctuation.delimiter)
 
-(pair
-  ":" @punctuation.delimiter)
-
 (index_signature
   ":" @punctuation.delimiter)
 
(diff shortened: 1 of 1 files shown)


queries/typescript/highlights.scm
```diff
@@ -38,6 +38,8 @@
 
 (template_literal_type) @string
 
+"!" @operator
```
lucario387, line 41: If anything, I'd say this should also be in ecma/highlights. Don't see why only typescript is allowed to have this highlighted

amaanq: well, in ecma it's distinguished on L257:
```(unary_expression ["!" "~" "-" "+"] @operator)```

Comment (lucario387) [reply queries/typescript/highlights.scm:41]: Yeah, but I don't remember any explicit case where "!" can be used in typescript that is not in javascript tho

Comment (amaanq) [reply queries/typescript/highlights.scm]:
```ts
const result = maybeundefinedreturn()!
```

lucario387: Well I don't know of that bizarre syntax :smile:

PR #4774, opened 2023-05-09
Add usd language

ColinKennedy:
Adding the [USD](https://www.pixar.com/usd) scene description language to nvim-treesitter.

Source: https://github.com/ColinKennedy/tree-sitter-usd

# Demonstrations
### Fold Via Blocks
https://user-images.githubusercontent.com/10103049/236990305-193a17c2-9d9f-4cd3-b232-fe121edea38c.mp4

### Fold Via Prim Definitions
https://user-images.githubusercontent.com/10103049/236990319-af47988b-9935-4547-8198-7a6afc89a534.mp4

## Indents
https://user-images.githubusercontent.com/10103049/236990331-2c701764-0d92-4c83-83ef-ef7a81357c3d.mp4

## Plug-ins
Currently using forked versions of existing plug-ins. But it is pretty gratifying to see it all working even with tools out in the wild.

### nvim-treesitter-context
https://user-images.githubusercontent.com/10103049/235326266-93c8e868-ed7f-47a7-bda9-238f02979e82.mp4

[nvim-treesitter-context](https://github.com/nvim-treesitter/nvim-treesitter-context)

### aerial.nvim
[aerial.nvim](https://github.com/stevearc/aerial.nvim)

https://user-images.githubusercontent.com/10103049/235325105-1490fb62-4c95-46bf-a170-50df4c7409ff.mp4

https://user-images.githubusercontent.com/10103049/235325115-a74d68c6-8f2d-40dd-a7ff-d58240f9b1cd.mp4

### winbar.nvim
[nvim-gps](https://github.com/SmiteshP/nvim-gps) + [winbar.nvim](https://github.com/fgheng/winbar.nvim)

https://user-images.githubusercontent.com/10103049/235326401-64be269b-5e96-4483-b6ee-995392603ef9.mp4

# Questions / Concerns For Reviewers
## textobjects.scm in existing documentation
I saw in the CONTRIBUTING.md file and a couple places in documentation that nvim-treesitter supports providing a ``textobjects.scm`` file. I checked around but no language has that. Is the new recommendation to define that file over at [nvim-treesitter-textobjects](https://github.com/nvim-treesitter/nvim-treesitter-textobjects) after this PR is merged?

## Folds - which do you prefer?
I'm wavering between whether to provide default folds which fold on each Prim definition or just fold on the Prim's text blocks. Personally I use "Fold Via Blocks" in my day to day life but as a default offering I could see others preferring / wanting to "Fold Via Prim Definitions". What do you think?


after/ftdetect/usda.lua
```diff
@@ -0,0 +1,4 @@
+vim.api.nvim_create_autocmd(
```
lucario387, line 1: Please don't add this here.

Keep it in userspace

after/ftdetect/usda.lua
lucario387, line 1: This should not exist here

lucario387 (changes requested):

queries/usd/indents.scm
```diff
@@ -0,0 +1,19 @@
...
+  (dictionary)
+  (list)
+  (list_proxy)
+  (metadata)
+  (timeSamples)
+  (tuple)
+  (variant_set_definition)
+] @indent.begin
+
+(block "}" @indent.end)
```
lucario387, line 12: Seems? to be missing `@indent.branch` for all of these`

queries/usd/locals.scm
```diff
@@ -0,0 +1,4 @@
+[
+  (block)
+  (metadata)
```
lucario387, line 3: Just a question, but should multiline tuples, arguments, ... be folded as well? And if so, should it be folded with the block, or not

after/ftdetect/usda.lua
lucario387: You should do a vim PR for the ft, then it will get patched by Neovim. 
This repo should not have any ftdetect logic

lucario387:
For indents, I'd like to look at the result of `gg=G` more than the video, but it looks fine for the most part

queries/usd/indents.scm
```diff
@@ -0,0 +1,19 @@
+[
+  (block)
+  (dictionary)
+  (list)
+  (list_proxy)
+  (metadata)
+  (timeSamples)
+  (tuple)
+  (variant_set_definition)
```
lucario387, line 9: Can you add a comment at EOL for a preview of what it is? i.e. `(block) ; { ... }`

after/ftdetect/usda.lua
ColinKennedy: My apologies about that. When I was looking for PR inspiration, I saw another person doing that over at https://github.com/nvim-treesitter/nvim-treesitter/pull/2159 so figured that was expected.

So the recommendation is

1. Remove here
2. Make [Vim](https://github.com/vim/vim) PR with the change
3. Request / PR https://github.com/neovim/neovim to patch it (what's the process like? Create an issue pointing to the commit(s) asking for them to be upstreamed?)

Like that?

lucario387: Yes, but with minor modifications
1. Remove here
2. Make a Vim PR using the old `au BufRead, BufNewFile *.usd* ...`
3. Create a `vim-patch` using vim-patch scripts in neovim/neovim, or ask a member there to port it to neovim. (zeertzq will probably port it before you even notice and request for one)

queries/usd/highlights.scm
```diff
@@ -0,0 +1,228 @@
+(None) @constant.builtin
+(asset_path) @string.special.asset
+(bool) @boolean
+(comment) @comment @spell
+(custom) @function.builtin
+(float) @float
+(integer) @number
+(namespace_identifier) @namespace
+(orderer) @function.call
+(prim_path) @string.special.prim
```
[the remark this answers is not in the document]
ColinKennedy: USD has at least 3-4 different "string-like" representations.

**Strings**
`"foo"` - A regular string. Depending on where in the file this is defined, this is actually a raw string, a docstring, or a comment

**Comments**
`# foo` - This is always an inline comment

**Asset Paths**
There are always external file references (e.g. loading another file)
`@./relative/path.usd@`
`@uri://www.resource.com/inner/path?query=here#fragment@`

**Prim Paths**
These are generally-but-not-always intra-file references (referring to another object in the same file or from another file)
`</path/to/foo_prim>`

**Asset Path + Prim Path**
The above two sections can be combined.
`@./relative/path.usd@</path/to/foo_prim>`

Because Asset Path and Prim Path can be combined or be separate, it would help a lot for readability's sake if they can be differentiated. The nvim-treesitter out-of-box literal recommendations are https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#literals but the `@string` literals don't map well to the different USD cases.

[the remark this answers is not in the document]
ColinKennedy: I don't have the link handy but I remember reading from either tree-sitter or nvim-treesitter's documentation that it's legal to append more text onto highlight groups and the longest highlight group will match. For people who just have  colorschemes that specify `@string.special`, these groups would work for those. But if someone wanted to add a buffer-local highlight group for just Asset Paths (for example) they still can.

If we can use "out of box" nvim-treesitter highlight groups for each of these 3-4 cases then I'm happy to do that though. If you have recommended alternatives I can try those out. Thank you :)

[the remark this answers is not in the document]
lucario387: I'm just saying that `prim_path` should be `@string.special`, or something markdown-url-like?. It will be aliased to `@string.special.filetype` anyways (Check your `:Inspect`)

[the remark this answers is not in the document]
amaanq: Strings - `@string`
Comments - `@comment`
Asset/Prim paths - `@string.special`, `@text.uri`, or `@string`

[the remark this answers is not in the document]
lucario387: @ObserverOfTime would give better recommendations here.

queries/usd/highlights.scm
```diff
@@ -0,0 +1,228 @@
...
+(uniform) @function.builtin
+
+[
+  "class"
+  "def"
+  "over"
+] @keyword.function
```
amaanq, line 17: Class is a function? :laughing: same question for over

queries/usd/highlights.scm
```diff
@@ -0,0 +1,228 @@
...
+)
+(relationship_declaration
+ (identifier) @variable
+)
+(relationship_assignment
+ (identifier) @variable
+)
+(qualified_identifier
+  (identifier) @variable)
+
```
amaanq, line 41: Just do `(identifier) @variable`

queries/usd/highlights.scm
```diff
@@ -0,0 +1,228 @@
...
+ .
+ (string) @comment.documentation)
+(metadata_assignment
+ (string) @string)
+(prim_definition
+ (string) @string)
+(variant_set_definition
+ (string) @string)
+(variant
+ (string) @string)
```
amaanq, line 139: ```(string) @string```, no need for all this

amaanq (changes requested):

queries/usd/highlights.scm
lucario387: for the section 
```scheme
(metadata
 (comment)*
 .
 (string) @comment.documentation)
```

The later it appears in the file, the higher priority it is

queries/usd/highlights.scm
ColinKennedy: It's a bit technical but I'll try to summarize. USD is a scene description format with a C++ and Python API, not a programming language itself, so "class", "def", and "over" don't refer to anything we might expect in programming terms even if you see a familiar word like "class".

If you're curious about a more technical explanation ...

`def` means "a concrete representation of data"
`over` means "something that (ideally) overrides another `def` which could be in the current file or referenced from some other file"
`class` means "a totally abstract piece of data that doesn't exist itself (like a `def`) but could be inherited by a `def`"

For an even greater technical breakdown, Pixar has a glossary on these terms which should be helpful for clarification

https://openusd.org/release/glossary.html#def

https://openusd.org/release/glossary.html#class

https://openusd.org/release/glossary.html#usdglossary-over

Please let me know if you have follow-up questions and I'll try to answer the best I can :)

queries/usd/highlights.scm
ColinKennedy: Just copying this comment here, because it's related

https://github.com/nvim-treesitter/nvim-treesitter/pull/4774#discussion_r1188790357

"foo" - A regular string. Depending on where in the file this is defined, this is actually a raw string, a docstring, or a comment

In USD, what we commonly think of as a string might be a string defining a docstring, a raw string, or a comment depending on where it's located in the file. My intent here was to be explicit about each case so that there's less chance of conflicting highlight groups matching. If I can express that intent more concisely please let me know

ColinKennedy: > The later it appears in the file, the higher priority it is

Oh okay, I didn't know that. I'll reorder the highlights to reflect that. Currently in a couple places it's the opposite order than it should be

after/ftdetect/usda.lua
ColinKennedy: Step (2) here, for reference: https://github.com/vim/vim/pull/12370

Step (1) I'll push once I've done most of the other stuff so I don't spam your inboxes too much :)

lucario387: Should I mark this as a draft PR for now?

queries/usd/indents.scm
ColinKennedy: Yep, that was a good catch. Just verified it on a test file. USD comes with a command that effectively formats the USD code. I created [this fairly complex USD file](https://gist.github.com/ColinKennedy/1997d1f0850031d22245f529cd92872f), intentionally messed up the indents, and my current indent logic doesn't match what USD's command outputs. But adding `@indent.branch` results in a perfect diff. Will fix!

queries/usd/highlights.scm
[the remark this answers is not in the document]
ColinKennedy: prim_path is now `@string.special` and asset_path is `@text.uri`

after/ftdetect/usda.lua
ColinKennedy: In case you were asking me and not maintainers - I think that's more a question for maintainers. Whether you want to always stall PRs to wait for Vim + Neovim to add ftdetect is a policy that maintainers should agree on.

Just weighing pros and cons - If we do merge this PR first before the filetype detection is added to vim / neovim and there's some sort of problem with nvim-treesitter then we wouldn't necessarily know about it until after this PR was merged. But on the other hand adding filetype info to nvim-treesitter that people have to "opt in" to match doesn't really hurt nvim-treesitter. There's always the possibility that the USD filetype detection PRs could be rejected though I'm not sure why that would happen. Just my two cents.

queries/usd/indents.scm
ColinKennedy: GitHub doesn't mark this out of date because I changed lines below but we should be good to go now. At my current place it's common practice for the reviewee to mark conversations as resolved but since I'm not sure that's common in OSS repos I will let a maintainer mark it instead

after/ftdetect/usda.lua
lucario387: I'm okay to merge this as long as the `ftdetect` is out of this PR, and the suggested changes by @amaanq are applied. indent/folds and such, if there's issues, can be fixed in following PRs

queries/usd/highlights.scm
```diff
@@ -0,0 +1,228 @@
...
+ (identifier) @variable
+)
+(relationship_declaration
+ (identifier) @variable
+)
+(relationship_assignment
+ (identifier) @variable
+)
+(qualified_identifier
+  (identifier) @variable)
```
lucario387, line 40: This whole section should just be `(identifier) @variable`

lucario387 (approved):

lucario387:
It's okay for me now

The highlight queries need changes, will let other maintainers decide on that

queries/usd/indents.scm
ColinKennedy: No problem. I'm assuming by EOL you mean an inline comment. I tried to summarize each ambiguous named node, please let me know if that's what you had in mind when you get the chance

ColinKennedy:
@lucario387 

With the latest changes to indents.scm here's the gg=G output

https://github.com/nvim-treesitter/nvim-treesitter/assets/10103049/2ac5d68c-8191-4447-8a85-c23296a6397b

I show and compare 2 files, `messed_up_indentation_fixed.usda`, scroll through it a bit just to show it's messed up, then `=G`. And I open `complex_usd_file.usda` which is effectively USD's auto-formatted version. Then I diff the two. The only differences shown are USD's preference to convert from `"""` to `"` and a part where it likes to put `(metadata)` and `(block)` on the same line as 

```usd
def "foo" (
    stuff = "here"
) {
}
```

rather than how I had it as

```usd
def "foo" (
    stuff = "here"
)
{
}
```

Which to be fair isn't really indent related. Otherwise that, indentation is identical

lucario387:
Lgtm on the indents

after/ftdetect/usda.lua
amaanq: ft detect being added to core does not matter for this pr, that has no bearing at all

queries/usd/highlights.scm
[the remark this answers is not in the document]
theHamsta: Our previous policy was to allow language specific suffixes when appropriate (not sure if this changes with unification efforts with helix). Just the prefix should follow contribution.md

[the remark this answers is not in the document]
ColinKennedy: I thought the language suffixes were automatic. e.g. If I call `:Inspect` while looking at a USD file the full highlight group is `@string.special.usd` even if I define it as `@string.special` in this file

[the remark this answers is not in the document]
clason: I think what was meant here were prefixes that are _not_ the language name but only makes sense for a specific language (like `@string.special.prim` here).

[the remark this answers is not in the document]
ColinKennedy: @clason Checking back on this comment again - "I'm just saying that prim_path should be @string.special, no need to say .prim, or something markdown-url-like, maybe @text.uri ?. It will be aliased to @string.special.filetype, which is @string.special.prim anyways (Check your :Inspect)" 

Yes it looks like there was that misunderstanding. But I think @amaanq recommendation works still.

Whether we use

prim_path - @string.special 
asset_path - @text.uri

or

prim_path - @string.special.prim
asset_path - @srting.special.asset

I'm okay with either approach as long as the highlights are distinguishable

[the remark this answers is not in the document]
lucario387: To better clarify my words, all hlgroups that you specify in `highlights.scm` file, for example, `@foo.bar`, when using in a file with ft `bax`, will have the hlgroups `@foo.bar.bax` providing the highlight. By default, this fallbacks to `@foo.bar`.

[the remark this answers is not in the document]
amaanq: The prim/asset nested capture are, with all due respect, useless as users/themes do **not** account for this. Hence why such an extension is best left up to user configuration/extension.

Imo both should be `@string.special` with one of them tacking on `@text.uri` as well if that makes sense 🤷

queries/usd/highlights.scm
ColinKennedy: Confirmed these edits + moving the (metadata ... (string) ...) to the end of the file passes unittests. Done!

queries/usd/highlights.scm
ColinKennedy: This is done now, thank you

queries/usd/highlights.scm
ColinKennedy: Done :)

ColinKennedy:
I think it's ready for a second look when you get the chance @amaanq 

Btw I confirmed that indent logic works well but for some reason my attempt at writing unittests for it aren't passing. It'd be cool to get those tests working before merging.

amaanq:
Query highlight tests are overkill, could you reduce that to 1-3 files? Each unique case need only be tested once, a fuller suite would be better suited in the grammar repo itself

lucario387:
@ColinKennedy on your end, `usd` is probably recognized as a filetype, while it's not recognized here

Add 
`vim.cmd [[au BufRead,BufNewFile *.usd *.usda set filetype=usd]]` to `scripts/minimal_init.lua` for now while waiting for upstream vim-patches

lucario387:
Yep, that's the problem

ColinKennedy:
> Query highlight tests are overkill, could you reduce that to 1-3 files? Each unique case need only be tested once, a fuller suite would be better suited in the grammar repo itself

@amaanq

I've reduced all tests down to be unique. Could we please keep the test files split up though? The current files have parity with the [tree-sitter-usd](https://github.com/ColinKennedy/tree-sitter-usd) repository which makes it really easy to bounce between them. The `20 files / 554 lines` are now `11 files / 142 lines` so I hope you'll find the new set-up to be much more manageable.

amaanq:
11 is still quite a lot, the most of any language and frankly looking at it I'd want to reduce hack and tiger's file counts too (9/10 respectively) , maybe for 1.0, so ideally keep it around 5 or so, it's just visual bloat for anyone navigating the repo

lucario387:
Also, please add the au BufRead,BufNewFile stuffs into here because it seems the nightly version fails to detect usda files rn for some reason ...

ColinKennedy:
Ah sorry @lucario387 I neglected to reply to your previous suggestion. I've added the `au` command now and looks like all tests are passing.

amaanq (approved):

amaanq:
Is it good to merge @ColinKennedy?

clason:
> Also, please add the au BufRead,BufNewFile stuffs into here because it seems the nightly version fails to detect usda files rn for some reason ...

You gotta wait for the nightly with the commit to actually be built...

> The current files have parity with the [tree-sitter-usd](https://github.com/ColinKennedy/tree-sitter-usd) repository which makes it really easy to bounce between them.

If the parser repo has tests, there's no sense in duplicating them here, though.

lucario387:
> You gotta wait for the nightly with the commit to actually be built...

Well it should be built since yesterday, but I tried 8 hours ago and it still fails ... Think it works now though.

ColinKennedy:
> If the parser repo has tests, there's no sense in duplicating them here, though.

That would be the case if this repo directly uses tree-sitter-usd's highlight queries but it doesn't. Though worth noting that the majority of the duplicate tests are already removed in any case.

ColinKennedy:
> Is it good to merge @ColinKennedy?

It should be. If you'd like to hold until tomorrow to see if the tests will pass without the `au` command, I can do that tomorrow. But either way tree-sitter-usd is ready and unittests pass both there and here. We should be good to go.

lucario387:
Just gonna merge this. We also need to remove gleam from that `minimal_init.lua` too.

PR #4780, opened 2023-05-10
comment: add support for parsing `DONE`

mortezadadgar:


ObserverOfTime:
What's the point of leaving it in the comment if it's done? :thinking:

mortezadadgar:
Yeah makes sense I should stop adding done in my codes

PR #4781, opened 2023-05-10
feat: add Odin language support

laytan:
https://odin-lang.org

### DIFF
--- a/queries/odin/indents.scm
+++ b/queries/odin/indents.scm
@@ -1,0 +1,17 @@
+[
+ (foreign_block)
+ (block)
+ (top_block)
+ (compound_literal)
+ (proc_call)
+ (assignment_statement)
+ (const_declaration)
+ (var_declaration)
+ (switch_statement)
+] @indent
+
+[
+ ")"
+ "}"
+] @indent_end @branch
+
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1014,6 +1014,7 @@ list.odin = {
   install_info = {
     url = "https://github.com/laytan/tree-sitter-odin",
     files = { "src/parser.c" },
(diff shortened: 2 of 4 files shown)


amaanq:
How coincidental! I just completed my own Odin parser today and was planning to PR it soon (I was not a fan of the "original" one, the one you forked) - taking a look at yours I noticed about ~440 files in the Odin core repo fail so I'd like to use mine instead

queries/odin/indents.scm
```diff
@@ -0,0 +1,17 @@
...
+ (foreign_block)
+ (block)
+ (top_block)
+ (compound_literal)
+ (proc_call)
+ (assignment_statement)
+ (const_declaration)
+ (var_declaration)
+ (switch_statement)
+] @indent
```
lucario387, line 11: This is using the old syntax
The equivalents are
- `@indent` -> `@indent.begin`
- `@indent_end` -> `@indent.end`
- `@branch` -> `@indent.branch`

queries/odin/indents.scm
```diff
@@ -0,0 +1,17 @@
...
+
+[
+ ")"
+ "}"
+] @indent_end @branch
+
```
lucario387, line 17: This is using the old syntax
The equivalents are

- `@indent` -> `@indent.begin`
- `@indent_end` -> `@indent.end`
- `@branch` -> `@indent.branch`

laytan:
@amaanq The grammar currently fails to parse the return types correctly, for example `-> (string, bool)` it for some reason tries to parse it as a group expression, and then errors when it doesn't know what to do with the `->`.

Not sure why, but I think when I figure that out that there are no more errors, I will get to that today hopefully. I will also double check all the core files, that's a good idea!

laytan:
Closing in favour of #4784

PR #4782, opened 2023-05-11
feat(t32): add indent rules

xasc:
- Adds indentation rules and a few tests
- Updates to new grammar version
- Extends the syntax highlighting tests


lua/nvim-treesitter/parsers.lua
```diff
@@ -1395,7 +1395,7 @@ list.t32 = {
   install_info = {
     url = "https://codeberg.org/xasc/tree-sitter-t32",
     files = { "src/parser.c", "src/scanner.c" },
-    revision = "675cd7de4eb80f2faa1e6d3bc9ee195fa9ef9790",
+    revision = "1dd98248b01e4a3933c1b85b58bab0875e2ba437",
```
amaanq, line 1398: Is there a reason you're using specific revisions instead of letting the ci auto update it?

tests/indent/t32/if_block.cmm
```diff
@@ -0,0 +1,49 @@
...
+)
+
+IF &f
+  IF &g
+    stop
+  ELSE
+    IF &h
+  (
+    continue
+  )
```
lucario387, line 48: Tbh, if-else chaining has been a problem for god knows how long. Without block scope delimiters it's impossible to know whether this is the elseif of the if inner or outer.

That said, this is fine.

lua/nvim-treesitter/parsers.lua
xasc: I am mainly developing the grammar on Linux and don't want to put in the effort to create an own CI agent for Windows or macOS builds. The result is that I broke Windows builds twice in a row without noticing, because I had not been able to add C99-compatible _static_assert()_ correctly.
The pinned revisions are checked more thoroughly to make sure that they build for _nvim-treesitter_ on all expected platforms and that all tests are passed. In addition, by pinning the version, I can work on the grammar without feeling pressured to bring the _nvim-treesitter_ parts back in sync after an auto update.

Please let me know if the _revision_ field should not be used in this way.

tests/indent/t32/if_block.cmm
xasc: Thank you for your feedback. Unfortunately, the TRACE32 scripting language has a lot of rough edges and that is surely one of them. I have added quite a few tests that I would consider as "nice-to-have", because I am still a bit unsure what the right expectation for the automatic indentation feature should be.

lua/nvim-treesitter/parsers.lua
amaanq: That's totally fine, I was just curious

queries/t32/indents.scm
```diff
@@ -0,0 +1,36 @@
...
+  (else_block
+    (if_block
+      _*
+      condition: (_)
+      . (_) @_then)) @indent.branch
+  (#not-has-type? @_then block)
+)
+
+(while_block
+  (command_expression) @indent.end .) @indent.begin
```
lucario387, line 30: Fwiw, with the current indent logic, this indent.end capture will most likely never matter.

indent.end must be the final and smallest range possible node of a line, i.e. ";" in C, "}"

lucario387: Note that indent.end as of now is strictly just for interactive typing. It has no effect on the result of "gg=G"

lucario387 (dismissed):

queries/t32/indents.scm
```diff
@@ -0,0 +1,36 @@
...
+
+(repeat_block
+  (command_expression) @indent.end .) @indent.begin
```
lucario387, line 33: I see `(command_expression)` here too, can you give me an example of what this node is?

queries/t32/indents.scm
xasc: This query is only covering the most basic case where we have a single command without delimiters after the IF:
```
WHILE &a
  Step
```

Removing `@indent.end` does not change the behavior, so it is gone now. I will probably have to revisit these queries in the future, if I get a better understanding of the differences between interactive typing and full file indentation.

queries/t32/indents.scm
xasc: This is the most basic case, too:
```
RePeaT &a
  print
```
A `command_expression` like `Data.Set VM:0x0--0xFF 0xA` starts with some command-dot-subcommands sequence and then continues with space- or comma-separated arguments. This command is for filling the memory range from `0` to `0xFF` with the value `0xA`.
There are many commands, so the possible arguments permutations are arbitrary. Commands terminate with the end of the line, so it is impossible to check for a special character sequence.

lucario387:
You should resolve the conflicts, then we can do a test run of this @xasc

amaanq:
unrelated

lucario387 (approved):

lucario387:
@xasc uhh.... This is my fault, but:

- `t32` is not a recognized filetype for neovim, which causes issues to the parser/installer
- After setting the filetype, using the indent module causes issues, is it okay for me to re write it?

PR #4784, opened 2023-05-11
feat: add Odin

amaanq:
cc: @laytan @WillEhrendreich

@laytan Give this a shot :grin:

### DIFF
--- a/queries/odin/highlights.scm
+++ b/queries/odin/highlights.scm
@@ -1,0 +1,293 @@
+; Preprocs
+
+[
+  (calling_convention)
+  (tag)
+] @preproc
+
+; Includes
+
+[
+  "import"
+  "package"
+] @include
+
+; Keywords
+
+[
+  "foreign"
+  "using"
+  "struct"
+  "enum"
+  "union"
+  "defer"
+  "cast"
+  "transmute"
+  "auto_cast"
+  "map"
+  "bit_set"
+  "matrix"
+] @keyword
+
+[
+  "proc"
+] @keyword.function
+
+[
+  "return"
+  "or_return"
+] @keyword.return
+
+[
+  "distinct"
+  "dynamic"
+] @storageclass
+
+; Conditionals
+
+[
+  "if"
+  "else"
+  "when"
+  "switch"
+  "case"
+  "where"
+  "break"
+  (fallthrough_statement)
+] @conditional
+
+((ternary_expression
+  [
+    "?"
+    ":"
+    "if"
+    "else"
+    "when"
+  ] @conditional.ternary)
+  (#set! "priority" 105))
+
+; Repeats
+
+[
+  "for"
+  "do"
+  "continue"
+] @repeat
+
+; Variables
+
+(identifier) @variable
+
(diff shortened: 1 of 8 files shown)


queries/odin/indents.scm
```diff
@@ -0,0 +1,26 @@
...
+] @indent.begin
+
+[
+  "}"
+  ")"
+  "]"
+] @indent.branch @indent.end
+
+[
+  (comment)
```
lucario387, line 23: `(block_comment)`

laytan:
This looks great and is a better implementation than I hacked together.

One thing I would like to see is a highlight for an explicit enum: `.Foo` or `.Test` (basically a member expression with only one identifier if I look at your grammar, would probably be an @ type, not sure if there is something for enums.

Another thing, make 'context' a @ builtin.variable.

I will close my PR

laytan:
One more thing I noticed is that this gets highlighted as a function, but it is actually a type (because it has no body):
`Handler_Proc :: proc(next: ^Handler, req: ^Request, res: ^Response)`

amaanq:
@laytan I fixed the freeze with regards to brackets, can you try it out a bit more and let me know of other potential bugs?

Also the proc types vs functions & context builtin issues are fixed, but can you elaborate more on the explicit enums?

laytan:
@amaanq In this snippet for example:
```odin
method_string :: proc(m: Method) -> string {
	switch m {
	case .Get:     return "GET"
	case .Head:    return "HEAD"
	case .Post:    return "POST"
	case .Put:     return "PUT"
	case .Patch:   return "PATCH"
	case .Trace:   return "TRACE"
	case .Delete:  return "DELETE"
	case .Connect: return "CONNECT"
	case .Options: return "OPTIONS"
	case:          return ""
	}
}
```

The compiler knows here that we are working with the Status enum, in the cases where it does, you can leave out the `Status` in `Status.Get`, so `Status.Get` == `.Get`. This is only possible with enums, so you can highlight that as an enum/type.

Here is the official overview's section about them: http://odin-lang.org/docs/overview/#implicit-selector-expression

Your other fixes and addition are working great btw!

amaanq:
so...would `@constant` be appropriate?

laytan:
@amaanq Its an enum, I have no clue what other languages with enums highlight that as, I think you know that better than I do ;)

amaanq:
yeah constant, regular member expressions (foo.bar) are fields, should be good now then.

queries/odin/highlights.scm
```diff
@@ -0,0 +1,293 @@
...
+
+((ternary_expression
+  [
+    "?"
+    ":"
+    "if"
+    "else"
+    "when"
+  ] @conditional.ternary)
+  (#set! "priority" 105))
```
lucario387, line 67: Hm, can't this be done in a way that doesn't require tweaking priority?

amaanq: It's because of ":" in punctuation as well, I could move it below punctuation, but I like where it is now

lucario387 (approved):

PR #4790, opened 2023-05-12
feat: add indents.scm for nickel

zhengpd:
This add basic indentation support for nickel files.


queries/nickel/indents.scm
```diff
@@ -0,0 +1,8 @@
+[
+  (term)
+] @indent.begin
+
+[
+  "]"
+  "}"
```
lucario387, line 7: This should also be captured with `@indent.end`

queries/nickel/indents.scm
```diff
@@ -0,0 +1,8 @@
+[
+  (term)
```
lucario387, line 2: Does this mean `(term)` consists of both `[...]` and `{...}` sections

tests/indent/nickel_spec.lua
```diff
@@ -0,0 +1,18 @@
+local Runner = require("tests.indent.common").Runner
+
+local run = Runner:new(it, "tests/indent/nickel", {
+  shiftwidth = 2,
+  expandtab = true,
+})
+
+describe("indent Nickel:", function()
```
lucario387, line 8: There should be a whole file test here, to check whether it's running or not

queries/nickel/indents.scm
zhengpd: > Does this mean `(term)` consists of both `[...]` and `{...}` sections

@lucario387 `(term)` is the container of all value types,  including `[...]` and `{...}`

lucario387 (approved):

amaanq:
squash as `feat(nickel): add indents` @lucario387

PR #4456, opened 2023-03-07
docs: Add documentation for running/writing tests

dwoznicki:
This is a rough draft of documentation for testing nvim-treesitter. It is incomplete (and possibly contains incorrect information) because I've only documented what I know: indentation tests using vim-plug as my package manager.

Addresses #3837

PR #4791, opened 2023-05-12
doc: add testing guide

zhengpd:
This PR adds some simple information about running tests on local development. The content is a summary from how I ran the tests when changing some scm files.


clason:
Thank you, that's good to put down. But could you make it non-specific to any plugin manager? The instructions should be generic (ideally without any plugin manager at all, so you can use `nvim --clean -u test.lua`).

lucario387:
Should this be an extension from #4456 , or a from ground-up version for the testing doc

PR #4798, opened 2023-05-13
highlights(markup): distinguish heading level in queries

clason:


clason:
@MDeiml

ObserverOfTime:
Isn't this too specific?

clason:
What do you mean? Specific is not bad in itself. And the legacy syntax files allow distinguishing H1--H6 for markup (HTML, Markdown, Vimdoc).

lucario387:
The goal is to align with https://docs.helix-editor.com/themes.html#syntax-highlighting @ObserverOfTime, so would be nice to have your input

amaanq (approved):

ObserverOfTime:
I understand the goal but I don't believe there's a reason to distinguish each heading level when it's not possible to change the font size in order to differentiate them.

lucario387:
We can differentiate using colors instead (compromise ...)

clason:
> I understand the goal but I don't believe there's a reason to distinguish each heading level when it's not possible to change the font size in order to differentiate them.

Colors, bold, italic are all useful for this. And you don't _have_ to, as all these default to the previous `@text.title`.

Again, this is about restoring parity with legacy syntax and Helix (as requested in https://github.com/MDeiml/tree-sitter-markdown/issues/96).

ObserverOfTime (approved):

lucario387 (approved):

MDeiml:
Lgtm, but there seem to be some changes unrelated to headings, specifically changing the highlight for code block info strings. Also an improved though probably.

clason:
Yes, that was noted in the commit message (not PR description): we used to conceal (not show) the code block delimiters -- including language -- by default, but that was controversial so I took the opportunity to revert it.

PR #4818, opened 2023-05-17
refactor(indent): backport improvements from main

clason:
* replace ts_utils.memoize_by_buftick by simpler implementation
* simplify construction of hash map
* prefer Neovim APIs

### DIFF
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -1,7 +1,5 @@
-local parsers = require "nvim-treesitter.parsers"
-local queries = require "nvim-treesitter.query"
-local tsutils = require "nvim-treesitter.ts_utils"
 local ts = vim.treesitter
+local parsers = require "lua.nvim-treesitter.parsers"
 
 local M = {}
 
@@ -14,6 +12,10 @@ M.comment_parsers = {
   jsdoc = true,
   phpdoc = true,
 }
+
+local function getline(lnum)
+  return vim.api.nvim_buf_get_lines(0, lnum - 1, lnum, false)[1] or ""
+end
 
 ---@param root TSNode
 ---@param lnum integer
@@ -29,8 +31,16 @@ end
 ---@param col? integer
 ---@return TSNode
 local function get_last_node_at_line(root, lnum, col)
-  col = col or (#vim.fn.getline(lnum) - 1)
+  col = col or (#getline(lnum) - 1)
   return root:descendant_for_range(lnum - 1, col, lnum - 1, col)
+end
+
+---@param node TSNode
+---@return number
+local function node_length(node)
+  local _, _, start_byte = node:start()
+  local _, _, end_byte = node:end_()
+  return end_byte - start_byte
 end
 
 ---@param bufnr integer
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/indent.lua
```diff
@@ -52,52 +61,59 @@ local function find_delimiter(bufnr, node, delimiter)
...
   local bufnr = vim.api.nvim_get_current_buf()
-  local parser = parsers.get_parser(bufnr)
+  local parser = ts.get_parser(bufnr)
   if not parser or not lnum then
     return -1
   end
 
-  local root_lang = parsers.get_buf_lang(bufnr)
+  local ft = vim.bo[bufnr].filetype
+  local root_lang = vim.treesitter.language.get_lang(ft) or ft
```
lucario387, line 116: Im pretty sure `get_lang(ft)` changes have yet to be backported, so this will be a regression to some fts (if any)

lucario387: Aside from this, I agree with the changes.
Should this also be changed for `memoize_by_buf_tick` directly, as `nvim-treesitter#foldexpr()` still relies on it?

clason: I wouldn't touch `foldexpr`, as it's already obsolete -- `vim.treesitter.foldexpr` was added in 0.9.0.

clason:
Indentation tests are now failing, but I don't quite see why they should. Do you have any idea?

lucario387:
The refactor seems to have missed somewhere... I'm checking the problems.

lucario387:
The `query:iter_captures`, alongside the `ts.is_in_node_range` is failing to find the smallest subtree that contains the current line
testing with `trailing_whitespace.html` shows that the node to start iterating from is `fragment`, which is the root node of html(the node above all nodes)

clason:
It's something to do with private nodes (starting with `_`), which `wrapped_iter` is somehow handling specially. But I don't really understand how...

lucario387:
@clason `query.iter_captures` in `query.lua` is getting the first subtree of the parser, and not the original tree
That may have been the issue

clason:
I don't think that branch is hit in `indent.lua`?

clason:
Hmm... doesn't look like a regression from these changes, actually.

lucario387:
Oh wow, `wgsl` and `tig` aren't recognized. This means the thing has been running without any queries for forever.

lucario387:
> Ah, so it's the buflang logic that's changed. Thanks! But why were the tests not failing before?

It never used treesitter indents :D, that's why

I didn't look at it before. Just comment the tiger tests out, the indent queries were missing some captures + the algorithm was probably different since then (12 months ago is a long time)

clason:
Yes, but why is it now?

lucario387:
I think it's probably because of vim.treesitter.get_parser not actually in parity to `parsers.get_parser`, maybe?

lucario387:
Ah, okay, @clason, Tiger's tests were expecting the text to also trigger a `==`, which it currently doesn't do, leading to failures.

Feel free to fix the tests that fails, or marked them as `XFAIL`, I'll have a check later

lucario387 (approved):

PR #4823, opened 2023-05-18
fix(query): allow multiline modeline (inherits/extends)

milisims:
`:h treesitter-query-modeline` reads:
```vimdoc
Note: These modeline comments must be at the top of the query, but can be
repeated, for example, the following two modeline blocks are both valid:
>query

    ;; inherits: foo,bar
    ;; extends

    ;; extends
    ;;
    ;; inherits: baz
<
```
This PR just allows multiple comments to start the block.

### DIFF
--- a/queries/query/highlights.scm
+++ b/queries/query/highlights.scm
@@ -27,8 +27,8 @@
 ((parameters (identifier) @number)
  (#match? @number "^[-+]?[0-9]+(.[0-9]+)?$"))
 
-((program . (comment) @include)
+((program . (comment)* . (comment) @include)
  (#match? @include "^;\ +inherits\ *:"))
 
-((program . (comment) @preproc)
+((program . (comment)* . (comment) @preproc)
  (#match? @preproc "^; +extends"))
--- a/queries/query/highlights.scm
+++ b/queries/query/highlights.scm
@@ -28,7 +28,7 @@
  (#match? @number "^[-+]?[0-9]+(.[0-9]+)?$"))
 
 ((program . (comment)* . (comment) @include)
- (#match? @include "^;\ +inherits\ *:"))
+ (#match? @include "^;+ +inherits *:"))
 
 ((program . (comment)* . (comment) @preproc)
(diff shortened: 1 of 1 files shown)


clason:
Can you show the difference this would make?

And note that this section in the docs was a minor error; on master this has already been corrected to
```
>query
    ;; inherits: foo,bar
    ;; extends
<
>query
    ;; extends
    ;;
    ;; inherits: baz
<
```

amaanq:
wouldn't . (comment)+ work without having to add a 2nd comment node

milisims:
It would highlight all of the comments rather than just the line that has the inherit or extend, tried it

milisims:
Old:
image
with changes:
image


> Can you show the difference this would make?
> 
> And note that this section in the docs was a minor error; on master this has already been corrected to
> 
> ```
> >query
>     ;; inherits: foo,bar
>     ;; extends
> <
> >query
>     ;; extends
>     ;;
>     ;; inherits: baz
> <
> ```

milisims:
Oh, problem. This also highlights this improperly:
image

clason:
Can you also fix it so the doc example (with multiple `;`) is highlighted correctly?

clason (approved):

milisims:
Oh, last find. testing `;extends` does extend properly, but the original query required whitespace.

queries/query/highlights.scm
```diff
@@ -27,8 +27,8 @@
 ((parameters (identifier) @number)
  (#match? @number "^[-+]?[0-9]+(.[0-9]+)?$"))
 
-((program . (comment) @include)
- (#match? @include "^;\ +inherits\ *:"))
+((program . (comment)* . (comment) @include)
+ (#match? @include "^;+ *inherits *:"))
```
lucario387, line 31: @amaanq Use `lua-match` where possible, run no.3? :smile:

amaanq: oh yeah :sweat_smile:

PR #4828, opened 2023-05-19
Add indents for java Annotations + notes

lucario387:
- Add indents for `@interface` 
  - `public @interface Foo` was missing indents before this
  - The parts inside `@Annotation(...)` was missing indents, and similarly, `{...}` inside `@Annotation`
- Add more notes to java indentations in case somebody else have a look at it in the future, they don't have to scroll the trees


lucario387:
@amaanq updated the PR to be more correct of its purposes.

I don't know why Lua tests are failing though.

amaanq (approved):

amaanq:
I still think this is a bit overkill and it's better for a contributor to familiarize himself w/ the actual grammar rules/repo to produce better contributions, but I'm not against it

PR #4834, opened 2023-05-20
highlights(go): add some highlights

smoka7:
- added missing [operators](https://go.dev/ref/spec#Operators_and_punctuation)
- added [iota](https://go.dev/ref/spec#Predeclared_identifiers) constant
- added comparable
- edited go keyword type

### DIFF
--- a/queries/go/highlights.scm
+++ b/queries/go/highlights.scm
@@ -64,6 +64,8 @@
   "&"
   "&&"
   "&="
+  "&^"
+  "&^="
   "%"
   "%="
   "^"
@@ -92,15 +94,12 @@
 
 [
   "break"
-  "chan"
   "const"
   "continue"
   "default"
   "defer"
-  "go"
   "goto"
   "interface"
-  "map"
   "range"
   "select"
   "struct"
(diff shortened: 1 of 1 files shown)


theHamsta (approved):
Nice thank you!

The only doubt that I currently have with `iota` being a `constant` as it is not a value in the traditional sense. But I would agree with your decision to use `@constant.builtin`.

Leaving this open for a moment, in case that there are other opinions.

smoka7:
Whats your opinion about adding `struct`,`map`,`interface` and `chan` to `@type`?

theHamsta:
`struct` keyword shouldn't be `@type`. I think also `interface` is used in that way that it only comes before the actual type identifier.

Adding `map`/`chan` to `@type.builtin` sounds good to me

amaanq:
comparable isn't exactly a type but more an interface.... (hint hint we could benefit from `@interface`)

theHamsta:
> comparable isn't exactly a type but more an interface.... (hint hint we could benefit from `@interface`)

But interface is a type. `@type.interface` would be a more compatible change

PR #4835, opened 2023-05-20
highlights(typescript): `is` and `:` in type predicate

zweimach:
| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/queries/typescript/highlights.scm
+++ b/queries/typescript/highlights.scm
@@ -13,6 +13,7 @@
   "satisfies"
   "module"
   "infer"
+  "is"
 ] @keyword
 
 (as_expression "as" @keyword)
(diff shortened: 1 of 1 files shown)


amaanq (approved):

amaanq:
Hmm, shouldn't `is` be `@keyword.operator`

zweimach:
Yes, I think it could be. I put `is` under `@keyword` because `keyof` and `as` is also `@keyword`.

amaanq:
`as` should be `@keyword.operator` too I'd think

zweimach:
Okay, I got it.
So I take this chance to make some keywords `@keyword.operator`.
From `ecma`: `in`, `instanceof`, and `typeof`.
From `typescript`: `as`, `is`, `keyof`, and `satisfies`.

amaanq:
I was fine with `is` as `@keyword` though :raised_eyebrow:

amaanq:
Oh well...

zweimach:
Thanks, and sorry about the delay.
I may have missed your comments, or I misunderstood your intentions.
Either way, I made a separate PR to clarify that `is` is not included as `@keyword.operator`.

PR #4837, opened 2023-05-20
Add ISPC language support

fab4100:
- Add ISPC parser config
- Add ISPC queries

The two commits add grammar and queries for the Intel implicit SPMD program
compiler ([ISPC](https://ispc.github.io/ispc.html)).

### DIFF
--- a/queries/ispc/highlights.scm
+++ b/queries/ispc/highlights.scm
@@ -1,0 +1,225 @@
+; inherits: c
+
+[
+  "soa"
+  "task"
+  "launch"
+  "unmasked"
+  "template"
+  "typename"
+  (sync_expression)
+] @keyword
+
+[
+  "in"
+  "new"
+  "delete"
+] @keyword.operator
+
+[
+  "cdo"
+  "cfor"
+  "cwhile"
+  "foreach"
+  "foreach_tiled"
+  "foreach_active"
+  "foreach_unique"
+] @repeat
+
+[
+  "cif"
+] @conditional
+
+[
+  "varying"
+  "uniform"
+] @type.qualifier
+
+"__regcall" @attribute
+
+(overload_declarator name: _ @function)
+(foreach_statement range_operator: _ @operator)
+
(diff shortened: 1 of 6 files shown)


queries/ispc/highlights.scm
```diff
@@ -0,0 +1,225 @@
...
+[
+  "cif"
+] @conditional
+
+[
+  "varying"
+  "uniform"
+] @type.qualifier
+
+"__regcall" @attribute
```
amaanq, line 38:
```suggestion
"__regcall" @storageclass
```

queries/ispc/injections.scm
```diff
@@ -0,0 +1,3 @@
+(preproc_arg) @c
+
+(comment) @comment
```
amaanq, line 3:
```suggestion
; inherits: c
```

amaanq (changes requested):

amaanq:
Mostly good, left a couple notes

queries/ispc/highlights.scm
fab4100: The `__regcall` attribute does not change the storage class of an object, it affects function argument passing in registers (https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-9/c-c-calling-conventions.html).  A similar attribute is `__vectorcall` which is already handled in `c/highlights.scm` as `@attribute`.  I would treat `__regcall` the same as `__vectorcall`.

amaanq: I was thinking of storageclass in a loose sense that a type of calling convention can be linked to its layout/storage at runtime, but I didn't see other similar ones already marked as attribute...seems fine then

amaanq (approved):

queries/ispc/highlights.scm
fab4100: Since Intel C/C++ compilers support `__regcall` it should probably be moved into `c/highlights.scm` anyway.

PR #4841, opened 2023-05-21
highlights(query): prefer lua-match

clason:


queries/query/highlights.scm
```diff
@@ -28,7 +28,7 @@
  (#match? @number "^[-+]?[0-9]+(.[0-9]+)?$"))
```
clason, line 28: would be nice to convert this as well, but I don't think Lua patterns support optional groups.

amaanq (approved):

queries/query/highlights.scm
amaanq: yeah it'd be two queries then

PR #4856, opened 2023-05-24
shim plenary error

lucario387:


clason:
@lucario387 can you report the issue at plenary? Conni2461 is active today, so this might be able to be fixed in a timely manner.

Conni2461:
did `c4d6e556f21b8e0afb8ee094e5f4ce4b90dae54b` break stuff?

lucario387:
> did `c4d6e556f21b8e0afb8ee094e5f4ce4b90dae54b` break stuff?

Yes, it was that commit that introduced the breakage.
`ee6fda7de0248d22853a510e329309e808ddc377` fails, `c4d6e556f21b8e0afb8ee094e5f4ce4b90dae54b` fails, while `7cc390e642d672a9789ad66637e77e06a0ae9522` passed (as can be seen from the workflow file changes)

Conni2461:
okay, give me 10 - 15, i try to fix it otherwise i would revert. you guys dont need to shim/pin a version

lucario387:
I don't really understand why, but the commit mentioned should not be leading to this error (if any). And if it does, the test only fails with lua files, and not happening to any other filetype

Conni2461:
i gonna fix it after lunch break

lucario387:
@Conni2461 shim'd it, the old `minimal_init` is now `init`, so I just changed it here. The `test_harness` commit should be marked as breaking imo, just because of that behavioral changes

Conni2461:
No i am gonna change plenary, i dont a breaking change. I just think --clean make problems and reverting back to --noplugin will fix it

lucario387:
I'm running on latest plenary.nvim, but needing to use `init` instead of `minimal_init` @Conni2461  @clason for the currently passing test.

Should this still be a plenary.nvim issue, or is it our issue?

clason:
Yes, as plenary will revert this breaking change on principle, we should do nothing.

Conni2461:
i have a fix, so i will not revert, but you dont have to do anything.

passed CI: https://github.com/Conni2461/nvim-treesitter/actions/runs/5068166798/jobs/9100118804

lucario387:
I'll close this PR then, as it's unnecessary anymore :smile:

wookayin:
The relevant code is here: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/scripts/run_tests.sh#L6-L9

```
nvim --headless --noplugin -u scripts/minimal_init.lua \
        -c "PlenaryBustedDirectory $1 { minimal_init = './scripts/minimal_init.lua' }"
```

But either case `--clean -u scripts/minimal_init.lua` or `--noplugin -u scripts/minimal_init.lua`, shouldn't it work the same? @lucario387, why was this a problem?

lucario387:
It doesn't as `--noplugin` also stops loading plugin files, which does matter for the CI.

Issue #4852, opened 2023-05-23
Add indentation rules for SQL


matthias-Q:
**Is your feature request related to a problem? Please describe.**
The SQL parser only has highlighting and simple injections, but not indentation rules.

**Describe the solution you'd like**
Some basic indentation rules should be added. Mostly for simple select queries, queries with CTEs, create table statements and such...

**Describe alternatives you've considered**
I am trying to get these queries working by myself, but I am struggling to get them working

**Additional context**
I am willing to add a MR, since I am contributing to the tree-sitter-sql parser a lot.

Here is what I was trying:
indents.scm:
```
(select_expression) @indent.begin
```

Test SQL query after automatically indenting it with tree-sitter:
```sql
select
a,
    b
from tab;
```
Note the column `a` is not indented, even though it is a node `term` and a child of `select_expression`

AST:
```
statement [0, 0] - [3, 9]
  select [0, 0] - [2, 5]
    keyword_select [0, 0] - [0, 6]
    select_expression [1, 0] - [2, 5]
      term [1, 0] - [1, 1]
        value: field [1, 0] - [1, 1]
          name: identifier [1, 0] - [1, 1]
      term [2, 4] - [2, 5]
        value: field [2, 4] - [2, 5]
          name: identifier [2, 4] - [2, 5]
  from [3, 0] - [3, 8]
    keyword_from [3, 0] - [3, 4]
    relation [3, 5] - [3, 8]
      table_reference [3, 5] - [3, 8]
        name: identifier [3, 5] - [3, 8]
```

lucario387:
I'll jump in if needed :smile: 

Feel free to PR it, I'll check if there's inconsistencies

Also, please comment a bit on why some queries are used, as I think there may be hacks needed(?, unsure, but most likely will happen)

PR #4857, opened 2023-05-24
indents(sql): initial support

matthias-Q:
As mentioned in #4852 I want to add indentation rules for SQL. 

Some of them are not working as intended and I don't know why. I have added test cases, but I also don't know how to run the tests locally and how the spec file should be written. Here is the indented SQL file for all the test cases with comments what is currently working/not working:
```SQL
-- first term in select_expression is not indented
select
a,
    b
from tab;

-- closing bracket of cte is not dedented
with data as (
    select
    a,
        b
    from tab
    )
select * from data;

-- this works!
create table my_table (
    id bigint,
    date date
);

-- case when then end works
-- issue is again the first term
-- of the select_expression
select
    case
        when a = 1 then '1'
        when a = 2 then '2'
        when a = 3 then '3'
        else '0'
    end as stmt1
from tab;

-- this may require a new node type
-- like a column_list or values_list
insert into mytable
( column1, column2)
values
('john', 123),
('jane', 124);

-- should indent the statement node
begin
create table foo (bar int);
end;

-- opening bracket of subquery should indent
-- or at least I am not sure if thats good style
-- if not, then this also works.. I guess?
select 
id
from foo
where id <
(
    select
    id
    from bar
    limit 1
);
```

I suspect, that for the `insert` statement, we require an extra layer in the AST for the column and values lists to be queriable, right?

Ping @lucario387

### DIFF
--- a/queries/sql/indents.scm
+++ b/queries/sql/indents.scm
@@ -1,0 +1,26 @@
+[
+ (select_expression)
+ (cte)
+ (column_definitions)
+ (case)
+ (keyword_when)
+ (keyword_else)
+ (subquery)
+] @indent.begin
+
+
+(compound_statement
+  (keyword_begin)
+  (statement) @indent.begin
+)
+
+(column_definitions "(" @indent.begin)
+(column_definitions ")" @indent.branch)
+(subquery ")" @indent.branch)
+(subquery "(" @indent.begin)
+
+[
+ (keyword_end)
+] @indent.branch
+
+ (keyword_from) @indent.align
--- a/queries/sql/indents.scm
+++ b/queries/sql/indents.scm
@@ -12,11 +12,9 @@
   (keyword_begin)
 ) @indent.begin
 
-(column_definitions "(" @indent.begin)
 (column_definitions ")" @indent.branch)
 
 (subquery ")" @indent.branch)
-(subquery "(" @indent.begin)
 
 (cte ")" @indent.branch)
 
--- a/queries/sql/indents.scm
+++ b/queries/sql/indents.scm
@@ -1,26 +1,29 @@
 [
- (select_expression)
+ (select)
  (cte)
  (column_definitions)
  (case)
- (keyword_when)
- (keyword_else)
  (subquery)
+ (insert)
 ] @indent.begin
 
 
 (compound_statement
   (keyword_begin)
-  (statement) @indent.begin
-)
+) @indent.begin
 
 (column_definitions "(" @indent.begin)
 (column_definitions ")" @indent.branch)
+
 (subquery ")" @indent.branch)
 (subquery "(" @indent.begin)
 
+(cte ")" @indent.branch)
+
 [
  (keyword_end)
+ (keyword_values)
+ (keyword_into)
 ] @indent.branch
 
- (keyword_from) @indent.align
+(keyword_end) @indent.end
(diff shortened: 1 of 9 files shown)


queries/sql/indents.scm
```diff
@@ -0,0 +1,26 @@
+[
+ (select_expression)
```
lucario387, line 2: This should be select

lucario387:
> I suspect, that for the insert statement, we require an extra layer in the AST for the column and values lists to be queriable, right?

From what I can see through the test files, yes.

queries/sql/indents.scm
```diff
@@ -0,0 +1,26 @@
...
+(column_definitions "(" @indent.begin)
+(column_definitions ")" @indent.branch)
+(subquery ")" @indent.branch)
+(subquery "(" @indent.begin)
+
+[
+ (keyword_end)
+] @indent.branch
+
+ (keyword_from) @indent.align
```
lucario387, line 26: This won't has any effects.

queries/sql/indents.scm
```diff
@@ -0,0 +1,26 @@
+[
+ (select_expression)
+ (cte)
+ (column_definitions)
+ (case)
+ (keyword_when)
+ (keyword_else)
```
lucario387, line 7: So are these

queries/sql/indents.scm
```diff
@@ -0,0 +1,26 @@
...
+(compound_statement
+  (keyword_begin)
+  (statement) @indent.begin
+)
+
+(column_definitions "(" @indent.begin)
+(column_definitions ")" @indent.branch)
+(subquery ")" @indent.branch)
+(subquery "(" @indent.begin)
+
```
lucario387, line 21: similarly, we can have `(cte ")" @indent.branch)` here

lucario387:
https://gist.github.com/lucario387/30b12c64492cc2a0399d5724454243b3

I have a version derived from your queries here, so you can check it out

queries/sql/indents.scm
lucario387: Oh, and don't mark "(" as indent.begin, that won't have any effects

queries/sql/indents.scm
matthias-Q: Are you sure? This is the tree for that statement:
```
statement [1, 0] - [4, 9]
  select [1, 0] - [3, 5]
    keyword_select [1, 0] - [1, 6]
    select_expression [2, 0] - [3, 5]
      term [2, 0] - [2, 1]
        value: field [2, 0] - [2, 1]
          name: identifier [2, 0] - [2, 1]
      term [3, 4] - [3, 5]
        value: field [3, 4] - [3, 5]
          name: identifier [3, 4] - [3, 5]
  from [4, 0] - [4, 8]
    keyword_from [4, 0] - [4, 4]
    relation [4, 5] - [4, 8]
      table_reference [4, 5] - [4, 8]
        name: identifier [4, 5] - [4, 8]
```

`keyword_select` is a child of `select` and should not be indented.

matthias-Q:
Thanks @lucario387 this works.  Even though, for some cases I do not understand why.

lucario387:
Lgtm, I'll push an extra commit to remove unnecessary queries, then merge this @matthias-Q

lucario387 (approved):

PR #4874, opened 2023-05-29
chore: update skipped langs in ci

amaanq:


clason:
Well, you gotta actually run the scripts and see if the updated parsers break...

PR #4878, opened 2023-05-30
highlights(ocaml): update operators

314eter:
Changes required because of tree-sitter/tree-sitter-ocaml#77.

### DIFF
--- a/queries/ocaml/highlights.scm
+++ b/queries/ocaml/highlights.scm
@@ -47,11 +47,11 @@
 
 (infix_expression
   left: (value_path (value_name) @function)
-  (infix_operator) @operator
+  operator: (concat_operator) @operator
   (#eq? @operator "@@"))
 
 (infix_expression
-  (infix_operator) @operator
+  operator: (rel_operator) @operator
   right: (value_path (value_name) @function)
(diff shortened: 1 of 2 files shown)


queries/ocaml/highlights.scm
```diff
@@ -47,11 +47,11 @@
 
 (infix_expression
   left: (value_path (value_name) @function)
-  (infix_operator) @operator
+  operator: (concat_operator) @operator
   (#eq? @operator "@@"))
```
lucario387, line 51: Is this ever needed now, considering `(concat_operator)` is already captured with the new changes

lucario387: I'd suggest in the case this isn't necessary anymore (which looks like the case), please use `@_...` capture instead of `@operator`, so the same node isn't captured twice, or remove it entirely

queries/ocaml/highlights.scm
```diff
@@ -47,11 +47,11 @@
...
 
 (infix_expression
-  (infix_operator) @operator
```
lucario387, line 54: Also this

lucario387 (approved):
LGTM :smile:

PR #4880, opened 2023-05-31
Bash: add "select" highlighting

fbearoff:
"select" is an opening keyword for select statements which are characterized as "for_statement" nodes per the bash language grammar. This adds the corresponding highlight for it under the @repeat group


ObserverOfTime (approved):

theHamsta:
Thank you!

PR #4883, opened 2023-05-31
chore(readme): update deprecated query functions

a-kenji:


lucario387:
Tyvm, but nvim-treesitter requirements are still 0.8+, and not yet 0.9+

Instead, you can use `require("nvim-treesitter.compat").get_query` for the time being

clason:
We've now bumped the requirement, but won't make any changes to adjust until the "big rewrite" for 0.10 (on `main`).

Thanks, though!

PR #4886, opened 2023-06-01
fix(make/highlights): add `define` and operators

tmillr:


queries/make/highlights.scm
```diff
@@ -38,17 +38,35 @@
...
+
+
+(define_directive
+ "define" @keyword
+ name: (word) @symbol
+ [
+  "="
+  ":="
+  "::="
+; ":::="
```
ObserverOfTime, line 50: What is `:::=`?

tmillr: [Immediately Expanded Variable Assignment Operator](https://www.gnu.org/software/make/manual/make.html#Immediate-Assignment). I had to comment it out because otherwise the query does not parse without erroring (the current `make` parser doesn't recognize it).

tmillr:
## Before

Before Screenshot

## After

After Screenshot

ObserverOfTime (approved):

ObserverOfTime:
Thank you for the contribution.

Issue #3061, opened 2022-06-24
[C/C++] Inconsistent highlighting of value and reference parameters


mliszcz:
### Describe the highlighting problem

I'm coming from: https://github.com/tree-sitter/tree-sitter-cpp/issues/164
Please see that one for some analysis and a potential solution.

Let me copy-paste the original report:

> Value parameters and reference parameters have different syntax highlight groups. Consider this example:
>
> (screenshot here)
>
> According to TSCaptureUnderCursor:
>   * arg1 is {TSVariable, TSParameter} in that order
>   * arg2 is {TSParameter, TSVariable} (order is reversed)
>
> The colorscheme I'm using indeed has different colors for TSVariable (fg) and TSParameter (red).
>
> I think that this inconsistency is not desired. This seems to be something that should be addressed by the parser, not the colorscheme.

### Example snippet that causes the problem

```cpp
void foo(int arg1, int& arg2, int const& arg3, int* arg4);
```

### Tree-sitter parsing result

```
declaration [0, 0] - [0, 58]
  type: primitive_type [0, 0] - [0, 4]
  declarator: function_declarator [0, 5] - [0, 57]
    declarator: identifier [0, 5] - [0, 8]
    parameters: parameter_list [0, 8] - [0, 57]
      parameter_declaration [0, 9] - [0, 17]
        type: primitive_type [0, 9] - [0, 12]
        declarator: identifier [0, 13] - [0, 17]
      parameter_declaration [0, 19] - [0, 28]
        type: primitive_type [0, 19] - [0, 22]
        declarator: reference_declarator [0, 22] - [0, 28]
          identifier [0, 24] - [0, 28]
      parameter_declaration [0, 30] - [0, 45]
        type: primitive_type [0, 30] - [0, 33]
        type_qualifier [0, 34] - [0, 39]
        declarator: reference_declarator [0, 39] - [0, 45]
          identifier [0, 41] - [0, 45]
      parameter_declaration [0, 47] - [0, 56]
        type: primitive_type [0, 47] - [0, 50]
        declarator: pointer_declarator [0, 50] - [0, 56]
          declarator: identifier [0, 52] - [0, 56]
```

### Example screenshot

![image](https://user-images.githubusercontent.com/6927259/175542680-45741073-c74a-4ed4-bbd8-fb7838dd437d.png)

### Expected behavior

arg1, arg2, arg3, arg4 all highlighted in red.

[...]

PR #4888, opened 2023-06-01
fix(c): default priority for identifier

### LINKED ISSUE #4887: cpp highlight in markdown incorrect

In a markdown file, if I write some cpp code, the `@text.literal.markdown` highlight overrides `@variable.cpp`, so the highlighting is off.
image
bob should be red
`:Inspect` - 
image

````markdown
```cpp
Person bob = {};
bob.name = "bob";
```
````

```
section [0, 0] - [4, 0] markdown
  fenced_code_block [0, 0] - [4, 0] markdown
    fenced_code_block_delimiter [0, 0] - [0, 3] markdown
    info_string [0, 3] - [0, 6] markdown
      language [0, 3] - [0, 6] markdown
    block_continuation [1, 0] - [1, 0] markdown
    code_fence_content [1, 0] - [3, 0] markdown
      declaration [1, 0] - [1, 16] cpp
        type: type_identifier [1, 0] - [1, 6] cpp
        declarator: init_declarator [1, 7] - [1, 15] cpp
          declarator: identifier [1, 7] - [1, 10] cpp
          "=" [1, 11] - [1, 12] cpp
          value: initializer_list [1, 13] - [1, 15] cpp
            "{" [1, 13] - [1, 14] cpp
            "}" [1, 14] - [1, 15] cpp
        ";" [1, 15] - [1, 16] cpp
      expression_statement [2, 0] - [2, 17] cpp
        assignment_expression [2, 0] - [2, 16] cpp
          left: field_expression [2, 0] - [2, 8] cpp
            argument: identifier [2, 0] - [2, 3] cpp
            operator: "." [2, 3] - [2, 4] cpp
            field: field_identifier [2, 4] - [2, 8] cpp
          operator: "=" [2, 9] - [2, 10] cpp
          right: string_literal [2, 11] - [2, 16] cpp
            """ [2, 11] - [2, 12] cpp
            """ [2, 15] - [2, 16] cpp
        ";" [2, 16] - [2, 17] cpp
      "=" [1, 11] - [1, 12] markdown
      "{" [1, 13] - [1, 14] markdown
      "}" [1, 14] - [1, 15] markdown
      ";" [1, 15] - [1, 16] markdown
      block_continuation [2, 0] - [2, 0] markdown
      "." [2, 3] - [2, 4] markdown
      "=" [2, 9] - [2, 10] markdown
      """ [2, 11] - [2, 12] markdown
      """ [2, 15] - [2, 16] markdown
      ";" [2, 16] - [2, 17] markdown
      block_continuation [3, 0] - [3, 0] markdown
    fenced_code_block_delimiter [3, 0] - [3, 3] markdown
```

### Example screenshot

image

colorscheme: https://github.com/olimorris/onedarkpro.nvim

### Expected behavior

`bob` should be highlighted red, not grayed out

```markdown

Installation ~
- OK `node` found v20.2.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.

OS Info:
{
  machine = "arm64",
  release = "21.6.0",
  sysname = "Darwin",
  version = "Darwin Kernel Version 21.6.0: Thu Mar  9 20:10:19 PST 2023; root:xnu-8020.240.18.700.8~1/RELEASE_ARM64_T8101"
} ~

  - c_sharp             ✓ ✓ ✓ . ✓
  - glsl                ✓ ✓ ✓ ✓ ✓
  - java                ✓ ✓ ✓ ✓ ✓
  - latex               ✓ . ✓ . ✓
  - wgsl                ✓ . ✓ ✓ .

```

```text
NVIM v0.10.0-dev-415+g55f6a1cab-Homebrew
Build type: Release
LuaJIT 2.1.0-beta3

  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/HEAD-55f6a1c/share/nvim"

```

### Additional context

Only happens with cpp as far as I've experienced.

clason:
Different fix for #3061
Fixes #4887

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -1,5 +1,4 @@
-; Lower priority to prefer @parameter when identifier appears in parameter_declaration.
-((identifier) @variable (#set! "priority" 95))
+(identifier) @variable
 
 [
   "default"
@@ -189,7 +188,7 @@
   declarator: (identifier) @parameter)
 
 (parameter_declaration
-  declarator: (pointer_declarator) @parameter)
+  declarator: (pointer_declarator) @parameter (#set! "priority" 110))
 
 (preproc_params (identifier) @parameter)
 
--- a/tests/indent/common.lua
+++ b/tests/indent/common.lua
@@ -74,7 +74,7 @@ assert:register(
 local function compare_indent(before, after, xfail)
   assert:add_formatter(format_indent)
   if xfail then
-    io.stdout:write "Warning! Known failure of this test! Please help to fix it! "
+    -- io.stdout:write "Warning! Known failure of this test! Please help to fix it! "
     assert.is_not.same_indent(before, after)
   else
     assert.is.same_indent(before, after)


clason:
Alternative would be to raise the priority of the desired captures.

queries/c/highlights.scm
```diff
@@ -191,6 +190,9 @@
 (parameter_declaration
   declarator: (pointer_declarator) @parameter)
 
+(parameter_declaration
+  declarator: (pointer_declarator (identifier) @parameter))
```
theHamsta, line 194: the problem with this is that pointers can nest arbitrarily. `*foo` `**foo` `***foo` `&foo`, `&&foo`.

`has_ancestor` was introduced for these and we had queries until nesting level 2 before. We successfully removed that costly operation by lowering the priority of `@variable`. I wouldn't want to change that just for issues with injections.

I think this is rather general issues with injections where there some situation where you want to have the injection having priority over everything in the parent language. Do we need `literal` highlighting in markdown when there's an injection. Probably the `literal` highlighting should have a very low priority when it's meant that injections render over it.

clason: `& foo` and all variations thereof already render correctly due to precedence; it's just `* foo` (and higher orders) that don't.

I would prefer to keep modifications within the language, if possible (since this is a C-specific issue; other injections have no problem).

Otherwise your suggestion works, too; we just need to make sure to keep high enough priority to avoid unwanted issues with other highlights (legacy syntax if enabled etc.)

clason: I tried raising the priority of the pointer declarator instead; this seems to highlight every variant your mention correctly.

theHamsta: Wouldn't that require to also raise the priority of the `*` operator? The idea of lowering variable vs increasing pointer declarator was that the operators would still paint over the actual to large declarator range. But if raising operators too would solve the problem we can of course do it.

In general, I'd argue that this isn't a C issue but a markdown one. It will happen for all code injections with languages with different priorities. Similar issues can also happen when injecting C into C macros.

In general, `(identifier) @variable` is so unspecific that it will also compete with other highlights and would then again just get it's priority over type identifiers (etc) only via ordering.

clason: > Wouldn't that require to also raise the priority of the * operator? The idea of lowering variable vs increasing pointer declarator was that the operators would still paint over the actual to large declarator range. But if raising operators too would solve the problem we can of course do it.

I didn't see an issue with the operators; the original issue was only about the arguments themselves?

(I can of course do that, too, but I'd prefer to only make changes that are _necessary_ for the issue at hand.)

> In general, I'd argue that this isn't a C issue but a markdown one. It will happen for all code injections with languages with different priorities. Similar issues can also happen when injecting C into C macros.

Actually, the issue is even more general, with _any_ injection into a node that also has highlights defined (as you point out). So if we want to solve this on that level (which is a perfectly reasonable idea), we need a more general solution; preferably one that doesn't involve having to reason about all possible combinations.

(This is what I'm worried about here: having to keep in mind all possible priorities of the injected language. So there needs to be one(!) priority for injection targets that is the same and works for all languages, so we can document it.)

> In general, (identifier) @variable is so unspecific that it will also compete with other highlights and would then again just get it's priority over type identifiers (etc) only via ordering.

Yes, but that is precisely what precedence by ordering is for?

clason:
>  It will happen for all code injections with languages with different priorities.

Do we have a list of injections where this is an issue, so we have some idea what we're looking at here?

* markdown `fenced_code_block`
* vimdoc `codeblock`
* latex `generic_environment` (?)
* c `preproc_argument`
* ...

(feel free to edit this comment to add more)

queries/c/highlights.scm
theHamsta: > (I can of course do that, too, but I'd prefer to only make changes that are necessary for the issue at hand.)

I agree, but `@variable` was never really supported by nvim-treesitter (was treated as unset highlight). It's possible that it has more problems on a global level and language injections in markdown seem like a side-issue when compared to issues that might be still in the actual languages. It could wait until we revise `@variable` globally and see that is works well and similar in most languages.

> I didn't see an issue with the operators; the original issue was only about the arguments themselves?

The solution with highlighting `pointer_declarator` as parameter has the problem that it will hightlight all `int* foo` as parameter instead of just `foo` which relies on `*` and `int` overwriting the highlight. So, we have to be a bit careful when raising prio of `pointer_declarator`-

> Yes, but that is precisely what precedence by ordering is for?

If it works, I'm fine with the change. We just need to do a smoke test that everything still works as intended.

theHamsta:
> > It will happen for all code injections with languages with different priorities.
> 
> Do we have a list of injections where this is an issue, so we have some idea what we're looking at here?
> 
>     * markdown `fenced_code_block`
> 
>     * vimdoc `codeblock`
> 
>     * latex `generic_environment` (?)
> 
>     * c `preproc_argument`
> 
>     * ...
> 
> 
> (feel free to edit this comment to add more)

Would you see any problem to make sure an individual languages have a certain tree-sitter priority range that they never underflow or overflow? Could we offset injection with a priority offset by that amount? Could this conflict with other priorities of users or highlight providers? I think it would be difficult to reason the priority of highlights of one language with another. Are there situations where the parent language would need to have a higher priority?

Offsetting language priorities could have disadvantages as mentioned above. But with issue at hand of injecting languages into markdown: do you think that languages should not go with their priority below 100 and only increase? Because otherwise there might be the same issue in all languages that use sub-100 prios (which might be ok, since those low prio highlights are rarer than `@variable`)

> Do we have a list of injections where this is an issue, so we have some idea what we're looking at here?

>    markdown fenced_code_block
>    vimdoc codeblock
>   latex generic_environment (?)
>    c preproc_argument
>  ...

We could decide this by capture type and treat all `@text.literal` the same.

queries/c/highlights.scm
clason: >  It's possible that it has more problems on a global level and language injections in markdown seem like a side-issue when compared to issues that might be still in the actual languages.

Well, not side-issue, exactly, but I agree that these are two different but interacting issues. Let's proceed with your suggestion then (lowering the priority of code blocks with possible injections).

clason:
I don't like the idea of "language priorities" either and would prefer to handle this on the level of individual captures. `@text.literal` is too broad for that; I'd suggest making this a new (sub)capture where we document that these should have priority (say) 90 to allow injected languages to override them (making sure that "real" captures have priority > 90, which I believe is the case).

This means this ties into the capture renaming plan, though.

queries/c/highlights.scm
theHamsta: increasing the prio of `@variable` for C (or just make it the same for all languages) could help to be more inline with other languages. We should try to apply both measures long term (not in a single PR)

theHamsta:
> I don't like the idea of "language priorities" either and would prefer to handle this on the level of individual captures. `@text.literal` is too broad for that; I'd suggest making this a new (sub)capture where we document that these should have priority (say) 90 to allow injected languages to override them (making sure that "real" captures have priority > 90, which I believe is the case).
> 
> This means this ties into the capture renaming plan, though. (But of course we can use a temporary capture like `@text.literal.codeblock` for now.)

Yeah, of course. I meant only `@text.literal` with injection. Meant this more as a grep criterion than by-capture policy that Neovim could apply, though such a policy could be possible for ``@text.literal.codeblock`

clason:
I wouldn't automate it; just do it manually for the relevant captures and add this to the capture documentation.

clason:
superseded by #4905

PR #4897, opened 2023-06-02
injections(query): add `#gsub!` pattern

lucario387:
Example: 

image

### DIFF
--- a/runtime/queries/query/injections.scm
+++ b/runtime/queries/query/injections.scm
@@ -12,5 +12,12 @@
  (#set! injection.language "luap")
  (#offset! @injection.content 0 1 0 -1))
 
+((predicate
+   name: (identifier) @_name
+   parameters: (parameters (string) @injection.content . (string) .))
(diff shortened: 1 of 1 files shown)


clason:
I'm always a bit concerned about the performance of these complex injections; could you say something about that?

lucario387:
I cannot really say anything, how exactly should/can I even perf these queries, though?

clason:
We need benchmarks... But for now, just anecdotal evidence ("I tested editing a reasonably large query before and after and didn't notice any lag") is sufficient.

theHamsta (approved):

clason:
@lucario387 a number of indentation tests are failing on the branch (unrelated to this PR) -- could you take a look at those when you have time?

amaanq:
I don't think performance would be an issue, is anyone writing thousand line query files with many potential injection candidates? ;) on a related note, a good consistent benchmarking ci tool would be nice...

PR #4898, opened 2023-06-03
feat(parsers): use gdamore/tree-sitter-d parser for D

ljmf00:
CC @gdamore @nawordar

This tree-sitter parser is better maintained (manual parser field names) and is now mentioned officially by the tree-sitter org. 

Perhaps, are you comfortable co-maintaining this with me @nawordar ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [css](https://github.com/tree-sitter/tree-sitter-css) (maintained by @TravonteD)
 - [x] [cuda](https://github.com/theHamsta/tree-sitter-cuda) (maintained by @theHamsta)
 - [x] [cue](https://github.com/eonpatapon/tree-sitter-cue) (maintained by @amaanq)
-- [x] [d](https://github.com/CyberShadow/tree-sitter-d) (experimental, maintained by @nawordar)
+- [x] [d](https://github.com/gdamore/tree-sitter-d) (maintained by @ljmf00)
 - [x] [dart](https://github.com/UserNobody14/tree-sitter-dart) (maintained by @akinsho)
 - [x] [devicetree](https://github.com/joelspadin/tree-sitter-devicetree) (maintained by @jedrzejboczar)
 - [x] [dhall](https://github.com/jbellerb/tree-sitter-dhall) (maintained by @amaanq)
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -289,13 +289,10 @@ list.cue = {
 
 list.d = {
   install_info = {
-    url = "https://github.com/CyberShadow/tree-sitter-d",
-    files = { "src/parser.c", "src/scanner.cc" },
-    requires_generate_from_grammar = true,
+    url = "https://github.com/gdamore/tree-sitter-d",
+    files = { "src/parser.c", "src/scanner.c" },
   },
-  -- Generating grammar takes ~60s
-  experimental = true,
-  maintainers = { "@nawordar" },
+  maintainers = { "@ljmf00" },
 }
 
 list.dart = {
--- a/queries/d/highlights.scm
+++ b/queries/d/highlights.scm
@@ -1,21 +1,75 @@
+;; Variables
+(identifier) @variable
+
 ;; Misc
 
-[
-  (line_comment)
-  (block_comment)
-  (nesting_block_comment)
-] @comment @spell
+(comment) @comment @spell
 
-((line_comment) @comment.documentation
+((comment) @comment.documentation
   (#lua-match? @comment.documentation "^///[^/]"))
-((line_comment) @comment.documentation
+
+((comment) @comment.documentation
   (#lua-match? @comment.documentation "^///$"))
 
-((block_comment) @comment.documentation
+((comment) @comment.documentation
   (#lua-match? @comment.documentation "^/[*][*][^*].*[*]/$"))
 
-((nesting_block_comment) @comment.documentation
+((comment) @comment.documentation
   (#lua-match? @comment.documentation "^/[+][+][^+].*[+]/$"))
+
+(directive) @preproc
+(shebang) @preproc
+
+;; Operators
+
+[
+  "/="
+  "/"
+  ".."
+  "..."
+  "&"
+  "&="
+  "&&"
+  "|"
+  "|="
+  "||"
+  "-"
+  "-="
+  "--"
+  "+"
+  "+="
+  "++"
+  "<"
+  "<="
+  "<<"
+  "<<="
+  ">"
+  ">="
+  ">>="
+  ">>>="
+  ">>"
+  ">>>"
+  "!"
+  "!="
+  "?"
+  "$"
+  "="
+  "=="
+  "*"
+  "*="
+  "%"
+  "%="
+  "^"
+  "^="
+  "^^"
+  "^^="
+  "~"
+  "~="
+  "@"
+  "=>"
+] @operator
+
+;; Punctuation
 
 [
   "(" ")"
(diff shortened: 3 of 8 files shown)


nawordar:
I'm not using D that much anymore, so I would be fine if I was delisted as a maintainer.

theHamsta:
The old parser took over a minute to compile (probably also had impact on runtime). So a change is welcome. I didn't have time to test the new parser yet.

clason:
As long as the queries look OK, I think that's a strict improvement (even if a breaking change, so please mark it as such) over the old, experimental, parser, so I'd be happy to merge.

queries/d/highlights.scm
```diff
@@ -1,22 +1,79 @@
...
+((comment) @comment.documentation
   (#lua-match? @comment.documentation "^/[+][+][^+].*[+]/$"))
 
+(directive) @preproc
+(shebang) @preproc
+
+;; Operators
+
+; these are listed first, because they override keyword queries
+(identity_expression (in) @operator)
```
theHamsta, line 23: `is`/`in` are they literal `in`/`is` in the language? Then perhaps `@keyword.operator`? What about `not_in`

ljmf00: Done

README.md
```diff
@@ -199,7 +199,7 @@ We are looking for maintainers to add more parsers and to write query files for
 - [x] [css](https://github.com/tree-sitter/tree-sitter-css) (maintained by @TravonteD)
 - [x] [cuda](https://github.com/theHamsta/tree-sitter-cuda) (maintained by @theHamsta)
 - [x] [cue](https://github.com/eonpatapon/tree-sitter-cue) (maintained by @amaanq)
-- [x] [d](https://github.com/CyberShadow/tree-sitter-d) (experimental, maintained by @nawordar)
+- [x] [d](https://github.com/gdamore/tree-sitter-d) (maintained by @nawordar and @ljmf00)
```
clason, line 202:
```suggestion
- [x] [d](https://github.com/gdamore/tree-sitter-d) (maintained by @ljmf00)
```

lua/nvim-treesitter/parsers.lua
```diff
@@ -289,13 +289,11 @@ list.cue = {
...
-    files = { "src/parser.c", "src/scanner.cc" },
-    requires_generate_from_grammar = true,
+    url = "https://github.com/gdamore/tree-sitter-d",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
   },
-  -- Generating grammar takes ~60s
-  experimental = true,
-  maintainers = { "@nawordar" },
+  maintainers = { "@nawordar", "@ljmf00" },
```
clason, line 296:
```suggestion
  maintainers = { "@ljmf00" },
```

We should take maintainer status seriously; if you are listed, you will get pinged anytime there is an issue with the language -- be it a reported issue, a PR to the queries, or a breaking lockfile update -- and are expected to react in a timely manner.

clason (changes requested):

queries/d/highlights.scm
```diff
@@ -35,254 +86,194 @@
...
+(function_declaration (identifier) @function)
 
-(class_declaration
-  (identifier) @type
-)
+(call_expression (identifier) @function.call)
+(call_expression (type (identifier) @function.call))
 
-(fundamental_type) @type.builtin
+;; Types
```
theHamsta, line 239: apparently `(type (identifier) @type)` is missing. I don't see any type highlights and also couldn't find `@type` in this `highlighs.scm`

theHamsta: strangely a function name in a call is also a `(type (identifier) @type)`

lua/nvim-treesitter/parsers.lua
ljmf00: Yes, its ok. I contribute a lot to D and nvim is my daily driver. I prefer for someone to directly ping me, because looking at the PRs is not feasible to me, but I have an RSS feed for whenever someone mentions me on github, even without a direct ping, or mentions `tree-sitter-d` or other relevant keywords for the projects I maintain. @gdamore is also my work colleague, so I can possibly cooperate with him if anything is related to the upstream.

clason: We need to look into automating that, as it's also not really feasible to manually ping maintainers for the 190+ languages we have...

gdamore: I don't use nvim, or I'd probably have done this myself -- thanks @ljmf00 for doing it.

I'm happy to help out.  Especially with the queries - though I'm not aware of the specific nvim selectors that are relevant.

clason: They're documented in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md (but we'll rework them a bit to be more in line with upstream/Helix in the near future).

ljmf00: There's a way to automatically add them via CODEOWNERS and also use github-actions with an heuristic to automatically mention possible maintainers, like based on the tree-sitter link or add a PR/issue template with a dropdown with the language and then map the language to a person.

queries/d/highlights.scm
gdamore: Grammatically in D -- the distinction between a function call and a type is not something you can easily distinguish using syntax alone in many circumstances.   (I don't love D's grammar, but I did the tree-sitter plugin because I'm forced to work in it at work.)

queries/d/highlights.scm
```diff
@@ -35,254 +86,194 @@
...
-  "&"
-  "&&"
-] @operator
-
-[
-  "catch"
-  "finally"
-  "throw"
-  "try"
-] @exception
```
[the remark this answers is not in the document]
gdamore: The @exception selector is missing in the other editor plugins I've done -- but it would be easy to fix that here by moving those words into a block from where they are located.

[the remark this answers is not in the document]
gdamore: Oooh, this looks like an error of omission.

lua/nvim-treesitter/parsers.lua
clason: CODEOWNERS only works for actual org members, so we'd have to add every (community) maintainer to the team. There are options with github actions, as you're mentioning; we are looking into the different options here (as the list of maintainers itself has to be maintained actively).

gdamore: > They're documented in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md (but we'll rework them a bit to be more in line with upstream/Helix in the near future).

I wrote the helix plugin for D. So when that happens it will be easier here.  Having said that it's possible that I am missing some there too - I will check on it.

clason:
@ljmf00 ping on the requested changes

README.md
ljmf00: Done

lua/nvim-treesitter/parsers.lua
ljmf00: Done

queries/d/highlights.scm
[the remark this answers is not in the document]
ljmf00: Added. Thanks for noticing!

lua/nvim-treesitter/parsers.lua
```diff
@@ -289,13 +289,11 @@ list.cue = {
 
 list.d = {
   install_info = {
-    url = "https://github.com/CyberShadow/tree-sitter-d",
-    files = { "src/parser.c", "src/scanner.cc" },
-    requires_generate_from_grammar = true,
+    url = "https://github.com/gdamore/tree-sitter-d",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
```
clason, line 294: If `main` is the default branch, this can be omitted.

queries/d/highlights.scm
ljmf00: Done

lua/nvim-treesitter/parsers.lua
ljmf00: I had some issues loading it on my nvim without it

clason: That doesn't sound like the actual issue; the problem was more likely (since it usually is) the stale lockfile commit hash.

Try dropping it.

queries/d/highlights.scm
ljmf00: I'm still not able to get types highlighted :/

ljmf00:
Don't merge, see https://github.com/nvim-treesitter/nvim-treesitter/pull/4898#discussion_r1230149947 .

ljmf00:
@gdamore I think your some of your highlights queries are wrong, or perhaps `nvim-treesitter` treats queries differently. This seem to override every identifier https://github.com/gdamore/tree-sitter-d/blob/main/queries/helix-highlights.scm#L230 .

@clason Look's like it's working correctly now:
image
Source code from DMD (`expression.d` file)

clason:
Yes, nvim-treesitter handles query order differently than upstream (last query wins, since all are applied in order, rather than first matching query short-circuits).

ljmf00:
> Yes, nvim-treesitter handles query order differently than upstream (last query wins, since all are applied in order, rather than first matching query short-circuits).

Is there any reason to do so? It seems to me that short circuiting is more efficient to me.

clason:
The reason is that this is easier to implement (and hence more efficient) and easier to reason about. Even upstream agrees now that this is better.

ljmf00:
Just squash to `feat!()` to mark as breaking change.

clason (approved):

amaanq:
@ljmf00 I made a lot of changes locally to the highlights to make it better (a bit too many to suggest changes here), would you mind if I force push those changes here for you to try out and review?

clason:
ping @ljmf00 are you willing to make the changes needed to avoid the regressions?

clason:
That's fine, there's no hurry, I just wanted to check whether you were still interested in this. (But if you are so busy, you probably want to look for a second maintainer as we do expect a reasonable reaction time in case of breakages.)

ljmf00:
> That's fine, there's no hurry, I just wanted to check whether you were still interested in this. (But if you are so busy, you probably want to look for a second maintainer as we do expect a reasonable reaction time in case of breakages.)

I am. I use this everyday on my nvim, as I develop in D including contributions to D language itself, but my time has been very limited, and I suppose my colleague @gdamore , who wrote this grammar implementation, too. If you want to take it @gdamore , just suggest some superseded PR and tell to drop this one. I added this to my personal task list and prioritize it, but it's been difficult to have some time for it.

clason:
Superseded by #5988 

Thanks for getting this started!

PR #4900, opened 2023-06-03
tests: fix tests on main

lucario387:
It is not a recognized filetype in Neovim, so adding/fixing the test case to the original author's desired intention

### DIFF
--- a/tests/indent/cpp_spec.lua
+++ b/tests/indent/cpp_spec.lua
@@ -7,7 +7,6 @@ local run = Runner:new(it, 'tests/indent', {
   shiftwidth = 4,
   softtabstop = 0,
   expandtab = true,
-  filetype = 'cpp',
 })
 
 describe('indent C++:', function()
--- a/tests/indent/javascript_spec.lua
+++ b/tests/indent/javascript_spec.lua
@@ -6,7 +6,6 @@ local run = Runner:new(it, 'tests/indent', {
   shiftwidth = 2,
   softtabstop = 0,
   expandtab = true,
-  filetype = 'javascript',
 })
 
 describe('indent JavaScript:', function()
--- a/tests/indent/jsx_spec.lua
+++ b/tests/indent/jsx_spec.lua
@@ -3,7 +3,6 @@ local runner = Runner:new(it, 'tests/indent/jsx', {
   tabstop = 2,
   shiftwidth = 2,
   expandtab = true,
-  filetype = 'jsx',
 })
 
 describe('indent JSX Elements:', function()


lucario387:
Oops, this should be on main
...

lucario387:
Not yet working ...

lucario387:
the test suite is running with neovim stable, while the code is in neovim nightly @clason

clason:
> the test suite is running with neovim stable, while the code is in neovim nightly @clason

No, that's not correct, it uses nightly -- but it also relies on changes from today's Neovim master, which won't be in the nightly release until tomorrow.

lucario387:
Some tests still failed, I'll have a look later today

scripts/minimal_init.lua
```diff
@@ -10,14 +10,17 @@ vim.filetype.add({
...
     usd = 'usd',
     usda = 'usd',
+    wgsl = 'wgsl',
   },
 })
 
 vim.o.swapfile = false
 vim.bo.swapfile = false
 
+vim.treesitter.language.register('javascript', 'javascriptreact')
```
clason, line 20: Hmm, that should already be done in `plugin/filetypes.lua`.

clason:
Looks like there's issues with the "inherited" languages (Javascript, Cpp). Will have to check whether (and why) the inherited queries are not installed.

lucario387:
From what I can see with the inherited languages, the lnum are all off by 1, which .. is interesting, to say the least

clason (approved):

clason:
bash injection tests are failing now -- related to the new injections?

lucario387:
Ahhhh, it's not you messing up, it's because there's a children containing the node text, so on `master`, `include-children` is there by default

lucario387:
Can I fix the queries with this branch as well? @clason

clason:
Sure, but I will probably want to squash that commit with your original one (for backporting).

lucario387:
@clason Can you apply this patch instead? That will be easier for both of us

```diff
diff --git a/runtime/queries/yaml/injections.scm b/runtime/queries/yaml/injections.scm
index 3939d403..9d174bfb 100644
--- a/runtime/queries/yaml/injections.scm
+++ b/runtime/queries/yaml/injections.scm
@@ -5,7 +5,8 @@
 (block_mapping_pair
   key: (flow_node) @_run (#any-of? @_run "run" "script" "before_script" "after_script")
   value: (flow_node
-           (plain_scalar) @injection.content
+           (plain_scalar 
+             (string_scalar) @injection.content)
            (#set! injection.language "bash")))
 
 (block_mapping_pair
@@ -20,7 +21,9 @@
   value: (block_node
            (block_sequence
              (block_sequence_item
-                (flow_node) @injection.content
+                (flow_node
+                  (plain_scalar 
+                    (string_scalar) @injection.content))
                 (#set! injection.language "bash")))))
 
 (block_mapping_pair
@@ -37,7 +40,8 @@
 (block_mapping_pair
   key: (flow_node) @_expr (#eq? @_expr "expr")
   value: (flow_node
-           (plain_scalar) @injection.content
+           (plain_scalar
+             (string_scalar) @injection.content)
            (#set! injection.language "promql")))
 
 (block_mapping_pair
@@ -52,7 +56,9 @@
   value: (block_node
            (block_sequence
              (block_sequence_item
-               (flow_node) @injection.content
+               (flow_node
+                 (plain_scalar
+                   (string_scalar) @injection.content))
                (#set! injection.language "promql")))))
 
 (block_mapping_pair
```

PR #4901, opened 2023-06-03
tests: fix failed tests

lucario387:
Similar to #4900 but for master, testing to see why something works/not work

These filetypes were not recognized, so the tests were always marked with `XFAIL`, because they never run.

Therefore, Removing XFAIL for most of the tests, alongside adding ft to these unrecognized extensions

### DIFF
--- a/queries/t32/indents.scm
+++ b/queries/t32/indents.scm
@@ -6,7 +6,6 @@
 ; Control flow statements
 (
   (if_block
-    _*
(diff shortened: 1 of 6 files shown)


lucario387:
@clason This is the same as #4900, but for master, that can be rebased to main. Some other issues that is specific to `main` will be pushed there

clason (approved):

PR #4916, opened 2023-06-06
Add if_statement to PHP indents.scm

rsynnest:
EDIT: this PR is incomplete, PHP `if`/`else`/`else_if` indentation is still kinda broken, will update if I can put together a fix. 

This PR fixes the indentation issue seen in the following if statement:

```php
<?php
if (true) {
    if (empty($a)) {
        echo "test";
        if (
        true                  // <----------- should be indented by 1 more
            || false
            || false
    ) {                       // <----------- should be indented by 1 more
            echo "hello";
        }
    } else {
        echo "elsebranch";
    }
}
```

Caveat: I'm very new to Treesitter queries, but ran this through some example files and didn't see any problems.

### DIFF
--- a/queries/php/indents.scm
+++ b/queries/php/indents.scm
@@ -13,6 +13,11 @@
   "["
 ] @indent.begin
 
+(
+ if_statement
+  condition: (_) @indent.begin
+)
+
 [
   ")"
   "}"


lucario387:
PR welcome! Upon first look, you can probably copy the `if/else` logic from `queries/c/indents.scm` for a test run, they seem to be the same structurally

tk-shirasaka:
Thanks for the PR!
After applying your modification, it seems to me that the `binary_expression` applies double indentation when using a compound condition and a new line.

e.g.
```php
if (
    $foo === 1 ||
        $bar === 2
) {
}
```

Please consider changing `binary_expression` in `@indent.begin` to `assignment_expression`.

Also, the following definitions may be necessary since the indentation seems to be broken when using `elseif`.
```query
(
 if_statement
  alternative: (else_if_clause) @indent.begin
)
```

This is not the purpose of this fix, but since there is unintended indentation after the end of the class, it would be nice if you could add the following definition.
```query
(declaration_list "}" @indent.end)
```

calebdw:
@lucario387,@clason this was fixed in https://github.com/nvim-treesitter/nvim-treesitter/pull/6293 so this can be closed

PR #4919, opened 2023-06-07
feat(lua): update highlights query for functions

MunifTanjim:
Make functions higlight a bit nicer. These will now be highlighted as function:

```lua
local func_one = function() end
--    ^

local tbl = {
  func_one = function() end,
  --  ^
}
```

---

On related note:

I just realized I can't test highlights in https://github.com/MunifTanjim/tree-sitter-lua using `tree-sitter test`, because upstream `tree-sitter` treats highlight queries differently than Neovim. Is there any plan to unify that?

### DIFF
--- a/queries/lua/highlights.scm
+++ b/queries/lua/highlights.scm
@@ -180,13 +180,40 @@
 
 (parameters (identifier) @parameter)
 
-(function_call name: (identifier) @function.call)
-(function_declaration name: (identifier) @function)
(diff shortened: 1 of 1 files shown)


clason:
> Is there any plan to unify that?

No plan from our side; we -- and upstream now -- prefer our approach. (I assume you refer to the precedence in ordering.)

clason (approved):

PR #4920, opened 2023-06-07
feat(scala): sync highlights with latest upstream

ckipp01:


clason:
Highlight query updates should still target `master` branch; it's only injections that we have switched to the new format that is not compatible with Neovim 0.8 and earlier.

clason:
Could I ask you to re-open this PR against the `master` branch?

ckipp01:
> Could I ask you to re-open this PR against the `master` branch?

Ah, ok, but I originally did open this against `master`. Then it was switch to main, then back to master, then again to main? I don't really understand what's going on here.

clason:
Yes, but you branched your PR from `main`, so I switched the base branch for the PR since I thought it was intentional (we do ask _injections_ changes to be made to `main`, since we already switched to the upstream format there).

(It was never switched back, but `main` gets frequent force-pushes and rebases, which might be confusing.)

ckipp01:
> we do ask injections changes to be made to main, since we already switched to the upstream format there)

Is this documented somewhere so contributors know this flow? I don't see it mentioned in the contributing docs unless I'm missing it.

clason:
No, it's not documented since it's temporary. We'll backport the changes once we drop 0.8 compatibility on the master branch.

PR #4932, opened 2023-06-08
feat(jsonnet): add "for", "function" keyword and update highlights for operators

folliehiyuki:
A simple Jsonnet snippet to verify this change:
```jsonnet
{
  local numbers = std.range(1, 10),
  a: [x ^ 2 for x in std.map(function(x) x + 2, numbers)],
  assert 5 % 2 == 1,
}
```

### DIFF
--- a/queries/jsonnet/highlights.scm
+++ b/queries/jsonnet/highlights.scm
@@ -30,11 +30,27 @@
   ":::"
 ] @punctuation.delimiter
 
-(expr
-  operator: (_) @operator)
+(unaryop) @operator
 [
   "+"
+  "-"
+  "*"
+  "/"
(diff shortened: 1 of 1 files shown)


queries/jsonnet/highlights.scm
```diff
@@ -30,18 +30,37 @@
...
 ] @operator
 
+"for" @repeat
+
 "in" @keyword.operator
 
 [
  (local)
  "assert"
+ "function"
```
amaanq, line 63: `@keyword.function`

FollieHiyuki: Done.

amaanq (approved):

PR #4933, opened 2023-06-08
update ts/ecma

amaanq:


queries/typescript/highlights.scm
```diff
@@ -52,6 +59,9 @@
 (type_parameters
   ["<" ">"] @punctuation.bracket)
 
+(object_type
+  ["{|" "|}"] @punctuation.bracket)
```
lucario387, line 63: This looking wrong, I don't remember object types having this syntax

amaanq: it's in the grammar

lucario387: Just found it

lucario387 (approved):

queries/typescript/highlights.scm
zweimach: @lucario387 is right about this. TypeScript doesn't have this syntax.
It is a syntax for [Flow](https://flow.org/en/docs/types/objects/#exact-and-inexact-object-types), and Flow (in Atom) uses the TS grammar.

zweimach: Thanks for the PR. I didn't know that I would need these additions. 👍🏽

PR #4935, opened 2023-06-09
add more support for glimmer

yads:
* add indents, locals and folds

### DIFF
--- a/queries/glimmer/indents.scm
+++ b/queries/glimmer/indents.scm
@@ -1,0 +1,22 @@
+[
+  (element_node (element_node_start))
+  (element_node_void)
+  (block_statement (block_statement_start))
+  (mustache_statement)
+] @indent.begin
+
+(element_node (element_node_end  [">"] @indent.end))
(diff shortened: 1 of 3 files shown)


queries/glimmer/indents.scm
```diff
@@ -0,0 +1,22 @@
...
+ "/>"
+ "</"
+ "{{/"
+ "}}"
+] @indent.branch
+
+(mustache_statement
+  (helper_invocation helper: (identifier) @_identifier (#lua-match? @_identifier "else"))
+  ) @indent.branch
+(mustache_statement ((identifier) @_identifier (#lua-match? @_identifier "else"))) @indent.branch
```
lucario387, line 21: Not familiar with the language, why does it have to be `lua-match?` and not `eq?` ?

yads: Yeah I guess it doesn't in this case. Thanks, I'll make the update here.

amaanq (approved):

PR #4941, opened 2023-06-11
highlights(commonlisp): include all symbols defined by the ANSI standard

theHamsta:
Before, those symbols were available via https://github.com/theHamsta/nvim-treesitter-commonlisp. Since then `#any-of?` has become more efficient via tables and clojure also added all builtin functions (https://github.com/theHamsta/nvim-treesitter/blob/72cbda75a901bf14a2eb98a47584fe5fc32689ab/queries/clojure/highlights.scm#L169-L278).

The list I'm adding is defined by the 1994 ANSI standard that has not changed since. This is why I don't expect any changes or maintenance burden due to adding that list.

Fixes https://github.com/theHamsta/tree-sitter-commonlisp/issues/18

### DIFF
--- a/queries/commonlisp/highlights.scm
+++ b/queries/commonlisp/highlights.scm
@@ -110,40 +110,9 @@
 (list_lit
  .
  (sym_lit) @function.macro
- ;; For a complete and more efficient version install theHamsta/nvim-treesitter-commonlisp
+ ;; Generated via https://github.com/theHamsta/nvim-treesitter-commonlisp/blob/22fdc9fd6ed594176cc7299cc6f68dd21c94c63b/scripts/generate-symbols.lisp#L1-L21
  (#any-of? @function.macro
-          "let"
-          "function"
-          "the"
-          "unwind-protect"
-          "labels"
-          "flet"
-          "tagbody"
-          "go"
-          "symbol-macrolet"
-          "symbol-macrolet"
-          "progn"
-          "prog1"
-          "error"
-          "or"
-          "and"
-          "defvar"
-          "defparameter"
-          "in-package"
-          "defpackage"
-          "case"
-          "ecase"
-          "typecase"
-          "etypecase"
-          "defstruct"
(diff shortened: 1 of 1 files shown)


queries/commonlisp/highlights.scm
```diff
@@ -156,22 +125,11 @@
...
...tion" "make-concatenated-stream" "defpackage" "class-of" "no-next-method" "logeqv" "deposit-field" "disassemble" "unuse-package" "copy-tree" "find" "asinh" "class-name" "rename-file" "values" "print-not-readable-object" "mismatch" "cadadr" "shadowing-import" "delete-if-not" "maplist" "listen" "return" "stream-element-type" "unintern" "merge" "make-synonym-stream" "prin1-to-string" "nsubst-if" "byte-position" "phase" "muffle-warning" "remhash" "continue" "load-time-value" "hash-table-size" "upgraded-complex-part-type" "char-lessp" "sbit" "upgraded-array-element-type" "file-length" "typecase" "cadddr" "first" "rationalize" "logtest" "find-if-not" "dpb" "mapc" "sinh" "char-greaterp" "shiftf" "denominator" "get-universal-time" "nconc" "setf" "lognand" "rename-package" "pprint-logical-block" "break" "symbol-macrolet" "the" "fresh-line" "clear-output" "assoc-if" "string/=" "princ" "directory-namestring" "stream-error-stream" "array-element-type" "setq" "copy-seq" "time" "restart-case" "prog*" "shared-initialize" "array-total-size" "simple-bit-vector-p" "define-method-combination" "write-byte" "constantly" "caddar" "print-object" "vector" "throw" "reverse" ">=" "upper-case-p" "nbutlast"))
```
theHamsta, line 133: pushed those all into a single line. since the list is auto-generated, I don't expect any value of reading it and instead let it occupy the least screen space possible.

amaanq (approved):

clason (approved):

Issue #4902, opened 2023-06-03
Cached and memoize values are incorrect when buffer changes without modifying b:changedtick


TheLeoP:
### Describe the bug

# General bug

Changing the contents of a neovim buffer without modifying `b:changedtick` results in wrong cached/memoize values being returned by nvim-treesitter.

# Particular example

I encountered this behaviour while using the `command-preview` feature of `nvim_create_user_command`. The content of the buffer is the same before and after the execution of the preview function, but the treesitter representation has changed. 

# Why this could be a problem

- Nodes in the current treesitter representation will have different `id`s to the nodes returned by memoized or cached functions (since they depend on `b:changedtick`).
- Comparisons like `vim.tbl_contains(some_list_of_nodes, some_node)` won't work as expected

### To Reproduce

1. Create the usercommand `:Example` with the following code (its important to note that the command does nothig, even tho it's preview function shows the buffer content being deleted)

```lua
vim.api.nvim_create_user_command("Example", function() end, {
	preview = function()
		local buf = vim.api.nvim_get_current_buf()
		vim.api.nvim_buf_set_lines(buf, 0, -1, true, {})
		return 1
	end,
})
```
2. Create the following keymap (it's important to note that this keymap is just for convenience, it isn't necesary to reproduce the bug):

```lua
vim.keymap.set("n", "<F5>", function()
	local ts_utils = require("nvim-treesitter.ts_utils")
	local ts_locals = require("nvim-treesitter.locals")
	local node = ts_utils.get_node_at_cursor(0)
	local buf = vim.api.nvim_get_current_buf()
	local definition = ts_locals.find_definition(node, buf)
	vim.print({ definition:range() })
end)
```

3. Open the file `example.lua` containing the following code:

```lua
-- example.lua
local function foo()
	local bar = "a"
	return bar
end

local bar = foo()

print(bar)
```

4. Put the cursor over the  `bar` identifier inside the `foo` function (line 3)
5. Press F5 (this will execute the keymap created before)
6. The range `{2,7,2,10}` is printed
7. Execute the `:Example` user command
8. With the cursor in the same position (over the  `bar` identifier inside the `foo` function (line 3)) press F5 again
9. The range `{6,6,6,9}` is printed (and will continued to be printed each time the function mapped to F5 gets executed until the value of `b:changedtick` changes)

### Expected behavior

 The range `{2,7,2,10}` should be printed on both situations.

### Output of `:checkhealth nvim-treesitter`

[...]

PR #4950, opened 2023-06-15
Make module `locals` work again (main branch)

TheLeoP:
As stated on #4902, the `locals` module wasn't working after the rewrite. This PR solves this.

I'm using `root:id` to cache the memoized values (to avoid the problem mentioned on #4902).

Also, I modified the behaviour of `find_usages`. Before (and also on the main branch), it only found definitions, I think (Evaluating `kind == nil or def_node == node` on the definition of a node to decide whether or not to add a node to the list of return values will only be true if the node is a definition or if a definition couldn't be found, so the same node gets inserted). I couldn't figure out if the previous behaviour was intended or not, at lest the name `find_usages` suggested a different behaviour in my mind. If this is wrong or causes issues, I'll revert this change.


lucario387:
You should `git pull --rebase` instead of fixing rebase like such

TheLeoP:
@lucario387 I did exactly that. But, there was a mess of conflicts. I thought I had solved them when I finished the rebase,  but it turns out I didn't (?). This was my first time using rebase :c

lucario387:
Then you should do a clean branch, `git pull upstream main --rebase --force` then re-apply the commits :D

TheLeoP:
Yeah, that makes sense. Thanks @lucario387. Should I close this PR and open a new one doing that?

lucario387:
You can do that on the same branch as well, just clear the commits, rebase, and then force-push back

clason:
Warning: the `main` branch gets frequent rebases and force-pushes. The best way to deal with them is `git rebase -i main`, and then `drop` all commits except your own.

clason:
Also, you should never make PRs from the base branch. Always `git branch fix/locals` first, then add commits there. Together with the merge, this PR is unusable for me unfortunately.

So please open a new PR from a feature branch (it's probably easiest to `git diff

PR #4952, opened 2023-06-15
Fix locals

TheLeoP:
Does the same as #4950, but without the rebase issue.

### DIFF
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -110,7 +110,7 @@ local function install_health()
   end
   if
     vim.iter(vim.api.nvim_list_runtime_paths()):any(function(p)
-      if installdir == p .. '/' then
+      if installdir == vim.fs.normalize(p) .. '/' then
         return true
       end
     end)
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -559,7 +559,7 @@ local function install(languages, options, _callback)
     return
   end
 
-  local cache_dir = vim.fn.stdpath('cache')
+  local cache_dir = vim.fs.normalize(vim.fn.stdpath('cache'))
   local install_dir = config.get_install_dir('parser')
 
   if not languages or type(languages) == 'string' then
--- a/lua/nvim-treesitter/locals.lua
+++ b/lua/nvim-treesitter/locals.lua
@@ -73,8 +73,8 @@ function M.iter_scope_tree(node, bufnr)
 end
 
 -- Gets a table of all nodes and their 'kinds' from a locals list
----@param local_def any: the local list result
----@return table: a list of node entries
+---@param local_def TSLocal[] the local list result
+---@return TSLocal[] a list of node entries
 function M.get_local_nodes(local_def)
   local result = {}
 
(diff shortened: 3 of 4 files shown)


clason:
Also, I'd prefer if you wouldn't group unrelated commits; this makes reviewing harder.

(For example, why did you make the changes in the other files? What precisely isn't working?)

clason:
(Test failure is unrelated)

TheLeoP:
> Also, I'd prefer if you wouldn't group unrelated commits; this makes reviewing harder.

Sorry. I thought it wouldn't make sence to create a new PR for small changes. In the future, would it be preffered to create separate PRs for each change, right?

> (For example, why did you make the changes in the other files? What precisely isn't working?)

- The change in [fix: always normalize paths](https://github.com/nvim-treesitter/nvim-treesitter/pull/4952/commits/6c78269aa69d339c81d57c3ff5184efa895e8b97) was because not doing it produced paths with a mix of `/` and `\` on Windows, so checks that compared two strings always failed.
- The change in [fix: check if config has install_info in filter](https://github.com/nvim-treesitter/nvim-treesitter/pull/4952/commits/463981cd42a9b3344ec843a6b3b5cdc1a116cc14) was because the line `return not M.configs[p].install_info.requires_generate_from_grammar` failed for configs that did not have the field `install_info` (like `jsx`, for example)

clason:
> Sorry. I thought it wouldn't make sence to create a new PR for small changes. In the future, would it be preffered to create separate PRs for each change, right?

Yup; makes it easier to review and merge -- some changes are simple and I can merge immediately, others I'd like to have another review for. Now one change blocks the other two.

TheLeoP:
> Yup; makes it easier to review and merge -- some changes are simple and I can merge immediately, others I'd like to have another review for. Now one change blocks the other two.

Makes a lot of sense. Is it ok to leave this PR as is and take this into account for the future? Or should I open separate PRs?

> Oh, you don't have the tree-sitter CLI or npm installed?

Huh, I didn't know that code was supposed to only got executed if I had neither node or treesitter cli installed. I do have node installed, but I don't have the treesitter cli installed (so `vim.fn.executable('tree-sitter') == 0 or vim.fn.executable('node') == 0` evaluates to true). If I'm understanding you correctly, that condition should have an `and` instead of an `or`, right?

clason:
> Makes a lot of sense. Is it ok to leave this PR as is and take this into account for the future? Or should I open separate PRs?

It's fine now; this was just for (hopefully) future PRs.

> Huh, I didn't know that code was supposed to only got executed if I had neither node or treesitter cli installed. I do have node installed, but I don't have the treesitter cli installed (so vim.fn.executable('tree-sitter') == 0 or vim.fn.executable('node') == 0 evaluates to true). If I'm understanding you correctly, that condition should have an and instead of an or, right?

The point is to only return parsers you can actually install. You are right that the conditional is wrong for that; basically,
* if you don't have tree-sitter installed, don't show parsers that have `requires_generate` set to true;
* if you have tree-sitter installed **but** don't have node installed, don't show parsers that have `requires_npm` set to true.

But that check is a bit expensive, so that's probably why there's this shortcut.

PR #4958, opened 2023-06-17
feat(bash): add various shell builtins

Akmadan23:
I noticed that many shell builtins were missing in the highlights, so I added them.

Moreover:
* added `break` and `continue` as conditionals because it just makes sense
* added `let` and moved `set` and `return` to keywords since they are more akin to things like `declare`, `local` and `unset` which were already there

### DIFF
--- a/queries/bash/highlights.scm
+++ b/queries/bash/highlights.scm
@@ -105,8 +105,15 @@
 
 ((command_name (word) @function.builtin)
  (#any-of? @function.builtin
-    "alias" "cd" "clear" "echo" "eval" "exit" "getopts" "popd"
-    "pushd" "return" "set" "shift" "shopt" "source" "test"))
+    "alias" "bg" "bind" "break" "builtin" "caller" "cd"
+    "command" "compgen" "complete" "compopt" "continue"
+    "coproc" "dirs" "disown" "echo" "enable" "eval"
+    "exec" "exit" "fc" "fg" "getopts" "hash" "help"
+    "history" "jobs" "kill" "let" "logout" "mapfile"
+    "popd" "printf" "pushd" "pwd" "read" "readarray"
+    "return" "set" "shift" "shopt" "source" "suspend"
+    "test" "time" "times" "trap" "type" "typeset"
+    "ulimit" "umask" "unalias" "wait"))
 
 (command
   argument: [


clason:
These are not nodes exposed by the parser, so you cannot add them to the query. This needs to be changed in the `tree-sitter/tree-sitter-bash` grammar first.

Akmadan23:
I took a look at the parser and there is no mention of `@function.builtin`, so I thought it

clason:
No, I mean the `break` etc. Queries work by assigning "captures" (like `@function.builtin`) to "nodes" (like `let`) that the parser generates. You cannot "capture" arbitrary strings.

Akmadan23:
Oh, I see... So if I just changed the `#any-of? @function.builtin` list it would work fine right?

clason:
Yes, any-of is different, that checks the _contents_ of the `command_name` node (and is done on the Neovim side, so less efficiently).

clason:
@TravonteD

amaanq:
I'm curious as to where'd you get the list from (it'd be nice if there was a standardized list of builtins for bash)? It'd also be nice to have chmod/chown added as well

amaanq (approved):

Akmadan23:
> I'm curious as to where'd you get the list from (it'd be nice if there was a standardized list of builtins for bash)?

In bash there is the `help` builtin which lists every builtin and keyword, but here I only added those shared between bash and zsh, because some of them (such as `help`) are bash-only. I know this is theoretically only a bash parser, but since there isn't a specific one for zsh I think most of the people use it for both of them. By the way `chmod` and `chown` are regular binaries, not shell builtins.

amaanq:
Ah...thanks for the clarification. Imo it'd be fine to add help as this is technically a bash parser (and I'd say most shell scripts target sh/bash anyways)..

clason:
Yes, this is a bash parser, and should strive to be as complete for bash as possible. If someone wants a zsh parser, they are free to fork it. (Since we have `; inherits` support, we can share queries; see the `ecma` queries for JS and friends.)

Akmadan23:
Ok, that makes sense. Do you prefer a new commit to the branch or a force push with a single commit?

PR #4962, opened 2023-06-18
chore: update cmake

uyha:
Fix crashes when opening CMake files

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -48,7 +48,7 @@
     "revision": "6e41628e9d18b19caea1cb1d72aae4ccff5bdfe9"
   },
   "cmake": {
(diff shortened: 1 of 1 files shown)


amaanq:
Is that so? do you have a file i can try on?

uyha:
let me try to create some file that I can share

uyha:
actually, just open an empty cmake file, type # and it crashes nvim immediately

amaanq:
caught it, I forgot advance/skip need to be static...

amaanq:
https://github.com/uyha/tree-sitter-cmake/pull/16

amaanq (approved):

PR #4974, opened 2023-06-20
improve scheme/racket quote highlight

6cdh:
Now it can highlight code like

```
'(list a b c)
(quote a b c)
```

as `@symbol`. It may require a specific colorscheme to work properly.

### DIFF
--- a/queries/racket/highlights.scm
+++ b/queries/racket/highlights.scm
@@ -47,10 +47,6 @@
 (extension) @keyword
 (lang_name) @variable.builtin
 
-;; quote ;;
-
-(quote) @symbol
-
 ;; list ;;
 
 ["(" ")" "[" "]" "{" "}"] @punctuation.bracket
@@ -124,7 +120,7 @@
 ;;                         Solve conflicts                          ;;
 ;;------------------------------------------------------------------;;
 
-;; See `:h treesitter`, and search `priority`
+;; See `:h treesitter-highlight-priority`
 
 (list
  .
@@ -132,10 +128,17 @@
  (#eq? @include "require")
  (#set! "priority" 101))
 
-(quote
-  .
-  (symbol)
-  (#set! "priority" 105)) @symbol
+;; quote
+
+(quote (#set! "priority" 105)) @symbol
+
+(list
+ .
+ (symbol) @_f
+ (#eq? @_f "quote")
+ (#set! "priority" 105)) @symbol
+
+
 
 ((sexp_comment) @comment
  (#set! "priority" 110))
--- a/queries/scheme/highlights.scm
+++ b/queries/scheme/highlights.scm
@@ -9,11 +9,6 @@
 [(comment)
  (block_comment)] @comment @spell
 
-;; highlight for datum comment
-;; copied from ../clojure/highlights.scm
-([(comment) (directive)] @comment
- (#set! "priority" 105))
-
 (escape_sequence) @string.escape
 
 ["(" ")" "[" "]" "{" "}"] @punctuation.bracket
@@ -78,17 +73,6 @@
 
 ((symbol) @conditional
  (#any-of? @conditional "if" "cond" "case" "when" "unless"))
-
-;; quote
-
-(quote
- "'"
- (symbol)) @symbol
-
-(list
- .
- (symbol) @_f
- (#eq? @_f "quote")) @symbol
 
 ;; library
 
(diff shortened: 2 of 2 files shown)


queries/racket/highlights.scm
```diff
@@ -124,18 +120,25 @@
...
-  (#set! "priority" 105)) @symbol
+;; quote
+
+(quote (#set! "priority" 105)) @symbol
+
+(list
+ .
+ (symbol) @_f
+ (#eq? @_f "quote")
+ (#set! "priority" 105)) @symbol
```
lucario387, line 139: Is it necessary to set a priority to these captures (and captures after it). Can't it be moved down instead?

6cdh: Yes. It's here for nested structure.

For example, It would be impossible to highlight this whole sexp as the same color without priority

```racket
(quote (list a b c (list d e f)))
```

lucario387: I'm still a bit not sold on the need for priorities, as captures applied earlier already has lower precedence

6cdh: > captures applied earlier already has lower precedence

precedence != priority.

from the docs `:h treesitter-highlight-priority`

> Treesitter uses |nvim_buf_set_extmark()| to set highlights with a default priority of 100

earlier query don't have lower priority.

Another reason is it works with priority, and don't work without priority.

To make sure we are talking about the same thing. The query without priority is:

```scheme
(list
 .
 (symbol) @_f
 (#eq? @_f "quote")) @symbol
```

A picture without priority:
Screenshot_20230623_190120

with priority:

Screenshot_20230623_190209

clason: but precedence is not orthogonal to priority -- priority is (only!) needed to work around unwanted precedence. If you can ensure the wanted precedence directly, that is always preferable.

6cdh: I'm sorry I can't as I'm not familiar with the internal implementation.
Could you suggest me a change that works without priority?

clason: Just try moving the query to after the other query it should have priority over ("last highlight wins", and they are applied in the order they appear in this file).

6cdh: I've heard that and tried it. But it doesn't work as I said above.

To make the context clearer,

`(list a b c 1 2 3)` should be highlight as this

```racket
(list a b c 1 2 3)
(@variable @variable @variable @variable @number @number @number)
```

And `list` should also be highlight as a builtin function.

while when a quote wraps it, like this:

```racket
(quote (list a b c 1 2 3))
```

everything should be highlight as `@symbol`.

I think priority is necessary here.

I don't know why you don't like it so much though.

clason: > I don't know why you don't like it so much though.

1. Because it doesn't do what most people believe it does (the priority applies to the extmark used for highlighting, not for the query/capture itself; these still get all applied in their order!).2
2. Because it can interfere with injections.

So it's always worth trying to do something without priority first.

lucario387: Similarly, from a quick look at the queries, the `priority` is put in such a way that correlates to the order of highlight application itself, so it should not be needed if all of them were to be removed

6cdh: > so it should not be needed if all of them were to be removed

@lucario387 How? Give me a change suggestion to make it works under this context.

@clason I tried it many times and can't do it anyway as the nested level of list can be infinity.

You could

1. accept current implementation, or
2. accept another implementation which only highlight `'abc` and only single nested level list `'(list a b c)`. This implementation don't need priority.

If you accept case 2, then I will do it later.

clason: All I was asking was to try a different order first; I wasn't saying priorities are inacceptable.

6cdh: Then how? I tried it many times and can't implement. I just think it's impossible to implement without priority.

clason: First, as I said, priorities are ok if ordering is not enough; all we did was ask you to confirm that ordering is in fact not enough. You did, so that's fine with me. I don't understand why we keep discussing this -- increasingly acrimoniously...

But there may be a misunderstanding: the query that raised _our_ eyebrows is
```
(quote (#set! "priority" 105)) @symbol
```
(In fact, I suspect -- and hence our question -- that _all_ these priorities are not really necessary in nvim-treesitter, even though they _are_ necessary in `tree-sitter highlight`, which uses a different precedence rule ("first match wins", instead of "all matches are applied in order").

PR #4976, opened 2023-06-20
fix(starlark): adapt to scanner rewrite

amaanq:


clason:
You need to change the lockfile commit to the rewrite.

clason (approved):

PR #4979, opened 2023-06-20
fix(sql): table_reference -> object_reference

dmfay:
the `table_reference` node will be changing soon once https://github.com/DerekStride/tree-sitter-sql/pull/150 lands

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -459,7 +459,7 @@
     "revision": "05f949d3c1c15e3261473a244d3ce87777374dec"
   },
   "sql": {
-    "revision": "7cbac0472e5b8f8486ce64ffbcf1982d5cd5fc8d"
+    "revision": "baab519f9781937e948cdb2b4ee4138a0dcda866"
   },
   "squirrel": {
     "revision": "3fefc6b9bb2b4de1b1c461783f675918cd957546"
--- a/queries/sql/highlights.scm
+++ b/queries/sql/highlights.scm
@@ -14,7 +14,7 @@
   (keyword_group_concat)
 ] @function.call
 
-(table_reference
+(object_reference
   name: (identifier) @type)
 
 (relation
--- a/queries/sql/highlights.scm
+++ b/queries/sql/highlights.scm
@@ -21,7 +21,6 @@
   alias: (identifier) @variable)
 
 (field
-  alias: (identifier) @variable
   name: (identifier) @field)
 
 (term
--- a/queries/sql/highlights.scm
+++ b/queries/sql/highlights.scm
@@ -1,6 +1,7 @@
 (invocation
-  name: (identifier) @function.call
-  parameter: [(field)]? @parameter)
+  (object_reference
+    name: (identifier) @function.call
+    parameter: [(field)]? @parameter))
 
 [
   (keyword_gist)
--- a/queries/sql/highlights.scm
+++ b/queries/sql/highlights.scm
@@ -18,10 +18,10 @@
   name: (identifier) @type)
 
 (relation
-  table_alias: (identifier) @variable)
+  alias: (identifier) @variable)
 
 (field
-  table_alias: (identifier) @variable
+  alias: (identifier) @variable
   name: (identifier) @field)
 
 (term


dmfay:
I've landed https://github.com/DerekStride/tree-sitter-sql/pull/150 , marking ready on the assumption that this needs to go next!

matthias-Q:
You need to update the lockfile.json with `baab519f9781937e948cdb2b4ee4138a0dcda866`

amaanq (approved):

amaanq:
Please test the queries locally before pushing, I don't want to have to rerun the workflow every time

dmfay:
> Please test the queries locally before pushing, I don't want to have to rerun the workflow every time

ah I didn't realize that was manual, sorry. I did work out how to run check_queries with a headless nvim and find where to put the generated .so, but I'm getting this error locally:

```
Check failed:
./scripts/check-queries.lua:96: sql (highlights): /usr/share/nvim/runtime/lua/vim/treesitter/query.lua:259: query: invalid node type at position 271 for language sql
```

with no indication _which_ highlight query was a problem or what "position 271" is relative to/in which file. "invalid node type" gets a mention in the readme but just says to make sure the parser is up to date (it is) and not duplicated (it isn't, I moved the old sql.so out). The lua line numbers aren't much help either.

amaanq:
Tldr clone ts-sql, copy the queries to queries/* in ts-sql, run `ts t` and you'll see where the issue is

matthias-Q:
Invocation has changed and should look like this:
```
(invocation
  (object_reference
    name: (identifier) @function.call
    parameter: [(field)]? @parameter))
```

Also this is broken:
```
(field
  alias: (identifier) @variable
  name: (identifier) @field)
```
but I don't know how this query should look with the new version...

clason:
> > Please test the queries locally before pushing, I don't want to have to rerun the workflow every time
> 
> ah I didn't realize that was manual, sorry. I did work out how to run check_queries with a headless nvim and find where to put the generated .so, but I'm getting this error locally:
> 
> ```
> Check failed:
> ./scripts/check-queries.lua:96: sql (highlights): /usr/share/nvim/runtime/lua/vim/treesitter/query.lua:259: query: invalid node type at position 271 for language sql
> ```
> 
> with no indication _which_ highlight query was a problem or what "position 271" is relative to/in which file. "invalid node type" gets a mention in the readme but just says to make sure the parser is up to date (it is) and not duplicated (it isn't, I moved the old sql.so out). The lua line numbers aren't much help either.

Yes, that's an unfortunate pain point in the Neovim treesitter code. But with Neovim 0.9+, if you have the parser installed, you can just open the query file in this repo, and the linter will tell you which query stanza has an error.

matthias-Q:
I was thinking for a while to add a copy of queries to the tree-sitter-sql repo, so tests will break already there. However queries are different for neovim, helix and emacs...

clason:
> I was thinking for a while to add a copy of queries to the tree-sitter-sql repo, so tests will break already there. However queries are different for neovim, helix and emacs...

Yes, which is unfortunate (and something we try to ameliorate, although perfect alignment will never be achievable until all editors become one in the future singularity).

Nevertheless, this would be useful since it's much easier to adapt existing, known to be working, queries to a specific editor than start from scratch. The main areas of divergence are
1. capture names: this is trivial to adapt
2. different predicates/directives: irrelevant if the upstream queries don't use any (or a lowest common denominator, which we can agree on)
3. query order: upstream/Helix has first matching query wins (because the highlighter short-circuits), while Neovim has last matching query wins (because they simply all applied and "paint over" each other).

Point 3 is an issue (we -- and upstream now -- believe that Neovim's approach is better) but can be dealt with by simply trying to not rely on query ordering for priority in favor of making captures more specific (which is preferable anyway).

dmfay:
>

thanks! I was very confused until I realized it detected the query file as Scheme but `:set ft=query` with the `query` parser installed fixed that.

@matthias-Q I've left `field` in -- we still use it in a couple of places e.g. assignment left-hand side in updates.

matthias-Q:
Because of the query order highlighting of `invocation` is now broken. It is overridden by `@type` and should be `@function.call`

I will see if I can make a PR later today

PR #4983, opened 2023-06-21
fix: prioritize comment keywords over semantic highlights

ribru17:
Allows the comment parser to work in tandem with semantic highlighting of comments


clason:
Thank you, but no. LSP semantic tokens have higher priority than treesitter highlighting; this is defined and documented in Neovim core. You are welcome to change these locally, but we will not do so in this repo.

ribru17:
Understandable. Perhaps we/one could document how to set these priorities locally in some wiki, discussion, or elsewhere on this repo? As you can imagine the lower priority behavior is quite unintuitive to the average user (for this specific parser).

ribru17:
That is indeed the issue, thank you both.

PR #4984, opened 2023-06-21
chore(matlab): bumps commit to latest.

acristoffers:
Bumps version to include changes in acristoffers/tree-sitter-matlab#2


clason:
Thank you, but we have an automated script for that. Manual bumping is only necessary if the queries need to be adapted at the same time.

PR #4987, opened 2023-06-21
lint(locals): use `#set!` to keep query linter happy

clason:


amaanq (approved):

clason:
Wait, that can't be it -- the other local queries also have `set!`, and the query check still fails locally.

clason:
Looks like these two are breaking (locally on main, so probably due to the refactor):
```lisp
(namespace_definition
  name: (namespace_identifier) @definition.namespace
  body: (_) @scope)

(namespace_definition
  name: (nested_namespace_specifier) @definition.namespace
  body: (_) @scope)
```

clason:
@ObserverOfTime

ObserverOfTime:
Works on my machine. :thinking: 

```
$ nvim -l scripts/check-queries.lua 
::group::Check parsers
Checking arduino highlights (131.52ms)
Checking arduino locals (34.48ms)
Checking arduino folds (10.35ms)
Checking arduino indents (19.01ms)
Checking arduino injections (6.30ms)
```

PR #4990, opened 2023-06-21
feat: add support for haskell persistent library

lykahb:
The syntax defines data types and the corresponding tables in a database. I've also created a PR for vim to add a filetype https://github.com/vim/vim/pull/12573, so the file would be recognized automatically once it gets merged and neovim ports it.

image


queries/haskell_persistent/highlights.scm
```diff
@@ -0,0 +1,38 @@
...
+;; Keywords, operators, includes
+
+[
+  "Id"
+  "Primary"
+  "Foreign"
+  "deriving"
+] @keyword
+
+"=" @punctuation.delimiter
```
amaanq, line 25: does `@operator` fit better?

lykahb: Yes, in HTML and Haskell syntaxes it is an @operator.

amaanq (approved):

PR #4993, opened 2023-06-22
fix(sql): fix highlights order for `invocation`

matthias-Q:
Fixes the query order in SQL highlighting introduced due to changes in #4979


clason:
Thanks!

PR #5009, opened 2023-06-25
fix(haskell_persistent): add to lockfile.json

stephen-huan:
e9984bb appears to have forgotten to update `lockfile.json`, causing `:TSUpdate haskell_persistent` to always update.


amaanq:
thanks!

amaanq (approved):

PR #5014, opened 2023-06-25
fix(sql): compound_statement -> block

dmfay:
a little more reorganizing; we've copied the indents file into tree-sitter-sql and it tests okay there.

### DIFF
--- a/queries/sql/indents.scm
+++ b/queries/sql/indents.scm
@@ -7,8 +7,7 @@
  (insert)
 ] @indent.begin
 
-
-(compound_statement
+(block
   (keyword_begin)
 ) @indent.begin
 
--- a/lockfile.json
+++ b/lockfile.json
@@ -462,7 +462,7 @@
     "revision": "05f949d3c1c15e3261473a244d3ce87777374dec"
   },
   "sql": {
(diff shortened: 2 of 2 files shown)


clason:
@matthias-Q

queries/sql/indents.scm
```diff
@@ -7,8 +7,7 @@
  (insert)
 ] @indent.begin
 
-
-(compound_statement
+(block
```
lucario387, line 10: Hmmm, which one is more narrowed down. Is it `compound_statement` or `block`

as in, is the s-expr tree `(compound_statement (block (keyword_begin)))` or `(block (compound_statement (keyword_begin)))`

lucario387: Ahhh, I see from the commit logs of tree-sitter-sql now, `compound_statement` is now renamed and refactored into `block`, is it correct? @dmfay

clason: I think this was just a rename in the grammar?

dmfay: that's correct, `block` is a more common name for the expression formerly called `compound_statement`. Only the name has changed, nothing's more specific or more general.

clason (approved):

lucario387 (approved):

PR #5015, opened 2023-06-25
fix(phpdoc): disable spell checking of attribute names

przepompownia:


przepompownia:
Before:
image

After:
image

lucario387:
I think https://github.com/nvim-treesitter/nvim-treesitter/blob/7e3f574f060125b4af71677e6c74acb20b5ef611/queries/phpdoc/highlights.scm#L45

can be narrowed down to only include description text instead of all types of text @przepompownia 

Okay with this approach tho

przepompownia:
@lucario387 I noticed that in my PR types are still checked against spell (e.g. `bool`).

Some popular IDE-s check also variable names. Your proposition seems to be more consistent with `queries/php/highlights.scm` so I did it similarly. 

Notice that it is still done indirectly. I could not find any working way to directly attach `@spell` to description text. If anyone can, please make a separate PR or propose something else here.

lucario387:
@przepompownia my suggestion was that to remove the `(text) @spell` from the bottom of the query file and move it inside the `(tag (description (text)))` region. Try that out and see if that meets your criteria

przepompownia:
@lucario387 I found why my previous approaches failed. At the `php` level the whole comments was checked. After disabling it there your proposition started to work.

queries/php/highlights.scm
```diff
@@ -142,7 +142,6 @@
 (null) @constant.builtin
 (integer) @number
 (float) @float
-(comment) @comment @spell
```
lucario387, line 145: This I don't agree with. It can be disabled by the user through `(comment) @nospell`, but unsure what would the effect be with other `@spell` of phpdoc tho?

przepompownia: I realize that disabling spell checking of comments at the php level can be controversial. IMHO it doesn't do its job well (it checks variable names, types etc.). At the phpdoc level we can do it more precisely.

lucario387: yep, it'll be controversial. Personally I don't use comment-related parsers (comment, phpdoc, ...) so some spellchecks that doesn't rely on the parsers will be more preferrable. 

Up to the php maintainers to decide. 
cc @tk-shirasaka

lucario387:
Please don't do merge commits, rebase instead.

przepompownia:
Can't they be squashed into one at merge?

queries/php/highlights.scm
przepompownia: No way to recognize if the comment is phpdoc or not (and enable spell on non-phpdoc comments?)?

lucario387: > No way to recognize if the comment is phpdoc or not (and enable spell on non-phpdoc comments)?

You can keep `(comment) @comment @spell` in `php`, while having a `(text) @nospell` at the top of the phpdoc file to counter that `@spell`, then have `(description (text) @spell)` to re-enable it for that section only, I believe

przepompownia: > You can keep `(comment) @comment @spell` in `php`, while having a `(text) @nospell` at the top of the phpdoc file to counter that `@spell`, then have `(description (text) @spell)` to re-enable it for that section only, I believe

It does not work for me. 

At the moment my best approach that resolves the problem without disabling `@spell` at the php level for the whole comment is to put `@nospell` for particular nodes like in the second commit (c2297a6ffd4396e2dce7ff36961d914c391e50ee).

lucario387: Let's try with the best approach you have :D 

If anything, it's probably the most correct one imo

jessarcher (commented):
I've been using the changes from this PR for a few weeks, and it's been solid :+1:

lucario387 (approved):

przepompownia:
Thank you all for review and checking.

lucario387:
Sorry for forgetting about the branch :bowing_man: 

It was fine for months

PR #5016, opened 2023-06-25
injections(nix): add nixosTest + home-manager neovim lua plugin config injections

mrcjkb:


lucario387:
Thank you very much for the PR, but I'd like you to base the PR off the `main` branch instead of `master`. Such improvements will be backported to `master` branch at a later date

lucario387:
Another recurring issue with Nix is the performance issue of the queries. If possible, please describe your editing experience with and without the queries in question, and how much it effects the experience.

Some relevant large files will be preferred

mrcjkb:
I'll have to get back to this in a few days.

Performance-wise, I haven't noticed any issues (I work with nix almost daily), but my laptop is quite beefy, so that might not be the best benchmark.

Is there a GitHub issue about the performance (e.g. in the nix parser repo)? So I can understand it better...

lucario387:
The performance and stability of the nix parser has lead to the parser version being locked for a while, and with PRs such as #4658, I have found some performance issues that lead to the initial parse of a 20k LOC nix file to be upwards of 20 seconds. Adding a `.` to the pattern, the time goes down to 2 second, removing the capture, the initial parse time is 1.5 second, so it's somewhat solved.

The problem here is somewhat similar to the PR mentioned above:
- How often are such patterns going to occur, during normal usage, or during reading an unusually large files.
- How many nodes are usually needed to match the pattern? Can it be narrowed down further? I'm not used to what is used in Nix, so that's outside of my control.
- What is the "feel" when editing Nix files, after adding such queries. This is entirely subjective, and I trust your judge on the feel.

mrcjkb:
I have applied the requested changes, but I would like to leave this as a draft for now, while I dogfeed my fork.

It would probably also be nice to add some tests.

runtime/queries/nix/injections.scm
```diff
@@ -129,3 +129,34 @@
...
+((binding
+  ((attrpath) @_attr_name) 
+    (#eq? _attr_name "nodes")
+  )
+  (binding
+   ((attrpath) @_func_name) (#eq? @_func_name "testScript")
+   (_
+     (string_fragment) @injection.content (#set! injection.language "python")
+   ))
+   (#set! injection.combined))
```
lucario387, line 143: Are testScripts always in python? If yes then lgtm

runtime/queries/nix/injections.scm
```diff
@@ -129,3 +129,34 @@
   ])
   (#match? @_func "(^|\\.)writeRust(Bin)?$")
   (#set! injection.combined))
+
+;; (nixosTest) testScript
+((binding
+  ((attrpath) @_attr_name)
```
lucario387, line 135: From the looks of other nix injection queries, `binding` has `expression` and such named fields, can you add them in as well?

runtime/queries/nix/injections.scm
mrcjkb: Yes, I'm matching 
```nix
{
 nodes = { ... }
 testScript = ''
 '';
}
```
where `testScript` should always be Python used to interact with virtual machines defined in `nodes`.

I just updated my local injections and they're no longer working, so I'll have to come back to this.
I see the directory structure has changed - maybe I'll have to update my local injections accordingly.

mrcjkb:
Update: Looks like my tests are succeeding now.
No idea why there are Javascript-related failures in the test suite now.

clason:
No, `main` gets frequent rebases and force-pushes, which breaks all PRs. Simplest is to `git rebase -i main`, and then `drop` all commits but your own and force-push.

I only do that for PRs under active review, though.

mrcjkb:
Ah, thanks for the info.
And thanks for taking care of it for me 😄

runtime/queries/nix/injections.scm
```diff
@@ -129,3 +129,35 @@
...
+   (#set! injection.combined))
+
+;; home-manager Neovim plugin config
+(attrset_expression
+  (binding_set
+    (binding
+      attrpath: (attrpath) @_ty_attr
+      (_
+        (string_fragment) @_ty
+      )
```
lucario387, line 152:
```suggestion
      (_
        (string_fragment) @_ty)
```

Same for all other parentheses that I don't want to list out

lucario387 (approved):
Only need to change the stylings to be lisp-style.

Aside from that I'm okay with the changes

runtime/queries/nix/injections.scm
mrcjkb: I think that should be all of them.

lucario387 (approved):

PR #4849, opened 2023-05-23
go: register `clear` as a new builtin function

110y:
The next Version of Go (1.21) introduces a new built-in function `clear`: https://github.com/golang/go/blob/8c445b7c9fe6738cbef2040a1011bd11489b0806/src/builtin/builtin.go#L230-L237. 
So, this PR just registers `clear` as a new built-in function.

PR #5018, opened 2023-06-26
go: register `max` and `min` as new builtin functions

110y:
The next version of Go (1.21) introduces new builtin functions: `max` and `min`:
https://github.com/golang/go/blob/a031f4ef83/src/builtin/builtin.go#L211-L221

### DIFF
--- a/queries/go/highlights.scm
+++ b/queries/go/highlights.scm
@@ -171,6 +171,8 @@
            "imag"
            "len"
            "make"
+           "max"
+           "min"
            "new"
            "panic"
            "print"


amaanq (approved):

lucario387 (approved):

lucario387:
Will wait till 1.21 goes out of rc :D

110y:
@lucario387 

Thank you for approving 👍 

Since my PR, which also adds a new builtin function introduced in 1.21, has already been merged, I think it's OK to merge this PR. What do you think about this...?
ref:  https://github.com/nvim-treesitter/nvim-treesitter/pull/4849

lucario387:
I personally believe the amount of times people write their `max` and `min` func are significantly more than `clear`, so it may cause some confusion in highlights if it's  marked as a builtin func.

Holding this back until it becomes the default will cause less confusion and related issues to highlights imo

110y:
@lucario387 

> I personally believe the amount of times people write their max and min func are significantly more than clear, so it may cause some confusion in highlights if it's marked as a builtin func.

Got it 👍 
By the way, though I would agree with your concern regarding this PR, I think treesitter project should have a solid policy that defines when (nightly, release candidate, stable release etc.) we should follow language changes 😄

ArtAndreev:
There is a mistake: these builtins are functions, not types.

lucario387:
Agreed on that

amaanq:
Go 1.21 is out now

PR #5021, opened 2023-06-26
feat(wing): add parser

gshpychka:
Add a parser for [Wing](https://github.com/winglang/wing)

I've copied the queries from the upstream repo. This is my first contribution here, let me know if I got it right.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1654,7 +1654,6 @@ list.wgsl_bevy = {
 list.wing = {
   install_info = {
     url = "https://github.com/winglang/wing",
-    branch = "main",
     files = { "src/parser.c", "src/scanner.c" },
     location = "libs/tree-sitter-wing",
     requires_generate_from_grammar = true,
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1651,6 +1651,18 @@ list.wgsl_bevy = {
   maintainers = { "@theHamsta" },
 }
 
+list.wing = {
+  install_info = {
+    url = "https://github.com/winglang/wing",
+    branch = "main",
+    files = { "src/scanner.c" },
+    location = "libs/tree-sitter-wing",
+    requires_generate_from_grammar = true,
+  },
+  maintainers = { "@gshpychka" },
+  experimental = true,
+}
+
 list.yaml = {
   install_info = {
     url = "https://github.com/ikatyang/tree-sitter-yaml",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1655,7 +1655,7 @@ list.wing = {
   install_info = {
     url = "https://github.com/winglang/wing",
     branch = "main",
-    files = { "src/scanner.c" },
+    files = { "src/parser.c", "src/scanner.c" },
     location = "libs/tree-sitter-wing",
     requires_generate_from_grammar = true,
   },
--- a/queries/wing/highlights.scm
+++ b/queries/wing/highlights.scm
@@ -14,7 +14,7 @@
   name: (identifier) @method
 )
 (inflight_method_definition
-  name: (identifier) @function
+  name: (identifier) @method
 )
 
 ; Functions
--- a/queries/wing/highlights.scm
+++ b/queries/wing/highlights.scm
@@ -1,3 +1,7 @@
+(identifier) @variable
+(reference_identifier) @variable
+(member_identifier) @property
+
 ; Classes
 
 (custom_type) @type
@@ -88,7 +92,3 @@
   "return"
   (inflight_specifier)
 ] @keyword
-
-(identifier) @variable
-(reference_identifier) @variable
-(member_identifier) @property
--- a/queries/wing/highlights.scm
+++ b/queries/wing/highlights.scm
@@ -1,0 +1,94 @@
+; Classes
+
+(custom_type) @type
+(class_field 
+  name: (identifier) @member
+) 
+(class_definition 
+  name: (identifier) @type
+)
+(method_definition
+  name: (identifier) @function
+)
+(inflight_method_definition
+  name: (identifier) @function
+)
+
+; Functions
+
+(keyword_argument_key) @variable.parameter
+(call 
+  caller: (reference 
+  	(nested_identifier 
+    	property: (member_identifier) @function.method)) 
+)
+(call 
+  caller: (reference 
+  	(reference_identifier) @function.method)
+)
+
+; Primitives
+
+[
+ (number)
+ (duration)
+] @constant.builtin
+(string) @string
+(bool) @constant.builtin
+(builtin_type) @type.builtin
+(json_container_type) @type.builtin
+
+; Special
+
+(comment) @comment
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+]  @punctuation.bracket
+
+[
+  "-"
+  "+"
+  "*"
+  "/"
+  "%"
+  "<"
+  "<="
+  "="
+  "=="
+  "!"
+  "!="
+  ">"
+  ">="
+  "&&"
+  "??"
+  "||"
+] @operator
+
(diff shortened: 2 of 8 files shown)


clason:
Thank you! Unfortunately, upstream queries cannot be used (unchanged); the two main differences are
1. we use different capture names, see https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md
2. the query order precedence is different: upstream has "first match wins", Neovim has "all matches are applied, and the last one wins (by painting over the older ones)".

lua/nvim-treesitter/parsers.lua
```diff
@@ -1651,6 +1651,18 @@ list.wgsl_bevy = {
   maintainers = { "@theHamsta" },
 }
 
+list.wing = {
+  install_info = {
+    url = "https://github.com/winglang/wing",
+    branch = "main",
```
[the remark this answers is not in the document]
gshpychka: @clason the docs imply that the default is "master", is that not the case anymore?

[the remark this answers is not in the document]
clason: No, I mean for the parser repo (this is configurable and looks to be `main` for winglang/wing).

gshpychka:
> Thank you! Unfortunately, upstream queries cannot be used (unchanged); the two main differences are
> 
>     1. we use different capture names, see https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md
> 
>     2. the query order precedence is different: upstream has "first match wins", Neovim has "all matches are applied, and the last one wins (by painting over the older ones)".



> Thank you! Unfortunately, upstream queries cannot be used (unchanged); the two main differences are
> 
>     1. we use different capture names, see https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md
> 
>     2. the query order precedence is different: upstream has "first match wins", Neovim has "all matches are applied, and the last one wins (by painting over the older ones)".

I believe I've fixed #1. Regarding #2, can you point me in a direction as to how to fix this? I'm not too familiar with TS.

clason:
Well, best way is to use the parser and see if it behaves as expected; if it doesn't, you can try to just shuffle the individual queries so the ones who should win come _later_ instead of earlier.

(You can also write tests, which are always welcome ;))

queries/wing/highlights.scm
```diff
@@ -0,0 +1,94 @@
...
+  "init"
+  "let"
+  "new"
+  "return"
+  (inflight_specifier)
+] @keyword
+
+(identifier) @variable
+(reference_identifier) @variable
+(member_identifier) @property
```
clason, line 94: Specifically, I believe these should be at the _top_ of the file here.

clason:
We also provide queries for indentation and folding, by the way. Indentation may be a bit involved, but folds at least should be easy to add (just look at some other ones in this repo).

queries/wing/locals.scm
```diff
@@ -0,0 +1,6 @@
+(block) @local.scope
+(variable_definition_statement
+  name: (identifier) @local.definition
```
clason, line 3: local queries are also different from upstream (see `CONTRIBUTING.md`; we don't use `@local` scope -- for now).

lua/nvim-treesitter/parsers.lua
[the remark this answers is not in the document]
gshpychka: @clason right, the documentation implies that the default value is "master", and should be overriden if it's something else (like "main"): https://github.com/nvim-treesitter/nvim-treesitter/blob/d49ba798c95398e62321b40041e8ac4654fea251/README.md?plain=1#L505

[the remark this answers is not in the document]
clason: Ah, indeed, that is misleading. You really only need this if the branch is different from the _default_ branch.

(I suspect this was written before the big `master`->`main` migration, when the default branch was fairly universally `master`.)

gshpychka:
> We also provide queries for indentation and folding, by the way. Indentation may be a bit involved, but folds at least should be easy to add (just look at some other ones in this repo).

Added some folds

queries/wing/highlights.scm
gshpychka: I committed this change, will need to research further if that alone is not sufficient

queries/wing/highlights.scm
```diff
@@ -0,0 +1,94 @@
...
+
+; Classes
+
+(custom_type) @type
+(class_field 
+  name: (identifier) @field
+)
```
ObserverOfTime, line 10: Lisp formatting convention. Same for the rest.

```suggestion
  name: (identifier) @field)
```

queries/wing/highlights.scm
```diff
@@ -0,0 +1,94 @@
...
+  name: (identifier) @field
+) 
+(class_definition 
+  name: (identifier) @type
+)
+(method_definition
+  name: (identifier) @method
+)
+(inflight_method_definition
+  name: (identifier) @function
```
ObserverOfTime, line 18: Why is this not `@method`?

queries/wing/highlights.scm
```diff
@@ -0,0 +1,94 @@
...
+)
+
+; Primitives
+
+[
+ (number)
+ (duration)
+] @constant.builtin
+(string) @string
+(bool) @constant.builtin
```
ObserverOfTime, line 41:
```suggestion
(number) @number
(duration) @constant
(string) @string
(bool) @boolean
```

ObserverOfTime (changes requested):

queries/wing/highlights.scm
```diff
@@ -0,0 +1,91 @@
...
+(class_field 
+  name: (identifier) @field)
+(class_definition 
+  name: (identifier) @type
+)
+(method_definition
+  name: (identifier) @method
+)
+(inflight_method_definition
+  name: (identifier) @function
```
gshpychka, line 17:
```suggestion
  name: (identifier) @method
```

queries/wing/highlights.scm
gshpychka: This one was my bad. Fixed.

gshpychka:
I am in the process of adding some tests for highlighting

gshpychka:
Added tests, but they fail if I move the last three highlighting queries to the top as suggested [here](https://github.com/nvim-treesitter/nvim-treesitter/pull/5021#discussion_r1242453572)

They pass without this change. Discussing this on matrix, posting here as well for posterity.

clason:
> Added tests, but they fail if I move the last three highlighting queries to the top as suggested https://github.com/nvim-treesitter/nvim-treesitter/pull/5021#discussion_r1242453572

Yes, you need to use _our_ test suite (which unfortunately is not trivial to run locally); `tree-sitter test` will give wrong results due to the different query ordering strategies.

tests/query/highlights/wing/class.w
```diff
@@ -0,0 +1,19 @@
+bring cloud;
+// <- @keyword
```
clason, line 2:
```suggestion
// <- keyword
```
omit the leading `@`

clason: everywhere; but the test still fails for some reason. @theHamsta ?

clason:
Oh, and you probably also need to add the `.w` extension for `wing` files to
https://github.com/nvim-treesitter/nvim-treesitter/blob/bad9f5bf5b4af5d97a9cffdc957f1b23dc91d851/scripts/minimal_init.lua#L5-L15

queries/wing/highlights.scm
```diff
@@ -0,0 +1,91 @@
...
+  name: (identifier) @field)
+(class_definition 
+  name: (identifier) @type
+)
+(method_definition
+  name: (identifier) @method
+)
+(inflight_method_definition
+  name: (identifier) @method
+)
```
ObserverOfTime, line 18:
```suggestion
  name: (identifier) @method)
```

ObserverOfTime (changes requested):
There are are more lines that should be reformatted. :)

ObserverOfTime (approved):

PR #5024, opened 2023-06-27
feat(pem): add parser

ObserverOfTime:
Should we highlight labels recognised by [OpenSSL](https://github.com/openssl/openssl/blob/master/include/openssl/pem.h#L35-L60) as `@label.builtin` or nah?


clason:
Is it possible to add fold queries? The `BEGIN...END` block sounds like a likely (the only?) candidate.

amaanq:
> Should we highlight labels recognised by [OpenSSL](https://github.com/openssl/openssl/blob/master/include/openssl/pem.h#L35-L60) as `@label.builtin` or nah?

up to you, I'm indifferent

PR #5032, opened 2023-06-28
feat(cmake): update cmake parser, adapt highlight.scm, and add indents.scm

uyha:


queries/cmake/indents.scm
```diff
@@ -0,0 +1,21 @@
...
+ (elseif_command)
+ (elseif_command)
+ (endif_command)
+ (endforeach_command)
+ (endfunction_command)
+ (endmacro_command)
+ (endblock_command)
+] @indent.branch
+
+(normal_command ")" @indent.branch)
```
lucario387, line 21: Maybe a `(normal_command ")" @indent.end)` after this to ensure it doesn't continue the previous indent level

uyha: already added

lucario387 (approved):

PR #5042, opened 2023-06-29
feat: add Hoon language support

urbit-pilled:
Adding a parser for the language [hoon](https://developers.urbit.org/reference/hoon)

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -694,6 +694,14 @@ list.hocon = {
   maintainers = { "@antosha417" },
 }
 
+list.hoon = {
+  install_info = {
+    url = "https://github.com/urbit-pilled/tree-sitter-hoon",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@urbit-pilled" },
+}
+
 list.html = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-html",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -700,6 +700,7 @@ list.hoon = {
     files = { "src/parser.c" },
   },
   maintainers = { "@urbit-pilled" },
+  experimental = true,
 }
 
 list.html = {
--- a/queries/hoon/highlights.scm
+++ b/queries/hoon/highlights.scm
@@ -1,0 +1,32 @@
+(number) @number
+
+(string) @string
+
+[
+  "("
+  ")"
+  "["
+  "]"
+]  @punctuation.bracket
+
+[
+  (coreTerminator)
+  (seriesTerminator)
+] @punctuation.delimiter
+
+
+(rune) @keyword
+
+(term) @constant
+
+(aura) @constant.builtin
+
+(Gap) @comment
+
+(boolean) @constant.builtin
+
+(date) @constant.builtin
+(mold) @constant.builtin
+(specialIndex) @constant.builtin
+(lark) @operator
+(fullContext) @constant.builtin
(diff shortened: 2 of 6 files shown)


amaanq:
experimental is set in parsers.lua (you can leave it in the readme, but an action just auto updates that bit)

lua/nvim-treesitter/parsers.lua
```diff
@@ -694,6 +694,14 @@ list.hocon = {
   maintainers = { "@antosha417" },
 }
 
+list.hoon = {
+  install_info = {
+    url = "https://github.com/urbit-pilled/tree-sitter-hoon",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@urbit-pilled" },
```
amaanq, line 702:
```suggestion
  maintainers = { "@urbit-pilled" },
  experimental = true,
```

queries/hoon/highlights.scm
```diff
@@ -0,0 +1,32 @@
...
+  "]"
+]  @punctuation.bracket
+
+[
+  (coreTerminator)
+  (seriesTerminator)
+] @punctuation.delimiter
+
+
+(rune) @keyword
```
amaanq, line 18: from the looks of it, isn't this an operator (or punctuation.(delimiter|special)?

queries/hoon/highlights.scm
```diff
@@ -0,0 +1,32 @@
...
+
+(term) @constant
+
+(aura) @constant.builtin
+
+(Gap) @comment
+
+(boolean) @constant.builtin
+
+(date) @constant.builtin
```
amaanq, line 28: Maybe `@string.special`?

queries/hoon/highlights.scm
```diff
@@ -0,0 +1,32 @@
...
+(mold) @constant.builtin
```
amaanq, line 29:
```suggestion
(mold) @operator
```
Or `@punctuation.special`

queries/hoon/highlights.scm
```diff
@@ -0,0 +1,32 @@
...
+(specialIndex) @constant.builtin
```
amaanq, line 30: Maybe `@number.builtin`? Need more context on this one

queries/hoon/highlights.scm
```diff
@@ -0,0 +1,32 @@
...
+(lark) @operator
+(fullContext) @constant.builtin
```
amaanq, line 32:
```suggestion
(fullContext) @punctuation.delimiter
```

amaanq (changes requested):

queries/hoon/highlights.scm
urbit-pilled: correct, a rune should be an operator

queries/hoon/highlights.scm
urbit-pilled: Here's the documentation for [molds](https://developers.urbit.org/reference/hoon/irregular#trivial-molds), it's a symbolic alias for a type, (noun, atom cell, boolean, null). So I think this should be `@symbol`

queries/hoon/highlights.scm
urbit-pilled: The specialIndex rule represents 3 different ways to index an array:
- `&n` nth element
- `|n` tail of list after nth element
- `+n` slot n (tree-addressing)

all of which are converted to an numerical index, so I think `@number.builtin` would be suitable.
https://developers.urbit.org/reference/hoon/irregular#list-addressing

queries/hoon/highlights.scm
urbit-pilled: This one is a similar to mold, it is a symbol alias. So I think `@symbol` would be suitable.

queries/hoon/highlights.scm
urbit-pilled: right

amaanq (approved):

PR #5043, opened 2023-06-29
feat(sql): json operator highlights

dmfay:
ref https://github.com/DerekStride/tree-sitter-sql/pull/170


amaanq:
this should target master

matthias-Q:
Nope, we have the release version on a separate branch. Since it is not merged yet, we do not know the commit hash for the lockfile. This is why it is a draft at the moment.

amaanq:
I assumed it was merged

matthias-Q:
It is merged, yes but we keep the parser.c in a separate branch to avoid merge conflicts. The relevant artifact is in the branch gh-pages

amaanq:
Ah - so is this just waiting for the workflow then? I thought main had a different branch for sql, but if that's the case master should be targeted for the pr then

dmfay:
oh that's worse, will reopen

PR #5053, opened 2023-07-01
fix: Add `force` to all custom predicates/directives

lucario387:
`trim!` was upstreamed. Adding `force = true` to skim error

### DIFF
--- a/lua/nvim-treesitter/query_predicates.lua
+++ b/lua/nvim-treesitter/query_predicates.lua
@@ -57,7 +57,7 @@ query.add_predicate("nth?", function(match, _pattern, _bufnr, pred)
   end
 
   return false
-end)
+end, true)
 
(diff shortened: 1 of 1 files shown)


clason:
Might as well do that for all of them. (I never understood why this was optional behavior.)

clason (approved):

lucario387:
> Might as well do that for all of them. (I never understood why this was optional behavior.)

Except for `exclude_children` and `set-lang*`, as i don't think they will ever be upstreamed anyways?

clason:
Probably not (but `set-lang-from-mimetype` might), but I don't see the drawback in either case?

lucario387:
> Probably not (but `set-lang-from-mimetype` might), but I don't see the drawback in either case?

Agreed, I just added it to all predicates/directives

PR #5061, opened 2023-07-04
refactor(v)!: update repository

amaanq:
Old one had a memory leak and was archived

relevant convo https://github.com/vlang/vls/pull/446

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1600,12 +1600,12 @@ list.uxntal = {
 
 list.v = {
   install_info = {
-    url = "https://github.com/vlang/vls",
+    url = "https://github.com/v-analyzer/v-analyzer",
     files = { "src/parser.c", "src/scanner.c" },
     location = "tree_sitter_v",
   },
   filetype = "vlang",
-  maintainers = { "@kkharji" },
+  maintainers = { "@kkharji", "@amaanq" },
 }
 
 list.vala = {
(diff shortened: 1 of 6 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -1600,12 +1600,12 @@ list.uxntal = {
...
 list.v = {
   install_info = {
-    url = "https://github.com/vlang/vls",
+    url = "https://github.com/v-analyzer/v-analyzer",
     files = { "src/parser.c", "src/scanner.c" },
     location = "tree_sitter_v",
   },
-  filetype = "vlang",
-  maintainers = { "@kkharji" },
+  filetype = "v",
```
clason, line 1607: That's an unrelated change. Any reason for it?

clason (commented):
Also needs a lockfile update.

lucario387:
Bit ootl here, but shouldn't the URL be https://github.com/v-analyzer/v-tree-sitter ? Or is there a reason why there are two separate tree-sitter parsers for v

clason:
foo-tree-sitter != tree-sitter-foo

(the former is bindings for tree-sitter in the foo language, the latter a tree-sitter parser for the foo language)

lucario387:
> (the former is bindings for tree-sitter in the foo language, the latter a tree-sitter parser for the foo language)

Yep, I didn't read that correctly at first. Should have been more aware

lucario387 (approved):

PR #5072, opened 2023-07-06
fix(ecma): Highlight regex flags

savq:
I don't know if the flag should be highlighted the same as the regex, but it does look weird without any highlight at all.

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -190,6 +190,7 @@
 (template_string) @string
 (escape_sequence) @string.escape
 (regex_pattern) @string.regex
+(regex_flags) @character.special
 (regex "/" @punctuation.bracket) ; Regex delimiters
 
 (number) @number


amaanq:
maybe `@symbol`, `@punctuation.special`, `@character.special`? Up to you though

lucario387:
I'd prefer it to not be string.regex, probably `@symbol` or `@property`, or some sort? As long as the capture usage here doesn't go too far from denoting `this is some extra stuffs for the regex`

ObserverOfTime:
I'd use `@character.special`.

savq:
I did a little bit of research and found that flags really are stored as strings [1], and the grammar parses them as `token.immediate(/[a-z]+/)`, so I'd consider using `@string.special`.

I suspect some colorschemes will use the same highlight for `@string.special` and `@string.regex`, so using `@character.special` is fine too. Let me know if I should change it.

[1]: https://262.ecma-international.org/#sec-get-regexp.prototype.flags

savq:
Makes sense. I updated the PR. It's ready to merge.

PR #4449, opened 2023-03-05
fix: Java indent for multiple line arguments

dwoznicki:
Fixes #4448 

Both `(constructor_declaration)` and `(constructor_body)` are marked for indent, but it looks like we only need to indent one of them.

NOTE: Java indent tests are not passing for this PR. When I run `./scripts/run_tests.sh indent/java_spec.lua`, I get a failure for the "whole file" test.

```
Fail	||	indent Java: whole file: ./constructor_with_arguments_on_multiple_lines.java
            ./tests/indent/common.lua:80: Incorrect indentation
                       Found:                                       Expected:
                        |public class Testo {                        |public class Testo {
                        |  public Testo(                             |  public Testo(
                        |    String a                                |    String a
                        |  ) {                                       |  ) {
              2 vs 0 => |  }                                         |}

            stack traceback:
            	./tests/indent/common.lua:80: in function 'compare_indent'
            	./tests/indent/common.lua:122: in function 'indent_whole_file'
            	./tests/indent/common.lua:187: in function <./tests/indent/common.lua:186>
```

The problem is that the expected indentation for the "whole file" test is wrong. The test expects an indentation of 0 for the closing `}`, but the actual + correct indentation is 2. I'm not sure how to solve this, so any advice would be appreciated.

PR #5078, opened 2023-07-07
feat: add Cairo

amaanq:
It supports both Cairo 0.x and 1.0 syntax

### DIFF
--- a/queries/cairo/highlights.scm
+++ b/queries/cairo/highlights.scm
@@ -1,0 +1,338 @@
+; Preproc
+
+[
+  "%builtins"
+  "%lang"
+] @preproc
+
+; Includes
+
+(import_statement [ "from" "import" ] @include module_name: (dotted_name (identifier) @namespace . ))
+
+[
+  "as"
+  "use"
+  "mod"
+] @include
+
+; Variables
+
+(identifier) @variable
+
+; Namespaces
+
+(namespace_definition (identifier) @namespace)
+
+(mod_item
+  name: (identifier) @namespace)
+
+(use_list (self) @namespace)
+
+(scoped_use_list (self) @namespace)
+
+(scoped_identifier
+  path: (identifier) @namespace)
+
+(scoped_identifier
+ (scoped_identifier
+  name: (identifier) @namespace))
+
+(scoped_type_identifier
+  path: (identifier) @namespace)
+
+((scoped_identifier
+  path: (identifier) @type)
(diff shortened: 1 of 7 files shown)


queries/cairo/indents.scm
```diff
@@ -0,0 +1,44 @@
+[
+  (import_statement)
```
lucario387, line 2: I don't think this should be here.

If it should, then restricting to `(import_statement (. "from"))` or it having braces/parenthesis will be better

amaanq: yeah you're right it needs a parenthesis

queries/cairo/indents.scm
```diff
@@ -0,0 +1,45 @@
...
+  (match_arm)
+  (match_block)
+  (call_expression)
+  (assignment_expression)
+  (arguments)
+  (block)
+  (use_list)
+  (field_declaration_list)
+  (enum_variant_list)
+  (tuple_pattern)
```
lucario387, line 18: Are the items in the list completely non-overlapping?

Example is like #4449, where two captures overlap with one another, leading to doubly-indented patterns

amaanq: Hmm, for the most part I'd say no, Cairo 0.x is python-like and Cairo 1.0 is rust-like, so I mostly just copied the indents that applied, though I'll check some of the potential candidates for double-indents

amaanq: In a Cairo file with just this
```cairo
struct Span<T> {
    snapshot: @Array<T>
}
```

Why would gg=G dedent the field?

lucario387: gg=G only considers the first node of each line, while o considers the last node of previous non blank line as well

queries/cairo/highlights.scm
```diff
@@ -0,0 +1,340 @@
...
+(namespace_definition (identifier) @namespace)
+
+(mod_item
+  name: (identifier) @namespace)
+
+(use_list (self) @namespace)
+
+(scoped_use_list (self) @namespace)
+
+(scoped_identifier (self) @namespace)
```
ObserverOfTime, line 33: Shouldn't `(self)` be `@variable.builtin`?

amaanq: I just took that from the Rust highlights since, functionally, they're the same in Cairo, but I would agree with you

ObserverOfTime (approved):

lucario387 (approved):

PR #5082, opened 2023-07-08
feat(markdown_inline): add queries for wiki links and tags

musjj:
Still new to treesitter queries, feedback would be appreciated.
image

### DIFF
--- a/queries/markdown_inline/highlights.scm
+++ b/queries/markdown_inline/highlights.scm
@@ -15,6 +15,8 @@
 
 (strikethrough) @text.strike
 
+(tag) @tag
+
 [
   (link_destination)
   (uri_autolink)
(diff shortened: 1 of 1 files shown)


amaanq:
this will fail ci as we don't compile with these options on, but you're free to add these in your own config extending nvim ts

musjj:
Alright.
I added them to my `~/.config/nvim/after/queries/highlights.scm` instead but it doesn't seem to work. I ran `vim.api.nvim_get_runtime_file("queries/*/*.scm", true)` and my path was listed there, but the highlights doesn't seem to be working.

amaanq:
First thing I noticed is that it should be in `queries/markdown/highlights.scm`, and if that doesn't fix it then did you add `; extends` at the top of the file? Lastly, not sure if it needs to be in after but that should work either way I'd guess

musjj:
The `; extends` did the trick, thanks!

PR #5086, opened 2023-07-08
feat(promql,yaml): add parser and yaml injections

MichaHoffmann:
This is an effort so we can get promql highlighted in prometheus rule manifests:

image

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1179,6 +1179,15 @@ list.prisma = {
   maintainers = { "@elianiva" },
 }
 
+list.promql = {
+  install_info = {
+    url = "https://github.com/MichaHoffmann/tree-sitter-promql",
+    files = { "src/parser.c" },
+    experimental = true,
+  },
+  maintainers = { "@MichaHoffmann" },
+}
+
 list.proto = {
   install_info = {
     url = "https://github.com/mitchellh/tree-sitter-proto",
--- a/queries/promql/highlights.scm
+++ b/queries/promql/highlights.scm
@@ -1,0 +1,54 @@
+; highlights.scm
+
+[
+  "*"
+  "/"
+  "%"
+  "+"
+  "-"
+  ">"
+  ">="
+  "<"
+  "<="
+  "="
+  "=~"
+  "!="
+  "!~"
+] @operator
+
+[
+  "and"
+  "unless"
+  "or"
+  "bool"
+] @keyword.operator
+
+[
+  "{"
+  "}"
+  "["
+  "]"
+  "("
+  ")"
+] @punctuation.bracket
+
+(float_literal) @float
+(string_literal) @string
+
+(metric_name) @type
+(range_selection) @text.strong @type
+(subquery_range_selection) @text.strong @type
+
+(label_name) @field
+(label_value) @text.underline @string.regex
+
+(function_name) @function.call
+
+[ 
+  "by"
+  "without"
+] @function
+
+(comment) @comment @spell
+(ERROR) @error
+
--- a/queries/yaml/injections.scm
+++ b/queries/yaml/injections.scm
@@ -25,3 +25,29 @@
              (block_sequence_item
                (block_node
                   (block_scalar) @bash (#offset! @bash 0 1 0 0))))))
+
+;; Prometheus Alertmanager ("expr")
+(block_mapping_pair
+  key: (flow_node) @_expr (#eq? @_expr "expr")
+  value: (flow_node
+           (plain_scalar) @promql))
+
+(block_mapping_pair
+  key: (flow_node) @_expr (#eq? @_expr "expr")
+  value: (block_node
+           (block_scalar) @promql (#offset! @promql 0 2 0 0)))
+
+(block_mapping_pair
+  key: (flow_node) @_expr (#eq? @_expr "expr")
+  value: (block_node
+           (block_sequence
+             (block_sequence_item
+               (flow_node) @promql))))
+
+(block_mapping_pair
+  key: (flow_node) @_expr (#eq @_expr "expr")
+  value: (block_node
+           (block_sequence
+             (block_sequence_item
+               (block_node
+                  (block_scalar) @promql (#offset! @promql 0 2 0 0))))))
--- a/lockfile.json
+++ b/lockfile.json
@@ -392,6 +392,9 @@
   "prisma": {
     "revision": "eca2596a355b1a9952b4f80f8f9caed300a272b5"
   },
+  "promql": {
+    "revision": "655afc4fe6813f38bde087d6493d8fd4920d6d4a"
+  },
   "proto": {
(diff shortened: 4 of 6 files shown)


queries/promql/highlights.scm
```diff
@@ -0,0 +1,47 @@
+; highlights.scm
+
+[
+  "\*"
```
lucario387, line 4: Is there any specific reason for escaped * here

queries/yaml/injections.scm
```diff
@@ -25,3 +25,14 @@
              (block_sequence_item
                (block_node
                   (block_scalar) @bash (#offset! @bash 0 1 0 0))))))
+
+;; Prometheus Alertmanager ("expr")
+(block_mapping_pair
+  key: (flow_node) @_expr (#eq? @_expr "expr")
+  value: (flow_node
+           (plain_scalar) @promql))
```
lucario387, line 33: I am a bit hesitant on this capture. Shouldn't this be more narrowed down as there should be lots of cases where the block in question here isn't promql?

MichaHoffmann: It could match non-promql stuff i think; but wouldnt same apply to the `bash` captures at the top of the file?

lucario387: Well yes, but bash is more used than promql (I believe).

I'm just a bit against it because I have no idea how often this is used, and how often it will lead to unwanted promql injections. For `bash` captures at least there's a ton of workflow files on Github and GitLab to check for usage to check for frequency

queries/promql/highlights.scm
MichaHoffmann: Ah, probably not; thanks! Was it necessary at some point? I think i remember that it was but dont know for sure!

queries/yaml/injections.scm
MichaHoffmann: Would a more specific match be acceptable?

lucario387: Anything that narrows it down to be more specific will be preferred.

If the current queries are of such state, I'd like to have some guidelines and documentations for promql usage that uses such captures.

MichaHoffmann: Ok, ill narrow it down some more and write a unittest! Sorry i didn't get the last part, can you rephrase please?

lua/nvim-treesitter/parsers.lua
```diff
@@ -1179,6 +1179,16 @@ list.prisma = {
   maintainers = { "@elianiva" },
 }
 
+list.promql = {
+  install_info = {
+    url = "https://github.com/MichaHoffmann/tree-sitter-promql",
+    files = { "src/parser.c" },
+    branch = "main",
```
amaanq, line 1186: if it's the default branch this can be removed

queries/promql/highlights.scm
```diff
@@ -0,0 +1,47 @@
...
+
+(function_name) @function.call
+(aggregation_operator) @function
+(binary_grouping) @function
+(aggregation_grouping) @function
+
+(range_selection) @string.special
+(subquery_range_selection) @string.special
+
+(comment) @comment
```
amaanq, line 46: `@spell` too?

lua/nvim-treesitter/parsers.lua
clason: ...but you need to add the parser to lockfile.json

queries/yaml/injections.scm
lucario387: What I mean is that if you believe that the current version of the queries are suffice, I'd like to see some documentations and guidelines on why that's the case. I'm okay with the queries right now, though that's with some concerns on the frequency of mis-injection

mhoffm-aiven: Oh , I mainly just tweaked until highlighting looked fine, it was not a rigorous process. Or do you mean specifically for the injections?

lucario387: Yes, I mean the injections. Is there any documentations or guidelines on the possible key-value pairs that should have its value be `promql` code, and if so, what are the chances of such key-value pairs not being promql code.

If the false injection rate is too high, I'd like to drop the injections, and you can add it locally in your config. The highlights looks fine to me

MichaHoffmann: The injections are intended for https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/ ( and alerting rules );

lua/nvim-treesitter/parsers.lua
MichaHoffmann: :heavy_check_mark:

queries/promql/highlights.scm
MichaHoffmann: :heavy_check_mark:

queries/yaml/injections.scm
lucario387: Lgtm.

MichaHoffmann: I didnt manage to write a more scoped injection and couldnt get dynamic detection with the "@language" directive to work. Should i take the injections out and define them locally?

clason: if in doubt, be conservative, especially with injections (which can be performance killers). it's always possible to add them later (which can then be in the upstream format, saving me from having to rewrite them for the `main` branch).

MichaHoffmann: Oh, sorry i didnt get the last part; whats upstream format? I dont feel too bad about the injections so far; if they are contentious i can take them out and define them locally but since there is no "promql" filetype; the only place where i would want the parser to be active is injected into yaml right now.

clason: I mean the format described in the [official tree-sitter docs](https://tree-sitter.github.io/tree-sitter/syntax-highlighting#language-injection), which we're switching to as soon as we drop support for Neovim 0.8.x.

And no filetype is not a blocker for adding the parser; you can always register it with `vim.filetype.add` (or `vim.treesitter.start` manually), but if injection is the _only_ real use case for this parser, leaving them out is indeed somewhat pointless.

MichaHoffmann: Ah, what i mean is that those promql statements do not appear as a format of a whole file; they only ever appear in those yaml injections ( to my knowledge at least )

amaanq (approved):

ObserverOfTime (approved):

amaanq:
@MichaHoffmann can I implore you to revisit https://github.com/MichaHoffmann/tree-sitter-hcl/pull/35?

MichaHoffmann:
> @MichaHoffmann can I implore you to revisit [MichaHoffmann/tree-sitter-hcl#35](https://github.com/MichaHoffmann/tree-sitter-hcl/pull/35)?

Im struggling with it somewhat since it will break downstream projects probably ( nvim-treesitter is not the only one; i know that semgrep is also using it and the emacs treesitter plugin i think also )

amaanq:
if it's any consolation - several upstream tree-sitter grammars got rewritten as well with little to no detected harm (python, cpp, bash, etc)

PR #5094, opened 2023-07-11
feat(julia): highlight docstrings before short function definitions

wangl-cc:
`string_literal` before a short function definition should be docstring:
```
In [1]: begin
            """
                foo()

            This is a docstring before short function definition.
            """
            foo() = nothing
        end
Out[1]: foo

help?> foo
search: foo ...

  foo()

  This is a docstring before short function definition.
```

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -286,6 +286,7 @@
       (abstract_definition)
       (struct_definition)
       (function_definition)
+      (short_function_definition)
       (assignment)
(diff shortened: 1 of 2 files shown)


lucario387:
This isn't correct.

Accoriding to your own words, `string_literal` should be the node to be captured here

But what `string_literal` to be captured, you haven't specified

savq:
LGTM.

clason (approved):

PR #5097, opened 2023-07-12
feat: add SystemTap

ok-ryoko:
[SystemTap] is a Red Hat-driven Linux instrumentation framework that’s been around since 2005. It exposes a scripting language inspired by C and awk. I’ve written a complete [grammar] and have shown up to contribute nvim-treesitter queries.

Here’s the highlights query in action on some of my code:

systemtap-nvim-treesitter-highlights

There’s very little SystemTap in the wild, so I would consider this PR to be a relatively low priority.

Your feedback is greatly appreciated, even if this PR doesn’t succeed. Thank you for your work on nvim-treesitter!

[grammar]: https://github.com/ok-ryoko/tree-sitter-systemtap
[SystemTap]: https://sourceware.org/systemtap/

### DIFF
--- a/queries/systemtap/highlights.scm
+++ b/queries/systemtap/highlights.scm
@@ -109,7 +109,6 @@
 
 [
   "delete"
-  "global"
   "limit"
   "next"
   "private"
@@ -150,3 +149,5 @@
 ] @preproc
 
 "@define" @define
+
+"global" @storageclass
--- a/queries/systemtap/highlights.scm
+++ b/queries/systemtap/highlights.scm
@@ -99,6 +99,8 @@
 ] @punctuation.delimiter
 
 [
+  "%{"
+  "%}"
   "("
   ")"
   "["
@@ -142,8 +144,6 @@
   "%)"
   "%:"
   "%?"
-  "%{"
-  "%}"
   (preprocessor_tokens)
   (embedded_code)
 ] @preproc
--- a/queries/systemtap/highlights.scm
+++ b/queries/systemtap/highlights.scm
@@ -1,7 +1,7 @@
 (identifier) @variable
 
 (preprocessor_macro_definition
-  name: (identifier) @define)
+  name: (identifier) @function.macro)
 
 (preprocessor_macro_expansion) @function.macro
 
@@ -26,7 +26,7 @@
 
 (type) @type.builtin
 
-(aggregation_operator) @function.macro
+(aggregation_operator) @attribute
 
 (member_expression
   member: (identifier) @field)
@@ -108,8 +108,6 @@
 ] @punctuation.bracket
 
 [
-  "break"
-  "continue"
   "delete"
   "global"
   "limit"
@@ -128,6 +126,8 @@
 ] @conditional
 
 [
+  "break"
+  "continue"
   "for"
   "foreach"
   "while"
--- a/queries/systemtap/highlights.scm
+++ b/queries/systemtap/highlights.scm
@@ -34,18 +34,14 @@
 (call_expression
   function: (identifier) @function.call)
 
-(
-  (call_expression
+((call_expression
     function: (identifier) @function.builtin)
   (#any-of? @function.builtin
     "print" "printd" "printdln" "printf" "println"
-    "sprint" "sprintd" "sprintdln" "sprintf" "sprintln")
-)
+    "sprint" "sprintd" "sprintdln" "sprintf" "sprintln"))
 
-(
-  (identifier) @variable.builtin
-  (#lua-match? @variable.builtin "^\$+[0-9A-Z_a-z]+\$*$")
-)
+((identifier) @variable.builtin
+  (#lua-match? @variable.builtin "^\$+[0-9A-Z_a-z]+\$*$"))
 
 (shebang_line) @preproc
 
--- a/queries/systemtap/highlights.scm
+++ b/queries/systemtap/highlights.scm
@@ -113,7 +113,6 @@
   "delete"
   "limit"
   "next"
-  "private"
   "probe"
 ] @keyword
 
@@ -150,4 +149,5 @@
 
 "@define" @define
 
+"private" @type.qualifier
 "global" @storageclass
--- a/queries/systemtap/highlights.scm
+++ b/queries/systemtap/highlights.scm
@@ -1,0 +1,156 @@
+(identifier) @variable
+
+(preprocessor_macro_definition
+  name: (identifier) @define)
+
+(preprocessor_macro_expansion) @function.macro
+
+(preprocessor_constant) @constant.macro
+
+(number) @number
+(string) @string
+(escape_sequence) @string.escape
+
+[
+  (script_argument_string)
+  (script_argument_number)
+] @constant
+
+(probe_point_component) @function
+
+(function_definition
+  name: (identifier) @function)
+
+(parameter
+  name: (identifier) @parameter)
+
+(type) @type.builtin
+
+(aggregation_operator) @function.macro
+
+(member_expression
+  member: (identifier) @field)
+
+(call_expression
+  function: (identifier) @function.call)
+
+(
+  (call_expression
+    function: (identifier) @function.builtin)
+  (#any-of? @function.builtin
+    "print" "printd" "printdln" "printf" "println"
+    "sprint" "sprintd" "sprintdln" "sprintf" "sprintln")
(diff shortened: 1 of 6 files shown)


queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
+(identifier) @variable
+
+(preprocessor_macro_definition
+  name: (identifier) @define)
```
ObserverOfTime, line 4:
```suggestion
  name: (identifier) @function.macro)
```

queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+
+(preprocessor_macro_expansion) @function.macro
+
+(preprocessor_constant) @constant.macro
```
ObserverOfTime, line 8:
```suggestion
(preprocessor_constant) @constant.builtin
```

queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+
+(function_definition
+  name: (identifier) @function)
+
+(parameter
+  name: (identifier) @parameter)
+
+(type) @type.builtin
+
+(aggregation_operator) @function.macro
```
ObserverOfTime, line 29:
```suggestion
(aggregation_operator) @attribute
```

queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+
+(member_expression
+  member: (identifier) @field)
+
+(call_expression
+  function: (identifier) @function.call)
+
+(
+  (call_expression
```
ObserverOfTime, line 38: Formatting.

```suggestion
((call_expression
```

queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+    function: (identifier) @function.builtin)
+  (#any-of? @function.builtin
+    "print" "printd" "printdln" "printf" "println"
+    "sprint" "sprintd" "sprintdln" "sprintf" "sprintln")
+)
```
ObserverOfTime, line 43:
```suggestion
    "sprint" "sprintd" "sprintdln" "sprintf" "sprintln"))
```

queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+
+(
+  (identifier) @variable.builtin
+  (#lua-match? @variable.builtin "^\$+[0-9A-Z_a-z]+\$*$")
+)
```
ObserverOfTime, line 48:
```suggestion
((identifier) @variable.builtin
  (#lua-match? @variable.builtin "^\$+[0-9A-Z_a-z]+\$*$"))
```

queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+  "catch"
+] @exception
+
+[
+  "%("
+  "%)"
+  "%:"
+  "%?"
+  "%{"
+  "%}"
```
ObserverOfTime, line 151: Why not `@punctuation.bracket`?

ObserverOfTime (changes requested):
I don't quite understand what these nodes represent:
```
(script_argument_string)
(script_argument_number)
(preprocessor_tokens)
(embedded_code)
```

queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+  "]"
+  "{"
+  "}"
+] @punctuation.bracket
+
+[
+  "break"
+  "continue"
+  "delete"
+  "global"
```
amaanq, line 118: Could be `@storageclass`

queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
+] @punctuation.bracket
+
+[
+  "break"
+  "continue"
```
amaanq, line 116: Should be `@repeat`

queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+  "("
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
+] @punctuation.bracket
+
+[
+  "break"
```
amaanq, line 115: Should be `@conditional` or `@repeat`

queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+  "continue"
+  "delete"
+  "global"
+  "limit"
+  "next"
+  "private"
```
amaanq, line 121: Could be `@type.qualifier`

queries/systemtap/highlights.scm
```diff
@@ -0,0 +1,156 @@
...
+] @exception
+
+[
+  "%("
+  "%)"
+  "%:"
+  "%?"
+  "%{"
+  "%}"
+  (preprocessor_tokens)
```
amaanq, line 152: Maybe this can be injected w/ itself? If it's anything like Rust/C

amaanq (changes requested):

amaanq:
Also you need an entry in `lockfile.json` w/ the latest revision @ok-ryoko

queries/systemtap/highlights.scm
ok-ryoko: I’m wary of using `@type` here since `private` can be used only for variable declarations and function definitions. Moreover, SystemTap has no user-defined types.

queries/systemtap/highlights.scm
ok-ryoko: The tokens `%(`, `%)`, `%:` and `%?` are meaningful only to the preprocessor, being interpreted in a separate pass before the SystemTap script gets parsed.

The delimiters `%{` and `%}` are used only for embedded C code, which is transcribed verbatim.

Another reason for choosing distinct captures for these tokens is to achieve consistency with the behavior of the language server.

queries/systemtap/highlights.scm
amaanq: private is a qualifier though, isn't it? `@type.qualifier` really works for any sort of qualifier

queries/systemtap/highlights.scm
ok-ryoko: I have addressed this in [my reply](https://github.com/nvim-treesitter/nvim-treesitter/pull/5097#issuecomment-1634546706) to the main conversation. Could you please take a look there?

queries/systemtap/highlights.scm
ok-ryoko: Is it the convention in nvim-treesitter to be able to use `@type.qualifier` to qualify entities other than types? If so, then I will update the capture.

ok-ryoko:
Would you also like me to set definition scopes in the locals query? Would that constitute a meaningful improvement to the submission quality?

queries/systemtap/highlights.scm
ObserverOfTime: One day it will be renamed to something more sensible…

ObserverOfTime:
> script_argument_string and script_argument_number refer to the special identifiers @1, …, @99 and $1, …, $99, respectively. These represent indexed command-line arguments to the script. For example, if your script accepts an integer as the first command-line argument, then you could access that integer using $1.

How about `@parameter` then?

>  Does nvim-treesitter expose a quick toggle for injections?

We have a toggle for highlights but not injections. Maybe we should. :thinking:

ok-ryoko:
In review of my comments, I have not made a compelling case against injecting C highlights. Having highlights for embedded C would help differentiate embedded blocks from preprocessor blocks. Moreover, the location of embedded C in SystemTap scripts is well-defined, whereas preprocessor blocks can appear virtually anywhere (although the grammar doesn’t allow that). Thus, I added an injections capture and switched to `@punctuation.bracket` for `%{` and `%}`.

> How about `@parameter` then?

My impression of the `@parameter` capture is that it is reserved for declarations, as in the C highlights query. If we used `@parameter` to color the script arguments in the script body, then I would expect to have to use `@parameter` to color every function’s parameters in the corresponding function body, locals notwithstanding. From my reading of nvim-treesitter queries, this doesn’t seem to be a common pattern.

ObserverOfTime (approved):

amaanq (approved):

PR #5113, opened 2023-07-17
highlights(typescript): decorators, default case, global declaration

zweimach:
| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/queries/typescript/highlights.scm
+++ b/queries/typescript/highlights.scm
@@ -136,6 +136,9 @@
 (arrow_function
   parameter: (identifier) @parameter)
 
+;; global declaration
+(ambient_declaration "global" @namespace)
+
 ;; function signatures
 (ambient_declaration
   (function_signature
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -158,6 +158,11 @@
 (namespace_import
   (identifier) @namespace)
 
+; Decorators
+;----------
+(decorator "@" @attribute (identifier) @attribute)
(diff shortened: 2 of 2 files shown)


queries/typescript/highlights.scm
```diff
@@ -136,6 +136,9 @@
 (arrow_function
   parameter: (identifier) @parameter)
 
+;; global declaration
+(ambient_declaration "global" @namespace)
```
amaanq, line 140:
```suggestion
(ambient_declaration "global" @storageclass)
```

amaanq (changes requested):

queries/typescript/highlights.scm
zweimach: Thanks for the suggestion.
I'm certainly not a TypeScript expert, but isn't `declare global {}` symmetrical to `declare namespace SomeClass {}`?

amaanq: Me neither - how about `@keyword` then?

zweimach: It could be, but I picked `@namespace` because of semantic tokens from `tsserver`.
I might be wrong and this is a semantic token from CoC.nvim.
Forgive me for my Neovim setup, I haven't finished my migration to Neovim LSP yet. 😆

image

ObserverOfTime: `@namespace` seems fine to me.

ObserverOfTime (approved):

amaanq (approved):

PR #5120, opened 2023-07-18
chore(c): update queries from upstream

amaanq:
@lucario387 (indents)

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -33,7 +33,7 @@
     "revision": "7f1a5df44861291d6951b6b2146a9fef4c226e14"
   },
   "c": {
-    "revision": "6adee194587678b250608cdb808544f06bcd26e7"
+    "revision": "51cd29c3d1b0f94dfd17ec4d94525344a1aeebab"
   },
   "c_sharp": {
     "revision": "1648e21b4f087963abf0101ee5221bb413107b07"
--- a/lockfile.json
+++ b/lockfile.json
@@ -69,7 +69,7 @@
     "revision": "f4b3cbc8b0bd4e13035d39940fef09f1392e8739"
   },
   "cpp": {
-    "revision": "0e7b7a02b6074859b51c1973eb6a8275b3315b1d"
+    "revision": "cbb9974efde90cbdc5d1027c97882df737aa3601"
   },
   "css": {
     "revision": "5f2c94b897601b4029fedcce7db4c6d76ce8a128"
--- a/lockfile.json
+++ b/lockfile.json
@@ -6,7 +6,7 @@
     "revision": "80ea622cf952a0059e168e5c92a798b2f1925652"
   },
   "arduino": {
-    "revision": "4de2f3e6235ee8659ecb7467c16ed13bde7fb272"
+    "revision": "3446bac828d5db2a6cf60ab31813c02ab53547a7"
   },
   "astro": {
     "revision": "e122a8fcd07e808a7b873bfadc2667834067daf1"
@@ -33,7 +33,7 @@
     "revision": "7f1a5df44861291d6951b6b2146a9fef4c226e14"
   },
   "c": {
-    "revision": "fcd123005b7c0bf3b3a63c1dae85590ef4740464"
+    "revision": "ad095896dd223f1c22b85ac5ec84ab11fb732b07"
   },
   "c_sharp": {
     "revision": "1648e21b4f087963abf0101ee5221bb413107b07"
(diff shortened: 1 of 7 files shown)


lucario387:
Do I just commit onto that branch directly?

amaanq:
Yeah that works

lucario387:
@amaanq I think it's fine now.
I only remove one if-else case of the 20 tested

amaanq:
@fab4100 can you update ispc's C base to 0.20.3 and then update queries here

clason (approved):

clason:
Is ispc still the blocker here? Maybe we should decouple (some) of the queries, at least temporarily.

clason:
@amaanq yes, let's decouple the queries (at least for now) and mark `ispc` as Tier 4.

lucario387:
cpp: 
> `invalid node type: else_clause`

So seems like the commit hash was incorrect

fab4100:
@amaanq: The ISPC parser is updated, should I push changes/revert for the queries onto  `amaanq:update-c` ? (apologies for the delay, another task required priority)

lucario387:
I'd say you should do the revert on this branch.

lucario387:
Well whatever ispc did, it's not working ...

clason:
Now only the (inherited) `indents` are failing.

lucario387:
Most likely the same problem of `invalid node type: else_clause`

clason:
Tests no longer pass, though...

clason:
What's weird is that the failing languages _only_ inherit `cpp` and don't have any other queries themselves. And `cpp` itself is passing the tests.

fab4100:
Failing indents for ISPC inherit from `c` and the file has additional queries, not just an `; inherits: c` statement.  It seems not only `cpp` inheritance fails.

lucario387:
> Failing indents for ISPC inherit from `c` and the file has additional queries, not just an `; inherits: c` statement.  It seems not only `cpp` inheritance fails.

The problem is the same as tree-sitter-cpp a few days ago. It was bumped to 0.20.3, but 0.20.3 doesn't actually have the new else_clause node, only while using tree-sitter-c HEAD does such node appear

clason:
So what you're saying is "tree-sitter-cpp needs another release" 😬

lucario387:
> So what you're saying is "tree-sitter-cpp needs another release" 

Actually, `tree-sitter-c` is the one that needs to be bumped, which means cpp and all of its child :grimacing:

clason:
All I'm hearing is "needs moar release"...

lucario387:
Or "living on the edge" if we're to follow `tree-sitter-cpp`'s footstep (it references ts-c's head for grammar update)

clason:
only arduino and ispc indents are still failing @ObserverOfTime @fab4100

amaanq:
cpp hasn't been released (yet, maybe later today once jdrouhard reviews it)

ObserverOfTime:
I'll wait for the release to avoid changing it back and forth.

clason:
Makes sense; but there is some urgency here as this is blocking scheduled updates (and upstream Neovim bumps). At least ispc has no excuse for waiting ;)

fab4100:
@clason I have re-generated the parser based on `tree-sitter-c 0.20.4`

lucario387:
@amaanq tree-sitter-c 0.20.4 still doesn't have `else_clause` node ... or ispc somehow generates a non-else_clause parser

clason:
Hmm, ispc indents are still breaking...

lucario387:
The breaker is for sure still `else_clause` node not available. Targeting tree-sitter-c's `HEAD` and such node does appear though ...

lucario387:
@clason I will just comment out all the if-else test for now, and only have if-else with braces to be supported

Once `else_clause` becomes available for a release again I will uncomment them

Let's not hold this longer

amaanq:
> @amaanq tree-sitter-c 0.20.4 still doesn't have `else_clause` node ... or ispc somehow generates a non-else_clause parser

I just checked out v0.20.4 and can definitely see sym_else_clause

image

lucario387:
Well the generated version with 0.20.4 of ispc doesn't have such node tho, that's very weird. Or is there something I'm missing here

amaanq:
ISPC's grammar modifies if_statement

lucario387:
well no wonder ...
Welp, @fab4100 You have to manually re-modify the if_statement then, or let's not inherit c for `indents.scm` for ispc anymore

amaanq:
@fab4100 https://github.com/fab4100/tree-sitter-ispc/pull/2

clason:
> @clason I will just comment out all the if-else test for now, and only have if-else with braces to be supported
> 
> Once `else_clause` becomes available for a release again I will uncomment them
> 
> Let's not hold this longer

To be clear: I'm talking about the query check, not the tests!

clason:
@ObserverOfTime `cpp` has a new release now.

ObserverOfTime:
Updated. (`3446bac828d5db2a6cf60ab31813c02ab53547a7`)

clason:
Why is `ispc` indent query now failing again... 😭

clason:
OK, that was painful. I'd be interested in a *post mortem* about what exactly got us into this mess, and how to _not_ get there again.

amaanq:
finally.. :sweat_smile:

lucario387:
How the addition of one syntax node breaks the entire ecosystem, part one:

ObserverOfTime:
Clearly the solution is to remove C and its children. :^)

PR #5136, opened 2023-07-22
fix(promql): bump parser, fix queries

MichaHoffmann:
The removed tokens are now case insensitive via an ugly regular expression; we cannot use them in captures here anymore.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -396,7 +396,7 @@
     "revision": "eca2596a355b1a9952b4f80f8f9caed300a272b5"
   },
   "promql": {
-    "revision": "655afc4fe6813f38bde087d6493d8fd4920d6d4a"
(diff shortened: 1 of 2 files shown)


MichaHoffmann:
cc @clason

clason:
That seems like a step backwards, from the point of view of highlighting?

MichaHoffmann:
> That seems like a step backwards, from the point of view of highlighting?

Somewhat yeah but i need to fix that in the parser; Ill fix it tomorrow or early next week. This is supposed to unbreak the automatic updates for now.

clason:
Makes sense; then we can leave the PR here for a bit and you can update it when you got it fixed; if it takes too long, I'm fine merging this. But a day or two doesn't matter much.

amaanq (approved):

clason:
I want lockfile CI to pass again.

PR #5138, opened 2023-07-22
scfg: new parser

WhyNotHugo:
This introduces a new parser for scfg. The highlights.scm file was copied from the repository for tree-sitter-scfg (and I have retained the license accordingly).

I don't think that there is a mechanism to copy the scm files from the parser repository itself, but that would be ideal in this case, especially since upstream targets neovim in this particular case.

Cc: @rockorager

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1367,6 +1367,15 @@ list.scala = {
   maintainers = { "@stevanmilic" },
 }
 
+list.scfg = {
+  install_info = {
+    url = "https://git.sr.ht/~rockorager/tree-sitter-scfg",
+    files = { "src/parser.c" },
+    requires_generate_from_grammar = true,
+  },
+  maintainers = { "@WhyNotHugo" },
+}
+
 list.scheme = {
   install_info = {
     url = "https://github.com/6cdh/tree-sitter-scheme",
--- a/queries/scfg/highlights.scm
+++ b/queries/scfg/highlights.scm
@@ -1,0 +1,8 @@
+[
+ "{"
+ "}"
+] @punctuation.bracket
+
+(comment) @comment @spell
+(directive_name) @type
+(directive_params) @parameter
--- a/README.md
+++ b/README.md
@@ -327,6 +327,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [ruby](https://github.com/tree-sitter/tree-sitter-ruby) (maintained by @TravonteD)
 - [x] [rust](https://github.com/tree-sitter/tree-sitter-rust) (maintained by @amaanq)
 - [x] [scala](https://github.com/tree-sitter/tree-sitter-scala) (maintained by @stevanmilic)
+- [x] [scfg](https://git.sr.ht/~rockorager/tree-sitter-scfg) (maintained by @WhyNotHugo)
 - [ ] [scheme](https://github.com/6cdh/tree-sitter-scheme)
 - [x] [scss](https://github.com/serenadeai/tree-sitter-scss) (maintained by @elianiva)
 - [x] [slint](https://github.com/jrmoulton/tree-sitter-slint) (experimental, maintained by @jrmoulton)
--- a/lockfile.json
+++ b/lockfile.json
@@ -461,6 +461,9 @@
   "scala": {
     "revision": "8062487fb3b7f3ce1bb7ce1fd1c84bed60c75203"
   },
+  "scfg": {
+    "revision": "6deae0cbb458c849a4d1e2985093e9c9c32d7fd0"
+  },
   "scheme": {
     "revision": "ca8af220aaf2a80aaf609bfb0df193817e4f064b"
   },
--- a/queries/scfg/injections.scm
+++ b/queries/scfg/injections.scm
@@ -1,0 +1,1 @@
+(comment) @comment


clason:
You do realize that Neovim query order is different from upstream treesitter (and the captures are different), so highlight queries cannot be copied as-is?

lua/nvim-treesitter/parsers.lua
```diff
@@ -1367,6 +1367,17 @@ list.scala = {
   maintainers = { "@stevanmilic" },
 }
 
+list.scfg = {
+  install_info = {
+    url = "https://git.sr.ht/~rockorager/tree-sitter-scfg",
+    files = { "src/parser.c" },
+    requires_generate_from_grammar = true,
+    branch = "main",
```
clason, line 1375: That is not necessary, if it's the default branch. You do, however, have to add a lockfile entry.

queries/scfg/highlights.scm
```diff
@@ -0,0 +1,30 @@
...
+; The above copyright notice and this permission notice shall be included in all
+; copies or substantial portions of the Software.
+;
+; THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+; IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+; FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+; AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+; LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+; OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+; SOFTWARE.
```
clason, line 21: That's a rather egregious license-to-content ratio. I'd prefer a clean room query. (Licenses should be compatible, in any case.)

lua/nvim-treesitter/parsers.lua
WhyNotHugo: > That is not necessary, if it's the default branch.

`TSInstall scfg` fails if I don't specify this:

```
nvim-treesitter[scfg]: Error while checking out revision                                      
error: pathspec 'master' did not match any file(s) known to git
```

I think it doesn't use the default branch, but `master` instead.

queries/scfg/highlights.scm
WhyNotHugo: I guess the original author can confirm if they're okay with dropping the licence, but I can't remove it myself since it's not my code.

WhyNotHugo:
> You do realize that Neovim query order is different from upstream treesitter (and the captures are different), so highlight queries cannot be copied as-is?

I realise that this is the general case, but as I said in the patch description:

> upstream targets neovim in this particular case.

WhyNotHugo:
How do I update the lockfile? I'm sure I've done this before, but I can't find it in CONTRIBUTING or README.

lua/nvim-treesitter/parsers.lua
clason: Yes, as I said, you need to add a lockfile entry.

queries/scfg/highlights.scm
clason: Just write your own code, then. I'm sure you ca do better than two or three random captures.

clason:
> How do I update the lockfile? I'm sure I've done this before, but I can't find it in CONTRIBUTING or README.

Add a new entry for the parser, and put the latest commit hash there.

queries/scfg/highlights.scm
rockorager: I'm perfectly fine removing the license, or @WhyNotHugo if you want to wholesale redo the queries or the parser I'm good with that.

WhyNotHugo: Thanks! I've removed the licence for now. I might try and improve upon this later, but, TBH, it works fine as-is so far.

WhyNotHugo:
Added entry to lockfile.

ObserverOfTime (approved):

queries/scfg/highlights.scm
```diff
@@ -0,0 +1,8 @@
+[
+ "{"
+ "}"
+ ] @punctuation.bracket
```
clason, line 4:
```suggestion
] @punctuation.bracket
```

queries/scfg/highlights.scm
```diff
@@ -0,0 +1,8 @@
...
+
+(comment) @comment @spell
```
clason, line 6: You probably want to add an injection for the comment parser, too.

WhyNotHugo: Done

queries/scfg/highlights.scm
WhyNotHugo: Done

WhyNotHugo:
Any particular reason for dropping the commit message and changing the author of the commit?

clason:
Just a short note: The parser does not include a `grammar.json`, which blocks using `tree-sitter` (consistently) for generating and building parsers. Unless this changes, we're probably going to drop the parser in 1.0 (on `main`).

(I'm also not crazy about parsers hosted on sr.ht; there's been too much outage in the past breaking our CI. It also prevents opening issues and PRs.)

PR #5145, opened 2023-07-25
feat: add requirements parser

ObserverOfTime:
Experimental because it's incomplete.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1310,6 +1310,15 @@ list.rego = {
   maintainers = { "@FallenAngel97" },
 }
 
+list.requirements = {
+  install_info = {
+    url = "https://github.com/ObserverOfTime/tree-sitter-requirements",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@ObserverOfTime" },
+  readme_name = "pip requirements",
+}
+
 list.rnoweb = {
   install_info = {
     url = "https://github.com/bamonroe/tree-sitter-rnoweb",
--- a/README.md
+++ b/README.md
@@ -320,6 +320,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [rasi](https://github.com/Fymyte/tree-sitter-rasi) (maintained by @Fymyte)
 - [x] [regex](https://github.com/tree-sitter/tree-sitter-regex) (maintained by @theHamsta)
 - [x] [rego](https://github.com/FallenAngel97/tree-sitter-rego) (maintained by @FallenAngel97)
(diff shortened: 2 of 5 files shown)


clason:
Open question: Do we want to keep adding "experimental" parsers? I think it would be good to have some sort of objective minimal standard to enforce for parsers (which this one very well might meet).


Also, note that "experimental" is no longer a status in Nvim-treesitter 1.0; we only have "unsupported" (which implies no auto-updates). To me, "experimental" would imply an even lower status (that allows failing tests), which I'm not sure we want to have?

ObserverOfTime:
The "experimental" status is currently rather arbitrary and I think it's fine to drop it for 1.0 and review the experimental parsers at that time.

But what would the minimal standard be? Supporting most language features and not breaking with every other change?

Our "worst" experimental parser I can think of is pug which not much worse than, say, bash in terms of features or c/cpp in terms of stability.

clason:
> The "experimental" status is currently rather arbitrary and I think it's fine to drop it for 1.0 and review the experimental parsers at that time.

I agree; "experimental" seems to me like a cop-out to avoid maintenance expectations -- it's precisely that I would like to push back on. Either it's good enough to be used -- then it's no longer experimental -- or it's not -- and then it needs some more cooking time _before_ being added to nvim-treesitter. 

For 1.0, I have just translated "experimental" to "tier 4", but I'm happy to change that (or would welcome a PR to `main` that bumps the relevant parsers' tier from 4 to 3).

In either case, the expectation for adding a parser to nvim-treesitter is that it will be actively supported. Tier 4 is meant as a temporary status to avoid removing useful work.


So TL;DR: If you think this parser is ready to be used (even though there's still some gaps and rough edges), just drop the "experimental" annotation.

clason (approved):

lua/nvim-treesitter/parsers.lua
```diff
@@ -1310,6 +1310,14 @@ list.rego = {
   maintainers = { "@FallenAngel97" },
 }
 
+list.requirements = {
+  install_info = {
+    url = "https://github.com/ObserverOfTime/tree-sitter-requirements",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@ObserverOfTime" },
```
clason, line 1318: Maybe add a `readme_note` explaining what this is for? ("requirements" is _very_ generic...) Does this need to register a filetype?

ObserverOfTime: > Does this need to register a filetype?

I've submitted a PR to Vim.

ObserverOfTime: I wouldn't mind leaving this open until that's merged.

ObserverOfTime: It's ready now.

PR #5147, opened 2023-07-25
feat: add some common comment tags

chrisgrieser:
`PERF` and `TEST` can be found as defaults for example in https://github.com/folke/todo-comments.nvim

all new tags are also common, for example as conventional commit keywords


amaanq (approved):

clason:
I personally don't like the comment parser at all, so no opinion either way.

PR #5148, opened 2023-07-25
fix(rust): move @namespace below @type

utkarshgupta137:
This is useful if one wishes to highlight enums (`@namespace`) differently from other types like structs. `@namespace` is more specific than `@type` in this case, so it should be defined last.
I've also changed the indentation to be more consistent.
This is my first tree-sitter contribution, so feedback is welcome!

### DIFF
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -78,47 +78,47 @@
 
 ; Assume that uppercase names in paths are types
 
-(scoped_identifier
-  path: (identifier) @namespace)
-
-(scoped_identifier
- (scoped_identifier
-  name: (identifier) @namespace))
-
-(scoped_type_identifier
-  path: (identifier) @namespace)
-
-(scoped_type_identifier
-  path: (identifier) @type
-  (#lua-match? @type "^[A-Z]"))
-
-(scoped_type_identifier
- (scoped_identifier
-  name: (identifier) @namespace))
-
 ((scoped_identifier
   path: (identifier) @type)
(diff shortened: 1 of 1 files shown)


utkarshgupta137:
cc: @amaanq

amaanq:
No, I have to disagree here. The captures for `@namespace` are above `@type` because `@type` is more specific in that it only matches CamelCase identifiers. And remember, Neovim queries are last wins, not first wins, so this is not right.

utkarshgupta137:
> And remember, Neovim queries are last wins, not first wins, so this is not right.

Yes. I personally want `@namespace` to win.

>  The captures for @namespace are above @type because @type is more specific in that it only matches CamelCase identifiers.

All structs/enums/traits/consts/statics etc. implement `@type`, but only enum implements `@namespace` - so I see it as more specific, regardless of the fact that `@namespace` can also apply to modules. Is there a better way for treesitter to differentiate enums from other types?

amaanq:
You can of course write your own queries to extend these, just add `; extends` at the top of the file

I don't think enums should be highlighted differently by default - they're a type after all

utkarshgupta137:
Can you point to an actual example? Are you suggesting that I can add custom tags like `@enum`? Will it be accepted into nvim-treesitter or are you suggesting a config option?

clason:
> Will it be accepted into nvim-treesitter

No, that's the whole point: you can make personal modifications that are not acceptable in general.

See `:h treesitter-query-modeline-extends`.

PR #5150, opened 2023-07-25
feat(t32): Add query support for extended HLL expression support

xasc:
Version 2.1 of the grammar has improved support for commands with HLL (high-level language, think C or C++) expressions. This change should update the existing queries accordingly.

Even though the syntax for HLL expressions is very close to C, it is, from my point of view, not really possible to add the support via language injection. The two main problems are that in contrast to C
- whitespaces or the lack thereof matters in weird ways, and that
- there are syntax extensions on top, like a slice operator for arrays (e.g. `Var.DRAW flags[0..16]`) or HLL variables that can be defined on the fly (e.g. `Var.set \x=func3(5,3)`).

That's why I tried to extract the relevant parts from tree-sitter-c, but keep most of the node definitions consistent.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1512,7 +1512,7 @@ list.t32 = {
   install_info = {
     url = "https://codeberg.org/xasc/tree-sitter-t32",
     files = { "src/parser.c", "src/scanner.c" },
-    revision = "1dd98248b01e4a3933c1b85b58bab0875e2ba437",
+    revision = "767f3c52fe649e4a6ab3551ac287e5b6038c9148",
   },
   maintainers = { "@xasc" },
 }
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1510,9 +1510,9 @@ list.systemtap = {
 
 list.t32 = {
   install_info = {
-    url = "https://codeberg.org/xasc/tree-sitter-t32",
+    url = "https://gitlab.com/xasc/tree-sitter-t32",
     files = { "src/parser.c", "src/scanner.c" },
-    revision = "767f3c52fe649e4a6ab3551ac287e5b6038c9148",
+    revision = "9d2520ae9886d3a768a352ec80db8762afb5232d",
   },
   maintainers = { "@xasc" },
 }
--- a/queries/t32/highlights.scm
+++ b/queries/t32/highlights.scm
@@ -1,3 +1,4 @@
+; Keywords, punctuation and operators
 [
   "="
   "^^"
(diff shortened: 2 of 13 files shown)


xasc:
Sorry, I hope that it's okay now.

queries/t32/highlights.scm
```diff
@@ -42,36 +55,125 @@
...
+  (identifier) @variable)
+
+(hll_unary_expression
+  (identifier) @variable)
+
+(hll_update_expression
+  (identifier) @variable)
+
+(hll_assignment_expression
+  left: (identifier) @variable)
```
lucario387, line 166: With these many `(identifier) @variable` I think it's okay to have a `(identifier) @variable` instead of all of these. Not against narrowing here though, this will allow for less extmarks

clason:
@xasc I think we want to drop the tests from our repo, as CodeBerg has proved to be too unreliable to have it in our CI (it's blocked all our pipelines several times in the last weeks, which is frankly unacceptable). We can still keep the parser and queries here as a compromise (if people hit install failures, maybe this will teach them not to blindly install everything...)

clason (changes requested):
Please rebase on master and drop the changes to the (now removed) tests.

Once upstream makes the intended changes to capture order and customization in `tree-sitter test`, you can just run those locally in your own CI.

queries/t32/highlights.scm
xasc: Amount of `(identifier) @variable` queries was reduced.

clason (approved):

xasc:
If something like a mirror repository would be acceptable, what would be the preference here
- bring back the tests or
- keep them out

amaanq:
a mirror on a reliable host would be great, I'd be okay to have the tests back then

clason:
If the mirror is more reliable (what did you have in mind?), tests are always good.

clason:
In particular, since Codeberg is a nice project, it would be great to have a general solution here so we can fully include parsers hosted there (which currently is not really feasible due to the CI -- in both directions, as our frequent CI also puts load on the hoster).

xasc:
As the list of parses mostly consists of GitHub and GitLab, I am going to pick one of these.
I will get back to you. Thank you very much.

amaanq:
sourcehut is fine too (pretty reliable w/ smali)

lucario387 (commented):
There seems to be some renamed nodes ... 

Shouldn't that mean the PR is a breaking change?

lua/nvim-treesitter/parsers.lua
```diff
@@ -1512,7 +1512,7 @@ list.t32 = {
   install_info = {
     url = "https://codeberg.org/xasc/tree-sitter-t32",
```
lucario387, line 1513: I can see that there exists a GitLab mirror for the language now. Very quick and clean work 😄 

I think it's fine to change the url to point directly to the GitLab mirror now

clason: and, more importantly, revert my last two commits

clason (approved):

lucario387 (approved):

xasc:
The installer is now pointing to a repository on GitLab and tests are re-enabled.

> There seems to be some renamed nodes ...
> 
> Shouldn't that mean the PR is a breaking change?

@lucario387 Yes, this is update is a breaking change. Is there anything that I missed?

clason:
We've just been trying to be better about this. Just mark the commit as breaking via `feat(t32)!: ...`.

xasc:
Ah alright, thank you for the hint.

lua/nvim-treesitter/parsers.lua
```diff
@@ -1510,9 +1510,9 @@ list.systemtap = {
 
 list.t32 = {
   install_info = {
-    url = "https://codeberg.org/xasc/tree-sitter-t32",
+    url = "https://gitlab.com/xasc/tree-sitter-t32",
     files = { "src/parser.c", "src/scanner.c" },
-    revision = "1dd98248b01e4a3933c1b85b58bab0875e2ba437",
+    revision = "9d2520ae9886d3a768a352ec80db8762afb5232d",
```
lucario387, line 1515: It's pretty obvious that nvim-treesitter will still pick up changes for your repo, even if you set the revision. So I believe it'd be better to remove this.

This can be re-added back from user-config level, and I think that's better

clason: No, this is how a parser is locked from automatic updates on `master`.

xasc: The latest of changes to the grammar was a bigger one. With these changes, the feature coverage of the grammar is quite complete, and I would expect things to slow down a bit now. The node names should also be quite stable.
I think I can continue without pinning the grammar to a certain version. At least I will try to be disciplined regarding changes that break things here.

clason: Then we can indeed drop the field from here on out (follow-up PR welcome!)

lucario387: > No, this is how a parser is locked from automatic updates on `master`.

I didn't check why it keeps showing in parser update list by the github-bot, looking into it and saw it was because of codeberg :smile:. Shouldn't be a problem anymore

So I take all the blame here

PR #5161, opened 2023-07-27
chore: update C, improve preproc_arg highlights

amaanq:


clason:
Could you try to avoid unrelated reformatting? Makes it harder to evaluate the diff.

amaanq:
for sure, updated

queries/c/highlights.scm
```diff
@@ -157,13 +158,17 @@
...
+  (#lua-match? @constant "^[A-Z][A-Z0-9_]+$"))
 (enumerator
   name: (identifier) @constant)
 (case_statement
   value: (identifier) @constant)
 
 ((identifier) @constant.builtin
     (#any-of? @constant.builtin "stderr" "stdin" "stdout"))
+(preproc_def (preproc_arg) @constant.builtin
+  (#any-of? @constant.builtin "stderr" "stdin" "stdout"))
```
clason, line 171: Would it make sense to keep these together?

(I guess this is a general question about structuring queries -- where not constrained by precedence -- by node or by capture.)

clason (approved):

queries/c/highlights.scm
amaanq: not sure I follow - are you asking does it make sense to also check for these in a preproc_arg? I'd like to say yes if so

clason: No, I just mean there's three different stanzas for the same node with different captures, in three different places (where the same capture is applied to different nodes).

amaanq: ah, yeah we could definitely keep these together for this one

clason: Just a thought; not a deal-breaker either way.

PR #5172, opened 2023-07-29
SQL Hightlighs new Keywords

matthias-Q:
Adds queries for SQL keywords introduced in https://github.com/DerekStride/tree-sitter-sql/pull/186


clason:
Default branch is still `master`; please open feature branches from that.

PR #5185, opened 2023-08-02
update ownership for proto parser

treywood:
it appears that the proto parser is unmaintained (no activity since 2021, [my PR](https://github.com/mitchellh/tree-sitter-proto/pull/6) unacknowledged for 2 weeks)  so am looking to take over ownership.

let me know if there's a more formal process for parser config changes like this


queries/proto/highlights.scm
```diff
@@ -1,15 +1,18 @@
...
+  "option"
   "optional"
   "repeated"
   "reserved"
+  "returns"
+  "required"
+  "rpc"
+  "service"
+  "syntax"
   "to"
```
treywood, line 16: sorting and adding a few new ones for proto2 support

nerdrew:
I can't approve the PR, but I'm excited about these improvements!

queries/proto/highlights.scm
```diff
@@ -1,15 +1,18 @@
...
   "enum"
+  "extend"
+  "extensions"
+  "message"
   "oneof"
+  "option"
   "optional"
   "repeated"
   "reserved"
+  "returns"
```
amaanq, line 11: `@keyword.return`?

queries/proto/highlights.scm
```diff
@@ -1,15 +1,18 @@
...
+  "required"
+  "rpc"
```
amaanq, line 13: related to returns, maybe `@keyword.function`?

queries/proto/highlights.scm
```diff
@@ -1,15 +1,18 @@
...
+  "extend"
+  "extensions"
+  "message"
   "oneof"
+  "option"
   "optional"
   "repeated"
   "reserved"
+  "returns"
+  "required"
```
amaanq, line 12: `@type.qualifier`? same for optional

amaanq (changes requested):

amaanq:
Also believe you need to add "proto2"

queries/proto/highlights.scm
treywood: this is not `return` as in a function return statement. it's a keyword for defining the return message type of a service RPC. for example:

```protobuf
service SearchService {
  rpc Search(SearchRequest) returns (SearchResponse);
}
```

amaanq: yeah I'm aware but imo it's similar enough to be warranted, and would allow a configurable yet clear distinction for users

queries/proto/highlights.scm
treywood: sure, that makes sense. updated

queries/proto/highlights.scm
treywood: alrighty I can get on board with that. updated.

amaanq (approved):

PR #5189, opened 2023-08-03
feat(perl): improve highlights

ObserverOfTime:
Mainly heredocs and builtin functions.

### DIFF
--- a/queries/perl/highlights.scm
+++ b/queries/perl/highlights.scm
@@ -30,15 +30,6 @@
 ; Keywords to define a function
 [ "sub" ] @keyword.function
 
-; Lots of builtin functions, except tree-sitter-perl doesn't emit most of
-;   these yet
-;[
-;  "print" "printf" "sprintf" "say"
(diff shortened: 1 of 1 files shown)


amaanq (approved):

ObserverOfTime:
Halfway through this, I noticed that there's [another perl parser](https://github.com/tree-sitter-perl/tree-sitter-perl) which may or may not be better. :thinking:

amaanq:
Interesting, it seems actively maintained and its scanner is in C

ObserverOfTime:
I'm thinking we should keep using this parser until tree-sitter-perl/tree-sitter-perl#85 is merged.
That way we won't have to deal with a potentially breaking change soon after switching.

amaanq:
Would be interesting to see if they're willing to upstream that parser

clason:
Yes, @leonerd is actively working on an official(!) Perl parser. It's not easy going.

PR #3296, opened 2022-08-13
highlights(xml): add highlights

theHamsta:
Alternative to https://github.com/nvim-treesitter/nvim-treesitter/pull/3297

https://github.com/nvim-treesitter/nvim-treesitter/issues/3295

PR #5194, opened 2023-08-04
feat: add xml parser

### LINKED ISSUE #3295: Adding support for generic XML files
Basically, as the title suggests, there is support for xml like languages - html for example. But nothing about supporting the general xml syntax. I would like to be able to use text objects such as attributes for example or even tags if possible, would be great.

A parser working on files with plain xml extensions

Forcing treesitter to use the html parser when working with generic xml files

**Additional context**
It would be great to also support things like tag text objects (not sure if there are markers for that) to be able to do something like [t or ]t to move to the next tag. By default, i imagine ]t can jump to the next child tag, if you end up with no child tags remaining, jump to the next parent tag (the next sibling of the parent of the current child tag) one level up (if one exists).

ObserverOfTime:
The parser is a messy fork of a messy fork of a messy source.
I'd mark it as experimental but @clason will complain. :^)

Closes #3297, closes #3296, fixes #3295.


ObserverOfTime:
Most of the highlights queries are for the DTD.
I wanna separate that into its own grammar at some point.

ObserverOfTime:
Honestly, I might as well rewrite it.

ObserverOfTime:
My rewrite has 2 issues at the moment:

1. Sometimes it freezes Neovim when typing. (very bad)
2. `"?>"` is not highlighted *only* in `(PI)` and *only* in Neovim. It's highlighted in all other nodes in Neovim and in all nodes, including `(PI)`, in the playground. (very weird)

I could use some help with solving those ~~and with writing a scanner for some [minor stuff](https://github.com/search?q=repo%3AObserverOfTime%2Ftree-sitter-xml%20FIXME&type=code)~~.

amaanq:
> My rewrite has 2 issues at the moment:
> 
> 1. It freezes Neovim when closing a start tag. (very bad)
> 2. `"?>"` is not highlighted _only_ in `(PI)` and _only_ in Neovim. It's highlighted in all other nodes in Neovim and in all nodes, including `(PI)`, in the playground. (very weird)
> 
> I could use some help with solving those ~and with writing a scanner for some [minor stuff](https://github.com/search?q=repo%3AObserverOfTime%2Ftree-sitter-xml%20FIXME&type=code)~.

I have some thoughts/explanations I'd like to share that you (or anyone reading this) might find helpful in how I debugged and fixed the infinite loop bug at least.

Generally, this comes from a regex that could match EOF (null char) or a regex that can be empty. The former has been fixed on master, the latter not yet..(maybe it should be ;) )

So, instead of trying every rule out one by one, I decided to just run a fuzzer and get a sample file that could crash. The fuzzer can be found [here](https://gist.github.com/amaanq/5a9bb5de2e19c4bfa2e9d80099518414) and can be run like so:
`./fuzz.sh xml 1 10 _`

The last arg is usually either `c` or `cpp` for a scanner file, but since there's none, I just set it to `_`

Now, immediately a test input fails:

<details>
<summary>Screenshot</summary>

image

</details>


Now we just parse that file using debug mode, `ts p fuzzer/timeout-0622d385e2781ed2ee15f904d4761c02539af8f5 -d`, and we can see what the issue is, we always skip an empty token "CharData" and so the lexer never advances past that:

<details>
<summary>Screenshot</summary>

image

</details>

So, the issue was CharData, and that was fixed by just ensuring at least one token matches, but that went into the external scanner afterwards anyways

ObserverOfTime:
All the issues should (hopefully) be resolved now, thanks to @amaanq's help.

clason:
Question: do we want to pull the plug on the old-time injections? Or switch new parsers over to the `main` branch, to be backported at a later date?

(Constantly rebasing new queries is getting old, fast...)

ObserverOfTime:
I'm fine with the latter.

clason:
> I just won't be able to use the new parsers until refactor and textobjects work with `main`. :^)

That would put some incentive on someone to start rewriting these plugins ;)

ObserverOfTime:
Honestly, we should probably move textobjects here.

clason:
> Honestly, we should probably move textobjects here.

No, that goes against the general goal of _removing_ functionality from this plugin to focus on parser and query management. (Indent is an exception since it's a self-contained module that doesn't require configuration and will be upstreamed anyway.)

clason:
New format please ;)

ObserverOfTime:
#### Old format

```scheme
((element
   (STag (Name) @_name)
   (content) @css @combined)
 (#eq? @_name "style"))
```

#### New format

```scheme
((element
   (STag (Name) @_name)
   (content) @injection.content)
 (#eq? @_name "style")
 (#set! injection.combined)
 (#set! injection.language "css"))
```

Good stuff. :^)

amaanq (approved):

PR #5209, opened 2023-08-08
highlights(ocaml): add all builtin functions

Akmadan23:
I noticed that there were only 4 builtin functions listed, so I added all the others.


amaanq:
@undu

ObserverOfTime:
FWIW the current builtins seem more like `@exception` to me.

XVilka:
I also should note that OCaml changes standard librar{y,ies} (compiler Stdlib, but also Base/Core as well) somewhat often, mostly by adding new functions though, but sometimes also deprecating.

Akmadan23:
Ok, got it. I just thought that since they are in scope by default they could be considered builtin, but I totally see the reason why they are not. Sorry for the inconvenience.

clason:
No worries. There could still be _some_ missing from the current list (if things change as often as they seem to do), so a PR adding only those might still be useful.

PR #5220, opened 2023-08-10
fix: update the rest of C's children

amaanq:


clason (approved):

clason:
typo in cuda commit message ;)

Although I think those can be squashed anyway.

amaanq:
yeah I wasn't sure whether to segment them or do something like`fix(arduino, cuda, glsl, hlsl): ...`

clason:
Yup, things go fast if you only change queries and don't have to compile 200 parsers...

PR #5221, opened 2023-08-10
feat(c): update highlights

amaanq:
This includes some changes I've been meaning to add for a while but always forgot about

### DIFF
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -17,6 +17,7 @@
   "sizeof"
   "offsetof"
 ] @keyword.operator
+(alignof_expression . _ @keyword.operator)
 
 "return" @keyword.return
 
@@ -142,8 +143,9 @@
 (storage_class_specifier) @storageclass
 
 [
-  (type_qualifier) 
+  (type_qualifier)
   (gnu_asm_qualifier)
+  "__extension__"
 ] @type.qualifier
 
 (linkage_specification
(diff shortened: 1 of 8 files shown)


amaanq:
Need to update objc 😅

clason:
Here we go again...

queries/c/highlights.scm
```diff
@@ -170,6 +172,28 @@
 (preproc_def (preproc_arg) @constant.builtin
   (#any-of? @constant.builtin "stderr" "stdin" "stdout"))
 
+((identifier) @variable.builtin
+  (#any-of? @variable.builtin
```
ObserverOfTime, line 176: These are technically constants. :)

```suggestion
((identifier) @constant.builtin
  (#any-of? @constant.builtin
```

theHamsta:
HLSL, cuda, glsl have updated to newest c/c++ changes

queries/c/highlights.scm
amaanq: it's really predefined macros, but, yeah constant works (`@macro.builtin` would be nice but it's too nice imo)

queries/c/highlights.scm
```diff
@@ -170,6 +172,28 @@
...
+    "__cplusplus" "__OBJC__" "__ASSEMBLER__"
+    "__BASE_FILE__" "__FILE_NAME__" "__INCLUDE_LEVEL__"
+    "__TIMESTAMP__" "__clang__" "__clang_major__"
+    "__clang_minor__" "__clang_patchlevel__"
+    "__clang_version__" "__clang_literal_encoding__"
+    "__clang_wide_literal_encoding__"
+    "__FUNCTION__" "__func__" "__PRETTY_FUNCTION__"
+    "__VA_ARGS__" "__VA_OPT__"))
+((identifier) @variable.builtin
+  (#has-ancestor? @variable.builtin attribute_specifier))
```
theHamsta, line 188: why not 
```
(attribute_specifier
  (identifier))
```
we should avoid `#has-ancestor?` being invoked on any `(identifier)`. `#has-ancestor?` will call `parent` until it reaches the top query level. Would be ideal if we could remove this predicate.

amaanq: Reaching `__word__` in here is hard:

`typedef int register_t __attribute__((__mode__(__word__)));`

 And gets harder w/ the more nesting. But yeah, I can agree has-ancestor might not be a great idea

theHamsta: Realistically, most people don't use compiler intrinsics but the predicate would penalize every single `identifier`.

It would probably be better when the query language would allow such nesting. It could be implemented by letting the outer query enable some state flag when it is matched that could enable matching the inner query. Maybe this could enable implementing nesting during a normal query pass in the Tree-Sitter C runtime. I know the query code is not a pleasure to work with.

amaanq: I was also thinking it'd be better if the query syntax allowed for this as well :)

amaanq:
Also gonna fix this case when I removed just injecting all preproc args:

```c
#define BAR                                                                    \
    do {                                                                       \
        if (lexer->lookahead == B) {                                           \
            advance(lexer);                                                    \
            return true;                                                       \
        }                                                                      \
    } while (0)
```

A good check to reduce most useless cases (like constants/flags/strings) is to check for \n being present (aka it spans multiple lines)

amaanq:
@fab4100 ispc needs to be updated (check latest PR)

also - for those with a C-family grammar, would you mind inviting me to the repo so I can keep them in sync more easily? :) no pressure if you wouldn't want to

ObserverOfTime:
I updated arduino and invited you.

ObserverOfTime (approved):

theHamsta:
> @fab4100 ispc needs to be updated (check latest PR)
> 
> also - for those with a C-family grammar, would you mind inviting me to the repo so I can keep them in sync more easily? :) no pressure if you wouldn't want to

I don't see why that's necessary. I have a cron GH action that updates deps of those three repos (tree-sitter-c, tree-sitter-cpp, tree-sitter CLI). When any of those update I get a PR. When tests pass I update, if tests fail I look up what changed in tree-sitter-c/tree-sitter-cpp and update my grammars accordingly

ObserverOfTime:
I also have a GH action that can run on schedule or manually but I don't mind.

fab4100:
I have merged the PR and added you as well.  Note that Intel expressed interest in hosting the parser repo on their ISPC github organization (currently they are doing paperwork on their end which seems slow).  I am updating w/r/t to this.  The parser for the ISPC compiler will likely not change anymore in the future, the `tree-sitter-ispc` parser could be made standalone if that is desired (I personally prefer the `tree-sitter-c` inheritance however).  GH action sounds good, I will add that feature.

clason:
Parsers are updated; rebase on `master` and you should be good to go.

amaanq:
> I don't see why that's necessary. I have a cron GH action that updates deps of those three repos (tree-sitter-c, tree-sitter-cpp, tree-sitter CLI). When any of those update I get a PR. When tests pass I update, if tests fail I look up what changed in tree-sitter-c/tree-sitter-cpp and update my grammars accordingly.
> 
> Or is this about doing the change in less than an hour since I would update only within the schedule of the cron job?

Right, maybe the others should employ the cron job too 😁 , I don't need it updated within an hour anyways

PR #5222, opened 2023-08-10
feat(perl)!: replace parser

rabbiveesh:
This replaces the perl parser from ganezdragon with an actively maintained + considerably
more correct version. There are some features that we're still working on, but by and
large this has already proven very useful for me in my day to day.

One of the things that we support is coloring scalars, arrays and hashes differently. Is
there a way to expose that in a simple way to users? We're doing it by making a
`@variable.scalar`, `@variable.array` and `@variable.hash`.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -414,7 +414,7 @@
     "revision": "e01767921df18142055d97407595329d7629e643"
   },
   "perl": {
-    "revision": "79e88f64681660f3961939bf764d8f3b4bbb0d27"
+    "revision": "99bbcb0faddfc79db3fd2996cb56689a4685e78c"
   },
   "php": {
     "revision": "92a98adaa534957b9a70b03e9acb9ccf9345033a"
--- a/lockfile.json
+++ b/lockfile.json
@@ -428,6 +428,9 @@
   "po": {
     "revision": "d6aed225290bc71a15ab6f06305cb11419360c56"
   },
+  "pod": {
+    "revision": "ea5d557cbd185cdcb5efcfdb6bc846fe909d86ae"
+  },
   "poe_filter": {
     "revision": "d7b43b51f92fb19efe8af45c2246087c611c8f63"
   },
--- a/queries/perl/highlights.scm
+++ b/queries/perl/highlights.scm
@@ -1,198 +1,146 @@
-; Misc keywords
-[
-  "my" "our" "local"
-  "next" "last" "redo"
-  "goto"
-  "package"
-;  "do"
-;  "eval"
-] @keyword
-
-; Keywords for including
-[ "use" "no" "require" ] @include
-
-; Keywords that mark conditional statements
-[ "if" "elsif" "unless" "else" ] @conditional
-(ternary_expression
-  ["?" ":"] @conditional.ternary)
-
-; Keywords that mark repeating loops
-[ "while" "until" "for" "foreach" ] @repeat
-
-; Keyword for return expressions
-[ "return" ] @keyword.return
-
-; Keywords for phaser blocks
-; TODO: Ideally these would be @keyword.phaser but vim-treesitter doesn't
-;   have such a thing yet
-[ "BEGIN" "CHECK" "UNITCHECK" "INIT" "END" ] @keyword.function
-
-; Keywords to define a function
-[ "sub" ] @keyword.function
-
-; Keywords that are regular infix operators
-[
-  "and" "or" "not" "xor"
-  "eq" "ne" "lt" "le" "ge" "gt" "cmp"
-] @keyword.operator
-
-; Variables
-[
-  (scalar_variable)
-  (array_variable)
-  (hash_variable)
-] @variable
-
-; Special builtin variables
-[
-  (special_scalar_variable)
-  (special_array_variable)
-  (special_hash_variable)
-  (special_literal)
-  (super)
-] @variable.builtin
-
-((scalar_variable) @variable.builtin
- (#eq? @variable.builtin "$#ARGV"))
-
-; Integer numbers
-[
-  (integer)
-  (hexadecimal)
-] @number
-
-; Float numbers
-[
-  (floating_point)
-  (scientific_notation)
-] @float
-
-; version sortof counts as a kind of multipart integer
-(version) @constant
-
-; Package names are types
-(package_name) @type
-
-; The special SUPER:: could be called a namespace. It isn't really but it
-;   should highlight differently and we might as well do it this way
-(super) @namespace
-
-; Comments are comments
-(comments) @comment @spell
-
-((source_file . (comments) @preproc)
+((source_file . (comment) @preproc)
   (#lua-match? @preproc "^#!/"))
 
-; POD should be handled specially with its own embedded subtype but for now
-;   we'll just have to do this.
-(pod_statement) @text
+[ "use" "no" "require" ] @include
 
-(method_invocation
-  function_name: (identifier) @method.call)
-(call_expression
-  function_name: (identifier) @function.call)
+[ "if" "elsif" "unless" "else" ] @conditional
 
-; Built-in functions
-((call_expression
-  function_name: (identifier) @function.builtin)
- (#any-of? @function.builtin
-   "print" "printf" "sprintf" "say"
-   "push" "pop" "shift" "unshift" "splice"
-   "exists" "delete" "keys" "values"
-   "each"))
+(conditional_expression [ "?" ":" ] @conditional.ternary) 
+
+[ "while" "until" "for" "foreach" ] @repeat
+
+"return" @keyword.return
+
+"sub" @keyword.function
+
+[ "map" "grep" ] @function.builtin
+
+"package" @include
 
 [
-  (map)
-  (grep)
-  (bless)
-] @function.builtin
+  "do"
+  "my" "our" "local"
+  "last" "next" "redo" "goto"
+  "undef"
+] @keyword
 
-;; ----------
+(_ operator: _ @operator)
+"\\" @operator
 
-(use_constant_statement
-  constant: (identifier) @constant)
+(yadayada) @exception
 
-(named_block_statement
-  function_name: (identifier) @function)
-
-(function_definition
-  name: (identifier) @function)
-
-(function) @function
+(phaser_statement phase: _ @keyword.phaser)
 
 [
-  "(" ")"
-  "[" "]"
-  "{" "}"
-  (standard_input_to_variable)
-] @punctuation.bracket
+  "or" "and"
+  "eq" "ne" "cmp" "lt" "le" "ge" "gt"
+  "isa"
+] @keyword.operator
 
-[ "`" "\"" ] @punctuation.special
+(eof_marker) @preproc
+(data_section) @comment
 
-[
-  "=~"
-  "!~"
-  "="
-  "=="
-  "+"
-  "-"
-  "."
-  "//"
-  "||"
-  "&&"
-  "<<"
-  (arrow_operator)
-  (hash_arrow_operator)
-  (array_dereference)
-  (hash_dereference)
-  (to_reference)
-  (type_glob)
-  (hash_access_variable)
-] @operator
+(pod) @text
 
 [
-  (regex_option)
-  (regex_option_for_substitution)
-  (regex_option_for_transliteration)
-] @parameter
-
-(type_glob
-  (identifier) @variable)
+  (number)
+  (version)
+] @number
 
 [
-  (word_list_qw)
-  (command_qx_quoted)
-  (string_single_quoted)
-  (string_double_quoted)
-  (string_qq_quoted)
-  (bareword)
-  (transliteration_tr_or_y)
+  (string_literal) 
+  (interpolated_string_literal) 
+  (quoted_word_list) 
+  (command_string) 
+  (heredoc_content)
+  (replacement)
+  (transliteration_content)
 ] @string
 
 [
- (heredoc_start_identifier)
- (heredoc_end_identifier)
+  (heredoc_token)
+  (command_heredoc_token)
+  (heredoc_end)
 ] @label
 
-(heredoc_body_statement) @text.literal
+[(escape_sequence) (escaped_delimiter)] @string.escape
 
-[
-  (pattern_matcher)
-  (regex_pattern_qr)
-  (patter_matcher_m)
-  (substitution_pattern_s)
+(_ modifiers: _ @character.special)
+[  
+ (quoted_regexp)
+ (match_regexp)
+ (regexp_content)
 ] @string.regex
 
-(escape_sequence) @string.escape
+(autoquoted_bareword _?) @string.special
 
-[
-  ","
-  (semi_colon)
-  (start_delimiter)
-  (end_delimiter)
-  (ellipsis_statement)
-] @punctuation.delimiter
+(use_statement (package) @type)
+(package_statement (package) @type)
+(require_expression (bareword) @type)
 
-(function_attribute) @field
+(subroutine_declaration_statement name: (bareword) @function)
+(attribute_name) @attribute
+(attribute_value) @string
 
-(function_signature) @type
+(label) @label
+
+(statement_label label: _ @label)
+
+(relational_expression operator: "isa" right: (bareword) @type)
+
+(function_call_expression (function) @function.call)
+(method_call_expression (method) @method.call)
+(method_call_expression invocant: (bareword) @type)
+
+(func0op_call_expression function: _ @function.builtin)
+(func1op_call_expression function: _ @function.builtin)
+
+([(function)(expression_statement (bareword))] @function.builtin
+ (#set! "priority" 101)
+ (#lua-match? @function.builtin
+   "^(accept|atan2|bind|binmode|bless|crypt|chmod|chown|connect|die|dbmopen|exec|fcntl|flock|getpriority|getprotobynumber|gethostbyaddr|getnetbyaddr|getservbyname|getservbyport|getsockopt|glob|index|ioctl|join|kill|link|listen|mkdir|msgctl|msgget|msgrcv|msgsend|opendir|print|printf|push|pack|pipe|return|rename|rindex|read|recv|reverse|say|select|seek|semctl|semget|semop|send|setpgrp|setpriority|seekdir|setsockopt|shmctl|shmread|shmwrite|shutdown|socket|socketpair|split|sprintf|splice|substr|system|symlink|syscall|sysopen|sysseek|sysread|syswrite|tie|truncate|unlink|unpack|utime|unshift|vec|warn|waitpid|formline|open|sort)$"
+))
+
+(function) @function
+
+(ERROR) @error
+
+(_
+  "{" @punctuation.special
+  (varname)
+  "}" @punctuation.special)
+
+(varname 
+  (block
+    "{" @punctuation.special 
+    "}" @punctuation.special))
+
+
+(
+  [(varname) (filehandle)] @variable.builtin
+  (#lua-match? @variable.builtin "^((ENV|ARGV|INC|ARGVOUT|SIG|STDIN|STDOUT|STDERR)|[_ab]|\\W|\\d+|\\^.*)$")
+)
+
+(scalar) @variable.scalar
+(scalar_deref_expression [ "$" "*"] @variable.scalar)
+[(array) (arraylen)] @variable.array
+(array_deref_expression [ "@" "*"] @variable.array)
+(hash) @variable.hash
+(hash_deref_expression [ "%" "*"] @variable.hash)
+
+(array_element_expression array:(_) @variable.array)
+(slice_expression array:(_) @variable.array)
+(keyval_expression array:(_) @variable.array)
+
+(hash_element_expression hash:(_) @variable.hash)
+(slice_expression hash:(_) @variable.hash)
+(keyval_expression hash:(_) @variable.hash)
+
+(comment) @comment
+
+([ "=>" "," ";" "->" ] @punctuation.delimiter)
+
+(
+  [ "[" "]" "{" "}" "(" ")" ] @punctuation.bracket
+  ; priority hack so nvim + ts-cli behave the same
+  (#set! "priority" 90))
--- a/queries/perl/injections.scm
+++ b/queries/perl/injections.scm
@@ -1,2 +1,14 @@
-((comments) @injection.content
+; an injections.scm file for nvim-treesitter
+((comment) @injection.content
  (#set! injection.language "comment"))
+ 
+((pod) @injection.content
+ (#set! injection.language "pod"))
+
+((substitution_regexp
+  (replacement) @injection.content
+  (substitution_regexp_modifiers) @_modifiers)
+    ; match if there's a single `e` in the modifiers list
+  (#lua-match? @_modifiers "e")
+  (#not-lua-match? @_modifiers "e.*e")
+  (#set! injection.language "perl"))
--- a/queries/pod/highlights.scm
+++ b/queries/pod/highlights.scm
@@ -1,0 +1,72 @@
+; A highlight file for nvim-treesitter to use
+
+[(pod_command)
+ (command)
+ (cut_command)] @keyword
+
+(command_paragraph
+  (command) @keyword
+  (#lua-match? @keyword "^=head")
+  (content) @text.title)
+
+(command_paragraph
+  (command) @keyword
+  (#lua-match? @keyword "^=over")
+  (content) @number)
+
+(command_paragraph
+  (command) @keyword
+  (#lua-match? @keyword "^=item")
+  (content) @text)
+
+(command_paragraph
+  (command) @keyword
+  (#lua-match? @keyword "^=encoding")
+  (content) @string.special)
+
+(command_paragraph
+  (command) @keyword
+  (#not-lua-match? @keyword "^=(head|over|item|encoding)")
+  (content) @string)
+
+(verbatim_paragraph (content) @text.literal)
+
+(interior_sequence
+  (sequence_letter) @character
+  ["<" ">"] @punctuation.delimiter
+)
+
+(interior_sequence
+  (sequence_letter) @character
+  (#eq? @character "B")
+  (content) @text.strong)
+
+(interior_sequence
+  (sequence_letter) @character
+  (#eq? @character "C")
+  (content) @text.literal)
+
+(interior_sequence
+  (sequence_letter) @character
+  (#eq? @character "F")
+  (content) @text.underline @string.special)
+
+(interior_sequence
+  (sequence_letter) @character
+  (#eq? @character "I")
+  (content) @text.emphasis)
+
+(interior_sequence
+  (sequence_letter) @character
+  (#eq? @character "L")
+  (content) @text.uri)
+
+(interior_sequence
+  (sequence_letter) @character
+  (#eq? @character "X")
+  (content) @text.reference)
+
+(interior_sequence
+  (sequence_letter) @character
+  (#eq? @character "E")
+  (content) @string.escape)
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1280,6 +1280,15 @@ list.po = {
   maintainers = { "@amaanq" },
 }
 
+list.pod = {
+  install_info = {
+    url = "https://github.com/tree-sitter-perl/tree-sitter-pod",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "release",
+  },
+  maintainers = { "@RabbiVeesh", "@LeoNerd" },
+}
+
 list.poe_filter = {
   install_info = {
     url = "https://github.com/ObserverOfTime/tree-sitter-poe-filter",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1239,10 +1239,11 @@ list.pem = {
 
 list.perl = {
   install_info = {
-    url = "https://github.com/ganezdragon/tree-sitter-perl",
-    files = { "src/parser.c", "src/scanner.cc" },
+    url = "https://github.com/tree-sitter-perl/tree-sitter-perl",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "release",
   },
-  maintainers = { "@lcrownover" },
+  maintainers = { "@RabbiVeesh", "@LeoNerd" },
 }
 
 list.php = {
--- a/queries/perl/folds.scm
+++ b/queries/perl/folds.scm
@@ -1,11 +1,19 @@
-[
-  (function_definition)
-  (if_statement)
-  (unless_statement)
-  (while_statement)
-  (until_statement)
-  (for_statement_1)
-  (for_statement_2)
-  (standalone_block)
-  (pod_statement)
-] @fold
+(comment)+ @fold
+(pod) @fold
+
+; fold the block-typed package statements only
+(package_statement (block)) @fold
+
+[(subroutine_declaration_statement)
+ (conditional_statement)
+ (loop_statement)
+ (for_statement)
+ (cstyle_for_statement)
+ (block_statement)
+ (phaser_statement)] @fold
+
+(anonymous_subroutine_expression) @fold
+
+; perhaps folks want to fold these too?
+[(anonymous_array_expression)
+ (anonymous_hash_expression)] @fold


clason:
Is this the one @leonerd has been working on?

queries/perl/injections.scm
```diff
@@ -1 +1,2 @@
-(comments) @comment
```
clason, line 1: We globally have that comment injection; any reason to remove it?

leonerd:
> Is this the one @leonerd has been working on?

It is, yes

queries/perl/injections.scm
leonerd: I'm not sure this is particularly a concsious decision to remove it, so much as just that our copy started off empty, and we added `(pod)`, and didn't add this one that you have.

That said, I don't think we have a `(comments)` definition, do we? Just a single-line `(comment)`.

clason:
So, first order of business:

1. we use a different query precedence than upstream (for now; they'll change eventually): latest match wins
2. we use different capture names, see https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights -- any other (toplevel) captures are illegal and trip CI.

queries/perl/injections.scm
clason: Well, whatever comment node you have should inject the `comment` parser.

queries/perl/highlights.scm
```diff
@@ -1,198 +1,89 @@
...
-
-; Keywords that are regular infix operators
 [
-  "and" "or" "not" "xor"
-  "eq" "ne" "lt" "le" "ge" "gt" "cmp"
-] @keyword.operator
+  "or" "and"
+  "eq" "ne" "cmp" "lt" "le" "ge" "gt"
+  "isa"
+] @operator
```
ObserverOfTime, line 23:
```suggestion
] @keyword.operator
```

queries/perl/highlights.scm
```diff
@@ -1,198 +1,89 @@
...
-  (floating_point)
-  (scientific_notation)
-] @float
+(string_literal) @string
+(interpolated_string_literal) @string
+(quoted_word_list) @string
+(command_string) @string
+[(heredoc_token) (command_heredoc_token)] @string.special
+(heredoc_content) @string
+(heredoc_end) @string.special
```
ObserverOfTime, line 41: We generally use `@label` for heredoc tokens.
```suggestion
[(heredoc_token) (command_heredoc_token)] @label
(heredoc_content) @string
(heredoc_end) @label
```

queries/perl/highlights.scm
```diff
@@ -1,198 +1,89 @@
...
+[(escape_sequence) (escaped_delimiter)] @string.special
 
-; version sortof counts as a kind of multipart integer
-(version) @constant
+[(quoted_regexp) (match_regexp)] @string.regexp
```
ObserverOfTime, line 44:
```suggestion
[(quoted_regexp) (match_regexp)] @string.regex
```

queries/perl/highlights.scm
```diff
@@ -1,198 +1,89 @@
...
-; Built-in functions
-((call_expression
-  function_name: (identifier) @function.builtin)
- (#any-of? @function.builtin
-   "print" "printf" "sprintf" "say"
-   "push" "pop" "shift" "unshift" "splice"
-   "exists" "delete" "keys" "values"
-   "each"))
+(subroutine_declaration_statement name: (_) @function)
+(attribute_name) @decorator
```
ObserverOfTime, line 70:
```suggestion
(attribute_name) @attribute
```

leonerd:
> * we use a different query precedence than upstream (for now; they'll change eventually): latest match wins

Ah as a total aside, I was curious about that for my own implementation. I think this seems to be the most common interpretation. Good.

But I'm unclear why you bring it up specifically now - I don't think we rely on that anywhere, do we? I've mostly been using/testing this via nvim anyway, so I believe nvim's interpretation is correct.

> * we use different capture names, see https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights -- any other (toplevel) captures are illegal and trip CI.

Ahyes; probably some adjusting needs doing there then.

queries/perl/highlights.scm
```diff
@@ -1,198 +1,89 @@
-; Misc keywords
 [
```
ObserverOfTime, line 1: Keep the previous captures for these:

```scheme
[ "use" "no" "require" ] @include

[ "if" "elsif" "unless" "else" ] @conditional

[ "while" "until" "for" "foreach" ] @repeat

"return" @keyword.return

"sub" @keyword.function
```

Also, "map" and "grep" could be `@function.builtin`.

queries/perl/highlights.scm
```diff
@@ -1,198 +1,89 @@
...
+(statement_label label: _ @label)
 
-(use_constant_statement
-  constant: (identifier) @constant)
+(relational_expression operator: "isa" right: (bareword) @type)
 
-(named_block_statement
-  function_name: (identifier) @function)
+(function_call_expression (function) @function)
+(method_call_expression (method) @function.method)
```
ObserverOfTime, line 81:
```suggestion
(function_call_expression (function) @function.call)
(method_call_expression (method) @method.call)
```

queries/perl/injections.scm
```diff
@@ -1 +1,2 @@
...
+; an injections.scm file for nvim-treesitter
+(pod) @pod
```
ObserverOfTime, line 2: Can you also add the pod parser (or comment this out till then)?

ObserverOfTime (changes requested):
Missing highlights for operators and punctuation.

Also missing this for shebangs:
```scheme
((source_file . (comment) @preproc)
  (#lua-match? @preproc "^#!/"))
```

clason:
> Ah as a total aside, I was curious about that for my own implementation. I think this seems to be the most common interpretation. Good.

It's the interpretation upstream will switch to, as far as I know.

> But I'm unclear why you bring it up specifically now - I don't think we rely on that anywhere, do we? I've mostly been using/testing this via nvim anyway, so I believe nvim's interpretation is correct.

It's just a very common gotcha when queries are downstreamed, so I thought I'd mention it just to potentially save some time.

queries/perl/injections.scm
leonerd: Oh; the POD parser is at https://github.com/tree-sitter-perl/tree-sitter-pod/

rabbiveesh: I'll do this today, I added a release branch there too just now

queries/perl/highlights.scm
```diff
@@ -1,198 +1,89 @@
...
+(method_call_expression invocant: (bareword) @type)
 
-(function_definition
-  name: (identifier) @function)
+(func0op_call_expression function: _ @function.builtin)
+(func1op_call_expression function: _ @function.builtin)
```
lucario387, line 85: Was there any specific reason for this weird syntax node? Or is it having separated nodes for these leads to more efficient parsing?

ObserverOfTime: I assume that's related to [this](https://www.perlmonks.org/?node_id=663393) quirk of Perl.

> ```perl
> whatever  / 25 ; # / ; die "this dies!";
> ```
>
> Schwartz's Snippet can parse two different ways: if `whatever` is nullary (that is, takes no arguments), the first statement is a division in void context, and the rest of the line is a comment. If `whatever` takes an argument, Schwartz's Snippet parses as a call to the `whatever` function with the result of a match operator, then a call to the `die()` function.

leonerd: > Was there any specific reason for this weird syntax node? Or is it having separated nodes for these leads to more efficient parsing?

Short answer: Because perl grammar is crazy.

Longer answer: Perl's parser (actually the lexer - yes really the lexer) categorizes the names of known functions into three separate sets - nullary functions, unary functions, and anything else. The parsing rules for these different categories take account of zero, at-most one or any number of arguments. Because of this, parsing of functions known to be in various categories has different precedence even without parens:

```perl
join $x, $y, $z   # is a regular list function, parses like  join($x, $y, $x)
int $x, $y, $z   # int is known to be a unary function, so this parses like  int($x), $y, $z
```

These rules don't matter for purely syntax-highlighting reasons, so previous technologies like `syntax/perl.vim` didn't have to care. But tooling that attempts to use the tree-sitter node tree result as some sort of AST and try to perform real code analysis will need to know about it, so it makes sense to attmept it here.

lucario387: TIL more about perl

queries/perl/injections.scm
rabbiveesh: has been added

queries/pod/highlights.scm
```diff
@@ -0,0 +1,40 @@
...
+
+(head_paragraph (content) @string)
+
+(over_paragraph (content) @string)
+(item_paragraph (content) @string)
+(encoding_paragraph (content) @string)
+
+(verbatim_paragraph (content) @verbatim)
+
+(interior_sequence) @variable
```
ObserverOfTime, line 19:
```suggestion
(interior_sequence
  (sequence_letter) @character
  (content) @text)
```
(maybe)

ObserverOfTime (changes requested):
You should also highlight the angle brackets as `@punctutation.delimiter` if possible.

I would suggest adding these as well:
```scheme
(interior_sequence
  (sequence_letter) @character
  (#eq? @letter "X")
  (content) @text.reference)
(interior_sequence
  (sequence_letter) @character
  (#eq? @letter "E")
  (content) @string.escape)
```

queries/pod/highlights.scm
```diff
@@ -0,0 +1,40 @@
...
+
+(interior_sequence
+  (sequence_letter) @letter
+  (#match? @letter "B")
```
clason, line 23:
```suggestion
  (#match? @letter "B")
```
prefer `#eq?`, `#any-of?`, and `#lua-match?` (in that order)

ObserverOfTime: I assume `#eq?` can't be used because it might have multiple letters. But `#lua-match?` can.

leonerd: `#eq?` would be fine, I think I just wrote this very early before I realised that `#eq?` even existed; having only seen `#match?` elsewhere

clason:
@rabbiveesh we've just switched the format of injection queries to that of upstream, so please rebase and adapt yours. (Holler if you need help, but should be straightforward.)

leonerd:
Damn, I literally just yesterday implemented injections in `Text::Treesitter`. Looks like I'm gonna have to alter it again then.

lucario387:
There isn't much to alter, just that only `@injection.language` and `@injection.content` now

queries/perl/injections.scm
```diff
@@ -1 +1,3 @@
-(comments) @comment
+; an injections.scm file for nvim-treesitter
+(comment) @comment
+(pod) @pod
```
clason, line 3:
```suggestion
((comment) @injection.content
 (#set! injection.language "comment"))
 
((pod) @injection.content
 (#set! injection.language "pod"))
```

clason:
> Damn, I literally just yesterday implemented injections in `Text::Treesitter`. Looks like I'm gonna have to alter it again then.

Oh, you were basing your own implementation on Neovim's custom format? That's not a good idea... The new format is the same that tree-sitter (and everybody else) is using, so you should, too ;)

queries/pod/highlights.scm
```diff
@@ -0,0 +1,40 @@
...
+
+[(pod_directive)
+ (head_directive)
+ (over_directive)
+ (item_directive)
+ (back_directive)
+ (encoding_directive)
+ (cut_directive)] @keyword
+
+(head_paragraph (content) @string)
```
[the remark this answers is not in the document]
clason: and similar below

rabbiveesh:
I'm working on the operator + punctuation highlighting RN, and I'm having an issue. We handle parsing semicolons in our scanner, and I can't for the life of me get tree-sitter to match the semicolon token. Any ideas? 
the obvious
`";" @punctuation.delimiter` is not working, b/c it doesn't come out as a token. And I can't get a match using `#eq?` to work either.

Worst comes to worst i can always rework the grammar to parse semicolon as a normal token, but looking fo rideas

amaanq:
alias $._PERLY_SEMICOLON to ';' wherever it's used

ObserverOfTime:
Or change the name so it isn't hidden.

rabbiveesh:
Rebased + force-pushed so CI can run.

Still open question about the `variable.scalar` and such highlights that i've moved to the bottom of the perl highlights file. Are those not allowed?

ObserverOfTime:
> Still open question about the `variable.scalar` and such highlights that i've moved to the bottom of the perl highlights file. Are those not allowed?

They are, but they won't be highlighted any differently by default.

queries/pod/highlights.scm
rabbiveesh: I left it alone for now b/c it didn't play too nicely with the other interiror seqs below it

rabbiveesh:
Okay, now it's ready!

queries/perl/highlights.scm
```diff
@@ -1,198 +1,107 @@
...
-; Built-in functions
-((call_expression
-  function_name: (identifier) @function.builtin)
- (#any-of? @function.builtin
-   "print" "printf" "sprintf" "say"
-   "push" "pop" "shift" "unshift" "splice"
-   "exists" "delete" "keys" "values"
-   "each"))
+(number) @number
+(version) @number
```
amaanq, line 47:
```suggestion
[
  (number)
  (version)
] @number
```

queries/perl/highlights.scm
```diff
@@ -1,198 +1,107 @@
...
 
-[
-  (map)
-  (grep)
-  (bless)
-] @function.builtin
+(string_literal) @string
+(interpolated_string_literal) @string
+(quoted_word_list) @string
+(command_string) @string
```
amaanq, line 52:
```suggestion
[
  (string_literal)
  (interpolated_string_literal)
  (quoted_word_list)
  (command_string)
] @string
```

queries/perl/highlights.scm
```diff
@@ -1,198 +1,107 @@
...
 
-;; ----------
+[(heredoc_token) (command_heredoc_token)] @label
+(heredoc_content) @string
+(heredoc_end) @label
```
amaanq, line 56:
```suggestion
[
  (heredoc_token)
  (command_heredoc_token)
  (heredoc_end)
] @label
(heredoc_content) @string
```

queries/perl/highlights.scm
```diff
@@ -1,198 +1,107 @@
...
-(named_block_statement
-  function_name: (identifier) @function)
+[(quoted_regexp) (match_regexp)] @string.regex
 
-(function_definition
-  name: (identifier) @function)
+(autoquoted_bareword _?) @string.special
 
-(function) @function
+(hash_element_expression key: (bareword) @string.special)
```
amaanq, line 64: hm, would be good to see an example of this, I'm not sure about this one

queries/perl/highlights.scm
```diff
@@ -1,198 +1,107 @@
...
-  (arrow_operator)
-  (hash_arrow_operator)
-  (array_dereference)
-  (hash_dereference)
-  (to_reference)
-  (type_glob)
-  (hash_access_variable)
-] @operator
+(goto_expression (label) @label)
+(loopex_expression (label) @label)
```
amaanq, line 75: Maybe just mark any label as a label?
```suggestion
(label) @label
```

amaanq (changes requested):

amaanq:
I'm also a bit concerned about state count, a 9000 large state count might be very laggy/slow on lower end machines, just something to consider

queries/perl/highlights.scm
```diff
@@ -1,198 +1,107 @@
...
 
-; Variables
 [
-  (scalar_variable)
-  (array_variable)
-  (hash_variable)
-] @variable
+  "package"
+  "do"
+  "my" "our" "local"
```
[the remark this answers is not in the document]
leonerd: It isn't a qualifier, though. A `local` expression temporarily modifies an existing variable, within a scope. It doesn't create a new variable. (It's badly named but we can't change that now ;) )

leonerd:
> I'm also a bit concerned about state count, a 9000 large state count might be very laggy/slow on lower end machines, just something to consider

Yeah; we're not happy about that either. A big part of that is because tree-sitter itself doesn't have a native concept of list-associative, or non-associative operators. When we tried to properly implement things like chaining comparisons `1 <= $x <= 10` or the non-associative `isa` operator, these all made the resulting parser.c file enormous, and very slow to generate and compile. We'd love to find a better way to implement them without some of the hackery we have, but we've bounced back and forward between lots of other tree-sitter related folks and not found a solution. I suspect to do it better needs tree-sitter *itself* to gain support for list-associative operators. It was rather beyond our scope to make such a feature addition.

But if you have any way on the process, or even know the right folks to talk to, please do get in touch :) We want that fixed.

queries/perl/highlights.scm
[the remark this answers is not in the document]
amaanq: yeah that is badly named, sorry I don't know Perl :)

queries/perl/highlights.scm
ObserverOfTime: `(heredoc_content)` can also be grouped here.

queries/pod/highlights.scm
```diff
@@ -0,0 +1,52 @@
...
+ (item_directive)
+ (back_directive)
+ (encoding_directive)
+ (cut_directive)] @keyword
+
+(head_paragraph (content) @text.title)
+
+(over_paragraph (content) @text.title)
+(item_paragraph (content) @text.title)
+(encoding_paragraph (content) @text.title)
```
ObserverOfTime, line 15: After looking into it a bit more, I'd go with this:

```suggestion
(over_paragraph (content) @number)
(item_paragraph (content) @text)
(encoding_paragraph (content) @string.special)
```

queries/perl/highlights.scm
```diff
@@ -1,198 +1,107 @@
...
-[
-  ","
-  (semi_colon)
-  (start_delimiter)
-  (end_delimiter)
-  (ellipsis_statement)
-] @punctuation.delimiter
+(array_element_expression [array:(_) "->" "[" "]"] @variable.array)
+(slice_expression [array:(_) "->" "[" "]"] @variable.array)
+(keyval_expression [array:(_) "->" "[" "]"] @variable.array)
```
ObserverOfTime, line 102: I would rather keep the punctuation highlighted as such.
```suggestion
(array_element_expression array: _ @variable.array)
(slice_expression array: _  @variable.array)
(keyval_expression array: _ @variable.array)
```

queries/perl/highlights.scm
```diff
@@ -1,198 +1,107 @@
...
 
-(function_attribute) @field
+(hash_element_expression [hash:(_) "->" "{" "}"] @variable.hash)
+(slice_expression [hash:(_) "->" "[" "]"] @variable.hash)
+(keyval_expression [hash:(_) "->" "[" "]"] @variable.hash)
```
ObserverOfTime, line 106: Same as above.
```suggestion
(array_element_expression hash: _ @variable.hash)
(slice_expression hash: _  @variable.hash)
(keyval_expression hash: _ @variable.hash)
```

queries/perl/highlights.scm
```diff
@@ -1,198 +1,107 @@
...
+(ERROR) @error
 
-(escape_sequence) @string.escape
+; these are the ones nvim doesn't like
+[(scalar) (arraylen)] @variable.scalar
+(scalar_deref_expression ["->" "$" "*"] @variable.scalar)
+(array) @variable.array
+(array_deref_expression ["->" "@" "*"] @variable.array)
+(hash) @variable.hash
+(hash_deref_expression ["->" "%" "*"] @variable.hash)
```
ObserverOfTime, line 98: The "->" should be removed from here so it's highlighted as punctuation.
The rest may or may not be kept as is.

ObserverOfTime (changes requested):
Perl is still missing `@punctuation.bracket` highlights, `$#` (maybe `@operator` ?), and regexp modifiers (`@character.special`). Built-in variables (e.g. `ARGV`) should be highlighted as `@variable.builtin` (or `@constant.builtin` where applicable).

Pod should include injections for `=begin` and `=for` blocks. You could also add folds but they aren't necessary.

ObserverOfTime:
Also, if you can make the regexp content a named node or field so it can be captured in textobjects, that'd be great.

queries/perl/highlights.scm
rabbiveesh: Truth is, i think that never happens, b/c anything that's plausibly a bareword would end up being `autoquoted_bareword` already. I'll remove this rule

queries/perl/highlights.scm
```diff
@@ -1,198 +1,107 @@
...
-] @keyword.operator
+[ "map" "grep" ] @function.builtin
 
-; Variables
 [
-  (scalar_variable)
-  (array_variable)
-  (hash_variable)
-] @variable
+  "package"
```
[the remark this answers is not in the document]
rabbiveesh: from what I just read about go it's similar, yes; it sets up the namespace for the following (or containing, depending on if it's block form)

queries/perl/highlights.scm
rabbiveesh: you can see the normal autoquoty behavior with `$hash->{key}`, `key` there is an autoquoted_bareword

queries/perl/highlights.scm
rabbiveesh: I think i'll make it do both `@variable` and also `@variable.$type`, or is that not necessary?

rabbiveesh:
> Perl is still missing `@punctuation.bracket` highlights, `$#` (maybe `@operator` ?), some operators (e.g. `&&`), and regexp modifiers (`@character.special`). Built-in variables (e.g. `ARGV`) should be highlighted as `@variable.builtin` (or `@constant.builtin` where applicable).
> 
> Pod should include injections for `=begin` and `=for` blocks. You could also add folds but they aren't necessary.

The comments that I've marked as resolved are a WIP at tree-sitter-perl/tree-sitter-perl#114 . Would like a second pair of eyes there, b/c in my latest commit, I had to alias a literal token to itself. I feel like either me or tree-sitter has done something horribly, horribly wrong, and I'm not sure which; i'd like to get to the bottom of it

queries/perl/highlights.scm
ObserverOfTime: It will fall back to `@variable` anyway.

ObserverOfTime:
> Would like a second pair of eyes there, b/c in my latest commit, I had to alias a literal token to itself.

If I'm not mistaken, literals directly behind hidden nodes cannot be targeted (or inlined).

rabbiveesh:
> Perl is still missing ... `$#` (maybe `@operator` ?), 

I'm on the fence if `$#` should be `operator` or `variable`. Syntactically it's a sigil, and even though it gives you information about the variable, i'm not sure if that makes it more of an operator than the `@` in scalar context (`my $len = @ary`) is.

I'm leaning more towards highlighting it as variable, which if I understand the defaulting behavior of the highlighting system, it already should do (b/c we have it captured as `variable.scalar`, unless of course it's suffering from a hidden token bug too)

rabbiveesh:
In the following `${name_of_var}`, should those brackets get `@variable` or `@punctuation.bracket`?

Additionally, things like `${_}` technically match a builtin var, should I try and get it to highlight as such or leave that?

ObserverOfTime:
> I'm on the fence if `$#` should be `operator` or `variable`.

Well, Lua uses `#` for length (`#foo`) and it's highlighted as  `@operator`.

> In the following `${name_of_var}`, should those brackets get `@variable` or `@punctuation.bracket`?

We typically use `@punctuation.special` for interpolation brackets.

rabbiveesh:
> > I'm on the fence if `$#` should be `operator` or `variable`.
> 
> Well, Lua uses `#` for length (`#foo`) and it's highlighted as `@operator`.
> 

Lua doesn't have sigils tho. I'm concerned that the coloring will be odd-ish b/c it'll be the only time that the highlight for the name of the var + the sigil don't match. Happy to do whatever you conclude so long as you're considering that particular point.
Picture for visualization (excuse the poor choice of color) 
operator-arylen


> > In the following `${name_of_var}`, should those brackets get `@variable` or `@punctuation.bracket`?
> 
> We typically use `@punctuation.special` for interpolation brackets.

Ah, well that's a different solution. Let's go for that, then

ObserverOfTime:
I'm fine with using `@punctuation.special` for either all or none of the sigils.

amaanq:
Status on this? Does upstream still need some changes @rabbiveesh?

rabbiveesh:
Ah, sorry, I've been adding a few more features. I should be able to get
back t this tonight.
The POD parser doesn't actually support for blocks yet, so I wanna add that
in, and then we should be good.
I'm working on the state count also, if I forgo list associative parsing
then state count goes down to 1k, so mainly I wish upstream supported it,
but I may have a way to do it with GLR that's less taxing re the state
count

rabbiveesh:
Sorry for the delay; here's the final final PR!

We did not do injections for POD yet, though. @leonerd says that the language names aren't standardized, so we're not sure how much usefulness it'll provide. If this is something that's important for `html`, say, then we'd have to update the parser, but I don't see that as a blocker on this PR as is.

The state count is down by an order of magnitutde b/c I finally gave up on getting listassoc to work, so that's a plus too.

There are some consistency questions that I have, so i'm happy for any + all input on this matter!

clason:
The tests already enforce some consistency, so looking at the failures would be a start ;) (You can run them locally with `./script/check-queries.lua`.)

rabbiveesh:
I actually meant INTERNAL consistency, not here.

I need to test a different part of the capture in one of the injections, and that's tripping the queries test. How can I do that w/o getting an error (basically, we inject perl into itself when the replacement of a regexp is a perl expression, ie when the modifiers on your `s` operator is a single `e`)

clason:
If you only need a capture to refer to it in a predicate, you need to mark it as private with an underscore: `@_modifiers`. _That_ is tripping the test.

rabbiveesh:
Ah, thanks a mil, fixed + repushed

queries/perl/highlights.scm
```diff
@@ -1,198 +1,153 @@
...
+(regexp_content) @string.regex
+(replacement) @string
 
-[
-  "(" ")"
-  "[" "]"
-  "{" "}"
-  (standard_input_to_variable)
-] @punctuation.bracket
+(transliteration_content) @string
```
amaanq, line 75: group w/ other strings together too?

queries/perl/highlights.scm
```diff
@@ -1,198 +1,153 @@
...
+(use_statement (package) @type)
+(package_statement (package) @type)
+(require_expression (bareword) @type)
 
-[
-  (regex_option)
-  (regex_option_for_substitution)
-  (regex_option_for_transliteration)
-] @parameter
+(subroutine_declaration_statement name: (_) @function)
```
amaanq, line 83: can a name be different rules? e.g. in Lua it could be like `function foo()` or `function M.foo()`, but you wouldn't want to capture `M.foo`, just `foo`, so ideally capturing identifiers are what you want, but maybe this is correct, not familiar w/ Perl

queries/perl/injections.scm
```diff
@@ -1,2 +1,17 @@
...
+(
+  (substitution_regexp
+    (replacement) @injection.content
+    (substitution_regexp_modifiers) @_modifiers
+    ; match if there's a single `e` in the modifiers list
+    (#match? @_modifiers "e")
+    (#not-match? @_modifiers "e.*e")
+  )
+  (#set! injection.language "perl")
+)
```
amaanq, line 17: keep formatting of parenthesis consistent, on the same line is preferred

queries/pod/highlights.scm
```diff
@@ -0,0 +1,67 @@
...
+  (#eq? @character "B")
+  (content) @text.strong)
+(interior_sequence
+  (sequence_letter) @character
+  (#eq? @character "C")
+  (content) @text.literal)
+(interior_sequence
+  (sequence_letter) @character
+  (#eq? @character "F")
+  (content) @text.underline @string.special)
```
amaanq, line 51: add spacing in between each one? gets a bit hard to find start/end w/ so many similar queries

queries/perl/highlights.scm
```diff
@@ -1,198 +1,153 @@
...
+    "{" @punctuation.special 
+    "}" @punctuation.special 
+  )
+)
+
+
+(
+  [(varname) (filehandle)] @variable.builtin
+  (#match? @variable.builtin "^((ENV|ARGV|INC|ARGVOUT|SIG|STDIN|STDOUT|STDERR)|[_ab]|\\W|\\d+|\\^.*)$")
+)
```
amaanq, line 126: same, formatting (keep on same line, don't dangle parenthesis)

queries/perl/injections.scm
clason: Apropos: how is the query formatter going? ;)

amaanq: Coming along..just getting really caught up w/ school n' stuff, hope to be done by December :) also it will be upstreamed, and I'm thinking of adding to the cli w/ `tree-sitter query format <path>`

rabbiveesh:
> > the language names aren't standardized, so we're not sure how much usefulness it'll provide
> 
> This is a problem that all markup languages share (#812). I don't know how useful language injections would be in pod's case, but if you can do it _without having to update the parser_ then why not?

well, only `for` blocks would work out of the box. Otherwise, we'd need to add parsing for `begin/end` sections, and you'll have to get @leonerd to do that b/c I tried and it didn't go well. But like i said, we can add it in later + take care of it then

ObserverOfTime:
That's fine then.

queries/pod/highlights.scm
```diff
@@ -0,0 +1,67 @@
+; A highlight file for nvim-treesitter to use
+
+[(pod_command)
+ (command)
+ (cut_command)] @keyword
+
+(command_paragraph
+  (command) @keyword
+  (#match? @keyword "^=head")
```
ObserverOfTime, line 9: You should use `lua-match` for these.

queries/pod/highlights.scm
```diff
@@ -0,0 +1,67 @@
...
+  (#not-match? @keyword "^=(head|over|item|encoding)")
+  (content) @string)
+
+(verbatim_paragraph (content) @text.literal)
+
+(interior_sequence
+  (sequence_letter) @character
+  "<" @punctuation.delimiter
+  ">" @punctuation.delimiter
+)
```
ObserverOfTime, line 38:
```suggestion
  [ "<" ">" ] @punctuation.delimiter)
```

queries/perl/injections.scm
```diff
@@ -1,2 +1,17 @@
...
+((pod) @injection.content
+ (#set! injection.language "pod"))
+
+(
+  (substitution_regexp
+    (replacement) @injection.content
+    (substitution_regexp_modifiers) @_modifiers
+    ; match if there's a single `e` in the modifiers list
+    (#match? @_modifiers "e")
+    (#not-match? @_modifiers "e.*e")
```
ObserverOfTime, line 14:
```suggestion
    (#lua-match? @_modifiers "e")
    (#not-lua-match? @_modifiers "e.*e")
```

ObserverOfTime (changes requested):

queries/perl/highlights.scm
```diff
@@ -1,198 +1,153 @@
...
-((scalar_variable) @variable.builtin
- (#eq? @variable.builtin "$#ARGV"))
+(yadayada) @exception
 
-; Integer numbers
-[
-  (integer)
-  (hexadecimal)
-] @number
+(phaser_statement phase: _ @keyword.phaser)
```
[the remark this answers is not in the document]
rabbiveesh: the phaser runs the block at a certain point in the program's lifecycle. for example `BEGIN` runs the code while perl is "compiling" (which is a VERY loose idea, b/c code is run even during "compile" time).
There are also other phasers that exist in the newer class feature, where they are code blocks that run in a specific part of the object's lifecycle. 
I think it makes sense, b/c it is distinct from other kinds of keywords for people that care about it. Looking thru the list of captures, it doesn't seem like there's one that exactly matches

queries/perl/highlights.scm
```diff
@@ -1,198 +1,153 @@
...
+  (heredoc_end)
+] @label
 
-(named_block_statement
-  function_name: (identifier) @function)
+[(escape_sequence) (escaped_delimiter)] @string.escape
 
-(function_definition
-  name: (identifier) @function)
+(_ modifiers: _ @character.special)
```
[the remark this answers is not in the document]
rabbiveesh: yes, it's exactly like that example

queries/pod/highlights.scm
rabbiveesh: is there a particular reason? I'd just rather be able to have the same captures available for when i test using the CLI

rabbiveesh:
Okay, only question remaining is if i can keep using plain-ol match or if we should switch it to `lua-match`

queries/pod/highlights.scm
ObserverOfTime: Performance.

rabbiveesh:
Is CI borked? these failures don't seem to have anything to do w/ my PR

ObserverOfTime:
Please avoid merge commits. :)

rabbiveesh:
oh sorry, i'm used to a squash merge flow. Let me rebase

queries/pod/highlights.scm
clason: _significant_ performance. I am afraid we hve to insist on that. 

(Queries are editor specific; I understand wanting to minimize the divergence, but it's not possible in general to keep them completely agnostic.)

lockfile.json
```diff
@@ -414,7 +414,7 @@
     "revision": "e01767921df18142055d97407595329d7629e643"
   },
   "perl": {
-    "revision": "79e88f64681660f3961939bf764d8f3b4bbb0d27"
+    "revision": "99bbcb0faddfc79db3fd2996cb56689a4685e78c",
```
clason, line 417:
```suggestion
    "revision": "99bbcb0faddfc79db3fd2996cb56689a4685e78c"
```

This is what's breaking CI.

ObserverOfTime: Gotta love JSON.

rabbiveesh: :facepalm:

clason:
> You can squash the commits if you want.

No, please do squash commits, especially if you want to keep perl and pod separate. (In that case, the pod commit needs to be first for the injections to not be broken.)

rabbiveesh:
Sure, but could I just squash it down to one? POD doesn't get too much use outside of perl, so it's effectively one commit. Unless you specifically want them separate

clason:
That's fine, too.

rabbiveesh:
k, squashed; did it in 2 commits b/c the commit log reads bettter that way

ObserverOfTime (approved):

clason:
@amaanq

amaanq (approved):

amaanq:
Thanks for your patience and contribution @rabbiveesh!

rabbiveesh:
:partying_face:  thanks for your patience!

PR #5225, opened 2023-08-11
fix: rename filetype for haskell persistent to match vim name

lykahb:
This is a follow-up for https://github.com/nvim-treesitter/nvim-treesitter/pull/4990 that I opened at about the same time as [PR for Vim](https://github.com/vim/vim/pull/12573) to recognize the filetype for the persistent library.
The recognition is now merged, but with another filetype name. I am grateful to Bram for taking time to explain why the filetype should not have a dot.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -655,7 +655,7 @@ list.haskell_persistent = {
     url = "https://github.com/MercuryTechnologies/tree-sitter-haskell-persistent",
     files = { "src/parser.c", "src/scanner.cc" },
   },
-  filetype = "haskell.persistent",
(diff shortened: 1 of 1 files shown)


lucario387:
Very sorry but I believe this will have to be parked for a while, or you will need to target `main` branch.
Why you may ask? It's because master branch is still targeting `0.9.x` versions of neovim, so this will be a breaking change to some haskell persistent user

amaanq:
I disagree - the language was just added recently anyways

amaanq (approved):

lucario387:
Hmmm, I too agree, now that I see the filetype was only added 2 months ago, so way after 0.9

lucario387 (approved):

PR #5230, opened 2023-08-11
fix downcase

clason:
@ObserverOfTime

### DIFF
--- a/plugin/query_predicates.lua
+++ b/plugin/query_predicates.lua
@@ -46,30 +46,23 @@ query.add_directive('set-lang-from-mimetype!', function(match, _, bufnr, pred, m
   end
 end, true)
 
+--- transform node text to lower case (e.g., to make @injection.language case insensitive)
+---
+---@param match (TSNode|nil)[]
+---@param _ string
+---@param bufnr integer
+---@param pred string[]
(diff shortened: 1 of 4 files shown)


plugin/query_predicates.lua
```diff
@@ -46,30 +46,23 @@ query.add_directive('set-lang-from-mimetype!', function(match, _, bufnr, pred, m
...
   end
 
-  if #pred == 3 then
-    metadata[id][key] = string.lower(text)
-  else
-    metadata[key] = string.lower(text)
+  local text = vim.treesitter.get_node_text(node, bufnr, { metadata = metadata[id] }) or ''
+  if not metadata[id] then
+    metadata[id] = {}
   end
```
ObserverOfTime, line 66: Wouldn't this make more sense?

```suggestion
  if not metadata[id] then
    metadata[id] = {}
  end
  local text = vim.treesitter.get_node_text(node, bufnr, { metadata = metadata[id] }) or ''
```

clason: I think it's better to pass `nil` instead of `{}` here -- at least that's the convention elsewhere.

ObserverOfTime (approved):

PR #5233, opened 2023-08-12
feat(swift): improve highlights

ordazgustavo:
- Add hl to “mutating” modifier
- Add hl to property access
- Add hl to argument name
- Add hl to context aware method call

I noticed that the bang operator (`!`) is not being highlighted but if I add `(bang) @operator` it works, I guess this is a parser thing (?). Should I add this highlight?

<details>
<summary>Before/After</summary>

**Before:**
Screenshot 2023-08-12 at 11 59 41

**After:**
Screenshot 2023-08-12 at 12 23 59
</details>

### DIFF
--- a/queries/swift/highlights.scm
+++ b/queries/swift/highlights.scm
@@ -17,6 +17,7 @@
   (property_modifier)
   (parameter_modifier)
   (inheritance_modifier)
+  (mutation_modifier)
 ] @type.qualifier
 
 (function_declaration (simple_identifier) @method)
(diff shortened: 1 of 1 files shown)


clason:
> I noticed that the bang operator (!) is not being highlighted but if I add (bang) @operator it works, I guess this is a parser thing (?). Should I add this highlight?

Sure (if it is an operator in Swift).

amaanq:
Yeah you can add it

queries/swift/highlights.scm
```diff
@@ -179,3 +185,4 @@
  "..<"
  "..."
 ] @operator
+(bang) @operator
```
amaanq, line 188:
```suggestion
 "..."
 (bang)
] @operator
```

PR #5236, opened 2023-08-13
feat(c): update parser, add K&R-style function parameter support

amaanq:
```c
int foo(bar, baz, qux)
int bar, baz;
char *qux;
{
}
```

Is now supported

Well an issue might be that C++ just doesn't support this at all..

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -33,7 +33,7 @@
     "revision": "7f1a5df44861291d6951b6b2146a9fef4c226e14"
   },
   "c": {
-    "revision": "54fa1febf4f3b1baaba1066a615e569f58c8e765"
+    "revision": "a2b7bac3b313efbaa683d9a276ff63cdc544d960"
   },
   "c_sharp": {
     "revision": "1648e21b4f087963abf0101ee5221bb413107b07"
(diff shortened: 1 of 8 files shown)


theHamsta:
Would be really glad if this style could be forgotten in history.

lucario387:
Think it's nicer to either add this to C only, and add a `c_common` query for all c-family languages then

theHamsta:
> Think it's nicer to either add this to C only, and add a `c_common` query for all c-family languages then

But the K&R style is currently enabled in all C language if it is not explicitly overwritten. At least this would be my guess (@amaanq?).  If they are explictly forbidden by the child languages then they would cause an error and we would need `c_common`, if they are allowed then we might as well also run the C queries for all languages (as they will probably never triggered).

amaanq:
> Would be really glad if this style could be forgotten in history.

I agree, but sometimes people do work with legacy C code. Maybe I can just leave it commented w/ a note saying "if you'd like to use this, add it in your own config with `; extends` at the top" to avoid the breakage

clason:
It's fine to enable it if it doesn't break anything else.

amaanq:
it breaks any cpp language since cpp doesn't support it, but the `c_common` route is doable

clason:
This will make syncing the bundled queries significantly harder...

amaanq:
Yeah so I'll just comment it out

ObserverOfTime:
> I agree, but sometimes people do work with legacy C code.

I doubt even Kernighan himself uses that style anymore.

ObserverOfTime (approved):

PR #5239, opened 2023-08-13
ci: dont hardcode the rockspec version

teto:
sry for that but picking up the old PR, I forgot to remove some cruft as mentioned by @mrcjkb see https://github.com/nvim-treesitter/nvim-treesitter/pull/4403#pullrequestreview-1575769713 .

Seems like I've lost the ability to set LUAROCKS_API_KEY (but I should have already added one) and tag a release so once this is merged, it would be nice if an admin could tag a release to make sure this works and the rockspec is properly uploaded to luarocks.org


clason:
We won't make a release until 1.0, I'm afraid, and use tags to mark "last compatible versions" before bumping Neovim requirements.

PR #5265, opened 2023-08-16
Update tree-sitter-robot version and maintainer

Hubro:
I have completed and polished the tree-sitter parser for Robot files. This PR:

- Updates tree-sitter-robot to the latest ref
- Removes "experimental" and sets me as maintainer
- Improves syntax highlighting
- Adds queries for indents and folds

### DIFF
--- a/queries/robot/folds.scm
+++ b/queries/robot/folds.scm
@@ -1,0 +1,5 @@
+[
+ (section)
+ (keyword_definition)
+ (test_case_definition)
+] @fold
--- a/queries/robot/folds.scm
+++ b/queries/robot/folds.scm
@@ -1,5 +1,5 @@
 [
- (section)
- (keyword_definition)
- (test_case_definition)
+  (section)
+  (keyword_definition)
+  (test_case_definition)
 ] @fold
--- a/queries/robot/highlights.scm
+++ b/queries/robot/highlights.scm
@@ -1,21 +1,53 @@
-(argument (dictionary_variable) @string.special)
-(argument (list_variable) @string.special)
-(argument (scalar_variable) @string.special)
-(argument (text_chunk) @string)
+(comment) @comment
 
-(keyword_invocation (keyword) @function)
-
-(test_case_definition (name) @property)
-
-(keyword_definition (body (keyword_setting) @keyword))
-(keyword_definition (name) @function)
-
-(variable_definition (variable_name) @variable)
+(section_header) @keyword
+(extra_text) @comment
 
 (setting_statement) @keyword
 
-(extra_text) @comment
-(section_header) @keyword
+(variable_definition (variable_name) @variable)
+
+(keyword_definition (name) @function)
+(keyword_definition (body (keyword_setting) @keyword))
+
+(test_case_definition (name) @property)
+
+(keyword_invocation (keyword) @function.call)
+(ellipses) @punctuation.delimiter
+
+(argument (text_chunk) @string)
+(argument (scalar_variable) @variable)
+(argument (list_variable) @variable)
+(argument (dictionary_variable) @variable)
+(argument (inline_python_expression) @string.special)
 
 (ellipses) @punctuation.delimiter
-(comment) @comment
+
+; Control structures
+(for_statement
+  "FOR" @repeat
+  "END" @repeat)
+(for_statement (in "IN" @repeat))
+(for_statement (in_range "IN RANGE" @repeat))
+(for_statement (in_enumerate "IN ENUMERATE" @repeat))
+(for_statement (in_zip "IN ZIP" @repeat))
+
+(while_statement
+  "WHILE" @repeat
+  "END" @repeat)
+
+(break_statement) @repeat
+(continue_statement) @repeat
+
+(if_statement
+  "IF" @conditional
+  "END" @conditional)
+(if_statement (elseif_statement "ELSE IF" @conditional))
+(if_statement (else_statement "ELSE" @conditional))
+
+(try_statement
+  "TRY" @exception
+  "END" @exception)
+(try_statement (except_statement "EXCEPT" @exception))
+(try_statement (else_statement "ELSE" @exception))
+(try_statement (finally_statement "FINALLY" @exception))
--- a/queries/robot/highlights.scm
+++ b/queries/robot/highlights.scm
@@ -1,53 +1,57 @@
-(comment) @comment
+[
+  (comment)
+  (extra_text)
+] @comment
 
-(section_header) @keyword
-(extra_text) @comment
-
-(setting_statement) @keyword
+[
+  (section_header)
+  (setting_statement)
+  (keyword_setting)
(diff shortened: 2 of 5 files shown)


queries/robot/folds.scm
```diff
@@ -0,0 +1,5 @@
+[
+ (section)
+ (keyword_definition)
+ (test_case_definition)
+] @fold
```
amaanq, line 5:
```suggestion
[
  (section)
  (keyword_definition)
  (test_case_definition)
] @fold
```

queries/robot/highlights.scm
```diff
@@ -1,21 +1,53 @@
...
-(argument (scalar_variable) @string.special)
-(argument (text_chunk) @string)
+(comment) @comment
 
-(keyword_invocation (keyword) @function)
+(section_header) @keyword
+(extra_text) @comment
 
-(test_case_definition (name) @property)
+(setting_statement) @keyword
```
amaanq, line 6: Can you group comments and keywords together? Like so:

```scm
[
  (comment)
  (extra_text)
] @comment
```

queries/robot/highlights.scm
```diff
@@ -1,21 +1,53 @@
...
-(setting_statement) @keyword
+(keyword_invocation (keyword) @function.call)
+(ellipses) @punctuation.delimiter
 
-(extra_text) @comment
-(section_header) @keyword
+(argument (text_chunk) @string)
+(argument (scalar_variable) @variable)
+(argument (list_variable) @variable)
+(argument (dictionary_variable) @variable)
```
amaanq, line 21: Does it need to be inside an argument to be a variable? Maybe the actual variables can just be variables alone and you can group them together

queries/robot/highlights.scm
```diff
@@ -1,21 +1,53 @@
...
-(comment) @comment
+
+; Control structures
+(for_statement
+  "FOR" @repeat
+  "END" @repeat)
+(for_statement (in "IN" @repeat))
+(for_statement (in_range "IN RANGE" @repeat))
+(for_statement (in_enumerate "IN ENUMERATE" @repeat))
+(for_statement (in_zip "IN ZIP" @repeat))
```
amaanq, line 33: Same here, does "FOR" appear in other places? If not, you can just make a group for all of these and mark them as repeat

```scm
[
  "FOR"
  "END"
  "IN"
  ; etc
] @repeat

File: queries/robot/highlights.scm [hunk 5]
```diff
@@ -1,21 +1,53 @@
...
+  "END" @conditional)
+(if_statement (elseif_statement "ELSE IF" @conditional))
+(if_statement (else_statement "ELSE" @conditional))
+
+(try_statement
+  "TRY" @exception
+  "END" @exception)
+(try_statement (except_statement "EXCEPT" @exception))
+(try_statement (else_statement "ELSE" @exception))
+(try_statement (finally_statement "FINALLY" @exception))
```
amaanq, line 53: Same as above

amaanq (changes requested):

queries/robot/highlights.scm
Hubro: I don't think I can do that, unfortunately Robot has "unquoted strings" and very lose naming rules for keywords (which are basically functions). This means that "END" can either be a built-in keyword, a user-defined keyword or a chunk of a string. Furthermore, the "END" in a loop is a different capture from the "END" of a conditional, which is a different capture from the "END" in a TRY/CATCH statement.

amaanq: Ah, I see. How about leaving the specific "END" capture as is and grouping the others like FOR, IN, etc?

Hubro: > Also, bare identifiers named "END" or "FOR" aren't captured here, it is literal string symbols that are used in the grammar, or error recovery nodes

I'm not sure I followed that. Are you saying that if I have a node (for example `text_chunk`) that contains the text `FOR`, then `"FOR" @repeat` won't match it? :thinking:

amaanq: yes, if it's a terminal symbol that relies on regex

Hubro: To clarify, you're saying `"FOO"` in a query will only match a string that is literally defined as `"FOO"` in the grammar?

In either case, I tested it and it seems to only match the keywords, so all good :+1:

amaanq: yep, exactly

Hubro: @amaanq I updated it now, though to be honest, I liked it better before. It clearly and explicitly followed the syntax of the language. Now it's more jumbled and it feels much easier to make a mistake. :thinking:

queries/robot/highlights.scm
```diff
@@ -1,21 +1,58 @@
...
-(test_case_definition (name) @property)
+(variable_definition (variable_name) @variable)
 
-(keyword_definition (body (keyword_setting) @keyword))
 (keyword_definition (name) @function)
+(keyword_definition (body (keyword_setting) @keyword))
 
-(variable_definition (variable_name) @variable)
+(test_case_definition (name) @function)
+(test_case_definition (body (test_case_setting) @keyword))
```
amaanq, line 17: Same question, can these two keyword nodes be keyword in every case or does it need to be this specific? If not, keyword_setting and test_case_setting can be moved to the chunk above with other keywords

queries/robot/highlights.scm
amaanq: Tbh, this is subjective for sure, I think in this case you can keep them as you had it before, my *main* want was to have groups of similar nodes (like keyword) grouped together. But since there's going to be a nested capture for the different END types, you can split up the literals like FOR, WHILE, and keep them with their respective nodes too, lua does this.

queries/robot/highlights.scm
Hubro: Those nodes can only occur in this location, so it felt natural to capture it like this. I've changed it now.

amaanq: It's easier to group them together as a "keyword" block if they can only be keywords :)

See my other comment about FOR/END etc, your old style is totally okay too (Lua does this)

amaanq (approved):

amaanq:
Awesome, looks good now and what I meant in my comment earlier is that you can go back to your old convention of the for/while stuff which was something like so (if you'd like)

```scm
(for_statement [ "FOR" "END" ] @repeat)
```

Hubro:
> what I meant in my comment earlier is that you can go back to your old convention of the for/while stuff

I moved the `"WHILE"` capture down so it's more logically grouped, I'm happy with it.

One question, I made the latest changes directly in `nvim-treesitter`, so now the lock file is pointing to an older version of `tree-sitter-robot` that has different queries. Should I update the lockfile or should I not bother?

amaanq:
your repos queries don't matter here, and we have a CI that updates parser revs daily, though I don't mind if you update it here

PR #5270, opened 2023-08-17
feat: add BitBake

amaanq:
Just a note that this language allows for python function definitions annoyingly, hence the major overlap

### DIFF
--- a/queries/bitbake/folds.scm
+++ b/queries/bitbake/folds.scm
@@ -1,0 +1,29 @@
+[
+  (function_definition)
+  (anonymous_python_function)
+  (python_function_definition)
+
+  (while_statement)
+  (for_statement)
+  (if_statement)
+  (with_statement)
+  (try_statement)
+
+  (import_from_statement)
+  (parameters)
+  (argument_list)
+
+  (parenthesized_expression)
+  (generator_expression)
+  (list_comprehension)
+  (set_comprehension)
+  (dictionary_comprehension)
+
+  (tuple)
+  (list)
+  (set)
+  (dictionary)
+
+  (string)
+  (python_string)
+] @fold
--- a/queries/bitbake/indents.scm
+++ b/queries/bitbake/indents.scm
@@ -1,0 +1,127 @@
+[
+  (import_from_statement)
+
+  (parenthesized_expression)
+  (generator_expression)
+  (list_comprehension)
+  (set_comprehension)
+  (dictionary_comprehension)
+
+  (tuple_pattern)
+  (list_pattern)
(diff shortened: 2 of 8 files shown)


queries/bitbake/indents.scm
```diff
@@ -0,0 +1,130 @@
...
+ (#set! indent.immediate 1))
+((try_statement) @indent.begin
+ (#set! indent.immediate 1))
+(ERROR "try" ":" @indent.begin (#set! indent.immediate 1))
+((python_function_definition) @indent.begin
+ (#set! indent.immediate 1))
+((function_definition) @indent.begin
+ (#set! indent.immediate 1))
+((anonymous_python_function) @indent.begin
+ (#set! indent.immediate 1))
```
lucario387, line 46: From bitbake docs (is it the doc? https://docs.yoctoproject.org/bitbake/bitbake-user-manual/bitbake-user-manual-metadata.html#functions)

Then functions like `func(){}` are allowed, which means normal C-like functions. If this is the normal function, then it probably doesn't need `indent.immediate`, while python style's function should have it

amaanq: yeah you're right :sweat_smile: 

only python_function should retain that

lucario387:
According to the doc, this is a valid if-else sequence:
```
if [condition] ; then
  # something
else 
  # something else
fi
```
This looks more bash-like than python-like

amaanq:
> According to the doc, this is a valid if-else sequence:
> 
> ```
> if [condition] ; then
>   # something
> else 
>   # something else
> fi
> ```
> 
> This looks more bash-like than python-like

https://github.com/amaanq/tree-sitter-bitbake/blob/43e55585c1fb60b9da3b7da821f584cf06d6d107/grammar.js#L219

It is, and the node is marked as shell_content and is injected w/ bash

amaanq:
I'm also thinking about getting rid of the python embedding and just considering every line that doesn't start at col 0 as python/bash code, but the spec isn't clear about how that works :smile:

ObserverOfTime:
Oh boy, another Python-like build tool.

queries/bitbake/folds.scm
```diff
@@ -0,0 +1,29 @@
...
+  (list_comprehension)
+  (set_comprehension)
+  (dictionary_comprehension)
+
+  (tuple)
+  (list)
+  (set)
+  (dictionary)
+
+  (string)
```
ObserverOfTime, line 27: Can non-Python strings be multiline?

amaanq: yes

PR #5272, opened 2023-08-17
feat: add CSV, PSV, & TSV

amaanq:


ObserverOfTime:
Where's PSV? :^)

amaanq:
Oh I can add that..didn't know it was a thing :)

ObserverOfTime:
I've seen it in some data dumps.

amaanq:
added PSV

ObserverOfTime (approved):

lucario387:
@amaanq for csv, all fields are text, as can be seen from the syntax tree

```
1, abcd, true, false, 1.2
```
The nodes in order are `text, text, boolean, boolean, text`

amaanq:
you're right, I wrote this in like 30m at 3am for fun :sweat_smile: will fix shortly

PR #5274, opened 2023-08-17
Update README.md for better folding settings

carschandler:
I think that the README's recommended settings for enabling treesitter's folding capabilities do not reflect how most people probably want their folds to behave. The current settings make it so that when a single section of code is folded, everything folds (due to how `foldlevel` works)... most people probably want the folds to be created for them automatically, sit in the background invisibly, and then only activate when they manually toggle them. By setting `foldlevelstart`, this is achieved, and the user can still use `:set foldlevel` to batch fold if desired. Most people probably do not already have this option set, so including it in the README rather than `nofoldenable` seems like it would be a desirable change to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,9 +474,9 @@ set foldlevelstart = 99
 or if you prefer lua:
 
 ```lua
-vim.wo.foldmethod = "expr"
-vim.wo.foldexpr = "nvim_treesitter#foldexpr()"
-vim.go.foldlevelstart = 99
+vim.o.foldmethod = "expr"
+vim.o.foldexpr = "nvim_treesitter#foldexpr()"
+vim.o.foldlevelstart = 99
 ```
 
 This will respect your `foldminlines` and `foldnestmax` settings.
--- a/README.md
+++ b/README.md
@@ -462,24 +462,29 @@ require'nvim-treesitter.configs'.setup {
 
 Tree-sitter based folding. _(Technically not a module because it's per windows and not per buffer.)_
 
-```vim
-set foldmethod=expr
-set foldexpr=nvim_treesitter#foldexpr()
-" With the above settings applied, vim will fold everything by default when
-" opening a file... setting foldlevelstart to a high value allows you to
-" start with no folds and then toggle them as desired with the za keymap.
-set foldlevelstart = 99
-```
-
-or if you prefer lua:
-
 ```lua
 vim.o.foldmethod = "expr"
-vim.o.foldexpr = "nvim_treesitter#foldexpr()"
+vim.o.foldexpr = "v:lua.vim.treesitter.foldexpr()"
+-- With the above settings applied, vim will fold everything by default when
+-- opening a file... setting foldlevelstart to a high value allows you to
+-- start with no folds and then toggle them as desired with the za keymap.
 vim.o.foldlevelstart = 99
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -465,7 +465,18 @@ Tree-sitter based folding. _(Technically not a module because it's per windows a
...
+" start with no folds and then toggle them as desired with the za keymap.
+set foldlevelstart = 99
+```
+
+or if you prefer lua:
+
+```lua
+vim.wo.foldmethod = "expr"
+vim.wo.foldexpr = "nvim_treesitter#foldexpr()"
+vim.go.foldlevelstart = 99
```
lucario387, line 479: Not that `vim.{wo,go}` is any wrong, just maybe not intended since `set ...` can also be global, so I'd say maybe `vim.o`

carschandler: You're right... it would be preferred to set the options globally with `vim.o` so that if a user wanted to change them locally and then revert to global settings with `:set option<` they would be able to do so. I'll make that change. Thanks!

lucario387: If anything, can you also change the `foldexpr` to `vim.o.foldexpr = 'v:lua.vim.treesitter.foldexpr()'`, as `nvim_treesitter#foldexpr()` is already deprecated/planned to remove. 

See https://github.com/nvim-treesitter/nvim-treesitter/tree/main#folds

carschandler: Thanks! I was looking for a more lua way to do that but gave up on it. I'll add that as well... and a comment about setting the fold options per-filetype.

amaanq (approved):

README.md
```diff
@@ -462,13 +462,29 @@ require'nvim-treesitter.configs'.setup {
...
+vim.o.foldlevelstart = 99
+```
+
+or if you prefer vimscript:
+
 ```vim
 set foldmethod=expr
-set foldexpr=nvim_treesitter#foldexpr()
-set nofoldenable                     " Disable folding at startup.
+set foldexpr=v:lua.vim.treesitter.foldexpr()
```
wookayin, line 478: Built-in foldexpr seems available from neovim 0.9+ and ported from this repo (https://github.com/neovim/neovim/pull/22299). Does this mean that `nvim_treesitter#foldexpr()` is deprecated and no longer used?

clason: That is correct. (I mentioned that the README is a mess, didn't I?)

DanilaMihailov:
I wanted to propose the same changes for README. And maybe add this options as well?
```lua
vim.opt.foldtext = '' -- "transparent folds - just text with syntax highlighting"
```

Is there anything else needed for this PR to be merged?

clason:
The problem is that folding has changed significantly in Nvim 0.10, so I want to wait for that to be released and (official) support for Nvim 0.9 can be dropped so we don't need to document for multiple versions in the README.

clason:
Updated the README with the minimal change (relevant for 0.9). Thanks!

carschandler:
Cool! I still think that it would be beneficial to include something about `foldlevelstart` since the options in the README currently still cause all folds to be fully collapsed by default which is rarely what the end user desires.

clason:
Yes, but that is not specific to the treesitter folding. We should not duplicate upstream documentation. It is an explicit project goal to move away from being (or even looking like) a separate thing rather than a collection of queries for Neovim features.

Issue #5244, opened 2023-08-14
(injection): the @injection.content of the regex injection in Python is wrong


TheLeoP:
### Describe the highlighting problem

`@injection.content` is set to the whole string (i.e. `r'...'`) instead of just the content of the string

### Example snippet that causes the problem

```python
import re

foo = ['this should not be parsed as a regex']

re.sub(r".*", "bar", "baz")
```

### Tree-sitter parsing result

```
(term) ; [1:1 - 5:27] regex
 (pattern_character) ; [1:1 - 1] regex
 (pattern_character) ; [1:2 - 2] regex
 (pattern_character) ; [1:3 - 3] regex
 (pattern_character) ; [1:4 - 4] regex
 (pattern_character) ; [1:5 - 5] regex
 (pattern_character) ; [1:6 - 6] regex
 (pattern_character) ; [1:7 - 7] regex
 (pattern_character) ; [1:8 - 8] regex
 (pattern_character) ; [1:9 - 9] regex
 (pattern_character) ; [3:1 - 1] regex
 (pattern_character) ; [3:2 - 2] regex
 (pattern_character) ; [3:3 - 3] regex
 (pattern_character) ; [3:4 - 4] regex
 (pattern_character) ; [3:5 - 5] regex
 (pattern_character) ; [3:6 - 6] regex
 (character_class) ; [3:7 - 46] regex
  "[" ; [3:7 - 7] regex
  (class_character) ; [3:8 - 8] regex
  (class_character) ; [3:9 - 9] regex
  (class_character) ; [3:10 - 10] regex
  (class_character) ; [3:11 - 11] regex
  (class_character) ; [3:12 - 12] regex
  (class_character) ; [3:13 - 13] regex
  (class_character) ; [3:14 - 14] regex
  (class_character) ; [3:15 - 15] regex
  (class_character) ; [3:16 - 16] regex
  (class_character) ; [3:17 - 17] regex
  (class_character) ; [3:18 - 18] regex
  (class_character) ; [3:19 - 19] regex
  (class_character) ; [3:20 - 20] regex
  (class_character) ; [3:21 - 21] regex
  (class_character) ; [3:22 - 22] regex
  (class_character) ; [3:23 - 23] regex
  (class_character) ; [3:24 - 24] regex
  (class_character) ; [3:25 - 25] regex
  (class_character) ; [3:26 - 26] regex
  (class_character) ; [3:27 - 27] regex
[...]

PR #5275, opened 2023-08-17
fix(lua): luap injection

TheLeoP:
This fixes a problem similar to #5244, but for lua.

In the following code, the injection of `luap` causes errors outside of the string because the string content node has children that are not being parsed in the injection (in this case, the `escape_sequence` node).

```lua
--- this isn't a lua pattern
string.match("123", "\0%d+")
```

To solve this, `(#set! injection.include-children)` was added to the captures.

Also, I made the captrues slightly more specific. If something is wrong with this logic, it can be reverted.

### DIFF
--- a/queries/lua/injections.scm
+++ b/queries/lua/injections.scm
@@ -39,30 +39,50 @@
   (#offset! @injection.content 0 1 0 0))
 
 ; string.match("123", "%d+")
+
 (function_call
   (dot_index_expression
     field: (identifier) @_method
     (#any-of? @_method "find" "match"))
-  arguments: (arguments (_) . (string content: _ @injection.content (#set! injection.language "luap"))))
+  arguments: (arguments
+               . (_)
+               . (string
+                   content: (string_content) @injection.content
+                   (#set! injection.language "luap")
+                   (#set! injection.include-children))))
 
 (function_call
   (dot_index_expression
     field: (identifier) @_method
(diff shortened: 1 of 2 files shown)


queries/lua/injections.scm
```diff
@@ -39,30 +39,50 @@
...
   (dot_index_expression
     field: (identifier) @_method
     (#any-of? @_method "find" "match"))
-  arguments: (arguments (_) . (string content: _ @injection.content (#set! injection.language "luap"))))
+  arguments: (arguments
+               . (_)
+               . (string
+                   content: (string_content) @injection.content
+                   (#set! injection.language "luap")
+                   (#set! injection.include-children))))
```
lucario387, line 52: I'd prefer the stylings to be more like this
```suggestion  
  arguments: 
    (arguments
      . (_)
      . 
      (string
        content: (string_content) @injection.content
        (#set! injection.language "luap")
        (#set! injection.include-children))))
```

But that is subjective.

lucario387 (approved):

lucario387:
If possible, is it fine for me/can you add these lines to `test.lua` in highlights?

```lua
-- Checking for incorrect hlgroup of injected luap
string.match(s, "\0%d[^\n]+")
--                       ^ !constant
```

amaanq:
weird that the linter checks the test file

PR #5277, opened 2023-08-18
chore(csv,psv,tsv): update parser to fix numbers

amaanq:
I know the lockfile can update this, but I wasted 30m on what should've been a 2m fix because Javascript, and maybe someone someday will stumble upon this w/ the same issue. When using `new RegExp`, \n and \r are escaped automatically, but not others like \s and \d, so I had a token that looked something like this:

```json
{
  "type": "PATTERN",
  "value": "[^,ds\"][^, \\n\\r\"]+"
},
```

ideally d and s would be escaped too, but who knows why it isn't...

Here's the modified rule, which needs 2 backslashes for d and s to process properly (and I did so for \n and \r for consistency)

```js
  text: _ => token(choice(
    new RegExp(`[^${separator}\\d\\s"][^${separator} \\n\\r"]+`),
    seq('"', repeat(choice(/[^"]/, '""')), '"'),
  )),
```

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -78,7 +78,7 @@
     "revision": "5f2c94b897601b4029fedcce7db4c6d76ce8a128"
   },
   "csv": {
-    "revision": "1f584470170b708880b024142d901b05b2b2b602"
+    "revision": "f1d35df780976721d3cd38f0b16538dd31f87a23"
   },
   "cuda": {
     "revision": "607851cc524d9f5b5a5cd31ad4044fd431cee54e"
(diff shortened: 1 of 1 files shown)


lucario387:
> When using new RegExp, \n and \r are escaped automatically

I never know this is the default behavior tbh, I have always been double escaping everything while using RegExp because of `no way JS is that smart`

amaanq:
I initially monitored for \n working (and it did) so I assumed \d and \s would as well

ObserverOfTime:
[String.raw](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw) is useful. :)

PR #5278, opened 2023-08-18
feat(java): add highlights to string interpolation

lucario387:
Introduced in tree-sitter/tree-sitter-java#157

`@punctuation.bracket` might not be the most fitting for this


amaanq:
use special not bracket

lucario387:
Right now closing `}` of string interpolation is highlighted with `punctuation.bracket`, so what should that be fixed into for consistency

amaanq (approved):

PR #5281, opened 2023-08-18
feat(jsonnet): update parser and queries

Duologic:
We've done a few refactoring rounds, improved highlighting and added folding and locals.

The queries in this PR are maintained and copied from upstream.

### DIFF
--- a/queries/jsonnet/highlights.scm
+++ b/queries/jsonnet/highlights.scm
@@ -1,21 +1,56 @@
+(id) @variable
+(comment) @comment
+
+; Literals
+(null) @constant.builtin
+(string) @string
+(number) @number
 [
   (true)
   (false)
 ] @boolean
 
-(comment) @comment
-(id) @variable
-(import) @include
-(null) @constant.builtin
-(number) @number
-(string) @string
+; Keywords
+"for" @repeat
+"in" @keyword.operator
+"function" @keyword.function
+[
+  "if"
+  "then"
+  "else"
+] @conditional
+[
+  (local)
+  (tailstrict)
+  "function"
+  "assert"
+  "error"
+] @keyword
 
-(fieldname (id) @field)
-(fieldaccess
-  last: (id) @field)
-(fieldaccess_super
-  (id) @field)
+[
+  (dollar)
+  (self)
+  (super)
+] @variable.builtin
+((id) @variable.builtin
+ (#eq? @variable.builtin "std"))
 
+; Operators
+[
+  (multiplicative)
+  (additive)
+  (bitshift)
+  (comparison)
+  (equality)
+  (bitand)
+  (bitxor)
+  (bitor)
+  (and)
+  (or)
+  (unaryop)
+] @operator
+
+; Punctuation
 [
   "["
   "]"
@@ -30,63 +65,100 @@
   ","
   ";"
   ":"
-  "::"
-  ":::"
 ] @punctuation.delimiter
 
-; Nested field inheritance
-("+" @punctuation.delimiter [":" "::" ":::"])
-
-"=" @operator
-(unaryop) @operator
-(binary
-  operator: (_) @operator)
-
-"for" @repeat
-
-"function" @keyword.function
-
-"in" @keyword.operator
-
 [
- (local)
- (tailstrict)
- "assert"
-] @keyword
+  "::"
+  ":::"
+] @punctuation.special
 
+(field
+  (fieldname) "+" @punctuation.special)
+
+; Imports
 [
-  "else"
-  "if"
-  "then"
-] @conditional
+  (import)
+  (importstr)
+] @include
 
-"error" @exception
+; References
 
-[
-  (dollar)
-  (self)
-  (super)
-] @variable.builtin
-((id) @variable.builtin
- (#eq? @variable.builtin "std"))
+; Make reference same color as parameter 
+; (may incur performance issues on big files)
+; Depends on locals.scm
+((id) @parameter.reference
+ (#is? @parameter.reference parameter))
 
-; Function declaration
+((id) @function.reference
+ (#is? @function.reference function))
+
+((id) @var.reference
+ (#is? @var.reference var))
+((id) @define
+ (#is? @var.reference var))
+
+; References do not apply to static field IDs
+; Workaround for `(#is-not? local)` not supported
+(fieldname (id) @field)
+(fieldname (string
+             (string_start) @text.strong
+             (string_content) @field
+             (string_end) @text.strong
+           ))
+
+; But it does apply if ID in an expression
+(fieldname
+ ("["
+  (id) @parameter.reference
+  "]"
+  (#is? @parameter.reference parameter)))
+(fieldname
+ ("["
+  (id) @define
+  "]"
+  (#is? @var.reference var)))
+
+; Functions
+(field
+  function: (fieldname (id) @function))
+(field
+  function: (fieldname
+              (string
+                (string_start) @text.strong
+                (string_content) @function
+                (string_end) @text.strong
+              )))
 (param
   identifier: (id) @parameter)
-(field
-  (fieldname (id) @function)
-  [(anonymous_function) (params)])
-(bind
-  function: (id) @function)
+
+(bind (id) @define)
+(bind function: (id) @function)
 
 ; Function call
-(functioncall (id) @function.call)
 (functioncall
   (fieldaccess
-    last: (id) @function.call))
-(functioncall
+    last: (id) @function.call
+  )?
   (fieldaccess_super
-    (id) @function.call))
-(named_argument (id) @parameter)
+    (id) @function.call
+  )?
+  (id)? @function.call
+  "("
+  (args
+    (named_argument
+      (id) @parameter
+    )
+  )?
+  ")"
+)
 
+; Emphasize implicit plus usage
+(implicit_plus
+  (_ "}"? @text.danger)
+  (object
+    "{" @text.danger
+  )
+)
+
+; ERROR
 (ERROR) @error
--- a/queries/jsonnet/locals.scm
+++ b/queries/jsonnet/locals.scm
@@ -1,0 +1,18 @@
+(parenthesis) @scope
+(anonymous_function) @scope
+(object) @scope
+(field) @scope
+(local_bind) @scope
+
+(field
+  function: (fieldname (id)  @definition.function)
+  (#set! "definition.function.scope" "parent")
+)
+
+(bind (id)  @definition.var)
+(bind function: (id)  @definition.function)
+
+(param (id) @definition.parameter)
+
+(id) @reference
+;(fieldname (id) (#is-not? local))
--- a/queries/jsonnet/folds.scm
+++ b/queries/jsonnet/folds.scm
@@ -1,0 +1,10 @@
+[
+ (field)
+ (object)
+ (array)
+ (parenthesis)
+ (bind)
+; (params)
+; (args)
+; (conditional)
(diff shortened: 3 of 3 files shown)


amaanq:
rebase

Duologic:
Rebased.

You should get rid of either the `main` or `master` branch, this is confusing.

lucario387:
> You should get rid of either the `main` or `master` branch, this is confusing.

If the users are all happy for us to develop and release that targets only `nightly` neovim, then yay

But for now, it's not that possible to do that.

clason:
> Rebased.
> 
> You should get rid of either the `main` or `master` branch, this is confusing.

No. This is a deliberate development strategy, and I don't see how that should be confusing:
1.  The `master` branch is marked as `default`, so that's what you get if you don't specify a branch (which you shouldn't).
2. The `main` branch is clearly marked as "WIP, do not use!".

clason:
You need to stick to  capture names from the list in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights (this is enforced in order to allow colorschemes to cover all languages equally).

queries/jsonnet/highlights.scm
```diff
@@ -30,63 +65,100 @@
...
-] @variable.builtin
-((id) @variable.builtin
- (#eq? @variable.builtin "std"))
+((id) @function.reference
+ (#is? @function.reference function))
+
+;((id) @var.reference
+; (#is? @var.reference var))
+;((id) @define
+; (#is? @var.reference var))
```
Duologic, line 98: I've commented this out now, any clue on how I can get references for things that aren't functions/parameters?

clason: You can't get references with treesitter highlighting (in the sense of LSP, if I understand you correctly). You'd need locals queries for that, but Neovim does not support that for highlighting (and upstream has soured on that as well).

Duologic: Mind running the CI checks?

Duologic:
re `main` vs. `master`, sorry about bringing it up, I didn't want to offend anyone, I feels like I annoyed y'all with this.

> I don't see how that should be confusing

I just got confused by it as in many cases when there is a `main` branch it is usually the 'new' default branch as that is what Github has been pushing us for. This let me spent quite a bit of time figuring out why my rebase was going bonkers, hence why I brought it up. Again sorry about that, should have explained it a bit more.

> If the users are all happy for us to develop and release that targets only `nightly` neovim, then yay

Perhaps a branch named `nightly` might be more descriptive? Just sayin' :)

queries/jsonnet/highlights.scm
Duologic: Thanks!! ;)

clason:
No worries; it just came across like a drive-by comment ("I don't know what you're doing, but you shouldn't be doing it"), which we get too many of. Hence the dismissive response.

Duologic:
Let me know if there is anything else I can do.

queries/jsonnet/highlights.scm
```diff
@@ -30,63 +65,100 @@
...
+  ")"
+)
+
+; Emphasize implicit plus usage
+(implicit_plus
+  (_ "}"? @text.danger)
+  (object
+    "{" @text.danger
+  )
+)
```
amaanq, line 161: what even is this?

queries/jsonnet/highlights.scm
```diff
@@ -1,21 +1,56 @@
...
-  last: (id) @field)
-(fieldaccess_super
-  (id) @field)
+[
+  (dollar)
+  (self)
+  (super)
+] @variable.builtin
+((id) @variable.builtin
+ (#eq? @variable.builtin "std"))
```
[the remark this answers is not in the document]
Duologic: I don't see `@namespace.builtin` in the CONTRIBUTING docs, also there is no concept of a 'namespace' in jsonnet.

queries/jsonnet/highlights.scm
Duologic: There are two ways to combine expressions:

A binaryop:
```
expr + { }
```

And an implicit way:

```
expr { }
```

This leads to the same result but from the AST perspective this is not a binaryop but a separate concept (with lack of a better name I called it `implicit_plus` in the parser.

As the implicit behavior should be discouraged, I've included this piece to mark those curly brackets so it becomes apparent something implicit is happening here and people should be aware.

Duologic: Can you point out what's wrong about it?

amaanq: I mean just because it's discouraged by the community doesn't mean text.danger fits it at all, maybe punctuation.special to differentiate it from other brackets

queries/jsonnet/highlights.scm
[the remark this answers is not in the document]
amaanq: you can tack on nested captures that make sense, the first level just has to be officially supported

also, what is std then? it's typically not a "variable"

queries/jsonnet/highlights.scm
Duologic: That doesn't feel expressive enough for me, if y'all don't want this, I'll rip it out and drop it in a plugin (with `after/`).

queries/jsonnet/highlights.scm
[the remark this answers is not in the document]
Duologic: >  ... object implicitly bound to the std variable

Source: https://jsonnet.org/ref/stdlib.html

queries/jsonnet/highlights.scm
amaanq: yeah, of course you yourself can highlight stuff however you like - but as a sane default I'd say remove it vs using text.danger

clason: Treesitter is _not_ a linter: our goal is to correctly highlight _correct_ syntax; highlighting _incorrect_ syntax is a non-goal (we have LSP for that). Please remove this.

queries/jsonnet/highlights.scm
Duologic: I've removed the references part from highlights, it was working but quite hacky, I'll put it in the nvim-jsonnet plugin so we can iterate at a different pace.

queries/jsonnet/highlights.scm
Duologic: It's gone.

Duologic:
Thanks for the thorough review, I think I've covered all concerns.

queries/jsonnet/locals.scm
```diff
@@ -0,0 +1,18 @@
+(parenthesis) @scope
+(anonymous_function) @scope
+(object) @scope
+(field) @scope
+(local_bind) @scope
+
+(field
+  function: (fieldname (id)  @definition.function)
+  (#set! "definition.function.scope" "parent")
+)
```
lucario387, line 10: I feel like you will be pinged by ObserverOfTime to fix the formatting to be lisp-like, so I will ping you first :D

Duologic: Any `fmt` tools you can recommend?

lucario387: no `fmt` tools afaik, mostly just don't let `)` be on new line, rather put all the trailing on the final node

```
; not like this
...
  bar
)))
; but rather like this
 bar)))
```

Duologic: done

lucario387: This is not the only place. I only pointed out one example. Please do so for all files. I can see some in `highlights.scm`, for example

lucario387: @Duologic

lucario387 (approved):

queries/jsonnet/highlights.scm
```diff
@@ -30,63 +70,64 @@
...
-  (self)
-  (super)
-] @variable.builtin
-((id) @variable.builtin
- (#eq? @variable.builtin "std"))
+(fieldname (id) @field)
+(fieldname (string
+             (string_start) @text.strong
+             (string_content) @field
+             (string_end) @text.strong
```
amaanq, line 95: Remove the text.strong here too and just highlight the whole (string) as `@string`

queries/jsonnet/highlights.scm
```diff
@@ -30,63 +70,64 @@
...
-; Function declaration
+; Functions
+(field
+  function: (fieldname (id) @function))
+(field
+  function: (fieldname
+              (string
+                (string_start) @text.strong
+                (string_content) @function
+                (string_end) @text.strong
```
amaanq, line 106: same

queries/jsonnet/highlights.scm
```diff
@@ -30,63 +70,64 @@
...
+              )))
 (param
   identifier: (id) @parameter)
-(field
-  (fieldname (id) @function)
-  [(anonymous_function) (params)])
-(bind
-  function: (id) @function)
+
+(bind (id) @define)
```
amaanq, line 111: remove this

queries/jsonnet/locals.scm
clason: There will be one soon, though, so not necessary to overdo it. Just avoid hanging parens.

queries/jsonnet/highlights.scm
clason: In general, the captures have to be purely syntactic; markup is the job of color schemes.

(I.e., do not use any @text captures unless the node contains text markup.)

Duologic: I'd like to agree, I just think displaying it just as `@string` in the fieldname context is just not expressive enough.

Do you know how the color scheme can make distinction between `string_start` and `string_content` so that I can do this kind of markup?

Duologic: Would keeping `(string_content) @field` be an option? This would allow the color scheme to make a distinction.

amaanq: does a string have a different meaning in this context compared to others? if not - then no, highlight the whole string node as `@string`

no other language differentiates the start/end quotes

Duologic: It is not a regular string, it is a key in an object:

```jsonnet
{
  key1: 'value'
  'key2': 'value',
  'functionname'(arg): arg,
}
```

`key2` has a different meaning than `value` but the same meaning as `key1`, but it is still a string so I'd find it useful to highlight the quotes differently from the content.

See also the highlight of the function below, simply displaying that as a string dismisses that the fact that it is a function name.

I'd love suggestions on how we can do this kind of distinction better.

queries/jsonnet/highlights.scm
Duologic: Considering the `bind` here is a local variable, would `@variable.local` be reasonable?

queries/jsonnet/highlights.scm
amaanq: okay, my bad, I see what you mean now

you can capture the inner string-content as `@field` or `@label` (what json does) 

if you captured string anywhere above as `@string` the quotes will still be highlighted as such.

queries/jsonnet/locals.scm
Duologic: I've checked and fixed, hopefully I haven't missed any.

queries/jsonnet/highlights.scm
amaanq: that's fine but this is more a locals thing

Duologic:
That's it for now I think. Again, thanks for all the feedback.

amaanq (approved):

PR #5300, opened 2023-08-22
feat: add re2c

amaanq:
Needs https://github.com/neovim/neovim/pull/24831

### DIFF
--- a/queries/re2c/indents.scm
+++ b/queries/re2c/indents.scm
@@ -1,0 +1,11 @@
+[
+ (re2c)
+ (action)
+] @indent.begin
+
+"}" @indent.branch @indent.end
+
+[
+ (comment)
+ (linedir)
+] @indent.ignore
--- a/queries/re2c/injections.scm
+++ b/queries/re2c/injections.scm
@@ -1,0 +1,12 @@
+((host_lang) @injection.content
+  (#set! injection.parent))
+
+((define
+   option: (option_name) @_yy
+   value: (dstring) @injection.content)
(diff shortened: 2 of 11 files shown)


queries/re2c/indents.scm
```diff
@@ -0,0 +1,6 @@
+(re2c) @indent
```
lucario387, line 1: auto or smth

queries/re2c/injections.scm
```diff
@@ -0,0 +1,5 @@
+((host_lang) @injection.content
+  (#set! injection.parent))
```
lucario387, line 2: I suppose this is the section that will have `injection.parent` ?

clason:
You could PR to main, which targets Nvim HEAD?

amaanq:
Yeah that'd be fine, though master could benefit from re2c as well except for host_code nodes not having proper parent injections

queries/re2c/injections.scm
```diff
@@ -0,0 +1,12 @@
...
+
+((define
+   option: (option_name) @_yy
+   value: (dstring) @injection.content)
+  (#any-of? @_yy "YYPEEK" "YYSKIP" "YYBACKUP" "YYBACKUPCTX" "YYRESTORE" "YYRESTORECTX" "YYFILL" "YYSHIFT")
+  (#offset! @injection.content 0 1 0 -1)
+  (#set! injection.parent))
```
lucario387, line 9: missing the language or the parent here

ObserverOfTime:
But it's broken on master.

amaanq:
> But it's going to be broken for anyone not using HEAD.

would the query not silently fail? It's not like the query itself is invalid, just the capture

amaanq:
the relevant PR has landed, this is good to go now

lucario387 (approved):

PR #5306, opened 2023-08-23
feat(bash): improve highlights

ObserverOfTime:
Waiting for tree-sitter/tree-sitter-bash#193 to land.

### DIFF
--- a/queries/bash/highlights.scm
+++ b/queries/bash/highlights.scm
@@ -1,36 +1,21 @@
-(simple_expansion) @none
-(expansion
-  "${" @punctuation.special
-  "}" @punctuation.special) @none
 [
- "("
- ")"
- "(("
- "))"
- "{"
- "}"
- "["
- "]"
- "[["
- "]]"
- ] @punctuation.bracket
+ "(" ")"
+ "{" "}"
+ "[" "]"
+ "[[" "]]"
+] @punctuation.bracket
 
 [
  ";"
  ";;"
- (heredoc_start)
- ] @punctuation.delimiter
-
-[
- "$"
-] @punctuation.special
+ "&"
+] @punctuation.delimiter
 
 [
  ">"
  ">>"
  "<"
  "<<"
- "&"
  "&&"
  "|"
  "||"
@@ -38,7 +23,7 @@
  "=~"
  "=="
  "!="
- ] @operator
+] @operator
 
 ; Do *not* spell check strings since they typically have some sort of
 ; interpolation in them, or, are typically used for things like filenames, URLs,
@@ -48,7 +33,13 @@
  (raw_string)
  (ansi_c_string)
  (heredoc_body)
+ (simple_heredoc_body)
 ] @string
+
+[
+ (heredoc_start)
+ (heredoc_end)
+] @label
 
 (variable_assignment (word) @string)
 
@@ -61,7 +52,7 @@
  "case"
  "in"
  "esac"
- ] @conditional
+] @conditional
 
 [
  "for"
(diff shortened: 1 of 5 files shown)


queries/bash/highlights.scm
```diff
@@ -89,17 +81,23 @@
...
 (comment) @comment @spell
-(test_operator) @string
 
-(command_substitution
-  [ "$(" ")" ] @punctuation.bracket)
+(test_operator) @operator
+
+(command_substitution "$(" @punctuation.bracket)
+(process_substitution "<(" @punctuation.bracket)
+(arithmetic_expansion "$((" @punctuation.bracket)
```
amaanq, line 92: hm, not special?

queries/bash/highlights.scm
```diff
@@ -1,29 +1,15 @@
...
 [
  ";"
  ";;"
- (heredoc_start)
- ] @punctuation.delimiter
-
-[
- "$"
-] @punctuation.special
+] @punctuation.delimiter
```
amaanq, line 12: add '&' as well

queries/bash/injections.scm
```diff
@@ -1,11 +1,13 @@
...
- (heredoc_body) @injection.content
- (#offset! @injection.content 0 0 -1 0)
- (#downcase! @injection.language))(#set! injection.language "regex"))
+((heredoc_redirect
+  [
+    (heredoc_body)
+    (simple_heredoc_body)
+  ] @injection.content
+  (heredoc_end) @injection.language)
+ (#downcase! @injection.language))
```
amaanq, line 13: do we want to support injecting when ${THIS_SH} is detected?

https://git.savannah.gnu.org/tree/tests/arith-for.tests?id=ec8113b9861375e4e17b3307372569d429dec814#n91

queries/bash/highlights.scm
ObserverOfTime: I kept it the same as it was and this way it matches the closing bracket.

amaanq: you can make the matching closing bracket in *this* case special, it'd be most useful for arith expansions imo

queries/bash/injections.scm
ObserverOfTime: That link isn't working.

queries/bash/highlights.scm
ObserverOfTime: What about "(("?

amaanq:
```scm
(arithmetic_expansion [ "$((" "((" ] @punctuation.special "))" @punctuation.special)
```

queries/bash/injections.scm
amaanq: updated

ObserverOfTime: There are too many cases to cover.

```bash
${THIS_SH} -c '...'
sh -c '...'; /bin/sh -c '...'; /usr/bin/sh -c '...'
bash -c '...'; /bin/bash -c '...'; /usr/bin/bash -c '...'
command sh -c '...' # plus path variants
command bash -c '...' # plus path variants
command -p sh -c '...' # plus path variants
command -p bash -c '...' # plus path variants
env sh -c '...' # plus path variants, plus env options
env bash -c '...' # plus path variants, plus env options
```

amaanq: hm yeah, nevermind

amaanq (approved):

amaanq:
sorta related, it'd be nice to match heredoc injections w/ the language ft/shorthand name too

e.g.
```bash
node <<JS
console.log("hi")
JS
```

not the end of the world

PR #5307, opened 2023-08-23
refactor: replace some match & vim-match usages

ObserverOfTime:
Also, Ruby currently highlights all `(constant)` nodes as `@type`.
I made it default to `@constant` and have `@type` *not* match `^[A-Z0-9_]+$`.

### DIFF
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -101,9 +101,9 @@
 
 ; Interop
 ((sym_lit) @method
- (#match? @method "^\\.[^-]"))
+ (#lua-match? @method "^[.][^-]"))
 ((sym_lit) @field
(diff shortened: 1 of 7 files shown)


clason (approved):

amaanq (approved):

NoahTheDuke:
This seems to have broken something in the Clojure highlighting. Now every symbol is marked as a field (in addition to the other matches). If I manually replace the new `@field` back to the old `(#match? @field "^\\.-")` impl, everything goes back to normal. I don't know enough about `#lua_match` to suggest an alternative, sadly.

amaanq:
Not a soln but you can learn about lua patterns here https://www.lua.org/pil/20.2.html

PR #5310, opened 2023-08-24
feat(injections): add printf format strings

ObserverOfTime:
Replaces #3573. I forked the parser so it's no longer limited to double-quoted strings.

### DIFF
--- a/runtime/queries/c/injections.scm
+++ b/runtime/queries/c/injections.scm
@@ -4,6 +4,73 @@
 ((comment) @injection.content
  (#set! injection.language "comment"))
 
+((call_expression
+  function: (identifier) @_function
+  arguments: (argument_list
+               . (string_literal (string_content) @injection.content)))
+ (#any-of? @_function "printf" "printf_s"
+                      "vprintf" "vprintf_s"
+                      "scanf" "scanf_s"
+                      "vscanf" "vscanf_s"
+                      "wprintf" "wprintf_s"
+                      "vwprintf" "vwprintf_s"
+                      "wscanf" "wscanf_s"
+                      "vwscanf" "vwscanf_s"
+                      "cscanf" "_cscanf"
+                      "printw"
+                      "scanw")
+ (#set! injection.language "printf"))
+
+((call_expression
+  function: (identifier) @_function
+  arguments: (argument_list
+               (_) . (string_literal (string_content) @injection.content)))
+ (#any-of? @_function "fprintf" "fprintf_s"
+                      "sprintf"
+                      "dprintf"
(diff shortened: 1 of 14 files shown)


runtime/queries/c/injections.scm
```diff
@@ -4,6 +4,28 @@
...
+ (#any-of? @_function "fprintf" "fscanf" "fprintf_s" "fscanf_s"
+                      "sprintf" "sscanf" "sscanf_s")
+ (#set! injection.language "printf"))
+
+((call_expression
+  function: (identifier) @_function
+  arguments: (argument_list
+               (_) . (_) . (string_literal (string_content) @injection.content)))
+ (#any-of? @_function "snprintf" "sprintf_s" "snprintf_s")
+ (#set! injection.language "printf"))
```
amaanq, line 27: there are a lot more than these

`man 3 printf` 

Missing:
dprintf (2nd arg)
vprintf (1st)
vfprintf (2nd)
vdprintf (2nd)
vsprintf (2nd)
vsnprintf (3rd)

runtime/queries/printf/highlights.scm
```diff
@@ -0,0 +1 @@
+(format) @character
```
amaanq, line 1: `@character.special`?

ObserverOfTime: I kinda want to differentiate format strings from other special characters.

amaanq: I'd like that too, but I feel character and string are somewhat oftenly used as the same color in colorschemes

A custom capture wouldn't be bad imo, like `@character.format`

ObserverOfTime: A custom capture would be pointless since `@character.printf` can be used anyway.

amaanq: Right, this is fine then

runtime/queries/c/injections.scm
ObserverOfTime: Just gonna pretend MS CRT is not a thing.

amaanq:
I was gonna jot more down but forgot, this has a pretty good list

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/vsprintf-vsprintf-l-vswprintf-vswprintf-l-vswprintf-l?view=msvc-170

amaanq (approved):

amaanq:
Added a few more missing ones, merge if you think it's good

clason:
Was it intentional to only add this to the `main` branch?

PR #5315, opened 2023-08-25
feat: add Doxygen

amaanq:


queries/c/injections.scm
```diff
@@ -8,6 +8,10 @@
   (#match? @injection.content "/\*!([a-zA-Z]+:)?re2c")
   (#set! injection.language "re2c"))
 
+((comment) @injection.content
+  (#match? @injection.content "/\*(!|<|\*)[^\w]")
```
lucario387, line 12: Shouldn't all of these patterns be `^pattern`

amaanq: /** /*! and /*< signify a potential doxygen blob

I added a not word after to avoid double injecting w/ re2c

lucario387, line 12: Also, shouldn't all of these be escaped twice i.e. `\\w`

lucario387 (approved):

amaanq:
Seems like it just needs indent.auto to play nicely

ObserverOfTime (commented):
Can you also highlight [links and references](https://www.doxygen.nl/manual/autolink.html)?

amaanq:
Done, though w/o self.parent to inject the function params the entire blob will be highlighted as function

ObserverOfTime:
nvim crashes when I open a cpp file with the doxygen parser installed. :thinking:

amaanq:
Hm, fuzzing shows nothing - can you provide a reproducer and potentially a backtrace? Maybe it's certain injections inside doxygen

ObserverOfTime:
Can you reproduce with this?
```cpp
#ifdef NDEBUG

#endif  // NDEBUG
```

amaanq:
I cannot

ObserverOfTime:
Then it's probably one of my plugins. :thinking:

amaanq:
does it crash just *opening* a file w/ that text or *after* you make some edits?

(leading to me wanting to add a fuzzer for edits feature upstream even more so if the latter..)

ObserverOfTime:
> does it crash just opening a file w/ that text or after you make some edits?

Both after opening a file w/ that text and after writing one (I think it's the comment specifically).

ObserverOfTime:
It crashes even with `--clean -u scripts/minimal_init.lua` so it's not a plugin that causes it.
That leaves either another parser or some other config file. :thinking:

ObserverOfTime:
Interesting. It only crashes when I write a word character in the comment.

amaanq:
maybe it's the comment parser if you use it? a gdb/lldb backtrace could be really insightful here

ObserverOfTime:
Works after escaping the asterisks twice. Wasn't my issue after all. :)

amaanq:
In the match? It's still really weird behavior though

amaanq:
Hmm, is it good now @ObserverOfTime?

ObserverOfTime:
Unfortunately not… I'm encountering another crash.

ObserverOfTime:
Write this in a cpp file and press enter anywhere after the `!`. 
```cpp
/*! \brief Help */
```

amaanq:
doesn't crash for me

ObserverOfTime:
This is very weird.

ObserverOfTime:
I'll leave it for tomorrow…

ObserverOfTime:
Here's the backtrace. I don't know if it helps.
image

amaanq:
Yeah that's very helpful, and wow, ironic how someone else reported that same bug and I fixed it as well

https://github.com/tree-sitter/tree-sitter/issues/2563

https://github.com/tree-sitter/tree-sitter/pull/2566

If possible - can you set the ts dep for neovim to my PR's branch and try it out then?

amaanq:
I see a potential spot where this can occur in the scanner too, I've push an update, can you try that now?

ObserverOfTime:
It's fixed after `!` but pressing enter before the closing `*/` still causes it to crash.

Also, you can change the matches to this (since re2c only matches letters):
```scheme
(#lua-match? @injection.content "/[*][!<*][^a-zA-Z]")
```

amaanq:
> It's fixed after `!` but pressing enter before the closing `*/` still causes it to crash.

Found another spot, but did you also build neovim w/ my ts fork?

ObserverOfTime:
> Found another spot, but did you also build neovim w/ my ts fork? (It'd be nice to confirm if that fixes it)

That does seem to fix it.

amaanq:
The latest doxygen fix (2nd one) or my ts pr?

ObserverOfTime:
Both.

amaanq:
awesome, thanks

ObserverOfTime:
The `tag_name` is sometimes not highlighted. Might need a higher priority.

```scheme
((tag_name) @keyword
 (#set! "priority" 101))
```

amaanq:
I notice that too - adding any item below it (like a function or something) fixes it. I'd want to say it's not a bug w/ priorities

ObserverOfTime:
Maybe, but the priority fixes it for me.

amaanq:
Just tag name, or all of them?

ObserverOfTime:
The others seem fine.

ObserverOfTime (approved):
Finally…

amaanq:
the joy of debugging tree-sitter

PR #5319, opened 2023-08-26
[clojure] Fix @field highlighting bug

NoahTheDuke:
Fixes error introduced in #5307.

Add tests for Clojure highlighting.

### DIFF
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -63,7 +63,7 @@
 
 ; Constructor
 ((sym_lit) @constructor
- (#lua-match? @constructor "^-\\>[^\\>].*"))
+ (#lua-match? @constructor "^-%>[^>].*"))
 
 ; Builtin dynamic variables
 ((sym_lit) @variable.builtin
@@ -100,11 +100,17 @@
  (#lua-match? @type "^[^/]+[.][^/]*$"))
 
 ; Interop
-((sym_lit) @method
- (#lua-match? @method "^[.][^-]"))
-((sym_lit) @field
- (#lua-match? @field "^%.-"))
-((sym_lit) @field
+(list_lit
+ .
+ (sym_lit) @method
+ (#lua-match? @method "^%.[^-]"))
+(list_lit
+ .
+ (sym_lit) @field
+ (#lua-match? @field "^%.%-.*"))
+(list_lit
+ .
+ (sym_lit) @field
  (#lua-match? @field "^[%u].*/.+"))
 (list_lit
  .
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -100,18 +100,17 @@
  (#lua-match? @type "^[^/]+[.][^/]*$"))
 
 ; Interop
-(list_lit
- .
- (sym_lit) @method
+; (.instanceMember instance args*)
+; (.instanceMember Classname args*)
+((sym_lit) @method
  (#lua-match? @method "^%.[^-]"))
-(list_lit
- .
- (sym_lit) @field
+; (.-instanceField instance)
+((sym_lit) @field
  (#lua-match? @field "^%.%-.*"))
-(list_lit
- .
- (sym_lit) @field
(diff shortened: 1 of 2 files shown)


queries/clojure/highlights.scm
```diff
@@ -103,7 +103,7 @@
 ((sym_lit) @method
  (#lua-match? @method "^[.][^-]"))
 ((sym_lit) @field
- (#lua-match? @field "^[.]-"))
+ (#lua-match? @field "^\\.-"))
```
lucario387, line 106: What about %. instead of \\., since it's lua-match

NoahTheDuke: Sure, I'll try that out.

lucario387:
```
Fail	||	highlight queries tests/query/highlights/clojure/test.clj	
            ...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:96: Error in at tests/query/highlights/clojure/test.clj:10:1: expected "keyword", captures: { "symbol" }", highlights: { "symbol" }
            Expected objects to be the same.
            Passed in:
            (nil)
            Expected:
            (boolean) true
            
            stack traceback:
            	...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:96: in function 'check_assertions'
            	...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:119: in function <...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:118>
```
Seems like something's wrong here

NoahTheDuke:
Indeed. It passes locally so I'm unsure how to approach a solution. Any advice?

lucario387:
> Indeed. It passes locally so I'm unsure how to approach a solution. Any advice?

I'm not sure how it passed locally on your end. Mine doesn't pass as well

Having `:Inspect` while the cursor at `clojure/test:clj:10:1` yields this

```
Treesitter                                                                                                                                                                                      
  - @symbol.clojure links to @symbol clojure    
```

lucario387:
The node that fails, has its name being `kwd_lit`, so not changed in this PR. You probably applied the wrong test for it

NoahTheDuke:
Turns out, for some reason, my custom `after` queries were being applied, so that's my fault. Hopefully I've gotten it right locally so the tests will now pass.

queries/clojure/highlights.scm
```diff
@@ -100,11 +100,17 @@
...
 
 ; Interop
-((sym_lit) @method
- (#lua-match? @method "^[.][^-]"))
-((sym_lit) @field
- (#lua-match? @field "^%.-"))
-((sym_lit) @field
+(list_lit
+ .
+ (sym_lit) @method
```
lucario387, line 105: Can I get some docs on how these methods/fields in clojure work?

And why is it necessary to limit it to the only the first `sym_lit` of each list

tests/query/highlights/clojure/test.clj
lucario387, line 1: Please don't add these many tests, only nodes that were relevant to this PR. No need to test for punctuations, brackets and such, so is include, string, variable ...

NoahTheDuke: Sure, I'll trim it down.

queries/clojure/highlights.scm
NoahTheDuke: I'll remove this change, put it in a different PR.

tests/query/highlights/clojure/test.clj
NoahTheDuke: If you don't mind, why remove those tests?

lucario387: Couple of reasons
- Highlight tests here should be used to reflect the specific differences between nvim-treesitter and the upstream parser creator. If there's no differences then there's no need for those tests as it's already tested from upstream.
- Tests can also be used to detect regressions, so tests for some changes in cases like this is nice, but only for those regressions

lucario387 (approved):

ObserverOfTime (approved):

PR #5321, opened 2023-08-26
fix(bash): update queries from upstream

amaanq:


lucario387:
And a lockfile update too

amaanq:
ci can take care of that

lucario387 (approved):

PR #5333, opened 2023-08-30
feat(bash): update parser and queries

amaanq:
ref https://github.com/tree-sitter/tree-sitter-bash/pull/211 :woozy_face: 

since string_content and heredoc_content are exposed, maybe we want to just apply `@string` to those


clason (approved):

ObserverOfTime:
How does `heredoc_content` differ from `heredoc_body`?

amaanq:
Those are in heredocs with expansions, the blobs of content in between them, vs just being a pure body blick

PR #5340, opened 2023-08-31
Feat: Add unison

tapegram:
Attempting to add unison

Appreciate this tool thanks for the help!

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1747,7 +1747,7 @@ list.unison = {
   install_info = {
     url = "https://github.com/kylegoetz/tree-sitter-unison",
     files = { "src/parser.c", "src/scanner.c", "src/maybe.c" },
-    branch = "main",
+    branch = "main"
   },
   maintainers = { "@tapegram" },
 }
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1747,7 +1747,7 @@ list.unison = {
   install_info = {
     url = "https://github.com/kylegoetz/tree-sitter-unison",
     files = { "src/parser.c", "src/scanner.c", "src/maybe.c" },
-    branch = "main"
+    branch = "main",
   },
   maintainers = { "@tapegram" },
 }
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1746,7 +1746,7 @@ list.ungrammar = {
 list.unison = {
   install_info = {
     url = "https://github.com/kylegoetz/tree-sitter-unison",
-    files = { "src/parser.c", "src/scanner.c", "src/maybe.c" },
+    files = { "src/parser.c", "src/scanner.c" },
     requires_generate_from_grammar = true,
   },
   maintainers = { "@tapegram" },
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1747,7 +1747,6 @@ list.unison = {
   install_info = {
     url = "https://github.com/kylegoetz/tree-sitter-unison",
     files = { "src/parser.c", "src/scanner.c", "src/maybe.c" },
-    branch = "main",
   },
   maintainers = { "@tapegram" },
 }
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1743,6 +1743,15 @@ list.ungrammar = {
   maintainers = { "@Philipp-M", "@amaanq" },
 }
 
+list.unison = {
+  install_info = {
+    url = "https://github.com/kylegoetz/tree-sitter-unison",
+    files = { "src/parser.c", "src/scanner.c", "src/maybe.c" },
+    branch = "main",
+  },
+  maintainers = { "@tapegram" },
+}
+
 list.usd = {
   install_info = {
     url = "https://github.com/ColinKennedy/tree-sitter-usd",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1747,6 +1747,7 @@ list.unison = {
   install_info = {
     url = "https://github.com/kylegoetz/tree-sitter-unison",
     files = { "src/parser.c", "src/scanner.c", "src/maybe.c" },
+    requires_generate_from_grammar = true,
   },
   maintainers = { "@tapegram" },
 }
--- a/queries/unison/highlights.scm
+++ b/queries/unison/highlights.scm
@@ -1,0 +1,75 @@
+;; Primitives
+(comment) @comment
+(nat) @constant.numeric
+(unit) @constant.builtin
+(literal_char) @constant.character
+(literal_text) @string
+(literal_boolean) @constant.builtin.boolean
+
+;; Keywords
+[
+  (kw_forall)
+  (unique_kw)
+  (type_kw)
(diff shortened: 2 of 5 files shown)


lucario387:
Please follow the captures listed in contributing.md

queries/unison/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+  (or)
+] @operator
+
+[
+  "if"
+  "else"
+  "then"
+  (match)
+  (with)
+  (cases)
```
lucario387, line 43: Can't the match, with, cases string themselves be used instead of nodes?

queries/unison/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+] @keyword.control.conditional
+
+(blank_pattern) @variable.builtin
+
+((pattern) @variable.other.member (constructor_or_variable_pattern) @type)
+
+(use_clause) @keyword.control.import
+
+;; Types
+(record_field name: (wordy_id) @variable.other.member type: (wordy_id) @type)
```
lucario387, line 53: I will take this as an example of following `CONTRIBUTING.md`

Only `@variable`, not the suffix .other.member

queries/unison/highlights.scm
tapegram: I tried with these as strings and it breaks, though I'm having trouble seeing the details of the error other than it just gets angry.

lua/nvim-treesitter/parsers.lua
```diff
@@ -1743,6 +1743,15 @@ list.ungrammar = {
   maintainers = { "@Philipp-M", "@amaanq" },
 }
 
+list.unison = {
+  install_info = {
+    url = "https://github.com/kylegoetz/tree-sitter-unison",
+    files = { "src/parser.c", "src/scanner.c", "src/maybe.c" },
+    requires_generate_from_grammar = true,
```
amaanq, line 1750: If you have parser.c requires_generate is not needed

Also, do not include maybe.c, if scanner.c includes it that's fine, but that's a really weird construct anyways

queries/unison/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+(literal_boolean) @boolean
+
+;; Keywords
+[
+  (kw_forall)
+  (unique_kw)
+  (type_kw)
+  (kw_equals)
+  (do)
+] @keyword
```
amaanq, line 16: forall and do sound like `@repeat`

equals sounds like `@keyword.operator`

queries/unison/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+
+(kw_let) @keyword.function
```
amaanq, line 18: does let declare *only* functions or a variable/other stuff?

queries/unison/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+(structural_kw) @storageclass
+(unique) @storageclass
+
+[ 
+  (type_constructor)
+] @constructor
```
amaanq, line 24: unnecessary brackets

queries/unison/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+
+(doc_block) @comment
```
amaanq, line 26:
```suggestion
(doc_block) @comment.documentation
```

queries/unison/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+
+
+[
+  (operator)
+  (pipe) 
+  (arrow_symbol) 
+  (">")
+  (or)
```
amaanq, line 34: or might be `@keyword.operator`

queries/unison/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+  "else"
+  "then"
+  (match)
+  (with)
+  (cases)
+] @conditional
+
+(blank_pattern) @variable.builtin
+
+((pattern) @variable (constructor_or_variable_pattern) @type)
```
amaanq, line 48: not sure what the wrapped parenthesis indicates, there's no parent node to these two

amaanq (changes requested):

queries/unison/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+  (kw_forall)
+  (unique_kw)
+  (type_kw)
+  (kw_equals)
+  (do)
+] @keyword
+
+(kw_let) @keyword.function
+(structural_kw) @storageclass
+(unique) @storageclass
```
ObserverOfTime, line 20: `@type.qualifier` if they don't affect memory.

ObserverOfTime (changes requested):

queries/unison/highlights.scm
tapegram: let opens a "block" wherever an expression is expected https://www.unison-lang.org/learn/language-reference/blocks-and-statements/

(though you assign variables/functions inside blocks, but that muddies the water a bit because I think *everything* is a function in unison).

Not sure what the best fit is though for capture.

queries/unison/highlights.scm
tapegram: `forall` - is a type level thing in unison (https://www.unison-lang.org/learn/language-reference/types/, https://www.unison-lang.org/learn/language-reference/scoped-type-variables/)
`do` - introduces a delayed computation block https://www.unison-lang.org/learn/language-reference/delayed-computations

I think you are right about equals, will update.

queries/unison/highlights.scm
tapegram: Going to move it into the @keyword block instead of keyword.function if that makes sense

queries/unison/highlights.scm
lucario387: Can these kind of patterns deduct consecutive nodes, or will they always be resolved to 2 separate query captures

tapegram: Removed the outer parens from this. Also tried the outer parens on the next line `(record_field name: (wordy_id) @variable type: (wordy_id) @type)` but that caused things to break.

queries/unison/highlights.scm
```diff
@@ -0,0 +1,75 @@
+;; Primitives
+(comment) @comment
```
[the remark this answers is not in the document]
ObserverOfTime: Still missing the injection.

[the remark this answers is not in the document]
tapegram: my b, just added

queries/unison/highlights.scm
amaanq: yeah that's good

queries/unison/highlights.scm
amaanq: Thanks, not familiar w/ the language so I just go off of preconceived notions about certain keywords I see often :)

queries/unison/highlights.scm
amaanq: Ah I see, my bad. Were you intending to match two items at the root level? Using the first/root rule as the parent would be much clearer then if so, otherwise if you meant for this to occur anywhere it'd be interesting to know the use-case/where it's used, I've never seen that kind of construct not wrapped in a parent rule of sorts except for immediate following matches w/ anchors (like several comments together)

amaanq: > Can these kind of patterns deduct consecutive nodes, or will they always be resolved to 2 separate query captures

Consecutive nodes allowing anything in-between as long as they're at the same level (top level, inside a definition of sorts, etc)

ObserverOfTime (approved):

lucario387 (approved):

amaanq (approved):

PR #5344, opened 2023-09-02
feat(c_sharp): add indent query

vlada-dudr:
I have been using it for some months now and seems to mostly work ok.

### DIFF
--- a/queries/c_sharp/indents.scm
+++ b/queries/c_sharp/indents.scm
@@ -1,0 +1,117 @@
+[
+ (property_declaration)
+ ; (field_declaration)
+ (variable_declaration)
+ (accessor_declaration)
+
+ ; (declaration_list)
+
+ (parameter_list)
+ (type_parameter_list)
+ (accessor_list)
+ (attribute_argument_list)
+ (base_list)
+ (argument_list)
+
+ (binary_expression)
+ (expression_statement)
+ (return_statement)
+ (local_declaration_statement)
+ (local_function_statement)
+ (yield_statement)
+ (fixed_statement)
+ (unsafe_statement)
+ (labeled_statement)
+ (for_each_statement)
+ (for_statement)
+ (if_statement)
+ (switch_expression)
+ (switch_section)
+ (switch_statement)
+ (while_statement)
+ (lambda_expression)
+ (do_statement)
+ (checked_statement)
+ (try_statement)
+ (using_statement)
+ (local_function_statement)
+ (object_creation_expression)
+ (array_creation_expression)
+ (implicit_object_creation_expression)
+
+ ] @indent.begin
+
+(method_declaration body: _ @indent.begin)
+(namespace_declaration body: _ @indent.begin)
+(record_declaration body: _ @indent.begin)
+(class_declaration body: _ @indent.begin)
+(struct_declaration body: _ @indent.begin)
+(constructor_declaration body: _ @indent.begin)
+(interface_declaration body: _ @indent.begin)
+(enum_declaration body: _ @indent.begin)
+
+(property_declaration (accessor_list "{" @indent.branch) @indent.dedent)
+
+(switch_section
+  (case_switch_label) (case_switch_label) @indent.branch)
+
+
+(if_statement (block "{" @indent.branch))
+(if_statement alternative: (if_statement) @indent.dedent)
+
+(while_statement
+  (block "{" @indent.branch))
+
+(array_creation_expression
+  (initializer_expression "{" @indent.branch))
+(object_creation_expression
+  (initializer_expression "{" @indent.branch))
+(implicit_object_creation_expression
+  (initializer_expression "{" @indent.branch))
+
+(for_each_statement 
+  body: (block "{" @indent.branch))
+
+(lambda_expression 
+  body: (block "{" @indent.branch))
+
+(for_statement 
+  body: (block "{" @indent.branch))
+
+(switch_expression "{" @indent.branch)
+
+(switch_statement
+  (switch_body "{" @indent.branch))
+
+(try_statement
+  body: (block "{" @indent.branch))
+(catch_clause
+  body: (block "{" @indent.branch))
+(finally_clause
+  (block "{" @indent.branch))
+
+(fixed_statement 
+  (block "{" @indent.branch))
+
+(using_statement 
+  body: (block "{" @indent.branch))
+
+["catch" "finally" "else" "}" "]" ")" ] @indent.branch
+
+["}" ")" "]" ] @indent.end
+
+(expression_statement ";" @indent.end)
+
+(local_function_statement ";" @indent.end)
+(local_declaration_statement ";" @indent.end)
+
+; this is fucked up, because grammer doesn't match # as part of the token
+[
+ (if_directive)
+ (else_directive)
+ (elif_directive)
+ (endif_directive)
+ (nullable_directive)
+ (pragma_directive)
+ (undef_directive)
+ ] @indent.zero


queries/c_sharp/indents.scm
```diff
@@ -0,0 +1,117 @@
...
+ ] @indent.begin
+
+(method_declaration body: _ @indent.begin)
+(namespace_declaration body: _ @indent.begin)
+(record_declaration body: _ @indent.begin)
+(class_declaration body: _ @indent.begin)
+(struct_declaration body: _ @indent.begin)
+(constructor_declaration body: _ @indent.begin)
+(interface_declaration body: _ @indent.begin)
+(enum_declaration body: _ @indent.begin)
```
lucario387, line 51: Why these and not `(block) @indent.begin`

queries/c_sharp/indents.scm
```diff
@@ -0,0 +1,117 @@
...
+
+(property_declaration (accessor_list "{" @indent.branch) @indent.dedent)
+
+(switch_section
+  (case_switch_label) (case_switch_label) @indent.branch)
```
lucario387, line 56: Is there any reason why only switch sections with 2 cases have branches, and only the second one is auto-dedented

queries/c_sharp/indents.scm
```diff
@@ -0,0 +1,117 @@
...
+(catch_clause
+  body: (block "{" @indent.branch))
+(finally_clause
+  (block "{" @indent.branch))
+
+(fixed_statement 
+  (block "{" @indent.branch))
+
+(using_statement 
+  body: (block "{" @indent.branch))
```
lucario387, line 97: I can see a lot of `(parent body: (block ... @capture))` here. Can it be replaced with only `(block ...)`

queries/c_sharp/indents.scm
```diff
@@ -0,0 +1,117 @@
...
+ (while_statement)
+ (lambda_expression)
+ (do_statement)
+ (checked_statement)
+ (try_statement)
+ (using_statement)
+ (local_function_statement)
+ (object_creation_expression)
+ (array_creation_expression)
+ (implicit_object_creation_expression)
```
lucario387, line 40: Imo, Some of these, from looking at the AST, can be replaced with a single `(block)` capture,

vlada-dudr: I did the whole thing incrementally over time, so i don't remember exactly all the reasons for everything. I did have the idea to use `block` captures, but it didn't work for some reason.

queries/c_sharp/indents.scm
lucario387: Personal guess: You need this because you capture the parent node with `@indent.begin`, which is probably not needed.

queries/c_sharp/indents.scm
vlada-dudr: Because:
```csharp
switch (var) {
    case "a": foo(); break;
}
```
is one `switch_section` starting on `case:` line. So `(switch_section) @indent.begin` start indenting on next line.

 ```csharp
switch (var) {
    case "b":
    case "a": foo(); break;
}
```
is also one `switch_section` with two `case_switch_label`s so if i don't deden't the next ones I end up with second case line indented too much.

File: queries/c_sharp/indents.scm [hunk 5]
```diff
@@ -0,0 +1,117 @@
...
+(enum_declaration body: _ @indent.begin)
+
+(property_declaration (accessor_list "{" @indent.branch) @indent.dedent)
+
+(switch_section
+  (case_switch_label) (case_switch_label) @indent.branch)
+
+
+(if_statement (block "{" @indent.branch))
+(if_statement alternative: (if_statement) @indent.dedent)
```
lucario387, line 60: The if_statement should be similar enough to C that I suggest you can use the C's indent queries version, with nodes modification (i.e. `(compound_statement) -> (block))

queries/c_sharp/indents.scm
vlada-dudr: I wish I remember the reason. But I clearly do remember that I couldn't make it work properly via `(block) @indent.begin`.

queries/c_sharp/indents.scm
lucario387: Couldn't it be something like

```
(switch_body) @indent.begin
"case" @indent.branch
```

queries/c_sharp/indents.scm
vlada-dudr: I remember I was looking onto C's version, but I understood only like half of it, so i did this, which I undestand. I think there is subtle difference in tree for C and C#. I generally had hard time to debug the thing so I ended up with kind of first working solution.

queries/c_sharp/indents.scm
lucario387: At its current state, I fear this is too messy for me, you, future you and others to maintain this.

If you think it's fine for you to dedicate some hours if issues pop up, and that the queries doesn't make editing C# feel sluggish, then I'm okay with it

queries/c_sharp/indents.scm
vlada-dudr: wouldn't it make it like this?
```csharp
switch (var) {
case "b":
case "a": foo(); break;
}
```
Thus without indenting the cases at all?

lucario387: it's our current version for C.

If that style is not a standard for C# then you can ignore this.

queries/c_sharp/indents.scm
vlada-dudr: hm, I will try what happens

queries/c_sharp/indents.scm
lucario387: Yours is fine :D. I don't think this case_switch is an issue. I have done the same for swift, C, and ocaml and haven't received complaints (yet)

queries/c_sharp/indents.scm
vlada-dudr: I see why I might do it this way. You may have most of these conditionals and loops not being followed by block, but by plain statement (or expression).

I had the idea to make it something like `(_ body: (block "{" @indent.branch))` to avoid repeating everything with this schema, but I couldn't figure out proper syntax or make it work.

queries/c_sharp/indents.scm
lucario387: Okay I can see better why this is declared as such. I'd still prefer to capture the node associated to `body` field instead

vlada-dudr: yeah, classes, structs and similar have `declaration_list` in their bodies, but e.g. methods and constructors can have `block` or `arrow_expression_clause`

queries/c_sharp/indents.scm
vlada-dudr: It is because of indenting inside of those statements, like:
```csharp
if (
     with()
     && many()
     && conditions()
)
{
    foo();
}
```
which seems to be broken also, as now my query will extra indent at `&&` lines.

lucario387: Ok, so another language like dart, swift and friends where the conditions of the if statement isn't a node :frowning_face: 

Consider this done

lucario387 (approved):
Edit: Consider this obsolete.

queries/c_sharp/indents.scm
```diff
@@ -0,0 +1,117 @@
...
+ (return_statement)
+ (local_declaration_statement)
+ (local_function_statement)
+ (yield_statement)
+ (fixed_statement)
+ (unsafe_statement)
+ (labeled_statement)
+ (for_each_statement)
+ (for_statement)
+ (if_statement)
```
lucario387, line 27: Using your example codeblock, I got this result with the if statement
```cs
if (
    with()
    && many()
    && conditions()
   )
{
  foo();
}
```
I don't think the ending `)` should be there

lucario387 (changes requested):

queries/c_sharp/indents.scm
vlada-dudr: hm, weird I am getting this:

vlada-dudr:
Can you point, how to run the tests? I could probably add some of them to make it easier to debug. I was looking into `CONTRIBUTING` but didn't see any pointer. And the script in `scrips/run_test.sh` expects presence of some non existing files...

lucario387:
> Can you point, how to run the tests

For this PR, you'll need to add example c_sharp files to `tests/indent/c_sharp/`, create that dir if it doesn't exist.

For running those file, you also need to create a `tests/indent/c_sharp_spec.lua`. The contents of that file can be similar to other `*_spec.lua` file in the same directory

vlada-dudr:
That's the part which I understood, but I have no clue how to run the thing

lucario387:
oh, just be inside the git repo, and do a `./scripts/run_tests.sh indent/c_sharp_spec.lua` and see the result

clason:
No, you also need to have plenary installed and in your standard runtimepath.

(I've said it before; I'll say it again: the test infrastructure is a pain point and needs to be streamlined.)

vlada-dudr:
Thanks @clason, that's the missing part - I expected that test machinery would grab it itself, if it needs it

queries/c_sharp/indents.scm
```diff
@@ -0,0 +1,117 @@
...
+; this is fucked up, because grammer doesn't match # as part of the token
+[
+ (if_directive)
+ (else_directive)
+ (elif_directive)
+ (endif_directive)
+ (nullable_directive)
+ (pragma_directive)
+ (undef_directive)
+ ] @indent.zero
```
theHamsta, line 117: Can you add some tests?

vlada-dudr: Yeah, I started writing something. But I probably don't get how it is supposed to work. I managed to start the test. I wrote some example files. But no matter what I put in those files, `whole_file` always passes. I'd expect it to fail when the file isn't indented in harmony with the query. `newline` fails sometimes, but obviously puts different indetation compared to `==` on such line. So, is it for testing opening new line and indentation triggers? Or just one of them?

queries/c_sharp/indents.scm
lucario387: Idk why but it's working like yours on my end now :D 
Probably some failed parsing somewhere up in the file

lucario387 (approved):

KiLLeRRaT:
I've been using this and my `if` blocks do some funny stuff.

Pipes added to make tabs visible.

Orig:
```
if (name == "john")
{
|  name = "sam";
}
```

Format using `3==`.  Adds a tab infront of `{` which it shouldn't.
```
if (name == "john")
|  {
|  name = "sam";
}
```


Same story with my `try` blocks, except it adds a tab infront of `{` and `}`
Orig:
```
try
{
|  name = "harry";
}
```

Becomes:
```
try
|  {
|  name = "harry";
|  }
```


Any ideas?

vlada-dudr:
Hm, that's wierd. It works for me. Don't you have so syntax error before this snippet?

I have to figure out how to create some reasonable tests. I will take me some time though.

lucario387:
> I have to figure out how to create some reasonable tests

In most cases, just a normal C# file will suffice. Don't bloat it up. I think for this around 100-200 lines may be enough. Double that if you want to test for multiple cases (braces on newline/eol, `else` after `}`, `else` in newline, ...)

vlada-dudr:
Yeah, I want to do mostly some tricky cases. I was struggling to make the testing machine work - i'd expect the `whole_file` thing takes file, tries to reindent it and passes if nothing changed. Sadly I was unable to break it - it always passed.

lucario387:
>  i'd expect the `whole_file` thing takes file, tries to reindent it and passes if nothing changed
That is what it does, and what it expects to be

> Sadly I was unable to break it - it always passed.
Which should mean you've done a great job :tada:

vlada-dudr:
> Which should mean you've done a great job 🎉
I wish you are right :-D Well, first thing I did was to test the test - put something really off to see what i get back... And got success. So I went ahead and tried to break tests for other languages - always passed, so either it is broken, or my setup is wonky.

lucario387:
You can publish it and let the CI/me test it as well.

KiLLeRRaT:
> Hm, that's wierd. It works for me. Don't you have so syntax error before this snippet?
> 
> I have to figure out how to create some reasonable tests. I will take me some time though.

No syntax errors here :)

clason:
Can you take care of the failing tests? There were some changes to the parser that apparently need adjustment for.

PR #5351, opened 2023-09-07
fix(lua): highlights for attribute

MunifTanjim:
- https://github.com/MunifTanjim/tree-sitter-lua/issues/39
- https://github.com/MunifTanjim/tree-sitter-lua/pull/40

### DIFF
--- a/queries/lua/highlights.scm
+++ b/queries/lua/highlights.scm
@@ -140,9 +140,10 @@
   (#eq? @keyword.coroutine "coroutine"))
 
 (variable_list
-   attribute: (attribute
-     (["<" ">"] @punctuation.bracket
-      (identifier) @attribute)))
+  (attribute
+    "<" @punctuation.bracket
+    (identifier) @attribute
+    ">" @punctuation.bracket))
 
 ;; Labels
 


queries/lua/highlights.scm
```diff
@@ -139,11 +145,6 @@
 ((identifier) @keyword.coroutine
   (#eq? @keyword.coroutine "coroutine"))
 
-(variable_list
-   attribute: (attribute
-     (["<" ">"] @punctuation.bracket
-      (identifier) @attribute)))
```
MunifTanjim, line 145: I remember this working before. Not sure if anything changed that make it not working anymore.

For some reason, now only `"<"` is being matched as `@punctuation.bracket` with this query.

amaanq: I don't think that'd have ever worked, and even the query execution not erroring out on an impossible pattern is interesting (">" is impossible before an identifier)

If it did work before a while ago, my guess is it was a bug that was fixed

This would be correct:

```scm
(variable_list
  (attribute
    "<" @punctuation.bracket
    (identifier) @attribute)
    ">" @punctuation.bracket)
```

amaanq (approved):

PR #5358, opened 2023-09-08
fix(julia): highlight for `end` keyword

wangl-cc:
At present, the `end` is defined as `@keyword` globally, causing every instance of `end` to be recognized as a keyword, even in cases where it should not serve as one, as in `if foo; end`. In such cases, `end` should be categorized as `@conditional`.

Upon reviewing the `grammar.js` file of `tree-sitter-julia`, I noticed that the `end` keyword is never used in isolation, except within the keyword list. Furthermore, this keyword list is solely employed in the definition of the `quote_expression`, where keywords should be treated as regular identifiers. Based on this analysis, I have removed `macro`, `struct`, and `end` from the global `@keyword` and now match them within their specific scopes.

Here is a comparison between the highlight before this change (left) and after this change (right). It's worth noting that prior to this update, the `end` keywords were shown in italics, while the corresponding `if`, `for`, and `function` keywords were not.

Screenshot 2023-09-09 at 00 57 40

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -157,9 +157,6 @@
 [
   "global"
   "local"
-  "macro"
-  "struct"
-  "end"
 ] @keyword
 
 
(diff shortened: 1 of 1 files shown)


clason:
@savq

savq:
LGTM. Each `end` should only be captured once.

clason (approved):

PR #5365, opened 2023-09-09
feat(eds): add parser and queries for EDS files

uyha:


queries/eds/highlights.scm
```diff
@@ -0,0 +1,63 @@
+["="] @punctuation.delimeter
+
+(
+ (section_name) @variable.builtin
+ (#any-of? @variable.builtin "[FileInfo]" "[DeviceInfo]" "[DummyUsage]" "[MandatoryObjects]" "[OptionalObjects]")
+)
```
amaanq, line 6: I can be wrong and maybe it's intended to have section_name include the brackets, but extracting the brackets to the outer node which is then followed by the section_name node of just text could be nicer (not an issue here)

queries/eds/highlights.scm
```diff
@@ -0,0 +1,63 @@
+["="] @punctuation.delimeter
```
amaanq, line 1:
```suggestion
"=" @punctuation.delimeter
```

queries/eds/folds.scm
```diff
@@ -0,0 +1,3 @@
+[
+ (section)
+] @fold
```
amaanq, line 3:
```suggestion
(section) @fold
```

queries/eds/highlights.scm
```diff
@@ -0,0 +1,63 @@
...
+
+(
+ (section_name) @variable.builtin
+ (#any-of? @variable.builtin "[FileInfo]" "[DeviceInfo]" "[DummyUsage]" "[MandatoryObjects]" "[OptionalObjects]")
+)
+
+(
+ (section_name) @variable.builtin
+ (#lua-match? @variable.builtin "^%[1")
+)
```
amaanq, line 11: Can we follow the style seen in pretty much every query? No dangling paren mainly..
```suggestion
((section_name) @variable.builtin
  (#lua-match? @variable.builtin "^%[1"))
```

queries/eds/highlights.scm
```diff
@@ -0,0 +1,63 @@
...
+
+(
+ (section
+   (section_name) @_name
+   (#lua-match? @_name "Comments")
+   ) @comment
+ )
```
amaanq, line 18: This marks the whole section itself as a comment, not the name/body (or other nodes that are children), just want to double check that that's intended, and also the extra parenthesis pair isn't needed here and in a bunch of places below, would be nice to tidy that up

```suggestion
(section
   (section_name) @_name
   (#lua-match? @_name "Comments")) @comment
```

amaanq (changes requested):

queries/eds/highlights.scm
uyha: yes, this is intentional that the whole section is a comment block, not just the section name.

amaanq (approved):

PR #5369, opened 2023-09-10
fix(swift): Fix property highlighting in value arguments after latest upstream PR

alex-pinkus:
- Update parsers: swift
- fix(swift): Fix property highlighting in value arguments

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -546,7 +546,7 @@
     "revision": "697bb515471871e85ff799ea57a76298a71a9cca"
   },
   "swift": {
-    "revision": "29541ac9bbe2090de75d0b1e70360b85bbda1fef"
+    "revision": "10eb01d29827f24b1271672e89790661d94da9e1"
   },
   "sxhkdrc": {
     "revision": "440d5f913d9465c9c776a1bd92334d32febcf065"
--- a/queries/swift/highlights.scm
+++ b/queries/swift/highlights.scm
@@ -63,7 +63,7 @@
 (navigation_expression
   (navigation_suffix (simple_identifier) @property))
 (value_argument
-  (simple_identifier) @property)
+  name: (value_argument_label) @property)
 
 (import_declaration ["import" @include])
 
@@ -186,3 +186,4 @@
  "..."
   (bang)
 ] @operator
+
--- a/queries/swift/highlights.scm
+++ b/queries/swift/highlights.scm
@@ -63,7 +63,7 @@
 (navigation_expression
   (navigation_suffix (simple_identifier) @property))
 (value_argument
-  name: (simple_identifier) @property)
+  (simple_identifier) @property)
 
 (import_declaration ["import" @include])
 


clason:
Isn't that exactly what I changed in the other PR? (Not complaining, just curious.)

clason (approved):

clason:
@alex-pinkus I've noticed that you stopped maintaining the queries here (in the sense of adapting them to breaking parser changes). I've tried my best to fix query errors, but the latest update is too much. What are your plans here going forward? I see two options:
1. Switch to tracking releases so there will be fewer updates -- but those will be more breaking, and I'd hope for a timely PR to nvim-treesitter here; for this, it would also be good if you added WASM builds to your releases (you can use the [upstream release workflow](https://github.com/tree-sitter/workflows) to automate that) so we can later bundle these with Neovim;
2. Drop the parser to "unmaintained" if you're no longer interested in maintaining the queries here, in which case there will be no more parser updates.

alex-pinkus:
I think my main problem is twofold:
* I don’t get notified when a change upstream is going to be incompatible with the highlight queries here
* There isn’t an atomic way for me to issue a release and update this at the same time and we haven’t discussed an SLA (if the expectation is that all changes be adapted immediately to work here then that seems rather restrictive)

In this case I’m surprised there was breakage since my change upstream was mostly additive. What broke, and how do I better capture it with in-repo tests so that I know to update it here?

clason:
Yeah, I understand that the situation is difficult; and I want to make it clear that I'm not complaining but genuinely trying to make the process as smooth as possible (for you as well as for me).

Hence my (non-rhetorical) question was how much you can and want to actually act as query maintainer here. This is a discussion we can have here or on Matrix/Discord, as you prefer.

But a quick first answer to your questions:
1. Unfortunately, we only find out when we try to update; if that happens, I first try to fix this myself (if the needed changes are clear, either from the grammar diff or -- better -- from the corresponding query changes in the breaking commit); only if I can't do that I will (manually) ping the maintainer. In this case, the issue is that the queries have diverged significantly so the breakage [here](https://github.com/nvim-treesitter/nvim-treesitter/blob/94ea4f436d2b59c80f02e293466c374584f03b8c/queries/swift/highlights.scm#L159-L179) do not occur in the upstream queries. I'm not sure where the divergence comes from; there's no real reason as far as I can see (the only necessary differences are different capture _names_ and prefering `#lua-match` over `#match` where possible).
2. There's no need for immediate updates here; it's enough for me to know that they are coming within, say, a week or two. And I would very much prefer to track releases (wherever possible) to get much less frequent but more meaningful updates. (That is one of the main components of the 1.0 refactor that I'm currently finishing.)



4.

alex-pinkus:
I see, that makes sense! I hadn't realized that those query updates had been added here; I've been manually copying changes back but hadn't checked in a while. Once we get back in sync (modulo `#lua-match`) I'd be perfectly happy to maintain a week-or-so SLA on updates.

What's the right way to switch to tracking releases? I think you've mentioned some sort of format for release zips before. Can easily snap to a standard if one exists.

clason:
Yeah, the constant updates were just untenable. 

On our side, tracking releases is as simple as changing the `tier` field in the parser entry from 2 to 1, which tells the script to check releases instead of commits to the default branch. (Right now this would be a downgrade, so I'll wait until your next release.)

On your side, nothing special is required except making versioned releases (which you already do) on whatever schedule works for you. We'd ask you to use "semver" for this from the point of queries:
1. `point` releases are just grammar or scanner (or binding) fixes or refactors that don't affect queries at all;
2. `minor` releases include changes to the grammar that are non-breaking but need query changes to make use of new features (e.g., adding a new node);
3. `major` releases include changes to the grammar that break queries (removing nodes, refactoring patterns etc.).

No special release format is required since we still check out the repo even on releases (I changed my mind on this); the only thing that would be good is to add (manually or via the [upstream workflow]) a WASM build. (This is what we will use for Tier 1 parsers in the future, when Neovim has support for WASM parsers by default -- probably by 0.13.)

clason:
Oh, and you would make things easier for users if you could keep the generated `grammar.json` (and `node-types.json`) in the (main) repository. Unlike `parser.c`, these are relatively small and have meaningful semantic changes, so tracking them in git is not a big burden, but you can `tree-sitter generate` the parser from them without requiring a node installation. (Swift is one of only two parsers where that is required.)

PR #5372, opened 2023-09-11
fix!: backport `ac9e755` from `main`

llllvvuu:
The signature for `LanguageTree:parse()` changed to `LanguageTree:parse(range)` in https://github.com/neovim/neovim/pull/24647. Lua ignores extra arguments. That means if we pass in the right `range`, we get unchanged behavior for pre- patch users, and the correct behavior for post- patch users.

Closes: neovim/neovim#25088


lucario387 (approved):
I personally have been using it for (as of now) 3 weeks and seen no issue.

clason (approved):

PR #5376, opened 2023-09-11
Mark all quoted symbols as @symbol

17451k:
I've noticed that Neovim does not distinguish quoted and unquoted lists and highlights them the same.

For example, `(a b c)` is a list that can be evaluated as a function call to `a` with `b` and `c` as arguments. But `'(a b c)` is a list of the symbols `a`, `b`, and `c`, not evaluated, just data.

How it looks now:
image

How it will look with this PR:
image

As you may note, this PR doesn't fix highlighting inside `(quote (a b c))` - this requires a minor change to the racket treesitter parser and will be implemented separately.

**Important disclaimer**: I am not a Racket expert, so it's possible that I understand `quote` semantics incorrectly. I've tried to look at how other editors highlight such code (Emacs, VS Code, Sublime Text, Helix), but they either does not highlight quotes at all, or highlight them approximately the same as Neovim now.


17451k:
@Lazerbeak12345 Thanks for checking! I've changed the PR to exclude changes to the Strings, now they will look more like in DrRacket.

image

Lazerbeak12345:
Oh, don't forget numbers too.

image

Lazerbeak12345:
Oh, and booleans. I'm pretty sure just most primitive values.

image

queries/racket/highlights.scm
```diff
@@ -58,7 +62,7 @@
 
 (list
  .
- (symbol) @function)
+ (symbol) @function (#not-has-ancestor? @function quote))
```
amaanq, line 65: just an fyi the ancestor functions are expensive - I'd pick one to use it on, and then another not - the latter one will override the looser former one, you can structure it how you'd like

PR #5384, opened 2023-09-12
perf(comment): remove spell capture

tomtomjhj:
Problem:
Comment highlight query produces too many `@spell` captures.
https://github.com/neovim/neovim/issues/25113#issuecomment-1715548267

Solution:
Remove the query. This is fine because comment parser is only used in injection and the parent language has `@spell` for comment.

### DIFF
--- a/queries/agda/highlights.scm
+++ b/queries/agda/highlights.scm
@@ -35,7 +35,7 @@
 
 (pragma) @preproc
 
-(comment) @comment
+(comment) @comment @spell
 
 ;; Keywords
 [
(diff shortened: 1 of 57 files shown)


lucario387:
We will need to add `@spell` to all `@comment` as well.

lucario387 (approved):

clason:
Should do it in this PR lest we forget.

tomtomjhj:
Added `@spell` to nodes that look like actual user-edited comment.

tomtomjhj:
Oh it shouldn't be added to injection queries

llllvvuu:
awesome, thank you!

does this need to be manually added to main as well or does that tend to just get rebased?

Issue #5057, opened 2023-07-03
Comment parser or queries are very slow for long comments


stasjok:
### Describe the bug

When there is a long comment (not multiple single line comments, but one long multiline) scrolling through it is very slow and laggy. 

### To Reproduce

Latest nvim.appimage. init.lua:

```lua
require("nvim-treesitter.configs").setup({
        ensure_installed = { "lua", "comment" },
        highlight = { enable = true },
})
```

1. `:e $VIMRUNTIME/lua/vim/lsp/protocol.lua`
2. Scroll down to the end of file with `<C-F>`. Experience very noticeable lag on comment for `protocol.make_client_capabilities()` function.

### Expected behavior

No lag.

### Output of `:checkhealth nvim-treesitter`

```markdown
Installation
- WARNING tree-sitter executable not found (parser generator, only needed for :TSInstallFromGrammar, not required    
for :TSInstall)
- OK node found v18.16.1 (only needed for :TSInstallFromGrammar)
- OK git executable found.
- OK cc executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: gcc (GCC) 12.3.0
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with
runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "4.19.0-17-amd64",
  sysname = "Linux",
  version = "#1 SMP Debian 4.19.194-3 (2021-07-18)"
}

Parser/Features         H L F I J
  - bash                ✓ ✓ ✓ . ✓
  - c                   ✓ ✓ ✓ ✓ ✓
  - comment             ✓ . . . .
  - lua                 ✓ ✓ ✓ ✓ ✓
  - markdown            ✓ . ✓ ✓ ✓
  - markdown_inline     ✓ . . . ✓
  - python              ✓ ✓ ✓ ✓ ✓
  - query               ✓ ✓ ✓ ✓ ✓
  - vim                 ✓ ✓ ✓ . ✓
  - vimdoc              ✓ . . . ✓

  Legend: H[ighlight], L[ocals], F[olds], I[ndents], In[j]ections
         +) multiple parsers found, only one will be used
         x) errors found in the query, try to run :TSUpdate {lang}
```

### Output of `nvim --version`

```text
NVIM v0.10.0-dev-618+g4fd852b8c
Build type: RelWithDebInfo
[...]

PR #5386, opened 2023-09-12
docs(README): call out the comment parser as slow

llllvvuu:
This issue keeps coming up: https://github.com/nvim-treesitter/nvim-treesitter/issues/5057

Adding a note here may cause less users to end up here.


clason:
We don't kink shame. _Every_ parser (unless they're bundled with Neovim) should be installed only when you know what you're doing. `comment` is by far not the worst offender here.

Issue #4918, opened 2023-06-07
Sometimes `jsx` files are not getting parsed - "No parser available for the given buffer"


primeapple:
### Describe the bug

Sometimes when working my highlighting for `jsx` files does not work. It may also be on `js` files or even others, but there the lsp is providing semantic information, so it doesn't catch my eye.

Typing `:Inspect` on a token in this file does only display a "Semantic tokens" and not a "Treesitter" section.

Running `:InspectTree` yields the following error:

```
Error executing Lua callback: /usr/share/nvim/runtime/lua/vim/treesitter/playground.lua:241: No parser available for the given buffer                                                                                                           
stack traceback:                                                                                                                                                                                                                                
        [C]: in function 'assert'                                                                                                                                                                                                               
        /usr/share/nvim/runtime/lua/vim/treesitter/playground.lua:241: in function 'inspect_tree'                                                                                                                                               
        /usr/share/nvim/runtime/lua/vim/treesitter.lua:496: in function 'inspect_tree'                                                                                                                                                          
        /usr/share/nvim/runtime/plugin/nvim.lua:18: in function </usr/share/nvim/runtime/plugin/nvim.lua:9> 
```

### To Reproduce

I don't know exactly how to reproduce. When I close Neovim and reopen it works for a while on `jsx` files.

It seems that once it fails once for a `.jsx` file it doesn't even try anymore and fails on all the other `.jsx` files as well.

### Expected behavior

_No response_

### Output of `:checkhealth nvim-treesitter`

```markdown
Error executing Lua callback: /usr/share/nvim/runtime/lua/vim/treesitter/playground.lua:241: No parser available for the given buffer                                                                                                           
stack traceback:                                                                                                                                                                                                                                
        [C]: in function 'assert'                                                                                                                                                                                                               
        /usr/share/nvim/runtime/lua/vim/treesitter/playground.lua:241: in function 'inspect_tree'                                                                                                                                               
        /usr/share/nvim/runtime/lua/vim/treesitter.lua:496: in function 'inspect_tree'                                                                                                                                                          
        /usr/share/nvim/runtime/plugin/nvim.lua:18: in function </usr/share/nvim/runtime/plugin/nvim.lua:9> 
```
```


### Output of `nvim --version`

```text
NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3
```


### Additional context

My Treesitter config (using lazy.nvim):

```lua
    {
[...]

PR #5393, opened 2023-09-13
fix(query): apply `parsers.ft_to_lang` to `vim.filetype.match`

llllvvuu:
This fixes an issue where `jsx` code blocks resolve to `javascriptreact` filetype, but there is no `javascriptreact` lang. So markdown JSX code blocks are not highlighted.

Tangential question regarding an edge case (will we ever hit this?):
- What if `lang` and `injection_alias` both exist? Currently, `lang` of the injection alias interpreted as a file extension is preferred regardless of whether a parser exists for it. Should we check nvim core `has_parser` API? And what if both `lang` and `injection_alias` have parsers? Which should be preferred?

Fixes: #4918

### DIFF
--- a/lua/nvim-treesitter/query_predicates.lua
+++ b/lua/nvim-treesitter/query_predicates.lua
@@ -1,4 +1,5 @@
 local query = require "vim.treesitter.query"
+local parsers = require "nvim-treesitter.parsers"
 
 local html_script_type_languages = {
   ["importmap"] = "json",
@@ -16,8 +17,9 @@ local non_filetype_match_injection_language_aliases = {
 }
 
 local function get_parser_from_markdown_info_string(injection_alias)
-  local match = vim.filetype.match { filename = "a." .. injection_alias }
-  return match or non_filetype_match_injection_language_aliases[injection_alias] or injection_alias
+  local ft = vim.filetype.match { filename = "a." .. injection_alias }
+  local lang = parsers.ft_to_lang(ft)
+  return lang or non_filetype_match_injection_language_aliases[injection_alias] or injection_alias
 end
 
 local function error(str)


clason:
Thank you, but I don't think we really want to make changes to code that is on the chopping block (that whole functionality has been upstreamed -- and much improved -- where it belongs).

llllvvuu:
is there nothing we can backport? the highlighting is broken against both stable and nightly Neovim and this little one-liner seems to have no downsides (it just fixes something that was obviously wrong - function should return a lang not an ft)

but also, how can upstream have this functionality if it only has `ft_to_lang` and not `lang_to_lang`?

llllvvuu:
> but also, how can upstream have this functionality if it only has ft_to_lang and not lang_to_lang?

On this (tangential) topic:

I checked `nvim-treesitter:main` and `neovim:master` and it looks like we're just [passing the code fence tag straight through as @injected.language](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/runtime/queries/markdown/injections.scm) (a one-to-one operation), and Neovim is just pulling `parser/{@injected.language}` (a one-to-one operation).

A many-to-one step is missing, ideally one which is user-configurable. `vim.treesitter.language.alias()` ?

llllvvuu:
Actually, I found a less deprecated code path to patch: https://github.com/neovim/neovim/pull/25143

Closing this.

clason:
> is there nothing we can backport?

Not really, since the changes are "rip everything out and rely on neovim master" (which we can't for the default branch until 0.10 is released and we drop 0.9 compatibility -- at which point `main` becomes the default. 

And making changes here makes my life rebasing main on harder...

Fixing this in Neovim -- as you are doing -- is the right approach, and very welcome.

PR #5394, opened 2023-09-14
fix(t32): update indent queries & test

xasc:
In the latest version of the grammar, handling of consecutive line breaks has changed. Tests inserting single lines will now fail (after next parser update). I am not quite sure why this is happening.

The updated queries are a bit simpler than before and seem to do the job.

### DIFF
--- a/tests/indent/t32/if_block.cmm
+++ b/tests/indent/t32/if_block.cmm
@@ -43,7 +43,7 @@ IF &f
     stop
   ELSE
     IF &h
-  (
-    continue
-  )
+    (
+      continue
+    )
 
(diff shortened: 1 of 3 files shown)


clason:
@xasc

tests/indent/t32/if_block.cmm
```diff
@@ -43,7 +43,7 @@ IF &f
     stop
   ELSE
     IF &h
-  (
-    continue
-  )
+    (
+      continue
+    )
```
lucario387, line 48: I don't think this is a problem or anything.

As it stands, handling any if-else cases without specific `elseif` nodes, or without brackets will always be a mess. As long as it's consistent during usage (which to me seems like it does, as the changes to the test file here are very minimal).

lucario387:
Needing a lockfile update as well @xasc. I agree with the changes.

clason:
Lockfile was just updated; you can rebase on master.

xasc:
I did the rebase.

lucario387 (approved):

PR #5396, opened 2023-09-14
add parsers and queries for apex, sosl, soql

xixiaofinland:


queries/apex/tags.scm
```diff
@@ -0,0 +1,24 @@
+(class_declaration
```
clason, line 1: Remove this file; we don't use these queries.

queries/apex/textobjects.scm
```diff
@@ -0,0 +1,70 @@
+(class_declaration
```
clason, line 1: Same; these queries need to be added to `nvim-treesitter/text-objects`

clason:
Missing injection queries?

clason:
Also, just to make sure: You can only use captures listed in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights

xixiaofinland:
> Missing injection queries?

no, I removed injection queries, there is no need for the time being.

clason:
Wait, wasn't that the whole point? That `sosl` can be nested in `apex` etc.?

xixiaofinland:
> Wait, wasn't that the whole point? That `sosl` can be nested in `apex` etc.?

no, soql is already baked into apex's parser in the compiling time. 
My `injections.scm` was meant to use SOQL highlighting only. I didn't know it will trigger the soql parsing.

xixiaofinland:
> Also, just to make sure: You can only use captures listed in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights

i think it's aligned now in the latest commit.

clason:
> no, soql is already baked into apex's parser in the compiling time.

That's weird; this is _not_ how tree-sitter is supposed to work.

xixiaofinland:
> > no, soql is already baked into apex's parser in the compiling time.
> 
> That's weird; this is _not_ how tree-sitter is supposed to work.

not sure what you mean. `soql` is a separate `sql` like language in Salesforce, but it's also natively part of `apex` so we can do things like below. So an apex parser naturally should have soql baked in, and this parser does so.

```
Account a = [select id from Account];
```

xixiaofinland:
not sure about those failed checks, do I need to take actions? @clason

clason:
Yes, see my comments above.

clason:
Ah, sorry, my comment got lost.

This is not how you specify parsers that don't live in the top-level of the repo; check, e.g., the `terraform` or `csv` parser.

lucario387:
If `soql` and `sosl` parsing is already handled with `apex`, then there's no need for `soql` and `sosl` queries, just throw them all into `apex` queries

aheber:
If it helps, SOQL and SOSL query structures are a slightly different dialect when embedded in Apex. Among other things you can have Apex embedded inside SOQL which is embedded inside of Apex, so on and so forth.

It isn't the pure external language dialect. To teach them to behave correctly inside of Apex I found it was better to include them directly with modified behavior flags rather than treat them as injection languages with fully abstracted parsers.

The reason we also have additional parsers are for the "external" dialect that can be used outside of the Apex language in independent files, because they have subtle differences I needed a slightly different parser behavior to work with those files.

lucario387:
Thanks for the thorough response. It clears a lot of questions I have with the queries of this PR

xixiaofinland:
> Thanks for the thorough response. It clears a lot of questions I have with the queries of this PR

Thanks for the questions too. It reminds us how one of a kind these three languages are both in combination and separation.

lua/nvim-treesitter/parsers.lua
```diff
@@ -91,6 +91,14 @@ list.agda = {
   maintainers = { "@Decodetalkers" },
 }
 
+list.apex = {
+  install_info = {
+    url = "https://github.com/aheber/tree-sitter-sfapex/tree/main/apex",
```
lucario387, line 96: This is still wrong, please check the `list.csv` for an example of a correct non-toplevel parser

xixiaofinland: @lucario387 These three parsers worked seamlessly with local installation in my nvim. Do they need to be restructured to align with standards or?

@aheber Not sure what to be changed. 
Your call here. https://github.com/amaanq/tree-sitter-csv

lucario387: No need for his call, you need to use `location = "apex"` instead of declaring `tree/main/apex`

lucario387: Example with `csv` parser

```lua
list.csv = {
  install_info = {
    url = "https://github.com/amaanq/tree-sitter-csv",
    files = { "src/parser.c" },
    location = "csv",
  },
  maintainers = { "@amaanq" },
}
```

xixiaofinland: Thanks, done

lucario387:
@xixiaofinland you probably can run the tests locally by doing `./scripts/check-queries.lua`

xixiaofinland:
> @xixiaofinland you probably can run the tests locally by doing `./scripts/check-queries.lua`

all pass locally now, thanks for the tip

xixiaofinland:
> @xixiaofinland you probably can run the tests locally by doing `./scripts/check-queries.lua`

I added the `pre-push` hook by following link below and it ran with all pass locally.  I will recheck why checks here fail still. @lucario387  
https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#style-checks-and-tests

lucario387:
`pre-push` scripts only check for formatting and simple linting issues for lua codes, not queries.

lucario387:
Running `./scripts/check-queries.lua` locally, I get the same result as CI's. So I suppose either:
- nvim-treesitter is not loaded on startup -> it just doesn't run properly ig
- Something else

xixiaofinland:
> Running `./scripts/check-queries.lua` locally, I get the same result as CI's. So I suppose either:
> 
> * nvim-treesitter is not loaded on startup -> it just doesn't run properly ig
> * Something else
I assume you mean run it in the console, right? I encountered this error and then tried the "pre-hook" way.
I know `parsers.lua` are in nvim package manager installed folder only, not sure how to get it working...

Image 008

lucario387:
Yup, your nvim-treesitter is not loaded on startup. You'll need to make sure it's loaded (idk how to do it with lazy.nvim, but `lazy = false` may suffice).

xixiaofinland:
> Yup, your nvim-treesitter is not loaded on startup. You'll need to make sure it's loaded (idk how to do it with lazy.nvim, but `lazy = false` may suffice).

Sorry, a bit confused. I'm new to running a lua test script. Am I supposed to execute this in a Nvim runtime like in `nvim --headless` and make sure nvim-treesitter module is loaded already, or in the console just like a `bashscript` in my screenshot (as it has #! at the beginning)?

lucario387:
> Am I supposed to execute this in a Nvim runtime like in nvim --headless

Yes. That's exactly what happened during CI

queries/soql/highlights.scm
```diff
@@ -0,0 +1,142 @@
+; https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#semantic-token-classification
```
ObserverOfTime, line 1: What's this comment here for?
```suggestion
```

queries/soql/highlights.scm
```diff
@@ -0,0 +1,142 @@
...
+  (field_list
+    (dotted_identifier
+      (identifier) @property )))
+
+(alias_expression
+  (identifier) @label)
+
+(storage_identifier) @type
+(function_name) @function
+(date_literal) @variable.readonly.defaultLibrary
```
ObserverOfTime, line 38: Date literals can be `@string.special`, `@number`, or `@constant`.

queries/soql/highlights.scm
```diff
@@ -0,0 +1,142 @@
...
+
+(else_expression
+  (field_list
+    (dotted_identifier
+      (identifier) @property )))
+
+(alias_expression
+  (identifier) @label)
+
+(storage_identifier) @type
```
ObserverOfTime, line 36: Would `@storageclass` make sense here?

queries/soql/highlights.scm
```diff
@@ -0,0 +1,142 @@
...
+  "NOT_IN"
+  "INCLUDES"
+  "EXCLUDES"
+] @operator
+(value_comparison_operator "<" @operator)
+"<=" @operator
+(value_comparison_operator ">" @operator)
+">=" @operator
+ @operator
+(set_comparison_operator "IN" @operator)
```
ObserverOfTime, line 65:
```suggestion
[
  "AND"
  "OR"
  "NOT"
  "LIKE"
  "NOT_IN"
  "INCLUDES"
  "EXCLUDES"
] @keyword.operator

[
  "="
  "!="
  "<="
  "=>"
] @operator

(value_comparison_operator  [ "<" ">" ] @operator)

(set_comparison_operator "IN" @keyword.operator)
```

queries/soql/highlights.scm
```diff
@@ -0,0 +1,142 @@
...
+
+(int) @number
+(decimal) @number
+(currency_literal) @number
```
ObserverOfTime, line 69:
```suggestion
[
  (int)
  (decimal)
  (currency_literal)
] @number
```

queries/soql/highlights.scm
```diff
@@ -0,0 +1,142 @@
...
+(string_literal) @string
+(date) @variable.readonly
+(date_time) @variable.readonly
```
ObserverOfTime, line 72: How do these differ from `(date_literal)`?

queries/soql/highlights.scm
```diff
@@ -0,0 +1,142 @@
...
+
+[
+  "TRUE"
+  "FALSE"
+  (null_literal)
+] @variable.readonly.defaultLibrary
```
ObserverOfTime, line 78:
```suggestion
[
  "TRUE"
  "FALSE"
] @boolean

(null_literal) @constant.builtin
```

queries/soql/highlights.scm
```diff
@@ -0,0 +1,142 @@
...
+  "TRACKING"
+  "TYPEOF"
+  "UPDATE"
+  "USING_SCOPE"
+  "VIEW"
+  "VIEWSTAT"
+  "WITH"
+  "WHERE"
+  "WHEN"
+] @keyword
```
ObserverOfTime, line 117: Some of these should be `@conditional` or `@repeat`.

queries/apex/highlights.scm
```diff
@@ -0,0 +1,340 @@
...
+
+
+;; Types
+
+(interface_declaration
+  name: (identifier) @type)
+(class_declaration
+  name: (identifier) @type)
+(class_declaration
+  (superclass) @storageclass) ;;
```
ObserverOfTime, line 52:
```suggestion
  (superclass) @type) 
```

queries/apex/highlights.scm
```diff
@@ -0,0 +1,340 @@
...
+(interface_declaration
+  name: (identifier) @type)
+(class_declaration
+  name: (identifier) @type)
+(class_declaration
+  (superclass) @storageclass) ;; 
+(enum_declaration
+  name: (identifier) @type)
+(enum_constant
+  name: (identifier) @variable);;
```
ObserverOfTime, line 56:
```suggestion
  name: (identifier) @constant)
```

queries/apex/highlights.scm
```diff
@@ -0,0 +1,340 @@
...
+
+(interfaces
+  (type_list
+    (type_identifier) @function ));;
```
ObserverOfTime, line 60: Why is this `@function`?

ObserverOfTime (changes requested):

queries/apex/highlights.scm
xixiaofinland: I copied the original `highlights.scm` which was tailor made for vscode.
Some annotations are not used by nvim, I need to quickly replace them while I don't have good knowledge of what new annotation to use.. so, some might seem funny like this...

queries/soql/highlights.scm
xixiaofinland: copied from the original scm, I will remove it

xixiaofinland:
> Yup, your nvim-treesitter is not loaded on startup. You'll need to make sure it's loaded (idk how to do it with lazy.nvim, but `lazy = false` may suffice).

I'm not sure how to get modules to load before running a cmd in `headless`, posted the question in Reddit.

queries/apex/highlights.scm
xixiaofinland: > Why is this `@function`?

what's the matching annotation for `interface` ? Thank you so much for correcting these many. My purpose was to learn and fix along the journey :)

queries/soql/highlights.scm
xixiaofinland: not sure here. @aheber comments?

queries/apex/highlights.scm
ObserverOfTime: It should be `@type`.

queries/soql/highlights.scm
```diff
@@ -59,7 +59,6 @@
   "="
   "!="
   "<="
-  "=>"
```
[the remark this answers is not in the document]
xixiaofinland: > Oh, whoops. This should be ">=".

Embarrassing, you eye-fixed this whichI didn't what's that error and removed it to get the linter passing.
Just added it back

queries/apex/highlights.scm
```diff
@@ -0,0 +1,340 @@
...
+(field_declaration
+  (modifiers (modifier ["final" "static"])(modifier ["final" "static"]))
+  (variable_declarator
+    name: (identifier) @constant))
+
+(variable_declarator
+  (identifier) @property)
+
+((identifier) @constant
+  (#match? @constant "^_*[A-Z][A-Z\\d_]+$"))
```
lucario387, line 241: We prefer lua-match wherever possible

Check section 5.4.1 of luadoc, named lua-patterns

queries/soql/highlights.scm
aheber: `date_literal` are macros for relative moments: `TODAY`, `YESTERDAY`, `LAST_YEAR`
https://github.com/aheber/tree-sitter-sfapex/blob/e63bcdcc26ae808b3fe79dfb8fa61bebdb95bda4/common/soql-grammar.js#L328

`date` is a raw date string `YYYY-MM-DD`
https://github.com/aheber/tree-sitter-sfapex/blob/e63bcdcc26ae808b3fe79dfb8fa61bebdb95bda4/common/soql-grammar.js#L450

`date_time` is a raw datetime string `YYYY-MM-DDThh:mm:ss.mmmmZ`
https://github.com/aheber/tree-sitter-sfapex/blob/e63bcdcc26ae808b3fe79dfb8fa61bebdb95bda4/common/soql-grammar.js#L452

I tagged them differently because the macro format receives a moment-relative value when processed, it doesn't have a constant inherent value like the `date` and `date_time` expressions do.

queries/apex/highlights.scm
xixiaofinland: solved with latest commit

queries/soql/highlights.scm
xixiaofinland: In this case, I guess no change is needed? @ObserverOfTime

queries/apex/highlights.scm
ObserverOfTime, line 1: Since soql is embedded in apex, its highlights should be inherited if possible.

queries/soql/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+  "STANDARD"
+  "THEN"
+  "TRACKING"
+  "TYPEOF"
+  "UPDATE"
+  "USING_SCOPE"
+  "VIEW"
+  "VIEWSTAT"
+  "WITH"
+] @keyword
```
ObserverOfTime, line 120: "ELSE" and "THEN" should also be `@conditional`.

ObserverOfTime (changes requested):

queries/apex/highlights.scm
lucario387: This may or may not have been a design decision I feel
https://github.com/nvim-treesitter/nvim-treesitter/pull/5396#issuecomment-1720377411

ObserverOfTime: Yes, but without inheritance, soql won't be highlighted.

ObserverOfTime, line 1: `; inherits: soql` would be preferable so long as all the queries are compatible.

ObserverOfTime: You should revert 81be228b09eb110baec4bc58dd2343f63b42a552.

ObserverOfTime (approved):

xixiaofinland:
Thank you all for the great help, I have learned a lot during this process :)! @ObserverOfTime @clason @lucario387

PR #5404, opened 2023-09-16
injections(lua): add various vim api highlights

### LINKED ISSUE #5399: Also inject syntax highlighting in `nvim_create_autocmd`'s `command`

#1922 added VimScript injection to `vim.cmd`/`nvim_command`/`nvim_exec`. The `command` field in the `opts` dictionary passed to `:h nvim_create_autocmd` is also VimScript, but doesn't currently get VimScript injected.

```lua
vim.api.nvim_create_autocmd({"BufEnter", "BufWinEnter"}, {
  pattern = {"*.c", "*.h"},
  command = "echo 'Entering a C or C++ file'", -- Doesn't currently get syntax highlighting
})
```

```lua
vim.api.nvim_create_autocmd({"BufEnter", "BufWinEnter"}, {
  pattern = {"*.c", "*.h"},
  callback = function()
    vim.cmd "echo 'Entering a C or C++ file'" -- Does get syntax highlighting
  end
})
```

The string that the `command` field passed to `nvim_create_autocmd` gets set to should also get VimScript injected.

lucario387:
Closes #5399

### DIFF
--- a/queries/lua/injections.scm
+++ b/queries/lua/injections.scm
@@ -28,6 +28,76 @@
   (#eq? @_method "nvim_exec_lua")
   (#set! injection.language "lua"))
 
+; vim.api.nvim_create_autocmd("FileType", { command = "injected here" })
+(function_call
+  name: (_) @_vimcmd_identifier
+  arguments:
+    (arguments
+      . (_)
+      . 
+      (table_constructor
+        (field
+          name: (identifier) @_command
+          value: 
+            (string content: (_) @injection.content)))
+      .) ; limit so only 2-argument functions gets matched before pred handle
+  (#eq? @_vimcmd_identifier "vim.api.nvim_create_autocmd")
+  (#eq? @_command "command")
+  (#set! injection.language "vim"))
+
+
+(function_call
+  name: (_) @_user_cmd
+  arguments: 
+    (arguments
+      . (_)
+      . 
+      (string
+        content: (_) @injection.content)
+      . (_) .)
+  (#eq? @_user_cmd "vim.api.nvim_create_user_command")
+  (#set! injection.language "vim"))
+
+(function_call
+  name: (_) @_user_cmd
+  arguments: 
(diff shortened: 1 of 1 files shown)


clason:
oof, that's a bit of a monster ;)

Can you also do `nvim_buf_set_keymap` (`vim.keymap.set` is probably not doable) and maybe `nvim_create_user_command`?

lucario387:
I find `nvim_buf_set_keymap` highlightings to be a bit misleading

```lua
vim.api.nvim_buf_set_keymap(0, "n", "abc", "xyz", {})
--                                          ^ keyword.vim
```

Will probably limit the content to capitalized words only or something else

clason:
Yeah, if it is not (easily) possible, just omit it. We can always point here if someone requests it.

lucario387:
Kindof scared to be pinged on lua files getting much slower so I will just narrow them as much as possible.

Downsides are these queries won't get its highlight until the final argument is given

```lua
vim.api.nvim_create_user_command("Foo", "echo 'Bar'")
--                                       ^ string
vim.api.nvim_create_user_command("Foo", "echo 'Bar'", {})
--                                       ^ keyword.vim
```

clason:
Yeah, I'd recommend leaving complicated queries for user configs. Maybe we can just comment them out, so that they're there for users to copy and uncomment?

lucario387:
I will still keep the narrowed down queries for `nvim_create_user_command`, as that can be used to signal users that they're not having enough arguments for that function

clason (approved):

clason:
I assume you did a test-drive and didn't notice any performance regression ;)

lucario387:
> I assume you did a test-drive and didn't notice any performance regression ;)

Which was the reason for the query narrowing :'). 

It's similar before and after this while editing `runtime/lua/util.lua` so I'll assume there isn't too much performance cost

PR #5408, opened 2023-09-17
fix: trees() may not be list-like

tomtomjhj:
See https://github.com/neovim/neovim/issues/25191


lucario387:
Don't think this is ever called here 🤔 and this will be removed soon once we drop support for 0.9

tomtomjhj:
The fixed function(s) are used in incremental selection. Before this fix, incremental selection on markdown_inline would fail (on latest nvim).

clason:
That function will also go away :)

clason (approved):

Issue #4801, opened 2023-05-14
markdown codeblock  conceal not work after update


glepnir:
### Describe the bug

just run plugin update then i find it not work anymore.  the codeblock still in markdown file.

### To Reproduce

1. `set conceal=2`
2. open a markdown file which have some code snippets like

```
```lua
print('hello')
\```

```

### Expected behavior

code block should be empty.

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- OK `tree-sitter` found 0.20.8 (parser generator, only needed for :TSInstallFromGrammar)
- OK `node` found v20.1.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: Apple clang version 14.0.3 (clang-1403.0.22.14.1)
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "22.4.0",
  sysname = "Darwin",
  version = "Darwin Kernel Version 22.4.0: Mon Mar  6 21:00:17 PST 2023; root:xnu-8796.101.5~3/RELEASE_X86_64"
} ~

Parser/Features         H L F I J
  - agda                ✓ . ✓ . .
  - awk                 ✓ . . . ✓
  - bash                ✓ ✓ ✓ . ✓
  - bass                ✓ ✓ ✓ ✓ ✓
  - c                   ✓ ✓ ✓ ✓ ✓
  - c_sharp             ✓ ✓ ✓ . ✓
  - capnp               ✓ ✓ ✓ ✓ ✓
  - chatito             ✓ ✓ ✓ ✓ ✓
  - clojure             ✓ ✓ ✓ . ✓
  - cmake               ✓ . ✓ . .
  - comment             ✓ . . . .
  - commonlisp          ✓ ✓ ✓ . .
  - cpp                 ✓ ✓ ✓ ✓ ✓
  - css                 ✓ . ✓ ✓ ✓
  - dart                ✓ ✓ ✓ ✓ ✓
  - devicetree          ✓ ✓ ✓ ✓ ✓
  - diff                ✓ . . . .
  - dockerfile          ✓ . . . ✓
  - dot                 ✓ . . . ✓
[...]

PR #5414, opened 2023-09-17
feat(markdown): conceal triple backticks and HTML entities

MariaSolOs:
Porting the queries that will be included in https://github.com/neovim/neovim/pull/25073.

### DIFF
--- a/queries/markdown/highlights.scm
+++ b/queries/markdown/highlights.scm
@@ -26,6 +26,14 @@
   (fenced_code_block_delimiter)
 ] @punctuation.delimiter
 
+;; Conceal backticks
+(fenced_code_block
+  (fenced_code_block_delimiter) @conceal
+  (#set! conceal ""))
+(fenced_code_block
+  (info_string (language) @conceal
+  (#set! conceal "")))
+
 (code_fence_content) @none
 
 [
--- a/queries/markdown_inline/highlights.scm
+++ b/queries/markdown_inline/highlights.scm
@@ -92,3 +92,11 @@
     "]"
   ] @conceal
   (#set! conceal ""))
+
+;; Replace common HTML entities.
+((entity_reference) @conceal (#eq? @conceal "&nbsp;") (#set! conceal ""))
+((entity_reference) @conceal (#eq? @conceal "&lt;") (#set! conceal "<"))
+((entity_reference) @conceal (#eq? @conceal "&gt;") (#set! conceal ">"))
+((entity_reference) @conceal (#eq? @conceal "&amp;") (#set! conceal "&"))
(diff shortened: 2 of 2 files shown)


clason (approved):

adrian5:
Not to beat some dead horse, but has there been a change of mind regarding [codeblock conceal](https://github.com/nvim-treesitter/nvim-treesitter/issues/4801#issuecomment-1546845938) or is this unavoidable due to the wider neovim PR this is a part of?

clason:
The latter.

mawkler:
Is this supposed to work in Markdown files? Triple backticks still aren't concealed for me.

This is the config I'm using:

<details>
<summary><b>Click to expand</b></summary>

```lua
local root = vim.fn.fnamemodify(vim.fn.expand('$HOME') .. '/code-other/nvim-test-config', ':p')
for _, name in ipairs({ 'config', 'data', 'state', 'cache' }) do
  vim.env[('XDG_%s_HOME'):format(name:upper())] = root .. '/' .. name
end

-- bootstrap lazy.nvim
local lazypath = root .. '/plugins/lazy.nvim'
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ 'git', 'clone', '--filter=blob:none', 'https://github.com/folke/lazy.nvim.git', lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

vim.g.mapleader = ' '

local plugins = {
  {
    'nvim-treesitter/nvim-treesitter',
    dependencies =  'nvim-treesitter/nvim-treesitter-textobjects',
    build = ':TSUpdate',
    opts = {
      ensure_installed = { 'markdown' },
      highlight = { enable = true },
    },
  },
}

vim.o.conceallevel = 2

require('lazy').setup(plugins, {
  root = root .. '/plugins',
})
```

</details>

Here's an example of a markdown file that I'm testing on:

    ```lua
    vim.o.conceallevel = 2
    ```

    ```
    foo bar
    ```

mawkler:
@clason I'm testing with the config posted in my previous comment, which includes the line `vim.o.conceallevel = 2` :)

lucario387:
Im having it concealed just fine, so not sure what's happening here

mawkler:
Never mind, it was a misconfiguration on my part. I forgot that I can't use Lazy's `opts` for nvim-treesitter in my test config. I changed to the following, which works:

```lua
{
  'nvim-treesitter/nvim-treesitter',
  dependencies =  'nvim-treesitter/nvim-treesitter-textobjects',
  build = ':TSUpdate',
  config = function()
      ensure_installed = { 'markdown' },
      highlight = { enable = true },
    })
  end
}
```

My guess is that the original problem was that I hadn't updated Neovim. Thanks for the help!

PR #5421, opened 2023-09-19
fix(highlight): apply @error to nodes under error, instead of parent node

llllvvuu:
feat: apply @error to nodes under error, instead of parent node

per @lucario387,
> When in error nodes, the tree is like (ERROR ","), so what's happening
> here is nvim is applying highlights onto ERROR node, but also having a
> separate highlight for ",", leading to it also getting applied. And
> because "," capture is more specific, it's taking priority

https://github.com/nvim-treesitter/nvim-treesitter/pull/5421#issuecomment-1724707540

Related: https://github.com/folke/tokyonight.nvim/issues/283

### DIFF
--- a/queries/ada/highlights.scm
+++ b/queries/ada/highlights.scm
@@ -192,5 +192,5 @@
 ;; Highlight errors in red. This is not very useful in practice, as text will
 ;; be highlighted as user types, and the error could be elsewhere in the code.
 ;; This also requires defining    :hi @error guifg=Red    for instance.
-(ERROR) @error
+(ERROR _ @error) ; up the specificity to nodes under error, instead of parent node
 
--- a/queries/bitbake/highlights.scm
+++ b/queries/bitbake/highlights.scm
@@ -357,4 +357,4 @@
 
 (comment) @comment @spell
 
-(ERROR) @error
+(ERROR _ @error) ; up the specificity to nodes under error, instead of parent node
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -283,4 +283,4 @@
   (attribute_declaration)
 ] @attribute
 
-(ERROR) @error
+(ERROR _ @error) ; up the specificity to nodes under error, instead of parent node
--- a/queries/cairo/highlights.scm
+++ b/queries/cairo/highlights.scm
@@ -335,4 +335,4 @@
 
 ; Errors
 
-(ERROR) @error
+(ERROR _ @error) ; up the specificity to nodes under error, instead of parent node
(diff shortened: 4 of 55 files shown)


lucario387:
I'm not really liking this. This is more of a specificity issue.

For `,` when in error nodes, the tree is like `(ERROR ",")`, so what's happening here is nvim is applying highlights onto `ERROR` node, but also having a separate highlight for `,`, leading to it also getting applied.

So it's still ordering issues.

lucario387:
For example, in json, this will work

```diff
diff --git a/queries/json/highlights.scm b/queries/json/highlights.scm
index 6163b209..ff261856 100644
--- a/queries/json/highlights.scm
+++ b/queries/json/highlights.scm
@@ -12,7 +12,6 @@
 
 (array (string) @string)
 
-(ERROR) @error
 
 ["," ":"] @punctuation.delimiter
 
@@ -28,3 +27,5 @@
 ((escape_sequence) @conceal
  (#eq? @conceal "\\\"")
  (#set! conceal "\""))
+
+ (ERROR _ @error) ; up the specificity to nodes under error, instead of parent node
```

llllvvuu:
I didn't know about that trick, thanks! gonna try it out and update

llllvvuu:
I've updated this PR to only address the specificity issue, and then I'll make a separate PR to add ERROR capture to queries that don't have it

lucario387:
I'll wait for other nvim-treesitter members to decide on this.

amaanq:
definitely a no on increasing priority, moving it to the bottom is good, I'm indifferent to using a more-specific wildcard capture

ObserverOfTime:
Is it possible for an error node not to have children?

clason:
Honestly, I see `(error)` nodes as purely diagnostic/debug entities that have no place in a highlight query. I would prefer them _all_ to be removed and handled (in a way I still need to think about) consistently by the highlighter in Neovim.

llllvvuu:
Whether we remove from the 1/3 of language that have it, or add it to the other 2/3, is whatever, as long as it's consistent. Definitely having it in Neovim would be the lowest-maintenance way to do the latter.

Otherwise, if we remove it without replacing the functionality in Neovim, people might have questions, but there are some fairly easy workarounds to point to.

clason:
> Otherwise, if we remove it without replacing the functionality in Neovim

But it's not intended functionality (and nvim-treesitter is experimental, yaddayadda).

clason:
Agreed, let's just remove all `@error` nodes from our queries. We have `:InspectTree` to check for parsing errors (in a better way).

PR #5423, opened 2023-09-19
feat: make `ERROR` node capture consistent across all langs

llllvvuu:
`ERROR` nodes are a concept fundamental to Tree-sitter. We capture them as `@error` in about a third of languages currently, which suggests that `@error` is an intended standard.

This patch extends it to all languages. Impact on users:

- Only certain Tree-sitter aware color schemes actually assign any visuals to parse errors; for example, catppuccin has it but tokyonight does not.
- Users of these themes already see errors prior to this PR; this just makes it consistent across all languages. I imagine if users found visual syntax errors annoying, they would've already experienced and disabled it.
- The color can be overwritten by users.

Note: whether we use `(ERROR) @error` or `(ERROR _ @error)` depends on the outcome of #5421. This can easily be find+replaced.

### DIFF
--- a/queries/agda/highlights.scm
+++ b/queries/agda/highlights.scm
@@ -79,3 +79,5 @@
 [
   "="
 ] @operator
+
+(ERROR _ @error) ; up the specificity to nodes under error, instead of parent node
--- a/queries/arduino/highlights.scm
+++ b/queries/arduino/highlights.scm
@@ -109,3 +109,5 @@
   (function_declarator
     declarator: (identifier) @constructor)
   (#eq? @type.builtin "SPISettings"))
+
+(ERROR _ @error) ; up the specificity to nodes under error, instead of parent node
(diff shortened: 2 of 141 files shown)


clason:
> Only certain Tree-sitter aware color schemes actually assign any visuals to parse errors; for example, catppuccin has it but tokyonight does not.

They should not. `@error` nodes are parsing errors and should not be highlighted (unless you like blinkenlights).

clason:
As discussed in #5421, we should remove all `ERROR` captures. If someone wants to highlight them for debugging purposes, we have other tools (tree inspector, query editor, or we could add this as an option to `vim.treesitter.start()`).

I'll close this PR to make clear the current change is not desired, but if you want to pivot to removing all `@error` highlights, you are welcome to force-push and I'll reopen. (A new PR might be simpler, though.)

PR #5424, opened 2023-09-19
feat(dart): Add base, final & sealed class modifier highlights

aidenlangley:


aidenlangley:
Just noticing this isn't quite right - closing, will re-open once ready.

ObserverOfTime (commented):
I don't know what "base" does but "sealed" should probably be `@type.qualifier`.

aidenlangley:
> I don't know what "base" does but "sealed" should probably be `@type.qualifier`.

I think you're right, that's more in-line with what's already going on.

ObserverOfTime (approved):

PR #5429, opened 2023-09-19
fix: use augroup for auto_install autocmds

wookayin:
Problem: The FileType autocmd to automatically install parsers on a new
filetype (introduced in #3130) did not have augroup. As a result, when
`ts_configs.setup{ }` is called multiple times (for some reason),
duplicated installation requests will be made.

Solution: Use augroup with `clear = true`.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -555,6 +555,7 @@ end
 function M.setup_auto_install()
   vim.api.nvim_create_autocmd("FileType", {
     pattern = { "*" },
+    group = vim.api.nvim_create_augroup("NvimTreesitter-auto_install", { clear = true }),
     callback = function()
       local lang = parsers.get_buf_lang()
       if parsers.get_parser_configs()[lang] and not is_installed(lang) and not is_ignored_parser(lang) then


lucario387 (approved):

lua/nvim-treesitter/install.lua
```diff
@@ -555,6 +555,7 @@ end
 function M.setup_auto_install()
   vim.api.nvim_create_autocmd("FileType", {
     pattern = { "*" },
+    group = vim.api.nvim_create_augroup('NvimTreesitter-auto_install', { clear = true }),
```
ObserverOfTime, line 558: stylua doesn't like the quotes and I don't like the name.

ObserverOfTime (changes requested):

lua/nvim-treesitter/install.lua
wookayin: I've fixed the stylua issue. The augroup name is chosen after `NvimTreesitter-<module_name>` structure. Do you have any other suggestion? I can't use `NvimTreesitter` because it's defined in `plugin/nvim-treesitter.lua`.

ObserverOfTime: The modules will be removed in 1.0.
You can use `group = "NvimTreesitter"`.

dundargoc:
@wookayin you didn't change the augroup name to "NvimTreesitter" as suggested. Is there a problem/conflict if you use that?

lua/nvim-treesitter/install.lua
wookayin: If we do not clear an own group upon `M.setup_auto_install()`, we still have the same problem: Call `setup_auto_install()` multiple times (from user config) and then multiple FileType events.

```diff
 function M.setup_auto_install()
   vim.api.nvim_create_autocmd("FileType", {
     pattern = { "*" },
-    group = vim.api.nvim_create_augroup('NvimTreesitter-auto_install', { clear = true }),
+    group = "NvimTreesitter",
     callback = function()
       ...
     end,
  })
 end
```

If we use `group = vim.api.nvim_create_augroup('NvimTreesitter', { clear = true })`, it will clear other autocmds under the group defined in `plugin/nvim-treesitter.lua`. So I think we should use a different augroup; what do you think?

wookayin:
I'm sorry this issue went out of my attention and has been stale; since some point I stopped using `auto_install`, but I believe this would still be a useful fix, despite the modules for 0.x being deprecated. Actually there were some issues with that simple (suggested) change, please see [the above comment](https://github.com/nvim-treesitter/nvim-treesitter/pull/5429#discussion_r1485206505).

ObserverOfTime (approved):

clason (approved):

PR #5433, opened 2023-09-21
Add liquidsoap.

toots:
This PR adds liquidsoap tree-sitter grammar and queries.

Liquidsoap is scripting language for creating media streams and more. It's got about [1.5k](https://github.com/search?type=code&q=NOT+is%3Afork+path%3A*.liq) script files on github.

Thanks for your consideration!


queries/liquidsoap/highlights.scm
```diff
@@ -0,0 +1,66 @@
...
+  "]"
+  "{"
+  "}"
+]  @punctuation.bracket
+
+(app name: (var) @function.call)
+(method) @method
+(method_app) @method.call
+
+(method) @function.method
```
lucario387, line 58: duplicate

queries/liquidsoap/locals.scm
```diff
@@ -0,0 +1,15 @@
...
+(labeled_argument label: (var) @local.parameter)
+
+(binding
+  defined: (var) @local.var)
+(def defined: (var) @local.var)
+(let defined: (var) @local.var)
+(meth_pattern (var) @local.var)
+(list_pattern (var) @local.var)
+(tuple_pattern (var) @local.var)
+(spread (var) @local.var)
```
toots, line 13:
```suggestion
(binding
  defined: (var) @definition.var)
(def defined: (var) @definition.var)
(let defined: (var) @definition.var)
(meth_pattern (var) @definition.var)
(list_pattern (var) @definition.var)
(tuple_pattern (var) @definition.var)
(spread (var) @definition.var)
```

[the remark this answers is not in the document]
toots: Fixed, thanks!

queries/liquidsoap/highlights.scm
```diff
@@ -0,0 +1,66 @@
...
+  "else"
+  "begin"
+  "end"
+  "while"
+  "do"
+  "to"
+  "for"
+  "fun"
+  "try"
+  "catch"
```
amaanq, line 20: these two could be `@exception`

queries/liquidsoap/highlights.scm
```diff
@@ -0,0 +1,66 @@
...
+"..." @keyword.operator
+
+[
+  "open"
+  "def"
+  "let"
+  "if"
+  "then"
+  "elsif"
+  "else"
```
amaanq, line 11: `@conditional`?

queries/liquidsoap/highlights.scm
```diff
@@ -0,0 +1,66 @@
+(op) @keyword.operator
+"..." @keyword.operator
+
+[
+  "open"
+  "def"
```
amaanq, line 6: `@keyword.function`?

queries/liquidsoap/highlights.scm
```diff
@@ -0,0 +1,66 @@
...
+  "if"
+  "then"
+  "elsif"
+  "else"
+  "begin"
+  "end"
+  "while"
+  "do"
+  "to"
+  "for"
```
amaanq, line 17: `@repeat`?

queries/liquidsoap/highlights.scm
```diff
@@ -0,0 +1,66 @@
...
+  "fun"
+  "try"
+  "catch"
+] @keyword
+
+["and" "or"] @conditional
```
amaanq, line 23: this looks like `@keyword.operator`

amaanq (changes requested):

queries/liquidsoap/highlights.scm
toots: Is `@exception` for raising calls? These are control statements, ended with and `end` so they should probably have the same color through and through?

The raising statement is `error.raise` but it can be overriden by a local variable. How does `locals` interact with `highlights` in the situations?

toots: Nevermind, I applied it and checked it looks great. Fixed!

queries/liquidsoap/highlights.scm
toots: Fixed!

queries/liquidsoap/highlights.scm
toots: Applied.

queries/liquidsoap/highlights.scm
toots: Applied!

queries/liquidsoap/highlights.scm
toots: Applied!

toots:
Thanks for the reviews y'all, this is much appreciated. All feedback has been checked and applied.

toots:
Hi! Is there any else that I can do to help moving things forward? Thanks for the help so far!

queries/liquidsoap/highlights.scm
```diff
@@ -0,0 +1,69 @@
+(op) @operator
+"..." @operator
+["and" "or" "not"] @keyword.operator
+
+[ "def"
+  "fun"
+  "let" ] @keyword.function
```
amaanq, line 7: Keep styling consistent, have brackets on their own line and use 2-space indentation for items

queries/liquidsoap/highlights.scm
```diff
@@ -0,0 +1,69 @@
...
+]  @punctuation.bracket
+
+(app name: (var) @function.call)
+(method) @method
+(method_app) @method.call
+
+(var) @variable
+(string) @string
+(integer) @number
+(float) @number
```
amaanq, line 65:
```suggestion
(integer) @number
(float) @float
```

queries/liquidsoap/locals.scm
amaanq, line 13: you never changed these to what they should be (in CONTRIBUTING.md) so CI is failing

amaanq (changes requested):

queries/liquidsoap/highlights.scm
toots: Done!

queries/liquidsoap/highlights.scm
toots: Good catch thanks.

queries/liquidsoap/locals.scm
toots: Ha darn sorry about that I think I did through GH and it got force-pushed out.

lua/nvim-treesitter/parsers.lua
```diff
@@ -973,6 +973,15 @@ list.llvm = {
   maintainers = { "@benwilliamgraham" },
 }
 
+list.liquidsoap = {
+  install_info = {
+    url = "https://github.com/savonet/tree-sitter-liquidsoap",
+    files = { "src/parser.c", "src/scanner.cc" },
+    branch = "main",
```
clason, line 980:
```suggestion
```

Not needed since it's the default branch.

lucario387 (commented):
Looking fine now.

Let's wait for @amaanq to give his opinion and then we can merge

toots:
Sounds good, thanks!

queries/liquidsoap/highlights.scm
```diff
@@ -0,0 +1,83 @@
+(op) @operator
+"..." @operator
```
lucario387, line 2: Not yet ig
```suggestion
[
  (op)
  "..."
] @operator
```

And why is `...` so special here. Can't the parser make it so it's part of `(op)`

toots: Indeed. Just pushed a change making it an `op`. It works great, `...` in types is not recognized as `@operator` now.

queries/liquidsoap/highlights.scm
```diff
@@ -8,6 +8,8 @@
   "def_end"
 ] @keyword.function
 
+(let_decoration) @keyword.coroutine
```
amaanq, line 11: doesn't look right to me but I could be wrong

lua/nvim-treesitter/parsers.lua
```diff
@@ -973,6 +973,14 @@ list.llvm = {
   maintainers = { "@benwilliamgraham" },
 }
 
+list.liquidsoap = {
+  install_info = {
+    url = "https://github.com/savonet/tree-sitter-liquidsoap",
+    files = { "src/parser.c", "src/scanner.cc" },
```
amaanq, line 979: not a blocker for this PR, but it'd be nice if you could rewrite the scanner in C

queries/liquidsoap/highlights.scm
```diff
@@ -0,0 +1,84 @@
+(op) @operator
+["and" "or" "not"] @keyword.operator
+
+[
+  "def"
+  "fun"
+  "let"
```
amaanq, line 7: is let just used to assign variables? or does it actually do something function-related

If the former it should just be `@keyword`

queries/liquidsoap/highlights.scm
```diff
@@ -0,0 +1,84 @@
...
+  "]"
+  "{"
+  "}"
+]  @punctuation.bracket
+
+(app name: (var) @function.call)
+(method) @method
+(method_app) @method.call
+
+(var) @variable
```
amaanq, line 76: Should be at the top, precedence is last wins so this will override function calls and parameters.

amaanq (changes requested):

queries/liquidsoap/highlights.scm
toots: Naming isn't great but it's correct.

lua/nvim-treesitter/parsers.lua
toots: I'm using the C++ regexp support in it. Waiting for this to switch to C: https://github.com/tree-sitter/tree-sitter/issues/2620

queries/liquidsoap/highlights.scm
toots: Done

queries/liquidsoap/highlights.scm
toots: Good to know thanks. Applied.

queries/liquidsoap/folds.scm
```diff
@@ -7,8 +7,9 @@
  (if)
  (if_condition)
  (if_then)
+ (elsif_condition)
+ (elseif_then)
```
lucario387, line 11: I don't see why there's a for `if_condition` and `elsif_condition` to be this specific? Isn't it still a normal expression?. Also typo probably :D

toots: Typo fixed, thanks.

The condition and body of each `then` clause can be full expressions with bindings, same expressions as the body of a function.
```liquidsoap
if
  x = f()
  x or g()
then
...
end

toots:
@amaanq all comments have been addressed!

queries/liquidsoap/folds.scm
lucario387: Well that's a surprising thing to learn then.

amaanq (approved):

toots:
Thanks for the help with this!

PR #5437, opened 2023-09-22
feat(nim): added nim parser and queries

aMOPel:
For [Nim parser](https://github.com/alaviss/tree-sitter-nim) added:
* highlights.scm
* locals.scm
* folds.scm
* injections.scm
    * regex and sql strings
    * emit pragma preceded by comment to specify language  
    * markdown in doc comments

Edit:
`indents.scm` seemed to be rather useless, since most syntax nodes are only parsed correctly, once the structure is complete, eg. `try:` needs `except:` etc.. Until then nodes can be merged in various ways with the previous ones and querying for every possibility seemed infeasible. Otherwise the indent behaviour would be very inconsistent, which provides little use for the end user. 
Edit End

And as a bonus for `std/strformat` format strings using [nim_format_string](https://github.com/aMOPel/tree-sitter-nim-format-string) parser:
* highlights.scm
* injections.scm

All testing has been performed locally and manually. 

Feedback is very welcome.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -362,6 +362,12 @@
   "nickel": {
     "revision": "b759233581fd8bae239e905c67a9ba453205da78"
   },
+  "nim": {
+    "revision": "35c0245439947e66239c1d59ea28b3fb0bba3d4c"
+  },
+  "nim_format_string": {
+    "revision": "b60f0e982a74fb5df93d26a468120cfe67e7e1af"
+  },
   "ninja": {
     "revision": "0a95cfdc0745b6ae82f60d3a339b37f19b7b9267"
   },
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1094,6 +1094,24 @@ list.nickel = {
   },
 }
 
+list.nim = {
+  install_info = {
+    url = "https://github.com/alaviss/tree-sitter-nim",
+    files = { "src/parser.c", "src/scanner.cc" },
+    cxx_standard = "c++14",
+    use_makefile = true,
+  },
+  maintainers = { "@aMOPel" },
+}
+
+list.nim_format_string = {
+  install_info = {
+    url = "https://github.com/aMOPel/tree-sitter-nim-format-string",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@aMOPel" },
+}
+
 list.ninja = {
   install_info = {
     url = "https://github.com/alemuller/tree-sitter-ninja",
--- a/queries/nim/highlights.scm
+++ b/queries/nim/highlights.scm
@@ -1,0 +1,690 @@
+; SPDX-FileCopyrightText: 2023 Leorize <leorize+oss@disroot.org>, aMOPel <>
+; SPDX-License-Identifier: MPL-2.0
+; SPDX-License-Identifier: Apache-2.0
+
+; =============================================================================
+; catch all rules
+
+((identifier) @variable (#set! "priority" 95))
+
+; =============================================================================
+; @comment               ; line and block comments
+
+[
+  (comment)
+  (block_comment)
+] @comment
+
+; =============================================================================
+; @comment.documentation ; comments documenting code
+
+[
+  (documentation_comment)
+  (block_documentation_comment)
+] @comment.documentation
+
+; =============================================================================
+; @none                  ; completely disable the highlight
+; unused
+
+; =============================================================================
+; @preproc               ; various preprocessor directives & shebangs
+
+(pragma_list ["{." "}" ".}"] @preproc)
+
+; =============================================================================
+; @define                ; preprocessor definition directives
+; unused
+
+; =============================================================================
+; @operator              ; symbolic operators (e.g. `+` / `*`)
+
+(operator) @operator
+
+; =============================================================================
+; @punctuation.delimiter ; delimiters (e.g. `;` / `.` / `,`)
+
+[ "." ";" "," ":" "=" ] @punctuation.delimiter
+
+; needs to be after @punctuation.delimiter
+(assignment "=" @operator)
+
+; =============================================================================
+; @punctuation.bracket   ; brackets (e.g. `()` / `{}` / `[]`)
+
+[ "(" ")" "[" "[:" "]" "{" "}" ] @punctuation.bracket
+
+; =============================================================================
+; @punctuation.special   ; special symbols (e.g. `{}` in string interpolation)
+
+(accent_quoted "`" @punctuation.special)
+
+(exported_symbol "*" @punctuation.special)
+
+; dereference operator
+(bracket_expression !right "[" @punctuation.special . "]" @punctuation.special)
+
+; =============================================================================
+; @string               ; string literals
+
+(string_literal) @string
+
+; =============================================================================
+; @string.documentation ; string documenting code (e.g. Python docstrings)
+; unused
+
+; =============================================================================
+; @string.regex         ; regular expressions
+; unused
+
+; =============================================================================
+; @string.escape        ; escape sequences
+
+(escape_sequence) @string.escape
+
+; =============================================================================
+; @string.special       ; other special strings (e.g. dates)
+; unused
+
+; =============================================================================
+; @character            ; character literals
+
+(char_literal) @character
+
+; =============================================================================
+; @character.special    ; special characters (e.g. wildcards)
+; unused
+
+; =============================================================================
+; @boolean              ; boolean literals
+
+((identifier) @boolean
+  (#any-of? @boolean "true" "false" "on" "off"))
+
+; =============================================================================
+; @number               ; numeric literals
+
+(integer_literal) @number
+
+(custom_numeric_literal) @number
+
+; =============================================================================
+; @float                ; floating-point number literals
+
+(float_literal) @float
+
+; =============================================================================
+; @function         ; function definitions
+
+(proc_declaration
+  name: [
+    (identifier) @function
+    (accent_quoted (identifier) @function)
+    (exported_symbol (identifier) @function)
+    (exported_symbol (accent_quoted (identifier) @function))
+  ])
+
+(func_declaration
+  name: [
+    (identifier) @function
+    (accent_quoted (identifier) @function)
+    (exported_symbol (identifier) @function)
+    (exported_symbol (accent_quoted (identifier) @function))
+  ])
+
+(iterator_declaration
+  name: [
+    (identifier) @function
+    (accent_quoted (identifier) @function)
+    (exported_symbol (identifier) @function)
+    (exported_symbol (accent_quoted (identifier) @function))
+  ])
+
+(converter_declaration
+  name: [
+    (identifier) @function
+    (accent_quoted (identifier) @function)
+    (exported_symbol (identifier) @function)
+    (exported_symbol (accent_quoted (identifier) @function))
+  ])
+
+; =============================================================================
+; @function.call    ; function calls
+
+(call
+  function: [
+    (identifier) @function.call
+    (accent_quoted (identifier) @function.call)
+    ; generic types
+    (bracket_expression left: (identifier) @function.call)
+    (bracket_expression left: (accent_quoted (identifier) @function.call))
+    ; dot accessor
+    (dot_expression right: (identifier) @function.call)
+    (dot_expression right: (accent_quoted (identifier) @function.call))
+    ; both
+    (bracket_expression left:
+      (dot_expression right: (identifier) @function.call))
+    (bracket_expression left:
+      (dot_expression right: (accent_quoted (identifier) @function.call)))
+  ])
+
+; generalized_string is a function call
+; `identifier"string literal"`
+; is short for
+; `identifier(r"string literal")`
+(generalized_string . [
+  (identifier) @function.call
+  (accent_quoted (identifier) @function.call)
+])
+
+; =============================================================================
+; @function.builtin ; built-in functions
+; unused
+
+; =============================================================================
+; @function.macro   ; preprocessor macros
+
+(template_declaration
+  name: [
+    (identifier) @function
+    (accent_quoted (identifier) @function)
+    (exported_symbol (identifier) @function)
+    (exported_symbol (accent_quoted (identifier) @function))
+  ])
+
+(macro_declaration
+  name: [
+    (identifier) @function
+    (accent_quoted (identifier) @function)
+    (exported_symbol (identifier) @function)
+    (exported_symbol (accent_quoted (identifier) @function))
+  ])
+
+; =============================================================================
+; @method           ; method definitions
+
+(method_declaration
+  name: [
+    (identifier) @method
+    (accent_quoted (identifier) @method)
+    (exported_symbol (identifier) @method)
+    (exported_symbol (accent_quoted (identifier) @method))
+  ])
+
+; =============================================================================
+; @method.call      ; method calls
+; unused
+
+; =============================================================================
+; @constructor      ; constructor calls and definitions
+
+(call
+  function: [
+    (identifier) @constructor
+    (accent_quoted (identifier) @constructor)
+    ; generic types
+    (bracket_expression left: (identifier) @constructor)
+    (bracket_expression left: (accent_quoted (identifier) @constructor))
+    ; dot accessor
+    (dot_expression right: (identifier) @constructor)
+    (dot_expression right: (accent_quoted (identifier) @constructor))
+    ; both
+    (bracket_expression left:
+      (dot_expression right: (identifier) @constructor))
+    (bracket_expression left:
+      (dot_expression right: (accent_quoted (identifier) @constructor)))
+  ]
+  (argument_list
+    (colon_expression)+))
+; NOTE: this cannot detect constructors with 0 arguments
+; those will be matched as @function.call instead
+
+; =============================================================================
+; @parameter        ; parameters of a function
+
+; named parameters when calling
+; call(parameter_name=arg)
+(argument_list
+  (equal_expression
+    left: [
+      (identifier) @parameter
+      (accent_quoted (identifier) @parameter)
+    ]))
+
+; parameters in function declaration
+(parameter_declaration_list
+  (parameter_declaration
+    (symbol_declaration_list
+      (symbol_declaration
+        name: [
+          (identifier) @parameter
+          (accent_quoted (identifier) @parameter)
+        ]))))
+
+; =============================================================================
+; @keyword             ; various keywords
+
+; unhandled but reserved keywords
+; end
+; interface
+
+; static expression
+; addr operator
+((call
+  function: (identifier) @keyword)
+  (#any-of? @keyword "static" "addr"))
+
+[
+  "const"
+  "let"
+  "var"
+  "type"
+  "concept"
+  "asm"
+  "bind"
+  "defer"
+  "do"
+  "mixin"
+  "static"
+  "object"
+  "tuple"
+  "enum"
+  "block"
+  "using"
+  "discard"
+] @keyword
+
+; =============================================================================
+; @keyword.coroutine   ; keywords related to coroutines (e.g. `go` in Go, `async/await` in Python)
+; unused
+
+; =============================================================================
+; @keyword.function    ; keywords that define a function (e.g. `func` in Go, `def` in Python)
+
+[
+  "proc"
+  "func"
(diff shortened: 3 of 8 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -1094,6 +1094,23 @@ list.nickel = {
   },
 }
 
+list.nim = {
+  install_info = {
+    url = "https://github.com/alaviss/tree-sitter-nim",
+    files = { "src/parser.c", "src/scanner.cc" },
+    branch = "main",
```
clason, line 1101:
```suggestion
```
Not needed if it's the default branch; you do need to add a lockfile entry for the parsers, though.

queries/nim/highlights.scm
```diff
@@ -0,0 +1,689 @@
+; SPDX-FileCopyrightText: 2023 Leorize <leorize+oss@disroot.org>, aMOPel <>
```
clason, line 1: Do you insist on this copyright statement? We'd prefer all queries to be available under the same license.

queries/nim/highlights.scm
```diff
@@ -0,0 +1,689 @@
...
+
+[
+  (documentation_comment)
+  (block_documentation_comment)
+] @comment.documentation
+
+; =============================================================================
+; @error                 ; syntax/parser errors
+
+(ERROR) @error
```
clason, line 28: move to the end

queries/nim/highlights.scm
aMOPel: @alaviss that is your copyright, right?

alaviss: I'd prefer having the MPL license kept for easier reintegration back into upstream.

Since @aMOPel wrote a lot of extra code for this, they can dual-license this file with `Apache-2.0` if they wish to do so.

alaviss:
> `indents.scm` seemed to be rather useless, since most syntax nodes are only parsed correctly, once the structure is complete, eg. `try:` needs `except:` etc.. Until then nodes can be merged in various ways with the previous ones and querying for every possibility seemed infeasible. Otherwise the indent behaviour would be very inconsistent, which provides little use for the end user.

The tokens can still be captured, like this `(ERROR "try")` which you could use to push an indent.

Many other nodes like `proc`, `if`, etc. can also have indentation trivially done because they're tolerant to missing nodes. The grammar allow for their body to be empty, similar to how the python grammar works.

lockfile.json
```diff
@@ -362,6 +362,12 @@
   "nickel": {
     "revision": "b759233581fd8bae239e905c67a9ba453205da78"
   },
+  "nim": {
+    "revision": "388b53f808eb162963cfe99bd2d1ed83c8b5f8a1"
```
alaviss, line 366: If you don't mind, please bump this to `35c0245439947e66239c1d59ea28b3fb0bba3d4c`.

I just tagged 0.1.0 for this occasion :partying_face:

clason: This file is autonated, so the revision here is not important (as long as the queries match). The only important thing for CI at this point is that there _is_ an entry.

queries/nim/highlights.scm
aMOPel: How would I do that?

alaviss: Add `; SPDX-License-Identifier: Apache-2.0` to the file

aMOPel: done

aMOPel:
> The tokens can still be captured, like this (ERROR "try") which you could use to push an indent.
> 
> Many other nodes like proc, if, etc. can also have indentation trivially done because they're tolerant to missing nodes. The grammar allow for their body to be empty, similar to how the python grammar works.

@alaviss I tried stuff like that.

The thing is that the AST goes wild as you type, eg

```nim
type
  Obj = object
    case x: int:
    of 5:
      b: int
```

```scheme
type_section [0, 0] - [6, 0]
  "type" [0, 0] - [0, 4]
  type_declaration [1, 2] - [6, 0]
    type_symbol_declaration [1, 2] - [1, 5]
      name: identifier [1, 2] - [1, 5]
    "=" [1, 6] - [1, 7]
    object_declaration [1, 8] - [6, 0]
      "object" [1, 8] - [1, 14]
      field_declaration_list [2, 4] - [6, 0]
        variant_declaration [2, 4] - [6, 0]
          "case" [2, 4] - [2, 8]
          variant_discriminator_declaration [2, 9] - [2, 15]
            symbol_declaration_list [2, 9] - [2, 10]
              symbol_declaration [2, 9] - [2, 10]
                name: identifier [2, 9] - [2, 10]
            type: ":" [2, 10] - [2, 11]
            type: type_expression [2, 12] - [2, 15]
              identifier [2, 12] - [2, 15]
          ":" [2, 15] - [2, 16]
          of_branch [3, 4] - [6, 0]
            "of" [3, 4] - [3, 6]
            values: expression_list [3, 7] - [3, 8]
              integer_literal [3, 7] - [3, 8]
            ":" [3, 8] - [3, 9]
            field_declaration_list [4, 6] - [6, 0]
              field_declaration [4, 6] - [4, 12]
                symbol_declaration_list [4, 6] - [4, 7]
                  symbol_declaration [4, 6] - [4, 7]
                    name: identifier [4, 6] - [4, 7]
                type: ":" [4, 7] - [4, 8]
                type: type_expression [4, 9] - [4, 12]
                  identifier [4, 9] - [4, 12]
```
```nim
type
  Obj = object
    case x: int:
    of 5:
      b: int
    of 6:               # <- added line
```

```scheme
ERROR [0, 0] - [7, 0]
  "type" [0, 0] - [0, 4]
  type_symbol_declaration [1, 2] - [1, 5]
    name: identifier [1, 2] - [1, 5]
  "=" [1, 6] - [1, 7]
  "object" [1, 8] - [1, 14]
  "case" [2, 4] - [2, 8]
  variant_discriminator_declaration [2, 9] - [2, 15]
    symbol_declaration_list [2, 9] - [2, 10]
      symbol_declaration [2, 9] - [2, 10]
        name: identifier [2, 9] - [2, 10]
    type: ":" [2, 10] - [2, 11]
    type: type_expression [2, 12] - [2, 15]
      identifier [2, 12] - [2, 15]
  ":" [2, 15] - [2, 16]
  of_branch [3, 4] - [5, 4]
    "of" [3, 4] - [3, 6]
    values: expression_list [3, 7] - [3, 8]
      integer_literal [3, 7] - [3, 8]
    ":" [3, 8] - [3, 9]
    field_declaration_list [4, 6] - [5, 4]
      field_declaration [4, 6] - [4, 12]
        symbol_declaration_list [4, 6] - [4, 7]
          symbol_declaration [4, 6] - [4, 7]
            name: identifier [4, 6] - [4, 7]
        type: ":" [4, 7] - [4, 8]
        type: type_expression [4, 9] - [4, 12]
          identifier [4, 9] - [4, 12]
  "of" [5, 4] - [5, 6]
  expression_list [5, 7] - [5, 8]
    integer_literal [5, 7] - [5, 8]
  ":" [5, 8] - [5, 9]
```

To indent correctly now, I just need to query something like:
```scheme
((ERROR "of" ":") @indent.begin (#set! indent.immediate 1))
```
right? Oh well. It captures the `@indent.begin` correctly, however because the AST is so messed up, all the previous indents
are incorrect, so the sum of the indents end up not being correct most of the time.

Stuff like this happens all the over the place and the tree can vary widely depending on preceding nodes in the AST.

I can write queries that work alright for some cases, but the end result is just gonna be inconsistent and I personally would prefer manual indenting over inconsistent autoindenting.

alaviss:
> Stuff like this happens all the over the place and the tree can vary widely depending on preceding nodes in the AST.

If you can nail down the weird ones I can fix them.

The problem with that `of` is just I forgot to make an empty branch valid, I think.

alaviss:
I pushed a few fixes wrt error recovery in case you wanted to try again with the indentation queries.

If you found error cases where the tree structure with regards to indentation would be thrown out of whack, feel free to file issues and I'll see if it's possible to make it better.

With https://github.com/alaviss/tree-sitter-nim/pull/40 merged I think you should see less broken trees during errors.

aMOPel:
Great! I'll give it another shot.

aMOPel:
@alaviss the CI broke when testing on MacOS. Any ideas?

theHamsta:
> @alaviss the CI broke when testing on MacOS. Any ideas?

Apple is quite conservative about C++ standards. You might be using C++14 or C++17 in your https://github.com/aMOPel/tree-sitter-nim/blob/main/src/scanner.cc . Can you try setting a C++ standard and set `use_makefile = true`? This solved the Mac problems for other parsers https://github.com/nvim-treesitter/nvim-treesitter/blob/b7f2dd5dfbd24a1239844e15637b637b990df164/lua/nvim-treesitter/parsers.lua#L1113-L1121

clason:
Or, you know, just rewrite the scanner in C, like many other parsers have been doing ;) @amaanq

theHamsta:
> Or, you know, just rewrite the scanner in C, like many other parsers have been doing ;) @amaanq

or just don't use MacOS

alaviss:
> @alaviss the CI broke when testing on MacOS. Any ideas?

The scanner is written with C++14 and macOS somehow can't update from their decades old C++03 default.

clason:
Just a heads-up: we have removed these options (and the norg parser) in the nvim-treesitter 1.0 rewrite. This means even if we merge this parser now, it will be removed again with the 1.0 release.

Even upstream is now pushing towards pure C scanners, for portability reasons.

aMOPel:
Oh well. I would still like to merge this pr, then users have the option to use the parser and not upgrade to 1.0.

How far is that release off?

clason:
> How far is that release off?

Shortly after the Neovim 0.10 release, so in the next few months.

> Oh well. I would still like to merge this pr, then users have the option to use the parser and not upgrade to 1.0.

That's a bit cheeky, assuming that your parser is somehow more important than a host of performance and usability improvements ;)

No, if C++14 is somehow a critical necessity, distribute the parser via a custom plugin with custom build steps, just like norg does. Neovim doesn't care _how_ the parser and queries end up on runtimepath.

aMOPel:
> assuming that your parser is somehow more important than a host of performance and usability improvements

You're the one who is assuming what is best for the user. I'm proposing to leave the decision to the user. There a quite a few nvim users who don't get the cutting edge stuff because it means to maintain your nvim config. 

Also I don't see the loss for this repo, as there are already a huge amount of parsers merged, that aren't using pure C scanners. So there is gonna be a big cleanup anyway, if I understand correctly.

Either way, if you enforce this constraint you should definitely update `CONTRIBUTING.md` to reflect that.

clason:
> Also I don't see the loss for this repo, as there are already a huge amount of parsers merged, that aren't using pure C scanners. So there is gonna be a big cleanup anyway, if I understand correctly.

It's not about pure C -- C++ is fine -- as about using specific features of newer standards that affect portability. This is only the second parser ever to make use of that (and the first has a dedicated plugin, so doesn't need to be bundled here), so this has not been an issue before.

Yes. CONTRIBUTING.md will be updated.

aMOPel:
I see.

alaviss:
> as about using specific features of newer standards that affect portability

Every single compiler save for Xcode now handles C++14 by default which is almost a decade old standard. Not having support for that is extremely limiting.

IMO, I think `nvim-treesitter` should just append `-std=c++14` to every C++ invocation to even out the compatibility between different platforms.

clason:
> Every single compiler save for Xcode now handles C++14 by default which is almost a decade old standard. Not having support for that is extremely limiting.

Gotta bring this up with Apple.

> IMO, I think nvim-treesitter should just append -std=c++14 to every C++ invocation to even out the compatibility between different platforms.

And you are 100% sure this will not break other platforms, even Windows or BSD or stock gcc on Debian 10? If it was that simple, we wouldn't care.

alaviss:
> And you are 100% sure this will not break other platforms, even Windows or BSD or stock gcc on Debian 10? If it was that simple, we wouldn't care.

Helix seems to be pretty happy with that: https://github.com/helix-editor/helix/blob/7702e130ba93f885ac56be5818b93c544a00538f/helix-loader/src/grammar.rs#L452-L554

Compiler wise:
- gcc: C++14 by [default since version 6.0](https://gcc.gnu.org/projects/cxx-status.html), fully feature completed since 5.0.
- clang: C++14 support [completed by version 3.4](https://clang.llvm.org/cxx_status.html#cxx14), default by 6.0.
- MSVC: C++14 is default since MSVC 2015, later versions doesn't even support pre-C++14: https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-140

And here's the distro compatibility list list:

- Debian 10 (2019): [gcc 8.x](https://packages.debian.org/buster/gcc) :heavy_check_mark: 
- Debian 9 (2017): [gcc 6.x](https://www.debian.org/releases/stretch/amd64/release-notes.en.txt) :heavy_check_mark:
- RHEL 6 with developer toolset version 4 (2015): [gcc 5.x](https://access.redhat.com/solutions/19458) :heavy_check_mark:
- FreeBSD 10.1 (2014): [clang 3.4.1](https://www.freebsd.org/releases/10.1R/relnotes/) :heavy_check_mark:
- DragonflyBSD 4.2 (2015): [gcc 5.1](https://www.dragonflybsd.org/release42/) :heavy_check_mark:
- OpenBSD 5.6 (2014): [clang 3.5](https://www.openbsd.org/56.html) :heavy_check_mark:
- NetBSD 8.0 (2018): [gcc 5.5](https://www.netbsd.org/releases/formal-8/NetBSD-8.0.html) :heavy_check_mark:
- Windows: MSVC 2015+ :heavy_check_mark:
- Mac OSX 10.9 (2016): [Xcode 6.0](https://developer.apple.com/library/archive/documentation/Xcode/Conceptual/WhatsNewXcode-Archive/Articles/xcode_6_0.html#//apple_ref/doc/uid/TP40014509-SW1) :heavy_check_mark:

So yes, I do believe that it will not break any other platforms.

aMOPel:
I was thinking:

```
     cxx_standard = "c++14", 
     use_makefile = true, 
```

Even if you're gonna remove those flags in the upcoming release, the CI tests for everything except MacOS worked fine, even before I added those flags. 

Can't we just add an entry to the "supported languages" readme section for the Nim where it says 

"except MacOS < 2016 / XCode < 6.0" 

Or something similar?

clason:
> So yes, I do believe that it will not break any other platforms.

Have you actually tested this? (Spoiler: you can't do that with mixed language compilation.)

> Can't we just add an entry to the "supported languages" readme section for Nim where it says

> "except MacOS < 2016 / XCode < 6.0"

> Or something similar?

You think we would drop a mainstream platform (or add yet more conditional compilation logic) in favor of a single parser? (This affects _current_ macOS, remember, not some legacy version you just invented.)

aMOPel:
Bummer. Separate plugin it is then I guess. 

If I run into issues I can just ask on the matrix server right?

clason:
> If I run into issues I can just ask on the matrix server right?

Sure! Sorry it didn't work out and you got stuck between a rock and a hard place...

aMOPel:
I made the plugin now:

https://github.com/aMOPel/nvim-treesitter-nim

Does the structure look good to you?

clason:
I'm not happy about that readme, but it's your decision. 

(The parser works just fine, it just cannot be compiled by a simple enough (for us, who have to maintain it across all platforms) invocation.)

clason:
And the whole point of the plugin is _not_ to depend on nvim-treesitter. Just steal our makefile and call that, and add an ftplugin that calls vim.treesitter.start.

aMOPel:
> And the whole point of the plugin is not to depend on nvim-treesitter

But I do depend on it. At the very least I am using `has-ancestor` (which I benchmarked and it is indeed sensible to use).

clason:
You can just define this in your plugin (with override = true), you know. Or wait for Neovim 0.10, which will have this builtin.

We are actively working on giving plugins zero reason for depending on nvim-treesitter (because they shouldn't).

aMOPel:
> I'm not happy about that readme

I (hopefully) corrected it.

aMOPel:
> Just steal our makefile

I'm a little lost on the licensing aspect of it.

My repo is also under apache-2.0. When I take code from this repo, is it sufficient, if I just give credit, i.e. write in a comment "from https://github.com/nvim-treesitter/nvim-treesitter" or what's the proper procedure here?

And when I then modify the code that has to be stated, too, correct?

lucario387:
There's actually a total of zero nim query code that's part of nvim-treesitter since it's still on your branch.

As for how it should be done, something like how Helix did in the past is fine I think.

> Borrowed/Referenced from nvim-treesitter

aMOPel:
> There's actually a total of zero nim query code that's part of nvim-treesitter since it's still on your branch.

No need to rub it in :D

It's not about the Nim query code, it's about your lua code and the one makefile. 

> the whole point of the plugin is not to depend on nvim-treesitter

When I don't depend on nvim-tresitter, then I need to have my own installation/uninstallation logic after all. I'm gonna borrow lua code from this repo heavily for that.

@clason Yes. I don't know what I was thinking. Of course I need my own installation logic since the whole makefile thing is gonna be dropped in this repo. Sorry.

> As for how it should be done, something like how Helix did in the past is fine I think.

> Borrowed/Referenced from nvim-treesitter

Thanks

clason:
> @clason Yes. I don't know what I was thinking. Of course I need my own installation logic since the whole makefile thing is gonna be dropped in this repo. Sorry.

No worries. You should be able to simplify this a lot since you don't need to support all manner of parsers, and will likely want to force using the Makefile.

Protip: if the _parser_  includes a proper Makefile, you can (and should) just rely on that. If every parser did that, nvim-treesitter would be much simpler (and weird scanners not a breaking deal).

Yes, you can just take the files you need, with blessing; adding 
```lua
-- (Adapted) from https://github.com/nvim-treesitter/nvim-treesitter
```
at the top of the respective files is fine.

(And you're always welcome back, if the situation changes. After all, no one should be writing a C++ scanner; parsers have been switching to pure C scanners since forever.)

aMOPel:
Is there some reason for me to keep 

* the download method with [tar and curl](https://github.com/aMOPel/nvim-treesitter/blob/add_nim_pr/lua/nvim-treesitter/shell_command_selectors.lua#L230)?
* the [`with_sync` alternatives](https://github.com/aMOPel/nvim-treesitter/blob/add_nim_pr/lua/nvim-treesitter/install.lua#L742)?

clason:
Your choice; tar and curl is much quicker but not always available; and people clamor for sync variants because they insist on installing/updating everything in a cronjob.

aMOPel:
> Protip: if the parser includes a proper Makefile, you can (and should) just rely on that. If every parser did that, nvim-treesitter would be much simpler (and weird scanners not a breaking deal).

Wait a minute. Does that mean, alaviss could just make a makefile for the parser with the c++14 flag and all would be good?

clason:
Not for us, since we don't use Makefiles (as they are not widely used and cannot be relied on). But for you who control both parser and plugin, you can of course do that.

aMOPel:
Ah. Bummer.

aMOPel:
But would that be a big deal to support? Naively I'm thinking it's just a few lines in the install logic to say "if there is a repo makefile present, use that". What am I missing?

clason:
It adds a lot of complexity (for a single upstream user). Makefiles aren't as portable as you would like to believe (just one example: `make` vs. `gmake`; never underestimate Windows' power to mess things up).

If they'd be widely used and standardized so we could blindly rely on them, that would be another thing. But this is not the world we live in. (The world I _want_ to live in is one where we can just punt to `tree-sitter build` for snowflake parsers.)

aMOPel:
Hm. So parsers would have to make a makefile and whatever `gmake` needs to support windows too, and then it would still create a hard dependency for the user to have make/gmake which is currently not the case.

clason:
Yes, exactly. We have to walk a tightrope here between requiring additional build tools at runtime and convenience (user as well as maintainer). We cannot do without a compiler (obviously) and -- for some but not all parsers -- tree-sitter CLI. Anything more would be a hard sell, especially if only a single parser requires it.

aMOPel:
Turns out @alaviss was convinced by tree-sitter upstream to convert the parser to pure c eventually. So I'm not gonna develop the independent plugin further. As a temporary measure it suffices.

https://github.com/alaviss/tree-sitter-nim/issues/44

Gonna come back and terrorize you with more nim queries when that is done :)

PR #5440, opened 2023-09-23
feat(julia): added missing String and AbstractString builtin types

lfenzo:
cc.: @savq


savq:
LGTM.

savq:
I should mention that the list still has some missing stuff.

In #4474, I wrote that it'd be a good idea to generate this list automatically. Although my first attempt was incomplete because I didn't include parametric types (`UnionAll`).

The following program will print all public built-in types:

```julia
istype(x) = typeof(x) === DataType || typeof(x) === UnionAll
get_types(m) = filter(x -> istype(Base.eval(m, x)), names(m))
type_names = sort(union(get_types(Core), get_types(Base)))
foreach(t -> println("\"$t\""), type_names)
```

The current list has some private types, so the generated list isn't a strict superset.

lucario387:
Do you want to add the other default builtin types to this PR as well @lfenzo.

If yes, then can you also add the code to print out all those types as comments in the query file?

lfenzo:
I've compared the built-in types in the current `julia/highlights.scm` file to the types returned by Sergio's script. In short, (see collapsable sections for details of each group):
- Public, but not highlighted: 45 types
- Highlighted, but not public: 23 types

I think it's fair to go with the union of these two sets of types; this increases the number of types in under `@type.builtin` from 177 to 222 and incorporates 100% of available public built-in types in `Core` and `Base`. For the private types, I believe we should keep the one we have for now and add others based on demand later on.

As suggested by @lucario387, I'll add this script as a comment in the `highlights.scm` file, thanks @savq for providing the snippet.

<details open>
  <summary>Public, but not highlighted</summary>

```
 45-element Vector{Symbol}:
 :ConcurrencyViolationError
 :Cvoid
 :Exception
 :GlobalRef
 :IO
 :LineNumberNode
 :Method
 :SegmentationFault
 :TypeVar
 :UInt
 :Unsigned
 :VecElement
 :WeakRef
 :(=>)
 :AbstractChannel
 :AbstractMatch
 :AbstractPattern
 :AbstractSlices
 :BitMatrix
 :BitVector
 :CanonicalIndexError
 :CapturedException
 :CartesianIndex
 :CartesianIndices
 :Colon
 :ColumnSlices
 :ComplexF16
 :ComplexF32
 :ComplexF64
 :HTML
 :IOContext
 :IndexCartesian
 :IndexLinear
 :IndexStyle
 :InvalidStateException
 :LazyString
 :LinearIndices
 :MissingException
 :PermutedDimsArray
 :Pipe
 :Returns
 :RowSlices
 :Slices
 :TaskFailedException
 :Text
```
</details>

<details open>
  <summary>Highlighted, but not public</summary>

```
25-element Vector{Symbol}:
 :Fix1
 :Fix2
 :EnvDict
 :AsyncCondition
 :ParseError
 :IteratorSize
 :IteratorEltype
 :ImmutableDict
 :Pairs
 :OneTo
 :Event
 :Semaphore
 :AbstractLniock
 :Atomic
 :SpinLock
 :Ordering
 :ReverseOrdering
 :By
 :Lt
 :Perm
 :Stateful
 :CFunction
 :Tmstruct
 :StackFrame
 :StackTrace
```
</details>

queries/julia/highlights.scm
```diff
@@ -114,35 +114,236 @@
...
-  "Vector" "Matrix" "VecOrMat" "DenseArray" "DenseVector" "DenseMatrix" "DenseVecOrMat"
-  "StridedArray" "StridedVector" "StridedMatrix" "StridedVecOrMat" "BitArray" "Dims"
-  "SubArray" "Task" "Condition" "Event" "Semaphore" "AbstractLniock" "ReentrantLock"
-  "Channel" "Atomic" "SpinLock" "RawFD" "IOStream" "IOBuffer" "AbstractDisplay" "MIME"
-  "TextDisplay" "PartialQuickSort" "Ordering" "ReverseOrdering" "By" "Lt" "Perm"
-  "Stateful" "CFunction" "Ptr" "Ref" "Cchar" "Cuchar" "Cshort" "Cstring" "Cushort"
-  "Cint" "Cuint" "Clong" "Culong" "Clonglong" "Culonglong" "Cintmax_t" "Cuintmax_t"
-  "Csize_t" "Cssize_t" "Cptrdiff_t" "Cwchar_t" "Cwstring" "Cfloat" "Cdouble" "Tmstruct"
-  "StackFrame" "StackTrace"))
+  "=>"
```
lfenzo, line 125: The `=>` was returned as a type, but I'm not sure if we should leave it here. Only saw it after submitting the commit...

lucario387: We should remove it. That doesn't look like a type to me (or is it?)

savq: `=>` is an alias for `Pair`, which is already listed. It can be removed.

savq:
There's some types in the current list that are private and also aren't part of either Core or Base, but rather submodules of Base:
```julia
Order => [:Ordering, :ReverseOrdering, :By, :Lt, :Perm]
Iterators => [:Stateful]
StackTraces => [:StackFrame, :StackTrace]
Threads => [:SpinLock, :Atomic]
Meta => [:ParseError]
Libc => [:Tmstruct]
```

Also, `AbstractLniock` is just a typo of `AbstractLock`.

That leaves a much smaller list of private types:
```julia
:Fix1
:Fix2
:EnvDict
:AsyncCondition
:IteratorSize
:IteratorEltype
:ImmutableDict
:Pairs
:OneTo
:Event
:Semaphore
:CFunction
```

I know the semantics of what is a built-in type are not super clear, but I think we simply shouldn't include private types on the list. They're very few, and rarely used explicitly as constructors (which is kinda the whole point of the list, because the types will be highlighted correctly in declarations and function signatures).

lfenzo:
Private types removed!

lucario387 (approved):

savq:
Thanks for the PR @lfenzo . Sorry for being so nitpicky!

lucario387:
That's the last confirmation I need to merge.
Thanks @lfenzo and @savq!

lfenzo:
Sure! Thanks @savq and @lucario387 for the comments and the effort!

PR #5450, opened 2023-09-26
feat(highlights): introduce @type.definition to java highlights

heimonsy:
RT.


amaanq:
That's not what `@type.definition` is for, it's more for C-like typedefs

heimonsy:
> That's not what `@type.definition` is for, it's more for C-like typedefs

I see, thanks.

PR #5451, opened 2023-09-26
feat(highlights): some haskell additions

mrcjkb:
- Re-add (module) @namespace
- Exception handling
- Debugging
- `otherwise` = boolean `True`
- `qq` string quasiquotes
- Documentation comments (draft)
- Function/lambda parameters
- Remove recently added @ (already defined as an operator)

### DIFF
--- a/queries/haskell/highlights.scm
+++ b/queries/haskell/highlights.scm
@@ -1,4 +1,16 @@
 ;; ----------------------------------------------------------------------------
+;; Parameters
+
+;; NOTE: These are at the top, so that they have low priority,
+;; and don't override destructured parameters 
+
+(function
+  patterns: (patterns (pat_name) @parameter))
+
+(exp_lambda
+  (pat_name) @parameter)
+
+;; ----------------------------------------------------------------------------
 ;; Literals and comments
 
 (integer) @number
@@ -11,6 +23,31 @@
 
 (comment) @comment
 
+; FIXME: The below documentation comment queries are inefficient
+; and need to be anchored, using something like
+; ((comment) @_first . (comment)+ @comment.documentation)
+; once https://github.com/neovim/neovim/pull/24738 has been merged.
+;
+; ((comment) @comment.documentation 
+;   (#lua-match? @comment.documentation "^-- |"))
+;
+; ((comment) @_first @comment.documentation 
+;  (comment) @comment.documentation 
+;   (#lua-match? @_first "^-- |"))
+;
+; ((comment) @comment.documentation 
+;   (#lua-match? @comment.documentation "^-- %^"))
+;
+; ((comment) @_first @comment.documentation 
+;  (comment) @comment.documentation 
+;   (#lua-match? @_first "^-- %^"))
+;
+; ((comment) @comment.documentation 
+;   (#lua-match? @comment.documentation "^{-"))
+;
+; ((comment) @_first @comment.documentation 
+;  (comment) @comment.documentation 
+;   (#lua-match? @_first "^{-"))
 
 ;; ----------------------------------------------------------------------------
 ;; Punctuation
@@ -77,6 +114,8 @@
   "@"
 ] @operator
 
+
+(module) @namespace
 (qualified_module (module) @constructor)
 (qualified_type (module) @namespace)
 (qualified_variable (module) @namespace)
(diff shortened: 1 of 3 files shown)


queries/haskell/highlights.scm
```diff
@@ -11,6 +11,26 @@
...
+((comment) @_first @comment.documentation (#set! "priority" 101)
+ (comment) @comment.documentation 
+  (#lua-match? @_first "^-- |"))
+
+((comment) @comment.documentation (#set! "priority" 101)
+  (#lua-match? @comment.documentation "^-- %^"))
+
+((comment) @_first @comment.documentation (#set! "priority" 101)
+ (comment) @comment.documentation 
+  (#lua-match? @_first "^-- %^"))
```
lucario387, line 26: This is not good.
For a chain of n comments node, the first will be matched n times, the second n-1 times, the third n-2 ... -> n^2 matches for n nodes.

Not to mention, this will also be matched by cases like this

```
(comment)
(n_random_arbitrary_nodes)
(comment)
```

Something like this is better
```
((comment) @_first . (comment)+ @comment.documentation)
```

queries/haskell/highlights.scm
```diff
@@ -150,13 +176,77 @@
 
 ; True or False
 ((constructor) @boolean (#any-of? @boolean "True" "False"))
-
+; otherwise (= True)
+((variable) @boolean (#set! "priority" 101)
```
lucario387, line 180: Is it a must for this to have priorities? Can't it be moved down a bit for nvim queries  since ours is "last wins"

lucario387 (changes requested):
I believe most of the `(#set! priority)` are unnecessary. Consider removing them alongside moving the queries further down the files, or some reordering to remove the need for those priorities setting

queries/haskell/highlights.scm
mrcjkb: :thinking: if I anchor with `(comment)+` or `(comment)*`, it only captures the last `comment`.
I'll have to revisit this later or remove them.

lucario387: > 🤔 if I anchor with `(comment)+` or `(comment)*`, it only captures the last `comment`. I'll have to revisit this later or remove them.

It's an issue with neovim (neovim/neovim#24738). But yes, the comment queries you provided here will lead to performance issues, unless it's anchored and quantified.

Better to comment them out and re-look at a later date

queries/haskell/highlights.scm
mrcjkb: You're right, this one (and the equivalent ones) are redundant.
Not sure why they weren't working for me yesterday.

mrcjkb:
I've moved the `@parameter` queries to the top, so they have a lower priority than destructured patterns.

lucario387 (approved):

clason:
Haskell is currently lacking a maintainer -- would you be interested in filling that role?

mrcjkb:
> Haskell is currently lacking a maintainer -- would you be interested in filling that role?

Sure :smile: 

I was thinking of adding some tests when I find time.

clason:
Then you could add yourself as maintainer in `parsers.lua` (maybe separate commit in this PR) :)

queries/haskell/highlights.scm
```diff
@@ -11,6 +23,31 @@
 
 (comment) @comment
 
+; FIXME: The below documentation comment queries are inefficient
+; and need to be anchored, using something like
+; ((comment) @_first . (comment)+ @comment.documentation)
+; once https://github.com/neovim/neovim/pull/24738 has been merged.
+;
+; ((comment) @comment.documentation 
+;   (#lua-match? @comment.documentation "^-- |"))
```
lucario387, line 32: This query is fine, you can add it if you want.

queries/haskell/highlights.scm
```diff
@@ -11,6 +23,31 @@
...
+;
+; ((comment) @comment.documentation 
+;   (#lua-match? @comment.documentation "^-- %^"))
+;
+; ((comment) @_first @comment.documentation 
+;  (comment) @comment.documentation 
+;   (#lua-match? @_first "^-- %^"))
+;
+; ((comment) @comment.documentation 
+;   (#lua-match? @comment.documentation "^{-"))
```
lucario387, line 46: So is this query. This is also fine

All 
```
((comment) ...
 (comment) @comment.documentation)
```
however, are not, they need anchors and quantifiers

queries/haskell/highlights.scm
lucario387: Ah, so the usage of this is supposed to be with the quantifiers.
Just ignore my comment and the one after

mrcjkb:
> Then you could add yourself as maintainer in `parsers.lua` (maybe separate commit in this PR) :)

Done. I also updated the readme (not sure if I should have?)

PR #5466, opened 2023-10-01
feat(haskell): highlights improvements + tests

mrcjkb:
This adds a test file for Haskell highlights + some improvements to the highlight queries.

### DIFF
--- a/queries/haskell/highlights.scm
+++ b/queries/haskell/highlights.scm
@@ -116,7 +116,8 @@
 
 
 (module) @namespace
-(qualified_module (module) @constructor)
+((qualified_module (module) @constructor)
+ . (module))
 (qualified_type (module) @namespace)
 (qualified_variable (module) @namespace)
 (import (module) @namespace)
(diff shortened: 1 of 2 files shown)


mrcjkb:
Marking this as draft until I'm sure the tests succeed.

lucario387:
>             ...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:96: Error in at tests/query/highlights/haskell/test.hs:157:19: expected "function", captures: { "variable", "function.call" }", highlights: { "variable", "function.call" }

mrcjkb:
> > ```
> >         ...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:96: Error in at tests/query/highlights/haskell/test.hs:157:19: expected "function", captures: { "variable", "function.call" }", highlights: { "variable", "function.call" }
> > ```

 :pray:  Fixed

mrcjkb:
Marking this as draft, because I have some more improvements to add.

@lucario387 if you want to merge this, I can continue in another PR.

lucario387:
I'll go with whatever you think is appropriate @mrcjkb

mrcjkb:
Then let's merge this and I'll continue in another PR :smile:

queries/haskell/highlights.scm
```diff
@@ -116,7 +116,8 @@
 
 
 (module) @namespace
-(qualified_module (module) @constructor)
+((qualified_module (module) @constructor)
+ (module))
```
lucario387, line 120: what is this second `module`

lucario387: Also, as this is some form of `((a) (b))` query, it will lead to unwanted captures, i.e. `a a b b` is 4 matches when I think you probably only want it to be one?

lucario387:
Aside from the above, the queries look fine to me

queries/haskell/highlights.scm
mrcjkb: This is to match
```haskell
import qualified Data.Map as Map
                -- ^ @constructor
                           -- ^ @namespace
```

(inspired by how it's done in Helix).
The second `(module)` should be anchored.

lucario387 (approved):

PR #5470, opened 2023-10-02
fix(haskell): highlight fixes

mrcjkb:
This is a follow-up to #5466.

Things done:

- Move `(variable) @variable` and `(pat_wildcard) @variable` to the top so they don't override `@parameter`.
- Refine infix-based `@function.call` queries.
- Remove queries for higher order function parameters (they don't scale well - this would be better left up to haskell-language-server when/if semantic highlighting is implemented).
- Merge separate queries for `(variable)` and `(qualified_variable (variable))` -> `[(variable) (qualified_variable (variable))]`
- Add more `@parameter` cases
- Add more `@field` queries

I'll leave this as draft while I dogfeed it.

### DIFF
--- a/queries/haskell/highlights.scm
+++ b/queries/haskell/highlights.scm
@@ -1,14 +1,23 @@
 ;; ----------------------------------------------------------------------------
-;; Parameters
+;; Parameters and variables
 
 ;; NOTE: These are at the top, so that they have low priority,
 ;; and don't override destructured parameters 
 
-(function
-  patterns: (patterns (pat_name) @parameter))
+(variable) @variable
+(pat_wildcard) @variable
 
-(exp_lambda
-  (pat_name) @parameter)
+(function
+  patterns: (patterns (_) @parameter))
+
+(exp_lambda (_) @parameter . (_))
+
+(function 
+  infix: (infix
+    lhs: (_) @parameter))
+(function 
+  infix: (infix
+    rhs: (_) @parameter))
 
 ;; ----------------------------------------------------------------------------
 ;; Literals and comments
@@ -121,6 +130,7 @@
 (qualified_type (module) @namespace)
 (qualified_variable (module) @namespace)
 (import (module) @namespace)
+(import (module) @constructor . (module))
 
 [
   (where)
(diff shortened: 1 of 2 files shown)


tests/query/highlights/haskell/test.hs
```diff
@@ -150,12 +152,22 @@ quasiQuotedString = [qq|Some string|]
...
 composition f g = f . g
-         -- ^ @function
-           -- ^ @function
                -- ^ @function
                    -- ^ @function
+qualifiedComposition = SomeModule.f . SomeModule.g
+                        -- ^ @namespace
+                               -- ^ @function
+                                      -- ^ @namespace
+                                              -- ^ @function
```
lucario387, line 162: I don't think it needs to be tested this carefully. The test right above seems to me, to have covered for this already. Don't need to overly test for `namespace` as well

queries/haskell/highlights.scm
```diff
@@ -186,59 +239,67 @@
...
-; qualified function calls with an infix operator
-(exp_infix (exp_name 
-  (qualified_variable (
+(exp_infix (exp_name [
+    ((variable) @function.call)
+    (qualified_variable (
       (module) @namespace
-      (variable) @function.call))) . (operator))
+      (variable) @function.call))
+  ]) . (operator))
```
lucario387, line 263: This may be easier to keep track of which part of the query is for which

```suggestion
(exp_infix
  (exp_name
    [
      ((variable) @function.call))
      (qualified_variable
        (module) @namespace
        (variable) @function.call)
    ])
  . (operator))
```

mrcjkb:
I think this is in a pretty stable state right now.
At this point I'm going down a rabbit hole of finding edge cases and minor improvements, which I can continue with in a follow-up PR.

tests/query/highlights/haskell/test.hs
mrcjkb: Oh, I thought I had replied to this...
The above test doesn't match `(qualified_variable (variable))`.
But the `@namespace` tests are overkill here, I agree. I've removed them.

queries/haskell/highlights.scm
```diff
@@ -1,14 +1,65 @@
...
+(pat_wildcard) @variable
+
 (function
-  patterns: (patterns (pat_name) @parameter))
+  patterns: (patterns [
+    (pat_name (variable) @parameter)
+    (_ (pat_name (variable) @parameter))
+    (_ (pat_apply (pat_name (variable) @parameter)))
+    (_ (pat_typed pattern: (pat_name) @parameter))
+    (_ (pat_typed pattern: (pat_apply (pat_name (variable) @parameter))))
```
lucario387, line 16: this query seems too nested to me. What stops it from being a separate `(pat_name (variable) @parameter)`. All these patterns seem to be relying on that.

@amaanq what's your thoughts about it

mrcjkb: If I just query `(pat_name (variable) @parameter)`, it'll match any pattern matching expressions, for example

```haskell
case mFoo of
  Just bar -> -- <...>
  Nothing -> -- <...>
```
which would require equally nested queries.

mrcjkb: 🤔 come to think of it, maybe it's better not to try to query for destructured patterns, because those can be arbitrarily nested, e.g.

```haskell
func (Left (Just a)) = -- <...>
func (Right (Just a)) = -- <...>
```

mrcjkb: Surprisingly, matching the entire expression appears to work for highlighting.

lucario387: > maybe it's better not to try to query for destructured patterns, because those can be arbitrarily nested, e.g.

I'd say just avoid it entirely. If you nested twice and it works, and then the 3rd nest doesn't work and people will randomly ask why doesn't this work

And then there will be some edge cases like https://github.com/projekt0n/github-nvim-theme/pull/250/files, which I do not want to happen at all

lucario387 (commented):
Looks fine to me

lucario387 (approved):

queries/haskell/highlights.scm
mrcjkb: I've just updated it to match the entire expression, and it works (see the screen shot).
I have my theme configured to italicize `@parameter` matches.

queries/haskell/highlights.scm
```diff
@@ -1,14 +1,23 @@
 ;; ----------------------------------------------------------------------------
-;; Parameters
+;; Parameters and variables
 
 ;; NOTE: These are at the top, so that they have low priority,
 ;; and don't override destructured parameters 
 
+(variable) @variable
+(pat_wildcard) @variable
```
ObserverOfTime, line 8: What is `pat_wildcard`?

queries/haskell/highlights.scm
```diff
@@ -186,59 +249,87 @@
...
+    ])
+    (exp_type_application)
+    (exp_parens)
+    (exp_record)
+  ]
+  . (exp_name [
+    ((variable) @variable)
+    (qualified_variable (variable) @variable)
+  ])
+)
```
ObserverOfTime, line 332:
```suggestion
  ]))
```

queries/haskell/highlights.scm
```diff
@@ -263,6 +354,13 @@
 (quasiquote 
   (quoter) @_name (#eq? @_name "qq")
   (quasiquote_body) @string)
+;; namespaced quasi-quoter
+(quasiquote
+  (_
+    (module) @namespace
+    . (variable) @function.call
+  )
+)
```
ObserverOfTime, line 363:
```suggestion
  ))
```

queries/haskell/highlights.scm
lucario387: Probably `_` in `match x ... _ -> somefunc`

mrcjkb: :point_up:

ObserverOfTime: So `@parameter` maybe?

mrcjkb: It can appear in any pattern matching expression. I wouldn't change it. I just moved these up from lines 154 and 153, so they don't override the other queries.

ObserverOfTime (approved):

PR #5475, opened 2023-10-03
feat: add nasm parser

ObserverOfTime:
I'm not very familiar with (N)ASM so the highlights can probably be improved.

### DIFF
--- a/queries/nasm/highlights.scm
+++ b/queries/nasm/highlights.scm
@@ -1,0 +1,141 @@
+;; adapted from https://github.com/naclsn/tree-sitter-nasm/blob/main/queries/highlights.scm
+
+(word) @variable
+((word) @constant
+  (#lua-match? @constant "^[A-Z_][?A-Z_0-9]+$"))
+((word) @constant.builtin
+  (#lua-match? @constant.builtin "^__%?[A-Z_a-z0-9]+%?__$"))
+
+[
+  (line_here_token)
+  (section_here_token)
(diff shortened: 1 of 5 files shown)


queries/nasm/highlights.scm
```diff
@@ -0,0 +1,141 @@
+;; adapted from https://github.com/naclsn/tree-sitter-nasm/blob/main/queries/highlights.scm
+
+((word) @constant
+  (#match? @constant "^[A-Z_][?A-Z_0-9]+$"))
```
clason, line 4: ouch. is there now way to make these more explicit (if necessary, as multiple patterns)?

ObserverOfTime: More explicit?

clason: Using simple Lua patterns, rather than a big (Vim) regex.

ObserverOfTime: I can change it to `lua-match` (I forgot to do that) but I don't know if I can simplify it.

clason: Ah, ok, that would be a good compromise (I assumed that the `match` was deliberate since it couldn't be done with `lua-match`).

amaanq:
Sorry for not responding sooner 😅 I pushed a few updates to the highlights, feels good to me if that's good w/ you

amaanq (approved):

PR #5476, opened 2023-10-03
fix(lua): add `vim` back to `namespace.builtin`

folke:
add `vim` back to `namespace.builtin`.

It was probably accidentally removed in https://github.com/nvim-treesitter/nvim-treesitter/commit/581fc14cea6e85c710ece75f89f17dffd816fbc4

This is annoying, since highlights change as soon as semantic tokens load

### DIFF
--- a/queries/lua/highlights.scm
+++ b/queries/lua/highlights.scm
@@ -134,7 +134,7 @@
   (#eq? @variable.builtin "self"))
 
 ((identifier) @namespace.builtin
-  (#any-of? @namespace.builtin "_G" "debug" "io" "jit" "math" "os" "package" "string" "table" "utf8"))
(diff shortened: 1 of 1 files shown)


clason:
No, this was intentional, since the parser and queries here are meant to be generic, not specific to Neovim development. 

Such local additions should go into local config.

folke:
Then you should also remove `jit`.

Be aware that everyoen that uses Neovim to edit their lua config:

- will see the color change every time semantic tokens load
- the new `foldtext` will have the wrong color as well

clason:
Semantic tokens will blinkenlight in any case; if it's not this, it's something else. That is not a compelling argument. I'll consider removing jit, though.

PR #5483, opened 2023-10-06
java(highlights): add missing import highlights

lucario387:
The pattern before does not occur if it's in import, leading to incorrect coloring

| Before | After |
| - | - |
| image| image|

### DIFF
--- a/queries/java/highlights.scm
+++ b/queries/java/highlights.scm
@@ -105,8 +105,8 @@
 ((field_access
   object: (identifier) @type)
   (#lua-match? @type "^[A-Z]"))
-((scoped_identifier
-  scope: (identifier) @type)
+(scoped_identifier
+  (identifier) @type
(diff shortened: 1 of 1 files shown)


queries/java/highlights.scm
```diff
@@ -105,8 +105,11 @@
 ((field_access
   object: (identifier) @type)
   (#lua-match? @type "^[A-Z]"))
-((scoped_identifier
-  scope: (identifier) @type)
+(scoped_identifier
+  scope: (identifier) @type
+  (#lua-match? @type "^[A-Z]"))
+(scoped_identifier
+  name: (identifier) @type
```
amaanq, line 112: why not just remove the field and capture (identifier) once

lucario387: agreed.

amaanq (approved):

PR #5487, opened 2023-10-06
refactor: remove support for 0.8 for indents

lucario387:
The minimum supported version is already 0.9, no need to keep the compatibility functions anymore


clason:
I have actually not planned to make any changes in this regard, since it will make rebasing `main` (the _real_ "drop obsolete code" refactor) quite a bit harder...

lucario387:
I will close this then.

clason:
But if this code is actually harmful (and not just esthetically displeasing), that would be a different issue!

lucario387:
I don't see how it's harmful (yet).

PR #5489, opened 2023-10-06
fix(php): fix indentation for in between braces

afonsocarlos:
This is an improvement fix on top of the #3591 so indentation also works properly in cases like:

```php
<?php

function test() {
    $array = [|] // insert new line in between braces
}
```

### DIFF
--- a/queries/php/indents.scm
+++ b/queries/php/indents.scm
@@ -10,7 +10,6 @@
   (switch_block)
   (match_block)
   (case_statement)
-  "["
 ] @indent.begin
 
 [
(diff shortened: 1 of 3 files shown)


afonsocarlos:
Additionally, the test for `unfinished-call.php` was incorrectly updated some commits ago. The expected behavior must be to keep the indentation level until closing "function call", not resetting the cursor position to 0. This PR also fixes that as a side effect.

queries/php/indents.scm
```diff
@@ -25,4 +24,5 @@
 
 (compound_statement "}" @indent.end)
 
-(ERROR) @indent.auto
+(ERROR "(" @indent.align (#set! indent.open_delimiter "(") (#set! indent.close_delimiter ")") . (_))
```
lucario387, line 27: Reading this is very hard to understand

```suggestion
(ERROR
  "(" @indent.align
  . (_)
  (#set! indent.open_delimiter "(")
  (#set! indent.close_delimiter ")"))
```

lucario387 (changes requested):

lucario387 (approved):

PR #5502, opened 2023-10-09
fix(haskell): prevent segfault due to anchorless queries

### LINKED ISSUE #5500: Haskell: Neovim segfaults after commit 6276cd9
### Describe the bug

Neovim crashes after jumping with `` `. `` to the latest change in one of my Haskell file. I bisected commits and found that the culprit is 6276cd9.

Here is the backtrace:

```GDB
(gdb) bt
#0  0x00007fa16ca70d25 in ts_query_cursor_next_capture (self=self@entry=0x565330b48e00, match=match@entry=0x7ffd3f42abb0, capture_index=capture_index@entry=0x7ffd3f42aba8)
    at lib/src/query.c:4061
#1  0x000056532fb129b7 in query_next_capture (L=0x7fa16c66e380) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/lua/treesitter.c:1308
#2  0x00007fa16c9b5a7b in lj_BC_FUNCC () at buildvm_x86.dasc:859
#3  0x00007fa16c9c8be0 in lua_pcall (L=L@entry=0x7fa16c66e380, nargs=nargs@entry=0, nresults=nresults@entry=0, errfunc=errfunc@entry=-2)
    at /usr/src/debug/luajit-2.1.0-0.27beta3.fc38.x86_64/src/lj_api.c:1116
#4  0x000056532fb070ec in nlua_pcall (lstate=lstate@entry=0x7fa16c66e380, nargs=nargs@entry=0, nresults=nresults@entry=0)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/lua/executor.c:159
#5  0x000056532fb0fb31 in nlua_typval_exec (lcmd=lcmd@entry=0x565331b98920 "require(\"treesitter-context\").update_context()", lcmd_len=46, name=name@entry=0x56532fc7d8f0 ":lua", 
    args=args@entry=0x0, argcount=argcount@entry=0, special=special@entry=false, ret_tv=0x0) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/lua/executor.c:1443
#6  0x000056532fb0ff23 in ex_lua (eap=<optimized out>) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/lua/executor.c:1639
#7  0x000056532fab6dd0 in execute_cmd0 (retv=retv@entry=0x7ffd3f42ae24, eap=eap@entry=0x7ffd3f42ae30, errormsg=errormsg@entry=0x7ffd3f42ae28, preview=preview@entry=false)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/ex_docmd.c:1620
#8  0x000056532faba4d1 in do_one_cmd (cmdlinep=cmdlinep@entry=0x7ffd3f42b088, flags=flags@entry=7, cstack=cstack@entry=0x7ffd3f42b110, fgetline=fgetline@entry=0x56532fa2b1af <getnextac>, 
    cookie=cookie@entry=0x7ffd3f42b6e0) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/ex_docmd.c:2279
#9  0x000056532fabb07a in do_cmdline (cmdline=<optimized out>, fgetline=0x56532fa2b1af <getnextac>, cookie=0x7ffd3f42b6e0, flags=7)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/ex_docmd.c:578
#10 0x000056532fa2bf54 in apply_autocmds_group (event=EVENT_USER, fname=<optimized out>, fname@entry=0x565330f6835c "CursorMovedVertical", fname_io=fname_io@entry=0x0, 
    force=force@entry=true, group=group@entry=-3, buf=buf@entry=0x565330b910d0, eap=0x0, data=0x0) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/autocmd.c:1940
#11 0x000056532fa2c312 in do_doautocmd (arg_start=0x565330f68357 "User CursorMovedVertical", do_msg=false, did_something=0x7ffd3f42b86f)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/autocmd.c:1287
#12 0x000056532fabdfa7 in ex_doautocmd (eap=<optimized out>) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/ex_docmd.c:4261
#13 0x000056532fab6dd0 in execute_cmd0 (retv=retv@entry=0x7ffd3f42b924, eap=eap@entry=0x7ffd3f42b930, errormsg=errormsg@entry=0x7ffd3f42b928, preview=preview@entry=false)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/ex_docmd.c:1620
#14 0x000056532faba4d1 in do_one_cmd (cmdlinep=cmdlinep@entry=0x7ffd3f42bb88, flags=flags@entry=7, cstack=cstack@entry=0x7ffd3f42bc10, fgetline=fgetline@entry=0x56532fbc44b2 <get_str_line>, 
    cookie=cookie@entry=0x7ffd3f42c2c0) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/ex_docmd.c:2279
#15 0x000056532fabb07a in do_cmdline (cmdline=<optimized out>, fgetline=0x56532fbc44b2 <get_str_line>, cookie=0x7ffd3f42c2c0, flags=7)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/ex_docmd.c:578
#16 0x000056532fbc4778 in source_using_linegetter (cookie=cookie@entry=0x7ffd3f42c2c0, fgetline=fgetline@entry=0x56532fbc44b2 <get_str_line>, traceback_name=<optimized out>)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/runtime.c:1906
#17 0x000056532fbc47fb in do_source_str (cmd=<optimized out>, traceback_name=<optimized out>) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/runtime.c:1955
#18 0x000056532fa21f26 in exec_impl (channel_id=9223372036854775809, src=..., opts=opts@entry=0x7ffd3f42c410, err=err@entry=0x7ffd3f42c3e0)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/api/vimscript.c:92
#19 0x000056532fa22072 in nvim_exec2 (channel_id=<optimized out>, src=..., opts=0x7ffd3f42c410, err=0x7ffd3f42c3e0) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/api/vimscript.c:59
#20 0x000056532f9f2e56 in nlua_api_nvim_exec2 (lstate=0x7fa16c66e380) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/redhat-linux-build/src/nvim/auto/lua_api_c_bindings.generated.c:5574
#21 0x00007fa16c9b5a7b in lj_BC_FUNCC () at buildvm_x86.dasc:859
#22 0x00007fa16c9c8be0 in lua_pcall (L=L@entry=0x7fa16c66e380, nargs=nargs@entry=0, nresults=nresults@entry=0, errfunc=errfunc@entry=-2)
    at /usr/src/debug/luajit-2.1.0-0.27beta3.fc38.x86_64/src/lj_api.c:1116
#23 0x000056532fb070ec in nlua_pcall (lstate=lstate@entry=0x7fa16c66e380, nargs=nargs@entry=0, nresults=nresults@entry=0)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/lua/executor.c:159
#24 0x000056532fb0fb31 in nlua_typval_exec (lcmd=lcmd@entry=0x5653310e75b0 "require(\"treesitter-context\").do_au_cursor_moved_vertical()", lcmd_len=59, 
    name=name@entry=0x56532fc7d8f0 ":lua", args=args@entry=0x0, argcount=argcount@entry=0, special=special@entry=false, ret_tv=0x0)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/lua/executor.c:1443
#25 0x000056532fb0ff23 in ex_lua (eap=<optimized out>) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/lua/executor.c:1639
--Type <RET> for more, q to quit, c to continue without paging--  
#26 0x000056532fab6dd0 in execute_cmd0 (retv=retv@entry=0x7ffd3f42c664, eap=eap@entry=0x7ffd3f42c670, errormsg=errormsg@entry=0x7ffd3f42c668, preview=preview@entry=false)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/ex_docmd.c:1620
#27 0x000056532faba4d1 in do_one_cmd (cmdlinep=cmdlinep@entry=0x7ffd3f42c8c8, flags=flags@entry=7, cstack=cstack@entry=0x7ffd3f42c950, fgetline=fgetline@entry=0x56532fa2b1af <getnextac>, 
    cookie=cookie@entry=0x7ffd3f42cf20) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/ex_docmd.c:2279
#28 0x000056532fabb07a in do_cmdline (cmdline=<optimized out>, fgetline=0x56532fa2b1af <getnextac>, cookie=0x7ffd3f42cf20, flags=7)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/ex_docmd.c:578
#29 0x000056532fa2bf54 in apply_autocmds_group (event=EVENT_CURSORMOVED, fname=<optimized out>, fname_io=<optimized out>, force=<optimized out>, group=group@entry=-3, buf=0x565330b910d0, 
    eap=0x0, data=0x0) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/autocmd.c:1940
#30 0x000056532fa2c3e7 in apply_autocmds (event=<optimized out>, fname=<optimized out>, fname_io=<optimized out>, force=<optimized out>, buf=<optimized out>)
    at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/autocmd.c:1603
#31 0x000056532fb54519 in normal_check_cursor_moved (s=s@entry=0x7ffd3f42d0b0) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/normal.c:1269
#32 0x000056532fb561b0 in normal_check (state=0x7ffd3f42d0b0) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/normal.c:1398
#33 0x000056532fbf6065 in state_enter (s=0x7ffd3f42d0b0) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/state.c:40
#34 0x000056532fb53ff1 in normal_enter (cmdwin=false, noexmode=false) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/normal.c:497
#35 0x000056532f9e4695 in main (argc=3, argv=<optimized out>) at /usr/src/debug/neovim-0.9.2-1.fc38.x86_64/src/nvim/main.c:642
```

```GDB
(gdb) f 0
#0  0x00007fa16ca70d25 in ts_query_cursor_next_capture (self=self@entry=0x565330b48e00, match=match@entry=0x7ffd3f42abb0, capture_index=capture_index@entry=0x7ffd3f42aba8)
    at lib/src/query.c:4061
4061	       self->states.contents[first_unfinished_state_index].capture_list_id
(gdb) l
4056	       first_unfinished_pattern_index,
4057	       first_unfinished_capture_byte
4058	     );
4059	     capture_list_pool_release(
4060	       &self->capture_list_pool,
4061	       self->states.contents[first_unfinished_state_index].capture_list_id
4062	     );
4063	     array_erase(&self->states, first_unfinished_state_index);
4064	   }
4065	
(gdb) p first_unfinished_state_index
$1 = 4294967295
```
The `first_unfinished_state_index` is apparently garbage.

Then I tried to spot the region in the source file after entering which treesitter crashes. Not sure if it can help, but here it is:

```Haskell
                touchPeers name now
            when (walkStart > 0 && walkStatus == NotAccessible &&
                     (len > 1 || length names > 1)) $
                         blacklistPeer name now
            getNext walk p now peersBlacklistInterval
        ) $ foundPeers' >>= findPeers walkSeqn
    where createNewPeers name len p t wt = do
              rr <- createRoundRobin len
              logG LogInfo $ C8.pack $
                  logCaption "getPeer" (Just walkConn) ++
                  "Create peers for service " ++ T.unpack name ++
                  ": " ++ show (map snd p)
              atomicModifyIORef' collectedPeers $ \curPeers@(_, ps) ->
                  let upd = isNothing $ M.lookup name ps
                  in (if upd
                          then let ps' = (t, ((0, (Nothing, (0, [])))
                                             ,(1, (rr, (len, p)))
                                             )
                                         )
                               in (wt, M.insert name ps' ps)
                          else curPeers
                     ,upd
                     )
          touchPeers name now =
              atomicModifyIORef' collectedPeers $
                  (, ()) . second (M.adjust (first $ second $ const now) name)
          blacklistPeer name now =
              atomicModifyIORef' collectedPeers $ \curPeers@(t, ps) ->
                  let ps' = M.lookup name ps
                  in (maybe curPeers
                         (blacklistPeer' name now ps curPeers t) ps'
                     ,()
                     )
```

### To Reproduce

Not sure how, this depends on the source code.

### Expected behavior

No segfault.

```markdown

Installation ~
- OK `tree-sitter` found 0.20.7 (parser generator, only needed for :TSInstallFromGrammar)
- OK `node` found v18.18.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
  Version: cc (GCC) 13.2.1 20230728 (Red Hat 13.2.1-1)

OS Info:
{
  machine = "x86_64",
  release = "6.5.5-200.fc38.x86_64",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Sun Sep 24 15:52:44 UTC 2023"
} ~

  - cmake               ✓ . ✓ ✓ .
  - haskell             ✓ . ✓ . ✓
  - latex               ✓ . ✓ . ✓
  - perl                ✓ . ✓ . ✓
  - r                   ✓ ✓ . ✓ ✓

```

```text
NVIM v0.9.2
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wvla -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -fno-common -Wno-unused-result -Wimplicit-fallthrough -fdiagnostics-color=auto -fstack-protector-strong -DUNIT_TESTING -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -I/usr/include/luajit-2.1 -I/usr/include -I/usr/include/luv -I/builddir/build/BUILD/neovim-0.9.2/redhat-linux-build/src/nvim/auto -I/builddir/build/BUILD/neovim-0.9.2/redhat-linux-build/include -I/builddir/build/BUILD/neovim-0.9.2/redhat-linux-build/cmake.config -I/builddir/build/BUILD/neovim-0.9.2/src -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include

```

### Additional context

_No response_

mrcjkb:
Fixed #5500.

### DIFF
--- a/queries/haskell/highlights.scm
+++ b/queries/haskell/highlights.scm
@@ -162,26 +162,11 @@
 ;; Functions and variables
 
 (signature name: (variable) @function)
-((signature name: (variable) @variable)
- (function 
-  name: (variable) @_name
-  rhs: [
-   (exp_literal)
-   (exp_apply 
-     (exp_name 
-       [(constructor)
-        (variable)
-       ]))
-   (quasiquote)
-   ((exp_name) . (operator))
-  ])
- (#eq? @variable @_name))
 
 ((signature name: (variable) @variable)
- (function 
-  name: (variable) @_name
-  rhs: (exp_infix 
-    [
+ . (function 
+    name: (variable) @_name
+    rhs: [
      (exp_literal)
      (exp_apply 
        (exp_name 
@@ -190,26 +175,36 @@
          ]))
      (quasiquote)
      ((exp_name) . (operator))
-    ]))
+    ])
+ (#eq? @variable @_name))
+
+((signature name: (variable) @variable)
+ . (function 
+    name: (variable) @_name
+    rhs: (exp_infix 
+      [
+       (exp_literal)
+       (exp_apply 
+         (exp_name 
+           [(constructor)
+            (variable)
+           ]))
+       (quasiquote)
+       ((exp_name) . (operator))
+      ]))
  (#eq? @variable @_name))
 
 ((signature name: (variable) @function)
- (function 
-  name: (variable) @_name
-  patterns: (patterns))
+ . (function 
+    name: (variable) @_name
+    patterns: (patterns))
  (#eq? @function @_name))
 
-(signature 
-  name: (variable) @function
-  type: (fun))
-((signature 
-   name: (variable) @_name
-   type: (fun))
- (function
-   name: (variable) @function)
- (#eq? @function @_name))
+;; For some reason, `(signature name: (variable) type: (fun))` doesn't work here
+(signature (variable) @function . (fun))
 
 (function name: (variable) @function)
+
 (function 
   name: (variable) @variable
   rhs: [
@@ -222,6 +217,7 @@
    (quasiquote)
    ((exp_name) . (operator))
   ])
+
 (function 
   name: (variable) @variable
   rhs: (exp_infix [
--- a/tests/query/highlights/haskell/test.hs
+++ b/tests/query/highlights/haskell/test.hs
@@ -111,16 +111,13 @@ someInfix x = fromIntegral x `myAdd` floatVal
              -- ^ @variable
                  -- ^ @variable
     floatVal :: Double
-    -- ^ @variable
     floatVal = 5.5
     -- ^ @variable
             -- ^ @float
     intVal :: Int
-    -- ^ @variable
     intVal = getInt 5
     -- ^ @variable
     boolVal :: Bool
-    -- ^ @variable
     boolVal = bool False True $ 1 + 2 == 3
     -- ^ @variable
     isInt :: Either Double Int -> Bool
@@ -171,12 +168,14 @@ someIOaction = do
         -- ^ @function.call
 
 intVal :: Int
--- ^ @variable
 intVal = 5
 -- ^ @variable
 
+intFun :: Int -> Int
+intFun = 5
+-- ^ @function
+
 mbInt :: Maybe Int
--- ^ @variable
 mbInt = Just 5
 -- ^ @variable
 


mrcjkb:
Marking as draft - It looks like this still needs some tweaking.

mrcjkb:
I think it's good now.
For some reason, the query

```scheme
(signature 
  name: (variable) 
  type: (fun))
```

stopped working :thinking:

queries/haskell/highlights.scm
```diff
@@ -190,26 +176,44 @@
...
+      ]))+
  (#eq? @variable @_name))
 
 ((signature name: (variable) @function)
- (function 
-  name: (variable) @_name
-  patterns: (patterns))
+ . (comment)*
+ . (function 
+    name: (variable) @_name
```
lucario387, line 202: This query wont be working very well right now for neovim as it's captures inside quantifiers. What's the result looking like while using helix (preferably helix nightly) with this query?

queries/haskell/highlights.scm
```diff
@@ -190,26 +176,44 @@
...
-   type: (fun))
- (function
-   name: (variable) @function)
+;; For some reason, `(signature name: (variable) type: (fun))` doesn't work here
+(signature (variable) @function . (fun))
+
+((signature (variable) @_name . (fun))
+ . ((comment)*
+ . (function (variable)))*
+ . (function (variable) @function)
```
lucario387, line 212: This last one doesn't seem necessary since there's already a `(function name: (variable) @function)` right below for all patterns of that

queries/haskell/highlights.scm
```diff
@@ -190,26 +176,44 @@
...
+      [
+       (exp_literal)
+       (exp_apply 
+         (exp_name 
+           [(constructor)
+            (variable)
+           ]))
+       (quasiquote)
+       ((exp_name) . (operator))
+      ]))+
```
lucario387, line 196: This section looks to me like a nearly exhaustive list of available function rhs.  I think we can compromise and just check for 
```
(function
  name: (variable)
  rhs: (exp_infix))
```
for this section of the query, instead of going deep down.

queries/haskell/highlights.scm
mrcjkb: This query isn't in Helix. I don't think I've actually encountered a case where this is necessary (yet). I just added it for better reliability.
The reason is as follows:

- In Haskell, everything is a function. So if you have `x = 5`, `x` is a function that evaluates to 5.
- The treesitter parser will parse `x` as a `variable` in  `foo = x`, but `foo` a function, and there are highlights that try to treat `foo` as a variable.
- There are some ways to detect this (e.g. if the lhs has parameters), and the tree-sitter queries are quite good at this.

A sure way to detect this is if there's an accompanying signature of `type: fun`.
For example:

```haskell
foo :: Int -> Int
foo = x
```

The problem is, signatures and functions are at the same level in the tree.
This makes sense, because in Haskell, signatures can be defined anywhere in the same module.
So something like

```haskell
foo :: Int -> Int

bar = 5

foo = x
```

is perfectly valid (although an unlikely use case).
That's why I had all those unanchored queries.

The query you commented on tries to match things like

```haskell
foo :: Int -> Int
-- some comment
foo 5 = 0
-- another comment
foo x = x + 1
```

But now that I'm writing this, I think there's no need for this.
The only case we need this for is 

```haskell
foo :: Int -> Int
-- some comment
-- maybe another comment
foo = x
```

because 

```haskell
foo :: Int -> Int
foo = 5
foo = x
```
is invalid haskell

and

```haskell
foo x = --<...>
```

is easy to capture without having to know about signatures.

mrcjkb: I just realised there were a bunch of queries that tested for this. They were just missing `. (comment)* .` in between the `(signature) . (function)` anchors (see my second commit).

queries/haskell/highlights.scm
mrcjkb: Yup. They were just missing ` . (comment)* `.

queries/haskell/highlights.scm
lucario387: I don't think having `(comment)*` here will be nice. Treesitter should not be responsible for this level of granular control. It should be left to the LSP.

Just `(signature)` and `(function)` that sits alongside one another will be better imo

queries/haskell/highlights.scm
mrcjkb: I've tried that. IIRC, it matched too much.
I'd have to play around with that again after work.

queries/haskell/highlights.scm
mrcjkb: It's quite common to have comments between the signature and the function (this is the case in the test file, too).
The language server doesn't support highlighting hints yet.

lucario387: I'm somewhat afraid for matching this kind of pattern, mostly because of the double quantifiers

something like 
`(signature) . (some 10 or more comment lines) . (10 consecutive functions)` can be quite costly

lucario387: Also, I'm pretty sure the original intention is not to match that far away. I don't think you need to capture all functions going right after a signature. I believe matching a function right after the signature is fine. Trying too much to make the highlights work well won't be worth it

queries/haskell/highlights.scm
mrcjkb: Shall we create a separate issue from this thread?
I think preventing the segfault is quite high priority.

lucario387: Yeah preventing the segfault is higher priority for sure.

queries/haskell/highlights.scm
```diff
@@ -190,26 +176,38 @@
          ]))
      (quasiquote)
      ((exp_name) . (operator))
-    ]))
+    ])+
```
lucario387, line 179: Do we still need to quantify `(function)` here ? I don't think it's that helpful anymore considering we're limiting it to the function right after its signature

queries/haskell/highlights.scm
mrcjkb: @lucario387 do you happen to know why the following signature queries (~line 241) are anchored to functions?

```scheme
((signature (variable) @function (fun)) . (function (variable)))
((signature (variable) @function (context (fun))) . (function (variable)))
((signature (variable) @function (forall (context (fun)))) . (function (variable)))
```

I believe this schould suffice:

```scheme
((signature (variable) @function (fun)))
((signature (variable) @function (context (fun))))
((signature (variable) @function (forall (context (fun)))))
```

I suppose it was done because they are each accompanied by something like this

```scheme
((signature (variable) @_type (fun)) . (function (variable) @function) (#eq? @function @_type))
```

Personally, I think a signature that has a function type, such as

```haskell
someFunc :: Int -> ReturnType
```

should always be highlighted as a function, even if it's accompanied by a function that for some reason doesn't get highlighted correctly.

The reason I'm asking is because I think I can get rid of the signature queries with long lists of rhs (see [your other comment](https://github.com/nvim-treesitter/nvim-treesitter/pull/5502#discussion_r1351612064)) by adding something like this:

```scheme
;; Consider signatures with only one value on the rhs as variables
(signature . (variable) @variable . (_) . )
;; but consider 'IO a' a function
(signature name: (variable) @function 
  . (type_apply (type_name) @_type)
    (#eq? @_type "IO"))
```

Which would result in

```haskell
intVal :: Int
-- ^ @variable

mbInt :: Maybe Int
-- ^ @variable

intFunc :: Int -> Int
-- ^ @function

getArgs :: IO [String]
-- ^ @function

main :: IO ()
-- ^ @function
```

... but this only works if I remove the anchors in the other queries.

The other function queries seem pretty reliable by now (apart from very few edge cases, so I don't think we really need to check for accompanying signatures).

I'll be dogfeeding the idea this week, but I wanted to get your input on it :smile:

lucario387: You are the maintainer for haskell queries now. If anything, you have the call to change some of the queries 😄 

For the anchored signature, I honestly do not know. If you believe the query can be improved, go for it 😄 . I'm all rooting for you

PR #5511, opened 2023-10-12
fix(elixir): highlight functions without parenthesis

xanderio:
Elixir does not require parenthesis after the function name for a valid function definition.

### DIFF
--- a/queries/elixir/highlights.scm
+++ b/queries/elixir/highlights.scm
@@ -112,6 +112,7 @@
   ))
   (arguments [
     (call (identifier) @function)
+    (identifier) @function
     (binary_operator left: (call target: (identifier) @function) operator: "when")])?)
 
 ; Kernel Keywords & Special Forms


xanderio:
As far as I can tell is just that the `()` are optional for `def*` declarations if the function (or macro, etc.) takes zero arguments.

For example these are the four ways to write this `foo` function that returns `:ok` but takes no arguments. 

```elixir
def foo() do
  :ok
end

def foo do 
  :ok
end

def foo(), do: :ok

def foo, do: :ok
```

I found the upstream elixir-tree-sitter project contain queries to match this case and as far as I can tell some other case currently not handled by our queries. https://github.com/elixir-lang/tree-sitter-elixir/blob/a2861e88a730287a60c11ea9299c033c7d076e30/queries/highlights.scm#L122-L142

amaanq (approved):

Issue #5512, opened 2023-10-12
'@' and '&&' shouldn't both be classified as operator in java files


Delice0:
### Describe the highlighting problem

In intellij, you won't see an operator like '&&' and annotation declarer '@' with the same color. Not in with the built-in themes as far as I know, therefor it shouldn't be the same case using treesitter but as-is you can't change the color of one without changing the other becase both are classified as 'operator.java'.

### Example snippet that causes the problem

1. Add a condition somewhere like: `if (true && true)`
2. Add an annotation somewhere like: `@Getter` (lombok stuff)

### Tree-sitter parsing result

operator: && 

```
        if_statement [188, 4] - [192, 41]
          "if" [188, 4] - [188, 6]
          condition: condition [188, 7] - [188, 21]
            "(" [188, 7] - [188, 8]
            binary_expression [188, 8] - [188, 20]
              left: true [188, 8] - [188, 12]
              operator: "&&" [188, 13] - [188, 15]
```

annotation declarer: @

```
    method_declaration [184, 2] - [417, 3]
      modifiers [184, 2] - [185, 8]
        marker_annotation [184, 2] - [184, 7]
          "@" [184, 2] - [184, 3]
```

### Example screenshot

![Screenshot 2023-10-12 at 16 12 08](https://github.com/nvim-treesitter/nvim-treesitter/assets/59118341/a3547e8b-dfcb-4f94-9b55-6ad61f515309)
![Screenshot 2023-10-12 at 16 12 17](https://github.com/nvim-treesitter/nvim-treesitter/assets/59118341/fdea81af-c5aa-4b27-8994-52410fb3f47e)

### Expected behavior

I expect treesitter differs between an actual operator like '&&' and an annotation declarer '@'. 

One example could be that the '@' is classified as 'annotation.java' 

### Output of `:checkhealth nvim-treesitter`

```markdown
irrelevant
```

### Output of `nvim --version`

```text
NVIM v0.9.2
Build type: Release
LuaJIT 2.1.0-beta3

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.9.2/share/nvim"
```

lucario387:
Not against such improvements. PR welcome.

ObserverOfTime:
`&&` is definitely an operator but `@` could be highlighted the same as the attribute instead.

Delice0:
[...]

PR #5514, opened 2023-10-12
Fix: (java) remove @ from operators highlight

Delice0:
Fixes: https://github.com/nvim-treesitter/nvim-treesitter/issues/5512


amaanq:
As discussed there, it would be better fit as `@annotation` then, can you do so?

queries/java/highlights.scm
```diff
@@ -35,14 +35,15 @@
 ; Annotations
 
 (annotation
-  name: (identifier) @attribute)
+  "@" @annotation
+  name: (identifier) @annotation.name)
 (marker_annotation
-  name: (identifier) @attribute)
+  "@" @annotation
+  name: (identifier) @annotation.name)
```
lucario387, line 42: All of these should be `@attribute`, the CI doesn't accept annotation as a valid name

Delice0: I don't understand. @amaanq asked it to be @annotation or am I missing something here?

amaanq: whoops, mixed up annotation and attribute, but yeah `@attribute` is fine

Delice0: I changed it. Please review again

amaanq (approved):

PR #5515, opened 2023-10-12
feat: (java) Add "var" as part of keyword

Delice0:
Nothing much to say about this rather than 'var' ~~is a keyword~~ is highlighted as keyword in Intellij and should be treated as such for highlighting.

### DIFF
--- a/queries/java/highlights.scm
+++ b/queries/java/highlights.scm
@@ -95,6 +95,8 @@
 (constructor_declaration
   name: (identifier) @type)
 (type_identifier) @type
+((type_identifier) @type.builtin
+  (#eq? @type.builtin "var"))
 ((method_invocation
   object: (identifier) @type)
  (#lua-match? @type "^[A-Z]"))


amaanq:
not how it works, "var" is not an anonymous node in java, look at the parse tree output to see how it's actually captured.

Delice0:
> not how it works, "var" is not an anonymous node in java, look at the parse tree output to see how it's actually captured.

Output:

```
        local_variable_declaration [30, 8] - [40, 10]
          type: type_identifier [30, 8] - [30, 11]
```

I tried adding a condition with the latest commit to make sure var is not part of that type_identifier but not sure if that helps

amaanq:
yeah, there's no "var" node, you need to *match* for var, not directly use it as a string, and iirc that's similar to c++'s 'auto' right?

lucario387:
It's the same as C++ 'auto', yes

Delice0:
> It's the same as C++ 'auto', yes
> 
> I don't see the point to highlight it as keyword tho
> 
> [Oracle docs](https://docs.oracle.com/en/java/javase/20/language/local-variable-type-inference.html):
> 
> > `var` is a reserved type name, not a keyword, which means that existing code that uses `var` as a variable, method, or package name is not affected. However, code that uses `var` as a class or interface name is affected and the class or interface needs to be renamed.

Using intellij as the best example; you can never see 'var' being a different color than the default keyword color and that's why I also want the option to define it as a keyword (or something else in that matter) that makes it easier for me to color it accordingly.

> yeah, there's no "var" node, you need to _match_ for var, not directly use it as a string, and iirc that's similar to c++'s 'auto' right?

I'm unsure about how to approach with this. I'm new to this scm syntax and lua in general so I'm trying my best. Any hints you can give that makes it easier to figure it out?

lucario387:
> Any hints you can give that makes it easier to figure it out?

In this PR, you modified a query to exclude only `var` from being captured as `@type`. So you can always do the reverse and add a pattern to capture only `var`

Delice0:
> > Any hints you can give that makes it easier to figure it out?
> 
> In this PR, you modified a query to exclude only `var` from being captured as `@type`. So you can always do the reverse and add a pattern to capture only `var`

Please check latest commit

queries/java/highlights.scm
```diff
@@ -94,8 +94,8 @@
   name: (identifier) @type)
 (constructor_declaration
   name: (identifier) @type)
-((type_identifier) @type
-  (#not-eq? @type "var"))
+(type_identifier) @type
+  (#eq? @type "var" (#set! @type @type.reserved))
```
clason, line 98: That is not how `#set!` should be used.

Delice0: Updated, please check latest commit

Delice0:
@clason Ready to be reviewed again.

Delice0:
@amaanq  Ready for review again

queries/java/highlights.scm
```diff
@@ -94,7 +94,10 @@
   name: (identifier) @type)
 (constructor_declaration
   name: (identifier) @type)
-(type_identifier) @type
+((type_identifier) @type
+  (#not-eq? @type "var"))
```
amaanq, line 98: unnecessary, later queries will just take precedence

queries/java/highlights.scm
```diff
@@ -94,7 +94,10 @@
...
+((type_identifier) @type.reserved
+  (#eq? @type.reserved "var"))
```
amaanq, line 100:
```suggestion
((type_identifier) @type.builtin
  (#eq? @type.builtin "var"))
```

amaanq (changes requested):

Delice0:
@amaanq Ready again

amaanq (approved):

queries/java/highlights.scm
```diff
@@ -94,7 +94,8 @@
   name: (identifier) @type)
 (constructor_declaration
   name: (identifier) @type)
-(type_identifier) @type
```
amaanq, line 97: huh, wait you'd want this by default for all type_identifiers, then override it if it's var

Delice0: Hmm yeah I thought so, therefor I pushed a new commit. Please check if it's correct

amaanq: No, I'll just fix it..

theHamsta (approved):

Delice0:
@amaanq  just a quick question. If any semantic tokens overrides treesitter's, how do I change the priority so that @type.builtin.java overrides lsp' semantic tokens?

amaanq:
:h lsp-semantic_tokens

PR #5520, opened 2023-10-13
fix(injections): fix injection for hurl for new injection syntax

kevindurb:
hurl injection currently throws an error of `No handler for inject-language` this fixes the injection to match new injection syntax. I pulled from the markdown injection as my example


kevindurb:
Fixes https://github.com/pfeiferj/tree-sitter-hurl/issues/6

kevindurb:
Here is a screenshot of the error 
image

clason (approved):

PR #5525, opened 2023-10-15
fix(typescript): do not highlight undefined as variable

MunifTanjim:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined

### DIFF
--- a/queries/typescript/highlights.scm
+++ b/queries/typescript/highlights.scm
@@ -98,10 +98,6 @@
 
 (conditional_type ["?" ":"] @conditional.ternary)
 
-; Variables
-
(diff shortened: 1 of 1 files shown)


MunifTanjim:
> I think having it highlighted as something else might be nice? 🤔

It's already highlighted as `@constant.builtin`

https://github.com/nvim-treesitter/nvim-treesitter/blob/f213fae3ddbced99a15e85ae387f2da897c4d1d7/queries/ecma/highlights.scm#L182-L185

lucario387 (approved):

PR #5527, opened 2023-10-15
fix(kotlin): avoid capturing jump_expression as @keyword.return

anthony-S93:
Without this PR, everything that comes after the `return` keyword will also be treated as a keyword, which is incorrect. As in: 

unfixed

This is due to the fact that the entire `jump_expression` node is captured as the `keyword.return`. This PR eliminates that and captures only the `return` keyword.

### DIFF
--- a/queries/kotlin/highlights.scm
+++ b/queries/kotlin/highlights.scm
@@ -309,14 +309,16 @@
 ] @keyword
 
 [
+  "return"
+] @keyword.return
+
+[
   "suspend"
 ] @keyword.coroutine
 
 [
   "fun"
 ] @keyword.function
-
-(jump_expression) @keyword.return
 
 [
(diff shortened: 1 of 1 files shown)


lucario387:
@SalBakraa

queries/kotlin/highlights.scm
```diff
@@ -308,6 +308,10 @@
 ;	"typeof" ; NOTE: It is reserved for future use
 ] @keyword
 
+[
+  "return"
+] @keyword.return
```
ObserverOfTime, line 313:
```suggestion
"return" @keyword.return
```

ObserverOfTime (changes requested):
"continue" and "break" should be added to `@repeat`.

Issue #5544, opened 2023-10-19
Misleading contribution docs for @type.definition


aMOPel:
### Describe the bug

[contribution docs](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#types) state

> @type.definition ; type definitions (e.g. `typedef` in C)

Which implies to capture `typedef` as `@type.definition`.
But highlight queries generally use `@type.definition` for the identifier of the defined type instead of the keyword.

### To Reproduce

[C](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/c/highlights.scm#L155)
[CPP](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/cpp/highlights.scm#L36)
[GO](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/go/highlights.scm#L8)

### Expected behavior

> @type.definition ; identifiers in type definitions (e.g. in `typedef <type> <identifier>` in C)

### Output of `:checkhealth nvim-treesitter`

```markdown
unrelated
```

### Output of `nvim --version`

```text
unrelated
```

lucario387:
PR welcome

PR #5545, opened 2023-10-19
docs: improve description of @type.definition capture

aMOPel:
related to #5544

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -160,7 +160,7 @@ As languages differ quite a lot, here is a set of captures available to you when
 ```scheme
 @type            ; type or class definitions and annotations
 @type.builtin    ; built-in types
-@type.definition ; identifiers in type definitions (e.g. in `typedef <type> <identifier>` in C)
+@type.definition ; identifiers in type definitions (e.g. `typedef <type> <identifier>` in C)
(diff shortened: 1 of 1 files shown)


lucario387 (commented):
Seems like a mistake that becomes the default for the capture 😨

aMOPel:
To be fair, it's equally ambiguous in the helpfiles (syntax.txt)

> 	 Typedef              A typedef

The way it is now, there are more parsers that capture the identifier, than parsers that capture the keyword.
It's up to you which way to correct things. But I believe the goal is to be consistent, which is currently not the case.

CONTRIBUTING.md
```diff
@@ -160,7 +160,7 @@ As languages differ quite a lot, here is a set of captures available to you when
 ```scheme
 @type            ; type or class definitions and annotations
 @type.builtin    ; built-in types
-@type.definition ; type definitions (e.g. `typedef` in C)
+@type.definition ; identifiers in type definitions (e.g. in `typedef <type> <identifier>` in C)
```
ObserverOfTime, line 163:
```suggestion
@type.definition ; identifiers in type definitions (e.g. `typedef <type> <identifier>` in C)
```

ObserverOfTime (approved):

PR #5548, opened 2023-10-20
Added objdump to the supported languages

ColinKennedy:
[tree-sitter-objdump](https://github.com/ColinKennedy/tree-sitter-objdump) parses [objdump](https://man7.org/linux/man-pages/man1/objdump.1.html) disassembly so users can quickly and easily parse / highlight / query the objdump text separately from the raw, generated Assembly code.

I've got some unittests / highlight tests which I can copy over to here if the maintainers are interested. The setup for running tests in nvim-treesitter changed I think since the last time I contributed a language though so if I could be directed to any documentation on how to run it now, that'd be helpful.

In a future PR, hopefully to github.com/vim/vim, I'd like to recognize .objdump / .cppobjdump etc files but I'll omit that from this PR since I hear maintainers prefer to receive those changes as downstreamed patches.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1174,7 +1174,7 @@ list.objc = {
 list.objdump = {
   install_info = {
     url = "https://github.com/ColinKennedy/tree-sitter-objdump",
-    files = { "src/parser.c", "src/scanner.cc" },
+    files = { "src/parser.c", "src/scanner.c" },
   },
   maintainers = { "@ColinKennedy" },
 }
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1171,6 +1171,14 @@ list.objc = {
   maintainers = { "@amaanq" },
 }
 
+list.objdump = {
+  install_info = {
+    url = "https://github.com/ColinKennedy/tree-sitter-objdump",
+    files = { "src/parser.c", "src/scanner.cc" },
+  },
+  maintainers = { "@ColinKennedy" },
+}
+
 list.ocaml = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-ocaml",
--- a/queries/objdump/highlights.scm
+++ b/queries/objdump/highlights.scm
@@ -1,0 +1,29 @@
+(byte) @constant
+
+[
+  (hexadecimal)
+  (integer)
+  (address)
+] @number
+
+(section_address) @number @text.underline
+
+(identifier) @variable
+(code_location (identifier) @function.call)
+(header (identifier) @keyword)
+(disassembly_section_label (identifier) @namespace)
+(disassembly_section (identifier) @namespace)
+
+[(file_offset) (discriminator)] @field
+
+(file_path) @string
+(instruction) @function
+(bad_instruction) @text.warning
+(label) @label
+(machine_code_bytes) @field
+
+["<" ">"] @punctuation.special
+["(" ")"] @punctuation.bracket
+["+" ":"] @punctuation.delimiter
+
+(comment) @comment.documentation
--- a/queries/objdump/highlights.scm
+++ b/queries/objdump/highlights.scm
@@ -6,6 +6,13 @@
   (address)
 ] @number
 
+[
+  "file" "format"
+  "File" "Offset:"
+  "discriminator"
+] @text
+"Disassembly of section " @text.title
+
 (section_address) @number @text.underline
 
 (identifier) @variable
@@ -20,7 +27,6 @@
 (instruction) @function
 (bad_instruction) @text.warning
 (label) @label
-(machine_code_bytes) @field
 
 ["<" ">"] @punctuation.special
 ["(" ")"] @punctuation.bracket
--- a/queries/objdump/injections.scm
+++ b/queries/objdump/injections.scm
@@ -1,7 +1,3 @@
-((comment) @injection.content
- (#set! injection.language "objdump"))
-
-
 ; TODO: https://github.com/nvim-treesitter/nvim-treesitter/pull/5548#issuecomment-1773707396
 ;
 ; To be added once a compatible Assembly parser is merged into nvim-treesitter
--- a/queries/objdump/injections.scm
+++ b/queries/objdump/injections.scm
@@ -7,4 +7,4 @@
 ; To be added once a compatible Assembly parser is merged into nvim-treesitter
 ;
 ; ((instruction) @injection.content
-;  (#set! injection.language "assembly"))
+;  (#set! injection.language "asm"))
--- a/queries/objdump/injections.scm
+++ b/queries/objdump/injections.scm
@@ -1,0 +1,10 @@
+((comment) @injection.content
+ (#set! injection.language "objdump"))
+
+
+; TODO: https://github.com/nvim-treesitter/nvim-treesitter/pull/5548#issuecomment-1773707396
+;
+; To be added once a compatible Assembly parser is merged into nvim-treesitter
+;
+; ((instruction) @injection.content
+;  (#set! injection.language "assembly"))
--- a/README.md
+++ b/README.md
@@ -303,6 +303,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [norg](https://github.com/nvim-neorg/tree-sitter-norg) (maintained by @JoeyGrajciar, @vhyrro)
 - [x] [nqc](https://github.com/amaanq/tree-sitter-nqc) (maintained by @amaanq)
 - [x] [objc](https://github.com/amaanq/tree-sitter-objc) (maintained by @amaanq)
+- [x] [objdump](https://github.com/ColinKennedy/tree-sitter-objdump) (maintained by @ColinKennedy)
 - [x] [ocaml](https://github.com/tree-sitter/tree-sitter-ocaml) (maintained by @undu)
 - [x] [ocaml_interface](https://github.com/tree-sitter/tree-sitter-ocaml) (maintained by @undu)
 - [x] [ocamllex](https://github.com/atom-ocaml/tree-sitter-ocamllex) (maintained by @undu)
--- a/lockfile.json
+++ b/lockfile.json
@@ -389,6 +389,9 @@
   "objc": {
     "revision": "62e61b6f5c0289c376d61a8c91faf6435cde9012"
   },
+  "objdump": {
+    "revision": "f002f632821d8006444d681abef0f024a9ef5cfe"
+  },
   "ocaml": {
     "revision": "694c57718fd85d514f8b81176038e7a4cfabcaaf"
   },
--- a/lockfile.json
+++ b/lockfile.json
@@ -390,7 +390,7 @@
     "revision": "62e61b6f5c0289c376d61a8c91faf6435cde9012"
   },
   "objdump": {
-    "revision": "f002f632821d8006444d681abef0f024a9ef5cfe"
+    "revision": "64e4741d58345c36ded639f5a3bcd7811be7f8f8"
   },
   "ocaml": {
     "revision": "694c57718fd85d514f8b81176038e7a4cfabcaaf"


clason:
Thank you. Yes, filetype detection is the job of the editor (and Neovim aligns with Vim, so that's the first -- and not difficult -- step).

Tests are only necessary for things that are specific to the queries _here_ (if they are different from your parser repo's).

lua/nvim-treesitter/parsers.lua
```diff
@@ -1170,6 +1170,14 @@ list.objc = {
   maintainers = { "@amaanq" },
 }
 
+list.objdump = {
+  install_info = {
+    url = "https://github.com/ColinKennedy/tree-sitter-usd",
```
clason, line 1175: copypasta, surely?

queries/objdump/highlights.scm
```diff
@@ -0,0 +1,20 @@
+(address) @number
+(byte) @constant
+(hexadecimal) @number
+(integer) @number
+(section_address) @number
```
ObserverOfTime, line 5:
```suggestion
(byte) @constant

[
  (hexadecimal)
  (integer)
  (section_address)
  (address)
] @number
```

queries/objdump/highlights.scm
```diff
@@ -0,0 +1,20 @@
...
+
+(disassembly_section_label) @text.title
+[(file_offset) (discriminator)] @field
+(file_path) @string
+(identifier) @namespace
```
ObserverOfTime, line 10:
```suggestion
(identifier) @variable
```
Maybe?

queries/objdump/highlights.scm
```diff
@@ -0,0 +1,20 @@
...
+(instruction) @text
```
ObserverOfTime, line 11:
```suggestion
(instruction) @function
```

queries/objdump/highlights.scm
```diff
@@ -0,0 +1,20 @@
+(address) @number
+(byte) @constant
+(hexadecimal) @number
+(integer) @number
+(section_address) @number
+
+(disassembly_section_label) @text.title
+[(file_offset) (discriminator)] @field
+(file_path) @string
```
ObserverOfTime, line 9:
```suggestion
(file_path) @string.special
```

queries/objdump/highlights.scm
```diff
@@ -0,0 +1,20 @@
...
+(identifier) @namespace
+(instruction) @text
+(label) @label
+(machine_code_bytes) @field
+(section_name) @namespace
+
+["<" ">"] @punctuation.special
+["(" ")"] @punctuation.bracket
+["+" ":"] @delimiter
```
ObserverOfTime, line 18:
```suggestion
["+" ":"] @punctuation.delimiter
```

queries/objdump/highlights.scm
```diff
@@ -0,0 +1,20 @@
...
+(integer) @number
+(section_address) @number
+
+(disassembly_section_label) @text.title
+[(file_offset) (discriminator)] @field
+(file_path) @string
+(identifier) @namespace
+(instruction) @text
+(label) @label
+(machine_code_bytes) @field
```
ObserverOfTime, line 13:
```suggestion
```

queries/objdump/highlights.scm
```diff
@@ -0,0 +1,20 @@
+(address) @number
+(byte) @constant
+(hexadecimal) @number
+(integer) @number
+(section_address) @number
+
+(disassembly_section_label) @text.title
+[(file_offset) (discriminator)] @field
```
ObserverOfTime, line 8:
```suggestion
[
  "file" "format"
  "File" "Offset:"
  "discriminator"
] @text

"Disassembly of section " @text.title
```

ObserverOfTime (changes requested):
Add an injections file for comments.
```clojure

((comment) @injection.content
 ```

File: queries/objdump/highlights.scm (hunk 3 above)
Comment (ColinKennedy) [reply queries/objdump/highlights.scm:10]: Hm. You could make a case for either I think

`(identifier)` comes up during section labels

https://github.com/ColinKennedy/tree-sitter-objdump/blob/master/test/corpus/sections.txt#L5-L8
```
Disassembly of section .text:

0000000000000000 <engine::world::World::createEntity()> (File Offset: 0x70):
   0:	41 55                                                 	push   r13
  1f0:	48 83 c3 08                                           	add    rbx,0x8
```

Everything under `.text` is part of its contents. And the lines indented `engine::world::World::createEntity()` are byte offsets from the starting address of `0000000000000000`. It felt very group-y to me. I use `(identifier)` for both `.text`  and `engine::world::World::createEntity()` so that's what led to `@namespace`. Maybe you'd prefer if I change the query to reflect the nesting? e.g. something like

```scm
(disassembly_section_label
  (identifier) @namespace)
(identifier) @variable
```

So if it's grouped, it's a namespace. But if it's free-form, it's a different highlight. What do you think?

lua/nvim-treesitter/parsers.lua
ColinKennedy: Yes, sorry. I should've waited until I was less tired before submitting. This PR notwithstanding, the parser is pretty solid, thankfully, haha. Thank you for your understanding

queries/objdump/highlights.scm
lucario387: That sounds better than the current queries

queries/objdump/highlights.scm
ColinKennedy: Oh good idea, I hadn't thought of it like that

queries/objdump/highlights.scm
ColinKennedy: Sometimes the `(instruction)` has garbage contents inside of it when objdump cannot read the instruction. For example it'll literally write `(bad)` instead of a real instruction like `pop rbx`. Do you still want `@function` in those cases or would you like me to add those exceptional cases as separate highlights to `[..."File Offset"...] @text`?

queries/objdump/highlights.scm
ColinKennedy: In retrospect this is obvious, I don't know why I didn't see it initially. Thank you for pointing it out!

queries/objdump/highlights.scm
ObserverOfTime: The free-form one should be on top.

queries/objdump/highlights.scm
ObserverOfTime: `pop` should be `@function`
`rbx` should be `@variable.builtin` (maybe)
`(bad)` should be `@text.warning`

ColinKennedy: At the moment the objdump parser doesn't attempt to parse inner assembly instructions like common x86 opcodes and registers like `pop` and `rbx`. We could have common instruction highlights even without explicitly adding parse rules but I was kind of hoping injections could solve that problem for me, since https://github.com/rush-rs/tree-sitter-asm and https://github.com/bearcove/tree-sitter-x86asm already exist.

Admittedly I haven't looked into whether injections are up to that task just yet. If it's important for a v1 to sort this now, I can start on that, otherwise I'd like to defer that decision to a v2.

If it's alright, for now though I can stick with `(instruction)` == `@function` and `(bad)` == `@text.warning`. If there's other cases I'll try to collect them and mark them separately too.

ColinKennedy: `(instruction)` from `@text` -> `@function` is done

queries/objdump/highlights.scm
ColinKennedy: > (identifier) @variable

> The free-form one should be on top.


This is done

ColinKennedy:
> Add an injections file for comments. ... `((comment) @injection.content (#set! injection.language "comment"))`

I could be misunderstanding the suggestion but my impression of injections is the ability to add extra parsing rules-within-rules using another parser language. (e.g. in Python you can get tree-sitter nodes for URLs and TODO notes, directly within comments)

In objdump comments have really specific usage so I'm not sure if injections make the most sense. For example you can have a line like this:

```
 129:	48 8d 3d 00 00 00 00                                  	lea    rdi,[rip+0x0]        # 130 <void std::vector<engine::entity::Entity, std::allocator<engine::entity::Entity> >::_M_realloc_insert<engine::entity::Entity const&>(__gnu_cxx::__normal_iterator<engine::entity::Entity*, std::vector<engine::entity::Entity, std::allocator<engine::entity::Entity> > >, engine::entity::Entity const&)+0x130> (File Offset: 0x40c)
```

Where `130` indicates the local byte offset and the `<stuff>` is the disassembly section read from and `0x40c` is the offset into that section. I've also discovered another comment flavor that allows only a hexadecimal

```
  40112c:	48 39 1d 55 06 0c 00 	cmp    QWORD PTR [rip+0xc0655],rbx        # 0x4c1788
```

These notes are directly from objdump and not part of the original assembly instruction. Would injections end up messing with those parse rules? I tried adding your suggested injection and it didn't seem to break existing parses but didn't recognize URLs like I would've expected. Probably I'm just not understanding something.

lucario387:
You can consider injection parsing as re-parsing such range of existing contents with a different parser. Hence, it will never break existing parses, but if comments are specific like this, maybe it'd be better to inject itself instead?

```
((comment) @injection.content
 (#set! injection.language "objdump"))

; Or for nightly
((comment) @injection.content
 (#set! injection.self))
```

ColinKennedy:
> maybe it'd be better to inject itself instead?

I'm not sure what reason we'd want to inject a parser into itself but I'll go with whatever recommendation you all. I'm happy to do whichever because it seems either way will safely include existing parses. That's the most important thing to me. My earlier comments were basically part of understanding your suggestion better. Please let me know what you'd like me to do and I'll update :)

queries/objdump/highlights.scm
ColinKennedy: I've added parse + highlight support for (bad). Please let me know if we're good to resolve this or if I need to start looking at assembly parsing (via injections or otherwise)

ObserverOfTime:
Would it be possible to parse the comments and inject cpp where it's relevant? Does objdump also show go or rust code?

queries/objdump/highlights.scm
ObserverOfTime: Then maybe go with `@text.literal` for instructions?

amaanq:
scanner doesn't need to be in c++, and the code was pretty much C code so I submitted a PR to change that @ColinKennedy

amaanq:
also you can add an injections.scm with a commented out injection for "asm", I'm in the works of writing a universal asm parser that can handle a bunch of flavors/architectures, x86* included

queries/objdump/highlights.scm
```diff
@@ -0,0 +1,30 @@
...
+[
+  (hexadecimal)
+  (integer)
+  (section_address)
+  (address)
+] @number
+
+(identifier) @variable
+(disassembly_section_label
+  (identifier)) @namespace
```
amaanq, line 12:
```suggestion
(identifier) @variable
(code_location (identifier) @function.call)
(header (identifier) @keyword)
(disassembly_section_label (identifier) @namespace)
(disassembly_section (identifier) @namespace)
```

queries/objdump/highlights.scm
```diff
@@ -0,0 +1,30 @@
+(byte) @constant
+
+[
+  (hexadecimal)
+  (integer)
+  (section_address)
+  (address)
+] @number
```
amaanq, line 8:
```suggestion
[
  (hexadecimal)
  (integer)
  (address)
] @number

(section_address) @number @text.underline
```

ColinKennedy:
> Would it be possible to parse the comments and inject cpp where it's relevant? Does objdump also show go or rust code?

objdump can display multiple call signatures for different languages.

<details>
<summary>Some quick proof that this works</summary>

objdump as far as I am aware is able to display the function signatures of any language compiled with debug symbols enabled. But the output quality may vary

Here's an example from rust, with `objdump --disassemble test.rs`

```
116dd:       e8 de 17 04 00          callq  52ec0 <_ZN4core5slice4sort26insertion_sort_shift_right17h022d47a99f017d4aE>
```

I read once that C and Java both have the same demangling rules. Possibly C++ too. `objdump` has a `--demangle=auto` that does a good job demangling function signatures which I tried on Rust and surprisingly it works

```
116dd:       e8 de 17 04 00          callq  52ec0 <core::slice::sort::insertion_sort_shift_right::h022d47a99f017d4a>
```

Of course even if `--demangle=auto` fails for a specific language there's still standalone programs like `c++filt` that do the same thing. All this to say objdump can display multiple call signatures for different languages.
</details>

> scanner doesn't need to be in c++, and the code was pretty much C code so I submitted a PR to change that @ColinKennedy

Thank you for your PR! And yep, I wrote as close to C as possible but since it was my first time writing a scanner I didn't know if I might need a C++ feature or two so I did scanner.cc. I'm happy to accept that PR as long as unittests pass, of course. I'll check out your PR later

queries/objdump/highlights.scm
ColinKennedy: > Then maybe go with @text.literal for instructions?

Done!

Regarding assembly highlights - It sounds like, according to https://github.com/nvim-treesitter/nvim-treesitter/pull/5548#issuecomment-1773707396, that we should be able to rely on injections after all. So I'm going to resolve this conversation but please unresolve if you think there's more things left to do here

ColinKennedy:
injections.scm is ready for those who wish to view + comment - https://github.com/nvim-treesitter/nvim-treesitter/pull/5548/files#diff-35db908ce6917a3575faad8e9898a7fdaf6d4b69d53529172e4b93a92d89bfd4

Noting that changing `scanner.cc` -> `scanner.c` from https://github.com/ColinKennedy/tree-sitter-objdump/pull/3 would require a change to both `parsers.lua` and `lockfile.json`. Or since lockfile.json is versioned by-commit, we can consider that change as separate to this PR and update those two files later as needed. I'm happy to do either (wait for the other PR to merge first or go ahead without it). Whichever maintainers here prefer.

clason:
> Or since lockfile.json is versioned by-commit, we can consider that change as separate to this PR and update those two files later as needed. I'm happy to do either (wait for the other PR to merge first or go ahead without it). Whichever maintainers here prefer.

Better to update this PR, even if it means having to wait a few days more.

ColinKennedy:
I've done all notes so far and the scanner refactor by @amaanq is now merged. Please let me know if there's more things needed for merging. Thank you to everyone who've contributed so far!

queries/objdump/injections.scm
```diff
@@ -0,0 +1,10 @@
+((comment) @injection.content
+ (#set! injection.language "objdump"))
+
+
+; TODO: https://github.com/nvim-treesitter/nvim-treesitter/pull/5548#issuecomment-1773707396
+;
+; To be added once a compatible Assembly parser is merged into nvim-treesitter
+;
+; ((instruction) @injection.content
+;  (#set! injection.language "assembly"))
```
amaanq, line 10:
```suggestion
; ((instruction) @injection.content
;  (#set! injection.language "asm"))
```

queries/objdump/injections.scm
```diff
@@ -0,0 +1,10 @@
+((comment) @injection.content
+ (#set! injection.language "objdump"))
```
ObserverOfTime, line 2: It might be more efficient to parse the comments initially than to self-inject.

ColinKennedy: I'm still not sure exactly what is meant by self-inject but if you mean "parse the comments for objdump's own purposes" then it's already doing that currently.

See [grammar.js](https://github.com/ColinKennedy/tree-sitter-objdump/blob/f002f632821d8006444d681abef0f024a9ef5cfe/grammar.js#L100-L106), [a full example parse](https://github.com/ColinKennedy/tree-sitter-objdump/blob/ee8f14c75e47eb24a2a24cea1ed47b4240e64da6/test/corpus/comment.txt#L53-L59) and [a more concise parse](https://github.com/ColinKennedy/tree-sitter-objdump/blob/ee8f14c75e47eb24a2a24cea1ed47b4240e64da6/test/corpus/comment.txt#L25-L26)

In which case I guess this injection can be removed. Just wanted to clarify in case there was a misunderstanding.

amaanq: Yea this can be removed, it's meant for stuff the parser doesn't handle, and in most normal languages a comment can have "comment" syntax - TODOs, FIXMEs, etc which you would inject the comment parser here then - but objdump won't (more like shouldn't?) have user comments

You *could* inject C/C++/Rust/Go in the demangled symbol, but detecting which language it is wouldn't be easy, so it might just be better to leave it as is

ColinKennedy: Thanks for the confirmation. Removed!

amaanq (approved):

ObserverOfTime (approved):

PR #5550, opened 2023-10-20
Highlight APKBUILD files as bash scripts

WhyNotHugo:
APKBUILDs are technically shell scripts, so this highlighting fits perfectly. `filetype=apkbuild` is commonly used because bash-related tools (e.g.: language servers) provide bogus results for it, and dedicated linting tools (e.g.: ale+apkbuild_lint) expect this filetype.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -30,6 +30,7 @@ for ft, lang in pairs {
   ecma = "javascript",
   jsx = "javascript",
   sh = "bash",
+  apkbuild = "bash",
   PKGBUILD = "bash",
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -26,6 +26,7 @@ local function register_lang(lang, ft)
 end
 
 for ft, lang in pairs {
+  apkbuild = "bash",
```
ObserverOfTime, line 29: Move it below `sh`.

WhyNotHugo: Done

amaanq (approved):

PR #5556, opened 2023-10-21
feat(nim): added nim parser and queries

aMOPel:
### Features
  
  - [x] Highlighting
  - [x] Language Injection
    - `SQL` and `RegEx` strings
    - any language you want in user defined generalized string literals
    - `std/strformat` strings, including debugging `=` and other
      [quirks](https://nim-lang.org/docs/strformat.html#standard-format-specifiers-for-strings-integers-and-floats) (see Notes)
    - `MarkDown` in documentation comments
    - `C/Cpp/ObjC/JavaScript` in emit pragma and `asm` statements (see Notes)
  - [x] Folds
  - [x] Locals

 ### NOTES

#### Format Strings

**Requires** `:TSInstall nim_format_string`.

In format strings `{`/`}` have to be escaped, which breaks the `Nim parser`,
thus expressions in format strings containing `\{`/`\}` will intentionally not be parsed.

#### Emit Pragma / asm statements

Parsing in the emit pragma requires the language name in comment on the preceding line.
E.g.:

```nim
#cpp           <- necessary
{.emit: """
...
<cpp code>
...
""".}

#javascript           <- necessary
asm """
<js code>
"""
```


lua/nvim-treesitter/parsers.lua
```diff
@@ -1128,6 +1128,22 @@ list.nickel = {
   },
 }
 
+list.nim = {
+  install_info = {
+    url = "https://github.com/alaviss/tree-sitter-nim",
+    files = { "src/parser.c", "src/scanner.cc" },
```
clason, line 1134:
```suggestion
    files = { "src/parser.c", "src/scanner.c" },
```

queries/nim/injections.scm
```diff
@@ -0,0 +1,147 @@
...
+; normal strings
+((comment) @injection.language
+  (#offset! @injection.language 0 2 0 0) . ; leading `# `
+  (pragma_statement
+    (pragma_list
+      (colon_expression
+        left: (identifier) @_emit_keyword (#eq? @_emit_keyword "emit")
+        right: (string_literal "\"" . "\"") @injection.content)
+          (#offset! @injection.content 0 1 0 -1) ; wrapping `"`
+    )))
```
lucario387, line 88: push this offset to the end so that it's easier to read the predicate and query. Also, I think the parser should be doing the heavy lifting and create a string contents node here, and then it's captured for injection

```suggestion
((comment) @injection.language
 . 
 (pragma_statement
   (pragma_list
     (colon_expression
       left: (identifier) @_emit_keyword
       right: (string_literal "\"" . "\"") @injection.content)))
 (#eq? @_emit_keyword "emit")
 (#offset! @injection.language 0 2 0 0)
 (#offset! @injection.content 0 1 0 -1))

```

aMOPel: @alaviss

queries/nim/folds.scm
```diff
@@ -0,0 +1,44 @@
...
+  (static_statement)
+
+  (try)
+  (except_branch)
+  (finally_branch)
+
+  (do_block)
+  (call
+    (argument_list
+      (statement_list)))
```
lucario387, line 31: In this case, is it better to just capture `argument_list` ? Or is the call supposed to be chained as well

aMOPel: That is about a special case in the syntax, where you can pass whole statements (not expressions) as arguments to a function. It is primarily used for macros and templates.

```nim
template_call(arg1:int, arg2:int):
  # here starts the (statement_list) as part of the (argument_list)
  arg3_statement_1 # here would be real statements
  arg3_statement_2
```

So yeah this is specifically about folding the whole call if there is a (statement_list) in the (argument_list)

queries/nim/injections.scm
alaviss: I can make this happen for strings, but only strings if that works for you.

Splitting comment would make this a non-terminal, which is not very nice in terms of `extras`.

lucario387: Only strings are more than enough.

I'm also okay with the current version, just that the code is a bit unreadable for me

aMOPel:
I'm gonna try to write indent queries after all now, since alaviss pushed some changes to the parser [just now](https://github.com/alaviss/tree-sitter-nim/pull/53)

lucario387 (commented):
These two overlaps with one another. You should add a `(#not-eq? @injection.language "fmt")` to the one above

queries/nim/injections.scm
```diff
@@ -0,0 +1,81 @@
...
+(generalized_string 
+  function: (identifier) @injection.language
+  (string_content) @injection.content)
+
+; format string in generalized_strings
+(generalized_string 
+  function: (identifier) @_string_prefix .
+  (string_content) @injection.content
+  (#set! injection.language "nim_format_string")
+  (#eq? @_string_prefix "fmt"))
```
lucario387, line 22: The second query is overlapped with the other two

```suggestion
(generalized_string 
  function: (identifier) @_string_prefix .
  (string_content) @injection.content
  (#set! injection.language "regex")
  (#any-of? @_string_prefix "re" "rex"))

; format string in generalized_strings
(generalized_string 
  function: (identifier) @_string_prefix .
  (string_content) @injection.content
  (#set! injection.language "nim_format_string")
  (#eq? @_string_prefix "fmt"))

; sql in generalized_strings
; and anything you like as long as the function name is the same as the injected language's parser
(generalized_string 
  function: (identifier) @injection.language
  (string_content) @injection.content
  (#not-any-of? @injection.language "re" "rex" "fmt"))
```

lucario387: Second thought, does injections for generalized strings with languages that's not sql, regex and fmt occur commonly? If not, I suggest to just narrow it to sql only, just like you've done for fmt and regex

aMOPel: Afaik there are no other injected languages in generalized string in the stdlib. (Beside peg, which doesn't have a tree sitter parser)

But you can always define your own generalized strings and I thought it would be cool to support injection of every language just like that. You don't think it's worth it?

@alaviss what's your opinion on this?

lucario387: > You don't think it's worth it?

I don't know to be honest. It may be nice, but if it's too niche in usage, I don't think it's worth to support it. Whatever the tree-sitter-nim authors thinks will be correct, I will be going with that opinion as well.

At least, remove the overlaps for the injections. It's never nice for 2 injection queries to overlap with one another.

aMOPel: yes will do. good catch

aMOPel: @lucario387 

A question aside:

I have been conservative with what I put into injection queries because of
https://github.com/neovim/neovim/issues/22426
> Currently we run the full injection query on every edit to a buffer.

Technically it would be correct to support a regex like that: `f[mM][tT]` and similar for other ones. But I thought it's not worth the performance cost.
Also look at this [issue](https://github.com/aMOPel/nvim-treesitter-nim/issues/1) which shows some alternative ways to write those generalized string, for which I currently didn't include queries.

Do you think this is a good decision or should I go for more correctness and performance will be fine?

lucario387: You can add whatever you think will be better, as long as the experience between keystrokes is not 100ms long

aMOPel: how can I measure that?

aMOPel: (also that is just subjective testing with my hardware)

queries/nim_format_string/highlights.scm
```diff
@@ -0,0 +1,19 @@
...
+  sign: (sign)? @operator
+  hash: (hash)? @punctuation.special
+  zero: (zero)? @field
+  min_width: (min_width)? @number
+  precision: (precision)? @number
+  type: (type)? @type)
+
+(matching_curlies 
+  nim_expression: (nim_expression 
+    escaped_curly: (escaped_curly)+ @string.escape) @error)
```
ObserverOfTime, line 19: Why `@error`?

queries/nim_format_string/injections.scm
```diff
@@ -0,0 +1,4 @@
+(
+  (matching_curlies (nim_expression !escaped_curly) @injection.content)
+  (#set! injection.language "nim")
+)
```
ObserverOfTime, line 4: Formatting.
```suggestion
((matching_curlies (nim_expression !escaped_curly) @injection.content)
  (#set! injection.language "nim"))
```

queries/nim/injections.scm
```diff
@@ -0,0 +1,82 @@
...
+  .
+  (pragma_statement
+    (pragma_list
+      (colon_expression
+        left: (identifier) @_emit_keyword (#eq? @_emit_keyword "emit")
+        right: [
+          (interpreted_string_literal (string_content) @injection.content)
+          (long_string_literal (string_content) @injection.content)
+          (raw_string_literal (string_content) @injection.content)
+        ]))))
```
ObserverOfTime, line 68: This should work.
```suggestion
        right: (_ (string_content) @injection.content)))))
```

queries/nim/injections.scm
```diff
@@ -0,0 +1,82 @@
...
+  (#eq? @_string_prefix "fmt"))
+
+; format string in normal strings with & prefix
+(prefix_expression
+  operator: (operator) @_string_prefix .
+  [
+    (interpreted_string_literal (string_content) @injection.content)
+    (long_string_literal (string_content) @injection.content)
+    (raw_string_literal (string_content) @injection.content)
+  ]
```
ObserverOfTime, line 25:
```suggestion
  (_ (string_content) @injection.content) 
```

queries/nim/injections.scm
```diff
@@ -0,0 +1,82 @@
...
+  function: (identifier) @injection.language
+  (string_content) @injection.content
+  (#not-any-of? @injection.language "re" "rex" "fmt"))
+
+; =============================================================================
+; asm statement
+
+; there is no ts parser for assembly
+; (assembly_statement (string_literal) @injection.content
+;   (#set! injection.language "assembly"))
```
ObserverOfTime, line 41:
```suggestion
;   (#set! injection.language "asm"))
```
Soon™

queries/nim/highlights.scm
```diff
@@ -0,0 +1,797 @@
+; SPDX-FileCopyrightText: 2023 Leorize <leorize+oss@disroot.org>, aMOPel <>
+; SPDX-License-Identifier: MPL-2.0
+; SPDX-License-Identifier: Apache-2.0
+
+; =============================================================================
+; catch all rules
+
+((identifier) @variable (#set! "priority" 95))
```
ObserverOfTime, line 8: Is the priority necessary?

queries/nim/highlights.scm
```diff
@@ -0,0 +1,797 @@
...
+
+(operator) @operator
+
+; =============================================================================
+; @punctuation.delimiter ; delimiters (e.g. `;` / `.` / `,`)
+
+[ "." ";" "," ":" "=" ] @punctuation.delimiter
+
+; needs to be after @punctuation.delimiter
+(assignment "=" @operator)
```
ObserverOfTime, line 45: Move the `(operator)` capture below instead?

queries/nim/injections.scm
lucario387: Yes, smoothness is also subjective. Mostly, just benchmark with plenary, or eye test :D

I'm okay with performance issues for now, we can always fix it going forward

queries/nim/highlights.scm
```diff
@@ -0,0 +1,797 @@
...
+; @keyword.function    ; keywords that define a function (e.g. `func` in Go, `def` in Python)
+
+[
+  "proc"
+  "func"
+  "method"
+  "converter"
+  "iterator"
+  "macro"
+  "template"
```
ObserverOfTime, line 363: Maybe `@preproc` for these two?

queries/nim/highlights.scm
```diff
@@ -0,0 +1,797 @@
...
+] @exception
+
+; =============================================================================
+; @type            ; type or class definitions and annotations
+
+((identifier) @type
+  (#has-ancestor? @type type_expression)
+  (#not-has-ancestor? @type pragma_list)
+  (#set! "priority" 98) ; for parameters in proc_type
+  )
```
ObserverOfTime, line 467:
```suggestion
  (#set! "priority" 98)) ; for parameters in proc_type
```

queries/nim/highlights.scm
```diff
@@ -0,0 +1,797 @@
...
+
+; =============================================================================
+; overrule things
+
+; left identifier in dot_expression
+(dot_expression left: [
+  (identifier) @none
+  (accent_quoted (identifier) @none)
+] (#set! "priority" 110))
+; NOTE: it can't be know what the left identifier is, so better leave it alone
```
ObserverOfTime, line 767:
```suggestion
; NOTE: we can't know what the left identifier is, so better leave it alone
```

queries/nim/highlights.scm
```diff
@@ -0,0 +1,797 @@
...
+      (array_construction)
+      (#eq? @_at "@"))
+    (tuple_construction)
+    (curly_construction)
+  ] @comment)
+
+; =============================================================================
+; @error                 ; syntax/parser errors
+
+(ERROR) @error
```
ObserverOfTime, line 797: See https://github.com/nvim-treesitter/nvim-treesitter/pull/5421#issuecomment-1768624719

queries/nim/highlights.scm
```diff
@@ -0,0 +1,797 @@
...
+  ]
+  right: [
+    (identifier) @function.call
+    (accent_quoted (identifier) @function.call)
+  ])
+; NOTE: will double capture as @function.call if it also has argument_list
+
+; =============================================================================
+; @function.builtin ; built-in functions
+; unused
```
ObserverOfTime, line 216: What about [these builtins](https://nim-lang.org/docs/manual.html#procedures-nonoverloadable-builtins)?

queries/nim/highlights.scm
ObserverOfTime, line 1: Please remove the unused blocks.

ObserverOfTime (changes requested):

queries/nim/highlights.scm
lucario387: Wait why do need priorities for these? Can't they be moved downwards

queries/nim_format_string/highlights.scm
clason: Do not use `@error`; that's a purely diagnostic capture that should not be used for actual highlights. Use `@text.danger` for error-type text instead.

aMOPel: alright. gonna go back to `@none` then

clason: What does that do? Why not simply skip the capture?

queries/nim/injections.scm
aMOPel: is that better for performance?

There are other places where I opted for explicit options over using `_` if they are limited.

queries/nim_format_string/highlights.scm
clason: (but, yes, `@none` is usually used for string interpolation, if that's what's happening here.)

aMOPel: https://github.com/aMOPel/tree-sitter-nim-format-string#limitations

It is supposed to highlight the nim expression as white or whatever (instead of string color), when we are not injecting the nim parser.

queries/nim/highlights.scm
aMOPel: huh.

queries/nim/highlights.scm
aMOPel: nice. I never found that reference. I was arguing with me which builtins (from system.nim) to include and didn't know where to draw the line so I left it completely.

queries/nim/injections.scm
clason: It's less likely to break if nodes are added or removed from the parser, for one thing.

aMOPel: doesn't feel like a strong argument for me. everything is likely to break, that's why the parser version is pinned.

more generic can be as bad more specific in terms of breakage.

Eg there could be a new node introduced that contains `(string_content)`

ObserverOfTime: > is that better for performance?

I don't know about performance but it's better for readability.

> Eg there could be a new node introduced that contains `(string_content)`

That would still be injectable, no?

clason: > doesn't feel like a strong argument for me. everything is likely to break, that's why the parser version is pinned.

Yes, but we still _try_ to update, and the less manual intervention is needed, the better for us and for you (as the language maintainer).

queries/nim_format_string/highlights.scm
clason: Then use `@none`.

queries/nim/injections.scm
aMOPel: > That would still be injectable, no?

Yes, but then it could be injected in the wrong places. And then it would be a silent error. This way we get immediate feedback if something is not working as intended

queries/nim/highlights.scm
aMOPel: I removed the prio here

queries/nim/highlights.scm
aMOPel: will do. But would you have thought about the @function.builtin without them? :b

queries/nim/injections.scm
lucario387: In this case, for the query to be injected in the wrong place, it means there has to be a new node with `(string_content)` that will be placed in the same scenario as this. And most likely that node will also be a string-esque that will need to be injected.

queries/nim/highlights.scm
aMOPel: Honestly I don't like this list. Put aside that it seems to be out of date, since `definedInScope` doesn't even exist apparently.

Like I said, there are just [sooo many builtin functions](https://nim-lang.org/docs/system.html#12) that could deserve to be captured here.
Eg, why should `sizeof` be in it, but not `typeof`?

Any line drawn would be arbitrary and the toll on the performance would grow and grow with the list of identifiers to check for.

queries/nim/injections.scm
aMOPel: Still. All we gain is subjective readability. But I can do it if you all insist.

aMOPel:
Just so we're on the same page, I'm gonna squash all those tiny extra commits be we merge right?

clason:
You can squash yourself to signal you're done; otherwise we squash-merge.

aMOPel:
Alright. Squash-merge sounds fine. But I'm still gonna do the indents.scm first.

aMOPel:
Is there some places I could put all my test code examples?

Also @clason , is the support for nvim query unit tests done yet?

clason:
> Is there some places I could put all my test code examples?

yes, under `/tests`. but only those that are specific to the queries you changed here; we neither need nor want to duplicate upstream parser tests.

> Also @clason , is the support for nvim query unit tests done yet?

What do you mean?

aMOPel:
> What do you mean?
 I believe a while ago you wrote on the matrix server that there is upstream (tree-sitter) support coming for unit test customization which would enable us to write actually automatic unit tests for nvim queries.

aMOPel:
> yes, under /tests. but only those that are specific to the queries you changed here; we neither need nor want to duplicate upstream parser tests.

I would take the relevant test code files from [here](https://github.com/aMOPel/nvim-treesitter-nim/tree/main/tests), that I use for manual testing of the queries.

lucario387:
> I believe a while ago you wrote on the matrix server that there is upstream (tree-sitter) support coming for unit test customization which would enable us to write actually automatic unit tests for nvim queries.

That requires a new tree-sitter release (or personal compilation of latest master) which will be arriving soon

clason:
> I believe a while ago you wrote on the matrix server that there is upstream (tree-sitter) support coming for unit test customization which would enable us to write actually automatic unit tests for nvim queries.

That would just allow us to replace our home-grown test crate with upstream tree-sitter, making it easier to run locally. It shouldn't make a difference for CI here.

clason:
> > yes, under /tests. but only those that are specific to the queries you changed here; we neither need nor want to duplicate upstream parser tests.
> 
> I would take the relevant test code files from [here](https://github.com/aMOPel/nvim-treesitter-nim/tree/main/tests), that I use for manual testing of the queries.

No, there's no need for that. If you already run these tests in _your_ CI, no need for us to run them twice. We _only_ want tests that are different from upstream (because you needed to adapt the queries).

queries/nim/highlights.scm
aMOPel: same goes for `@type.builtin` btw. With that one it would be also much more impactful on performance, because I would need to duplicate all the `@type` queries to check for all the possible identifiers.

aMOPel:
> No, there's no need for that. If you already run these tests in your CI, no need for us to run them a second time. We only want tests that are different from upstream (because you needed to adapt the queries).

I think you misunderstood. That link I posted is from the repo I am gonna archive once I could merge all my queries. That is no upstream repo.

clason:
But the upstream repo has queries, that are presumably tested. Anything that _can_ be tested there, _should_ be tested there. We just cannot run full CI (on every single PR!) for all 230+ parsers that we distribute.

aMOPel:
What upstream repo are you talking about?

queries/nim/highlights.scm
aMOPel: this priority is necessary, because the capture is most specific 
and will overrule things like (type_expression) which contain (identifier)s

I'm gonna put this in a comment

queries/nim/highlights.scm
aMOPel: sure

queries/nim/highlights.scm
lucario387: Then we can move `(identifier)` up, and then add a `(type_expression (identifier) @type)` instead of needing priorities

queries/nim/highlights.scm
aMOPel: doesn't even make a difference for my theme.

`macro` and `template` share the syntax with the other function keywords though and I am already capturing the identifier following `macro` and `template` keywords as `@function.macro` and the keywords don't appear standalone.

clason:
> What upstream repo are you talking about?

https://github.com/alaviss/tree-sitter-nim, of course.

queries/nim/highlights.scm
ObserverOfTime: The colour is irrelevant. If these define preprocessor functions, then `@preproc` is appropriate.

queries/nim/highlights.scm
ObserverOfTime: I actually thought about it from the static/addr highlights.

aMOPel:
> But the upstream repo has queries, that are presumably tested

but there are only rudimentary highlight queries (that have little in common with mine) and basically no tests in there... And those tests mean nothing in the context of nvim queries. My test files are for the nvim captures. Frankly I haven't written the comments in there yet, just the code examples since I couldn't use the unit testing locally anyways.

clason:
If upstream doesn't want your queries (which would be a bit baffling), then here is fine.

queries/nim/highlights.scm
aMOPel: `@variable` capture is already at the top.

I had it wrong in my mind. the situation is this:

1. since `(type_expression)`s can infinitely nest, we have to capture the `(identifier)`s inside as `@type` with `has-ancestor` directive,
2. to have `@variable` captures not override `@type` captures those need to come before
3. NOW to have `@type` capture not override `@parameter` captures inside `(proc_type)`s, `@type` captures need to be before `@parameter`  captures.

But you're right, I can remove both priorities and push the `@parameter` block behind the `@type` block

aMOPel:
I made a PR to @alaviss a while ago and we settled on putting the queries here only, since they are nvim specific and the tree sitter cli tests simply don't work correctly, on my queries since I use stuff like `has-ancestor` (and used to use `set priority`)

queries/nim/highlights.scm
```diff
@@ -427,8 +385,8 @@
 
 ((identifier) @type
   (#has-ancestor? @type type_expression)
-  (#not-has-ancestor? @type pragma_list)
-  (#set! "priority" 98)) ; for parameters in proc_type
+  (#not-has-ancestor? @type pragma_list))
+; NOTE: needs to be after @variable
```
lucario387, line 389: This will be costly, much more than most queries in this file

aMOPel: you're talking about the has-ancestor?

lucario387: Yep
You basically are traversing the entire depth of the tree with (not-has-ancestor), so this may lead to slowdowns

queries/nim/highlights.scm
```diff
@@ -0,0 +1,756 @@
...
+      (_ (_ (_ (identifier) @type)))
+      (_ (_ (_ (_ (identifier) @type))))
+      (_ (_ (_ (_ (_ (identifier) @type)))))
+      (_ (_ (_ (_ (_ (_ (identifier) @type))))))
+      (_ (_ (_ (_ (_ (_ (_ (identifier) @type)))))))
+      (_ (_ (_ (_ (_ (_ (_ (_ (identifier) @type))))))))
+      (_ (_ (_ (_ (_ (_ (_ (_ (_ (identifier) @type)))))))))
+      (_ (_ (_ (_ (_ (_ (_ (_ (_ (_ (identifier) @type))))))))))
+      (_ (_ (_ (_ (_ (_ (_ (_ (_ (_ (_ (identifier) @type)))))))))))
+      (_ (_ (_ (_ (_ (_ (_ (_ (_ (_ (_ (_ (identifier) @type))))))))))))
```
lucario387, line 425: I know this is to workaround tree-sitter/tree-sitter#880 but I don't think it's acceptable for nvim-treesitter

aMOPel: Why not? What else am I supposed to do?

clason: Nothing for now, and wait until tree-sitter has figured this out.

queries/nim/highlights.scm
aMOPel: I knew that. I was just being a little dumb, I benchmarked this and it looked fine, but my benchmarking was done with the ts cli, which doesn't know the has-ancestor directive.

queries/nim/highlights.scm
aMOPel: @lucario387 have a look at the latest commit, if you like that better

lucario387: Guess I have to pick a poison, priority or deep nesting 😅 

I prefer the former than the latter here.

@clason will decide.

aMOPel: Correctness wise, this solution is even a little better since it catches everything no matter how deep.
Also it got rid of the the performance hungry `has-ancestor` and `not-has-ancestor` call on every `(identifier)`

clason: Priority is tricky to get right because of injections. You may have to tweak the exact values. 

But avoiding hacky nestings or expensive `ancestor` calls are more important.

aMOPel: Tweak how? My injection queries look fine.

clason: You never know if someone doesn't inject Nim into a codeblock somewhere.

lucario387: At least make sure of this one injection: Nim codeblocks in Markdown files have correct highlighting

aMOPel:
I think something is wrong with the CI, since some cpp unit tests seems to be failing

queries/nim/highlights.scm
aMOPel: was indeed not working but I think I fixed it

queries/nim/indents.scm
lucario387, line 1: Can I get some tests for indents as well

aMOPel: I removed indents again. `indents.scm` just annoys me like crazy. There are too many cases and I have to manually test every single one. Not willing to do that. Also the queries are just super weird sometimes and I can't figure out why the indent or dedent won't happen. On top of that [alaviss's plugin](https://github.com/alaviss/nim.nvim) already does a good enough job at indenting.

aMOPel:
Alright. If you got nothing else. I'm ready to merge

ObserverOfTime (approved):

lucario387 (approved):

lucario387:
Here we go. 

Ty @alaviss and @aMOPel for the parser, queries and changes as per requests

PR #5559, opened 2023-10-22
feat: add GN (Generate Ninja)

amaanq:
https://gn.googlesource.com/gn/+/main/docs/reference.md#grammar

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -641,6 +641,15 @@ list.glsl = {
   maintainers = { "@theHamsta" },
 }
 
+list.gn = {
+  install_info = {
+    url = "https://github.com/amaanq/tree-sitter-gn",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@amaanq" },
+  readme_name = "GN (Generate Ninja)",
+}
+
 list.go = {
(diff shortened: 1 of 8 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -69,7 +69,7 @@ end
 ---@field experimental boolean|nil
 ---@field readme_name string|nil
 
----@type ParserInfo[]
```
lucario387, line 72: shouldn't this still be type?

amaanq: oops, that's extraneous but I did that to shut lua_ls up :sweat_smile:

queries/gn/highlights.scm
```diff
@@ -0,0 +1,74 @@
...
+
+(scope_access field: (identifier) @field)
+
+; Literals
+
+(string) @string
+
+(escape_sequence) @string.escape
+
+(expansion) @none
```
lucario387, line 53: Is `@none` really needed here considering it's just to nullify highlights, which can be done by not capturing at all? 🤔

amaanq: expansions are potentially inside strings, it's to remove the string hl

lucario387 (approved):

PR #5563, opened 2023-10-23
fix(leo): minor coloring fixes constant highlight, finalize function highlights, affine_group highlights

r001:
- make sure affine_group has priority, so parens and comma will not be miscolored; added `annotation` as `@attribute`
- `then` is not used as conditional, so moved to keywords
- finalize function highlight added
- differentiate between constant and variable variable identifiers

### DIFF
--- a/queries/leo/highlights.scm
+++ b/queries/leo/highlights.scm
@@ -143,6 +143,9 @@
 (transition_declaration
   name: (identifier) @function.builtin)
 
+(finalizer
+  name: (identifier) @function.builtin)
+
 (free_function_call
   (identifier) @function.call)
 
--- a/queries/leo/highlights.scm
+++ b/queries/leo/highlights.scm
@@ -13,6 +13,7 @@
  "record"
  "self"
  "struct"
+ "then"
 ] @keyword
 
  "in" @keyword.operator
@@ -43,7 +44,6 @@
 [ 
   "else"
   "if"
-  "then"
 ] @conditional
 
 [
--- a/queries/leo/highlights.scm
+++ b/queries/leo/highlights.scm
@@ -168,10 +168,12 @@
 
 [ 
   (address_literal)
-  (affine_group_literal) 
+  ((affine_group_literal) (#set! "priority" 101))
   (field_literal) 
   (product_group_literal) 
   (scalar_literal) 
   (signed_literal) 
   (unsigned_literal) 
 ] @number
+
+(annotation) @attribute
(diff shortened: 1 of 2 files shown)


queries/leo/highlights.scm
```diff
@@ -168,10 +172,12 @@
 
 [ 
   (address_literal)
-  (affine_group_literal) 
+  ((affine_group_literal) (#set! "priority" 101))
```
lucario387, line 175: Why priority for this?

Can't it be moved downwards instead

r001: Because an example affine_group_literal looks like this: `(1,2)group` unfortunately it has both parentheses and comma. While we want affine_group_literal to be colored evenly,if it is not prioritized, then the first paren and the comma will have different color - no good. After prioritization it all looks as intended: nice and even.

r001: BTW your speed is amazing. Just made the PR, and you commented... Wow... :)

amaanq: are you sure you want it to be colored evenly? it looks like some sort of point type/value

r001: Yes. This should be colored like a number literal. Evenly. This value represents a single point on a Twisted Edwards curve. It has addition and subtraction defined. So it behaves like number. So YES.

r001: I know it is weird, but the whole construct is weird.

amaanq (approved):

queries/leo/highlights.scm
lucario387: What about capturing everything below it?
Can something like `(affine_group_literal _ @capture_name)` work?

r001: Let me give you a real example: the literal `2group` is equal to `(2, 5553594316923449299484601589326170487897520766531075014687114064346375156608)group`. They are just a different notation of the same literal. The second one is NOT an expression, it is a literal. This is what is emphasized with even coloring. If we would allow different colors for the parens and comma, we would give the impression that user has entered some kind of expression. With even colors it is immediately obvious that it is just a group literal. 
The coordinates are numbers without type. But in leo there no such thing as `2` only `2u32` or the likes. We shouold not color them the same as number literals. 
We are discussing a thing that is not important. There is no programmer in the world, that will use the second notation instead of the first. It is just nice that we are prepared for this. I know its hard to to get the feeling of a language that has `(x,z)group` literals, where `then` is not a conditional, and has four different kinds of function-like constructs. If you insist I can do sub coloring of course, but that is against the languages logic.

lucario387 (approved):

PR #5564, opened 2023-10-23
feat(typoscript) Added parser and queries for TypoScript

Teddytrombone:
### Features
- [x] Highlighting
- [x] Language injection (comment)
- [x] Folds
- [x] Indents
- [ ] Locals (there is no concept like locals in TypoScript)


queries/typoscript/highlights.scm
```diff
@@ -0,0 +1,54 @@
...
+  (condition) 
+  (condition_end) 
+  (condition_else)
+] @conditional
+
+
+[
+  "@import"
+  "INCLUDE_TYPOSCRIPT"
+] @keyword
```
lucario387, line 18: Repeated and overridden by the capture a few lines below

lua/nvim-treesitter/parsers.lua
```diff
@@ -1874,6 +1874,14 @@ list.typescript = {
   maintainers = { "@steelsojka" },
 }
 
+list.typoscript = {
+  install_info = {
+    url = "https://github.com/Teddytrombone/tree-sitter-typoscript", -- local path or git repo
+    files = { "src/parser.c" }, -- note that some parsers also require src/scanner.c or src/scanner.cc
```
amaanq, line 1880:
```suggestion
    url = "https://github.com/Teddytrombone/tree-sitter-typoscript",
    files = { "src/parser.c" },
```

queries/typoscript/folds.scm
```diff
@@ -0,0 +1,3 @@
+[
+ (block)
+] @fold
```
amaanq, line 3:
```suggestion
(block) @fold
```

queries/typoscript/highlights.scm
```diff
@@ -0,0 +1,48 @@
...
+  ">"
+  "<"
+  ":="
+  "=<"
+  (condition_bool)
+] @operator
+
+[
+  ","
+] @punctuation.delimiter
```
amaanq, line 42:
```suggestion
"," @punctuation.delimiter
```

queries/typoscript/indents.scm
```diff
@@ -0,0 +1,7 @@
+[
+  (block)
+] @indent.begin
```
amaanq, line 3:
```suggestion
(block) @indent.begin
```

queries/typoscript/indents.scm
```diff
@@ -0,0 +1,7 @@
...
+
+("}") @indent.branch @indent.end
```
amaanq, line 5:
```suggestion
"}" @indent.branch @indent.end 
```

amaanq (changes requested):

queries/typoscript/highlights.scm
```diff
@@ -0,0 +1,48 @@
+(identifier) @field
```
amaanq, line 1: are all identifiers a field? seems a bit weird but that could be the case after all

Teddytrombone: Yes imho every identifier is a field. A TypoScript configuration is a hierarchical collection of configuration blocks (allthough "configuration" is maybe not the correct word). Each block has an identifier and contains fields. Each of this fields can also be a block of configuration items. The identifier can be the identifier of the block itself or a path in the configuration tree. So for example

```
foo {
    bar {
        test = 1
    }
}
```
is the same as `foo.bar.test = 1`
So in my opinion `foo` and `bar` and `test` are fields inside the configuration. Maybe you find `@property` a better match?

clason:
"locals" are for plugins like nvim-treesitter-refactor. It's OK to leave them out.

queries/typoscript/highlights.scm
amaanq: thanks! field is good as is

amaanq (approved):

queries/typoscript/highlights.scm
```diff
@@ -0,0 +1,46 @@
...
+
+[
+  "@import"
+  "INCLUDE_TYPOSCRIPT"
+] @include
+
+[
+  (comment)
+  (single_line_comment)
+] @comment
```
ObserverOfTime, line 24:
```suggestion
] @comment @spell
```

ObserverOfTime (approved):

PR #5572, opened 2023-10-25
refactor: Remove all `(ERROR) @error` captures

lucario387:
As discussed in #5421, capturing `@error` is inconsistent, requiring
deep nesting (or priority) in order to correctly have red backgrounds to
it.

Some queries has this capture, some don't. For consistency purposes,
removing all of them is more preferable.

For re-enabling error, either add `(ERROR) @error` or `(ERROR _ @error)`
to your custom queries.


amaanq (approved):

amaanq:
I'd probably just delete the error stuff instead of commenting it out

amaanq:
I'd probably just delete the error stuff instead of commenting it out

lucario387:
If I were to remove it, I'll only remove `@error` for `(ERROR)` queries, while some niche ones like redundant_escape for git and such, I'll just comment it out

queries/gitattributes/highlights.scm
```diff
@@ -42,12 +42,12 @@
...
-[
-  (pattern_negation)
-  (redundant_escape)
-  (trailing_slash)
-] @error
+; [
+;   (pattern_negation)
+;   (redundant_escape)
+;   (trailing_slash)
+; ] @error
```
ObserverOfTime, line 49: These are explicitly parsed as errors, similar to some vim syntax files. Do we want to keep them or should I remove them from the grammar?

ObserverOfTime: Maybe we could change them to `@text.warning`?

lucario387: I'll go with whatever you and clason choose

lucario387:
> I wouldn't remove all @error captures, just (hard) remove all ERROR nodes.

Just did that. Also un-comment @ObserverOfTime's gitattribute nodes

queries/rnoweb/highlights.scm
lucario387, line 1: I don't know whether deleting the file is ok but it only has `(ERROR) @error` capture, so I remove it

lucario387:
I don't particularly see how automatically adding `(ERROR)` pattern will solve anything, when specificity trumps everything (bar priorities). And `ERROR` nodes, as it's designed right now, will always be the least specific possible.

clason:
> I don't particularly see how automatically adding (ERROR) pattern will solve anything, when specificity trumps everything (bar priorities). And ERROR nodes, as it's designed right now, will always be the least specific possible.

It will solve the issue of 
1. nobody wants these for highlighting
2. some (may) want these for (tree-sitter) debugging

For case 2, one would add them with high(est) priority.

ObserverOfTime (approved):

clason:
Beat you to it ;)

PR #5577, opened 2023-10-27
injections(vue): inject typescript instead of js

lucario387:
Interpolations (`{{valid js code}}`) and directive attributes (`<element :directive="valid js code"`) can also be typescript code, hence we can inject typescript instead of javascript here and still ok with highlights (bar some changes between ts and js)

### DIFF
--- a/queries/vue/injections.scm
+++ b/queries/vue/injections.scm
@@ -57,12 +57,12 @@
 
 ((interpolation
   (raw_text) @injection.content)
-  (#set! injection.language "javascript"))
+  (#set! injection.language "typescript"))
 
 (directive_attribute
   (quoted_attribute_value
     (attribute_value) @injection.content
-    (#set! injection.language "javascript")))
+    (#set! injection.language "typescript")))
 
 (template_element
     (start_tag
--- a/tests/query/injections/vue/test-vue-injections.vue
+++ b/tests/query/injections/vue/test-vue-injections.vue
@@ -1,12 +1,12 @@
 <template>
   <span>{{"Text inside interpolation"}}</span>
-  <!--      ^ javascript -->
+  <!--      ^ typescript -->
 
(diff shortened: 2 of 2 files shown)


theHamsta (approved):

WhyNotHugo:
A project can potentially be vue+javascript without typescript. In that case, this might end up highlighting "invalid" javascript code as if it were typescript.

Without this patch, I think that valid typescript code won't be highlighted correctly. So I'd lean towards including it. Pointing out when code is invalid is the job of a language server.

Out of curiosity: what typescript-only syntax would be used in interpolated code?

lucario387:
>  In that case, this might end up highlighting "invalid" javascript code as if it were typescript.

Most (if not all) typescript-only highlights are for the types system only. Otherwise they share the same highlight with javascript

> Out of curiosity: what typescript-only syntax would be used in interpolated code?

For example, type narrowing again in
```vue
{{ (function(arg: Type) return some_oddly_uninferrable_type {})() as some_type
```

Or, for my case, primevue type definition

```vue
<slot #data="{data} : {data: SomeType}">
<!-- 
note that in this scenario, typescript parser consider this an error, 
so there may not be much of a reason to inject typescript here, 
as compared to interpolations from above
-->
```

As the version I have to run on doesn't have correct type narrowing, so I usually narrow them down myself to get the correct autocompletion

lucario387:
Also, 
> In that case, this might end up highlighting "invalid" javascript code as if it were typescript.

With #5572, we have already removed nearly all default supports for highlighting error nodes. So the goal should be more towards highlighting correct code, and let the wrong parts handle its own

theHamsta (approved):
@lucario387 anything left to do on this one?

lucario387:
@theHamsta That's all the changes

lucario387:
Pinging @clason to change `main` branch requires for `vue` from `javascript` to `typescript` :smile:

PR #5579, opened 2023-10-27
Fix Julia indentation after end and with parenthesis at the beginning of line

ronisbr:
Hi!

This pull request is my attempt to fix two Julia indentation problems.

The first one happens only when you start a new line (typing enter or `o` for example) after the word end. The following 
picture shows the before and after using the current queries:

julia-indentation-end

Notice that the cursor is positioned with indentation increased.

The other problem occurs when you have expressions like `(a == 2)` (with parenthesis) at the beginning of the line. In the current queries, the indentation is wrongly decreased. For example:

```julia
function teste(a)
    (a == 2) || (a == 3) && return 1
     return 2
end
```

is indented as follows:

```julia
function teste(a)
(a == 2) || (a == 3) && return 1
    return 2
end
```

With this PR, the indentation works as expected.


lucario387 (approved):

ronisbr:
Wow that was fast! Thanks @lucario387

PR #5583, opened 2023-10-28
feat: Another try adding asynchronous module attachment

wbthomason:
This PR attempts to revive the asynchronous module attachment feature initially added in #824 and reverted in #845.
It's motivated by the very slow initial load times of some complex grammars (e.g., `latex`), which can block the first screen update.
For me, this PR reduces time to first update on `latex` files by ~200ms - a very noticeable difference.

I believe that I've addressed at least some of the issues with the original implementation by (1) moving the async handle creation to inside the existing attach logic as much as possible, (2) explicitly closing the async handle, and (3) adding two-stage tracking of module attachment status.
I cannot reproduce #838 with this PR. However, further testing is absolutely called for before merging this work - at this point, I have "works for me"-level confidence, but my `treesitter` setup isn't very advanced.

Additionally, I think that this should be made opt-in, potentially on a per-language basis - there is a visible "flicker" when `treesitter` highlights load for the first time, and not all users will prefer this to the existing blocking load.

If there's interest in trying this feature again, I'll look into how to set this up with the config system in `nvim-treesitter`, as well as figuring out a less hacky design for the two-stage attachment tracking.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -76,6 +76,7 @@ local builtin_modules = {
 }
 
 local attached_buffers_by_module = caching.create_buffer_cache()
+local completed_attached_buffers_by_module = caching.create_buffer_cache()
 
 ---Resolves a module by requiring the `module_path` or using the module definition.
 ---@param mod_name string
@@ -510,6 +511,7 @@ function M.attach_module(mod_name, bufnr, lang)
     handle = vim.loop.new_async(vim.schedule_wrap(function()
       if not handle:is_closing() then
         resolved_mod.attach(bufnr, lang)
+        completed_attached_buffers_by_module.set(mod_name, bufnr, true)
         handle:close()
       end
     end))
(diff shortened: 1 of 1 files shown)


wbthomason:
Test failures appear to all be `jsx`/`js` related - will need to figure out what's wrong there also.

lucario387:
indent module is attached too late

wbthomason:
Yes - but I think that failure may be spurious. In manual testing on the same cases, the module attaches correctly before I can run `gg=G` - I suspect the async load just doesn't happen before the test function is ready. If this is opt-in, I'd argue that's OK - presumably users are not expecting to rely on blocking loading in order to immediately run indent on editor startup if they have chosen to use async attachment. What do you think?

clason:
Thank you, but we do not want this -- in fact, the whole module system is going to be removed (and already is on the `main` branch). Individual features should be configured and enabled individually, whichever makes most sense for each (instead of "one size fits none").

PR #5584, opened 2023-10-28
Improve indentation in Julia language

ronisbr:
This commit improves the indentation in the function declaration, function call, and in the assignment expressions.

We now have the possibility to align functions declaration in two ways, both very common in Julia community:

```julia
function teste(a,
               b,
               c;
               d,
               e)
end

function teste(
    a,
    b,
    c;
    d,
    e
)
end
```

which is controlled by the first argument position.

It is also working in the function call:

```julia
function teste(a, b, c, d)
    teste1(a,
           b,
           c,
           d)

    teste1(
        a,
        b,
        c,
        d
    )
end
```

In the previous version, the first scenario was not possible, leading to:

```julia
function teste(a,
    b,
    c;
    d,
    e)
end

    teste1(a,
        b,
        c,
        d)
```

This commit also improves how assignment expressions with multiple lines are indented. Before this commit, we have:

```julia
function teste(a, b, c, d)
    var =
        a +
            b *
                c /
                d
end
```

After this commit, we have the much nicer result:

```julia
function teste(a, b, c, d)
    var =
        a +
        b *
        c /
        d
end
```


queries/julia/indents.scm
```diff
@@ -48,3 +48,10 @@
   (block_comment)
 ] @indent.ignore
 
+((argument_list) @indent.align
+  (#set! indent.open_delimiter "(")
+  (#set! indent.close_delimiter ")"))
+
+((parameter_list) @indent.align
+  (#set! indent.open_delimiter "(")
+  (#set! indent.close_delimiter ")"))
```
theHamsta, line 57: Could you add the expected behavior as a test case? It is sufficient to place your sample as a Júlia filé in the indent test folder for Julia (if it exists already for that language)

ronisbr: Perfect! I will do this!

ronisbr: Hi @theHamsta !

I think I add all the required tests :)

theHamsta (approved):
Thank you!

PR #5585, opened 2023-10-29
feat(python): inject comment into docstring (string.documentation)

wookayin:


wookayin:
Failing tests seem irrelevant (lang = starlark)

```
Last error: 
Check failed:
scripts/check-queries.lua:96: starlark (injections): ...-linux64/share/nvim/runtime/lua/vim/treesitter/query.lua:248: Query error at 21:2. Invalid node type "class_definition":
(class_definition
```

lucario387:
Seems like so

wookayin:
Discussion: injecting `markdown_inline` instead of `comment` might be more useful for some use cases but this can be controversial. Not all python docstrings are markdown; rst can be preferred (#806, #917, and then removed in #1204).

I also just found that in https://github.com/nvim-treesitter/nvim-treesitter/pull/1204#issuecomment-1450177812 @amaanq, @clason prefers not injecting `comment` into python docstrings, possibly rejecting the idea in this PR.

lucario387:
Markdown inline also doesn't make sense. There's definitely a big enough community using rst for docs.

wookayin:
Yes, I agree. If one wants to highlight python docstrings, dynamic injection would be needed. But nevertheless I think we can benefit from having `comments` injected as a minimum viable default. Not only it highlights some common marks FIXME, TODO, NOTE, it also highlights URL links (`https://...`). Decision would be up to the core maintainers.

amaanq:
Yeah this should be in a user's config, not here, highlighting doc strings as a string, special kind of string, or comment is very much user-preference, which is why `@string.documentation` exists for users/colorschemes to highlight as they'd like

wookayin:
For more context, can you please explain what I can do with `@string.documentation`? It's just a highlight group right? How does it help to achieve nested highlight via injection? To my understanding injections still requires custom query on the user config.

ObserverOfTime:
See #4199 for rst.

amaanq:
> To my understanding injections still requires custom query on the user config.

`:h treesitter-query-modeline`

wookayin:
I know `; inherits: `, and `; extends: ` directives, but what does these help in this situation? Could you please explain more a bit specifically?

`@string.documentation` is defined in `python/highlights.scm` and in order to customize injections one would need to write `python/injections.scm`. Can `injections` include or inherit from `highlights`? Even if so we need to write some custom queries, right?

amaanq:
oh, no injections can't inherit/extend from highlights - I see what you're saying now, you want to inject all string.docs with comment right? You'd have to duplicate the queries unfortunately

wookayin:
I see, thanks. Yes, my understanding seems correct. Currently, similar (almost duplicate) rules need to be written for injections and highlights (which is actually very usual and common the case), but that's OK with me. This PR was exactly proposing that, but the maintainers didn't want to have *even* comments although I thought it'd be good to have, it's neither markdown nor RST).

amaanq:
yeah along w/ that having string.documentation *not* be injected with comments was explicit, they're a similar concept but not exactly the same, I don't think it's even the norm to have stuff like TODO FIXME etc in string docs but rather as comments even in Python

PR #5596, opened 2023-10-31
docs: remove a description about the outdated way of overriding query

wookayin:
This saying that "all queries found in the runtime directories will be
combined" was true prior to 0.8.0, but since neovim/neovim#20117 query
files are no longer concatenated but the first query file on runtimepath
will be used.

This documentation section has already been fixed in the `main` branch
but we still have to correct the wrong, outdated information for the
current versions of neovim and nvim-treesitter until it becomes the new
default.

(The old docs was written in 2021 for neovim 0.7.x, see #2011)

### DIFF
--- a/README.md
+++ b/README.md
@@ -583,18 +583,13 @@ like the `queries` folder of this plugin, e.g. `queries/{language}/{locals,highl
 Other modules may require additional queries such as `folding.scm`. You can find a
 list of all supported capture names in [CONTRIBUTING.md](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#parser-configurations).
 
-All queries found in the runtime directories will be combined.
-By convention, if you want to write a query, use the `queries/` directory,
-but if you want to extend a query use the `after/queries/` directory.
+Queries can be placed anywhere in your runtimepath under `queries/<language>`, with earlier directories taking precedence unless the queries are marked with `; extends` (see `:h treesitter-query`).
 
-If you want to completely override a query, you can use `:h set_query()`.
-For example, to override the `injections` queries from `c` with your own:
-
-```lua
-require("vim.treesitter.query").set_query("c", "injections", "(comment) @comment")
-```
-
-Note: when using `set_query`, all queries in the runtime directories will be ignored.
+Note: Since Neovim 0.8.0, the first query file on `runtimepath` will be used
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -583,18 +583,13 @@ like the `queries` folder of this plugin, e.g. `queries/{language}/{locals,highl
 Other modules may require additional queries such as `folding.scm`. You can find a
 list of all supported capture names in [CONTRIBUTING.md](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#parser-configurations).
 
-All queries found in the runtime directories will be combined.
-By convention, if you want to write a query, use the `queries/` directory,
-but if you want to extend a query use the `after/queries/` directory.
+Queries can be placed anywhere in your runtimepath under queries/<language>, with earlier directories taking precedence unless the queries are marked with ; extends; see :h treesitter-query.
```
amaanq, line 586:
```suggestion
Queries can be placed anywhere in your runtimepath under queries/<language>, with earlier directories taking precedence unless the queries are marked with ; extends - see :h treesitter-query.
```

I'd use a hyphen or period - semicolon there can be confusing there as well and might seem like it's needed after extends

amaanq (approved):

README.md
lucario387: Also nitpicked, wrap the `; extend`

clason:
The whole readme is an obsolete mess... I've refrained from touching it to avoid (more) rebase conflicts on the `main` branch, though.

README.md
wookayin: I will fix as suggested. FYI this sentence was copied literally from the main branch by @clason, I didn't change any single letter taking conflicts to consideration but these are all good to address. Anyway we cannot avoid any conflicts while rebasing (sorry about it) let's try to make the best form possible. I will modify text tomorrow.

wookayin:
Will this be merged?

camoz:
Just wanted to say that I think it would be nice if this was merged, since right now what is said in the Readme is wrong (which is kind of worse than no documentation at all).

I don't know about the 1.0 timeline, but if significantly modifying the current (master branch) Readme is really not an option, maybe just put a big fat warning at the top that it's partly outdated?

I think it would be very helpful, at least for me if would have been.

Thanks!

clason:
Thanks, I added it.

PR #5606, opened 2023-11-01
Escape shell command arguments

### LINKED ISSUE #5605: Install commands fail when stdpath contains whitespace
### Describe the bug

The `TSInstall` commands fail when `vim.fn.stdpath("data")` contains whitespace characters. This is because the command line args are concatenated into a single space-separated string:

https://github.com/nvim-treesitter/nvim-treesitter/blob/efec7115d8175bdb6720eeb4e26196032cb52593/lua/nvim-treesitter/install.lua#L294-L297

https://github.com/nvim-treesitter/nvim-treesitter/blob/efec7115d8175bdb6720eeb4e26196032cb52593/lua/nvim-treesitter/install.lua#L268-L284

### To Reproduce

1. have `XDG_DATA_HOME` with a path containing whitespace. e.g. `export XDG_DATA_HOME="/Users/foo/Library/Application Support/nvim"`
2. Run any install command, e.g. `:TSInstallSync ada`
3. Command fails because `mv` shell command errors:

```
Downloading tree-sitter-ada...
Creating temporary directory
Extracting tree-sitter-ada...
Compiling...
mv: Support/nvim/treesitter/parser/ada.so is not a directory
{
  cmd = "mv",
  opts = {
    args = { "-f", "parser.so", "/Users/foo/Library/Application Support/nvim/treesitter/parser/ada.so" },
    cwd = "/Users/foo/Library/Caches/nvim/tree-sitter-ada"
  }
}
```

### Expected behavior

`TSInstall` commands succeed without failure

```markdown

Installation ~
- OK `node` found v21.0.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
  Version: Apple clang version 15.0.0 (clang-1500.0.40.1)

OS Info:
{
  machine = "arm64",
  release = "23.0.0",
  sysname = "Darwin",
  version = "Darwin Kernel Version 23.0.0: Fri Sep 15 14:41:43 PDT 2023; root:xnu-10002.1.13~1/RELEASE_ARM64_T6000"
} ~

  - diff                ✓ . . . .
  - gitignore           ✓ . . . .
  - hcl                 ✓ . ✓ ✓ ✓
  - hurl                ✓ . ✓ ✓ ✓
  - jq                  ✓ . . . ✓
  - lua                 ✓ ✓ ✓ ✓ x
  - make                ✓ . ✓ . ✓
  - regex               ✓ . . . .
  - rst                 ✓ ✓ . . ✓
  - sql                 ✓ . . ✓ ✓

- ERROR lua(injections): ...im/0.9.4/share/nvim/runtime/lua/vim/treesitter/query.lua:259: query: invalid structure at position 2992 for language lua
  lua(injections) is concatenated from the following files:
  | [ERROR]:"/Users/foo/Library/Application Support/nvim/lazy/nvim-treesitter/queries/lua/injections.scm", failed to load: ...im/0.9.4/share/nvim/runtime/lua/vim/treesitter/query.lua:259: query: invalid structure at position 2992 for language lua
```

```text
NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1696795921

  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.9.4/share/nvim"

```

### Additional context

Changing this line:

https://github.com/nvim-treesitter/nvim-treesitter/blob/efec7115d8175bdb6720eeb4e26196032cb52593/lua/nvim-treesitter/install.lua#L275

to:

```lua
 options = string.format("%s %s", options, vim.fn.shellescape(opt)) 
```

seems to do the trick. Although I haven't read through all the code paths so i'm not sure if there's any side-effects.

likewise, the paths in `make_directory_change_for_command` should be similarly escaped:

https://github.com/nvim-treesitter/nvim-treesitter/blob/efec7115d8175bdb6720eeb4e26196032cb52593/lua/nvim-treesitter/shell_command_selectors.lua#L323-L333

baod-rate:
Fixes #5605

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -272,7 +272,7 @@ local function get_command(cmd)
       vim.list_extend(cmd.opts.args, M.command_extra_args[cmd.cmd])
     end
     for _, opt in ipairs(cmd.opts.args) do
-      options = string.format("%s %s", options, opt)
+      options = string.format("%s %s", options, vim.fn.shellescape(opt))
     end
   end
 
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -272,7 +272,11 @@ local function get_command(cmd)
       vim.list_extend(cmd.opts.args, M.command_extra_args[cmd.cmd])
     end
     for _, opt in ipairs(cmd.opts.args) do
-      options = string.format("%s %s", options, vim.fn.shellescape(opt))
+      if fn.has "win32" == 1 then
+        options = string.format("%s %s", options, opt)
+      else
+        options = string.format("%s %s", options, vim.fn.shellescape(opt))
+      end
     end
   end
 
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -328,7 +328,7 @@ function M.make_directory_change_for_command(dir, command)
       return string.format("pushd %s ; %s ; popd", cmdpath(dir), command)
     end
   else
-    return string.format("cd %s;\n %s", dir, command)
+    return string.format("cd %s;\n %s", vim.fn.shellescape(dir), command)
   end
 end
 


clason:
This seems to fail on Windows...

baod-rate:
Excluded the win32 platform from shell escaping to fix the windows tests.

baod-rate:
@clason is there anything else that needs to be addressed before this can be merged?

clason:
Is the conditional really needed here? Shouldn't Vim handle the platform dependency for `shellescape`?

baod-rate:
The tests definitely [failed before I added the conditional](https://github.com/nvim-treesitter/nvim-treesitter/actions/runs/6724542500/job/18277039084). It seems to be because the additional `shellescape` caused the ["Compile parsers" step to fail on windows builds](https://github.com/nvim-treesitter/nvim-treesitter/actions/runs/6724542500/job/18277039084#step:8:55):

```
'"if"' is not recognized as an internal or external command,
operable program or batch file.
{
  cmd = "cmd",
  opts = {
    args = { "/C", "if", "exist", "C:\\Users\\runneradmin\\AppData\\Local\\nvim-data\\tree-sitter-ada", "rmdir", "/s", "/q", "C:\\Users\\runneradmin\\AppData\\Local\\nvim-data\\tree-sitter-ada" }
  }
}
```

I'm not very familiar with windows (nor do I have a windows environment set up to debug this). Advice would be appreciated

clason:
Yeah, no -- looking deeper into this, this is definitely not the right fix at all, sorry. Only actual file names/paths should be escaped, not generic options. 

Before continuing, I'd suggest two things:
1. could you test the `main` branch of nvim-treesitter (which is the upcoming, totally rewritten, 1.0 release)?
2. macOS (for this very reason) has an `ApplicationSupport` alias without spaces -- can you just `export` that in the meantime?

baod-rate:
> Only actual file names/paths should be escaped, not generic options.

Not sure about windows' cmd, but for unix shells it's perfectly normal to quote any words on the command line (e.g. `foo --bar=baz` is equivalent to `foo "--bar=baz"`), unless we're depending on things like glob-expansion.

If that's the case, I can `shellescape` only the file paths, but they're generated at many different places and then appended on to the `command_list`, so there's a greater chance one will be missed.

> could you test the main branch of nvim-treesitter (which is the upcoming, totally rewritten, 1.0 release)?

will do

> macOS (for this very reason) has an ApplicationSupport alias without spaces -- can you just export that in the meantime?

I don't believe this is true? Neither `/Library/ApplicationSupport` nor `~/Library/ApplicationSupport` exist on my system (macOS Sonoma 14.1.1 on M1). Some google suggests that such a symlink may at point have been [created by iTerm2](https://gitlab.com/gnachman/iterm2/-/issues/8141)

baod-rate:
Good news is that this PR is superseded by the rewrite on `main`. The improved command execution (using `vim.system`) does not require any escaping, since arguments are passed directly instead of concatenated into a command line.

Unfortunately, `main` has quite a few dependencies on functions/modules that don't exist in the current release of neovim (v0.9.4). Namely, `vim.fs.joinpath`, `vim.uv` (f.k.a `vim.loop`), `vim.list_contains` (f.k.a `vim.tbl_contains`), `vim.iter`, and `vim.system`. Which makes using it quite difficult.

Do you know what the timeline is for those changes in neovim to be released? I'll leave it up to you whether this PR should be closed. If not, I can finish it up per the suggestions. Otherwise, this patch is functional enough for my personal purposes

clason:
Indeed, `main` deliberately targets Nvim 0.10 and will not become the default before that is released.

> Do you know what the timeline is for those changes in neovim to be released?

We don't currently have a fixed ETA for Nvim 0.10, as this is one of the bigger releases with a lot of heavy work on internals. Probably early next year, would be my guess.

As I wrote earlier, I don't think this patch is safe since it will escape all options, not just file paths. So I'm afraid I can't merge it as is.

clason:
Closing since this approach is not correct (and the issue is fixed -- differently -- on main).

Issue #5599, opened 2023-10-31
Clojure: dispatch character, `#`, highlight


kaiuri:
In [queries/clojure/highlights.scm#L16C1-L19C2](https://github.com/nvim-treesitter/nvim-treesitter/blob/efec7115d8175bdb6720eeb4e26196032cb52593/queries/clojure/highlights.scm#L16C1-L19C2), the Clojure `#_` highlights the next form as a comment.

```scheme
(
 (dis_expr) @comment
 (#set! "priority" 105) ; Higher priority to mark the whole sexpr as a comment
)
```

Although it has similar effects to the `;` character, they're not the same thing. `#` is a special [dispatch](https://clojure.org/guides/weird_characters#dispatch) character that when prefixed with `_`, _i.e._ `#_`, marks the next form for discarding. This is different from the `comment` macro, which yields `nil`. `#` and its prefixes tell Clojure to use a _read table_ for interpretation of the next character, e.g. `#(...)` expands into an anonymous function, `#'` calls the `var` function and so on.

- Would it be possible to remove the marking of s-expressions preceded by `#_` as comments?

- Furthermore, if possible, could we add a highlighting for `#`?

lucario387:
PR welcome

PR #5613, opened 2023-11-02
feat(clojure): highlight Clojure's `#_` dispatch character.

kaiuri:
Closes issue #5599.
- Remove `@comment` highlight for s-expressions preceded by `#_`.
- Add `@preproc` highlight for `#_`,

### DIFF
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -10,7 +10,9 @@
 ;; This can mean that sometimes things are highlighted weirdly because they
 ;; have multiple highlight groups applied to them.
 
-
+;; dispatch `discard` reader macro
+((dis_expr "#_" @preproc))
+ 
 ;; >> Literals
 
 (kwd_lit) @symbol


lucario387:
@NoahTheDuke what do you think about the change

NoahTheDuke:
I'm of two minds. On the one hand, it would be nice to have full highlighting of discarded forms. I sometimes miss them when reading code I've marked as such. On the other, I think it's important that discarded forms look like and are treated like whitespace. It should be plainly obvious from even a quick peek. If you want to have highlighting on discarded forms, you can remove the `#_` and then add it back in.

What are your thoughts?

lucario387:
> What are your thoughts?

I'm following your choices here. I don't know Clojure at all so I won't be giving opinions

NoahTheDuke:
Sorry, I should have tagged @kaiuri directly.

NoahTheDuke:
Cool, I think that's reasonable.

lucario387:
Is it a must to have both with priorities? Can we not use ir

NoahTheDuke:
I played around with this a bit and I'm not sure if there's a bug in nvim-treesitter or if I just don't understand how tree-sitter queries work (i don't really).

Your suggested change works:
```lua
((dis_expr) @comment
 (#set! "priority" 105))

((dis_expr "#_" @preproc)
 (#set! "priority" 105))
```

I would expect this to work as well, but it does not:
```lua
(
 dis_expr "#_" @preproc
 value: (_) @comment
 (#set! "priority" 105)
)
```
For some reason, this does not work. Neither does it work if I wrap the value query in parens: `value: ((_) @comment (#set! "priority" 105))`.

I'd prefer for something that has a single declaration as two declarations feels messy and like we're doing something wrong.

I'm not opposed to two queries if there's no other way, just want to avoid it where we can.

As for messy PRs, you can always rebase your branch and force-push to get clean git history.

lucario387:
Priority > specificity > order of appearance

NoahTheDuke:
Okay, I think this is worth it then.

NoahTheDuke:
Hey @kaiuri, still wanna work on this?

kaiuri:
> Hey @kaiuri, still wanna work on this?

No, I share the opinion that a change to [tree-sitter-clojure/grammar.js#L265](https://github.com/sogaiu/tree-sitter-clojure/blob/master/grammar.js#L265) might be a better solution.

clason:
Closing (for now) then.

PR #5622, opened 2023-11-03
rnoweb: add highlights query

bamonroe:
Add an empty highlights query to rnoweb. Without the empty query, the highlighting for injected languages is disabled.

### DIFF
--- a/queries/rnoweb/highlights.scm
+++ b/queries/rnoweb/highlights.scm
@@ -1,0 +1,1 @@
+;; This file is a placeholder to enable the highlighting of injected languages.


clason:
Is there really _nothing_ that can be highlighted?

bamonroe:
Not really. Rnoweb files are just latex files with injected R code. The only thing that could be highlighted in an Rnoweb file is the bit of markup that signifies an R code block. A 100% latex file is therefore a valid Rnoweb file. But without a highlights.scm file the latex highlighting doesn't kick in. This is due to some recent change I suspect from you guys, because it worked fine without this blank file until I updated this morning.

For what it's worth, I'm the maintainer of the rnoweb parser and of rnoweb-nvim, and I've never received a request to add any highlights to that injection markup, and I don't find it necessary.  I could invent something I guess, but this pull request is only really in response to nvim-treesitter requiring that file to be present to highlight the injected languages. I'm sure there was a good reason behind that change, and this seems the most trivial way around this problem for filetypes like rnoweb with no distinct highlighting.

bamonroe:
No problem, I just added some text. Does the new comment text work?

clason:
Sure, although the second line by itself would suffice.

bamonroe:
Just changed it to be the second line only.

The fold query is over an chunk/block of injected R code. There is no special rnoweb code present there, so nothing to highlight.

lucario387 (approved):
I guess there is a reason to keep that file then. Sorry for the file deletion :bowing_man:

bamonroe:
No problem. Thanks for all the hard work you guys put into this project!

PR #5624, opened 2023-11-03
feat(query): highlight parameters as `@parameter`

Akmadan23:
I feel like this just makes sense, and it's something that was missing.

### DIFF
--- a/queries/query/highlights.scm
+++ b/queries/query/highlights.scm
@@ -24,6 +24,7 @@
 ["@" "#"] @punctuation.special
 "_" @constant
 
+(parameters (identifier) @parameter)
 ((parameters (identifier) @number)
  (#match? @number "^[-+]?[0-9]+(.[0-9]+)?$"))
 


clason:
How so? There's more targeted captures for the components of  this node.

Akmadan23:
Well, the only query that involves parameters highlights them only if they are numbers, whereas in something like `(#set! conceal "whatever")` the parameter (`conceal`) is not captured.

clason:
Please show a concrete example that currently is not correctly highlighted.

clason:
OK, but this specific example shouldn't be a `@parameter` -- it's an `identifier` (which usually isn't highlighted specifically).

Akmadan23:
> it's an `identifier` (which usually isn't highlighted specifically)

Exactly, I probably said it wrong, but the fact is that identifiers inside a `parameters` group are not highlighted unless they're numbers.

> OK, but this specific example shouldn't be a `@parameter`

Why tho? How would you highlight it?

clason:
Not at all, or `@identifier`. Parameter is simply wrong.

Akmadan23:
Alright, so if I replace `@parameter` with `@identifier` it should be ok, right?

clason:
`@variable`, if anything. Still not convinced that this is a good idea, though.

lucario387:
This is capturing only `(identifier)` though. But still, I'm also not convinced. This capture, from what I can tell, is only the `(#set! identifier identifier ...)`. But this can either be a string, a number, or a named node. And there is zero way to distinguish named nodes and strings from the parser's view, so I don't see how highlighting a string in a different color helps

Akmadan23:
Ok, I didn't consider the ambiguity of the various `(identifier)`s here. Sorry for the inconvenience

amaanq:
You can always extend ts queries to have it in your own config

Akmadan23:
> You can always extend ts queries to have it in your own config

Yeah, that's exactly what I was doing and continue to do.

PR #5626, opened 2023-11-04
Add PureScript parser and queries

postsolar:


lua/nvim-treesitter/parsers.lua
```diff
@@ -1410,6 +1410,14 @@ list.puppet = {
   maintainers = { "@amaanq" },
 }
 
+list.purescript = {
+  install_info = {
+    url = "https://github.com/postsolar/tree-sitter-purescript",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers - { "@postsolar" },
```
clason, line 1418:
```suggestion
  maintainers = { "@postsolar" },
```

clason:
Textojects queries live in their own repository; please drop them here and contribute them to nvim-treesitter/nvim-treesitter-textobjects.

Also, we use different capture names and predicates than upstream; please read CONTRIBUTING.md and adjust the queries accordingly.

lucario387 (changes requested):
All of these are helix-compatible queries, but not neovim or nvim-treesitter compatible. Please change the queries to match what's listed as allowed captures in `CONTRIBUTING.md`

queries/purescript/locals.scm
```diff
@@ -0,0 +1,4 @@
+(signature name: (variable)) @definition
+(function name: (variable)) @definition
+(pat_name (variable)) @definition
+(exp_name (variable)) @reference
```
ObserverOfTime, line 4: Should these not capture the variable node?

queries/purescript/highlights.scm
```diff
@@ -0,0 +1,149 @@
...
+ (exp_record_access (variable) @field)
+
+ (signature name: (variable) @type)
+ (function name: (variable) @function)
+ (class_instance (instance_name) @function)
+ (derive_declaration (instance_name) @function)
+
+ ; true or false
+((variable) @boolean
+ (#match? @boolean "^(true|false)$"))
```
ObserverOfTime, line 130:
```suggestion
 (#any-of? @boolean "true" "false"))
```

queries/purescript/highlights.scm
```diff
@@ -0,0 +1,149 @@
...
+ ; The former one works for `tree-sitter highlight` but not in Helix/Kakoune.
+ ; The latter two work in Helix (but not Kakoune) and are a good compromise between not highlighting anything at all
+ ; as an operator and leaving it to the child nodes, and highlighting everything as an operator.
+ (exp_ticked (_) @operator)
+ (exp_ticked (exp_name (variable) @operator))
+ (exp_ticked (exp_name (qualified_variable (variable) @operator)))
+
+ (variable) @variable
+
+ ("@" @namespace)  ; "as" pattern operator, e.g. x@Constructor
```
ObserverOfTime, line 141: This is already captured as `@operator` above.

queries/purescript/highlights.scm
```diff
@@ -0,0 +1,149 @@
...
+   "<="
+   "⇐"
+   "->"
+   "→"
+   "<-"
+   "←"
+   "\\"
+   "`"
+   "@"
+ ] @operator
```
ObserverOfTime, line 79: Some of these might be `@punctuation.delimiter` or `@punctuation.special`.
`(wildcard)` should be `@parameter`.

queries/purescript/highlights.scm
```diff
@@ -0,0 +1,149 @@
...
+ (integer) @number
+ (exp_negation) @number
+ (exp_literal (number)) @float
+ (char) @character
+ [
+   (string)
+   (triple_quote_string)
+ ] @string
+
+ (comment) @comment
```
ObserverOfTime, line 13:
```suggestion
 (comment) @comment @spell
```

queries/purescript/highlights.scm
```diff
@@ -0,0 +1,149 @@
...
+ ";"
+ ] @punctuation.delimiter
+
+
+; ----------------------------------------------------------------------------
+; Keywords, operators, includes
+
+ ; This needs to come before the other "else" in
+ ; order to be highlighted correctly
+ (class_instance "else" @keyword)
```
ObserverOfTime, line 39: Vice-versa in neovim.

queries/purescript/highlights.scm
```diff
@@ -0,0 +1,149 @@
...
+   "as"
+   "hiding"
+   "do"
+   "ado"
+   "forall"
+   "∀"
+   "infix"
+   "infixl"
+   "infixr"
+ ] @keyword
```
ObserverOfTime, line 107: Some of these might be `@type.qualifier` (i.e. modifiers). "forall" could be `@repeat`.

ObserverOfTime (changes requested):

queries/purescript/locals.scm
postsolar: Semantically it's something else. It's just that it appears in the tree under this name.

queries/purescript/highlights.scm
postsolar: Almost all of them are used specifically as operators, and I don't really see which ones would be a good fit for `@punctuation.*`, except for `@` perhaps. Could you elaborate on which ones you have in mind?

`(wildcard)` has a bit of an unfortunate name, it is actually a lambda constructor, similar to `=>` in JS, so I wouldn't call it a parameter. Would you suggest something else for it?

queries/purescript/highlights.scm
postsolar: Sorry, it's not obvious to me which ones you have in mind for `@type.qualifier`?

"forall" is "for all" precisely as in logic (universal quantification), not the for-loop.

queries/purescript/locals.scm
ObserverOfTime: All right. The function one should be `@definition.function`. I don't know what the other two are.

queries/purescript/highlights.scm
ObserverOfTime: The dot is definitely a delimiter. "::" and "∷" could be as well.
"\\\\" and "`" might be either delimiters or special depending on their usage.

queries/purescript/highlights.scm
ObserverOfTime: I was thinking the infix ones and maybe also "foreign".

queries/purescript/locals.scm
```diff
@@ -0,0 +1,4 @@
+(signature name: (variable)) @definition
+(function name: (variable)) @definition
+(pat_name (variable)) @definition
```
ObserverOfTime, line 3:
```suggestion
(signature name: (variable)) @definition.type
(function name: (variable)) @definition.function
(pat_name (variable)) @definition
```
I don't know if the last one can be more specific.

queries/purescript/highlights.scm
postsolar: Re dot: agreed, will fix.

I think special-casing "∷" relative to operators would look weird in source code. PureScript would often be formatted like this:
image

`\` is used to introduce a lambda: `\a -> expr`, compare it with JS' `a => expr`. `` ` `` is used to make a function a named infix operator: `add 1 2` would be "regular" notation, and ``1 `add` 2`` is the infix notation with "`".

Wildcards can be used in different contexts:
```purescript
f = case _ of -- lambda-case, the wildcard creates an anonymous function
  x → x

f = if _ then _ else _  -- this is a 3-argument anonymous function
f = if _ then 1 else 2  -- 1-argument anonymous function, etc...

f = (_ + 2)
f = (2 `add` _)         -- operator sections use wildcards to create lambdas

f = _.field             -- anonymous function which accesses field "field" of a record

f = { field: _ }        -- anonymous function which puts a value into field "field"

f = _ { field = _ }     -- two-argument lambda which updates the field "field"
```

queries/purescript/highlights.scm
postsolar: The infix ones are not related to types, they declare an operator (syntax), like this:
```
infixl 6 add as + -- now you can use '+'
```

`foreign` also does not modify the type of what it's directed at, it's a keyword (or more specifically, part of it) for using foreign function interface, so the type of a value does not depend on whether it's coming from `foreign` or not. I also don't think it fits well with `@include` because the actual inter-module imports in PS are strictly separated from the values declaration and can only happen at the top of a module, whereas foreign definitions can occur anywhere. also foreign definitions must come from a file with the same name but different extension, so it's more of a "same place" thing rather than a proper "import" which deals with actual outside files.

queries/purescript/highlights.scm
ObserverOfTime: "∷" can remain an operator.
"\\\\" and "`" could be `@punctuation.special`.
`(wildcard)` is definitely `@parameter`.

queries/purescript/highlights.scm
ObserverOfTime: The `@type.qualifier` is an unfortunate name. Think of them as modifiers.

queries/purescript/highlights.scm
```diff
@@ -27,6 +27,8 @@
  [
  (comma)
  ";"
+ (qualified_module)  ; grabs the `.` (dot), ex: import System.IO
+ "."
```
ObserverOfTime, line 31: Do you need both?

postsolar: Yes, dots are also used for record access, not just modules.

theHamsta:
I tried to run vigoux's fuzzing action on this project but apparently there's a upstream bug in tree-sitter regarding some Unicode characters https://github.com/theHamsta/tree-sitter-purescript/actions/runs/6755617856/job/18364229460

queries/purescript/highlights.scm
postsolar: After thinking about it more, I'm strictly against doing anything to `` ` ``. Consider the examples above (highlighted by Github): this is the classic representation that's been in use for many years, both by PureScript and Haskell and which is based on ergonomics. The function in infix form serves the role of an operator and as such is highlighted as an operator, where the ticks, highlighted the same color, visually extend it and make it easily recognizable. I ran a number of Vim themes with ticks being `@punctuation.special`, but with this change it feels like these ergonomics are gone and the ticks only serve a distracting role.

Wrt the backslash, it makes no sense to edit it without also editing the `->` arrow, since they work together to create a visually distinguishable structure (`\arg -> expr`). But then to preserve the coherence we'd also need to edit `->` in case expression and probably special-case it as a type operator and in functional dependencies. If it's special-cased as a type operator it'd look really weird, on top of not making sense semantically. I am absolutely sure 99% of PureScript users will not be happy about it at all.

Wildcards being `@parameter` seems to be a good suggestion, I will add it.

theHamsta: Would also agree than infix ` makes sense as operator. This is also how it is currently handled for Haskell https://github.com/nvim-treesitter/nvim-treesitter/blob/96f55f304332ca6ea1b7dde32d3ec04b5298c316/queries/haskell/highlights.scm#L122

queries/purescript/highlights.scm
postsolar: I see, but in `CONTRIBUTING.md` it's listed in the `Types` section, while `infix[lr]` and `foreign import` don't in any way act on types and are semantically closer to JS' `function` or Python's `def` keyword. They don't _define_ a function however because the definition happens elsewhere, so I think `@keyword` is the most sensible and least ambiguous choice here.

ObserverOfTime: Kotlin uses `@type.qualifier` for `infix`.

postsolar: Haskell uses `@keyword`, as well as Github's and VS Code's Atom-based grammars and [TryPureScript](https://try.purescript.org) highlights. I tried running it with `@type.qualifier` and I noticed that even among the most popular and "recommended" (listed in the wiki) themes, only a few of them define a separate color for `@type.qualifier` (everforest, nord). The rest either just don't define it which makes it fallback to `@type`, which in turn looks really bad because it's a keyword blending into non-keyword "area", or define it the same as `@keyword` (kanagawa, catppuccin, sonokai). In fact among 11 themes only 2 make `@type.qualifier` styled differently from `@type`/`@keyword`, so effectively it's a keyword 40% of the time and type (wrong) 40% of the time, which IMO is not worth it.

queries/purescript/highlights.scm
```diff
@@ -0,0 +1,150 @@
+; ----------------------------------------------------------------------------
+; Literals and comments
+
+ (integer) @number
+ (exp_negation) @number
```
amaanq, line 5:
```suggestion
[
  (integer)
  (exp_negation)
] @number
```

queries/purescript/highlights.scm
amaanq: Isn't that an issue w/ the themes and not us then? It's nice to be more accurate here, and while I don't know purescript, some of these keywords do look like qualifiers/modifiers

queries/purescript/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+
+ (type_role_declaration
+   "role" @keyword
+   role: (type_role) @type.qualifier)
+
+ (hole) @character.special
+
+ ; `_` wildcards in if-then-else and case-of expressions,
+ ; as well as record updates and operator sections
+ (wildcard) @parameter
```
amaanq, line 112: Maybe `@character.special`?

queries/purescript/highlights.scm
amaanq: Isn't that an issue w/ the themes and not us then? It's nice to be more accurate here, and while I don't know purescript, some of these keywords do look like qualifiers/modifiers

queries/purescript/highlights.scm
ObserverOfTime: It's more of a parameter (see scala).

queries/purescript/highlights.scm
ObserverOfTime: It's also an issue with us having an inaccurate name but there's not much of a reason to change it before 1.0.

postsolar: I think accuracy is nice but I see neither an accuracy nor a looks gain here, and I described above what could be lost. Of course it would be impossible by definition for different languages to have the same mapping to Neovim's capture groups, but it seems more natural to me in the face of ambiguity to opt for the choice that is more generic and also results in better UX.

amaanq (approved):

ObserverOfTime (approved):

lucario387 (approved):

PR #5639, opened 2023-11-08
fix(kotlin): improve highlights

ObserverOfTime:
Built-in variables, jumps, and type arguments.

Replaces #5527 (CC @anthony-S93)


lucario387:
what's the correct commit id here

amaanq (approved):

ObserverOfTime:
Well the lockfile was already updated so I'll drop that change.

PR #5649, opened 2023-11-10
fix(htmldjango): improve highlights

ObserverOfTime:
Instead of increasing all the priorities, I opted for reducing the string priority in html_tags instead.
This shouldn't cause an issue in html and it might even help other template languages as well.


lucario387 (approved):
Definitely wont affect html and vue, svelte and astro idk, but they will probably like that ad well since webdev land is a bunch of javascript inside attribute value

PR #5659, opened 2023-11-11
parsers: add tree-sitter-slang

theHamsta:


queries/cmake/locals.scm
```diff
@@ -0,0 +1,11 @@
...
+  (identifier) @_command
+  (argument_list
+    .  (argument) @definition.var)
+  (#any-of? @_command
+   "set" "SET"
+   "option" "OPTION"))
+
+(variable_ref) @reference
+
+(source_file) @scope
```
lucario387, line 11: Hmm, I don't think this file was intended to be here

theHamsta:
Will probably rename to `shaderslang` as Neovim already has a syntax file for `slang` (probably S-lang https://en.wikipedia.org/wiki/S-Lang)

queries/cmake/locals.scm
theHamsta: sorry. fixed

theHamsta:
For context: some ecosystem PRs that I plan to do or already did
- [x] fix slangd (LSP for slang) to work with Neovim
- [ ] add slangd config to lspconfig
- [ ] add slang ftdetect to Vim (and subsequently port patch to Neovim)

Any recommendation on the filetype name?

At the moment Vim/Neovim recognize `.sl` as `slang` (S-lang). I thought I could use `shaderslang` for `.slang` to disambiguate.

clason:
I think `shaderlang` would be better, yes (if in doubt, be explicit). Is there precedent in other editors? What is the official "long name" for the language (link, pls)?

theHamsta:
It goes officially by slang but the org-name is shader-slang https://github.com/shader-slang/slang. I'm not aware of any editor integration apart from VSCode via their extension.

The language is build to extend HLSL with some convenience features and you can also compile to other shader languages (relation to HLSL, is similar like TS to JS, or fennel/typed Lua to Lua)

I think this is the first 4th level derived parse :stuck_out_tongue:

clason:
Then I'd go with `shaderslang`. 

> I think this is the first 4th level derived parse 😛

"Yo, dawg, we heard you like shaders!"

theHamsta:
Anything missing here?

queries/slang/highlights.scm
```diff
@@ -0,0 +1,42 @@
+; inherits: hlsl
+
+[
+ "var"
+ "let"
+ "This"
```
amaanq, line 6: is `This` like js/ts? maybe `@variable.builtin` would be nicer then (I could be completely wrong)

amaanq (approved):

PR #5664, opened 2023-11-12
fix(nim): literals in pragmas in types would be captured as `@identifier`

aMOPel:


lucario387:
Not again ... :)

This kindof means priority >100 still has lower precedence than specificity? Something is wrong here

lucario387:
Also, I don't see any `@identifier` in this highlight query. Are you sure it's not something in your custom config?

aMOPel:
My mistake, of course it's about `@variable` not `@identifier`, since that's not a thing in nvim-treesitter.

aMOPel:
> This kindof means priority >100 still has lower precedence than specificity? Something is wrong here

No. Not really. I could get rid of the priority sort of as a side effect, because I had to be specific about capturing only `(identifier)`s.

Before, I captured the whole content of the `(pragma_list)` as `@variable`, but that was just wrong, since there can be literals in there, which would then captured wrongly.

aMOPel:
> Not again ... :)

You thought the nightmare was over? :D

ObserverOfTime (approved):

PR #5665, opened 2023-11-12
Leo language fix: highlights program_id and this_program_id correctly

r001:
Minor fix for Leo language: now filenames are highlighted correctly.


amaanq:
If it's just a lockfile update (no query change) then there's no need for a PR as that's automatically updated every 24hrs

PR #5669, opened 2023-11-12
feat(markdown): conceals for bullets, block quotes

ribru17:
Adds conceals for Markdown bullets and block quotes. Preview:

moremdconceals

**KNOWN LIMITATION:** When using a bullet with an immediate indented code block, e.g.:

```md
-     print('code here')
```

the conceals will display this as:

```md
- print('code here')
```

As far as I know there is no (easy?) way to fix this, as (unlike with the block quote markers) the bullet point capture fills whitespace until the first non-whitespace content is reached. However I think little to no users will be affected by this, and even so, indented blocks still display all preceding whitespace on a new line in a bullet. E.g.:

```md
- This works
  
      print('code here')
```

### DIFF
--- a/queries/markdown/highlights.scm
+++ b/queries/markdown/highlights.scm
@@ -39,9 +39,17 @@
   @punctuation.special
   (#offset! @punctuation.special 0 0 0 -1)
   (#set! conceal "•"))
+([(list_marker_plus) (list_marker_star)]
+  @punctuation.special
+  (#any-of? @punctuation.special "+" "*")
+  (#set! conceal "•"))
 ((list_marker_minus)
   @punctuation.special
   (#offset! @punctuation.special 0 0 0 -1)
+  (#set! conceal "—"))
+((list_marker_minus)
+  @punctuation.special
+  (#eq? @punctuation.special "-")
   (#set! conceal "—"))
 
 (code_fence_content) @none
--- a/queries/markdown/highlights.scm
+++ b/queries/markdown/highlights.scm
@@ -34,28 +34,15 @@
   (info_string (language) @conceal
   (#set! conceal "")))
 
-;; Conceal blockquote markers
-((block_quote_marker) @punctuation.special
-  (#offset! @punctuation.special 0 0 0 -1)
-  (#set! conceal "▐"))
-((block_continuation) @punctuation.special
-  (#eq? @punctuation.special ">")
-  (#set! conceal "▐"))
-((block_continuation) @punctuation.special
-  (#eq? @punctuation.special "> ")
-  (#offset! @punctuation.special 0 0 0 -1)
-  (#set! conceal "▐"))
-((block_continuation) @punctuation.special
-  ; for indented code blocks
-  (#eq? @punctuation.special ">     ")
-  (#offset! @punctuation.special 0 0 0 -5)
-  (#set! conceal "▐"))
-
 ;; Conceal bullet points
-([(list_marker_minus) (list_marker_plus) (list_marker_star)]
+([(list_marker_plus) (list_marker_star)]
   @punctuation.special
   (#offset! @punctuation.special 0 0 0 -1)
   (#set! conceal "•"))
+((list_marker_minus)
+  @punctuation.special
+  (#offset! @punctuation.special 0 0 0 -1)
+  (#set! conceal "—"))
 
 (code_fence_content) @none
 
--- a/queries/markdown/highlights.scm
+++ b/queries/markdown/highlights.scm
@@ -34,6 +34,29 @@
   (info_string (language) @conceal
   (#set! conceal "")))
 
+;; Conceal blockquote markers
+((block_quote_marker) @punctuation.special
+  (#offset! @punctuation.special 0 0 0 -1)
+  (#set! conceal "▐"))
+((block_continuation) @punctuation.special
+  (#eq? @punctuation.special ">")
+  (#set! conceal "▐"))
+((block_continuation) @punctuation.special
+  (#eq? @punctuation.special "> ")
+  (#offset! @punctuation.special 0 0 0 -1)
+  (#set! conceal "▐"))
+((block_continuation) @punctuation.special
+  ; for indented code blocks
+  (#eq? @punctuation.special ">     ")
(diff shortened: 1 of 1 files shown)


queries/markdown/highlights.scm
```diff
@@ -34,6 +34,29 @@
...
+((block_continuation) @punctuation.special
+  (#eq? @punctuation.special ">")
+  (#set! conceal "▐"))
+((block_continuation) @punctuation.special
+  (#eq? @punctuation.special "> ")
+  (#offset! @punctuation.special 0 0 0 -1)
+  (#set! conceal "▐"))
+((block_continuation) @punctuation.special
+  ; for indented code blocks
+  (#eq? @punctuation.special ">     ")
```
lucario387, line 50: Why 4, what if it's 2, 3, ...

queries/markdown/highlights.scm
```diff
@@ -34,6 +34,29 @@
...
+  (#offset! @punctuation.special 0 0 0 -5)
+  (#set! conceal "▐"))
+
+;; Conceal bullet points
+([(list_marker_minus) (list_marker_plus) (list_marker_star)]
+  @punctuation.special
+  (#offset! @punctuation.special 0 0 0 -1)
+  (#set! conceal "•"))
```
lucario387, line 58: This will lead to some confusion, as there's a reason after all to have `-`, `+` and `*` list markers. Forcing GFM-style markers here is not favorable imo

ribru17: I see. Which node(s) do you think this query should apply to?

lucario387: imo, There's an argument for any of the 3 to be applied by this query, just not all 3.

queries/markdown/highlights.scm
ribru17: Due to the way the parser is implemented, there can never be such a case. The only `block_continuation` marker contents are `>`, `> `, and `>     ` (if there is an indented code block). Any variable amount of spaces other than these are not captured in the `block_continuation` group, and thus they work nicely here. For example, in:
```md
>   block continuation with three spaces
```
The first `> ` (caret with one space) will be captured under the block continuation node, and the rest of the spaces are captured under the next `paragraph` node.

amaanq:
To me, some of these conceals feel like they should be in a plugin or just in your own user config - not everyone might like it

ribru17:
> To me, some of these conceals feel like they should be in a plugin or just in your own user config - not everyone might like it

That makes sense. Maybe we should wait on this until #2825 is fixed?

queries/markdown/highlights.scm
ribru17: My vote would be `-` (or `-` and `*`), I guess we'll see what others say

clason:
This will never be "fixed", unfortunately, so user config (or plugin) is the only path in the foreseeable future,

Personally, I wouldn't mind the bullet for star (and plus, while hyphen gets a unicode dash) as default -- also keeping in mind the use of these queries for LSP docs -- but would definitely draw the line at the block quotes which are just too much. (A big issue here is that terminal fonts may not have all such unicode glyphs.)

ribru17:
That makes sense. I have removed block quote conceals and changed the `list_marker_minus` conceal to Em Dash (`U+2014`). See what you think. I wish there was an easier way to see how well-supported certain unicode characters are, I have yet to find one...

clason:
@MariaSolOs would that interfere (or improve) LSP Markdown rendering?

MariaSolOs:
@clason I think that conceal for bullet lists would be nice. I don't have strong opinions about block quotes since I don't think it's even a common scenario in LSP Markdown responses.

queries/markdown/highlights.scm
```diff
@@ -34,6 +34,24 @@
...
+  (#any-of? @punctuation.special "+" "*")
+  (#set! conceal "•"))
+((list_marker_minus)
+  @punctuation.special
+  (#offset! @punctuation.special 0 0 0 -1)
+  (#set! conceal "—"))
+((list_marker_minus)
+  @punctuation.special
+  (#eq? @punctuation.special "-")
+  (#set! conceal "—"))
```
lucario387, line 53: As there can be arbitrary amount of whitespaces after `-*+`, can something like this be better?

```suggestion
((list_marker_minus)
  @punctuation.special
  (#gsub! @punctuation.special "(.)%s*" "%1")
  (#set! conceal "—"))
```

lucario387:
lines with em-dash looks a bit bad imo
```
—List
  •List
  •List
```
If there's a way to have spaces for conceal, it would look a bit less cramped.

queries/markdown/highlights.scm
lucario387: nvm I misunderstood what the code does

ribru17:
@clason I am realizing it would be very useful in this case to have some way to grab the first `n` (1 in this case) columns of the node's range. Would you be opposed to me making a separate PR (or amending this one) to add a directive like the following? (If you oppose I think the current solution still works fine, unless you count the indented code block conceal behavior mentioned above.)

```lua
-- something like the following would be quite helpful here
local offset_first_n = function(match, _, _, pred, metadata)
  ---@cast pred integer[]
  local capture_id = pred[2]
  if not metadata[capture_id] then
    metadata[capture_id] = {}
  end

  local range = metadata[capture_id].range
    or { match[capture_id]:range() }
  local offset = pred[3] or 0

  range[4] = range[2] + offset
  metadata[capture_id].range = range
end

vim.treesitter.query.add_directive(
  'offset-first-n!',
  offset_first_n,
  true
)
```

lucario387:
It can be done here, but the usage seems to niche, and is too similar to offset imo

ribru17:
> It can be done here, but the usage seems to niche, and is too similar to offset imo

Ah, I see, thank you for the information.

clason:
Yes, we need to be _very_ conservative in adding new predicates, as this makes the queries even more tied to Neovim and harder to share with upstream and other editors.

ribru17:
Sounds good. In any case I think this PR is ready to go without it

amaanq (approved):

PR #5686, opened 2023-11-17
chore(rust): update queries from upstream

amaanq:


clason:
apropos: you probably want to update upstream tests as well ;)

amaanq:
> apropos: you probably want to update upstream tests as well ;)

I did :grin:

clason (approved):

PR #5695, opened 2023-11-19
Fix robot parser metadata in parsers.lua

Hubro:


lucario387 (approved):
Lovely <3

Always loved it when the parser maintainer decided to also maintain this downstream nvim-treesitter

PR #5696, opened 2023-11-19
Add parser for "Just"

Hubro:
**Edit:** Currently blocked by the upstream custom scanner, which is written in C++ and fails to compile on MacOS. See issue: https://github.com/IndianBoy42/tree-sitter-just/issues/23. Unfortunately I don't know C or C++ so I can't fix the issue myself, and the original author is not active.

---

I did not make this parser, but I use it constantly and it would be nice to have it in nvim-treesitter.

I wrote the injections and added a couple of highlight queries, otherwise this is all @IndianBoy42's work.

I have listed myself as maintainer, as I'm not sure if it refers to the maintainer of the parser or the maintainer of the associated configuration in this repository, and I guessed the latter.

The injections could be improved, but I'm not sure how.

To very roughly explain, a justfile can look like this:

```just
deploy: deploy-foo deploy-bar
    echo "This is a regular recipe and defaults to Bash"

deploy-foo:
    #!/usr/bin/env bash
    echo "This is a shebang recipe and can be set to any language"

deploy-bar:
    #!/usr/bin/env python
    print("Including Python!")
```

The injections currently capture the executable name in the shebang line ("bash" and "python" in these examples) and sets that to the injection language, which works phenomenally!

The only issue is that the executable name doesn't necessarily match the language, such as with "python3" and probably many others. Are there any good ways to deal with that currently, or do I just have to add a whole bunch of extra queries to capture most error scenarios?

The same is also true for the injection used for normal recipes when a shell is explicitly set:

```just
set shell := ["python"]
```

### DIFF
--- a/queries/just/highlights.scm
+++ b/queries/just/highlights.scm
@@ -1,39 +1,102 @@
 
-(assignment (NAME) @variable)
-(alias (NAME) @variable)
-(value (NAME) @variable)
-(parameter (NAME) @variable)
-(setting (NAME) @keyword)
-(setting "shell" @keyword)
-
-(call (NAME) @function)
-(dependency (NAME) @function)
-(depcall (NAME) @function)
-(recipeheader (NAME) @function)
-
-(depcall (expression) @parameter)
-(parameter) @parameter
-(variadic_parameters) @parameter
-
-["if" "else"] @conditional
-
-(string) @string
-
-(boolean ["true" "false"]) @boolean
-
 (comment) @comment
+(string) @string
+(boolean ["true" "false"]) @boolean
+["if" "else"] @conditional
+["export" "alias" "set"] @keyword
+["@" "==" "!=" "+" ":="] @operator
+[ "(" ")" "[" "]" "{{" "}}" "{" "}"] @punctuation.bracket
 
+[
+  (assignment (NAME))
+  (alias (NAME))
+  (value (NAME))
+  (parameter (NAME))
+] @variable
+
+; Recipe definitions
+[
+  (recipeheader (NAME))
+  (dependency (NAME))
+  (dependency (depcall (NAME)))
+] @function
+(parameter) @parameter
+(depcall (expression (value (NAME) @parameter)))
+
+; Fallback highlighting for recipe bodies
 (recipe
   (body) @text.literal.block (#set! "priority" 90))
-
 (shebang_recipe
   (shebang interpreter: (TEXT) @keyword) @comment
   (shebang_body) @text.literal.block (#set! "priority" 90))
 
-["export" "alias" "set"] @keyword
+; Ref: https://just.systems/man/en/chapter_26.html
+(setting (NAME) @error)
+(setting (NAME) @constant.builtin
+  (#any-of? @constant.builtin 
+    "allow-duplicate-recipes"
+    "dotenv-filename"
+    "dotenv-load"
+    "dotenv-path"
+    "export"
+    "fallback"
+    "ignore-comments"
+    "positional-arguments"
+    "shell"
+    "tempdir"
+    "windows-powershell"
+    "windows-shell"))
+; Special handling for the shell setting
+(setting "shell" @constant.builtin)
+(setting lang: (NAME) @string.special)
 
-["@" "==" "!=" "+" ":="] @operator
-
-[ "(" ")" "[" "]" "{{" "}}" "{" "}"] @punctuation.bracket
-
-(ERROR) @error
+; Ref: https://just.systems/man/en/chapter_31.html
+(call (NAME) @error)
+(call (NAME) @function
+  (#any-of? @function 
+    "arch"
+    "num_cpus"
+    "os"
+    "os_family"
+    "env_var"
+    "env_var_or_default"
+    "env"
+    "invocation_directory"
+    "invocation_directory_native"
+    "justfile"
+    "justfile_directory"
+    "just_executable"
+    "quote"
+    "replace"
+    "replace_regex"
+    "trim"
+    "trim_end"
+    "trim_end_match"
+    "trim_end_matches"
+    "trim_start"
+    "trim_start_match"
+    "trim_start_matches"
+    "capitalize"
+    "kebabcase"
+    "lowercamelcase"
+    "lowercase"
+    "shoutykebabcase"
+    "shoutysnakecase"
+    "snakecase"
+    "titlecase"
+    "uppercamelcase"
+    "uppercase"
+    "absolute_path"
+    "extension"
+    "file_name"
+    "file_stem"
+    "parent_directory"
+    "without_extension"
+    "clean"
+    "join"
+    "path_exists"
+    "error"
+    "sha256"
+    "sha256_file"
+    "uuid"
+    "semver_matches"))
--- a/queries/just/highlights.scm
+++ b/queries/just/highlights.scm
@@ -27,7 +27,6 @@
   (shebang_body) @text.literal.block (#set! "priority" 90))
 
 ; Ref: https://just.systems/man/en/chapter_26.html
-(setting (NAME) @error)
 (setting (NAME) @constant.builtin
   (#any-of? @constant.builtin 
     "allow-duplicate-recipes"
@@ -47,7 +46,6 @@
 (setting lang: (NAME) @string.special)
 
 ; Ref: https://just.systems/man/en/chapter_31.html
-(call (NAME) @error)
 (call (NAME) @function
   (#any-of? @function 
     "arch"
--- a/queries/just/highlights.scm
+++ b/queries/just/highlights.scm
@@ -7,19 +7,15 @@
 ["@" "==" "!=" "+" ":="] @operator
 [ "(" ")" "[" "]" "{{" "}}" "{" "}"] @punctuation.bracket
 
-[
-  (assignment (NAME))
-  (alias (NAME))
-  (value (NAME))
-  (parameter (NAME))
-] @variable
+(assignment (NAME) @variable)
+(alias (NAME) @variable)
+(value (NAME) @variable)
+(parameter (NAME) @variable)
 
 ; Recipe definitions
-[
-  (recipeheader (NAME))
-  (dependency (NAME))
-  (dependency (depcall (NAME)))
-] @function
+(recipeheader (NAME) @function)
+(dependency (NAME) @function)
+(dependency (depcall (NAME) @function))
 (parameter) @parameter
 (depcall (expression (value (NAME) @parameter)))
 
--- a/queries/just/highlights.scm
+++ b/queries/just/highlights.scm
@@ -1,7 +1,7 @@
 
 (comment) @comment
 (string) @string
-(boolean ["true" "false"]) @boolean
+[ "true" "false" ] @boolean
 ["if" "else"] @conditional
 ["export" "alias" "set"] @keyword
 ["@" "==" "!=" "+" ":="] @operator
--- a/queries/just/highlights.scm
+++ b/queries/just/highlights.scm
@@ -1,0 +1,39 @@
+
+(assignment (NAME) @variable)
+(alias (NAME) @variable)
+(value (NAME) @variable)
+(parameter (NAME) @variable)
+(setting (NAME) @keyword)
+(setting "shell" @keyword)
+
+(call (NAME) @function)
+(dependency (NAME) @function)
+(depcall (NAME) @function)
+(recipeheader (NAME) @function)
+
+(depcall (expression) @parameter)
+(parameter) @parameter
+(variadic_parameters) @parameter
+
+["if" "else"] @conditional
+
+(string) @string
+
+(boolean ["true" "false"]) @boolean
+
+(comment) @comment
+
+(recipe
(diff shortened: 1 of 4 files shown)


queries/just/highlights.scm
```diff
@@ -0,0 +1,39 @@
...
+(boolean ["true" "false"]) @boolean
+
+(comment) @comment
+
+(recipe
+  (body) @text.literal.block (#set! "priority" 90))
+
+(shebang_recipe
+  (shebang interpreter: (TEXT) @keyword) @comment
+  (shebang_body) @text.literal.block (#set! "priority" 90))
```
lucario387, line 31: Is this a deliberate priority reducing for bash/python3? or what

lucario387: Because if it's only to be injected, it's better to just not have any highlights applied to the node at all

queries/just/highlights.scm
```diff
@@ -0,0 +1,39 @@
+
+(assignment (NAME) @variable)
+(alias (NAME) @variable)
+(value (NAME) @variable)
+(parameter (NAME) @variable)
+(setting (NAME) @keyword)
+(setting "shell" @keyword)
```
lucario387, line 7: Why can't this `shell` be at the bottom with other keywords

queries/just/highlights.scm
```diff
@@ -0,0 +1,39 @@
+
+(assignment (NAME) @variable)
+(alias (NAME) @variable)
+(value (NAME) @variable)
+(parameter (NAME) @variable)
```
lucario387, line 5:
```suggestion
(NAME) @variable
```

queries/just/highlights.scm
```diff
@@ -0,0 +1,39 @@
...
+  (shebang interpreter: (TEXT) @keyword) @comment
+  (shebang_body) @text.literal.block (#set! "priority" 90))
+
+["export" "alias" "set"] @keyword
+
+["@" "==" "!=" "+" ":="] @operator
+
+[ "(" ")" "[" "]" "{{" "}}" "{" "}"] @punctuation.bracket
+
+(ERROR) @error
```
lucario387, line 39: No error capture, we recently changed our mind on how to handle it, #5572

queries/just/highlights.scm
Hubro: @lucario387 It's intended to be overridden by injected syntax highlighting, yes. I followed the instructions here:

https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#text

It works as a nice fallback if the user picks a language that doesn't have an installed parser.

clason:
> I have listed myself as maintainer, as I'm not sure if it refers to the maintainer of the parser or the maintainer of the associated configuration in this repository, and I guessed the latter.

It's the latter (as in, the person that gets pinged and is expected to fix things if an update is breaking or someone makes a PR to the queries).

queries/just/highlights.scm
lucario387: I don't see why it's a necessity to highlight raw text.

queries/just/highlights.scm
Hubro: These are not my queries, but I took a loot at this today. Seems like the `set shell := ...` setting is parsed differently from other settings, because part of the value of this setting is used to set the default injection language for regular Just recipes.

Essentially, "shell" is not a keyword and should actually be captured by `(NAME)` on the line above, but isn't because it's parsed weirdly.

queries/just/highlights.scm
Hubro: I took a look at the parser today, and it seems `(NAME)` is used all over the place. It's basically a generic identifier match. Capturing all of them as `@variable` is too broad.

Hubro:
I rewrote the highlight query, would appreciate a new review :slightly_smiling_face:

queries/just/highlights.scm
```diff
@@ -0,0 +1,102 @@
...
+    "dotenv-load"
+    "dotenv-path"
+    "export"
+    "fallback"
+    "ignore-comments"
+    "positional-arguments"
+    "shell"
+    "tempdir"
+    "windows-powershell"
+    "windows-shell"))
```
Hubro, line 48: Is this usage of `@error` acceptable?

amaanq:
if anything the scanner *should* be in C as it's just easier to deal with and more portable, and just looking at the scanner code it absolutely does not need c++ for its containers

clason:
Indeed, C++ scanners aren't welcome anymore. We have a small handful of holdouts, but these aren't top tier languages and may be dropped (or marked "unsupported") in the near future.

lucario387:
If there's a big enough community writing scanners in Rust, then yes, otherwise no

theHamsta:
> I don't suppose nvim-treesitter supports or plans to support scanners written in Rust?

Would be cool! But we would probably need to improve our story (or better the story about upstream tree-sitter) about letting parser developers release binaries or auto-build verified binaries from the parser ecosystem. The nice thing about C is that it comes preinstalled with Linux (and probably also on Mac?). An advance of Rust vs C++ is that you don't need a matching shared standard library C++ installed which can cause problems on Windows.

clason:
Most definitely not.

queries/just/highlights.scm
lucario387: Well this seems to be very explicit, I'm okay with it.

clason: `@error` is specifically a parsing error, so it shouldn't in general be used at all. The goal of this(!) project is syntax highlighting, not linting -- this means we want to correctly highlight all _correct_ syntax, but highlighting all _incorrect_ syntax is an explicit non-goal-

Hubro: > `@error` is specifically a parsing error, so it shouldn't in general be used at all. The goal of this(!) project is syntax highlighting, not linting -- this means we want to correctly highlight all _correct_ syntax, but highlighting all _incorrect_ syntax is an explicit non-goal.

That's fair, but what's the best way to make users aware of typos like this (short of LSP)?

Should the treesitter parser fail to parse function- and setting names that don't exist, at which point they would be automatically highlighted as errors in Neovim?

lucario387: ask your parser maintainer to add that

clason: You're always welcome to put these captures in your own personal queries. They just don't fit in _our_ curated, consistent, queries.

queries/just/highlights.scm
```diff
@@ -0,0 +1,102 @@
+
+(comment) @comment
+(string) @string
+(boolean ["true" "false"]) @boolean
```
lucario387, line 4: Why can't it just be 
```suggestion
[ "true" "false" ] @boolean
```

queries/just/highlights.scm
```diff
@@ -0,0 +1,102 @@
...
+["export" "alias" "set"] @keyword
+["@" "==" "!=" "+" ":="] @operator
+[ "(" ")" "[" "]" "{{" "}}" "{" "}"] @punctuation.bracket
+
+[
+  (assignment (NAME))
+  (alias (NAME))
+  (value (NAME))
+  (parameter (NAME))
+] @variable
```
lucario387, line 15: Are you sure you want to capture the entire assignment with variable, or just it's name?

queries/just/highlights.scm
```diff
@@ -0,0 +1,102 @@
...
+
+; Recipe definitions
+[
+  (recipeheader (NAME))
+  (dependency (NAME))
+  (dependency (depcall (NAME)))
+] @function
```
lucario387, line 22: Similar to this

queries/just/highlights.scm
Hubro: Oh yeah, that was a brain fart. I guess I have to capture each of those individually, huh?

queries/just/injections.scm
```diff
@@ -0,0 +1,30 @@
...
+
+; For shebang recipes, use the shebang executable name as the language
+(shebang_recipe
+  (shebang interpreter: (TEXT) @_lang)
+  (shebang_body) @injection.content
+  (#eq? @_lang "python3")
+  (#set! injection.language "python")
+  (#set! injection.include-children))
+(shebang_recipe
+  (shebang interpreter: (TEXT) @injection.language)
```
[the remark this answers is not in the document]
Hubro: Is that needed? :thinking: I thought the first and/or most specific query "wins", or is that just with highlights?

queries/just/highlights.scm
Hubro: > The goal of this(!) project is syntax highlighting, not linting -- this means we want to correctly highlight all correct syntax, but highlighting all incorrect syntax is an explicit non-goal.

Sorry for harping on this but I still don't fully understand.

Following the goals of this project, what is the correct course of action when you're writing highlight queries for a parser that is "too nice", i.e. successfully parses files even if they are technically invalid?

I feel like highlighting such files naively is actively unhelpful, as it helps mask obvious errors, such as typos in keywords, indentation errors in Python code, etc.

Simultaneously, it's *much* easier and also more performant to write parsers that are a bit too nice, since properly handling errors in tree-sitter without temporarily borking the entire document can be very difficult.

Obvious example, Python files are parsed just fine with invalid indentation (a syntax error):

image

It seems like pointing out certain syntax errors is just way easier with syntax highlighting, which this project has moved away from. (I still don't understand why, but I'm sure it's documented in the issues somewhere, I'll try to read up on it.)

Does nvim-treesitter need a new linter module? :smile:

queries/just/injections.scm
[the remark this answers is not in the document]
lucario387: For neovim, it never was the first query "wins". It was always `all queries are applied`. Priority -> specificity -> order of apperance. For injections this means all injection queries are applied.

queries/just/highlights.scm
Hubro: From reading this thread: https://github.com/nvim-treesitter/nvim-treesitter/pull/5572#issuecomment-1779113178

It seems like the consensus is that capturing `(ERROR)` nodes with `@error` is a bad thing, for several reasons that make a lot of sense. It sounds like applying `@error` to very specific captures that are *definitely* errors is still purely positive though, even if it's not the goal of the project :thinking:

What's the argument for strictly not allowing the `@error` capture anywhere?

clason: > Following the goals of this project, what is the correct course of action when you're writing highlight queries for a parser that is "too nice", i.e. successfully parses files even if they are technically invalid?

Do something outside the scope of this project.

> What's the argument for strictly not allowing the @error capture anywhere?

As I stated: Stick to one thing, and do that (as) well (as possible). For us, that thing is syntax highlighting, _not_ linting. 

> Does nvim-treesitter need a new linter module? 😄

Feel free to write a tree-sitter based linter. (An "nvim-treesitter" module is not welcome; we are actively working on getting rid of them.)

lucario387: The argument here is moreso: Let the parser handle the heavyweight of finding the error. We don't need/want to do it here. Users can add the reverse of your queries (`((NAME) @error (#not-eq? @error "..."))`) if they want error highlighting

Hubro: > > Following the goals of this project, what is the correct course of action when you're writing highlight queries for a parser that is "too nice", i.e. successfully parses files even if they are technically invalid?
> 
> Do something outside the scope of this project.

What I mean is, what is the correct course of action when writing the highlight queries?

- Should I highlight code that is *obviously* wrong as if it was correct?
  - I feel like this is actively harmful, people have an expectation that if it highlights as a keyword, it's a keyword
- Should I highlight code that is correct, while "leaving out" code that is wrong?
  - This feels like it's almost exactly the same as just capturing it with `@error`, but less helpful. At least if it was captured, users could choose to highlight it.

Are there any modules or plugins you know of that replaces the error highlighting that used to be in nvim-treesitter highlights? Personally I found it generally helpful to get live feedback while typing, without waiting for- and reading an error message from the LSP, assuming what I'm writing even has an LSP.

clason: > Should I highlight code that is correct, while "leaving out" code that is wrong?

Yes. (The big problem with highlighting syntax errors is that with treesitter, you get live feedback as you type. And (almost) any incomplete statement will be a syntax error, so you get constant blinking Christmas lights while typing. This is _horrible_ UX and not acceptable to us. And since consistency across queries is a primary project goal, this means _no_ query will highlight errors.)

> Are there any modules or plugins you know of that replaces the error highlighting that used to be in nvim-treesitter highlights? 

No. But you can do so very easily in your own config by adding queries that start with `; extends`.

Hubro: Fair enough, I can see how that would be really annoying with some color schemes.

> No. But you can do so very easily in your own config by adding queries that start with ; extends.

Right :thinking: So then it would be technically possible to write a plugin that extends all the highlights queries in nvim-treesitter to add the error captures back?

clason: Of course.

mkatychev:
@Hubro any status on this? There's been upstream changes made that for example removed the cpp generated code that need to be resolved.

clason:
Closing as stale. If people are interested in this, they are encouraged to make their own PR based on the current state of the parser and our contributing guidelines.

PR #5698, opened 2023-11-20
feat(regex): any_character, start end assertion hl differentiation

ribru17:
This PR highlights the `class_range` operator (`-`) as `@operator`. It also moves `start_assertion` and `end_assertion` (and adds `any_character` (`.`)) to the `@variable.builtin` highlight. Hopefully this change isn't too controversial but I understand if it gets shot down. I figured this made more sense than the escape highlight group since (unlike the other nodes in the highlight group) these nodes do not begin with an actual escape character (`\`). If the start and end assertion changes get shot down I still think it would be beneficial to highlight `.` as either `@string.escape` or `@variable.builtin`.
Before:
beforeregex

After:
afterregex

### DIFF
--- a/queries/regex/highlights.scm
+++ b/queries/regex/highlights.scm
@@ -25,10 +25,14 @@
  (control_letter_escape)
  (character_class_escape)
  (control_escape)
- (start_assertion)
- (end_assertion)
  (boundary_assertion)
  (non_boundary_assertion)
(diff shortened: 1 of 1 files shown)


amaanq:
personally I'm fine with this and do like the added distinctions

amaanq (approved):

PR #5700, opened 2023-11-21
feat: add properties parser

ObserverOfTime:


amaanq (approved):

clason:
I think we should strive towards more descriptive language names...

PR #5707, opened 2023-11-22
feat(bash) Highlight concatenated words as strings

milesfrain:
In this shell snippet, it would be helpful for the `.y` parts to be a different color than the `foo` variable to clarify that the variable name is not `foo.y`. Only the second line `$foo.y` was originally misleading, but I edited the third line too for consistency. I think it's most correct to highlight as `@string`. The 5th line is a bit of an outlier because it's still highlighted as `@parameter`, which seems fine.

Before:

After:

With `nvim --clean foo.sh` (no treesitter, for reference):

### DIFF
--- a/queries/bash/highlights.scm
+++ b/queries/bash/highlights.scm
@@ -66,6 +66,13 @@
 (command
   argument: "$" @string) ; bare dollar
 
+(concatenation
+  [
+    (simple_expansion)
+    (expansion)
+  ]
+  (word) @string)
+
 [
   "if"
   "then"


milesfrain:
I found this bash `queries/highlights.scm` file in three different repos. I'm not clear on the relationship between these. Are changes distributed as follows:
[tree-sitter-bash](https://github.com/tree-sitter/tree-sitter-bash/blob/7331995b19b8f8aba2d5e26deb51d2195c18bc94/queries/highlights.scm)  -->  [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter/blob/cd583cfb5b492c4f81edc5e1916c9f59b7b0ef16/queries/bash/highlights.scm)  -->  [neovim](https://github.com/neovim/neovim/blob/879617c9bbbacb0d0f778ff6dd53cc7c95794abe/runtime/queries/bash/highlights.scm)

Should I be making this PR to `tree-sitter-bash` instead?

milesfrain:
This is another problematic case:
```bash
bar=$one/$two/string
```


Expected highlighting:


It seems like the syntax tree is not quite right:
```lisp
(command) ; [1:1 - 20]
 (variable_assignment) ; [1:1 - 10]
  name: (variable_name) ; [1:1 - 3]
  value: (concatenation) ; [1:5 - 10]
   (simple_expansion) ; [1:5 - 8]
    (variable_name) ; [1:6 - 8]
   (word) ; [1:9 - 9]
 name: (command_name) ; [1:11 - 20]
  (word) ; [1:11 - 20]
```

ObserverOfTime:
@amaanq

amaanq (approved):

PR #5711, opened 2023-11-24
docs: update parsers that come with highlights module

ribru17:


amaanq:
the lacking x means no maintainer

ribru17:
Ah I see, thank you. Should [this line](https://github.com/nvim-treesitter/nvim-treesitter/blob/71bdf97bf6dafc776ad957169533f2f669a8c562/README.md?plain=1#L167C102-L167C197) be changed then?

clason:
Yes, and already is on the `main` branch. (This list is autogenerated, so manual changes are not acceptable.)

PR #5721, opened 2023-11-26
twig queries: add combined injections by default

sheepwall:
As it is in https://github.com/gbprod/tree-sitter-twig/blob/main/queries/injections.scm

What changes for the user?

image

instead of

image

I didn't realise this was fixed after reading some issues in gbprod/tree-sitter-twig. Until I compared the injections.scm in both repositories and seeing a similar update was made for PHP which had the same issue.

### DIFF
--- a/queries/twig/injections.scm
+++ b/queries/twig/injections.scm
@@ -1,3 +1,4 @@
 ((content) @injection.content
- (#set! injection.language "html"))
+ (#set! injection.language "html")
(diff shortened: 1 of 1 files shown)


lucario387:
The version that was in nvim-treesitter didn't have `injection.combined` or `@combined`, so I didn't add it in.
LGTM

lucario387 (approved):

Issue #3488, opened 2022-09-12
Incremental selection doesn't work with exclusive selections


megabug:
### Describe the bug

Setting the nvim option `selection` to `exclusive` makes visual mode selections not include the character under the cursor.

In this case, nvim-treesitter's incremental selections do not contain the final character of the node.

For example, here the cursor is on the `e`.
<img width="647" alt="Screen Shot 2022-09-12 at 17 24 24" src="https://user-images.githubusercontent.com/15011974/189579503-451fc05f-1983-4bef-be82-9bddca842c49.png">

Starting incremental selection shows that the final `r` in `query_nameserver` is not selected:
<img width="647" alt="Screen Shot 2022-09-12 at 17 24 38" src="https://user-images.githubusercontent.com/15011974/189579527-2145a88f-4319-4930-b923-fa673ca6d9f7.png">

After pressing `c` to change the selection:
<img width="508" alt="Screen Shot 2022-09-12 at 17 24 51" src="https://user-images.githubusercontent.com/15011974/189579553-7edd20e2-06dd-4f70-beec-754b2a111a77.png">

### To Reproduce

Start an incremental selection with the nvim option `selection` set to `exclusive`.

### Expected behavior

The last character of the incremental selection node is selected.

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: `tree-sitter` found 0.20.7 (parser generator, only needed for :TSInstallFromGrammar)
  - OK: `node` found v18.6.0 (only needed for :TSInstallFromGrammar)
  - OK: `git` executable found.
  - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
    Version: Apple clang version 13.1.6 (clang-1316.0.21.2.5)
  - OK: Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

## Parser/Features H L F I J
  - ruby           ✓ ✓ ✓ ✓ ✓
  - markdown       ✓ . ✓ . ✓
  - query          ✓ ✓ ✓ ✓ ✓
  - rust           ✓ ✓ ✓ ✓ ✓
  - css            ✓ . ✓ ✓ ✓
  - scss           ✓ . . ✓ .
  - scheme         ✓ . ✓ . ✓
  - rst            ✓ ✓ . . ✓
[...]

PR #5724, opened 2023-11-27
Fix incremental selection omitting last character if vim selection=exclusive

VanaIgr:
As described in #3488, incremental selection doesn't include the last character if vim's `selection` option is set to `exclusive`. This happens because [`ts_utils.update_selection`](https://github.com/nvim-treesitter/nvim-treesitter/blob/b056e4227b1c5d3ecfe96941352364c0c10668df/lua/nvim-treesitter/ts_utils.lua#L278C55-L278C55) converts treesitter node range to an inclusive range, and then sets the start and end markers to the boundaries of that range.
Updating this function to consider `selection` could break other scripts as they might expect both selection ends to be inclusive (as it is the case with incremental selection), so whether the end is exclusive is now controlled by an additional parameter. 
Incremental selection script is changed to take `selection` into account and call `update_selection` with the correct new parameter.

The `end_col = end_col - 1` part is not guaranteed to work with multibyte characters for arbitrary position, as `end_col` can be on the second+ byte of the character and subtracting 1 will leave position the same, but from my tests '< and '> marks of the last visual selection always stay on the first byte of the character, even if the cursor in visual mode was set to the second byte of that character. I couldn't find any vim function that would give starting position of a multibyte character from a byte offset. Besides, two other plugins I know work with `selection=exclusive` both do it the same way ([nvim-surround](https://github.com/kylechui/nvim-surround/blob/0855a89e00a5822c3a482a82e5223fcf2e9ede13/lua/nvim-surround/init.lua#L130) and [targets](https://github.com/wellle/targets.vim/blob/642d3a4ce306264b05ea3219920b13ea80931767/autoload/targets/target.vim#L47)).

### DIFF
--- a/lua/nvim-treesitter/incremental_selection.lua
+++ b/lua/nvim-treesitter/incremental_selection.lua
@@ -15,7 +15,7 @@ function M.init_selection()
   local buf = api.nvim_get_current_buf()
   local node = ts_utils.get_node_at_cursor()
   selections[buf] = { [1] = node }
-  ts_utils.update_selection(buf, node)
+  ts_utils.update_selection(buf, node, nil, ts_utils.is_selecton_end_exclusive())
 end
 
 -- Get the range of the current visual selection.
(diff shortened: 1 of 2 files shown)


clason:
Sorry for not responding sooner. I'm afraid this module is deprecated, as the whole module system (and ts_utils) will be removed in 1.0. 

If there's sufficient interest in this module as it is (instead of nvim-treesitter-textobjects, which has a much more powerful approach), someone would have to refactor it into a standalone similar to indent.lua.

PR #5729, opened 2023-11-27
feat(make): give targets the function highlight

ribru17:
Highlights target words as `@function` to match up with builtin target words being `@function.builtin`
Before:
beforemakefile
After:
aftermakefile

### DIFF
--- a/queries/make/highlights.scm
+++ b/queries/make/highlights.scm
@@ -10,6 +10,8 @@
  ] @conditional)
  "endif" @conditional)
 
+(rule (targets (word) @function))
+
 (rule (targets
        (word) @function.builtin
        (#any-of? @function.builtin


ObserverOfTime (commented):
Could the references also be highlighted (i.e. `all`, `clean` in `.PHONY`)?

ribru17:
Sure, do you think it would be satisfactory just to highlight any prerequisite of the `.PHONY` rule as a function?

ribru17:
Sounds good. Would you mind if I still added function highlights to `.PHONY` prerequisites? I could be wrong but I've looked around and couldn't find any instances of prerequisites being other rules except in the special case of `.PHONY`.

ObserverOfTime:
I don't mind.

PR #5734, opened 2023-11-28
feat(angular): switch to parser supporting v17

dlvandenberg:
Updated the parser for Angular which has support for v17.
Added injections and highlights for both HTML and TypeScript files (for inline templates)

### DIFF
--- a/queries/angular/highlights.scm
+++ b/queries/angular/highlights.scm
@@ -21,10 +21,14 @@
   function: ((identifier) @function.builtin
     (#eq? @function.builtin "$any")))
 
+(annotation) @punctuation.special
+
 [
-"let"
-"as"
+  (control_flow_keyword)
+  "let"
 ] @keyword
+
+(special_block_keyword) @type
 
 [
 "("
@@ -43,9 +47,9 @@
 ] @punctuation.delimiter
 
 ((identifier) @boolean
-  (#any-of? @boolean "true" "false"))
+  (#vim-match? @boolean "^(true|false)$"))
 ((identifier) @variable.builtin
-  (#any-of? @variable.builtin "this" "\$event" "null"))
+  (#vim-match? @variable.builtin "^(this|\$event|null)$"))
 
 [
   "-"
@@ -65,3 +69,5 @@
   "||"
   "%"
 ] @operator
+
+(ternary_operator) @conditional.ternary
--- a/queries/angular/highlights.scm
+++ b/queries/angular/highlights.scm
@@ -21,7 +21,7 @@
   function: ((identifier) @function.builtin
     (#eq? @function.builtin "$any")))
 
-(annotation) @punctuation.special
+(annotation) @type
 
 [
   (control_flow_keyword)
@@ -31,25 +31,26 @@
 (special_block_keyword) @type
 
 [
-"("
-")"
-"["
-"]"
-"{"
-"}"
+  "("
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
 ] @punctuation.bracket
 
 [
-";"
-"."
-","
-"?."
+  ";"
+  "."
+  ","
+  "?."
 ] @punctuation.delimiter
 
 ((identifier) @boolean
   (#vim-match? @boolean "^(true|false)$"))
 ((identifier) @variable.builtin
   (#vim-match? @variable.builtin "^(this|\$event|null)$"))
+
 
 [
   "-"
@@ -70,4 +71,7 @@
   "%"
 ] @operator
 
-(ternary_operator) @conditional.ternary
+[
+ (ternary_operator)
+ (conditional_operator)
+] @conditional.ternary
--- a/queries/angular/highlights.scm
+++ b/queries/angular/highlights.scm
@@ -47,9 +47,9 @@
 ] @punctuation.delimiter
 
 ((identifier) @boolean
-  (#vim-match? @boolean "^(true|false)$"))
+  (#any-of? @boolean "true" "false"))
 ((identifier) @variable.builtin
-  (#vim-match? @variable.builtin "^(this|\$event|null)$"))
+  (#any-of? @variable.builtin "this" "\$event" "null"))
 
 
 [
--- a/queries/html/injections.scm
+++ b/queries/html/injections.scm
@@ -25,3 +25,112 @@
   (text) @injection.content
   (#eq? @_py_config "py-config")
   (#set! injection.language "toml"))
+
+;------ Angular queries ------
+
+; {{ someBinding }}
+((text) @injection.content 
+  (#lua-match? @injection.content "%{%{.+%}%}")
+  (#set! injection.language "angular")
+)
+
+; }
+((text) @injection.content 
+  (#lua-match? @injection.content "%}.*")
+  (#set! injection.language "angular")
+)
+
+; Angular control flow statements
+; @if (condition) {
+((text) @injection.content 
+  (#lua-match? @injection.content "%@if%s+%(.+%)%s+%{$")
+  (#set! injection.language "angular")
+)
+
+; @for (item of items) {
+((text) @injection.content 
+  (#lua-match? @injection.content "%@for%s+%(.+%)%s+%{$")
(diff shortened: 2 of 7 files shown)


queries/angular/highlights.scm
```diff
@@ -21,31 +21,36 @@
...
-"?."
+  ";"
+  "."
+  ","
+  "?."
 ] @punctuation.delimiter
 
 ((identifier) @boolean
-  (#any-of? @boolean "true" "false"))
+  (#vim-match? @boolean "^(true|false)$"))
```
lucario387, line 50: Keep the any-of here

queries/angular/highlights.scm
```diff
@@ -21,31 +21,36 @@
...
 ((identifier) @variable.builtin
-  (#any-of? @variable.builtin "this" "\$event" "null"))
+  (#vim-match? @variable.builtin "^(this|\$event|null)$"))
```
lucario387, line 52: So is this

queries/html/injections.scm
```diff
@@ -25,3 +25,112 @@
   (text) @injection.content
   (#eq? @_py_config "py-config")
   (#set! injection.language "toml"))
+
+;------ Angular queries ------
+
+; {{ someBinding }}
+((text) @injection.content
```
lucario387, line 32: With this many text node injections. i kindof feel like there'll be performance issues

clason: Yeah, that's way too many. We either need to add support for conditional injections, or leave them for user config.

(Alternatively -- and I'm tending more and more towards that -- we should punt all "framework" handling to an `nvim-treesitter-webdev` plugin that can be smarter about what to inject and where.)

ObserverOfTime: > (Alternatively -- and I'm tending more and more towards that -- we should punt all "framework" handling to an `nvim-treesitter-webdev` plugin that can be smarter about what to inject and where.)

`nvim-treesitter-angular` is dead.
Long live `nvim-treesitter-webdev`.

clason: (Joke aside, that was my first thought when I looked at nvim-treesitter-angular, which would have been a reasonable thing. Only when I learned that it doesn't do anything that nvim-treesitter wouldn't do did I press for upstreaming and archiving.)

ObserverOfTime: I do have an idea that I'll post on matrix for discussion later ~~provided I don't forget~~.

lucario387:
I think @dlvandenberg is writing an angular parser for specifically `*.component.html` files that extends html parser, so I think with that new parser injecting html will not be necessary

Theaninova:
> I think @dlvandenberg is writing an angular parser for specifically `*.component.html` files that extends html parser, so I think with that new parser injecting html will not be necessary

Hmm, I think I understand most of the issues here, though the `.component.html` extension isn't mandatory, just a suggestion. One of the projects I work on inadvertently ended up having just the `.html` extension. Given what I think I understand it seems like the best thing to just enforce the extension from the angular style guide, though I think it's something that should be mentioned very explicitly somewhere, otherwise it can be super unclear why the syntax highlighting isn't showing up...

ObserverOfTime:
> One of the projects I work on inadvertently ended up having just the `.html` extension.

You can manually set the filetype to `html.angular` in that case.

ObserverOfTime:
Removing the html injections and applying this patch mostly works:

```diff
diff --git i/lua/nvim-treesitter/parsers.lua w/lua/nvim-treesitter/parsers.lua
index ddab5498..16a9bc59 100644
--- i/lua/nvim-treesitter/parsers.lua
+++ w/lua/nvim-treesitter/parsers.lua
@@ -97,8 +97,8 @@ list.angular = {
   install_info = {
     url = "https://github.com/dlvandenberg/tree-sitter-angular",
     files = { "src/parser.c" },
-    requires_generate_from_grammar = true,
   },
+  filetype = "html.angular",
   maintainers = { "@dlvandenberg" },
   experimental = true,
 }
diff --git i/queries/angular/injections.scm w/queries/angular/injections.scm
new file mode 100644
index 00000000..5081859c
--- /dev/null
+++ w/queries/angular/injections.scm
@@ -0,0 +1,4 @@
+((program) @injection.content
+ (#set! injection.combined)
+ (#set! injection.include-children)
+ (#set! injection.language "html_tags"))
diff --git i/queries/html_tags/highlights.scm w/queries/html_tags/highlights.scm
index 92d5897b..10ad8f2b 100644
--- i/queries/html_tags/highlights.scm
+++ w/queries/html_tags/highlights.scm
@@ -5,7 +5,8 @@
 ((attribute
   (quoted_attribute_value) @string)
  (#set! "priority" 99))
-(text) @text @spell
+((text) @text @spell
+ (#set! "priority" 99))
 
 ((element (start_tag (tag_name) @_tag) (text) @text.title)
  (#eq? @_tag "title"))
```
But modeline (`<!-- vim:set ft=html.angular: -->`) highlights flicker between html and angular.

lucario387:
Andd we are back to square one. vim/vim#13604

queries/html/injections.scm
```diff
@@ -25,3 +25,112 @@
...
+
+; } @loading (conditions) {
+((text) @injection.content
+  (#lua-match? @injection.content "%}%s+%@loading%s+%(.+%)%s+%{$")
+  (#set! injection.language "angular")
+)
+
+; } @error {
+((text) @injection.content
+  (#lua-match? @injection.content "%}%s+%@error%s+%{$")
```
lucario387, line 131: @dlvandenberg Will a pattern of `(@error|@loading|@placeholder|@empty|@default|@defer|@if|@else|@switch)` be decent enough to detect html angular files? (given you can only check for the first 10 lines)

lucario387: tldr: the way to detect angular is to let vscode do it :sob:

ObserverOfTime: The parser should either extend html (in which case it probably won't need any injections), or do it the [htmldjango](https://github.com/interdependence/tree-sitter-htmldjango/blob/8873e3df89f9ea1d33f6235e516b600009288557/grammar.js#L149-L150) way and let html be injected in the remaining content.

clason: Either way is fine; we just need a decision and documentation (even if just in in the PR description here) to point people to if they ask.

clason: Yeah, and a PR would surely be welcome to add to that for angular (if necessary, increasing the number of lines). It doesn't have to be foolproof and perfect; you can always override this via a plugin or (project-specific) config.

ObserverOfTime: > Filetype would be `htmlangular` then?

If someone wants to contribute the syntax (or at least extend html).

lucario387: (not me)

Also, for highlighting angular files correctly, let's wait for https://github.com/dlvandenberg/tree-sitter-angular-2 and then see what would be the best

dlvandenberg: > > @dlvandenberg Will a pattern of `(@error|@loading|@placeholder|@empty|@default|@defer|@if|@else|@switch)` be decent enough to detect html angular files? (given you can only check for the first 10 lines)
> 
> There are other patterns like the `[attr]=""` or `*structuralDirective` or `@animation` or `{{template}}` that all indicate an Angular file and may appear anywhere in the file. Even worse the a bunch of these are not unique to Angular.
> 
> But even with that, maybe, until the first file with a license header comes around...
> 
> I think the only reliable way to know if the file is an Angular template is if a parent directory contains an `angular.json` file (though even that is not technically correct... It's an Angular Template file if it is referenced by an Angular component)
> 
> The Angular LSP does it that way...

In case of a Nx monorepo, you don’t have any `angular.json` files. I’m not sure how to handle file detection yet. 

But the parser that I’m rewriting is extending the HTML one and assumes that the file is detected as an Angular template file.

dlvandenberg:
I'm closing this PR as most of the discussion here is either about the many injection queries or detecting angular files. I will create a new PR that will add the Angular parser that extends the HTML one.

PR #5735, opened 2023-11-28
feat: add tree-sitter-styled

mskelton:
I created a parser for styled-components or other CSS-in-JS languages to highlight CSS within `styled.div` type expressions.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1779,6 +1779,15 @@ list.strace = {
   maintainers = { "@amaanq" },
 }
 
+list.styled = {
+  install_info = {
+    url = "https://github.com/mskelton/tree-sitter-styled",
+    files = { "src/parser.c", "src/scanner.c" },
+    generate_requires_npm = true,
+  },
+  maintainers = { "@mskelton" },
+}
+
 list.supercollider = {
   install_info = {
     url = "https://github.com/madskjeldgaard/tree-sitter-supercollider",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1783,7 +1783,6 @@ list.styled = {
   install_info = {
     url = "https://github.com/mskelton/tree-sitter-styled",
     files = { "src/parser.c", "src/scanner.c" },
-    generate_requires_npm = true,
   },
   maintainers = { "@mskelton" },
 }
--- a/queries/styled/highlights.scm
+++ b/queries/styled/highlights.scm
@@ -1,0 +1,4 @@
+; inherits: css
+
+(plain_value) @identifier
+(js_expression) @identifier
--- a/queries/styled/highlights.scm
+++ b/queries/styled/highlights.scm
@@ -1,4 +1,4 @@
 ; inherits: css
 
-(plain_value) @identifier
-(js_expression) @identifier
+(plain_value) @variable
+(js_expression) @variable
--- a/queries/styled/highlights.scm
+++ b/queries/styled/highlights.scm
@@ -1,4 +1,1 @@
 ; inherits: css
-
-(plain_value) @variable
-(js_expression) @variable
--- a/lockfile.json
+++ b/lockfile.json
@@ -606,7 +606,7 @@
     "revision": "d819cdd5dbe455bd3c859193633c8d91c0df7c36"
   },
   "styled": {
-    "revision" : "05b795ad742c3ad0dd212c5e601d824760a52a8f"
+    "revision": "e51e673efc860373167680b4bcbf418a11e4ed26"
   },
   "supercollider": {
     "revision": "3b35bd0fded4423c8fb30e9585c7bacbcd0e8095"
--- a/lockfile.json
+++ b/lockfile.json
@@ -606,7 +606,7 @@
     "revision": "d819cdd5dbe455bd3c859193633c8d91c0df7c36"
   },
   "styled": {
-    "revision" : "e51e673efc860373167680b4bcbf418a11e4ed26"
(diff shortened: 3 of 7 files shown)


queries/styled/highlights.scm
```diff
@@ -0,0 +1,4 @@
+; inherits: css
+
+(plain_value) @identifier
+(js_expression) @identifier
```
clason, line 4:
```suggestion
(plain_value) @variable
(js_expression) @variable
```

See https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights for allowed capture names.

clason (dismissed):

queries/styled/highlights.scm
mskelton: Thanks, updated

queries/styled/highlights.scm
```diff
@@ -0,0 +1,4 @@
+; inherits: css
+
+(plain_value) @variable
+(js_expression) @variable
```
ObserverOfTime, line 4: Shouldn't this be highlighted as `@none` and injected with javascript?

mskelton: I ended up removing them. At one point it was necessary to preventing CSS keywords such as `color: red` from coloring the `red` as a string. I'm not seeing that anymore if I remove them though, so going to remove for now.

ObserverOfTime: I still think `(js_expression)` might need to be injected with javascript.

lua/nvim-treesitter/parsers.lua
```diff
@@ -1751,6 +1751,15 @@ list.strace = {
   maintainers = { "@amaanq" },
 }
 
+list.styled = {
+  install_info = {
+    url = "https://github.com/mskelton/tree-sitter-styled",
+    files = { "src/parser.c", "src/scanner.c" },
+    generate_requires_npm = true,
```
lucario387, line 1758: Can't the generation be done from parser repo's side?

lucario387:
Rn I don't know when this parser will be injected into other languages?. Is it just 
```js
styled`somecss`
```
in js/ts?

lua/nvim-treesitter/parsers.lua
clason: That's only _if_ you generate the parser from the grammar, which you don't _need_ to do.

mskelton: The docs say this:

```
generate_requires_npm = false, -- if stand-alone parser without npm dependencies
```

Does that mean if the grammar has npm dependencies then `generate_requires_npm` should be `true`?  That's how that reads to me, and this grammar does have dependencies since it extends the CSS grammar.

I'm not an expert on this though, so very possible I'm wrong about this!

mskelton:
> Rn I don't know when this parser will be injected into other languages?. Is it just
> 
> ```js
> styled`somecss`
> ```
> 
> in js/ts?

@lucario387 So, it requires additional injections to make that work.

```scm
; css`<css>`, keyframes`<css>`
(call_expression
  function: (identifier) @_name
    (#match? @_name "^(css|keyframes)")
  arguments: ((template_string) @styled
  (#offset! @styled 0 1 0 -1)))

; styled.div`<css>`
(call_expression
 function: (member_expression
   object: (identifier) @_name
     (#eq? @_name "styled"))
 arguments: ((template_string) @styled
  (#offset! @styled 0 1 0 -1)))

; styled(Component)`<css>`
(call_expression
 function: (call_expression
   function: (identifier) @_name
     (#eq? @_name "styled"))
 arguments: ((template_string) @styled
  (#offset! @styled 0 1 0 -1)))
```

I'm not really sure if these should be part of nvim-treesitter or not, if you want me to add them to the ECMA grammar, I'm happy to, just not sure if that makes sense.

queries/styled/highlights.scm
mskelton: From my testing it's not really required, highlighting in the template string works fine.

image

Even nested expressions work just fine.

image

lucario387:
Well if there isn't a query to inject it, then the parser is zero used everywhere ...

Also, when adding ecma injection queries, please also move to treesitter upstream style's for injection queries. #4675

lua/nvim-treesitter/parsers.lua
clason: Yes, that is correct. This field is used (and required) if and only if you do `:TSInstallFromGrammar` (which some but not all parsers require).

clason:
Please don't rebase on `main`, that's the (not yet ready) development branch. Pull requests should always be made from `master` for now.

lua/nvim-treesitter/parsers.lua
mskelton: Gotcha, updated the PR to remove `generate_requires_npm`

mskelton:
@clason Just realized that 🙂 I got that fixed up in my latest rebase.

mskelton:
@lucario387 I updated the existing ecma injections to use `styled` instead of `css` since the injections did exist for styled components, but since they use plain CSS, they would not properly support all styled components features (e.g. top level properties, template strings, etc.)

lucario387:
It seems like the default branch hasn't gotten any updates for a while. I'll wait until it's updated with the changes

mskelton:
@lucario387 I've been working on a branch locally updating the CSS parser and was having some issues with the external scanner. I got those resolved today and merged the changes so the grammar is now updated with the latest underlying CSS grammar.

ObserverOfTime (approved):

lucario387 (approved):

clason:
@lucario387 if you're fine with the changed ecma injection, I have no objections; feel free to merge.

PR #5745, opened 2023-11-30
feat: add rbs parser support

joker1007:
I would like to add rbs (Ruby type signature) parser support.

I see that there was a pull request in the past to support rbs, but it wasn't merged.
Sorry, this pull request is for a different implementation by myself.

### DIFF
--- a/queries/rbs/highlights.scm
+++ b/queries/rbs/highlights.scm
@@ -82,8 +82,6 @@
 
 (type (integer_literal) @number)
 
-(ERROR) @error
-
 [
  "="
  "->"
--- a/queries/rbs/highlights.scm
+++ b/queries/rbs/highlights.scm
@@ -1,8 +1,3 @@
-(use_directive
- "use" @keyword)
-(use_clause
- "as" @keyword)
-
 (use_clause
   [
     (type_name)
@@ -19,6 +14,8 @@
 ] @constant.builtin
 
 [
+  "use"
+  "as"
   "class"
   "module"
   "interface"
@@ -35,9 +32,11 @@
  "def"
 ] @keyword.function
 
-(include_member "include" @function.method)  
-(extend_member "extend" @function.method)  
-(prepend_member "prepend" @function.method)  
+[
+ "include"
+ "extend"
+ "prepend"
+] @function.method
 
 (visibility) @type.qualifier
 
@@ -52,8 +51,7 @@
      (setter)
     ] @function.method))
 
-(ivar_member (ivar_name) @property)
-(ivar_member (cvar_name) @property)
+[(ivar_name) (cvar_name)] @property
 
 (alias_member (method_name) @function)
 
--- a/queries/rbs/highlights.scm
+++ b/queries/rbs/highlights.scm
@@ -1,0 +1,111 @@
+(use_directive
+ "use" @keyword)
+(use_clause
+ "as" @keyword)
+
+(use_clause
+  [
+    (type_name)
+    (simple_type_name)
+  ] @type)
+
+[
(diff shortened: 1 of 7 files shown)


queries/rbs/highlights.scm
```diff
@@ -0,0 +1,111 @@
...
+(self) @variable.builtin
+
+(type (symbol_literal) @symbol)
+
+(type (string_literal (escape_sequence) @string.escape))
+(type (string_literal) @string)
+
+(type (integer_literal) @number)
+
+(ERROR) @error
```
lucario387, line 85: (ERROR) should now be on users' side, we don't really want the node here. See #5572

queries/rbs/highlights.scm
```diff
@@ -0,0 +1,111 @@
...
+  "alias"
+] @keyword
+
+[
+ "def"
+] @keyword.function
+
+(include_member "include" @function.method)  
+(extend_member "extend" @function.method)  
+(prepend_member "prepend" @function.method)
```
lucario387, line 40: Can't this be

```suggestion
[ "include" "extend" "prepend" ] @function.method
```

queries/rbs/highlights.scm
```diff
@@ -0,0 +1,111 @@
...
+  (method_name
+    [
+     (identifier)
+     (constant)
+     (operator)
+     (setter)
+    ] @function.method))
+
+(ivar_member (ivar_name) @property)
+(ivar_member (cvar_name) @property)
```
lucario387, line 56: Is this also possible?

```suggestion
[ (ivar_name) (cvar_name) ] @property
```

joker1007:
@lucario387 
Thank you very much for your reviews! It's so fast!
I have accepted your suggestions and have fixed them.

queries/rbs/highlights.scm
```diff
@@ -0,0 +1,110 @@
+(use_directive
+ "use" @keyword)
+(use_clause
+ "as" @keyword)
```
lucario387, line 4: This can probably be reduced

joker1007:
I addressed the next review comments. I have fixed them.
Thanks!!

queries/rbs/highlights.scm
```diff
@@ -0,0 +1,107 @@
...
+(namespace (constant) @namespace)
+
+(builtin_type) @type.builtin
+
+(const_name (constant) @constant)
+(global_name) @property
+
+(parameter (var_name) @variable.parameter)
+
+(keyword) @variable.parameter
```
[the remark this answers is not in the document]
joker1007: It's not a keyword.
 `(keyword)` is a part of "keyword arguments" which is a feature in Ruby.
Probably, it's a `@parameter` token.

queries/rbs/highlights.scm
```diff
@@ -67,9 +63,9 @@
 (const_name (constant) @constant)
 (global_name) @property
 
-(parameter (var_name) @variable.parameter)
+(parameter (var_name) @parameter)
 
-(keyword) @variable.parameter
+(keyword) @parameter
```
lucario387, line 68: I think having a comment for this case is kindof needed, just to clear away the misunderstandings

Something like `keyword arguments` as a comment right on top of this capture will be nice

joker1007: You are right. I have added comments.

joker1007:
Thank you for the approval!

By the way, a test (terraform_spec) failed on CI. It does not appear to be related to this pull request.
I'm sorry, I don't know the cause.
Is it sometimes flaky?

PR #5755, opened 2023-12-02
Add tree-sitter-disassembly

ColinKennedy:
A couple months ago I made added tree-sitter-objdump (https://github.com/nvim-treesitter/nvim-treesitter/pull/5548). One oversight of that repository is that you cannot parse just disassembly on its own, nor memory dumps. So I pulled that portion out and made a separate parser for it.Thanks @amaanq for bringing that to my attention.

Since then, I've started adding a disassembly view to nvim-dap-ui which is working really well. You can see a demo of the parser in action, here. 

https://github.com/nvim-treesitter/nvim-treesitter/assets/10103049/52a66458-ae0c-4f68-81d0-f1fbed43252b

And adding this parser should pave the way for viewing raw memory with nvim-dap-ui, too. I also took a different approach for parsing disassembly in tree-sitter-disassembly than I had in tree-sitter-objdump so it should future proof it better.

### About Maintaining tree-sitter-objdump
Since most of the hard stuff about disassembly parsing is solved in tree-sitter-disassembly and both tree-sitter-objdump and tree-sitter-disassembly will benefit from an eventual Assembly parser, I wonder if it makes sense to convert tree-sitter-objdump into a "don't parse anything, use inject to disassembly and have it parse everything instead", rather than maintain two separate parsers. I maintain both repositories in this case but was hoping to get some advice on that point (e.g. is it worth it, is nested injections `objdump -> disassembly -> asm` advisable, etc)

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [devicetree](https://github.com/joelspadin/tree-sitter-devicetree) (maintained by @jedrzejboczar)
 - [x] [dhall](https://github.com/jbellerb/tree-sitter-dhall) (maintained by @amaanq)
 - [x] [diff](https://github.com/the-mikedavis/tree-sitter-diff) (maintained by @gbprod)
+- [x] [disassembly](https://github.com/ColinKennedy/tree-sitter-disassembly) (maintained by @ColinKennedy)
 - [x] [dockerfile](https://github.com/camdencheek/tree-sitter-dockerfile) (maintained by @camdencheek)
 - [x] [dot](https://github.com/rydesun/tree-sitter-dot) (maintained by @rydesun)
 - [x] [doxygen](https://github.com/amaanq/tree-sitter-doxygen) (maintained by @amaanq)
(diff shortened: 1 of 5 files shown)


lucario387:
@amaanq

theHamsta (commented):
From this PR it is a bit unclear what kind of input this parser expects (e.g. assembly notation and the used three column layout). Even though `asm` is injected into the third column, the three column parser has to know which `asm` to inject.

Otherwise looks good to me!

ColinKennedy:
As far as I'm aware (note: mostly novice) disassembly only contains vanilla x86_64 assembly instructions. So no NASM / MASM directives, for example. Though occasionally I've seen disassembly contain `.foo` labels. The unittests for tree-sitter-disassembly are pretty thorough though. Of course, things change if the disassembly is from Motorolla CISC or ARM. Since disassembly doesn't have a syntax to indicate a language, I guess one would have to manage the injections themselves if they wanted to change it. I recall a Neovim plug-in for that but I'm blanking on the name right now.

amaanq (approved):

lucario387:
This breaks all future PRs from master due to queries not formatted

amaanq:
oops, my bad

clason:
Man, that'll make rebasing `main` even more painful... 😞

PR #5762, opened 2023-12-04
fix(luadoc): Assign `@field.lua`, not `@field`, in member_type

wookayin:
TL;DR) This patch makes the `@field.luadoc` highlight group more usable,
in distinction with `@field.lua.luadoc`.

Problem: Previously, there is no way to distinguish "field name" and
"fields" in lua type nodes with a nested namespace (both captured as
`@field.luadoc`).

For example,

```lua
---@class MyClass
---@field opts? lsp.ClientConfig | { buffer: integer? } an option table.
---       ^^^^      ^^^^^^^^^^^^     ^^^^^^
---        (1)           (2)           (3)
```

- First, (1) (field_annotation) should be captured as `@field.luadoc`.
- But (2) should not. `lsp.ClientConfig` (member_type) is a type with
  some namespaces; this is captured `@type` but we are not
  interested in the most significant segment (e.g. `ClientConfig`).
- Also, (3) is (table_literal_type) another example that is correctly
  captured as `@field.luadoc`.

So if one highlights `@field.luadoc`, all of (1), (2), and (3) will be
highlighted. Instead, we want (1) and (3) only.

Solution: Capture the cases like (2) in (member_type) nodes as
`@field.lua.luadoc`. This will allow users to have two separate
highlight groups for (1) and (3) v.s. (2).

### DIFF
--- a/queries/luadoc/highlights.scm
+++ b/queries/luadoc/highlights.scm
@@ -85,7 +85,7 @@
 
 (table_literal_type field: (identifier) @field)
 
-(member_type ["#" "."] . (identifier) @field)
+(member_type ["#" "."] . (identifier) @field.lua)
 
 ; Types
 


wookayin:
Screenshot:

image

(Previously all the three groups were `@field.luadoc`)

clason:
Sorry, but this is not a standard capture. A better fix would be to make the parser distinguish these, if they are actually syntactically distinct.

wookayin:
If so, I'd suggest just removing the `(member_type ["#" "."] . (identifier) @field)` query so that (2) won't be captured. IMHO it's a not highlight group.

clason:
@amaanq

wookayin:
Here is the syntax tree:

```scheme
(field_annotation ; [1:1 - 3:39]                 ; @field
  name: (identifier) ; [1:8 - 11]                ; opts          <-- (1)
  (union_type ; [1:14 - 52]
    (member_type ; [1:14 - 29]                   ; lsp.ClientConfig
      (identifier) ; [1:14 - 16]                 ; lsp
      (identifier)) ; [1:18 - 29]                ; ClientConfig  <-- (2)
    (table_literal_type ; [1:33 - 52]            ; { .. }
      field: (identifier) ; [1:35 - 40]          ; buffer        <-- (3)
      (optional_type ; [1:43 - 50]               ; integer?
        (identifier)))) ; [1:43 - 49]            ; integer
```

clason:
Sorry, I'm not convinced; in Lua, _everything_ is a table, so _everything_ is a field. But the maintainer (@amaanq) is the one to decide.

wookayin:
Hmm, then do you have any idea how these are supposed to be distinguished, logically speaking?

clason:
No, but I'm not fond of syntax highlighted comments at all (and wish that lua-language-server -- which also does this, and better, since it can use semantic information which tree-sitter can't -- wouldn't do that, either).

wookayin:
I see. But some users will find the color in luadoc useful. And this is something that can be done in a "syntactic" level, not in a semantic level.

An alternative approach is to capture `(field_annotation name: (identifier))` with another highlight group (it's probably going to be a non-standard but a suffix-ed group name, say, `@field.name`), leaving `@field.luadoc` for (2) intact.

wookayin:
Closing, `@field` is gone. Will revisit if I'd need this again.

PR #5767, opened 2023-12-06
fix(vim): highlight 'scriptencoding' command

ObserverOfTime:
Should also be upstreamed to neovim & tree-sitter-vim.

### DIFF
--- a/queries/vim/highlights.scm
+++ b/queries/vim/highlights.scm
@@ -83,6 +83,7 @@
   "delcommand"
   "comclear"
   "colorscheme"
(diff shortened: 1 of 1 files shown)


clason (approved):

clason:
I'll sync the query changes in bulk (after the Big Renaming).

lucario387:
Did the Big Renaming just happen 🤔

clason:
No, I'm just saying that I'll sync the queries then; it's fine to merge her now (as it's a relatively minor divergence that most people won't notice).

PR #5774, opened 2023-12-08
feat(gitcommit): add highlight group for subject_prefix

gbprod:
After introducing a new token in gitcommit tree-sitter grammar, I think we should add a highlight group.

Ref: https://github.com/gbprod/tree-sitter-gitcommit/pull/48

### DIFF
--- a/queries/gitcommit/highlights.scm
+++ b/queries/gitcommit/highlights.scm
@@ -9,6 +9,7 @@
 
 (subject) @text.title @spell
 (subject (overflow) @text @spell)
+(subject (subject_prefix) @function @nospell)
 (prefix (type) @keyword @nospell)
 (prefix (scope) @parameter @nospell)
 (prefix [


queries/gitcommit/highlights.scm
```diff
@@ -9,6 +9,7 @@
 
 (subject) @text.title @spell
 (subject (overflow) @text @spell)
+(subject (subject_prefix) @function @spell)
```
lucario387, line 12: Does it really need `@spell` when it's that limited (only `amend!` or `fixup!`)

gbprod: Oh! you're right, it's a mistake, I wanted to add `@nospell` :sweat_smile:

lucario387 (approved):

wookayin:
Probably it's only me, but this has been causing a problem for me
```
$VIMRUNTIME/lua/vim/treesitter/query.lua:248: Query error at 12:11. Invalid node type "subject_prefix":
(subject (subject_prefix) @function @nospell)
          ^
```

even if I **re-installed** the treesitter parser for gitcommit again.

The new node type were added in https://github.com/gbprod/tree-sitter-gitcommit/pull/48, with the commit https://github.com/gbprod/tree-sitter-gitcommit/commit/7e3ad5fdc61cd701e146ef78e4fc6dcdf6dbca0e.

#5772 includes lockfile.json updated to the correct revission 7e3ad5fdc61cd701e146ef78e4fc6dcdf6dbca0e, and `parser-info/gitcommit.revision` reads the same correct SHA, so it should work. But strangely, I keep getting the same error -- cache issues?

Slotos:
Facing exactly the same issue here.

lucario387:
@gbprod 
I can also revert if needed

clason:
Why did CI not catch this?

gbprod:
I don't see any problem with the grammar, I don't know where this issue could come from. This works for me as well as in different CI tests. Maybe the parser is not installed in the latest version ?
Anyway, if many people have this problem, we can revert the commit

Slotos:
So far there are only two reported cases and one (mine) is all about keeping in sync with `main`.

I left my comment as a possible solution for others to follow. Submitted it on the first sentence accidentally and had to edit, which probably landed as a red flag in your mailboxes.

PR #5777, opened 2023-12-10
leo - indents added new language highlights added

r001:
Added indents. Added someminor changes to highlights.scm.

### DIFF
--- a/queries/leo/indents.scm
+++ b/queries/leo/indents.scm
@@ -1,0 +1,33 @@
+[
+ (record_declaration) 
+ (struct_declaration)
+ (mapping_declaration)
+ (constant_declaration)
+ (return_statement)
+ (expression_statement)
+ (variable_declaration)
+ (loop_statement)
+ (assignment_statement)
+ (assert_statement)
+ (struct_expression)
+ (array_expression)
+ (tuple_expression)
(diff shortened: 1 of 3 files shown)


queries/leo/indents.scm
```diff
@@ -0,0 +1,33 @@
...
+ (array_expression)
+ (tuple_expression)
+ (parenthesized_expression)
+ (items_block)
+ (block)
+] @indent.begin
+
+; if "if" statement and conditional statement are in separate lines 
+; conditional should be indented and when the conditional block 
+; starts dedented
```
lucario387, line 22: Does this mean the tree is like

```
branch
  block
   branch
    block
```

lucario387: Ah I understand now.
Can you make the condition of the if statement be a singular node, and if not, can you explain why

Something like

```
if_statement
  condition: (_)
  consequence: (block)
```

r001: No, I do not do it to indent the block, the condition after "if" statement must be indented if written in a new line. And it  was hard first as it has no opening and closing parens as C does, so the block opening brace will end the indent.

lucario387 (approved):

PR #5784, opened 2023-12-12
fix(nim): minor improvements to queries and parser

aMOPel:
- fix(nim): bump parser again to include minor fix
- fix(nim): shifting around for precedence
- fix(nim): added new fields from recent parser version


aMOPel:
@lucario387 feel free to merge now

lucario387 (approved):
Let's get this in before the temporary branch lock

PR #5785, opened 2023-12-12
feat(nix): Slightly improve injections

dasJ:
This adds support for # comments, so one could do something like
```nix
{
  test = # bash
    ''
      rmdir $out
    '';

  test2 =
    # bash
    ''
      rmdir $out
    '';
}
```

This is both a useful feature in general but would also help with formatters that convert the `/**/` comments.

The second change adds injection.combined to scripts to make sure that interpolations in shell scripts don't break quoting.

### DIFF
--- a/queries/nix/injections.scm
+++ b/queries/nix/injections.scm
@@ -1,6 +1,7 @@
 ((comment) @injection.content
  (#set! injection.language "comment"))
 
+; /**/-style comments
 ((comment) @injection.language
   . ; this is to make sure only adjacent comments are accounted for the injections
   [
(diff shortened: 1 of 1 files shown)


queries/nix/injections.scm
```diff
@@ -29,7 +40,8 @@
     (indented_string_expression 
       ((string_fragment) @injection.content (#set! injection.language "bash")))
   ]
-  (#match? @_path "(^\\w+(Phase|Hook)|(pre|post)[A-Z]\\w+|script)$"))
+  (#match? @_path "(^\\w+(Phase|Hook)|(pre|post)[A-Z]\\w+|script)$")
+  (#set! injection.combined))
```
lucario387, line 44: The version before moving to upstream style doesn't have this. Can you briefly explain why it's needed?

lucario387: Consider me blind :+1: it's a good reason why.

lucario387 (approved):
I do not know about possible performance issues. But here we go

queries/nix/injections.scm
lucario387: I'm back :smile:, sorry for not mentioning this. @dasJ can you add a comment above this to note why `injection.combined` is required? It'll be easier to track than reading commit message/PR text. After that I'll merge this PR. Doesn't cause much more performance issues from what I can notice (not that Nix files have good performance in the first place)

clason:
Needs a rebase (and format).

clason:
Closing as stale.

PR #5789, opened 2023-12-14
fix(python): highlight variadic lambda parameters

ObserverOfTime:
The `args` / `**kwargs` in `lambda *args, **kwargs: ...`


lucario387 (approved):

ObserverOfTime:
Added one more fix while I was at it.

Issue #1741, opened 2021-08-23
mv: cannot stat 'tree-sitter-*' : No such file or directory


basicfunc:
Hey!!
I was setting up neovim from scratch and while configuring tree-sitter the error show below happened. I guess, it's confused about `mv` command. 

**Output of `:checkhealth nvim_treesitter`**

```
health#nvim_treesitter#check
  2 ========================================================================
  3 ## Installation
  4   - WARNING: `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :              TSInstall)
  5   - WARNING: `node` executable not found (only needed for :TSInstallFromGrammar, not required for :TSInstall)
  6   - OK: `git` executable found.
  7   - OK: `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl" }
  8   - OK: Neovim was compiled with tree-sitter runtime ABI version 13 (required >=13). Parsers must be compatible with runtime ABI.
  9 
 10 ## Parser/Features H L F I J
 11 
 12  Legend: H[ighlight], L[ocals], F[olds], I[ndents], In[j]ections
 13          +) multiple parsers found, only one will be used
 14          x) errors found in the query, try to run :TSUpdate {lang} 

```
**Output of `nvim --version`**
```
NVIM v0.5.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/lib/ccache/bin/cc -fstack-clash-protection -D_FORTIFY_SOURCE=2 -mtune=generic -O2  -g -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/builddir/neovim-0.5.0/build/config -I/builddir/neovim-0.5.0/src -I/usr/include -I/builddir/neovim-0.5.0/build/src/nvim/auto -I/builddir/neovim-0.5.0/build/include
[...]

PR #5790, opened 2023-12-14
Use main branch for HCL parser installation

heitorPB:
https://github.com/MichaHoffmann/tree-sitter-hcl uses `main` instead of `master`.

Without this change, I get an error similar to #1741 when `:TSInstall hcl`:

```
nvim-treesitter[jsonc]: Failed to execute the following command:
{
  cmd = "mv",
  opts = {
    args = { "tree-sitter-hcl-tmp/tree-sitter-hcl-master", "tree-sitter-hcl" },
    cwd = "/home/user/.local/share/nvim",
    stdio = {
      [2] = <userdata 1>,
      [3] = <userdata 2>
    }
  }
}
mv: cannot stat 'tree-sitter-hcl-tmp/tree-sitter-hcl-master': No such file or directory
```

I lost the error message, so I adapted from the linked issue.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -779,6 +779,7 @@ list.hcl = {
   install_info = {
     url = "https://github.com/MichaHoffmann/tree-sitter-hcl",
     files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
   },
   maintainers = { "@MichaHoffmann" },
 }


clason:
You do not need to (and should not) specify the default branch; it's enough for the lockfile to point to a commit on the right branch (which it does).

Please try removing the full local plugin and reinstalling (you might have a stale manifest lying around).

heitorPB:
Thanks for the clarification, @clason! do you mind pointing out where manifests and lockfiles are saved?

clason:
Just delete the whole checked out `nvim-treesitter` directory (location depends on your plugin manager).

PR #5798, opened 2023-12-17
fix(markdown): apply spell checking to tables

ribru17:


clason:
I'm not sure this is unconditionally a good idea; not all -- and I wouldn't even say the majority -- of tables contain prose that can meaningfully be spellchecked. If you propose a change, please always give some justification (why is the current situation bad and must be changed?)

ribru17:
> please always give some justification

Apologies, I will do this in the future. 

> and I wouldn't even say the majority -- of tables contain prose that can meaningfully be spellchecked

In retrospect I guess I'd have to agree; this sounds like it would be better to change on a per-user-config basis. My original justification was just that maybe most would expect spellchecking to affect tables as well (or maybe just non-title cells?) but I could only find [one complaint about this](https://www.reddit.com/r/neovim/comments/11gprcc/spellcheck_doesnt_work_in_markdown_tables/) so I think this change would do more harm than good.

clason:
Yeah, there's been too much back-and-forth around this (and conceal); we need a better (or stricter) strategy for such things going forward. My current stance is being extremely conservative with setting these by default and telling people to set these in user queries instead. (Something missing is a way to make such adjustments global instead of per language; but this is a Neovim core problem.)

PR #5807, opened 2023-12-19
feat:  support injection with incremental selection

atusy:
This PR adds the `ignore_injections` option to support injections with incremental selection.
By default, injections are ignored for the backword-compatibility.

``` lua
  incremental_selection = {
    enable = true,
    keymaps = {
      init_selection = "gnn", -- set to `false` to disable one of the mappings
      node_incremental = "grn",
      scope_incremental = "grc",
      node_decremental = "grm",
      ignore_injections = function(bufnr)
        return false
      end
    },
  },
}
```

### DIFF
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -191,6 +191,8 @@ Supported options:
     (as defined in `locals.scm`). Defaults to `grc`.
   - node_decremental: in visual mode, decrement to the previous named node.
     Defaults to `grm`.
+- ignore_injections: a function that takes bufnr and returns boolean. Defaults
+  to `true`.
 
 >
   require'nvim-treesitter.configs'.setup {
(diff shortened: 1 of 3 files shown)


clason:
Thank you, but this functionality is actually being removed from nvim-treesitter for 1.0 (to focus on parser and query management), together with the whole module system.

If this functionality is still useful to some (and not replaced by textobjects, which is more powerful), then it needs to be refactored into a (leaner) stand-alone module, similar to `indent.lua` on the `main` branch. This feature could then be included.

atusy:
Oh I see! Then I'll try make a new PR later!! Thank you for the suggestion.

atusy:
I just noticed that https://github.com/folke/flash.nvim can do the something similar.
It also works as textobj.
So I think it is okay nvim-treesitter removes the feature in the future.

``` lua
-- increment with ; and decrement with ,
vim.keymap.set({ "x", "o" }, "<leader><leader>", function()
  require("flash").treesitter({ labels = "" })
end)
```

clason:
We might still eventually have this in core, so if you change your mind, a PR would be welcome.

atusy:
Alright!

PR #5811, opened 2023-12-20
fix(cpp): incorrect indent of class with opening { on new line

jedrzejboczar:
Removed `@indent.begin` from `(class_specifier)`, otherwise it indents as:
```cpp
class Foo
    {
        int x;
        class Bar {
            int y;
        };
        Bar z;
    };
```
Added a test for this.

I also removed  `@indent.begin`  from `(condition_clause)` as I couldn't find why this is there. No tests fail due to removing it and I tried some variants of `if (x) ...` and indenting still seems to work correctly, so maybe this was some legacy code that is no longer needed due to changes in `c/indents.scm`?

### DIFF
--- a/tests/indent/cpp/class_newline.cpp
+++ b/tests/indent/cpp/class_newline.cpp
@@ -1,0 +1,9 @@
+class Foo
+{
+    int x;
+    class Bar
+    {
+        int y;
+    };
+    Bar z;
+};
--- a/queries/cpp/indents.scm
+++ b/queries/cpp/indents.scm
@@ -1,11 +1,9 @@
 ; inherits: c
 
 [
-  (class_specifier)
   (condition_clause)
 ] @indent.begin
 
 ((field_initializer_list) @indent.begin
   (#set! indent.start_at_same_line 1))
 (access_specifier) @indent.branch
-


lucario387:
When is `condition_clause` appearing though? I do not know about C++. Last time I fixed `c/indents.scm`, I just made sure that all C++ tests also passed, I did not check what is needed/not needed in C++

tests/indent/cpp/class_newline.cpp
lucario387, line 1: :+1:

jedrzejboczar:
Here are all the `condition_clause` occurrences: https://github.com/tree-sitter/tree-sitter-cpp/blob/master/grammar.js#L770-L788, so it's all just valid C.

All the following cases seem to still work for me (`switch` without braces doesn't work but I think it's invalid syntax):
```cpp
if (1)
    foo();

if (1) {
    foo();
}

if (1)
{
    foo();
}

switch (1) {
    case 1: break;
    default: break;
}

switch (1) 
{
    case 1: break;
    default: break;
}

while (1) 
    foo();


while (1) {
    foo();
}

while (1)
{
    foo();
}
```

lucario387:
We'll need to keep `condition_clause`, or adding a pattern to support mutliline switch condition in C (if any more than 1% of C programmer do that I'd be surprise)

If you deem that it's not worth it to keep it then it's ok to remove

jedrzejboczar:
This is a syntax error in treesitter parser (and possibly for a compiler too?) Screenshot from 2023-12-20 12-09-36

jedrzejboczar:
Ok, suprisingly this is valid C... `gcc -Wall test.cpp -o test` succeeds. Still I think I've never seen something like this and it makes no sense to write a switch with a single case, so I wouldn't care.

lucario387:
> This is a syntax error in treesitter parser (and possibly for a compiler too?)
> 

I mean the `switch (condition_clause)`. It's obvious not having {} is not allowed

lucario387:
I'd vote for keeping condition_clause, while class_specifier needs to go.

jedrzejboczar:
But `condition_clause` is the node with condition that is in every switch/if/while, e.g. here it is `(1)`
image
So even if we should keep `condition_clause` then I think it should be in `c/indents.scm` not in cpp, as this is just C syntax.

jedrzejboczar:
But it looks to me like it is not needed, something else is indenting all those if/while/switch statements.

lucario387:
It's for indenting inside the condtions. For if/else/while I did it from C. For switch's conditions there's nothing yet

jedrzejboczar:
Ok, I reverted removing of the `condition_clause` as I'm not sure where I should add it in c/indents.scm.

But overall, I think it is not used anyway. I added print statements in `indent.lua` to see which nodes are used to increase the indent and then tried indenting line 4 (using `=` operator) in the following code - it looks like it just uses `compound_statement` (so the curly braces `{ ... }`):
Screenshot from 2023-12-21 20-28-11

lucario387:
It's very niche, but I like this version.

lucario387 (approved):

PR #5815, opened 2023-12-21
fix(julia): pipe function highlight

degnbol:
In Julia when piping into a function the function should be highlighted as a function call, and not a variable.
This change adds `@function.call` to any identifier that are piped into by matching the operator in a binary_expression against the string "|>". It also annotates the operator with `@pipe`, as a side-effect.

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -47,6 +47,12 @@
 (broadcast_call_expression
   (field_expression (identifier) @function.call .))
 
+(binary_expression
+  (_)
+  (operator) @pipe
+  (identifier) @function.call
+  (#lua-match? @pipe "|>"))
+
 ;; Builtins
 
 ((identifier) @function.builtin
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -49,9 +49,9 @@
 
 (binary_expression
   (_)
-  (operator) @pipe
+  (operator) @_pipe
   (identifier) @function.call
-  (#lua-match? @pipe "|>"))
+  (#eq? @_pipe "|>"))
 
 ;; Builtins
 


queries/julia/highlights.scm
```diff
@@ -47,6 +47,12 @@
 (broadcast_call_expression
   (field_expression (identifier) @function.call .))
 
+(binary_expression
+  (_)
+  (operator) @pipe
+  (identifier) @function.call
+  (#lua-match? @pipe "|>"))
```
ObserverOfTime, line 54:
```suggestion
  (operator) @_pipe
  (identifier) @function.call
  (#eq? @_pipe "|>"))
```

ObserverOfTime (changes requested):

clason:
@savq

savq:
The change is ok. I agree with ObserverOfTime suggested changes.


\<rant>
In PLs with first-class functions, highlighting a function identifier is a best effort thing. I think `@function.call` should signify the _application_, not for the function itself.


This is the type of thing we wouldn't need to worry about if LS.jl had semantic highlights tho.
\</rant>

clason:
<rant>
Don't get me started on LS.jl...

"Oh, let's write a language server for Julia in Julia! It'll be grand!" ... *Runs away giggling*.
</rant>

ObserverOfTime:
> "Oh, let's write a language server for Julia in Julia! It'll be grand!" ... _Runs away giggling_.

Well R's language server is written in R and it's pretty good. :^)

clason:
@savq instead of `@pipe`, would `@operator` be suitable here (in case you _want_ to capture it)?

ObserverOfTime:
I assumed it's already captured elsewhere.

clason:
Ah, yes, it is.

clason (approved):

ObserverOfTime (approved):

PR #5819, opened 2023-12-21
feat(luap): highlights more consistent with regex

ribru17:
This PR adds some highlights similar to #5698 to be a bit more consistent. It also adds logic to make sure that any instances of `.` that are to be matched as a literal full stop will be highlighted as `@constant`, as other characters in a set should be.
Before:
beforeluapreg
After:
afterluapreg

### DIFF
--- a/queries/luap/highlights.scm
+++ b/queries/luap/highlights.scm
@@ -4,8 +4,10 @@
   "."
 ] @variable.builtin
 
-((character "." @constant) @_
- (#has-parent? @_ set negated_set))
+(set
+  (character "." @constant))
+(negated_set
+  (character "." @constant)) 
 
 [
   "[" "]"
--- a/queries/luap/highlights.scm
+++ b/queries/luap/highlights.scm
@@ -1,7 +1,7 @@
 [
   (anchor_begin)
   (anchor_end)
-] @variable.builtin
+] @punctuation.delimiter
 
 (pattern
   (character "." @variable.builtin))
--- a/queries/luap/highlights.scm
+++ b/queries/luap/highlights.scm
@@ -8,6 +8,8 @@
   (character "." @constant))
 (negated_set
   (character "." @constant)) 
+(range
+  (character "." @constant))
 
 [
   "[" "]"
@@ -28,12 +30,12 @@
 
 (set
   (character) @constant)
+(negated_set
+  (character) @constant)
 
 (class) @keyword
 
-(negated_set
-  "^" @operator
-  (character) @constant)
+(negated_set "^" @operator)
 
 (balanced_match
   (character) @parameter)
--- a/queries/luap/highlights.scm
+++ b/queries/luap/highlights.scm
@@ -1,9 +1,11 @@
-"." @character
-
 [
   (anchor_begin)
   (anchor_end)
-] @string.escape
+  "."
+] @variable.builtin
+
+(character "." @constant
+ (#has-ancestor? @constant set negated_set))
 
 [
   "[" "]"
@@ -19,7 +21,7 @@
 
 (range
   from: (character) @constant
-  "-" @punctuation.delimiter
+  "-" @operator
   to: (character) @constant)
 
 (set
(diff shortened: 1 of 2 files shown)


queries/luap/highlights.scm
```diff
@@ -1,9 +1,11 @@
...
-
 [
   (anchor_begin)
   (anchor_end)
-] @string.escape
+  "."
+] @variable.builtin
+
+(character "." @constant
+ (#has-ancestor? @constant set negated_set))
```
lucario387, line 8:
```suggestion
((character "." @constant) @_
 (#has-parent? @_ set negated_set))
```

I'd rather avoid `has-ancestor` if possible

queries/luap/highlights.scm
```diff
@@ -1,9 +1,11 @@
...
-
 [
   (anchor_begin)
   (anchor_end)
-] @string.escape
+  "."
+] @variable.builtin
+
+((character "." @constant) @_
+ (#has-parent? @_ set negated_set))
```
lucario387, line 8: The comment hours prior from me was weird.

For this use case, I like two queries rather using predicates. It generates less captures than the current version

```suggestion
(set
  (character "." @constant))
(negated_set
  (character "." @constant)) 
```

ribru17: Sounds good, thanks for the input! While we're at it, I noticed this query:
```scheme
(negated_set
  "^" @operator
  (character) @constant)
```
highlights `"^"` as `@operator` for each character matched in the negated set, e.g. `[^ffffffffff]` will have the `^` highlighted as (per `:Inspect`):
```
Treesitter
  - @string.lua links to String lua
  - @operator.luap links to Operator luap
  - @operator.luap links to Operator luap
  - @operator.luap links to Operator luap
  - @operator.luap links to Operator luap
  - @operator.luap links to Operator luap
  - @operator.luap links to Operator luap
  - @operator.luap links to Operator luap
  - @operator.luap links to Operator luap
  - @operator.luap links to Operator luap
  - @operator.luap links to Operator luap
  ```
Do you think this should also be split into two queries? Or is this not much of an issue

Comment (ribru17) [reply queries/luap/highlights.scm:8]: Another follow up; I think I may have to add the following query if it's alright:
```scheme
(range
  (character "." @constant))
```
Right now for something like `[.-0]`, the `.` is still highlighted as `@variable.builtin` rather than `@constant`. I am remembering now that this is why I had the ancestor query originally

lucario387: > Do you think this should also be split into two queries? Or is this not much of an issue

Yes it should. This should not be happening for any language/query

lucario387: > Another follow up; I think I may have to add the following query if it's alright:
> 
> ```scheme
> 
> (range
> 
>   (character "." @constant))
> 
> ```
> 
> Right now for something like `[.-0]`, the `.` is still highlighted as `@variable.builtin` rather than `@constant`. I am remembering now that this is why I had the ancestor query originally

What about the reverse?

```scheme
(character "." @constant) 
; Top-level dot
(pattern
  (character "." @variable.builtin))
```

ribru17: Oh nice! This is great

ribru17: Let me know if I should squash all of these commits if there are now too many

queries/luap/highlights.scm
```diff
@@ -1,9 +1,10 @@
-"." @character
-
 [
   (anchor_begin)
   (anchor_end)
-] @string.escape
```
lucario387, line 6: This change was not described in the PR. I'll inquire @ObserverOfTime on these kinds of decisions

ObserverOfTime: I feel like these should be `@punctuation.delimiter`. Same for their regex counterparts.

ribru17: Sounds good. Should the `.` be kept as is or also included in this capture?

ObserverOfTime: It should be kept as is.

ObserverOfTime (approved):

PR #5820, opened 2023-12-21
feat: more `@string.regex` highlights

ribru17:
Applies `@string.regex` highlights to places where `luap` is injected. Also applies it where `regex` is injected in `query` files.

**NOTE:** For the query used in the `query` files, I didn't have any success using the `#offset!` directive for some reason (maybe due to https://github.com/neovim/neovim/issues/23664? I really have no clue), even though it seemed to be working in the `injections` file... hence the hacky workaround by highlighting the actual `"` nodes

### DIFF
--- a/queries/lua/highlights.scm
+++ b/queries/lua/highlights.scm
@@ -247,3 +247,23 @@
 (string) @string
 
 (escape_sequence) @string.escape
+
+; string.match("123", "%d+")
+(function_call
+  (dot_index_expression
+    field: (identifier) @_method
+    (#any-of? @_method "find" "match" "gmatch" "gsub"))
+  arguments: (arguments
+               . (_)
+               .
+               (string
+                 content: (string_content) @string.regex)))
+
+;("123"):match("%d+")
+(function_call
+  (method_index_expression
+    method: (identifier) @_method
+    (#any-of? @_method "find" "match" "gmatch" "gsub"))
+  arguments: (arguments
+               . (string
+                   content: (string_content) @string.regex)))
--- a/queries/lua/injections.scm
+++ b/queries/lua/injections.scm
@@ -114,19 +114,7 @@
 (function_call
   (dot_index_expression
     field: (identifier) @_method
-    (#any-of? @_method "find" "match"))
-  arguments: (arguments
-               . (_)
-               .
-               (string
-                 content: (string_content) @injection.content
-                 (#set! injection.language "luap")
-                 (#set! injection.include-children))))
-
-(function_call
-  (dot_index_expression
-    field: (identifier) @_method
-    (#any-of? @_method "gmatch" "gsub"))
+    (#any-of? @_method "find" "match" "gmatch" "gsub"))
   arguments: (arguments
                . (_)
                .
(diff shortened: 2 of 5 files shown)


clason:
Nit: Lua patterns are not regular expressions...

ribru17:
Man this is unforunate... :smile: I guess for this PR we can at least keep the part that highlights the (actual) regex for `query` files? As for the `luap` highlights it doesn't break my heart to drop these.

Do you think down the line it would be possible to perhaps have a broader capture group (e.g. `@string.regex` $\rightarrow$ `@string.pattern` (or another appropriate name)) so as to include the Lua patterns? I agree with the nit, but it is a shame because conceptually I think it would be nice to have such strings highlighted the same as (or similar to) the actual regular expressions

clason:
It was more about the comments; we don't need (or want) super-specific capture names.

ribru17:
Ah I see, thanks for the clarification

clason:
@MunifTanjim

queries/lua/highlights.scm
```diff
@@ -247,3 +247,41 @@
...
+
+(function_call
+  (dot_index_expression
+    field: (identifier) @_method
+    (#any-of? @_method "gmatch" "gsub"))
+  arguments: (arguments
+               . (_)
+               .
+               (string
+                 content: (string_content) @string.regex)))
```
MunifTanjim, line 270: These look kinda identical (unless I'm missing something) 🤔

Can't these be combined together?

```scm
(function_call
  (dot_index_expression
    field: (identifier) @_method
    (#any-of? @_method "find" "match" "gmatch" "gsub"))
  (arguments
     . (_)
     . (string (string_content) @string.regex)))
```

clason: (and same below)

ribru17: I merged the identical queries in this PR. This feedback I think is also applicable to the `luap` injection counterparts that I copied the queries from. I may address this in a following PR (unless anyone would like it to be done in this one as well)

clason: Did you skip the luau queries intentionally?

clason: And if you squash these commits, you can add the luap changes as a separate one; no need for a new PR.

ribru17: > Did you skip the luau queries intentionally?

Yes, it turns out that each of the queries was actually slightly different (two using `method_index_expression` and two using `dot_index_expression`, with each query in the different groups using either `(arguments (_) . (string content: _ @string.regex))` or `(arguments (_) (string content: _ @string.regex))`). But I think this is a bug, the `luau`  highlights (and injections) are too liberal (unlike their `lua` counterparts). Example:
luauhighlightbugwithluap
The `lua` highlights and injections would not apply to the third argument on line two.

> And if you squash these commits, you can add the luap changes as a separate one; no need for a new PR.

:+1: I will do this to address the above.

Out of curiosity, is there any reason `luau` doesn't just inherit from `lua`?

clason: Gotta ask the maintainer.

clason (approved):

queries/lua/highlights.scm
lucario387: Lua has neovim-specific highlights, obviously luau won't have and won't need them

ribru17: @lucario387 That makes sense, thank you

PR #5831, opened 2023-12-24
feat!: align with upstream captures

### LINKED ISSUE #4799: Tracking issue: Moving towards Helix captures
In an effort to make query collaboration and sharing easier, some of our captures would be better off following upstream's and/or Helix's convention. However, not all of them offer benefits, and some potentially make us worse off. With that being said, the following list below is what I feel would be *good* changes to follow and conform to, discourse welcome!

### [Tree-sitter "standard capture names"](https://github.com/tree-sitter/tree-sitter/blob/3f44b896852eb7daaa6df4fb778c9bb52c70c815/highlight/src/lib.rs#L20-L72):

- [ ] `@parameter` -> `@variable.parameter`
- [ ] `@field` -> `@variable.member`
- [ ] `@namespace` -> `@module`
- [ ] `@float` -> `@number.float`
- [ ] `@symbol` -> `@string.special.symbol`
- [ ] `@string.regex` -> `@string.regexp`
- [ ] `@text.*` -> `@markup.*` (`strong`, `italic`, `link`, `strikethrough`; with exceptions; see below)
- [ ] `@text.title` -> `@markup.heading`
- [ ] `@text.literal` -> `@markup.raw`
- [ ] `@text.reference` -> `@markup.link`
- [ ] `@text.uri` -> `@markup.link.url` (in markup links)
- [ ] `@string.special` -> `@markup.link.label` (non-url links)
- [ ] `@punctuation.special` -> `@markup.list` (markdown lists only; move subitems from `@text.todo`)

### [Helix captures](https://docs.helix-editor.com/master/themes.html#syntax-highlighting):

- [ ] `@method` -> `@function.method`
- [ ] `@method.call` -> `@function.method.call`
- [ ] `@text.{todo,warning,note,danger}` -> `@comment.{error,warning,hint,info,ok}` (Helix uses `@error` etc., but that conflicts with parsing errors)
- [ ] `@text.diff.*` -> `@diff.*` (`plus`, `minus`, `delta`)
- [ ] `@text.uri` -> `@string.special.url` (outside markup)
- [ ] `@preproc` -> `@keyword.directive`
- [ ] `@storageclass` -> `@keyword.storage`

#### Non-changes:

- move control captures to `@keyword.control.*` (too much nesting)
- use other "standard capture names" of questionable value (`@escape`, `@carriage-return`)

### Possible follow-ups

- [ ] distinguish `@comment.line` and `@comment.block`
- [ ] distinguish `@markup.raw.inline` and `@markup.raw.block`
- [ ] distinguish `@type.enum`, `@type.enum.variant`, `@type.interface`
- [ ] deprecate redundant captures (`@property`, ...)

## Roadmap

The following steps will be taken over the next days/weeks:

1. freeze master
2. make changes on main
3. make main the default target for any changes (I'll have to see about moving the CI over)
4. upstream changes (including documentation)
5. make a compat release from master
6. backport query change commits from main to master.

We can discuss whether we want to move the target back to master then (which would be a bit of a mess when rebasing main again, but only once) or keep backporting query and lockfile changes periodically.

clason:
Closes #4799

**Note: This targets `main` branch; `master` is frozen while working on this. See #4799 for the roadmap.**

### [Tree-sitter "standard capture names"](https://github.com/tree-sitter/tree-sitter/blob/3f44b896852eb7daaa6df4fb778c9bb52c70c815/highlight/src/lib.rs#L20-L72):

- [x] `@parameter` -> `@variable.parameter`
- [x] `@field` -> `@variable.member`
- [x] `@namespace` -> `@module`
- [x] `@float` -> `@number.float`
- [x] `@symbol` -> `@string.special.symbol`
- [x] `@string.regex` -> `@string.regexp`
- [x] `@text.*` -> `@markup.*` (`strong`, `italic`, `link`, `strikethrough`; with exceptions; see below)
- [x] `@text.title` -> `@markup.heading`
- [x] `@text.literal` -> `@markup.raw`
- [x] `@text.reference` -> `@markup.link`
- [x] `@text.uri` -> `@markup.link.url` (in markup links)
- [x] `@string.special` -> `@markup.link.label` (non-url links)
- [x] `@punctuation.special` -> `@markup.list` (markdown lists only; move subitems from `@text.todo`)

### [Helix captures](https://docs.helix-editor.com/master/themes.html#syntax-highlighting):

- [x] `@method` -> `@function.method`
- [x] `@method.call` -> `@function.method.call`
- [x] `@text.{todo,warning,note,danger}` -> `@comment.{error,warning,hint,info,ok}` (Helix uses `@error` etc., but that conflicts with parsing errors)
- [x] `@text.diff.*` -> `@diff.*` (`plus`, `minus`, `delta`)
- [x] `@text.uri` -> `@string.special.url` (outside markup)
- [x] `@preproc` -> `@keyword.directive`
- [x] `@define` -> `@keyword.directive`(`.define`?)
- [x] `@storageclass` -> `@keyword.storage`
- [x] `@conditional` -> `@keyword.conditional`
- [x] `@debug` -> `@keyword.debug`
- [x] `@exception` -> `@keyword.exception`
- [x] `@include` -> `@keyword.import`
- [x] `@repeat` -> `@keyword.repeat`

### Cleanup

- [x] remove redundant `@conceal` (but still allow it for conceal-only patterns)
- [x] remove obsolete `@error` (syntax linting is out of scope for this repo)
- [x] sort, cleanup capture list in `CONTRIBUTING.md`


runtime/queries/blueprint/highlights.scm
```diff
@@ -36,7 +36,7 @@
 
 (import_statement (gobject_library) @module)
 
-(import_statement (version_number) @float)
+(import_statement (version_number) @number.float)
 
 (float) @float
```
lucario387, line 41: missing

runtime/queries/luau/highlights.scm
```diff
@@ -252,7 +252,7 @@
   (dot_index_expression
     field: (identifier) @_method
     (#any-of? @_method "find" "format" "match" "gmatch" "gsub"))
-  arguments: (arguments . (_) . (string content: _ @string.regex)))
+  arguments: (arguments . (_) . (string content: _ @string.regexp)))
 
 ; ("123"):match("%d+")
 (function_call
```
lucario387, line 258: missing string.regexp for the one below

runtime/queries/awk/highlights.scm
```diff
@@ -107,7 +107,7 @@
   "@load"
 ] @include
 
-"@namespace" @preproc
+"@module" @preproc
```
lucario387, line 110: unwanted change

runtime/queries/blueprint/highlights.scm
clason: That's... worrying.

runtime/queries/awk/highlights.scm
clason: Yeah, I just realized I gotta watch out for those (luadoc, objc, and phpdoc are also tricky).

clason:
Ugh, I just realized that the tests sometimes don't use the `@` for the expected captures. That'll make things very tedious... Maybe I'll need to change them all to include them first :/

lucario387:
My current script to only change the captures. Just put it in `scripts/` or put it under main and change `fnamemodify`

```lua

local function replace(bufnr, from, to)
  local query_pattern = string.format([[
    (capture
      (identifier) @replace
      (#eq? @replace "%s")
      (#set! replace "%s"))
  ]], from, to)
  local root = vim.treesitter.get_parser(bufnr, "query"):parse(true)[1]:root()
  local query = vim.treesitter.query.parse("query", query_pattern)
  for _, node, metadata in query:iter_captures(root, bufnr) do
    local srow, scol, erow, ecol = node:range()
    vim.api.nvim_buf_set_text(bufnr, srow, scol, erow, ecol, { metadata["replace"] })
  end
end

local from = "string.regex"
local to = "string.regexp"

local source = vim.fn.fnamemodify(string.sub(debug.getinfo(1, "S").source, 2), ":p:h:h")
---@type string[]
local files = vim.fn.glob(source .. "/runtime/queries/**/highlights.scm", true, true, true)
for _, file in ipairs(files) do
  local bufnr = vim.fn.bufadd(file)
  vim.fn.bufload(file)
  vim.print(file)
  replace(bufnr, from, to)
end

vim.cmd("wqa!")


```

lucario387:
It can easily be extended to multi pattern if you want.

lucario387:
Oh, and the list above has dupliced `@text.uri` @clason

clason:
Ok, I'm mostly through and will push when I have network access again. 

Open questions:

1. There's a bunch of captures that seem to use text.underline purely for markup (in addition to other syntactic queries). I strongly feel that these should be removed and proper syntax captures be used. 

2. There are a number of pure text captures. Some of them may be needed to override a containing capture (mostly LaTeX, which is too different to fit into the current framework), and some don't seem to be useful at all (mostly used in combination with spell). I suggest to change the former to none and remove the latter.

clason:
(The test failure is a weird Rust indent regression that does not happen on `master`.)

ObserverOfTime:
> There's a bunch of captures that seem to use text.underline purely for markup (in addition to other syntactic queries). I strongly feel that these should be removed and proper syntax captures be used.

Most of those should be filenames.

lucario387:
> (The test failure is a weird Rust indent regression that does not happen on `master`)

Will look at it tomorrow

ofseed:
> - [x] `@string.special` -> `@markup.link.label` (non-url links)

Seems that all `@markup.link.label` were `@text.reference` before, and no `@string.special` become `markup.link.label` ?

lucario387:
> Seems that all `@markup.link.label` were `@text.reference` before, and no `@string.special` become `markup.link.label` ?

Can be done in a follow up to this. This PR already is too big :)

Of course, you can also note which `@string.special` should become `@markup.link.label`

clason:
It's about future captures (documentation) as much as about current ones. Not all captures were actually used as such.

ofseed:
Oh, thank you for the reply. I thought the checked todos were marked as completed.

CONTRIBUTING.md
```diff
@@ -186,7 +186,7 @@ As languages differ quite a lot, here is a set of captures available to you when
 @type.definition ; identifiers in type definitions (e.g. `typedef <type> <identifier>` in C)
 @type.qualifier  ; type qualifiers (e.g. `const`)
 
-@storageclass    ; modifiers that affect storage in memory or life-time
+@keyword.storage    ; modifiers that affect storage in memory or life-time
 @attribute       ; attribute annotations (e.g. Python decorators)
 @field           ; object and struct fields
 @property        ; similar to `@field`
```
lucario387, line 192: `@field` is changed to `@variable.member`, from what I can see. The doc should also be updated

runtime/queries/capnp/highlights.scm
```diff
@@ -14,7 +14,7 @@
   "using"
 ] @include
 
-(import_path) @string @text.uri
+(import_path) @string @string.special.url
```
lucario387, line 17:
```suggestion
(import_path) @string.special.url
```

CONTRIBUTING.md
clason: How should we document `@property` then?

clason:
There seem to be a few queries that (illegally) use "highlight captures" (i.e., captures based on how they should look) rather than "syntax captures" (i.e., captures that describe the closest syntax analogue). One prime offender is RST (@stsewd), but there may be others.

(LaTeX is also a bit of a mess in this regard; some languages just do not fit into any syntax scheme that is built with programming languages in mind. A follow-up PR should look at these specifically and extend the `@markup` hierarchy, I think.)

clason:
OK, I'm through with the plan now. Please review, especially the last four commits.

Regarding the last commit: I've opted to err in the direction of caution and left `@none` for `@text` when I suspected this might be injected. Let me know if some of these are superfluous. (I've also commented out instead of removing captures that were pointless in my view; again, let me know if those should be restored or deleted.)

CONTRIBUTING.md
lucario387: Very much unsure. `@property` usage is all over the place right now.

clason: That's not the only one...

lucario387: We will need community support to reliably convert `@property` into actually meaningful(relatable) captures ... So a backlog type of stuff I feel.

clason: I think I may have been too optimistic here; it's not enough to simply rename a bunch of captures, we need to also deprecate some and push for changes on every PR. Those won't be as breaking so can indeed be done down the line, but it would be good to get the ball rolling here and identify the ones we want to see the back of.

lucario387: This would also mean to **separate** from tree-sitter upstream queries, which goes against the intention of the renaming. As upstream also sometime uses `@property` incorrectly.

clason: Yes, that's what I suspect.

clason: (It's always the same queries that are fishy: re2c, chatito, capnp, git_commit, prompql, ini, ...)

CONTRIBUTING.md
```diff
@@ -178,90 +190,83 @@ As languages differ quite a lot, here is a set of captures available to you when
...
+
+```query
+@none    ; completely disable the highlight
+@error   ; syntax/parser errors
+```
+
+##### Conceal
 
 ```query
 @conceal ; for captures that are only used for concealing
```
lucario387, line 264: `@conceal` is not needed for concealing, only the `#set!` is needed. We can use conceal alongside something like markdown's list checked/unchecked markers to have both the highlight and the conceal. So should this PR also removes/deprecates the usage of this query?

clason: You're right; the capture is completely unnecessary -- if you _need_ a placeholder because no other capture fits, you can always use `@_conceal`.

runtime/queries/json/highlights.scm
```diff
@@ -19,10 +19,9 @@
...
 
-(("\"" @conceal)
+(("\"" @string)
  (#set! conceal ""))
 
-(escape_sequence) @string.escape
-((escape_sequence) @conceal
- (#eq? @conceal "\\\"")
+((escape_sequence) @string.escape
+ (#eq? @string.escape "\\\"")
```
lucario387, line 26: These aren't equivalent queries. We still need to separate them
```suggestion
(escape_sequence) @string.escape
((escape_sequence) @string.escape
 (#eq? @string.escape "\\\"")
```

CONTRIBUTING.md
lucario387: `@_conceal` still passed the highlights down...
I tested this query for `json`
```query
((escape_sequence) @_conceal
 (#eq? @_conceal "\\\"")
 (#set! conceal "\""))
```
And using `:Inspect` gives the result:
```
Treesitter                                                                                     
  - @string.json links to @string json                                                         
  - @string.escape.json links to @string.escape json                                           
  - @_conceal.json links to @_conceal json
```

I'm definitely surprised by this behavior, should this be documented in a later PR?

runtime/queries/markdown_inline/highlights.scm
```diff
@@ -29,11 +27,8 @@
 ] @string.escape
 
 ; Conceal codeblock and text style markers
-([
-  (code_span_delimiter)
-  (emphasis_delimiter)
-] @punctuation.delimiter
-(#set! conceal ""))
+(code_span_delimiter) @markup.raw (#set! conceal "")
+(emphasis_delimiter) @markup.strong (#set! conceal "")
```
lucario387, line 31: Possible readability issue. They should be inside a grouping, similar to other queries with just `(#set!)`
```suggestion
((code_span_delimiter) @markup.raw 
  (#set! conceal ""))
((emphasis_delimiter) @markup.strong 
  (#set! conceal ""))
```

CONTRIBUTING.md
```diff
@@ -107,66 +107,78 @@ you can mark the language as optional (by putting it between parenthesis).
...
-@function.macro   ; preprocessor macros
+@function             ; function definitions
+@function.builtin     ; built-in functions
+@function.call        ; function calls
+@function.macro       ; preprocessor macros
 
-@method           ; method definitions
-@method.call      ; method calls
+@function.method      ; method definitions
+@function.method.cal  ; method calls
```
lucario387, line 165: Typos

CONTRIBUTING.md
ObserverOfTime: `@property` should be used for css properties and keys in data/config languages (e.g. json/toml/etc.)

runtime/queries/chatito/highlights.scm
ObserverOfTime, line 1: `(alias)` can use `@keyword.directive` and `@label` can be replaced with `@property` instead.

runtime/queries/cmake/highlights.scm
```diff
@@ -21,7 +21,7 @@
 
 (normal_command (identifier) @function)
 
-["ENV" "CACHE"] @storageclass
+["ENV" "CACHE"] @keyword.storage
```
ObserverOfTime, line 24:
```suggestion
["ENV" "CACHE"] @module
```

runtime/queries/cpon/highlights.scm
ObserverOfTime, line 1: `@label` should be replaced with `@property`.

runtime/queries/css/highlights.scm
```diff
@@ -57,7 +57,7 @@
  (attribute_name)
  ] @property
```
ObserverOfTime, line 58: These highlights should be fixed if that's not out of scope.
I'll take another look later to figure out what they should be.

CONTRIBUTING.md
```diff
@@ -107,156 +107,158 @@ you can mark the language as optional (by putting it between parenthesis).
...
-#### Conceal
+#### Non-highlighting captures
 
 ```query
-@conceal ; for captures that are only used for concealing
+@none    ; completely disable the highlight
+@error   ; syntax/parser errors
 ```
 
-`@conceal` must be followed by `(#set! conceal "")`.
```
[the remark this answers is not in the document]
lucario387: `:h treesitter-highlight-conceal` maybe

runtime/queries/cmake/highlights.scm
lucario387: Env and cache in cmake isn't exactly module if i understand correctly

ObserverOfTime: They're containers.

runtime/queries/css/highlights.scm
lucario387: I'd vote for keeping all `@property` stuffs for another PR, or to fix them all in this one. Might be hard as it's not just a rename though.

CONTRIBUTING.md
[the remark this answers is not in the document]
clason: Yes, `conceal` is completely handled by upstream, but it doesn't hurt to mention (especially the "private" capture). I'll add a separate section.

runtime/queries/css/highlights.scm
clason: Yes, I think we should keep fixup (changing as opposed to renaming captures) to separate PR(s); that'll make it easier to get the language maintainers involved and can happen in parallel. The more that needs to be done in this PR, the longer everything gets blocked.

Remember that the main point of this PR is to change the _allowed_ captures in CONTRIBUTING.md; changing the queries is just a necessary consequence thereof.

runtime/queries/awk/highlights.scm
```diff
@@ -47,21 +47,21 @@
...
-(regex) @string.regex
+(regex) @string.regexp
 (escape_sequence) @string.escape
 
 (comment) @comment @spell
 
-((program . (comment) @preproc)
-  (#lua-match? @preproc "^#!/"))
+((program . (comment) @keyword.directive)
+  (#lua-match? @keyword.directive "^#!/"))
```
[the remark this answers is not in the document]
clason: _That_ is the only kind of change (direct consequence of capture renaming) that's in scope here. (And the reason why it's better to review by comment rather than by file.)

I'd appreciate a list of patterns that should be changed; it's not always obvious what's a shebang and what not.

Personally, I'm not sure I agree, though; it feels more like a directive to me (and Helix uses `@keyword`, too).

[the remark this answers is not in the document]
ObserverOfTime: > I'd appreciate a list of patterns that should be changed; it's not always obvious what's a shebang and what not.

Any `@preproc` comments.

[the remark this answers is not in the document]
clason: so any `@keyword.directive` that applies to a `(comment)` node?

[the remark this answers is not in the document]
ObserverOfTime: Yeah.

lucario387:
@clason I added a commit to also convert all injection tests to `@[!]capture`, for consistency with highlight tests

runtime/queries/glimmer/highlights.scm
ObserverOfTime, line 1: **Out of scope:** should `(attribute_name)` be  `@attribute`?

runtime/queries/gosum/highlights.scm
```diff
@@ -8,11 +8,11 @@
 ] @keyword
 
 
-(module_path) @string @text.uri
+(module_path) @string.special.path
```
ObserverOfTime, line 11: Is this not a URL?

clason: I don't know; from what I could tell it's a local path. But happy to change if you know better!

ObserverOfTime: Seems to be a URL.

runtime/queries/http/highlights.scm
ObserverOfTime, line 1: **Out of scope:** `(scheme)` should be `@type`.

runtime/queries/hurl/highlights.scm
ObserverOfTime, line 1: **Out of scope:** the sections (L4-L11) should be `@module` (maybe).

runtime/queries/json/highlights.scm
ObserverOfTime, line 1: `@label` -> `@property`

runtime/queries/kotlin/highlights.scm
ObserverOfTime, line 1: `@property` -> `@variable.member`

runtime/queries/awk/highlights.scm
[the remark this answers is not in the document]
ObserverOfTime: Also any `shebang` / `hashbang` nodes that some languages have.

runtime/queries/m68k/highlights.scm
ObserverOfTime, line 1: `(path)` should be `@string.special.path`

runtime/queries/make/highlights.scm
```diff
@@ -41,10 +41,10 @@
 
 (export_directive "export" @keyword)
 (override_directive "override" @keyword)
-(include_directive ["include" "-include"] @include)
+(include_directive ["include" "-include"] @keyword.import)
```
ObserverOfTime, line 44:
```suggestion
(include_directive
  ["include" "-include"] @keyword.import
  filenames: (list (word) @string.special.path))
```

runtime/queries/matlab/highlights.scm
ObserverOfTime, line 1: `@property` -> `@variable.member`

runtime/queries/mlir/highlights.scm
```diff
@@ -328,8 +328,8 @@
 (func_dialect name: (symbol_ref_id) @function)
 (llvm_dialect name: (symbol_ref_id) @function)
 
-(func_arg_list (value_use) @parameter)
-(block_arg_list (value_use) @parameter)
+(func_arg_list (value_use) @variable.parameter)
+(block_arg_list (value_use) @variable.parameter)
 
 (caret_id) @tag
```
ObserverOfTime, line 334: **Out of scope:** I don't know about this one.

runtime/queries/objc/highlights.scm
ObserverOfTime, line 1: `@property` -> `@variable.member`

runtime/queries/ocaml/highlights.scm
ObserverOfTime, line 1: `@property` -> `@variable.member` (except `(attribute_id)` which should be `@attribute`)

runtime/queries/pascal/highlights.scm
ObserverOfTime, line 1: `@property` -> `@variable.member`

runtime/queries/passwd/highlights.scm
```diff
@@ -1,12 +1,12 @@
...
-(auth) @symbol
+(auth) @string.special.symbol
 
 (gecos) @string
 
-(home) @text.uri @constant
+(home) @string.special.url @constant
 
-(shell) @text.uri @string.special
+(shell) @string.special.url
```
ObserverOfTime, line 9:
```suggestion
(shell) @string.special.path
```

runtime/queries/passwd/highlights.scm
```diff
@@ -1,12 +1,12 @@
-(user) @namespace
+(user) @module
 
-(auth) @symbol
+(auth) @string.special.symbol
 
 (gecos) @string
 
-(home) @text.uri @constant
+(home) @string.special.url @constant
```
ObserverOfTime, line 7:
```suggestion
(home) @string.special.path @constant
```
Should we remove the `@constant`?

runtime/queries/phpdoc/highlights.scm
ObserverOfTime, line 1: `@property` -> `@variable.member`

runtime/queries/poe_filter/highlights.scm
```diff
@@ -36,6 +36,6 @@
 "\"" @punctuation.delimiter
 
 ; conceal unnecessary quotes
-("\"" @conceal
-  (#not-has-parent? @conceal string file)
+("\"" @string
+  (#not-has-parent? @string string file)
```
ObserverOfTime, line 40:
```suggestion
("\"" @punctuation.delimiter
  (#not-has-parent? @punctuation.delimiter string file)
```
Would a `@_conceal` capture work for conceal-only queries?

runtime/queries/proto/highlights.scm
ObserverOfTime, line 1: **Out of scope:** "proto3" & "proto2" should be `@string.special`.

runtime/queries/pug/highlights.scm
```diff
@@ -18,7 +18,7 @@
 (id) @constant
 (class) @property
```
ObserverOfTime, line 19: **Out of scope:** should be changed to match whatever we use in css (`@type`?).

runtime/queries/puppet/highlights.scm
ObserverOfTime, line 1: `@property` -> `@attribute` probably

runtime/queries/poe_filter/highlights.scm
lucario387: I'd prefer `@conceal`, as `@_conceal` will still be applied

ObserverOfTime: But we'd have to document `@conceal` for it to work.

runtime/queries/passwd/highlights.scm
clason: Yes. Double captures are never acceptable; either it's a constant or a custom `@string.special.path.builtin` (which I'd prefer to leave for user customization).

runtime/queries/poe_filter/highlights.scm
clason: So will `@conceal`? There's nothing special about that in the treesitter highlighter.

clason: And the point of these non-conceal captures is to have the expected highlighting if `conceal` is not set.

runtime/queries/rasi/highlights.scm
ObserverOfTime, line 1: **Out of scope:** should `(property_name)` be `@variable.member` instead of `@variable`?

runtime/queries/rbs/highlights.scm
ObserverOfTime, line 1: `(ivar_name)` & `(cvar_name)` should be `@variable.member`
`(global_name)` should be `@variable` (or `@variable.global`)

runtime/queries/ron/highlights.scm
ObserverOfTime, line 1: `@property` -> `@variable.member`

runtime/queries/ruby/highlights.scm
ObserverOfTime, line 1: **Out of scope:**
`(class_variable)` & `(instance_variable)` should be `@variable.member`
 `(heredoc_beginning)` & `(heredoc_end)` should be `@label`

runtime/queries/poe_filter/highlights.scm
lucario387: > But we'd have to document `@conceal` for it to work.

Do we need to? I thought @clason has already mentioned conceals in `CONTRIBUTING.md`

clason: I have documented that you do _not_ need `@conceal`.

ObserverOfTime: > And the point of these non-conceal captures is to have the expected highlighting if `conceallevel` is not set.

But they're already captured before.

runtime/queries/scala/highlights.scm
ObserverOfTime, line 1: **Out of scope:**
`@property` -> `@variable.member`
`(symbol_literal)` should be `@string.special.symbol`
`(character_literal)` should be `@character`

runtime/queries/smali/highlights.scm
ObserverOfTime, line 1: **Out of scope:** `(annotation_key)` should probably be `@attribute`.

runtime/queries/snakemake/highlights.scm
ObserverOfTime, line 1: **Out of scope:** `@label` -> `@property` (probably)

runtime/queries/soql/highlights.scm
ObserverOfTime, line 1: **Out of scope:** `@property` -> `@variable.member`

runtime/queries/squirrel/highlights.scm
ObserverOfTime, line 1: **Out of scope:** `@property` -> `@variable.member`

runtime/queries/supercollider/highlights.scm
ObserverOfTime, line 1: **Out of scope:** `@property` -> `@variable.member`

runtime/queries/swift/highlights.scm
ObserverOfTime, line 1: **Out of scope:** `@property` -> `@variable.member`

runtime/queries/poe_filter/highlights.scm
ObserverOfTime: > And the point of these non-conceal captures is to have the expected highlighting if `conceallevel` is not set.

But they're already highlighted.

runtime/queries/teal/highlights.scm
ObserverOfTime, line 1: **Out of scope:** the `string.format` comments should be removed and replaced with `format` injections.

runtime/queries/templ/highlights.scm
ObserverOfTime, line 1: **Out of scope:** the `css_*` nodes should be changed to match whatever we use in css.

runtime/queries/tiger/highlights.scm
ObserverOfTime, line 1: **Out of scope:** `@property` -> `@variable.member`

runtime/queries/tlaplus/highlights.scm
ObserverOfTime, line 1: **Out of scope:** the capture names comment is outdated and should be removed.

runtime/queries/verilog/highlights.scm
ObserverOfTime, line 1: **Out of scope:**
"begin" & "end" should be `@keyword`
`(port_direction)` should be `@keyword.storage`

runtime/queries/vhs/highlights.scm
```diff
@@ -31,8 +31,8 @@
 
 [ "@" ] @operator
 (control) @function.macro
-(float) @float
+(float) @number.float
 (integer) @number
 (comment) @comment @spell
 [(path) (string) (json)] @string
```
ObserverOfTime, line 37:
```suggestion
[(string) (json)] @string
(path) @string.special.path
```

runtime/queries/vhs/highlights.scm
```diff
@@ -31,8 +31,8 @@
...
-(time) @symbol
+(time) @string.special.symbol
```
ObserverOfTime, line 38:
```suggestion
(time) @string.special
```

runtime/queries/poe_filter/highlights.scm
clason: Then the patterns should be combined.

ObserverOfTime: No because they shouldn't always be concealed.

runtime/queries/wing/highlights.scm
ObserverOfTime, line 1: **Out of scope:** 
`(member_identifier)` should be `@variable.member`
"return" should be `@keyword.return`
"if" & "else" should be `@keyword.conditional`
"for" should be `@keyword.repeat`

runtime/queries/yaml/highlights.scm
ObserverOfTime, line 1: **Out of scope:**
`@variable.member` -> `@property`
`(anchor_name)` & `(alias_name)` should be `@label`

runtime/queries/yang/highlights.scm
ObserverOfTime, line 1: **Out of scope:**  `(enum_value)` should probably be `@constant`.

runtime/queries/json/highlights.scm
ObserverOfTime: This could also use `@_conceal`
```clojure
((escape_sequence) @_conceal
 (#eq? @_conceal "\\\"")
 ```

File: runtime/queries/poe_filter/highlights.scm [hunk 23]
```diff
@@ -36,6 +36,6 @@
 "\"" @punctuation.delimiter
 
 ; conceal unnecessary quotes
-("\"" @conceal
-  (#not-has-parent? @conceal string file)
+("\"" @_conceal
+  (#not-has-parent? @_conceal string file)
   (#set! conceal ""))
```
lucario387, line 41: Just noting: This will still be captured as `@_conceal.poe_filter`, so we can still give it highlight groups

clason: But we don't want to, in this case, since the highlight is already applied in the previous pattern?

ObserverOfTime: :thinking:

lucario387: I'd go for duplicating the highlight captures. At least it's consistent highlighting with or without conceal

runtime/queries/vim/highlights.scm
ObserverOfTime, line 1: **Out of scope:** `(literal_key)` should be `@property` (almost missed it).

CONTRIBUTING.md
```diff
@@ -107,156 +107,158 @@ you can mark the language as optional (by putting it between parenthesis).
...
 @tag.delimiter ; XML tag delimiters
 ```
 
-#### Conceal
+#### Non-highlighting captures
 
 ```query
-@conceal ; for captures that are only used for concealing
+@none    ; completely disable the highlight
+@error   ; syntax/parser errors
```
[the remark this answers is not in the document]
clason: we do not use it anymore at all -- or at least shouldn't.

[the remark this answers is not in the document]
clason: (it's only used in `html_tags` and `gitattributes` would you be OK with removing them/commenting them out?)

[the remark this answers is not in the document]
ObserverOfTime: No. :^)

runtime/queries/poe_filter/highlights.scm
clason: > Oh, it's true, the @_conceal capture shows up in Inspect.

But so does `@conceal`? (I'm fine with keeping the highlight, though.)

runtime/queries/json/highlights.scm
```diff
@@ -19,7 +19,7 @@
  "{" "}"
 ] @punctuation.bracket
 
-(("\"" @conceal)
+(("\"" @string)
```
ObserverOfTime, line 22:
```suggestion
(("\"" @conceal)
```

ObserverOfTime (changes requested):

ObserverOfTime (approved):

lucario387 (approved):
Let the ship sail

clason:
Follow-ups (separate PR each):

* `@property`/`@label`/`@variable.member` cleanup as outlined in the comments (and improving the guidance in `CONTRIBUTING.md`)
* possibly `@comment.shebang` (I'm skeptical on principle, but maybe it makes sense when actually seeing it)

wookayin:
Status: this was merged to main, and backported to master (#5895, 1ae9b0e4558fe7868f8cda2db65239cfb14836d0) with additional changes and fixups. The main branch is rebased to master, so the list of changes described are not up-to-date. See 1ae9b0e4558fe7868f8cda2db65239cfb14836d0 for the most up-to-date list of changes.

PR #5836, opened 2023-12-28
fix(julia): add missing capture of ":" in selected import

degnbol:
If writing a selected import, e.g. `using BSON: @load`, the colon isn't currently captured. This change captures it as `@punctuation.delimiter`.

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -398,6 +398,8 @@
   "as" @include)
 (export_statement
   "export" @include)
+(selected_import
+  ":" @punctuation.delimiter)
 
 (struct_definition
   ["struct" "end"] @keyword)


theHamsta (approved):

clason:
@savq 

(We currently don't accept PRs to `master`, so we'll have to park this and then change the target to `main` later.)

savq:
LGTM

> (We currently don't accept PRs to master, so we'll have to park this and then change the target to main later.)

Don't know what you mean by this. Is the repo changing the integration branch?

clason:
> Don't know what you mean by this. Is the repo changing the integration branch?

Yes, but this is about a current massive change in capture naming (see #4799) during which we have frozen both branches. This change should apply cleanly, though, so we can try merging and rebasing it.

PR #5837, opened 2023-12-28
feat(bash): highlight commands called via keyword

ribru17:
Highlights commands that are called with the `command` keyword as `@function.call`, as regular commands are highlighted.
Before:
beforebashcommand
After:
afterbashcommand

### DIFF
--- a/queries/bash/highlights.scm
+++ b/queries/bash/highlights.scm
@@ -144,6 +144,12 @@
              (concatenation (word) @parameter)
              ])
 
+(command
+  name: (command_name) @_name
+  .
+  argument: (word) @function.call
+  (#eq? @_name "command"))
+
 (number) @number
 ((word) @number
  (#lua-match? @number "^[0-9]+$"))


clason:
@TravonteD

queries/bash/highlights.scm
```diff
@@ -144,6 +144,12 @@
              (concatenation (word) @parameter)
              ])
 
+(command
+  name: (command_name) @_name
+  .
+  argument: (word) @function.call
+  (#eq? @_name "command"))
```
ObserverOfTime, line 151: I assume this doesn't cover `command -[pVv] foo`.
There are also other similar commands like `which`, `whereis`, `hash`, `sudo`, `pkexec` …
So while I'm not against this change, it might be too complicated to properly implement.

ribru17: Dang, you're right. I am fine to close this then :+1:

PR #5852, opened 2024-01-04
feat: add hyprlang parser & queries

luckasRanarison:
Added [hyprlang](https://github.com/luckasRanarison/tree-sitter-hyprlang/) parser and queries. 

It's my first contribution to nvim-treesitter, feedbacks are welcome!

### DIFF
--- a/queries/hyprlang/highlights.scm
+++ b/queries/hyprlang/highlights.scm
@@ -49,7 +49,7 @@
 
 (hex) @number
 
-[ "," ] @punctuation
+[ "," ] @punctuation.delimiter
 
 [
   "("
--- a/queries/hyprlang/highlights.scm
+++ b/queries/hyprlang/highlights.scm
@@ -1,0 +1,65 @@
+(comment) @comment
+
+(source "source" @keyword)
+
+(exec "exec" @keyword)
+(exec "exec-once" @keyword)
+
+(keyword
+  (name) @keyword
+)
+
+(assignment
+  (name) @property
+)
+
+(section
+  (name) @namespace
+)
+
+(section
+  device: (name) @type
+)
+
+(variable) @variable
+(variable "$" @punctuation.special)
+
+(number) @number
+
+(boolean) @boolean
+
+(mod) @constant
+
+(vec2
+  (number) @number
+  (number) @number
+)
+
+[
+  "rgb"
+  "rgba"
+] @function
+
+(color
+  (legacy_hex) @number
+)
+
+(angle) @number
+(angle "deg" @type)
+
+(hex) @number
+
+[ "," ] @punctuation
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+] @punctuation.bracket
+
+[
+  "="
+  "-"
+  "+"
+] @operator
--- a/queries/hyprlang/highlights.scm
+++ b/queries/hyprlang/highlights.scm
@@ -1,6 +1,10 @@
 (comment) @comment
 
-[ "source" "exec" "exec-once" ] @keyword
+[
+  "source"
+  "exec"
+  "exec-once"
+] @keyword
 
 (keyword
   (name) @keyword)
@@ -15,7 +19,9 @@
   device: (name) @type)
 
 (variable) @variable
-(variable "$" @punctuation.special)
+
+(variable
+  "$" @punctuation.special)
 
 (number) @number
 
@@ -32,11 +38,13 @@
   (legacy_hex) @number)
 
 (angle) @number
-(angle "deg" @type)
+
+(angle
+  "deg" @type)
 
 (hex) @number
 
-[ "," ] @punctuation.delimiter
+"," @punctuation.delimiter
 
 [
   "("
--- a/queries/hyprlang/highlights.scm
+++ b/queries/hyprlang/highlights.scm
@@ -5,20 +5,16 @@
 [ "exec" "exec-once" ] @keyword
 
 (keyword
-  (name) @keyword
-)
+  (name) @keyword)
 
 (assignment
-  (name) @property
-)
+  (name) @property)
 
 (section
-  (name) @namespace
-)
+  (name) @namespace)
 
 (section
-  device: (name) @type
-)
+  device: (name) @type)
 
 (variable) @variable
 (variable "$" @punctuation.special)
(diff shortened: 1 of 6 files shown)


queries/hyprlang/highlights.scm
```diff
@@ -0,0 +1,65 @@
...
+
+(vec2
+  (number) @number
+  (number) @number
+)
+
+[
+  "rgb"
+  "rgba"
+] @function
```
theHamsta, line 41: Might be even function.builtin

queries/hyprlang/highlights.scm
```diff
@@ -0,0 +1,65 @@
+(comment) @comment
+
+(source "source" @keyword)
+
+(exec "exec" @keyword)
+(exec "exec-once" @keyword)
```
lucario387, line 6:
```suggestion
(exec [ "exec" "exec-once" ] @keyword)
```

You can probably do this as well
```suggestion
[ "exec" "exec-once" ] @keyword
```

queries/hyprlang/highlights.scm
```diff
@@ -0,0 +1,65 @@
...
+
+(keyword
+  (name) @keyword
+)
```
lucario387, line 10: Stylings
```suggestion
(keyword
  (name) @keyword)
```

Same for other lines

queries/hyprlang/highlights.scm
```diff
@@ -0,0 +1,65 @@
...
+
+(number) @number
+
+(boolean) @boolean
+
+(mod) @constant
+
+(vec2
+  (number) @number
+  (number) @number
```
lucario387, line 35: One should be enough. This creates n^2 captures for n number nodes

clason:
Note that we are currently in a feature freeze due to the rewrite of capture groups, so we cannot merge this right now. You can either wait until we have backported the rewrite to `master`, or target the `main` branch (with the new capture names).

queries/hyprlang/highlights.scm
clason: In fact, this pattern is not needed at all, since the capture in line 27 already covers this.

luckasRanarison:
> Note that we are currently in a feature freeze due to the rewrite of capture groups, so we cannot merge this right now. You can either wait until we have backported the rewrite to `master`, or target the `main` branch (with the new capture names).

Oh, I just read about nvim-treesitter rewrite. If main will eventually be merged to master I think could adapt this to the new capture groups and nvim-treesitter changes and wait.

clason:
> If main will eventually be merged to master

That won't happen, for the record -- only those commits will be cherry-picked and backported.

luckasRanarison:
So should I directly target main instead?

queries/hyprlang/highlights.scm
luckasRanarison: Done

clason:
If you want to get a head start, yes.

queries/hyprlang/highlights.scm
luckasRanarison: You're right, I fixed it

queries/hyprlang/highlights.scm
```diff
@@ -0,0 +1,54 @@
+(comment) @comment
+
+(source "source" @keyword)
+
+[ "exec" "exec-once" ] @keyword
```
theHamsta, line 5: probably `@keyword` is correct for `source`, `exec`, `exec-one`, `bind`  which probably fits nicely with the highlighting as then every line starts with a keyword. I was wondering whether they could also be something `function.builtin` like has they have something of a command. I have currently no opinion which better fits the semantics.

theHamsta:
Apparently, the parser has problems with trailing `,` for some commands. Could you check whether this is allowed in the hyperland config (with meaning empty value for that item)?

queries/hyprlang/highlights.scm
luckasRanarison: Oh, I forgot to add source with the other ones. 

I just used the keyword group because they're called [keywords](https://wiki.hyprland.org/Configuring/Keywords/) but I also think function makes more sense in a way but it could also be confusing because it's just a key value relation but not a real programming language.

luckasRanarison:
> Apparently, the parser has problems with trailing `,` for some commands. Could you check whether this is allowed in the hyperland config (with meaning empty value for that item)?

Empty values are represented with an empty space followed by a line break, so trailing comma will throw an error.

theHamsta:
> > Apparently, the parser has problems with trailing `,` for some commands. Could you check whether this is allowed in the hyperland config (with meaning empty value for that item)?
> 
> Empty values are created by adding nothing after the equal sign and putting a linebreak , so trailing comma will throw an error.

But you could have `env foo,`

luckasRanarison:
> But you could have `env foo,`
Oh, you're right. Thanks for pointing that out!

clason:
The backport PR has been merged; please **rebase on master** and note 1. the new list of valid capture names and 2. the mandatory query format (run `./scripts/format-queries.lua` on the new queries).

queries/hyprlang/highlights.scm
```diff
@@ -0,0 +1,60 @@
+(comment) @comment
```
amaanq, line 1:
```suggestion
(comment) @comment @spell
```

queries/hyprlang/highlights.scm
```diff
@@ -0,0 +1,60 @@
...
+(section
+  (name) @module)
+
+(section
+  device: (device_name) @type)
+
+(variable) @variable
+
+(variable
+  "$" @punctuation.special)
```
amaanq, line 24: If `$` isn't used as a literal anywhere else, then you can just directly query it

```suggestion
"$" @punctuation.special
```

queries/hyprlang/highlights.scm
```diff
@@ -0,0 +1,60 @@
...
+
+[
+  "rgb"
+  "rgba"
+] @function.builtin
+
+(color
+  (legacy_hex) @number)
+
+(angle) @number
```
amaanq, line 40: Same here, if `legacy_hex` is used only here, then there's no need to be specific, and on top of that, you can combine the number queries

```scm
[
  (number)
  (legacy_hex)
  (angle)
  (hex)
] @number
```

queries/hyprlang/highlights.scm
luckasRanarison: I see, thanks for the tips!

amaanq (approved):

PR #5854, opened 2024-01-04
feat: query formatting script

lucario387:
Usage: `nvim --clean -l scripts/format-queries.lua`

Related: https://github.com/nvim-treesitter/nvim-treesitter/pull/5831#issuecomment-1868475852


scripts/format-queries.lua
```diff
@@ -0,0 +1,215 @@
+-- #!/usr/bin/env -S nvim -l
+
+local ts = vim.treesitter
+local get_node_text = ts.get_node_text
+local root_path = vim.fn.fnamemodify(string.sub(debug.getinfo(1, 'S').source, 2), ':p:h:h')
+
+vim.opt.runtimepath:append(root_path)
+vim.cmd.runtime({ 'plugin/nvim-treesitter.lua', bang = true })
```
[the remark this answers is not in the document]
lucario387: I don't know. But it seems necessary. Without manually adding it in the code seems to not run

scripts/format-queries.lua
```diff
@@ -0,0 +1,216 @@
...
+---@type string[]
+local files = vim.fn.split(vim.fn.glob(root_path .. '/runtime/queries/**/*.*'))
+
+for _, file in ipairs(files) do
+  local buf = vim.fn.bufadd(file)
+  vim.fn.bufload(file)
+  format(buf, format_queries)
+end
+
+vim.cmd('wqa!')
```
clason, line 216: I don't think that's needed with `-l`, either.

clason (commented):
Feel free to commit the formatted queries (in a separate commit) so we can see how it looks!

scripts/format-queries.lua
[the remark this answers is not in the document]
clason: Oh, you're using `#not-has-type`; then you indeed need to source the `query_predicates` (or just add it locally here; I think that would be simplest).

scripts/format-queries.lua
lucario387: `-l` doesn't actually write any file. I don't know exactly how to parse a raw string so I opted for `bufload` -> `bufadd` -> `write`

lucario387:
> Feel free to commit the formatted queries (in a separate commit) so we can see how it looks!

Done :smile:

runtime/queries/apex/highlights.scm
```diff
@@ -36,43 +32,46 @@
...
 ((field_access
   object: (identifier) @type))
 
 (field_access
   field: (identifier) @property)
 
-((scoped_identifier
-  scope: (identifier) @type)
- (#match? @type "^[A-Z]"))
+(
```
clason, line 66: Can we keep the "collapsed" opening parens (since we collapse the corresponding closing parens)?

runtime/queries/apex/locals.scm
```diff
@@ -1,41 +1,51 @@
...
 (enum_declaration) @local.scope
+
 (enhanced_for_statement) @local.scope
 
 ; if/else
-
 (if_statement) @local.scope
+
 (if_statement
-  consequence: (_) @local.scope) ; if body in case there are no braces
```
clason, line 13: I think we should leave inline comment alone, or at least move them _above_ the line

runtime/queries/arduino/highlights.scm
```diff
@@ -1,96 +1,24 @@
...
-           "highByte"
-           "lowByte"
-           ; External Interrupts
-           "attachInterrupt"
-           "detachInterrupt"
-           ; Interrupts
-           "interrupts"
-           "noInterrupts"
-           ))
+  (#any-of? @function.builtin ; Digital I/O
```
clason, line 3: OK; _that_ is an argument for allowing to ignore queries (or patterns)...

scripts/format-queries.lua
clason: Then of course you have to `vim.cmd.write()`; you just don't need the `quit()`.

clason:
I'm a bit conflicted about introducing so many newlines. I think some grouping is helpful?

But I don't feel strongly about it.

runtime/queries/arduino/highlights.scm
clason: If it's feasible, supporting `; format-ignore` for individual patterns would be best!

runtime/queries/apex/highlights.scm
lucario387: We can.

runtime/queries/arduino/highlights.scm
lucario387: > If it's feasible, supporting `; format-ignore` for individual patterns would be best!

That's easy. Adding a pattern to support that and I think it should be good to go

lucario387:
I am okay with not introducing newlines. But then there'll need to be some ways to separate query patterns from one another. Right now I cannot think of any pattern to only exclude some of it

clason:
> I am okay with not introducing newlines. But then there'll need to be some ways to separate query patterns from one another. Right now I cannot think of any pattern to only exclude some of it

Yeah, it's either "newline after every pattern" or "never add newlines" (except possibly around comments). Not sure which is preferable.

runtime/queries/scheme/highlights.scm
```diff
@@ -1,182 +1,76 @@
...
-
-(number) @number
-(character) @character
-(boolean) @boolean
-(string) @string
-[(comment)
- (block_comment)] @comment @spell
-
-;; highlight for datum comment
+(number) @number(character) @character(boolean) @boolean(string) @string[
```
clason, line 4: Here you're eating newlines, breaking the query (not sure why, as it doesn't affect other queries)?

runtime/queries/nix/injections.scm
```diff
@@ -1,153 +1,193 @@
...
 
-((comment) @injection.language
-  . ; this is to make sure only adjacent comments are accounted for the injections
-  [
-    (string_expression (string_fragment) @injection.content)
-    (indented_string_expression (string_fragment) @injection.content)
-  ]
-  (#gsub! @injection.language "/%*%s*([%w%p]+)%s*%*/" "%1")
-  (#set! injection.combined))
+((comment) @injection.language.
```
clason, line 4: Where does the trailing dot come from (again, breaking the query)? 🤔

lucario387: I ran the code with the non-updated query :facepalm: 
Will push the fix

runtime/queries/scheme/highlights.scm
lucario387: This is also because it's being run with an outdated version of the query.

runtime/queries/c/injections.scm
```diff
@@ -1,83 +1,61 @@
...
+  (#set! injection.self))
 
 ((comment) @injection.content
- (#set! injection.language "comment"))
+  (#set! injection.language "comment"))
 
 ((comment) @injection.content
-  (#match? @injection.content "/\\*!([a-zA-Z]+:)?re2c")
-  (#set! injection.language "re2c"))
+(#match? @injection.content "\\")
```
lucario387, line 8: Oops

runtime/queries/nix/injections.scm
clason: Looks good now!

runtime/queries/cue/highlights.scm
```diff
@@ -150,11 +152,15 @@
...
+  (_)
+  ")" @punctuation.special) @none
 
-(interpolation "\\(" @punctuation.special (_) ")" @punctuation.special) @none
-
-(interpolation "\\(" (identifier) @variable ")")
+(interpolation
+  "\\"
+  (identifier) @variable
+  ")")
```
clason, line 163: I think you're missing some escapes in the code...

runtime/queries/arduino/highlights.scm
lucario387: Added the pattern.

runtime/queries/cue/highlights.scm
lucario387: Re-added them.

scripts/format-queries.lua
```diff
@@ -1,12 +1,18 @@
...
-vim.cmd.runtime({ 'plugin/query_predicates.lua', bang = true })
+ts.query.add_predicate('has-type?', function(match, _, _, pred)
+  local node = match[pred[2]]
+  if not node then
+    return true
+  end
+
+  local types = { unpack(pred, 3) }
+  return vim.list_contains(types, node:type())
+end, true)
```
lucario387, line 15: I'm also using `not-has-parent`, to achieve the collapsing opening parentheses

clason: That one is already in Neovim core (in 0.10, which we're targeting here).

clason:
Very nice! Some small nitpicks around comment placement, but this is almost done and a massive improvement on the status quo 🚀

runtime/queries/agda/highlights.scm
```diff
@@ -62,20 +71,16 @@
...
   "{"
-  "}"]
-@punctuation.bracket
+  "}"
+] @punctuation.bracket
 
-[
-  "="
-] @operator
+[ "=" ] @operator
```
lucario387, line 86: Nitpick: Should I be removing the braces here, or keep it as this style. The implementation in nvim-ts-format does support removing 1-sized list like this.

lucario387: (Just finished washing the dishes. Going to sleep nwo)

clason: I'm fine with either; keeping makes it slightly easier to add, but removing looks better.

(I'm assuming this is specific to anonymous nodes?)

lucario387:
Just pushed some extra changes
```scheme
; Before
(foo
 field: (bar
   (baz))
(#pred?))

; After
(foo
  field:
    (bar
      (baz))
  (#pred?))
```

lucario387:
I've done a quick checkup to all query files in commit #2 @clason 

Will be nice if you can checkup that again

runtime/queries/agda/highlights.scm
lucario387: Going to keep it as it is right now. It's simpler and we can always look at this again at a later date.

runtime/queries/apex/locals.scm
```diff
@@ -1,41 +1,51 @@
...
+  alternative: (_) @local.scope)
 
 ; try/catch
+; covers try+catch, individual try and catch are covered by (block)
+(try_statement) @local.scope
 
-(try_statement) @local.scope ; covers try+catch, individual try and catch are covered by (block)
-(catch_clause) @local.scope ; needed because `Exception` variable
+; needed because `Exception` variable
+(catch_clause) @local.scope
```
clason, line 28: Something is still not working correctly with inline comments. They should either all move to the line above, or just be left alone (possibly with spacing adjusted and empty comments trimmed, if that is feasible). My preference is for the latter, as that seems easiest and most robust.

runtime/queries/agda/highlights.scm
```diff
@@ -1,38 +1,47 @@
-
 ;; Constants
 (integer) @number
 
 ;; Variables and Symbols
```
clason, line 4: Question: do we want to standardize single vs. double semicolon for comments? (I'd say yes, especially if we leave inline comments.)

lucario387: I don't know. Whichever you prefer will be what I follow.

runtime/queries/apex/locals.scm
lucario387: I'll try. Probably need a new predicate for that

runtime/queries/agda/highlights.scm
clason: Depends on how simple it is to implement ;)

runtime/queries/apex/locals.scm
clason: I'd be open to improving the query parser to be able to distinguish inline comments.

runtime/queries/agda/highlights.scm
lucario387: I see it as a `(#gsub! @capture "^;+(.*)" ";;%1")`, then modifying a bit to also apply metadata for get_node_text

clason: That sounds feasible. The two options are

1. Top-level comments get two `;;`; inline comments get (properly) moved and `;`.
2. All comments get one `;`;  inline comments stay where they are.

My (swayable) preference would be for option 2.

lucario387: gsub to the rescue (soon) then. I'll go with 2

clason: (We can always revisit later; the nice part about a full formatter is that it always yields the intended format, no matter the original state.)

lucario387: Well from a syntax tree POV, both nodes are adjacent and nothing else. Lines aren't really a thing for non-indent-aware languages. Still, I'll try to go for 2

lucario387: > Can't we distinguish with `#not-has-parent?`

We can. But as you said you're more in favor of 2., it won't be needed

lucario387: I celebrated too soon. `#gsub!` doesn't seem to be handling stuffs quite well. On the other hand, end of line comments are now preserved @clason

scripts/format-queries.lua
```diff
@@ -0,0 +1,286 @@
+-- #!/usr/bin/env -S nvim -l
```
clason, line 1:
```suggestion
#!/usr/bin/env -S nvim -l
```
Is it intentional that you keep reverting this change (and the `+x`)?

runtime/queries/apex/locals.scm
```diff
@@ -1,37 +1,48 @@
...
 
 (if_statement) @local.scope
+
 (if_statement
   consequence: (_) @local.scope) ; if body in case there are no braces
+
 (if_statement
   alternative: (_) @local.scope) ; else body in case there are no braces
-
 ; try/catch
```
clason, line 22: Hmm, this empty line should ideally stay

scripts/format-queries.lua
lucario387: No, I forgot to apply it

runtime/queries/markdown/highlights.scm
```diff
@@ -71,17 +104,17 @@
...
 
 [
   (block_continuation)
   (block_quote_marker)
 ] @punctuation.special
 
-[
-  (backslash_escape)
-] @string.escape
+[ (backslash_escape) ] @string.escape
```
clason, line 118: Your choice whether to keep or strip these brackets.

scripts/format-queries.lua
clason: Just wondering.

runtime/queries/markdown/highlights.scm
lucario387: I'm keeping it. Can always look back at it later anyways

runtime/queries/agda/highlights.scm
lucario387: It works now!

runtime/queries/go/locals.scm
```diff
@@ -1,79 +1,97 @@
...
-
-(var_spec 
+((function_declaration
+  name: (identifier) @local.definition.function)
+   ;@function 
+  )
+
+((method_declaration
+  name: (field_identifier) @local.definition.method)
+   ; @function.method
```
clason, line 8: Same?

runtime/queries/go/locals.scm
```diff
@@ -1,79 +1,97 @@
...
-)
-
-(short_var_declaration 
-  left: (expression_list
-          (identifier) @local.definition.var)) 
-
-(var_spec 
+((function_declaration
+  name: (identifier) @local.definition.function)
+   ;@function
```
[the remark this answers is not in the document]
lucario387: Nope, just me making a handling error :sob:

runtime/queries/apex/locals.scm
lucario387: It should be better now.

runtime/queries/swift/highlights.scm
```diff
@@ -58,60 +95,118 @@
...
 (call_expression
-  (prefix_expression (simple_identifier) @function.call)) ; .foo()
+  (prefix_expression
+    (simple_identifier) @function.call)) ; .foo()
+
 ((navigation_expression
-   (simple_identifier) @type) ; SomeType.method(): highlight SomeType as a type
-   (#lua-match? @type "^[A-Z]"))
+  (simple_identifier) @type)
+   ; SomeType.method(): highlight SomeType as a type
```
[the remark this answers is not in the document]
lucario387: I forgot to handle it. Should work now

runtime/queries/go/locals.scm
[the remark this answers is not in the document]
lucario387: Should be resolved now.

runtime/queries/apex/locals.scm
```diff
@@ -1,37 +1,48 @@
...
 
+(catch_clause) @local.scope ; needed because `Exception` variable
 ; loops
 
 (for_statement) @local.scope
-(for_statement         ; "for" body in case there are no braces
+
+; "for" body in case there are no braces
+
+(for_statement
```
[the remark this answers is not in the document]
lucario387: Resolved for now, I believe.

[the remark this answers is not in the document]
clason: Not quite; it's still moved (now below with no whitespace).

runtime/queries/clojure/highlights.scm
```diff
@@ -1,360 +1,373 @@
...
 
 ; Types
-;; TODO: improve?
+; TODO: improve?
 ((sym_lit) @type
- (#lua-match? @type "^[%u][^/]*$"))
-;; Symbols with `.` but not `/`
+  (#lua-match? @type "^[%u][^/]*$"))
+
+; Symbols with `.` but not `/`
```
clason, line 117: another spurious newline

runtime/queries/clojure/highlights.scm
```diff
@@ -1,360 +1,373 @@
...
 (list_lit
- .
- (sym_lit) @function.method
- (#lua-match? @function.method "^[%u].*/.+"))
-;; TODO: Special casing for the `.` macro
+  .
+  (sym_lit) @function.method
+  (#lua-match? @function.method "^[%u].*/.+"))
 
+; TODO: Special casing for the `.` macro
```
clason, line 141: "inline" comment moved out of the capture (not sure how to deal with this; maybe it's OK to prepare by hand (turn into inline comment)?

runtime/queries/apex/locals.scm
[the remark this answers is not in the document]
lucario387: I moved the comment to outside the query during commit no.2. I can move it back to its original position if it's needed

runtime/queries/clojure/highlights.scm
lucario387: I definitely moved them out of the capture before. Just need to clean the checkup for it.

runtime/queries/ispc/highlights.scm
```diff
@@ -37,189 +34,53 @@
...
...mic_xor_global" "atomic_xor_local" "avg_down" "avg_up" "broadcast" "ceil" "clamp" "clock" "cos" "count_leading_zeros" "count_trailing_zeros" "doublebits" "exclusive_scan_add" "exclusive_scan_and" "exclusive_scan_or" "exp" "extract" "fastmath" "float16bits" "floatbits" "float_to_half" "float_to_half_fast" "float_to_srgb8" "floor" "frandom" "frexp" "half_to_float" "half_to_float_fast" "insert" "intbits" "invoke_sycl" "isnan" "ISPCAlloc" "ISPCLaunch" "ISPCSync" "lanemask" "ldexp" "log" "max" "memcpy" "memcpy64" "memmove" "memmove64" "memory_barrier" "memset" "memset64" "min" "none" "num_cores" "or" "packed_load_active" "packed_store_active" "packed_store_active2" "packmask" "popcnt" "pow" "prefetch_l1" "prefetch_l2" "prefetch_l3" "prefetch_nt" "prefetchw_l1" "prefetchw_l2" "prefetchw_l3" "print" "random" "rcp" "rcp_fast" "rdrand" "reduce_add" "reduce_equal" "reduce_max" "reduce_min" "rotate" "round" "rsqrt" "rsqrt_fast" "saturating_add" "saturating_div" "saturating_mul" "saturating_sub" "seed_rng" "select" "shift" "shuffle" "signbits" "sign_extend" "sin" "sincos" "soa_to_aos2" "soa_to_aos3" "soa_to_aos4" "sqrt" "streaming_load" "streaming_load_uniform" "streaming_store" "tan" "trunc")
```
clason, line 85: may want to `; format-ignore`?

runtime/queries/jq/highlights.scm
```diff
@@ -1,258 +1,64 @@
...
...t" "getpath" "gmtime" "group_by" "gsub" "halt" "halt_error" "has" "hypot" "implode" "in" "index" "indices" "infinite" "input" "input_filename" "input_line_number" "inputs" "inside" "isempty" "isfinite" "isinfinite" "isnan" "isnormal" "iterables" "j0" "j1" "jn" "join" "keys" "keys_unsorted" "last" "ldexp" "leaf_paths" "length" "lgamma" "lgamma_r" "limit" "localtime" "log" "log10" "log1p" "log2" "logb" "ltrimstr" "map" "map_values" "match" "max" "max_by" "min" "min_by" "mktime" "modf" "modulemeta" "nan" "nearbyint" "nextafter" "nexttoward" "normals" "not" "now" "nth" "nulls" "numbers" "objects" "path" "paths" "pow" "pow10" "range" "recurse" "recurse_down" "remainder" "repeat" "reverse" "rindex" "rint" "round" "rtrimstr" "scalars" "scalars_or_empty" "scalb" "scalbln" "scan" "select" "setpath" "significand" "sin" "sinh" "sort" "sort_by" "split" "splits" "sqrt" "startswith" "stderr" "strflocaltime" "strftime" "strings" "strptime" "sub" "tan" "tanh" "test" "tgamma" "to_entries" "todate" "todateiso8601" "tojson" "tonumber" "tostream" "tostring" "transpose" "trunc" "truncate_stream" "type" "unique" "unique_by" "until" "utf8bytelength" "values" "walk" "while" "with_entries" "y0" "y1" "yn"))
```
clason, line 60: same?

runtime/queries/clojure/highlights.scm
clason: I checked the diff; it was certainly inline before this PR?

clason:
We now have trailing whitespace after colons.

runtime/queries/nix/injections.scm
```diff
@@ -1,155 +1,221 @@
 ((comment) @injection.content
- (#set! injection.language "comment"))
-
-((comment) @injection.language
-  . ; this is to make sure only adjacent comments are accounted for the injections
+  (#set! injection.language "comment"))
+((comment) @injection.language. ; this is to make sure only adjacent comments are accounted for the injections
```
clason, line 3: 😭

lucario387: :')

Well at least that's easy

runtime/queries/clojure/highlights.scm
lucario387: image
Not really, it was under a line

clason: Right, I must have confused this with the other moved query. Sorry about that

lucario387:
> We now have trailing whitespace after colons (e.g. ecma/injections.scm).

I resolved this for real this time :tm:

clason:
Still some left: elixir locals (format-ignored) and lua injections (commented out); maybe we can include those in the manual cleanup commit?

lucario387:
> Still some left: elixir locals (format-ignored) and lua injections (commented out); maybe we can include those in the manual cleanup commit?

I already noticed it and forced-push the fix. Should be good to go now.

scripts/format-queries.lua
```diff
@@ -0,0 +1,316 @@
...
+end, true)
+
+ts.query.add_predicate('is-start-of-line?', function(match, _, _, pred)
+  local node = match[pred[2]]
+  if not node then
+    return true
+  end
+  local start_row, start_col = node:start()
+  local foo = vim.fn.indent(start_row + 1) == start_col
+  return foo
```
clason, line 24:
```suggestion
  return vim.fn.indent(start_row + 1) == start_col
```

clason:
Ok, I think we're looking pretty! Could you squash your commits (1. manual pre-cleanup, 2. adding script, 3. running the formatter, 4. adding CI)?

Then we can let the others take a look at it (@theHamsta @ObserverOfTime @amaanq). I plan to merge this over the weekend and see how painful backporting will be.

lucario387:
I have squashed and re-order the commits as suggested. If there's any nitpicks or changes needed to be done, I will try to finish it in Saturday's night/Sunday's morning.

runtime/queries/apex/locals.scm
```diff
@@ -22,7 +22,8 @@
 ; loops
 
 (for_statement) @local.scope
-(for_statement         ; "for" body in case there are no braces
+; "for" body in case there are no braces
+(for_statement
```
clason: sorry, missed that: can you drop that change?

runtime/queries/agda/highlights.scm
```diff
@@ -62,20 +71,16 @@
...
-]
-@keyword
-
-;;;(expr 
-;;;	f_name: (atom) @function)
-;; Brackets
+] @keyword
 
+;(expr 
+;	f_name: (atom) @function)
```
ObserverOfTime, line 77:
```suggestion
; (expr f_name: (atom) @function)
```

runtime/queries/bash/highlights.scm
```diff
@@ -1,84 +1,88 @@
 [
- "(" ")"
- "{" "}"
- "[" "]"
- "[[" "]]"
- "((" "))"
+  "("
```
ObserverOfTime, line 2: Matching brackets should remain grouped (in other languages as well).

runtime/queries/bitbake/highlights.scm
```diff
@@ -99,150 +111,167 @@
...
-           ;; https://docs.python.org/3/library/constants.html
-           "NotImplemented"
-           "Ellipsis"
-           "quit"
-           "exit"
-           "copyright"
-           "credits"
-           "license"))
+  (#any-of? @constant.builtin ; https://docs.python.org/3/library/constants.html
+   "NotImplemented" "Ellipsis" "quit" "exit" "copyright" "credits" "license"))
```
ObserverOfTime, line 129:
```suggestion
  ; format-ignore
  (#any-of? @constant.builtin
   ; https://docs.python.org/3/library/constants.html
   "NotImplemented" "Ellipsis" "quit" "exit" "copyright" "credits" "license"))
```

runtime/queries/bitbake/highlights.scm
```diff
@@ -329,30 +362,44 @@
...
-              ;; https://docs.python.org/3/library/stdtypes.html
-              "bool" "int" "float" "complex" "list" "tuple" "range" "str"
-              "bytes" "bytearray" "memoryview" "set" "frozenset" "dict" "type" "object"))
+([
+  (identifier)
+  (python_identifier)
+] @type.builtin
+  ; format-ignore
+  (#any-of? @type.builtin
+      ;; https://docs.python.org/3/library/exceptions.html
```
ObserverOfTime, line 389:
```suggestion
      ; https://docs.python.org/3/library/exceptions.html
```

runtime/queries/bitbake/highlights.scm
```diff
@@ -329,30 +362,44 @@
...
+      "KeyboardInterrupt" "MemoryError" "NameError" "NotImplementedError" "OSError" "OverflowError" "RecursionError"
+      "ReferenceError" "RuntimeError" "StopIteration" "StopAsyncIteration" "SyntaxError" "IndentationError" "TabError"
+      "SystemError" "SystemExit" "TypeError" "UnboundLocalError" "UnicodeError" "UnicodeEncodeError" "UnicodeDecodeError"
+      "UnicodeTranslateError" "ValueError" "ZeroDivisionError" "EnvironmentError" "IOError" "WindowsError"
+      "BlockingIOError" "ChildProcessError" "ConnectionError" "BrokenPipeError" "ConnectionAbortedError"
+      "ConnectionRefusedError" "ConnectionResetError" "FileExistsError" "FileNotFoundError" "InterruptedError"
+      "IsADirectoryError" "NotADirectoryError" "PermissionError" "ProcessLookupError" "TimeoutError" "Warning"
+      "UserWarning" "DeprecationWarning" "PendingDeprecationWarning" "SyntaxWarning" "RuntimeWarning"
+      "FutureWarning" "ImportWarning" "UnicodeWarning" "BytesWarning" "ResourceWarning"
+      ;; https://docs.python.org/3/library/stdtypes.html
```
ObserverOfTime, line 401:
```suggestion
      ; https://docs.python.org/3/library/stdtypes.html
```

runtime/queries/bitbake/indents.scm
```diff
@@ -1,101 +1,144 @@
...
+(parenthesized_expression
+  ")" @indent.end)
+
+(generator_expression
+  ")" @indent.end)
+
+(list_comprehension
+  "]" @indent.end)
+
+(set_comprehension
```
ObserverOfTime, line 125: Can encapsulated literal captures like these remain in a single line?

runtime/queries/cairo/locals.scm
```diff
@@ -16,53 +16,51 @@
...
+    (scoped_identifier
+      name: (identifier) @local.definition.import))
 
-(use_as_clause 
+(use_as_clause
   alias: (identifier) @local.definition.import)
 
 (use_list
-    (identifier) @local.definition.import) ; use std::process::{Child, Command, Stdio};
+  (identifier) @local.definition.import) ; use std::process::{Child, Command, Stdio};
```
ObserverOfTime, line 33:
```clojure
 ; use std::process::{Child, Command, Stdio};
(use_list
  (identifier) @local.definition.import)
```
*(can't make this one a suggestion)*

runtime/queries/cairo/locals.scm
```diff
@@ -16,53 +16,51 @@
...
+  ] @local.definition.parameter)
 
 ; Types
-
 (struct_item
   name: (type_identifier) @local.definition.type)
 
-(constrained_type_parameter 
+(constrained_type_parameter
   left: (type_identifier) @local.definition.type) ; the P in  remove_file<P: AsRef<Path>>(path: P)
```
ObserverOfTime, line 56:
```suggestion
 ; the P in  remove_file<P: AsRef<Path>>(path: P)
(constrained_type_parameter
  left: (type_identifier) @local.definition.type)
```

runtime/queries/bibtex/folds.scm
```diff
@@ -1,3 +1 @@
-[
- (entry)
-] @fold
+[ (entry) ] @fold
```
ObserverOfTime, line 1: Can unnecessary arrays be stripped?

runtime/queries/cpon/highlights.scm
```diff
@@ -26,21 +27,27 @@
...
 
-[ "<" ">" ] @punctuation.bracket
+[
+  "<"
+  ">"
+] @punctuation.bracket
 
-(("\"" @string)
- (#set! conceal ""))
+(("\"" @string)(#set! conceal ""))
```
ObserverOfTime, line 50: What happened here? :thinking: 
```suggestion
(("\"" @string)
  (#set! conceal ""))
```

runtime/queries/agda/highlights.scm
clason: I don't think commented code is -- or can be -- formatted.

runtime/queries/bash/highlights.scm
clason: Disagree. This is not automatable.

runtime/queries/bibtex/folds.scm
clason: Can, but were decided not to. I'm open to either choice.

runtime/queries/cairo/locals.scm
clason: Inline comments must be conserved; manual changes should only be made in this PR if they are required for the auto-formatter to work properly.

runtime/queries/agda/highlights.scm
ObserverOfTime: Not automatically but we might as well format the ones we notice.

runtime/queries/cpon/highlights.scm
clason: Probably backslash escape tripping things up.

runtime/queries/bash/highlights.scm
ObserverOfTime: It doesn't have to be when `; format-ignore` exists.

runtime/queries/cairo/locals.scm
clason: See below re: inline comments.

runtime/queries/cpon/highlights.scm
lucario387: Just me didn't think there'll be a pattern of `(grouping (grouping) (node))`

runtime/queries/agda/highlights.scm
clason: Not in this PR; it's big enough as it is.

runtime/queries/bash/highlights.scm
clason: Yes, but these defeat the purpose. We should use them _very_ sparingly, if autoformatting produces inacceptable results.

runtime/queries/cpon/highlights.scm
clason: Is this something we can standardize as well? This seems like a (needlessly) unusual pattern.

runtime/queries/bitbake/indents.scm
lucario387: Resolved :+1:

lucario387:
The fixup is for inlining `(foo "bar")` pattern. Feel free to drop if you want.

runtime/queries/cpon/highlights.scm
lucario387: Resolved

lucario387:
Rebased some of the suggestions from @ObserverOfTime into commit no.1 (format-ignore comments stuffs)

The first 4 commits' contents are similar to before

The last 3's are for styling suggestions from @ObserverOfTime.

runtime/queries/apex/locals.scm
lucario387: Sorry. Fixed that again today

runtime/queries/bitbake/indents.scm
clason: I think we should be consistent here -- either _all_ "simple" patterns are collapsed or none. I don't think it makes a difference whether the node is anonymous are not, as far as formatting is concerned.

(I have a slight preference for keeping the current pattern so the nesting is consistent across _all_ patterns, no matter how simple, but don't feel strongly enough to fight about it.)

runtime/queries/ada/highlights.scm
```diff
@@ -135,41 +135,33 @@
...
 
-(with_clause
-  "limited" @keyword.import)
+(with_clause "limited" @keyword.import)
 
 (use_clause
   (_) @module)
 
 (with_clause
   (_) @module)
```
clason, line 146: Here you can see why special-casing anonymous nodes makes no sense.

runtime/queries/bitbake/indents.scm
lucario387: I only added back the inlining for anonymous captures without preds/directives, as they cannot be deeper than 1 layer.

For named nodes, lists, groupings, fields and such, they are all kept as before the changes.

Also I do believe it can be disruptive so it's currently put into a separate commit for easier review.

runtime/queries/ada/highlights.scm
lucario387: Rn, There's no reasonable query pattern to check for nodes without any child. I don't want to create a new predicate just for this use case

runtime/queries/bitbake/indents.scm
clason: > I only added back the inlining for anonymous captures without preds/directives/anchors/quantifiers, as they cannot be deeper than 1 layer.

I understand that those "simple" captures (without etc.) look nicer collapsed; I'm just saying if you do it for anonymous nodes, you should also do it for named nodes -- i.e., no special case for anonymous nodes.

runtime/queries/ada/highlights.scm
clason: Then I'm strongly in favor of leaving things as before.

clason: Although I don't quite see why it's impossible? Can't you just do `[ (anonymous_node) (named_node)]`?

runtime/queries/bitbake/indents.scm
lucario387: It's just a bit too hard to cover for multiple depths, and another is "what is considered as a simple capture". I can only reliably support anonymous nodes as they are defined depths

clason: Then I am against this change, I'm afraid.

lucario387: Some issues related to this:
anchors (`.`) only considers named nodes, and doesn't support `(foo . .)`. So there's zero way to declare if there's no child here. I don't think having a multi-depth queries will be nice

runtime/queries/ada/highlights.scm
lucario387: named nodes can have child.

You will need a lot more queries than just that.

Eg:
```
(foo (bar (baz (abc (xyz)))))
```

While anonymous nodes cannot have any more nested layers

lucario387: Also need to ignore other patterns that contains the named nodes :)
I'll try, but will only support 1-depth named nodes, no more than that

clason: >  will only support 1-depth named nodes

Oh yes, sorry in case that wasn't clear. I _only_ want
```
(with_clause (_) @module)
```
(or other named nodes), nothing else. But if it requires complications, then I'd argue for leaving named _and_ anonymous nodes alone.

runtime/queries/bitbake/highlights.scm
ObserverOfTime: These changes should also be applied to python and starlark.

runtime/queries/ada/highlights.scm
lucario387: Added support for 

```
(foo (_) @capture)
```

This still stays
```
(foo
  (bar) @capture)
```

clason: Sorry, it's all or nothing for me. (Though I appreciate the effort.)

lucario387: > Sorry, it's all or nothing for me; that just makes it worse. Formatting should be based solely on the syntax, not the contents. (Though I appreciate the effort.)

Well now it's both then :)

lucario387: Only 1-depth. If a pattern is

```
(foo
  (bar
    (baz)))
```
it should stay as is

runtime/queries/cpp/highlights.scm
```diff
@@ -15,14 +15,11 @@
   declarator: (_) @variable.parameter)
 
 ;(field_expression) @variable.parameter ;; How to highlight this?
-(((field_expression
-  (field_identifier) @function.method))@_parent(#has-parent? @_parent template_method function_declarator))
+(((field_expression (field_identifier) @function.method))@_parent(#has-parent? @_parent template_method function_declarator))
```
clason, line 18: Here's another instance where the predicate gets smushed together with the captures (not new in this commit, just noticed).

runtime/queries/ada/highlights.scm
clason: I can live with this (although I still prefer the original, nicely nested, format). @ObserverOfTime is this an acceptable compromise?

runtime/queries/cpp/highlights.scm
clason: There's a number of `)@` patterns in the queries introduced by the formatter?

runtime/queries/bitbake/highlights.scm
lucario387: Done.

runtime/queries/cpp/highlights.scm
lucario387: I just didn't expect people to do captures after a grouping :bow: 

fyi, `((foo) (bar)) @capture` is equivalent to `(foo) @capture` so I'm unsure why people write it like that

lucario387: Fixed with latest force push

clason: People have no clue and just spam parens until it works. (At least, that's what I do.)

Is this something we can normalize in the formatter (always replace the former by the latter)? (Not worth complicating the script significantly, but would be nice to have if it's simple).

lucario387: It's too much. The formatter will also need to do refactoring work.

Formatters should not be allowed to move nodes around that change how the syntax tree works like that tbh, that's a linters/others jobs

lucario387: (With that said). It can help remove groupings (paren spamming). But I will rather look at it later with list `[]` removals.

clason: That makes sense, yes.

lucario387:
I'm not squashing the new commits just yet :thinking:. This is easier for others to view.

runtime/queries/bibtex/folds.scm
clason: Looking at the Apex highlights, I think I'm now (slightly) in favor of stripping arrays. @lucario387 what do you think?

runtime/queries/ada/highlights.scm
ObserverOfTime: I'm not too fond of it…

runtime/queries/bibtex/folds.scm
lucario387: I can remove it if you want :)
But I won't be removing parentheses for grouping just yet. That's just another realm of stuffs (AST manip)

runtime/queries/ada/highlights.scm
clason: Me neither, but we have to compromise. Otherwise I'll invoke executive privilege.

ObserverOfTime: Let's just go back to the nested format.

runtime/queries/bibtex/folds.scm
clason: No, I'm just talking about removing arrays around single nodes, as we talked about a few days ago. I agree with you that AST manipulation is out of scope here.

runtime/queries/ada/highlights.scm
clason: None of this is set in stone anyway. As long as the transformation is consistent and not relying on the input format (which is what I really care about), we can always change the format later.

runtime/queries/bibtex/folds.scm
lucario387: Do note that removing `[]` here alters the sytna tree. So running twice might give different results

clason: Hmm, that's an argument. But to me consistency is more important; having to run the script twice is an acceptable price.

lucario387: @clason Done. :) 33148b6f23e057408fe25c6c55c3b83edcc59ed2

runtime/queries/ada/highlights.scm
lucario387: So which pattern should I be going for now :thinking:

clason: Your call; I can live with both and @ObserverOfTime can live with neither... (Or wait for @theHamsta to have yet a third opinion.)

lucario387: I'm gonna wait then. Both styles are fine to me

runtime/queries/cpp/highlights.scm
lucario387: Still not going to look at grouping stuffs. I cannot truly confirm the behavior for this.

runtime/queries/ada/highlights.scm
clason: Other considerations:

1. simpler implementation/faster formatting
2. smaller diff

lucario387: 1. leads to bigger diffs

2. bit more implementations. But right now it's gone from 4->5 second for all files on GitHub actions, but still not too slow i guess

clason: Yeah, I don't think speed is relevant here (this was more a general comment). But I thought the diff against `HEAD` was _larger_ yesterday?

lucario387: Well yes, 2k lines larger. 

1k line was removed from this inlining. Another 1k line is from only `elixir/locals.scm`

clason: > Another 1k line is from only elixir/locals.scm

Which is irredeemable.

lucario387: Also why the file is blasted with `; format-ignore` cause that's the most I can care about it.

clason: Clearly the right move, and the reason we have that ("this is a dumpster fire; do not touch")

ObserverOfTime: > I can live with both and @ObserverOfTime can live with neither...

I can live with one node (of any kind) per line. At least it's consistent.

lucario387: I don't know of the mess I've created so I won't spend some time rebasing it, just gonna push :sob:

lucario387:
Oops it's 2k lines more ...

runtime/queries/ada/highlights.scm
lucario387: Welp it's back now :)
If you want to bring back the inlining, just remove the `fixup: removal ...` and the commit following that

clason:
That's what I thought :)

lucario387:
Well I think that's about it for now. If it's fine then I'll squash the commits after

clason:
Fine with me! (Can you document in a comment the "compact" query that we had at the end for possible later or personal use?)

lucario387:
> Can you document in a comment the "compact" query that we had at the end for possible later or personal use

Do you mean comments for the query, or for the `iter()` func

clason:
Ah, if the changes were more extensive than just adding a pattern or two in the query, then that's not a good idea.

clason:
Other idea for a compromise: 
1. if the full pattern fits in a line (`textwidth` or some fraction thereof), keep it in one line
2. otherwise expand it (fully) as now

Would that be doable and acceptable?

lucario387:
Yes. I do have a partially working version of that inside nvim-ts-format. But it's partial and require walking the tree upwards. So keeping it simple for now.

clason:
> Acceptable or not? Probably only for predicates and lists. Otherwise not

Again, it's "all or nothing". I suspect that -- with a reasonable line width threshold -- this will only really apply to patterns where we want to have a compact format.

Anyway, let's keep it in mind for the future then. As I said, the exact format is not so important as long as we agree to have them consistently autoformatted at all -- then it's trivial to reformat them anytime we come up with a better implementable format.

lucario387:
> Again, it's "all or nothing". I suspect that -- with a reasonable line width threshold -- this will only really apply to patterns where we want to have a compact format.

That sounds like another pattern to be added. Backlog areas then

clason:
I think you can squash down to the separate commits already.

lucario387:
Latest commit fixes the 
```
((pat) @capture . ;comment
 )
``` issue

```
; Before
((pat) @capture. ;comment)
; After
((pat) @capture
  . ; comment)
```

**Comment by clason** [comment]
Shouldn't that be 
```
((pat) @capture
  . ; comment
)
```
?

**Comment by lucario387** [comment]
The "backlog" will be in nvim-ts-format first, just like this version here :)

And unless I can reproduce a reliable and less choring maintenance queries, I won't try to create one for nvim-treesitter just yet

> Shouldn't that be
> 
> ```
> ((pat) @capture
>   . ; comment
> )
> ```
> 
> ?

Yes. Typo'd

**Comment by clason** [comment]
> And unless I can reproduce a reliable and less choring maintenance queries, I won't try to create one for nvim-treesitter just yet

Oh, sure, no hurry! Just the fact that we _have_ a formatter (in CI) now is already a great help with maintenance.

**Comment by lucario387** [comment]
Here's to hope nothing breaks

**Comment by lucario387** [comment]
Latest force-push is a fixup to make sure inline comments are respected for real this time :sob: 

```
; Orig
(foo
  !field . ;comment
)

; Before
(foo
  !field. ;comment
  )

; After
(foo
  !field
  . ;comment
  )
```

**Comment by ObserverOfTime** [comment]
Would this be possible?
```clojure
(foo
  !field
  .) ;comment
  ```

**Comment by lucario387** [comment]
I don't know how to re order nodes just yet. 🤔

File: runtime/queries/dart/indents.scm [hunk 40]
```diff
@@ -12,7 +13,6 @@
   (arguments)
   (try_statement)
 ] @indent.begin
-
```
clason, line 15: Missing empty line?

clason: This seems to be gone in _every_ `indents.scm` after the `@indent.begin`...

lucario387: That doesn't make sense :thinking: Guess I messed up somewhere

lucario387: Well guess I was dumb :)
```
(program
  . (_)
  (_) @_comment
  .
  [
    (list)
    (grouping)
    (named_node)
    (anonymous_node)
  ] @format.prepend-newline
  (#not-has-type? @_comment comment))
```

This completely ignores the second node, all the time. Which happens to be what this pattern matches

lucario387: It's okay again now :). Sorry for the mistake

lucario387: https://github.com/nvim-treesitter/nvim-treesitter/compare/cfcc62d53a7c246493c3dd1fa80f22ed43b1e120..aad089b3adc0c481ffd3bc89148b850e83a05a87

This diff shows that all files were missing that separation of the first pattern. Got lost in the sauce

clason: Yup. Thanks for the fast fix!

clason:
I think this looks very nice now! Should we mention this (in particular the `; format-ignore`, and what exactly it applies to) in `CONTRIBUTING.md`?

lucario387:
I don't know how to write the opening. Will write the parts related to `; format-ignore` and let you handle the rest.

clason:
Ok, will add a commit with a suggestion, and you can fix all my factual errors :)

CONTRIBUTING.md
```diff
@@ -103,6 +88,15 @@ line of your file_.
...
 you can mark the language as optional (by putting it between parenthesis).
 
+#### Formating
+
+All queries are expected to follow a standard format, with every node on a single line and indented by two spaces for each level of nesting. You can automatically format the bundled queries by running the provided formatter `./scripts/format-queries.lua` from the repo root.
+
+Should you need to preserve a specific format for a node, you can exempt it (and all contained nodes) by placing before it
+```query
+; format-ignore
+```
```
lucario387, line 98: I don't know what needs to be fact-checked here :clinking_glasses:. Cannot be anymore correct than this

runtime/queries/query/highlights.scm
```diff
@@ -63,6 +63,9 @@
   (comment) @keyword.directive)
   (#lua-match? @keyword.directive "^;+ *extends *$"))
 
+((comment) @keyword.directive
+  (#eq? @keyword.directive "; format-ignore"))
```
lucario387, line 67: Maybe more lenient?
```suggestion
  (#lua-match? @keyword.directive "^;%s*format-ignore"))
```

clason: But maybe more expensive? Now that we can rely on the format ;))

lucario387: The formatter right now is only checking if you have `format-ignore` in that comment or not :thinking: 
I think I will fixup the pattern inside the formatter to match this. It should be lenient to avoid unexpected results.

clason:
Last suggestion: I think it would be useful to be able to format specific files/directories? (With `runtime/queries` being default, why not.)

lucario387:
> Last suggestion: I think it would be useful to be able to format specific files/directories? (With `runtime/queries` being default, why not.)

Yes, I did mention that before in Matrix chatbox but I think I forgot to implement it.

What should be the args then ?

clason:
> What should be the args then ?

That is a good question. Ideally either files or directories, but that may not be worth the complication? I'd say start with either a single file or a single directory, for simplicity. 

If we take a directory, we can avoid hardcoding the query location in the script and instead move it to CI, which makes it easier to backport.

lucario387:
I'm thinking of something like this maybe

```lua
-- Format a file/dir 
-- Usage:
-- nvim -l scripts/format-queries.lua /path/to/file # Single file formatting
-- nvim -l scripts/format-queries.lua --glob runtime/queries/**/*.scm
-- nvim -l scripts/format-queries.lua --dir /path/to/dir # Will check for all `*.scm` files undir this dir
```

clason:
I was thinking just to check the argument whether it ends in `.scm` (single file) or not (directory)? I don't think we need to support globs.

clason:
@lucario387 if my approach looks fine to you, you can squash the commits again (may need a bit of un- and restaging to parcel up the changes correctly -- let me know if I should do it), then I'll merge!

lucario387:
I think that's it :thinking:

lucario387:
Actually, I will force pushed again to make the manual adding of `format-ignore` to be after the formatter. Doesn't make sense to have it before the formatter.

clason:
Excellent work, @lucario387! Thank you so much for your effort and patience, that will be a huge boon for maintaining!

PR #5855, opened 2024-01-04
Highlight HTML character references (a.k.a. entities)

cpmsmith:
HTML entities, e.g. `&nbsp;`, are detected by the HTML parser, but not currently highlighted. It's not totally clear to me what highlight group they should get – I can see an argument for `@string.escape`, `@escape`, `@character.special`, `@constant`, the list goes on – and I'm open to input on that, but they should get _something_.

### DIFF
--- a/queries/html/highlights.scm
+++ b/queries/html/highlights.scm
@@ -3,3 +3,5 @@
 (doctype) @constant
 
 "<!" @tag.delimiter
+
+(entity) @character.special


cpmsmith:
Hmm, I guess I can't query for a node in `html` highlights unless all the other highlights which extend `html` (`astro`, `svelte`, `vue`) support the same node? That's a bit of a nuisance.

lucario387:
You can, just don't put it inside `html_tags`

clason:
And `@character.special` sounds best to me (it's what's used in Markdown).

Also, be aware that the `master` branch is currently frozen; we won't merge PRs until the big (breaking) capture renaming PR has been backported.

cpmsmith:
> You can, just don't put it inside `html_tags`

Whoops, I definitely meant to put it in `html`. I'll move it, though I think the issue will still appear on `astro`, as [it extends `html`, unlike the other two](https://github.com/cpmsmith/nvim-treesitter/blob/0da349ed303bea955942f409d29059cdb89dbe2c/queries/astro/highlights.scm#L1), and apparently [hasn't updated its dependency on `tree-sitter-html` to 0.20.0](https://github.com/virchau13/tree-sitter-astro/blob/e122a8fcd07e808a7b873bfadc2667834067daf1/package-lock.json#L57), which is when HTML entity support was added. I'll open another PR over there, I guess.

> And `@character.special` sounds best to me (it's what's used in Markdown).

Will do 👍 

> Also, be aware that the `master` branch is currently frozen; we won't merge PRs until the big (breaking) capture renaming PR has been backported.

Right – should I make this PR against `main` instead?

clason:
I think just having a bit of patience is the best course of actions. 

I am also leaning toward declaring web template languages out of scope here (beyond the simple basics) and punting to filetype plugins that can set up the specific combinations in a project specific way.

cpmsmith:
Since this is blocked on virchau13/tree-sitter-astro#23, I'll convert to draft for now, and re-ready it when it's unblocked.

lucario387 (approved):

clason:
Astro parser is bumped on `master`, so just rebase. I think this PR doesn't conflict with the refactors, so we can merge it to `master`.

cpmsmith:
Rebased ✅

PR #5858, opened 2024-01-05
fix(latex): only highlight inner env text as math

ribru17:
Removes `@text.math` highlights from the actual math environment boundaries, preventing highlight overwriting.

Before:
beforemathenv

After:
aftermathenv

### DIFF
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -111,17 +111,7 @@
 ] @text.math
 
 (math_environment
-  (begin
-   command: _ @text.math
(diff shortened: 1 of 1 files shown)


lucario387:
Making treesitter latex reaches 10% of vimtex's highlighting, one step at a time :confetti_ball:

ribru17:
Perfect, I was fiddling around with the query ordering and I think I found a way to make the highlights more robust/better while being less specific/hacky. I'll keep working on it and update after the backport like you say

PR #5871, opened 2024-01-07
fix(php): adjust to parser refactor

clason:


lucario387:
So do we need to inject html for php anymore? Seems like not

clason:
According to https://github.com/tree-sitter/tree-sitter-php/pull/192, nothing has changed with the `php` parser; it was just moved?

lucario387:
`php_only` is the old one, which parses only php. `php` is the updated one which also parsers HTML directly, not needing injections.

clason:
From the diff, it looks otherwise:

> src/parser.c → php/src/parser.c

Actually, I don't see any differences between the two grammars -- maybe this is just the separation, with no deviation yet?

Let me wait for CI to conclude, then try `php_only`. If they're indeed the same, let's stick with `php_only` for now and let a language maintainer decide (later) what to do.

lucario387:
I'm wrong :)
`php` is the old one, `php_only` is the new one that doesn't support any html-related taggings inside php

clason:
Related note: we really need better parser compilation feedback in CI; the step always shows as "succeeded" even if it fails.

lucario387 (approved):

PR #5884, opened 2024-01-10
fix: improve highlight consistency

ObserverOfTime:
The first commit is technically unrelated but I didn't want to separate it.

~~I didn't manage to run the tests locally so I'll update them if they fail.~~ Done.

### DIFF
--- a/runtime/queries/css/highlights.scm
+++ b/runtime/queries/css/highlights.scm
@@ -5,19 +5,32 @@
   "@supports"
   "@keyframes"
   (at_keyword)
+] @keyword.directive
+
+"@import" @keyword.import
+
+[
   (to)
   (from)
 ] @keyword
 
-"@import" @keyword.import
-
 (comment) @comment @spell
 
+(tag_name) @tag
+
+(class_name) @type
+
+(id_name) @constant
+
 [
-  (tag_name)
+  (property_name)
+  (feature_name)
+] @property
+
+[
   (nesting_selector)
   (universal_selector)
-] @type
+] @character.special
 
 (function_name) @function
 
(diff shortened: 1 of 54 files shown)


clason:
@TravonteD could you look at the CSS changes?

runtime/queries/css/highlights.scm
```diff
@@ -50,37 +63,30 @@
 
 (pseudo_element_selector
   "::"
-  (tag_name) @property)
+  (tag_name) @attribute)
```
lucario387, line 66: Shouldn't this also be `@tag`?

runtime/queries/ruby/highlights.scm
```diff
@@ -122,7 +122,7 @@
 [
   (class_variable)
   (instance_variable)
-] @label
+] @variable.member
 
 ((identifier) @constant.builtin
   (#match? @constant.builtin "^__(callee|dir|id|method|send|ENCODING|FILE|LINE)__$"))
```
lucario387, line 128: Can you also change this to `any-of` as well? :thinking:

runtime/queries/snakemake/highlights.scm
```diff
@@ -18,7 +18,7 @@
     (_
       (directive
         name:
-          _ @label)))
+          _ @property)))
```
lucario387, line 21: This looks bad :thinking: 
I think I f'd up something in the formatter. This kind of pattern shouldn't happen

Will fix later

runtime/queries/ruby/highlights.scm
lucario387: Will find a way to do it :+1:

runtime/queries/snakemake/highlights.scm
lucario387: #5885

clason:
@ObserverOfTime I rebased your PR on the rebased `main`.

ObserverOfTime:
Oh, whoops. :)

clason:
Well, then you get to rebase yourself ;)

runtime/queries/css/highlights.scm
lucario387: Agreed

lucario387:
I think `pug` should also be updated too.

ObserverOfTime:
It was. :thinking:

clason:
Can we improve the guidance in `CONTRIBUTING.md` as well?

ObserverOfTime:
If you have something to suggest.

clason:
I was hoping you would...

lucario387:
I think it's okay 🤔

clason:
Merging this now to allow preparing the Great Backport (which will include this, and any other query improvements that are safe for Neovim 0.9).

PR #5886, opened 2024-01-10
fix(config)!: always install parsers bundled with nvim

clason:
Problem: Nvim-treesitter ships with queries that may be incompatible with parsers bundled by Neovim and used by default (in particular `vimdoc`, whence the recent flood of issues).

Solution: Force installation of bundled parsers that may be used by Neovim.

This is a bandaid fix to tide us over until 1.0, which addresses this issue more cleanly.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -20,7 +20,7 @@ local M = {}
 local config = {
   modules = {},
   sync_install = false,
-  ensure_installed = {},
+  ensure_installed = { "lua", "query", "vimdoc", "vim", "c", "python", "bash", "markdown", "markdown_inline" },
   auto_install = false,
   ignore_install = {},
   parser_install_dir = nil,
@@ -421,6 +421,7 @@ function M.setup(user_data)
   end
 
   local ensure_installed = user_data.ensure_installed or {}
+  vim.list_extend(ensure_installed, config.ensure_installed)
   if #ensure_installed > 0 then
     if user_data.sync_install then
       require("nvim-treesitter.install").ensure_installed_sync(ensure_installed)


lua/nvim-treesitter/configs.lua
```diff
@@ -20,7 +20,7 @@ local M = {}
 local config = {
   modules = {},
   sync_install = false,
-  ensure_installed = {},
+  ensure_installed = { "lua", "query", "vim", "vimdoc", "python", "bash", "markdown", "markdown_inline" },
```
clason, line 23: * Lua, query, and vimdoc are used for highlighting on `master` by default (via `vim.treesitter.start()` in a ftplugin)
* Python and bash are pulled in via injections in help files
* Markdown is used for LSP hover on `master`.

lucario387 (approved):

lua/nvim-treesitter/configs.lua
wookayin:
```suggestion
  ensure_installed = { "c", "lua", "query", "vim", "vimdoc", "python", "bash", "markdown", "markdown_inline" }
```

Is this list supposed to be the same as the list of all "bundled" parser?

nvim 0.9.4 has: c, lua, query, vim, vimdoc
nvim 0.10 has: Plus help, bash, markdown, markdown_line, python

If so, `c` might also need to be added?

clason: * `vimdoc` _is_ `help`
* good point about `c`; it's supposed to be the list of all languages for which TS is enabled by default, but then I realized we also need to account for injections (which I forgot `c` for).

dimbleby:
this is broken if `ensure_installed` is set to the string "all", isn't it?

clason:
Is it?

dimbleby:
```
vim/shared.lua:0: dst: expected table, got string
```

clason:
How about now?

dimbleby:
lgtm, thanks!

Liquidmantis:
I'm still getting the table extend error when `ensure_installed` is `all`.  I commented it out for now.

Liquidmantis:
Can #2293 be repinned?

clason:
Once we have unpinned one of the other issues (can only have three).

And, please, "I have this issue still" doesn't help; we need details to know which edge case is still left.

mehalter:
Could anyone shine light to why this ended up getting reverted? Was there a pitfall of some kind that was identified?

clason:
No, just enough people complained that this broke their snowflake setup. Since this problem is already solved (differently) on main, I didn't want to waste more time on it.

mehalter:
Ok, thanks for the insight! Appreciate all of the work going on here by you and the `nvim-treesitter` team! Looking forward to the v1 release in the future 😄

PR #5889, opened 2024-01-12
Fix(julia): support broadcasting in piped function call

degnbol:
When piping into a function call there was no capture when broadcasting, i.e. `println` is captured as `@function.call` in `x |> println` but not in `x .|> println`. This commit makes it captured in both cases.

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -51,7 +51,7 @@
   (_)
   (operator) @_pipe
   (identifier) @function.call
-  (#eq? @_pipe "|>"))
+  (#lua-match? @_pipe "\.?|>"))
 
 ;; Builtins
 


queries/julia/highlights.scm
```diff
@@ -51,7 +51,7 @@
   (_)
   (operator) @_pipe
   (identifier) @function.call
-  (#eq? @_pipe "|>"))
+  (#lua-match? @_pipe "\.?|>"))
```
lucario387, line 54:
```suggestion
  (#any-of? @_pipe "|>" ".|>"))
```

clason:
@savq

clason:
ping. And please rebase on master.

clason:
closing as stale.

PR #5904, opened 2024-01-15
Add :TSEnsure command

HiPhish:
I have a testing setup where an embedded headless Neovim (`nvim --embed --headless`) runs inside an isolated environment (by setting the XDG environment variables). Aside from the plugin under test there is also nvim-treesitter in this environment because I need it to download parsers. But I want to download each parser at most once, synchronously, without need for confirmation.

None of the existing commands meet all of these criteria, so I have added the commands `:TSEnsure` (asynchronous) and `:TSEnsureSync` (synchronous). I use the synchronous command in my tests.

### DIFF
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -347,6 +347,17 @@ Perform the |:TSInstall| operation synchronously.
 
 List information about currently installed parsers
 
+								   *:TSEnsure*
+:TSEnsure {language} ...~
+
+Install {language} like with |:TSInstall|, but without need for confirmation.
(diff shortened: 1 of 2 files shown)


clason:
Sorry, that is too niche. We are actively reducing complexity. (And the main branch is frozen in any case.)

HiPhish:
Do you have an alternative suggestion? I could write code of my own code to interface with nvim-treesitter, but I don't know which parts are "public API".

clason:
I would suggest waiting for the main branch to become default and then look at the exported API.

PR #5908, opened 2024-01-17
fix: allow merge config with 'all'

ofseed:
There's a special case, if users set `ensure_installed` to `"all"` in their config, and every other call of `setup` will try to merge a table with a string.

### DIFF
--- a/lua/nvim-treesitter/configs.lua
+++ b/lua/nvim-treesitter/configs.lua
@@ -421,6 +421,9 @@ function M.setup(user_data)
   end
 
   local ensure_installed = user_data.ensure_installed or {}
+  if type(config.ensure_installed) ~= "table" then
+    ensure_installed = config.ensure_installed
+  end
   if type(ensure_installed) == "table" then
     vim.list_extend(ensure_installed, config.ensure_installed)
   end


lua/nvim-treesitter/configs.lua
```diff
@@ -421,6 +421,9 @@ function M.setup(user_data)
   end
 
   local ensure_installed = user_data.ensure_installed or {}
+  if type(config.ensure_installed) ~= "table" then
+    ensure_installed = config.ensure_installed
+  end
```
lucario387, line 426: Does this ever run? When `config.ensure_installed` is `{"lua", "vim", "vimdoc", ...}`

lua/nvim-treesitter/configs.lua
```diff
@@ -421,6 +421,9 @@ function M.setup(user_data)
   end
 
   local ensure_installed = user_data.ensure_installed or {}
```
lucario387, line 423: I think the fix should be
```lua
local ensure_installed = user_data.ensure_installed
ensure_installed = ensure_installed and type(ensure_installed) == "table" and ensure_installed or { ensure_installed }
```

lua/nvim-treesitter/configs.lua
ofseed: This is intentionally only allowed to run when the value is a string, when `config.ensure_installed` is a table, continue to execute, and then it may merge with the new config, aka `user_data.ensure_installed`

Run this only when `config.ensure_installed` is a string, that is "all".

lua/nvim-treesitter/configs.lua
ofseed: I am unsure whether this plugin could handle "all" like a normal language, because `ensure_installed = { "all", "c", "lua", }` is meaningless. but I think it can handle `ensure_installed = "all"` at least.

lua/nvim-treesitter/configs.lua
lucario387: but `config.ensure_installed` is always `{ "vim", "lua", "luadoc", ... }`

ofseed: No, the definition above is just the initial value, it's current `config.ensure_installed` value which is modified by users in https://github.com/nvim-treesitter/nvim-treesitter/blob/97ba59c6f532072e456b3d1f2413560e689c132b/lua/nvim-treesitter/configs.lua#L436

a simple proof by nvim-dap:

image

lua/nvim-treesitter/configs.lua
clason: `"all"` (which is really not recommended for users, and will be heavily discouraged in 1.0) already includes the necessary parsers and should be left alone. Similarly, I don't want to support, e.g., `ensure_installed = "lua"`. So the logic should be "if (and only if) `ensure_installed` is a table, make sure it contains the mandatory parsers".

If I missed a case in the control flow, (only) this should be fixed.

clason:
Can you please comment on the exact situation when this fix is needed (i.e., when the extended table is overwritten)?

ofseed:
When I call `require"nvim-treesitter".setup{ ensure_installed = "all" }` somewhere in my config,
the `config.ensure_installed` is `"all"`, as you can see from the screenshot above.

In this case, if I call `setup` again anywhere with a table without `ensure_installed`, due to
https://github.com/nvim-treesitter/nvim-treesitter/blob/94cb036ba7bd2b4086c04c945da5e983b5e8b322/lua/nvim-treesitter/configs.lua#L423
`ensure_installed` will be an empty table, and because it is still a table
https://github.com/nvim-treesitter/nvim-treesitter/blob/94cb036ba7bd2b4086c04c945da5e983b5e8b322/lua/nvim-treesitter/configs.lua#L425
will be executed, and the error occurs, because it tries to merge a table with a string.

ObserverOfTime:
Why would you call setup twice? :thinking:

ObserverOfTime:
Although, #5910 seems to be running into the same issue.

lua/nvim-treesitter/configs.lua
clason: But is it a _reasonable_ control flow?

ofseed:
For plugins that rely on the existing module system, the `setup` must be called. If you don't want to put their settings in the only file, you need to call it multiple times.

What you say makes me realize that it may not have been intentionally designed to be used this way, but it worked fine until today.

clason:
> If you don't want to put their settings in the only file, you need to call it multiple times.

No, I don't think so; you can work with guards, for example. And wanting to set up modules in different files is not a compelling use case for us to support with additional code.

(And this is just another reason that the module system has to be removed, which it will be in 1.0).

> What you say makes me realize that it may not have been intentionally designed to be used this way, but it worked fine until today.

In which case there was never a guarantee that it would. Nevertheless, a silent error is not good, and I'll try to think of something. (Probably an explicit error if setup is called twice.)

ofseed:
>(And this is just another reason that the module system has to be removed, which it will be in 1.0).

Agree with it. The above is what this PR can fix. If you feel it is not needed, please feel free to close this PR.

clason:
I any case, I appreciate your pointing this out; this was useful information (and helped exposing plugins with bad design like go.nvim).

clason:
I pushed a simpler change that should fix the issue for you.

ofseed:
> (and helped exposing plugins with bad design like go.nvim)

🤣It exactly is...

Thanks.

PR #5924, opened 2024-01-19
fix(comment): wrong todo highlight

ofseed:
The backported `master` branch includes the `@comment.todo` highlight, but the `main` branch does not, I'm wondering if it is a mistake.


ofseed:
Found it before in the previous PR to `main`, and just want to continue my contribution to colorscheme support now. Always thank you for your great work!

PR #5925, opened 2024-01-19
Removed pre installed parsers 

jpx40:
Hey, 
i removed changes introduced in https://github.com/nvim-treesitter/nvim-treesitter/commit/fe89bf17197fc5daead3878fb1328911b30fb022.

Reason:
It breaks nvim-treesitter with manually added parsers. I am using LazyVim as basis for my Config LazyVim adds most Parsers already, this commits breaks the manaual adding of Parsers. This is a little difficult if you using an distro.
I used Lazy.nvim to test Lazy.nvim to test it pre this Commit. My Config started to work again. It was tested on an clean Nvim build.


lucario387:
It's already reverted.

lucario387:
https://github.com/nvim-treesitter/nvim-treesitter/commit/f197a15b0d1e8d555263af20add51450e5aaa1f0

jpx40:
> It's already reverted.

it is? Cool man. i had an missunderstanding with one of maintainer, but this just brakes to much configs. Nvim-treesitter shoulden't dicide which parsers are installed.

lucario387:
Well we received a lot of issues related to neovim bundled parsers not compatible with our queries. That was our way (although unsuccesfully) to resolve the root cause.

jpx40:
> Well we received a lot of issues related to neovim bundled parsers not compatible with our queries. That was our way (although unsuccesfully) to resolve the root cause.

yeah, thanks for reverting this.

jpx40:
> Well we received a lot of issues related to neovim bundled parsers not compatible with our queries. That was our way (although unsuccesfully) to resolve the root cause.

Pls test pushes to master in the future to prevent issues like this.

lucario387:
Not that we don't test, we don't expect other consumers to call `setup()` multiple times (also lazyloading ...)

jpx40:
> Not that we don't test, we don't expect other consumers to call `setup()` multiple times (also lazyloading ...)

my, fault woth lazy loading. my whole config is built around lazy.nvim.

jpx40:
ok,  might try this

PR #5927, opened 2024-01-19
fix(highlights): use `markup.raw.delimiter` for fences

clason:


wookayin:
FYI, in other langs (sql, etc.) there are `@punctuation.delimiter` are used (was the capture group on v0.9.2). A reasonable default in the colorscheme will be: to link to `@puncutation.delimiter`.

clason:
> FYI, in other langs (sql, etc.) there are `@punctuation.delimiter` are used (was the capture group on v0.9.2). A reasonable default in the colorscheme will be: to link to `@puncutation.delimiter`.

To be honest, I don't see that (for SQL)? I'm not so sure about the reasonable default (old capture notwithstanding); I suspect any choice will make other people angry.

(There's been similar discussions about coloring quotes around strings differently...)

ribru17:
What should determine whether something's delimiters gets `@punctuation.delimiter` versus a `.delimiter` suffix highlight? E.g. right now bolded markdown text delimiters are also highlighted as `@markup.strong` (this is different from before the upstream capture PR merge) and I am wondering if all similar captures should have their own `.delimiter` if we do it for `@markup.raw` (too complicated?). Or if not, would it be odd to only have it for `@markup.raw`?

clason:
Punctuation.delimiter is about internal separators. Begin/end markers should inherit their block. But I don't think that is worth it for every block; codeblocks are a compromise. 

I'll look at the markdown emphasis delimiters whether it's easy (and sensible) to revert the change. (It's usually concealed anyway...)

CONTRIBUTING.md
```diff
@@ -193,11 +193,11 @@ As languages differ quite a lot, here is a set of captures available to you when
...
 
-@comment.error         ; error-type comments (e.g., `DEPRECATED:`)
-@comment.warning       ; warning-type comments (e.g., `WARNING:`, `FIX:`)
-@comment.hint          ; note-type comments (e.g., `NOTE:`)
+@comment.error         ; error-type comments (e.g. `DEPRECATED:`)
+@comment.warning       ; warning-type comments (e.g. `WARNING:`, `FIX:`)
+@comment.hint          ; note-type comments (e.g. `NOTE:`)
 @comment.info          ; info-type comments
-@comment.todo          ; todo-type comments (e.g-, `TODO:`, `WIP:`)
+@comment.todo          ; todo-type comments (e.g. `TODO:`, `WIP:`)
```
lucario387, line 200: Let's not do unrelated changes, shall we. Or at least put it in a different commit

lucario387: Actually you are clason so feel free :smile: I thought it was others doing the change.

lucario387:
I'm fine with either choices. The best way is probably to merge this and let users decide whether it's necessary or not.

CONTRIBUTING.md
clason: Yeah, that was a regression that somehow slipped in. I'm fine with pulling this into a separate commit.

clason:
Ok, here's my thinking:
1. `@punctuation.delimiter` is wrong here; these things are like strings (special content with start and end markers);
2. hence by default, the whole thing should be highlighted uniformly;
3. which means `.delimiters` or `.marker` (my favorite, but we already have `@tag.delimiter`) is the right thing;
4. having a separate group for emphasis markers would be too noisy (as they are concealed in Neovim anyway), but may be needed for consistency (but not documented);
5. an alternative would be `@markup.delimiter` (which I'd probably link to `Comment`, as a poor man's conceal).
6. alternatively, I am fine with reverting the changes to conceal captures (so they capture as `@conceal`, meaning no highlighting).

CONTRIBUTING.md
clason: I think there's a few doc improvements here; I'll put this in a separate PR.

clason:
OK, after thinking about it, here's what I want to do:
* use `@markup.raw.delimiter` for codespan markers (this will allow people to highlight them differently if they choose)
* do not document this capture (it's too niche for a third-level capture, since it has a reasonable fallback)
* revert the (incorrect) change for `(emphasis_marker)`.

I feel this is a reasonable compromise.

queries/markdown_inline/highlights.scm
```diff
@@ -16,10 +16,10 @@
 ] @string.escape
 
 ; Conceal codeblock and text style markers
-((code_span_delimiter) @markup.raw
+((code_span_delimiter) @markup.raw.delimiter
   (#set! conceal ""))
 
-((emphasis_delimiter) @markup.strong
+((emphasis_delimiter) @conceal
```
lucario387, line 22: If this is reverted dont we need to have something for it when not concealed

clason: No, it'll inherit from the containing `(emphasis)` etc. nodes. I simply should not have changed this.

clason: The issue here is that this node is shared between _all_ text formatting nodes (strong, italic, underlined, struck-through). So the choices are
1. no separate highlighting;
2. introduce `@markup.strong.delimiter` etc. (and restructure this pattern to fold it into the individual patterns, which might break due to directives with multiple nodes)
3. introduce a new higher-level `@markup.delimiter` capture.

I don't like 3., and 2. seems like too much work (and potential for issues) for this. Hence my choice for 1.

lucario387 (approved):
i'm fine with this change :+1:

PR #5930, opened 2024-01-19
feat(html): inject comment

ribru17:
Allows TODO comments and comment URIs, etc. to be recognized by comments in html, markdown, etc.

### DIFF
--- a/queries/html_tags/injections.scm
+++ b/queries/html_tags/injections.scm
@@ -1,3 +1,5 @@
+((comment) @injection.content
+  (#set! injection.language "comment"))
 ; <style>...</style>
 ; <style blocking> ...</style>
 ; Add "lang" to predicate check so that vue/svelte can inherit this


queries/html/injections.scm
```diff
@@ -1,4 +1,7 @@
 ; inherits html_tags
+((comment) @injection.content
+  (#set! injection.language "comment"))
+
```
lucario387, line 4: Should this be put in html_tags 🤔

ribru17: It has been moved. Also, after formatting, no new line is given after the queries (between the next comment). If I add a comment above the query that was just added, then a new line will be given after it. Is this intended? Just thought I would point it out.

lucario387: Kindof yes. It's easier that way to handle adjacent comments

lucario387: Ah I see what you mean.

I can fixup this later.

lucario387 (approved):

PR #5933, opened 2024-01-20
Backport rust changes from `main`

lucario387:
The rust indents commit from `main` duplicated the `indent.end` queries as well. This also remove that


lucario387:
I do want to backport injection changes as well, but it has `injection.combined`, which acts differently on nightly and stable.

clason (approved):

PR #5935, opened 2024-01-20
fix(markdown_inline): prioritize link URI

ribru17:
Right now the conceal code is overriding the `@markup.link.url` highlight for link URIs with `@markup.link`. This PR applies the more specific, non-conceal queries _after_ the conceal ones so they are given higher precedence.

**NOTE:** This is not necessary right now for the link label query, but I moved them both so they are near each other still.

### DIFF
--- a/queries/markdown_inline/highlights.scm
+++ b/queries/markdown_inline/highlights.scm
@@ -7,20 +7,8 @@
 
 (strikethrough) @markup.strikethrough
 
-[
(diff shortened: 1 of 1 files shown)


lucario387:
It surprisingly worked :?! :smile:

clason (approved):

lucario387:
I'll go with this for now :thinking:

PR #5942, opened 2024-01-20
fix(highlights): clean up tag highlights

ObserverOfTime:
Should yuck be using `@tag` for widget highlights?
If so, do we have any other widget languages that can use it?


clason (approved):
I trust you on this; I have zero knowledge of these languages.

ObserverOfTime:
> I trust you on this; I have zero knowledge of these languages.

So do I…

clason:
don't we have maintainers for these languages? 🤔

ObserverOfTime:
Let's see if they respond.
- cooklang: @addcninblue
- godot_resource: @pierpo
- thrift: @duskmoon314

lucario387:
Not all yuck symbols are tags from what I can tell through the docs. Only those listed in https://elkowar.github.io/eww/widgets.html seems to be tags. So we can separate them into two

ObserverOfTime:
I'll leave that to you or the maintainer (@Philipp-M). :)
(should also handle the built-in [variables](https://elkowar.github.io/eww/magic-vars.html) (constants?) and [jq](https://elkowar.github.io/eww/expression_language.html) injection)

lucario387:
I'm just gonna let yuck's maintainer handle it. I don't know/use eww anyways

lucario387 (approved):

pierpo:
> * godot_resource: @pierpo

My memory is quite blurry, I haven't touched this in a very long time :joy: 
What is the point of removing the `@tag` highlight? Can we replace it with something else? Isn't it language-specific?

Excuse my dumb questions :grin:

ObserverOfTime:
It's specific to XML-like languages.

pierpo:
OK, so that's the reason why we remove it? Is it because using `@tag` is not relevant here?

It still makes sense in the godot-resource files. Here `node` would be a tag. Do you think I could use something else than `@tag`?


```
[node parent="game/test"]
script = ExtResource( 1 )
editor/display_folded = true
float = 10000390.03230329
bool = false
str = "hello
       world \"okay\""
```

ObserverOfTime:
I don't see `node` in the highlights.

clason:
it's about the `[ ... ]` -- if `node` _is_ a `@tag`, these _are_ `@tag.delimiter`s

ObserverOfTime:
I looked into the grammar and I guess `@tag` works. Also fixed some other stuff.

Philipp-M:
Ok, just landed here, but I don't see any changes in yuck ^^.
So is there anything todo here?
It's some time ago (so memory is blurry), but I think I used `@tag` because yuck uses some kind of DOM like markup model similar to XML/HTML (with a lisp syntax), so I think I thought it was natural to use it there as well. AFAIK all symbols (other than some special keywords like `(for in)`) are some kind of widget (often custom). Not sure if the language evolved significantly (should maybe check that), but I think I covered all keywords with a different label, and otherwise all the symbols are some kind of DOM like widget.

I think the question remains: What label to use for widgets (other than `@tag`)?

lucario387:
I see something like `defwindow` inside yuck syntax, it can be `@function` or something similar

duskmoon314:
> * thrift: @duskmoon314

@amaanq, any ideas?

ObserverOfTime:
Surely you can review the one capture change on your own…

ObserverOfTime:
> I see something like `defwindow` inside yuck syntax, it can be `@function` or something similar

`defwindow` is highlighted as keyword.
I added builtin variable highlights and regex & jq injections in the last commit.

queries/yuck/highlights.scm
```diff
@@ -10,8 +10,6 @@
     (#eq? @keyword.import "include")))
 
 ; Keywords
-; I think there's a bug in tree-sitter the anchor doesn't seem to be working, see
-; https://github.com/tree-sitter/tree-sitter/pull/2107
 (list
   .
   ((symbol) @keyword
```
lucario387, line 15: While we're at it, can you also remove the unnecessary `()` at line 15-16

clason: or just run `./scripts/format-queries`...

ObserverOfTime: > or just run `./scripts/format-queries`...

I did…

lucario387: I haven't add it :)

`grouping` is kindof dangerous is what I think

clason:
Good to go?

ObserverOfTime:
I suppose.

clason:
You can always open a new PR with more changes, which are always welcome ;)

PR #5943, opened 2024-01-20
fix(comments)!: consolidate note captures

clason:
Distinguishing `hint` and `info` is too confusing; it's enough to have `error`, `warning`, `todo`, and `note`.

Alternative to #5929 

(@ribru17 thanks for pointing this out!)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -193,11 +193,10 @@ As languages differ quite a lot, here is a set of captures available to you when
 @comment               ; line and block comments
 @comment.documentation ; comments documenting code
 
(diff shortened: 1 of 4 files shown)


lucario387:
I personally think `@comment.error` are still useful. It looks weird for `ERROR` to be `warning`

clason:
Can you show me examples in the wild? (And it feels weird for a comment to be an error ;))

clason:
Another alternative is just to take the Helix queries verbatim; then it's not our problem ;)

lucario387 (approved):
`@text.*` -> `@comment.*` 2: Electric Boogaloo

PR #5944, opened 2024-01-20
feat(perl): highlight `continue`

rabbiveesh:


queries/perl/highlights.scm
```diff
@@ -29,6 +29,9 @@
   "foreach"
 ] @keyword.repeat
 
+("continue" @keyword.repeat
```
lucario387, line 32: Can't it just be `"continue" @keyword.repeat`

lucario387: I understand why now. https://perldoc.perl.org/functions/continue#continue-BLOCK

clason (approved):

PR #5947, opened 2024-01-21
feat(latex): more comprehensive highlights

ribru17:
Better precedence for highlighting environments (supporting different nesting levels), and operator highlights for super- and subscripts. Some examples below.

Before:
beforelatex

After:
afterlatex

A lot of the changes are simply from query reordering which applies highlights with different precedence. Other than that, this PR makes it so all nodes (not just `text`) in a math environment are highlighted as such, and the aforementioned operator highlights.

### DIFF
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -156,6 +156,12 @@
     (curly_group_text
       (_) @markup.link))
 
+; Formatting
+(text_mode
+  content:
+    (curly_group
+      (_) @none @spell))
+
 ; Math
 [
   (displayed_equation)
@@ -167,17 +173,17 @@
     command: _ @markup.math
     name:
       (curly_group_text
-        (text) @markup.math)))
+        (_) @markup.math)))
 
 (math_environment
-  (text) @markup.math)
+  (_) @markup.math)
 
 (math_environment
   (end
     command: _ @markup.math
     name:
       (curly_group_text
-        (text) @markup.math)))
+        (_) @markup.math)))
 
 ; Sectioning
 (title_declaration
@@ -275,12 +281,6 @@
     (curly_group
       (text) @markup.heading))
   (#eq? @_name "\\frametitle"))
-
-; Formatting
-(text_mode
-  content:
-    (curly_group
-      (_) @none @spell))
 
 ((generic_command
   command: (command_name) @_name
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -28,6 +28,8 @@
 [
   (operator)
   "="
+  "_"
+  "^"
 ] @operator
 
 "\\item" @punctuation.special


lucario387:
`@none` currently doesn't do anything :thinking: . It doesn't even have a hlgroup def.

clason:
Note that these highlights were deliberate to match VimTeX syntax highlighting (which I consider the reference): In particular, `$`, `$$`, `\[`, and `\begin{equation}` all are highlighted the same, namely as "math mode delimiters" -- which, like other delimiters, inherit the highlight of what they delimit.

As I personally don't use the tree-sitter parser for highlighting, I don't want to block further changes, **but** only if someone then takes over the maintenance of these queries and promises to extend them to be (much) more comprehensive.

ribru17:
> `@none` currently doesn't do anything 🤔 . It doesn't even have a hlgroup def.

I don't understand, I thought this was determined by colorscheme/user? None is used in other places like `gitcommit` to indicate highlight removal

> Note that these highlights were deliberate to match VimTeX syntax highlighting (which I consider the reference): In particular, `$`, `$$`, `\[`, and `\begin{equation}` all are highlighted the same, namely as "math mode delimiters" -- which, like other delimiters, inherit the highlight of what they delimit.

I am fine to revert this change, this makes sense. I think the other changes should be kept unless someone disagrees

> As I personally don't use the tree-sitter parser for highlighting, I don't want to block further changes, **but** only if someone then takes over the maintenance of these queries and promises to extend them to be (much) more comprehensive.

I would quite like to but I don't have a strong enough grasp of LaTeX I'm afraid :smile:

clason:
> I don't understand, I thought this was determined by colorscheme/user? None is used in other places like gitcommit to indicate highlight removal

That doesn't mean they work... ;) That's a different issue though. It doesn't do any harm so you can leave it in if you prefer (but it's not required for the `@spell` to work).

> I am fine to revert this change, this makes sense. I think the other changes should be kept unless someone disagrees

No, I think the other changes (operators, nesting) are good.

queries/latex/highlights.scm
```diff
@@ -1,3 +1,9 @@
+; Formatting
+(text_mode
+  content:
+    (curly_group
+      (_) @none @spell))
```
wookayin, line 5: Do we really need to move this to top?

wookayin (commented):
After the suggested changes, it looks like we have two main points of change. It'd be good if the title of the commit/PR can be more specific.

queries/latex/highlights.scm
ribru17: It just needs to be before the `@markup.math` highlights, I'll just put it right above them. I think grouping them like that will make more sense. I will break up the commits as well like you say :+1:

clason (approved):

petobens:
Hi @ribru17. Great work here! I have two questions:
1. As it can be seen in the image for some reason `@markup.math.latex` still take precedence over `@markup.environment.latex`. Do I need to add some priority setting?
2. Is it possible to highlight the `$$` operators differently?

Thanks in advance!

clason:
> Is it possible to highlight the $$ operators differently?

In screaming red? (You should _never_ use `$$` in LaTeX. And the parser is not meant for plain TeX.)

petobens:
Sure. I meant within a markdown file.

ribru17:
> 1. As it can be seen in the image for some reason `@markup.math.latex` still take precedence over `@markup.environment.latex`. Do I need to add some priority setting?

This was kept intentionally to match the vimtex highlighting (there was some back and forth here), but yeah you can override it with a priority change. Also for the `$$` you can use something like:
```query
(displayed_equation "$$" @myhighlight)
```

clason:
> Sure. I meant within a markdown file.

Yeah, but that's not the primary target here. It's a LaTeX parser first and foremost.

PR #5950, opened 2024-01-21
fix(format-scripts): Remove parens if possible

lucario387:
Dangerland but somehow it doesn't seem to break anything

### DIFF
--- a/queries/wgsl_bevy/highlights.scm
+++ b/queries/wgsl_bevy/highlights.scm
@@ -14,7 +14,7 @@
 
 (function_declaration
   (import_path
-    ((identifier) @function.)))
+    (identifier) @function.))
 
 (import_path
   (identifier) @module
--- a/queries/apex/highlights.scm
+++ b/queries/apex/highlights.scm
@@ -57,8 +57,8 @@
 (type_arguments
   ">" @punctuation.delimiter)
 
-((field_access
-  object: (identifier) @type))
+(field_access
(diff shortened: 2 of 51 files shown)


clason (approved):

lucario387:
Guess it does :cry:

queries/wgsl_bevy/highlights.scm
```diff
@@ -14,7 +14,7 @@
 
 (function_declaration
   (import_path
-    ((identifier) @function.)))
+    (identifier) @function.))
```
ObserverOfTime, line 17: Where's the space? :thinking:

lucario387: It was a remnant from #5854 that was not looked at

ObserverOfTime (commented):
```
queries/css/indents.scm
9:("}") @indent.dedent

queries/jq/highlights.scm
31:("null") @type.builtin

queries/tlaplus/highlights.scm
122:("return") @keyword.return
124:("print") @function.macro
```

lucario387:
> ```
> queries/css/indents.scm
> 9:("}") @indent.dedent
> 
> queries/jq/highlights.scm
> 31:("null") @type.builtin
> 
> queries/tlaplus/highlights.scm
> 122:("return") @keyword.return
> 124:("print") @function.macro
> ```

Should be resolved now

ObserverOfTime (approved):

PR #5952, opened 2024-01-22
feat(git_config): shell command alias injections

ribru17:
Provides `bash` injections for Git aliases that run external commands, e.g. things like:
```.gitconfig
[alias]
	poc = !git push origin $(git symbolic-ref --short HEAD)
	squash = "!f() { git rebase -i HEAD~$1; }; f"
```
(excludes the leading `!` character and surround quotes, if any)


queries/git_config/injections.scm
```diff
@@ -6,3 +6,26 @@
...
+
+(section
+  (section_header
+    (section_name) @_section)
+  (variable
+    value: (string) @injection.content)
+  (#eq? @_section "alias")
+  (#lua-match? @injection.content "^!")
+  (#offset! @injection.content 0 1 0 0)
+  (#set! injection.language "bash"))
```
lucario387, line 18: Firstly, is there a documentation for this? 
Secondly, can this be anything else other than bash (I'll presume not because even windows have git-bash)

queries/git_config/injections.scm
```diff
@@ -6,3 +6,26 @@
   value: (string) @injection.content)
   (#any-of? @_name "cmd" "command" "textconv")
   (#set! injection.language "bash"))
+
+(section
+  (section_header
+    (section_name) @_section)
+  (variable
+    value: (string) @injection.content)
+  (#eq? @_section "alias")
```
wookayin, line 15:
```suggestion
    (section_name) @_section)
  (#eq? @_section "alias")
  (variable
    value: (string) @injection.content)
```

queries/git_config/injections.scm
wookayin: I think this is good --  https://git-scm.com/docs/git-config#Documentation/git-config.txt-alias

> If the alias expansion is prefixed with an exclamation point, it will be treated as a shell command

queries/git_config/injections.scm
lucario387: I think it's fine for predicates and directives to sit all at the end of the query. Just styling preferences that the formatter can't enforce right now.

queries/git_config/injections.scm
lucario387: So from the doc, I assume the command can also be a powershell command. I'm somewhat against this because of that

ribru17: Ah, I see. I will close then. Maybe one day we can get a fancy directive that gets the current user shell and injects that one for things like this :joy: However, I think it is worth noting that `bash` is injected for other commands:
```scheme
((variable
  (name) @_name
  value: (string) @injection.content)
  (#any-of? @_name "cmd" "command" "textconv")
  (#set! injection.language "bash"))
```
From the docs it looks like these are also just run in the general shell. Should this be kept?

lucario387: :thinking: I guess if it's for consistency purposes then your PR should also be kept as well.

ribru17: Sounds good. I am curious how Github handles this, it seems to have some sort of injection for the alias/command things.

queries/git_config/injections.scm
```diff
@@ -6,3 +6,26 @@
...
+(section
+  (section_header
+    (section_name) @_section)
+  (variable
+    value:
+      (string
+        "\""
+        "\"") @injection.content)
+  (#eq? @_section "alias")
+  (#lua-match? @injection.content "^\"!")
```
ObserverOfTime, line 29: What about single quotes?

ribru17: I think single quotes may not be an option, I tried them out in my config and single quotes cause errors where double quotes do not. Same with adding single quotes around an unquoted command

ObserverOfTime: Tree-sitter errors or git errors?

ObserverOfTime (approved):

queries/git_config/injections.scm
wookayin: Yup it's fine, but it'd better in style (although we do not enforce). As in other queries, predicates are not necessarily placed at the end, so I thought it'd be better to follow the style of them.

ObserverOfTime: I place them at the end in my queries. ¯\\\_(ツ)\_/¯

clason: In general, please avoid stylistic preferences that 1. don't make an observable difference and 2. are not enforced through our formatter. If a convention is worth having, it's worth enforcing; otherwise we'll end up with inconsistent back-and-forth.

PR #5954, opened 2024-01-22
fix(kotlin): temporary fix for treesitter bug that causes delay in file opening

### LINKED ISSUE #5810: Slight delay in the opening of Kotlin files
https://github.com/nvim-treesitter/nvim-treesitter/assets/69449791/bcf5dcfe-17bc-4ae1-a1bc-201830d1ffa1

As you can see, there is a slight delay in the loading of kotlin files, which used to be instantaneous. I must also point out that this delay has nothing to do with the size of the file being opened. Here is a video of the same delay occuring for a kotlin file with only two lines: 

https://github.com/nvim-treesitter/nvim-treesitter/assets/69449791/f52d4165-7a88-4070-80fa-d532b7fac3cf

I did open a similar issue here last month but was advised to open an issue at `tree-sitter-kotlin` instead, which I did: fwcd/tree-sitter-kotlin#107

The tests conducted by the developers over at `tree-sitter-kotlin` yielded nothing unusual (mean parsing time of `8.1ms`), which led me to believe that something in the plugin is causing the delay after all. 

**Additional Info:** 
Note that the delay disappears completely after I rolled back the plugin to commit `7b26b085` and reinstalled the parser.
 
https://github.com/nvim-treesitter/nvim-treesitter/assets/69449791/4b652647-9c71-44d1-a870-0827aac280fe

After rolling back to `7b26b085`, the opening of kotlin files is back to being instantaneous. 

These tests are conducted using the following minimal config:
```lua
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({"config", "data", "state", "cache"}) do
    vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

--bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
    vim.fn.system({
        "git",
        "clone",
        "--filter=blob:none",
        "--single-branch",
        "https://github.com/folke/lazy.nvim.git",
        lazypath,
    })
end
vim.opt.runtimepath:prepend(lazypath)


--install plugins
local plugins = {
    {
        build = ":TSUpdate",
    },
}

require("lazy").setup(plugins, { root = root .. "/plugins", })

--setup treesitter
local ok, treesitter = pcall(require, "nvim-treesitter.configs")
if ok then
    treesitter.setup({
        ensure_installed = { "kotlin" },
        highlight = { enable = true },
    })
end
```

I will pin the plugin to `7b26b085` for now until this issue is resolved.

anthony-S93:
Fixes #5810 

The following query inside `query/kotlin/highlights.scm` leads to noticeable delay in the opening of kotlin files.
```
(class_body
	(property_declaration
		(variable_declaration
			(simple_identifier) @property)))
```

### DIFF
--- a/queries/kotlin/highlights.scm
+++ b/queries/kotlin/highlights.scm
@@ -27,9 +27,11 @@
 (class_parameter
   (simple_identifier) @variable.member)
 
-((simple_identifier) @property
-  (#has-parent? @property variable_declaration))
-
+; NOTE: temporary fix for treesitter bug that causes delay in file opening
+;(class_body
+;  (property_declaration
+;    (variable_declaration
+;      (simple_identifier) @variable.member)))
 ; id_1.id_2.id_3: `id_2` and `id_3` are assumed as object properties
 (_
   (navigation_suffix
--- a/queries/kotlin/highlights.scm
+++ b/queries/kotlin/highlights.scm
@@ -27,10 +27,8 @@
 (class_parameter
   (simple_identifier) @variable.member)
 
-(class_body
-  (property_declaration
-    (variable_declaration
-      (simple_identifier) @variable.member)))
+((simple_identifier) @property
+  (#has-parent? @property variable_declaration))
 
 ; id_1.id_2.id_3: `id_2` and `id_3` are assumed as object properties
 (_
--- a/queries/kotlin/locals.scm
+++ b/queries/kotlin/locals.scm
@@ -32,11 +32,11 @@
     (variable_declaration
       (simple_identifier) @local.definition.parameter)))
 
-(class_body
-  (property_declaration
-    (variable_declaration
-      (simple_identifier) @local.definition.field)))
-
+; NOTE: temporary fix for treesitter bug that causes delay in file opening
+;(class_body
+;  (property_declaration
+;    (variable_declaration
+;      (simple_identifier) @local.definition.field)))
 (class_declaration
   (primary_constructor
     (class_parameter


queries/kotlin/highlights.scm
```diff
@@ -27,10 +27,8 @@
 (class_parameter
   (simple_identifier) @variable.member)
 
-(class_body
-  (property_declaration
-    (variable_declaration
-      (simple_identifier) @variable.member)))
+((simple_identifier) @property
+  (#has-parent? @property variable_declaration))
```
lucario387, line 31: We need to move this up.

This overrides the query right above it.

Please move it upper

ObserverOfTime (commented):
It should be `@variable.member` instead of `@property` and probably use `property_declaration` rather than `variable_declaration` (since `variable_declaration` will likely match regular variables as well).

queries/kotlin/highlights.scm
theHamsta: This has to be some kind of bug in the tree-sitter query implementation.`#has-parent` does actually very inefficient look up of it's parent. Is the load/query delay also there if you just have (variable_declaration (simple_identifier))?

anthony-S93: > Is the load/query delay also there if you just have (variable_declaration (simple_identifier))?

Can you clarify this? Do you mean in the original query or the new one?

lucario387:
>  probably use `property_declaration` rather than `variable_declaration`

The issue here is writing a query with `(variable_declaration)` is the culprit to slowing down massively. Because of that I opted to do `has-parent?` to workaround it

queries/kotlin/highlights.scm
lucario387: > Is the load/query delay also there if you just have (variable_declaration (simple_identifier))?

From what I tested, yes, it's that pattern causing issue

ObserverOfTime:
I don't mind using `has-parent?` to solve the issue, but `(variable_declaration)` is not limited to properties afaik.

lucario387:
> I don't mind using `has-parent?` to solve the issue, but `(variable_declaration)` is not limited to properties afaik. 

This is true :thinking: It's also in for loops and other places

theHamsta:
> I don't mind using `has-parent?` to solve the issue, but `(variable_declaration)` is not limited to properties afaik.

Yes, it's no problem to use it as a work-around. But we should also look for the underlying problem in Tree-Sitter query engine for this grammar/query combination

clason:
> Yes, it's no problem to use it as a work-around. But we should also look for the underlying problem in Tree-Sitter query engine for this grammar/query combination

There's a few such random-seeming issues; for example the Dart indent pattern for conditional branches.

lucario387:
Probably not for a while. I will be inactive for the next month or more so I cannot do the fixup for those Dart indents. 

In the mean time you can comment out all if blocks and just put a 'idnent.auto' for (if_statement)

clason:
> In the mean time you can comment out all if blocks and just put a 'idnent.auto' for (if_statement)

That's what I meant. I'd be more comfortable with you doing that, as you know the indent subsystem well. (Provided you have time for that, of course.)

clason:
There's a similar issue with the `locals.scm` query -- could you check that as well?

queries/kotlin/highlights.scm
clason: And, yes, that's a parser (most likely) or tree-sitter bug. Nevertheless, it's acceptable as a workaround **iff** there's a comment before it explaining that.

@anthony-S93 can you do that (and also make the corresponding change in `locals.lua`)?

anthony-S93: @lucario387 Should `@property` be replaced with `@variable.member` as @ObserverOfTime suggested? 

@clason I have added a comment to clarify that the change is only a temporary workaround. Also, do you mean `locals.scm`? Inside `locals.scm`, I see two queries that follow the same pattern in the discussion. 

```txt
(lambda_literal
  (lambda_parameters
    (variable_declaration
      (simple_identifier) @local.definition.parameter)))

(class_body
  (property_declaration
    (variable_declaration
      (simple_identifier) @local.definition.field)))
```

I am not familiar enough with Treesitter's query syntax to confidently make the changes you requested. Should I replace those two queries with something like this?

```txt
((simple_identifier) @local.definition.field
  (#has-parent? @local.definition.field) variable_declaration)
```

clason: It's only the `class_body` that's problematic, but since you need to distinguish the two, you probably need both. Your second variant doesn't make any difference since it still has the nesting; the first variant works, though.

@lucario387 ?

lucario387: Idk how other plugins use locals, but pretty sure it will affect all other locals queries.

Maybe we should just comment it out.

clason: Yeah, sorry, can't read. The first variant has identical patterns, so the second overwrites the first. You need to check for `lambda_parameters` and `property_declaration` as `#has-ancestor?` (which will be even more expensive, but probably still faster than what we have now).

lucario387: I'd vouch for just commenting it out. `has-ancestor` may have unwanted side-effects

clason: Sounds good to me (but just the `class_declaration` one; the `lambda_literal` can stay).

anthony-S93: > I'd vouch for just commenting it out. `has-ancestor` may have unwanted side-effects

So we comment out both of these?
```txt
(lambda_literal
  (lambda_parameters
    (variable_declaration
      (simple_identifier) @local.definition.parameter)))

(class_body
  (property_declaration
    (variable_declaration
      (simple_identifier) @local.definition.field)))
```

clason: > We need to move this up.
> 
> This overrides the query right above it.
> 
> Please move it upper

Just to be clear: the current state of the PR is fine, or does this still need to be done?

clason: > So we comment out both of these in locals.scm?

No, just the second one.

queries/kotlin/highlights.scm
```diff
@@ -24,14 +24,13 @@
...
+
 (class_parameter
   (simple_identifier) @variable.member)
 
-(class_body
-  (property_declaration
-    (variable_declaration
-      (simple_identifier) @variable.member)))
```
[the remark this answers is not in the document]
clason: Why? Just squash and force-push. It's better to keep the change together with the discussion, for future reference.

queries/kotlin/highlights.scm
lucario387: > Just to be clear: the current state of the PR is fine, or does this still need to be done?

That part is done.

queries/kotlin/highlights.scm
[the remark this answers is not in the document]
clason: Or we can squash when merging, if you prefer that.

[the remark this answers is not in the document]
anthony-S93: @clason  @lucario387 

Please check.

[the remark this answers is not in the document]
clason: You can (and should) keep the simplified pattern in the highlight (also with a comment).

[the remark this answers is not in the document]
anthony-S93: Done. Please verify.

clason (approved):

queries/kotlin/highlights.scm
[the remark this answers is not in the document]
lucario387: Please run `./script/format-queries queries/kotlin` as well

queries/kotlin/highlights.scm
```diff
@@ -24,6 +24,10 @@
 (super_expression
   "@" @variable.builtin)
 
+; NOTE: simplified pattern to replace (temporarily) problematic query that causes delay (see below)
+((simple_identifier) @property
```
lucario387, line 28: variable.member, yes no maybe?

queries/kotlin/highlights.scm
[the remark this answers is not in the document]
clason: Looks good. Just run `./scripts/format-queries.lua queries/kotlin` and I'll merge.

queries/kotlin/highlights.scm
clason: Yes, needs to be `@variable.member`; `@property` is a regression from the capture cleanup.

clason (changes requested):
* switch to `@variable.member`
* run the query formatter

Then I'll merge.

ObserverOfTime (changes requested):
I'll say it again:  `(variable_declaration)` is not limited to properties.

anthony-S93:
> * switch to `@variable.member`
> 
>     * run the query formatter
> 
> 
> Then I'll merge.

I keep getting errors when running `./scripts/format-queries.lua queries/kotlin`: 
```txt
E5113: Error while calling lua chunk: /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:93: no parser for 'query' language, see :help treesitter-parsers
stack traceback:
	[C]: in function 'error'
	/usr/share/nvim/runtime/lua/vim/treesitter/language.lua:93: in function 'add'
	/usr/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:98: in function 'new'
	/usr/share/nvim/runtime/lua/vim/treesitter.lua:61: in function '_create_parser'
	/usr/share/nvim/runtime/lua/vim/treesitter.lua:131: in function 'get_parser'
	./scripts/format-queries.lua:403: in function 'format'
	./scripts/format-queries.lua:419: in main chunk
```
I'm pretty sure that the parser for the `query` language is already installed (there is a `query.so` file in the parser directory).

anthony-S93:
> I'll say it again: `(variable_declaration)` is not limited to properties.

So we replace `variable_declaration` with `property_declaration`? Can you confirm this? @clason  @lucario387 ?

ObserverOfTime:
> So we replace `variable_declaration` with `property_declaration`?

Unfortunately, `property_declaration` is not limited to properties either. The parser is a mess.

clason:
You know what? Let me fix this for you.

ObserverOfTime:
Fix the parser? :^)

clason:
The PR.

clason (approved):

ObserverOfTime:
Just remove the workaround and wait for the issue to be fixed.
Otherwise every variable will be highlighted as `@variable.member`.
(The LSP does that too but that doesn't mean we should *also* be wrong.)

clason:
@ObserverOfTime good to go now?

PR #5963, opened 2024-01-23
fix(incremental_selection): mute deleting possibly non-existent keymaps

przepompownia:
It can happen when the buffer gets some change from outside e.g. `git stash`.

### DIFF
--- a/lua/nvim-treesitter/incremental_selection.lua
+++ b/lua/nvim-treesitter/incremental_selection.lua
@@ -162,12 +162,18 @@ end
 
 function M.detach(bufnr)
   local config = configs.get_module "incremental_selection"
+  ---@type string
+  local mode
   for f, mapping in pairs(config.keymaps) do
     if mapping then
       if f == "init_selection" then
-        vim.keymap.del("n", mapping, { buffer = bufnr })
+        mode = "n"
       else
-        vim.keymap.del("x", mapping, { buffer = bufnr })
+        mode = "x"
+      end
+      local ok, err = pcall(vim.keymap.del, mode, mapping, { buffer = bufnr })
+      if not ok then
+        vim.notify(('%s "%s" for mode "%s"'):format(err, mapping, mode), vim.log.levels.DEBUG, {title = "nvim-treesitter"})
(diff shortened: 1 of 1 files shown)


clason:
Thank you, but that function is already slated for removal in nvim-treesitter 1.0 (together with the whole module system). So unless someone rewrites it as a standalone module similar to nvim-treesitter-context or the indent module (on `main`), this is to be considered unmaintained.

przepompownia:
I proposed such change and use it only to avoid displaying stack traces while the main branch seems to be still not ready to use it as the master (i.e. without losing any functionality).

lucario387:
> be still not ready to use it as the master (i.e. without losing any functionality).

I'd argue even a week after 0.10 is released, some functionalities will still be missing

przepompownia:
Independently it would be good to fix this issue (not necessary with `vim.notify` included here) while the `master` branch without `main` is useful.

clason:
I don't think this is a good fix; you're just putting lipstick on a pig (silencing an error and instead outputting your own warning). A proper fix would be to find out _why_ it's trying to delete a non-existing keymap and prevent that from happening.

(Or, just don't mess with mappings at all and let the user handle this themselves -- which is the point of the rewrite on `main`. This module system was written before we had a good Lua API for mappings and user commands.)

przepompownia:
For me this is not a good fix either. Maybe it's a bit of a provocation to write something better instead. 

I started from looking if we can check single buffer keymap existence. `nvim_buf_get_keymap` returns the whole array for given mode. I have not checked yet where and when `detach()` is called (quick search leads me to https://github.com/nvim-treesitter/nvim-treesitter/blob/023a8ee9a2e3fcb134f0155d9390e068e85bb4f4/lua/nvim-treesitter/ts_utils.lua#L374). 

It's probably an issue similar to calling a custom diagnostic handler when buffer contents get some change from outside.

clason:
And this ties into my earlier point: rather than adding hacks on top of hacks, this needs a fundamental rewrite that avoids these issues. This is much less work in the end.

PR #5966, opened 2024-01-24
chore: update luadoc parser & highlights

amaanq:


queries/luadoc/highlights.scm
lucario387, line 1: What about member_type :thinking: 

I have this in my highlight (java-like scoping)
```scheme
(member_type
  (identifier) @namespace)
(member_type
  (identifier) @type .)
```

amaanq: yeah that could be a nice addition

lucario387: Oops
It should be module :smile:

queries/luadoc/highlights.scm
```diff
@@ -93,6 +98,12 @@
   .
   (identifier) @variable.member)
 
+(member_type
+  (identifier) @namespace)
```
lucario387, line 102:
```suggestion
  (identifier) @module)
```

PR #5968, opened 2024-01-24
chore(c_sharp): highlight tweaks

amaanq:


queries/c_sharp/highlights.scm
```diff
@@ -4,6 +4,9 @@
   (#eq? @keyword "value")
   (#has-ancestor? @keyword accessor_declaration))
 
+((identifier) @constant
+  (#eq? @constant "_"))
```
lucario387, line 8: Shouldn't this only be constants in some cases :thinking: 

I don't think all cases are fine, even though no normal person will use it as a variable

amaanq: Nah it's like that everywhere, it's a special variable in c#

lucario387: but that doesn't necessarily mean it's a constant variable :thinking: 

Im just skimming through C# docs: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards

> _ is also a valid identifier. When used outside of a supported context, _ is treated not as a discard but as a valid variable. If an identifier named _ is already in scope, the use of _ as a standalone discard can result in:
> - Accidental modification of the value of the in-scope _ variable by assigning it the value of the intended discard. For example:

amaanq: sure, I'm open to something else like `@variable.builtin`

lucario387: I'd be more happy with that.

lucario387 (approved):

PR #5974, opened 2024-01-24
fix(jsx): proper tag and builtin tag distinction

ribru17:
This PR changes the uppercase JSX tag highlights from `@constructor` to `@tag`. There was discussion about this before and it was decided to keep it this way so that they were differentiated from builtin tags, but now we have `@tag.builtin` so I figured this would be OK. This PR changes existing highlights of `@tag` to the builtin variant, so users and colorscheme authors can still have proper differentiation. Using constructor is especially annoying for colorschemes like [onedark](https://github.com/navarasu/onedark.nvim) that use bolded constructors or other styles that cascade through just color highlights.

### DIFF
--- a/queries/jsx/highlights.scm
+++ b/queries/jsx/highlights.scm
@@ -24,42 +24,42 @@
   (property_identifier) @tag.attribute)
 
 (jsx_opening_element
-  name: (identifier) @tag)
+  name: (identifier) @tag.builtin)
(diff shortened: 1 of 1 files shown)


amaanq (approved):

amaanq:
Fine w/ me but I don't do web dev stuff

PR #5979, opened 2024-01-25
fix(dart): replace slow indent with simpler strategy

clason:
Pattern for dedenting else blocks is painfully slow (~20% of total test
time), so replace with `@indent.auto` for the whole `if_statement` for
now.


lucario387 (approved):
I will try to figure out a better way to fix this after the hiatus

For now this is good

Issue #5980, opened 2024-01-25
"locals" update broke functions which iterates locals


leftas:
### Describe the bug

Similar bug to what https://github.com/nvim-treesitter/nvim-treesitter/pull/5956 PR fixed in `get_*` functions - account for new “local” nesting. 

I am experiencing breaking of the `find_usages` function. As it tries to check whether matches have reference and node,  however, it does not account for a change in new structure being:
`{["local"] = { reference = { node = { ... }}}}`
```lua
  for match in M.iter_locals(bufnr, scope_node) do
    if match.reference and match.reference.node and ts.get_node_text(match.reference.node, bufnr) == node_text then
```

Mainly, I could submit PR myself, but I have trouble identifying whether something else will be/is touched by this and whether this fix should be in iter_locals or find_usages itself.

### To Reproduce

You can use any module which uses get_usages to do any work, in this case I used nvim-treesitter-refactor's navigation, specifically `goto_next_usage`.

1. Enable navigation submodule: https://github.com/nvim-treesitter/nvim-treesitter-refactor?tab=readme-ov-file#navigation
2. Open a file supported by TS.
3. Go onto any variable/function or any TS node for that matter.
4. Use "goto_next_usage" functionality with your defined bindings (by default ALT-*). OR call method manually.

Or 

1. Call nvim-treesitter.locals' `find_usages` method manually supplying a node.

### Output of `:checkhealth nvim-treesitter`

```markdown
==============================================================================
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- OK `tree-sitter` found 0.20.8 (parser generator, only needed for :TSInstallFromGrammar)
- OK `node` found v20.3.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: Apple clang version 15.0.0 (clang-1500.0.40.1)
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "arm64",
  release = "23.1.0",
  sysname = "Darwin",
  version = "Darwin Kernel Version 23.1.0: Mon Oct  9 21:28:45 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T6020"
} ~

Parser/Features         H L F I J
[...]

PR #5982, opened 2024-01-25
Fix find_usages

leftas:
Mostly I explained everything in my issue - https://github.com/nvim-treesitter/nvim-treesitter/issues/5980. 

Should close https://github.com/nvim-treesitter/nvim-treesitter/issues/5980 if merged.

Although, I haven't put `if` for checking whether that “local” exists, I don't think we should as per other lines: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/locals.lua#L44

### DIFF
--- a/lua/nvim-treesitter/locals.lua
+++ b/lua/nvim-treesitter/locals.lua
@@ -254,6 +254,7 @@ function M.find_usages(node, scope_node, bufnr)
   local usages = {}
 
   for match in M.iter_locals(bufnr, scope_node) do
+    match = match["local"]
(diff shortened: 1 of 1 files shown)


clason (approved):

clason:
Thanks!

clason:
(conventional commit would have been nice, but it fixes a nasty regression, which is more important)

leftas:
Ahh, whoops! My bad, didn't look at other PRs and couldn't find any info on PRs. Sorry!

clason:
nah, if anyone is at fault, it's me for not squash-merging it.

PR #5985, opened 2024-01-25
fix(slint)!: Change slint tree-sitter to official slint repo

jrmoulton:
Points slint to the official repository as I haven't maintained the other one in a while. 

Marking as a draft for now because I believe that slint needs to add the generated c source files first. 

@hunger I've marked you as the maintainer. Does that work?

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1697,7 +1697,8 @@ list.slang = {
 list.slint = {
   install_info = {
     url = "https://github.com/slint-ui/slint",
-    files = { "editors/tree-sitter-slint/src/parser.c" },
+    files = { "src/parser.c" },
+    location = "editors/tree-sitter-slint",
   },
   maintainers = { "@hunger" },
   experimental = true,
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1696,10 +1696,10 @@ list.slang = {
 
 list.slint = {
   install_info = {
-    url = "https://github.com/jrmoulton/tree-sitter-slint",
-    files = { "src/parser.c" },
+    url = "https://github.com/slint-ui/slint",
+    files = { "editors/tree-sitter-slint/src/parser.c" },
   },
-  maintainers = { "@jrmoulton" },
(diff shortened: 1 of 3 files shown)


clason:
@theHamsta

lua/nvim-treesitter/parsers.lua
```diff
@@ -1696,10 +1696,10 @@ list.slang = {
 
 list.slint = {
   install_info = {
-    url = "https://github.com/jrmoulton/tree-sitter-slint",
-    files = { "src/parser.c" },
+    url = "https://github.com/slint-ui/slint",
+    files = { "editors/tree-sitter-slint/src/parser.c" },
```
clason, line 1700: you should set the `location` field instead.

clason:
> Marking as a draft for now because I believe that slint needs to add the generated c source files first.

You can always set `generate_from_grammar = true` (check other parsers in this list).

clason:
Also, if the queries remain unchanged, this is not a breaking change.

clason:
can you update the lockfile and README entries?

And do you expect the parser to be committed to the repo in the near future?

clason:
@hunger looks like there _were_ breaking changes in the grammar, so the (highlight and indent) queries need to be adapted.

clason:
Actually, the grammar looks ver different; you'll need to completely rewrite the queries.

clason:
>  I assumed nvim-treesitter would have some config somewhere to get them from there.

That is very wrong. Queries are editor-specific, so you need to adapt the ones _in this repo_.

clason:
I've already moved the SHA; that's how the tests are failing and I found out.

clason:
>  Would it be OK to add an injection into the rust queries?

Sure.

lucario387:
For only `nvim-treesitter`, `locals.scm` doesn't do anything. It's used in nvim-treesitter-refactor and possibly other plugins, so it is kept.

clason:
the official support channel is https://matrix.to/#/#nvim-treesitter:matrix.org

clason:
Best rebase on master yet again; I changed the macos workflow to use M1 so the required checks changed.

clason:
If it's easier for you, you can just open a new PR from your branch; I don't mind.

PR #5986, opened 2024-01-26
fix(lua): inject luadoc into comments immediately beginning with a `|`

amaanq:
context - https://github.com/amaanq/tree-sitter-luadoc/issues/8#issuecomment-1911045824

I can understand this being rejected for false positives with normal comments just happening to start w/ a `|`, but I'd assume that'd be rare

### DIFF
--- a/queries/lua/injections.scm
+++ b/queries/lua/injections.scm
@@ -131,7 +131,7 @@
 
 (comment
   content: (_) @injection.content
-  (#lua-match? @injection.content "^[-][%s]*@")
+  (#lua-match? @injection.content "^[-][%s]*[@|]")
   (#set! injection.language "luadoc")
   (#offset! @injection.content 0 1 0 0))
 


queries/lua/injections.scm
```diff
@@ -135,6 +135,12 @@
   (#set! injection.language "luadoc")
   (#offset! @injection.content 0 1 0 0))
 
+(comment
+  content: (_) @injection.content
+  (#lua-match? @injection.content "^[-][|]")
```
lucario387, line 140: Why not combine with the previous pattern?
```
(#lua-match? @injection.content "^[-][%s]*[|]")
```

amaanq: I was debating that but my rationale was that it might let in more false positives - but I did check lua-ls and spaces don't seem to matter, my bad

queries/lua/injections.scm
```diff
@@ -135,6 +135,12 @@
   (#set! injection.language "luadoc")
   (#offset! @injection.content 0 1 0 0))
 
+(comment
+  content: (_) @injection.content
+  (#lua-match? @injection.content "^[-][%s]*[|]")
```
lucario387, line 140: I mean, is there a reason for this `[|]` to not be put with the query pattern right above it? :thinking:

amaanq: lua match doesn't have an OR operator?

lucario387: I don't see why this pattern won't work
```
(comment
  content: (_) @injection.content
  (#lua-match? @injection.content "^[-][%s]*[|@]")
  (#set! injection.language "luadoc")
  (#offset! @injection.content 0 1 0 0))
```

amaanq: right...

lucario387 (approved):

PR #5988, opened 2024-01-26
feat(d)!: switch upstream parser to `gdamore/tree-sitter-d`

amaanq:
Supersedes #4898

### DIFF
--- a/queries/d/highlights.scm
+++ b/queries/d/highlights.scm
@@ -1,21 +1,317 @@
-; Misc
+; Keywords
 [
-  (line_comment)
-  (block_comment)
-  (nesting_block_comment)
-] @comment @spell
+  (directive)
+  (shebang)
+] @keyword.directive
 
-((line_comment) @comment.documentation
-  (#lua-match? @comment.documentation "^///[^/]"))
+[
+  (import)
+  (module)
+] @keyword.import
 
-((line_comment) @comment.documentation
-  (#lua-match? @comment.documentation "^///$"))
+[
+  (alias)
+  (asm)
+  (class)
+  (delegate)
+  (delete)
+  (enum)
+  (interface)
+  (invariant)
+  (mixin)
+  (pragma)
+  (struct)
+  (template)
+  (union)
+  (unittest)
+  (version)
+  (with)
+  (traits)
+  (vector)
+  (parameters_)
+  (default)
+  (goto)
+] @keyword
 
-((block_comment) @comment.documentation
-  (#lua-match? @comment.documentation "^/[*][*][^*].*[*]/$"))
+(function) @keyword.function
 
-((nesting_block_comment) @comment.documentation
-  (#lua-match? @comment.documentation "^/[+][+][^+].*[+]/$"))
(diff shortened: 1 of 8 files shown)


clason:
Not related to this PR, but I think we should add performance regression tests. At the least, we could ask people making big PRs to show the output of `./scripts/check-queries.lua <parser>` before and after.

(I'll make a PR allowing `check-queries` to take a single language argument; I've wanted this for a while.)

queries/d/highlights.scm
```diff
@@ -1,275 +1,373 @@
...
+  (#lua-match? @comment.documentation "^///[^/]"))
+
+((comment) @comment.documentation
+  (#lua-match? @comment.documentation "^///$"))
+
+((comment) @comment.documentation
+  (#lua-match? @comment.documentation "^/[*][*][^*].*[*]/$"))
+
+((comment) @comment.documentation
+  (#lua-match? @comment.documentation "^/[+][+][^+].*[+]/$"))
```
ObserverOfTime, line 373: Wouldn't it be better to merge these?

amaanq: hmm, how so with lua match? `[+*]` could work but that'd allow for `/*+`

ObserverOfTime: Can you test the performance with multiple lua-match queries vs 1 or 2 match queries?

clason:
@amaanq could you rebase? I'm happy with a quick `check-queries d` A/B test not showing significant regressions -- let's not block useful PRs forever with minor quibbles; it's fine to leave some improvements for followup PRs (by those that care).

ObserverOfTime (approved):
Honestly fine to merge as is.

clason:
Needs a rebase for the check, though.

amaanq:
Yeah sorry I wasn't sure how to test for perf between 2 lua matches and 1 vim match, I'm not great at all that lua stuff :grimacing:

amaanq:
Wow so I think anchored wildcard named node patterns are extremely expensive if there's a lot of options I think

e.g. something like `(_) .)`

Before:
```
5. 0.03ms d locals
4. 9.72ms d indents
3. 10.21ms d injections
2. 11.16ms d folds
1. 23.21ms d highlights
```

After:
```
5. 4.54ms d indents
4. 4.57ms d folds
3. 6.62ms d injections
3. 34.76ms d locals
1. 198.42ms d highlights
```

I changed one pattern, namely this one
```
(type
  .
  (_)
  .
  (identifier) @variable.member
  .
  (_))
  ```
  
  to this 
  ```
(type
  "."
  (identifier) @variable.member)
 ```
 
 And now highlights are down to 70-80ms. Maybe we should have a warning if they go above 100ms per query, D's grammar is just complex in both cases and this one definitely feels more full, so some patterns probably just have more possibilities.

clason:
We removed the worst offenders, so now everything is under 500ms. Getting below 100ms seems tedious, given how many currently trip that bar.

ljmf00:
Thanks for taking it :heart: . I didn't had much time to work on it and I'm glad you grab it.

PR #5991, opened 2024-01-26
feat(tests): allow check-queries to take list of parsers

clason:
This allows contributors to run the tests on some languages only via

`./scripts/check-queries.lua foo bar baz`

Makes it more feasible to ask them for before/after timings on PRs that might have a performance impact.

### DIFF
--- a/scripts/check-queries.lua
+++ b/scripts/check-queries.lua
@@ -46,8 +46,8 @@ end
 
 local function do_check()
   local timings = {}
-  local parsers = require("nvim-treesitter.info").installed_parsers()
   local queries = require "nvim-treesitter.query"
+  local parsers = #_G.arg > 0 and { unpack(_G.arg) } or require("nvim-treesitter.info").installed_parsers()
   local query_types = queries.built_in_query_groups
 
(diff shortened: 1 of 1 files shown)


scripts/check-queries.lua
```diff
@@ -46,8 +46,8 @@ end
 
 local function do_check()
   local timings = {}
-  local parsers = require("nvim-treesitter.info").installed_parsers()
   local queries = require "nvim-treesitter.query"
+  local parsers = #_G.arg > 0 and { unpack(_G.arg) } or require("nvim-treesitter.info").installed_parsers()
```
ObserverOfTime, line 50:
```suggestion
  local parsers = #_G.arg > 0 and _G.arg or require("nvim-treesitter.info").installed_parsers()
```
Would this not work?

clason: No, because `_G.arg[0]` is the script name (which `unpack` drops).

ObserverOfTime (approved):

clason:
(Note: I'm further cleaning up that script on `main`, but `master` gets minimal changes only.)

PR #5994, opened 2024-01-26
feat: register aliases defined for set-lang-from-info-string

ribru17:
~~There are several [Chart.js](https://github.com/chartjs/Chart.js) Markdown previewing plugins, and they all (or all I have seen) take JSON data passed to a `chart` codeblock to generate the graphs. Since the chart info string isn't recognized in any way, it would be nice to alias it to JSON to get nice injections still.~~

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -33,6 +33,8 @@ for ft, lang in pairs {
   py = "python",
   erl = "erlang",
   typ = "typst",
+  pl = "perl",
+  uxn = "uxntal",
 } do
   ts.language.register(lang, ft)
 end


lucario387:
Can't you just `vim.treesitter.language.register("json", {"chart"})` :thinking:

clason:
Yes, going forward, that's the way this should be handled. (But it's Neovim 0.10 only.)

This is a general decision on when we freeze `master` completely. I am mostly worried about changes that conflict with rebasing `main`, though, and this is a change to something I've already ripped out completey.

ribru17:
Thank you both.

clason:
> I am mostly worried about changes that conflict with rebasing main, though, and this is a change to something I've already ripped out completely.

Which means it's not something I'd want to block necessarily, if the argument for a short-term fix is strong enough.

ribru17:
I see. I will reopen this then, and if anyone has an argument against then they can shut it down.

clason:
Just be aware that it'll be gone again in 1.0.

ribru17:
Should've made a new PR for this in hindsight (I changed my mind on what I was doing after some iteration):

This now ports the rest of the aliases in `query_predicates.lua` to be recognized when using the new injection styles. Also I gave up on the `chart` recognition because sadly obsidian treats them as yaml, and regular markdown treats them as JSON. It will have to be up to user config

clason:
Looks like I missed Perl 🤦 (Couldn't care less about uxntal...)

PR #5999, opened 2024-01-27
fix(notification): use `vim.notify` instead of `print` for installs

chrisgrieser:
The utils contain a function for `vim.notify`, but it is not used when installing parsers.
https://github.com/nvim-treesitter/nvim-treesitter/blob/b4138891b3454beeb14eef171c91c92377fcd715/lua/nvim-treesitter/utils.lua#L7-L14

This PR fixes that, so that `vim.notify` is consistently used.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -39,7 +39,7 @@ local function reset_progress_counter()
 end
 
 local function get_job_status()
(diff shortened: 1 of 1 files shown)


clason:
Thank you, but these functions have been completely rewritten for 1.0 on the main branch, and we do not want to make conflicting changes.

PR #6001, opened 2024-01-27
fix(incremental-selection): use cmd instead of :lua for command

MariaSolOs:
Fixes the issue described in https://github.com/neovim/neovim/issues/27234.

This is the solution proposed by @luukvbaal in https://github.com/neovim/neovim/issues/27234#issuecomment-1913332244.

### DIFF
--- a/lua/nvim-treesitter/incremental_selection.lua
+++ b/lua/nvim-treesitter/incremental_selection.lua
@@ -148,7 +148,7 @@ function M.attach(bufnr)
         mode = "x"
         -- We need to move to command mode to access marks '< (visual area start) and '> (visual area end) which are not
         -- properly accessible in visual mode.
-        rhs = string.format(":lua require'nvim-treesitter.incremental_selection'.%s()<CR>", funcname)
+        rhs = string.format("<cmd>lua require'nvim-treesitter.incremental_selection'.%s()<CR>", funcname)
       end
       vim.keymap.set(
         mode,


luukvbaal:
(untested by me)

clason:
Full disclosure: this whole module will be gone in the next version anyway.

clason (approved):

MariaSolOs:
Oh oh there's no error anymore but now the selection is kinda broken (it doesn't increment/decrement scope-wise).

It's fine though. No point in fixing it if this will soon be dead code.

clason:
Sounds like it needs a proper rewrite (as a standalome module)...?

luukvbaal:
FWIW (I never used the mapping before), both `<Esc><Cmd>` and `:<C-U>` seem to work.

jtledon:
Before submitting a bug report, I thought I would ask here, is anyone else having problems leaving visual mode (normal, line or block) using <Esc> after this commit was merged? I narrowed it down to this commit as the breaking change for me. In the comment above the changed piece of code, it seems to mention a workaround due to being in visual mode

wookayin:
@jtledon Feel free to open a new issue describing the problem. By the way can you please be specific in the problem you're having after this commit -- how exactly did it break?

jtledon:
https://github.com/nvim-treesitter/nvim-treesitter/assets/81050146/aa999f70-b23a-4ec6-9ca0-d1d36f6bc88c

I've been following both this thread and the revert (which is now the second attempt at the fix). You can see in the video, with the original string using `:` as the command key, when in any of the visual modes (I only demonstrated visual line in the video) I can press the `<esc>` key and have it return me back to normal mode as expected. With the first fix, and now second proposed fix, pressing `<esc>` does nothing, and I have to either go into normal visual mode and press v to exit, or cmd mode and then press `<esc>` to exit.

MariaSolOs:
> esc-from-visual.mp4 
> I've been following both this thread and the revert (which is now the second attempt at the fix). You can see in the video, with the original string using `:` as the command key, when in any of the visual modes (I only demonstrated visual line in the video) I can press the `<esc>` key and have it return me back to normal mode as expected. With the first fix, and now second proposed fix, pressing `<esc>` does nothing, and I have to either go into normal visual mode and press v to exit, or cmd mode and then press `<esc>` to exit.

Hmmm I don't seem to be able to reproduce this. Could you please share your incremental selection mappings?

jtledon:
I can definitely share my TS config, and specifically my incremental selection mappings.

I was playing with it a bit more, and it turns out I was right to not file an issue. When running without my config `nvim -u NORC`, any of the 3 proposed solutions works. Not sure what in my specific config is causing the issue though.

jtledon:
```lua
return {
    event = { "BufReadPre", "BufNewFile" },
    build = ":TSUpdate",
    dependencies = {
        "nvim-treesitter/nvim-treesitter-textobjects",
        "windwp/nvim-ts-autotag",
    },
    config = function ()
        local ts_config = require("nvim-treesitter.configs")
        ts_config.setup({
            ensure_installed = {
                "c",
                "cpp",
                ...
            },

            sync_install = false,
            auto_install = true,

            highlight = {
                enable = true,

                disable = function(lang, buf)
                    local max_filesize = 100 * 1024 -- 100 KB
                    local ok, stats = pcall(vim.loop.fs_stat, vim.api.nvim_buf_get_name(buf))
                    if ok and stats and stats.size > max_filesize then
                        return true
                    end
                end,

                additional_vim_regex_highlighting = false,
            },

            incremental_selection = {
                enable = true,
                keymaps = {
                    init_selection = '<c-]>',
                    node_incremental = '<c-]>',
                    scope_incremental = '<c-}>',
                    node_decremental = '<c-[>',
                },
            },

            textobjects = {
                select = {
                    enable = true,
                    lookahead = true, -- Automatically jump forward to textobj, similar to targets.vim
                    keymaps = {
                        -- You can use the capture groups defined in textobjects.scm
                        ['aa'] = '@parameter.outer',
                        ['ia'] = '@parameter.inner',
                        ['af'] = '@function.outer',
                        ['if'] = '@function.inner',
                        ['ac'] = '@class.outer',
                        ['ic'] = '@class.inner',
                    },
                },
                move = {
                    enable = true,
                    set_jumps = true, -- whether to set jumps in the jumplist
                    goto_next_start = {
                        [']m'] = '@function.outer',
                        [']]'] = '@class.outer',
                    },
                    goto_next_end = {
                        [']M'] = '@function.outer',
                        [']['] = '@class.outer',
                    },
                    goto_previous_start = {
                        ['[m'] = '@function.outer',
                        ['[['] = '@class.outer',
                    },
                    goto_previous_end = {
                        ['[M'] = '@function.outer',
                        ['[]'] = '@class.outer',
                    },
                },
            },

            autotag = {
                enable = true,
            },
        })

        -- register pde files as processing filetype
        vim.treesitter.language.register("java", "processing")

        vim.treesitter.language.register("markdown", "mdx")

        vim.api.nvim_create_user_command("TSInfo", function() print(vim.treesitter.get_parser():lang()) end, { desc = "Print out the treesitter parser for the current buffer" })
        vim.keymap.set("n", '<leader>PG', vim.treesitter.inspect_tree, { noremap = true, desc = 'Treesitter [P]lay [G]round, prior to being built-in' })
    end
}

```

jtledon:
@MariaSolOs I wasn't. That video was strictly using visual-line mode (`V`), but it also applies to visual-block (`C-v`) and normal visual mode

MariaSolOs:
> @MariaSolOs I wasn't. That video was strictly using visual-line mode (`V`), but it also applies to visual-block (`C-v`) and normal visual mode

I think we're talking about different issues then? The bug I'm describing is the visual selection using the increment/decrement TS commands not working anymore.

MariaSolOs:
@jtledon are you using nightly Neovim? Did you also make the other change in lines 26-27 as described [here](https://github.com/nvim-treesitter/nvim-treesitter/pull/6014#issuecomment-1917690081)?

jtledon:
@MariaSolOs Oh, no I agree that is the issue that you have, I was simply saying that the proposed fixes from this thread and the revert break my ability to leave visual mode by pressing escape. I was simply commenting here to see if this was a problem specific to me, and if so, try to understand why this change might be having that impact

I did make the changes to 26-27, and it didn't seem to have an impact on the functionality of any of the three options, but I admittedly don't have the context to understand what they were supposed to be doing.

I've tried it on both the latest nightly and latest tagged (v0.9.5)

MariaSolOs:
> @MariaSolOs Oh, no I agree that is the issue that you have, I was simply saying that the proposed fixes from this thread and the revert break my ability to leave visual mode by pressing escape. I was simply commenting here to see if this was a problem specific to me, and if so, try to understand why this change might be having that impact

Interesting, solution 1 (the one in this PR) didn't properly fix incremental selection for me (I again apologize for my eagerness in providing an incomplete fix). https://github.com/nvim-treesitter/nvim-treesitter/pull/6014 does fix incremental selection, and `<esc>` exits visual mode as expected (with my setup).

Despite this maybe something specific to your setup, I would still like to understand it better because I don't want to break anyone's workflow either :)

jtledon:
@MariaSolOs I completely understand, I'm going to spend some time now going from a fully removed TS, building up my config to try and find what breaks it. I'm glad the newly proposed fix works for your setup though!

jtledon:
@MariaSolOs I found the issue in my setup. For whatever reason, my specific mapping for `node_decremental` which was `<C-[>` broke everything for me. If I had to guess, it would be because `[` is the start of lua's string literal `[[`, and that was in some was interfering with the command string? I really have no idea. Once I swapped that remapping, it seemed to work perfectly for me. 
https://github.com/nvim-treesitter/nvim-treesitter/assets/81050146/12be1e57-2c77-4304-bd05-d1d769792b6c

MariaSolOs:
@jtledon https://vi.stackexchange.com/questions/24852/how-to-remove-the-mapping-from-ctrl-to-escape might be a good read for you ;)

jtledon:
Yeaa, it might be worth reading about more of these haha. If nothing else I learned something new, thank you!

pidgeon777:
This also fixed my issue, incremental selection wasn't working anymore using LazyVim (at first it visually selected the right scope, then a subsequent command would select all of the code, or a completely different section, apparently in a random way).

PR #6003, opened 2024-01-28
ci: use token for automated PRs

clason:
This should allow automated `update-lockfile` and  `update-readme` PRs to run CI automatically (following https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens, but using the new official Github token action). 

It's possible to exempt the bot from the branch protection rule; this would allow fully automated (non-breaking) lockfile updates. In this case, the update-readme workflow could also push straight to master so we don't have to deal with the noise of a PR.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,8 +1,9 @@
 name: Linting and style checking
 
 on:
-  push:
   pull_request:
+    branches:
+      - "master"
 
 jobs:
   luacheck:
--- a/.github/workflows/update-lockfile.yml
+++ b/.github/workflows/update-lockfile.yml
@@ -14,6 +14,12 @@ jobs:
         with:
           ref: master
 
+      - uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ vars.TOKEN_ID }}
+          private-key: ${{ secrets.TOKEN_PRIVATE_KEY }}
+
       - name: Prepare
         env:
           NVIM_TAG: stable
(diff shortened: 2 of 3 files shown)


.github/workflows/lint.yml
```diff
@@ -2,7 +2,11 @@ name: Linting and style checking
 
 on:
   push:
+    branches:
+      - "master"
```
dundargoc, line 6: I'm a bit confused why this is needed for push on master. Isn't linting only useful before merging a PR?

clason: It was there when I found it... I'm not sure why "push" is needed at all? (Assuming you don't push straight to master -- which sometimes happens...)

.github/workflows/update-lockfile.yml
```diff
@@ -32,21 +38,20 @@ jobs:
...
-          git commit -m "Update parsers: $UPDATED_PARSERS" || echo 'No commit necessary!'
-          git clean -xf
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v5
         with:
+          token: ${{ steps.app-token.outputs.token }}
+          commit-message: "Update parsers: ${{ env.UPDATED_PARSERS }}"
           title: "Update lockfile.json: ${{ env.UPDATED_PARSERS }}"
+          body: "[beep boop](https://github.com/peter-evans/create-pull-request)"
```
dundargoc, line 50: Guessing the "beep boop" is intentional and not something you intended to remove? :)

clason: It is, although I'm not 100% I'll leave it in ;) (But I can't think of any more useful body, without adding more scripting for, e.g., changelogs.)

.github/workflows/lint.yml
dundargoc: Even in that case it's too late since it's already merged. I think you can remove the entire push section tbh.

.github/workflows/update-lockfile.yml
dundargoc: Eh, I'm perfectly fine with it as is.

.github/workflows/lint.yml
clason: For both actions?

dundargoc: Hmm, I don't think so. The `lint.yml` workflow doesn't need it but it looks like `update-readme.yml` needs it (since it uses a push to master to check if readme should be updated.)

clason: Gotcha; I was thinking of the tests workflows.

dundargoc: Could be worth it since PRs can use the cache from the master branch workflow. Aka not guaranteed useless but haven't benchmarked or anything.

dundargoc (commented):
Looks good. I haven't tested it myself so take the approval with some salt, but I can't see anything obviously egregious.

clason:
I tested the lockfile workflow and it works as expected: https://github.com/nvim-treesitter/nvim-treesitter/pull/6013

dundargoc:
Well technically it's the schedule trigger that was the problem originally as it's the only trigger that is not done by a human, so you'd need to check for that as well. But just being pedantic because I can be lol, I don't see why this shouldn't work for that as well :)

PR #6014, opened 2024-01-28
fix(incremental-selection): second attempt at fixing command

MariaSolOs:
I was a bit too eager with #6001. As mentioned in there, the change doesn't fully fix the issue anyway (it does get rid of the error though) and it also seems to cause other problems (like https://github.com/nvim-treesitter/nvim-treesitter/pull/6001#issuecomment-1913719917).

This solution is the one proposed by @monkoose

### DIFF
--- a/lua/nvim-treesitter/incremental_selection.lua
+++ b/lua/nvim-treesitter/incremental_selection.lua
@@ -23,8 +23,8 @@ end
 -- The range starts with 1 and the ending is inclusive.
 ---@return integer, integer, integer, integer
 local function visual_selection_range()
-  local _, csrow, cscol, _ = unpack(vim.fn.getpos "'<") ---@type integer, integer, integer, integer
-  local _, cerow, cecol, _ = unpack(vim.fn.getpos "'>") ---@type integer, integer, integer, integer
+  local _, csrow, cscol, _ = unpack(vim.fn.getpos "v") ---@type integer, integer, integer, integer
+  local _, cerow, cecol, _ = unpack(vim.fn.getpos ".") ---@type integer, integer, integer, integer
 
   local start_row, start_col, end_row, end_col ---@type integer, integer, integer, integer
 
@@ -146,9 +146,7 @@ function M.attach(bufnr)
         rhs = M[funcname]
       else
         mode = "x"
-        -- We need to move to command mode to access marks '< (visual area start) and '> (visual area end) which are not
-        -- properly accessible in visual mode.
-        rhs = string.format(":lua require'nvim-treesitter.incremental_selection'.%s()<CR>", funcname)
+        rhs = M[funcname] ---@type function
       end
       vim.keymap.set(
         mode,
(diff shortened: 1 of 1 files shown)


clason:
Sorry, we don't look back (sharks don't look back, because sharks don't have necks).

Is there no way to fix this? The whole `keymap` thing is sus, of course, but can't we export a Lua function that people on master can map directly and avoid the issue?

MariaSolOs:
> Sorry, we don't look back (sharks don't look back, because sharks don't have necks).

Sad. They won't ever be able to enjoy a good neck massage.

> Is there no way to fix this? The whole `keymap` thing is sus, of course, but can't we export a Lua function that people on master can map directly and avoid the issue?

Unfortunately this surpasses my Treesitter knowledge.

zeertzjq:
Does adding `<C-U>` after `:` not work?

monkoose:
Is there a reason why getpos('v') isn't used instead of `'<` and `'>`? So hacks to enter command line could be dropped.

clason:
It has nothing to do with treesitter, actually; it's just about the (horrible, _horrible_) way that nvim-treesitter customizes mappings.

(Which, to be fair, predated Nvim's API for mapping Lua functions directly by quite a bit. Unfortunately, the design was such that adapting it was infeasible.)

clason:
> Is there a reason why `getpos('v')` isn't used instead of `'<` and `'>`? So hacks to enter command line could be dropped.

Worth a try. PR welcome!

MariaSolOs:
As a shark, I'll close this PR for now.

clason:
> Does adding `<C-U>` after `:` not work?

You could try this first, though?

MariaSolOs:
> > Does adding `<C-U>` after `:` not work?
> 
> You could try this first, though?

True, it's worth a try :)

MariaSolOs:
> Does adding <C-U> after : not work?

Okay no, it doesn't work. The error still appears, and the scope isn't correctly incremented.

clason:
Ah, OK, pity. Are you game for more tests?

MariaSolOs:
> Ah, OK, pity. Are you game for more tests?

Always ;)

MariaSolOs:
> Then try @monkoose's suggestion:
> 
> 1. replace here
>    https://github.com/nvim-treesitter/nvim-treesitter/blob/5f950cdcb82d2cf74e6b30338897b3a4897f52b3/lua/nvim-treesitter/incremental_selection.lua#L26-L27
>    
>    `"'<"` and `"'>"` by `"v"` and `"."`, respectively (compare `range_from_selection` in `vim.lsp.buf`);
> 2. replace here  https://github.com/nvim-treesitter/nvim-treesitter/blob/5f950cdcb82d2cf74e6b30338897b3a4897f52b3/lua/nvim-treesitter/incremental_selection.lua#L151
>    
>    the `string.format(...)` by `M[funcname]`.

It worked!

sp3ctum:
Nice!
Not sure about the shark stuff (feeling optimistic though) but this does seem to fix the issue for me. Thanks so much, this was driving me nuts. 😄

PR #5594, opened 2023-10-30
feat(gnuplot): added gnuplot parser and queries

dpezto:

PR #6018, opened 2024-01-30
fix(format): handle list with 300 nodes

lucario387:
Cut down on the amount of captures required for a file to support this case.
Related: https://github.com/nvim-treesitter/nvim-treesitter/pull/5594#discussion_r1467326738

### DIFF
--- a/queries/gnuplot/highlights.scm
+++ b/queries/gnuplot/highlights.scm
@@ -165,8 +165,7 @@
     "background"
     "rounded"
     "butt"
-  ]*
-     @attribute)
+  ]* @attribute)
 
 ; (t_canvas)
 ; (t_cgm)
@@ -288,7 +287,6 @@
 ] @variable.member
 
 ; Workaround because formatter cannot handle 300 list nodes
-; format-ignore
 [
   (angles)
   (clip)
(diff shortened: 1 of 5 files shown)


queries/gnuplot/highlights.scm
```diff
@@ -165,8 +165,7 @@
     "background"
     "rounded"
     "butt"
-  ]*
-     @attribute)
+  ]* @attribute)
```
lucario387, line 168: This is intentional :). The previous one looked wrong

scripts/format-queries.lua
```diff
@@ -106,25 +106,19 @@ local format_queries = [[
...
   .
   "]" @format.remove)
 ;; [ ... ] @capture1 @capture2
-(list
-  (capture) @format.prepend-space)
 ;; Append newlines for nodes inside the list
 (list
   (_) @format.append-newline
-  (#not-has-type? @format.append-newline capture))
+  (#not-has-type? @format.append-newline capture quantifier))
```
lucario387, line 121: Just to be clear, this line is responsible for the `?` and `*` after a list to add a newline. Oversight happened

clason (approved):

Issue #6025, opened 2024-01-31
Regression in PR #6014: error "rhs: expected string|function, got nil" on incremental_selection 


GR3YH4TT3R93:
### Describe the bug

When opening a file with nerd-tree or Dashboard errors appear and ts highlighting doesn't work

### To Reproduce

1.) Install Neo-tree or Dashboard-nvim
2.) Checkout commit 5f950cdcb82d2cf74e6b30338897b3a4897f52b3
3.) Restart Neovim
4.) Use Neo-tree or Dashboard to navigate to a file (should be working as expected)
5.) Update to commit ba6454783493ac3a5dd209c25e491640b07bd8de (#6014)
6.) Restart neovim
7.) Use Neo-tree or Dashboard to navigate to a file
8.) An error should appear (the error will be different based on whether you use Dashboard or Neo-tree but only occurs after this commit)

### Expected behavior

The file to open without errors and highlighting to work.

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- OK `tree-sitter` found 0.20.9 (parser generator, only needed for :TSInstallFromGrammar)
- OK `node` found v20.2.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: clang version 17.0.6
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "aarch64",
  release = "5.10.168-android12-9-00003-gc873b6b86254-ab10823632",
  sysname = "Linux",
  version = "#1 SMP PREEMPT Mon Sep 18 13:31:53 UTC 2023"
} ~

Parser/Features         H L F I J
  - bash                ✓ ✓ ✓ . ✓
  - c                   ✓ ✓ ✓ ✓ ✓
  - css                 ✓ . ✓ ✓ ✓
  - go                  ✓ ✓ ✓ ✓ ✓
  - html                ✓ ✓ ✓ ✓ ✓
  - json                ✓ ✓ ✓ ✓ .
  - lua                 ✓ ✓ ✓ ✓ ✓
  - markdown            ✓ . ✓ ✓ ✓
  - query               ✓ ✓ ✓ ✓ ✓
[...]

PR #6029, opened 2024-02-01
refactor(inc-selection): simplify keymap and add error messages

MariaSolOs:
Follow up of #6014. After looking at the code I realized that the if-statement looked a bit odd. I also added a notification message for better error messages in cases like #6025.

EDIT: Also including the changes from #5963

### DIFF
--- a/lua/nvim-treesitter/incremental_selection.lua
+++ b/lua/nvim-treesitter/incremental_selection.lua
@@ -5,6 +5,7 @@ local ts_utils = require "nvim-treesitter.ts_utils"
 local locals = require "nvim-treesitter.locals"
 local parsers = require "nvim-treesitter.parsers"
 local queries = require "nvim-treesitter.query"
+local utils = require "nvim-treesitter.utils"
 
 local M = {}
 
@@ -138,22 +139,14 @@ function M.attach(bufnr)
   local config = configs.get_module "incremental_selection"
   for funcname, mapping in pairs(config.keymaps) do
     if mapping then
-      ---@type string, string|function
-      local mode, rhs
-      if funcname == "init_selection" then
-        mode = "n"
-        ---@type function
-        rhs = M[funcname]
+      local mode = funcname == "init_selection" and "n" or "x"
+      local rhs = M[funcname] ---@type function
+
+      if not rhs then
+        utils.notify("Unknown keybinding: " .. funcname .. debug.traceback(), vim.log.levels.ERROR)
       else
-        mode = "x"
-        rhs = M[funcname] ---@type function
+        vim.keymap.set(mode, mapping, rhs, { buffer = bufnr, silent = true, desc = FUNCTION_DESCRIPTIONS[funcname] })
       end
-      vim.keymap.set(
-        mode,
-        mapping,
-        rhs,
-        { buffer = bufnr, silent = true, noremap = true, desc = FUNCTION_DESCRIPTIONS[funcname] }
-      )
     end
   end
 end
--- a/lua/nvim-treesitter/incremental_selection.lua
+++ b/lua/nvim-treesitter/incremental_selection.lua
@@ -155,10 +155,10 @@ function M.detach(bufnr)
   local config = configs.get_module "incremental_selection"
   for f, mapping in pairs(config.keymaps) do
     if mapping then
-      if f == "init_selection" then
-        vim.keymap.del("n", mapping, { buffer = bufnr })
-      else
-        vim.keymap.del("x", mapping, { buffer = bufnr })
+      local mode = f == "init_selection" and "n" or "x"
+      local ok, err = pcall(vim.keymap.del, mode, mapping, { buffer = bufnr })
+      if not ok then
+        utils.notify(string.format('%s "%s" for mode %s', err, mapping, mode), vim.log.levels.ERROR)
       end
     end
   end


lucario387 (approved):

lua/nvim-treesitter/incremental_selection.lua
```diff
@@ -138,16 +138,13 @@ function M.attach(bufnr)
...
-        mode = "x"
-        rhs = M[funcname] ---@type function
+      local mode = funcname == "init_selection" and "n" or "x"
+      local rhs = M[funcname] ---@type function
+      
+      if not rhs then
+        utils.notify("Unknown keybinding: " .. funcname)
       end
+      
       vim.keymap.set(
```
lucario387, line 148: the keymap.set should be in the else block

MariaSolOs: I thought about it, since the `vim.keymap.set` would fail anyway. That being said I think that the full stack trace is helpful? I don't have strong feelings about it though.

lua/nvim-treesitter/incremental_selection.lua
```diff
@@ -138,16 +138,13 @@ function M.attach(bufnr)
...
-        ---@type function
-        rhs = M[funcname]
-      else
-        mode = "x"
-        rhs = M[funcname] ---@type function
+      local mode = funcname == "init_selection" and "n" or "x"
+      local rhs = M[funcname] ---@type function
+      
+      if not rhs then
+        utils.notify("Unknown keybinding: " .. funcname)
```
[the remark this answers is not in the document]
MariaSolOs: Oops good catch!

lua/nvim-treesitter/incremental_selection.lua
```diff
@@ -138,16 +139,14 @@ function M.attach(bufnr)
...
-        rhs = M[funcname]
-      else
-        mode = "x"
-        rhs = M[funcname] ---@type function
+      local mode = funcname == "init_selection" and "n" or "x"
+      local rhs = M[funcname] ---@type function
+
+      if not rhs then
+        utils.notify("Unknown keybinding: " .. funcname, vim.log.levels.ERROR)
+        return
```
wookayin, line 147: Isn't `return` too aggressive? Keymapping sets in the previous iterations of the for loop will still remain. I suggest: just skip `vim.keymap.set` (putting in the `else` block) or continue. Oh well, it's Lua :) ~~`::continue::`~~

MariaSolOs: I mean, at this point the module configuration is just wrong. We can have a more "graceful" setup but things are going to be whack anyway...

wookayin: Yeah that also sounds good, thanks!

MariaSolOs: Thanks to YOU <3

lua/nvim-treesitter/incremental_selection.lua
```diff
@@ -138,22 +139,19 @@ function M.attach(bufnr)
...
-      local mode, rhs
-      if funcname == "init_selection" then
-        mode = "n"
-        ---@type function
-        rhs = M[funcname]
+      local mode = funcname == "init_selection" and "n" or "x"
+      local rhs = M[funcname] ---@type function
+
+      if not rhs then
+        utils.notify("Unknown keybinding: " .. funcname, vim.log.levels.ERROR)
```
wookayin, line 146: Not mandatory, but this might be slightly more helpful:

```suggestion
        utils.notify("Unknown keybinding: " .. funcname .. "\n" .. debug.traceback(), vim.log.levels.ERROR)
```

wookayin (approved):
LGTM, and thanks a lot for the follow-up. 

One minor comment: this is somehow related to #5963 -- detach will also raise error messages upon an invalid configuration, but this should be already fine (@clason didn't want to merge #5963 given its current form) and should not block this PR.

lua/nvim-treesitter/incremental_selection.lua
```diff
@@ -138,22 +139,19 @@ function M.attach(bufnr)
...
+      if not rhs then
+        utils.notify("Unknown keybinding: " .. funcname .. debug.traceback(), vim.log.levels.ERROR)
       else
-        mode = "x"
-        rhs = M[funcname] ---@type function
+        vim.keymap.set(
+          mode,
+          mapping,
+          rhs,
+          { buffer = bufnr, silent = true, noremap = true, desc = FUNCTION_DESCRIPTIONS[funcname] }
```
clason, line 152:
```suggestion
          { buffer = bufnr, silent = true, desc = FUNCTION_DESCRIPTIONS[funcname] }
```
(not a thing with `vim.keymap.set`, and `remap` defaults to `false` anyway)

clason:
> Actually @clason didn't want to merge https://github.com/nvim-treesitter/nvim-treesitter/pull/5963 given its current form); so should not block this PR.

For the record, I'm fine to include it in this PR, while we're already touching the file. It still is going to be deleted for nvim-treesitter 1.0 unless someone volunteers to rewrite it into a standalone module.

przepompownia:
The good news is that my loose proposition (#5963) led to fix the issue (even for a few weeks of use). Thank you all.

PR #6043, opened 2024-02-02
feat(faust): add new parser tree-sitter-faust

khiner:


queries/faust/highlights.scm
```diff
@@ -0,0 +1,317 @@
+; See: https://tree-sitter.github.io/tree-sitter/syntax-highlighting#highlights
```
clason, line 1: I think it's better to remove these comments here in our repo (which already includes the guide -- which may change over time!)

clason (commented):
How about injections (at least comments), indents, folds?

queries/faust/highlights.scm
khiner: Good call, removed these

khiner:
Can I add injections, indents, and folds later?

queries/faust/highlights.scm
```diff
@@ -0,0 +1,242 @@
...
+  "["
+  "]"
+  "{"
+  "}"
+] @punctuation.bracket
+
+; Comments
+(comment) @comment
+
+; @tag           ; XML-style tag names (and similar)
```
clason, line 231: missed some ;)

khiner: oops! :)

khiner: fixed

clason:
> Can I add injections, indents, and folds later?

Sure, but no time like the present ;) Things put off are things left off...

queries/faust/highlights.scm
```diff
@@ -0,0 +1,239 @@
+; Identifiers
+(process) @variable.builtin
+
+(parameters
+  (identifier) @variable.parameter)
+
+(identifier) @variable
```
amaanq, line 7: flip these, you'll never have parameters highlighted properly (queries have a last-wins precedence)

queries/faust/highlights.scm
```diff
@@ -0,0 +1,239 @@
...
+  filename: (string)) @string.special.path
+
+(soundfile
+  filename: (string)) @string.special.path
+
+(component
+  filename: (string)) @string.special.path
+
+(library
+  filename: (string)) @string.special.path
```
amaanq, line 20:
```suggestion
(_
  filename: (string)) @string.special.path
```

queries/faust/highlights.scm
```diff
@@ -0,0 +1,239 @@
...
+  (lshift)
+  (rshift)
+  (lt)
+  (le)
+  (gt)
+  (ge)
+  (eq)
+  (neq)
+  (delay)
+  (one_sample_delay)
```
amaanq, line 170: are these operators like `+` or words literally like `add`? If the latter, these should be `@keyword.operator`

khiner: They're operators lik `+'`.. Except 'xor' - I thought about that, should `xor` be `@keyword.operator` instead?

queries/faust/highlights.scm
khiner: Ahh that explains the issue I was seeing with parameters - thanks!!

queries/faust/highlights.scm
khiner: ❤️ exactly what I was looking for, thanks!

khiner: That reminds me - can I somehow use list brackets `[...]` to simplify
```
(file_import filename: (string)) @string.special.path
(soundfile filename: (string)) @string.special.path
(component filename: (string)) @string.special.path
(library filename: (string)) @string.special.path
```
 - does it work with parent rules?

queries/faust/highlights.scm
amaanq: sure

queries/faust/highlights.scm
amaanq: nah, sorry - it'd be difficult to know which rule you're targeting then

khiner: Ok that's what I figured, thanks

khiner:
@amaanq I addressed all feedback and added injections. Thanks for the review!

wookayin:
It'd be great if you could follow the conventional commit messsage formats (please have a look at existing commits and PRs). e.g. `feat(faust): ...`

khiner:
> It'd be great if you could follow the conventional commit messsage formats (please have a look at existing commits and PRs). e.g. `feat(faust): ...`

@wookayin Thanks, updated

queries/faust/highlights.scm
```diff
@@ -0,0 +1,214 @@
...
+  "("
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
+] @punctuation.bracket
+
+; Comments
+(comment) @comment
```
amaanq, line 204:
```suggestion
(comment) @comment @spell
```

queries/faust/highlights.scm
```diff
@@ -0,0 +1,214 @@
...
+(parameters
+  (identifier)) @variable.parameter
+
+; Literals
+(_
+  filename: (string)) @string.special.path
+
+(function_names) @string.special
+
+(documentation) @string.documentation
```
amaanq, line 15:
```suggestion
(documentation) @string.documentation @spell
```

queries/faust/highlights.scm
```diff
@@ -0,0 +1,214 @@
...
+    "~")
+  (sequential
+    ":")
+  (split
+    "<:")
+  (merge
+    ":>")
+  (parallel
+    ",")
+] @operator
```
amaanq, line 159: this won't target the operator inside, but rather the outer nodes (recursive/sequential), is there a reason you just don't put the literals with all the others right above it? like `~`, `<:`, etc

khiner: Good catch, it should just be with the others

khiner: Actually, I think I do want to only match these inside their respective parent nodes. How would I accomplish that?

khiner: ah I think I have it
```
(recursive "~" @operator)
(sequential ":" @operator)
(split "<:" @operator)
(merge ":>" @operator)
(parallel "," @operator)
```

khiner: pushed

queries/faust/highlights.scm
khiner: pushed

queries/faust/highlights.scm
khiner: pushed

queries/faust/highlights.scm
amaanq: Yeah exactly, if you have other instances where these literals can appear *in a different context*, then the specific query makes total sense. But otherwise, just use the literal as is

khiner:
The test failures seem unrelated?
```
Fail	||	indent Swift: whole file: ./statements.swift	
            vim/_editor.lua:0: command line..script nvim_exec2() called at command line:0: Vim(edit):E576: Error while reading ShaDa file: there is an item at position 15819 that must not be there: Missing items are for internal uses only
```

queries/faust/highlights.scm
```diff
@@ -0,0 +1,216 @@
...
+; Keywords
+[
+  (par)
+  (seq)
+  (sum)
+  (prod)
+] @keyword.repeat
+
+(file_import
+  "import") @keyword.import
```
ObserverOfTime, line 172:
```suggestion
  "import" @keyword.import)
```

queries/faust/highlights.scm
```diff
@@ -0,0 +1,216 @@
...
+(process) @variable.builtin
+
+(parameters
+  (identifier)) @variable.parameter
+
+; Literals
+(_
+  filename: (string)) @string.special.path
+
+(function_names) @string.special
```
ObserverOfTime, line 13: Not `@function`?

queries/faust/highlights.scm
```diff
@@ -0,0 +1,216 @@
...
+  (string)
+  (fstring)
+] @string
+
+(real) @number.float
+
+[
+  (int)
+  (real)
+] @number
```
ObserverOfTime, line 27:
```suggestion
(int) @number
```

queries/faust/highlights.scm
```diff
@@ -0,0 +1,216 @@
...
+
+; Types
+(global_metadata
+  key: (identifier)) @property
```
ObserverOfTime, line 31:
```suggestion
  key: (identifier) @variable.member)
```

queries/faust/highlights.scm
```diff
@@ -0,0 +1,216 @@
...
+
+(function_metadata
+  key: (identifier)) @property
```
ObserverOfTime, line 34:
```suggestion
  key: (identifier) @variable.member)
```

queries/faust/highlights.scm
```diff
@@ -0,0 +1,216 @@
...
+; Comments
+(comment) @comment @spell
+
+[
+  (doc_metadata)
+  (doc_equation)
+  (doc_diagram)
+  (doc_listing)
+  (doc_notice)
+] @tag
```
ObserverOfTime, line 214: You should only capture the tags and not their contents.

ObserverOfTime (changes requested):

queries/faust/highlights.scm
khiner: This is a language feature for declaring foreign functions. This node matches e.g. `asinhf|asinh|asinhl|asinfx` in
```
asinh = ffunction(float asinhf|asinh|asinhl|asinfx(float), <math.h>, "");
```
as `(function_names)`

ObserverOfTime: So… `@function`.

khiner: I suppose so, it's a custom syntax for specifying a list of foreign function names by concatenating them with `|`. Happy to change this to `@function` if you think it matches better

khiner: Pushed

queries/faust/highlights.scm
```diff
@@ -0,0 +1,216 @@
...
+  (vgroup_type)
+  (hgroup_type)
+  (tgroup_type)
+  "button"
+  "checkbox"
+  "soundfile"
+  "inputs"
+  "outputs"
+  "route"
+] @function.builtin
```
wookayin, line 121: Do each of the builtin functions need a separate node type? This could be a limitation for writing queries. In many other languages built-in functions are managed as (a list of) literal strings. (Using `#any-of?`)

queries/faust/highlights.scm
khiner: nice! pushed

queries/faust/highlights.scm
khiner: Yeah I'm experimenting with being detailed enough to support using the AST for generating other downstream intermediates, e.g. with MLIR. I know it's not the intended use-case for tree-sitter, but the idea of only maintaining one parser/lexer is too appealing not to try :)

queries/faust/highlights.scm
khiner: Thanks - I'll just skip tags for now

queries/faust/highlights.scm
khiner: pushed

queries/faust/highlights.scm
khiner: pushed

queries/faust/highlights.scm
ObserverOfTime: I mean, if you only have a subset of tags, it's better to just highlight those instead of injecting xml.

queries/faust/highlights.scm
wookayin: I am sorry I don't get your point. What's blocking here from fixing the grammar and what does have it to do with 'one' lexers/parser?

khiner: My goal is to support a consumer of the AST being able to, in theory, fully compile the AST into its corresponding Faust program. E.g. instead of having a Flex grammar/Bison parser that tokenizes and matches tokens to execute actions on those matches, I want to enable walking the tree-sitter generated AST instead and executing actions to compile the program. So the AST needs to somehow hold the information that a built-in function is e.g. a `cos` instead of a `sin`.

khiner: I'm open to ideas for better ways of achieving this!

khiner: Like if there was a way to walk into hidden nodes, that would be perfect

wookayin: tree-sitter generated AST would also allow visiting the token or the literal string of the identifiers or function name if needed. TS can visit unnamed nodes too. In my personal opinion, I am not fully convinced why they need to be "types" in the grammar.

wookayin (commented):
Another minor feedback: I tried examples from https://faustdoc.grame.fr/examples/ambisonics/, but there a few parsing errors that you might want to fix by improving the grammar.

khiner:
> Another minor feedback: I tried examples from https://faustdoc.grame.fr/examples/ambisonics/, but there a few parsing errors that you might want to fix by improving the grammar.

Thanks for checking - can you please provide the code that produces parse errors? I tried both examples on that page and couldn't reproduce parse errors.

queries/faust/highlights.scm
khiner: If it's frowned upon to inject xml parsing for subsets of the language that are defined in XML if only a handful of tags are expected, I'll just remove the XML injection for now and the tag highlight names for now.

queries/faust/highlights.scm
khiner: Interesting, I didn't know you could visit unnamed nodes. One thing that bugs me is that unnamed rules/fields can't be tested in the corpus since they don't show up. So you have untested outputs of the grammar that are only meant as... metadata for programmatically visiting the AST and not conceptually part of the AST itself? I see this is common practice (e.g. all the 'operator' named fields in the cpp grammar). Do folks usually maintain a separate test corpus to programmatically visit the expected unnamed nodes/fields?

What are the downsides of naming more rules than are strictly necessary? My understanding is it's mostly about performance?

queries/faust/highlights.scm
ObserverOfTime: It's not frowned upon per se, but users would have to install a parser they may not otherwise need.
You can remove the injection but keep the highlights (just apply them to the actual tags).

khiner: I see, so it's a requirement, not optional, to have the injected language installed. Ok then definitely best not to have it. I'll resolve this then.

ObserverOfTime: > I see, so it's a requirement, not optional, to have the injected language installed.

Well, if it's not installed, it won't be highlighted.

queries/faust/highlights.scm
clason: Performance, yes. State count and conflicts. You can run `./scripts/check-queries.lua faust` to get some timings for parsing the queries. See also https://github.com/tree-sitter/tree-sitter/wiki/Tips-and-Tricks-for-a-grammar-author

khiner: This aspect has been the most challenging part of tree-sitter for me to make design decisions about, given my dual-use interests (incremental, real-time highlighting + first stage of compiler infrastructure). I've been thinking about it more with your valuable feedback in mind. Regardless of my issues with losing easy testability with unnamed/hidden nodes/fields, I will adapt my approach and
1. Rewrite my operators/terminal primitives to generally not be named nodes, in compliance with standard tree-sitter conventions.
2. Continue pursuing my hopes to use tree-sitter to generate a more detailed AST for compiling purposes, but do this in two passes - first, a "low res" pass using unnamed nodes conventionally, and then a programmatic transformation step walking through the tree, visiting unnamed nodes to decorate the low-res AST with more detail. I realize for my purposes, I'll need much more than I can expect to put into tree-sitter anyway (such as full literal values and variable names).

I'll make those grammar changes soon and post back - thanks for the feedback!

khiner: @clason Do you have any suggestions for a case like this?
```
foo = f : + ~ g;
```
`~` is the "recursive" composition operator, and ':' is sequential composition. Before, this included the add operator on the lhs of the recursion:
```
  (definition
    variable: (identifier)
    value: (expression: (sequential
        left: (identifier)
        right: (recursive left: (add) right: (identifier))))
```
If I turn add into an unnamed '+' string, I lose it, which seems acceptable for binary infix operations, but here it seems wrong to take it out of the syntax tree:
```
  (definition
    variable: (identifier)
    value: (expression: (sequential
        left: (identifier)
        right: (recursive right: (identifier))))
```

clason: Honestly? You're asking the wrong person. We use tree-sitter for highlighting (and folding, and indenting), which requires low latency. A full syntax parser is very much a non-goal for us.

khiner: Roger. I am working to give up on the full syntax parsing, as I explained in my comment, and I don't expect you to help me accomplish that goal, but I'm trying to find a way to do that and still keep it sensible. Do you consider this feedback of yours on having too many operator/primitive named rules a blocker for getting my grammar/highlighter used nvim-treesitter because of the performance implications? Or can I stick with my current grammar while I work on ways to strike a better balance?

clason: Well, the parser is all yours; these are feedbacks on the queries, which we'll have to maintain. I personally will not block anything.

As I wrote, you can check the performance with `./scripts/check-queries.lua`.

khiner:
I think I've addressed all feedback if folks want to take another pass.

queries/faust/highlights.scm
khiner: @clason I updated the grammar and highlights to not used named rules for almost every built-in function.

khiner:
```
faust (highlights): ...-linux64/share/nvim/runtime/lua/vim/treesitter/query.lua:259: query: invalid node type at position 449 for language faust
```

Any tips on how to debug this? I don't see any errors when I run my grammar tests.

clason:
Install the parser and open the query in Neovim. The query linter should show you which pattern is wrong exactly.

khiner:
> Install the parser and open the query in Neovim. The query linter should show you which pattern is wrong exactly.

How do I install parsers that aren't published yet?

clason:
You've got your PR checked out, don't you?

khiner:
> You've got your PR checked out, don't you?

I'm just using my brew-installed neovim. How do I switch to using a local source dir?

clason:
I mean this PR here; just switch the checkout of nvim-treesitter to your feature branch (or use `gh pr checkout`).

khiner:
I have my fork of nvim-treesitter that I've been developing on, with the code on this PR. If I edit `nvim ~/.config/nvim/init.lua` to add `"faust"` to the `nvim-treesitter.configs -> ensure_installed` config, I get error:
```
Error detected while processing /Users/khiner/.config/nvim/init.lua:
Installation not possible: .../plugged/nvim-treesitter/lua/nvim-treesitter/install.lua:86: Parser not available for language "faust"
See https://github.com/nvim-treesitter/nvim-treesitter/#adding-parsers on how to add a new parser!
```

clason:
You should only have one nvim-treesitter installed, of course.

khiner:
> Either point it to your local directory (check your plugin manager config

This is what I need, looking into it...

khiner:
Ok I got past that - configured my nvim-treesitter to use my local dir, and successfully installed this faust parser. Now I'm trying to get past this:
```
$ ./scripts/check-queries.lua
E5113: Error while calling lua chunk: ./scripts/check-queries.lua:104: module 'nvim-treesitter.parsers' not found:
	no field package.preload['nvim-treesitter.parsers']
	no file './nvim-treesitter/parsers.lua'
	no file '/opt/homebrew/share/luajit-2.1/nvim-treesitter/parsers.lua'
	no file '/usr/local/share/lua/5.1/nvim-treesitter/parsers.lua'
	no file '/usr/local/share/lua/5.1/nvim-treesitter/parsers/init.lua'
	no file '/opt/homebrew/share/lua/5.1/nvim-treesitter/parsers.lua'
	no file '/opt/homebrew/share/lua/5.1/nvim-treesitter/parsers/init.lua'
	no file './nvim-treesitter/parsers.so'
	no file '/usr/local/lib/lua/5.1/nvim-treesitter/parsers.so'
	no file '/opt/homebrew/lib/lua/5.1/nvim-treesitter/parsers.so'
	no file '/usr/local/lib/lua/5.1/loadall.so'
	no file './nvim-treesitter.so'
	no file '/usr/local/lib/lua/5.1/nvim-treesitter.so'
	no file '/opt/homebrew/lib/lua/5.1/nvim-treesitter.so'
	no file '/usr/local/lib/lua/5.1/loadall.so'
stack traceback:
	[C]: in function 'require'
	./scripts/check-queries.lua:104: in main chunk
```

clason:
Well, you need to have nvim-treesitter on `runtimepath` for this to work. Easier to open `nvim`, `:TSInstall faust`, and then open your query.

khiner:
Gotcha. How do I open the query given the failure:
```
faust (highlights): ...im-macos/share/nvim/runtime/lua/vim/treesitter/query.lua:259: query: invalid node type at position 449 for language faust
```

I've tried `nvim lua/nvim-treesitter/query.lua` and see the erroring line is the beginning of a for loop:
```for id, node in pairs(match) do ...```

Sorry, this is the first time I've used nvim & nvim-treesitter & lua. I'm making this for others in the Faust community.

**Comment by clason** [comment]
Just edit your `queries/faust/highlights.scm`; there's a query linter that runs on these.

**Comment by khiner** [comment]
Ah I gotcha! Ok I did `nvim queries/faust/highlights.scm` and see no errors. I do have the "query" parser installed.

**Comment by clason** [comment]
You need to have _your_ faust parser installed, using the correct lockfile entry.

**Comment by khiner** [comment]
> You need to have _your_ faust parser installed, using the correct lockfile entry.

I do - I meant that this PR hadn't pulled my recent npm patch - but this is exactly what I was missing, I need to bump the lockfile entry here in my PR 🤦‍♂️

**Comment by khiner** [comment]
Ok we're good. _Now_ ready for re-review whenever you fine folks get the chance 😅

File: queries/faust/highlights.scm
Comment (ObserverOfTime) [inline queries/faust/highlights.scm:1]: Please re-add the tag highlights. :)

Comment (khiner) [reply queries/faust/highlights.scm:1]: Ok. So for this, we don't want XML language injection, we want to highlight just the tags (not their inner contents), so just making custom tag strings. Something like `["<foo>", "</foo>", "<bar/>", "<bar />"]` - that kind of thing?

Comment (khiner) [reply queries/faust/highlights.scm:1]: And how best to handle properties with simple rules like these, `"<foo"` without the closing bracket?

Comment (ObserverOfTime) [reply queries/faust/highlights.scm:1]: Just highlight those as tags as well.

Comment (khiner) [reply queries/faust/highlights.scm:1]: > Just highlight those as tags as well.

Can you say more? What's the behavior you're looking for for e.g. `<foo biz="buzz" /><bar fiz="fuzz"></bar>`?

Comment (khiner) [reply queries/faust/highlights.scm:1]: and doing this with a simple list of string literal matches

Comment (khiner) [reply queries/faust/highlights.scm:1]: This seems close
```
(documentation
  [
    "<metadata>"
    "<equation>"
    "</equation>"
    "<diagram>"
    "</diagram>"
    "<listing"
    "<notice"
    "/>"
  ] @tag)
```
But produces
```
    Query error at 208:3. Impossible pattern:
      [
      ^
```

Comment (khiner) [reply queries/faust/highlights.scm:1]: Got it!
```
(documentation
  (_
    [
      "<metadata>"
      "<equation>"
      "</equation>"
      "<diagram>"
      "</diagram>"
      "<listing"
      "<notice"
      "/>"
    ] @tag))
    ```

khiner: Pushed

ObserverOfTime:
```scheme
[
  "<mdoc>"
  "</mdoc>"
  "<metadata>"
  "</metadata>"
  "<equation>"
  "</equation>"
  "<diagram>"
  "</diagram>"
  "<listing"
  "<notice"
  "/>"
] @tag
```

khiner: Are you saying you'd like to see this
```
[
  "<mdoc>"
  "</mdoc>"
  "<metadata>"
  "</metadata>"
  "<equation>"
  "</equation>"
  "<diagram>"
  "</diagram>"
  "<listing"
  "<notice"
  "/>"
] @tag
```

Instead of what I've pushed up?
```
(documentation
  (_
    [
      "<metadata>"
      "<equation>"
      "</equation>"
      "<diagram>"
      "</diagram>"
      "<listing"
      "<notice"
      "/>"
    ] @tag))
    ```
    
And if so, why?

Comment (khiner) [reply queries/faust/highlights.scm:1]: oops I missed a metadata end tag

Comment (khiner) [reply queries/faust/highlights.scm:1]: I also pushed up this addition to get the root `<mdoc>`
```
(documentation
  [
    "<mdoc>"
    "</mdoc>"
  ] @tag)
```

Comment (khiner) [reply queries/faust/highlights.scm:1]: but I don't understand why the suggestion to not scope these inside of the `documentation` rule (the only place where xml-style tags are present in the language). This would be less performant and cause ambiguities, like `<listing`, which would be a less-than comparator with the identifier `listing` outside of documentation.

Comment (ObserverOfTime) [reply queries/faust/highlights.scm:1]: > This would be less performant

Which version is actually more performant? @clason :thinking: 

> and cause ambiguities, like `<listing`, which would be a less-than comparator with the identifier `listing` outside of documentation.

There wouldn't be any ambiguities since "<" "listing" is 2 nodes.

Comment (khiner) [reply queries/faust/highlights.scm:1]: > There wouldn't be any ambiguities since "<" "listing" is 2 nodes.

Oh I see

Comment (khiner) [reply queries/faust/highlights.scm:1]: Maybe you could explain the value add, or why you’re suggesting this code change?

Comment (ObserverOfTime) [reply queries/faust/highlights.scm:1]: Simpler queries = better queries.

Comment (clason) [reply queries/faust/highlights.scm:1]: > Which version is actually more performant? @clason 🤔

`./scripts/check-queries.lua faust` before and after should give you a general idea if one is significantly worse. (Omit the `faust` to see how other queries fare; Kotlin queries are considered unacceptably slow, as a benchmark.)

Comment (khiner) [reply queries/faust/highlights.scm:1]: > Simpler queries = better queries.

I see, thanks. Simplicity is always nice, and if this was a matter of choosing between two functionally equivalent choices, I would be with you and choose the simpler of the two. But in this case, we have to weigh the criterion of simplicity against:
* Clarity/self-documenting code: Keeping HTML tag matching inside the `documentation` rule makes it clear that HTML tags are _only_ expected/valid within that section.
* Avoiding unnecessary CPU work: If it is the case that it's slower to only match these tokens _within a single rule_ than _across an entire code unit_, then this would be odd - I'm not claiming it's empirically the case that it's faster in every case in tree-sitter, but in principle, it should be. Take an extreme case - if I tell you to look for appearances of the string "foo" in a 10M line code base, it will take you longer than if I give you the hint that this string only appears in one of the files. This is the same principle - I'm less concerned about whether tree-sitter actually happens to make use of this extra information as it should, but why not provide it so it has the option?
* Robustness: Why tempt fate and say all "/>" tokens in Faust should be highlighted as a tag, even if we know for sure it's not a tag since it's not in a documentation block? What if Faust wants to add a new "fancy-merge" feature, "/>", that's a play on the ":>" merge feature? I know it's a stretch, just pointing to the _kind_ of robustness I'm pointing at.

It feels odd that this PR should be held back by the point "your queries only search for HTML tags where they're valid, rather than also where they're not valid, because that rule takes fewer characters to express."

Can you please explain why you feel so strongly about this beyond "Simpler queries = better queries"?

**Review by ObserverOfTime — APPROVED** [review]

File: queries/faust/highlights.scm
Comment (clason) [reply queries/faust/highlights.scm:1]: 1. Linting (highlighting incorrect code as such) is an explicit non-goal of this project; we only care about highlighting _correct_ code.
2. Performance is critical: perceptible latency on keystrokes (the tree is reparsed on every edit) is inaccaptable. Also, precedence can be tricky to reason about, which is another reason why simpler is better.
3. If the language changes, the parser and queries have to be adapted. It doesn't make sense to implement for the unknown future.


But in the end, this is just feedback based on the 250+ parsers we've already added queries for.

Comment (khiner) [reply queries/faust/highlights.scm:1]: Thanks for providing some more detail. Can you say more about why it would bad to only search for HTML tags where they're expected? How would this make the parsing less performant? Why does it make precedence harder to reason about? My intuition is that having more rules searching across the entire file would be harder to reason about than knowing for sure it only applies in one code context. 

> If the language changes, the parser and queries have to be adapted. It doesn't make sense to implement for the unknown future.

Fair enough, and that example I provided isn't the main reason I'm asking for help to educate me on why I should make this change.

> But in the end, this is just feedback based on the 250+ parsers we've already added queries for.

So many parsers! I'm completely new to tree-sitter, this is my first one, and yesterday was the first time I ever wrote a tree-sitter query :)

Comment (khiner) [reply queries/faust/highlights.scm:1]: I tried `./scripts/check-queries.lua faust`, and usually get around `0.03ms`, but when I change my `queries/faust/highlights.scm` , even to something nonvalid, it doesn't change. I've tried reinstalling between changes with
```shell
$ nvim
$ TSInstall faust
> faust parser already available: would you like to reinstall ? y
```

but it still runs successfully, so doesn't seem to be picking up my changes 🤔

ObserverOfTime: TSUpdate faust

khiner: Thanks. Still couldn't get it to pick things up, I'm sure it's something about my env. I made this change, sorry to be a pain I know it's a lot of work reviewing without needing to explain every little thing. I'm going to be messing with queries a lot for my next project so I'm sure I'll develop my own intuition over time - in the meantime, I see tags are defined matched globally in other languages so I made that change to look for doc tags outside of docs.

PR #6050, opened 2024-02-04
feat: gotemplate and helm grammar

qvalentin:
This adds the gotmpl grammar and also support for helm which uses the same grammar but different injections. This is a little bit tricky and we haven't found the perfect solution for having a second language use the same grammar. 

One possible solution would be to have the parser.c file include a second TSLanguage defintion, as proposed here: https://github.com/ngalaiko/tree-sitter-go-template/pull/15

If you have any other idea, how this could be solved, feel free to make a suggestion.


lua/nvim-treesitter/parsers.lua
```diff
@@ -727,6 +727,23 @@ list.gowork = {
...
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@qvalentin" },
+}
+
+list.helm = {
+  install_info = {
+    url = "https://github.com/qvalentin/tree-sitter-go-template",
+    files = { "src/parser.c" },
+    branch = { "feat/helm-lang" },
```
qvalentin, line 742: This would need to be set back to the original repo.

lucario387:
You can just `; extends` for one language's highlights/indents, and have different injection queries for them

qvalentin:
> You can just `; inherits lang` for one language's highlights/indents, and have different injection queries for them

This is being used in this PR. But the problem is more of how you add the parser for the second language or configure a second filetype to use the same parser but different queries.

lucario387:
That's just a filetype thing for neovim no?

You are one `vim.treesitter.language.register` away from what you wanted

qvalentin:
> vim.treesitter.language.register

I'm afraid that won't work. As the readme states:

> vim.treesitter.language.register('python', 'someft')  -- the someft filetype will use the python parser **and queries**.

But I need to use only the same parser with different queries.

lucario387:
I think you are right here. In that case, you can follow the way `hcl` and `terraform` handles, they are the same language (for the most part) with some minor differences.

- hcl: https://github.com/nvim-treesitter/nvim-treesitter/blob/8b99311ca1171015bda7d2bb2250a1aaf2d9ea8a/lua/nvim-treesitter/parsers.lua#L794-L800
- terraform: https://github.com/nvim-treesitter/nvim-treesitter/blob/8b99311ca1171015bda7d2bb2250a1aaf2d9ea8a/lua/nvim-treesitter/parsers.lua#L1899-L1906

qvalentin:
Thanks for the example, good to know we are not the only ones with that problem. Looks like a rational solution.

lua/nvim-treesitter/parsers.lua
qvalentin: Also the sorting needs to be fixed.

kmoschcau:
Yeah that's the same solution as I already come to here: https://github.com/ngalaiko/tree-sitter-go-template/pull/13
Just registering the parser twice essentially. Sadly that means we potentially download/compile it twice.

queries/gotmpl/highlights.scm
```diff
@@ -0,0 +1,125 @@
+; Identifiers
+([
+  (field)
+  (field_identifier)
+] @property
+  (#set! priority 101))
```
clason, line 6: What is the reason for spamming `priority` here? That is normally not needed.

qvalentin: I removed them for now. @kmoschcau can you tell why you added them?

clason:
Please run `./scripts/format-queries.lua queries/`

queries/gotmpl/highlights.scm
kmoschcau: I added these because I use background color for literals in all languages. (strings, numbers, etc.)
gotmpl building on plain text might not need this. But helm building on YAML I think needed it, so that the gotmpl specific highlights were applied on top of the highlights that YAML adds.

clason: Then you should _lower_ the priority of the YAML literals (see markdown).

kmoschcau: That definitely sounds more reasonable. :D

queries/gotmpl/highlights.scm
```diff
@@ -0,0 +1,106 @@
...
+  method:
+    (selector_expression
+      field: (field_identifier) @function))
+
+; Builtin functions
+(function_call
+  function: (identifier) @function.builtin
+  (#set! priority 101)
+  (#match? @function.builtin
+    "^(and|call|html|index|slice|js|len|not|or|print|printf|println|urlquery|eq|ne|lt|ge|gt|ge)$"))
```
lucario387, line 23: Chang to any-of

queries/helm/injections.scm
lucario387, line 1: There should be a gotmpl file with comment injection and then helm inheriting it

queries/gotmpl/highlights.scm
```diff
@@ -0,0 +1,106 @@
...
+
+(method_call
+  method:
+    (selector_expression
+      field: (field_identifier) @function))
+
+; Builtin functions
+(function_call
+  function: (identifier) @function.builtin
+  (#set! priority 101)
```
lucario387, line 21: Priority isn't necessary i'd say. Just move the queries down. The effect can also be seen with tree-sitter-cli 0.21.0

queries/gotmpl/highlights.scm
```diff
@@ -0,0 +1,106 @@
...
+  ] @keyword.conditional)
+
+; Literals
+[
+  (interpreted_string_literal)
+  (raw_string_literal)
+  (rune_literal)
+] @string
+
+(escape_sequence) @string.special
```
ObserverOfTime, line 91:
```suggestion
(escape_sequence) @string.escape
```

queries/gotmpl/highlights.scm
clason: I think that's meant to overwrite the highlighting of the YAML node this is injected to (which, as I wrote earlier, should be handled by setting the priority of the YAML node to `90`).

queries/gotmpl/highlights.scm
```diff
@@ -0,0 +1,106 @@
...
+
+[
+  (int_literal)
+  (float_literal)
+  (imaginary_literal)
+] @number
```
ObserverOfTime, line 97:
```suggestion
[
  (int_literal)
  (imaginary_literal)
] @number

(float_literal) @number.float
```

queries/gotmpl/highlights.scm
```diff
@@ -0,0 +1,106 @@
...
+    "else"
+    "end"
+  ] @keyword.conditional)
+
+; Literals
+[
+  (interpreted_string_literal)
+  (raw_string_literal)
+  (rune_literal)
+] @string
```
ObserverOfTime, line 89:
```suggestion
[
  (interpreted_string_literal)
  (raw_string_literal)
] @string

(rune_literal) @string.special.symbol
```

queries/gotmpl/highlights.scm
lucario387: Another thing is that there is html and js here. So does that mean the arguments for the function can be valid html/js? If yes, more injections!

ObserverOfTime: And printf. ;)

qvalentin: image

Do you know why prinf is not really working? In the tree-sitter playground my query was looking good. Printf injection also does not seem to work for bash in my nvim config.

Regarding html and js, I think you could them with almost the same query, but it seems like the quotes are also part of the `interpreted_string_literal` node (see picture). This would require modifying the grammar.

image

ObserverOfTime: > Do you know why prinf is not really working?

Did you `:TSInstall printf` ?

> `it seems like the quotes are also part of the interpreted_string_literal`

Use `(#offset! @injection.content 0 1 0 -1)`

qvalentin: I installed it.
JS and html worked fine after installing them.

queries/helm/highlights.scm
```diff
@@ -0,0 +1,44 @@
...
+      name: (identifier) @constant.builtin
+      (#any-of? @constant.builtin "Values" "Chart" "Release"))
+  (field_identifier))
+
+; {{ $.Values.test }}
+(selector_expression
+  operand: (variable)
+  field:
+    ((field_identifier) @constant.builtin
+      (#any-of? @constant.builtin "Values" "Chart" "Release")))
```
lucario387, line 44:
```suggestion
    (field_identifier) @constant.builtin
      (#any-of? @constant.builtin "Values" "Chart" "Release"))
```

queries/helm/highlights.scm
```diff
@@ -0,0 +1,46 @@
...
+    "mergeOverwrite" "mustMergeOverwrite" "keys" "pick" "omit" "values" "deepCopy" "mustDeepCopy"
+    "b64enc" "b64dec" "b32enc" "b32dec" "list" "first" "mustFirst" "rest" "mustRest" "last"
+    "mustLast" "initial" "mustInitial" "append" "mustAppend" "prepend" "mustPrepend" "concat"
+    "reverse" "mustReverse" "uniq" "mustUniq" "without" "mustWithout" "has" "mustHas" "compact"
+    "mustCompact" "index" "slice" "mustSlice" "until" "untilStep" "seq" "add" "add1" "sub" "div"
+    "mod" "mul" "max" "min" "len" "addf" "add1f" "subf" "divf" "mulf" "maxf" "minf" "floor" "ceil"
+    "round" "getHostByName" "base" "dir" "clean" "ext" "isAbs" "kindOf" "kindIs" "typeOf" "typeIs"
+    "typeIsLike" "deepequal" "semver" "semverCompare" "urlParse" "urlJoin" "urlquery" "lookup"
+    "include")
+  )
```
ObserverOfTime, line 30:
```suggestion
    "include"))
```
The formatter messed up again.

lucario387: :facepalm: Time for me to refactor the queries again

lucario387: Not my fault but rahter neovim :D 

```scheme
(program ; [0, 0] - [5, 0]
  (comment) ; [0, 0] - [0, 19]
  (list ; [1, 0] - [4, 2029]
    (symbol) ; [1, 1] - [1, 14]
    (symbol) ; [2, 2] - [2, 11]
    (list ; [2, 12] - [2, 24]
      (symbol)) ; [2, 13] - [2, 23]
    (symbol) ; [2, 25] - [2, 42]
    (list ; [3, 2] - [4, 2028]
      (ERROR) ; [3, 3] - [3, 11]
      (symbol) ; [3, 12] - [3, 29]
      (string) ; x 200
      (string)))) ; [4, 2018] - [4, 2027]
```

lucario387: Tbh I'm more than surprised that the formatter doesn't go boom with it. So I take that as a win

clason:
Could you rebase (and squash) again?

queries/helm/highlights.scm
clason: Linter has the last word. Either we wait until the upstream issue https://github.com/neovim/neovim/issues/27705 is fixed, or we merge the messed up version and reformat later (probably easier).

ObserverOfTime (approved):

kmoschcau:
Good job getting this over the finish line. This will also help me out a lot.

kmoschcau:
I think we need a different PR for this. It seems that query order doesn't really affect the highlighting order:
image
For me `helm` parts in YAML strings are highlighted as strings. I already tried setting the priority on the injection query, but that doesn't seem to work. Setting priority on single helm highlights does work though.

clason:
Yes, as I wrote, the YAML(!) highlight for the node with helm injection needs to have lower priority. (Not injection; there is no such thing as priority for injections.)

PR welcome.

kmoschcau:
> Yes, as I wrote, the YAML(!) highlight for the node with helm injection needs to have lower priority. (Not injection; there is no such thing as priority for injections.)
> 
> PR welcome.

How do you do that though? The only idea I would have is to then lower the priority for everthing in the YAML highlighting.

And I have no problem with the background color of the string staying there. In fact for me that's the intention. Just the foreground colors of the helm tokens should be actually highlighted.

clason:
> How do you do that though? The only idea I would have is to then lower the priority for everthing in the YAML highlighting.

`:InspectTree` will show you the YAML node; then look at the YAML query and set the priority for the corresponding pattern to `90`. (Compare the markdown highlight query.) Some trial and error is probably required.

kmoschcau:
So you mean in this case I would need to edit `queries/yaml/highlights.scm` and lower the priority of `(double_quote_scalar)` to 90?
image

kmoschcau:
Also overall this kinda seems backwards to me. We are now lowering the priority of all YAML highlights no matter whether there's helm code in it or not. Helm is ever only layered on top of YAML. To me it seems more logical to increase the helm and gotmpl highlights instead.

clason:
You know what? Why don't you open a proper issue with an actual, minimal, example showing code, parsed tree, and `:Inspect` results that you want changed. This is getting a bit confusing -- earlier you talked about injecting helm into yaml; now it's the other way around?

kmoschcau:
It's a bit confusing how this is built, I know. Both `gotmpl` and `helm` are built as their own parsers. Pretty much anything inside and including double curly braces is treated as `gotmpl` or `helm`. Anything outside is treated as plain text in case of `gotmpl`. In the case of `helm`, anything outside of those curlies is injected as one combined `yaml` template. So `yaml` is injected in the `helm` document, but it usually physically encloses the `helm` parts.

clason:
>  So yaml is injected in the helm document, but it usually physically encloses the helm parts.

No, that is _not_ happening here -- if there is no injection query for it, there's no injection. Instead, everything is parsed as a single "helm" language, which includes syntax elements from YAML. So what you need to control is the relative priority _within_ the `helm` query. This priority is determined by -- in order -- three things:
1. order (later patterns overwrite previous ones)
2. explicit `priority`

You should rely on 1. wherever possible and only use 2. when absolutely necessary -- which is usually (but not here!) the case when you need to control priorities across different (tree-sitter) languages.

kmoschcau:
But this PR added the the queries to inject YAML into helm here: https://github.com/nvim-treesitter/nvim-treesitter/pull/6050/files#diff-78359e0d0cfddf3453540bcbf5b74c3ccf59a093f44939e3a4dfdace40481960R3-R5

I'm not sure order would work in this case. So to me it still seems like raising the priority of the `helm` highlight queries would be the right way.

clason:
Oh, you realize helm -> YAML injection by injecting YAML into basically every helm node? That's a bit... not my problem. (This is something I'd have seen in a parsed tree, hence my repeated requests for that.)

Anyway, then I see two solutions:
1. make the injection more specific so it _doesn't_ apply to actual helm nodes
2. raise the helm pattern priorities.

I would strongly prefer 1. since it's robust to changes in the YAML highlights. (Otherwise use 110 instead of 101 to give yourself a bit of room.)

kmoschcau:
If I understand that correctly, solution one would also require changing the parser. Also from earlier experience, `yaml` documents tend to completely break when they are not parsed as one combined thing. So I'm not sure having them as separate regions is going to work.

clason:
No, I'm talking about the injection query here. Right now, all `(text)` nodes are unconditionally injected with `yaml`. I'd start with trying to find out why you get YAML injected where helm should be and fixing that.

kmoschcau:
Isn't that because of the `injection.combined`?

clason:
That just says "treat all captured nodes as a single region"; it does not say _what_ is captured.

kmoschcau:
@qvalentin Do you have a preference which route to take?

qvalentin:
> No, I'm talking about the injection query here. Right now, all `(text)` nodes are unconditionally injected with `yaml`. I'd start with trying to find out why you get YAML injected where helm should be and fixing that.

So I noticed the following:

Given this text

```yaml
test: {{ .Values.test }}b
```

it produces this tree:

```query
(document) ; [1:1 - 2:0]
 (block_node) ; [1:1 - 2:0]
  (block_mapping) ; [1:1 - 2:0]
   (block_mapping_pair) ; [1:1 - 25]
    key: (flow_node) ; [1:1 - 4]
     (plain_scalar) ; [1:1 - 4]
      (string_scalar) ; [1:1 - 4]
    value: (flow_node) ; [1:7 - 25]
     (plain_scalar) ; [1:7 - 25]
      (string_scalar) ; [1:7 - 25]
(text) ; [1:1 - 6]
(selector_expression) ; [1:10 - 21]
 operand: (field) ; [1:10 - 16]
  name: (identifier) ; [1:11 - 16]
 field: (field_identifier) ; [1:18 - 21]
(text) ; [1:25 - 2:0]
```

The text nodes are as expected, only those should be injected with yaml. But the yaml tree also includes nodes that are outside the range of the text nodes.

Given this I would not know how to achieve the first suggestion 

> 1. make the injection more specific so it doesn't apply to actual helm nodes
                                                                                                                     

Regarding if we are wrongly injection yaml into helm, when we should inject helm into yaml, I would like to know if it would be possible to do it the other way around. Detecting if a template is helm or gotemplate or something completely different would be quite hard.

clason:
> The text nodes are as expected, only those should be injected with yaml. But the yaml tree also includes nodes that are outside the range of the text nodes.

Yes, that is what is confusing me as well. I think the first step should be understanding why this happens.

qvalentin:
I did some more experiments and found out the following:

Calling ts_parser_set_included_ranges (what I assume is used for injections) does not prevent the parser from creating nodes that include ranges that are outside the included ranges.

With this I see no other way then using priorities. I would suggest raising the gotemplate priorities, since this is specific to gotemplate and should not change yaml.

clason:
That seems... like a bug somewhere? But, ok, assuming that can't be solved quickly, it's OK to raise the gotmpl priorities
1. to 110 (not 101)
2. and add a comment at the top about the rationale so this can be removed when the issue is fixed.

PR #6051, opened 2024-02-04
fix(latex): give comments higher precedence

### LINKED ISSUE #6047: [latex] Some comments highlighted incorrectly in math regions
### Discussed in https://github.com/nvim-treesitter/nvim-treesitter/discussions/6038

<div type='discussions-op-text'>

<sup>Originally posted by **gi1242** February  1, 2024</sup>
Hi All,

In Latex files, is it possible for comments inside the file to be highlighted as comments. I'm attaching a screenshot of what it looks like now:

image

For the comment that is highlighted correctly, `:Inspect` shows:

    Treesitter
      - @markup.math.latex links to Special latex
      - @spell.latex links to @spell latex
      - @comment.latex links to Comment latex
      - @spell.latex links to @spell 

For the comment that is highlighted incorrectly, `:Inspect` shows:

    Treesitter
      - @comment.latex links to Comment latex
      - @spell.latex links to @spell latex
      - @markup.math.latex links to Special latex

Thanks in advance,

GI</div>

@clason  suspects this is a regression caused by a recent commit of @ribru17

ribru17:
This commit also moves things around to give text better precedence in the case of `textit` and `textbf`.

Fixes #6047 

I created a comprehensive example file to show some differences. Before:
latexfixupsbefore

After:
latexfixupsafter

Note that this was tested in Markdown only but I did a few tests in a `.tex` file and the same results were noticed. Hopefully this fixes the last of this weirdness, but I am not certain that every edge case is caught.

### DIFF
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -298,15 +298,6 @@
       (_) @markup.link.url))
   (#any-of? @_name "\\url" "\\href"))
 
-[
-  (line_comment)
-  (block_comment)
-  (comment_environment)
-] @comment @spell
-
-((line_comment) @keyword.directive
-  (#lua-match? @keyword.directive "^%% !TeX"))
-
 ; File inclusion commands
 (class_include
   command: _ @keyword.import
@@ -369,3 +360,15 @@
   (displayed_equation)
   (inline_formula)
 ] @markup.math
+
+[
+  (line_comment)
+  (block_comment)
+  (comment_environment)
+] @comment @spell
+
+((line_comment) @keyword.directive
+  (#lua-match? @keyword.directive "^%% !TeX"))
+
+((line_comment) @keyword.directive
+  (#lua-match? @keyword.directive "^%%&"))
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -12,15 +12,6 @@
   value: (_))
 
 [
-  (line_comment)
-  (block_comment)
-  (comment_environment)
-] @comment @spell
-
-((line_comment) @keyword.directive
-  (#lua-match? @keyword.directive "^%% !TeX"))
-
-[
   (brack_group)
   (brack_group_argc)
 ] @variable.parameter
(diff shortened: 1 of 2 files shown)


queries/latex/highlights.scm
```diff
@@ -313,6 +298,15 @@
...
   (#any-of? @_name "\\url" "\\href"))
 
+[
+  (line_comment)
+  (block_comment)
+  (comment_environment)
+] @comment @spell
+
+((line_comment) @keyword.directive
+  (#lua-match? @keyword.directive "^%% !TeX"))
```
lucario387, line 308: Can this line directive happen everywhere, or only the first few lines of the tex file

If it's the latter then I suggest limiting it to first few line at top-level only

ribru17: I am not sure, I searched around but couldn't find whether it can only be used at the top (though I think it usually is only used at the top). Maybe someone of more expertise can help here

wookayin: This, so-called the TeX "magic comments", should come in the first few lines *by convention*, although it's not strictly mandatory from a technical perspective.

https://tex.stackexchange.com/questions/78101/when-and-why-should-i-use-tex-ts-program-and-tex-encoding/84687#84687
https://www.texdev.net/2011/03/24/texworks-magic-comments/

wookayin (approved):

queries/latex/highlights.scm
lucario387, line 308: I'd say just put this to the end of the file. Who knows how many queries we will add in the future.

gi1242: Another special comment is for pre-compiled headers. (This is a LaTeX feature, and not tied to TeX works or other conventions.) A document that starts with

    %& filename

signals the document is using precomplied headers, and tex looks for filename.fmt

Would it be possible to add these special comments here too? Or do they go into the grammar.js upstream?

Thanks again for fixing this so quickly!

gi1242: PS: Comments can come before the `%&` comment; but I don't think any regular commands can. I didn't find a reference for this (but tested it).

gi1242:
Thanks a lot. I tested and it works for me 😄. If I figure out the syntax I'll hopfully fix it instead of bothering you in future 😄

GI

PR #6054, opened 2024-02-06
Try to fix indentation for <template> ... </template>

NullVoxPopuli:
Currently, when I press `Enter` in a 
```gjs
// test.gjs
<template>

</template>
```
block, 
the cursor goes all the way to the left, rather than being indented.

I'm still trying to figure out how to test things locally, so this isn't ready for review or merge.

### DIFF
--- a/queries/ecma/indents.scm
+++ b/queries/ecma/indents.scm
@@ -16,6 +16,7 @@
   (switch_statement)
   (template_substitution)
   (ternary_expression)
+  (glimmer_opening_tag)
 ] @indent.begin
 
 (arguments
@@ -68,6 +69,8 @@
   "}"
   "]"
 ] @indent.end
+
+(glimmer_closing_tag) @indent.end
 
 (template_string) @indent.ignore
 


lucario387:
Are you sure? I think it's just the filetype being incorrect here.

`test.gjs` shows `javascript.glimmer` as the filetype, not `glimmer`

NullVoxPopuli:
correct -- maybe I need to configure an offset? (I don't know how to do that)
Right now, tree-sitter-glimmer has no idea if it's being used between `<template>...</template>` or not

NullVoxPopuli:
I don't think that's correct. tree-sitter-glimmer is for the sub-language within `<template> ... </template>`, not the whole gjs file.


Injection declared here: https://github.com/nvim-treesitter/nvim-treesitter/blob/aa805bf22f4a2fb9f429de3a22e9dc45ec444833/queries/ecma/injections.scm#L58

> Consider adding a line of vim.treesitter.language.register("glimmer", {"javascript.glimmer"}) somewhere in your file.

Is this something other language libraries do? (registering themselves?) 
Or is this only done to override associations with parsers?

If so, tree-sitter-javascript is the correct parser for gjs.
But maybe I misunderstand what that register command does? how does it relate to the indentexpr? Is that something I should PR to tree-sitter-glimmer?


One thing I've been trying to figure out, is where is the right place to fix indentation of

```gjs
const Two = <template>|</template>
```
(press enter)
```gjs
const Two = <template>
  |</template>
```
it should be this instead:
```gjs
const Two = <template>
  |
</template>;  
```
something needs to tell the inner syntax (tree-sittter-glimmer) to offset all indentation by an additional $OFFSET_SIZE (whatever that is)

lucario387:
That's not really because of the indentexpr.
Auto newlining the end template tag is the formatexpr's thing

clason:
Agree, this is a filetype issue. Glimmer nodes don't belong in the basic shared ecma queries.

PR #6055, opened 2024-02-06
queries/templ: add the spread_attributes node

vrischmann:
I recently added a new `spread_attributes` node in the `tree-sitter-templ` parser that requires specific highlighting, this PR adds it.


vrischmann:
(closing for now, need to figure out a bug)

PR #6063, opened 2024-02-09
fix(unison): adapt to breaking parser change

clason:


queries/unison/highlights.scm
```diff
@@ -14,7 +14,7 @@
 ; Keywords
 [
   (kw_forall)
-  (unique_kw)
+  (uniquew)
```
lucario387, line 17: Typo'd

lucario387 (approved):

PR #6066, opened 2024-02-09
feat(ecma): highlight decorator and destructuring

zweimach:
| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -12,10 +12,12 @@
 
 (private_property_identifier) @variable.member
 
-(variable_declarator
-  name:
-    (object_pattern
-      (shorthand_property_identifier_pattern))) @variable
+(object_pattern
+  (shorthand_property_identifier_pattern) @variable)
+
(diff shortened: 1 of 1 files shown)


queries/ecma/highlights.scm
```diff
@@ -138,6 +140,17 @@
   (call_expression
     (identifier) @attribute))
 
+(decorator
+  "@" @attribute
+  (member_expression
+    (property_identifier) @attribute))
```
lucario387, line 146: Can there be `@A.B.C`? Because if yes, I think we might be in a sticky situation.

zweimach: I think it's fine:
image

zweimach: I'm unsure if `@A.B.C` is semantically correct, but it's usually just `@SomeModule.Decorate()`.  
IMO from a syntactical perspective it's good enough.

lucario387 (approved):

PR #6075, opened 2024-02-10
chore(bash): update parser and highlights

amaanq:
totally not an excuse to update the parser early


lucario387 (approved):
I read like 2/3 of tree-sitter-bash, but looks good

PR #6077, opened 2024-02-10
ci: fix invalid workflow

dundargoc:
Context variables aren't always available, so we can't rely on them for
e.g. environment variables.

Also merge all install scripts into `script/ci-install.sh`, that way it
can be called the same way regardless of system.

### DIFF
--- a/scripts/ci-install.sh
+++ b/scripts/ci-install.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -e
+
+os=$(uname -s)
+if [[ $os == Linux ]]; then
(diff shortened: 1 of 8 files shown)


clason:
You also need to rename the scripts, which are tied to the workflow name. (Which is why we use `foo-latest` if possible.)


If you have a better idea for setup, go ahead!

clason:
In fact, why are they even external scripts -- can't we just put those few lines in the workflow itself?

scripts/ci-install.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
```
wookayin, line 2: Would `set -e` be needed here?

dundargoc: Eh sure, why not

PR #6088, opened 2024-02-13
fix(cpp): add back comment injection

tomtomjhj:
accidentally removed in https://github.com/nvim-treesitter/nvim-treesitter/commit/57a8acf0c4ed5e7f6dda83c3f9b073f8a99a70f9#diff-7e47e0991335c05bf503d5603c5c047b7648299bb0f6fe6d2f08230827cece77


lucario387:
Thanks. On `main` branch it's removed because cpp inherits c queries so it was not thought of then

PR #6096, opened 2024-02-14
feat: add some ft, lang pairs

Freed-Wu:
ebuild comes from https://github.com/gentoo/gentoo-syntax which is installed by Gentoo/Linux
tutor comes from $VIMRUNTIME/syntax/tutor.vim
mysql comes from $VIMRUNTIME/syntax/mysql.vim
text comes from $VIMRUNTIME/ftplugin/text.vim
expect comes from $VIMRUNTIME/syntax/expect.vim

I am not sure should we add

brewfile => ruby, comes from https://github.com/bfontaine/Brewfile.vim
altera => tcl, comes from https://github.com/vim-scripts/altera.vim
xdc => tcl, comes from https://github.com/amal-khailtash/vim-xdc-syntax

Because they don't come from $VIMRUNTIME
or preinstalled by any GNU/Linux distributions.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -30,14 +30,10 @@ for ft, lang in pairs {
   ecma = "javascript",
   jsx = "javascript",
   sh = "bash",
-  apkbuild = "bash",
-  PKGBUILD = "bash",
   html_tags = "html",
   ["typescript.tsx"] = "tsx",
   ["html.handlebars"] = "glimmer",
   systemverilog = "verilog",
-  cls = "latex",
-  sty = "latex",
   pandoc = "markdown",
   rmd = "markdown",
   quarto = "markdown",
@@ -46,6 +42,8 @@ for ft, lang in pairs {
   svg = "xml",
   xsd = "xml",
   xslt = "xml",
+  expect = "tcl",
+  mysql = "sql",
   sbt = "scala",
 } do
   register_lang(lang, ft)


clason:
We will only ever consider filetypes that Neovim recognizes by default, and filetypes which have the _exact_ same syntax (i.e., where the syntax file includes the other syntax file verbatim). I also want to keep this list small as users can always use `vim.treesitter.language.register()` for niche filetypes.

So that's a definite no for `text` (which is _not_ comment!), `tutor` (which is _not_ Markdown!) and `brewfile`. I'm also very skeptical about the other mappings. (@lewis6991 for tcl?)

amaanq:
ebuild/class is a no as well (belongs in nvim ft detect)

Freed-Wu:
> ebuild/eclass for Gentoo

I notice:

```lua
  apkbuild = "bash",
  PKGBUILD = "bash",
```

They are not belongs nvim ftdetect.

amaanq:
I'm inclined to agree

clason:
Yes, and the correct step would be to _remove_ these, not add new ones.

amaanq:
isn't expect a superset of tcl? this is similar (but not exactly) to zsh vs bash

ObserverOfTime:
`cls` and `sty` are not actual filetypes either as far as I can tell.

clason:
Indeed; they're recognized (if they are) as `tex` files.

clason:
I think many of these entries were just cargo-culted from back when nvim-treesitter also handled filetype detection.

ObserverOfTime (approved):

PR #6097, opened 2024-02-14
fix(python): better try-except indentation

### LINKED ISSUE #6072: (python) Incorrect try/except indentation in certain cases
### Describe the bug

When a try statement is inserted before a comment followed by indented statement (`if`, `while`, `for`, etc.), it is given 0 indentation.

### To Reproduce

1. Open a new python file and paste this example:
```python
def foo():
    print('hi')


    # comment
    if True:
        pass
```

2. Place the cursor on the line after the print statement and type `try:` in insert mode.

4. The new try statement will be given no indentation:
```python
def foo():
    print('hi')
try:

    # comment
    if True:
        pass
```

### Expected behavior

The try block is indented at the same level as the rest of the function:
```python
def foo():
    print('hi')
    try:

    # comment
    if True:
        pass
```

```markdown

Installation ~
- OK `node` found v20.11.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
  Version: Apple clang version 15.0.0 (clang-1500.1.0.2.5)

OS Info:
{
  machine = "x86_64",
  release = "23.3.0",
  sysname = "Darwin",
  version = "Darwin Kernel Version 23.3.0: Wed Dec 20 21:28:58 PST 2023; root:xnu-10002.81.5~7/RELEASE_X86_64"
} ~

```

```text
$ nvim --version
NVIM v0.10.0-dev-2317+g44ec4b5b1
Build type: RelWithDebInfo
LuaJIT 2.1.1707061634
```

### Additional context

Removing either the if statement or the comment causes the indentation to behave as expected.

ribru17:
Fixes #6072 (mostly). There is still an issue with something like the following:

```python
def foo():
    print('hi')
    try:
        print('hi')
        # This if statement messes up the following except statement
        if True:
            pass

        except:

    # comment
    if True:
        pass
```

I think tests should be added but I'm not sure how to do this as it is sort of needs to test while the user is typing, hopefully someone can offer some insight


lucario387:
You can add a `run:new_line("if_else file", { text = try:}`

lucario387:
You can add a `run:new_line("if_else file", { text = try:}`

queries/python/indents.scm
```diff
@@ -38,8 +38,19 @@
...
+
+(ERROR
+  "try"
+  .
+  ":"
+  (ERROR
+    (block
+      (expression_statement
+        (identifier) @_except) @indent.branch))
+  (#eq? @_except "except"))
```
lucario387, line 53: After trying a bit, this covers both cases:

```suggestion
(
  "try"
  .
  ":"
  .
  (ERROR
    (block
      (expression_statement
        (identifier) @_except) @indent.branch))
  (#eq? @_except "except"))
```

lucario387: And if you decided to apply this suggestion, remember to add some more tests. Even the current one looks/is good to me

ribru17:
Will `run:new_line` remember previous calls? Like can I do a newline of `try:` and later do a new line of `except:`? I just added two versions in case it does not work this way but I don't know if it can be more concise.

lucario387:
Nope it cannot for the time being.

Someday I/we will have to port nvim's screen test to here

ribru17:
Interesting behavior in the tests. Typing (hitting enter for new lines, using `cc` command) has correct indentation, but `=ip` over everything gives different (wrong) indents. Now I have found a case where `=ip` gives wrong indents, running it again re-corrects them, running it again breaks again, etc. :laughing:

ribru17:
After more tries I think it is not worth it to fix this aspect. The indentation changes the actual nodes of the syntax tree, which have different indentation strategies, causing the weird oscillation behavior. And I think as long as it works while typing, most users will probably complete the block with an `except` before doing any sort of manual indentation formatting. However I don't know how to test inserting `except:` text in the tests without it also checking to see if the above statements hold after some `=` operation, just want to tell it to not worry about that as long as except works lol

lucario387:
Yep. It working while typing is enough.

lucario387 (approved):
Great work!

ribru17:
Thank you!

PR #6100, opened 2024-02-14
Add indent and fold files to mermaid

aareman:
I added basic indentation and folding to mermaid files. The language only had highlights previously. (I think there is room for markdown injections by node names, but I wasn't planning to add that here)

First time contributing here, I read the Contributing.md file and still need to run the formatting scripts. Please let me know if there is anything else I need to change.

----
Thank you for this amazing plugin. :slightly_smiling_face:


clason:
The mermaid queries are currently orphaned. Would you like to volunteer as maintainer?

queries/mermaid/indents.scm
```diff
@@ -0,0 +1,13 @@
+[
+  (diagram_flow)
+  (flow_stmt_subgraph)
+] @indent.begin
+
+[
+  "end"
+] @indent.branch
```
lucario387, line 8: `@indent.end` seems to be missing here.

Also you need to run `scripts/format-queries queries/mermaid`

aareman:
> The mermaid queries are currently orphaned. Would you like to volunteer as maintainer?

What would that entail? I would potentially be interested.

lucario387:
> What would that entail

To be expected to be pinged and to have some responsiblity in fixing things when some mermaid changes happen that breaks the current queries.

lucario387:
Also, this PR is only for `flowgraph` and `subgraph` of mermaid, is it correct? What about the other mermaid graphs, do they not exist in the queries?

clason:
> > What would that entail
> 
> Expected to be pinged and to have some responsiblity in fixing things when some mermaid changes happen that breaks the current queries.

And ideally be proactive: monitor the mermaid parser and make PRs to adapt or extend the queries here _before_ our CI fails. (Especially the latter -- if there are new nodes or such which queries need to be rewritten to make use of; our CI catches breaking changes but not possible improvements.)

queries/mermaid/indents.scm
aareman: > `@indent.end` seems to be missing here.

I'm not sure what you mean here. Is `@indent.end` a required part of an indent query in neovim (I didn't see that in the docs)? By making it branch it does the following

```{mermaid}
flowchart BT
  subgraph DA
    direction BT
    a --> b
  end
```
```mermaid
flowchart BT
  subgraph DA
    direction BT
    a --> b
  end
```
In terms of the "end" of the first indent. There appears to be no "end" in mermaid js, since I think its one chart per file. And the TS Query Tree is rooted at document. I can't find anything contrary to this with some cursory looking. (I mostly use it in markdown -- one chart per code block)

> Also you need to run `./scripts/format-queries.lua queries/mermaid`

I was waiting to do this until I knew that I had a general direction to get the pr going.

aareman:
> Also, this PR is only for `flowgraph` and `subgraph` of mermaid, is it correct? What about the other mermaid graphs, do they not exist in the queries?

That is a good point. I was thinking that there are other graphs. I can add them to the list.

aareman:
Added other graph types And fixed all linting/formatting issues. 

I'm going to pass, for now, on the maintainer position for mermaid.

queries/mermaid/indents.scm
```diff
@@ -0,0 +1,18 @@
+([
+  (diagram_sequence)
+  (diagram_class)
+  (diagram_state)
+  (diagram_gantt)
```
lucario387, line 5: Can this do
```
gantt
    title A Gantt Diagram
    dateFormat YYYY-MM-DD
    section Section
        A task          :a1, 2014-01-01, 30d
        Another task    :after a1, 20d
    section Another
        Task in Another :2014-01-12, 12d
        another task    :24d
```

clason:
@aareman any updates?

aareman:
I couldn't get the indentation working for the gantt example posted. I don't understand the treesitter indentation rules well enough. Do you have any pointers how to do that example?

lucario387:
Eh I also can't make it works, which means this is good enough :)

Sorry for the wait.

lucario387:
Closing in favor of #7123. Thank you for the PR!

PR #6106, opened 2024-02-16
fix: update add_predicate and add_directive calls for upstream

gpanders:
Update custom predicates and directives to handle multiple nodes per capture ID per changes upstream.

Ref: https://github.com/neovim/neovim/pull/24738

### DIFF
--- a/plugin/query_predicates.lua
+++ b/plugin/query_predicates.lua
@@ -1,21 +1,22 @@
 local query = vim.treesitter.query
 
 local predicates = {
-  ---@param nodes TSNode[]
+  ---@param match TSMatch
   ---@param pred string[]
   ---@param any boolean
   ---@return boolean
-  ['has-type'] = function(nodes, pred, any)
+  ['has-type'] = function(match, pred, any)
+    local nodes = match[pred[2]]
     if not nodes or #nodes == 0 then
       return true
     end
 
     local types = { unpack(pred, 3) }
     for _, node in ipairs(nodes) do
-      local match = vim.list_contains(types, node:type())
-      if any and match then
+      local res = vim.list_contains(types, node:type())
+      if any and res then
         return true
-      elseif not any and not match then
+      elseif not any and not res then
         return false
       end
     end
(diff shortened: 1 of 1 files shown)


plugin/query_predicates.lua
```diff
@@ -6,14 +6,25 @@ local query = vim.treesitter.query
...
 
   local types = { unpack(pred, 3) }
-  return vim.list_contains(types, node:type())
-end, true)
+  for _, node in ipairs(nodes) do
+    if vim.list_contains(types, node:type()) then
+      return true
+    end
+  end
+  return false
```
lucario387, line 23: Isn't this the behavior of `any-has-type?` i.e. only need 1 node of the quantifiers to match.

gpanders: Yes you're right, good catch. I'll add a separate `any-has-type?` predicate for the "any" behavior.

plugin/query_predicates.lua
```diff
@@ -1,19 +1,59 @@
...
 ---@param pred string[]
 ---@return boolean|nil
 query.add_predicate('has-type?', function(match, _, _, pred)
-  local node = match[pred[2]]
-  if not node then
-    return true
+  local nodes = match[pred[2]]
+  if type(nodes) ~= 'table' then
+    -- Check for old versions of Nvim that map capture IDs to a single node
+    nodes = { nodes }
```
lucario387, line 35: This branch targets neovim 0.10+, and I believe according to neovim/neovim#24738, this guard is unnecessary

gpanders: The guard is there in case someone updates nvim-treesitter before updating neovim (assuming both this PR and https://github.com/neovim/neovim/pull/24738 are merged). But you're right, since this targets 0.10 and we're talking about nightly/unstable versions, we probably don't need a guard and we can just tell people to update neovim.

plugin/query_predicates.lua
```diff
@@ -31,11 +62,17 @@ local mimetype_aliases = {
 ---@return boolean|nil
 query.add_directive('set-lang-from-mimetype!', function(match, _, bufnr, pred, metadata)
   local id = pred[2]
-  local node = match[id]
-  if not node then
+  local nodes = match[id]
+  if type(nodes) ~= 'table' then
+    nodes = { nodes }
+  end
```
lucario387, line 68: Similarly, I believe this is also unnecessary.

lucario387 (approved):

wookayin:
This should've not been merged until the upstream issue is merged because we don't have the guard?

clason:
It's fine; `main` is work in progress and can be (mildly) broken for a day.

PR #6108, opened 2024-02-16
feat(rust): add doc comments

clason:
replace old lua-match patterns with newly added nodes

### DIFF
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -220,20 +220,11 @@
   (block_comment)
 ] @comment @spell
 
-((line_comment) @comment.documentation
-  (#lua-match? @comment.documentation "^///[^/]"))
+(line_comment
+  (doc_comment)) @comment.documentation
 
-((line_comment) @comment.documentation
-  (#lua-match? @comment.documentation "^///$"))
-
-((line_comment) @comment.documentation
-  (#lua-match? @comment.documentation "^//!"))
-
-((block_comment) @comment.documentation
-  (#lua-match? @comment.documentation "^/[*][*][^*].*[*]/$"))
-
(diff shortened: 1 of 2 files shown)


queries/rust/highlights.scm
```diff
@@ -220,20 +220,11 @@
   (block_comment)
 ] @comment @spell
 
-((line_comment) @comment.documentation
-  (#lua-match? @comment.documentation "^///[^/]"))
+(line_comment
+  (doc_comment)) @comment.documentation
```
lucario387, line 224: comment.documentation should only be captured for doc_comment I think

clason: Hmm... Won't that mean you highlight the comment markers and the comments differently?

lucario387: Yes.
I personally will do it like that. Unsure if that's the better behavior.

clason: I have no skin in this game, so defer to the majority opinion. How do other languages handle that? (Consistency wins.)

clason: I'm fine with either; I just want to hear the opinions before merging it to avoid later "this bad we need to change" PRs ;)

amaanq: I think it should be the entire `line_comment`/`block_comment` node for consistency

ObserverOfTime: I don't mind either way.

lucario387: I'll go with @amaanq here then.

clason: Then I'll go with the Rust maintainer's opinion. (This can always be revisited.)

ObserverOfTime (approved):

PR #6109, opened 2024-02-16
feat: improve queries/muttrc/highlights.scm

Freed-Wu:
I found almost colorscheme doesn't distinguish `@variable` and `@variable.builtin` .
So I am not sure is this PR useful.

### DIFF
--- a/queries/muttrc/highlights.scm
+++ b/queries/muttrc/highlights.scm
@@ -20,10 +20,12 @@
 
 (regex) @string.regexp
 
-[
-  (option)
-  (command_line_option)
-] @variable.builtin
+(option) @variable
+
+(command_line_option) @variable.builtin
+
+((option) @variable.builtin
+  (#not-lua-match? @variable.builtin "^my_"))
 
 (command) @keyword
 


lucario387:
Don't care about what colorschemes do to the captures. Just capture it with what you think/believe is correct.

queries/muttrc/highlights.scm
```diff
@@ -20,10 +20,11 @@
 
 (regex) @string.regexp
 
-[
-  (option)
-  (command_line_option)
-] @variable.builtin
+(command_line_option) @variable.builtin
+
+((option) @variable.builtin (#not-match? @variable.builtin "^my_"))
```
clason, line 25: match is unacceptable. either use `#eq?` or `#lua-match!`.

And make sure it's really builtin.

clason (changes requested):

queries/muttrc/highlights.scm
Freed-Wu: There are many queries use `#match?`:

```sh
❯ rg '#match' queries/*/highlights.scm
queries/vala/highlights.scm
33:  (#match? @type "^[A-Z][A-Za-z_0-9]{2,}$"))
115:  (#match? @function.macro "^assert[A-Za-z_0-9]*|error|info|debug|print|warning|warning_once$"))
...
```

lucario387: And most of those are usage where there doesn't exist an equivalent lua pattern

The match pattern in your PR, however, can be translated to lua pattern

Freed-Wu: Fixed.

clason: Lint is still failing.

Freed-Wu:
Fixed.

queries/muttrc/highlights.scm
```diff
@@ -20,10 +20,13 @@
...
-  (option)
-  (command_line_option)
-] @variable.builtin
+(command_line_option) @variable.builtin
+
+((option) @variable.builtin
+  (#not-lua-match? @variable.builtin "^my_"))
+
+((option) @variable
+  (#lua-match? @variable "^my_"))
```
amaanq, line 29: having both be a lua match is redundant, pick a base case and a more specific case to follow below

Freed-Wu: Fixed.

PR #6110, opened 2024-02-16
feat: add the typst parser

RaafatTurki:
This adds basic folding and indenting queries for typst to make it more bearable to work with.
I plan on adding more w time.

I believe I should also add comment injections for spellsitter.

### DIFF
--- a/queries/typst/folds.scm
+++ b/queries/typst/folds.scm
@@ -2,4 +2,6 @@
 
 (let) @fold
 
+(show) @fold
+
 (call) @fold
--- a/queries/typst/folds.scm
+++ b/queries/typst/folds.scm
@@ -1,3 +1,5 @@
 (set) @fold
+
 (let) @fold
+
 (call) @fold
--- a/queries/typst/folds.scm
+++ b/queries/typst/folds.scm
@@ -1,7 +1,6 @@
-(set) @fold
-
-(let) @fold
-
-(show) @fold
-
-(call) @fold
+[
+  (set)
+  (let)
+  (show)
+  (call)
+] @fold
--- a/queries/typst/folds.scm
+++ b/queries/typst/folds.scm
@@ -1,0 +1,3 @@
+(set) @fold
+(let) @fold
+(call) @fold
--- a/queries/typst/highlights.scm
+++ b/queries/typst/highlights.scm
@@ -1,0 +1,79 @@
+;; All scopes: comments
+(comment) @comment @spell
+
+;; Code
+;; punctuation
+["#"] @punctuation.special
+[":" ";" ","] @punctuation.delimiter
+["(" ")" "{" "}"] @punctuation.bracket
+;; TODO: context blocks?
+[ "[" "]" ] @punctuation.bracket
+;; operators
+[
+ "-"
+ "+"
+ "*"
+ "/"
+
+ "=="
+ "!="
+ "<"
+ "<="
+ ">"
+ ">="
+
+ "="
+
+ "in"
+ "and"
+ "or"
+ "not"
+] @operator
+;; keywords
+[ "import" "include" ] @include
+[ "let" "set" "show" ] @keyword
+;; control flow
+[ "for" "while" "break" "continue" ] @repeat
+[ "if" "else" ] @conditional
+;; special case: #for (ident) in (expr)
+(for "in" @repeat)
+;; type literals
+(number) @number
+(string) @string
+(bool) @boolean
+;; identifiers
+(builtin) @namespace
+(ident) @identifier
+;; name-value pairs
+(tagged
+  field: (ident)) @field
+;; function definitions, calls, etc.
+;; TODO: support for functions such as #calc.cos, tree view appears as
+;; (call)
+;;  item: (field)
+;;   (builtin)
+;    field: (ident)
+(call
+  item: (builtin)) @function.builtin
+(call
+  item: (ident)) @function.call
+
+;; Text
+(heading) @text.title
+(text) @text
+(strong) @text.strong
+(emph) @text.emph
+(url) @text.uri
+;; code blocks
+(raw_span) @text.literal
+(raw_span
+  (blob)) @text.literal
+(raw_blck) @text.literal
+(raw_blck
+  (blob)) @text.literal.block
+;; refs and labels
+(label) @text.reference
+(ref) @text.reference
+
+;; Math
+(math) @text.math
--- a/queries/typst/highlights.scm
+++ b/queries/typst/highlights.scm
@@ -1,13 +1,25 @@
-;; All scopes: comments
+;; comments
 (comment) @comment @spell
 
-;; Code
 ;; punctuation
-["#"] @punctuation.special
-[":" ";" ","] @punctuation.delimiter
-["(" ")" "{" "}"] @punctuation.bracket
-;; TODO: context blocks?
-[ "[" "]" ] @punctuation.bracket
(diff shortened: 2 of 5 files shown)


clason:
This is useless without a parser. If you're willing to make the effort (and maintain) queries, please add the parser and highlight queries.

RaafatTurki:
Hmm I thought the typst parser was already added.
I will add [it](https://github.com/frozolotl/tree-sitter-typst) and maintain the queries since I personally care about the ts typst support in nvim.

I'm going to put @frozolotl and myself in the maintainers list for now as they've made the parser and I'm going to maintain the queries.

clason:
No, someone made a half-effort and then didn't follow up, so I closed the PR after some time.

clason:
It was for a different parser, though, and I'd appreciate an explanation why we should take this one and not one of the others (https://github.com/uben0/tree-sitter-typst, https://github.com/SeniorMars/tree-sitter-typst).

frozolotl:
> It was for a different parser, though, and I'd appreciate an explanation why we should take this one and not one of the others (https://github.com/uben0/tree-sitter-typst, https://github.com/SeniorMars/tree-sitter-typst).

As the developer and maintainer of the grammar referenced here, I think the one by uben0 is the best choice at the moment.
It is actively being developed and works rather well overall.
The one by SeniorMars is very incomplete and unmaintained.

And last, mine, is in my opinion closer to the actual Typst parser, as that was my primary goal when developing it. However, it is not complete due to an issue with implementing a small, but important part of the grammar.
Which means that highlighting is much better on average with uben0's grammar. I would wholeheartedly recommend using that one, at least until I finally fix the issue I mentioned.

RaafatTurki:
The man himself!

I've picked the frozolotl parser for that exact reason (being the closest to the original parser)
I guess the plan of attack is using the uben0 parser until frozolotl's parser is finished (I'm willing to rework the queries when that happens)

Thanks for your honest feedback btw

queries/typst/highlights.scm
RaafatTurki, line 1: I'm going to base the highlight queries on the work of @Nagefire as he has made a beautiful set

clason: But please be aware of the different conventions in this project (see CONTRIBUTING.md).

clason: Also note the different query ordering in Helix from the rest of the ecosystem (in particular Neovim).

clason:
You should also add relevant injections (at the very least, for `comment`).

clason:
You can just run `./scripts/format-queries.lua queries/typst`.

RaafatTurki:
I've tried, It spits out
```bash
E5113: Error while calling lua chunk: /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:93: no parser for 'query' language, see :help treesitter-parsers
stack traceback:
        [C]: in function 'error'
        /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:93: in function 'add'
        /usr/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:98: in function 'new'
        /usr/share/nvim/runtime/lua/vim/treesitter.lua:61: in function '_create_parser'
        /usr/share/nvim/runtime/lua/vim/treesitter.lua:131: in function 'get_parser'
        ./scripts/format-queries.lua:392: in function 'format'
        ./scripts/format-queries.lua:408: in main chunk
```
even tho I do have the query parser installed and listed in checkhealth

clason:
You need to have the parser installed with Neovim (it relies on the bundled parser, nothing installed via nvim-treesitter).

RaafatTurki:
hmm, I don't seem to have it bundled w my neovim
after `:TSUninstall query` running `:checkhealth vim.treesitter` no longer lists the `query` parser

`:treesitter.txt` says it doesn't matter where a parser comes from and that it will load the first it finds in order from `user config > plugins > bundled` which wasn't what was happening.

> You need to have the parser installed with Neovim
I'm unsure on how to "have it"

clason:
> hmm, I don't seem to have it bundled w my neovim

Then you should complain to who you got it from ;) (And make sure you have the latest version.)

queries/typst/highlights.scm
RaafatTurki: I've changed the queries to match the neovim standard, however I don't understand what you mean by "query ordering".

Are you referring to the order way they're laid out inside the `highlights.scm` in comparison to the order they're listed in within `CONTRIBUTING.md`?

clason: I mean the order of patterns; in Neovim, all matching patterns are applied and the last wins while e.g. in Helix, the first match wins.

RaafatTurki: ahaa, I think in the case of this parser all whats needed regarding the reversed cascading specificity is moving `(comment) @comment @spell` all the way to the bottom

clason: It's not "reversed", it's the one true way ;) (Only Helix uses a different ordering still.)

RaafatTurki:
please review

queries/typst/highlights.scm
```diff
@@ -0,0 +1,131 @@
...
+  ")"
+  "{"
+  "}"
+] @punctuation.bracket
+
+; TODO: context blocks?
+[
+  "["
+  "]"
+] @punctuation.bracket
```
amaanq, line 21: combine these?

queries/typst/folds.scm
```diff
@@ -0,0 +1,7 @@
+(set) @fold
+
+(let) @fold
+
+(show) @fold
+
+(call) @fold
```
lucario387, line 7: Group these

queries/typst/highlights.scm
```diff
@@ -0,0 +1,128 @@
...
+(bool) @boolean
+
+; identifiers
+(builtin) @module.builtin
+
+(ident) @constant
+
+; name-value pairs
+(tagged
+  field: (ident)) @variable.member
```
amaanq, line 82: Based on the above `@constant` capture you'll never really see `@variable.member` highlights because it's not as specific as it should be

```suggestion
(tagged
  field: (ident) @variable.member)
```

queries/typst/highlights.scm
```diff
@@ -0,0 +1,128 @@
...
+
+(emph) @markup.italic
+
+(url) @markup.link.url
+
+; code blocks
+(raw_span) @markup.raw
+
+(raw_span
+  (blob)) @markup.raw
```
amaanq, line 112: Isn't the second one useless since you have the first one?

queries/typst/highlights.scm
```diff
@@ -0,0 +1,128 @@
...
+
+(raw_blck) @markup.raw
+
+(raw_blck
+  (blob)) @markup.raw.block
```
amaanq, line 117:
```suggestion
(raw_blck
  (blob) @markup.raw.block)
```

queries/typst/highlights.scm
RaafatTurki: hmmm, that makes sense yes

lucario387 (approved):

amaanq (approved):

clason:
@amaanq That should have been squashed ;)

amaanq:
I noticed just as I hit rebase, oops

RaafatTurki:
Ahahah, at least the commit messages were semantic.

Thanks to all who reviewed the PR and to both the current and future parser maintainers!

PR #6115, opened 2024-02-17
feat(go): add back cgo injections

lucario387:
Now that neovim have fixed quantifier issues, we can add this back


lucario387:
Nope

PR #6117, opened 2024-02-17
feat: Use json to store parse information

Freed-Wu:
Sorry for this PR.

This is a fast experiment for #5588, without any strict check.

I hope it can provide a json file to store information of parsers.

1. We can use tools like `jq -S` to format json and sort json keys
   alphabetically in linter or CI/CD. For lua, sort it is hard by machine and
   easy to make error by human.
2. A json to store information can be helpful for other projects in the
   ecosystem of tree-sitter to reuse it. Backwardly, it can also improve the
   ecosystem of treesitter and nvim-treesitter.
3. Some tree-sitter-X repo provide queries/highlights.scm and package.json with the key "tree-sitter".
   For any user who want to add tree-sitter-X to nvim-treesitter, just

- Copy queris/*.scm to nvim-treesitter_path/queries/lang_name/
- Modify parse.json according to `package.json`'s `tree-sitter`

It should can make adding a new parser easier.

This just be a demo. Discussion is acknowledged.


clason:
No thank you.

Freed-Wu:
Can we have a method to export `lua/nvim-treesitter/parse.lua`'s `list`? It should have similar function and will not break code.

clason:
Obviously. That's how all the code in this repo works. Please familiarize yourself with a project before making such sweeping changes. (You might want to start with the scripts, especially the lockfile-update and check-queries.)

lucario387:
Somewhat unrelated but still relevant: I don't want my startuptime to have +10ms or more from this change :)

Freed-Wu:
> Can we have a method to export 

Oh, I means can we provide a API to get parser's information. I think it will not increase start time.

Or any other method to get the parser's information.

lucario387:
what information? If it's just the parsers', you can always `require("nvim-treesitter.parsers").get_parsers_config()["parser_name"]`. Of course versioning for parsers are not available because tree-sitter doesn't provide a way to retrieve that info

clason:
Again, please read the code. This is _exactly_ how this plugin works, and why the information is stored in the form it is. Feature proposals that are unrelated to this plugin are not welcome.

Freed-Wu:
> what information? If it's just the parsers', you can always

Oh, I know your means:

```lua
#!/usr/bin/env -S nvim -l
print(vim.fn.json_encode(require "nvim-treesitter.parsers".get_parser_configs()))
```

> Of course versioning for parsers are not available

No problem. They can be get from `lockfile.json`.

PR #6118, opened 2024-02-17
chore(parsers): update URLs to transfer repos by theHamsta

theHamsta:
This updates the URLs for all parser repos that I have transfered to tree-sitter-grammars. This is not an user visible change since the old URLs already redirect to the new repos.

theHamsta/tree-sitter-gstlaunch has not moved yet. I tested that cloning from the new ULRs stil work.


clason:
Thanks, but I wanted to do that in one big batch (all ~70 transferred parsers). Note that you'll also have to update the lockfile, I suspect.

PR #6129, opened 2024-02-18
feat(format): extra newline between modelines

lucario387:
```diff
; inherits:...
+                                                                       
...
```


lucario387:
I won't respect newlines inside actual queries :) But top-level makes sense for queries. This will lead to less whitespaces :tada:

clason:
No strong feelings either way (although I don't mind the no-space-after-comment), so if you think this is an improvement, go ahead!

lucario387:
This just allows grouping of queries (more like before, no newline separating them). with comments being as such. I'm indifferent to this so hope others can chime in to give their view. 

This certainly will reduce the amount of unnecessary newlines, but may make reading queries harder.

ObserverOfTime (approved):
We do need a blank line after `; inherits` comments.

clason:
In general, a formatter should strive to be input-independent, so "respecting" newlines is a bit of a regression. If we could _just_ allow (and enforce!) newlines after top modeline comments, that would be preferable.

(You can always add "visual newlines" with
```
; grouping
;
(foo) @bar
```
anyway.)

lucario387:
I can reuse the PR for that

clason (approved):

lucario387:
With this PR, I found an "interesting" tidbit

`contains` and `lua-match` are technically similar, but somehow different

`(#not-contains? "foo" "; inherits")` return false on nightly
But `(#not-lua-match? "foo" "; inherits")` returns true

PR #6147, opened 2024-02-19
Add env option to install_info for parser configuration

dawsers:
This PR adds support for the `env` variable in `install_info` for parser configuration.

Certain parsers may have different grammar options at build time that may be triggered using environment variables. When `requires_generate_from_grammar` is `true`, the `env` table can be used to set those options understood by `tree-sitter generate`.

For example, the markdown parser supports `tags` and `wiki links`, but are disabled by default. To enable the extensions, the parser can be configured to be built from grammar and enable those optional features:

```lua
parser_config.markdown_inline.install_info.requires_generate_from_grammar = true
parser_config.markdown_inline.install_info.env = {
  EXTENSION_TAGS = 1,
  EXTENSION_WIKI_LINK = 1
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -576,6 +576,7 @@ parser_config.zimbu = {
     branch = "main", -- default branch in case of git repo if different from master
     generate_requires_npm = false, -- if stand-alone parser without npm dependencies
     requires_generate_from_grammar = false, -- if folder contains pre-generated src/parser.c
+    env = {}  -- add any environment variables you may need when requires_generate_from_grammar is true
   },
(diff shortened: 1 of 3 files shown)


clason:
Sorry, but this is adding too much complication for relatively little gain (and `master` is effectively frozen). If you have custom needs, it's better to install the parser manually.

dawsers:
OK, I understand,. but it is not really adding "too much complication". It is an option that is disabled by default; the code is barely 20 lines long, and provides flexibility for any parser to be built with special options.

I will do as you suggest.

Thank you.

PR #6149, opened 2024-02-20
feat(ecma): folding for function arguments

ribru17:
Adds folds for arguments passed to functions, when there are many on multiple lines. Like:
```js
let total = sum(
  123,
  99,
  36,
  346,
  100,
  22,
);
```


amaanq (approved):

amaanq:
nice thanks

PR #6154, opened 2024-02-20
feat(query): indents that reflect the formatter

ribru17:
Indentation queries that reflect the indentation of lucario's query formatter. Only applicable if `vim.opt_local.lispoptions = 'expr:1'`


lucario387:
I will not be able to check this for a while, so can you write some test files to reflect that? I think it's great as is right now.

ribru17:
Ok I think I have everything working now, but the queries are needlessly complicated and the indentation was basically fine before. I think this nonsense was a waste of time and I need to learn how to not procrastinate and go to bed. Feel free to close if too much extra code, I'm not even sure if I covered all edge cases (and 99% of users will probably be using the default lisp mode indentation anyway?)

lucario387:
We are the 1% :) 
That's more than enough tbh

lucario387 (approved):
Updated the queries to better reflect the intention.

Great thanks for the work!

.github/workflows/lint.yml
```diff
@@ -46,5 +46,5 @@ jobs:
 
       - name: Lint
         run: |
-          nvim -l scripts/format-queries.lua
+          nvim -l scripts/format-queries.lua queries
```
clason, line 49:
```suggestion
          nvim -l scripts/format-queries.lua
```

You need to rebase; I changed the script to default to current dir.

lucario387:
@clason Can I do a "partial" revert for the formatting scripts to only target `queries`? It will help if there are extra scheme/query test files in the future.

.github/workflows/lint.yml
lucario387: It also tries to format the test file, so I deliberately do that again

lucario387: Maybe there's a better solution here :thinking:

lucario387: Decided to `; format-ignore` pilled that file

ribru17:
Thanks for salvaging this!

PR #6162, opened 2024-02-21
fix(highlights): highlight @type in phpdoc

gbprod:
Currently, phpdoc types that are not union type are not highlighted, this change allows to apply `@type` group to all types.

Eg:

Without this change
2024-02-21-150834_723x177_scrot

With this change
2024-02-21-150817_728x262_scrot


amaanq (approved):

theHamsta (approved):

theHamsta:
Thank you!

PR #6170, opened 2024-02-22
slint: Update SHA of tree-sitter parser

hunger:
Fix some MISSING nodes being generated by the parser that slipped through the cracks before.


clason:
Thank you, but these sort of updates are automated; a manual PR is only needed if the queries need to be adapted to breaking changes.

PR #6174, opened 2024-02-22
feat(faust): Better highlighting of function calls, built-in variables, and member access

khiner:
Better highlighting of function calls, built-in variables, and member access.


lucario387 (approved):
:+1:

PR #6190, opened 2024-02-24
feat(fennel): in-depth form support and fixes

alexmozaidze:
This commit adds support for destructuring in **all** built-in forms/macros, but it involves a lot of tedious and ugly hacks.

# Problems

This PR has to be a draft because of the issues with queries it makes apparent.

## Limitations

Fennel is a Lisp, there is a lot of nesting, and comments can appear anywhere. Not only that, but some forms/macros have sequential syntax and key/value syntax combined (`each`, `collect`, etc), so implementing highlighting for every form/macro and have it consistent is incredibly, **incredibly** hard. I had to explicitly state where a comment might appear, as to not break the highlighting for the simplest scenarios.

The comments part may be fixed sometime in the future with https://github.com/tree-sitter/tree-sitter/issues/1454 being resolved.

## Rendering

Rendering is f_cked. Due to a lot of overriding highlights, their sheer amount and the large amount of queries, the highlighting is overloaded and sometimes it doesn't render correctly, especially on captures with large amount of text.

I do not know why rendering is inconsistent, but I assume it's because of https://github.com/neovim/neovim/issues/26325, though, I'm not sure, since `:Inspect` shows that all the extmarks are in place.

## Duplicate extmarks

There is a lot of duplicate highlights because of keyword checking, and that can't be helped. This is [expected behaviour](https://github.com/tree-sitter/tree-sitter/discussions/2101) on Tree-sitter's part.

So the only thing that can be done is to fold the duplicate captures into one and add a number of how many times it repeated for `:Inspect`.

## Maintainability

As already stated, it is incredibly hard to maintain queries, since most of them are duplicates of one another, but with slight variations. Unless a superset query language appears (like SCSS is to CSS) there is not much to be done; queries are explicit by nature.

### Potential fix

Add a new way of highlighting. Mark certain regions to be parsed with a custom Lua module that is given the sub-tree of what it needs to highlight, and returns ranges and their highlight name:

```lua
-- custom-highlighter.lua

return {
  highlight = function(tree)
    -- code...

    return {
      {20, 26, name = "@keyword"},
      -- more highlights...
    }
  end,
}
```

### DIFF
--- a/queries/fennel/highlights.scm
+++ b/queries/fennel/highlights.scm
@@ -29,13 +29,12 @@
   member: (symbol_fragment) @variable.member)
 
 (list
-  .
-  (symbol) @function.call)
+  call: (symbol) @function.call)
 
 (list
-  .
-  (multi_symbol
-    member: (symbol_fragment) @function.call .))
+  call:
+    (multi_symbol
+      member: (symbol_fragment) @function.call .))
 
 (multi_symbol_method
   ":" @punctuation.delimiter
@@ -61,7 +60,7 @@
     ; comparison
     ">" "<" ">=" "<=" "=" "~="
     ; other
-    "#" "." "?." ".."))
+    "#" ":" "." "?." ".."))
 
 ((symbol) @keyword.operator
   (#any-of? @keyword.operator
@@ -84,12 +83,12 @@
   (#any-of? @keyword.repeat "for" "each" "while"))
 
 ((symbol) @keyword.conditional
-  (#any-of? @keyword.conditional "if" "when" "match" "case"))
+  (#any-of? @keyword.conditional "if" "when" "case" "case-try" "match" "match-try"))
 
 ((symbol) @keyword
   (#any-of? @keyword
-    "global" "local" "let" "set" "var" "comment" "do" "doc" "eval-compiler" "lua" "macros" "unquote"
-    "quote" "tset" "values" "tail!"))
+    "global" "local" "let" "set" "var" "comment" "do" "doc" "eval-compiler" "lua" "macro" "macros"
+    "macrodebug" "unquote" "quote" "tset" "values" "tail!"))
 
 ((symbol) @keyword.import
   (#any-of? @keyword.import "require" "require-macros" "import-macros" "include"))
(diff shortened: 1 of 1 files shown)


clason:
> Potential fix

Yeah, no.

queries/fennel/highlights.scm
```diff
@@ -128,67 +130,479 @@
...
+  call: (symbol) @keyword
+  (#eq? @keyword "comment")) @comment
+  ; BUG: Offset for an unclosed list errors on Nightly *and* Stable:
+  ; Invalid 'end_col': out of range
+  ;
+  ; With offset, parentheses are colored properly, without it, they're part
+  ; of the comment (which is incorrect, but ok for now)
+  ;
+  ; (#offset! @comment 0 1 0 -1)
+  (#set! "priority" 101))
```
alexmozaidze, line 603: Oh, and also this bug occurs, which is unpleasant.

alexmozaidze:
Tree-sitter is the right tool, it's just that queries aren't. I suppose I should make a separate plugin for the "potential fix" that would make that work.

clason:
> Tree-sitter is the right tool, it's just that queries aren't.

That makes no sense. Queries are part of tree-sitter. Although you may enjoy https://github.com/vigoux/ratatoskr.nvim (warning: neither functional nor maintained).

clason:
Also, consider performance (`check-queries` before/after will tell you of inacceptable regressions).

alexmozaidze:
How can I check the performance? I cannot find `check-queries` anywhere in `:help`.

clason:
it's a script in this repo (also run by CI).

`./scripts/check-queries.lua fennel`

alexmozaidze:
I should probably change the way of doing things. Making a separate plugin for such a small subset of uses is not a great idea, especially when such things can be done with vanilla `vim.treesitter`.

I'm gonna go over the grammar and see if I can make a framework for adding new forms and macros so that the process of updating it is simple (like I originally intended).

I will also look into how [tree-sitter-markdown](https://github.com/tree-sitter-grammars/tree-sitter-markdown) does its parser separation and see if it's a viable option. If it is, I may be able to add new forms without making any breaking changes to the grammar, leaving working queries as-is.

clason:
Yes, the parser should do the heavy lifting, not the queries.

PR #6194, opened 2024-02-25
Special highlight group for Rust's `?` operator

AlphaKeks:
The `?` or "try" operator in Rust acts as syntax sugar for conditional checks
followed by early returns. This means that it manipulates control flow, which
is unusual for operators. Because it's a postfix operator and just a single
character, it can be hard to see, especially if your `@operator` highlights
aren't very bright colors. `?` is used heavily in Rust code, and an important
detail, so in my opinion it should stand out more.

By simply removing it from the list for `@operator` tokens and giving it a
special highlight group, which I call `@operator.controlFlow`, users can link
it to `Conditional` or anything else they see fit, to make `?` stand out
more. Because it is a subgroup of `@operator`, this will not change / break
anything for current users, but give more flexibility to those who want it.

I'm not hung up on the name; I don't know of another language which has control
flow operators, so I wasn't sure what to name it. If there is already an
established naming pattern for this, I'm happy to change the name!

### DIFF
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -372,7 +372,6 @@
   ">="
   ">>"
   ">>="
-  "?"
   "@"
   "^"
   "^="
@@ -380,6 +379,8 @@
   "|="
   "||"
 ] @operator
+
+"?" @operator.controlFlow
 
 ; Punctuation
 [


clason:
Thank you, but the captures used in this repo are standardized (upstream). This is something better done in your own config (or a stand-alone plugin).

clason:
@amaanq ?

ribru17:
Is `@keyword.exception` satisfactory? (not technically exceptions, but...)

AlphaKeks:
It's neither a keyword nor an exception.

I would definitely advocate for `@operator.<something>` considering it is currently (correctly) regarded an operator. If I remember correctly, rust-analyzer uses `operator.controlFlow` in their semantic highlights as well.

clason:
To be clear, users can already highlight it differently by adding a custom query. The question is whether _everybody_ should (would want to) highlight it differently, to the point that this should be included in "standard" colorschemes.

In this project, we value consistency across languages more highly than expressiveness in individual languages (when these two goals conflict). If someone values the latter of the former, they are welcome to create a dedicated language plugin with custom queries (and highlights).

(And we do not want to duplicate LSP semantic highlighting; it's another goal to keep these complementary.)

amaanq:
I disagree with this, it is a more special operator compared to the typical ones, but it doesn't need a special highlight/capture by default. You can freely tune this in your own config though

AlphaKeks:
> To be clear, users can already highlight it differently by adding a custom query.

I know, and I did. It is a 1-line change however, and I doubt many people are even going

lucario387:
Yes it is a 1 line change, but it's a 1line change that requires breaking the standard(as of now). And if this breaks the standard, you can be certain there will be other language queries that will try to follow the offspec standard

PR #6201, opened 2024-02-26
python highlights: make @type.builtin consistent with @type

pappasam:
Language: Python

Problem: when a `@type.builtin` is "called", its color does not change from that of a "type" to that of a "function". It should change to a function, because the symbol now represents a function being called (the constructor), not an expression of a type.

Reason: For Python's query highlights, the highlight match query for `@type.builtin` is currently placed much lower than the one for `@type`. This causes inconsistent highlighting behavior for the two.

Solution: This pull request moves the match query for `@type.builtin` to be next to the match query for `@type`.

Before:

image

After:

image

### DIFF
--- a/queries/python/highlights.scm
+++ b/queries/python/highlights.scm
@@ -10,6 +10,26 @@
 ((identifier) @type
   (#lua-match? @type "^[A-Z].*[a-z]"))
 
+((identifier) @type.builtin
+  (#any-of? @type.builtin
+    ; https://docs.python.org/3/library/exceptions.html
+    "BaseException" "Exception" "ArithmeticError" "BufferError" "LookupError" "AssertionError"
+    "AttributeError" "EOFError" "FloatingPointError" "GeneratorExit" "ImportError"
+    "ModuleNotFoundError" "IndexError" "KeyError" "KeyboardInterrupt" "MemoryError" "NameError"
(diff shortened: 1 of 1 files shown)


clason:
That may have been intentional @theHamsta

amaanq:
this PR is incorrect, and `@type.builtin` is intentionally lower to have a higher precedence than `@type`, since it's more specific

pappasam:
@amaanq Based on the inconsistent layout of `python/highlights.scm`, I'm not sure I agree. I still put `type.builtin` lower, I just don't put it WAYYYYY lower.

`@constant.builtin` and `@constant` are co-located, while `@type` and `@type.builtin` are not. Is the existing co-location a mistake?

pappasam:
Example script:

```python
class Hello:
    pass
y = Hello()
y = str()
```

Output of `:Inspect` on `Hello()`

```
Treesitter
  - @variable.python links to @variable python
  - @type.python links to Type python
  - @function.call.python links to Function python
  - @constructor.python links to Function python
```

Output of `:Inspect` on `str()`:

```
Treesitter
  - @variable.python links to @variable python
  - @function.call.python links to Function python
  - @function.builtin.python links to Function python
  - @type.builtin.python links to Type python
```

pappasam:
After some thought, I don't think this PR was the correct fix, but there's a consistency issue to address with Python queries that cause some weird match-group hierarchies between constructors, built-ins, and functions

PR #6213, opened 2024-02-28
Update Clojure highlights to be metadata aware

NoahTheDuke:
Clojure has the concept of metadata, where a symbol can hold "hidden" data. This is done with a ^ and data in front: `^java.util.Long some-foo`. The current clojure treesitter parser produces a tree that stores the metadata node on the `sym_lit` node. When a query matches on `sym_lit`, it's looking at the metadata too, which means that any queries that match the metadata will in fact leak across to the following symbol.

For example:

```
((sym_lit) @type
  (#lua-match? @type "^[%u][^/]*$"))
```

We can fix this by being more specific about what we're matching against, selecting the `name` or `namespace` field:

```
(sym_lit
  name: (sym_name) @_name
  (#lua-match? @_name "^[%u][^/%s]*$")) @type
```

### DIFF
--- a/queries/clojure/highlights.scm
+++ b/queries/clojure/highlights.scm
@@ -77,7 +77,10 @@
 
 ; Inline function variables
 ((sym_lit) @variable.builtin
-  (#lua-match? @variable.builtin "^%%"))
+  (#lua-match? @variable.builtin "^%%%d*$"))
+
+((sym_lit) @variable.builtin
+  (#eq? @variable.builtin "%&"))
 
 ; Constructor
 ((sym_lit) @constructor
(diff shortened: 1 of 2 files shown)


queries/clojure/highlights.scm
```diff
@@ -77,7 +77,8 @@
 
 ; Inline function variables
 ((sym_lit) @variable.builtin
-  (#lua-match? @variable.builtin "^%%"))
+  ; Use #match? to rely on vim's regex for alternations
+  (#match? @variable.builtin "^\\%([0-9][0-9]*|\\&)?$"))
```
clason, line 81: It's preferable to have multiple patterns instead of a single vim-match (which is _really_ slow) in this case.

NoahTheDuke: Good to know, thanks.

NoahTheDuke:
How dos reviewing work when I'm both the maintainer of the filetype and the PR author?

lucario387:
We have more reasons to click the approve button without much reasonings

queries/clojure/highlights.scm
```diff
@@ -77,7 +77,10 @@
 
 ; Inline function variables
 ((sym_lit) @variable.builtin
-  (#lua-match? @variable.builtin "^%%"))
+  (#lua-match? @variable.builtin "^%%%d*$"))
+
+((sym_lit) @variable.builtin
+  (#lua-match? @variable.builtin "^%%&$"))
```
lucario387, line 83: Possibly

```suggestion
  (#eq? @variable.builtin "%&"))
```

NoahTheDuke: Oh of course

NoahTheDuke:
Any way to get someone to look at this?

clason:
You should ping the maintainer for clojure queries.

NoahTheDuke:
That's me.

clason:
Then squashing commits would go a long way.

NoahTheDuke:
Thank you!

clason:
Nah, I can just exercise my dark side powers.

PR #6231, opened 2024-03-02
feat: parser for man pages

ribru17:
Best-effort parser for man pages, based on some common standards and such

### DIFF
--- a/queries/man/injections.scm
+++ b/queries/man/injections.scm
@@ -1,0 +1,13 @@
+(section_heading
+  (section_title) @_title
+  (block) @injection.content
+  (#any-of? @_title "SYNOPSIS" "SYNTAX" "SINTASSI" "SKŁADNIA" "СИНТАКСИС" "書式")
+  ; a heuristic to determine if we should inject c
+  (#lua-match? @injection.content "^%s+#include")
+  (#set! injection.language "c"))
+
+(subsection_heading
+  (subsection_title) @_title
+  (block) @injection.content
+  (#eq? @_title "Program source")
+  (#set! injection.language "c"))
--- a/README.md
+++ b/README.md
@@ -303,6 +303,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [luau](https://github.com/amaanq/tree-sitter-luau) (maintained by @amaanq)
 - [x] [m68k](https://github.com/grahambates/tree-sitter-m68k) (maintained by @grahambates)
 - [x] [make](https://github.com/alemuller/tree-sitter-make) (maintained by @lewis6991)
+- [x] [man](https://github.com/ribru17/tree-sitter-man) (maintained by @ribru17)
 - [x] [markdown (basic highlighting)](https://github.com/MDeiml/tree-sitter-markdown) (experimental, maintained by @MDeiml)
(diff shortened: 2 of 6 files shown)


queries/man/injections.scm
```diff
@@ -0,0 +1,13 @@
+(section_heading
+  (section_title) @_title
+  (block) @injection.content
+  (#any-of? @_title "SYNOPSIS" "SYNTAX" "SINTASSI" "SKŁADNIA" "СИНТАКСИС" "書式")
+  ; a heuristic to determine if we should inject c
+  (#lua-match? @injection.content "^%s+#include")
```
clason, line 6: I don't like this. Can the parser not do this more rigorously?

queries/man/injections.scm
```diff
@@ -0,0 +1,13 @@
+(section_heading
+  (section_title) @_title
+  (block) @injection.content
+  (#any-of? @_title "SYNOPSIS" "SYNTAX" "SINTASSI" "SKŁADNIA" "СИНТАКСИС" "書式")
```
clason, line 4: That's a bit of an arbitrary selection of languages. I'd prefer to stick to `en_US` only here.

clason:
Good idea, though!

queries/man/injections.scm
ObserverOfTime: This will also include any "Feature Test Macro Requirements" lines.

clason:
Thinking more about this, I'm a bit lukewarm. Not _everything_ needs a tree-sitter parser, and for a simple static man page Vim's simple regex syntax file seems a better tool for the job. (Especially if you need hackery to achieve feature parity.)

lucario387:
And it's still not feature parity :)
Because who said manpages can only have C code

ObserverOfTime:
I do think a man parser would be nice, but it would require a lot more development.
You'll know it's ready when manpages like `BASH(1)` and `printf(3)` look better than they do now.

clason:
> And it's still not feature parity :) Because who said manpages can only have C code

Bram. Because the `man` syntax file only considers C, too. (But that would indeed be something we could do better and worth doing.)

lucario387:
A man parser should be parsing the `man[1-9]/*` files, not the result of parsing that file though, no? :thinking:

clason:
Yes, but how is highlighting only C _not_ feature parity, if the syntax file doesn't do any more? For me, that is the definition of "parity".

ObserverOfTime:
> A man parser should be parsing the `man[1-9]/*` files, not the result of parsing that file though, no? 🤔

Technically that would be a (n)roff parser.

clason:
Anyway, `man` (output) is an even worse format than vimdoc. I don't think it makes much sense putting lipstick on _that_ pig. A better use of time would be improving the `man` plugin to do better rendering (with an `(n)roff` tree-sitter parser, natch). But if you go that route, you should just render as markdown and use _that_ parser, while you're at it.

(Neovim already has a markdown renderer for LSP hover, signature help, and (soon) diagnostics.)

ribru17:
Sounds good, I will close this. I think it is too difficult to get proper parity without the Markdown solution anyway: there are some sections that are not really syntactically different from regular text sections, yet they require C injections and the regular syntax file knows how to do it somehow. I don't know what magic is going on under the hood but it seems to be too tall of an order for this parser lol

Issue #6205, opened 2024-02-28
Consider adding a highlight capture for struct/enum/union keywords


ebkalderon:
**Is your feature request related to a problem? Please describe.**

I'd like `nvim-treesitter` to offer greater control for highlighting the keywords for product and sum types (`struct`, `enum`, `union`) separately from other generic keywords. Editors such as Sublime Text, TextMate, and Vim/Neovim offer such functionality in their syntax highlighters, allowing color schemes to color these keywords differently from other keywords, if they so choose:

![241819697-7fbda943-6775-4526-9651-ab8734401399](https://github.com/nvim-treesitter/nvim-treesitter/assets/322831/92f01733-e5af-48c3-abd3-527211b7d8a7)

At the moment, most languages' `highlights.scm` lump these keywords in with other unrelated keywords in the general-purpose `@keyword` capture. Vim/Neovim includes a highlight group name for structs/enums/unions right out of the box ([`Structure`](https://neovim.io/doc/user/syntax.html#group-name)), but `nvim-treesitter` parsers lack an equivalent to this. This makes creating a Treesitter-based color scheme which highlights the `struct` keyword in a unique color, as in the screenshot above, impossible while `nvim-treesitter` is enabled.

**Describe the solution you'd like**

We could add a new `@keyword.structure` entry to the list of standard highlight captures in the ["Keywords" section of CONTRIBUTING.md](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#keywords). This capture would map to the keywords `struct`, `enum`, and `union` in all languages that have such concepts. Color scheme authors could then link `@keyword.structure` to the standard `Structure` Vim/Neovim highlight and specify unique colors for the `struct`, `enum`, and `union` keywords separately from other language keywords.

**Describe alternatives you've considered**

We could do nothing and leave the existing set of standard highlight captures unchanged.

**Additional context**

N/A

lucario387:
Neovim also provide the functionality to do that: Overriding with your custom queries from local. So I don't see an issue with this?

ebkalderon:
Indeed it does. I was wondering if there is any appetite towards adding something like this upstream, similar to the existing `@keyword.function`, `@keyword.storage`, etc. special-case captures that `nvim-treesitter` already uses? Considering that Neovim itself has an equivalent `Structure` highlight group for this exact purpose, it feels a bit odd to me that `nvim-treesitter` doesn't do the same as well.

[...]

PR #6232, opened 2024-03-02
feat(highlights): enforce documented captures

clason:
Problem: Allowing undocumented "secret" (sub)captures makes it harder to write comprehensive colorschemes and catch inconsistent captures.

Solution: Only allow captures listed in CONTRIBUTING.md. Add useful (cross-language) subcaptures and drop language-specific or too niche ones.

Language-specific subcaptures should instead be added in user config or a custom language plugin.

**NOTE:**

* `@markup.environment.name` ~~very LaTeX specific, but without it, environments don't make sense. So I used that for language annotations as well (did I miss one); I'll upstream that and link to `Label` by default.~~ changed to `@label`
* `@markup.environment` is correspondingly changed to `@module` (as originally suggested in the Great Renamening)
* `@keyword.type` added; see #6205 (and already used in `ada`)

Missing/removed some unused (or very little used) `*.builtin` subcaptures. These should be re-added in a follow-up PR that applies these to all appropriate queries (and adds it to the documentation).

### DIFF
--- a/queries/cue/locals.scm
+++ b/queries/cue/locals.scm
@@ -19,13 +19,13 @@
 (package_identifier) @local.definition.namespace
 
 (for_clause
-  (identifier) @local.definition.variable
+  (identifier) @local.definition.var
   (expression))
 
 (for_clause
   (identifier)
-  (identifier) @local.definition.variable
+  (identifier) @local.definition.var
   (expression))
 
 (let_clause
-  (identifier) @local.definition.variable)
+  (identifier) @local.definition.var)
--- a/queries/soql/highlights.scm
+++ b/queries/soql/highlights.scm
@@ -82,9 +82,10 @@
 
 (string_literal) @string
 
-(date) @variable.readonly
-
-(date_time) @variable.readonly
+[
+  (date)
+  (date_time)
+] @string.special
 
 [
   "TRUE"
--- a/queries/vimdoc/highlights.scm
+++ b/queries/vimdoc/highlights.scm
@@ -7,11 +7,11 @@
 (column_heading) @markup.heading.4
 
 (column_heading
-  "~" @markup.heading.4.marker
+  "~" @markup.heading.4
   (#set! conceal ""))
 
 (tag
-  "*" @markup.heading.5.marker
+  "*" @label
   (#set! conceal "")
   text: (_) @label)
 
@@ -24,7 +24,7 @@
   text: (_) @markup.link)
 
 (codespan
-  "`" @markup.raw.delimiter
+  "`" @markup.raw
   (#set! conceal "")
   text: (_) @markup.raw)
 
@@ -32,14 +32,12 @@
   (#set! "priority" 90))
 
 (codeblock
-  [
-    ">"
-    (language)
-  ] @markup.raw.delimiter
+  ">" @markup.raw
+  (language) @label
   (#set! conceal ""))
 
 (block
-  "<" @markup.raw.delimiter
+  "<" @markup.raw
   (#set! conceal ""))
 
 (argument) @variable.parameter
--- a/queries/vimdoc/highlights.scm
+++ b/queries/vimdoc/highlights.scm
@@ -33,6 +33,9 @@
 
 (codeblock
   ">" @markup.raw
+  (#set! conceal ""))
+
+(codeblock
   (language) @label
   (#set! conceal ""))
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,10 +98,11 @@ As languages differ quite a lot, here is a set of captures available to you when
 #### Identifiers
 
 ```query
-@variable            ; various variable names
-@variable.builtin    ; built-in variable names (e.g. `this`)
-@variable.parameter  ; parameters of a function
-@variable.member     ; object and struct fields
+@variable                    ; various variable names
+@variable.builtin            ; built-in variable names (e.g. `this`)
+@variable.parameter          ; parameters of a function
+@variable.parameter.builtin  ; special parameters (e.g. `_`, `it`)
+@variable.member             ; object and struct fields
 
 @constant          ; constant identifiers
 @constant.builtin  ; built-in constant values
@@ -111,6 +112,7 @@ As languages differ quite a lot, here is a set of captures available to you when
 @module.builtin    ; built-in modules or namespaces
 @label             ; GOTO and other labels (e.g. `label:` in C), including heredoc labels
 ```
+
 #### Literals
 
 ```query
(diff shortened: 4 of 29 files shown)


queries/cue/locals.scm
```diff
@@ -19,13 +19,13 @@
 (package_identifier) @local.definition.namespace
 
 (for_clause
-  (identifier) @local.definition.variable
+  (identifier) @local.definition.var
```
lucario387, line 22: I don't think this is intended :thinking:

clason: Yes it is: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#locals

(This is exactly why we need to enforce captures at all levels.)

ObserverOfTime:
> * `@markup.environment.name` very LaTeX specific, but without it, environments don't make sense. So I used that for language annotations as well (did I miss one); I'll upstream that and link to `Label` by default.

We could just use `@label` instead and document it.

Missing `@constructor.builtin` (should be used in arduino).

clason:
> Missing @constructor.builtin (should be used in arduino).

Where exactly? Please give _exact_ suggestions; I haven't even heard of 99% of these languages.

ObserverOfTime:
https://github.com/nvim-treesitter/nvim-treesitter/blob/a47540fd737eb5c03ee21ee69eb8134ce5568fb6/queries/arduino/highlights.scm#L45-L47

clason:
Hmm... I think one of the guiding principles here is "only add captures that are used for several languages". No more special snowflakes!

queries/cue/locals.scm
lucario387: To me this seems weird. All other `local.definition.*` are full name, but only `local.definition.var` is shorted :thinking:

clason: Agreed. I'll extend it. (But this will be breaking again.)

queries/vimdoc/highlights.scm
```diff
@@ -33,13 +33,13 @@
 
 (codeblock
   [
-    ">"
-    (language)
-  ] @markup.raw.delimiter
+    ">" @markup.raw
+    (language) @markup.environment.name
+  ] @markup.raw
   (#set! conceal ""))
```
lucario387, line 39: I know this works, but probably better if we split this into two. Reading this seems awkward because `(language)` has 2 captures for it

queries/soql/highlights.scm
```diff
@@ -82,9 +82,9 @@
 
 (string_literal) @string
 
-(date) @variable.readonly
+(date) @variable
 
-(date_time) @variable.readonly
+(date_time) @variable
```
lucario387, line 87: Grouping these 2 captures maybe

queries/vimdoc/highlights.scm
clason: Whoops, this was me changing my mind mid-change. The grouping needs to go of course.

queries/soql/highlights.scm
clason: Should these be `@string.special` instead?

clason:
> > We could just use @Label instead and document it.
> I like that suggestion; Helix has it as well.

Turns out we already have this 🤦

queries/soql/highlights.scm
lucario387: Seems like that https://github.com/aheber/tree-sitter-sfapex/blob/857077f9e6bb04df0f769c18d32bfe036911adc8/common/soql-grammar.js#L415

But I don't think this PR should remove the fallback highlight (if possible).

clason: > But I don't think this PR should remove the fallback highlight (if possible).

What do you mean?

lucario387: As in, I think this PR should just remove the undocumented captures and use fallback if possible. Of course if the fallback no longer fits the usage criteria then we have to use another capture. 

But for this, I think it still fits, and will rather let the grammar maintainers to decide.

lucario387: Nevermind, I second-thought and would rather you to switch it to string.special :D

Doesn't make sense for a datestring to be a variable

clason: So which is it? Adapt to the right usage, or not? ;)

ObserverOfTime:
> I think one of the guiding principles here is "only add captures that are used for several languages".

It could be.

clason:
Not against it, but not a change I'm willing to make myself. I'll leave out unused "builtin" captures (constructor, attributes) in this PR and let you make a separate PR adding it to the docs and queries.

clason:
@ObserverOfTime I'd also settle for a (copy-pasteable!) suggestion for the corresponding entry in CONTRIBUTING.md, together with the promise to do that in the near future. I'm just not comfortable adding new captures that I personally don't see the usefulness of (as in "can't tell", not "doubt it").

ObserverOfTime:
I'm fine with removing the unused builtin captures for now and adding them later.

clason:
What about the three still TBD? ;)

ObserverOfTime:
Yeet.

clason:
I'd still appreciate suggestions (especially useful example) for `@attribute.builtin` and `@tag.builtin`.

ObserverOfTime:
`@attribute.builtin` could be used for e.g. `@property` in Python.
`@tag.builtin` would require an extensive `any-of` list for HTML5 tags.

clason:
I just want something to put in the docstring ;)

lucario387:
> `@tag.builtin` would require an extensive any-of list for HTML5 tags.

Can be a fallback reverse. Just have all tags as `@tag.builtin` and all capitalized tags to be `@tag`

ObserverOfTime:
Custom tags in HTML are generally lowercase. Not everything is Vue.

lucario387:
Vue-land is kebab-case and PascalCase, so generally safe from that.

any-of? is better I'd agree

ObserverOfTime:
> I just want something pithy to put in the docstring ;)

```scheme
@attribute.builtin  ; builtin annotations (e.g. `@property`)
@tag.builtin   ; builtin tag names (e.g. HTML5 tags)
```

queries/vimdoc/highlights.scm
```diff
@@ -24,22 +24,20 @@
...
   (#set! "priority" 90))
 
 (codeblock
-  [
-    ">"
-    (language)
-  ] @markup.raw.delimiter
+  ">" @markup.raw
+  (language) @label
   (#set! conceal ""))
```
lucario387, line 37: Missing an equivalent query for `language` here

clason: oh, right, because conceal will only apply to one capture.

Do multiple directives in a single pattern work yet?

lucario387: Still not. :sob: 

Would require wookayin's PR

PR #6233, opened 2024-03-02
feat(vue): switch to new parser

clason:
The new branch is renamed to `main`; the old to `fork` but still default so previous commits of this plugin remain working. We'll switch the default (and delete the old branch) after a grace period.

@amaanq queries may need adaptation; feel free to push to this PR (and mark as breaking).

### DIFF
--- a/queries/vue/highlights.scm
+++ b/queries/vue/highlights.scm
@@ -1,7 +1,5 @@
 ; inherits: html_tags
 
-(dynamic_directive_inner_value) @tag
-
 [
   "["
   "]"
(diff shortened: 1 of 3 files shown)


queries/vue/highlights.scm
```diff
@@ -1,9 +1,11 @@
 ; inherits: html_tags
 
+(dynamic_directive_inner_value) @tag
```
lucario387, line 3: it's not a tag :thinking: If anything, it is something to be injected with js

clason: Feel free to push any changes to the PR! That's quicker.

queries/vue/highlights.scm
```diff
@@ -1,7 +1,5 @@
 ; inherits: html_tags
 
-(dynamic_directive_inner_value) @tag
```
lucario387: We don't really need to inject nor highlight this with anything. The current vue parser/compiler (official vue compiler too) will fail if it's anything other than a variable name. Maybe `@variable` highlight for it

lucario387 (approved):

Issue #6238, opened 2024-03-03
Misleading capture names for type qualifiers and  lifetimes


tomtomjhj:
Some highlight capture names are scoped inappropriately
* `@type.qualifier`
    * Problem: Most type qualifiers, such as `const`, `public`, `final` are language keywords.
    * Suggested solution: Use `@keyword.type.qualifier`.
* Rust lifetime (`@keyword.storage`)
    * Problem: Rust lifetimes (except for `'static` and `'_`) are not language keywords.
    * Suggested solution
        * normal lifetimes: Use `@type.storage`, `@type.qualifier`, or `@type.qualifier.storage`.
        * `'static` and `'_`: Use `@keyword.type.qualifier` or `@type.{whatever-chosen-above}.builtin`

clason:
Queries are a community effort. Please open a PR instead.

tomtomjhj:
ok then i'll make a pr that
* changes most of `@type.qualifier` to `@keyword.type` and
* changes normal rust lifetime to `@type.qualifier`
* changes builtin rust lifetime to `@keyword.type`

clason:
To be clear, the first point makes sense to me (but someone -- not me -- will have to go through all the captures to make sure a simple rename will not break anything).

The second I leave to the maintainer @amaanq to comment on (which is always easier in the context of a PR).

clason:
> ok then i'll make a pr that

_two_ PRs. These are separate issues.

PR #6240, opened 2024-03-03
feat: `@keyword.modifier`,  `@type.modifier`

tomtomjhj:
Changes most of `@type.qualifier` to a new capture name `@keyword.modifier`.

Discussion: the first bullet point in #6238. (The second point is not addressed in the PR)

These are not changed as they seem to be used for user-provided identifiers
* https://github.com/nvim-treesitter/nvim-treesitter/blob/f8d4e5c1ba2b24b7d13830eeca394fc24c808d83/queries/blueprint/highlights.scm#L53
* https://github.com/nvim-treesitter/nvim-treesitter/blob/f8d4e5c1ba2b24b7d13830eeca394fc24c808d83/queries/chatito/highlights.scm#L33
* https://github.com/nvim-treesitter/nvim-treesitter/blob/f8d4e5c1ba2b24b7d13830eeca394fc24c808d83/queries/eds/highlights.scm#L43

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -139,7 +139,6 @@ As languages differ quite a lot, here is a set of captures available to you when
 @type             ; type or class definitions and annotations
 @type.builtin     ; built-in types
 @type.definition  ; identifiers in type definitions (e.g. `typedef <type> <identifier>` in C)
-@type.qualifier   ; type qualifiers (e.g. `const`)
 
 @attribute          ; attribute annotations (e.g. Python decorators)
 @attribute.builtin  ; builtin annotations (e.g. `@property` in Python)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -171,6 +171,7 @@ As languages differ quite a lot, here is a set of captures available to you when
 @keyword.import            ; keywords for including modules (e.g. `import` / `from` in Python)
 @keyword.storage           ; modifiers that affect storage in memory or life-time
 @keyword.type              ; keywords describing composite types (e.g. `struct`, `enum`)
+@keyword.modifier          ; keywords modifying other constructs (e.g. `const`, `static`, `public`)
 @keyword.repeat            ; keywords related to loops (e.g. `for` / `while`)
 @keyword.return            ; keywords like `return` and `yield`
 @keyword.debug             ; keywords related to debugging
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -169,7 +169,6 @@ As languages differ quite a lot, here is a set of captures available to you when
 @keyword.function          ; keywords that define a function (e.g. `func` in Go, `def` in Python)
 @keyword.operator          ; operators that are English words (e.g. `and` / `or`)
 @keyword.import            ; keywords for including modules (e.g. `import` / `from` in Python)
-@keyword.storage           ; modifiers that affect storage in memory or life-time
 @keyword.type              ; keywords describing composite types (e.g. `struct`, `enum`)
 @keyword.modifier          ; keywords modifying other constructs (e.g. `const`, `static`, `public`)
 @keyword.repeat            ; keywords related to loops (e.g. `for` / `while`)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,7 +140,7 @@ As languages differ quite a lot, here is a set of captures available to you when
 @type.builtin     ; built-in types
 @type.definition  ; identifiers in type definitions (e.g. `typedef <type> <identifier>` in C)
 
-@attribute          ; attribute annotations (e.g. Python decorators)
+@attribute          ; attribute annotations (e.g. Python decorators, Rust lifetimes)
 @attribute.builtin  ; builtin annotations (e.g. `@property` in Python)
 @property           ; the key in key/value pairs
 ```
--- a/queries/lalrpop/highlights.scm
+++ b/queries/lalrpop/highlights.scm
@@ -72,7 +72,14 @@
 ] @punctuation.delimiter
 
 (lifetime
-  (identifier) @keyword.modifier)
+  "'" @keyword.modifier)
+
+(lifetime
+  (identifier) @attribute)
+
+(lifetime
+  (identifier) @attribute.builtin
+  (#any-of? @attribute.builtin "static" "_"))
 
 (string_literal) @string
 
--- a/queries/lalrpop/highlights.scm
+++ b/queries/lalrpop/highlights.scm
@@ -72,7 +72,7 @@
 ] @punctuation.delimiter
 
 (lifetime
-  (identifier) @keyword.storage)
+  (identifier) @keyword.modifier)
 
 (string_literal) @string
 
--- a/queries/ada/highlights.scm
+++ b/queries/ada/highlights.scm
@@ -54,7 +54,7 @@
   "aliased"
   "constant"
   "renames"
-] @keyword.storage
+] @keyword.modifier
 
 [
   "with"
--- a/queries/apex/highlights.scm
+++ b/queries/apex/highlights.scm
@@ -202,7 +202,7 @@
   "protected"
   "public"
   "static"
-] @type.qualifier
+] @keyword.modifier
 
 [
   "if"
--- a/queries/bitbake/highlights.scm
+++ b/queries/bitbake/highlights.scm
@@ -86,7 +86,7 @@
 [
   "before"
   "after"
-] @keyword.storage
+] @keyword.modifier
 
 [
   "append"
--- a/queries/bitbake/highlights.scm
+++ b/queries/bitbake/highlights.scm
@@ -92,7 +92,7 @@
   "append"
   "prepend"
   "remove"
-] @type.qualifier
+] @keyword.modifier
 
 ; Variables
 [
--- a/queries/blueprint/highlights.scm
+++ b/queries/blueprint/highlights.scm
@@ -49,9 +49,6 @@
 (menu_item
   "item" @function.macro)
 
-(template_definition
-  (template_name_qualifier) @type.qualifier)
-
 (import_statement
   (gobject_library) @module)
 
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -175,7 +175,7 @@
   (type_qualifier)
   (gnu_asm_qualifier)
   "__extension__"
-] @type.qualifier
+] @keyword.modifier
 
 (linkage_specification
   "extern" @keyword.storage)
--- a/queries/c/highlights.scm
+++ b/queries/c/highlights.scm
@@ -169,7 +169,7 @@
   (type_descriptor)
 ] @type
 
-(storage_class_specifier) @keyword.storage
+(storage_class_specifier) @keyword.modifier
 
 [
   (type_qualifier)
@@ -178,7 +178,7 @@
 ] @keyword.modifier
 
 (linkage_specification
-  "extern" @keyword.storage)
+  "extern" @keyword.modifier)
 
 (type_definition
   declarator: (type_identifier) @type.definition)
--- a/queries/c_sharp/highlights.scm
+++ b/queries/c_sharp/highlights.scm
@@ -450,7 +450,7 @@
   "partial"
   "sealed"
   "virtual"
-] @type.qualifier
+] @keyword.modifier
 
 (parameter_modifier) @operator
 
--- a/queries/c_sharp/highlights.scm
+++ b/queries/c_sharp/highlights.scm
@@ -439,7 +439,7 @@
   "static"
   "volatile"
   "required"
-] @keyword.storage
+] @keyword.modifier
 
 [
   "abstract"
--- a/queries/cairo/highlights.scm
+++ b/queries/cairo/highlights.scm
@@ -138,7 +138,7 @@
 [
   "tempvar"
   "extern"
-] @keyword.storage
+] @keyword.modifier
 
 [
   "if"
--- a/queries/capnp/highlights.scm
+++ b/queries/capnp/highlights.scm
@@ -27,7 +27,7 @@
 ] @keyword
 
 ; Builtins
-"const" @type.qualifier
+"const" @keyword.modifier
 
 [
   (primitive_type)
--- a/queries/chatito/highlights.scm
+++ b/queries/chatito/highlights.scm
@@ -30,7 +30,7 @@
 
 (slot) @type
 
-(variation) @type.qualifier
+(variation) @attribute
 
 (alias) @keyword.directive
 
--- a/queries/cmake/highlights.scm
+++ b/queries/cmake/highlights.scm
@@ -136,7 +136,7 @@
   (argument_list
     .
     (argument)
-    ((argument) @_cache @keyword.storage
+    ((argument) @_cache @keyword.modifier
       .
       (argument) @_type @type
       (#any-of? @_cache "CACHE")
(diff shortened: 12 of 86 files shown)


ObserverOfTime:
If we're gonna change it I'd prefer `@keyword.modifier` (with subcaptures later).

tomtomjhj:
> `@keyword.modifier`

hmm that makes sense. things like `public` isn't really about types.

clason:
FWIW, Helix has `@keyword.storage.type` (for our `@keyword.type`) and `@keyword.storage.modifier` (for the proposed `@keyword.type.qualifier`). I proposed changing to `@keyword.type.qualifier` at first because that is easiest to do and easily reversible. In any case, I'd prefer something that would give us a 1:1 mapping to Helix captures, so `type` and `modifier` makes sense.

(Follow-up: Create a Neovim--Helix Rosetta Stone somewhere.)

ObserverOfTime:
Helix's captures in this case are almost as nonsensical as our `@type.qualifier`.

clason:
yeah, I don't like the `storage` bit, either. My concern is only making it easier to convert from one to the other, not following their example *per se*.

tomtomjhj:
What about Zed? Do they follow Helix's scheme?

clason:
No.

ObserverOfTime:
We have `@keyword.function` for `function` and `@keyword.type` for `class` and others. So the `@keyword.storage.type` capture is pretty much useless to us. We should use `@keyword.modifier` instead of `@keyword.storage.modifier` and rename *our* `@keyword.storage` to `@keyword.modifier.storage`. We can then add more subcaptures if necessary.

clason:
> We can then add more subcaptures if necessary.

I would like to avoid that, if possible -- third-level captures are best left for user customization. Also, this is widening the scope quite a bit now, so let me ask concretely:

1. `@keyword.type` -- that should not be changed?
2. `@type.qualifier` -- that should be `@keyword.modifier`?
3. `@keyword.storage` -- should that be `@keyword.modifier` as well? Or can that stay as `@keyword.storage`?

ObserverOfTime:
1. Yes.
2. Yes, since it doesn't only apply to types.
3. Depends on whether we want more subcaptures.

tomtomjhj:
Sounds good. I'll change `@keyword.type.qualifier` to `@keyword.modifier`. 

But what about remaining non-keyword `@type.qualifer`s (https://github.com/nvim-treesitter/nvim-treesitter/pull/6240#issue-2165345620)? Keep them as is or change to `@type.modifier`?

clason:
> But what about remaining non-keyword @type.qualifers (https://github.com/nvim-treesitter/nvim-treesitter/pull/6240#issue-2165345620)? Keep them as is or change to @type.modifier?

No opinion; I'll leave it to the resident esolangs afficionados whether to keep them (if some other standard capture would fit, I would prefer removing the capture entirely). Otherwise, yes, consistent renaming `qualifier`->`modifier` makes sense 👍

CONTRIBUTING.md
```diff
@@ -139,7 +139,7 @@ As languages differ quite a lot, here is a set of captures available to you when
 @type             ; type or class definitions and annotations
 @type.builtin     ; built-in types
 @type.definition  ; identifiers in type definitions (e.g. `typedef <type> <identifier>` in C)
-@type.qualifier   ; type qualifiers (e.g. `const`)
+@type.qualifier   ; type qualifiers that are not language keywords
```
clason, line 142: could you add a concrete example for that?

tomtomjhj: I don't understand what the 3 remaining `@type.qualifier` are for, but anyway Rust lifetime would be a good example. In that case I'd like to add a separate commit to change rust lifetime to `@type.modifier` in this PR (instead of a dedicated PR).

lucario387: :thinking: But at least from rustlang doc, `'static` is still a keyword, just not reserved keywords 

https://doc.rust-lang.org/reference/keywords.html

clason: OK, if we can use it for Rust, it must be important ;) I'd recommend two commits:
1. adding/using `@keyword.modifier`
2. renaming/expanding `@type.modifier`

ObserverOfTime: I'd rather not have *another* modifier capture. It's fine to use `@keyword.*` for not-exactly-keywords sometimes.

clason: Yes, if `@keyword.storage` fits well enough, no sense in adding more.

But really, this is starting to look like we need to address this mess holistically by reconsidering _all_ captures and queries that are somewhat related. I'm not too keen on that, to be honest.

And I stand by my point: Removing captures is not (easily) reversible, so if in doubt, leave it in (just renaming is fine).

tomtomjhj: > I'd rather not have another modifier capture. It's fine to use `@keyword.*` for not-exactly-keywords sometimes.

I am strongly against this. <sub>Specifically, the recent commit that removed the `.lifetime` subcapture that differentiates rust lifetimes from keywords like `const` is what triggered me to raise the issue and PR.</sub>

> Yes, if `@keyword.storage` fits well enough, no sense in adding more.

I noticed that `@keyword.storage` is used for things like `public`, `const`, etc, which basically overlaps the proposed role of `@keyword.modifier`. Even the CONTRIBUTING.md says it's for "modifiers that affect storage in memory or life-time". I'm fine with either of them as long as non-keywords don't fall into `@keyword.*`.

So here's a yet another proposal(s)
* Option 1: If the maintainers want to stick to exising new captures
    * Commit 1: Replace `@type.qualifier` used for language keyword with `@keyword.storage`
    * Commit 2: For rust lifetimes
        * use `@keyword.storage` for `'static` and `'_`
        * use `@type.qualifier` for others
* Option 2: If the maintainers don't mind introducing a new capture
    * Commit 1: Replace `@type.qualifier` used for language keyword with `@keyword.modifier`
    * Commit 2: Replace remaining `@type.qualifier` with `@type.modifier`
    * Commit 3: Replace `@keyword.storage` with `@keyword.modifier`
    * Commit 4: For rust lifetimes
        * use `@keyword.modifier` for `'static` and `'_`
        * use `@type.modifier` for others

clason: To be clear, we have to walk a line here between expressiveness for any given language and consistency across languages; these are directly competing goals. This project has decided to prioritize the second wherever they conflict.

In the past, we used to be permissive with captures (only enforce top-level captures so every capture is guaranteed to be highlighted _somehow_), but we realized that this lead to us missing too many inconsistencies (see above re: goal 2). So we locked things down so only the explicitly listed captures are valid; these are documented in Neovim so can be considered stable.

That being said, this doesn't preclude adding new ones provided they are **generally useful**; captures that only make sense for a single language are too much and better defined in a standalone filetype plugin. (This is similar to how Vim treats syntax files.) The burden of proof here rests on the proposer: if you can demonstrate (in a PR) that the capture can be used consistently in multiple languages, I'm inclined to accept it.

clason: See also https://github.com/nvim-treesitter/nvim-treesitter/issues/6205, which this should align with.

tomtomjhj: After reviewing some more queries, it seems there is no meaningful difference between the current use of `@keyword.storage` and the proposed role of `@keyword.modifier`, and generally `@keyword.modifier` fits better due to its loose definition. So I'm inclined to the Option 2 (pro-`@keyword.modifier`).

But I have a very minor concern regarding `@keyword.modifier`. The term "modifier" has a connotation that it's optional, yet some keywords that fall under `@keyword.modifier` may not be optional in some cases. The `const` keyword in Rust is a good example.

- [const function](https://doc.rust-lang.org/reference/const_eval.html?highlight=const#const-functions): `const` is optional, so `@keyword.modifier` fits.
- [const item](https://doc.rust-lang.org/reference/items/constant-items.html#constant-items): Without `const`, it's a syntax error. Maybe `@keyword.storage` fits better?
- [`*const T` pointer](https://doc.rust-lang.org/reference/types/pointer.html?highlight=const#raw-pointers-const-and-mut): Without `const`, it's a syntax error because types cannot have the form of `* T`.
- [const generic parameter](https://doc.rust-lang.org/reference/items/generics.html?highlight=const#generic-parameters): Without `const`, it may or may not be a syntax error, because it might be a syntactically valid type parameter with type bound.

Addressing this issue would require writing more complex queries. I'm not really interested in that and I think it's fine to ignore it.

clason: The question is more about guidance:
1. when would query authors use one over the other?
2. how should colorschemes color one or the other?

(Remember these are _highlight_ queries, so that's the goal here. Coloring everything differently does not add value, so it's OK to -- meaningfully -- group things even if it's semantically not fully correct.)

Concretely, I would **not** let

> The term "modifier" has a connotation that it's optional

(which is a connotation I don't see, personally) be a blocker here; this can be handled by documentation. (For me, a "modifier" is simply something that doesn't make sense in itself but only in connection with a "base" -- which `const` fits in every case, doesn't it?)

clason: (Just to be clear: I don't want to block anything here; your initiative is very welcome. I just want to find the best long-time viable strategy, given the project goals.)

tomtomjhj: ## current uses of `@type.qualifier`

I'm trying to write a description for the non-keyword `@type.qualifier`, but its uses and my proposal to apply it to rust lifetime don't seem to be consistent.

### blueprint
https://github.com/nvim-treesitter/nvim-treesitter/blob/f8d4e5c1ba2b24b7d13830eeca394fc24c808d83/queries/blueprint/highlights.scm#L52-L53

Example: `Gtk.HeaderBar` in the following.
```
template $MapsHeaderBar : Gtk.HeaderBar {
  /* probably a lot of buttons ... */
}
```
(from <https://jwestman.pages.gitlab.gnome.org/blueprint-compiler/reference/templates.html>)

`Gtk.HeaderBar` looks like some sort of inheritance base. `Gtk.HeaderBar` itself would be a valid type. Queries for other language don't specifically highlight inheritance base (C++, Java) or something similar (Rust trait bound).

#### chatito

<https://github.com/nvim-treesitter/nvim-treesitter/blob/f8d4e5c1ba2b24b7d13830eeca394fc24c808d83/queries/chatito/highlights.scm#L33>

Example: `#time_in_hours` and `#relative_time` in the following.
```
%[ask_for_delivery]
    my parcel should be delivered in @[delivery_time#time_in_hours]
    my parcel should be delivered @[delivery_time#relative_time]

@[delivery_time#time_in_hours]
    3 days
    5 hours

@[delivery_time#relative_time]
    as fast as possible
    quickly
```
(from <https://github.com/rodrigopivi/Chatito/blob/master/spec.md#212---slot>)

They modifiy the base `delivery_time`, and do not make sense without the base. This is quite different from blueprint's use.

### eds
<https://github.com/nvim-treesitter/nvim-treesitter/blob/f8d4e5c1ba2b24b7d13830eeca394fc24c808d83/queries/eds/highlights.scm#L40-L45>
This filetype is just configurations with [simple list of key/values](https://github.com/uyha/tree-sitter-eds/blob/main/grammar.js). It doesn't have much syntactic variations, so it seems to use random capture names for some specific built-in keys. I think we can ignore this case.

## what to do with `@type.qualifier`
I think this group can be expanded to other languages or removed.
* Expand: Declare that `@type.qualifier` is for things that are types themselves. Apply `@type.qualifier` for inheritance base and type bounds in the other languages.
* Remove: Don't specially treat inheritance base and type bounds.

In both cases, we are left with chatito's variation and rust lifetime. I can think of two solutions that don't involve inventing a new capture name.
* `@attribute`: This is used for things decorating existing stuff, so kinda makes sense? It also has `@attribute.builtin` subcapture, which `'static` and `'_` may fit. Also chatito's variation seems to fit.
* Alternatively for rust lifetime, use `@label`: Queries in some other project highlights rust lifetime as `@label` ([Helix](https://github.com/helix-editor/helix/blob/fd89c3c8335399e344e038f1141ea0657653a591/runtime/queries/rust/highlights.scm#L51-L56), [tree-sitter-rust](https://github.com/tree-sitter/tree-sitter-rust/blob/2d7bac5164c344c95530c16dd5c8fa1319e34308/queries/highlights.scm#L98)) (not in [Zed](https://github.com/zed-industries/zed/blob/main/crates/languages/src/rust/highlights.scm)). This seems to be an ok-ish choice as it looks the same as loop labels.

## other notes
* verilog has [`lifetime`](https://github.com/tree-sitter/tree-sitter-verilog/blob/2dfddfcb891f74a1cdb9d8ce480a997d11846e17/grammar.js#L1282), but it's highlighted `@label` in nvim-treesitter. It should be `@keyword.modifier`.
* lalrpop has lifetime, but it's a parser generator for rust, so it's basically Rust.

clason: Current usage is not consistent (for historical reasons). I'm not against changing things up if that is required for consistency. Agree that `@type.modifier` makes sense for chatito but not for blueprint.

(Also, blueprint and chatito are esolangs, feel free to adjust as needed for Rust.)

clason:
@tomtomjhj I would like to get this merged (in some form) relatively soon so I can update the upstream highlight groups before the 0.10 release. I don't mind having two captures `keyword.modifier` and `type.modifier` (but has to be "modifier", not "qualifier"), with these broad rules:
1. `keyword.modifier` is for keywords independent of the type (i.e., the modify _another keyword_, like a -- fictitious -- `const new`)
2. `type.modifier` is (only) for tokens that are part of the type _name_ (i.e., if you could speak of a -- fictitious -- `volatile int`).

I would suggest just making changes based on this guidance and see how it actually looks. Worst case, I agree that `@attribute.builtin` will do for 2. in a pinch (but the documentation should then include that). I'd prefer having both modifiers, though.

tomtomjhj:
I have implemented the changes for `@keyword.modifier` and `@type.modifier` based on the discussion so far.

> 2\. `type.modifier` is (only) for tokens that are part of the type _name_ (i.e., if you could speak of a -- fictitious -- `volatile int`).

I skimmed through more grammars and queries, but I couldn't find more *non-keyword* tokens that are part of type name.

ObserverOfTime:
So where is `@keyword.storage` used now?

lucario387:
~~Gone, disappered, reduced to atom~~

ObserverOfTime:
Oh, I didn't notice it was removed from CONTRIBUTING.md

queries/lalrpop/highlights.scm
```diff
@@ -72,7 +72,14 @@
...
 (lifetime
-  (identifier) @keyword.storage)
+  "'" @keyword.modifier)
+
+(lifetime
+  (identifier) @type.modifier)
+
+(lifetime
+  (identifier) @keyword.modifier
+  (#any-of? @keyword.modifier "static" "_"))
```
ObserverOfTime, line 82: Make them all the same.

ObserverOfTime:
I still think there's no point in a capture that will only be used in 3 places.

queries/lalrpop/highlights.scm
tomtomjhj: `@type.modifier` or `@keyword.modifier`?

tomtomjhj:
> I still think there's no point in a capture that will only be used in 3 places.

I agree

ObserverOfTime:
So replace it with another appropriate capture.
Whether `@keyword.modifier`, or `@attribute`, or something else.

clason:
I don't care either way, but if there's a chance that these will be used in the future, let's keep them. 

Otherwise attribute is preferred, and please extend its documentation accordingly. Whatever you pick will be merged, I'm calling curtains on the discussion.

ObserverOfTime (approved):

tomtomjhj:
* Removed `@type.modifier`. Use `@type` or `@attribute`.
* Use `@attribute` for Rust lifetime.

clason:
Thanks!

PR #6243, opened 2024-03-03
fix(yaml): lower string priority for helm injections

qvalentin:
see also https://github.com/nvim-treesitter/nvim-treesitter/pull/6050

### DIFF
--- a/queries/yaml/highlights.scm
+++ b/queries/yaml/highlights.scm
@@ -2,12 +2,14 @@
 
 (null_scalar) @constant.builtin
 
-(double_quote_scalar) @string
+((double_quote_scalar) @string
+  (#set! "priority" 90))
 
-(single_quote_scalar) @string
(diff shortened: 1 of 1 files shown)


kmoschcau:
You might want to change the priority to 99 instead to keep in line with the rest.

clason:
No, 90 is the documented priority for captures that should give precedence to injections.

kmoschcau:
Also if you want to make this really work, you'd need to lower ALL YAML highlights, because helm can appear anywhere in a YAML file.

clason:
Let me close this until we have a proper issue describing what is and what should be happening.

lucario387:
This is just another chain of events that happened, previous case was with PHP highlights, where there was no way to have HTML highlights without breaking PHP. Do note that this does work with old treesitter (first-win order).

PR #6244, opened 2024-03-03
Improved error message when unknown keybinding

jaydorsey:
Makes a small change to the unknown keybinding message to improve readability

image

The original concatenates the `funcname` and the traceback with no space which makes it a little harder for someone who's just googling for the error (as I was) to find something.

In my example, the original error contained `scope_decrementalstack traceback` which was kind of weird, but also as someone who isn't familiar with the codebase it didn't seem unreasonable that `scope_decrementalstack` was a key part of the error.

I added the mapping itself too, which I thought was useful when grepping my configs.

Reproduction steps: the root cause of this was I copied a treesitter configuration for reference, and it had these lines:

```lua
  incremental_selection = {
    enable = true,
    keymaps = {
      -- other stuff
      scope_decremental = '<s-cr>', --- this was breaking
    },
  },
```

This could also happen with someone making a typo with a keymap.

Update: 
Reformatted it to meet the style check. Retested it locally, I think this is still correct (my shell theme changed between this and first)

image

### DIFF
--- a/lua/nvim-treesitter/incremental_selection.lua
+++ b/lua/nvim-treesitter/incremental_selection.lua
@@ -143,7 +143,7 @@ function M.attach(bufnr)
       local rhs = M[funcname] ---@type function
 
       if not rhs then
-        utils.notify("Unknown keybinding: " .. funcname .. debug.traceback(), vim.log.levels.ERROR)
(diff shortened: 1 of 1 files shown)


clason:
The issue here is that this functionality is deprecated and will be removed (or moved to nvim-treesitter-textobjects) in 1.0. I am therefore hesitant to make any changes here that will make rebasing the `main` branch harder.

jaydorsey:
@clason thank you for the feedback. I'll close this PR

PR #6245, opened 2024-03-03
feat: add just parser

Hubro:
Following up on: https://github.com/nvim-treesitter/nvim-treesitter/pull/5696

The parser has gotten some attention since last time. Most importantly, it no longer uses c++ for the scanner, so it should now be possible to add it to nvim-treesitter.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1033,6 +1033,14 @@ list.julia = {
   maintainers = { "@theHamsta" },
 }
 
+list.just = {
+  install_info = {
+    url = "https://github.com/IndianBoy42/tree-sitter-just",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@Hubro" },
+}
+
 list.kconfig = {
   install_info = {
     url = "https://github.com/amaanq/tree-sitter-kconfig",
--- a/queries/just/folds.scm
+++ b/queries/just/folds.scm
@@ -1,0 +1,6 @@
+([
+  (recipe)
+  (string)
+  (external_command)
+] @fold
+  (#trim! @fold))
--- a/queries/just/highlights.scm
+++ b/queries/just/highlights.scm
@@ -1,0 +1,133 @@
+[
+  "true"
+  "false"
+] @boolean
+
+[
+  "if"
+  "else"
+] @keyword.conditional
+
+[
+  "alias"
+  "set"
+  "shell"
+  "mod"
+] @keyword
+
+[
+  "import"
+  "export"
+] @keyword.import
+
+[
+  ":="
+  "?"
+  "=="
+  "!="
+  "=~"
+  "@"
+  "="
+  "$"
+  "*"
+  "+"
+  "&&"
+  "@-"
+  "-@"
+  "-"
+  "/"
+  ":"
+] @operator
+
+[
+  "("
+  ")"
+  "["
+  "]"
+  "{{"
+  "}}"
+  "{"
+  "}"
+] @punctuation.bracket
+
+[
+  "`"
+  "```"
+] @punctuation.special
+
+"," @punctuation.delimiter
+
+(shebang) @keyword.directive
+
+(comment) @comment @spell
+
+[
+  (string)
+  (external_command)
+] @string
+
+(escape_sequence) @string.escape
+
+(module
+  (identifier) @module)
+
+(assignment
+  (identifier) @variable)
+
+(alias
+  (identifier) @variable)
+
+(value
+  (identifier) @variable)
+
+; Recipe definitions
+(recipe_header
+  (identifier) @function)
+
+(dependency
+  (identifier) @function.call)
+
+(dependency_expression
+  (identifier) @function.call)
+
+(parameter
+  (identifier) @variable.parameter)
+
+(dependency_expression
+  (expression
+    (value
+      (identifier) @variable.parameter)))
+
+; Fallback highlighting for recipe bodies
+(recipe
+  (recipe_body) @string
+  (#set! "priority" 90))
+
+; Ref: https://just.systems/man/en/chapter_26.html
+;(setting (identifier) @error)
+(setting
+  (identifier) @constant.builtin
+  (#any-of? @constant.builtin
+    "allow-duplicate-recipes" "dotenv-filename" "dotenv-load" "dotenv-path" "export" "fallback"
+    "ignore-comments" "positional-arguments" "tempdir" "windows-powershell" "windows-shell"))
+
+; Ref: https://just.systems/man/en/chapter_32.html
+;(recipe (attribute (identifier) @error))
+(recipe
+  (attribute
+    (identifier) @attribute)
+  (#any-of? @attribute
+    "confirm" "linux" "macos" "no-cd" "no-exit-message" "no-quiet" "private" "unix" "windows"))
+
+; Ref: https://just.systems/man/en/chapter_31.html
+;(function_call (identifier) @error)
+(function_call
+  (identifier) @function.call
+  (#any-of? @function.call
+    "arch" "num_cpus" "os" "os_family" "env_var" "env_var_or_default" "env" "invocation_directory"
+    "invocation_directory_native" "justfile" "justfile_directory" "just_executable" "quote"
+    "replace" "replace_regex" "trim" "trim_end" "trim_end_match" "trim_end_matches" "trim_start"
+    "trim_start_match" "trim_start_matches" "capitalize" "kebabcase" "lowercamelcase" "lowercase"
+    "shoutykebabcase" "shoutysnakecase" "snakecase" "titlecase" "uppercamelcase" "uppercase"
+    "absolute_path" "extension" "file_name" "file_stem" "parent_directory" "without_extension"
+    "clean" "join" "path_exists" "error" "sha256" "sha256_file" "uuid" "semver_matches"))
--- a/queries/just/indents.scm
+++ b/queries/just/indents.scm
@@ -1,0 +1,14 @@
+; Source: https://github.com/IndianBoy42/tree-sitter-just/blob/main/queries/just/indents.scm
+[
+  (recipe)
+  (string)
+  (external_command)
+] @indent.begin
+
+(string
+  [
+    "'''"
+    "\"\"\""
+  ] @indent.branch @indent.end)
+
+(comment) @indent.auto
--- a/queries/just/injections.scm
+++ b/queries/just/injections.scm
@@ -1,0 +1,41 @@
+; The right side of =~ literals
+(regex_literal
+  (_) @injection.content
+  (#set! injection.language "regex"))
+
+; Default to bash highlighting for non-shebang recipes and commands
+(recipe
+  (recipe_body
+    !shebang) @injection.content
+  (#set! injection.include-children)
+  (#set! injection.language "bash"))
+
+(external_command
+  (command_body) @injection.content
+  (#set! injection.language "bash"))
+
+; For shebang recipes, use the shebang executable name as the language by default
+(recipe
+  (recipe_body
+    (shebang
+      (language) @injection.language)) @injection.content
+  (#not-any-of? @injection.language "python3" "nodejs" "node")
+  (#set! injection.include-children))
+
+; python3 -> python
+(recipe
+  (recipe_body
+    (shebang
+      (language) @_lang)) @injection.content
+  (#eq? @_lang "python3")
+  (#set! injection.language "python")
+  (#set! injection.include-children))
+
+; node/nodejs -> javascript
+(recipe
+  (recipe_body
+    (shebang
+      (language) @_lang)) @injection.content
+  (#any-of? @_lang "node" "nodejs")
+  (#set! injection.language "javascript")
+  (#set! injection.include-children))
--- a/lockfile.json
+++ b/lockfile.json
@@ -344,6 +344,9 @@
   "julia": {
     "revision": "e84f10db8eeb8b9807786bfc658808edaa1b4fa2"
   },
+  "just": {
+    "revision": "5fe40d3622042c66465c4673b209a71e8376f902"
+  },
   "kconfig": {
     "revision": "486fea71f61ad9f3fd4072a118402e97fe88d26c"
   },
--- a/queries/just/locals.scm
+++ b/queries/just/locals.scm
@@ -1,0 +1,35 @@
+; Source: https://github.com/IndianBoy42/tree-sitter-just/blob/main/queries/just/locals.scm
+; Scope
+(recipe) @local.scope
+
+; Definitions
+(alias
+  left: (identifier) @local.definition.var)
+
+(assignment
+  left: (identifier) @local.definition.var)
+
+(module
+  name: (identifier) @local.definition.namespace)
+
+(parameter
+  name: (identifier) @local.definition.var)
+
+(recipe_header
+  name: (identifier) @local.definition.function)
+
+; References
+(alias
+  right: (identifier) @local.reference)
+
+(function_call
+  name: (identifier) @local.reference)
+
+(dependency
+  name: (identifier) @local.reference)
+
+(dependency_expression
+  name: (identifier) @local.reference)
+
+(value
+  (identifier) @local.reference)


tgross35:
Hey! Thanks for adding this.

I would really love to keep these queries in sync with the queries at the repo located at https://github.com/IndianBoy42/tree-sitter-just/tree/main/queries/just, just so there is a single source of truth as much as possible. Especially since upstream has the highlight tests.

I'm happy to make needed changes there and incorporating review, but would you mind explaining the divergences? I.e. what is incorrect with https://github.com/IndianBoy42/tree-sitter-just/blob/main/queries/just/highlights.scm

Also there are queries for folds, locals, and textobjects

clason:
Queries are editor-specific, especially the captures. There cannot be a "single source of truth" for all editors.

queries/just/highlights.scm
```diff
@@ -0,0 +1,137 @@
...
+[
+  "`"
+  "```"
+] @punctuation.special
+
+"," @punctuation.delimiter
+
+[
+  (comment)
+  (shebang)
```
lucario387, line 62: Shebang is a directive

I think it should be the same capture with Bash shebangs

queries/just/highlights.scm
```diff
@@ -0,0 +1,137 @@
...
+  (identifier) @module)
+
+[
+  (assignment
+    (identifier))
+  (alias
+    (identifier))
+  (value
+    (identifier))
+] @variable
```
lucario387, line 84: Do you mean capture for identifier, or its parenr?

queries/just/injections.scm
```diff
@@ -0,0 +1,35 @@
+; The right side of =~ literals
+((regex_literal
+  (_) @injection.content)
+  (#set! injection.language "regex"))
```
lucario387, line 4:
```suggestion
(regex_literal
  (_) @injection.content
  (#set! injection.language "regex"))
```

queries/just/injections.scm
```diff
@@ -0,0 +1,35 @@
...
+  (#set! injection.language "python")
+  (#set! injection.include-children))
+
+(recipe
+  (recipe_body
+    (shebang
+      (language) @_lang)) @injection.content
+  (#any-of? @_lang "node" "nodejs")
+  (#set! injection.language "python")
+  (#set! injection.include-children))
```
lucario387, line 35: :?, JavaScript?

tgross35:
In the upstream repo I have the source queries that use `tree-sitter-cli`'s captures https://github.com/IndianBoy42/tree-sitter-just/tree/main/queries-src then a fixup script that turns them into flavors (currently only nvim) https://github.com/IndianBoy42/tree-sitter-just/blob/main/build-flavored-queries.py. If this is generating incorrect or bad queries then I'd like to try to fix it there

queries/just/highlights.scm
Hubro: Oops, good catch

queries/just/injections.scm
Hubro: Oops again! Good catch

Hubro:
@tgross35 Since you're currently only supporting Neovim anyway, a sensible course of action would be to tweak the queries and translation script in treesitter-just to produce the same output.

I'm guessing it's not a good idea to try to maintain a translation script though. The script is likely going to be much more work to maintain over time than separate queries would be. It might be an interesting idea for a separate project though, since all that work would only need to be done once :slightly_smiling_face:

lucario387:
And then reuse the script to convert Neovim -> helix (or vice versa)

lua/nvim-treesitter/parsers.lua
```diff
@@ -944,6 +944,17 @@ list.julia = {
   maintainers = { "@theHamsta" },
 }
 
+list.just = {
+  install_info = {
+    url = "https://github.com/IndianBoy42/tree-sitter-just",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
```
clason, line 951:
```suggestion
```

Not needed if `main` is the default branch.

lua/nvim-treesitter/parsers.lua
```diff
@@ -944,6 +944,17 @@ list.julia = {
...
+  },
+  filetype = "just",
```
clason, line 953:
```suggestion
```
Not needed if the filetype is the same as the parser name.

lua/nvim-treesitter/parsers.lua
```diff
@@ -944,6 +944,17 @@ list.julia = {
...
+  maintainers = { "@Hubro" },
+  experimental = true,
```
clason, line 955: Why is this marked as "experimental"? These days we expect parser to be somewhat stable and usable _before_ they are added here.

clason:
> Also there are queries for folds, locals, and textobjects

Folds and local (using the upstream captures!) should be added here; textobjects live in a separate plugin.

lua/nvim-treesitter/parsers.lua
Hubro: Primarily because of the large changes that have happened in the parser lately, plus the fact that the README opens with "WIP:".

I've been using it for a long time though, it works great (at least for my use cases) but I thought the experimental flag was fitting.

Should I remove it?

clason: I you think it's fit for use (which is a prerequisite for it being fit for inclusion), yes.

clason: Ping; this is blocking this PR.

queries/just/injections.scm
```diff
@@ -0,0 +1,36 @@
+; The right side of =~ literals
+(regex_literal
+  (_) @injection.content
+  (#set! injection.language "regex"))
+
+; Default to bash highlighting for non-shebang recipes
+(recipe
+  (recipe_body) @injection.content
+  (#set! injection.include-children)
+  (#set! injection.language "bash"))
```
ObserverOfTime, line 10: Won't they *all* be injected with bash this way?

Hubro: I tried long and hard to find a way to match `(recipe_body)` only when it _doesn't_ have a `(shebang)` child node, but I couldn't. I found the `has-parent` and `has-ancestor` predicates in Neovim, but nothing similar for matching on child nodes.

When I tested this, it seemed to work anyway. I assumed the more specific queries below got precedence over this one, kind of like CSS. It didn't occur to me both could apply to the exact same region of text at the same time :thinking:

Do you know of a solution to this? Did I overlook anything in the regular query syntax, or are there any Neovim predicates that can solve this?

Hubro:
@tgross35 FYI:

```
Check failed!
Errors:
(x) Invalid capture @definition.namespace in locals for just.
(x) Invalid capture @definition.function in locals for just.
(x) Invalid capture @reference.variable in locals for just.
(x) Invalid capture @reference.function in locals for just.
```

queries/just/injections.scm
tgross35: There is something kind of weird with the grammars, TS and helix correctly capture the first line of the recipe as (shebang (lang)), but for some reason nvim doesn’t seem to capture that and just calls it part of the recipe body (meaning that e.g. `#!/use/bin/env python3` gets highlighted as a comment in the default language bash).

I just updated the queries yesterday to correctly handle defaults including `set shell` so that might be worth trying, but I haven’t yet gotten the chance to figure out this specific bug.

tgross35:
> @tgross35 FYI:
> 
> ```
> Check failed!
> Errors:
> (x) Invalid capture @definition.namespace in locals for just.
> (x) Invalid capture @definition.function in locals for just.
> (x) Invalid capture @reference.variable in locals for just.
> (x) Invalid capture @reference.function in locals for just.
> ```

Any chance that is this? https://github.com/IndianBoy42/tree-sitter-just/issues/118#issuecomment-1975024061

Queries are kind of in limbo right now 🙂 since TS changed their capture names and Helix is changing their precedence. I don’t want to update tree-sitter-just until there is a new version of nvim-treesitter released, just since most people are using it via git, but I assume the locals query from https://github.com/IndianBoy42/tree-sitter-just/pull/133 should work

queries/just/injections.scm
clason: >  TS and helix correctly capture the first line of the recipe as (shebang (lang))

Remember that TS (used to) and Helix (still) has a "first match wins" strategy, while Neovim (and TS and Helix will) has an "apply all and let the last match win" strategy. You cannot just copy queries.

clason:
Again, queries have to be written specifically for the editor following _their_ guidelines. You can (and by all means should) use other queries as a base, but you _cannot_ simply copy them.

(You should also be prepared to improve and adapt _these_ queries going forward; we no longer accept new languages without a dedicated maintainer.)

queries/just/injections.scm
Hubro: @clason In that case it sounds like this injection query is already acceptable for Neovim (?)

It applies a sane default first, then applies better matches if the shebang node is available. It seems to work perfectly.

Hubro:
> Any chance that is this? [IndianBoy42/tree-sitter-just#118 (comment)](https://github.com/IndianBoy42/tree-sitter-just/issues/118#issuecomment-1975024061)

Maybe, I just got the check error from GitHub Workflows, I didn't get any Lua errors.

In any case, it sounds like the idea of maintaining generic queries and translating them to various editors is unrealistic. It assumes that the only differences are capture names, which is not the case. Different editors seem to have different capture names, precedence rules, directives, predicates etc. and they are in active development. It's definitely a better idea to maintain separate queries for each editor.

clason:
Agreed, bit we _are_ working towards reducing divergences -- fairly unilaterally, but at least the precedence will be uniform since treesitter (and soon Helix) switch to the Neovim order.

tgross35:
> Again, queries have to be written specifically for the editor following _their_ guidelines. You can (and by all means should) use other queries as a base, but you _cannot_ simply copy them. Please read CONTRIBUTING.md; that file makes it very clear how queries for Neovim should look.

Yes, I have no argument with this at all. A large part of the confusion comes from the recent changes to nvim queries that is currently unreleased, meaning I am trying to figure out how to reconcile _two_ NeoVim flavors. This will get better whenever nvim-treesitter v0.10 is released, but I cannot suddenly abandon the old+stable NeoVim queries because most `tree-sitter-just` users install via git - which will hopefully change after this PR merges.

> In any case, it sounds like the idea of maintaining generic queries and translating them to various editors is unrealistic. It assumes that the only differences are capture names, which is not the case. Different editors seem to have different capture names, precedence rules, directives, predicates etc. and they are in active development. It's definitely a better idea to maintain separate queries for each editor.

The only relevant divergence I know of, aside from capture names, is precedence and I am ok keeping the helix queries outdated here. Queries were reworked a while ago to have fallbacks when e.g. `gsub` is not available. To help resolve the above issue, I added a new flavor nvim-next that aligns pretty reasonably with what you have https://github.com/IndianBoy42/tree-sitter-just/tree/main/queries-flavored/nvim-next:

- `folds.scm`: identical
- `highlights.scm`:
  - The queries in this PR call out specific function calls to highlight. I don't call out any specific functions upstream since they change pretty frequently, e.g. this is already missing the recent `home_directory` and other XDG functions, `blake3`, etc
  - https://github.com/nvim-treesitter/nvim-treesitter/pull/6245/files#diff-60755c67f56f932ceaf593e18fd2800eb10ea961400e807b2e60c9cfe5c92c69R87-R92 may have the same thing as https://github.com/nvim-treesitter/nvim-treesitter/pull/6245#discussion_r1510411676
  - Does `recipe` need a priority? I thought injections always take precedence over highlighting by default if they match
  - Just different ordering in the file. I don't care about this but Amaan did the ordering/sections in ts-just, so I have no strong reason to change the upstream
- `indents.scm`: similar, I pulled your `python3` and node overrides into upstream (removed one level of unneeded nesting). Upstream also has a default language for `set shell := ...` that you may like to pull in here.
- `locals.scm`: identical
- `textobjects.scm`: exists in upstream and works but is incomplete, not sure if you want to add it here

No luck getting formatting in the tree-sitter-just repo yet unfortunately, but that's pretty insignificant.

To be clear - I am absolutely not blocking this PR from merging if you feel maintaining separate queries is definitely a better idea. My only goal is to lower the effort for testing with different editors on my end, and to make future synchronization easier (knowing there will probably always be small tweaks needed) - but I do not intend to be the one doing the synchronization here, so what actually gets merged makes little difference to me. And this looks fine to me as long as the above comments in `highlights.scm` sound OK to you

I just find it easier to update a best effort script than to make the same changes in 4 different locations :) but most grammars don't need to worry about this because they have 0-1 query flavors in-tree, ts-just is kind of unique in having more.

clason:
> A large part of the confusion comes from the recent changes to nvim queries that is currently unreleased, meaning I am trying to figure out how to reconcile two NeoVim flavors. 

What do you mean? There's only one Neovim flavor -- the current one.

> The only relevant divergence I know of, aside from capture names, is precedence 

That is no longer a divergence; tree-sitter already changed, and Helix will, too.

> textobjects.scm: exists in upstream and works but is incomplete, not sure if you want to add it here

These are not maintained here but in `nvim-treesitter-textobjects`.

> To be clear - I am absolutely not intending to block this PR from merging if you feel maintaining separate queries is definitely a better idea. 

It's not like you could stop us ;) So, yes, please let's focus on the queries _in this repo_, which **must** align to our CONTRIBUTING.md, since the goal of this project is consistency across languages. If your goal is different, that is fine; it's perfectly possible to have different queries to choose from.

tgross35:
> > A large part of the confusion comes from the recent changes to nvim queries that is currently unreleased, meaning I am trying to figure out how to reconcile two NeoVim flavors.
> 
> What do you mean? There's only one Neovim flavor -- the current one.

We have just been using the latest tag as a checkpoint for available captures and my work adding query checks in the ts-just repo. There isn't a tag with the new queries yet, but I suppose that doesn't matter much.

> > The only relevant divergence I know of, aside from capture names, is precedence
> 
> That is no longer a divergence; tree-sitter already changed, and Helix will, too.

Right, I just meant that as an exhaustive list since it is still a divergence for now (Helix will not have this change in stable until 24.09). 

> > textobjects.scm: exists in upstream and works but is incomplete, not sure if you want to add it here
> 
> These are not maintained here but in `nvim-treesitter-textobjects`.

Thanks, nevermind that then.

> > To be clear - I am absolutely not intending to block this PR from merging if you feel maintaining separate queries is definitely a better idea.
> 
> It's not like you could stop us ;) So, yes, please let's focus on the queries _in this repo_, which **must** align to our CONTRIBUTING.md, since the goal of this project is consistency across languages. If your goal is different, that is fine; it's perfectly possible to have different queries to choose from.

Yup, just trying to understand the reasoning for any differences in case there is something to be improved in ts-just too. Everything is pretty consistent between ts-just and this PR now.

(Fyi `injection.include-children` isn't mentioned in CONTRIBUTING.md but does seem to work)

clason:
> We have just been using the latest tag as a checkpoint for available captures and my work adding query checks in the ts-just repo. There isn't a tag with the new queries yet, but I suppose that doesn't matter much.

Then you misunderstood the tag: they are not releases but convenience labels for people who don't want to upgrade to breaking changes. Since this plugin is still experimental, we don't make maintained releases; there is only "following HEAD". 

> (Fyi injection.include-children isn't mentioned in CONTRIBUTING.md but does seem to work)

Yes, because that is upstream Neovim, not nvim-treesitter.

Hubro:
> * https://github.com/nvim-treesitter/nvim-treesitter/pull/6245/files#diff-60755c67f56f932ceaf593e18fd2800eb10ea961400e807b2e60c9cfe5c92c69R87-R92 may have the same thing as [Add "just" parser #6245 (comment)](https://github.com/nvim-treesitter/nvim-treesitter/pull/6245#discussion_r1510411676)

Good catch! Fixed :slightly_smiling_face: 

> * Does `recipe` need a priority? I thought injections always take precedence over highlighting by default if they match

Not sure, but the contribution guide says:

```
@markup.raw.block      ; literal or verbatim text as a stand-alone block
                       ; (use priority 90 for blocks with injections)
```

I took this to mean that if you have a block that is supposed to get injected, use priority 90 when highlighting it.

> indents.scm: similar, I pulled your python3 and node overrides into upstream (removed one level of unneeded nesting). Upstream also has a default language for set shell := ... that you may like to pull in here.

I didn't know about the `external_command` node, I added it to injections.scm now :slightly_smiling_face:

I don't have the energy to look into the `set shell := ...` powered injections right now :smile: What happens if the first list item is "/usr/bin/env", for example. It's probably possible to write some good queries to match 99.9% of use cases, but it will take a bit of effort.

clason:
If this is done, please squash your commits.

queries/just/folds.scm
```diff
@@ -0,0 +1,7 @@
+; Source: https://github.com/IndianBoy42/tree-sitter-just/blob/main/queries/just/folds.scm
+[
+  (recipe)
+  (string)
+  (external_command)
+] @fold
+(#trim! @fold)
```
lucario387, line 7: The trim doesn't work here without grouping

queries/just/injections.scm
lucario387: Not just conflicting, but random. Random highlight, random fold, random text objects, random indents, and so forth

lucario387: Yes. Last-wins only applies for highlight queries. For all other queries all are applied.

clason: No, _all_ matching queries (including highlight queries) are always applied, in order of their appearance in the query file. In the highlight case, this means that later matches "paint over" earlier ones and only "win" in this sense -- which is very weak, since if the first capture applies a bold font and the second capture applies a red color, you get bold red text.

lucario387: But here I'm talking about the effects of overlapping injections will have on other functionalities such as highlights, indents, and more though. Svelte and Vue were heavily affected by overlapped injections before, having Typescript code to be parsed as TS and JS, breaking everything.

So yes, please try your best to ensure no two injection queries have even a same matched node

clason: Sure, I was not saying anything about injections; just that there's no magic special-casing going on: all matching captures are always applied in all cases (and not just in Neovim).

Hubro: > @Hubro a couple days ago I added the named field needed to add `!shebang` to the default field which was needed for Helix, that may help here too https://github.com/IndianBoy42/tree-sitter-just/blob/5fe40d3622042c66465c4673b209a71e8376f902/queries/just/injections.scm#L17-L21. You will need to bump to the latest grammar sha.

That helps! :smile:

queries/just/injections.scm
```diff
@@ -0,0 +1,41 @@
+; The right side of =~ literals
+(regex_literal
+  (_) @injection.content
+  (#set! injection.language "regex"))
+
+; Default to bash highlighting for non-shebang recipes and commands
+(recipe
+  (recipe_body
+    !shebang) @injection.content
```
lucario387, line 9: This does not work. From what I can see shebang is a node, not a field. That syntax is field-only

tgross35: I did add a named field to support this upstream before suggesting it, https://github.com/IndianBoy42/tree-sitter-just/pull/151. I believe @Hubro updated to something more recent than that PR already

lucario387: Sou. That'll work then

Hubro:
> If this is done, please squash your commits.

Done!

queries/just/indents.scm
```diff
@@ -0,0 +1,8 @@
+; Source: https://github.com/IndianBoy42/tree-sitter-just/blob/main/queries/just/indents.scm
+[
+  (recipe)
+  (string)
+  (external_command)
+] @indent.begin
```
lucario387, line 6: String with indent begin seems off.
I'd say just let it be

tgross35: This is because Just strips common leading whitespace for multiline strings, e..g.

```just
# becomes `foo\nbar\n`
x := '''
  foo
  bar
'''
```

ls is nice that it continues indent while you're typing a string, but indeed a bit weird that it keeps the indent after the string. Not sure if there is a better way to do this.

ObserverOfTime (approved):

Hubro:
(Squashed)

lucario387 (approved):

PR #6247, opened 2024-03-04
fix: sourcepawn grammar update

Sarrus1:
Fixes the issue raised in https://github.com/nvim-treesitter/nvim-treesitter/pull/6246.


clason:
Thanks! You need to update the hash in the lockfile, too.

lockfile.json
```diff
@@ -797,4 +797,4 @@
   "zig": {
     "revision": "0d08703e4c3f426ec61695d7617415fff97029bd"
   }
-}
+}
```
lucario387, line 800: Hehh

clason (approved):

lockfile.json
clason: We'll fix it in post.

Sarrus1: Autoformat was on :/

PR #6252, opened 2024-03-04
feat(swift): improved locals

wojciech-kulik:
I'm not very familiar with the Treesitter syntax, some detailed code review would be appreciated :)

### DIFF
--- a/queries/swift/locals.scm
+++ b/queries/swift/locals.scm
@@ -1,8 +1,53 @@
-(import_declaration
-  (identifier) @local.definition.import)
-
+; Constants
+; Constants in Swift are more like readonly variables in other languages
+; Functions
 (function_declaration
   name: (simple_identifier) @local.definition.function)
+
+; Methods
+(function_declaration
+  name: (simple_identifier) @local.definition.method)
+
+; Variables
+(statements
+  (property_declaration
+    name:
+      (pattern
+        bound_identifier: (simple_identifier) @local.definition.var)))
+
+(for_statement
+  (pattern
+    bound_identifier: (simple_identifier) @local.definition.var))
+
+; Parameters
+(function_declaration
+  (parameter
+    name: (simple_identifier) @local.definition.parameter))
+
+(lambda_function_type_parameters
+  (lambda_parameter
+    (simple_identifier) @local.definition.parameter))
+
+; Types
+name: (type_identifier) @local.definition.type
+
+; Fields
+(class_body
+  (property_declaration
+    name:
+      (pattern
+        bound_identifier: (simple_identifier) @local.definition.field)))
+
+; Enums
+(class_declaration
+  name: (type_identifier) @local.definition.enum
+  body: (enum_class_body))
+
+; Namespaces
+; In Swift there are no namespaces
+; Imports
+(import_declaration
+  (identifier) @local.definition.import)
 
 ; Scopes
 [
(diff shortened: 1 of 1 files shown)


clason:
@alex-pinkus

queries/swift/locals.scm
```diff
@@ -19,3 +64,40 @@
...
+(switch_statement
+  expr: (simple_identifier) @local.reference)
+
+(self_expression) @local.reference
+
+condition: (simple_identifier) @local.reference
+
+collection: (simple_identifier) @local.reference
+
+value: (simple_identifier) @local.reference
```
wojciech-kulik, line 103: I noticed that some queries declare just:
```
(simple_identifier) @local.reference
```

However, I would like to avoid capturing things that are declarations, so I guess this is a better approach?

wojciech-kulik:
Any chance to merge it? :)

lucario387:
If there's any swifters to review it :)
I don't

wojciech-kulik:
If there are no reviewers, I guess we will find them if we merge it and something isn't working then 😂

wojciech-kulik:
Can we merge it?

lucario387:
Will need someone that know Swift to review this :)
@Saafo May I ask you to do a favor and review this PR?

Saafo:
@lucario387 Sorry I do not use dap and I'm not familiar with scheme, so I couldn't understand this pr 🥲

wojciech-kulik:
Can someone else review it? I guess the familiarity with Swift is not very important. Not sure what's the procedure of testing locals, but anyone knowing treesitter could take a sample swift file from GitHub and check if it behaves as expected.

clason:
Just for the record: It would have helped tremendously if you had explained _what_ you were actually "improving". Lacking an explanation, we have to rely on the community maintainers of the language to evaluate things. (And locals are particularly tricky since they are not actually used in this repo and in fact being deprecated upstream.)

wojciech-kulik:
Maybe "improved" wasn't the right word :).

As you can see in the diff, I just added missing things. Some plugins use them to add extra features, like nvim-dap-virtual-text uses it to show values inline while debugging.

clason:
That would have been useful context. Without knowing which plugins, there's no way of evaluating the changes for us.

wojciech-kulik:
I decided to close this PR, because I'm not sure how these locals really work and If I paste all of it into `EditQuery` its very laggy. I'm not sure whether it's `EditQuery` problem or if these rules are not well optimized. So I decided, that it's better not to break the general performance and I found out that it's possible to have it locally instead.

lucario387:
It can be both :)

Swift parser is slow, very slow if compared to things like C

clason:
Yeah, the Swift parser is one of the worst (performance-wise) we have. But we're mostly concerned about highlighting, which locals are not used for.

wojciech-kulik:
Yeap, when I'm testing these rules in `EditQuery` my whole Neovim almost blows up :D It's completely frozen for a couple of seconds after updating queries.

clason:
Then we definitely don't want to add these ;)

wojciech-kulik:
But queries don't seem to be "complex" so I guess it's either EditQuery issue or Swift parser. Nevertheless, I guess locals are not commonly used, so we can just leave it as it is.

lucario387:
not complex doesn't mean it isn't slow :)

Most likely, the pattern you were writing were either matching too many nodes, or the pattern requires lots of node conflict resolution.

wojciech-kulik:
@alex-pinkus if you are available now, you can take a look here if there is anything worth merging from here.

Or you can just investigate why `EditQuery` is so extremely slow with these queries. Possibly speeding up these queries could speed up a general performance of the parser?

alex-pinkus:
I'm afraid I'm in the same boat as others that I don't have familiarity with locals queries.

In terms of performance, I've spent a lot of time optimizing this over the years to get it to where it is now. This includes not only reducing the complexity of the grammar (e.g. making pattern matching more permissive than it is in the compiler itself) but also finding and fixing perf bottlenecks in the tree-sitter c library. Swift is simply an extremely complex language that takes a lot of work to parse and query accurately. If you find improvements that will simplify the grammar then I'll happily review them.

wojciech-kulik:
Sure, thank you for the explanation!

PR #6255, opened 2024-03-05
Add ECMA injections for additional CSS-in-JS frameworks

mskelton:
This adds support for `css` and `keyframes` tagged templates to support additional CSS-in-JS frameworks like Emotion.

```js
const style = css`
  border: 1px solid red;
`

const animation = keyframes`
  from { opacity: 0; }
  to { opacity: 1; }
`
```

https://emotion.sh/docs/introduction

### DIFF
--- a/queries/ecma/injections.scm
+++ b/queries/ecma/injections.scm
@@ -57,6 +57,16 @@
 ((glimmer_template) @injection.content
   (#set! injection.language "glimmer"))
 
+; css`<css>`, keyframes`<css>`
+(call_expression
+  function: (identifier) @_name
+  (#any-of? @_name "css" "keyframes")
+  arguments:
+    ((template_string) @injection.content
+      (#offset! @injection.content 0 1 0 -1)
+      (#set! injection.include-children)
+      (#set! injection.language "styled")))
+
 ; styled.div`<css>`
 (call_expression
   function:
--- a/queries/ecma/injections.scm
+++ b/queries/ecma/injections.scm
@@ -17,9 +17,12 @@
   (#lua-match? @injection.language "^[a-zA-Z][a-zA-Z0-9]*$")
   (#offset! @injection.content 0 1 0 -1)
   (#set! injection.include-children)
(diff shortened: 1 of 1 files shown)


queries/ecma/injections.scm
```diff
@@ -57,6 +57,16 @@
 ((glimmer_template) @injection.content
   (#set! injection.language "glimmer"))
 
+; css`<css>`, keyframes`<css>`
+(call_expression
+  function: (identifier) @_name
+  (#match? @_name "^(css|keyframes)")
```
lucario387, line 63:
```suggestion
  (#any-of? @_name "css" "keyframes")
```

mskelton: Yes and no, it is injected as part of the more generalized injection for html, gql, etc. However the purpose of using `css` here is to inject with the `styled` grammar which supports JS interpolation which is very common in CSS-in-JS frameworks:

```js
const style = css`
  padding: 32px;
  border: 1px solid ${color};
`
```

lucario387:
@mskelton Ping, please also change the `not-eq? svg` in some queries up above to `not-any-of? svg css`. It's better to have only one injection tree per node

mskelton:
@lucario387 Sorry for the delay, just pushed a commit to disable the auto-injection of the CSS parser.

urrang:
Hi! I've been trying to track down an issue with highlighting of tagged templates for css in [Lit](https://lit.dev/) projects, and it seems to originate from this PR. Any clue what might have caused this?

When locking nvim-treesitter to 76f699dcc89642eedf69fafaef79bc235b7ef51e highlighting works fine:

image


However, if I update to 35c1a32b267c969a024ef651c8a25d7ff9018d22 it stops working:

image

Repo used for reproducing the issue: https://github.com/urrang/treesitter-lit-minimal-repro

lucario387:
please install `styled` parser, as per the commit diffs

urrang:
Ah, missed that.. Thanks a lot!

PR #6257, opened 2024-03-06
fix!: use a different angular filetype

dlvandenberg:
When the filetype is `angular.html` instead of `angular`, other HTML functionality keeps working. 

Things like using `%` to jump between opening and closing work again. 
Also specific plugin configuration for the `angular` filetype is no longer needed. Users had to write custom filehandling for plugins like:
> VonHeikemen/lsp-zero.nvim
> tpope/vim-commentary
> L3MON4D3/LuaSnip
> nvimtools/none-ls.nvim

Note: I do not know if this change should be marked as breaking. It does not change any treesitter code directly, but it does break angular highlighting for active users.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -100,6 +100,7 @@ list.angular = {
     generate_requires_npm = true,
   },
   maintainers = { "@dlvandenberg" },
+  filetype = "angular.html",
   experimental = true,
 }
 


lucario387 (approved):
Remember, this is a filetype removed from Vim because it broke HTML ftplugins and the likes. But that doesn't gatekeep adding explicit support for the filetype.

clason:
Is this an actual filetype? Otherwise it doesn't make sense to hard-code it like that, and users should instead do `vim.treesitter.start(0, 'angular')` themselves in a suitable FileType autocommand or ftplugin (or plugin).

clason:
Maybe, but if it's not a filetype, it shouldn't be declared as "filetype".

dlvandenberg:
All I need is treesitter to use the `angular` parser when I do `:set filetype=angular.html` (or `autocmd BufRead,BufEnter *.component.html set filetype=angular.html`).

If there is another way than the code in this PR, let me know. Without these changes, nvim-treesitter complains that there is no parser for the `angular.html` filetype.

Seeing as in nvim you set the filetype to `angular.html`, it would make sense (imo) to declare the filetype in nvim-treesitter as well..

clason:
That was exactly my question. Then this _is_ an official filetype, and this is fine. But be aware that this will no longer trigger on `angular` filetyped, which is breaking. If this id not desired, instead add the new filetype to the table at the top of this file.

clason:
So, back at a computer now so I could check: this is indeed not a standard Vim filetype, so the detection must come from a plugin. This means that we cannot merge this change. My recommendation is to either get the new filetype detection into Vim (which may be less controversial, but I doubt it), or let the plugin also handle treesitter. (All it needs to do is put a `vim.treesitter.start()` in their ftplugin, or your `BufRead` autocommand.)

PR #6261, opened 2024-03-07
feat: add vento support

wrapperup:
Adds support for the [Vento](https://vento.js.org/) template language
Grammar can be found here: https://github.com/ventojs/tree-sitter-vento

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2184,6 +2184,15 @@ list.vala = {
   maintainers = { "@Prince781" },
 }
 
+list.vento = {
+  install_info = {
+    url = "https://github.com/ventojs/tree-sitter-vento",
+    files = { "src/parser.c" },
+  },
+  filetype = "vto",
+  maintainers = { "@wrapperup", "@oscarotero" },
+}
+
 list.verilog = {
   install_info = {
     url = "https://github.com/tree-sitter/tree-sitter-verilog",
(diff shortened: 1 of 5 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -2184,6 +2184,16 @@ list.vala = {
   maintainers = { "@Prince781" },
 }
 
+list.vento = {
+  install_info = {
+    url = "https://github.com/ventojs/tree-sitter-vento",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
```
clason, line 2191:
```suggestion
```
Not needed if that is the default branch. You _do_ need to add an entry to the lockfile, though, to have CI run.

clason:
Please also add `comment` injections. The highlights are a bit barebones; is this really all there is to the syntax?

wrapperup:
> Please also add `comment` injections. The highlights are a bit barebones; is this really all there is to the syntax?

Yep, it's a bit like EJS where it's just JavaScript inside tags, though sometimes tags can have keywords _followed_ by JavaScript code to do special transformations on it.

lua/nvim-treesitter/parsers.lua
```diff
@@ -2184,6 +2184,15 @@ list.vala = {
   maintainers = { "@Prince781" },
 }
 
+list.vento = {
+  install_info = {
+    url = "https://github.com/ventojs/tree-sitter-vento",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  filetype = "vto",
```
clason, line 2192: You might want to make a PR to Vim adding this as a recognized filetype.

ObserverOfTime (approved):

PR #6263, opened 2024-03-07
highlights(php): fix property access using nullsafe operator

Hippo0o:
With this change, properties that are accessed using nullsafe operator will be highlighted correctly.

After | Before
Screenshot_20240307_161558

### DIFF
--- a/queries/php_only/highlights.scm
+++ b/queries/php_only/highlights.scm
@@ -153,6 +153,14 @@
 (member_access_expression
   name: (name) @property)
 
+(nullsafe_member_access_expression
+  name:
+    (variable_name
+      (name)) @property)
+
+(nullsafe_member_access_expression
+  name: (name) @property)
+
 ; Variables
 (relative_scope) @variable.builtin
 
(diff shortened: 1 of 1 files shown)


queries/php_only/highlights.scm
```diff
@@ -153,6 +153,14 @@
 (member_access_expression
   name: (name) @property)
 
+(nullsafe_member_access_expression
```
ObserverOfTime, line 156: Please make them all `@variable.member` instead of `@property` (the ones above too).

Hippo0o: i dont understand why i should change it for `nullsafe_member_access_expression` when the `member_access_expression` is `@property`

lucario387: You can change all of them, that's the point :)

ObserverOfTime: Because `@variable.member` is the correct capture. I may have missed these when I fixed the rest.

ObserverOfTime (approved):

PR #6271, opened 2024-03-09
fix(markdown): @conceal for code fence delimiters

ribru17:
All other concealed delimiters that I can see are given `@conceal` except for the code fence ones, resulting in them being the same color as the actual code fence content (not counting injectoins). Not sure if this was intentional? This PR applies the `@conceal` highlight to these delimiters as it is applied to the others

### DIFF
--- a/queries/markdown/highlights.scm
+++ b/queries/markdown/highlights.scm
@@ -54,7 +54,7 @@
   (#set! "priority" 90))
 
 (fenced_code_block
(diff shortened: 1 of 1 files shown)


clason:
This was intentional. `@conceal` is not a highlight group and only used for patterns that conceal previously captured nodes.

ribru17:
I see, thanks!

PR #6279, opened 2024-03-11
feat(swift): add init/deinit indents

lucario387:


lucario387:
Wing just keep failing :angry: 

@clason This already fixes the indent tests for swift but wing just says no

clason (approved):

clason:
Why is the wing test suddenly failing?

lucario387:
I don't know :) 
It just happens to fail across PRs

clason:
It's a tree-sitter error:
```
Error processing rule _escape_sequence
Caused by:
    regex parse error:
        u{[0-9a-fA-F]+}
          ^
    error: repetition quantifier expects a valid decimal
Error during "tree-sitter generate"
```
(I _hate_ that the compile step never reports failures!)

lucario387:
Can we do a dumb counting of "list of uninstalled parsers" and use it as exit code? 🤔

clason:
We'd probably need to adapt this every time we add parsers?

I think it's better to abort(?) or at least keep track of a bool if the exit code is nonzero (or just sum the exit code of each job).

PR #6280, opened 2024-03-11
added #set-lang-by-filetype! directive

hankthetank27:
In some languages, we'd like to be able to inject a different languages depending on the file extension/file type. For example -- in the Liquid template language, if the extension is simply `.liquid` we would like to inject HTML as the template content. But, if we have a `.js.liquid` file, we would like to inject JavaScript as the template content. As such, I added a query directive that will inject a language if the file extension of the current buffer matches the specified extension.

Usage example:
``` lisp
((template_content) @injection.content
 (#set-lang-by-filetype! "liquid" "html")
 (#set-lang-by-filetype! "js.liquid" "javascript")
 (#set-lang-by-filetype! "css.liquid" "css")
 (#set-lang-by-filetype! "scss.liquid" "scss")
 (#set! injection.combined))
```

### DIFF
--- a/lua/nvim-treesitter/query_predicates.lua
+++ b/lua/nvim-treesitter/query_predicates.lua
@@ -175,6 +175,26 @@ query.add_directive("set-lang-from-info-string!", function(match, _, bufnr, pred
   metadata["injection.language"] = get_parser_from_markdown_info_string(injection_alias)
 end, true)
 
+---@param _ (TSNode|nil)[]
+---@param _ string
+---@param bufnr integer
+---@param pred string[]
+---@param metadata table
+---@return boolean|nil
+query.add_directive("set-lang-by-filetype!", function(_, _, bufnr, pred, metadata)
+  if not valid_args("set-lang-by-filetype!", pred, 2, true) then
+    return
+  end
+  local filename = vim.fn.expand("#" .. bufnr .. ":t")
+  local extension_index = filename:find "%."
+  if not extension_index then
+    return
+  end
+  if pred[2] == filename:sub(extension_index + 1) then
+    metadata["injection.language"] = pred[3]
+  end
+end, true)
+
 -- Just avoid some annoying warnings for this directive
 query.add_directive("make-range!", function() end, true)
 


lucario387:
I don't see too much a reason to use this over `vim.treesitter.language.register`

clason:
Sorry, but what @lucario387 said. We are actively removing custom directives.

hankthetank27:
@lucario387 unless I'm misunderstanding its usage, the issue here is that `vim.treesitter.language.register` will parse the entire file with registered language. I'd like to _only_ parse the specific nodes (ie .`template_content` in the example I gave) with the injected language dependent on the file type -- as the the rest of the file should still be parsed with the liquid parser.

@clason understood that custom directives are being removed. I'm just not sure if the suggestion will solve this issue.

hankthetank27:
@clason I'm sorry but can you possibly elaborate on what you mean by that? What I'm saying is that as far as I can tell there is no way to inject a language on a node such that the language that is injected depends on the filetype...

clason:
You are welcome to write your own plugin for this; we do not want this in nvim-treesitter.

hankthetank27:
I'm aware of that and take no issue in it. Was simply seeking clarification as you insinuated there may be an existing solution for the issue I was attempting to address with my PR. I guess that didn't come across.

clason:
Sorry, the fault was mine for mobile reviewing. I didn't quite understand your point on first reading. 

In general, template languages are a poor fit for treesitter; it's oit of scope for this plugin to address this. Ideally, tree-sitter itself should address this in some way. (I believe a similar ask comes from treesitter tests.) If upstream adds new directives for this, Neovim will implement them of course. But we are discouraging adding more divergent directives that makes sharing queries harder.

hankthetank27:
No worries -- understandable. Thanks for re-reviewing.

Yeah, that all makes a lot of sense. It's certainly been a bit of a pain to implement/integrate the parser for this language correctly. Maybe I will look into on the TS end. Thank you!

lucario387:
This can be a chore, but I believe it's easier to wait for neovim core to add dynamic queries. Then you can implement this much easier

clason:
And neovim core is waiting for tree-sitter core to implement them ;)

PR #6281, opened 2024-03-11
Sync with tree-sitter-scala 8062487

rochala:
Change `lua-match` to `match` to maintain sync with `tree-sitter-scala` https://github.com/tree-sitter/tree-sitter-scala/commit/65fdb1743fa50ca47ca34feabd9cbb3a1505aebf

Add "conditional" to case in indented_cases https://github.com/tree-sitter/tree-sitter-scala/commit/b7a63ca032e24295f00233cf946694603ee9a503

### DIFF
--- a/queries/scala/highlights.scm
+++ b/queries/scala/highlights.scm
@@ -65,11 +65,11 @@
 
 ((import_declaration
   path: (identifier) @type)
-  (#lua-match? @type "^[A-Z]"))
+  (#match? @type "^[A-Z]"))
 
 ((stable_identifier
   (identifier) @type)
-  (#lua-match? @type "^[A-Z]"))
+  (#match? @type "^[A-Z]"))
 
 (export_declaration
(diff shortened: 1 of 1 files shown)


queries/scala/highlights.scm
```diff
@@ -269,23 +269,27 @@
...
+
 (operator_identifier) @operator
 
 ((identifier) @type
-  (#lua-match? @type "^[A-Z]"))
+  (#match? @type "^[A-Z]"))
 
 ((identifier) @variable.builtin
-  (#lua-match? @variable.builtin "^this$"))
+  (#match? @variable.builtin "^this$"))
```
lucario387, line 289: Upstream could have used eq? instead here

clason:
Sorry, but that's not acceptable. These are `#lua-match` for a (performance) reason. 

If you want to switch _some_ to `#eq?` or `#any-of?`, that would be welcome.

PR #6283, opened 2024-03-11
fix(scripts): actually skip lockfile updates

ObserverOfTime:
Finally.


clason (approved):

clason:
Not that we're currently skipping any files on master (and it already works -- differently -- on main)...

ObserverOfTime:
For local updates.

clason:
I use main for that.

PR #6285, opened 2024-03-12
feat: ✨ add nushell

melMass:
and register nuon as extra


clason:
Drop the gitmoji, then we can talk.

PR #6292, opened 2024-03-14
fix(vimdoc): adapt to breaking Neovim change

clason:
Problem:
https://github.com/neovim/neovim/commit/cb46f6e467268edf917cc3617b4c024a66b256de#diff-edf997b0c2d42d31828800641ac1f5a46487fe24a0d46ccf70bcd58a14f3868f
introduced a regression on how `#set!` behaves with several matches in a
single pattern, breaking conceals in, e.g., taglinks.

Solution: Separate taglink etc. patterns into concealed marker and
contents. Also add URL support for concealed markdown links.

### DIFF
--- a/queries/markdown_inline/highlights.scm
+++ b/queries/markdown_inline/highlights.scm
@@ -33,6 +33,11 @@
   ] @markup.link
   (#set! conceal ""))
 
+(inline_link
+  (link_text) @markup.link.label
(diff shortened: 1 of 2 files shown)


lucario387:
Does this happen to other places with conceal with multiple matches 🤔

clason:
Those are the only places, but yes.

PR #6300, opened 2024-03-15
feat(v): prefer to generate from grammar

ttytm:


clason:
Why? You should prefer to compile C, as this does not require treeitter CLI

ttytm:
The goal is more convince. Not keeping the generated files not in the repository but having it automatically generated at the latest development state at client site has some advantages.

From what I experienced at the `v-analyzer` repos: 
Updating the generated files in the repo upon changes to the grammar is additional effort and adds error potential. As well in terms of the growth of the repository size. E.g. simple changes that are part the commit history but produce a diff of ~200k lines. 
https://github.com/v-analyzer/v-analyzer/pull/65
https://github.com/v-analyzer/v-analyzer/pull/74

During early development this has more impact and the project is fairly young.

So I thought it's a good way and an approach e.g. Swift is choosing.

clason:
As long as the parser.c is committed, you should not set this flag. It's only for repos that do not contain the generated parser.

ttytm:
You checked the related PR at v-analyzer? It would remove the `parser.c`

ttytm:
Also, there are tree-sitter projects that have `parser.c` in their repo but `requires_generate_from_grammar` set to true.

clason:
Then you should make a PR to remove the flag.

clason:
> You checked the related PR at v-analyzer? It would remove the `parser.c`

Once this is merged, we will need to adapt. But not a second sooner.

PR #6301, opened 2024-03-15
feat(incremental_selection): Add countable motions

Joniator:
Currently it is not possible to use a count before the motions that manipulate the incremental_selection, e.g. to select 4 nodes out with 4grn

This PR changes the created keymaps to respect the count before the motion if it was set. This should not affect the current behavior if count is omitted.


clason:
Thank you, but that functionality will be removed in 1.0, so we prefer not to make changes. It's free to a good home!

PR #6311, opened 2024-03-17
feat(rust): move pub to `@keyword.modifier`

ribru17:
Trying to match this to e.g. C++ `public` keyword. Also consolidates most of the captures into one set


amaanq (approved):

amaanq:
thanks

PR #6315, opened 2024-03-17
feat(http): switch to new parser version

NTBBloodbath:
Hey, I recently updated the HTTP parser with a lot of breaking changes. This PR is intended to update the parser here so that it works correctly.

Some of the changes:
- Add `branch` field in the `install_info` table as the default branch name is `main` (I've had problems trying to install the parser if I don't set the branch manually as it tries to use `master` which does not exist, not sure if this is necessary)
- Update `http` revision in the lockfile
- Update queries
- Remove `generate_requires_npm` field in the `install_info` table, not required anymore

I have also added myself again as one of the maintainers, I now have free time again to help with that work. I hope there is no problem with that :)

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -961,9 +961,9 @@ M.configs = {
     install_info = {
       url = 'https://github.com/rest-nvim/tree-sitter-http',
       files = { 'src/parser.c' },
-      generate_requires_npm = true,
+      branch = 'main',
     },
-    maintainers = { '@amaanq' },
+    maintainers = { '@amaanq', '@NTBBloodbath' },
     tier = 2,
   },
 
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -961,7 +961,6 @@ M.configs = {
     install_info = {
       url = 'https://github.com/rest-nvim/tree-sitter-http',
       files = { 'src/parser.c' },
-      branch = 'main',
     },
     maintainers = { '@amaanq', '@NTBBloodbath' },
     tier = 2,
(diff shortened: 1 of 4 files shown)


clason:
> Add branch field in the install_info table as the default branch name is main (I've had problems trying to install the parser if I don't set the branch manually as it tries to use master which does not exist, not sure if this is necessary)

That's not needed; you _do_ need to update the commit in the lockfile, though.

lua/nvim-treesitter/parsers.lua
```diff
@@ -961,9 +961,9 @@ M.configs = {
     install_info = {
       url = 'https://github.com/rest-nvim/tree-sitter-http',
       files = { 'src/parser.c' },
-      generate_requires_npm = true,
+      branch = 'main',
```
clason, line 964:
```suggestion
```

clason:
PR should go to `master`; I'm rebasing `main` on top of all changes manually.

clason:
Also, be aware of the new (mandatory!) capture names and query format (you can run `./scripts/format-queries.lua runtime/queries/http` to automate the latter).

NTBBloodbath:
It seems like switching branches caused some weird conflicts in the parsers file. Do you want me to close this PR and make another one with the master branch as the base?

clason:
That's probably easier, yes (you can use `git checkout --theirs` to avoid the massive conflict, but it's not a large PR and you can cherry-pick the important changes).

runtime/queries/http/highlights.scm
```diff
@@ -1,48 +1,86 @@
...
+(external_body
+  file_path: (path) @text.uri)
 
 ; Comments
+
 (comment) @comment @spell
+
+; Errors
+
+(ERROR) @error
```
clason, line 86: And that _definitely_ needs to go.

NTBBloodbath: Noted, thanks for the heads-up!

NTBBloodbath:
Done in #6316, I've also formatted the queries by using the script you mentioned above and updated the highlights query capture names as well

PR #6316, opened 2024-03-17
feat(http): switch to new parser version

NTBBloodbath:
Hey, I recently updated the HTTP parser with a lot of breaking changes. This PR is intended to update the parser here so that it works correctly.

Some of the changes:
- Update `http` revision in the lockfile
- Update queries
- Remove `generate_requires_npm` field in the `install_info` table, not required anymore

I have also added myself again as one of the maintainers, I now have free time again to help with that work. I hope there is no problem with that :)

---

Supersede #6315

### DIFF
--- a/queries/http/highlights.scm
+++ b/queries/http/highlights.scm
@@ -1,5 +1,5 @@
-; Schemes
-(scheme) @type
+; Keywords
+(scheme) @module
 
 ; Methods
 (method) @function.method
@@ -7,12 +7,23 @@
 ; Constants
 (const_spec) @constant
 
+; Headers
+(header
+  name: (name) @constant)
+
 ; Variables
-(identifier) @variable
+(variable_declaration
+  name: (identifier) @variable)
 
 ; Fields
 (pair
   name: (identifier) @variable.member)
+
+; URL / Host
+(host) @string.special.url
+
+(path
+  (identifier) @string.special.url)
 
 ; Parameters
 (query_param
@@ -24,17 +35,19 @@
   "?"
   "&"
   "@"
+  "<"
 ] @operator
 
 ; Literals
-(string) @string
-
 (target_url) @string.special.url
+
+(http_version) @constant
+
+(string) @string
 
 (number) @number
 
-; (boolean) @boolean
-(null) @constant.builtin
+(boolean) @boolean
 
 ; Punctuation
 [
@@ -43,6 +56,10 @@
 ] @punctuation.bracket
 
 ":" @punctuation.delimiter
+
+; external JSON body
+(external_body
+  file_path: (path) @string.special.path)
 
 ; Comments
 (comment) @comment @spell
--- a/queries/http/injections.scm
+++ b/queries/http/injections.scm
@@ -1,11 +1,17 @@
+; Comments
 ((comment) @injection.content
   (#set! injection.language "comment"))
 
+; Body
 ((json_body) @injection.content
   (#set! injection.language "json"))
 
 ((xml_body) @injection.content
   (#set! injection.language "xml"))
 
-; ((graphql_body) @injection.content
-;  (#set! injection.language "graphql")) ; Not used as of now..
+((graphql_body) @injection.content
+  (#set! injection.language "graphql"))
+
+; Lua scripting
+((script_variable) @injection.content
+  (#set! injection.language "lua"))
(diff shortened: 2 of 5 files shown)


clason:
Please check the documented captures in CONTRIBUTING.md (which have been massively reworked since your last PR); these are the only allowed ones.

queries/http/injections.scm
```diff
@@ -1,11 +1,14 @@
-((comment) @injection.content
-  (#set! injection.language "comment"))
```
clason, line 2: Please keep the comment injection; these are standard across all languages.

NTBBloodbath:
> Please check the documented captures in CONTRIBUTING.md (which have been massively reworked since your last PR); these are the only allowed ones.

I updated the captures but it seems that I forgot one before, the CIs should already pass.

clason:
Yup, sorry!

queries/http/highlights.scm
```diff
@@ -1,19 +1,42 @@
-; Schemes
-(scheme) @type
+; Keywords
+(scheme) @module
 
 ; Methods
-(method) @function.method
+(method) @keyword
```
ObserverOfTime, line 5: Keep it as `@function.method`

NTBBloodbath:
I already removed the merging message and squashed it, let me know if anything else is needed and thanks for the guidance :)

clason:
(You might also update the README, while you're at it -- saves one bot PR.)

NTBBloodbath:
> (You might also update the README, while you're at it -- saves one bot PR.)

Done

queries/http/highlights.scm
```diff
@@ -1,19 +1,42 @@
...
 
 ; Fields
 (pair
   name: (identifier) @variable.member)
 
+; URL / Host
+(host) @string.special.url
+
+(host
+  (identifier) @string.special.url)
```
lucario387, line 35: Seems redundant

queries/http/highlights.scm
```diff
@@ -1,19 +1,42 @@
...
+  name: (identifier) @variable)
+
+(variable_declaration
+  value: (number) @number)
+
+(variable_declaration
+  value: (boolean) @boolean)
+
+(variable_declaration
+  value: (string) @string)
```
lucario387, line 25: There is already string/number/boolean below, this also seems redundant

NTBBloodbath:
I hope those were the last changes, thank you very much for the feedback.

Just a few questions

1. It seems that all the CIs are passing except for the Windows one because that one does not install the XML parser while the Linux and MacOS CI do install it, will that be a problem or can it be skipped?
2. The testing CI also seems to be failing, but from what I see it's a Rust parser issue so it shouldn't block this PR, right?

queries/http/highlights.scm
```diff
@@ -1,19 +1,39 @@
...
+  name: (identifier) @variable)
+
+(variable_declaration
+  value: (number))
+
+(variable_declaration
+  value: (boolean))
+
+(variable_declaration
+  value: (string))
```
lucario387, line 25: You don't need to write this anymore if there isn't any captures with it

lucario387 (approved):

lucario387:
> 1. It seems that all the CIs are passing except for the Windows one because that one does not install the XML parser while the Linux and MacOS CI do install it, will that be a problem or can it be skipped?
No it's fine, just ignore it
> 2. The testing CI also seems to be failing, but from what I see it's a Rust parser issue so it shouldn't block this PR, right?
Yes

queries/http/highlights.scm
NTBBloodbath: Done, thanks 😅

queries/http/highlights.scm
```diff
@@ -24,17 +35,19 @@
...
 (target_url) @string.special.url
 
+(http_version) @constant
+
+(string) @string
+
 (number) @number
 
-; (boolean) @boolean
-(null) @constant.builtin
```
ObserverOfTime, line 37: So what happened to null?

NTBBloodbath: That node had been a cleanup error when I made the initial parser version to be included here, since before the parser had its own JSON parsing. After that, the node was removed and ceased to exist in the grammar since it had no use case, so I've removed it from the highlights query as well.

ObserverOfTime (approved):

PR #6320, opened 2024-03-19
fix(php): fix highlights, add keyword captures, add tests

calebdw:
Hello!

This PR adds highlight tests for PHP, adds missing keywords, and fixes incorrect keywords.

Thanks!


lucario387:
Looks like some failed

```
            ...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:96: Error in at tests/query/highlights/php/keywords.php:76:33: expected "type.builtin", captures: { "type" }", highlights: { "type" }
            Expected objects to be the same.
            Passed in:
            (nil)
            Expected:
            (boolean) true
            
            stack traceback:
            	...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:96: in function 'check_assertions'
            	...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:119: in function <...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:118>
            	
Success	||	highlight queries tests/query/highlights/php/literals.php	
Fail	||	highlight queries tests/query/highlights/php/types.php	
            ...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:96: Error in at tests/query/highlights/php/types.php:18:23: expected "type.builtin", captures: { "type", "variable.parameter" }", highlights: { "type", "variable.parameter" }
            Expected objects to be the same.
            Passed in:
            (nil)
            Expected:
            (boolean) true
            
            stack traceback:
            	...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:96: in function 'check_assertions'
            	...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:119: in function <...eesitter/nvim-treesitter/tests/query/highlights_spec.lua:118>
```

Also shouldn't most of this be done from upstream php parser? This repo should be for specific captures that diverges from the upstream

calebdw:
Yes, I can't seem to run highlight tests locally so I created a draft pr so that the tests would run.

> Also shouldn't most of this be done from upstream php parser?

No I've already tried that---these queries completely overwrite the upstream queries (instead of extending) so the same fixes need to be applied to both places

calebdw:
@lucario387, this is ready to review now

queries/php_only/highlights.scm
```diff
@@ -265,12 +286,17 @@
   "protected"
   "public"
   "readonly"
-  "static"
+  (static_modifier)
+  (function_static_declaration
+    "static")
```
lucario387, line 291: Hmm, why

queries/php_only/highlights.scm
```diff
@@ -265,12 +286,17 @@
...
 ] @keyword.modifier
 
 [
   "return"
+  "exit"
   "yield"
+  (yield_expression
+    "from")
 ] @keyword.return
```
lucario387, line 300: I think you meant to highlight `from` in this case?

queries/php_only/highlights.scm
lucario387: Shouldn't this be capturing static

queries/php_only/highlights.scm
calebdw: "from" can only be used in the context of "yield from"---and php technically considers "yield from" a single keyword

queries/php_only/highlights.scm
calebdw: Because static is used for many things---type.builtin, variable.builtin, and keyword.modifer.

clason:
>> Also shouldn't most of this be done from upstream php parser?
> No I've already tried that---these queries completely overwrite the upstream queries (instead of extending) so the same fixes need to be applied to both places

It's about the tests, though. If only the capture names are different, there's no point in duplicating the tests. New tests _here_ only make sense if there's necessary difference, especially with predicates/directives.

(Query order is no longer a difference, since upstream changed to the Neovim order.)

queries/php_only/highlights.scm
calebdw: This is fixed now!

clason:
We can't add complete test coverage here; that's simply not feasible with nearly 300 languages.

calebdw:
Alrighty, I'll trim down the tests to only cover the more important / tricky scenarios

queries/php_only/highlights.scm
```diff
@@ -1,3 +1,168 @@
...
+  "unset"
+] @keyword
+
+[
+  (abstract_modifier)
+  "const"
+  (final_modifier)
+  (visibility_modifier)
+  (readonly_modifier)
+  (static_modifier)
```
lucario387, line 43: I still think capturing keywords is better than its parent

calebdw: How is it better?

lucario387: It's still more specific. And for neovim, priority > specificity > precedence. Because yours isn't fully specific, those which override queries may have weird highlights

calebdw: Alrighty, reverted

queries/php_only/highlights.scm
```diff
@@ -1,3 +1,170 @@
...
+  "xor"
+] @keyword.operator
+
+[
+  "fn"
+  "function"
+] @keyword.function
+
+[
+  "break"
```
lucario387, line 16: break is like `break` in for, while loop, is it correct? If yes, please use `keyword.repeat` for it, just like other languages

calebdw: I can do that!

lucario387 (approved):

calebdw:
Thanks folks!

PR #6323, opened 2024-03-19
feat(fennel): update parser and queries

alexmozaidze:
WIP version of the parser. It works fine, but there are some bugs.

If you think I should do something that's not in TODO, please let me know ^^

## TODO

- [ ] Correctly distinguish function docstring/metadata from function body (function body's first string/table gets parsed as docstring/metadata)
- [x] Detect double-quoted special table metadata keys (`:fnl/docstring` is detected, while `"fnl/docstring"` is parsed as a regular string)
- [x] Separate `;` from the comment body
- [ ] Support `comment` form (if possible)
- [ ] Ensure form error recovery doesn't leak outside its own paren scope
- [ ] Refactor internals

## Queries

I used `#has-ancestor?`, which is available in Neovim >0.10. Removing it would only make highlighting a *little* incorrect, which would be fine.
I also Removed queries I couldn't use and test, so that there are no broken queries.

## Notes on forms support

I only support forms that have destructuring, complex syntax, or forms that might need to be queried with `#has-ancestor?`/`#has-parent?`.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -531,7 +531,7 @@ list.faust = {
 list.fennel = {
   install_info = {
     url = "https://github.com/alexmozaidze/tree-sitter-fennel",
-    files = { "src/parser.c" },
+    files = { "src/parser.c", "src/scanner.c" },
   },
   maintainers = { "@alexmozaidze" },
 }
--- a/queries/fennel/highlights.scm
+++ b/queries/fennel/highlights.scm
@@ -12,44 +12,90 @@
   "]"
 ] @punctuation.bracket
 
-(nil) @constant.builtin
+[
+  (nil)
+  (nil_binding)
+] @constant.builtin
 
-(boolean) @boolean
+[
+  (boolean)
+  (boolean_binding)
+] @boolean
 
-(number) @number
+[
+  (number)
+  (number_binding)
+] @number
 
-(string) @string
+[
+  (string)
+  (string_binding)
+] @string
+
+[
+  (symbol)
+  (symbol_binding)
+] @variable
+
+(symbol_option) @keyword.directive
 
 (escape_sequence) @string.escape
-
-(symbol) @variable
 
 (multi_symbol
   "." @punctuation.delimiter
   member: (symbol_fragment) @variable.member)
 
 (list
-  .
-  (symbol) @function.call)
+  call: (symbol) @function.call)
 
 (list
-  .
-  (multi_symbol
-    member: (symbol_fragment) @function.call .))
+  call:
+    (multi_symbol
+      member: (symbol_fragment) @function.call .))
 
 (multi_symbol_method
   ":" @punctuation.delimiter
-  method: (symbol_fragment) @function.method.call .)
+  method: (symbol_fragment) @function.method.call)
 
-; Just `&` is only used in destructuring
-((symbol) @punctuation.special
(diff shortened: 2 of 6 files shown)


clason:
Why did you delete fold and local queries?

And, yes, avoiding `#has-ancestor?` would be good.

lua/nvim-treesitter/parsers.lua
```diff
@@ -531,7 +531,8 @@ list.faust = {
 list.fennel = {
   install_info = {
     url = "https://github.com/alexmozaidze/tree-sitter-fennel",
-    files = { "src/parser.c" },
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "dev",
```
clason, line 535: I don't think we want a prerelease branch here; it's OK to leave the PR up for testing, but whatever gets merged must refer to the default branch.

alexmozaidze: This is a draft. I will either force-push or make a separate PR for when the above TODO list is mostly done, and there are no major bugs.

alexmozaidze:
I couldn't configure Neovim to use tree-sitter folds, so I can't test those. I will try to add fold queries once I configure tree-sitter folds.

As for locals: I don't know much things that make use of it, so I don't think it's viable to support it. If it's absolutely necessary to support them, I will re-add the queries back, but I won't be able to guarantee that they will be correct.

clason:
> I couldn't configure Neovim to use tree-sitter folds, so I can't test those. I will try to add fold queries once I configure tree-sitter folds.

They're simple enough to easily adapt to changes; removing them is an unacceptable regression.

> As for locals: I don't know much things that make use of it, so I don't think it's viable to support it. If it's absolutely necessary to support them, I will re-add the queries back, but I won't be able to guarantee that they will be 100% correct.

They're used in `incremental_selection` and in downstream plugins; most notably `nvim-treesitter-refactor`. Again, better to leave them in, as long as tests pass.

alexmozaidze:
I seem to have fixed all the major issues.

I'm not sure if I can implement the comment form nicely, since I don't think Tree-sitter can reduce a sequence of expressions into a single one, so making it work won't be pretty. I will most likely resort to `#offset!` since it's the easiest solution I can think of.

Sorry for the radio silence, I recently got myself a job and was pretty busy. I'll refactor the parser and open a new PR in the near future with everything sorted out.

clason:
Please rebase on latest `master`.

queries/fennel/highlights.scm
```diff
@@ -12,44 +12,99 @@
...
-  (symbol) @function.call)
+  call: (symbol) @function.call)
 
 (list
-  .
-  (multi_symbol
-    member: (symbol_fragment) @function.call .))
+  call:
+    (multi_symbol
+      member: (symbol_fragment) @function.call .))
```
lucario387, line 54: After you rebase, remember to re-run the formatter as well

alexmozaidze:
I was mistaken when I said that I fixed *all* the major issues. The issue in the first todo bullet point is only partly fixed. The docstring works fine after I applied some patches, but the metadata table is not fixed; it parses a regular table as metadata if it's the only thing in function's body.

Anyhow, this version of the parser works well enough for most use-cases. I'll be fixing problems as I go.

alexmozaidze:
I'll move this to a new PR.

alexmozaidze:
Moved to #6460

PR #6329, opened 2024-03-20
feat(julia): broadcast pipe

degnbol:
Add support for capturing function call after a broadcasted pipe. Currently works for unbroadcasted pipe.
There might be a better more complicated way to do this, since this will capture the broadcasting dot as part of the `@_pipe` capture.


clason:
@savq

savq:
LGTM.

amaanq (approved):

PR #6339, opened 2024-03-21
Add multiple filetypes support for the parser configs

reegnz:
Add multiple filetypes support for individual parser configs.
This allows defining a table of multiple filetypes instead of a
single filetype.

The change also cleans up some hackery that existed before to
support parsers that had to work with multiple filetypes.


clason:
Thank you, but we're actually already going the other way for 1.0: deprecate the `filetype` field (because it mixes install info with runtime info) in favor of putting everything in the filetype table; see the `main` branch.

reegnz:
Ah, thanks for the info!

PR #6353, opened 2024-03-24
feat(bash): align substitutions with expansions

michaelPotter:
Hi, I think that command and process substitutions [ie. $(), >(), and <() ] should be highlighted the same as other expansions, which use `@punctuation.special`. This PR sets that up.

In defense of this approach, the [bash manual](https://www.gnu.org/software/bash/manual/html_node/Shell-Expansions.html) considers command and process substitutions to be forms of expansions, so they should be highlighted as such. This is traditionally how they have been highlighted in the past too.

## Screenshots
The default colorscheme links both `@punctuation.special` and `@punctuation.default` to `Special` so they DO happen to look the same in that particular colorscheme, but this might not be the case for others. I ran this command to make the difference more apparent:

```colorscheme default | hi link @punctuation.special PreProc | hi PreProc guifg=#b854d4```

### DIFF
--- a/queries/bash/highlights.scm
+++ b/queries/bash/highlights.scm
@@ -124,10 +124,15 @@
 (test_operator) @operator
 
 (command_substitution
-  "$(" @punctuation.bracket)
+  "$(" @punctuation.special
+  ")" @punctuation.special)
 
 (process_substitution
-  "<(" @punctuation.bracket)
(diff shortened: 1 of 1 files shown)


amaanq (approved):

amaanq:
thanks

PR #6359, opened 2024-03-25
feat(liquid): add new parser tree-sitter-liquid

hankthetank27:
~~experimental~~ parser for liquid template language

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1135,10 +1135,8 @@ list.liquid = {
   install_info = {
     url = "https://github.com/hankthetank27/tree-sitter-liquid",
     files = { "src/parser.c", "src/scanner.c" },
-    branch = "main",
   },
   maintainers = { "@hankthetank27" },
-  experimental = true,
 }
 
 list.liquidsoap = {
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1131,6 +1131,16 @@ list.linkerscript = {
   maintainers = { "@amaanq" },
 }
 
+list.liquid = {
+  install_info = {
+    url = "https://github.com/hankthetank27/tree-sitter-liquid",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
+  },
+  maintainers = { "@hankthetank27" },
+  experimental = true,
+}
+
 list.liquidsoap = {
   install_info = {
     url = "https://github.com/savonet/tree-sitter-liquidsoap",
--- a/queries/liquid/highlights.scm
+++ b/queries/liquid/highlights.scm
@@ -53,8 +53,8 @@
   "render"
   "section"
   "sections"
-  ]@keyword.import
-   (#set! priority 110))
+] @keyword.import
+  (#set! priority 110))
 
 ([
   "case"
@@ -66,8 +66,8 @@
   "if"
   "unless"
   "when"
-  ] @keyword.conditional
-   (#set! priority 110))
+] @keyword.conditional
+  (#set! priority 110))
 
 ([
   "break"
@@ -80,7 +80,7 @@
   "for"
   "paginate"
   "tablerow"
-  ] @keyword.repeat
+] @keyword.repeat
   (#set! priority 110))
 
 ([
--- a/queries/liquid/highlights.scm
+++ b/queries/liquid/highlights.scm
@@ -107,5 +107,5 @@
   (raw_content) @spell
   (#set! priority 111))
 
-((comment) @comment
+((comment) @comment @spell
   (#set! priority 111))
--- a/queries/liquid/highlights.scm
+++ b/queries/liquid/highlights.scm
@@ -1,27 +1,24 @@
-([
-  "{{"
-  "}}"
-  "{{-"
-  "-}}"
-  "{%"
-  "%}"
-  "{%-"
-  "-%}"
-] @tag.delimiter
+((comment) @comment @spell
   (#set! priority 110))
 
-([
-  "]"
-  "["
-  ")"
-  "("
-] @punctuation.bracket
+(raw_statement
+  (raw_content) @spell
   (#set! priority 110))
 
-([
-  ","
-  "."
-] @punctuation.delimiter
+((identifier) @variable
+  (#set! priority 110))
+
+((string) @string
+  (#set! priority 110))
+
+((boolean) @boolean
+  (#set! priority 110))
+
+((number) @number
+  (#set! priority 110))
+
+(filter
+  name: (identifier) @function.call
   (#set! priority 110))
 
 ([
@@ -46,14 +43,6 @@
   "style"
   "with"
 ] @keyword
-  (#set! priority 110))
-
-([
-  "include"
-  "render"
-  "section"
-  "sections"
-] @keyword.import
   (#set! priority 110))
 
 ([
@@ -92,32 +81,50 @@
   (#set! priority 110))
 
 ([
+  "{{"
+  "}}"
+  "{{-"
+  "-}}"
+  "{%"
+  "%}"
+  "{%-"
+  "-%}"
+] @tag.delimiter
+  (#set! priority 110))
+
+[
+  "include"
+  "render"
+  "section"
+  "sections"
+] @keyword.import
+
+[
   "|"
   ":"
   "="
-  (predicate)
+  "+"
+  "-"
+  "*"
+  "/"
+  "%"
+  "^"
+  "=="
+  "<"
+  "<="
+  "!="
+  ">="
+  ">"
 ] @operator
-  (#set! priority 110))
 
-((identifier) @variable
-  (#set! priority 110))
+[
+  "]"
+  "["
+  ")"
+  "("
+] @punctuation.bracket
 
-((string) @string
-  (#set! priority 110))
-
-((boolean) @boolean
-  (#set! priority 110))
-
-((number) @number
-  (#set! priority 110))
-
-(filter
-  name: (identifier) @function.call
-  (#set! priority 110))
-
-(raw_statement
-  (raw_content) @spell
-  (#set! priority 110))
-
-((comment) @comment @spell
-  (#set! priority 110))
+[
+  ","
+  "."
+] @punctuation.delimiter
--- a/queries/liquid/highlights.scm
+++ b/queries/liquid/highlights.scm
@@ -109,4 +109,3 @@
 
 ((comment) @comment
   (#set! priority 111))
-
--- a/queries/liquid/highlights.scm
+++ b/queries/liquid/highlights.scm
@@ -27,48 +27,60 @@
 ([
   "as"
   "assign"
-  "break"
-  "by"
   "capture"
-  "case"
-  "continue"
-  "cycle"
   "decrement"
   "echo"
-  "else"
-  "elsif"
   "endcapture"
-  "endcase"
-  "endfor"
   "endform"
-  "endif"
   "endjavascript"
-  "endpaginate"
   "endraw"
   "endschema"
   "endstyle"
-  "endtablerow"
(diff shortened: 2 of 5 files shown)


clason:
Thank you, but we no longer accept experimental parsers. If this is ready to be used, it should be marked as such; if not, please wait until it is.

lua/nvim-treesitter/parsers.lua
```diff
@@ -1131,6 +1131,16 @@ list.linkerscript = {
   maintainers = { "@amaanq" },
 }
 
+list.liquid = {
+  install_info = {
+    url = "https://github.com/hankthetank27/tree-sitter-liquid",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
```
clason, line 1138:
```suggestion
```
`branch` is only needed if it's not the default branch (regardless of the name).

queries/liquid/injections.scm
```diff
@@ -0,0 +1,21 @@
+((template_content) @injection.content
+  (#set! injection.language "html")
+  (#set! injection.combined))
```
lucario387, line 3: Please ensure that there is never any two injection queries that capture the same node

queries/liquid/highlights.scm
```diff
@@ -0,0 +1,111 @@
...
+  "{{"
+  "}}"
+  "{{-"
+  "-}}"
+  "{%"
+  "%}"
+  "{%-"
+  "-%}"
+] @tag.delimiter
+  (#set! priority 101))
```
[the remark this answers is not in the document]
hankthetank27: Since it's a template language, there were instances where the highlight group of the embedded language would conflict with injected "parent" language -- IE. it would prioritize the JavaScript highlights over the Liquid highlights when some Liquid appeared in a script tag.

Maybe this isn't the right approach? It seemed sensible to me though.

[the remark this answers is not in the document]
clason: I see. We should have a consistent approach for such template languages then.

At the very least, the priority should be higher to indicate that this is cross-language (110, in analogy for the documented value of 90 for captures that injections should override).

The different priority 102 further down should not be needed then; instead use pattern order ("last wins").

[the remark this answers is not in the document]
clason: (Note also that this approach can have unintended side effects with languages that are injected into, e.g., css.)

hankthetank27:
@clason Understandable re: experimental parsers. 

One major reason I marked it as such is related to  #6280. But as discussed it seems there is no clear way to resolve the issue of queries behaving differently on different file types until updates made upstream. That said, the parser does work as expect for the standard `.liquid` file, so I wasn't sure at what point this would be considered "usable" given the constraints.

hankthetank27:
Sure, I get that. 

I'd say that it doesn't _need_ to accommodate for this multiple file extension situation in order to be useful. In general, a `.liquid` file is the average use-case, and I believe the parser would be of benefit for anyone who wanted to use this language in nvim/nvim-treesitter. I use it every day for work, even before I implemented the custom directives, if that's of any persuasion. 

So, yeah I'm personally fine with not expecting this change to be made down the line sometime, and to have it implemented standalone with some inherent constraints, if that is what you're suggesting.

clason:
No persuasion needed; that is fine and doesn't need the "experimental" label. Just trying to be brutally clear about expectations. Another thing to keep in mind is that nvim-treesitter is not meant to be the gatekeeper for treesitter content in Neovim -- quite the opposite: we're actively working towards minimizing the reliance on this plugin; having (non-trivial) treesitter-based filetype plugins is part of this strategy.

(Basically, the long-term plan is for this repo to be an incubator for queries to be bundled in Neovim's runtime -- nothing more, nothing less. This implies that things that wouldn't be acceptable in core shouldn't be added here either.)

hankthetank27:
Totally -- thanks for clarifying & makes sense :)

Working on some of the suggestions/requests y'all made and will get back. Cheers.

queries/liquid/highlights.scm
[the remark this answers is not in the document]
hankthetank27: Right - I went ahead and changed the priorities to 110 from 101.

I tried going with 110 from 102 using the "last wins" approach, but for some reason I cannot not get the desired result.. Basically what I'm trying to achieve there is to have the tag delimiters for comments highlight as comments. The comments query is already below the delimiter query and I'm frankly quite confused as to why the the delimiters seem to still take priority.

For the time being I left the comments and raw statement 1 higher for consistency but I'll see what I can do to remove.

queries/liquid/injections.scm
hankthetank27: I went ahead and explicitly defined these node types here separately in the grammar to prevent them from being captured multiple times.

queries/liquid/highlights.scm
```diff
@@ -69,43 +69,44 @@
...
   name: (identifier) @function.call
-  (#set! priority 101))
+  (#set! priority 110))
 
 (raw_statement
   (raw_content) @spell
-  (#set! priority 102))
+  (#set! priority 111))
 
 ((comment) @comment
```
lucario387, line 110: comment should also have spell

queries/liquid/injections.scm
```diff
@@ -0,0 +1,18 @@
...
+  (#set! injection.language "json")
+  (#set! injection.combined))
+
+((style_content) @injection.content
+  (#set! injection.language "css")
+  (#set! injection.combined))
+
+((js_content) @injection.content
+  (#set! injection.language "javascript")
+  (#set! injection.combined))
```
lucario387, line 15: I see that it's inside a `javascript_statement` node, so I think it should be
```suggestion
(javascript_statement
  (js_content) @injection.content
  (#set! injection.language "javascript")
  (#set! injection.combined))
```

Otherwise all js_content from all js blocks are combined into one, which I don't think is the original intention?

hankthetank27: Good point, I didn't consider this. This applies to the others as well so will adjust.

lucario387: Edit: I just uodated the suggested query @hank. On stable version of neovim injection.combined doesn't really worked, so it's better to comment the query and re-enable it once 0.10 becomes stable

hankthetank27: lol, ok sounds good.

lucario387 (dismissed):

queries/liquid/injections.scm
```diff
@@ -0,0 +1,22 @@
+((template_content) @injection.content
+  (#set! injection.language "html")
+  (#set! injection.combined))
+
+;TODO we can switch to quantifiers once neovim 0.10 becomes stable
```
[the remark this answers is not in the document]
lucario387: Nvm it has been like that since the first version of the query it seems :?

queries/liquid/highlights.scm
```diff
@@ -0,0 +1,111 @@
...
+  "endjavascript"
+  "endpaginate"
+  "endraw"
+  "endschema"
+  "endstyle"
+  "endtablerow"
+  "endunless"
+  "for"
+  "form"
+  "if"
```
lucario387, line 54: This section should be separated into keyword.repeat, import, ...

queries/liquid/highlights.scm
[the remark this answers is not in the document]
hankthetank27: I was able to fix this by making some small adjustments in the parser

queries/liquid/highlights.scm
```diff
@@ -0,0 +1,123 @@
...
+  "{{"
+  "}}"
+  "{{-"
+  "-}}"
+  "{%"
+  "%}"
+  "{%-"
+  "-%}"
+] @tag.delimiter
+  (#set! priority 110))
```
lucario387, line 11: Is this still intended for all highlights to be 110 priority? 🤔

lucario387: Also, it doesn't seem like these ranges are ever injected, so it should mean priorities here are unnecessary

hankthetank27: Because of this language is embedded and injections are done with `injection.combined` this is (afaik) the only way I can see to achieve the correct highlighting, as I eluded to @clason above.

To clarify,  highlight groups the sometimes conflict with injected "parent" language. For example, JavaScript highlights will be prioritized over the Liquid highlights when Liquid is embedded in a `<script>` tag. 

What you pointed out makes a perfect example of what I'm saying here.
With priority 110 (how its intended to be highlighted):
Screenshot 2024-03-27 at 10 36 26 PM

Without any priority:
Screenshot 2024-03-27 at 10 52 12 PM

lucario387: but this should mean the js tree is broken or not 🤔

queries/liquid/highlights.scm
```diff
@@ -0,0 +1,123 @@
...
+  ")"
+  "("
+] @punctuation.bracket
+  (#set! priority 110))
+
+([
+  ","
+  "."
+] @punctuation.delimiter
+  (#set! priority 110))
```
lucario387, line 25: Im pretty sure these one have the same highlights as js and html, so probably unnecessary to add priorities to these

queries/liquid/highlights.scm
hankthetank27: Yes, unfortunately this does mean the js tree (or any injected tree for that matter) is broken during these portions. Frankly, I don't really see a way around this using `injection.combined` without writing the parser to explicitly include the language constructs for html/js/css without injection (open to suggestions ofc -- I am new to ts and may be missing something..).

That being said, I haven't encountered any issues with the injected languages rendered incorrectly in terms of highlights, and since `injection.combined` is implemented upstream I honestly assumed this tree breaking behavior was to be expect when using injected languages this way.

hankthetank27: I did a bit more digging into this issue. I think that because the html parser injects js itself, js trys to parse the entire range within the script tag, rather than the combined `template_content` nodes..

you can see here that the html content is correctly captured and parsed..
Screenshot 2024-03-28 at 10 03 03 AM

but here, the js/script content is correctly captured, but not parsed as such -- all the content in the script tag is attempted to be parsed rather than just the combined `template_content` nodes like in its html counterpart.
Screenshot 2024-03-28 at 10 02 32 AM
Screenshot 2024-03-28 at 10 27 47 AM

It's a bit messy. Still, the only way I see around this is to more explicitly handle how the html is parsed in the template, and avoid leaning on just injecting the html parser over all template content.

hankthetank27: Yeah definitely.  It certainly does feel a bit like a "putting a square peg in a round hole" type of situation lol.

Anyways, I'd be willing to try some alternatives out I suppose, but I do feel using injections in this way is the most logical and (maybe?) idiomatic approach given the tools. Other template languages here do the same and suffer from the same issue, it just may not have been quite as apparent or investigated.

For example, the Vento language (added recently to this repo) also cannot correctly parse when dropping into a script tag either: 
Screenshot 2024-03-28 at 1 45 26 PM

Same with htmldjango (though experimental).

I guess in concussion to me it makes sense to make a concession with the higher priorities and at least have it appear to be correct.  But, I do understand if breaking the tree of an injected language is not acceptable.  Like you said I do think it's ultimately a just hard problem bc upstream implementation.

clason: Yes, this applies to all template languages, and my primary concern is to have a _consistent_ sensible strategy (given the constraints, which we can't do anything about).

On the other hand, I do not care so much about correctness; our goal here is syntax highlighting, not a syntax tree for code validation or transformation.

So priorities are not a deal-breaker (especially if this _is_ the consistent strategy), but should be used surgically, so only if absolutely necessary (and the check is per pattern).

hankthetank27: I removed what I see to be the unnecessary highlights after doing some testing, and left the remaining queries at 110 priority.

lucario387 (approved):
Looks good

PR #6360, opened 2024-03-25
feat(markdown): yaml, toml metadata highlights

ribru17:
Highlights metadata blocks such as the following:

```md
---
title: A title
author: An author
---
```

The highlight is given a lower priority because everything except the delimiting dashes/pluses will be injected with the appropriate metadata language.

### DIFF
--- a/queries/markdown/highlights.scm
+++ b/queries/markdown/highlights.scm
@@ -114,7 +114,7 @@
 ([
   (plus_metadata)
   (minus_metadata)
-] @string.special
+] @keyword.directive
   (#set! "priority" 90))
 
 [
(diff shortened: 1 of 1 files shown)


queries/markdown/highlights.scm
```diff
@@ -111,6 +111,12 @@
 ((block_quote) @markup.quote
   (#set! "priority" 90))
 
+([
+  (plus_metadata)
+  (minus_metadata)
+] @string.special
```
ObserverOfTime, line 117:
```suggestion
] @keyword.directive
```
Might be better.

ObserverOfTime (approved):

PR #6364, opened 2024-03-26
fix: qmljs hightlight

Decodetalkers:
Since ecma is still not fixed, should we add the part of highlight ourself?

### DIFF
--- a/queries/qmljs/highlights.scm
+++ b/queries/qmljs/highlights.scm
@@ -1,5 +1,12 @@
 ; temporarily removed the ecma inherit due to it breaking
-"pragma" @keyword.import
+[
+  "pragma"
+  "import"
+] @keyword.import
+
+(function_declaration
+  "function" @keyword.function
+  name: (identifier) @property)
 
 ; Annotations
 (ui_annotation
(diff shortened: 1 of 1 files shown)


queries/qmljs/highlights.scm
```diff
@@ -1,6 +1,13 @@
 ; temporarily removed the ecma inherit due to it breaking
 "pragma" @keyword.import
 
+(ui_import
+  "import" @keyword.import)
```
lucario387, line 5: Can't this just be `"import" @keyword.import`

Decodetalkers: Emm. I just worry that maybe it will find other place..

queries/qmljs/highlights.scm
```diff
@@ -1,6 +1,12 @@
 ; temporarily removed the ecma inherit due to it breaking
 "pragma" @keyword.import
 
+"import" @keyword.import
```
lucario387, line 4: Grouping this, possibly

Decodetalkers: done

ObserverOfTime (commented):
If whatever issue blocks inheritance cannot be fixed then *all* relevant parent highlights should be added.

Issue #5756, opened 2023-12-03
Windows unable to uninstall parsers


EvilGiraffes:
### Describe the bug

When trying to uninstall a parser, an error is thrown, looking at the path its likely due to the path using both `/` and `\` and on windows paths are `\`
![image](https://github.com/nvim-treesitter/nvim-treesitter/assets/85831955/037cbcc8-9980-46a3-9d02-730f16dab544)

### To Reproduce

1. Install any parser on a windows machine
2. Uninstall it again

### Expected behavior

Uninstalling the parser, without having to do it manually

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
- OK `node` found v18.18.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `gcc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: gcc.exe (x86_64-posix-seh-rev2, Built by MinGW-W64 project) 12.2.0
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "10.0.19045",
  sysname = "Windows_NT",
  version = "Windows 10 Home"
} ~

Parser/Features         H L F I J
  - bash                ✓ ✓ ✓ . ✓
  - c                   ✓ ✓ ✓ ✓ ✓
  - c_sharp             ✓ ✓ ✓ . ✓
  - gdscript            ✓ ✓ ✓ ✓ ✓
  - html                ✓ ✓ ✓ ✓ ✓
  - javascript          ✓ ✓ ✓ ✓ ✓
  - jsdoc               ✓ . . . .
  - json                ✓ ✓ ✓ ✓ .
  - lua                 ✓ ✓ ✓ ✓ ✓
  - luadoc              ✓ . . . .
  - luap                ✓ . . . .
  - markdown            ✓ . ✓ ✓ ✓
  - markdown_inline     ✓ . . . ✓
  - python              ✓ ✓ ✓ ✓ ✓
  - query               ✓ ✓ ✓ ✓ ✓
  - regex               ✓ . . . .
  - rust                ✓ ✓ ✓ ✓ ✓
[...]

PR #6369, opened 2024-03-27
fix(uninstall): use correct path separator on windows

qgates:
Addressed bug/issue #5756

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -632,6 +632,7 @@ function M.uninstall(...)
         )
       end
 
+      install_dir = install_dir:gsub("/", utils.get_path_sep()) -- fix path separator for windows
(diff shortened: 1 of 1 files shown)


clason:
I think this should use the `clean_path` function below (and same for other paths that need to be normalized on Windows -- can you check for those and give them the same treatment if necessary?)

qgates:
Ok thanks, will take a look. Fine with lua but new to working on nvim plugins (go easy on me pls!) question:

Installed to nvim via Lazy, but git head detached by default. What's the best workflow for fixing/contrib on this (and other) plugin repos... manually clone into default lazy install dir? or is there a Lazy config that allows me to work with cloned repos instead of default plugin install path?

Apologies if this is explained somewhere, didn't find it.

🙂

qgates:
@clason commit 79cf503 updates top level funcs install / uninstall to use clean_path as suggested

clason (approved):

PR #6375, opened 2024-03-29
feat(djot): update queries

treeman:
I went through the query files for Djot and improved and updated them:

- Fill in some missing highlights
- Adjust some highlights to fit context better
- Conceal inline content
- Add locals
- Add folds
- Add indents

@NoahTheDuke

### DIFF
--- a/queries/djot/highlights.scm
+++ b/queries/djot/highlights.scm
@@ -21,7 +21,7 @@
   (code_block)
   (raw_block)
   (frontmatter)
-] @markup.raw
+] @markup.raw.block
   (#set! "priority" 90))
 
 ; Remove @markup.raw for code with a language spec
@@ -29,7 +29,8 @@
   .
   (code_block_marker_begin)
   (language)
-  (code) @none)
+  (code) @none
+  (#set! "priority" 90))
 
 [
   (code_block_marker_begin)
@@ -38,7 +39,11 @@
   (raw_block_marker_end)
 ] @punctuation.delimiter
 
-(language) @label
+(language) @attribute
+
+(inline_attribute
+  _ @conceal
+  (#set! conceal ""))
 
 ((language_marker) @punctuation.delimiter
   (#set! conceal ""))
@@ -60,7 +65,7 @@
 (table_caption
   (marker) @punctuation.special)
 
-(table_caption) @markup.heading
+(table_caption) @markup.italic
 
 [
   (list_marker_dash)
@@ -85,10 +90,14 @@
 ] @markup.list
 
 (list_marker_task
-  (unchecked) @constant.builtin) @markup.list.unchecked
+  (unchecked)) @markup.list.unchecked
 
 (list_marker_task
-  (checked) @constant.builtin) @markup.list.checked
+  (checked)) @markup.list.checked
+
+; Colorize `x` in `[x]`
+((checked) @constant.builtin
+  (#offset! @constant.builtin 0 1 0 -1))
 
 [
   (ellipsis)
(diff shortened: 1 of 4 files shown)


clason:
Please read `CONTRIBUTING.md` and fix the failing tests.

treeman:
Oh, yeah. These changes depends on a newer version of the Djot parser. Bumped the revision in the lockfile.

queries/djot/highlights.scm
```diff
@@ -85,106 +90,163 @@
 ] @markup.list
 
 (list_marker_task
-  (unchecked) @constant.builtin) @markup.list.unchecked
+  (unchecked)) @markup.list.unchecked
```
lucario387, line 93: Are you marking the unchecked, or the marker, or is the marker has some other child nodes other than unchecked/checked?

treeman: `unchecked` here is just marking a single space. So `- [ ]` is the `list_marker_task` and the space is the `unchecked` element. It didn't make much sense to mark the space, but I'm not sure what the best practice here is.

clason: I think the convention is to capture the whole `[  ]` and `[X]` (in case people want to conceal it).

treeman:
I got this failing test that I don't think is related to my changes:

```
Fail	||	injections tests/query/injections/svelte/test-svelte-injections.svelte	
            Vim:E576: Error while reading ShaDa file: last entry specified that it occupies 92 bytes, but file ended earlier
            
            stack traceback:
            	...reesitter/nvim-treesitter/tests/query/injection_spec.lua:9: in function 'check_assertions'
            	...reesitter/nvim-treesitter/tests/query/injection_spec.lua:83: in function <...reesitter/nvim-treesitter/tests/query/injection_spec.lua:82>
```

queries/djot/highlights.scm
treeman: Oh okay, I guess this would need an update to the parser then.

clason: I think the idea is to replace `[x]`, not just the `x` (check markdown).

lucario387: yes.

lucario387: (mostly just the queries, not the parser). Haskell parser also have unicode characters as nodes, but using them in queries break it.

clason: That's different, and will be fixed by https://github.com/tree-sitter/tree-sitter/pull/3223.

treeman: I've removed all the special characters from conceal.

queries/djot/highlights.scm
```diff
@@ -85,106 +90,168 @@
...
+
+((quotation_marks) @string.special
+  (#eq? @string.special "'}")
+  (#offset! @string.special 0 1 0 0)
+  (#set! conceal ""))
+
+((quotation_marks) @string.special
+  (#any-of? @string.special "\\\"" "\\'")
+  (#offset! @string.special 0 0 0 -1)
+  (#set! conceal ""))
```
lucario387, line 137: These can be combined into 2 patterns

```suggestion
; Conceal { and } but leave " and '
((quotation_marks) @string.special
  (#any-of? @string.special "{\"" "{'" "\\\"" "\\'")
  (#offset! @string.special 0 0 0 -1)
  (#set! conceal ""))

((quotation_marks) @string.special
  (#any-of? @string.special "\"}" "'}")
  (#offset! @string.special 0 1 0 0)
  (#set! conceal ""))
```

Also, are these like markdown backticks, or string quotes? If the latter, then I don't think we should hide it

queries/djot/highlights.scm
```diff
@@ -85,106 +90,168 @@
...
+
+(hard_line_break) @string.escape
+
+(backslash_escape) @string.escape
```
lucario387, line 141:
```suggestion
[
  (hard_line_break)
  (backslash_escape)
] @string.escape

```

queries/djot/highlights.scm
```diff
@@ -85,106 +90,168 @@
...
 
-(strong
-  (strong_begin) @punctuation.delimiter)
+(highlighted) @string.special
 
-(strong
-  (strong_end) @punctuation.delimiter)
+(superscript) @string.special
 
+(subscript) @string.special
```
lucario387, line 164: Group these

queries/djot/highlights.scm
```diff
@@ -85,106 +90,168 @@
...
+  ] @punctuation.delimiter
+  (#set! conceal ""))
 
 (delete
   [
     "{-"
     "-}"
-  ] @punctuation.delimiter)
+  ] @punctuation.delimiter
+  (#set! conceal ""))
```
lucario387, line 187: These look like they can be grouped into one

```suggestion
([
    "{="
    "=}"
    "{+"
    "+}"
    "{-"
    "-}"
] @punctuation.delimiter
  (#set! conceal ""))
```

lucario387: So are the ones below it seems

queries/djot/highlights.scm
treeman: These are string to control start/end quotes like `“` and `”` (`{'` forces the "left" variant).

queries/djot/highlights.scm
treeman: I don't think so, it's important to target child nodes specifically (e.g. `(delete "{-")` but not `{-` globally because there exists an anonymous node that captures `{-` that would den get wrongly highlighted and concealed.

lucario387: As you are the writer of the language, is there any other `{-` node?  From what I can see it's only here and symbol_fallback, which doesn't matter

treeman: Yes I am, and it's symbol_fallback I'm referring to.

I also didn't think it shouldn't matter, but when I use the global version it does get highlighted and concealed and I can only assume that symbol_fallback is the culprit.

queries/djot/highlights.scm
clason: Yeah, don't conceal quotes. (Markdown is special and gets a pass.) If in doubt, leave it for users to add.

clason: Ah, djot is also a markdown-style language; that changes things. Then you can use `conceal` to "render" it as much as you want.

lucario387: Can you group some of these queries @treeman. I don't think it's necessary to separate these captures into 6 queries, 2 queries + `any-of` should be enough

treeman: Yes sorry about that, I thought i accepted the suggestion but I must have screwed up somehow.

Fixed not.

treeman:
Awesome, thank you all.

PR #6379, opened 2024-03-30
feat(snakemake): add indentation queries

osthomas:
This PR adds indentation queries to snakemake.


lucario387 (approved):
Looks good. Just skimmed through the corpus in the repo. It looks like these are similar to python def and class def enough

PR #6381, opened 2024-03-30
feat(roc): add new parser tree-sitter-roc

nat-418:
Parser for the [Roc] programming language.

[Roc]: https://www.roc-lang.org/

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,6 +370,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [pip requirements](https://github.com/ObserverOfTime/tree-sitter-requirements) (maintained by @ObserverOfTime)
 - [x] [rnoweb](https://github.com/bamonroe/tree-sitter-rnoweb) (maintained by @bamonroe)
 - [x] [robot](https://github.com/Hubro/tree-sitter-robot) (maintained by @Hubro)
+- [x] [roc](https://github.com/nat-418/tree-sitter-roc) (maintained by @nat-418)
 - [x] [ron](https://github.com/amaanq/tree-sitter-ron) (maintained by @amaanq)
 - [x] [rst](https://github.com/stsewd/tree-sitter-rst) (maintained by @stsewd)
 - [x] [ruby](https://github.com/tree-sitter/tree-sitter-ruby) (maintained by @TravonteD)
--- a/queries/roc/highlights.scm
+++ b/queries/roc/highlights.scm
@@ -1,0 +1,189 @@
+;---Most generic types---
+(module) @module
+
+(identifier) @variable
+
+(concrete_type) @type
+
+;---annotations----
+(annotation_type_def
+  (annotation_pre_colon
+    (identifier) @type))
+
+(annotation_type_def
+  (annotation_pre_colon
+    (identifier) @function)
+  (function_type))
+
+;----decleration types----
+(value_declaration
+  (decl_left
+    (identifier_pattern
+      (identifier) @variable.parameter)))
+
+;---records----
+(field_name) @variable.member
+
+(record_field_pattern
+  (_
+    (identifier) @variable))
+
+;matches the second identifier and all subsequent ones
+(field_access_expr
+  (identifier) @variable.member)
+
+;----comments----
+(line_comment) @comment @spell
+
+(doc_comment) @comment.documentation @spell
+
+;-----Punctuation----
+[
+  "?"
+  (arrow)
+  (back_arrow)
+  (backslash)
+  ","
+  ":"
+] @punctuation.delimiter
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+  "["
+  "]"
+] @punctuation.bracket
+
+[
+  "|"
+  "&"
+  (operator)
+] @operator
+
+(wildcard_pattern) @constant.builtin
+
+[
+  "if"
+  "then"
+  "else"
+] @keyword.conditional
+
+[
+  (implements)
+  (when)
+  (is)
+  "as"
+  (to)
+] @keyword
+
+;----headers-----
+(interface_header
+  (name) @type.definition)
+
+(imports
+  (imports_entry
+    (module) @module))
+
+(packages
+  (record_pattern
+    (record_field_pattern
+      (field_name) @module)))
+
+(app_name) @string.special
+
+(import_path) @string.special.path
+
+[
+  "app"
+  "packages"
+  "provides"
+  "interface"
+  "exposes"
+  "expect"
+] @keyword
+
+[
+  (import_as)
+  "imports"
+] @keyword.import
+
+(backpassing_expr
+  assignee: (identifier_pattern
+    (identifier) @variable.parameter))
+
+(value_declaration
+  (decl_left
+    (identifier_pattern
+      (identifier) @function))
+  (expr_body
+    (anon_fun_expr)))
+
+;----tags----
+(tags_type
+  (apply_type
+    (concrete_type) @constructor))
+
+[
+  (tag)
+  (opaque_tag)
+] @constructor
+
+;-----builtins----
+(variable_expr
+  (module) @module
+  (identifier) @boolean
+  (#any-of? @boolean "true" "false")
+  (#eq? @module "Bool"))
+
+"dbg" @keyword.debug
+
+;----function invocations ----
+(function_call_expr
+  caller: (variable_expr
+    (identifier) @function.call))
+
+(function_call_expr
+  caller: (field_access_expr
+    (identifier) @function.call .))
+
+(bin_op_expr
+  (operator
+    "|>") @operator
+  (variable_expr
+    (identifier) @function))
+
+;----function arguments----
+(argument_patterns
+  (identifier_pattern
+    (identifier) @variable.parameter))
+
+(argument_patterns
+  (_
+    (identifier_pattern
+      (identifier) @variable.parameter)))
+
+(argument_patterns
+  (_
+    (_
+      (identifier_pattern
+        (identifier) @variable.parameter))))
+
+;-----consts-----
+[
+  (int)
+  (uint)
+  (iint)
+  (xint)
+  (natural)
+] @number
+
+[
+  (decimal)
+  (float)
+] @number.float
+
+(string) @string
+
+(char) @character
--- a/queries/roc/locals.scm
+++ b/queries/roc/locals.scm
@@ -1,0 +1,40 @@
+(expr_body) @local.scope
+
+(argument_patterns
+  (identifier_pattern
+    (identifier) @local.definition))
+
+; (argument_patterns(long_identifier)@local.definition)
+(exposes_list
+  (ident) @local.reference)
+
+(opaque_type_def
+  (apply_type
+    (concrete_type) @local.definition.type))
+
+(alias_type_def
+  (apply_type
+    (concrete_type) @local.definition.type))
+
+(value_declaration
+  (decl_left
+    (identifier_pattern
+      (identifier) @local.definition.function))
+  (expr_body
+    (anon_fun_expr)))
+
+(value_declaration
+  (decl_left
+    (identifier_pattern
+      (identifier) @local.definition.var)))
+
+(identifier_pattern
+  (identifier) @local.definition)
+
+(exposes
+  (ident) @local.reference)
+
+(identifier) @local.reference
+
+(tag_expr
+  (tag)) @local.reference
--- a/lockfile.json
+++ b/lockfile.json
@@ -590,6 +590,9 @@
   "robot": {
     "revision": "322e4cc65754d2b3fdef4f2f8a71e0762e3d13af"
   },
+  "roc": {
+    "revision": "14c66eeceb7db2d5c98616036d861863b273bab8"
+  },
   "ron": {
     "revision": "ce6086b2c9e8e71065b8129d6c2289c5f66d1879"
   },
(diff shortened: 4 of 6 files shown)


clason:
1. tags are not used in Neovim, and textobjects live in another repo.
2. please read CONTRIBUTING.md and adjust the queries accordingly; we have much tighter rules for capture names (and our indent module works very differently).
3. consider adding folds as well.

clason:
also run `./scripts/format-queries.lua queries/roc`

README.md
```diff
@@ -371,6 +371,7 @@ We are looking for maintainers to add more parsers and to write query files for
 - [x] [rnoweb](https://github.com/bamonroe/tree-sitter-rnoweb) (maintained by @bamonroe)
 - [x] [robot](https://github.com/Hubro/tree-sitter-robot) (maintained by @Hubro)
 - [x] [ron](https://github.com/amaanq/tree-sitter-ron) (maintained by @amaanq)
+- [x] [roc](https://github.com/nat-418/tree-sitter-roc) (maintained by @nat-418)
```
lucario387, line 374: Position in the list seems wrong

clason: bot will take care of it, but might as well do it right.

clason: (Also lockfile)

queries/roc/highlights.scm
```diff
@@ -0,0 +1,228 @@
...
+;----decleration types----
+(value_declaration
+  (decl_left
+    (identifier_pattern
+      (identifier) @parameter.definition))
+  (#set! "priority" 99))
+
+(backpassing_expr
+  assignee: (identifier_pattern
+    (identifier) @parameter.definition))
```
lucario387, line 98: Captures doesn't seem nvim-treesitter like
Please follow CONTRIBUTING.md guideline on allowed captures

queries/roc/highlights.scm
```diff
@@ -0,0 +1,228 @@
...
+            (identifier) @parameter))))))
+
+(argument_patterns
+  (_
+    (_
+      (_
+        (_
+          (_
+            (identifier_pattern
+              (identifier) @parameter)))))))
```
lucario387, line 184: Long story short: no
Why 5 depths, why not 6,7,8,9 depths is an example question. At worst please do queries like these that go as much as 2 depth, at best just the direct child

queries/roc/locals.scm
```diff
@@ -0,0 +1,40 @@
+(expr_body) @scope
```
lucario387, line 1: Please follow upstream treesitter locals naming, and CONTRIBUTING.md guidelines

queries/roc/highlights.scm
clason: Or use `#has-ancestor?`. (Expensive, but will be improved upstream, so it's better than copy-pasting variations on a theme.)

queries/roc/highlights.scm
nat-418: I have read that document and I do not understand what I need to change here.

clason: `@parameter.definition` is not an allowed capture name. You can **only** use the captures listed in CONTRIBUTING.md, nothing else.

queries/roc/locals.scm
nat-418: I have changed the naming, please let me know if this is correct.

queries/roc/highlights.scm
```diff
@@ -0,0 +1,201 @@
...
+
+;---Most generic types---
+(module
+  (#set! "priority" 99)) @module
+
+(identifier
+  (#set! "priority" 98)) @variable
+
+(concrete_type
+  (#set! "priority" 99)) @type
```
clason, line 201: Rely on order rather than explicit priority wherever possible. Try moving them to the top of the file instead.

queries/roc/highlights.scm
```diff
@@ -0,0 +1,201 @@
...
+  (name) @type.definition)
+
+(imports
+  (imports_entry
+    (module) @module))
+
+(packages
+  (record_pattern
+    (record_field_pattern
+      (field_name) @@module)))
```
clason, line 61:
```suggestion
      (field_name) @module)))
```
typo.

queries/roc/highlights.scm
```diff
@@ -0,0 +1,194 @@
...
+      (field_name) @module)))
+
+(app_name) @string
+
+(import_path) @string
+
+[
+  "app"
+  "packages"
+  "imports"
```
ribru17, line 104: Maybe merge this list with the previous list that also targets `@keyword`. Also should `"imports"` be instead highlighted as `@keyword.import`?

queries/roc/highlights.scm
```diff
@@ -0,0 +1,194 @@
...
+  (uint)
+  (iint)
+  (xint)
+  (natural)
+] @constant
+
+[
+  (decimal)
+  (float)
+] @constant
```
ribru17, line 190: `@number.float`

queries/roc/highlights.scm
```diff
@@ -0,0 +1,194 @@
...
+        (identifier) @variable.parameter))))
+
+;-----consts-----
+[
+  (int)
+  (uint)
+  (iint)
+  (xint)
+  (natural)
+] @constant
```
ribru17, line 185: `@number`

queries/roc/highlights.scm
nat-418: Here is an example of the use of `imports` in  a simple Roc program:

```
app "hello"
    packages { pf: "https://github.com/roc-lang/basic-cli/releases/download/0.8.1/x8URkvfyi9I0QhmVG98roKBUs_AZRkLFwFJVJ3942YA.tar.br" }
    imports [pf.Stdout]
    provides [main] to pf

main =
    Stdout.line "I'm a Roc application!"
 ```
 
 My understanding is that, as part of the module header, it should be `@module` rather than `@keyword.imports` which I would associate for example with the Go style of importing. If I am wrong, please let me know.

File: queries/roc/highlights.scm [hunk 10]
```diff
@@ -0,0 +1,194 @@
...
+
+[
+  (decimal)
+  (float)
+] @constant
+
+(string) @string
+
+(char) @constant
```
ribru17, line 194: Maybe `@character` (or `@character.special`, don't know this language well) for this

queries/roc/highlights.scm
```diff
@@ -0,0 +1,194 @@
...
+    (module) @module))
+
+(packages
+  (record_pattern
+    (record_field_pattern
+      (field_name) @module)))
+
+(app_name) @string
+
+(import_path) @string
```
ribru17, line 99: Would `@string.special.path` make more sense here?

queries/roc/highlights.scm
clason: the `imports` is a keyword, so should be `@keyword.import`. The `[pf.Stdout]` would be the `@module` (assuming I understand correctly).

queries/roc/highlights.scm
```diff
@@ -0,0 +1,194 @@
...
+    (anon_fun_expr)))
+
+;----tags----
+(tags_type
+  (apply_type
+    (concrete_type) @constructor))
+
+(tag) @constructor
+
+(opaque_tag) @constructor
```
ribru17, line 130: Nit: merge these two captures into a list pls :)

queries/roc/highlights.scm
```diff
@@ -0,0 +1,194 @@
...
+  "?"
+  (arrow)
+  (back_arrow)
+  (backslash)
+] @punctuation.delimiter
+
+[
+  ","
+  ":"
+] @punctuation.delimiter
```
ribru17, line 52: Also maybe merge this list with the one above

queries/roc/highlights.scm
ribru17: While you're at it, is it easy / would it make sense to highlight the tarball string as `@string.special.url`?

nat-418: > the `imports` is a keyword, so should be `@keyword.import`. The `[pf.Stdout]` would be the `@module` (assuming I understand correctly).

In Roc, the whole top block in my example  is known as a module header. Every `.roc`  file is a module, and there are different types of modules. We know this particular one is an application module because it begins with the `app` keyword.

clason: So should the whole block be highlighted in a single color?

nat-418: > While you're at it, is it easy / would it make sense to highlight the tarball string as `@string.special.url`?

In Roc, packages may be loaded either via URL or from a local filesystem path. So I am not sure it is appropriate to mark that string as a URL.

nat-418: > So should the whole block be highlighted in a single color?

I think different colors could be fine. It is comparable to the usual beginning in Go:

```go
package main

import (
	"fmt"
	"os"
)

// ...
```

In my Neovim setup, `package` and `import` are both colored red; `main` , `(`, and `)` are white; `"fmt"` and `"os"` are green. Nothing per se marks them as "top matter". In Roc, the same could work—or, the whole block could be visually distinct in some way. I am open to whatever seems like it fits in with the norms of the `nvim-treesitter` project.

clason: Then `import` is `@keyword.import`, period.

clason: (`.import`, not `.import**s**` -- that's exactly why we have that CI job ;))

queries/roc/indents.scm
ObserverOfTime, line 1: If it's broken it should not be included.

queries/roc/highlights.scm
```diff
@@ -0,0 +1,193 @@
...
+
+;matches the second identifier and all subsequent ones
+(field_access_expr
+  (identifier) @variable.member)
+
+;----comments----
+[
+  (line_comment)
+  (doc_comment)
+] @comment
```
ObserverOfTime, line 39:
```suggestion
(line_comment) @comment @spell

(doc_comment) @comment.documentation @spell
```

queries/roc/highlights.scm
```diff
@@ -0,0 +1,193 @@
...
+  "}"
+  "["
+  "]"
+] @punctuation.bracket
+
+[
+  "|"
+  "&"
+  (operator)
+  (wildcard_pattern)
```
ObserverOfTime, line 64: `(wildcard_pattern)` may be `@variable.parameter` or `@constant.builtin`.

queries/roc/highlights.scm
```diff
@@ -0,0 +1,193 @@
...
+  (implements)
+  (when)
+  (is)
+  "as"
+  (to)
+] @keyword
+
+;----headers-----
+(interface_header
+  (name) @type.definition)
```
ObserverOfTime, line 83: Is this akin to a typedef?

queries/roc/highlights.scm
```diff
@@ -0,0 +1,193 @@
...
+(import_path) @string.special.path
+
+[
+  "app"
+  "packages"
+  "provides"
+  "interface"
+  "exposes"
+  "expect"
+  (import_as)
```
ObserverOfTime, line 105: `(import_as)` should be `@keyword.import`.

queries/roc/highlights.scm
```diff
@@ -0,0 +1,193 @@
...
+  (module) @module
+  (identifier) @constant.builtin
+  (#eq? @constant.builtin "true")
+  (#eq? @module "Bool"))
+
+(variable_expr
+  (module) @module
+  (identifier) @constant.builtin
+  (#eq? @constant.builtin "false")
+  (#eq? @module "Bool"))
```
ObserverOfTime, line 142:
```suggestion
(variable_expr
  (module) @module
  (identifier) @boolean
  (#any-of? @boolean "true" "false")
  (#eq? @module "Bool"))
```

queries/roc/highlights.scm
```diff
@@ -0,0 +1,193 @@
...
+
+"dbg" @constant.builtin
```
ObserverOfTime, line 144:
```suggestion
"dbg" @keyword.debug
```

ObserverOfTime (changes requested):
Missing `injections.scm`:

```clojure
([
  (line_comment)
  (doc_comment)
] @injection.content
```

queries/roc/highlights.scm
nat-418: Here is an example:

```
interface Nullable
    exposes [Nullable, null, full, nullableDecode, nullableEncode]
    imports [
        json.Core.{ json },
    ]

Nullable val := [Null, Full val] implements [
        Decoding { decoder: nullableDecode },
        Encoding { toEncoder: nullableEncode },
        Inspect, # auto derive
        Eq, # auto derive
    ]

null = @Nullable Null
full = \a -> @Nullable (Full a)
```

queries/roc/highlights.scm
```diff
@@ -0,0 +1,189 @@
...
+(imports
+  (imports_entry
+    (module) @module))
+
+(packages
+  (record_pattern
+    (record_field_pattern
+      (field_name) @module)))
+
+(app_name) @string
```
ObserverOfTime, line 94:
```suggestion
(app_name) @string.special
```

ObserverOfTime (approved):

clason:
Thank you for your patience!

clason:
@nat-418 I notice that your fork is behind upstream. Can we switch to that? (Especially since the symlinked nix files in your fork break installing on Windows, where symlinks are not supported out of the box.)

PR #6382, opened 2024-03-30
feat: support quantifiers for folding

ayroblu:
Solves #3196

With https://github.com/neovim/neovim/pull/24738 merged in nightly neovim, quantifiers can be all matched and folded as one

Tested with:

```lua
local folds_query = [[
  ((import_statement)+ @fold)
]]

require("vim.treesitter.query").set("typescript", "folds", folds_query)
```


clason:
Folding is now implemented in Neovim itself; this module is just left here (for now) for backward compatibility.

ayroblu:
Dumb question, how do I use the built in version?

clason:
`:h vim.treesitter.foldexpr()`

PR #6384, opened 2024-04-01
Misc nix fixes

ribru17:
Adds some punctuation highlights that seem to be missing, and also changes the ellipsis parameter to parameter.builtin to match other languages that do this to their ellipsis params, like lua e.g.


lucario387:
Wait why aren't you a member yet 😆

ribru17:
Lol  :laughing: I would love to be a member, not sure what the process is to join though

PR #6393, opened 2024-04-03
fix(angular): add highlights for static member expression

dlvandenberg:


queries/angular/highlights.scm
```diff
@@ -4,7 +4,10 @@
 
 (pipe_operator) @operator
 
-(string) @string
+[
+ (string)
+ (static_member_expression)
```
lucario387, line 9: Formatting seems wrong

lucario387 (approved):

PR #6395, opened 2024-04-03
add earthfile language

glehmann:


glehmann:
related to https://github.com/neovim/neovim/pull/28166, to be fully usable

queries/earthfile/highlights.scm
```diff
@@ -0,0 +1,193 @@
...
+(wait_command
+  [
+    "WAIT"
+    "END"
+  ] @keyword)
+
+[
+  (comment)
+  (line_continuation_comment)
+] @comment
```
ObserverOfTime, line 135:
```suggestion
] @comment @spell
```

queries/earthfile/highlights.scm
```diff
@@ -0,0 +1,193 @@
...
+(expansion
+  [
+    "$"
+    "{"
+    "}"
+    "("
+    ")"
+  ] @punctuation.special)
+
+(build_arg) @variable
```
ObserverOfTime, line 168: `@variable.parameter` ?

queries/earthfile/highlights.scm
ObserverOfTime, line 1: Nodes that are never highlighted differently don't need to be wrapped. e.g.

```scheme
[
  "ARG"
  "BUILD"
  ...
] @keyword

"=" @operator
```

queries/earthfile/highlights.scm
glehmann: that’s to be able to visually differentiate the build args and the options, so, if you don’t mind, I think it’s better to keep it as `@variable`

queries/earthfile/highlights.scm
glehmann: mm, I wrapped them because `TRY` for example was highlighted in the middle of other words.
See https://github.com/glehmann/tree-sitter-earthfile/issues/6 for a screenshot

queries/earthfile/highlights.scm
ObserverOfTime: Maybe you can use `@property` for options.

queries/earthfile/highlights.scm
clason: Then your parser is not doing its job properly ;) It (not the queries!) should only expose `TRY` in the proper context. Otherwise you're back at basic bitch regex parsing.

ObserverOfTime: That sounds like a parser issue.

glehmann: the parser must have done something wrong at that time. I’ve rechecked and it works without problem now without wrapping the keyword.

I’m fixing the queries

queries/earthfile/highlights.scm
```diff
@@ -0,0 +1,126 @@
...
+  [
+    "$"
+    "{"
+    "}"
+    "("
+    ")"
+  ] @punctuation.special)
+
+(build_arg
+  (variable) @variable.parameter) @variable.parameter
```
ObserverOfTime, line 119:
```suggestion
  (variable) @variable.parameter)
```

glehmann: in `--foo=bar` that would make `--` and `foo` be highlighted differently:

Capture d’écran 2024-04-04 à 17 24 19

I clearly prefer to have it highlighted uniformly, but I’ll change it if that’s a common practice — just checking that we are well aligned :)

clason: Why not 
```
(build_arg) @variable.parameter
```
then?

ObserverOfTime: As long as it doesn't capture the value.

clason: (or `(buildarg (variable)) @variable.parameter`, if it makes a difference)

glehmann: neither `(build_arg) @variable.parameter` or `(buildarg (variable)) @variable.parameter` produce an uniformly highlighted `--foo`. That’s because the `build_arg` node contains a `variable` node which is already highlighted as `@variable`.

`(build_arg (variable) @variable.parameter) @variable.parameter` applies the `@variable.parameter` to both the `build_arg` node, with is visible as the `--` and the `variable` node which is visible as `foo`.

Capture d’écran 2024-04-04 à 22 23 09

The value is a `string` which itself is made of one or more `unquoted_string`, `single_quoted_string` or `double_quoted_string` which are themselves highlighted with other queries.

Is there something wrong with that query?

lucario387: So what im seeing is there is no dedicated anon node for - and --

glehmann: there is a anonymous `--` node. Is it better to highlight that node instead of the whole `build_arg` node?

something like:

```
(build_arg
  [
    "--"
    (variable)
  ] @variable.parameter)
```

lucario387: Yes

glehmann: ok, works for me 👍

glehmann: Is there a documentation with the good practices on writing tree-sitter queries that I may have missed?

ObserverOfTime (approved):

queries/earthfile/highlights.scm
clason: CONTRIBUTING.md is the best documentation we currently have.

lucario387 (approved):

glehmann:
Thanks!

PR #6402, opened 2024-04-04
fix(rust): indentation in multi-line tuple pattern

ganhaque:
previous:
```rust
    for (
    abc,
    def
) in thing.iter() {
        // ...
    }
```

new:
```rust
    for (
        abc,
        def
    ) in thing.iter() {
        // ...
    }
```


lucario387:
Please also add in a test for this case :)

lucario387 (approved):

PR #6405, opened 2024-04-05
fix(bash): fix printf injection queries

datwaft:
The `(string)` node has a `(string_content)` child which breaks the injection of `printf`. The solution was to use `(#set! injection.include-children)`.

I also added support for `(concatenation)` nodes.

### Before

image

### After

image

### DIFF
--- a/queries/bash/injections.scm
+++ b/queries/bash/injections.scm
@@ -14,13 +14,15 @@
   name: (command_name) @_command
   .
   argument: [
-    (string)
+    (string
+      (string_content) @injection.content)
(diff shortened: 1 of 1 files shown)


lucario387:
Can't we just target string_content directly 🤔

datwaft:
> Can't we just target string_content directly 🤔

Done!

image

image

datwaft:
Forgot to remove the include-children, once second

datwaft:
Done!

ObserverOfTime (approved):

PR #6407, opened 2024-04-05
Format awk code in bash files

Trulsaa:


ribru17:
This is good but I think it needs to be more restrictive: right now, in something like `awk -f './script.awk'`, the string './script.awk' will also be given awk injection

Trulsaa:
I have tried to restrict  the query to not include raw_string nodes preceded with a flag. But as far as I have been able to read, this is not possible. Any suggestions? As far as I can se gather, the Negated Fields ! does not work in nvim.

ribru17:
It may not be possible to do this perfectly. Even if we could do some negating pattern like that, there are some cases where it would be valid to have the string injected if it follows a regular flag, such as `-b`. Maybe we could just inject if the first argument is a raw string, and then anything after that will just be ignored. Obviously this is not exhaustive though. I am not sure if it is worth doing if we cannot do it fully properly, but maybe someone else will give some valid input on how to go forward.

Trulsaa:
I agree. This might not be worth doing. We could leave this here and see if someone have any suggestions. I can set up a reminder to come back and close it if nothing happens. Feel free to tag anyone if you know someone that could have some insight.

lucario387:
I don't think it's inclusion-worthy (yet) if we can't target only the awk pattern nodes. For now I will close the PR. Feel free to ping/re-open it once there is a better way to do it

ViniiR:
Hi there, i'm 2 years late but i think my solution could be worth implementing, i've made it so it does not match when a raw string is preceded by ```-f```(in any arrangement e.g. ```-bef```) or ```--file```

i'd be happy to try fixing any possible edge cases.

image

PR #6410, opened 2024-04-06
feat(luap): Highlight class captures similar to regex captures

ValdezFOmar:


queries/luap/highlights.scm
```diff
@@ -32,7 +32,10 @@
 (negated_set
   (character) @constant)
 
-(class) @keyword
+(class) @string.escape
+
+(class
+  (escape_char)) @string.luap
```
lucario387, line 38: Why not just 
```suggestion
(class
  (escape_char) @string.escape)
```

clason: In any case, `@string.luap` is not an allowed capture name.

ValdezFOmar: I use that capture because I was trying to do something similar to this:

https://github.com/nvim-treesitter/nvim-treesitter/blob/a2d6678bb21052013d0dd7cb35dffbac13846c98/queries/regex/highlights.scm#L20

clason: That `regexp` is not the language name, though (which is `regex`, in fact).

ValdezFOmar: Oh, honestly didn't know. Should I use that capture instead? Since the `(escape_char)` node is similar to `(identity_escape)`

clason: Yes, that's exactly what @lucario387 suggested.

ValdezFOmar: He suggested using `@string.escape`, not `@string.regexp`

lucario387: Well it's an escape character, I'd expect the result to be either literal string or escaped strings, not a regex/lua pattern

ValdezFOmar: `(escape_char)` is for a escaped special char (`.` -> `%.`) and since the equivalent in regex (`.` -> `\.`) is captured as `@string.regexp` I'd expect luap to do the same

ValdezFOmar: So I change it to `@string.escape` or leave it as `@string.regexp`?

queries/luap/highlights.scm
```diff
@@ -32,7 +32,11 @@
 (negated_set
   (character) @constant)
 
-(class) @keyword
+(class) @string.escape
+
+(class
+  "%" @string.regexp
+  (escape_char) @string.regexp)
```
lucario387, line 39: Question. Why only % and escape_char are regexp? It should be the reverse

clason: Concrete examples always go a long way.

ValdezFOmar: Consider the following regex `\w*\.\s\*.` with the equivalent lua pattern `%w*%.%s%*.`

regex | lua pattern | meaning
---   | ---         | ---
`\w`  | `%w`        | character class
`*`   | `*`         | zero or more times
`\.`  | `%.`        | literal dot
`\s`  | `%s`        | character class
`\*`  | `%*`        | literal asterisk
`.`   | `.`         | any character

The regex queries capture character classes as `@string.escape` and escaped special characters (`\.`, `\*`, etc) as `@string.regexp`, but the luap queries captures both as `@keyword`.

Using `(class) @string.escape` will capture character classes. And for escaped special characters: 

```scm
(class
  "%" @string.regexp
  (escape_char) @string.regexp)
```

clason: Makes sense to me. @amaanq ?

clason:
While waiting for (re-)review, can you rebase on latest master?

clason:
Having consistency is worth merging this. We can revisit how to handle this (across all regex-like languages) down the line.

PR #6415, opened 2024-04-08
docs: invalid structure at position -> do not lazy load nvim-treesitter

ecerulm:
Explicitly mention `query: invalid structure at position` in the docs with the "do not use lazy-loading" advice. 


that way if will be more clear that when the user gets: 
```
Error executing vim.schedule lua callback: ...im/0.9.5/share/nvim/runtime/lua/vim/treesitter/query.lua:259: query: invalid structure at position 3158 for language lua                                                                                                                                            
stack traceback:                                                                                                                                                                                                                                                                                                  
        [C]: in function '_ts_parse_query'                                                                                                                                                                                                                                                                        
        ...im/0.9.5/share/nvim/runtime/lua/vim/treesitter/query.lua:259: in function 'get'                                                                                                                                                                                                                        
        ...5/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:114: in function 'new'                                                                                                                                                                                                                        
        ...r/neovim/0.9.5/share/nvim/runtime/lua/vim/treesitter.lua:61: in function '_create_parser'                                                                                                                                                                                                              
        ...r/neovim/0.9.5/share/nvim/runtime/lua/vim/treesitter.lua:131: in function 'get_parser'                                                                                                                                                                                                                 
        ...r/neovim/0.9.5/share/nvim/runtime/lua/vim/treesitter.lua:460: in function 'ts_highlighter'                                                                                                                                                                                                             
        ...m/lazy/telescope.nvim/lua/telescope/previewers/utils.lua:152: in function 'highlighter'                                                                                                                                                                                                                
        ...scope.nvim/lua/telescope/previewers/buffer_previewer.lua:234: in function ''                                                                                                                                                                                                                           
```

it's probably because they are using  `lazy.nvim` plugin manager and they set lazy loading for nvim-treesitter (like I did inadvertently by using `event='VeryLazy'` which still loads the nvim-treesitter but "too late".

### DIFF
--- a/README.md
+++ b/README.md
@@ -717,16 +717,19 @@ Try updating the parser that you suspect has changed (`:TSUpdate {language}`) or
 If the error persists after updating all parsers,
 please [open an issue](https://github.com/nvim-treesitter/nvim-treesitter/issues/new/choose).
 
-#### I get `query error: invalid node type at position`
+#### I get `query error: invalid node type at position` or `query: invalid structure at position xxx for language yyy`
 
 This could be due a query file outside this plugin using outdated nodes,
 or due to an outdated parser.
 
 - Make sure you have the parsers up to date with `:TSUpdate`
+- Make sure that you don't lazy load `nvim-treesitter`
 - Make sure you don't have more than one `parser` runtime directory.
   You can execute this command `:echo nvim_get_runtime_file('parser', v:true)` to find all runtime directories.
   If you get more than one path, remove the ones that are outside this plugin (`nvim-treesitter` directory),
   so the correct version of the parser is used.
+
+
 
 #### I experience weird highlighting issues similar to [#78](https://github.com/nvim-treesitter/nvim-treesitter/issues/78)
 


ribru17:
Hm, do you get this immediately when opening a file? I don't have this issue when using `VeryLazy` to load this plugin. I wonder if perhaps something else is at play (maybe caused by another plugin trying to parse a query) that only comes out when you defer loading of `nvim-treesitter`?

ecerulm:
> Hm, do you get this immediately when opening a file? 

No, I don't get it when opening a file at all. I believe it only happens when I open `telescope.nvim` and certain files are show on the `File Preview`

image

For example, I opened `:e ~/.config/nvim/lua/ecerulm/plugins/vim-fugitive.lua` with no problems, but if I `:Telescope find_files` and navigate to the same file I get the error


If I disable lazy loading `lazy=false` and do the same `:Telescope find_files` the `File Preview` will load properly (with treesitter syntax highlighting) 

image

> I wonder if perhaps something else is at play (maybe caused by another plugin trying to parse a query) that only comes out when you defer loading of nvim-treesitter?

Sure,  I guess it is the interaction with telescope. But anyway @clason (a maintainer of this project) said to me in [this discussion](https://github.com/nvim-treesitter/nvim-treesitter/discussions/5289#discussioncomment-9045341) that lazy loading nvim-treesitter is NOT supported: 

>> (DO NOT lazy-load nvim-treesitter; that is not supported!)

I don't think this is super clear anywhere, maybe it should also be warned about at https://github.com/nvim-treesitter/nvim-treesitter/wiki/Installation but I can't write to that wiki page, maybe @clason  can add a warning in that wiki page saying " don't lazy load nvim-treesiter."

In any case, I think mentioning in the FAQ section of the README it's at least an improvement,

clason:
I'll think about. We can't guard against _every_ misconfiguration; nothing official recommends or even suggests lazy-loading, so it's on users (or whatever content creators) to push their features responsibly.

ecerulm:
>  it's on users (or whatever content creators) to push their features responsibly

yes, I'm not putting any blame on nvim-treesitter.  I just lazy loaded  nvim-treesitter following a content creator "tutorial" (https://www.youtube.com/watch?v=jWZ_JeLgDxU and accompanying https://github.com/bcampolo/nvim-starter-kit/blob/e76b5c2c7b6f73bdbcfab7f5fd1467842b9bc32e/.config/nvim/lua/plugins/nvim-treesitter.lua#L5)

But since people are already doing the wrong thing (lazy loading nvim-tresitter) it's possible to warn explicitly  against lazy-loading on the docs to fight the misinformation. 

For example mason.nvim in https://github.com/williamboman/mason.nvim?tab=readme-ov-file#setup recommends explicitly not to use lazy loading 
>> Lazy-loading the plugin, or somehow deferring the setup, is not recommended.

ObserverOfTime:
*Never follow youtube tutorials.*

clason:
We'll revisit this for the (completely rewritten) 1.0 release. Until then, the pinned issue is the best documentation.

PR #6418, opened 2024-04-09
Nix cleanups

ribru17:
This PR gives two changes:

1. First commit fixes an issue where strings inside string interpolations weren't working:

    ```nix
    {
      test = "this ${would + "not work"}";
    }
    ```

    It does this by only applying the string highlights to the fragments themselves (which do not include the 
    interpolations) and the quote delimiters.

2. Second commit removes the special `derivation` query, lumping it in as a `@function.builtin` (as it is defined [here](https://nixos.org/manual/nix/stable/language/derivations)), and it also applies the `@keyword.import` queries at the end of the file so as to remove any need for using `priority` workarounds.

    It also adds tests.

### DIFF
--- a/queries/nix/highlights.scm
+++ b/queries/nix/highlights.scm
@@ -8,17 +8,6 @@
   "with"
 ] @keyword
 
-(variable_expression
-  name: (identifier) @keyword
-  (#eq? @keyword "derivation")
-  (#set! "priority" 101))
-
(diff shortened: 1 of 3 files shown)


lucario387:
Can we also add the injections in #5785 as well

lucario387 (approved):

PR #6420, opened 2024-04-09
feat(java): highlight when operator

guilhas07:
Add `when` operator so that it gets highlighted inside the switch guard.


ribru17:
Can you put it as `@keyword.conditional` instead? Seems like other languages with a similar keyword mostly do this

ribru17 (approved):

PR #5402, opened 2023-09-15
feat(textproto): Add Textproto parser and queries

PorterAtGoogle:

PR #6427, opened 2024-04-10
fix(textproto): use 'pbtxt' filetype

ambroisie:
This is the actual name of the detected filetype.

### DIFF
--- a/scripts/minimal_init.lua
+++ b/scripts/minimal_init.lua
@@ -8,7 +8,6 @@ vim.filetype.add {
     cmm = "t32",
     hurl = "hurl",
     ncl = "nickel",
-    textproto = "pbtxt",
     tig = "tiger",
     usd = "usd",
     usda = "usd",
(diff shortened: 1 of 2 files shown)


ambroisie:
Once https://github.com/vim/vim/pull/14463 is merged into NeoVim, we can even remove the `vim.filetype.add` call.

scripts/minimal_init.lua
```diff
@@ -8,7 +8,7 @@ vim.filetype.add {
     cmm = "t32",
     hurl = "hurl",
     ncl = "nickel",
-    textproto = "textproto",
+    textproto = "pbtxt",
```
clason, line 11:
```suggestion
```
If `pbtxt` is _the_ filetype, then this is not needed.

ambroisie: The test files are using `*.textproto` as their extension.

ambroisie: I thought #5402 had added test files using the `*.textproto`  extension, I was mistaken.

clason: Even if so, these could be changed ;)

PR #6437, opened 2024-04-13
feat(t32): use filetype "trace32"

xasc:
Neovim prerelease builds contain the filetype "trace32" for PRACTICE scripts (see [neovim#23967](https://github.com/neovim/neovim/pull/23967)). I would like to claim it for the `t32` parser.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2004,6 +2004,7 @@ list.t32 = {
     url = "https://gitlab.com/xasc/tree-sitter-t32.git",
     files = { "src/parser.c", "src/scanner.c" },
   },
+  filetype = "trace32",
   maintainers = { "@xasc" },
 }
 


lucario387:
Do we also rename the parser to trace32?

lucario387:
```diff
-list.t32
+list.trace32
```

clason:
No, since we don't control the symbol of the parser. Do not mess with the language name.

lua/nvim-treesitter/parsers.lua
```diff
@@ -2004,6 +2004,7 @@ list.t32 = {
     url = "https://gitlab.com/xasc/tree-sitter-t32.git",
     files = { "src/parser.c", "src/scanner.c" },
   },
+  filetype = "trace32",
```
clason, line 2007: This means it will no longer start for a `t32` filetype.

lucario387: Vimland doesn't have t32 as a filetype anymore so I don't think t32 fieltype is needed

clason: Just wanted to make sure that's intended.

xasc: I don't think Vim really had a file type `t32` in the past, so this change shouldn't cause many problems for users.

The only complaints I received were that the new file type may break custom file type definitions in Vim and that it was not immediately obvious what had been causing this.

clason (approved):

PR #6439, opened 2024-04-13
fix: update queries to work with public expression nodes

JoranHonig:


clason (approved):

clason:
Test failure is unrelated (wing parser can't be generated using the latest tree-sitter CLI). Gotta park this for a bit.

PR #6440, opened 2024-04-13
fix(t32): update queries from upstream

xasc:
Adds support for new node type `elif_block`.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -684,7 +684,7 @@
     "revision": "1af543a96d060b1f808982037bfc54cc02218edd"
   },
   "t32": {
-    "revision": "8487a14874f3e3138877655f303322d9e7222cec"
+    "revision": "6182836f4128725f1e74ce986840d7317021a015"
   },
   "tablegen": {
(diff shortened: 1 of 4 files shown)


clason:
Please run `scripts/format-queries.lua queries/t32`.

xasc:
Autoformatter applied. Thanks for the tip.

clason (approved):

clason:
@xasc I know this is a big ask, but would you consider setting up a Github mirror for tree-sitter-t32? We had issues with gitlab being flaky or inaccessible from, e.g., Brazil, and that would reduce significant friction on our end. (I know relying on one -- and arguably this -- codeforge is not great, but the automation requires high reliability which the other codeforges simply don't (yet) have.)

(t32 is one out of only four parsers we pull from gitlab, and the only maintained one, so this would make a big difference for us.)

xasc:
@clason I fully understand your perspective. Thanks for letting me know. I guess then I will just have to bite the bullet.

clason:
If it's not too much to ask, that would be appreciated (other codeberg-hosted parsers also have github mirrors so I thought it'd be easy to set up).

PR #6441, opened 2024-04-13
feat(mojo): add mojo language

austincummings:
Copy over Python queries and extend with Mojo specific syntax.

### DIFF
--- a/queries/mojo/highlights.scm
+++ b/queries/mojo/highlights.scm
@@ -273,15 +273,15 @@
 
 [
   "def"
-  "fn"
+  "fn" ; from mojo
   "lambda"
 ] @keyword.function
 
 [
   "assert"
   "class"
-  "struct"
-  "trait"
+  "struct" ; from mojo
+  "trait" ; from mojo
   "exec"
   "global"
   "nonlocal"
@@ -290,11 +290,11 @@
   "with"
   "as"
   "type"
-  "var"
-  "alias"
-  "borrowed"
-  "inout"
-  "owned"
+  "var" ; from mojo
+  "alias" ; from mojo
+  "borrowed" ; from mojo
+  "inout" ; from mojo
+  "owned" ; from mojo
 ] @keyword
 
 [
@@ -342,7 +342,7 @@
   "except"
   "except*"
   "raise"
-  "raises"
+  "raises" ; from mojo
   "finally"
 ] @keyword.exception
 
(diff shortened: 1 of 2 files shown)


ribru17:
Instead of copying applicable queries, please use `; inherits: python` at the top of the file (unless maybe they aren't perfectly compatible)

clason:
In particular, does the `mojo` _grammar_ inherit from Python? If not, why not?

lucario387:
mojo-python is like typescript-javascript. All valid python codes are valid mojo code, not the other way around.

Seeing how the parser is written, I have some feeling there will be some decent breakage from python :? So keeping it separated will mean less maintanence effort

lua/nvim-treesitter/parsers.lua
```diff
@@ -1278,6 +1278,15 @@ list.mlir = {
   maintainers = { "@artagnon" },
 }
 
+list.mojo = {
+  install_info = {
+    url = "https://github.com/austincummings/tree-sitter-mojo",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
```
clason, line 1285:
```suggestion
```
Only needed if it's not the default branch.

clason:
> that's a valid concern or not though since I haven't worked much with this.

That _is_ a valid concern, and something you'd have to keep an eye on as parser (and query) maintainer.

clason:
@austincummings please also test that `TSInstallFromGrammar mojo` works without errors. (`tree-sitter generate src/grammar.json`, basically)

queries/mojo/highlights.scm
```diff
@@ -0,0 +1,476 @@
...
+
+((call
+  function: (attribute
+    attribute: (identifier) @constructor))
+  (#lua-match? @constructor "^%u"))
+
+; Decorators
+((decorator
+  "@" @attribute)
+  (#set! "priority" 101))
```
lucario387, line 62: Wait, why?
If anything, move this down so you don't need priorities for it

queries/mojo/highlights.scm
lucario387, line 1: Can you note in this file (and others, if possible) which part are mojo-specific? That'd be helpful

queries/mojo/highlights.scm
```diff
@@ -0,0 +1,480 @@
...
+; from mojo
+(struct_definition
+  body: (block
+    (function_definition
+      name: (identifier) @function.method)))
+
+; This one causes a crash for some reason
+; (struct_definition
+;   superclasses: (argument_list
+;     (identifier) @type))
```
clason, line 427: 👀 

You might want to run the workflows from https://github.com/tree-sitter/workflows against your scanner, in particular.

lucario387:
Whichever you think is more suitable for you and us.

lucario387:
Is it ready for review now?

clason:
@austincummings friendly ping

PR #6459, opened 2024-04-17
ci: add back upload-artifact

KortanZ:
Since #2320, artifact uploading had been removed with "binary release archives soonish?". But till then, we still have no parser binary release archive, and it's really inconvenient for users who have trouble to compile parsers. 
Could we keep artifacts until we figure out a more proper solution?


clason:
Sorry, this is not maintainable, and I don't want to offer anything we cannot support. (And note that macOS CI runs on M1 hardware now.)

PR #6460, opened 2024-04-17
feat(fennel)!: update parser and queries

alexmozaidze:
> A continuation of #6323

Updated parser to support:
1. complex forms (`case`, `case-try`, `fn`, `lambda`)
2. forms with destructuring (`global`, `local`, `var`, `set`)
3. forms that may be viable to query via `#has-parent?`/`#has-ancestor?` (`hashfn`, `quote`)

The queries are updated as well to support the new parser.

> [!NOTE]\
> `locals.scm` and `folds.scm` are supported on a best-effort basis. If you encounter any issues regarding those queries, open an issue and @ me so I can fix it as quickly as I can.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -547,7 +547,7 @@ list.faust = {
 list.fennel = {
   install_info = {
     url = "https://github.com/alexmozaidze/tree-sitter-fennel",
-    files = { "src/parser.c" },
+    files = { "src/parser.c", "src/scanner.c" },
   },
   maintainers = { "@alexmozaidze" },
 }
--- a/lockfile.json
+++ b/lockfile.json
@@ -168,7 +168,7 @@
     "revision": "f3b9274514b5f9bf6b0dd4a01c30f9cc15c58bc4"
   },
   "fennel": {
-    "revision": "215e3913524abc119daa9db7cf6ad2f2f5620189"
+    "revision": "63e16deb8e66bff61a09f7582c60945ed216cd65"
   },
   "fidl": {
     "revision": "0a8910f293268e27ff554357c229ba172b0eaed2"
--- a/queries/fennel/folds.scm
+++ b/queries/fennel/folds.scm
@@ -1,34 +1,49 @@
+; compounds
 [
   (list)
   (table)
   (sequence)
 ] @fold
 
-(list
-  .
-  (symbol) @_let
-  (#eq? @_let "let")
-  .
-  (sequence) @fold) @fold
+; sub-forms / special compounds
+[
+  (list_binding)
+  (table_binding)
+  (sequence_binding)
+  (table_metadata)
+  (sequence_arguments)
+  (let_vars)
+  (case_guard_or_special)
+  (case_guard)
+  (case_catch)
+] @fold
 
-(list
-  .
-  (symbol) @_local
-  (#eq? @_local "local")) @fold
+; forms
+[
+  (quote_form)
+  (unquote_form)
+  (local_form)
+  (var_form)
+  (set_form)
+  (global_form)
+  (let_form)
+  (fn_form)
+  (lambda_form)
+  (hashfn_form)
+  (each_form)
+  (collect_form)
+  (icollect_form)
+  (accumulate_form)
+  (for_form)
+  (fcollect_form)
+  (faccumulate_form)
+  (case_form)
+  (match_form)
+] @fold
 
-(list
-  .
-  (symbol) @_global
-  (#eq? @_global "global")) @fold
+; reader macros
+(quote_reader_macro
+  expression: (_) @fold)
 
-(list
-  .
-  (symbol) @_fn
-  (#any-of? @_fn "fn" "lambda" "λ" "hashfn")) @fold
-
-(reader_macro
-  macro: [
-    "'"
-    "`"
-  ]
+(quasi_quote_reader_macro
   expression: (_) @fold)
--- a/queries/fennel/highlights.scm
+++ b/queries/fennel/highlights.scm
@@ -12,44 +12,87 @@
   "]"
 ] @punctuation.bracket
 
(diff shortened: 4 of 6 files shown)


clason:
Looks like you already merged the breaking PR, so we need to follow suit ASAP.

alexmozaidze:
What breaking PR?

clason:
In your repo.

alexmozaidze:
I don't quite get it. Do you mean the parser repo itself, or the rebased branch? If the second one, then could you point me to the breaking commit? Also, what do you mean by "we need to follow suit"?

clason:
Your parser repo. If we don't merge your PR, all future automated updates are blocked. (This is not your problem.)

ObserverOfTime (approved):
Do [textobjects](https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/master/queries/fennel/textobjects.scm) and [context](https://github.com/nvim-treesitter/nvim-treesitter-context/blob/master/queries/fennel/context.scm) also need to be updated?

alexmozaidze:
Most likely. You cannot refer to some forms with the following manner anymore:
```scm
(list
  call: (symbol) @keyword
  (#eq? @keyword "some-form"))
```

All special forms that the parser supports are wrapped in their own node, along with some forms having specialized nodes. (like function's arguments list being `sequence_arguments` instead of `sequence`)

I'm working on the documentation along with refactors, but for now, you can open the [extensions/builtin-forms.js](https://github.com/alexmozaidze/tree-sitter-fennel/blob/main/extensions/builtin-forms.js) file and search for `rules\['[^_]\|\[\n\t'[^_]`, which will show you all the public nodes.

clason:
If you would consider making prs to these two repos, that would be very much appreciated (though not required).

alexmozaidze:
Sure! Queries seem simple enough.

Also, I just realized that I forgot to add `item:` field to tables (which is useful for querying empty compounds), I'll do that and get to the queries.

alexmozaidze:
While working on textobjects queries, I realized that the `if` special should be supported on the parser's side since it's associative, and associativity is hard to capture reliably in queries. I'll make that change and finish up the queries, though, be warned that it'll be a breaking change on parser's side.

clason:
okie, parking this until then

lucario387:
It's already a breaking change if there's a renamed node

alexmozaidze:
Yeah, and, unfortunately, it will always be a breaking change to support some form, since it will make `(list)` representation of that form an impossible pattern to match, even though it's not an error from queries perspective.

clason:
(we use the label for changes which require reinstalling the parser, e.g., because the repo changed. breaking changes to queries happen all the time, so the constant signaling would remove the point of the label.)

clason:
@alexmozaidze you might want to fix your regexes while you're at it: `u{[\da-fA-F]+}` is now illegal since tree-sitter switched to proper javascript regex (@amaanq can explain in detail).

alexmozaidze:
Weren't regexes always surrounded by slashes? Isn't it, like, *the* syntax of regex literals?

clason:
(@lucario387 sorry, you were right: it's breaking because it adds a scanner and now requires npm. basically, any change to the parsers.lua table is breaking).

@alexmozaidze please mark your parser as "generates_requires_npm" and make sure to always commit an updated `grammar.json`. Please test `TSInstallFromGrammar fennel`.)

clason:
> Weren't regexes always surrounded by slashes? Isn't it, like, _the_ syntax of regex literals?

Or they might be removed now, or have to be added in specific places (hence my ping for @amaanq to explain). I just know this has changed in recent versions, and `tree-sitter generate` fails because of it.

clason:
Ah, right, you need to escape the braces now:
```
    /u\{[\da-fA-F]+\}/,
```

alexmozaidze:
I reckon it would need the latest git version of the cli, I'll update and see what I can do.
Also, is it possible to avoid marking the PR as a breaking one if I remove the npm dependency, or is it also influenced by the now present `scanner.c`?

clason:
Don't worry about the breaking label; that's fine -- we just want to signal people to pay attention.

> I reckon it would need the latest git version of the cli, I'll update and see what I can do.

No, latest release (0.22.5) is fine.

lua/nvim-treesitter/parsers.lua
```diff
@@ -547,7 +547,7 @@ list.faust = {
 list.fennel = {
   install_info = {
     url = "https://github.com/alexmozaidze/tree-sitter-fennel",
-    files = { "src/parser.c" },
+    files = { "src/parser.c", "src/scanner.c" },
```
clason, line 550:
```suggestion
    files = { "src/parser.c", "src/scanner.c" },
    generate_requires_npm = true,
```

clason (changes requested):

alexmozaidze:
It's best I polish the parser a bit and document it before making changes here. Whilst making my finishing touches, I found a few bugs.

I'll do the required changes once it's all *done* done.

clason:
Honestly, I'd prefer to rip off the bandaid now, since the breaking changes are adding friction to the update workflow. You can always make further changes later.

In particular, I want to get the regex fix in to unbreak `TSInstallFromGrammar`.

alexmozaidze:
I think all the necessary things are done.

clason:
Except merging my PR ;)

clason:
No worries! You missed a small gotcha, but that's not breaking for nvim-treesitter.

clason (approved):

alexmozaidze:
I'll make a PR to textobjects shortly, but I do wonder what was the gotcha that I've missed?

clason:
The `binding.gyp` got overwritten by a new one that was missing the `scanner.c`. (Don't you get notified by PRs on your own repository?)

alexmozaidze:
I apparently had the `Automatically watch repositories` option off, and I never even touched that option.

PR #6462, opened 2024-04-17
ci: Don't run `Update lockfile` on forks

savq:
Prevent the bot from opening PRs on forks of this repo.

See [jobs.<job_id>.if](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif)

### DIFF
--- a/.github/workflows/update-lockfile.yml
+++ b/.github/workflows/update-lockfile.yml
@@ -9,6 +9,7 @@ jobs:
   update-lockfile:
     name: Update lockfile
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


clason:
Just don't enable workflows on your fork? That's a standard setting.

savq:
Yeah, I always forget to update my settings so I thought it'd be nice if it were the default. 😅

clason:
Thank you, but I don't think we need to single out this workflow. It really is better to just disable all workflows on forks.

PR #6478, opened 2024-04-20
feat: change astro injections to match parser rewrite

virchau13:
The Astro parser's extension logic has been almost entirely rewritten to match up with the actual Astro compiler; these are the new injections after the rewrite.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -18,7 +18,7 @@
     "revision": "afdc3d5da18d42cbb471c0f40527dbed9cace7ad"
   },
   "astro": {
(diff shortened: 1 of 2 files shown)


clason:
If the rewrite _just_ landed, you'll have to bump the lockfile entry to match. Are you sure your branch is (re)based on the very latest `master`?

virchau13:
Oops, forgot about the lockfile bump.
Yep, I pulled down the latest master before making this PR.

clason (approved):

PR #6482, opened 2024-04-20
feat(php): fold consecutive imports

przepompownia:
- It has no effect in Neovim's `vim.treesitter.foldexpr()` before https://github.com/neovim/neovim/commit/5e6240ffc24e55ecf7721fd9dc3f33c6f178be8c
- It has no effect for the current `nvim_treesitter#foldexpr()`

### DIFF
--- a/queries/php_only/folds.scm
+++ b/queries/php_only/folds.scm
@@ -13,4 +13,5 @@
   (enum_declaration)
   (function_static_declaration)
   (method_declaration)
+  (namespace_use_declaration)+
 ] @fold
--- a/queries/php_only/folds.scm
+++ b/queries/php_only/folds.scm
@@ -13,5 +13,5 @@
   (enum_declaration)
   (function_static_declaration)
   (method_declaration)
-  (namespace_use_declaration)+
+  ((comment)* (namespace_use_declaration))+
 ] @fold
--- a/queries/php_only/folds.scm
+++ b/queries/php_only/folds.scm
@@ -13,5 +13,6 @@
   (enum_declaration)
   (function_static_declaration)
   (method_declaration)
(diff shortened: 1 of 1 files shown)


przepompownia:
It's rather not typical in PHP because of automatic import by a language server or any equivalent. I can't remember any project from the past. 

I didn't even check what happens with comments after such automatic import or sorting with any formatter (php-cs-fixer for example). 

If I find a moment to focus I can try to fix it here or provide it separately unless you want to add such interleaving earlier.

ribru17:
I think it's fine then if it is uncommon. In my testing the comments break up the folds (even if they are on the same line as the `use` statements) but it looks like that will rarely be an issue so maybe it is just better to leave it :+1: What do you think?

przepompownia:
The fix turned out to be trivial. Before fixing I didn't sure if `*` will work as expected.

ribru17:
With the current solution I get folds looking like this:
phpfoldissue
I think queries with quantifiers around parentheses aren't well handled at this time, because the very top part should not be folded. I also tried:
```scheme
[(comment) (namespace_use_declaration)]+ @fold
```
but then that also folds groups of comments :/

przepompownia:
In other words an empty space cannot be a separator between comments. 

I would be willing to restore the initial version assuming almost no use in practice.

ribru17 (approved):

ribru17:
Thanks!

przepompownia:
Thanks for critical eye on it. It couldn't hurt to check how embracing comments works.

lucario387:
Just a quick note: `((comment)* (namespace_use_declaration))+`, from what I can read, equates to `(comment)*+ (namespace_use_declaration)`

PR #6483, opened 2024-04-21
feat(astro): add backtick string injections

virchau13:
This also updates a highlight to be more accurate and bumps the lockfile to fix a few bugs.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -18,7 +18,7 @@
     "revision": "afdc3d5da18d42cbb471c0f40527dbed9cace7ad"
   },
   "astro": {
-    "revision": "1aab99ed2af477fe4283365226589908809a696d"
+    "revision": "506515c1dce4220fc9db278ce9eae2ad4e239cff"
   },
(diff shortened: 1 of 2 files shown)


queries/astro/highlights.scm
```diff
@@ -19,3 +19,9 @@
 ((erroneous_end_tag
   (erroneous_end_tag_name) @type)
   (#lua-match? @type "^[A-Z]"))
+
+(start_tag
+  (fragment_tag_name) @tag.delimiter)
```
lucario387, line 24: tag.delimiter is for `<`, `>` and the like, and fragment is, I'd assume, `</>`? So naming it as tag_name and then capturing with tag.delimiter seems a bit off

virchau13:
Yeah that probably wasn't best-practices. Fixed.

clason:
(may want to rebase on master)

virchau13:
Rebased.

lucario387 (approved):

PR #6487, opened 2024-04-21
feat(install)!: generate from json instead of requiring node

clason:
Problem: Many parsers require node/npm to evaluate the `grammar.js` before being able to generate a parser from it.

Solution: Generate from `grammar.json` instead, which is fully resolved. Drops `node` and `npm` as (optional) requirements for nvim-treesitter.

Note that this requires parsers to commit the generated json **iff** the grammar requires evaluation (which is currently the case for all tracked languages).

Part #2 of my series of PRs simplifying the install code. While testing this, I found out that a good 20% of the listed parsers actually fail to generate from grammar, mostly due to breaking changes in `tree-sitter` that parsers didn't adapt to. The failures roughly fall into three classes:

* **Invalid regex:** Upstream switched to Rust regex, so the common `escape_sequence` pattern `/u{[\da-fA-F]+}/` now needs to escape the braces: `/u\{[\da-fA-F]+\}/`. This affects

- [ ] bass
- [ ] blueprint
- [ ] dart
- [x] fennel
- [ ] fsh
- [ ] nasm
- [ ] qmljs
- [ ] r
- [ ] rbs
- [ ] smithy
- [ ] supercollider
- [ ] usd

* **Invalid `package.json`:** Upstream now errors out if the `package.json` doesn't contain a `"tree-sitter"` key. This should only affect generating bindings (which we don't), so this is in fact an upstream bug. Nevertheless, grammars may want to fix that for other users. This affects

- [ ] beancount
- [ ] comment
- [ ] eds
- [ ] elvish
- [ ] gdscript
- [ ] godot_resource
- [ ] gowork
- [ ] graphql
- [ ] hjson
- [ ] htmldjango
- [ ] jq
- [ ] json5
- [ ] jsonc
- [ ] kusto
- [ ] liquid
- [ ] menhir
- [ ] nim_format_string
- [ ] org
- [ ] passwd
- [ ] pioasm
- [ ] prisma
- [ ] prql
- [ ] rnoweb
- [ ] surface
- [ ] turtle
- [ ] verilog
- [ ] yang

* If  there is a `tree-sitter` key, it must be an array: ` "tree-sitter": [{...}]`, not `"tree-sitter": {...}`. This affects

- [ ] angular
- [ ] cooklang

* **Upstream bug**: (probably https://github.com/tree-sitter/tree-sitter/issues/768)

- [ ] haskell_persistent

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -31,7 +31,7 @@ local M = {}
 ---@type table<string, LockfileInfo>
 local lockfile = {}
 
-local max_jobs = 50
+local max_jobs = 10
 
 local iswin = uv.os_uname().sysname == 'Windows_NT'
 local ismac = uv.os_uname().sysname == 'Darwin'
@@ -145,24 +145,17 @@ end
 --- @param repo InstallInfo
 --- @param compile_location string
 --- @return string? err
-local function do_generate_from_grammar(logger, repo, compile_location)
+local function do_generate(logger, repo, compile_location)
   if not executable('tree-sitter') then
     return logger:error('tree-sitter CLI not found: `tree-sitter` is not executable')
   end
 
-  if repo.generate_requires_npm then
-    if not executable('npm') then
-      return logger:error('NPM requires to be installed from grammar.js')
-    end
-
-    logger:info('Installing NPM dependencies')
-    local r = system({ 'npm', 'install' }, { cwd = compile_location })
-    if r.code > 0 then
-      return logger:error('Error during `npm install`: %s', r.stderr)
-    end
-  end
-
-  logger:info('Generating source files from grammar.js...')
+  logger:info(
+    string.format(
+      'Generating source files from %s...',
+      repo.generate_from_json and 'grammar.json' or 'grammar.js'
+    )
+  )
 
   local r = system({
     fn.exepath('tree-sitter'),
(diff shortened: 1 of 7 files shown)


lua/nvim-treesitter/install.lua
```diff
@@ -31,7 +31,7 @@ local M = {}
 ---@type table<string, LockfileInfo>
 local lockfile = {}
 
-local max_jobs = 50
+local max_jobs = 10
```
clason, line 34: 50 was way too much once `tree-sitter` got into the mix; the memory pressure skyrocketed. This is also friendlier for CI (macos went from 1:40 to 1:10 for compiling all parsers). @lewis6991

clason:
@gabmus since your blueprint parser is on gitlab, I can't make a PR to fix this for you. Please update or we'll have to drop it from nvim-treesitter.

GabMus:
Sorry, what change needs to be made? Also you can still do PRs on gitlab, you just have to sign up or send it in via email or something

clason:
> Also you can still do PRs on gitlab, you just have to sign up or send it in via email or something

Yeah, sorry, but I won't do that, that's too much for me (and the diff is too large for email).

> Sorry, what change needs to be made?

See the "invalid regex" item above: in your `grammar.js`, the `escape_sequence` pattern `/u{[\da-fA-F]+}/` now needs to escape the braces: `/u\{[\da-fA-F]+\}/`. Then `tree-sitter generate --no-bindings` with the latest tree-sitter CLI (0.22.5).

GabMus:
Should be fixed with the latest commit, please let me know if there's anything else needed on my end

clason:
Yes, looks good, thank you!

PR #6488, opened 2024-04-22
feat: import statement folds for common languages

ribru17:
Some fold queries for import statements (or the equivalent) across some common languages. Pre-0.9 nvim is unaffected by these queries, except in cases (like the rust one) where the quantifier was added to a node already in the fold set. In these cases the fold will no longer apply to that node. I think this isn't too big a deal, but feel free to suggest only partially merging if it is :+1:

Also I think these are the only additions needed for the parsers bundled in neovim, so once they are upstreamed, there should be nothing left to do

### DIFF
--- a/queries/c/folds.scm
+++ b/queries/c/folds.scm
@@ -16,6 +16,7 @@
   (preproc_function_def)
   (initializer_list)
   (gnu_asm_expression)
+  (preproc_include)+
 ] @fold
 
 (compound_statement
--- a/queries/java/folds.scm
+++ b/queries/java/folds.scm
@@ -4,4 +4,5 @@
   (constructor_declaration)
   (argument_list)
   (annotation_argument_list)
+  (import_declaration)+
 ] @fold
--- a/queries/python/folds.scm
+++ b/queries/python/folds.scm
@@ -21,3 +21,8 @@
   (dictionary)
   (string)
 ] @fold
+
+[
+  (import_statement)
+  (import_from_statement)
+]+ @fold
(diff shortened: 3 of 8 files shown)


queries/python/folds.scm
```diff
@@ -20,4 +20,8 @@
   (set)
   (dictionary)
   (string)
+  [
+    (import_statement)
+    (import_from_statement)
+  ]+
```
lucario387, line 26: Let's just separate them outside of this. Reading this is a bit hard

queries/c/folds.scm
```diff
@@ -16,6 +16,7 @@
   (preproc_function_def)
   (initializer_list)
   (gnu_asm_expression)
+  (preproc_include)+
```
lucario387, line 19: This breaks for the closest use case: Neovim codebase. (IWYU pragma: keep)

queries/java/folds.scm
```diff
@@ -4,4 +4,5 @@
   (constructor_declaration)
   (argument_list)
   (annotation_argument_list)
+  (import_declaration)+
```
lucario387, line 7: Java is fine, I like this change.

lucario387 (commented):
Aside from the C change, I'm fine with the added queries (maybe rusteaceans have decent enough use-and-comment mixing?)

queries/java/folds.scm
ribru17: I scanned around github for things like this and it seemed like in the majority of cases there were no comments in between the rust use statements

queries/c/folds.scm
lucario387: Well if it didn't fully break for neovim, it's fine for me :D

lucario387 (approved):

PR #6491, opened 2024-04-22
feat(idl): add idl language

cathaysia:
add omg idl language https://github.com/cathaysia/tree-sitter-idl

image


clason:
Thank you, but we don't accept "experimental" languages anymore. Either this is ready to be used (in which case the experimental label should be dropped), or it needs more work before submitting it here.

cathaysia:
This is not an experimental language, vim already has its syntax files.  This language is part of the omg specification

please check the readme of https://github.com/cathaysia/tree-sitter-idl

lua/nvim-treesitter/parsers.lua
```diff
@@ -948,6 +948,26 @@ list.hyprlang = {
...
 }
 
+list.idl = {
+  install_info = {
+    url = "https://github.com/cathaysia/tree-sitter-idl",
+    location = "tree-sitter-idl",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@cathaysa" },
+  experimental = true,
```
clason, line 958: I'm talking about this: If the parser is not ready for production use, we don't want to include it (yet).

clason:
And where is the `grammar.js` in that repo? Please test whether `:TSInstallFromGrammar` works.

queries/idl/injections.scm
```diff
@@ -0,0 +1,3 @@
+((comment) @injection.content
```
clason, line 1: Normally you'd inject the `comment` parser here; you should then add an injection for `(annotation_comment)` in `idl_annotation`.

clason:
> And where is the `grammar.js` in that repo? Please test whether `:TSInstallFromGrammar` works.

Before you put too much time into this, can you answer this question? This could be a deal-breaker for us.

cathaysia:
> And where is the `grammar.js` in that repo? Please test whether `:TSInstallFromGrammar` works.

the grammar.js is generated by `npm build`, maybe I should add it to git?

clason:
Yes. We do not want to depend on node if it can't be helped.

queries/idl/injections.scm
cathaysia: what you means is this?
```scheme
((comment) @injection.content

((comment) @injection.content
  (#lua-match? @injection.content "/[*\/][!*\/]<?[^a-zA-Z]")
  (#set! injection.language "doxygen"))

((comment) @injection.content
  (#lua-match? @injection.content "//@[a-zA-Z]")
  (#set! injection.language "idl_annotation"))
```

cathaysia: so it should be this?

```scheme
((comment) @injection.content
  (#lua-match? @injection.content "/[*\/][!*\/]<?[^a-zA-Z]")
  (#set! injection.language "doxygen"))

((comment) @injection.content
  (#lua-match? @injection.content "//@[a-zA-Z]")
  (#set! injection.language "idl_annotation"))
```

These two injections will not be selected at the same time

lucario387: And a 
```
((comment) @injection.content
  (#not-lua-match? @injection.content "/[*\/][!*\/]<?[^a-zA-Z]")
  (#not-lua-match? @injection.content "//@[a-zA-Z]")
```

lucario387: (Tbh comment parser is a special case, you can have multiple injections with comment nodes, but only comment nodes)

clason: No, I mean add injections for the injected `idl_annotation` language.

queries/idl_annotation/highlights.scm
```diff
@@ -0,0 +1,20 @@
...
+[
+  "("
+  ")"
+] @punctuation.bracket
+
+"=" @operator
+
+[
+  (char_literal)
+  (string_literal)
```
ribru17, line 17: Should this one be `@string`?

queries/idl_annotation/highlights.scm
```diff
@@ -0,0 +1,20 @@
+; https://neovim.io/doc/user/treesitter.html#treesitter-highlight-groups
+(annotation_comment) @comment
```
ribru17, line 2: Would `@comment.documentation` make more sense?

queries/idl/highlights.scm
```diff
@@ -0,0 +1,206 @@
...
+  "map"
+] @type.builtin
+
+(boolean_literal) @boolean
+
+(number_literal) @number
+
+[
+  (char_literal)
+  (string_literal)
```
ribru17, line 66: Same as below

queries/idl/highlights.scm
```diff
@@ -0,0 +1,206 @@
...
+  "exception"
+] @keyword.type
+
+[
+  "module"
+  "switch"
+  "case"
+  "default"
+  "const"
+  "void"
```
ribru17, line 22: Should `"void"` be considered a builtin type rather than a keyword?

queries/idl/highlights.scm
```diff
@@ -0,0 +1,206 @@
...
+  "@annotation"
+  "interface"
+  "sequence"
+  "exception"
+] @keyword.type
+
+[
+  "module"
+  "switch"
+  "case"
```
ribru17, line 19: Case and switch should usually get `@keyword.conditional`

cathaysia: as you see, switch and case is a part of union declare, I'm not sure whether it should be `@keyword.conditional` or not.

```idl
union A switch(long) {
    case 1:
        u8 a;
    case 2:
    case 3:
        u16 b;
    default:
        u32 c[10];
};
```

ribru17: I personally think they should still, but I don't have strong feelings either way. Also I think that `char_literal` and `string_literal` should be broken up, getting `@character` and `@string` highlights, respectively

queries/idl/injections.scm
cathaysia: > Normally you'd inject the `comment` parser here; you should then add an injection for `(annotation_comment)` in `idl_annotation`.

Sorry, maybe my naming caused confusion. In fact, idl_annotation is used to parse annotation_comment, so it should not need injection:

```js
module.exports = grammar({
  name: 'idl_annotation',
  rules: {
    specification: $ => repeat1($.annotation_comment),
  })
```

queries/idl/highlights.scm
lucario387: This looks similar enough to C-switch/case, so it should follow what C's switch-case captures

cathaysia: The color looks a bit strange:
image
image

lucario387: We don't care about the color here :) Only the correctness of the capture

cathaysia: ok

ribru17: Also maybe it would help if `default` was also `@keyword.conditional`? Apparently C does not do this but many other languages I have seen here do, and maybe C should be doing this as well...

cathaysia: sure, I mark the default also be @keyworkd.conditional

ribru17 (approved):

cathaysia:
The code is almost completed, I think it could be merge.

queries/idl/highlights.scm
```diff
@@ -0,0 +1,206 @@
...
+  (simple_declarator) @property)
+
+(attr_declarator) @property
+
+(annotation_appl
+  "@" @property
+  (scoped_name) @property)
+
+(op_dcl
+  (identifier) @function)
```
ObserverOfTime, line 103:
```suggestion
(op_dcl
  (identifier) @function.method)
```
I think.

ObserverOfTime (changes requested):

queries/idl_annotation/highlights.scm
```diff
@@ -0,0 +1,42 @@
...
+
+(char_literal) @character
+
+(string_literal) @string
+
+(number_literal) @number
+
+(boolean_literal) @boolean
+
+(identifier) @property
```
ObserverOfTime, line 42:
```suggestion
(identifier) @variable.member
```

queries/idl/highlights.scm
```diff
@@ -0,0 +1,206 @@
...
+] @punctuation.delimiter
+
+(readonly_attr_declarator
+  (simple_declarator) @property)
+
+(attr_declarator) @property
+
+(annotation_appl
+  "@" @property
+  (scoped_name) @property)
```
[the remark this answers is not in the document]
cathaysia: The code for the two parsers is basically the same. The only difference between the two is that idl_annotation will only appear in the form of code injection.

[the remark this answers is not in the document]
ObserverOfTime: Is idl_annotation needed if idl already parses the annotations?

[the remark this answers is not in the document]
cathaysia: image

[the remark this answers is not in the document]
cathaysia: idl_annotation is an extension of idl annotation, allowing annotation to appear in comments. However, tree-sitter always parse //@ as a comment. So I separated it out as a separate parser. This problem is solved through language injection.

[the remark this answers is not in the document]
cathaysia: Although all node names are the same in both languages. But each language uses a separate highlight.scm, so I need to implement a highlight.scm for them.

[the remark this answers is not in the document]
ObserverOfTime: Then use `@attribute` in both if they are the same.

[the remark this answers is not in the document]
clason: Please try that; we'd like to avoid split parsers if we can.

queries/idl/highlights.scm
```diff
@@ -0,0 +1,205 @@
...
+  "@annotation"
+  "interface"
+  "exception"
+  "typedef"
+] @keyword.type
+
+[
+  "module"
+  "const"
+  "readonly"
```
ribru17, line 19: I think `"const"` and `"readonly"` should be `@keyword.modifier`

queries/idl/highlights.scm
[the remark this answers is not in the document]
cathaysia: I had try `prec.left` before, but it won't work also. :(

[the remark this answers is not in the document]
cathaysia: I think this problem can be put aside for now and solved later when better methods are available.

cathaysia:
not, it does not works. socped name is not a terminal char, treesitter reject it
________________________________

queries/idl/highlights.scm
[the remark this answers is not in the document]
cathaysia: not, it does not works. socped name is not a terminal char, treesitter reject it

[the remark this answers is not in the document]
cathaysia: yes, that works. I remove idl_annotation. :) thanks

queries/idl/highlights.scm
```diff
@@ -0,0 +1,205 @@
...
+] @punctuation.delimiter
+
+(readonly_attr_declarator
+  (simple_declarator) @variable.member)
+
+(attr_declarator) @variable.member
+
+(annotation_appl
+  "@" @property
+  (scoped_name) @property)
```
ObserverOfTime, line 102:
```suggestion
  "@" @attribute
  (scoped_name) @attribute)
```

ObserverOfTime (approved):

cathaysia:
@amaanq @lucario387 Any suggestions about this?

queries/idl/highlights.scm
```diff
@@ -0,0 +1,205 @@
...
+
+[
+  "enum"
+  "struct"
+  "union"
+  "bitmask"
+  "bitset"
+  "@annotation"
+  "interface"
+  "exception"
```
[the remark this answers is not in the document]
cathaysia: exception also be used defining composite types(just like struct, enum and etc). Maybe it's better to use @keyword.type?

from https://neovim.io/doc/user/treesitter.html#treesitter-highlight-groups

> @keyword.type           keywords defining composite types (e.g. struct, enum)
> @keyword.exception      keywords related to exceptions (e.g. throw, catch)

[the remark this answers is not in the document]
cathaysia: as a example, here is a exception define:

```idl
exception HelloWorld {
    u8 a;
    u16 b[10];
    string c[10][20];
};
```

PR #6496, opened 2024-04-22
feat!: use tree-sitter build

clason:
quick test to compare build times on CI:

* ubuntu: 2:09 vs 1:50
* macos: 1:40 vs 1:20 (locally, 95s vs 85s)
*  windows: 3:55 vs 5:09

network/load adds large variance, but the windows improvement is significant (and welcome on CI).

Also, adding or removing scanners is no longer a breaking change.

Also also, look at those diffstats 🤩 

Notes on possible (upstream) improvements:
1. Ideally, we could build directly into the target location, obviating the need for the move task, but `ts b` litters debug symbols in the build directory.
2. We could combine `generate` and `build` step if `ts b` generated if necessary. For our purposes, though, we'd need more control:
* `ts b` builds from `parser.c`
* `ts b src/grammar.json` calls `ts g src/grammar.json` first
* `ts b grammar.js` calls `ts g` first

(and `ts b` would need to take all flags for `ts g` as well).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 
 # WARNING
 
-**This branch is a [full, incompatible, rewrite of `nvim-treesitter`](https://github.com/nvim-treesitter/nvim-treesitter/issues/4767) and [work in progress](TODO.md). This branch REQUIRES (the latest commit on) Neovim `master`.**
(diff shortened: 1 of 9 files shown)


clason:
Looks like the highlight crate is not compatible with the parsers built using the latest `tree-sitter` CLI? @theHamsta

ObserverOfTime:
Is it time to use `tree-sitter highlight --check` yet?

ObserverOfTime:
This issue is actually fixed by tree-sitter/tree-sitter#3188. We need it in 0.22.6 after all. ;)

Issue #6497, opened 2024-04-23
file types with dot notation are highlighted incorrectly


perryrh0dan:
### Describe the bug

I was just wondering if this behaviour is correct. For example, I have a normal html file. I set the file type to `set ft=html` and everything is highlighted correctly. Now I change the file type to `set ft=html.asdf`. Still the same highlighting, so all is well. But if I rotate the file type to `set ft=asdf.html`, the highlighting changes. I would expect to just skip all invalid/unknown filetypes until we find a valid one.

### To Reproduce

- Open `.html` file
- `set ft=asdf.html`
- Different/Wrong highlighing

### Expected behavior

Skip all unknown file types until we find a valid one, fallback to the first if none is valid.

### Output of `:checkhealth nvim-treesitter`

```markdown
==============================================================================
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
- OK `node` found v18.20.2 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: cc (Debian 12.2.0-14) 12.2.0
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "5.10.102.1-microsoft-standard-WSL2",
  sysname = "Linux",
  version = "#1 SMP Wed Mar 2 00:30:59 UTC 2022"
} ~

Parser/Features         H L F I J
  - c                   ✓ ✓ ✓ ✓ ✓
  - cpp                 ✓ ✓ ✓ ✓ ✓
  - css                 ✓ . ✓ ✓ ✓
  - diff                ✓ . . . .
  - dockerfile          ✓ . . . ✓
  - git_rebase          ✓ . . . ✓
  - gitignore           ✓ . . . .
  - go                  ✓ ✓ ✓ ✓ ✓
  - gomod               ✓ . . . ✓
  - gosum               ✓ . . . .
  - graphql             ✓ . . ✓ ✓
  - hcl                 ✓ . ✓ ✓ ✓
  - html                ✓ ✓ ✓ ✓ ✓
  - http                ✓ . . . ✓
[...]

PR #6499, opened 2024-04-23
feat!: Iterate filetype parts and search for available lang

perryrh0dan:
Like mentioned here: https://github.com/nvim-treesitter/nvim-treesitter/issues/6497

We could iterate over the different file types, separated by `.`, and take the first one that is available.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2393,10 +2393,17 @@ function M.ft_to_lang(ft)
   local result = get_lang(ft)
   if result then
     return result
-  else
-    ft = vim.split(ft, ".", { plain = true })[1]
-    return get_lang(ft) or ft
   end
+
+  local parts = vim.split(ft, ".", { plain = true })
+  local i = 1
+  local lang = nil
+  while lang == nil and parts[i] do
+    lang = get_lang(parts[i])
+    i = i + 1
+  end
+
+  return lang or parts[1]
 end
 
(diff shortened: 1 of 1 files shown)


clason:
Sorry, that function is obsolete. Any changes going forward must be made on the Neovim side.

perryrh0dan:
Could you point me in the direction of where such logic could be placed on the Neovim site? I am completely new to neovim and treesitter.

clason:
Not really, because I don't think it _should_ be placed. First step is arguing why the function I provided is not good enough.

perryrh0dan:
Okay will try, I have following use case: 

I have a file named `index.spec.ts`. The file type of this file is typescript. Now I want to add some code snippets specifically for my test(spec) files. So I thought an elegant solution would be to use the dot separate file type. 

So whenever I open a file named `*.spec.ts` I set the file type to `spec.typescript`. Now all my lsps still see this as a typescript file, so they still work, but I was also able to add some snippets for the file type `spec` that are now only shown in all test(spec) files. So not only in `.spec.ts` but also in `spec.js` for example. At the moment only treesitter does now not work anymore like expected.

But i am also not quiet sure how i could do that with `vim.treesitter.language.add` ?
I would just add new languages called 'spec.typescript' and 'spec.javascript' and provide a path to the `normal` typescript and javascript parser ?

clason:
`vim.treesitter.language.register('javascript', 'spec.javascript')` (sorry, the `add` was a persistent brain-typo.)

PR #6501, opened 2024-04-24
feat: more `@keyword.type` captures

ribru17:
Migrates many (to my knowledge, all) keywords relating to composite types to `@keyword.type`.

I gave a quick google search over all of these languages to make sure these changes made sense, and some I didn't apply for this reason. That said I'm not 100% confident about each of these changes


queries/matlab/highlights.scm
```diff
@@ -7,14 +7,15 @@
   "arguments"
   "classdef"
```
ObserverOfTime, line 8: Add "classdef"

queries/nim/highlights.scm
```diff
@@ -416,14 +416,17 @@
   "do"
   "mixin"
   "static"
-  "object"
```
ObserverOfTime, line 419: Add "type"

queries/ocaml/highlights.scm
```diff
@@ -132,16 +131,20 @@
   "method"
   "module"
   "new"
-  "object"
   "of"
   "sig"
-  "struct"
   "type"
```
ObserverOfTime, line 136: Add "type"

queries/php_only/highlights.scm
```diff
@@ -13,13 +13,11 @@
 ] @keyword.function
 
 [
-  "class"
   "clone"
   "declare"
   "default"
   "echo"
   "enddeclare"
-  "enum"
```
ObserverOfTime, line 22: Add "interface"

queries/pony/highlights.scm
```diff
@@ -5,11 +5,9 @@
 [
   "type"
   "actor"
-  "class"
   "primitive"
   "interface"
   "trait"
```
ObserverOfTime, line 10: Add "type", "interface", "trait"

queries/prisma/highlights.scm
```diff
@@ -2,13 +2,14 @@
 
 [
   "datasource"
-  "enum"
   "generator"
   "model"
   "type"
```
ObserverOfTime, line 7: Add "type"

queries/puppet/highlights.scm
```diff
@@ -13,14 +13,15 @@
 
 ; Keywords
 [
-  "class"
   "inherits"
   "node"
   "type"
```
ObserverOfTime, line 18: Add "type"

queries/purescript/highlights.scm
```diff
@@ -90,7 +90,6 @@
   (where)
   "let"
   "in"
-  "class"
```
ObserverOfTime, line 93: Add "type", "newclass"

queries/python/highlights.scm
```diff
@@ -311,6 +310,8 @@
   "type"
```
ObserverOfTime, line 310: Add "type"

queries/ql/highlights.scm
```diff
@@ -1,7 +1,6 @@
 [
   "as"
   "by"
-  "class"
```
ObserverOfTime, line 4: Add "newtype"

queries/qmljs/highlights.scm
```diff
@@ -138,7 +138,6 @@
   "property"
   "signal"
   "declare"
-  "enum"
   "export"
   "implements"
   "interface"
```
ObserverOfTime, line 143: Add "interface", "type"

queries/rbs/highlights.scm
```diff
@@ -16,7 +16,6 @@
 [
   "use"
   "as"
-  "class"
   "module"
   "interface"
   "type"
```
ObserverOfTime, line 21: Add "interface", "type"

queries/rust/highlights.scm
```diff
@@ -243,18 +243,21 @@
 
 [
   "default"
-  "enum"
   "impl"
   "let"
   "move"
-  "struct"
   "trait"
   "type"
```
ObserverOfTime, line 250: Add "trait", "type"

queries/scala/highlights.scm
```diff
@@ -170,8 +170,6 @@
 
 [
   "case"
-  "class"
-  "enum"
```
ObserverOfTime, line 174: Add "trait", "type"

queries/solidity/highlights.scm
```diff
@@ -145,8 +145,6 @@
   "interface"
```
ObserverOfTime, line 145: Add "interface", "contract"

queries/sourcepawn/highlights.scm
```diff
@@ -276,21 +276,24 @@
...
-  "funcenum"
   "functag"
   "get"
   "methodmap"
   "new"
   "property"
   "public"
   "set"
-  "struct"
   "typedef"
```
ObserverOfTime, line 286: Add "typedef"

queries/swift/highlights.scm
```diff
@@ -68,10 +68,7 @@
 
 [
   "typealias"
```
ObserverOfTime, line 70: Add "typealias"

ObserverOfTime (changes requested):
I think typedef/interface-like keywords should also use it.

There are also cases where the corresponding "end" keyword needs to be captured but we could just leave those to people familiar with the languages.

queries/fortran/highlights.scm
```diff
@@ -61,7 +61,6 @@
   "endselect"
   "endsubmodule"
   "endtype"
-  "enum"
```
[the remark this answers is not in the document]
ribru17: In fortran it looks like `class` is very similar to `type`, according to [this source](https://stackoverflow.com/questions/29652620/difference-between-type-and-class-in-fortran-2003). I think the "class" equivalent is `module` (with `endmodule`). Maybe we should apply it there also?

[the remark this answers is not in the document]
ObserverOfTime: Sure.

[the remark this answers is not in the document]
ObserverOfTime: Still missing "class"

queries/rbs/highlights.scm
```diff
@@ -28,6 +25,21 @@
...
+] @keyword.type
+
+(class_decl
+  "end" @keyword.type)
+
+(interface_decl
+  [
+    "interface"
+    "end"
+  ] @keyword.type)
```
ObserverOfTime, line 41:
```suggestion
  "end" @keyword.type)
```
"interface" is captured above.

queries/verilog/highlights.scm
```diff
@@ -267,8 +270,20 @@
 
 (struct_union) @type
```
ObserverOfTime, line 271: Can the identifier be captured?

queries/fortran/highlights.scm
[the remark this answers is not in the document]
ribru17: oops, thanks!

queries/verilog/highlights.scm
ribru17: To make sure I understand, you are talking about identifier of the struct/union which should be `@variable` (unless  there is a `typedef`, in which case it should be `@type` as it is now)?

ObserverOfTime: Something like this:
````
(struct_union
  [
    "struct"
    "union"
  ] @keyword.type)
  (identifier) @type)
```
(and "enum" should not be captured as `@type` at all)

ribru17: Even with no `typedef`? I think the current reasoning is that without `typedef`, the identifier is a variable name itself, with type `enum`. With `typedef`, the identifier is a *type* name that other variables will be typed as

ObserverOfTime: That's not the case in C (as far as I can tell) so I don't see why it should be in verilog.

ObserverOfTime (approved):

ribru17:
@amaanq thanks! also should the capture description in `CONTRIBUTING.md` be updated to say that it can denote namespace keywords as well as composite types?

clason:
If it does, it should!

PR #6502, opened 2024-04-24
feat(qml): add back `ecma` inheritance

ribru17:
Fixed after the grammar dependencies were updated. This time inheritance is given to the injections queries also.


queries/qmljs/highlights.scm
lucario387, line 1: Some of the queries below is also a duplicate of ecma queries so you probably have to remove them as well

lucario387 (approved):

PR #6504, opened 2024-04-24
feat(koto): Add koto support

irh:
This PR introduces [tree-sitter-koto](https://github.com/koto-lang/tree-sitter-koto).

### DIFF
--- a/queries/koto/highlights.scm
+++ b/queries/koto/highlights.scm
@@ -1,0 +1,131 @@
+[
+  "="
+  "+"
+  "-"
+  "*"
+  "/"
+  "%"
+  "+="
+  "-="
+  "*="
+  "/="
+  "%="
+  "=="
+  "!="
+  "<"
+  ">"
+  "<="
+  ">="
+  ".."
+  "..="
+] @operator
+
+[
+  "and"
+  "not"
+  "or"
+] @keyword.operator
+
+[
+  "return"
+  "yield"
+] @keyword.return
+
+[
+  "if"
+  "then"
+  "else"
+  "else if"
+  "match"
+  "switch"
+] @keyword.conditional
+
+[
+  (break)
+  (continue)
+  "for"
+  "in"
+  "loop"
+  "until"
+  "while"
+] @keyword.repeat
+
+[
+  "throw"
+  "try"
+  "catch"
+  "finally"
+] @keyword.exception
+
+[
+  "export"
+  "from"
+  "import"
+  "as"
+] @keyword.import
+
+[
+  "("
+  ")"
+  "["
+  "]"
+  "{"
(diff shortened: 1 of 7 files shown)


clason:
Please check `CONTRIBUTING.md` and ensure that all(!) captures align with the guidance therein. (You can _only_ use captures from that list.)

irh:
@clason Thanks, I've updated the captures now.

queries/koto/highlights.scm
```diff
@@ -0,0 +1,131 @@
...
+[
+  "and"
+  "not"
+  "or"
+] @keyword.operator
+
+[
+  "return"
+  "yield"
+] @keyword.return
```
ObserverOfTime, line 32:
```suggestion
"return" @keyword.return

"yield" @keyword.coroutine
```

queries/koto/highlights.scm
```diff
@@ -0,0 +1,131 @@
...
+  (false)
+] @boolean
+
+(comment) @comment @spell
+
+(debug) @keyword.debug
+
+(string) @string
+
+(fill_char) @string
```
ObserverOfTime, line 105:
```suggestion
(fill_char) @punctuation.delimiter
```
Maybe?

queries/koto/highlights.scm
irh: The docs in CONTRIBUTING.md say:

```
@keyword.return            ; keywords like `return` and `yield`
```

I don't have a strong opinion here, happy to go either way.

lucario387: yield depending on the language can be for return value of iterator or of coroutine. So which one is this `yield` meaning?

irh: Currently it's only for iterator output, but I think the coroutine description is appropriate so let's go with your suggestion. Thanks for the review!

ObserverOfTime: Oh yeah for iterators `@keyword.return` is fine.

queries/koto/highlights.scm
```diff
@@ -26,10 +26,9 @@
   "or"
 ] @keyword.operator
 
-[
-  "return"
-  "yield"
-] @keyword.return
+"return" @keyword.return
+
+"yield" @keyword.coroutine
```
lucario387, line 31: I think you can revert this change if `yield` is only used in iterators. If you know it can be a part of a coroutine call then please only narrow coroutine to that.

irh: Ok that makes sense. `yield` will be used in coroutines later on, but for now they're only used in custom iterators. When things change I'll update the capture used here.

ObserverOfTime (approved):

PR #6505, opened 2024-04-24
feat(asm): update queries from upstream

RubixDev:
adds support for TC assembly


clason:
If these were _just_ added, you'll have to update the lockfile.json entry to the latest commit so these nodes are available.

clason (approved):

PR #6526, opened 2024-04-28
make sass to be highlighted in vue files

morr:
The reason for this patch is that I found it to be impossible to find instructions in internet on how to enable `sass` syntax highlight for `vue` files.
I spent couple of hours both googling it and asking chart-gpt-4 how to do this. 

Since there is no `sass` parser for Treesitter lets use `css` or `scss` if such is available. It is not ideal solution, but it is so much better to have it this way than to have no syntax highlighting at all.

before patch
image

after patch
image

### DIFF
--- a/queries/vue/injections.scm
+++ b/queries/vue/injections.scm
@@ -20,7 +20,7 @@
         (attribute_value) @_scss)))
   (raw_text) @injection.content
   (#eq? @_lang "lang")
-  (#any-of? @_scss "less" "postcss")
+  (#any-of? @_scss "less" "postcss" "sass")
   (#set! injection.language "scss"))
 
 ; <script lang="js">


lucario387:
Can you use `:Inspect` to show the capture for those highlights?

queries/vue/injections.scm
```diff
@@ -8,7 +8,7 @@
         (attribute_value) @injection.language)))
   (raw_text) @injection.content)
   (#eq? @_lang "lang")
-  (#any-of? @injection.language "css" "scss"))
+  (#any-of? @injection.language "css" "scss" "sass"))
```
lucario387, line 11: No. Only one of the two places. Only the one below should be enough

lucario387 (dismissed):

morr:
@lucario387 
`Inspect`
image
image
image

`InspectTree`
image
`TSPlaygroundToggle`
image

queries/vue/injections.scm
morr: @lucario387 done

lucario387 (approved):

PR #6531, opened 2024-04-28
Fix compact constructor highlights for Java

BlueZeeKing:
This makes the record name in record constructors highlight properly because they are currently highlighted as `@variable` and not `@type`


ribru17 (approved):

ribru17:
Thanks!

PR #6547, opened 2024-05-01
fix(html_tags): stop indenting void elements

### LINKED ISSUE #6332: Cursor is indented after HTML void elements
Currently, inserting a newline after an HTML void element will indent the cursor:
```html
<img src="img.png">
    |
```
It doesn't, however, indent after void elements `meta` and `link` because those have a special case:
https://github.com/nvim-treesitter/nvim-treesitter/blob/36860295919fb18e57cd637b2a5f22e16f750c2f/queries/html_tags/indents.scm#L1-L14

To fix this, all of the void elements should be added to this list of "special" elements (and maybe rename that to void element instead?). Though to stay consistent with the way that `tree-sitter-html` parses void elements, we should use all of the following tags (deprecated or not) that functionally cannot have children:

https://github.com/tree-sitter/tree-sitter-html/blob/cc53b6878a003cf26d213025af49c78388cebe21/src/tag.h#L7-L29

I had considered opening a pull request, but I'm interested in how we should deal with the case-insensitivity of HTML. `` will be treated the same way as `` (and ``, etc.) by HTML parsers (and by extension `tree-sitter-html`), so I would think that we should handle that. I don't see a clear way to do that which wouldn't be very computationally expensive, _especially_ considering how many void-like elements that there are, so I'm wondering how I should proceed. Thanks.

LeviOP:
closes #6332

### DIFF
--- a/queries/html_tags/indents.scm
+++ b/queries/html_tags/indents.scm
@@ -1,17 +1,19 @@
 [
   ((element
     (start_tag
-      (tag_name) @_not_special))
-    (#not-any-of? @_not_special "meta" "link"))
+      (tag_name) @_not_void_element))
+    (#not-any-of? @_not_void_element
+      "area" "base" "basefont" "bgsound" "br" "col" "command" "embed" "frame" "hr" "image" "img"
+      "input" "isindex" "keygen" "link" "menuitem" "meta" "nextid" "param" "source" "track" "wbr"))
   (element
     (self_closing_tag))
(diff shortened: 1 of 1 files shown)


lucario387 (approved):

lucario387 (approved):

amaanq:
I was gonna make this easier upstream...

clason:
You snooze you lose :)

PR #6549, opened 2024-05-01
Add support for Inko

yorickpeterse:
This adds support for [Inko](https://inko-lang.org/) using [tree-sitter-inko](https://github.com/inko-lang/tree-sitter-inko/).

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -966,6 +966,14 @@ list.ini = {
   experimental = true,
 }
 
+list.inko = {
+  install_info = {
+    url = "https://github.com/inko-lang/tree-sitter-inko",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@yorickpeterse" },
+}
+
 list.ispc = {
   install_info = {
     url = "https://github.com/fab4100/tree-sitter-ispc",
--- a/queries/inko/folds.scm
+++ b/queries/inko/folds.scm
@@ -1,0 +1,17 @@
+[
+  (call)
+  (class)
+  (closure)
+  (define_case)
+  (define_constant)
+  (define_field)
+  (define_variable)
+  (external_function)
+  (if)
+  (import)+
+  (loop)
+  (match)
+  (method)
+  (trait)
+  (while)
+] @fold
--- a/queries/inko/highlights.scm
+++ b/queries/inko/highlights.scm
@@ -1,0 +1,195 @@
+; Brackets and operators
+[
+  "("
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
+] @punctuation.bracket
+
+[
+  ","
+  "->"
+  "."
+  ":"
+] @punctuation.delimiter
+
+[
+  "!="
+  "%"
+  "%="
+  "&"
+  "&="
+  "*"
+  "**"
+  "**="
+  "*="
+  "+"
+  "+="
+  "-"
+  "-="
+  "/"
+  "/="
+  "<"
+  "<<"
+  "<<="
+  "<="
+  "<="
+  "=="
+  ">"
+  ">="
+  ">="
+  ">>"
+  ">>="
+  ">>>"
+  ">>>="
+  "^"
+  "^="
+  "|"
+  "|="
+] @operator
+
+; Keywords
+[
+  "as"
+  "for"
+  "impl"
+  "let"
+  "mut"
+  "ref"
+  "uni"
+  "move"
+  "recover"
+] @keyword
+
+"fn" @keyword.function
+
+"import" @keyword.import
+
+[
+  "and"
+  "or"
+] @keyword.operator
+
+[
+  "class"
+  "trait"
+] @keyword.type
+
+[
+  "extern"
+  (modifier)
+  (visibility)
+] @keyword.modifier
+
+[
+  "loop"
+  "while"
+  (break)
+  (next)
+] @keyword.repeat
+
+"return" @keyword.return
+
+[
+  "throw"
+  "try"
+] @keyword.exception
+
+[
+  "case"
+  "else"
+  "if"
+  "match"
+] @keyword.conditional
+
+; Identifiers/variable references
+(identifier) @variable
+
+; Comments
+(line_comment) @comment @spell
+
+; Literals
+(self) @variable.builtin
+
+(nil) @constant.builtin
+
+[
+  (true)
+  (false)
+] @boolean
+
+(integer) @number
+
+(float) @number.float
+
+(string) @string
+
+(escape_sequence) @string.escape
+
+(interpolation
+  "${" @punctuation.special
+  "}" @punctuation.special) @none
+
+(constant) @constant
+
+; Patterns
+(integer_pattern) @number
+
+(string_pattern) @string
+
+(constant_pattern) @constant
+
+; Types
+(generic_type
+  name: _ @type)
+
+(type) @type
+
+; Imports
+(extern_import
+  path: _ @string)
+
+(import
+  (path
+    (identifier) @module))
+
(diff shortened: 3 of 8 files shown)


queries/inko/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+  "move"
+  "or"
+  "recover"
+  "return"
+  "throw"
+  "try"
+  "while"
+  (modifier)
+  (visibility)
+] @keyword
```
ObserverOfTime, line 62: Some of these should be more specific (e.g. `@keyword.conditional`). Check CONTRIBUTING.md.

queries/inko/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+
+; Comments
+(line_comment) @comment
```
ObserverOfTime, line 65:
```suggestion
(line_comment) @comment @spell
```

queries/inko/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+
+; Literals
+[
+  (nil)
+  (self)
+  (break)
+  (next)
+] @keyword
```
ObserverOfTime, line 73:
```suggestion
(nil) @constant.builtin

(self) @variable.builtin
```
`(break)` & `(next)` should be `@keyword.repeat` (above).

queries/inko/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+
+[
+  (true)
+  (false)
+] @boolean
+
+(integer) @number
+
+(float) @number
```
ObserverOfTime, line 82:
```suggestion
(float) @number.float
```

queries/inko/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+  name: _ @function)
+
+(external_function
+  name: _ @function)
+
+(argument
+  name: _ @variable)
+
+(named_argument
+  name: _ @variable)
```
ObserverOfTime, line 143:
```suggestion
(argument
  name: _ @variable.parameter)

(named_argument
  name: _ @variable.parameter)
```

queries/inko/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+
+(call
+  name: _ @function)
+
+(field) @property
```
ObserverOfTime, line 148:
```suggestion
(field) @variable.member
```

queries/inko/highlights.scm
```diff
@@ -0,0 +1,150 @@
...
+
+(identifier) @variable
```
ObserverOfTime, line 150: Move this above the rest so it doesn't overwrite them.

queries/inko/injections.scm
```diff
@@ -0,0 +1,2 @@
+(line_comment) @injection.content
+(#set! injection.language "comment")
```
ObserverOfTime, line 2:
```suggestion
((line_comment) @injection.content
  (#set! injection.language "comment"))
```

lua/nvim-treesitter/parsers.lua
```diff
@@ -966,6 +966,15 @@ list.ini = {
   experimental = true,
 }
 
+list.inko = {
+  install_info = {
+    url = "https://github.com/inko-lang/tree-sitter-inko",
+    files = { "src/parser.c" },
+    branch = "main",
```
ObserverOfTime, line 973:
```suggestion
```
Not needed.

ObserverOfTime (changes requested):

queries/inko/highlights.scm
yorickpeterse: @ObserverOfTime Identifiers are only used for variable references, for symbol names (e.g. `foo` in `fn foo ...`) the node type is `(name)`. As such, I think this query won't overwrite anything unintentionally?

queries/inko/highlights.scm
```diff
@@ -0,0 +1,169 @@
...
+  "^"
+  "=="
+  "!="
+] @operator
+
+; Keywords
+[
+  "as"
+  "case"
+  "for"
```
ObserverOfTime, line 36: "as" -> `@keyword.import`
"case" -> `@keyword.conditional`
"for" -> `@keyword.repeat`
(maybe)

queries/inko/highlights.scm
```diff
@@ -0,0 +1,169 @@
...
+  "if"
+  "match"
+] @keyword.conditional
+
+; Comments
+(line_comment) @comment @spell
+
+; Literals
+[
+  (nil)
```
ObserverOfTime, line 90: `(nil)` -> `@constant.builtin`

queries/inko/highlights.scm
yorickpeterse: @ObserverOfTime `as` is used for both type casts (`expr as SomeType`) and aliasing symbols in imports (`import some.module (SomeType as Whatever)`), hence I left it as a generic keyword.

`case` is used for pattern matching branches, so you're right in that that should be `keyword.conditional`, seems I forget that.

`for` is used for implementing traits, i.e. `impl SomeTrait for SomeClass { ... }`. There are no traditional for loops in Inko.

ObserverOfTime (approved):

queries/inko/folds.scm
```diff
@@ -0,0 +1,17 @@
...
+  (call)
+  (class)
+  (closure)
+  (define_case)
+  (define_constant)
+  (define_field)
+  (define_variable)
+  (external_function)
+  (if)
+  (import)
```
ribru17, line 11:
```suggestion
  (import)+
```

queries/inko/highlights.scm
```diff
@@ -0,0 +1,168 @@
...
+  ">="
+  "<<"
+  ">>"
+  ">>>"
+  "&"
+  "|"
+  "^"
+  "=="
+  "!="
+] @operator
```
ribru17, line 30: Can we get `"->"` here too? Also `","`, `"."`, and `":"` as `@punctuation.delimiter`

queries/inko/highlights.scm
ribru17: In my testing it seems that this query *does* override e.g. `@variable.parameter`

ribru17: Also for something like `fn main { ... }` `main` gets the regular variable highlight over the function one

queries/inko/highlights.scm
```diff
@@ -0,0 +1,168 @@
...
+] @keyword.operator
+
+[
+  "class"
+  "trait"
+] @keyword.type
+
+[
+  (modifier)
+  (visibility)
```
ribru17, line 61:
```suggestion
  (visibility)
  "extern"
```

queries/inko/highlights.scm
```diff
@@ -0,0 +1,168 @@
...
+
+; Types
+(generic_type
+  name: _ @type)
+
+(type) @type
+
+; Imports
+(extern_import
+  path: _ @string)
```
ribru17, line 127: I think it would also be nice to have 
```query
(import
  (path
    (identifier) @module))
```

queries/inko/highlights.scm
ribru17: Also looks like assignment operators like `"+="`, `"-="`, etc. are missing

yorickpeterse: @ribru17 `->` isn't an actual operator though, it's used to signal the return type of a function, i.e. `fn foo -> Int { ... }`, or does that also count as an operator?

ribru17: Ah, I see sorry. Maybe it would be better as `@punctuation.delimiter`

ribru17 (approved):

PR #6555, opened 2024-05-03
feat(ecma): inject template methods

DimensionalDot:
Added injection into tagged templates for method calls (e.g., `` client.sql`...` ``) as mentioned in #6550.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -6,7 +6,7 @@
     "revision": "d3dc807692e6bca671d4491b3bf5c67eeca8c016"
   },
   "angular": {
-    "revision": "b96a0d1605da3492f6474245098b6f0c503e596d"
+    "revision": "31182d43b062a350d4bd2449f2fc0d5654972be9"
   },
   "apex": {
     "revision": "c99ad4b16d112fea91745e3f1b769754239fdaba"
@@ -120,7 +120,7 @@
     "revision": "0229c0211dba909c5d45129ac784a3f4d49c243a"
   },
   "djot": {
-    "revision": "1aacaf472117fb855e2f80f744b7704ef4dc75dc"
+    "revision": "886601b67d1f4690173a4925c214343c30704d32"
   },
   "dockerfile": {
     "revision": "087daa20438a6cc01fa5e6fe6906d77c869d19fe"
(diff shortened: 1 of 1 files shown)


queries/ecma/injections.scm
```diff
@@ -9,8 +9,14 @@
 (call_expression
   function: [
     (await_expression
-      (identifier)) @injection.language
+      [
+        (identifier) @injection.language
+        (member_expression
+          property: (property_identifier) @injection.language)
```
lucario387, line 15: This kindof open a weird can of wormholes. What about `foo.bar.sql`, `foo.bar.baz.sql`, ....

I'd prefer to keep this in my own config, but not in nvim-treesitter.

Unless you can give clear reasonings why this won't lead to the can of worms above

DimensionalDot: The `foo.sql` pattern appears in one of the leading Javascript frameworks SQL SDKs (see [Vercel's PostgreSQL SDK](https://vercel.com/docs/storage/vercel-postgres/sdk#db)). This makes me feel that it is common enough to be covered here instead of in each user’s individual config (similar to how we already cover things like Styled Components via the `styled.div` injection).

On the other hand, I can’t think of a non user-constructed scenario (or any library) that would have a pattern like `foo.bar.sql`, `foo.bar.baz.sql`, `sql.bar`. This makes me feel that these are niche/rare enough to not be covered here but instead in a given user’s config.

While this doesn’t address why this wouldn’t open a can of worms, it does address why we might not feel compelled make any further changes related to it.

If you prefer the route of covering specific libraries (similar to Styled Components) let me know and I can amend this to have a (potentially separate) query only covering `sql` or even just `client.sql`.

lucario387: I will go for only covering for the most popular (like top 500 node packages).

Vercel's Postgres is popular enough so I'd give it a pass. Not sure about other cases though.

lucario387: Preferably writing a new query covering whatever that vercel postgres will be for + comments close to the new query

clason:
@DimensionalDot any updates?

clason:
never merge master into a branch; always rebase.

DimensionalDot:
@clason sorry, I just clicked the button to sync the fork on it's web page. Didn't realize. What should I do to remedy this?

clason:
if your local branch is still OK, sync your fork, `git rebase master`, and then `git push -f`.

DimensionalDot:
My local branch is still OK. So to double check I just synced my fork's master. Now I should locally do `git pull master` before rebasing and force pushing?

clason:
Yes, `git pull master` (but not this branch!).

clason:
If you force-push after closing, that breaks Github. You'll have to open a new PR now.

DimensionalDot:
I didn't manually close. I think that may have happened when I force pushed because my only two commits were the original and a revert, both of which had merge conflicts with master. I skipped them because there didn't seem to be a way to merge them that would make sense but now this happened. Sorry for all the issues I just caused. I'm new to this part of GitHub.

clason:
No worries.

PR #6565, opened 2024-05-03
fix(commonlisp): correctly highlight quotes

johannesrld:
the previous queries caused issues like highlighting `(quoting_lit)` and `(format_directive_type "'")` twice which made them really annoying to overwrite.

### DIFF
--- a/queries/commonlisp/highlights.scm
+++ b/queries/commonlisp/highlights.scm
@@ -105,24 +105,19 @@
   (#lua-match? @variable.builtin "^[*].+[*]$"))
 
 ; quote
-"'" @string.escape
-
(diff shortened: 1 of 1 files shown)


ribru17:
Would it work to only highlight anonymous nodes? With something like:
```lisp
(let ((tmp (gensym)))
    ``(lambda (,tmp ,,tmp ,',tmp) ()))
```
I get many duplicate highlights with either change. But keeping the anonymous nodes and then removing the other `@string.escape` and `@none` highlights gives a cleaner result (but I don't know much about lisp so this may be unacceptable due to some edge cases)

ribru17 (approved):

PR #6568, opened 2024-05-04
feat(c_sharp): update parser & queries

amaanq:


queries/c_sharp/highlights.scm
```diff
@@ -222,60 +296,57 @@
...
+  (preproc_define)
+  (preproc_undef)
+  (preproc_line)
+  (preproc_pragma)
+  (preproc_nullable)
+  (preproc_region)
+  (preproc_endregion)
+  (preproc_if)
+  (preproc_elif)
+  (preproc_else)
```
ribru17, line 328: I feel like the `#if` `#endif` etc. should be `@keyword.directive` similar to C

queries/c_sharp/highlights.scm
```diff
@@ -157,7 +216,7 @@
   (identifier) @type)
 
 ; Generic Types.
-(type_of_expression
+(typeof_expression
   (generic_name
     (identifier) @type))
```
ribru17, line 221: Could the `generic_name` be generalized or no? E.g. I see right now for 
```csharp
        protected override Task<QuickFixResponse> TranslateResponse(QuickFixResponse response, GotoFileRequest request)
        {
            return response.TranslateAsync(Workspace, request);
        }
```
the `Task` part of `Task<QuickFixResponse>` is highlighted as a variable instead of a type

queries/c_sharp/highlights.scm
```diff
@@ -366,9 +439,11 @@
   ] @punctuation.bracket)
 
 [
-  (this_expression)
-  (base_expression)
+  "assembly"
+  "module"
   "this"
+  "base"
```
ribru17, line 445: Right now this gets `@constructor` and `@variable.builtin`, not sure if this is intended

queries/c_sharp/highlights.scm
```diff
@@ -114,8 +140,7 @@
   (identifier) @type)
 
 (using_directive
```
ribru17, line 142: Can `(using_directive)` identifiers (and ones in `(qualified_name)`) be `@module` instead of `@type`? Also is it intended that namespace declarations like `namespace Foo.Bar.Baz`, only `Baz` is given just module while `Foo` and `Bar` are given `@module` overridden by `@type`?

queries/c_sharp/highlights.scm
```diff
@@ -449,9 +531,14 @@
   "partial"
   "sealed"
   "virtual"
+  "global"
+  (modifier)
 ] @keyword.modifier
 
-(parameter_modifier) @operator
+(modifier) @keyword.modifier
```
ribru17, line 538: This is covered above (right now some modifiers are highlighted 3 times); also `"override"` is highlighted as `@keyword` above but also covered in `(modifier)` so highlighted as a modifier later

queries/c_sharp/highlights.scm
```diff
@@ -82,6 +96,15 @@
...
 (null_literal) @constant.builtin
 
+(calling_convention
+  [
+    (identifier)
+    "Cdecl"
+    "Stdcall"
+    "Thiscall"
+    "Fastcall"
+  ] @constant.builtin)
```
ObserverOfTime, line 106: `@keyword.modifier` ?

queries/c_sharp/highlights.scm
```diff
@@ -357,6 +429,7 @@
   "}"
   "("
   ")"
+  (interpolation_brace)
```
ObserverOfTime, line 432: Interpolation braces should be `@punctuation.special`.

clason:
@amaanq can you make (or otherwise address) these changes? I want to unbreak the automatic lockfile updates sooner rather than later.

amaanq:
yeah, didn't forget about this, just making a couple more tweaks to the grammar & I'll address this later today

clason:
Feel free to merge when you're done. I have removed the review requirement for a reason.

clason:
@amaanq friendly ping.

queries/c_sharp/highlights.scm
amaanq: no, consider this case - `MyFunction<A,B>(1);` where MyFunction is the identifier of a generic name. It is not a type but rather a function

queries/c_sharp/highlights.scm
amaanq:
```cs
public DerivedClass() : base()
    {
    }
```

vs `base.foo` (similar to `this.foo`/`self.foo` in other languages)

but you bring up a good point, this should be lower in the list to override the catch-all base query

queries/c_sharp/highlights.scm
amaanq: I don't really think it's a modifier, it's an attribute/constant that determines what abi/calling convention the function uses (determines the memory layout/register usage)

clason: `@attribute.builtin` then? ;)

queries/c_sharp/highlights.scm
amaanq: using ones can't be determined without semantic info, so no point trying there - but the namespace ones could probably be all `@module`, it's just not easy due to how the children are structured but we could use has-ancestor here since there won't be many anyways

queries/c_sharp/highlights.scm
amaanq: but yeah it's annoying to have to account for all these spots with generic name

ribru17: ah i see that makes sense!

queries/c_sharp/highlights.scm
ribru17: ah man. maybe who cares then :laughing: I had hoped it would be an easy change

queries/c_sharp/highlights.scm
```diff
@@ -222,60 +322,63 @@
...
+[
+  "#define"
+  "#undef"
+] @keyword.directive.define
 
-(pragma_directive
-  (preproc_string_literal) @string) @constant.macro
+[
+  "#if"
+  "#elif"
```
ribru17, line 346: Should `#endif` be here? for me looks like it just gets no highlight

queries/c_sharp/highlights.scm
```diff
@@ -46,11 +64,22 @@
...
+  name: (identifier) @module)
+
+(file_scoped_namespace_declaration
+  name: (identifier) @module)
+
+(qualified_name
+  (identifier) @module
+  (#not-has-ancestor? @module method_declaration)
+  (#not-has-ancestor? @module record_declaration)
+  (#has-ancestor? @module namespace_declaration))
```
ribru17, line 76: I think this can be combined with the below one, like `(#has-ancestor? @module namespace_declaration file_scoped_namespace_declaration))`? Also would it make sense to add `using_directive` to this list? (And then also I think we would have to add 
```scheme
(using_directive
  (identifier) @module)
```
but towards the end: I tested adding it next to this query and it wasn't applied with high enough priority)

amaanq: > I think this can be combined with the below one, like (#has-ancestor? @module namespace_declaration file_scoped_namespace_declaration))

Wasn't sure if that'd work - nice!

As for using directives, again you cannot assume it's a namespace - it probably is oftentimes but other times it could be just types

ribru17: > As for using directives, again you cannot assume it's a namespace - it probably is oftentimes but other times it could be just types

ahhh ok sry i get it now :sweat_smile:

ribru17 (approved):

PR #6572, opened 2024-05-04
fix(inko): update parser revision

yorickpeterse:
This revision includes a fix for parsing certain string literals.


clason:
Thanks, but we automate these updates (with a daily CI action). You only need to worry about it if an update breaks queries (node is renamed or removed); then a timely PR bumping lockfile _and_ queries is much appreciated.

PR #6574, opened 2024-05-05
fix(dockerfile): fix shell_command bash injection

### LINKED ISSUE #6530: Dockerfile bash injection is too broad

The bash injection in `Dockerfile`s is not isolated per `RUN` statement

```dockerfile
FROM scratch
RUN echo a
RUN echo b
```

```lisp
(source_file ; [0, 0] - [3, 0] dockerfile
  (program ; [1, 4] - [2, 10] bash
    (command ; [1, 4] - [2, 10] bash
      name: (command_name ; [1, 4] - [1, 8] bash
        (word)) ; [1, 4] - [1, 8] bash
      argument: (word) ; [1, 9] - [2, 8] bash
      argument: (word))) ; [2, 9] - [2, 10] bash
  (from_instruction ; [0, 0] - [0, 12] dockerfile
    "FROM" ; [0, 0] - [0, 4] dockerfile
    (image_spec ; [0, 5] - [0, 12] dockerfile
      name: (image_name))) ; [0, 5] - [0, 12] dockerfile
  "\n" ; [0, 12] - [1, 0] dockerfile
  (run_instruction ; [1, 0] - [1, 10] dockerfile
    "RUN" ; [1, 0] - [1, 3] dockerfile
    (shell_command ; [1, 4] - [1, 10] dockerfile
      (shell_fragment))) ; [1, 4] - [1, 10] dockerfile
  "\n" ; [1, 10] - [2, 0] dockerfile
  (run_instruction ; [2, 0] - [2, 10] dockerfile
    "RUN" ; [2, 0] - [2, 3] dockerfile
    (shell_command ; [2, 4] - [2, 10] dockerfile
      (shell_fragment))) ; [2, 4] - [2, 10] dockerfile
  "\n") ; [2, 10] - [3, 0] dockerfile
```

### Example screenshot

image

Colorscheme: ellisonleao/gruvbox.nvim

`:Inspect` for `RUN` on the last line:

image

### Expected behavior

I'd expect there to be two bash injecections, one for each `RUN` statement

```markdown

Installation ~
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
- OK `node` found v21.7.3 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
  Version: cc (GCC) 13.2.1 20240417

OS Info:
{
  machine = "x86_64",
  release = "6.6.28-1-lts",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Wed, 17 Apr 2024 10:11:09 +0000"
} ~

- ERROR properties(highlights): /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:112: Failed to load parser for language 'properties': uv_dlopen: no error
  properties(highlights) is concatenated from the following files:
  | [ERROR]:"/home/jannis/.local/share/nvim/plugged/nvim-treesitter/queries/properties/highlights.scm", failed to load: /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:112: Failed to load parser for language 'properties': uv_dlopen: no error
- ERROR properties(locals): /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:112: Failed to load parser for language 'properties': uv_dlopen: no error
  properties(locals) is concatenated from the following files:
  | [ERROR]:"/home/jannis/.local/share/nvim/plugged/nvim-treesitter/queries/properties/locals.scm", failed to load: /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:112: Failed to load parser for language 'properties': uv_dlopen: no error
- ERROR properties(injections): /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:112: Failed to load parser for language 'properties': uv_dlopen: no error
  properties(injections) is concatenated from the following files:
  | [ERROR]:"/home/jannis/.local/share/nvim/plugged/nvim-treesitter/queries/properties/injections.scm", failed to load: /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:112: Failed to load parser for language 'properties': uv_dlopen: no error
```

```text
NVIM v0.10.0-dev-3032+g61063653b0
Build type: RelWithDebInfo
LuaJIT 2.1.1713773202
```

### Additional context

Unrelated to this issue: The `Output of :checkhealth nvim-treesitter` section makes it quite hard to search for issues. Because of that section, every highlighting issue mentions a lot of unrelated languages, making it really difficult to narrow down if an issue has been reported already. I reduced my output to the languages relevant for this issue.

faergeek:
Fixes #6530

### DIFF
--- a/queries/dockerfile/injections.scm
+++ b/queries/dockerfile/injections.scm
@@ -1,10 +1,9 @@
 ((comment) @injection.content
   (#set! injection.language "comment"))
 
-((shell_command
-  (shell_fragment) @injection.content)
+((shell_command) @injection.content
   (#set! injection.language "bash")
-  (#set! injection.combined))
+  (#set! injection.include-children))
 
 ((run_instruction
   (heredoc_block) @injection.content)


queries/dockerfile/injections.scm
```diff
@@ -1,10 +1,9 @@
 ((comment) @injection.content
   (#set! injection.language "comment"))
 
-((shell_command
-  (shell_fragment) @injection.content)
+((shell_command) @injection.content
```
ObserverOfTime, line 4: There's a reason it was implemented the way it was.

This includes the dockerfile heredoc in the bash injection.
```dockerfile
RUN <<EOF
echo 'hello world'
EOF

**Review by ObserverOfTime — CHANGES_REQUESTED** [review]

File: queries/dockerfile/injections.scm (hunk 1 above)
Comment (faergeek) [reply queries/dockerfile/injections.scm:4]: But `heredoc_marker` is also under `shell_fragment` anyway, right now at least. So it doesn't seem it will change much in that regard. Or am I missing something?

On the other hand highlighting individual `shell_fragment` nodes results in a broken highlighting for snippets like this one:

```dockerfile
RUN ls \
  --tree
```

Here each line results in a separate `shell_fragment`, which results in `--tree` being treated as a separate `command` instead of being an `argument`. Not a big deal probably, but not exactly correct still.

faergeek: The tree with the change from PR:

```scm
(run_instruction) ; [1:1 - 2:8]
 (shell_command) ; [1:5 - 2:8]
  (command) ; [1:5 - 2:8]
   name: (command_name) ; [1:5 - 6]
    (word) ; [1:5 - 6]
   argument: (word) ; [2:3 - 8]
  (shell_fragment) ; [1:5 - 7]
  (line_continuation) ; [1:8 - 2:0]
  (shell_fragment) ; [2:3 - 8]
```

The tree with original `shell_fragment` matching:

```scm
(run_instruction) ; [1:1 - 2:8]
 (shell_command) ; [1:5 - 2:8]
  (shell_fragment) ; [1:5 - 7]
   (command) ; [1:5 - 6]
    name: (command_name) ; [1:5 - 6]
     (word) ; [1:5 - 6]
  (line_continuation) ; [1:8 - 2:0]
  (shell_fragment) ; [2:3 - 8]
   (command) ; [2:3 - 8]
    name: (command_name) ; [2:3 - 8]
     (word) ; [2:3 - 8]
```

ObserverOfTime: Your version will work if `heredoc_marker` is moved.

faergeek: >Your version will work if heredoc_marker is moved.

Sorry, I don't get if it's for or against the change. Should I revert that part anyway?

ObserverOfTime: Against, until the parser is fixed.

ObserverOfTime: This does not change anything.

faergeek: The start of a heredoc is still captured by it, by the way, so that didn't change

faergeek: But that injection breaks highlighting at the moment and I don't see why using `injection.combined` is necessary here. Treating all of the nodes as a big `bash` script doesn't seem to make sense here. They're meant to be distinct separate scripts. As I understand, `injection.combined` is meant for languages like PHP, where you can stop PHP at basically any point (like at the start of the loop), put a bunch of HTML and the get back to PHP. For dockerfile+bash that doesn't seem to make sense.

iyzana:
This will also break Dockerfiles that contain the escape directive, for example

```
# escape=%
FROM scratch
RUN ls %
  --tree
```

Here the content below shell_script can't be treated as one script. It is interspersed by the Dockerfile line_continuation.

None of the existing injection types are completely correct here. include-children is wrong, because it includes the Dockerfile line_continuations in the script. And combined is wrong, because it treats all RUN statements as one large script.

But I think treating it as one large script is the less bad alternative.

faergeek:
>This will also break Dockerfiles that contain the escape directive, for example

Great point :+1: Though this also doesn't seem to be parsed correctly by dockerfile parser alone, without any injections. From what I see it just seems to stop parsing after `%`.

faergeek:
And it's a known issue https://github.com/camdencheek/tree-sitter-dockerfile/issues/2

queries/dockerfile/injections.scm
ObserverOfTime: > But that injection breaks highlighting at the moment

So does the alternative.

> Treating all of the nodes as a big `bash` script doesn't seem to make sense here. They're meant to be distinct separate scripts.

This has multiple nodes but is a single script:

```dockerfile
RUN apt-get update && \
    apt-get install foo
```

---

The **only** solution here is to fix the parser first and *then* use your injection.

faergeek: >So does the alternative.

Not sure what you mean. Can you show an example where an implementation in this PR breaks the highlighting?

>This has multiple nodes but is a single script:

In the case you're showing it makes sense. But multiple `RUN` statements should not be treated as a single giant script.

ObserverOfTime: I did. In my very first comment.

faergeek:
Just read a bit more about docker heredocs (never used them myself). They can be used to supply the whole script, but also they can be in the middle of a command like in normal shell scripts. Which means it's not quite right to always inject bash in heredocs, but we have no way to separate those cases. With heredocs in the middle of a command it would be ok to use `include-children` it seems, but on a `shell_fragment`, not `shell_command`.

clason:
So what is the resolution here? Is there something actionable remaining?

faergeek:
Seems like there's nothing actionable at this stage. This upstream PR would help making it all easier, but there's no response so far and I have no idea what else can be done without it:

https://github.com/camdencheek/tree-sitter-dockerfile/pull/52

geigerzaehler:
@faergeek, @ObserverOfTime Would you consider re-opening this PR? I know that this fix breaks with heredocs and custom line breaks, but I think it is still an improvement for most users.

Without this fix, everything in a Dockerfile between two RUN statements will have messed up highlighting and this situation is very common. On the other hand, with this fix, only `<<EOF` in `RUN <<EOF \ ...` will be parsed and highlighted incorrectly. The rest of the shell script inside the heredoc will be highlighted correctly.

I think this PR provides a good trade-off between improving the situation for a broad and common case while breaking highlighting in a limited way in less common cases.

faergeek:
@geigerzaehler you can just take the file with the changes made by this PR and put it in `~/.config/nvim/queries/dockerfile/injections.scm` if you want. That's what I do myself at the moment.

But a proper solution would be to merge a tree-sitter-dockerfile PR I mentioned above.

PR #6576, opened 2024-05-05
feat(c): Inject printf grammar in concatenated strings

ValdezFOmar:


ribru17:
Would you mind showing an example of when this injection would be applied?

amaanq (approved):

PR #6577, opened 2024-05-05
fix(haskell): update parser, queries and tests for rewritten parser

mrcjkb:
This PR updates the queries and tests for the rewritten Haskell parser grammar (https://github.com/tree-sitter/tree-sitter-haskell/pull/120).

### Notes:

- I based these on the updated queries in tree-sitter-haskell (which were based on the old ones in this repo), with some tweaks and fixes (using the test file as a reference). 
I think a lot of the queries can be simplified/rewritten to be more concise and readable. If you prefer, I can mark this as draft and update them in this PR later, or I could do it in a follow-up.
- It would be nice to have some time to update the queries in other plugins (neotest-haskell, iswap.nvim, vim-matchup, ...). But if I understand correctly, the parser lockfile is updated automatically anyway?

### DIFF
--- a/runtime/queries/haskell/highlights.scm
+++ b/runtime/queries/haskell/highlights.scm
@@ -4,66 +4,37 @@
 ; and don't override destructured parameters
 (variable) @variable
 
-(pat_wildcard) @variable
+(pattern/wildcard) @variable
 
-(function
+(decl/function
   patterns: (patterns
     (_) @variable.parameter))
 
-(exp_lambda
+(expression/lambda
   (_)+ @variable.parameter
   "->")
 
-(function
-  infix: (infix
-    lhs: (_) @variable.parameter))
-
-(function
-  infix: (infix
-    rhs: (_) @variable.parameter))
+(decl/function
+  (infix
+    (pattern) @variable.parameter))
 
 ; ----------------------------------------------------------------------------
 ; Literals and comments
 (integer) @number
 
-(exp_negation) @number
+(negation) @number
 
-(exp_literal
+(expression/literal
   (float)) @number.float
 
 (char) @character
 
 (string) @string
 
-(con_unit) @string.special.symbol ; unit, as in ()
-
 (comment) @comment
 
-; FIXME: The below documentation comment queries are inefficient
-; and need to be anchored, using something like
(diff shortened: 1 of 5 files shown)


clason:
> But if I understand correctly, the parser lockfile is updated automatically anyway?

Not if it breaks the queries. Syncing downstream plugins is a big issue that we don't have a good solution for anyway (except pushing for versioned releases, which should slow down the pace and more clearly signal breakage), together with raising awareness in downstream maintainers and users.

Not sure what's going on; from what I can see it should work. Let me test.

runtime/queries/haskell/highlights.scm
```diff
@@ -4,66 +4,37 @@
 ; and don't override destructured parameters
 (variable) @variable
 
-(pat_wildcard) @variable
+(pattern/wildcard) @variable
```
clason, line 7: I don't think Neovim can deal with these; it's the first time I've seen nodes using slashes. I would strongly recommend sticking to word-like characters.

clason: (Protip: `:TSUpdate` the parser, then open the query in Neovim; the query linter will tell you exactly what it doesn't like.)

mrcjkb: Hmm, interesting. 
I get the lint warnings, but highlighting seems to be working fine.
Those are `<super>/<node>` queries. 
I'll see if I can find an alternative in a few days. 

(cc @tek)

clason: Interesting. If it's just the tests (and query linter) that are affected, then maybe we can fix this. Still, sticking to the general snake_case convention would be preferable.

clason:
Oh, and please do not make your PR agains `main` -- the default branch is still `master`.

runtime/queries/haskell/highlights.scm
clason: (It doesn't error, but I don't think it's working correctly, either.)

clason:
Yeah, could be that only the query linter needs changes; the failures here could very likely be from targeting the wrong branch (which no longer looks at `lockfile.json`).

Could you make a new PR against `master`? Hopefully that will just pass.

PR #6581, opened 2024-05-05
feat(lalrpop): highlight touchups, folds

ribru17:
Adds a lot of missing highlights that weren't covered by rust injections. Reference file: https://github.com/vosen/ZLUDA/blob/bdc652f9ebcac9a79849eeee84a391a4ac107913/ptx/src/ptx.lalrpop

### DIFF
--- a/queries/lalrpop/highlights.scm
+++ b/queries/lalrpop/highlights.scm
@@ -1,3 +1,5 @@
+(comment) @comment @spell
+
 "grammar" @keyword
 
 [
@@ -8,6 +10,7 @@
 [
   "pub"
   "extern"
+  (mut)
 ] @keyword.modifier
 
 [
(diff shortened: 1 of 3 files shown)


queries/lalrpop/highlights.scm
```diff
@@ -46,6 +51,31 @@
...
+
+(parametrized_type
+  [
+    "<"
+    ">"
+  ] @punctuation.bracket)
+
+(nonterminal_name
+  [
+    "<"
```
amaanq, line 71: seems like `<` and `>` aren't used as an operator anywhere, might as well just have a generic capture for them instead of being specific 5x

```
[
  "<"
  ">"
] @punctuation.bracket
```

amaanq (approved):

PR #6580, opened 2024-05-05
fix(haskell): update queries and tests for rewritten parser

mrcjkb:
#6577, but based on `master`, as requested in https://github.com/nvim-treesitter/nvim-treesitter/pull/6577#issuecomment-2094883010

PR #6582, opened 2024-05-06
fix(format): fix supertypes formatting

lucario387:
Fix the issues with formatting raised in #6580. I'm still wondering whether the parser should adapt to add a new node for supertypes, but for the mean time this will help with that.


clason:
That's possible (and probably a good idea). Would you know how?

lucario387:
Superseded by #6586

PR #6586, opened 2024-05-06
fix(query): update to support supertypes

clason:


clason:
@lucario387 supersedes #6582

PR #6587, opened 2024-05-06
fix(java): add highest priority to "var" keyword

idelice:
Currently, the `var` keyword highlight will get changed depending of the right side of the equal sign, but in fact, it shouldn't do that because `var` is basically to declare a variable _without_ specifying its data type - therefor, its priority should be higher than _any_ for a consistent highlight.

### Examples

**Before this PR's changes**
Screenshot 2024-05-06 at 15 28 43

**After this PR's changes**
Screenshot 2024-05-06 at 15 28 24

### DIFF
--- a/queries/java/highlights.scm
+++ b/queries/java/highlights.scm
@@ -93,7 +93,8 @@
 (type_identifier) @type
 
 ((type_identifier) @type.builtin
-  (#eq? @type.builtin "var"))
+  (#eq? @type.builtin "var")
+  (#set! "priority" 155))
 
 ((method_invocation
   object: (identifier) @type)


idelice:
> It seems to have the same highlight for me. Would you mind running `:Inspect` over the keyword? If this is caused by LSP highlights then I think it would be best not to mess with it

It's because you are not having different return types for `var` declaration.

Use a `String` and `int`of return types to see the difference - just like I've attached in the description.

idelice:
Screenshot 2024-05-06 at 20 35 55


Screenshot 2024-05-06 at 20 36 10

idelice:
Here's an example from Intellij (without any custom highlighting!):

Screenshot 2024-05-06 at 20 39 39

idelice:
And here's one from VScode (without any custom highlighting!):

ribru17:
> > It seems to have the same highlight for me. Would you mind running `:Inspect` over the keyword? If this is caused by LSP highlights then I think it would be best not to mess with it
> 
> It's because you are not having different return types for `var` declaration.
> 
> Use a `String` and `int`of return types to see the difference - just like I've attached in the description.

I meant my `var` is the same even with different value types on the right. As you showed in your screenshot the discrepancy is due to LSP semantic highlighting, not because of something on nvim-treesitter's end. When possible, we want to avoid using priority hacking to override LSP highlights

clason:
This.

idelice:
> > > It seems to have the same highlight for me. Would you mind running `:Inspect` over the keyword? If this is caused by LSP highlights then I think it would be best not to mess with it
> 
> > 
> 
> > It's because you are not having different return types for `var` declaration.
> 
> > 
> 
> > Use a `String` and `int`of return types to see the difference - just like I've attached in the description.
> 
> 
> 
> I meant my `var` is the same even with different value types on the right. As you showed in your screenshot the discrepancy is due to LSP semantic highlighting, not because of something on nvim-treesitter's end. When possible, we want to avoid using priority hacking to override LSP highlights 

I don't understand how yours can be different than mine.

Are you simply saying that my config is responsible for the different highlighting?

clason:
Yes. You have an LSP that is adding semantic highlighting that overrides tree-sitter highlighting.

idelice:
> Yes. You have an LSP that is adding semantic highlighting that overrides tree-sitter highlighting.

Hmm.. I can't remember ever touching any LSP semantic highlightings so certanly doubt it, but will check tomorrow and get back.

ribru17:
r

> Hmm.. I can't remember ever touching any LSP semantic highlightings so certainly doubt it, but will check tomorrow and get back.

The LSP adds it for you, you don't need to have touched it. You can, if you want, just disable the `@lsp.type.type.java` group (which is making it purple for you)

idelice:
> @lsp.type.type.java

I tried `vim.api.nvim_set_hl(0, "@lsp.type.type.java", {})` with no success. The color won't change.

idelice:
> It's _literally_ right there in your `:Inspect` output.

Doesn't necessarily mean that _I_ added the custom highlight. And I just looked through my config to see if I did and can confirm I haven't

clason:
That's irrelevant since it's enabled by default (by Neovim).

idelice:
The highlight color - purple -  was added by the theme I'm using and I've just overridden it. 

Now, the keyword `var` is the same color regardless of its datatype.

idelice:
> That's irrelevant since it's enabled by default (by Neovim).

Of course it's not irrelevant. I'm simply trying to say that _I_ haven't touched any LSP highlighting inside my personal configurations. 

But it's fixed now

PR #6590, opened 2024-05-07
fix(ecma): template string injections

zweimach:
This PR fixes a regression that used to work before #6550.

| Before (#6550) | After (This PR) |
| ------ | ----- |
| image | image |


ribru17 (approved):

ribru17:
Thank you!

PR #6594, opened 2024-05-07
feat(query): pattern injections for new predicates

ribru17:
Gives `regex` and `luap` injections to the new 0.10 predicates.

Sorry, this should have been done before https://github.com/neovim/neovim/commit/3d4eb9d544cbbe39544586890b5de83a48de3680 :grimacing:


clason:
I drop these injections when upstreaming anyway.

PR #6600, opened 2024-05-08
fix(inko): indent call nodes

yorickpeterse:
This ensures that when you add a new call line in the middle of a call chain, the line is indented like so:

    foo
      .bar
      .new_line_here
      .baz

Instead of it being indented like so:

    foo
      .bar
    .new_line_here
      .baz

### DIFF
--- a/queries/inko/indents.scm
+++ b/queries/inko/indents.scm
@@ -7,6 +7,7 @@
   (binary)
   (block)
   (bounds)
+  (call)
   (cast)
   (class)
   (class_pattern)


yorickpeterse:
Note that indentation is still funky when you try to add a new line at the end _and_ parentheses are used for the existing line, i.e.

```
foo
  .bar(10)
  | <- new line here
```

Results in this:

```
foo
  .bar(10)
.baz
```

Instead of this:

```
foo
  .bar(10)
  .baz
```

This isn't specific to Inko however and affects other languages as well (e.g. Java and Javascript). The core issue there is that at the end of the expression, the cursor isn't part of the `call` node (or whatever the parser uses), so it doesn't know the new line needs to be indented, nor does there seem to be a way to indent a newly typed line _after_ you finish typing it.

lucario387 (approved):
Wait why is it not requiring approval for merge anymore :D 
I'll trust the maintainer of the parser to know what they're doing

clason:
Because I removed that stupid restriction; all it did was show I don't trust the other maintainers. (Contributors can't approve anyway, so a single maintainer can just approve-and-merge -- pointless extra work -- so it only affects PRs by maintainers.)

PR #6619, opened 2024-05-12
feat(diff): more comprehensive highlights

ribru17:
More highlights for complicated diffs. Reference file:

```diff
diff --git a/CONTRIBUTING.md b/CONTRIBUTING_HERE.md
similarity index 100%
rename from CONTRIBUTING.md
rename to CONTRIBUTING_HERE.md
diff --git a/LICENSE b/LICENSE
old mode 100644
new mode 100755
diff --git a/Makefile b/Makefile
deleted file mode 100644
index 338c7546..00000000
--- a/Makefile
+++ /dev/null
@@ -1,7 +0,0 @@
-# https://github.com/luarocks/luarocks/wiki/Creating-a-Makefile-that-plays-nice-with-LuaRocks
-build:
-	echo "Do nothing"
-
-install:
-	mkdir -p $(INST_LUADIR)
-	cp -r lua/* $(INST_LUADIR)
diff --git a/hi.txt b/hi.txt
new file mode 100644
index 00000000..45b983be
--- /dev/null
+++ b/hi.txt
@@ -0,0 +1 @@
+hi
diff --git a/queries/rust/highlights.scm b/queries/rust/highlights.scm
index c392c030..1609a81a 100644
--- a/queries/rust/highlights.scm
+++ b/queries/rust/highlights.scm
@@ -29,7 +29,10 @@
 (mod_item
   name: (identifier) @module)

-(self) @variable.builtin
+[
+  (self)
+  "_"
+] @variable.builtin

 (label
   [
@@ -45,7 +48,10 @@
   (identifier) @function)

 (parameter
-  (identifier) @variable.parameter)
+  [
+    (identifier)
+    "_"
+  ] @variable.parameter)

 (closure_parameters
   (_) @variable.parameter)
```

### DIFF
--- a/queries/diff/highlights.scm
+++ b/queries/diff/highlights.scm
@@ -12,4 +12,35 @@
 
 (location) @attribute
 
-(command) @function
+(command
+  "diff" @function
+  (argument) @variable.parameter)
+
+(filename) @string.special.path
+
+(mode) @number
+
+[
+  ".."
+  "+"
+  "++"
+  "+++"
+  "++++"
+  "-"
+  "--"
+  "---"
+  "----"
+] @punctuation.special
+
+[
+  (binary_change)
+  (similarity)
+  (file_change)
+] @label
+
+(index
+  "index" @keyword)
+
+(similarity
+  (score) @number
+  "%" @number)
--- a/lockfile.json
+++ b/lockfile.json
@@ -111,7 +111,7 @@
     "revision": "affb6ee38d629c9296749767ab832d69bb0d9ea8"
   },
   "diff": {
-    "revision": "710e881127512bda8157096e26c1f3e91702823a"
+    "revision": "629676fc3919606964231b2c7b9677d6998a2cb4"
   },
   "disassembly": {
     "revision": "0229c0211dba909c5d45129ac784a3f4d49c243a"


queries/diff/highlights.scm
```diff
@@ -13,3 +13,30 @@
 (location) @attribute
 
 (command) @function
+
+(context) @none
```
clason, line 17: I don't think that does _anything_. Can you try omitting it (or giving it a proper capture)?

ribru17: For sure; it indeed does nothing :laughing:

clason: (I know we have a bunch of uses still in our queries; they predate the Neovim highlighter, and some are due to my removal of `@text` -- I didn't want to delete the captures entirely, so I left them as placeholders.)

ribru17: Gotcha, that makes sense :+1:

clason: (Doesn't mean a cleanup -- where possible -- wouldn't be welcome ;))

ribru17: Any excuse to procrastinate my other work!

amaanq (approved):

ribru17:
At least for this repo, we only care about capture correctness. The highlights of the captures themselves are up to the individual colorscheme authors. For what it's worth, I tested out that colorscheme and for me it seems that the punctuation gets delimiter highlights, which are orange and not grey. Maybe try updating? I think the orange looks alright. But if you still think it needs a change feel free to make a PR over there to differentiate `@punctuation.special` (or even just `@punctuation.special.diff`)

ambroisie:
Yes, I was missing a patch which made them orange.

> But if you still think it needs a change feel free to make a PR over there to differentiate @punctuation.special (or even just @punctuation.special.diff)

My big issue is that the colorscheme can't make the difference between the different `special` operators, they all get the same color... Whereas the previous version of the query had different groups for `+` and `-`.

ribru17:
That's fair, colorschemes can't change queries (kinda...); but I don't think they used to get different groups before, they just both inherited their respective `@diff.minus` and `@diff.plus`, which they still do now. Now they just also get the special punctuation highlight to differentiate them from normal diffed text. That said, if you want them to be extra different, you can add something like the following to your `~/.config/nvim/queries/diff/highlights.scm`:
```query
; extends
"+" @myhighlightforplus

"-" @myhighlightforminus
```
and then highlight your groups accordingly (or just link the queries to something else you know works for you)

ambroisie:
> Now they just also get the special punctuation highlight to differentiate them from normal diffed text.

Yeah, my point is it looks like the punctuation one is getting priority over `diff.minus` and `diff.plus` :-/

lucario387:
https://github.com/the-mikedavis/tree-sitter-diff/blob/629676fc3919606964231b2c7b9677d6998a2cb4/grammar.js#L89-L102

It seems like the queries should have taken these into account then :?

Also, I believe we all have lived through our early programming days with the `+` and `-` of a diff patch with the same highlighting as the line, so differentiating from what has been the norm will be weird.

PR #6620, opened 2024-05-12
feat(zig): misc improvements

ribru17:
Moves some keyword highlights around, highlights parameter types and `=>`, and specifies only the `|` brackets in a payload receive the bracket highlight

### DIFF
--- a/queries/zig/highlights.scm
+++ b/queries/zig/highlights.scm
@@ -100,6 +100,8 @@ field_constant: (IDENTIFIER) @constant
   "test"
   "opaque"
   "error"
+  "const"
+  "var"
 ] @keyword
 
 [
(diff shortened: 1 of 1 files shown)


amaanq (approved):

clason:
Unfortunately, conflicts with your export PR. Rebase please?

diocletiann:
@ribru17 thanks for this. Would it be possible to fix `@function` so one lowercase word is also tagged correctly?

ribru17:
@diocletiann Sorry, not 100% sure what you mean by this, but if you are referring to highlighting something like `myFuncCall` in `myFuncCall(foo)`, I think I am unable right now. I tried a bit but it just seems that the parser is too all over the place...

diocletiann:
> @diocletiann Sorry, not 100% sure what you mean by this, but if you are referring to highlighting something like `myFuncCall` in `myFuncCall(foo)`, I think I am unable right now. I tried a bit but it just seems that the parser is too all over the place...

referring to something like `print("hello world", .{}` where the function call is one, lowercase word. This gets tagged `@variable` currently and requires an lsp semantic token.

PR #6621, opened 2024-05-13
feat: standardize and document export keywords

ribru17:
Many export keywords are captured as `@keyword.import`. This commit makes it so they are all captured like that, and mentions it in the documentation.


amaanq:
nice cleanup

amaanq (approved):

PR #6630, opened 2024-05-13
feat(corn): more highlights, comment injections

ribru17:
Reference file:
```corn
let {
    $base = {
        greeting = 12.0e+2
        subject.hi.adsf = "world"
    }
} in {
  ..$base
  subject = "github"
  hey = null
}
// vim: ft=corn
// TODO: corn
```

### DIFF
--- a/queries/corn/highlights.scm
+++ b/queries/corn/highlights.scm
@@ -1,6 +1,7 @@
-"let" @keyword
-
-"in" @keyword
+[
+  "let"
+  "in"
+] @keyword
 
 [
   "{"
@@ -11,7 +12,14 @@
 
 "." @punctuation.delimiter
 
+[
+  ".."
+  "="
+] @operator
+
 (input) @constant
+
+(null) @constant.builtin
 
 (comment) @comment @spell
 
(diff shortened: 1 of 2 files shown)


queries/corn/highlights.scm
```diff
@@ -1,6 +1,8 @@
-"let" @keyword
-
-"in" @keyword
+[
+  "let"
+  "in"
+  (null)
```
amaanq, line 4: not `@constant.builtin`?

queries/corn/highlights.scm
```diff
@@ -9,7 +11,12 @@
   "]"
 ] @punctuation.bracket
 
-"." @punctuation.delimiter
+[
+  "."
+  "="
```
amaanq, line 16: is `=` not an operator?

ribru17: in this language i felt like it was more of a delimiter, like `:` in js objects, but i think treating it as an operator makes sense too. also i only found one other place where `=` is treated as `@punctuation.delimiter`

amaanq (approved):

PR #6653, opened 2024-05-18
fix: do not use flatten from neovim std

hoofcushion:
It deprecated again.
image
```
$ nvim --version
NVIM v0.11.0-dev-6+g06135cc21
Build type: RelWithDebInfo
LuaJIT 2.1.1713484068
```


hoofcushion:
https://github.com/neovim/neovim/blob/62eb7e79a5f9b5f476f034b5446d2972c840ef87/runtime/lua/vim/iter.lua#L231-L237

hoofcushion:
The test failed, But it has nothing to do with my changes.

hoofcushion:
I pushed the wrong repo...

PR #6654, opened 2024-05-18
fix: do not use flatten from neovim std

hoofcushion:
The last pr is closed, probably because I did something stupid.
I am sorry about that.
#6653

I open this pr because vim.iter():flatten() deprecated again.
Maybe it's better to write an own one.
image
```
$ nvim --version
NVIM v0.11.0-dev-6+g06135cc21
Build type: RelWithDebInfo
LuaJIT 2.1.1713484068
```

### DIFF
--- a/lua/nvim-treesitter/compat.lua
+++ b/lua/nvim-treesitter/compat.lua
@@ -28,12 +28,24 @@ function M.require_language(lang, opts)
   return (ts.language.add or ts.language.require_language)(lang, opts)
 end
 
-function M.flatten(t)
-  if vim.fn.has "nvim-0.11" == 1 then
-    return vim.iter(t):flatten():totable()
-  else
-    return vim.tbl_flatten(t)
+local insert = table.insert
+---@param list any[]
+---@return table Flattened
+local function flatten(list, l)
+  -- lua std does not using ipairs
+  for i = 1, #l do
+    local v = l[i]
+    if type(v) == "table" then
+      flatten(list, v)
+    else
+      insert(list, v)
+    end
   end
+  return list
+end
+
+function M.flatten(t)
+  return flatten({}, t)
 end
 
 return M


hoofcushion:
I've tried running tests in the normal version, but they also fail, and the test failures should have nothing to do with my changes

wookayin:
> vim.iter():flatten() deprecated again.

I don't think this is true --- we didn't deprecate it!

clason:
Thank you, but no. The current master works(!) on all platforms (unless I made another mistake), and we don't care about luals having a hissy fit.

zjykzk:
i encounter the same problem...

hoofcushion:
> I don't think this is true --- we didn't deprecate it!

But why there is only an error in it?

wookayin:
You never talked about what the error looks like nor the (non-existent) source saying vim.iter():flatten is deprecated.

hoofcushion:
That's my fault, I didn't look it carefully.

But my point stands still. There is an inconsistency in these two flatten.

https://github.com/nvim-treesitter/nvim-treesitter/blob/7c9c5bed8340031f8a6dad47a58a26eaf15b9a56/lua/nvim-treesitter/compat.lua#L31-L37

This issue mentioned the difference bettwen tbl_flatten and Iter:flatten

https://github.com/neovim/neovim/issues/28777

E.g.

```lua
vim.iter({1, nil, 3}):flatten():totable() -- ERROR: flatten() requires a list-like table
```

And tbl_flatten would be weird:

```lua
vim.tbl_flatten({1, nil, 3}) -- {1}
```

They can not handle the potential map-like table, and to avoid this inconsistency, the best solution is just not use them at all.

clason:
Again, this has been fixed on Neovim master; `vim.iter` now works with array-like tables. Your code snippet does not give an error for me.

hoofcushion:
My bad, I must meet this error before that update. It works fine now. Thanks.

PR #6671, opened 2024-05-21
feat: rescript parser and queries

ribru17:
Adds [rescript](https://rescript-lang.org/) parser. Queries adapted/improved from the base repo (which currently doesn't track `parser.c`, but there is a PR for it).

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1804,6 +1804,14 @@ list.requirements = {
   readme_name = "pip requirements",
 }
 
+list.rescript = {
+  install_info = {
+    url = "https://github.com/rescript-lang/tree-sitter-rescript",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@ribru17" },
+}
+
 list.rnoweb = {
   install_info = {
     url = "https://github.com/bamonroe/tree-sitter-rnoweb",
--- a/queries/rescript/highlights.scm
+++ b/queries/rescript/highlights.scm
@@ -1,0 +1,340 @@
+(comment) @comment @spell
+
+; Identifiers
+;------------
+; Escaped identifiers like \"+."
+((value_identifier) @constant.macro
+  (#lua-match? @constant.macro "^%.*$"))
+
+(value_identifier) @variable
+
+[
+  (type_identifier)
+  (unit_type)
+  (list)
+  (list_pattern)
+] @type
+
+((type_identifier) @type.builtin
+  (#any-of? @type.builtin "int" "char" "string" "float" "bool" "unit"))
+
+[
+  (variant_identifier)
+  (polyvar_identifier)
+] @constructor
+
+(record_type_field
+  (property_identifier) @property)
+
+(record_field
+  (property_identifier) @property)
+
+(object
+  (field
+    (property_identifier) @property))
+
+(object_type
+  (field
+    (property_identifier) @property))
+
+(module_identifier) @module
+
+(member_expression
+  (property_identifier) @variable.member)
+
+(value_identifier_path
+  (module_identifier)
+  (value_identifier) @variable)
+
+(record_pattern
+  (value_identifier_path
+    (value_identifier) @variable.member))
+
+(record_pattern
+  (value_identifier) @variable)
+
+(labeled_argument
+  label: (value_identifier) @variable.parameter)
+
+; Parameters
+;----------------
+(list_pattern
+  (value_identifier) @variable.parameter)
+
+(spread_pattern
+  (value_identifier) @variable.parameter)
+
+; String literals
+;----------------
+[
+  (string)
+  (template_string)
+] @string
+
+(character) @character
+
+(escape_sequence) @string.escape
+
+; Other literals
+;---------------
+[
+  (true)
+  (false)
+] @boolean
+
+(number) @number
+
+(polyvar) @constructor
+
+(polyvar_string) @constructor
+
+; Functions
+;----------
+; parameter(s) in parens
+(parameter
+  (value_identifier) @variable.parameter)
+
+(labeled_parameter
+  (value_identifier) @variable.parameter)
+
+; single parameter with no parens
+(function
+  parameter: (value_identifier) @variable.parameter)
+
+; first-level descructuring (required for nvim-tree-sitter as it only matches direct
+; children and the above patterns do not match destructuring patterns in NeoVim)
+(parameter
+  (tuple_pattern
+    (tuple_item_pattern
+      (value_identifier) @variable.parameter)))
+
+(parameter
+  (array_pattern
+    (value_identifier) @variable.parameter))
+
+(parameter
+  (record_pattern
+    (value_identifier) @variable.parameter))
+
+; function identifier in let binding
+(let_binding
+  pattern: (value_identifier) @function
+  body: (function))
+
+; function calls
+(call_expression
+  function: (value_identifier_path
+    _
+    (value_identifier) @function.call))
+
+(call_expression
+  function: (value_identifier) @function.call)
+
+; highlight the right-hand side of a pipe operator as a function call
+(pipe_expression
+  _
+  [
+    (value_identifier_path
+      _
+      (value_identifier) @function.call)
+    (value_identifier) @function.call
+  ])
+
+; Meta
+;-----
+(decorator_identifier) @attribute
+
+(extension_identifier) @keyword
+
+"%" @keyword
+
+; Misc
+;-----
+(polyvar_type_pattern
+  "#" @constructor)
+
+[
+  "include"
+  "open"
+] @keyword.import
+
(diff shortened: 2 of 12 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -1748,6 +1748,16 @@ list.requirements = {
   readme_name = "pip requirements",
 }
 
+list.rescript = {
+  install_info = {
+    url = "https://github.com/rescript-lang/tree-sitter-rescript",
+    files = { "src/parser.c", "src/scanner.c" },
+    requires_generate_from_grammar = true,
+    use_makefile = true, -- macOS specific instruction
```
clason, line 1756: Sorry, but that is not acceptable (anymore). If it doesn't build out of the box, it needs its own plugin.

clason (changes requested):

lua/nvim-treesitter/parsers.lua
clason: I don't think that's required, though -- but they do need to fix the include header for their scanner (and probably could use a bunch of other updates @amaanq)

ribru17: There is a PR to add all the generated files (and maybe it also fixes what you are talking about?). I'll mark as draft  and wait for that to get merged, that way no `generate_from_grammar` needed either

clason: The `generate_from_grammar` is fine; it's the `use_makefile` that is unacceptable.

(And the PR does not fix the build issue since it doesn't touch the scanner.)

I would recommend making a parallel PR _just_ fixing that (and maybe committing the two json files and headers); I have total sympathy for maintainers noping out on a +200K diffstat PR. (The bindings were a mistake. `ts g --no-bindings` 4tw.)

ribru17: Sounds good. Also is it possible the build issue is mac only? For me the parser builds when checking out this PR, and `ts g --no-bindings && ts t` works fine

clason: Could be platform (compiler) specific, yes. But it still should be fixed sooner this would pick up a system treesitter header (which could be any old version).

Emilios1995:
@ribru17 The queries and highlights on the main repo's `main` branch definitely need some improvement, but it would be great for those improvements to be on the parser's repo itself. We've recently moved the repo to `rescript`'s GH organization and are looking into trying to actively maintain it, so we will look at incoming PRs. 

Actually, we have been working on improving the queries on this [PR](https://github.com/rescript-lang/tree-sitter-rescript/pull/241/), which we're testing specifically on Neovim, so it would be even better if you made improvements over _that_.  (I haven't followed up on the PR for a few weeks, but I intend to do so and get it merged soon.)

clason:
@ribru17 What's the status here? We "own" our queries (which are editor-specific) so we don't want to get blocked by slow upstream maintenance unless absolutely necessary.

lua/nvim-treesitter/parsers.lua
```diff
@@ -1748,6 +1748,15 @@ list.requirements = {
   readme_name = "pip requirements",
 }
 
+list.rescript = {
+  install_info = {
+    url = "https://github.com/rescript-lang/tree-sitter-rescript",
+    files = { "src/parser.c", "src/scanner.c" },
+    requires_generate_from_grammar = true,
```
clason, line 1755:
```suggestion
```
That doesn't seem to be true? The repo contains an (up to date) `parser.c`.

ribru17:
> @ribru17 What's the status here? We "own" our queries (which are editor-specific) so we don't want to get blocked by slow upstream maintenance unless absolutely necessary.

Just saw this, sorry. Upstream PR has been merged now thankfully, fixing this one up right now

queries/rescript/highlights.scm
```diff
@@ -0,0 +1,340 @@
+(comment) @comment @spell
+
+; Identifiers
+;------------
+; Escaped identifiers like \"+."
+((value_identifier) @constant.macro
+  (#match? @constant.macro "^\\.*$"))
```
clason, line 7: Can this be replaced with a Lua pattern?

clason:
I worry a bit that these queries are overly specific at the cost of performance -- is this really necessary for acceptable highlighting?

queries/rescript/highlights.scm
```diff
@@ -0,0 +1,340 @@
...
+
+; highlight the right-hand side of a pipe operator as a function call
+(pipe_expression
+  _
+  [
+    (value_identifier_path
+      _
+      (value_identifier) @function.call)
+    (value_identifier) @function.call
+  ])
```
lucario387, line 141: How does the tree look like under `pipe_expression`? This looks like it can become hellish very quickly

queries/rescript/highlights.scm
```diff
@@ -0,0 +1,340 @@
...
+
+(labeled_parameter
+  (value_identifier) @variable.parameter)
+
+; single parameter with no parens
+(function
+  parameter: (value_identifier) @variable.parameter)
+
+; first-level descructuring (required for nvim-tree-sitter as it only matches direct
+; children and the above patterns do not match destructuring patterns in NeoVim)
```
clason, line 105: 👀 
```suggestion
; first-level destructuring (required for Neovim as it only matches direct
; children and the above patterns do not match destructuring patterns)
```

👀  Can you comment on that?

clason:
Testing `ts g` with HEAD CLI gives the dreaded
```
Warning: rule number is just a `seq` or `choice` rule with a single element. This is unnecessary.
```

This often (but not always) means that the generated parser does not correctly expose the corresponding node to the queries. That should be fixed before inclusion.

ribru17:
@Emilios1995 would you mind commenting on some of these requests? You are more familiar with these queries

queries/rescript/highlights.scm
clason: (I think it can, so it should.)

ribru17: Yes I agree, if I resolved this without fixing it was by accident haha. Apologies

ribru17:
> Testing `ts g` with HEAD CLI gives the dreaded
> 
> ```
> Warning: rule number is just a `seq` or `choice` rule with a single element. This is unnecessary.
> ```
> 
> This often (but not always) means that the generated parser does not correctly expose the corresponding node to the queries. That should be fixed before inclusion.

I remember when this issue was first floating around the root cause was kind of murky. Have there been false positives? Based on the grammar and the json file produced it seems like there are no states where `number` has a redundant `seq` or `choice`. In terms of the node being exposed in the grammar, it seems like it is - it exists in a lot of spots in the corpus tests at least

clason:
>  Based on the grammar and the json file produced it seems like there are no states where number has a redundant seq or choice

There are no false positive reported so far, but it only reports the top rule even if the `seq` is in some other rule in a call sequence. In this case, the issue is https://github.com/rescript-lang/tree-sitter-rescript/blob/444c127686714b2358622427c3bdba7eb09021c6/grammar.js#L1367

ribru17:
Ahh thanks I see, I'll make a PR

queries/rescript/highlights.scm
ribru17: Yeah, the queries were weird but I think all that was needed was an anchor here. I have split things up in a fixup commit to make them simpler, I think they are better now but let me know what you think

queries/rescript/highlights.scm
ribru17: Good catch, think I will just remove that comment. Looking through the queries+grammar looks like this comment is basically just trying to say that those queries are necessary because, by default, something like `(parent (child) @cap)` uses immediate parent matching and not something like `has-ancestor` to match `child`. But that goes without saying and this comment just serves to increase confusion I think

clason (approved):

ribru17:
> I worry a bit that these queries are overly specific at the cost of performance -- is this really necessary for acceptable highlighting?

I have broken up and simplified some of the weirder queries, most are the same as you can see in the fixup commit but hopefully performance won't be too bad. Unless just the quantity of queries themselves has a non-negligible effect

clason:
> Unless just the quantity of queries themselves has a non-negligible effect

It very much does, but I don't think these are excessive (but I haven't tested these myself). `check-queries` output will give you some idea.

queries/rescript/indents.scm
```diff
@@ -0,0 +1,39 @@
...
+  ")"
+  (jsx_closing_element)
+] @indent.branch @indent.end
+
+(jsx_self_closing_element
+  "/" @indent.branch)
+
+(jsx_self_closing_element
+  "/"
+  ">" @indent.end)
```
lucario387, line 28: Is there some problem with merging these 2 queries?
```suggestion
(jsx_self_closing_element
  "/" @indent.branch
  ">"? @indent.end)
```

lucario387 (approved):

PR #6675, opened 2024-05-22
feat: Add `ralph`

tdroxler:
`Ralph` is the smart contract programming language for the [Alephium blockchain](https://alephium.org/)


queries/ralph/highlights.scm
```diff
@@ -0,0 +1,91 @@
...
+(alph_token_id) @type.builtin
+
+(var_declaration) @variable
+(struct_field) @variable
+(arg_def (identifier) @variable)
+(struct_constructor_field (identifier) @variable)
+
+(hex_num) @number
+(integer) @number
+(typed_num) @number
```
ribru17, line 12: Would you mind grouping the captures to make things easier? E.g. for this one it would be:
```query
[
  (hex_num)
  (integer)
  (typed_num)
] @number
```
Also, please run this file through the formatter found at `scripts/format-queries.lua`

queries/ralph/highlights.scm
```diff
@@ -0,0 +1,91 @@
...
+(op_ne) @operator
+(op_lt) @operator
+(op_le) @operator
+(op_gt) @operator
+(op_ge) @operator
+(op_and) @operator
+(op_or) @operator
+(op_not) @operator
+
+(line_comment) @comment
```
ribru17, line 90: Please also give the `@spell` capture to comments

queries/ralph/highlights.scm
```diff
@@ -0,0 +1,91 @@
...
+"struct" @keyword
+"fn" @keyword
+"pub" @keyword
+"let" @keyword
+"while" @keyword
+"for" @keyword
+"if" @keyword
+"else" @keyword
+"emit" @keyword
+"enum" @keyword
```
ribru17, line 41: Please check `CONTRIBUTING.md`, some of these keywords should be more specific (e.g., `"enum"` should be `@keyword.type`)

ribru17:
Please also add an `injections.scm` file, at least for injecting the `comment` parser

ribru17:
Thank you! I think a few more changes maybe and it will be good to go

queries/ralph/highlights.scm
```diff
@@ -0,0 +1,135 @@
...
+  "Interface"
+  "AssetScript"
+  "struct"
+  "enum"
+  "event"
+  "mapping"
+] @keyword.type
+
+[
+  "const"
```
ribru17, line 64: I think this one can be just `@keyword` (Sorry if this was confusing: we usually highlight `"const"` as a modifier in the case of something like `pub const function myfunc() ...`, but not in variable declarations like `const myvar = 3`. I should mention this in the docs!)

queries/ralph/highlights.scm
```diff
@@ -0,0 +1,135 @@
...
+
+(annotation_field
+  (identifier) @variable)
+
+(annotation_field
+  (identifier) @variable)
+
+(map_def
+  (identifier) @variable)
+
```
ribru17, line 49: What do you think about:
```query
(arg_def
```
EDIT: Maybe also `(event_field)` identifiers?

queries/ralph/highlights.scm
```diff
@@ -0,0 +1,135 @@
...
+  "{"
+  "}"
+] @punctuation.bracket
+
+[
+  "."
+  ","
+  ":"
+  ";"
+  "="
```
ribru17, line 103: I think this one should be `@operator`

queries/ralph/highlights.scm
```diff
@@ -0,0 +1,135 @@
...
+[
+  (var_declaration)
+  (struct_field)
+] @variable
+
+(arg_def
+  (identifier) @variable)
+
+(struct_constructor_field
+  (identifier) @variable)
```
ribru17, line 17: Seems like sometimes variables in random places (like `key` in `let depositor = key`) aren't highlighted as `@variable`; would it be easy to place (at the top of the file) just: `(identifier) @variable` and then override the rest (like `(func_id (identifier) @function)`) later? Genuinely asking. If there are too many other non-variable identifier nodes then it's fine

ribru17:
Also it doesn't matter for this PR but just as an FYI it seems that the parser cannot yet handle `import "..."`

tdroxler:
> Also it doesn't matter for this PR but just as an FYI it seems that the parser cannot yet handle import "...

oh nice catch, totally forgot this one :sweat_smile:

queries/ralph/highlights.scm
```diff
@@ -0,0 +1,135 @@
...
+  (bool)
+  (bytes)
+  (address)
+] @constant.builtin
+
+(func_id) @function
+
+[
+  (mutable)
+  (annotation)
```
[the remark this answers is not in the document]
tdroxler: thx for your time, I'm new to tree-sitter so it's a bit a "trial-and-error" experience ^^

Would it make sense to have:
```
(annotation
 [ "@" (identifier) ] @attribute)
```
so the full `@myannotation` is highlighted the same?

image

queries/ralph/highlights.scm
tdroxler: definitively, maybe also `->`

queries/ralph/highlights.scm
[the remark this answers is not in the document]
lucario387: Generally yes. I'd prefer both the `@` and its following identifier to have the same capture applied

queries/ralph/highlights.scm
tdroxler: I gave a try, it looks good with that.

tdroxler: Hopefully I made everything well this time :)

ribru17: Thanks!

ribru17 (approved):

lua/nvim-treesitter/parsers.lua
```diff
@@ -1804,6 +1804,14 @@ list.rust = {
...
 }
 
+list.ralph = {
+  install_info = {
+    url = "https://github.com/alephium/tree-sitter-ralph",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@tdroxler" },
+}
+
```
amaanq, line 1814: put it in its correct spot alphabetically - below racket and above rasi

clason: also in the readme and the lockfile

tdroxler: oops...Done

clason: and the lockfile ;)

tdroxler: -_-'

tdroxler: done

queries/ralph/highlights.scm
```diff
@@ -0,0 +1,139 @@
...
+  (identifier) @function)
+
+(annotation
+  [
+    "@"
+    (identifier)
+  ] @attribute)
+
+(annotation_field
+  (identifier) @variable)
```
lucario387, line 48: The (identifier) in the first few lines should be covering for this already

tdroxler: Indeed, I just push a fix.

queries/ralph/highlights.scm
```diff
@@ -0,0 +1,136 @@
...
+[
+  (bool)
+  (bytes)
+  (address)
+] @constant.builtin
+
+(func_id) @function
+
+(func_id
+  (identifier) @function)
```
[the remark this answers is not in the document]
tdroxler: The outer node is needed as some of our `func_id` can have an extra `!`, e.g. `callerAddress!()` which isn't capture by `identifier`.  
I thought I needed also the inner node to override the top `(identifier @variable)` as proposed [here](https://github.com/nvim-treesitter/nvim-treesitter/pull/6675#discussion_r1611968385)
But indeed I gave a try and keeping only the outer node highlight well  the functions, sorry I'm still a bit new to tree-sitter.

I'll push a fix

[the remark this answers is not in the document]
lucario387: Then you should be capturing the `!` as well. This probably will lead to the function identifier to be highlighted as variable


```diff
-(func_id) @function
+(func_id
+  [
+    (identifier)
+    "!"
+  ] @function)
```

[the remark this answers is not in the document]
tdroxler: but there's not always a `!`, should I have both:

```
(func_id) @function

(func_id
  [
    (identifier)
    "!"
  ] @function)
```
?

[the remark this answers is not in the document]
lucario387: The things inside `[]` are : If there exists any node that matches those in the `[]`, then capture it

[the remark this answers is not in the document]
tdroxler: This works great, thx a lot for your time

lucario387 (approved):

PR #6676, opened 2024-05-22
Update README.md | Update Auto-disable Example

alirezaalavi87:
since `vim.loop` is deprecated, updated the function.
This is also more concise and easier to understand.

Written with help from chatGPT3.5 turbo

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,12 +131,15 @@ require'nvim-treesitter.configs'.setup {
     disable = { "c", "rust" },
     -- Or use a function for more flexibility, e.g. to disable slow treesitter highlight for large files
     disable = function(lang, buf)
-        local max_filesize = 100 * 1024 -- 100 KB
-        local ok, stats = pcall(vim.loop.fs_stat, vim.api.nvim_buf_get_name(buf))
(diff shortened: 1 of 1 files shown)


clason:
That's about the sort of PR I would expect from a glorified Markov chain.

alirezaalavi87:
Wow!
You are so insecure that you get offended by PRs?

Having outdated examples in your document isn't a good thing.
I just spent some time trying to fix it for myself and thought it will be helpful for other people too.

you seem like a small person.

README.md
```diff
@@ -131,13 +131,16 @@ require'nvim-treesitter.configs'.setup {
...
-        if ok and stats and stats.size > max_filesize then
-            return true
-        end
+      local max_filesize = 100 * 1024 -- 100 KB
+      local filename = vim.api.nvim_buf_get_name(buf)
+      local file_size = vim.fn.getfsize(filename)
+
+      if file_size > max_filesize then
+        return true
+      end
```
lucario387, line 140: `getfsize` returns -1, which doesn't seem to be handled here?

Additionally, why not just move vim.loop to vim.uv now that vim.uv is in latest release version of Neovim?

PR #6677, opened 2024-05-22
docs: clarify luap identifier in readme

svvac:


svvac:
test failure seems inherited from master (30de5e7e9486fb1b1b8c2a1e71052b13f94f1cb0)

ribru17:
I think you would also need to change the name [here](https://github.com/nvim-treesitter/nvim-treesitter/blob/30de5e7e9486fb1b1b8c2a1e71052b13f94f1cb0/lua/nvim-treesitter/parsers.lua#L1206), but I'm not certain this would be helpful? Seems less descriptive than the readme name

svvac:
I was after the identifier to pas to `:TSInstall` and had to dig in the repo to find it (it appears I mistakenly took the space for an underscore due to the link on the rendered md).

May I suggest `lua patterns (luap)` instead? Or maybe this is just not meant to be an index, in which case please close :—)

ribru17:
Ah that makes sense. I personally like the second suggestion! Up to @clason.

ribru17 (approved):

clason:
Honestly, I want this list gone (which it already is on `main`). Maybe just replace the whole thing with a link to https://github.com/nvim-treesitter/nvim-treesitter/blob/main/SUPPORTED_LANGUAGES.md for now.

PR #6683, opened 2024-05-24
feat(query)!: upstream changes

amaanq:
pending https://github.com/tree-sitter-grammars/tree-sitter-query/pull/36

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -36,15 +36,16 @@ jobs:
   format-queries:
     name: Lint queries
     runs-on: ubuntu-latest
+    env:
+      NVIM_TAG: stable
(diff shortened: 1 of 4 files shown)


lucario387:
Oh this will be a lot...

amaanq:
lint fails because it uses neovim w/ the bundled parser I think

clason:
And now you see why I'm so hesitant to make breaking changes to that parser...

clason:
I guess we need to add a `:TSInstallSync query` step to the lint workflow.

clason:
Not sure how to do this on `main`, though. I'll make a tree-sitter-query release and bump the parser in Neovim (nightly and release).

PR #6687, opened 2024-05-26
feat(cpp): update highlights

amaanq:
arduino and the others probably need to update


clason:
Look to be fine on my end!

PR #6689, opened 2024-05-26
fix(eds)!: adapt to breaking repo change

clason:


uyha:
hi, don't move move on with this PR, I'm sorry for deleting the repo (didn't remember that I added it to `nvim-treesitter`, I restored it. I will make a PR later when the new parser is created.

PR #6699, opened 2024-05-28
rust: increase highlight priority for return

WhyNotHugo:
LSP semantic tokens set all keywords to `Keyword` with priority 125. This overrides the more-specific match in this treesitter query. As a result, the 'return' and 'yield' keywords are not highlighted as such.

The :Inspect command indicates:

  Treesitter
    - @keyword.return.rust links to @keyword.return rust

  Semantic Tokens
    - @lsp.type.keyword.rust links to Keyword priority: 125
    - @lsp.mod.controlFlow.rust links to @lsp priority: 126 - @lsp.typemod.keyword.controlFlow.rust links to @lsp priority: 127

Change the priority of the tree sitter query so is is not overridden by LSP semantic tokens.


WhyNotHugo:
v2: Fixed formatting

clason:
This is intentional. Semantic tokens have priority over treesitter, period. You are welcome to override this in your configuration.

clason:
Clear reasoning: priorities come in brackets according to upstream documentation, and treesitter's bracket is 51-124. Within that bracket, you can be sure of no interference from other sources.

clason:
Again, you can change this in your own config, but the default priorities **must** respect their brackets.

WhyNotHugo:
> priorities come in brackets according to upstream documentation, and treesitter's bracket is 51-124

Again, what's the intention behind that?

WhyNotHugo:
Just trying to fully understand context here.

WhyNotHugo:
Keeping my own forked version (and continuously tracking upstream) is a bit of a pain, so I keep that option as a last resort.

Do you think there's any other approach to fix the underlying bug here?

ribru17:
> Do you think there's any other approach to fix the underlying bug here?

See `:h treesitter-query-modeline-extends`. Just put
```query
; extends
([
  "return"
  "yield"
] @keyword.return
  (#set! "priority" 130))
```
in `~/.config/nvim/queries/rust/highlights.scm`. Or just use `vim.api.nvim_set_hl(0, '@lsp.type.keyword.rust', {})`

clason:
Or, if you believe there is a bug here (in rust-analyzer), go tell them about it. From the point of view of Neovim, there is no bug; everything is working as intended (legacy < treesitter < semantic < extmarks, by level of information available).

WhyNotHugo:
@ribru17 Thanks, that works well enough for me!

@clason I can't really confirm where cause of the bug lies. As an end user, I see that the `return` keyword is not highlighted with the expected highlight group (`@keyword.return`).

PR #6704, opened 2024-05-29
[injections/yaml]: bash injections in ansible files

magnuslarsen:
Brings some color to Ansible's command/shell modules:

image

...without polluting other modules.

I've added a tests file based on the others; though I'm not sure how it works - please let me know if I should do it differently


magnuslarsen:
Running the `format-queries.lua` script produces non-idempotent changes on these queries :S

magnuslarsen:
My Ansible injections, if somebody finds this PR in the future: https://github.com/magnuslarsen/dotfiles/blob/master/dot_config/nvim/queries/yaml/injections.scm

PR #6709, opened 2024-05-31
fix(odin): highlight types with underscores

Player01osu:
Odin's [naming convention](https://github.com/odin-lang/Odin/wiki/Naming-Convention) calls for types to be in Ada_Case. It therefore makes sense for types with underscores to also be highlighted as such.


amaanq:
nice, thanks

PR #5640, opened 2023-11-08
feat: add asciidoc parser

ObserverOfTime:
The [parser](https://github.com/cathaysia/tree-sitter-asciidoc) is still a WIP and the [fork](https://github.com/cpkio/tree-sitter-asciidoc) is…worse.
Just leaving this here for later since I gave it a try.

PR #6710, opened 2024-05-31
feat(parser): add asciidoc

cathaysia:
The asciidoc parser is a (in)completed parser. There are still some cases where it cannot parse, but it should be able to parse most cases.

Since there is still no official grammar specification, the parser may have some deviate from the official one.

[the state of the parser](https://github.com/cathaysia/tree-sitter-asciidoc/issues/2)

https://github.com/nvim-treesitter/nvim-treesitter/pull/6710#issuecomment-2147622735

### DIFF
--- a/queries/asciidoc/highlights.scm
+++ b/queries/asciidoc/highlights.scm
@@ -1,0 +1,168 @@
+(title0) @markup.heading.1
+
+(title1) @markup.heading.2
+
+(title2) @markup.heading.3
+
+(title3) @markup.heading.4
+
+(title4) @markup.heading.5
+
+(title5) @markup.heading.6
+
+(email) @markup.link.url @markup.link
+
+(author_line
+  ";" @punctuation.delimiter)
+
+(revision_line
+  "," @punctuation.delimiter
+  ":" @punctuation.delimiter)
+
+[
+  (firstname)
+  (middlename)
+  (lastname)
+] @attribute
+
+(revnumber) @number
+
+(revdate) @string.special
+
+(revremark) @string
+
+(table_block_marker) @punctuation.special
+
+(table_cell_attr) @attribute
+
+(table_cell
+  "|" @punctuation.special)
+
+[
+  (breaks)
+  (hard_wrap)
+  (quoted_block_md_marker)
+  (quoted_paragraph_marker)
+  (open_block_marker)
+  (listing_block_marker)
+  (literal_block_marker)
+  (passthrough_block_marker)
+  (quoted_block_marker)
+  (ntable_block_marker)
+  (anno_marker)
+] @punctuation.special
+
+(ntable_cell
+  "!" @punctuation.special)
+
+(task_list_marker_unchecked) @markup.list.unchecked
+
+(task_list_marker_checked) @markup.list.checked
+
+[
+  (list_marker_star)
+  (list_marker_hyphen)
+  (list_marker_dot)
+  (list_marker_digit)
+  (list_marker_geek)
+  (list_marker_alpha)
+] @markup.list
+
+[
+  (line_comment)
+  (comment_block)
+] @comment @spell
+
+[
+  (document_attr_marker)
+  (element_attr_marker)
+] @punctuation.delimiter
+
+(document_attr
+  (attr_name) @property)
+
+(element_attr
+  (attr_value) @attribute)
+
+(block_title
+  (block_title_marker) @punctuation.special) @attribute
+
+(ident_block) @markup.raw.block
+
+(anno_list_marker) @punctuation.special
+
+(delimited_block) @punctuation.bracket
+
+(block_macro
+  (block_macro_name) @keyword
+  "::" @punctuation.delimiter
+  (target)? @markup.link
+  "[" @punctuation.bracket
+  "]" @punctuation.bracket)
+
+(block_macro_attr
+  (name) @attribute
+  "=" @operator
+  (value) @variable.parameter)
+
+(admonition
+  (admonition_important) @comment.error
+  ":" @comment.error)
+
+(admonition
+  (admonition_warning) @comment.warning
+  ":" @comment.warning)
+
+(admonition
+  (admonition_caution) @comment.warning
+  ":" @comment.warning)
+
+(admonition
+  (admonition_note) @comment.note
+  ":" @comment.note)
+
+(admonition
+  (admonition_tip) @comment.note
+  ":" @comment.note)
+
+((section_block
+  (element_attr
+    (element_attr_marker) @comment.note
+    (attr_value) @attribute @comment.note
+    (element_attr_marker) @comment.note)
+  (delimited_block
+    (delimited_block_marker) @comment.note
+    (delimited_block_marker) @comment.note))
+  (#any-of? @attribute "NOTE" "TIP"))
+
+((section_block
+  (element_attr
+    (element_attr_marker) @comment.warning
+    (attr_value) @attribute @comment.warning
+    (element_attr_marker) @comment.warning)
+  (delimited_block
+    (delimited_block_marker) @comment.warning
+    (delimited_block_marker) @comment.warning))
+  (#any-of? @attribute "CAUTION" "WARNING"))
+
+((section_block
+  (element_attr
+    (element_attr_marker) @comment.error
+    (attr_value) @attribute @comment.error
+    (element_attr_marker) @comment.error)
+  (delimited_block
+    (delimited_block_marker) @comment.error
+    (delimited_block_marker) @comment.error))
+  (#eq? @attribute "IMPORTANT"))
+
+((element_attr
+  (element_attr_marker)
+  (attr_value) @_attr
+  (element_attr_marker))
+  (listing_block
+    (listing_block_body) @markup.raw.block)
+  (#not-lua-match? @_attr "^[^,]*,%s*(%w+)")
+  (#not-any-of? @_attr
+    "a2s" "barcode" "blockdiag" "bpmn" "bytefield" "d2" "dbml" "diagrams" "ditaa" "dpic" "erd"
+    "gnuplot" "graphviz" "lilypond" "meme" "mermaid" "msc" "nomnoml" "pikchr" "plantuml" "shaape"
+    "smcat" "structurizr" "svgbob" "symbolator" "syntrax" "tikz" "umlet" "vega" "wavedrom"))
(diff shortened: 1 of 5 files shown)


clason:
Thank you, but we don't take "experimental" or incomplete parsers anymore. Can you please detail (on the parser repo README, e.g., as a list) what does and doesn't work?)

cathaysia:
I update the README of parser's repo.

Since there is no official specification, this parser may not be complete for a long time. In addition, this grammar relies heavily on context-sensitive parsing. Using LR(1) may not be able to fully parse its grammar.

This parser is much better than the built-in highlighting of vim. So I think it can be merged into

clason:
> This parser is much better than the built-in highlighting of vim.

Maybe, but we don't have to maintain that, while _this_ we would have. So that argument holds no weight. And what you describe makes me wary of adding it here.

(Just to be clear: Our goal is not (anymore) to bundle all possible parsers; we have to work on raising the **total** quality since even a single "bad" parser -- in functionality or, much worse, performance, will have people spouting off about "tree-sitter being slow/bad/disaster" on Reddit.)

clason:
I would like some other maintainer's opinion on this. And obviously the parser is welcome when it is finished (and you are no longer concerned about performance); asciidoc has been requested often.

ObserverOfTime:
Tried it in #5640. No point if it's still incomplete.

clason:
To be fair, half a year of work has gone into since then. I'd say that merits another look.

ObserverOfTime:
I have two asciidoc files at hand and it doesn't properly parse either (though it doesn't ERROR out).
1. It doesn't consider the line break character (`+`) as special.
2. It fails to parse code blocks with more than 4 dashes.
It does seem usable now though, and both of those should be fairly easy to fix.

cathaysia:
Parsing nested blocks or blocks with more than 4 dashes requires maintaining a stack. This is a bit cumbersome, so I have not done it yet.
The line continue case is a bit more complicated. It requires maintaining a stack in the inline parser. This conflicts with the pass_through symbol.

queries/asciidoc_inline/injections.scm
ObserverOfTime, line 1: Add an injection for comment nodes.

ObserverOfTime (commented):
Fold queries would also be nice to have, but not required.

queries/asciidoc/injections.scm
```diff
@@ -0,0 +1,43 @@
...
+  (element_attr
+    (element_attr_marker)
+    (attr_value) @injection.language
+    (element_attr_marker))
+  (raw_block
+    (raw_block_marker)
+    (raw_block_body) @injection.content
+    (raw_block_marker)))
+  (#vim-match? @injection.language
+    "a2s\|barcode\|blockdiag\|bpmn\|bytefield\|d2\|dbml\|diagrams\|ditaa\|dpic\|erd\|gnuplot\|graphviz\|graphviz\|lilypond\|meme\|mermaid\|msc\|nomnoml\|pikchr\|plantuml\|shaape\|smcat\|structurizr\|svgbob\|symbolator\|syntrax\|tikz\|umlet\|vega\|wavedrom"))
```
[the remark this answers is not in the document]
clason: Or `#any-of?`, if they need to be hardcoded for some reason.

cathaysia:
If this PR cannot be merged, is it better to change these in the original repository? Since this PR, I have pushed some new commits

clason:
Nobody said that this PR cannot be merged.

cathaysia:
> Nobody said that this PR cannot be merged.

I think we may first need to clarify under what circumstances this parser is allowed to be merged. If the parser is required to parse the entire asciidoc grammar correctly, I think it will be difficult, especially when lookahead and lookaround cannot be used.

I'm not complaining, of course. It's just that clarifying this will save us from too much disagreement and a lot of duplication of work.

cathaysia:
@mojavelinux @asciidoctor Do you have any suggestions on this?

cathaysia:
OK, in that case this discussion should be meaningful. I need to improve the documentation and testing for a while. I will submit relevant evidence after the API is stable.

queries/asciidoc/injections.scm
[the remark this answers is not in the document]
cathaysia: yes, `#any-of?`, here cannot be a regex.

queries/asciidoc/injections.scm
```diff
@@ -0,0 +1,43 @@
...
+  (element_attr
+    (element_attr_marker)
+    (attr_value) @injection.language
+    (element_attr_marker))
+  (raw_block
+    (raw_block_marker)
+    (raw_block_body) @injection.content
+    (raw_block_marker)))
+  (#gsub! @injection.language "%s*,%s*(%w+)%s*" "%1")
+  (#vim-match? @injection.language "\\s*,\\s*.*\\s*"))
```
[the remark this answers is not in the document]
cathaysia: There are two forms of syntax:

```adoc
[, cpp]
----
int a  = 10;
----
```

or 
```adoc
[source, cpp]
----
int a = 10;
----
```

Since Lua does not have the concept of optional group, vim-regex is used here.

cathaysia:
| element      | ts | vim |
| ----------- | ----------- |---|
| callout | image | image |
| title0 | image | image|
| block macro | image |image |
| admonition | image| image |
| inline | image| image|

Besides that, vim syntax often renders wrong colors when scrolling.

mojavelinux:
> @mojavelinux @asciidoctor Do you have any suggestions on this?

I encourage you to ask in https://chat.asciidoc.org and to also follow the development on the language specification.

ObserverOfTime:
> Maintaining a stack is essential to parse such case.

You can see how we handled a similar case in the [TS test scanner](https://github.com/tree-sitter-grammars/tree-sitter-test/blob/master/src/scanner.c).
I assume asciidoc doesn't have arbitrary suffixes so it should be even simpler.

> Although prec.right sounds like it would work in this situation, it actually doesn't.

That was for `++++` vs `+`.

queries/asciidoc/injections.scm
ObserverOfTime, line 1:
```clojure
([
  (line_comment)
  (comment_block)
] @injection.content

File: queries/asciidoc/injections.scm [hunk 3]
```diff
@@ -0,0 +1,44 @@
...
+  (element_attr
+    (element_attr_marker)
+    (attr_value) @injection.language
+    (element_attr_marker))
+  (raw_block
+    (raw_block_marker)
+    (raw_block_body) @injection.content
+    (raw_block_marker)))
+  (#gsub! @injection.language "%s*source%s*,%s*(%w+)%s*" "%1")
+  (#vim-match? @injection.language "\\s*source\\s*,\\s*.*\\s*"))
```
ObserverOfTime, line 30: It would be better to handle this by special-casing it in the parser if possible.
```clojure
(section_block
  (element_attr
    (element_attr_marker)
    (attr_value
      (source_language) @injection.language)
    (element_attr_marker))
  (raw_block
    (raw_block_marker)
    (raw_block_body) @injection.content
    (raw_block_marker)))
```

queries/asciidoc_inline/injections.scm
ObserverOfTime, line 1: Don't include the file here if it will just be commented out.
Also you can merge the two cases: `[ "stem" "latexmath" ]` .

queries/asciidoc/injections.scm
ObserverOfTime: If that's not possible, then I think the (singular) query should look like this:
```clojure
((section_block
  (element_attr
    (element_attr_marker)
    (attr_value) @injection.language
    (element_attr_marker))
  (raw_block
    (raw_block_marker)
    (raw_block_body) @injection.content
    (raw_block_marker)))
  (#gsub! @injection.language "^[^,]*,%s*(%w+)" "%1")
  (#lua-match? @injection.language "^[^,]*,%s*(%w+)"))
```

queries/asciidoc/highlights.scm
```diff
@@ -0,0 +1,155 @@
...
+(document_attr
+  (attr_name) @property)
+
+(element_attr
+  (attr_value) @attribute)
+
+(block_title
+  (block_title_marker) @punctuation.special) @attribute
+
+(raw_block_body) @markup.raw.block
```
ObserverOfTime, line 89: This will interfere with injected syntax and should be `@none` or removed entirely.

queries/asciidoc/injections.scm
cathaysia: Similar to the following. `attr_value` cannot be any value. Otherwise it is not a code block.

ObserverOfTime: [But it can.](https://docs.asciidoctor.org/asciidoc/latest/verbatim/highlight-lines/)

queries/asciidoc/highlights.scm
ObserverOfTime: Those are also injected.

queries/asciidoc/injections.scm
```diff
@@ -0,0 +1,33 @@
...
+    (attr_value) @injection.language
+    (element_attr_marker))
+  (listing_block
+    (listing_block_marker)
+    (listing_block_body) @injection.content
+    (listing_block_marker)))
+  (#any-of? @injection.language
+    "a2s" "barcode" "blockdiag" "bpmn" "bytefield" "d2" "dbml" "diagrams" "ditaa" "dpic" "erd"
+    "gnuplot" "graphviz" "graphviz" "lilypond" "meme" "mermaid" "msc" "nomnoml" "pikchr" "plantuml"
+    "shaape" "smcat" "structurizr" "svgbob" "symbolator" "syntrax" "tikz" "umlet" "vega" "wavedrom"))
```
ObserverOfTime, line 33:
```suggestion
    "gnuplot" "graphviz" "lilypond" "meme" "mermaid" "msc" "nomnoml" "pikchr" "plantuml" "shaape"
    "smcat" "structurizr" "svgbob" "symbolator" "syntrax" "tikz" "umlet" "vega" "wavedrom"))
```
"graphviz" is duplicated.

queries/asciidoc/highlights.scm
cathaysia: I understand what you mean. You don't want any rule be covered.

cathaysia:
Here is a benchmark about this parser:

https://cathaysia.github.io/tree-sitter-asciidoc/#benchs

The files used in this bench come from https://github.com/asciidoctor/asciidoctor .

cathaysia:
Hi everyone. Thank you very much for your help here. But for some personal reasons and to avoid trademark disputes. I decided to rename this parser. As a price, I can no longer make this PR under the name of asciidoc. So I will close this PR, but all issues in the current PR will be merged into the parser repository. I am deeply sorry for the inconvenience caused to everyone.

PR #6712, opened 2024-05-31
feat(gotmpl,helm): raise highlight priorities

kmoschcau:


qvalentin:
Thanks @kmoschcau.

The reason for this change was already discussed in https://github.com/nvim-treesitter/nvim-treesitter/pull/6050.

The problem can be seen in this image:
image

kmoschcau:
To explain this screenshot further, this is highlighted with my own theme. In it, literals have a background color. The YAML string is highlighted green and overlaps the highlighting of gotmpl/helm. Raising the priorities of the gotmpl/helm highlights will make them show up.

queries/gotmpl/highlights.scm
lucario387, line 1: It doesn't make sense if almost all of them have priorities raised.

kmoschcau: Why not? We have been over this already in the linked thread. You were part of this discussion. Solving this in other ways requires a rewrite of the parser, to my understanding.

qvalentin:
Hi @lucario387 can you suggest an alternative if this makes not sense for you? This solution fixes the problem, so why should we not use it?

lucario387:
> Hi @lucario387 can you suggest an alternative if this makes not sense for you? This solution fixes the problem, so why should we not use it?

It was a dumb moment from me, I was testing this with both Helix and Neovim, and didn't realize why priorities was needed (on Helix 23.10 it doesn't need priorities, possibly because of reverse highlight ordering). 

Now that I look at it again, it makes sense, so please forget whatever I said this time last month

lucario387:
Please rebase to latest master and let it run again :)

I had enabled auto-merge if the test pass

kmoschcau:
@lucario387 I rebased and pushed it. You'll have to look again. Auto-merge does not survive such pushes.

PR #6714, opened 2024-05-31
test: fix tests broken after 0.10 update

ribru17:


ribru17:
Looks like `highlight_spec` no longer checks for injections. Need to fix that. Other than that everything is working (html indent test failure is accurate)

lucario387 (approved):

PR #6716, opened 2024-06-01
feat(predicates)!: rename `has-type` to `kind-eq` to align with Helix

clason:
No point in having a different name for the same predicate.

(Helix uses quotes for the second argument even though those are not literals. I prefer that, to be honest.)


lucario387:
I'm not against it.

lucario387 (approved):
More alignments with other treesitter consumers = less chores for nvim :D 

How can one hate that

NoahTheDuke:
This is breaking multiple plugins I use. Can the original function be left as an alias at least for a little while?

clason:
Please pin the v0.9.2 tag until plugins catch up.

PR #6733, opened 2024-06-06
Update branch info for parsers.

jackruder:
# Issue:
Several parsers would not automatically install because the install defaults to the "master" branch. 
# Solution:
Update the parsers with correct branches. 

Updated gitcommit, gitignore, r, lua, norg. 

# New behavior:
parsers auto-install as desired.


clason:
Thank you, but that is not true: The default branch is `main`, so the branch need _not_ be specified (as long as the lockfile points to the correct commit). So something must be broken on your side.

PR #6735, opened 2024-06-06
feat(ruby): add highlights

omohokcoj:
Add highlight groups for ruby special methods

Before
image
image

After
image
image

### DIFF
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -127,7 +127,7 @@
 ((identifier) @keyword.function
   (#any-of? @keyword.function
     "include" "extend" "prepend" "attr_reader" "attr_writer" "attr_accessor" "module_function"
-    "refine" "using")))
+    "refine" "using"))
 
 ((identifier) @keyword.exception
   (#any-of? @keyword.exception "raise" "fail" "catch" "throw"))
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -124,8 +124,8 @@
   (#any-of? @constant.builtin
     "__callee__" "__dir__" "__id__" "__method__" "__send__" "__ENCODING__" "__FILE__" "__LINE__"))
 
-((identifier) @keyword.function
-  (#any-of? @keyword.function
+((identifier) @function.builtin
+  (#any-of? @function.builtin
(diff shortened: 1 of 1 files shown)


queries/ruby/highlights.scm
```diff
@@ -127,6 +124,14 @@
   (#any-of? @constant.builtin
     "__callee__" "__dir__" "__id__" "__method__" "__send__" "__ENCODING__" "__FILE__" "__LINE__"))
 
+((identifier) @keyword.function
+  (#any-of? @keyword.function
+    "include" "extend" "prepend" "attr_reader" "attr_writer" "attr_accessor" "module_function"
```
lucario387, line 129: `include` looks like it should be `keyword.import`. The other doesn't really look like functions, moreso just general keywords?

omohokcoj: @lucario387 in ruby "include" "extend" "prepend" "attr_reader" "attr_writer" "attr_accessor" "module_function" are not technically keywords but special class/module methods https://www.rubydoc.info/stdlib/core/Object:extend - so it's more like a function with a keyword meaning that's why i thought `@keyword.function` is the closes available group to it (also this group gives the right color)

lucario387: Got it, so include is moreso "adding things to base class", rather than typical imports. Got it

omohokcoj:
@lucario387 please let me know if there is anything that needs to be adjusted for those changes to be merged

queries/ruby/highlights.scm
amaanq: why not `@function.builtin`?

omohokcoj: @amaanq currently highlights.scm doesn't include `@function.builtin` so i just don't want to introduce new highlight groups. Also `@function.builtin` doesn't return the correct color by default so it will require adjusting this highlight group color for ruby

clason: Colors are irrelevant here; that's a matter for individual colorschemes. Important is that the captures are _semantically_ correct. And any capture in CONTRIBUTING.md can (and should, if appropriate) be used.

omohokcoj: @function.builtin group has been added

lucario387 (approved):

PR #6737, opened 2024-06-06
Fix example of the parser install directory

lambdalisue:
I've encountered an `Invalid node type` error on Neovim 0.10.0 using the latest nvim-treesitter.

image

This issue occurred because my `parser_install_dir` was **appended** to my `runtimepath`, causing Neovim to use its own parsers instead of those from nvim-treesitter.


lambdalisue:
Thanks 🎉

linsui:
Does [this line](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/README.md?plain=1#L122) also need to be changed?

PR #6741, opened 2024-06-07
feat(php): php 8.4 support

calebdw:
Hello!

This PR:
- adds php 8.4 support
- updates the php queries to be compatible with the BCs in https://github.com/tree-sitter/tree-sitter-php/pull/247
- fixes import highlights

I added myself to the list of php maintainers seeing as I also help maintain the tree-sitter-php grammar.

Thanks!


ribru17:
Thanks! I think you need up to update the lockfile to get checks passing

calebdw:
I'm aware :), I was going to update the lockfile when the upstream pr was merged

ribru17:
Ah sorry I see this is draft now :sweat_smile:

calebdw:
This should be good to go!

calebdw:
Thank you!

mehalter:
Hey! Glad to see these updates :D should this commit/PR have been marked as a breaking change since query groups are removed in the new revision? Just wanted to let it be known since this did break queries in other plugins and seeing breaking change indicators in the changelog is very useful for maintainers/developers.

clason:
_All_ updates have to be considered as breaking, as this is still zerover. This is something we want to change after the 1.0 release, but for now it's caveat emptor.

mehalter:
Awesome, thanks so much for the clarification!

clason:
(To be clear, the difficulty here is that the breaking update happens on the parser side, not the query side -- and we have no control, and often no information, of that as the majority of parsers are still not versioned. That needs to change for us to receive and propagate downstream that information.)

PR #6742, opened 2024-06-08
feat(markdown_inline): Conceals for character escapes

ValdezFOmar:


ribru17:
I think you could change all of the paragraphs to
```query
((backslash_escape) @string.escape
  (#offset! @string.escape 0 0 0 -1)
  (#set! conceal ""))
```
This also fixes the fact that the current queries are missing some other possible escape sequences, like `\$`

queries/markdown_inline/highlights.scm
```diff
@@ -87,6 +87,11 @@
   (uri_autolink)
 ] @markup.link.url @nospell
 
+[
+  (link_label)
+  (email_autolink)
```
ribru17, line 92: Looks like we also have no regular highlights for email autolinks. Would you mind giving it one?

queries/markdown_inline/highlights.scm
```diff
@@ -87,6 +87,11 @@
...
+] @nospell
```
ribru17, line 93: While we're at it, can you add `(entity_reference)` to the nospell list?

clason:
To be honest, I'm not sure we want this (but also not sure yet we don't). The issue is that the markdown parser is used by Neovim as a renderer for LSP documentation, and it seems to me that such strings are something you would want to see literally (for -- manual -- copy&pasting). Unfortunately, conceal is all-or-nothing and already used for rendering, so we can't play the "just don't enable conceal" card here.

clason:
Also, can you pull out the unrelated changes to a separate PR? The `@nospell` for `link_label` is also debatable, as the label often _is_ normal text that should be spell-checked.

ValdezFOmar:
> Also, can you pull out the unrelated changes to a separate PR? The `@nospell` for `link_label` is also debatable, as the label often _is_ normal text that should be spell-checked.

Done https://github.com/nvim-treesitter/nvim-treesitter/pull/6750

clason:
Closing since I still think we need to be careful about default conceal; I'd prefer to leave this driven by upstream. Thanks, though!

PR #6744, opened 2024-06-08
feat: more comment injections

ribru17:
Adds comment injections to parsers missing them (where appropriate). I also gave `@spell` captures to comments that were missing them, in parsers that I touched. Also the `hoon` comment node seemed inappropriate so I changed that one as well. Other than that, the rest of the changes are basically all the same

### DIFF
--- a/queries/ada/injections.scm
+++ b/queries/ada/injections.scm
@@ -1,0 +1,2 @@
+((comment) @injection.content
+  (#set! injection.language "comment"))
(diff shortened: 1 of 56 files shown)


clason:
> Also the hoon comment node seemed inappropriate so I changed that one as well.

Who can tell with that one? It's more of a practical joke than a language...

ObserverOfTime (approved):

Issue #5353, opened 2023-09-07
:TSInstall while committing may break current git repo


staticssleever668:
### Describe the bug

Running `:TSInstall gitcommit` while committing in git broke my repo.
I use `require("nvim-treesitter.install").prefer_git = true`.

### To Reproduce

After encountering it on remote server, I immediately retried and reproduced it once on my home PC, but on subsequent attempts I can't reproduce it again.
Rough steps:
```
mkdir ~/Downloads/test
cd ~/Downloads/test
git init
touch asdf
git add asdf
git commit --no-gpg-sign
nvim asdf # ijklsdfjklsd:x
git add asdf
git commit # ikljsadflkjsfadjlk<ESC>:TSInstall gitcommit:x
git status
# > see error "fatal: unable to read 11544a188ec3192eb3aab635520020d5f175bd0b"
```

### Output of `:checkhealth nvim-treesitter`

Initial encounter:
```
==============================================================================
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
- OK `node` found v12.22.9 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "6.2.0-31-generic",
  sysname = "Linux",
  version = "#31~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Aug 16 13:45:26 UTC 2"
} ~

Parser/Features         H L F I J
  - bash                ✓ ✓ ✓ . ✓
  - c                   ✓ ✓ ✓ ✓ ✓
  - cmake               ✓ . ✓ ✓ .
  - cpp                 ✓ ✓ ✓ ✓ ✓
  - lua                 ✓ ✓ ✓ ✓ ✓
  - markdown            ✓ . ✓ ✓ ✓
  - markdown_inline     ✓ . . . ✓
  - python              ✓ ✓ ✓ ✓ ✓
  - query               ✓ ✓ ✓ ✓ ✓
[...]

PR #6745, opened 2024-06-08
Clear env when invoking git commands.

mathieulj:
If neovim is started during a git commit, the GIT environment variables will point to the host index and attempting git operations without clearing the environment can corrupt the parent index.

We weren't setting env on the git commands, setting them to an empty table instructs libuv to clear the environment.

Resolves https://github.com/nvim-treesitter/nvim-treesitter/issues/5353

### DIFF
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -1,5 +1,6 @@
 local fn = vim.fn
 local utils = require "nvim-treesitter.utils"
+local luv = vim.loop
 
 -- Convert path for cmd.exe on Windows.
 -- This is needed when vim.opt.shellslash is in use.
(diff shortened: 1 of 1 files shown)


clason:
I'm afraid that this will simply break other uses that _rely_ on environment variables. Can you explain what the actual issue here is? I'm very inclined right now just to drop `git` support for 1.0 as this was never intended as a full alternative to `curl` and `tar`.

mathieulj:
My apologies, I added more context [on the existing issue's comments](https://github.com/nvim-treesitter/nvim-treesitter/issues/5353#issuecomment-2156098296) but I should have added more here as well.

The root of the issue is that git will a) set environment variables when an editor is spawned to edit the commit message and b) override defaults when it sees those environment variables. When neovim was spawned by git, attempting to use git to download a plugin can corrupt the index of the repository being commited to. Since the environment variables are telling git to use that index while checking out the plugin's commit, it's effectively modifying the wrong index. Downloading the git plugin fails because the incorrect index doesn't contain the requested commit hash:

image


While we could make this change more focused, we should remove `GIT_INDEX_FILE` and `GIT_DIR`. Anything set by the parent process is guaranteed to be invalid for the plugin clone/checkout since we are executing git with the paths we intend in the arguments list (different from the parent directory).

> I'm very inclined right now just to drop git support for 1.0 as this was never intended as a full alternative to curl and tar.

Removing this feature entirely would indeed be an alternate path forward. I don't think it should remain in a broken state but I'd be happy with either path being taken.

clason:
If there are a small handful of variables we can explicitly clear that would 1. fix this and 2. be sure not to break valid use, let's go with that for now.

mathieulj:
I've updated the implementation to selectively exclude the git location environment variables since those are guaranteed to be invalid for the plugin fetch. 

Since libuv will clear environment variables when any are passed in, we have to actually map through and manually filter env variables if we want just a subset of the inherited variables.

mathieulj:
Also, going back to you comment about a rewrite on `main` (rather than `master`). It looks like the same issue is still present there. I can put up another PR targeting main with these same changes if you decide to keep git fetching.

clason:
Ok, that is more effort than I expected (and am comfortable with). Let me think about it; I'm leaning toward deprecating the git alternative.

clason:
Going with deprecating  `git` (which is already done on `main` so will be gone with 1.0). Thanks, though!

PR #6753, opened 2024-06-10
fix: declare fields of TSConfig optional

Lattay:
Hi,
I propose to define all the fields of `TSConfig` as nillable to allow the user to not declare some of them without the LS freaking out.
Indeed, with the current trunk lua_ls displays a big diagnostics on the whole setup argument that hides all other possible information (more interesting diagnostics and syntactic coloration).


clason:
Thank you, but that code is essentially frozen. All future development happens on the main branch.

clason:
No worries, it's an awkward situation. Your change makes sense but would amend rebasing harder (and the warnings are welcome for pushing people to migrate ;))

PR #6757, opened 2024-06-11
Remove incorrect builtin highlight for Gleam

lpil:
Hello!

I'm the creator and lead developer of Gleam. Neovim has been highlighting Gleam types slightly incorrectly, and this has been identified as the issue.

I'm not very familiar with this configuration, so please do let me know if this is not the right way to correct this.

Thanks,
Louis

### DIFF
--- a/queries/gleam/highlights.scm
+++ b/queries/gleam/highlights.scm
@@ -147,9 +147,6 @@
   (type_var)
 ] @type
 
-((type_identifier) @type.builtin
-  (#any-of? @type.builtin "Int" "Float" "String" "List"))
-
 ; Type Qualifiers
 [
   "const"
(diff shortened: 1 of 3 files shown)


clason:
Are these not builtin types? We mostly care about correct syntax groupings; the actual colors are handled by colorschemes.

lpil:
There are built in types with those names but it is not possible to tell if it is referring to them by only looking at the name, and that is not all the built in types. To identify built in types the grammar would need to perform some simple type inference and keep track of the type scope for the module in question.

lpil:
Good morning folks! Do let me know if there's anything else I can do to help out here. Thank you.

ObserverOfTime (approved):

ribru17 (approved):

PR #6761, opened 2024-06-13
feat(install)!: drop support for git

clason:
Problem: Using git for installing parsers can lead to data loss if in a
git commit buffer.

Solution: Only support downloading via curl+tar, which are installed on
all supported platforms (since Windows 10). Curl will also be required
for WASM parsers (and for `vim.net.download()`).

### DIFF
--- a/lua/nvim-treesitter/health.lua
+++ b/lua/nvim-treesitter/health.lua
@@ -67,20 +67,19 @@ local function install_health()
     end
   end
 
-  do -- curl+tar or git check
+  do -- curl+tar check
     local curl = check_exe('curl')
-    local tar = check_exe('tar')
-
-    if curl and tar and vim.uv.os_uname().sysname ~= 'Windows_NT' then
-      health.ok(string.format('tar %s (%s)', tar.version, tar.path))
+    if curl then
       health.ok(string.format('curl %s (%s)\n%s', curl.version, curl.path, curl.out))
     else
-      local git = check_exe('git')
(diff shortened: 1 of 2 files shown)


theHamsta:
The reason for prefering git on Windows were symlinks present in some of the repos representing multiple languages. I think it was only typescript/tsx and Ocaml/Ocaml interface. Windows doesn't allow symlinks unless enabled via Developer mode which leadings to failure when untarring the typescript tarball while it is somehow handled by git.

Could we remove symlinks from those repos via pull requests to avoid problems for Windows users?

clason:
Thank you for adding this context. Yes, symlinks should be removed from repos, for this and other reasons.

Typescript/tsx has already removed the symlinks; just ocaml has been... recalcitrant. Both are tree-sitter projects so @amaanq should be able to have a word. I'd rather drop parsers than cater for deliberately non-standard projects.

clason:
Anybody on Windows can do a quick `:TSInstall! all` from this branch and list the failing parsers?

clason:
Actually, `ocaml` seems to be fine, but the following are not:

* tree-sitter-twig
* tree-sitter-roc
* tree-sitter-rbs
* tree-sitter-qmljs
* tree-sitter-janet-simple
* tree-sitter-hyprlang
* tree-sitter-scfg (sr.ht, so this could just be a brownout)

clason:
All parsers now install on Windows without using git \o/

PR #6775, opened 2024-06-15
feat(systemverilog): Add systemverilog support

zhangwwpeng:
Parse taken from https://github.com/gmlarumbe/tree-sitter-systemverilog so people can install it with nvim-treesitter

Upstream for the queries remains https://github.com/gmlarumbe/tree-sitter-systemverilog


queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,528 @@
...
+(string_literal
+  (quoted_string) @string)
+
+(include_compiler_directive) @keyword.import
+
+(include_compiler_directive
+  (quoted_string) @string.special.path)
+
+(seq_block
+  (simple_identifier) @labal)
```
clason, line 274:
```suggestion
  (simple_identifier) @label)
```

Typo (also below).

clason:
Please run `./scripts/format-queries.lua queries/systemverilog`.

nvim-treesitter
clason, line 1: This has to go.

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,528 @@
...
+  "else"
+  (case_keyword)
+  "endcase"
+] @keyword.conditional
+
+[
+ "?"
+ ":"
+ "#"
+] @keyword.conditional.ternary
```
ribru17, line 121: Would you mind only applying these changes to children of `(conditional_expression)`? At least for `":"`, it should be highlighted as `@punctuation.delimiter` broadly (and then get the ternary highlight only when it is part of a ternary statement)

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,528 @@
...
+  "+"
+  "/"
+  "*"
+  "^"
+  "&"
+  "|"
+  "&&"
+  "||"
+  "{"
+  "}"
```
ribru17, line 135: I think these two brackets should be `@punctuation.bracket` (maybe also the one with an apostrophe below it?)

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,528 @@
...
+
+(severity_system_task) @function.builtin
+
+(string_literal
+  (quoted_string) @string)
+
+(include_compiler_directive) @keyword.import
+
+(include_compiler_directive
+  (quoted_string) @string.special.path)
```
ribru17, line 270: I think `(system_lib_string)` should also be handled (probably just as `@string`)

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,528 @@
...
+
+(attribute_instance
+  (attr_spec
+    (simple_identifier) @property))
+
+[
+  (integral_number)
+  (unbased_unsized_literal)
+  (fixed_point_number)
+] @number
```
ribru17, line 243: I think `(unsigned_number)` should be here too. Also `(time_unit)` should be highlighted as something (CSS has units as `@string`, but maybe something like `@type.builtin` would work better? your call)

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,528 @@
...
+  "release"
+  "timeunit"
+  "timeprecision"
+  "sequence"
+  "endsequence"
+  "property"
+  "endproperty"
+  "clocking"
+  "endclocking"
+  "iff"
```
ribru17, line 58: Should this be `@keyword.conditional` or no?

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,527 @@
...
+(port
+  (simple_identifier) @variable)
+
+(list_of_port_identifiers
+  (simple_identifier) @variable)
+
+(net_lvalue
+  (simple_identifier) @variable)
+
+(clocking_event) @attribute
```
ribru17, line 221: Is this only given the attribute highlight for the `@` sign? If so could you just replace this with `"@" @attribute`? Seems the `@` character is not highlighted in something like `always @(*) begin ...`

nvim-treesitter
clason, line 1: Please remove the submodule.

clason (changes requested):

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,528 @@
...
+"return" @keyword.return
+
+[
+  (always_keyword)
+  "for"
+  "foreach"
+  "repeat"
+  "forever"
+  "initial"
+  "while"
```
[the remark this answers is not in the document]
zhangwwpeng:
```
(loop_statement
  [
    "begin"
    "end"
  ] @keyword.repeat)

```
like this?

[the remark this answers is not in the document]
ribru17: Yes! :)

zhangwwpeng:
Most submodules have been fixed, but some submodules are limited by parse

clason:
All (git) submodules have to be removed; this is non-negotiable.

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,565 @@
...
+  name: (simple_identifier) @function)
+
+(function_prototype
+  name: (simple_identifier) @function)
+
+(type_assignment
+  name: (simple_identifier) @type.definition)
+
+(enum_name_declaration
+  (simple_identifier) @type.parameter)
```
clason, line 499: This is not a legal capture; you can only use the ones listed in CONTRIBUTING.md.

clason:
To be honest, I don't have a high degree of confidence in a parser without any tests or CI...

zhangwwpeng:
Thank you for your patience. I am not familiar with git, but tree-sitter-verilog has many bugs. After my test, this parse is much more stable than tree-sitter-verilog.

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,562 @@
...
+(string_literal
+  (quoted_string) @string)
+
+; include
+(include_statement
+  (file_path_spec) @string.special.path)
+
+; directive
+; TODO:
+; Parse cannot distinguish `define `include etc,
```
ribru17, line 528: Actually I think this is not true: \`define is captured as the `"directive_define"` anonymous node and \`include is `"directive_include"`. Would you mind removing the `(include_compiler_directive)` and `(text_macro_usage)` nodes form below and use the above anonymous nodes to highlight just the keywords themselves (as `@keyword.directive.define` and `@keyword.import`, respectively)?

queries/systemverilog/folds.scm
```diff
@@ -0,0 +1,20 @@
...
+  (config_declaration)
+  (covergroup_declaration)
+  (clocking_declaration)
+  (interface_declaration)
+  (module_declaration)
+  (package_declaration)
+  (module_instantiation)
+  (program_declaration)
+  (specify_block)
+  (generate_region)
```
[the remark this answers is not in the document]
ribru17: Sorry, I think you may have forgotten the `+` after the nodes here. This is intentional so that, e.g., multiple include statements can be folded

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,562 @@
...
+  (conditional_compilation_directive)
+  (undefine_compiler_directive)
+  (undefineall_compiler_directive)
+  (text_macro_usage)
+] @keyword.directive
+
+(include_compiler_directive
+  [
+    (quoted_string)
+    (system_lib_string)
```
ribru17, line 544: I think this one should just be `@string`, not `@string.special.path` since it is not a file path

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,562 @@
...
+  (integral_number)
+  (unbased_unsized_literal)
+  (fixed_point_number)
+  (unsigned_number)
+] @number
+
+[
+  (net_type)
+  (data_type)
+  (timescale_compiler_directive)
```
ribru17, line 229: This node is also captured earlier as `@attribute`. I think both these nodes should be removed, with `"directive_timescale"` captured as `@keyword.directive`, and `(time_unit)` as `@type.builtin`

ribru17:
Thanks for your work, I have left a few more comments and resolved the appropriate comments above (some still need a change)

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,562 @@
...
+  "|->"
+  ">>>"
+  "<<<"
+  "->>"
+  "->"
+  "=>"
+  "*>"
+  ".*"
+  "?"
+  ":"
```
ribru17, line 153: This should be `@punctuation.delimiter`. You can add another rule:
```scm
(conditional_expression
  [
    "?"
    ":"
  ] @keyword.conditional.ternary)
```

queries/systemverilog/injections.scm
```diff
@@ -0,0 +1,2 @@
+((comment) @injection.content
+  (#set! injection.language "comment"))
```
ribru17, line 2: Maybe we should add
```scm
((macro_text) @injection.content
  (#set! injection.language "systemverilog"))
```
for macro definition highlights

queries/systemverilog/highlights.scm
```diff
@@ -0,0 +1,562 @@
...
+  "virtual"
+  "extends"
+  "implements"
+  "super"
+  (class_item_qualifier)
+  "parameter"
+  "localparam"
+  "defparam"
+  "assign"
+  "typedef"
```
ribru17, line 27: Maybe this should be `@keyword.type`

queries/systemverilog/injections.scm
zhangwwpeng: great idea

queries/systemverilog/highlights.scm
zhangwwpeng: all done

queries/systemverilog/folds.scm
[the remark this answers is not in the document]
ribru17: Also after `(package_import_declaration)` please :smile:

ribru17 (approved):

queries/systemverilog/folds.scm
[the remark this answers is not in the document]
zhangwwpeng: (include_compiler_directive) and (package_import_declaration) are not the same. When I add (package_import_declaration)+, it does not pass CI. When I test locally, their generated tree-sitter structures are different.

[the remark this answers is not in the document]
ribru17: Ah, I see. In that case maybe you can just remove that one then. Sorry about that - thanks for checking!

clason:
@zhangwwpeng your repo fails to build with `tree-sitter build` (which is required going forward). I also noticed that it contains 200Mb worth of test, and an 80Mb `parser.c` (which takes several minute to generate) -- I don't think that is reasonable to ask for people to download every time.

I would strongly encourage (in fact, ask) you to spend some effort on the grammar to minimize it.

clason:
Just to give some idea what I'm talking about, here's the timings from a current CI run:
```
10. 189.02ms hlsl highlights
9. 226.60ms slang highlights
8. 231.24ms swift indents
7. 235.60ms cuda highlights
6. 237.15ms perl highlights
5. 287.58ms haskell highlights
4. 295.67ms agda highlights
3. 369.34ms swift highlights
2. 493.11ms nim highlights
1. 820.02ms systemverilog highlights
```

This means just opening a systemverilog file (just loading the queries, not even parsing!) takes almost a second, which is nearly twice as slow as the previous "frontrunner". I don't think people will find that acceptable editing experience (and we will get the complaints about it...)

zhangwwpeng:
Thank you. I originally planned to write indent.scm. Now I plan to learn how to optimize parse. This is a big project for me and will take a long time in the future, but I will stick to it. (Although the current parese performance is very weak, it is good for Verilog users)

clason:
At least merge the PR soon, then.

henrykvdb:
Hi all,

I have some comments related to this PR and I wanted to hear you guy's opinions before I make a new PR with my proposed change.

Before this PR we used the same parser (https://github.com/tree-sitter/tree-sitter-verilog) for SystemVerilog and Verilog. This makes sense since Verilog is essentially a subset of SystemVerilog;

In this PR zhangwwpeng added a second parser (https://github.com/zhangwwpeng/tree-sitter-systemverilog) for SystemVerilog but not for Verilog. From my personal experience/opinion this one seems to work a lot better, seems closer to the relevant standard (LRM) and is better tested.

I think it would be sensible to use this new parser for both languages. Then zhangwwpeng's personal fork could also be replaced by the maintained upstream repository.
As far as I can tell the fork only exists to rename the grammar (from 'Verilog' to 'SystemVerilog', which would be no longer necessary) and to build to compile the parser.c file (the upstream author added a compiled version in the meantime).

@clason  With regards to the query loading slowdown, do you have any indications as to what could be causing that? Are there any helpful commands to help debug this further? Is it purely because the grammar is rather big? Personally I lazy load treesitter (hence it does not bother me much), but clearly this is not a general solution :)

clason:
> @clason With regards to the query loading slowdown, do you have any indications as to what could be causing that? Are there any helpful commands to help debug this further? Is it purely because the grammar is rather big? Personally I lazy load treesitter (hence it does not bother me much), but clearly this is not a general solution :)


Yes, it's because the grammar is _massive_. This is a regression over the already huge and slow `verilog` grammar. I'm sure it's possible to bring that down, but there's no silver bullet; some good resources can be found here: https://github.com/nvim-treesitter/nvim-treesitter/wiki/Parser-Development

And if the parser (and queries) work better for verilog, I'm happy to switch; just make a PR and give some objective examples _why_.

henrykvdb:
For the old highlighter; It really seemed to struggle with parsing even simple programs so I was using it with highlighting off (before switching).

Quickly wrote up a the most simple verilog program I could come up with (tested against Icarus v10.0 simulator). Program consists of a function that assigns it's input value to it's output (using bit slicing). Even for this the :InspectTree already contains errors  😅 

old-parser

I can submit a PR in the coming week (including the necessary changes to the textobjects) removing the old parser and switching over to the upstream repo of the new one (which is still receiving minor updates every once in a while)

---

For the performance; feared it would indeed be the grammar size, SystemVerilog is quite an complex language with lots of keywords, optional syntax and edge cases (there is a >1300p document detailing all of it called the LRM, not too enjoyable to read). It's not very suprising to me that sticking closer to the spec yields a bigger grammar, so I am not sure if there are many easy optimizations left to make (I never wrote a parser).
The author told me he uses Emacs and that he did not notice a performance regression compared to the old parser (besides in initial compile time), not sure if that is helpful info?

clason:
Not really; I trust my benchmarks more than Emacs very early days anecdotes ;)

PR #6779, opened 2024-06-17
fix(apex): highlights: change the constant variable order to recognize it correctly

asknet:
This changes the order of how the constant variable is recognized. Without this, it applies property highlights to constant types. 

Kindly review & share the feedback.


lucario387 (approved):
Makes sense. Thanks for the PR

PR #5260, opened 2023-08-16
feat(janet): update highlights

sogaiu:
This mostly includes newer constructs (e.g. `catseq`, `has-key?`, etc.).

There's some minor reformatting as well.

PR #6789, opened 2024-06-19
feat(janet): update highlights

sogaiu:
This mostly includes newer constructs since the last update, which I think may have been #5260.

_Update_: Changed the following.

~There's some reformatting as well.  The grouping of items is now mostly programmatically generated and formatted to fit a bit more nicely for horizontally-challenged situations.~

I used the `./scripts/format-queries.lua` tool to format the file as quoted [below](https://github.com/nvim-treesitter/nvim-treesitter/pull/6789#issuecomment-2178697306).


sogaiu:
Ah, why is it not recommended?

sogaiu:
Thanks a lot for the explanation!

Much appreciated.

For this time, I've used the recommended tool for formatting.

I'm not that keen on going over 80 columns, but this file is not too big and if the tool produces results that help with maintenance for everyone, I think that works for me.

lucario387:
As you are the maintainer of the language queries, I'll vote to do what is best/easiest for you to manage the queries :)

If the maintainer cannot track it, so can't we.

PR #6803, opened 2024-06-24
improve statusline() implementation

### LINKED ISSUE #1931: Improve statusline / parsing for multiline node declarations
I'm working in a C++ codebase that often makes multiline node declarations, such as
```
void
MavlinkParametersManager::handle_message(const mavlink_message_t *msg)
{
```
and have found that this results in `M.statusline` just returning `"void"` for that node.

I don't understand treesitter in detail, but my first idea was to increase how many lines `M.statusline` uses for each node and pair down the result via `:gsub`. That worked ok, but obviously it's quite hacky and prone to a few undesired behaviors. E.g. if your cursor is over `MavlinkParametersManager`, treesitter seems to think both the first and second lines are nodes, so `M.statusline` returns
```
"void MavlinkParametersManager::handle_message(const mavlink_message_t *msg) -> MavlinkParametersManager::handle_message(const mavlink_message_t *msg)"
```

Next, I looked for some kind of node name property to use instead, but I'm not sure if there is one.

It's unclear to me if this is a shortcoming of treesitter or this plugin, but I looking for advice on how to help improve this functionality.

Thanks!

QikLiang:
Fixes #1931 
This PR contains three changes.
1. expand default `opt.type_patterns`: add `struct` and `namespace` as scopes.
2. Handle bug from false-positive node type:
Consider the following C++ code:
```
int f(int a, int b){
    return 1;
}
```
The above produces following `:InspectTree` output:
```
(function_definition ; [0, 0] - [2, 1]
  type: (primitive_type) ; [0, 0] - [0, 3]
  declarator: (function_declarator ; [0, 4] - [0, 12]
    declarator: (identifier) ; [0, 4] - [0, 5]
    parameters: (parameter_list ; [0, 5] - [0, 12]
      (parameter_declaration ; [0, 6] - [0, 11]
        type: (primitive_type) ; [0, 6] - [0, 9]
        declarator: (identifier)))) ; [0, 10] - [0, 11]
  body: (compound_statement ; [0, 12] - [2, 1]
    (return_statement ; [1, 4] - [1, 13]
      (number_literal)))) ; [1, 11] - [1, 12]
```
When the cursor is on the symbol `a`, the output of `nvim_treesitter#statusline()` is `int f(int a) -> f(int a)`. Looking at the tree, it's evident that this is because `opt.type_patterns` ends up matching both `function_definition` and `function_declarator`. When both a node and its parent matches the `type_pattern`, merge their text output.

3. More robust way of extracting signature from a node: The current implementation uses the first line in the range of a node as the signature of that node, which causes the problem demonstrated in #1931 
```
int
foo(very_long_parameter_type_that_takes_up_a_whole_line const& param){
    // statusline() here returns "int"
}
```
This new implementation extracts signature from a node by iterating over its children, filtering out the part that we don't want, and concatenating the text of the other nodes. The default option is to ignore the field named "body", which seems to work for C++, java, Python, C#, rust, lua, and is sufficiently flexible to customization. I also looked at using `@locals.scope` for this purpose, but the scope generally tags the function node instead of the function body node to include parameters.
Removed `@private` function `ts_utils._get_line_for_node()` which is currently only used by `statusline()`, but I didn't touch the `@deprecated` function `ts_utils.get_node_text()`.

### DIFF
--- a/doc/nvim-treesitter.txt
+++ b/doc/nvim-treesitter.txt
@@ -549,8 +549,9 @@ Default options (lua syntax):
 >
   {
     indicator_size = 100,
-    type_patterns = {'class', 'function', 'method'},
-    transform_fn = function(line, _node) return line:gsub('%s*[%[%(%{]*%s*$', '') end,
+    type_patterns = {'namespace', 'class', 'struct', 'function', 'method'},
+    ignore_fields = {'body'}
+    transform_fn = function(line, _node) return line:gsub('%s*[%[%(%{]*%s*$', ''):gsub('%s+', ' ') end,
     separator = ' -> ',
     allow_duplicates = false
   }
@@ -558,8 +559,11 @@ Default options (lua syntax):
 - `indicator_size` - How long should the string be. If longer, it is cut from
   the beginning.
 - `type_patterns` - Which node type patterns to match.
+- `ignore_fields` - for a node that matches the type pattern, which fields to
+  ignore to extract a signature
 - `transform_fn` - Function used to transform the single item in line. By
-  default it removes opening brackets and spaces from end. Takes two arguments:
+  default it removes opening brackets and spaces from end, and replace
+  consecutive whitespaces with a single space. Takes two arguments:
   the text of the line in question, and the corresponding treesitter node.
 - `separator` - Separator between nodes.
 - `allow_duplicates` - Whether or not to remove duplicate components.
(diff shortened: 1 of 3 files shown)


clason:
Thank you, but this code (and functionality) will be completely removed in the next version. Nvim-treesitter is focusing on installing parsers and curating queries. Feel free to create and maintain your own plugin using this code if you find it useful!

PR #6810, opened 2024-06-26
Feat: Pro(b)log support

foxyseta:
Added Prolog and ProbLog parsers and queries. The two languages do not have proper scopes or injections, as far as I know (used the ISO standard to implement the parsers), so those queries are intentionally missing.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1564,6 +1564,24 @@ list.prisma = {
   maintainers = { "@elianiva" },
 }
 
+list.problog = {
+  install_info = {
+    url = "https://codeberg.org/foxy/tree-sitter-prolog",
+    files = { "src/parser.c" },
+    location = "grammars/problog",
+  },
+  maintainers = { "@foxyseta" },
+}
+
+list.prolog = {
+  install_info = {
+    url = "https://codeberg.org/foxy/tree-sitter-prolog",
+    files = { "src/parser.c" },
+    location = "grammars/prolog",
+  },
+  maintainers = { "@foxyseta" },
+}
+
 list.promql = {
   install_info = {
     url = "https://github.com/MichaHoffmann/tree-sitter-promql",
--- a/queries/prolog/highlights.scm
+++ b/queries/prolog/highlights.scm
@@ -1,0 +1,43 @@
+(comment) @comment @spell
+
+(atom) @constant
+
+((atom) @boolean
+  (#any-of? @boolean "true" "false"))
+
+(functional_notation
+  function: (atom) @function.call)
+
+(integer) @number
+
+(float_number) @number.float
+
+(directive_head) @operator
+
+(operator_notation
+  operator: _ @operator)
+
+[
+  (open)
+  (open_ct)
+  (close)
+  (open_list)
+  "|"
+  (close_list)
+  (open_curly)
+  (close_curly)
+] @punctuation.bracket
+
+[
+  (arg_list_separator)
+  (comma)
+  (end)
+  (list_notation_separator)
+] @punctuation.delimiter
+
+(operator_notation
+  operator: (semicolon) @punctuation.delimiter)
+
+(double_quoted_list_notation) @string
+
+(variable_term) @variable
--- a/README.md
+++ b/README.md
@@ -350,6 +350,8 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [pony](https://github.com/amaanq/tree-sitter-pony) (maintained by @amaanq, @mfelsche)
 - [x] [printf](https://github.com/ObserverOfTime/tree-sitter-printf) (maintained by @ObserverOfTime)
 - [x] [prisma](https://github.com/victorhqc/tree-sitter-prisma) (maintained by @elianiva)
+- [x] [problog](https://codeberg.org/foxy/tree-sitter-prolog) (maintained by @foxyseta)
+- [x] [prolog](https://codeberg.org/foxy/tree-sitter-prolog) (maintained by @foxyseta)
 - [x] [promql](https://github.com/MichaHoffmann/tree-sitter-promql) (maintained by @MichaHoffmann)
 - [x] [properties](https://github.com/tree-sitter-grammars/tree-sitter-properties) (maintained by @ObserverOfTime)
 - [x] [proto](https://github.com/treywood/tree-sitter-proto) (maintained by @treywood)
--- a/lockfile.json
+++ b/lockfile.json
@@ -530,6 +530,12 @@
   "prisma": {
     "revision": "eca2596a355b1a9952b4f80f8f9caed300a272b5"
   },
+  "problog": {
+    "revision": "d8bc22c007825d3af3d62b4326f9d8f9ca529974"
+  },
+  "prolog": {
+    "revision": "d8bc22c007825d3af3d62b4326f9d8f9ca529974"
+  },
   "promql": {
     "revision": "77625d78eebc3ffc44d114a07b2f348dff3061b0"
   },
--- a/queries/problog/folds.scm
+++ b/queries/problog/folds.scm
@@ -1,0 +1,1 @@
+; inherits: prolog
--- a/queries/problog/highlights.scm
+++ b/queries/problog/highlights.scm
@@ -1,0 +1,4 @@
+; inherits: prolog
+
+(probability_label
+  _ @attribute)
--- a/queries/problog/indents.scm
+++ b/queries/problog/indents.scm
@@ -1,0 +1,1 @@
+; inherits: prolog
--- a/queries/problog/injections.scm
+++ b/queries/problog/injections.scm
@@ -1,0 +1,1 @@
+; inherits: prolog
--- a/queries/prolog/folds.scm
+++ b/queries/prolog/folds.scm
@@ -1,0 +1,6 @@
+[
+  (directive_term)
+  (clause_term)
+  (arg_list)
+  (list_notation)
+] @fold
--- a/queries/prolog/indents.scm
+++ b/queries/prolog/indents.scm
@@ -1,0 +1,16 @@
+(directive_term) @indent.zero
+
+(clause_term) @indent.zero
+
+(functional_notation
+  (atom)
+  (open_ct) @indent.begin
+  (close) @indent.end)
+
+(list_notation
+  (open_list) @indent.begin
+  (close_list) @indent.end)
+
+(curly_bracketed_notation
+  (open_curly) @indent.begin
+  (close_curly) @indent.end)
--- a/queries/prolog/injections.scm
+++ b/queries/prolog/injections.scm
@@ -1,0 +1,2 @@
+((comment) @injection.content
+  (#set! injection.language "comment"))


foxyseta:
Not sure how to make sense of the failed pipelines logs.

ribru17:
You need to add lockfile entries for the two parsers (not sure if this will fix all the tests)

foxyseta:
Done!

lua/nvim-treesitter/parsers.lua
```diff
@@ -1564,6 +1564,24 @@ list.prisma = {
   maintainers = { "@elianiva" },
 }
 
+list.problog = {
+  install_info = {
+    url = "https://codeberg.org/foxy/tree-sitter-problog",
+    files = { "src/parser.c" },
+    branch = "feat/problog",
```
clason, line 1571: No need to specify the branch if it's the default.

clason:
Also, we expect to be able to download archives as with Github (cloning and checking out revisions is deprecated for security reasons). Does codeberg support that?

lua/nvim-treesitter/parsers.lua
clason, line 1571:
```suggestion
```

lua/nvim-treesitter/parsers.lua
```diff
@@ -1564,6 +1564,24 @@ list.prisma = {
...
+  },
+  maintainers = { "@foxyseta" },
+}
+
+list.prolog = {
+  install_info = {
+    url = "https://codeberg.org/foxy/tree-sitter-prolog",
+    files = { "src/parser.c" },
+    branch = "main",
```
clason, line 1580:
```suggestion
```

clason:
> Not sure how to make sense of the failed pipelines logs.

Here's your issue: https://codeberg.org/foxy/tree-sitter-problog/src/commit/3687c8579bad69c0143852032f3830363f151312/grammar.js#L306

lua/nvim-treesitter/parsers.lua
foxyseta: You are right, thank you! I got confused with the local install (from init.lua)!

foxyseta:
> > Not sure how to make sense of the failed pipelines logs.
> 
> Here's your issue: https://codeberg.org/foxy/tree-sitter-problog/src/commit/3687c8579bad69c0143852032f3830363f151312/grammar.js#L306

Thanks! Fixed in the parser repo and updated the lockfile accordingly.

clason:
You still might like to consider keeping both grammars in a single repository and inheriting from a common grammar, like tree-sitter-ocaml does.

foxyseta:
> Also, we expect to be able to download archives as with Github (cloning and checking out revisions is deprecated for security reasons). Does codeberg support that?

Yes, you can do stuff like:

https://codeberg.org/foxy/tree-sitter-prolog/archive/main.zip
https://codeberg.org/foxy/tree-sitter-prolog/archive/main.tar.gz

And also

https://codeberg.org/foxy/tree-sitter-prolog/archive/main.bundle,

which is not supported by GitH*b.

foxyseta:
> You still might like to consider keeping both grammars in a single repository and inheriting from a common grammar, like tree-sitter-ocaml does. I think this would be a better strategy.

All in the same branch? Is there any guide for grammar inheritance anywhere?

clason:
> All in the same branch? Is there any guide for grammar inheritance anywhere?

Yes, just different directories. There's no official guidance (yet), but you can look at the upstream ocaml parser as an example.

> Yes, you can do stuff like:

I need `archive/<sha>.tar.gz`, but that probably works as well then.

> which is not supported by GitH*b.

🙄

foxyseta:
> Yes, just different directories. There's no official guidance (yet), but you can look at the upstream ocaml parser as an example.

Okay, thank you! I can work on that soon and then do the relevant changes to this PR once I am ready.  
 
> I need `archive/<sha>.tar.gz`, but that probably works as well then.

Yes, that works too.

foxyseta:
@clason Done! Hopefully I followed the example correctly.

queries/prolog/highlights.scm
```diff
@@ -0,0 +1,59 @@
+(comment) @comment
+
+(atom) @constant
+
+((atom) @boolean
+  (#eq? @boolean "true"))
+
+((atom) @boolean
+  (#eq? @boolean "false"))
```
ObserverOfTime, line 9:
```suggestion
((atom) @boolean
  (#any-of? @boolean "true" "false"))
```

queries/prolog/highlights.scm
```diff
@@ -0,0 +1,59 @@
...
+
+(functional_notation
+  function: (atom) @function.call)
+
+[
+  (integer)
+  (negative_number
+    (integer))
+] @number
```
ObserverOfTime, line 18:
```suggestion
(integer) @number
```

queries/prolog/highlights.scm
```diff
@@ -0,0 +1,59 @@
...
+
+[
+  (float_number)
+  (negative_number
+    (float_number))
+] @number.float
```
ObserverOfTime, line 24:
```suggestion
(float_number) @number.float
```

queries/prolog/highlights.scm
```diff
@@ -0,0 +1,59 @@
+(comment) @comment
```
ObserverOfTime, line 1:
```suggestion
(comment) @comment @spell
```

queries/prolog/highlights.scm
ObserverOfTime: Oh I see.

foxyseta: @lucario387 are you talking about making `-` part of `integer` and removing the `negative_number` non-terminal altogether? Although I do agree that it is very weird to use `integer` to only refer to non-negative integer numbers, the current state of the parser is meant to reflect how the ISO specification defines the language. The currenty query is built keeping in mind we do not want to change the parser.

lucario387: Yes, that's what I meant. It's kindof weird for the ISO spec of a language to specifically denote negative number though. Is there any special usage for it?

foxyseta: I cannot find an explicit motivation anywhere, sorry. Could this has been done for unification/pattern matching, maybe? This way, `delta_is(-1)` can match `delta_is(-x)` for example. In the original specs, negative numbers are specified as:

```
6.3.1.2 Negative numbers
term = name, integer;
term = name, float number;
(Condition: name is "-")
```

So the non-terminal "name" is used, but then a side condition is specified.

I personally think sticking to the language itself is beneficial in this case, but I am still open to rewriting this part of the parser as long as I can document the change with a valid reason.

clason: In general, sticking too closely to the language spec is not a good idea; often a more pragmatic approach that focuses on the end goal (which for us is syntax highlighting) leads to a more robust and performant parser. [(Github says so.)](https://github.com/github/semantic/blob/3baf0b47ff2117a4724771687d458151ce2b92a7/docs/grammar-development-guide.md#rely-on-language-specs-but-not-too-much)

That being said, in the end it's your decision; this is not a critical blocker.

ribru17:
Would you mind adding comment injections please?

queries/prolog/highlights.scm
```diff
@@ -0,0 +1,59 @@
...
+  (operator_1100xfy)
+  (operator_1050xfy)
+  (operator_1000xfy)
+  (operator_900fy)
+  (operator_700xfx)
+  (operator_500yfx)
+  (operator_400yfx)
+  (operator_200xfx)
+  (operator_200xfy)
+  (operator_200fy)
```
ribru17, line 38: I think `":-"` and `"|"` should be added. Also `(comma)` should be given `@punctuation.delimiter` somewhere after this query

queries/prolog/highlights.scm
```diff
@@ -0,0 +1,59 @@
...
+[
+  (float_number)
+  (negative_number
+    (float_number))
+] @number.float
+
+[
+  (operator_1200xfx)
+  (operator_1200fx)
+  (operator_1100xfy)
```
ribru17, line 29: 1100 (semicolon) should be moved to `@punctuation.delimiter`. To be honest I think most or all of the `operator_*` nodes should be prefixed with underscores in the grammar, these node names seem pretty unintuitive :sweat_smile:

foxyseta: Unlike periods, commas and semicolons are considered operators in Prolog, but I can see how thinking about them as punctuation would be more intuitive, so I'll change this.

I thought hiding nodes would prevent me from referring to them in queries?

queries/prolog/highlights.scm
foxyseta: I would maybe give `|` the same highlight group as its braces. Currently, only binary :- is colored as an operator, so I can color unary :- as one, too.

queries/prolog/highlights.scm
foxyseta: I understand! You make a valid point. I'll need some time to make the changes, and then ping you all when I am ready. Thanks.

queries/prolog/highlights.scm
ribru17: That would be a great compromise, thanks! Also if things like semicolons are meant to be operators then it's fine to keep it that way; my mistake :)

foxyseta: Whoopsie I hadn't seen this commit in time and committed already. It's okay, I like it better your way anyway.

foxyseta:
I think I addressed all of the comments yesterday. Let me know if I misunderstood/forgot something!

queries/prolog/highlights.scm
```diff
@@ -0,0 +1,43 @@
...
+  function: (atom) @function.call)
+
+(integer) @number
+
+(float_number) @number.float
+
+(directive_head) @operator
+
+(operator_notation
+  operator: (_) @operator)
```
ribru17, line 18: Thank you, I think this PR is almost good to go! One thing: 
```suggestion
  operator: _ @operator)
```
This way anonymous nodes can be captured. Also, would you mind updating the grammar (and then the lockfile) so that things like:
```js
        prec.left(
          1,
          seq(
            $._term,
            field(fields.operator, /(:\-)|(\-\->)/),
            $._term,
          ),
        ),
```
can be indexed? Sadly regexes become invisible in the grammar and thus not query-able. Something like the following would allow for operators that use regex rules to be highlighted:
```js
        prec.left(
          1,
          seq(
            $._term,
            field(fields.operator, alias(/(:\-)|(\-\->)/, $.arrow)),
            $._term,
          ),
        ),
```
(or you can of course replace `$.arrow` with something quoted if you prefer the node be anonymous, your call)

foxyseta: Should have fixed in https://github.com/nvim-treesitter/nvim-treesitter/pull/6810/commits/d95626829a0cda0c2b002dceb0745d6516ca37db

ribru17 (approved):

clason:
@foxyseta please rebase on master (do not merge master into your branch!) and squash your commits

lucario387:
woah woah, what's happening here, why's there all these commits? Did you forget to drop unrelated commits while interactive rebasing?

foxyseta:
Apparently I did, sorry 😅  It's my first time dealing with unrelated commits during a rebase

clason:
> > Yes, just different directories. There's no official guidance (yet), but you can look at the upstream ocaml parser as an example.
> 
> Okay, thank you! I can work on that soon and then do the relevant changes to this PR once I am ready.
> 
> > I need `archive/<sha>.tar.gz`, but that probably works as well then.
> 
> Yes, that works too.

Not the same way, though. This breaks the refactored installation on the `main` branch (which will become the default branch in the near future). There's a limit how many different codeforges we can (and want to) support.

foxyseta:
I am sorry. Luckily, compatibility with Codeberg means compatibility with any ForgeJo (and maybe Gittea) instance. Are you considering rolling back this PR or adding compatibility to the `main` branch? Is there any way I can help?

lucario387:
Having a GitHub/GitLab mirror of the repo will be nice (and probably required, unless there's enough parsers that's hosted on codeberg)

foxyseta:
Yes, I had considered that option and it is definitely not a problem for me. Here you go:
https://github.com/foxyseta/tree-sitter-prolog. I'll make sure this mirror stays up to date. So discussion for ForgeJo support will open only if enough parsers are hosted on it? I usually push all my code to Codeberg, but I do not mind mirroring to GitHub my future parsers, too.

What precisely are the differences between Codeberg and GitHub breaking this? If it's an objective limitation of ForgeJo, maybe a patch could be submitted. If it's just different design choices between the two, I guess not.

lucario387:
> What precisely are the differences between Codeberg and GitHub breaking this

We had bad experience with Codeberg before, such as flaky CIs due to codeberg being down at certain times. Another reason is different installation support, which `main` branch significantly reduced (on `master` there is support for makefile which was removed for `main` branch, for example).

clason:
I already pushed a compat commit to `main`, but the flakiness of "smaller" codeforges (due to more frequent outages due to smaller resources) is a significant issue for our CI: if a test fails due to not being able to install a parser, our automated actions are blocked. So thank you for the mirror; I'll switch the parser to that.

clason:
@foxyseta that doesn't look like a full mirror; it's missing the problog parser!

The difference is really small: the tarball for a commit (e.g., `https://codeforge.org/foxyseta/tree-sitter-prolog/<hash>.tar.gz`) when unpacked contains a `tree-sitter-prolog` directory with the repo contents, while Github and Gitlab append the hash to it (`tree-sitter-prolog-<hash>`).

foxyseta:
Cool, thank you. Will make sure to also provide mirrors in the future.

PR #6814, opened 2024-06-26
fix(v): use `source_file` as parent node for comment highlight query

ttytm:
This PR fixes a huge performance impairment related to doc comments. This is particularly noticeable with larger files.

To give an example that can be locally tested: with the current highlight query it is barely possible to work in https://github.com/vlang/v/blob/094c30c34785597de7238e06895d153d8d121e4d/vlib/builtin/string.v - even on good machines. With the changes, things work smoothly.

The `source_file` parent node is what e.g. the Go queries use for doc comments too.

### DIFF
--- a/queries/v/highlights.scm
+++ b/queries/v/highlights.scm
@@ -288,8 +288,9 @@
   (block_comment)
 ] @comment @spell
 
-(source_file
+(_
   (line_comment)+ @comment.documentation
+  .
   [
     (function_declaration)
     (type_declaration)
--- a/queries/v/highlights.scm
+++ b/queries/v/highlights.scm
@@ -288,7 +288,7 @@
   (block_comment)
 ] @comment @spell
 
-(_
+(source_file
   (line_comment)+ @comment.documentation
   [
     (function_declaration)


queries/v/highlights.scm
```diff
@@ -288,7 +288,7 @@
   (block_comment)
 ] @comment @spell
 
-(_
```
lucario387, line 291: I don't think this is the correct fix. doc comments are comments that is followed by a func/type/enum right?

The fix should be
```diff
(_
  (line_comment)+ @comment.documentation
+ .
  [
    (function_declaration)
    (type_declaration)
    (enum_declaration)
  ])
```

lucario387:
This doesn't look like it solves anything, it just restricts to top-level comments for captures.


Something like
```
comments
comments
......
comments
type
function
type
enum

comments
comments
.....
comments
function
enum
enum
```

With this kind of top-level nodes, the first bunch of comments is captured 7 times, each for every type/function/enum node. The second bnunch of comments is 3 times. In reality, they should only be captured once each

queries/v/highlights.scm
lucario387: A different fix:
```diff
(_
+ (line_comment) @comment.documentation
+ .
  (line_comment)* @comment.documentation
+ .
  [
    (function_declaration)
    (type_declaration)
    (enum_declaration)
  ])
```

ttytm:
Right `@comment.documentation` is not recognized correctly with the changes.

> This doesn't look like it solves anything

Wrong. The performance change is worlds of a difference.

queries/v/highlights.scm
clason: We can start thinking about dropping 0.9 compatibility, although I usually prefer to wait for a 0.x.1  that has ironed out most of the rough bits of a major (zerover) release.

lucario387:
>> This doesn't look like it solves anything
> Wrong. The performance change is worlds of a difference.

The performance change is because the query now captures less nodes, not because it's more correct. 

From what I can see, multiple lines of comments adjacent to a function/type suggests it's the documentation for such rhs.
What about type/enum declaration inside functions? Should the documentation highlights for those be gone because of this change?

ttytm:
> The performance change is because the query now captures less nodes, not because it's more correct.

Yes. Didn't said it's the correct fix, just said it fixes something.

---

Using the prior  `_` in combination with the suggested `.` helps with performance while still querying doc comments.

lucario387:
> Using the prior _ in combination with the suggested . helps with performance while still querying doc comments.

So is the proposed fix now works for large files with lots of comments? If there's some test file for that, it'd be nice

ttytm:
> > Using the prior _ in combination with the suggested . helps with performance while still querying doc comments.
> 
> So is the proposed fix now works for large files with lots of comments? If there's some test file for that, it'd be nice

@lucario387 I linked a file in the PR description when opening it that can be used for local reproduction and verification.

If there are programmatically run CI tests, can you please share a link? Since I'm not too deep into (nvim-)treesitter development, I would have to see and base it of other test implementations first. Otherwise, with my current time constraints, it won't be possible for me to look for what a good implementation for a test would be.

PR #6826, opened 2024-06-28
fix(r): adapt indent queries

lucario387:
- Anchor parameter query. There's no need for multiple indent.align captures
- Narrow down binary_operator indent.begin. From tree-sitter-r corpus and highlight tests, this seems to be appropriate.


clason:
Thanks!

PR #6830, opened 2024-06-29
feat: Robots.txt support

opa-oz:
Hello, it's me again 🙇

Brought to you lang support for [robots.txt](https://en.wikipedia.org/wiki/Robots.txt)

**Before**
Screenshot 2024-06-29 at 15 44 54

**After**
Screenshot 2024-06-29 at 15 44 31

- [x] `tree-sitter build`
Screenshot 2024-06-29 at 15 45 47

- [x] `TSInstallFromGrammar robots`
Screenshot 2024-06-29 at 15 46 07

- [x] Alphabetically sorted
- [x] Injections & highlights in place

### DIFF
--- a/queries/robots/highlights.scm
+++ b/queries/robots/highlights.scm
@@ -1,0 +1,9 @@
+(comment) @comment @spell
+
+(directive) @constant
+
+(unknownDirective) @attribute
+
+(value) @string
+
+":" @punctuation.delimiter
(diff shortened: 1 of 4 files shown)


queries/robots/highlights.scm
```diff
@@ -0,0 +1,9 @@
+(comment) @comment @spell
+
+(directive) @constant
+
+(unknownDirective) @attribute
```
ObserverOfTime, line 5: I think that either both should be `@property`, or `(unknownDirective)` should not be highlighted by default.

opa-oz: Done

ObserverOfTime (approved):

PR #6832, opened 2024-06-30
Add folds for elm.

degustaf:


queries/elm/folds.scm
```diff
@@ -0,0 +1,13 @@
...
+
+[
+  (case_of_branch)
+  (case_of_expr)
+  (value_declaration)
+  (type_declaration)
+  (type_alias_declaration)
+  (list_expr)
+  (record_expr)
+  (parenthesized_expr)
```
ribru17, line 12: Can we please also get `(import_clause)+`?

queries/elm/folds.scm
```diff
@@ -0,0 +1,13 @@
+((function_call_expr) @_fn
+  (#not-has-parent? @_fn parenthesized_expr)) @fold
```
ribru17, line 2: Out of curiosity, why no `parenthesized_expr` parents here?

degustaf: Elm-format will place the closing paren on the following line. So with just the function folded it looks like
```
+ fold
    )
```
This seemed like it was worth the extra work to remove that case and just fold on the `parenthesized_expr`.

queries/elm/folds.scm
degustaf: Added

PR #6846, opened 2024-07-02
fix(ecma): indent single-line if statements

### LINKED ISSUE #5898: Incorrect indentation for javascript/typescript single-line `if` statement bodies
### Describe the bug

nvim-treesitter incorrectly indents the body of `if` statements (and also `else`, `for`, `while`, etc.) if it is not wrapped in curly braces.

### To Reproduce

1. Create a typescript file.
2. Paste the following code
```typescript
function foo() {
  if (true)
    return 0;
}
```
3. Re-indent the 3rd line (`return`) with `==`. Code becomes
```typescript
function foo() {
  if (true)
  return 0;
}
```

### Expected behavior

Code stays unchanged.

```markdown

Installation ~
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
- OK `node` found v21.5.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
  Version: cc (GCC) 13.2.1 20230801

OS Info:
{
  machine = "x86_64",
  release = "6.6.11-1-lts",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Wed, 10 Jan 2024 16:41:56 +0000"
} ~

  - tsx                 ✓ ✓ ✓ ✓ ✓

```

```text
NVIM v0.10.0-dev-2086+g965dbd0e0
Build type: RelWithDebInfo
LuaJIT 2.1.1703358377
```

### Additional context

_No response_

ribru17:
Fixes #5898. Currently only works after-the-fact, i.e. indentation will not work with `o`.

### DIFF
--- a/queries/ecma/indents.scm
+++ b/queries/ecma/indents.scm
@@ -59,10 +59,11 @@
 (statement_block
   "{" @indent.branch)
 
-(parenthesized_expression
-  ("("
-    (_)
-    ")" @indent.end))
+((parenthesized_expression
+  "("
+  (_)
+  ")" @indent.end) @_outer
+  (#not-has-parent? @_outer if_statement))
 
 [
   "}"
@@ -75,3 +76,7 @@
   (comment)
   (ERROR)
 ] @indent.auto
+
+(if_statement
+  consequence: (_) @indent.dedent
+  (#not-kind-eq? @indent.dedent statement_block)) @indent.begin
--- a/tests/indent/ecma/if_else.js
+++ b/tests/indent/ecma/if_else.js
@@ -11,3 +11,9 @@ if (cond1) {
 } else {
   do_fallback()
 }
+
+if (true)
+  console.log('hi')
+console.log('hi')
+
+if (true)
--- a/tests/indent/javascript_spec.lua
+++ b/tests/indent/javascript_spec.lua
@@ -75,6 +75,8 @@ describe("indent JavaScript:", function()
       { 9, 2 },
       { 12, 2 },
       { 13, 0 },
+      { 16, 0 },
+      { 19, 2 },
     } do
       run:new_line("ecma/if_else.js", { on_line = info[1], text = "hello()", indent = info[2] }, info[3], info[4])
     end


lucario387:
https://github.com/nvim-treesitter/nvim-treesitter/blob/a80fe081b4c5890980561e0de2458f64aaffbfc7/queries/c/indents.scm#L36-L56

What about something like this

ribru17:
It's so beautiful... :smiling_face_with_tear: 

I have added dedent on the end node but the `o` indentation still doesn't work if followed by nothing:
```js
function foo() {
  if (true)
  // no indent here after "o"
}
```
Is there an easy solution to this? Every other case works

lucario387:
> Even with only (if_statement) @indent.begin the empty statement case still doesn't indent :( 

indent.immediate, probably

ribru17:
I tried that also, still nothing somehow. Bright side is this PR works for most edge cases but the last one is really bothering me lol

ribru17:
Found the culprit. Removing this query allows it to work:
```scm
(parenthesized_expression
  ("("
    (_)
    ")" @indent.end))
```
The `end` indentation carries over...

ribru17:
Yes there is :muscle: Will add tests later need to go to bed

queries/ecma/indents.scm
```diff
@@ -75,3 +76,15 @@
...
+  consequence: (_) @indent.dedent
+  (#not-kind-eq? @indent.dedent statement_block)) @indent.begin
+
+(while_statement
+  body: (_) @indent.dedent
+  (#not-kind-eq? @indent.dedent statement_block)) @indent.begin
+
+(for_statement
+  body: (_) @indent.dedent
+  (#not-kind-eq? @indent.dedent statement_block)) @indent.begin
```
lucario387, line 90: I personally think this doesn't happen frequent enough in modern javascript land to really need to care about. Especially when most formatters and autocompletion when doing `foo.for` will result in `for (const bar in foo) {}`

ribru17: Sounds good: same for while statement?

lucario387: I'd say so.

queries/ecma/indents.scm
```diff
@@ -62,7 +62,8 @@
 (parenthesized_expression
   ("("
     (_)
-    ")" @indent.end))
+    ")" @indent.end)
+  !condition)
```
lucario387, line 66: Wait what 🤔 

Isn't parenthesized expression the condition itself, how can it check for whether it contains a condition field or not

ribru17: The syntax is weird, I think to check against
```query
anythingbutthis: (node)
```
you use
```query
(node !anythingbutthis) @yay
```

ribru17: Yeah, after rereading the docs I think I misunderstood the negated fields part completely

queries/ecma/indents.scm
```diff
@@ -59,10 +59,11 @@
 (statement_block
   "{" @indent.branch)
 
-(parenthesized_expression
+((parenthesized_expression
   ("("
     (_)
-    ")" @indent.end))
+    ")" @indent.end)) @_outer
+  (#not-has-parent? @_outer if_statement))
```
lucario387, line 66: Can you also remove the unnecessary grouping here

```diff
((parenthesized_expression
-  ("("
-    (_)
-    ")" @indent.end)) @_outer
+  "("
+  (_)
+  ")" @indent.end) @_outer
  (#not-has-parent? @_outer if_statement))
`

Comment (ribru17) [reply queries/ecma/indents.scm]: Is this a pattern that the formatter should handle? or could there be cases where it can't tell if the grouping is redundant? don't know if it's different for non-top-level groupings

Comment (lucario387) [reply queries/ecma/indents.scm]: It's kinda hard to detect useless grouping. Because it may be needed. I think it can be detected with

```
(named_node
  (grouping
    "(" @format.remove
    ")" @format.remove
    .)
```

I have tried to remove with this before, but because this cannot reliably detect quantifiers. May have been fixed since.

Comment (ribru17) [reply queries/ecma/indents.scm]: Ah, makes sense

**Comment by IronBlood** [comment]
If fact the indents for `while` and `for` are also incorrect:

```js
function foo() {
  let a;
  while (true)
  a = 42;
  for (let i = 0; i < 5; i++)
  a = 42;
}
```

ribru17:
Yeah, there was a [discussion here](https://github.com/nvim-treesitter/nvim-treesitter/pull/6846#discussion_r1663535695) about it. Maybe if enough people want that we should just add it as well

IronBlood:
Sorry, didn't read the review carefully, and also thank you for implement that. But I think this test case should be added:

```js
function foo() {
  if (true)
    if (true)
      if (true)
        foo();
  bar();
}
```

The current indent is like:

```js
  if (true)
    if (true)
    if (true)
    foo();
```

ribru17:
Ah, true. I have found this is mostly covered by changing the end query to two:
```scheme
(if_statement
  consequence: (_) @indent.dedent
  (#not-kind-eq? @indent.dedent statement_block if_statement))

(if_statement
  consequence: (_) @_cons
  (#not-kind-eq? @_cons statement_block)) @indent.begin
```
but this query is a lot more complicated and I think that the above test case is so rare anyways (I fear anyone who writes code like that lol)

PR #6855, opened 2024-07-03
feat: editorconfig parser and queries

ValdezFOmar:


clason:
Please run `./scripts/format-queries.lua queries/editorconfig`.

queries/editorconfig/highlights.scm
```diff
@@ -0,0 +1,51 @@
...
+; consistency with toml and ini
+(section
+  (section_name) @type)
+
+(character_choice
+  (character) @constant)
+
+(character_range
+  start: (character) @constant
+  end: (character) @constant)
```
ObserverOfTime, line 12:
```suggestion
(character) @character
```

queries/editorconfig/highlights.scm
```diff
@@ -0,0 +1,51 @@
...
+
+[
+  "-"
+  "="
+  ".."
+  (negation)
+  (wildcard_characters)
+  (wildcard_any_characters)
+  (wildcard_single_character)
+] @operator
```
ObserverOfTime, line 34:
```suggestion
[
  "-"
  "="
  ".."
  (negation)
] @operator

[
  (wildcard_characters)
  (wildcard_any_characters)
  (wildcard_single_character)
] @character.special
```

queries/editorconfig/highlights.scm
```diff
@@ -0,0 +1,51 @@
...
+
+(escaped_character) @string.escape
+
+(pair
+  key: (identifier) @property)
```
ObserverOfTime, line 39: Values should be highlighted as `@string` by default.

queries/editorconfig/highlights.scm
```diff
@@ -0,0 +1,51 @@
...
+
+(integer) @number
+
+[
+  (unset)
+  (spelling_language)
+  (indent_style)
+  (end_of_line)
+  (charset)
+] @constant.builtin
```
ObserverOfTime, line 51: Personally, I'd prefer this:
```suggestion
(unset) @constant.builtin

[
  (spelling_language)
  (indent_style)
  (end_of_line)
  (charset)
] @string.special
```
Also, does `indent_style` include the `tab` in `indent_size`?

queries/editorconfig/highlights.scm
ValdezFOmar: Yeah, that would make more sense.

queries/editorconfig/highlights.scm
ValdezFOmar: Yes

queries/editorconfig/highlights.scm
ValdezFOmar: Here `character_choice` and `character_range` are the same as character classes and ranges, which in the queries of regex and gitignore are highlighted as `@constant`. Also, the `character` node in this grammar is more of a placeholder than something meaningful, I don't think is appropriate to highlight it like that.

queries/editorconfig/highlights.scm
ValdezFOmar: wait, here `..` doesn't mean 'parent directory' but is used to indicate an integer range: `{1..10}`.

ObserverOfTime (approved):

queries/editorconfig/highlights.scm
clason: Why? That's not how the legacy syntax works, and I don't see any documentation supporting that behavior.

ValdezFOmar: I at least get this highlighting when opening a `.editorconfig` file with `nvim --clean`, numbers *are* highlighted as numbers but not any other special value (space, true, lf).

image

clason: Yeah, I was thinking of numbers. We _could_ distinguish (override) these with a `#lua-match?`.

clason: But it's a question of semantics: if the format treats every value -- regardless of content -- as a literal, `@string` is arguably correct. But then it's not a question of "default"; _everything_ must be a string. I lack the domain knowledge here to decide. @gpanders ?

queries/editorconfig/highlights.scm
```diff
@@ -0,0 +1,56 @@
+(comment) @comment @spell
+
+; consistency with toml and ini
+(section
+  (section_name) @type)
```
clason, line 5: I disagree; this should be `@markup.heading` or `@module` ("namespaces").

queries/editorconfig/highlights.scm
```diff
@@ -0,0 +1,56 @@
...
+  "-"
+  "="
+  (negation)
+] @operator
+
+[
+  (wildcard_characters)
+  (wildcard_any_characters)
+  (wildcard_single_character)
+] @character.special
```
clason, line 37: These also apply in section headings -- is that intended? In fact, these together with `character` completely override the `section_name`.

ValdezFOmar: Pretty much intended to apply them in headings, actually that is the *only* place where they appear.

The `character` node is not captured in headings for reasons that I mentioned in another comment (https://github.com/nvim-treesitter/nvim-treesitter/pull/6855#discussion_r1665553487). In retrospective, maybe I shouldn't have exposed `character` that much.

clason: > The character node is not captured in headings for reasons that I mentioned in another comment (https://github.com/nvim-treesitter/nvim-treesitter/pull/6855#discussion_r1665553487).

But it _is_: see the following example.
```editorconfig
[*.py]
indent_size = 4
```
Tree:
```
  (section ; [12, 0] - [15, 0]
    (section_name ; [12, 1] - [12, 5]
      (wildcard_characters) ; [12, 1] - [12, 2]
      (character) ; [12, 2] - [12, 3]
      (character) ; [12, 3] - [12, 4]
      (character)) ; [12, 4] - [12, 5]
    (pair ; [13, 0] - [14, 0]
      key: (identifier) ; [13, 0] - [13, 11]
      value: (integer))) ; [13, 14] - [13, 15]
```

Again, trying to understand the intention here to see if the queries can (should) be improved.

clason: Yeah, if they're not meant to be captured, it's better to keep them private (underscored or inlined).

clason: (It just looks weird to me, but if that's what people expect, I don't mind.)

queries/editorconfig/highlights.scm
clason: In any case, since the parser exposes them, we should capture them.

queries/editorconfig/highlights.scm
clason: Actually, aren't these always paths or globs? Shouldn't we use `@string.special.path` or `@string.regexp` in that case (and possibly inject the `regex` parser)?

PR #6856, opened 2024-07-03
feat: add Cedar language support

chrnorm:
Adds support for [Cedar](https://cedarpolicy.com/) policy language.

### DIFF
--- a/queries/cedar/highlights.scm
+++ b/queries/cedar/highlights.scm
@@ -1,0 +1,67 @@
+; Function calls
+
+(call_expression
+  function: (selector_expression
+    field: (field_identifier) @function.method))
+
+[
+    "contains"
+    "containsAll"
+] @function
+
+; Identifiers
+
+[
+    (entity)
+    (principal)
+    (action)
+    (resource)
+    (context)
+    (identifier)
+    (annotation)
+] @variable
+
+(path) @type
+(field_identifier) @property
+
+; Operators
+
+[
+  "-"
+  "!"
+  "!="
+  "*"
+  "&&"
+  "+"
+  "<"
+  "<="
+  "=="
+  ">"
+  ">="
+  "||"
+  "in"
+  "has"
+] @operator
+
+; Keywords
+
+[
+  (permit)
+  (forbid)
+  (when)
+  (unless)
+] @keyword
+
+
+; Literals
+
+(str) @string
+
+(int) @number
+
+[
+  (true)
+  (false)
+] @constant.builtin
+
+(comment) @comment
--- a/queries/cedar/highlights.scm
+++ b/queries/cedar/highlights.scm
@@ -8,6 +8,9 @@
   "containsAll"
 ] @function
 
+; Annotations
+(annotation) @attribute
+
 ; Identifiers
 [
   (entity)
@@ -16,7 +19,6 @@
   (resource)
   (context)
   (identifier)
-  (annotation)
 ] @variable
 
 (path) @type
--- a/queries/cedar/highlights.scm
+++ b/queries/cedar/highlights.scm
@@ -1,31 +1,29 @@
 ; Function calls
-
 (call_expression
   function: (selector_expression
     field: (field_identifier) @function.method))
 
 [
-    "contains"
-    "containsAll"
+  "contains"
+  "containsAll"
 ] @function
 
 ; Identifiers
-
 [
-    (entity)
-    (principal)
-    (action)
-    (resource)
-    (context)
-    (identifier)
-    (annotation)
+  (entity)
+  (principal)
+  (action)
+  (resource)
+  (context)
+  (identifier)
+  (annotation)
 ] @variable
 
 (path) @type
+
 (field_identifier) @property
 
 ; Operators
-
 [
   "-"
   "!"
@@ -44,7 +42,6 @@
 ] @operator
 
 ; Keywords
-
 [
   (permit)
   (forbid)
@@ -52,9 +49,7 @@
   (unless)
 ] @keyword
 
-
 ; Literals
-
 (str) @string
 
 (int) @number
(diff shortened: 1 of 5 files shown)


clason:
Please run `./scripts/format-queries.lua queries/cedar`.

queries/cedar/highlights.scm
```diff
@@ -0,0 +1,62 @@
...
+  (entity)
+  (principal)
+  (action)
+  (resource)
+  (context)
+  (identifier)
+  (annotation)
+] @variable
+
+(path) @type
```
ObserverOfTime, line 22: Is this a file path?

queries/cedar/highlights.scm
```diff
@@ -0,0 +1,62 @@
...
+; Identifiers
+[
+  (entity)
+  (principal)
+  (action)
+  (resource)
+  (context)
+  (identifier)
+  (annotation)
+] @variable
```
ObserverOfTime, line 20: `(annotation)` should be `@attribute`.
I'm sure the others (barring `(identifier)`) can also be something else.

queries/cedar/highlights.scm
```diff
@@ -0,0 +1,62 @@
...
+
+; Literals
+(str) @string
+
+(int) @number
+
+[
+  (true)
+  (false)
+] @constant.builtin
```
ObserverOfTime, line 60:
```suggestion
] @boolean
```

queries/cedar/highlights.scm
```diff
@@ -0,0 +1,62 @@
...
+
+(comment) @comment
```
ObserverOfTime, line 62:
```suggestion
(comment) @comment @spell
```

queries/cedar/highlights.scm
chrnorm: In Cedar `path` refers to a namespace and a type of an authorization entity. For example, `MyApplication::User` or `Google::Drive::Document`.

Where possible I tried to keep names consistent with the [Cedar grammar](https://docs.cedarpolicy.com/policies/syntax-grammar.html), although this is my first Tree Sitter grammar so I'm unsure if `path` is a reserved word.

ObserverOfTime: Nothing is reserved. You might wanna highlight the namespace(s) as `@module` though.

queries/cedar/highlights.scm
chrnorm: Thanks, I've fixed `annotation` to be `@attribute`. For the others, `principal`, `action`, `resource` and `context` are "well-known variables" that are always defined for a particular policy evaluation. Seems like `@variable.builtin` could be a good fit?

ObserverOfTime: Whatever makes sense for the grammar according to [CONTRIBUTING.md](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights).

queries/cedar/highlights.scm
chrnorm: Thanks. I added `::` as `@punctuation.delimiter`. For consistency with the existing VS Code plugin (see screenshot below), I've opted to keep the path in place as is. Cedar doesn't have fully fledged namespaces yet, so the namespace is essentially part of the type name.

Thanks for all your fast feedback on this.

image

ObserverOfTime (commented):
There's more punctuation that should be highlighted.
(e.g. commas as `@punctuation.delimiter` and brackets as `@punctuation.bracket`)

queries/cedar/highlights.scm
```diff
@@ -0,0 +1,69 @@
...
+  "&&"
+  "+"
+  "<"
+  "<="
+  "=="
+  ">"
+  ">="
+  "||"
+  "in"
+  "has"
```
ribru17, line 46: These last two should be `@keyword.operator`

queries/cedar/highlights.scm
```diff
@@ -0,0 +1,69 @@
...
+  (entity)
+  (identifier)
+] @variable
+
+[
+  (principal)
+  (action)
+  (resource)
+  (context)
+] @variable.builtin
```
ribru17, line 25: These two variable queries should be the very first to prevent overriding things like `(annotation)`

queries/cedar/highlights.scm
```diff
@@ -0,0 +1,69 @@
...
+  function: (selector_expression
+    field: (field_identifier) @function.method))
+
+[
+  "contains"
+  "containsAll"
+] @function
+
+; Annotations
+(annotation) @attribute
```
ribru17, line 12: This needs to be `(annotation (identifier @attribute))`; as of now it gets overridden by `@variable`. Then you'll also need a separate one for `(annotation "@" @attribute)`

queries/cedar/highlights.scm
```diff
@@ -0,0 +1,69 @@
...
+] @variable
+
+[
+  (principal)
+  (action)
+  (resource)
+  (context)
+] @variable.builtin
+
+(path) @type
```
ribru17, line 27: This is also getting overridden by `@variable`, try `(path (identifier) @type)`

queries/cedar/highlights.scm
```diff
@@ -0,0 +1,69 @@
...
+
+; Identifiers
+[
+  (entity)
+  (identifier)
+] @variable
+
+[
+  (principal)
+  (action)
```
ribru17, line 22: Should these nodes be the anonymous node equivalents? Looks like these highlights don't get applied because sometimes an action is not exposed as `(action)`, but rather as `"action"` in, e.g., `(action_eq_constraint)`

ObserverOfTime: That sounds like a consistency issue that should be fixed in the parser.

clason:
@chrnorm any updates?

clason:
Closing as stale. @chrnorm if you want to pick this up (and can commit to maintaining the queries in the future), just comment here (**before force-pushing** any updates; that would break the PR).

Anybody else is welcome to pick this up if they are interested in this language.

PR #6857, opened 2024-07-03
fix(indent): indent empty lines correctly

### LINKED ISSUE #5037: Python indent is wrong for files indented with tabs
### Describe the bug

```python
# vim: ft=python et sw=4
def foo():
    print()
```
When editing the above file with Treesitter indent enabled, adding a new line below `print()` either with `o` or `A<cr>` correctly puts the cursor on a line already indented to the same level as `print()`.

When editing a file indented with tabs, autoindent works correctly when adding lines in the middle of a scope, but indents to the level of the enclosing scope when adding lines at the end of a scope. With `:TSDisable indent`, indentation is correct with or without tab-based indent.

### To Reproduce

```python
# vim: ft=python noet sw=0
def foo():
	print()
```
1. Ensure Treesitter indentation is enabled and the `python` parser is installed.
2. Save the above to a file and open it in Nvim.
3. Place the cursor in insert mode after `print()` and press return.
4. Observe incorrect autoindent.

### Expected behavior

Automatic indentation should behave the same regardless of whether tabs or spaces are used for indentation.

```markdown
Installation ~
- OK `node` found v20.3.1 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
  Version: cc (GCC) 13.1.1 20230429

OS Info:
{
  machine = "x86_64",
  release = "6.3.9-arch1-1",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Wed, 21 Jun 2023 20:46:20 +0000"
} ~

  - luadoc              ✓ . . . .

```

```text
NVIM v0.10.0-dev-596+g7d0a23973
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc-10 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wvla -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -fno-common -Wno-unused-result -Wimplicit-fallthrough -fdiagnostics-color=always -fstack-protector-strong -DUNIT_TESTING -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_TS_HAS_SET_MAX_START_DEPTH -I/__w/neovim/neovim/.deps/usr/include/luajit-2.1 -I/usr/include -I/__w/neovim/neovim/.deps/usr/include -I/__w/neovim/neovim/build/src/nvim/auto -I/__w/neovim/neovim/build/include -I/__w/neovim/neovim/build/cmake.config -I/__w/neovim/neovim/src -I/usr/include -I/__w/neovim/neovim/.deps/usr/include -I/__w/neovim/neovim/.deps/usr/include -I/__w/neovim/neovim/.deps/usr/include -I/__w/neovim/neovim/.deps/usr/include -I/__w/neovim/neovim/.deps/usr/include -I/__w/neovim/neovim/.deps/usr/include

  fall-back for $VIM: "/__w/neovim/neovim/build/nvim.AppDir/usr/share/nvim"
```

### Additional context

It doesn't seem to matter whether `expandtab` is set. The bug seems to present itself when either the *first* or *last* indented line of a scope contains at least one tab character. The indent composition of the lines between doesn't seem to matter.

If you take a Python file indented with tabs, set `et` and run `:%retab`, the parse tree remains identical save for column numbers. It thus seems unlikely that this bug is caused by the parser.

abhinavnatarajan:
Fixes #5037.

### DIFF
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -17,12 +17,19 @@ local function getline(lnum)
   return vim.api.nvim_buf_get_lines(0, lnum - 1, lnum, false)[1] or ""
 end
 
+---@param lnum integer
+---@return integer
+local function get_indentcols_at_line(lnum)
+  local _, indentcols = getline(lnum):find "^%s*"
+  return indentcols or 0
+end
+
 ---@param root TSNode
 ---@param lnum integer
 ---@param col? integer
 ---@return TSNode
 local function get_first_node_at_line(root, lnum, col)
-  col = col or vim.fn.indent(lnum)
+  col = col or get_indentcols_at_line(lnum)
   return root:descendant_for_range(lnum - 1, col, lnum - 1, col + 1)
 end
 
@@ -152,20 +159,20 @@ function M.get_indent(lnum)
   local node ---@type TSNode
   if is_empty_line then
     local prevlnum = vim.fn.prevnonblank(lnum)
-    local indent = vim.fn.indent(prevlnum)
+    local indentcols = get_indentcols_at_line(prevlnum)
     local prevline = vim.trim(getline(prevlnum))
     -- The final position can be trailing spaces, which should not affect indentation
-    node = get_last_node_at_line(root, prevlnum, indent + #prevline - 1)
+    node = get_last_node_at_line(root, prevlnum, indentcols + #prevline - 1)
     if node:type():match "comment" then
       -- The final node we capture of the previous line can be a comment node, which should also be ignored
       -- Unless the last line is an entire line of comment, ignore the comment range and find the last node again
-      local first_node = get_first_node_at_line(root, prevlnum, indent)
+      local first_node = get_first_node_at_line(root, prevlnum, indentcols)
       local _, scol, _, _ = node:range()
       if first_node:id() ~= node:id() then
         -- In case the last captured node is a trailing comment node, re-trim the string
-        prevline = vim.trim(prevline:sub(1, scol - indent))
+        prevline = vim.trim(prevline:sub(1, scol - indentcols))
         -- Add back indent as indent of prevline was trimmed away
-        local col = indent + #prevline - 1
+        local col = indentcols + #prevline - 1
         node = get_last_node_at_line(root, prevlnum, col)
       end
     end


lua/nvim-treesitter/indent.lua
```diff
@@ -17,12 +17,19 @@ local function getline(lnum)
   return vim.api.nvim_buf_get_lines(0, lnum - 1, lnum, false)[1] or ""
 end
 
+---@param lnum integer
+---@return integer
+local function get_indentcols_at_line(lnum)
+  local _, indentcols = getline(lnum):find "^%s*"
+  return indentcols or 0
+end
```
lucario387, line 25: For a line of
`\t\t\t\tfoo`

It should return tabstop * 4, but instead it's returning tabstop.

ribru17:
What about just
```lua
  local indent_size = vim.lsp.util.get_effective_tabstop(bufnr)
```
in line 179?

lua/nvim-treesitter/indent.lua
abhinavnatarajan: Wouldn't that cause the exact same issue as before? In (new) lines 160-178, the number of indentation columns is used as the position argument to get a treesitter node via `get_first_node_at_line` and `get_last_node_at_line`, which both call `TSNode:descendant_for_range`. That function expects a column number, not a "visual" column number. For example consider
```python
def foo():
				x = 1 # indent using 4 tabs
```
For this example, `root:get_descendent_for_range(1, 4):type()` (i.e., 2nd line 5th column, corresponding to the character `x`) correctly returns `identifier`. On the other hand `root:get_descendent_for_range(1, 16):type()` returns `module`. 

The point is that `indentcols` is the number of actual columns, not visual indentation, and this behaviour is required to get the correct treesitter node. The function `get_indentcols_at_line` is only being wherever a column position is required to get a treesitter node, not to get the actual indent width.

lucario387: Makes perfect sense honestly. This is correct

lucario387 (approved):

lucario387:
Test failures are relevant

lucario387:
Lemme fix the indent issue first, then we can merge this PR. Will need to update Go's queries first

lucario387:
@abhinavnatarajan I've pushed your commit up to master branch. Thanks for the contribution :D

PR #6858, opened 2024-07-03
docs: recommending against starting treesitter for all parsers

mbromell:
**NOTE: This is for the `main` branch**

I use a handy auto command in my own config which will enable treesitter highlighting for any language that nvim-treesitter can parse. Trying to enable treesitter on every `FileType` event with pattern `*` would cause a chain of errors when opening oil.nvim plugin windows, or Telescope windows -- so I needed a way to check if treesitter can actually parse the file type.

This auto command I've outlined in the README causes no errors for me and enables treesitter highlighting on all the languages I work with so far. Thought it would be good to document it there for others in the future since it's something I searched for but couldn't find any info on it, and I imagine many others would want this functionality too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,23 @@ vim.api.nvim_create_autocmd('FileType', {
 })
 ```
 
+You may also want to enable treesitter highlighting for all languages. But removing the `pattern` option or using `pattern = '*'` will cause issues when opening buffers that cannot be parsed by treesitter (such as Telescope windows). To enable highlighting for all languages that nvim-treesitter can parse, you can use the following auto command anywhere **after** nvim-treesitter's `setup` function:
+
+```lua
+vim.api.nvim_create_autocmd("FileType", {
+  callback = function(opts)
+    local language = vim.treesitter.language.get_lang(opts.match)
+    if language == nil then
+      return
+    end
+    local parsers = vim.tbl_keys(require("nvim-treesitter.parsers"))
+    if vim.list_contains(parsers, language) then
+      vim.treesitter.start()
+    end
+  end,
+})
+```
+
 ## Folds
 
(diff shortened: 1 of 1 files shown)


clason:
I don't want to endorse this, sorry. (If you really want to yolo enable treesitter highlighting for anything some random plugin puts on runtime path, just pcall the vim.treesitter.start call.)

mbromell:
> I don't want to endorse this, sorry. (If you really want to yolo enable treesitter highlighting for anything some random plugin puts on runtime path, just pcall the vim.treesitter.start call.)

I might be misunderstanding something here, is there a security risk in what I'm doing or do you disagree with the idea?
The purpose of this autocommand is to specifically **not** start treesitter on any random plugin's file type, instead to start it only if its on the parser table from this repo https://github.com/nvim-treesitter/nvim-treesitter/blob/caf275382f91ec8a2498d455c4d7d3fd0265ebd3/lua/nvim-treesitter/parsers.lua#L73-L84

clason:
Yes, and you might not want to do that. For example, we track the C or vimscript parsers which Neovim deliberately does not enable by default. Enabling should be an explicit and deliberate action. There is too much cargo cult around already.

mbromell:
Yes this makes perfect sense, I didn't consider the security risks until now. I'm flipping this PR upside down and instead I updated the README summarising what we discussed.

When I saw the documented way to enable highlighting, I assumed it was an incomplete feature and immediately tried to find a way to do what `highlight = true` did on master -- and I can imagine others thinking the same thing. So documenting that this isn't recommended can save the headache of dealing with more PRs and issues trying to get that `highlight = true` functionality reimplemented when 1.0 is released.

clason:
We'll burn that bridge when the `main` branch actually becomes the default. Right now, this is a pure development branch not meant for production use. I appreciate the enthusiasm, though.

PR #6864, opened 2024-07-04
feat(jsdoc): add missing highlights

ribru17:
Some missing bracket and delimiter highlights, and nicer identifier highlights for certain tags. [Reference](https://jsdoc.app/)


amaanq:
nice

PR #6872, opened 2024-07-05
fix(ecma): improve object argument indentation

### LINKED ISSUE #6036: ecma function call arguments indents are incorrect
While programming in javascript/typescript, I have noticed many problems with function call argument indentation. I finally decided to look into why I'm having problems, and discovered three different, but related problems with ecma indents that all result in over-indentation.

I've decided not to make any pull requests as javascript and typescript are some of the most used languages today, and changing the way that they are indented could potentially effect thousands of people. 

First, the bug that made me look into this. The following is code indented by nvim-treesitter:
```js
console.log({
    key: "value"
}, {
        key: "value"
    });
```
The reason the code is indented like this is because the children of `arguments` nodes are indented, but also the children of `array`, `object` and `parenthesized_expression` nodes:
https://github.com/nvim-treesitter/nvim-treesitter/blob/e7ea07e42c478cb466cf96124693b447add84011/queries/ecma/indents.scm#L1-L19
The first object is not indented because part of the object node (the opening `{`) starts on the same line as the `arguments` node. Subsequent nodes, though, are indented once because they are a child of the `arguments` node, and then again their contents are indented as they should be.

We can see the reasoning for these all being indented if we format the code in the way that it expects:
```js
console.log(
    {
        key: "value"
    }, {
        key: "value"
    }
);
```

The problem is that tree-sitter doesn't know that the only reason the children of the `arguments` node are on a new line is because the nodes themselves are multi-line. How to solve this (or if it even is possible to solve), I'm not sure. But while exploring the ecma indents, I discovered another related issue which causes multiple indents. The following is code indented by nvim-treesitter:
```js
console.log
(
        {
            key: "value"
        }, {
            key: "value"
        }
    );
```
There are two reasons the code appears this way. First, all `arguments` nodes which have an `object` node as a child will branch their indentation:
https://github.com/nvim-treesitter/nvim-treesitter/blob/e7ea07e42c478cb466cf96124693b447add84011/queries/ecma/indents.scm#L51-L57
I cannot think of any reason why this would be the case. This statement was included in the [original commit of the ecma indents file](https://github.com/nvim-treesitter/nvim-treesitter/blob/3b8c2ea492917fcb3c0e88ad6682dbd355cc0644/queries/ecma/indents.scm), and seems to have been kept around because no one knows what it does. Removing it makes the code indent like this:
```js
console.log
    (
        {
            key: "value"
        }, {
            key: "value"
        }
    );
```
This looks better, but still has a problem in my eyes: The `arguments` node (the parentheses and everything inside) is indented when not on the same line as the function identifier (as can be seen above). This doesn't look _bad_, but isn't in line with how indenting traditionally works. Usually, an indent occurs when a block (or other multi-line expression) starts, but does not end on the same line. In the above code, there is no block that starts in the first line that isn't continued that would create need for an indent. _(Note: Not necessarily an authority, but code formatters like eslint/stylistic and prettier suggest that there should be no indentation for arguments on a new line, as I have suggested.)_

This could easily be solved by removing these queries: (added in [this commit](https://github.com/nvim-treesitter/nvim-treesitter/commit/7a6d93ca5b95b4485dd70c1a1d813bb5a80fcc07))
https://github.com/nvim-treesitter/nvim-treesitter/blob/e7ea07e42c478cb466cf96124693b447add84011/queries/ecma/indents.scm#L21-L28
Removing them makes indentation look like this:
```js
console.log
(
    {
        key: "value"
    }, {
        key: "value"
    }
);
```
In my mind, this is correct.

In summary, there are three issues:
1. tree-sitter indents all children of the `arguments` node, even though we don't intend for them to be "multi-line" arguments (in the sense that each argument goes on its own line). I have no idea how this could be solved
2. `arguments` nodes that include an object are branched for seemingly no reason. This can easily be solved.
3. `arguments` nodes that are on a separate line than their function identifier are indented. This could be easily fixed, but it could be argued that this is a matter of personal preference, and it appears that this functionality was added on purpose.

I would appreciate opinions/insight on the things I have brought up. Thanks.

ribru17:
Fixes #6036

### DIFF
--- a/tests/indent/ecma/func.js
+++ b/tests/indent/ecma/func.js
@@ -49,3 +49,11 @@ function prepare_list_fetcher(filter) {
     { key: ['/list', filter] }
   )
 }
+
+console.log({
+  key: 'value',
+}, {
+  key: 'value',
+},
+  hi
+);
--- a/queries/ecma/indents.scm
+++ b/queries/ecma/indents.scm
@@ -6,7 +6,6 @@
   (export_clause)
   (formal_parameters)
   (named_imports)
-  (object)
   (object_pattern)
   (parenthesized_expression)
   (return_statement)
@@ -17,6 +16,9 @@
   (template_substitution)
   (ternary_expression)
 ] @indent.begin
+
+((object) @indent.begin
+  (#not-has-parent? @indent.begin arguments))
 
 (arguments
   (call_expression) @indent.begin)


tests/indent/ecma/func.js
```diff
@@ -49,3 +49,11 @@ function prepare_list_fetcher(filter) {
...
   )
 }
+
+console.log({
+  key: 'value',
+}, {
+  key: 'value',
+},
+  hi
+);
```
lucario387, line 59: Umm
Pretty sure this will fail, and it's pretty common as well
```
console.log(
  {
    key: 'value',
  },
  {
    key: 'value',
  },
  hi
);
```

What we need to check here is whether the start of the object/array/function node is on the same line of the place where ( is placed

lucario387 (dismissed):

tests/indent/ecma/func.js
ribru17: Ah, damn. You're right, looks like this one needs more thought

PR #6878, opened 2024-07-07
feat: tree-sitter-nginx

opa-oz:
Hello, it's me again 🙇

(this week I went to edit nginx.conf, heh)

## Before
Screenshot 2024-07-07 at 19 27 06

## After
Screenshot 2024-07-07 at 19 43 05

Screenshot 2024-07-07 at 19 29 59

# Checklist
- [x] `tree-sitter build`
Screenshot 2024-07-07 at 19 28 35

- [x] `TSInstallFromGrammar nginx`
Screenshot 2024-07-07 at 19 29 13

- [x] Alphabetically sorted
- [x] Injections & highlights in place

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1374,6 +1374,14 @@ list.nix = {
   maintainers = { "@leo60228" },
 }
 
+list.nginx = {
+  install_info = {
+    url = "https://github.com/opa-oz/tree-sitter-nginx",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@opa-oz" },
+}
+
 list.norg = {
   install_info = {
     url = "https://github.com/nvim-neorg/tree-sitter-norg",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1335,6 +1335,14 @@ list.nasm = {
   maintainers = { "@ObserverOfTime" },
 }
 
+list.nginx = {
+  install_info = {
+    url = "https://github.com/opa-oz/tree-sitter-nginx",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@opa-oz" },
+}
+
 list.nickel = {
   install_info = {
     url = "https://github.com/nickel-lang/tree-sitter-nickel",
@@ -1372,14 +1380,6 @@ list.nix = {
     files = { "src/parser.c", "src/scanner.c" },
   },
   maintainers = { "@leo60228" },
-}
-
-list.nginx = {
-  install_info = {
-    url = "https://github.com/opa-oz/tree-sitter-nginx",
-    files = { "src/parser.c", "src/scanner.c" },
-  },
-  maintainers = { "@opa-oz" },
 }
 
 list.norg = {
--- a/queries/nginx/highlights.scm
+++ b/queries/nginx/highlights.scm
@@ -32,9 +32,13 @@
 ] @variable.builtin
 
 [
-  (string_literal)
   (file)
   (mask)
+] @string.special.path
+
+[
+  (string_literal)
+  (quoted_string_literal)
 ] @string
 
 (directive
--- a/queries/nginx/highlights.scm
+++ b/queries/nginx/highlights.scm
@@ -2,9 +2,6 @@
 
 (value) @variable
 
-(attribute
-  (keyword) @attribute)
-
 [
   (location_modifier)
   "="
@@ -28,7 +25,7 @@
   (level)
   (connection_method)
   (var)
-  condition: (condition)
+  (condition)
 ] @variable.builtin
 
 [
@@ -40,9 +37,6 @@
   (string_literal)
   (quoted_string_literal)
 ] @string
-
-(directive
-  (variable) @variable.parameter)
 
 (directive
   (variable
--- a/queries/nginx/highlights.scm
+++ b/queries/nginx/highlights.scm
@@ -1,0 +1,64 @@
+(comment) @comment @spell
+
+(value) @variable
+
+(attribute 
+  (keyword) @attribute) 
+
+[
+  (location_modifier)
+  "="
+] @operator
+
+[
+  (keyword)
+  "location"
+] @keyword
+
+[
+  "if"
+  "map"
+] @keyword.conditional
+
+(directive 
+  (keyword) @constant) 
+
+(boolean) @boolean
+
+[
+  (auto)
+  (constant)
+  (level)
+  (connection_method)
+  (var)
+  condition: (condition)
+] @variable.builtin
+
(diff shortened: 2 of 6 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -1374,6 +1374,14 @@ list.nix = {
   maintainers = { "@leo60228" },
 }
 
+list.nginx = {
```
clason, line 1377: Please keep the lists alphabetized.

opa-oz: >g comes before i in my alphabet...

:( sorry

clason:
> Warning: Some amount of directives are in different color. Debatable, but it's because these directives are [core functionality](https://nginx.org/en/docs/ngx_core_module.html#directives)

That's not acceptable in general; color doesn't matter (that's on colorschemes), only semantically correct captures. So if they're directives, they should be captured as such -- or you need to add a new `directive.builtin` **and use it consistently across all languages**.

opa-oz:
@clason fair, removed custom value for core directives

lua/nvim-treesitter/parsers.lua
opa-oz: done

queries/nginx/highlights.scm
```diff
@@ -0,0 +1,61 @@
...
+
+(value) @variable
+
+(attribute
+  (keyword) @attribute)
+
+[
+  (location_modifier)
+  "="
+] @operator
```
ObserverOfTime, line 11: Should `(location_modifier)` be `@keyword.modifier` ?

queries/nginx/highlights.scm
```diff
@@ -0,0 +1,61 @@
...
+  (connection_method)
+  (var)
+  condition: (condition)
+] @variable.builtin
+
+[
+  (string_literal)
+  (file)
+  (mask)
+] @string
```
ObserverOfTime, line 38: `(file)` should be `@string.special.path`

queries/nginx/highlights.scm
opa-oz: @ObserverOfTime idk, what do you think?:
```js
    location_modifier: (_) =>
      choice(
        "=", // An exact match of the URI path.
        "~", // A case-sensitive regular expression match.
        "~*", //A case-insensitive regular expression match.
        "^~", //Indicates that the following path should be considered the best non-regular expression match.
      ),
```

queries/nginx/highlights.scm
```diff
@@ -0,0 +1,61 @@
+(comment) @comment @spell
+
+(value) @variable
```
ribru17, line 3: I am wondering if this would make more sense as `@string`, if it would be more similar to like an `ini` type file?

ribru17:
It seems that the parser may need a bit more work, as I have found some examples that aren't parsed properly:
```nginx
events {
   worker_connections 32768;
   multi_accept off; # this comment causes the closing bracket below to be an (ERROR); putting just "#" with no content causes it to be parsed as a (comment)

http {
   include /etc/nginx/mime.types;
   access_log off;
   server_tokens off;
   msie_padding off;

   # errors here
   sendfile off; #default
   tcp_nopush off; #default
   tcp_nodelay on; #default
   keepalive_timeout 65;
   keepalive_disable none; #default msie6
   keepalive_requests 300000; #default 100

   js_path "/etc/nginx/njs/";

   #js_import utils.js;
   js_import main from /njs/hello.js;

   server {
      listen 8080 default_server reuseport deferred fastopen=4096;
      # 8080 default_server reuseport deferred backlog=65535 fastopen=4096;

      location = /plaintext {
         js_content main.hello;
      }

      location = /json {
         js_content main.json;
      }

      #location = /version {
      #   js_content utils.version;
      #}
   }
}

}
```
```nginx
server {
  # multiple "." cannot be handled
  listen 50.116.11.184:80;
  server_name filldisk.com;
  rewrite ^ http://www.filldisk.com$request_uri permanent;
  include /home/feross/config/nginx/_server.include;
}
```
Also I think things like `js_import` and `js_path` should be handled, rather than just being parsed as `(attribute (keyword) (value) (value) (value))`

One more thing: not a necessity but it would be nice if values starting with `http` or maybe ending in `.com` could be parsed as URLs, but maybe that is out of scope.

opa-oz:
@ribru17 
Lemme check errors and fix

>Also I think things like js_import and js_path should be handled, rather than just being parsed as (attribute (keyword) (value) (value) (value))

Don't wanna be a nerd, but it seems like LSP suff. There are a lot of rules, that can be handled individually 
For example 
- `gzip_types       text/plain;` and `default_type  application/octet-stream;` probably have list of possible values and all combinations of them
- `proxy_set_header   Host             $host;` except for rare custom-headers also can have only pre-defined values

And I suppose it's much more. Shall we pre-define all of them?

>One more thing: not a necessity but it would be nice if values starting with http or maybe ending in .com could be parsed as URLs, but maybe that is out of scope.

- Nginx serves not only http
- `.com` isn't the only existing domain (it's a couple of hundreds of them rn, if I recall correctly)
- `localhost` is also URL

opa-oz:
Screenshot 2024-07-08 at 12 32 43

@ribru17 comments fixed, double quotes as well

queries/nginx/highlights.scm
```diff
@@ -0,0 +1,65 @@
...
+
+(boolean) @boolean
+
+[
+  (auto)
+  (constant)
+  (level)
+  (connection_method)
+  (var)
+  condition: (condition)
```
lucario387, line 31:
```suggestion
  (condition)
```

Doesn't seem necessary when you only have one usage of it in the grammar file

queries/nginx/highlights.scm
```diff
@@ -0,0 +1,65 @@
...
+  (string_literal)
+  (quoted_string_literal)
+] @string
+
+(directive
+  (variable) @variable.parameter)
+
+(directive
+  (variable
+    (keyword) @variable.parameter))
```
lucario387, line 49: One of these should be enough. No need to capture both of them with the capture.

ribru17:
> Don't wanna be a nerd, but it seems like LSP suff. There are a lot of rules, that can be handled individually
> For example

Sounds good :+1:

queries/nginx/highlights.scm
```diff
@@ -0,0 +1,65 @@
+(comment) @comment @spell
+
+(value) @variable
+
+(attribute
+  (keyword) @attribute)
```
ribru17, line 6: This is overridden by the query below that highlights `(keyword)` as `@keyword`

queries/nginx/highlights.scm
opa-oz: If someone wants my comment to re-review PR - idk 

Please share your wisdom what is better - `@variable` or `@string`? 🙇

ribru17 (approved):

PR #6890, opened 2024-07-08
fix(typst): code block injection now works correctly

Andrew15-5:
Since this project has a separate `injections.scm` per language, this means that it is also have to be fixed. See https://github.com/uben0/tree-sitter-typst/issues/34.

### DIFF
--- a/queries/typst/injections.scm
+++ b/queries/typst/injections.scm
@@ -2,5 +2,6 @@
   (#set! injection.language "comment"))
 
 (raw_blck
-  (ident) @injection.language
-  (blob) @injection.content)
+  lang: (ident) @_lang
+  (blob) @injection.content
+  (#set-lang-from-info-string! @_lang))


clason:
Thank you, but that directive is deprecated because it's non-standard. A better approach would be to  add the list of known language aliases [here](https://github.com/nvim-treesitter/nvim-treesitter/blob/c1ad655b6a0c83ab48e55240f367e2bc0c15af31/lua/nvim-treesitter/parsers.lua#L28-L52) (assuming it's small).

Generic "extension recognition" is out of scope for this plugin.

clason:
(It is small -- only Rust, Elixir, JavaScript, Typescript, Haskell, Python, Erlang, and -- of course -- Typst.)

clason:
https://github.com/nvim-treesitter/nvim-treesitter/pull/6897

Andrew15-5:
The problem is that Markdown works better than Typst. Just from looking at it, I think it uses the same thing that I proposed:

https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries%2Fmarkdown%2Finjections.scm#L1-L5

So shouldn't Markdown also be patched from using this approach?

P.S. I'm completely new to all this.

clason:
Yes, and it shouldn't (and won't in the future). My PR adds support for _all_  (standard) `@injection.language` directives, no custom directives needed.

Andrew15-5:
OK, thanks!

PR #6897, opened 2024-07-09
feat(parser): register common short-hand as filetypes

clason:
Problem: Cannot use short-hands for @injection.language (e.g., `py` for
Python).

Solution: Register common short-hand as filetypes for some languages.

List taken from https://github.com/helix-editor/helix/blob/master/languages.toml

Also remove deprecated custom filetype->parser mapping.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1,36 +1,11 @@
 local api = vim.api
 local ts = vim.treesitter
 
-local new_lang_api = ts.language.register ~= nil
-
-local filetype_to_parsername = {}
-
-if new_lang_api then
-  filetype_to_parsername = setmetatable({}, {
-    __newindex = function(_, k, v)
-      require("nvim-treesitter.utils").notify(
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -1,36 +1,11 @@
...
-  end
-  filetype_to_parsername[ft] = lang
-end
-
 for ft, lang in pairs {
   automake = "make",
   javascriptreact = "javascript",
   ecma = "javascript",
   jsx = "javascript",
-  sh = "bash",
```
clason, line 33: This alias is already set via the `filetype` key for `bash` below.

ribru17:
LGTM. Out of curiosity, what is the rationale of this method vs. adding something like the filetype detection in core?

ribru17 (approved):

clason:
Simplicity, and backward compatibility. _If_ a `vim.filetype.match` fallback to language resolution was added in core (which isn't a given), it'd only be available after 0.11 is released.

PR #6898, opened 2024-07-09
Configuration fix/enhancement

CheeseTurtle:
# In `configs.setup` function
 
`tbl_deep_extend` `config.modules` with `user_data.modules` rather than with `user_data`. This not only avoids introducing unnecessary/non-module items into `config.modules`, but also correctly processes `user_data` tables that correctly follow the `TSConfig` type definition specified in the annotations.

# In `nvim-treesitter.setup` function

Added optional parameters to pass to `configs.setup()` so that `require"nvim-treesitter".setup(opts)` works as expected.

### DIFF
--- a/lua/nvim-treesitter.lua
+++ b/lua/nvim-treesitter.lua
@@ -9,11 +9,13 @@ require "nvim-treesitter.query_predicates"
 
 local M = {}
 
-function M.setup()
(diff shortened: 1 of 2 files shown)


clason:
Thank you, but the whole modules system is deprecated and will be removed in 1.0.

CheeseTurtle:
You don't think it's worth repairing this in the meantime?

clason:
I haven't seen anything that's worth repairing so far? Can you explain what exactly the problem is you're trying to solve?

PR #6901, opened 2024-07-10
fix(idl): fix break changes

cathaysia:


queries/idl/highlights.scm
```diff
@@ -106,7 +106,11 @@
 
 (boolean_literal) @boolean
 
-(number_literal) @number
+[
+  (integer_literal)
+  (floating_pt_literal)
+  (fixed_pt_literal)
```
clason, line 112: Those should be `@number.float` (at least the first one).

PR #6909, opened 2024-07-11
feat(parser): add amber language

Mte90:
The Amber language is quite new and we are working on that.

We have an experimental tree-sitter integration, so in the meantime we have this way to test it in neovim using my fork with that change.

https://github.com/amber-lang/tree-sitter-amber
https://docs.amber-lang.com

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -32,6 +32,7 @@ for ft, lang in pairs {
   py = "python",
   erl = "erlang",
   typ = "typst",
+  ab = "amber",
 } do
   ts.language.register(lang, ft)
 end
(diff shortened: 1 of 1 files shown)


clason:
Also, no queries is a deal-breaker -- these are the main things this repo provides!

Mte90:
Yeah I was thinking the same that is experimental, I left as draft for that reason :-)

lua/nvim-treesitter/parsers.lua
```diff
@@ -78,6 +79,16 @@ list.agda = {
   maintainers = { "@Decodetalkers" },
 }
 
+list.amber = {
+  install_info = {
+    url = "https://github.com/amber-lang/tree-sitter-amber",
+    files = { "src/parser.c" },
+    generate_requires_npm = true,
```
clason, line 86: For the record, that's not needed as far as I can tell.

lua/nvim-treesitter/parsers.lua
```diff
@@ -32,6 +32,7 @@ for ft, lang in pairs {
   py = "python",
   erl = "erlang",
   typ = "typst",
+  ab = "amber",
```
clason, line 35: Duplicate.

clason (commented):
You might want to look at https://github.com/tree-sitter-grammars/template if you haven't already.

clason:
>  I left as draft for that reason :-)

Yeah, but it's not even ready for that yet. Just take your time and open a new pull request when the grammar is stable and you have a set of queries.

(You don't need this repo for testing; just `tree-sitter generate` and `tree-sitter build -o amber.so`, then put the `.so` in a `parser` directory anywhere on your runtimepath -- same for the queries.)

PR #6929, opened 2024-07-16
fix(facility): Add highlight for event keyword

bryankenote:


clason:
Please run the formatter `./scripts/format-queries.lua queries/facility` (or fix the indentation manually).

PR #6944, opened 2024-07-19
feat(latex): add additional include highlights

jdujava:
These were recognized by `latex` grammar, but not highlighted.


clason:
If these are filenames, they should be `@string.special.path` (consistently).

jdujava:
Will try to look into which are really meant to be filenames, and which are mainly package names or something like that.

jdujava:
Tried to choose highlight groups corresponging to the most frequent use cases of the commands.

PR #6945, opened 2024-07-19
feat(latex): capture placeholder as variable

jdujava:


clason:
Can you show examples (text and screenshot before/after)?

jdujava:
Before:
image
After:
image

clason:
Hmm, I don't think that's correct. `@variable.parameter` is for the parameters of a function in the signature, not the body (compare C). These should just be `@variable`.

jdujava:
When working with `lua`, I see for example this
image

I thought that it would be nice to have it similarly for `latex`. My colorscheme doesn't do anything special for `@variable` (I think this is usually the default behavior), and I don't want to change this behavior for all languages.

Since `#num` is really special in `latex`, isn't it fine to make an "exeption" and capture it as `@variable.parameter`, or do you propose another solution to obtain nice highlighting of `latex` placeholders?

clason:
The highlight you show is not from tree-sitter, though, but from LSP semantic tokens (which are different and out of our control). In general, we frown upon exceptions; consistent use of captures is the core value that this project provides.

Maybe you can find another capture group that would fit? We can interpret broadly for such different languages as LaTeX, but we won't directly contradict the intended usage (otherwise consistent colorschemes become impossible).

jdujava:
What do you think about `@variable.parameter.builtin`?

clason:
That has the same problem (this is not a parameter).

jdujava:
Sorry, I thought that would mean something different.. Maybe "@punctuation.special"?

clason:
The captures are hierarchical: subcaptures are always a special case. Punctuation doesn't fit, either; it's really just a variable...

My suggestion would be to use `@variable` (which is the semantically correct capture) and just adding `@variable.latex` to your colorscheme.

jdujava:
Nice, I didn't know about such a workaround!

queries/latex/highlights.scm
```diff
@@ -1,6 +1,8 @@
 ; General syntax
 (command_name) @function
 
+(placeholder) @variable.parameter
```
clason, line 4:
```suggestion
(placeholder) @variable
```

PR #6947, opened 2024-07-19
feat(bash): Readline injections

ValdezFOmar:


queries/bash/injections.scm
```diff
@@ -65,3 +65,19 @@
...
+  (#eq? @_command "bind")
+  (#set! injection.language "readline"))
+
+((command
+  name: (command_name) @_command
+  .
+  argument: (raw_string) @injection.content)
+  (#eq? @_command "bind")
+  (#offset! @injection.content 0 1 0 -1)
+  (#set! injection.language "readline"))
```
ribru17, line 83:
```suggestion
((command
  name: (command_name) @_command
  .
  argument: [
    (string)
    (raw_string)
  ] @injection.content)
  (#eq? @_command "bind")
  (#offset! @injection.content 0 1 0 -1)
  (#set! injection.include-children)
  (#set! injection.language "readline"))
```

Might be cleaner to use one query

ribru17 (approved):

ribru17:
Thanks!

PR #6949, opened 2024-07-20
feat(ecma): inject sql template methods

DimensionalDot:
Added injection into tagged templates for SQL method calls (e.g., `` client.sql`...` ``) in order to add compatibility with Vercel's PostgreSQL SDK.

This is a continuation of https://github.com/nvim-treesitter/nvim-treesitter/pull/6555 which I inadvertently closed. That was originally opened to address the issue raised in mentioned in https://github.com/nvim-treesitter/nvim-treesitter/pull/6550.

### DIFF
--- a/queries/ecma/injections.scm
+++ b/queries/ecma/injections.scm
@@ -43,6 +43,26 @@
   (#set! injection.include-children)
   (#set! injection.language "html"))
 
+; Vercel PostgreSQL
+; foo.sql`...` or foo.sql(`...`)
+(call_expression
+  function: [
+    (await_expression
+      (member_expression
+        property: ((property_identifier) @injection.language)))
+    (member_expression
+      property: ((property_identifier) @injection.language)
+    )
+  ]
+  arguments: [
+    (arguments
+      (template_string) @injection.content)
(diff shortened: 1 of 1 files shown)


queries/ecma/injections.scm
```diff
@@ -43,6 +43,25 @@
   (#set! injection.include-children)
   (#set! injection.language "html"))
 
+; Vercel PostgreSQL
+; foo.sql`...` or foo.sql(`...`)
+(call_expression
+  function: [
+    (await_expression
+      (member_expression
+        property: (property_identifier) @injection.language))
```
ribru17, line 52: Out of curiosity, what does this capture try to achieve? Looks like without it the injection still works for something like ``await client.sql`...`;``

ribru17 (approved):

queries/ecma/injections.scm
DimensionalDot: Without that line the query doesn't capture async calls with type arguments (e.g., `` await client.sql<T>`...` ``).

lucario387 (approved):

PR #6951, opened 2024-07-20
feat(snakemake): add wildcard flag

osthomas:
Wildcard interpolations support the `:q` flag to quote list/tuple elements with whitespace.

```snakemake
rule x:
    input: 
        "the first file.txt",
        "the second file.txt"
    output: "output.txt"
    shell:
       """
       cat {input:q} > {output}
       """
```

With `{input:q}`, this expands to `cat 'the first file.txt' 'the second file.txt' > output.txt`

With `{input}`, this expands to `cat the first file.txt the second file.txt > output.txt`

The `(wildcard (flag))` node was added in the latest version of the `snakemake` parser.

### DIFF
--- a/queries/snakemake/highlights.scm
+++ b/queries/snakemake/highlights.scm
@@ -44,7 +44,8 @@
 
 ; Wildcard names
 (wildcard
-  (identifier) @variable)
+  (identifier) @variable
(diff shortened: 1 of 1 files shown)


ribru17:
Needs a lockfile update as well

PR #6954, opened 2024-07-21
feat(latex): command definition without curly braces

jdujava:
Support both
```latex
\newcommand\foo{bar}
```
and
```latex
\newcommand{\foo}{bar}
```

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -387,7 +387,7 @@
     "revision": "854a40e99f7c70258e522bdb8ab584ede6196e2e"
   },
   "latex": {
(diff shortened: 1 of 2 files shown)


clason:
Would help to have examples of the targeted syntax before each pattern.

And can this be combined with the previous (using grouping)?

jdujava:
Is this what you had in mind?

clason:
If it works :)

jdujava:
Fixed (forgot to format also after adding the comment :).

PR #6955, opened 2024-07-21
feat(latex): simplify captures of font changing

jdujava:
Additionally, add `\bm` (bold symbols in maths mode).

### DIFF
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -239,13 +239,7 @@
   command: (command_name) @_name
   arg: (curly_group
     (_) @markup.italic))
-  (#eq? @_name "\\emph"))
-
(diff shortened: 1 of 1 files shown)


clason:
˚\bm` is deprecated (pure TeX), same as `\it` ;)

jdujava:
I have in mind the following: https://ctan.org/pkg/bm?lang=en
As far as I know, this is usually the way to go for bold symbols in math mode.

clason:
Hmm, yes, packages can do any number of things. We have already decided that adding package-specific highlights is out-of-scope for this project; that is left for a (so far hypothetical) nvimtex plugin.

jdujava:
Oh, alright. Only the simplification then.

PR #6956, opened 2024-07-22
fix(lua): use @property for keys in table literals

tmillr:
`@property` should be used for key-value pairs, while `@variable.member` is used for field-access/index-expressions. Using different captures also provides the benefit of allowing the two to be highlighted differently.


ribru17 (approved):

ribru17:
Thanks!

PR #6961, opened 2024-07-22
fix(filetype): register csharp for `c_sharp` parser

seblj:
Hi :)

I was a bit unsure if I should add this to `main` or `master` branch. I am maintaining a fork that enables compatibility with the "new" language server for csharp, and it currently won't highlight with treesitter as the server is sending codeblocks with ```csharp.

Let me know if you instead want this to the master branch

Another a bit unrelated question I have been wondering on for some weeks. Do you know when the main branch will become the default branch here?


seblj:
I understand, I can register it in the plugin then

seblyng:
@clason Could this be reconsidered now? `roslyn_ls` is now added as a part of `nvim-lspconfig` per: https://github.com/neovim/nvim-lspconfig/pull/3753

I think it would be nice if this could be supported "out of the box" now that this server is in `nvim-lspconfig`

clason:
Sorry, no. Nvim-lspconfig is a completely unrelated project.

seblyng:
> (And the master branch is -- finally -- frozen anyway.)

That's good to hear. Is there any ETA of the main branch being default?

> Sorry, no. Nvim-lspconfig is a completely unrelated project and what they do has no bearing on this project. Our policy remains that we support the official upstream filetypes; anything else is user config.

I would say that I quite strongly disagree about this one since I consider both `nvim-treesitter` and `nvim-lspconfig` as official plugins, and I would think that since neovim now supports that language server "out of the box", it would make sense to improve the "out of box" experience with very little extra code. I also see that `c-sharp` is registered, but I can't find a single reference to that in neovim codebase?

clason:
> That's good to hear. Is there any ETA of the main branch being default?

We're getting there, is all I can say. The main blocker is getting the tests to pass (which are broken on Nvim 0.11 due to an unmaintained upstream dependency, which we need to replace). There's also one or two open questions on which features remain supported, which I need to make a decision on before declaring it "done".

>  I consider both nvim-treesitter and nvim-lspconfig as official plugins

You might do so; I don't (and certainly not _this_ one, which isn't even under the Neovim org).

PR #6962, opened 2024-07-22
fix(latex): refactor highlights

jdujava:
There is no need to capture `declaration` field when defining new commands, since it is already captured by `(command_name)`.

Supersedes #6954.

### DIFF
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -1,17 +1,19 @@
 ; General syntax
-(command_name) @function
-
-(placeholder) @variable
-
-(text_mode
-  [
-    "\\text"
-    "\\intertext"
-    "\\shortintertext"
-  ] @function)
+(command_name) @function @nospell
 
 (caption
   command: _ @function)
+
+(placeholder) @variable
+
+; Turn spelling on for text
+(text) @spell
+
+; \text, \intertext, \shortintertext, ...
+(text_mode
+  command: _ @function @nospell
+  content: (curly_group
+    (_) @none @spell))
 
 (key_value_pair
   key: (_) @variable.parameter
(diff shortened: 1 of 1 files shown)


queries/latex/highlights.scm
```diff
@@ -1,17 +1,19 @@
 ; General syntax
-(command_name) @function
+(command_name) @function @nospell
```
clason, line 2: Shouldn't nospell be the default?

queries/latex/highlights.scm
```diff
@@ -137,26 +140,6 @@
...
-(math_environment
-  (begin
-    command: _ @markup.math
-    name: (curly_group_text
-      (_) @markup.math)))
-
-(math_environment
-  (_) @markup.math)
-
-(math_environment
```
clason, line 154: These were deliberate to mimic VimTeX syntax highlighting (which is the high watermark).

I'm not averse on principle, but that looks like a bigger change that should be discussed.

queries/latex/highlights.scm
jdujava: But when some command is called inside a sentence, its parent node can be `(text)` -> `@spell`. Anyway, it was there before, I just moved it to one capture.

queries/latex/highlights.scm
jdujava: But weren't `(math_environment (begin/end ...` straight up redundant, when
```
(math_environment
  (_) @markup.math)
```
(which is still there, just moved) gives `@markup.math` to all children of `(math_environment)`?

jdujava: Also, `(text_mode content:...` was just moved/combined.

clason:
You seem to have a plan, so I'll trust you ;) (mobile review)

PR #6968, opened 2024-07-23
feat(latex): improve parameters

jdujava:
Before:
image
After:
image
Also, `(brack_group)` appeared in different contexts (optional arguments for environments), where it definitely wasn't `@variable.parameter`. What do you think?


clason:
Makes sense! The LaTeX queries have not seen a lot of dedicated effort beyond simple local fixes.

PR #6970, opened 2024-07-24
feat(julia): Add syntax highlighting for markdown and bash `prefixed_string_literals`

kdheepak:
This PR adds markdown syntax highlighting and bash command line syntax highlighting for the appropriate Julia expressions.

bash syntax highlighting:

image

markdown syntax highlighting:

image

This PR also adds a comment for each injection.

### DIFF
--- a/queries/julia/injections.scm
+++ b/queries/julia/injections.scm
@@ -13,14 +13,27 @@
   (#set! injection.language "markdown")
   (#offset! @injection.content 0 3 0 -3))
 
+; Inject comments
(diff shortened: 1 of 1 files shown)


clason:
@savq

savq:
Note that julia command literals only parse simple commands. Pipelines, control operators, etc. are not parsed.

Beginners regularly get confused when Julia commands don't work like their shell, so let's hope this doesn't cause more confusion 😅

PR #6979, opened 2024-07-26
feat(parsers): Highlight wildcard imports

ValdezFOmar:


ribru17:
Thank you, I think this is good. Would you mind also changing this for other languages that wildcards in imports, e.g. rust, java, etc.

clason:
(Sorry to pile more work onto you, but consistency across languages is one of the primary project goals, and it helps to keep PRs/commits "atomic" in that sense.)

ValdezFOmar:
No problem, it'll just take a bit of time since I'm not that familiar with all of those languages.

ValdezFOmar:
I've added highlights for the languages I could find that allow wildcard imports, let me know if there's more missing. Captures for kotlin and groovy are missing because the parser exposes wildcard imports as a single anonymous node `".*"`, I'll open pull requests upstream to make it more convenient to highlight.

ribru17 (approved):

ribru17:
Thank you!

ValdezFOmar:
> I'll open pull requests upstream to make it more convenient to highlight.

- [ ] https://github.com/fwcd/tree-sitter-kotlin/pull/130#issue-2433253133

clason:
Let's give it until the end of the week; if there's no upstream movement, I'd say we merge what we have and add a TODO for the remaining languages so we can pick this up in the future.

clason:
@ValdezFOmar groovy parser has been updated, too; let me know when the queries are ready to be merged!

ValdezFOmar:
@clason It should be good to go now

clason:
Thanks!

PR #6981, opened 2024-07-26
fix(latex): environment name should be label

jdujava:
Environment names are captured as `@label` in `(begin)` and `(end)` nodes, and so should they be when we are defining environments.

Before:
image
After:
image

### DIFF
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -67,7 +67,7 @@
 (environment_definition
   command: _ @function.macro @nospell
   name: (curly_group_text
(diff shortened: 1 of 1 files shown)


clason:
What did we say about PR descriptions ;) Can you explain _why_ this should be changed? Before/after screenshots would be helpful, too.

clason:
(Don't get me wrong, I really appreciate the effort you put into the LaTeX treesitter support! But the current queries are opinionated (sorry not sorry) but consistent, so if there is a change -- to which I am open -- it must be consistent as well.)

PR #6982, opened 2024-07-26
feat(query): Highlight key in set! directive

ValdezFOmar:
Setting any property like `injection.language` or `injection.language` doesn't highlight it. This fixes it.

### DIFF
--- a/queries/query/highlights.scm
+++ b/queries/query/highlights.scm
@@ -12,7 +12,7 @@
   name: (identifier) @variable)
 
 (field_definition
-  name: (identifier) @property)
+  name: (identifier) @variable.member)
 
 (negated_field
   "!" @operator
@@ -62,6 +62,15 @@
 
 ((comment) @keyword.directive @nospell
   (#lua-match? @keyword.directive "^;+%s*format%-ignore%s*$"))
+
+((predicate
+  name: (identifier) @_name
+  parameters: (parameters
+    .
+    (capture)?
+    .
+    (identifier) @property))
+  (#eq? @_name "set"))
 
 ((predicate
   name: (identifier) @_name


queries/query/highlights.scm
```diff
@@ -63,6 +63,15 @@
 ((comment) @keyword.directive @nospell
   (#lua-match? @keyword.directive "^;+%s*format%-ignore%s*$"))
 
+((predicate
+  name: (identifier) @_name
+  parameters: (parameters
+    .
+    (capture)?
+    .
```
clason, line 71: Is that really needed?

lucario387: It should be to cover for `(#set! foo)` and `(#set! @capture foo)`. Then the question should be: will we need to add capture for `(#set "foo")` and `(#set! @capture "foo")`

clason: Both work fine for me without the anchoring.

clason: Also, is `@property` the best capture here? It's also used for field names (like `name` in this pattern).

clason:
+1 to adding a capture for that, tho!

queries/query/highlights.scm
ValdezFOmar: Docs say that:

https://github.com/nvim-treesitter/nvim-treesitter/blob/e265fec94c7dc0c8c64cb86820ff5ad3ee135c7d/CONTRIBUTING.md?plain=1#L145

So that's the best capture that I can think for this.

clason: Yeah, fair. But then the field name is wrong ;)

ValdezFOmar: Maybe `@variable.member` for the field? But `@label` also sounds good.

clason: > The anchoring is to prevent highlighting value in (#set! @foo key value), but I don't know how common that is.

I tested that as well; `value` remains captured as `@string`.

 > Maybe @variable.member for the field? But @label also sounds good

I think the former is better; `@label` is more for GOTO and friends.

clason: > The anchoring is to prevent highlighting value in (#set! @foo key value), but I don't know how common that is.

Ah, you're right, you do need it for 
```
(#set! "priority" 90)
```
which is pretty common. Sorry! Can you put the anchors back then (and rename the field name)?

clason: It's a bit of an annoyance that the metadata can also be specified as a string. What do we want to do about that? Add an alternation, or leave it like that?

ValdezFOmar: What about capturing just the string content without quotes? Something like:

```scm
[
  (identifier) @property
  (string
    (string_content) @property)
]
```

image

clason: I'm half-tempted to just remove the quotes (and possibly support for the quoted form in Neovim). Two ways of doing the same thing is bad.

ValdezFOmar: Then I'll just capture `(identifier)` for now

clason:
Cleaning up string properties (in the docs, too!) can be a follow-up.

PR #6987, opened 2024-07-27
feat(latex): capture left/right math delimiters

jdujava:
<del>
Capture math delimiter commands such as `\left` and `\right`.

Not sure wheter it should be `@function.macro` or `@function` or perhaps something else. I am open to suggestions.

Before:
image
After:
image

</del>

Edit:
Capture math delimiters provided by commands such as
`\left` and `\right`.

### DIFF
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -36,6 +36,12 @@
 
 (delimiter) @punctuation.delimiter
 
+(math_delimiter
+  left_command: _ @punctuation.delimiter
+  left_delimiter: _ @punctuation.delimiter
+  right_command: _ @punctuation.delimiter
+  right_delimiter: _ @punctuation.delimiter)
+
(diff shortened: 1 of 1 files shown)


clason:
Well, that was deliberate: `\left(` is a single delimiter.

jdujava:
But should that mean that `\left` is highlighted depending on whether it is inside math mode, or outside (like in the command definition in first screenshot)? Isn't it still it's own macro/function (deserving it's own capture :)?.

Or am I missing something?

clason:
The issue here is that `\langle` etc. is not highlighted as `@punctuation.delimiter`, as it should ;)

(Again, the highlighting is explicitly designed to mirror VimTeX's; if you've never used that plugin, I _highly_ recommend checking it out.)

jdujava:
Does this sound alright?

PR #6988, opened 2024-07-27
refactor(queries): Remove quotes from properties in set! directive

ValdezFOmar:
follow-up from https://github.com/nvim-treesitter/nvim-treesitter/pull/6982

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -287,7 +287,7 @@ A conceal can be restricted to part of the capture via the [`#offset!` directive
 #### Priority
 
 Captures can be assigned a priority to control precedence of highlights via the
-`#set! "priority" <number>` directive (see `:h treesitter-highlight-priority`).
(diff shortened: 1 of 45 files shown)


ValdezFOmar:
@clason should this be done also in neovim? The docs use quoted properties:

https://github.com/neovim/neovim/blob/bdff50dee56ebf6de58d58315920abf2f8e262f7/runtime/doc/treesitter.txt#L340

clason:
I would handle the docs when syncing the queries.

ValdezFOmar:
Also, while doing this I notice this strange query, is this correct?

https://github.com/nvim-treesitter/nvim-treesitter/blob/f97e0de0053df1cb3b9f74b9b1239acced1b7d9a/queries/strace/injections.scm#L1-L2

clason:
No, that's an obvious copypasta. A lot of queries are abandonware, unfortunately.

clason:
Can you fix this (in a separate commit) while you're at it?

PR #6995, opened 2024-07-29
Correct Godot Resources language name in README.md

Weefa:
Adjustment is to change the language name hint from `gdresource`, which is unrecognized, to `godot_resource`. Should save the next person some time looking up the appropriate name.


clason:
Thank you, but that file is autogenerated.

PR #6997, opened 2024-07-30
Add VRL support for nvim-treesitter

belltoy:
Add [VRL](https://vrl.dev) support for nvim-treesitter

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2398,6 +2398,14 @@ list.vimdoc = {
   maintainers = { "@clason" },
 }
 
+list.vrl = {
+  install_info = {
+    url = "https://github.com/belltoy/tree-sitter-vrl",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@belltoy" },
+}
+
 list.vue = {
   install_info = {
     url = "https://github.com/tree-sitter-grammars/tree-sitter-vue",
--- a/queries/vrl/folds.scm
+++ b/queries/vrl/folds.scm
@@ -1,0 +1,6 @@
+[
+  (block)
+  (object)
+  (array)
+  (arguments)
+] @fold
--- a/queries/vrl/highlights.scm
+++ b/queries/vrl/highlights.scm
@@ -1,0 +1,108 @@
+(comment) @comment @spell
+
+(null) @constant.builtin
+
+(timestamp) @constant
+
+(closure_variables
+  (ident) @variable.parameter)
+
+(integer) @number
+
+(float) @number.float
+
+[
+  (string)
+  (raw_string)
+] @string
+
+[
+  (raw_string_escape_sequence)
+  (escape_sequence)
+  (regex_escape_sequence)
+] @string.escape
+
+(string_template
+  "{{" @punctuation.special
+  (_)
+  "}}" @punctuation.special)
+
+(regex) @string.regexp
+
+(boolean) @boolean
+
+(ident) @variable
+
+(noop) @variable.builtin
+
+(function_call
+  (ident) @function.call)
+
+; VRL queries
+(query
+  [
+    (event)
+    (metadata)
+  ] @variable.builtin)
+
+(query
+  (path
+    [
+      (field) @variable
+      (string) @string
+      (index) @number
+      "." @punctuation.delimiter
+    ]))
+
+"return" @keyword.return
+
+"abort" @keyword.exception
+
+[
+  "if"
+  "else"
+] @keyword.conditional
+
+[
+  "="
+  "=="
+  "!="
+  "|="
+  ">"
+  ">="
+  "<"
+  "<="
+  "+"
+  "-"
+  "*"
+  "/"
+  "&&"
+  "||"
+  "??"
+  "|"
+  "!"
+] @operator
+
+[
+  "->"
+  ":"
+  ";"
+  ","
+] @punctuation.delimiter
+
+[
+  "("
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
+] @punctuation.bracket
+
+(closure_variables
+  "|" @punctuation.bracket)
+
+(function_call
+  (ident) @keyword.exception
+  "!"
+  (#any-of? @keyword.exception "assert" "assert_eq"))
--- a/queries/vrl/injections.scm
+++ b/queries/vrl/injections.scm
@@ -1,0 +1,6 @@
+((comment) @injection.content
+  (#set! injection.language "comment"))
+
+((regex) @injection.content
+  (#offset! @injection.content 0 2 0 -1)
+  (#set! injection.language "regex"))
--- a/README.md
+++ b/README.md
@@ -451,6 +451,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [vhs](https://github.com/charmbracelet/tree-sitter-vhs) (maintained by @caarlos0)
 - [x] [vim](https://github.com/neovim/tree-sitter-vim) (maintained by @clason)
 - [x] [vimdoc](https://github.com/neovim/tree-sitter-vimdoc) (maintained by @clason)
+- [x] [vrl](https://github.com/belltoy/tree-sitter-vrl) (maintained by @belltoy)
 - [x] [vue](https://github.com/tree-sitter-grammars/tree-sitter-vue) (maintained by @WhyNotHugo, @lucario387)
 - [x] [wgsl](https://github.com/szebniok/tree-sitter-wgsl) (maintained by @szebniok)
 - [x] [wgsl_bevy](https://github.com/theHamsta/tree-sitter-wgsl-bevy) (maintained by @theHamsta)
--- a/lockfile.json
+++ b/lockfile.json
@@ -833,6 +833,9 @@
   "vimdoc": {
     "revision": "2249c44ecd3f5cf22da3dcccfb74f816ddb29245"
   },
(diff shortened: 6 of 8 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -2398,6 +2398,15 @@ list.vimdoc = {
   maintainers = { "@clason" },
 }
 
+list.vrl = {
+  install_info = {
+    url = "https://github.com/belltoy/tree-sitter-vrl",
+    files = { "src/parser.c" },
+  },
+  filetype = "help",
```
clason, line 2406: No. `help` is the filetype for Vim's builtin help.

queries/vrl/folds.scm
```diff
@@ -0,0 +1,8 @@
+;; vim:ft=query
```
clason, line 1:
```suggestion
```
Not needed.

queries/vrl/injections.scm
```diff
@@ -0,0 +1,5 @@
+;; vim:ft=query
+
```
clason, line 2: please add a `comment` injection.

clason:
This line in your grammar does not work correctly on the latest tree-sitter CLI: https://github.com/belltoy/tree-sitter-vrl/blob/c875e0aa123336883df02215235e1d2791b1f783/grammar.js#L440-L442

(Single child `seq` or `choice` nodes will not be exposed to queries.) Please remove that and adapt the queries if necessary.

clason:
And please run `./scripts/format-queries.lua queries/vrl`.

queries/vrl/folds.scm
```diff
@@ -0,0 +1,8 @@
+; vim:ft=query
```
clason, line 1:
```suggestion
```
Modelines are not needed if these files are in a `queries` subdirectory.

lua/nvim-treesitter/parsers.lua
```diff
@@ -2398,6 +2398,15 @@ list.vimdoc = {
   maintainers = { "@clason" },
 }
 
+list.vrl = {
+  install_info = {
+    url = "https://github.com/belltoy/tree-sitter-vrl",
+    files = { "src/parser.c" },
+  },
+  filetype = "vrl",
```
clason, line 2406:
```suggestion
```
That entry is only needed if the language name is different from the filetype.

belltoy:
> And please run `./scripts/format-queries.lua queries/vrl`.

```
./scripts/format-queries.lua queries/vrl
E5113: Error while calling lua chunk: ...m/0.10.0/share/nvim/runtime/lua/vim/treesitter/query.lua:252: Query error at 163:17. Impossible pattern:
(anonymous_node (string) @format.keep)
                ^

stack traceback:
        [C]: in function '_ts_parse_query'
        ...m/0.10.0/share/nvim/runtime/lua/vim/treesitter/query.lua:252: in function 'fn'
        ...ovim/0.10.0/share/nvim/runtime/lua/vim/func/_memoize.lua:58: in function 'parse'
        ./scripts/format-queries.lua:432: in function 'format'
        ./scripts/format-queries.lua:447: in main chunk
```

@clason Any suggestions?

clason:
You need to make sure you are using the query parser in the same version as this plugin installs. (If you upgrade to the latest stable Nvim 0.10.1, that should also fix this.)

clason:
And in case you didn't know: You can only use the capture names listed in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights, no others.

belltoy:
Thanks, @clason. I've updated the parser and this PR.

queries/vrl/highlights.scm
```diff
@@ -0,0 +1,118 @@
+(comment) @comment
```
ObserverOfTime, line 1:
```suggestion
(comment) @comment @spell
```

queries/vrl/highlights.scm
```diff
@@ -0,0 +1,118 @@
...
+
+[
+  (null)
+  (boolean)
+] @constant.builtin
```
ObserverOfTime, line 6:
```suggestion
(null) @constant.builtin
```

queries/vrl/highlights.scm
```diff
@@ -0,0 +1,118 @@
...
+
+(closure_variables
+  "|" @punctuation.bracket)
+
+(closure_variables
+  (ident) @variable.parameter)
+
+(integer) @number
+
+(float) @number
```
ObserverOfTime, line 18:
```suggestion
(float) @number.float
```

queries/vrl/highlights.scm
```diff
@@ -0,0 +1,118 @@
...
+  ; "self"
+  ; "std"
+  ; "then"
+  ; "this"
+  "true"
+  ; "type"
+  ; "until"
+  ; "use"
+  ; "while"
+] @keyword
```
ObserverOfTime, line 83: Use more specific captures and avoid duplicates.

ObserverOfTime (changes requested):

queries/vrl/highlights.scm
```diff
@@ -0,0 +1,118 @@
...
+] @string
+
+[
+  (raw_string_escape_sequence)
+  (escape_sequence)
+  (regex_escape_sequence)
+] @string.escape
+
+(string_template
+  (ident) @variable)
```
lucario387, line 32: This is not needed as ident capture a few lines below this is more relevant

queries/vrl/highlights.scm
clason: and `(boolean) @boolean`?

ObserverOfTime: It's duplicated below (along with "true" and "false").

queries/vrl/highlights.scm
```diff
@@ -0,0 +1,118 @@
...
+
+[
+  (null)
+  (boolean)
+] @constant.builtin
+
+(timestamp) @constant
+
+(closure_variables
+  "|" @punctuation.bracket)
```
ribru17, line 11: Can you move this after the `@operator` query? Otherwise that one overrides this one

queries/vrl/highlights.scm
```diff
@@ -0,0 +1,118 @@
...
+  "+"
+  "-"
+  "*"
+  "/"
+  "&&"
+  "||"
+  "??"
+  "|"
+  "!"
+  "."
```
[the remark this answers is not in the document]
belltoy: Thanks, @ribru17.

Btw, how can I deal with the `"."` used as both query path delimiter and the default builtin `event` variable?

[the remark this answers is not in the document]
belltoy: I think I solved this by adding predicates to the `event` query. But that's a little weird.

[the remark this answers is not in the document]
clason: Just do
```
((path)
  "." @punctuation.delimiter)
```

to only capture the "." within `(path)`.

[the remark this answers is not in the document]
belltoy: That's a simpler query, but the `"."` in `(query)` will match both of the two queries. It only works for highlighting, because seems the @punctuation.delimiter priority is higher than @variable.builtin.

For variable reference, the `"."` in `(query)` will be select as `event` references if  the `(event)` query is not predicated with `#not-has-parent?`.

image

I don't want the interior points to be visible in the tree. Do I have to, to prevent `not-has-parent` in `locals.scm`?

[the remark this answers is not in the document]
clason: Not in my tests; there the first dot and the following dots were separately captured as (I believe) expected.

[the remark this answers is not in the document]
clason: Screenshot 2024-07-30 at 19 03 54
Screenshot 2024-07-30 at 19 04 03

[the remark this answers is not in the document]
clason: (I'm less concerned about locals, as these are currently not really used in nvim-treesitter.)

clason (approved):

queries/vrl/highlights.scm
```diff
@@ -0,0 +1,110 @@
...
+(regex) @string.regexp
+
+(boolean) @boolean
+
+(ident) @variable
+
+[
+  (null)
+  (noop)
+] @variable.builtin
```
ObserverOfTime, line 39: Duplicate
```suggestion
(noop) @variable.builtin
```

queries/vrl/highlights.scm
```diff
@@ -0,0 +1,110 @@
...
+
+[
+  "->"
+  ":"
+  ";"
+  ","
+] @punctuation.delimiter
+
+(path
+  "." @punctuation.delimiter)
```
clason, line 93: oh, and maybe check if you can include this in the `(query (path))` pattern above? (And same for the `(query ".")` pattern.)

Not necessary, but maybe at least bring them closer together.

queries/vrl/highlights.scm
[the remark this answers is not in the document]
belltoy: Resolved by putting `(event)` inside `(query)`. Thanks, @clason :)

The variable reference highlights feature is provided by [vim-illuminate](https://github.com/RRethy/vim-illuminate/blob/master/lua/illuminate/providers/treesitter.lua#L16), which is useful if there is not an LSP available.

queries/vrl/highlights.scm
belltoy: +1

ObserverOfTime (approved):

PR #7000, opened 2024-07-31
feat(powershell): added data section name highlights

L2jLiga:
added highlight for variable name in data section definition

```powershell
Data TextMsgs {
 ^---- highlight "data" as @keyword.modifier
       ^---- highlights variable definition as @constant
  "Message"
  "Another message"
}
```

data sections can be used to isolate read-only data and text strings so I think `@constant` is most suitable tag for this

docs: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_data_sections?view=powershell-7.4

### DIFF
--- a/queries/powershell/highlights.scm
+++ b/queries/powershell/highlights.scm
@@ -143,6 +143,9 @@
 ((variable) @variable.builtin
   (#lua-match? @variable.builtin "^\$env:"))
 
+(data_name
+  (simple_name) @constant)
+
 (comment) @comment @spell
 
 ((program
--- a/queries/powershell/highlights.scm
+++ b/queries/powershell/highlights.scm
@@ -67,7 +67,10 @@
   "enum"
 ] @keyword.type
 
-(class_attribute) @keyword.modifier
+[
+  "data"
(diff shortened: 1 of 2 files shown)


clason:
> read-only data

sounds more like  `@attribute.builtin`
> ; attribute annotations (e.g. Python decorators, Rust lifetimes)

or `@keyword.modifier`
> ; keywords modifying other constructs (e.g. `const`, `static`, `public`)

to me?

L2jLiga:
> sounds more like @attribute.builtin
> or @keyword.modifier

Definitely no, added query highlights variable name itself, not modifier or keyword

> Concrete examples in the PR description always go a long way and help cut down on stupid comments based on guessing ;

My bad, sorry 😅 

```powershell
Data TextMsgs {
       ^---- query highlights this variable definition
  "Message"
  "Another message"
}
```

clason:
But the DATA could be? ;)

L2jLiga:
> But the DATA could be? ;)

Actually, yes :)

ribru17 (approved):

PR #7002, opened 2024-08-01
fix(format): field_def inside grouping

lucario387:
Related: https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/645


clason:
Thanks!

PR #7017, opened 2024-08-05
feat: update snakemake parser and queries

osthomas:
Update snakemake parser and queries to accommodate breaking change (newly included `exclude` keyword in rule imports).

Also includes minor improvements to highlight and indent queries.


queries/snakemake/highlights.scm
```diff
@@ -13,7 +13,7 @@
     name: _ @keyword))
 
 ; Subordinate directives (eg. input, output)
-(_
+((_)
   body: (_
     (directive
       name: _ @label)))
```
lucario387, line 19: This changes from finding `body` field that are child of any nodes, to finding `body` fields that have any node before it, making a lot of captures. I don't think this is correct

lucario387 (changes requested):

queries/snakemake/highlights.scm
osthomas: The old parser broke when interspersing comments in rules. This is fixed.
However, when a comment is inserted immediately after opening a rule, the comment node is a child of `rule_definition`, not of `rule_body`. This is the same structure as when inserting a comment immediately after opening a vanilla Python class. 

The old capture (with `_` instead of `(_)`) does not work there.
queries

lucario387: Then can't the query only be
```suggestion
body: (_
  (directive
    name: _ @label))
```

osthomas: That seems to work. I did not know this was valid syntax.

queries/snakemake/highlights.scm
```diff
@@ -33,6 +32,7 @@
   "use" @keyword.import
   "rule" @keyword.import
   "from" @keyword.import
+  "exclude"? @keyword.import
   "as"? @keyword.import
   "with"? @keyword.import)
```
lucario387, line 37: Can this be reduced to

```suggestion
[
  "use"
  "rule"
  "from"
  "exclude"
  "as"
  "with"
] @keyword.import
```

queries/snakemake/indents.scm
```diff
@@ -9,8 +9,18 @@
...
 ((module_definition) @indent.begin
   (#set! indent.immediate 1))
 
 ((directive) @indent.begin
   (#set! indent.immediate 1))
+
+; end indentation after last parameter node (no following ',')
+(directive_parameters
+  (_)*
+  (_) @indent.end)
```
lucario387, line 26: It seems like this is only for the last child of `directive_parameters`. In that case, something like this will be better
```suggestion
(directive_parameters
  (_) @indent.end .)
```

queries/snakemake/highlights.scm
osthomas: The scoping to `(rule_import)` is required, but grouping is fine.

lucario387 (approved):

PR #6960, opened 2024-07-22
fix(install): abort installation using git in active git session

clason:
Problem: Installing a parser using git in an active git session (e.g., when editing a commit message) can corrupt the corresponding repository.

Solution: Check for typical environment variables first and abort installation if found.

Closes #5353

PR #7026, opened 2024-08-08
Fix missing format argument in shell_command_selectors

Kilenaitor:
Problem: [#6960](https://github.com/nvim-treesitter/nvim-treesitter/pull/6960) introduced a fix to bail installs if an active git session is detected. An error message is supposed to be printed using `string.format`. However, the error message has two placeholder %s args but only one provided argument. This causes a runtime error.

Solution: Add project_name again as a third argument to string.format to hydrate the second %s. This fixes the issue and the error message prints as-expected.


clason:
I've been meaning to remove this code completely but got held up on other things. Thanks!

PR #7028, opened 2024-08-09
Fix to remain compatible with nvim 0.9

svermeulen:


svermeulen:
Ah yes, true.  Fixed

clason:
Last PR I'm merging for this; next issue I'll just drop git support.

clason:
But thanks ;)

PR #7033, opened 2024-08-11
fix(ini): Fix injection in comments

ValdezFOmar:


clason:
Can you explain why you made that change? (Before/after screenshots)

ValdezFOmar:
The inner node is preventing injections since `include-children` is not set, so I just capture that node directly.

Before:

image

After:

image

clason:
I'd rather set `include-children`, to be honest, to keep consistency with all the other comment injections. (Unless they're broken as well? Can you check at least a sample? Or is ini special for some reason?)

ribru17:
Lua is the only other language I know of that exposes comment content differently from the comment marker, and it injects this way ([without include children](https://github.com/nvim-treesitter/nvim-treesitter/blob/079af15600ceaca92e6ce96e40d37955307aa11e/queries/lua/injections.scm#L186))

ribru17 (approved):

clason:
I have also found vhdl in the meantime. So I guess there's enough precedent.

PR #7036, opened 2024-08-13
feat(json): injection for scripts field in package.json

ValdezFOmar:
Before:

image

After:

image

### DIFF
--- a/queries/json/injections.scm
+++ b/queries/json/injections.scm
@@ -1,0 +1,11 @@
+; scripts field in package.json
+((object
+  (pair
+    key: (string
+      (string_content) @_key)
+    value: (object
+      (pair
+        value: (string
+          (string_content) @injection.content)))))
+  (#eq? @_key "scripts")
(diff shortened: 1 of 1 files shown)


ObserverOfTime:
Some scripts are batch rather than bash.

ValdezFOmar:
Strange, the [documentation](https://docs.npmjs.com/cli/v10/using-npm/scripts#exiting) only mentions `sh`:

> Scripts are run by passing the line as a script argument to sh.

ObserverOfTime:
See [_this_](https://docs.npmjs.com/cli/v10/commands/npm-run-script#script-shell) documentation.

ValdezFOmar:
Then I agree this might not be appropriate.

clason:
I was thinking there might be a smarter injection we can use to distinguish that, but JSON files are (often) _huge_ so performance is a critical concern. So I on the whole I agree we should not do this. It was a good idea, though!

PR #7038, opened 2024-08-14
feat(jinja2): add jinja2

cathaysia:
add https://github.com/cathaysia/tree-sitter-jinja2

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1021,6 +1021,24 @@ list.javascript = {
   maintainers = { "@steelsojka" },
 }
 
+list.jinja2 = {
+  install_info = {
+    url = "https://github.com/cathaysia/tree-sitter-jinja2",
+    location = "tree-sitter-jinja2",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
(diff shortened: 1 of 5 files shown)


queries/queries/jinja2
```diff
@@ -0,0 +1 @@
+../tree-sitter-jinja2/queries
```
lucario387, line 1: uhhh ...

lua/nvim-treesitter/parsers.lua
```diff
@@ -1021,6 +1021,24 @@ list.javascript = {
   maintainers = { "@steelsojka" },
 }
 
+list.jinja2 = {
+  install_info = {
+    url = "https://github.com/cathaysia/tree-sitter-jinja2",
+    location = "tree-sitter-jinja2",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@steelsojka" },
```
clason, line 1030: No.

clason:
Make sure your repo does not contain any symlinks; these don't work on Windows so we can't include the parser.

clason (changes requested):
You need to use only the captures listed in CONTRIBUTING.MD.

PR #7049, opened 2024-08-17
feat(surrealql): add parser and queries

Ce11an:
## Description

Hey hey 👋🏻 

I've been working on an experimental tree-sitter parser for [SurrealQL](https://surrealdb.com/docs/surrealdb/surrealql), the query language used by [SurrealDB](https://surrealdb.com/). While this is still a work in progress, I've laid the groundwork for the grammar, though not all statements are fully supported just yet. I wanted to share this with the community 🙏🏻

P.S. Not 100% sure how to update the `lockfile.json` - any advice?


clason:
Not sure this is the best way of sharing WIP projects -- we need to raise the standards to ensure a reasonable out of the box experience. (It's ok if we're talking about a week or two of effort, but months of unactionable PRs puts a definite strain on maintenance.)

Ce11an:
> Not sure this is the best way of sharing WIP projects -- we need to raise the standards to ensure a reasonable out of the box experience. (It's ok if we're talking about a week or two of effort, but months of unactionable PRs puts a definite strain on maintenance.)

Totally understandable! Will revisit this once the grammar has more complete. Shall I leave this in draft or close the PR and later reopen?

clason:
I'd say finish it first, as it looks like it'll take longer. (You should also consider more fine-grained captures.)

You don't need to add it here to test the parser with Neovim; just add the checked-out repo to `runtimepath` and make sure that
1. the parser is built into a `parser` subdirectory (`tree-sitter build -o parser/surrealql.so`)
2. there are queries in a `queries/surrealql` subdirectory.

Ce11an:
> I'd say finish it first, as it looks like it'll take longer. (You should also consider more fine-grained captures.)
> 
> 
> 
> You don't need to add it here to test the parser with Neovim; just add the checked-out repo to `runtimepath` and make sure that
> 
> 1. the parser is built into a `parser` subdirectory (`tree-sitter build -o parser/surrealql.so`)
> 
> 2. there are queries in a `queries/surrealql` subdirectory.

Thanks for the advice. Yeah, already have that set-up locally 👍🏻

PR #7053, opened 2024-08-18
feat(latex): capture label_definition as label

DimitrisDimitropoulos:
This PR attempts to enhance the highlighting of labels by capturing label_definition as a label. Here are some before and after:  

The label capture does not fit the strict guidelines provided by the repo, but it is fitted in the context of LSP navigation. Please feel free to close this PR if it not fitted.
Thanks for your work and for your time reviewing :)

### DIFF
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -86,6 +86,9 @@
     (_) @function))
 
 (label_definition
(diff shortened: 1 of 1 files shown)


clason:
Thank you, but this is already captures below as `@markup.link`, which is more appropriate in the context of tree-sitter highlighting.

PR #7054, opened 2024-08-18
ci(tests): add optional workflow for generating grammars

clason:
run on PR by adding label `ci:generate` or manually via workflow dispatch

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,4 +1,4 @@
-name: Linting and style checking
+name: Lint
 
 on:
   pull_request:
@@ -34,7 +34,7 @@ jobs:
           args: --check .
 
   format-queries:
-    name: Lint queries
+    name: Format queries
     runs-on: ubuntu-latest
     env:
       NVIM_TAG: stable
--- a/.github/workflows/test-core.yml
+++ b/.github/workflows/test-core.yml
@@ -1,0 +1,55 @@
+on:
+  workflow_call:
+    inputs:
+      type:
+        type: string
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  check_compilation:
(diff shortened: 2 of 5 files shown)


clason:
Looks like CI can't handle the (memory) pressure:
* `ubuntu` gets OOM killed out of the gate, it looks like
* `macos` hangs but gets there eventually (1:40h)
* `windows` is... fine (surprisingly)? (0:30h) Let's see if it isn't a fluke.

clason:
Also, @dlvandenberg @lucario387 indent tests fail for `angular` on the main branch (comments don't get indentented) -- does one of you see what's going on?

dundargoc:
There's a significant overlap between `test-generate.yml` and `test-queries.yml`. I would just combine these two workflows into a single workflow or find a way to reuse the code parts.

clason:
How would I do that? There's significant differences.

clason:
The culprit is https://github.com/yuja/tree-sitter-qmljs (@yuja), which eats ~18Gb RAM on `tree-sitter generate src/grammar.json).

clason:
Ok, things look better with tree-sitter 0.23.0 _but_:
* ubuntu still gets OOM-killed due to spawning multiple tree-sitter jobs in parallel
* macos still needs 1:40 because the runners are absolute potatoes (7 Gb -- thanks, Apple, I see you still believe in RAM austerity)
* windows is now done in 10min -- you go, girl!

clason:
Setting the maximum number of jobs to 1 works but takes 20:15/18:30/22:30 -- okay for an optional test, but generating from grammar will become required eventually when `parser.c` is dropped from the repos.

clason:
@dundargoc I think this is useful now -- if you have suggestions for improving the workflow(s) (in general), I'm all ears!

dundargoc (approved):

PR #7056, opened 2024-08-18
feat(scala): And indentation queries for Scala

susliko:


ribru17:
Can you please add some tests?

clason:
and run `./scripts/format-queries.lua queries/scala`?

susliko:
@ribru17 I'll gladly add tests - haven't noticed at first that indent queries are tested.

What are prerequisites for running tests locally (haven't found them in CONTRIBUTING.md)?
Execution of`./scripts/run_tests.sh` fails with
```
E492: Not an editor command: PlenaryBustedDirectory tests/ { minimal_init = './scrip
ts/minimal_init.lua' }
```

clason:
https://github.com/nvim-treesitter/nvim-treesitter/wiki/Testing is the best we have.

ribru17:
Yeah it's not the best way but I usually just run gg=G and if the text stays the same it's good to go (though this doesn't account for line additions)

susliko:
@ribru17 @clason Added some tests. Please, take a look if you can

clason:
@lucario387

ribru17:
Thanks, this looks great overall! I would suggest just a few things:

---

indentations for if ... else

from what I can see, dedenting `if` consequence blocks (that have a following `else` section) changes (breaks) the parser output, so this one is pretty important. Maybe add a test for:

```scala
  def cloudsmithCredentials(validate: Boolean): Seq[Credentials] = {
    (sys.env.get("PUBLISH_USER"), sys.env.get("PUBLISH_PASSWORD")) match {
      case (Some(user), Some(password)) =>
        Seq(Credentials("Cloudsmith API", "maven.cloudsmith.io", user, password))
      case _ =>
        if (validate)
          throw new Exception("Publishing credentials expected in `PUBLISH_USER` and `PUBLISH_PASSWORD`.")
        else
          Nil
    }
  }
```

---
```scheme
(block_comment) @indent.auto
```
to prevent bullets like

```scala
/*
 * Copyright (C) 2009-2023 Lightbend Inc. <https://www.lightbend.com>
 */
```
from being dedented

---

It would also be good to have indentation after `case` blocks, and in cases like the following:

```scala
  def show(player: FidePlayer, tours: Paginator[RelayTour.WithLastRound])(using Context) =
    ui.player.show(
      player,
      (tours.nbResults > 0).option(
        views.relay.tour.renderPager(views.relay.tour.asRelayPager(tours)): page =>
          routes.Fide.show(player.id, player.slug, page) // this should be indent3ed
      )
    )
```

queries/scala/indents.scm
```diff
@@ -0,0 +1,32 @@
+[
+  (template_body)
+  (block)
+  (parameters)
+  (arguments)
+  (match_expression)
+  (splice_expression)
+  (import_declaration)
+  (function_definition)
```
ribru17, line 9: I think function definitions should only be indented like this if it doesn't have a `(block)` within it, otherwise the block will get doubly indented. Maybe replace this with a
```query
(function_definition
  body: (_) @_body
  (#not-kind-eq? @_body "block")) @indent.begin
```
below? Or only apply `(block)` indents when not in a `function_definition`? Whatever you think makes the most sense given the grammar.

susliko: For some reason it doesn't cause any problems. Consider
```scala
def foo() = {
  println() // Indented only once
```

So, I guess we could leave it like this 🤔

susliko:
@ribru17 Thank you for making these queries more robust!
I've addressed your comments regarding block comments, cases in pattern matching and colon arguments.

But I had no luck making proper indents for `if` expression. 

I tried 
```
(if_expression 
  condition: (_)) @indent.begin
```

But I haven't found a way to issue a dedent before `else`
```
   if true then
       1
        else // should be dedent here instead
```
`"else" @indent.end` does not help.

queries/scala/indents.scm
lucario387: what about 
```scala
def foo() = 
  {
    println()
}
```

susliko: Scala has scalafmt which is a standard formatter that is included in the language server and is used almost in every Scala project.
So, I perceive indentation queries only as a tool to guess non-irritating and grammatically correct indentation while typing.

ribru17: > For some reason function_definitions with blocks are not double indented. Consider
> 
> ```scala
> def foo() = {
>   println() // Indented only once
> }
> ```
> 
> So, I guess we could leave it like this, with simpler queries 🤔

Interesting... looks like it only applies when using multiline parameters. See the following:

```scala
  def foo(
    sourceDir: File
  ) = {
      println() // Indented twice
    }
```

ribru17:
That's unfortunate... looks like it isn't feasible right now given the way the parser handles these things. If it could recognize `else` keywords that are indented ahead of the `if` then I think it would work, but maybe that's not how it should parse things. I guess we can not worry about it for now

PR #7064, opened 2024-08-20
Glimmer grammar updates

NullVoxPopuli:
This is stemming from an effort to move `<template>` support out of tree-sitter-javascript.
See this discussion: https://github.com/tree-sitter/tree-sitter-typescript/pull/308

These 3 grammars are related
- https://github.com/NullVoxPopuli/tree-sitter-glimmer-javascript
  - JavaScript, no JSX, w/ one extra node for `<template>` parsing
- https://github.com/NullVoxPopuli/tree-sitter-glimmer-typescript
  - TypeScript, no TSX, w/ one extra node for `<template>` parsing
- https://github.com/ember-tooling/tree-sitter-glimmer
  - the language for between the `<template>` tags parsed in the above grammars

<details><summary>I tested this locally with packer, via</summary>

```lua
  use {
    -- 'nvim-treesitter/nvim-treesitter',
    '~/Development/OpenSource/nvim-treesitter',
    branch = "glimmer-updates",
    config = function()
      require("nvim-treesitter.install").prefer_git = true
      require 'nvim-treesitter.configs'.setup {
        autotag = {
          enable = true,
          filetypes = {
            "html",
            "javascript", "typescript",
            "typescript.glimmer", "javascript.glimmer",
            "javascriptreact", "typescriptreact",
            "markdown",
            "glimmer", "handlebars", "hbs", "svelte", "vue"
          }
        },
        ensure_installed = {
          -- Web Languages
          -- "javascript", "typescript",
          "html", "css", "regex",
          "ejs",
          -- Web Framework Languages
          "svelte",
          "glimmer", "glimmer_javascript", "glimmer_typescript",
          -- Web Transport Languages
          "graphql",
          -- Documentation Languages
          "markdown", "markdown_inline",
          -- "help", -- missing?
          -- "comment", -- slow?
          "jsdoc",
          -- Configuration Languages
          "toml", "jsonc", "yaml",
          "dockerfile",
          "lua", "vim",
          -- Scripting Languages
          "commonlisp",
          "bash",
          "jq",
          -- Systems Languages
          "c", "cmake",
          "rust",
          "go",
          -- Specifically for the treesitter AST
          "query",
          -- Utility Syntaxes
          "diff",
          "jq",
          "git_rebase", "gitcommit", "gitignore"
        },
        ignore_install = {
          "json" -- jsonc is better
        },
        highlight = {
          enable = true,
        },
        indent = {
          enable = true
        },
      }

      require('ts_context_commentstring').setup({})
      vim.g.skip_ts_context_commentstring_module = true
    end
  }
```

</details>


queries/glimmer/highlights.scm
```diff
@@ -1,113 +1,84 @@
...
 ; Tags that start with a lower case letter are HTML tags
 ; We'll also use this highlighting for named blocks (which start with `:`)
 ((tag_name) @tag
-  (#lua-match? @tag "^:?[%l]"))
-
+  (#match? @tag "^(:)?[a-z]"))
 ; Tags that start with a capital letter are Glimmer components
 ((tag_name) @constructor
-  (#lua-match? @constructor "^%u"))
+  (#match? @constructor "^[A-Z]"))
```
clason, line 9: No. That uses the vim regex engine, which kills performance.

clason (changes requested):

queries/glimmer/highlights.scm
NullVoxPopuli: oh excellent! I wondered why this change existed -- I'll revert upstream, too

lua/nvim-treesitter/parsers.lua
```diff
@@ -11,6 +11,8 @@ for ft, lang in pairs {
   ["typescript.tsx"] = "tsx",
   ["terraform-vars"] = "terraform",
   ["html.handlebars"] = "glimmer",
+  ["javascript.glimmer"] = "glimmer_javascript",
```
clason, line 14: You are duplicating the information in the parsers table. This is only needed for _additional_ filetypes.

clason (changes requested):

queries/glimmer/highlights.scm
```diff
@@ -1,113 +1,84 @@
...
 
-(attribute_name) @attribute
+(attribute_name) @property
 
 (string_literal) @string
-
```
clason, line 14: Don't make formatting changes; we enforce style via the format-queries script, just run that locally.

queries/glimmer/highlights.scm
NullVoxPopuli, line 1: it looks like this file and tree-sitter-glimmer got out of sync :(

queries/glimmer/highlights.scm
```diff
@@ -1,113 +1,84 @@
...
+  (path_expression (identifier) @variable)
+  (identifier) @variable
   ])
-  (#not-eq? @variable "this"))
-
+  (#not-match? @variable "this"))
 ; `this` should be highlighted as a built-in variable
 ((identifier) @variable.builtin
-  (#eq? @variable.builtin "this"))
+  (#match? @variable.builtin "this"))
```
clason, line 57: That is completely backwards. Prefer eq over any-of over lia-match over match (if absolutely unavoidable).

clason (changes requested):

queries/glimmer_javascript/tags.scm
```diff
@@ -0,0 +1,99 @@
...
+      name: (identifier) @name)
+    (function_declaration
+      name: (identifier) @name)
+    (generator_function
+      name: (identifier) @name)
+    (generator_function_declaration
+      name: (identifier) @name)
+  ] @definition.function
+  (#strip! @doc "^[\\s\\*/]+|^[\\s\\*/]$")
+  (#select-adjacent! @doc @definition.function)
```
clason, line 38: That predicate does not exist.

clason (changes requested):

queries/glimmer/highlights.scm
NullVoxPopuli: this is literally a `cp` -- what I need to do is run your formatting tool in the tree-sitter-glimmer* repos

clason: Please do that before opening a PR.

NullVoxPopuli: it _is_ a draft PR -- it wasn't meant to be ready for review. 
I open draft PRs to learn what CI requirements are in repos I'm not familiar with.

queries/glimmer_javascript/tags.scm
NullVoxPopuli: resolved with inherits from javascript

NullVoxPopuli:
Looks like the tig grammar/queries are failing

queries/glimmer_typescript/indents.scm
lucario387, line 1: Can't this file be `; extends glimmer_javascript`

NullVoxPopuli: I suppose -- it's more happenstance that the nodes are named the same -- tree-sitter-glimmer-typescript doesn't extend from tree-sitter-glimmer-javascript (because they extend JavaScript and Typescript and TS has a lot of modifications on top of JS) :thinking:

lucario387: I see.

queries/glimmer/highlights.scm
NullVoxPopuli: no longer draft! :tada:

lua/nvim-treesitter/parsers.lua
```diff
@@ -704,14 +704,39 @@ list.gleam = {
 
 list.glimmer = {
   install_info = {
-    url = "https://github.com/alexlafroscia/tree-sitter-glimmer",
+    url = "https://github.com/ember-tooling/tree-sitter-glimmer",
+    branch = "main",
```
clason, line 708: Not needed if it's the default branch.

clason:
Missing lockfile entries.

queries/glimmer_typescript/injections.scm
```diff
@@ -0,0 +1,22 @@
+; inherits: typescript
+
+; extends
```
clason, line 3: No.
```suggestion
```

And avoid unnecessary (especially whitespace)  comments.

lua/nvim-treesitter/parsers.lua
```diff
@@ -704,14 +704,39 @@ list.gleam = {
...
+  filetype = "javascript.glimmer",
+  maintainers = { "@NullVoxPopuli" },
+  readme_name = "Glimmer-flavored JavaScript",
+}
+
+list.glimmer_typescript = {
+  install_info = {
+    url = "https://github.com/NullVoxPopuli/tree-sitter-glimmer-typescript",
+    branch = "main",
+    files = { "src/parser.c", "src/scanner.c", "src/tree-sitter-typescript/scanner.h" },
```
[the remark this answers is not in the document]
NullVoxPopuli: It is, it's just two fires in the same repo instead of one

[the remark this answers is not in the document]
NullVoxPopuli: ah wait! I think I get it -- that's why tree-sitter-typescript has a makefile here: https://github.com/tree-sitter/tree-sitter-typescript/blob/master/typescript/Makefile

the compile their stuff in to one file, rather than pass it downstream.
I will fix

lua/nvim-treesitter/parsers.lua
NullVoxPopuli: removed, thanks!

queries/glimmer_typescript/injections.scm
NullVoxPopuli: Removed, thanks!

lua/nvim-treesitter/parsers.lua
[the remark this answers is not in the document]
clason: Makefiles are also not acceptable. Just copy the header.

queries/glimmer_typescript/injections.scm
clason: `extends` is very much a thing; it's just not appropriate here.

lua/nvim-treesitter/parsers.lua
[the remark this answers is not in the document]
NullVoxPopuli: I mean, I did copy the header -- here: https://github.com/NullVoxPopuli/tree-sitter-glimmer-typescript/blob/main/src/tree-sitter-typescript/scanner.h :sweat_smile: 

Why is a Makefile not acceptable? how would a consumer even know, if the output ultimately ends up being what defaults are?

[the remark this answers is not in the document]
clason: Because we rely on a simple, sane build step. We simply cannot support 300 snowflake parsers. Copy the scanner **including its header** into each "dialect" subdirectory; this is non-negotiable and I don't want to waste time arguing about it.

(And, no, symlinks are not an option; these don't work on Windows.)

clason (changes requested):

lua/nvim-treesitter/parsers.lua
```diff
@@ -704,14 +704,36 @@ list.gleam = {
...
 
+list.glimmer_javascript = {
+  install_info = {
+    url = "https://github.com/NullVoxPopuli/tree-sitter-glimmer-javascript",
+    files = { "src/parser.c", "src/scanner.c" },
+    generate_requires_npm = true,
+  },
+  filetype = "javascript.glimmer",
+  maintainers = { "@NullVoxPopuli" },
+  readme_name = "Glimmer-flavored JavaScript",
```
clason, line 723:
```suggestion
```
That is sufficiently clear from the language name.

queries/glimmer_javascript/highlights.scm
```diff
@@ -0,0 +1,65 @@
+; inherits: ecma
+
+(glimmer_opening_tag) @tag.builtin
+
+(glimmer_closing_tag) @tag.builtin
+
+;
+;
+; Copied from the javascript queries
```
clason, line 9: Why not simply inherit from Javascript?

NullVoxPopuli: JavaScript includes JSX, which causes errors, because JSX is not valid is GJS files (I removed all the JSX nodes, too).

queries/glimmer_javascript/tags.scm
```diff
@@ -0,0 +1 @@
+; inherits: javascript
```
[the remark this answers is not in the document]
NullVoxPopuli: done, thanks!

lua/nvim-treesitter/parsers.lua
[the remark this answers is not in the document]
NullVoxPopuli: Gotchya -- I appreciate you're patience with me on this -- I've had a lot of loosey goosey misconceptions in my head to work through.

.h is removed now.

ObserverOfTime (approved):

clason (approved):

queries/glimmer_typescript/injections.scm
```diff
@@ -0,0 +1,20 @@
+; inherits: typescript
+
+;
+;
+; Glimmer injections are here
+;
+;
```
clason, line 7:
```suggestion
```

queries/glimmer_javascript/highlights.scm
```diff
@@ -0,0 +1,65 @@
...
+;
+;
```
clason, line 11:
```suggestion
; Copied from javascript
```

ribru17:
Sorry for my ignorance: I am having trouble understanding glimmer and in what files it should apply, should something like the following be valid?
```tsx
import { useStore } from '@builder.io/mitosis';

export default function SpecialTags() {
  const state = useStore({
    get scriptStr(): string {
      return `console.log('hello from script tag.')`;
    },
    styleStr: '.wrap { background-color: rgb(255, 0, 0); }',
  });
  return (
    <div>
      <template>
        <div>Template Tag Div</div>
      </template>
      <style data-testid="wrap-style" innerHTML={state.styleStr} />
      <div className="wrap">red content</div>
      <script data-testid="special-script" innerHTML={state.scriptStr} />
    </div>
  );
}
```
would this be a `typescript.glimmer` because of the `<template>` tag? Because right now the parser errors a bit for this ft. I may be misunderstanding in this PR in which case I think I don't have the expertise required to review it :joy:

NullVoxPopuli:
> should something like the following be valid?

it would not, that is more JSX, and JSX is not allowed.


> would this be a typescript.glimmer because of the `<template>` tag? 

If the `<template>` is used in one of two ways: as a regular expressoin, and as a class body property/expression (I don't know the exact name of things that go in the class body)

- an expression
  e.g.: 
  
  ```gjs
  const example = 1;
  const Demo = <template>{{example}}</template>;
  const Demo2 = <template>
    <Demo />
    <div>{{example}}</div>
  </template>;
  ```
- as a class body expression

  ```gjs
  const example = 1;

  class Demo {
    property = 2;

    <template>
       <p>{{this.property}}</p>
       <div>{{example}}</div>
    </template>       
  }
  ```

For more information,
GitHub already supports this if you want to look at the linguist grammars:
~~~md

```gjs
code here
```

~~~
And the TS variant:
~~~md


```glimmer-ts

// Linguist already bound gts to Gnome something, so we couldn't use the shorthand

```

~~~


Here is the community RFC where we explored several other grammars/syntaxes for having something with the benefits with both JSX and Static templates while also not having the downsides of those tools: https://github.com/emberjs/rfcs/pull/779

ribru17:
Ah, that makes sense. Thanks for clarifying!

ribru17 (approved):

PR #7071, opened 2024-08-22
fix(latex): make sure math_environment does not get double capture

DimitrisDimitropoulos:
This PR is an adaptation to the way generic environments are captured in order to make sure math_envrironment do not get double captures, thus highlighting them differently. I think this improves the highlighting, since many engines follow a similar paradigm, like legacy vim syntax. Moreover, it indicates a core difference in the code, indicated by the grammar itself. 
Here are some before and after images to showcase the differences:

I am not sure if the default behavior is to be expected, if it is feel free to close this PR.
Keep up the great work and thanks for your contribution!

### DIFF
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -53,14 +53,12 @@
 (begin
   command: _ @module
   name: (curly_group_text
-    (text) @label @nospell)
-  (#not-has-ancestor? @label math_environment))
+    (text) @label @nospell))
 
 (end
   command: _ @module
   name: (curly_group_text
-    (text) @label @nospell)
-  (#not-has-ancestor? @label math_environment))
+    (text) @label @nospell))
 
 ; Definitions and references
 (new_command_definition
@@ -320,6 +318,14 @@
 
 (math_environment
   (_) @markup.math)
+
+(math_environment
+  begin: (begin
+    name: (curly_group_text
+      text: (text) @markup.math))
+  end: (end
+    name: (curly_group_text
+      text: (text) @markup.math)))
 
 ; Comments
 [
--- a/queries/latex/highlights.scm
+++ b/queries/latex/highlights.scm
@@ -52,12 +52,14 @@
 ; General environments
 (begin
   command: _ @module
-  name: (curly_group_text (text) @label @nospell)
+  name: (curly_group_text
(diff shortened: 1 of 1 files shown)


clason:
@jdujava

queries/latex/highlights.scm
```diff
@@ -53,12 +53,14 @@
 (begin
   command: _ @module
   name: (curly_group_text
-    (text) @label @nospell))
+    (text) @label @nospell)
+  (#not-has-ancestor? @label math_environment))
```
clason, line 57: `has-ancestor` is relatively expensive, so try to avoid that at all cost. It's better to rely on query ordering to get the desired final result. (It's perfectly fine to have multiple captures apply, as long as the right one "wins", i.e., is last.)

DimitrisDimitropoulos: Thanks for the helpful insight. However the order of the queries seems to be corrected, i.e. the math_environment is second to the generic environment, but it is not the one to "win".

clason: Yes, because that only covers the environment itself. You need to override the `name` capture there. But I'm not sure this is wanted -- there's been some back and forth over having the math environment names be the same as the contents or the same as the other environments.

clason:
For the record, this pattern was deliberately removed in https://github.com/nvim-treesitter/nvim-treesitter/pull/6962/files

DimitrisDimitropoulos:
> For the record, this pattern was deliberately removed in https://github.com/nvim-treesitter/nvim-treesitter/pull/6962/files

Reading the conversation it seems like it was removed, because it was seen as redundant, which to my understanding isn't

jdujava:
As already pointed out, I deliberately removed different highlighting of `math_environment` names. I don't really remember if it was to remove the assumed redundancy, or to make highlighting of environment names the same.

Nonetheless, as of now, personally I prefer the current behavior. I don't really see why `math_environment` names should be handled differently, since there is already a different highlighting of the body. This seems to me like something one can opt in by including it in their own config, but not making it the default (to me it seems less consistent).

clason:
Well, the original rationale was that this is how VimTeX (the gold standard for LaTeX syntax highlighting) behaves. But I care most about consistency, which means we'll stick with the current behavior for now.

PR #7078, opened 2024-08-25
feat(http): update injections queries

boltlessengineer:
Updated `http/injections.scm` file following latest release of [tree-sitter-http](https://github.com/rest-nvim/tree-sitter-http).

This enables response handler script language injections for cases like:

```http
GET example.com

# @lang=lua
{%
-- lua code
%}
```

### DIFF
--- a/queries/http/injections.scm
+++ b/queries/http/injections.scm
@@ -13,28 +13,12 @@
   (#set! injection.language "graphql"))
 
 ; Script (default to javascript)
-((script) @injection.content
-  (#offset! @injection.content 0 2 0 -2)
-  (#set! injection.language "javascript"))
-
-; Script with other languages
-((comment
+((#set! injection.language "javascript")
+(comment
   name: (_) @_name
   (#eq? @_name "lang")
-  value: (_) @injection.language)
-  .
-  (_
-    (script) @injection.content
-    (#offset! @injection.content 0 2 0 -2)))
-
-; post-request scripts for requests without body
-((request
-  !body
-  (comment
-    name: (_) @_name
-    (#eq? @_name "lang")
-    value: (_) @injection.language) .)
-  .
-  (res_handler_script
-    (script) @injection.content
-    (#offset! @injection.content 0 2 0 -2)))
+  value: (_) @injection.language)?
+.
+(_
+  (script) @injection.content
+  (#offset! @injection.content 0 2 0 -2)))
--- a/queries/http/injections.scm
+++ b/queries/http/injections.scm
@@ -13,12 +13,12 @@
   (#set! injection.language "graphql"))
 
 ; Script (default to javascript)
-((#set! injection.language "javascript")
-(comment
+((comment
   name: (_) @_name
   (#eq? @_name "lang")
   value: (_) @injection.language)?
-.
-(_
-  (script) @injection.content
-  (#offset! @injection.content 0 2 0 -2)))
+  .
+  (_
+    (script) @injection.content
+    (#offset! @injection.content 0 2 0 -2))
+  (#set! injection.language "javascript"))
--- a/queries/http/injections.scm
+++ b/queries/http/injections.scm
@@ -26,3 +26,15 @@
   (_
     (script) @injection.content
     (#offset! @injection.content 0 2 0 -2)))
+
+; post-request scripts for requests without body
+((request
+  !body
+  (comment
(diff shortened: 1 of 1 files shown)


queries/http/injections.scm
```diff
@@ -26,3 +26,15 @@
   (_
     (script) @injection.content
     (#offset! @injection.content 0 2 0 -2)))
+
+; post-request scripts for requests without body
+((request
+  !body
```
clason, line 32: I don't think that actually works, though...

boltlessengineer: It is there to match `script` tag in cases like:
```http
GET localhost:9999

# @lang=lua
> {%
-- comment
%}
```
IMG_1337

comment `# @lang=lua` should not close the `request` node to be used as body wrapper like `#@_RES … # @_END`. But we need “script node with variable declaration comment on top of it” query.
So that is what this query does; match `script` node inside `res_handler_script` that comes right after the `request` node ending with `comment` node with `lang` variable declaration.

ribru17:
Due to an earlier query, now this will inject both javascript and whatever the intended language is. I think the javascript one should either be removed or, if it should be the default, it should only apply when not setting a specific language

queries/http/injections.scm
clason: I mean the `!body` negated node -- I don't think that's supported in Neovim. (At least I've never seen it used in any query here.)

boltlessengineer: Didn’t know about that, thank you for pointing out.
I’ll refactor the parser to simplify the queries.

Will this query be fine to be merged?

```scheme
; Script (default to javascript)
((#set! injection.language "javascript")
  (comment
    name: (_) @_name
    (#eq? @_name "lang")
    value: (_) @injection.language)?
  .
  (_
    (script) @injection.content
    (#offset! @injection.content 0 2 0 -2)))
```

boltlessengineer:
@ribru17 I updated the parser and query to not make duplicated matches.
Can you review the PR again?

queries/http/injections.scm
```diff
@@ -13,16 +13,12 @@
   (#set! injection.language "graphql"))
 
 ; Script (default to javascript)
-((script) @injection.content
-  (#offset! @injection.content 0 2 0 -2)
-  (#set! injection.language "javascript"))
-
-; Script with other languages
-((comment
+((#set! injection.language "javascript")
```
lucario387, line 16: can you move this downwards? i.e. after the `(_ (script))` block

boltlessengineer: Done. I thought it won’t work.
I put it there to overwrite after setting `injection.language` as javascript, but seems like order doesn’t matter?

ribru17: I think the order doesn't matter, but now for instance I only get javascript injections. I am using
```http
# vim: ft=http
### post-request script
GET https://jsonplaceholder.typicode.com/posts/3

# @lang=lua
> {%
local body = vim.json.decode(response.body)
client.global.set("postId", body.id)
%}

### run request with variable

GET https://jsonplaceholder.typicode.com/posts/{{postId}}

> {%
  client.global.set("auth_token", response.body.jwt);
%}
```

I think `javascript` and generalized injections should be two separate queries, if you can manage to get them to always be disjoint

boltlessengineer: Did you tested with latest http parser?
I updated the parser itself too.

IMG_1344

I opened your snippet with latest http parser installed manually but the language injection seems to be working. No error node in `:InspectTree` and `:Inspect` points to lua captures.

ribru17: Oh oops, nah didnt update the parser sorry. I'll test again in one moment

ribru17 (approved):

clason:
Needs a rebase on `master`, then LGTM.

boltlessengineer:
Done. I forgot there was a cron job for auto updating parsers. Sorry about that.

clason:
No worries!

lucario387:
This PR does show that one can set a fallback injection language for a query :thinking: 

With an example query of
```scheme
(_
  (foo)? @injection.language
  .
  (bar) @injection.content
  ; One of these will work
  ; (#set! injection.language "markdown")
  ; (#set! injection.self)
  ; (#set! injection.parent)
```

Will this be useful in other scenarios :thinking:

PR #7086, opened 2024-08-26
leo/bugfix small bugfixes of variables, records, and struct highlighting

r001:
small bugfixes for leo language variables, records, and structs

### DIFF
--- a/queries/leo/highlights.scm
+++ b/queries/leo/highlights.scm
@@ -150,6 +150,14 @@
   (locator
     (identifier) @function))
 
+(associated_function_call
+  (named_type
+    (identifier
+      (constant_identifier) @function)))
+
+(associated_function_call
+    (identifier) @function.call)
+
 (record_type
   (locator
     (identifier) @variable.member))
--- a/queries/leo/highlights.scm
+++ b/queries/leo/highlights.scm
@@ -176,7 +176,7 @@
   name: (identifier) @type.definition)
 
 (variable_declaration
-    (identifier) @variable)
+  (identifier) @variable)
 
 [
   (address_literal)
--- a/queries/leo/highlights.scm
+++ b/queries/leo/highlights.scm
@@ -1,7 +1,3 @@
-(variable) @variable
-
-(constant_identifier) @constant
-
 [
   "assert"
   "assert_eq"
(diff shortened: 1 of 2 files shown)


queries/leo/highlights.scm
```diff
@@ -173,11 +173,10 @@
   (identifier) @variable.parameter)
 
 (struct_declaration
-  name: (identifier) @variable.member)
+  name: (identifier) @type.definition)
```
ribru17, line 176: This gets overridden by `@constant` if the name is capitalized, is this intended? Same with records

clason: Unless that is an official part of the grammar (in which case the parser should handle it), we should stop this sort of case-dependent convention captures.

r001: Nope, it is part of my learning curve! Fixing it now.

queries/leo/highlights.scm
```diff
@@ -115,7 +113,15 @@
 (boolean_literal) @boolean
 
 (constant_declaration
-  (identifier) @constant)
+  (identifier
+    (constant_identifier) @constant))
+
+(variable
+  (constant_identifier) @constant)
```
lucario387, line 120: Question: Can this be reduced to
```suggestion
(constant_identifier) @constant
```

r001:
Ddont merge yet please!

r001:
Now it can be merged if you don't find any errors

queries/leo/highlights.scm
clason: No, see below.

r001: @lucario387 This was my original plan, but unfortunately for me, records, and structs both have in their struct_component_declarations constant_identifiers, but I dont want them to LOOK like constants.
If I accept your change this will override struct_component_declarations. If you check the comments of @ribru17 above, he had issues with exactly this. 
So I would like to keep things as they are.

queries/leo/highlights.scm
r001: Guys I think I did the necessary changes, will resolve this conversation.

ribru17 (approved):

PR #7094, opened 2024-08-29
(leo) fix: import file extension changed from .leo to .aleo

r001:
before:
import program.leo

now:
import program.aleo


clason:
I don't see what the actual change has to do with the PR description? In any case, I made the change in the automated update lockfile.

PR #7097, opened 2024-08-30
feat!: switch upstream Zig parser

amaanq:
The new parser is faster, does not lag while editing, correctly parses the entire Zig code base, is much easier to write queries for, and the queries have much better highlighting.


queries/zig/highlights.scm
```diff
@@ -221,19 +261,20 @@ field_constant: (IDENTIFIER) @constant
...
+  "."
+  ","
+  ":"
+  "=>"
+  "->"
+] @punctuation.delimiter
 
-(PtrIndexPayload
+(payload
   "|" @punctuation.bracket)
```
ribru17, line 274: Would this make sense also?

```query
(payload
```

queries/zig/folds.scm
```diff
@@ -1,14 +1,22 @@
...
+  (if_statement)
+  (while_statement)
+  (for_statement)
+  (if_expression)
+  (else_clause)
+  (for_expression)
+  (while_expression)
+  (if_type_expression)
+  (function_signature)
+  (parameters)
```
ribru17, line 16: Maybe `(call_expression)` here as well? Or too much

ribru17 (approved):

queries/zig/folds.scm
amaanq: why not

PR #7102, opened 2024-08-30
feat(edl): highlights for EDL

mrdgo:
EDL is intel's enclave definition language to define SGX enclaves. These files probably never need anything more than highlights.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -470,7 +470,6 @@ list.edl {
   install_info = {
     url = "https://github.com/Andrade/tree-sitter-edl",
     files = { "src/parser.c" },
-    branch = "master",
   },
   maintainers = { "@mrdgo" },
 }
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -466,6 +466,15 @@ list.editorconfig = {
   maintainers = { "@ValdezFOmar" },
 }
 
+list.edl {
+  install_info = {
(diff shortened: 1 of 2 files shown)


clason:
Well, they do need a parser, otherwise they are worthless...

mrdgo:
> Well, they do need a parser, otherwise they are worthless...

[not mine](https://github.com/Andrade/tree-sitter-edl)

clason:
Well, so, you need to add this parser here. Otherwise the queries won't work.

lua/nvim-treesitter/parsers.lua
```diff
@@ -466,6 +466,15 @@ list.editorconfig = {
   maintainers = { "@ValdezFOmar" },
 }
 
+list.edl {
+  install_info = {
+    url = "https://github.com/Andrade/tree-sitter-edl",
+    files = { "src/parser.c" },
+    branch = "master",
```
clason, line 473: Not needed unless it's a non default branch

clason (changes requested):

mrdgo:
happy to remove the branch. But I am not sure if I want to fight that pipeline

clason:
> happy to remove the branch. But I am not sure if I want to fight that pipeline 

It's not optional, though. Quality control is important. If you are not willing to make the effort (and keep doing so when updates are required), it's better to close the PR now to avoid wasting both our time.

mrdgo:
> Also, never open a PR from your master branch. Create a feature branch first, make your changes, and then open a PR.

history beautification reasons, right?

clason:
No, so CI works properly and maintainers can push to the PR to fix issues such as the current ones.

PR #7104, opened 2024-08-31
feat(ini): highlight setting value as `@string`

ValdezFOmar:
Before (no highlight):

image

After:

image

### DIFF
--- a/queries/ini/highlights.scm
+++ b/queries/ini/highlights.scm
@@ -13,4 +13,4 @@
 (setting
   (setting_name) @property)
 
-(setting_value) @string ; grammar does not support subtypes
+(setting_value) @string
--- a/queries/ini/highlights.scm
+++ b/queries/ini/highlights.scm
@@ -13,4 +13,4 @@
 (setting
   (setting_name) @property)
 
-; (setting_value) @none ; grammar does not support subtypes
+(setting_value) @string ; grammar does not support subtypes


queries/ini/highlights.scm
```diff
@@ -13,4 +13,4 @@
 (setting
   (setting_name) @property)
 
-; (setting_value) @none ; grammar does not support subtypes
+(setting_value) @string ; grammar does not support subtypes
```
clason, line 16: Is the comment still relevant?

clason: What if the `option = 1.0`? That should _not_ be captured as `@string`. I think the comment still applies and prevents this capture.

clason:
Sorry, we can't do this until the parser starts distinguishing different value types.

queries/ini/highlights.scm
ValdezFOmar: But that is not guaranteed to be a number, a program may read that as whatever it wants since there is no spec to how to read an ini file. Also, for the [editorconfig queries](https://github.com/nvim-treesitter/nvim-treesitter/pull/6855#discussion_r1665476308) it was point out that default values should be highlighted as strings by default.

clason: Hmm, not sure I agree with that comment. The legacy syntax file also distinguishes them. If anything, it should be `@identifier` then.

ValdezFOmar: I don't see how that should be an `@identifier` and I don't see that capture in `:h treesitter-highlight-groups` or `CONTRIBUTING.md`

clason: Sorry, name is `@variable`.

clason: If the argument is that the value is _always_ treated as a string literal in these files, that's different, of course. I just disagree with the "default value" reasoning.

clason:
```
(section_name
 (text) @type) ; consistency with toml
```

Unrelated, but I don't think that is correct -- it's a literal title so it should be `@markup.title`. If it has programmatic meaning, `@module` is the right capture. @ribru17 ?

clason:
Sorry for putting the brakes on such a simple PR, but this surfaces a generic issue with "ini-like" languages I would like to clear up first and handle consistently. (My concern is that we're blurring the line here between syntax -- which should be the sole focus here -- and semantics by imposing a distinction on value types that the language specification explicitly doesn't make.)

clason:
The issue here is that the parser is minimal and follows the spec, so doesn't make any distinction between the types of values. (Neither does the editorconfig spec, for the record, but your parser does make a distinction -- so it's your problem, not ours ;))

We could make the same distinction here using `#lua-match?` (for numbers) and `#any-of?` (for booleans) -- the question is whether we _should_, given the specification. (The parser is unlikely to make such a change, although one could open an issue and ask.)

clason:
In any case, this PR is an improvement so can be merged (just remove the comment for now, it's not helpful); further type distinction can still be added in a follow-up PR.

I'll leave this open for a bit to give the general questions a bit more visibility and will merge later.

(Another follow-up will have to take care of the section naming across grammars.)

ribru17:
> ```
> (section_name
>  (text) @type) ; consistency with toml
> ```
> 
> Unrelated, but I don't think that is correct -- it's a literal title so it should be `@markup.title`. If it has programmatic meaning, `@module` (for "namespaces") is the right capture. @ribru17 ?

Yeah `markup.title` makes sense to me, I could also see label maybe but I think title makes more sense

clason:
(I looked at editorconfig and TOML; these are quite different so should not matter here.)

PR #7107, opened 2024-09-01
feat(wing): add many missing highlights

ribru17:
...as well as some fold and injection queries.

Test code:

```wing
bring expect;
let then: datetime = datetime.utcNow();
let now: datetime = then;


let isOlderOrEqual = inflight (x: datetime, y: datetime) => {
 return y.timestampMs >= x.timestampMs;
}; 

test "datetime is a valid type" {
  let later: datetime = datetime.fromComponents({
      day: now.dayOfMonth, 
      hour: now.hours, 
      year: now.year + 1, 
      min: now.min, 
      month: now.month, 
      ms: now.ms, 
      sec: now.sec, 
      tz: now.timezone
    });

  expect.ok(isOlderOrEqual(then, now));
  expect.ok(isOlderOrEqual(now, later));

  hi += 5;
}


let decimal: regex = regex.compile("^[0-9]*$");

let testString = inflight (re: regex, s: str) => {
  return re.test(s);
};

test "regex is valid type" {
  let binary: regex = regex.compile("^[0-1]*$");
  expect.equal(testString(decimal, "24"), true);
  expect.equal(testString(decimal, "340523"), true);
  expect.equal(testString(decimal, "a23"), false);

  expect.equal(testString(binary, "01101"), true);
  expect.equal(testString(binary, "0"), true);
  expect.equal(testString(binary, "120010"), false);
}
// vim: ft=wing
```

```wing
bring cloud;
bring "./subdir/util.w" as myutil;

pub struct StoreOptions {
  name: str?;
}

pub class Store {
  handlers: MutArray<inflight (str): void>;
  data: cloud.Bucket;

  new(options: StoreOptions?) {
    this.data = new cloud.Bucket();
    this.handlers = MutArray<inflight (str): void> [];
  }

  extern "./util.js" pub static makeKey(name: str): str;
  extern "./util.js" pub static inflight makeKeyInflight(name: str): str;

  pub inflight set(message: str) {
    this.data.put("data.txt", myutil.double(message));
    for handler in this.handlers {
      handler(message);
    }
  }

  pub onSet(handler: inflight (str): void) {
    this.handlers.push(handler);
  }
}

pub interface PublicInterface {}

internal interface InternalInterface {}

internal class InternalClass {
  static internal internalStaticMethod() {}
}

pub class PublicClass impl PublicInterface, InternalInterface {
  static internal internalStaticMethod() {}
  pub publicField: num;
  internal internalField: num;

  new() {
    this.publicField = 42;
    this.internalField = 42;
  }

  pub publicMethod() {}
  internal internalMethod() {}
}

pub struct PublicStruct {}
internal struct InternalStruct {}
// vim: ft=wing
```

### DIFF
--- a/queries/wing/highlights.scm
+++ b/queries/wing/highlights.scm
@@ -8,16 +8,33 @@
 (custom_type) @type
 
 (class_field
-  name: (identifier) @variable.member)
+  name: (identifier) @property)
+
+(struct_field
+  name: (identifier) @property)
 
 (class_definition
+  name: (identifier) @type)
+
+(struct_definition
+  name: (identifier) @type)
+
+(interface_definition
   name: (identifier) @type)
 
 (method_definition
   name: (identifier) @function.method)
 
+(json_literal_member
+  (identifier) @property)
+
 ; Functions
 (keyword_argument_key) @variable.parameter
+
+(parameter_definition
+  name: (identifier) @variable.parameter)
+
+(variadic) @variable.parameter.builtin
 
 (call
   caller: (reference
@@ -37,7 +54,17 @@
 
 (bool) @boolean
 
-(builtin_type) @type.builtin
+[
+  (builtin_type)
+  "MutSet"
+  "MutMap"
+  "MutArray"
+  "Json"
+  "Set"
+  "Map"
+  "Array"
+  "MutJson"
+] @type.builtin
 
 (json_container_type) @type.builtin
 
@@ -45,16 +72,12 @@
 (comment) @comment @spell
 
 [
-  "("
-  ")"
-  "{"
-  "}"
-] @punctuation.bracket
-
-[
   "-"
+  "-="
   "+"
+  "+="
   "*"
+  "**"
   "/"
   "%"
   "<"
@@ -68,23 +91,54 @@
   "&&"
   "??"
   "||"
+  "?"
 ] @operator
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+  "["
+  "]"
+] @punctuation.bracket
+
+(mutable_container_type
+  [
+    "<"
+    ">"
+  ] @punctuation.bracket)
+
+(immutable_container_type
+  [
+    "<"
+    ">"
+  ] @punctuation.bracket)
 
 [
   ";"
   "."
   ","
+  ":"
+  "=>"
 ] @punctuation.delimiter
 
 [
   "as"
-  "bring"
   "let"
   "new"
   (inflight_specifier)
+  "impl"
+  "test"
 ] @keyword
 
-"class" @keyword.type
+"bring" @keyword.import
+
+[
+  "class"
+  "struct"
+  "interface"
+] @keyword.type
 
 [
   "for"
@@ -100,6 +154,26 @@
   "pub"
   "protected"
   "internal"
+  "extern"
+  (static)
 ] @keyword.modifier
 
 "return" @keyword.return
+
+(import_statement
+  module_name: (identifier) @module)
+
+(import_statement
+  alias: (identifier) @module)
+
+(call
+  (reference
+    (nested_identifier
+      object: (reference) @_ref
+      property: (member_identifier) @_ident))
+  (argument_list
+    (positional_argument
+      (string) @string.regexp))
+  (#eq? @_ref "regex")
+  (#eq? @_ident "compile")
+  (#offset! @string.regexp 0 1 0 -1))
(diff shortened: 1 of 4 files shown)


clason:
@gshpychka, @MarkMcCulloh

queries/wing/highlights.scm
```diff
@@ -37,24 +54,31 @@
...
+  (builtin_type)
+  "MutSet"
+  "MutMap"
+  "MutArray"
+  "Json"
+  "Set"
+  "Map"
+  "Array"
+  "MutJson"
+  "Promise"
```
MarkMcCulloh, line 67:
```suggestion
```

`Promise` is not yet a valid type in wing

queries/wing/highlights.scm
```diff
@@ -100,6 +155,26 @@
...
+  (reference
+    (nested_identifier
+      object: (reference) @_ref
+      property: (member_identifier) @_ident))
+  (argument_list
+    (positional_argument
+      (string) @string.regexp))
+  (#eq? @_ref "regex")
+  (#eq? @_ident "compile")
+  (#offset! @string.regexp 0 1 0 -1))
```
MarkMcCulloh, line 180: While this is very cool I think highlighting implies that the string is special or is handled in a special way (e.g. different escape sequences) when it's really just a normal string.

MarkMcCulloh (commented):
Really nice improvements, thank you!

queries/wing/highlights.scm
clason: No, you just get "smarter" highlighting for free; it's still a string in the end. Unless the language has no support at all for regular expressions?

MarkMcCulloh: It has no support at a language level, no. The standard library has a type that allows you to write `regex.compile("^[0-1]*$")` but the string there is a simple string, it can also be interpolated or you can also pass a variable.

clason: Still, users would like to have the regex highlighted (consistently) to make it easier to parse -- which is what this parser injection ensures.

(Remember, the goal here is to use tree-sitter for better and more consistent syntax highlighting, not linting or otherwise enforcing an abstract grammar.)

clason: > it can also be interpolated or you can pass a variable.

But if there's a serious risk of mis-parsing, that's different of course.

MarkMcCulloh: Yeah the risk is misparsing which will be misleading to users, specifically with regards to the interpolation braces `{}` and escape sequences.
Wing also has a non-interpolated string literal, but escape sequences will still be a problem there.

clason: Then the function name is misleading, too ;) If you can't assume that the argument of  `regex.compile` is a regular expression, something has gone wrong in your standard library design...

MarkMcCulloh: The function will throw if the argument is not a valid regex string. There's no actual intrinsic regex type nor syntax, so there's no way to convey that more strictly in the type signature. Wing is still an early language so that's all subject to change.

clason: Again, the purpose here is _not_ linting, that is out-of-scope. We want to correctly highlight _correct_ code and do not care about incorrect code. So just because the function _can_ take non-regex and give an error instead of crashing, that is _not_ sufficient reason to not give the (expected) argument a more useful highlight.

The situation is different if users are _expected_ to put in a non-regexp string in a _correct_ situation (in which case the function name is weird, but you do do you).

MarkMcCulloh: I fully understand that, I'm trying to convey that wing's strings with interpolation and escapes will both look and become valid regex, but the highlights will be misleading:

```wing
let someText = "hi";
let validRegex = regex.compile("{someText[0]}+");
```

In this example, the final regex will actually be `/h+/` due to interpolation, but the highlighting will make it seem like the regex is `/{someText[0]}+/ because it's seeing it all as a regex.

clason: An example says more than a thousand words :) Yes, then this injection is unsafe and should not be added.

ribru17: I'm not desperate to add this, but I think the injection is still safe: interpolations are exposed under child of string (`(template_substitution)`) and thus the injection is not applied to them. The only weird case I can think of would be something like `regex.compile("{var}-9]")` where `var` is something like `"[0"` (causing unmatched brackets from the injection's point of view), but I imagine cases like that would be insanely uncommon. If this is still to dangerous, I don't mind removing it

MarkMcCulloh: > interpolations are exposed under child of string and thus the injection is not applied to them

Ah ok nice, I didn't know injections worked like that. In that case I'm cool with this and it should be fine with escape sequences as well. 👍

PR #7108, opened 2024-09-01
feat(highlights): capture headings as `@markup.heading` in ini-like gr…

ValdezFOmar:
…ammars

Follow up of https://github.com/nvim-treesitter/nvim-treesitter/pull/7104#issuecomment-2322857504

### DIFF
--- a/queries/editorconfig/highlights.scm
+++ b/queries/editorconfig/highlights.scm
@@ -1,7 +1,7 @@
 (comment) @comment @spell
 
 (section
-  (section_name) @markup.heading)
+  (section_name) @string.special.path)
 
 (character_choice
   (character) @constant)
--- a/queries/editorconfig/highlights.scm
+++ b/queries/editorconfig/highlights.scm
@@ -1,8 +1,7 @@
 (comment) @comment @spell
 
-; consistency with toml and ini
 (section
-  (section_name) @type)
+  (section_name) @markup.heading)
 
 (character_choice
   (character) @constant)
--- a/queries/toml/highlights.scm
+++ b/queries/toml/highlights.scm
@@ -1,6 +1,6 @@
 ; Properties
 ;-----------
-(bare_key) @markup.heading
+(bare_key) @type
 
 (quoted_key) @string
 
--- a/queries/toml/highlights.scm
+++ b/queries/toml/highlights.scm
@@ -1,6 +1,6 @@
 ; Properties
 ;-----------
-(bare_key) @type
+(bare_key) @markup.heading
 
 (quoted_key) @string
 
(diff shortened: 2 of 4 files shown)


ValdezFOmar:
Unrelated, but I find the toml highlight queries a bit strange compared to others, would it be worth to make a PR refactoring it?

queries/toml/highlights.scm
```diff
@@ -1,6 +1,6 @@
 ; Properties
 ;-----------
-(bare_key) @type
+(bare_key) @markup.heading
```
ribru17, line 3: Maybe could we change this to 
```
Maybe could we change this to
```suggestion
(table
  (bare_key)) @markup.heading
```
many color schemes have bolded headings, and it would be annoying for the bold attribute to carry over to regular properties. Shame the grammar defines them as the same node...

ribru17:
> Unrelated, but I find the toml highlight queries a bit strange compared to others, would it be worth to make a PR refactoring it?

Sure! Thanks

queries/toml/highlights.scm
ValdezFOmar: Yeah, this is one of the reasons why I said this https://github.com/nvim-treesitter/nvim-treesitter/pull/7108#issuecomment-2323113344 , the refactor that I have in my mind solves this.

ValdezFOmar: I think it would be better to leave it as is in this PR and solve it in the refactor since there is more special cases to this problem (dotted keys and tables array headings).

clason: I had planned on tackling this, since I spent some time reading the TOML specs. As far as I can tell, headings should _not_ be marked as headings here, since they are in fact tables (i.e., common dotted keys). So my solution would be to remove this highlight entirely and mark _all_ keys -- whether in pairs or in tables -- the same.

queries/editorconfig/highlights.scm
```diff
@@ -1,8 +1,7 @@
 (comment) @comment @spell
 
-; consistency with toml and ini
 (section
-  (section_name) @type)
+  (section_name) @markup.heading)
```
clason, line 4: Again, I don't think these are actual headings -- they are globs and should be highlighted as such: `@string.special.path` or `@regexp`.

ValdezFOmar: Changed to `@string.special.path`

queries/toml/highlights.scm
clason: Maybe leave this change out of this PR for now and make it part of the TOML refactor (where we can discuss specs and intention in a holistic fashion)?

queries/editorconfig/highlights.scm
clason: @gpanders

queries/toml/highlights.scm
ValdezFOmar: I'm fine with that.

Issue #5445, opened 2023-09-25
`indent.branch` not working properly


lucario387:
### Describe the bug

Like the title describe

### To Reproduce

1. Create a `test.html` file with the following code
```html
<html>
  <body>
</html>
```
3. On line 2, press `o</body>`

### Expected behavior

Expect `</body>` to get dedented to the same level as `<body>`

### Output of `:checkhealth nvim-treesitter`

```markdown
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- OK `tree-sitter` found 0.20.8 (d4c1bf7ce78051b7f4a381d1508d68928512ed5f) (parser generator, only needed for :TSInstallFromGrammar)
- OK `node` found v18.16.1 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: cc (GCC) 13.2.1 20230801
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "6.5.4-arch2-1",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Thu, 21 Sep 2023 11:06:39 +0000"
} ~

Parser/Features         H L F I J
  - html                ✓ ✓ ✓ ✓ ✓

  Legend: H[ighlight], L[ocals], F[olds], I[ndents], In[j]ections
         +) multiple parsers found, only one will be used
         x) errors found in the query, try to run :TSUpdate {lang} ~
```

### Output of `nvim --version`

```text
NVIM v0.10.0-dev-8195c1800
Build type: Release
LuaJIT 2.1.1695409640
Run "nvim -V1 -v" for more info
```

### Additional context

This also happens to #5437 and was first reported on nvim-treesitter matrix server

PR #7110, opened 2024-09-01
feat(gap): Add gap and gaptst parser and queries

reiniscirpons:
[GAP](https://www.gap-system.org/) is an interpreted programming language for computational discrete algebra. This PR adds support for the GAP system programming language (filetype `gap`) and the GAP system test file format (filetype `gaptst`).

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -633,6 +633,24 @@ list.fusion = {
   maintainers = { "@jirgn" },
 }
 
+list.gap = {
+  install_info = {
+    url = "https://github.com/gap-system/tree-sitter-gap",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@reiniscirpons" },
+  readme_name = "GAP system",
+}
+
+list.gaptst = {
+  install_info = {
+    url = "https://github.com/gap-system/tree-sitter-gaptst",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@reiniscirpons" },
+  readme_name = "GAP system test files",
+}
+
 list.gdscript = {
   install_info = {
     url = "https://github.com/PrestonKnopp/tree-sitter-gdscript",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -646,7 +646,6 @@ list.gaptst = {
   install_info = {
     url = "https://github.com/gap-system/tree-sitter-gaptst",
     files = { "src/parser.c", "src/scanner.c" },
-    branch = "main",
   },
   maintainers = { "@reiniscirpons" },
   readme_name = "GAP system test files",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -646,6 +646,7 @@ list.gaptst = {
   install_info = {
     url = "https://github.com/gap-system/tree-sitter-gaptst",
     files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
   },
   maintainers = { "@reiniscirpons" },
   readme_name = "GAP system test files",
--- a/queries/gap/highlights.scm
+++ b/queries/gap/highlights.scm
@@ -1,0 +1,219 @@
+; NOTE: (reiniscirpons) in case multiple queries match, last query wins. So
+; queries should go from least specific to most specific. (This is the default
+; behaviour since tree-sitter 0.22.2)
+(identifier) @variable
+
+; Constants
+; convention: constants are of the form ALL_CAPS_AND_UNDERSCORES and have length at least 2
+((identifier) @constant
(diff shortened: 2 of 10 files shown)


queries/gap/highlights.scm
```diff
@@ -0,0 +1,219 @@
+; NOTE: (reiniscirpons) in case multiple queries match, last query wins. So
+; queries should go from least specific to most specific. (This is the default
+; behaviour since tree-sitter 0.22.2)
```
clason, line 3:
```suggestion
```
Not needed here ;)

queries/gap/highlights.scm
```diff
@@ -0,0 +1,219 @@
...
+(identifier) @variable
+
+; Constants
+; convention: constants are of the form ALL_CAPS_AND_UNDERSCORES and have length at least 2
+((identifier) @constant
+  (#match? @constant "^[A-Z_][A-Z_]+$"))
```
clason, line 9:
```suggestion
  (#lua-match? @constant "^[A-Z_][A-Z_]+$"))
```
Avoid regex at any cost; these are really expensive in Neovim.

Prefer `#eq?` > `#any-of?` >> `#lua-match?` >>> `#match?`.

queries/gap/highlights.scm
```diff
@@ -0,0 +1,219 @@
+; NOTE: (reiniscirpons) in case multiple queries match, last query wins. So
+; queries should go from least specific to most specific. (This is the default
+; behaviour since tree-sitter 0.22.2)
+(identifier) @variable
+
+; Constants
+; convention: constants are of the form ALL_CAPS_AND_UNDERSCORES and have length at least 2
```
clason, line 7: We're trying to get away from this convention-based highlighting. It's not worth the headache, all things considered.

lua/nvim-treesitter/parsers.lua
```diff
@@ -646,6 +646,7 @@ list.gaptst = {
   install_info = {
     url = "https://github.com/gap-system/tree-sitter-gaptst",
     files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
```
clason, line 649:
```suggestion
```
Only needed if that's not the default branch. What is missing is `lockfile.json` entries.

reiniscirpons: Ah that makes sense, thanks! I've removed the branch and updated the lockfile, hopefully CI passes now.

clason: Looks good, yes! (Sorry, didn't catch this earlier.)

clason:
Would it be possible to add indents?

clason:
Oh, and please add injections, at least for `comment`.

reiniscirpons:
> Would it be possible to add indents?

I tried to add queries for indents earlier, however was not able to get a satisfactory result. One thing I kept running into was an issue akin to issue https://github.com/nvim-treesitter/nvim-treesitter/issues/5445 where the branch query would not properly dedent closing tags, e.g. I would get

```
if x = 2 then
  Print("a");
  fi;
```

Instead of 

```
if x = 2 then
  Print("a");
fi;
```

Even though I had a query of the form `"fi" @indent.branch @indent.end`. In the end the indentation I was getting seemed like it would get in the way of the end users so I decided to not include indent queries for now, at least until some of the issues are ironed out on the nvim-treesitter side. Happy to make a PR for these later though!

> Oh, and please add injections, at least for `comment`.

Good point, added injections now.

clason:
Yeah, no sense fighting the indentations right now if they don't work out of the box.

queries/gaptst/highlights.scm
```diff
@@ -0,0 +1,19 @@
+(output_line) @string.documentation
```
ribru17, line 1: I am wondering if there is a better capture for this? Not sure if plain `@string` would make it too confusing with regular code but it also doesn't really seem like documentation. But I guess test output is in a way documenting what should happen :thinking:

clason: Good question; I don't think we have any guidance for "session/notebook" languages yet.

Unless we come up with something better, I think `@markup.raw.block` would be my suggestion.

queries/gap/highlights.scm
```diff
@@ -0,0 +1,211 @@
...
+    (identifier) @variable.parameter.builtin) .)
+  (#eq? @variable.parameter.builtin "arg"))
+
+((lambda_parameters
+  .
+  (identifier) @variable.parameter.builtin .)
+  (#eq? @variable.parameter.builtin "arg"))
+
+(locals
+  (identifier) @variable.parameter)
```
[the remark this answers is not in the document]
reiniscirpons: Thanks for pointing this out. The language is not special, I was just overzealous with the captures :)

queries/gaptst/highlights.scm
reiniscirpons: `@markup.raw.block` definitely conveys the semantics of the `(output_line)` nodes better. They are essentially the raw expected output of the test cases. I initially put `string.documentation` since I couldn't find a better option. Thanks for the suggestion!

ribru17 (approved):

PR #7118, opened 2024-09-03
refactor(toml): highlight queries

ValdezFOmar:
Follow up of https://github.com/nvim-treesitter/nvim-treesitter/pull/7108

I've used `@markup.heading` for now since that was the original idea, but this should be discussed further.

### DIFF
--- a/queries/toml/highlights.scm
+++ b/queries/toml/highlights.scm
@@ -1,23 +1,32 @@
-; Properties
-;-----------
-(bare_key) @type
+(table
+  [
+    (bare_key) @markup.heading
+    (dotted_key
+      (bare_key) @markup.heading)
+  ])
 
-(quoted_key) @string
+(table_array_element
+  [
+    (bare_key) @markup.heading
+    (dotted_key
+      (bare_key) @markup.heading)
+  ])
 
 (pair
-  (bare_key)) @property
+  [
+    (bare_key) @property
+    (dotted_key
+      (bare_key) @property)
+  ])
 
-(pair
-  (dotted_key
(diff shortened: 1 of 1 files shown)


clason:
As I commented there, my reading of the spec is that "headings" are actually not headings but "common dotted keys", so I think they should be highlighted as such. This would mean to completely ignore the `(table)` nodes and rely on the existing `(key)` highlight.

ribru17:
> As I commented there, my reading of the spec is that "headings" are actually not headings but "common dotted keys", so I think they should be highlighted as such. This would mean to completely ignore the `(table)` nodes and rely on the existing `(key)` highlight.

I think this seems like the way to go, especially when looking at something like

```toml
[summaries.default.host-platform]
triple = "x86_64-unknown-linux-gnu"
target-features = "all"
```

In this PR `summaries` and `default` don't get a highlight, which they should. And I think it would make sense to give all of these `@variable.member` or `@property` (for the first one?). I think now I am overthinking it but something like that.

---

That said, I do think that practically it would be helpful for it to have some differentiating (additional?) highlight, compared to regular keys... maybe we could give it a title or label highlight as a lower priority? :laughing: Idk if that is discouraged here or not, if it is oh well

clason:
> That said, I do think that practically it would be helpful for it to have some differentiating (additional?) highlight, compared to regular keys... maybe we could give it a title or label highlight as a lower priority? 😆 Idk if that is discouraged here or not, if it is oh well

I think the brackets are enough to make them stand out.

ValdezFOmar:
Done!

PR #7119, opened 2024-09-05
feat(latex): parse ```math as LaTeX

jyc:
Thanks for making nvim-treesitter! I use it every day; it's very useful.

This change makes <code>```math</code> fenced code blocks render using LaTex syntax highlighting.

For context, GitHub-flavored Markdown renders <code>```math</code> fenced code blocks by first parsing them as LaTeX (c.f. <code>```latex</code> which will render as raw LaTeX source): https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions#writing-expressions-as-blocks

I've been using the following injection at `after/queries/markdown/injections.scm`:

```scm
; extends
; Use :Inspect, :InspectTree, and :EditQuery

; Highlight ```math fenced code blocks using the LaTeX highlighter.
(fenced_code_block
  (info_string (language) @_lang)
  (code_fence_content) @injection.content
  (#eq? @_lang "math")
  (#set! injection.language "latex"))
```

But I noticed that we use `#set-lang-from-info-string!` with some built-in aliases: https://github.com/nvim-treesitter/nvim-treesitter/blob/bfe74a4899882a4ef45abb80813f14644a110a34/queries/markdown/injections.scm#L5

### DIFF
--- a/lua/nvim-treesitter/query_predicates.lua
+++ b/lua/nvim-treesitter/query_predicates.lua
@@ -13,6 +13,7 @@ local non_filetype_match_injection_language_aliases = {
   sh = "bash",
   uxn = "uxntal",
   ts = "typescript",
+  math = "latex",
 }
 
 -- compatibility shim for breaking change on nightly/0.11


clason:
Sorry, these directives are deprecated. And the parser (and thus by extension the queries) follows CommonMark. 

It's perfectly fine to have this in your personal config, of course!

jyc:
@clason No worries, that makes sense!

Do you have any advice for improving the performance of the injection? The reason I started investigating this is I noticed that <code>```latex</code> blocks seemed to have much less of an impact on typing latency than the <code>```math</code> blocks after my injection.

My suspicion was that nvim-treesitter was more efficient at dealing with the built-in `set-lang-from-info-string!` query than my `eq? @_lang "math"` logic; do you think that's the case?

clason:
Well, the latex injection is not using that codepath but standard language lookup. But I am surprised it's measurable, much less noticeable.

clason:
You can try registering "math" as a filetype for latex, :h vim.treesitter.language.register.

jyc:
Thanks, that was very useful for testing!

Interesting, unfortunately that seems to be just as laggy as my injection, so you're right that that isn't the origin of the performance difference. That is a bummer. I will keep an eye on [Treesitter performance (tracking issue) · Issue #22426 · neovim/neovim](https://github.com/neovim/neovim/issues/22426).

Thanks for the help!

PR #7130, opened 2024-09-09
feat(zine): add zine-ssg parsers and queries

rockorager:
Add parsers and queries for filetypes used by the
[Zine](https://github.com/kristoff-it/zine) static site generator.

- [Ziggy](https://github.com/kristoff-it/ziggy)
- [SuperMD](https://github.com/kristoff-it/supermd)
- [SuperHTML](https://github.com/kristoff-it/superhtml)

The initial state of the queries are copies of those provided in the Zine repo.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2101,6 +2101,18 @@ list.supercollider = {
   maintainers = { "@madskjeldgaard" },
 }
 
+list.superhtml = {
+  install_info = {
+    url = "https://github.com/kristoff-it/superhtml",
+    files = {
+      "src/parser.c",
+      "src/scanner.c",
+    },
+    location = "tree-sitter-superhtml",
+  },
+  maintainers = { "@rockorager" },
+}
+
 list.supermd = {
   install_info = {
     url = "https://github.com/kristoff-it/supermd",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2101,6 +2101,30 @@ list.supercollider = {
   maintainers = { "@madskjeldgaard" },
 }
 
+list.supermd = {
+  install_info = {
+    url = "https://github.com/kristoff-it/supermd",
+    files = {
+      "src/parser.c",
+      "src/scanner.c",
+    },
+    location = "tree-sitter/supermd",
+  },
+  maintainers = { "@rockorager" },
+}
+
+list.supermd_inline = {
+  install_info = {
+    url = "https://github.com/kristoff-it/supermd",
+    files = {
+      "src/parser.c",
+      "src/scanner.c",
+    },
+    location = "tree-sitter/supermd-inline",
+  },
+  maintainers = { "@rockorager" },
+}
+
 list.surface = {
   install_info = {
     url = "https://github.com/connorlay/tree-sitter-surface",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2561,6 +2561,15 @@ list.ziggy = {
   maintainers = { "@rockorager" },
 }
 
+list.ziggy_schema = {
+  install_info = {
+    url = "https://github.com/kristoff-it/ziggy",
+    files = { "src/parser.c" },
+    location = "tree-sitter-ziggy-schema",
+  },
+  maintainers = { "@rockorager" },
+}
+
 list.templ = {
   install_info = {
     url = "https://github.com/vrischmann/tree-sitter-templ",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2552,6 +2552,15 @@ list.zig = {
   maintainers = { "@amaanq" },
 }
 
+list.ziggy = {
+  install_info = {
+    url = "https://github.com/kristoff-it/ziggy",
+    files = { "src/parser.c" },
+    location = "tree-sitter-ziggy",
+  },
+  maintainers = { "@rockorager" },
+}
+
 list.templ = {
   install_info = {
     url = "https://github.com/vrischmann/tree-sitter-templ",
--- a/queries/superhtml/highlights.scm
+++ b/queries/superhtml/highlights.scm
@@ -1,0 +1,47 @@
+(doctype) @constant
+
+(comment) @comment
+
+((tag_name) @string.special
+  (#any-of? @string.special "super" "extend"))
+
+(tag_name) @tag
+
+((element
+  (start_tag
+    (attribute
+      (attribute_name) @attribute
+      [
+        (attribute_value) @markup.link.url
+        (quoted_attribute_value
+          (attribute_value) @markup.link.url)
+      ]))
+  (element
+    (start_tag
+      (tag_name) @tag)))
+  (#eq? @tag "super")
+  (#eq? @attribute "id"))
+
+((element
+  (start_tag
+    (tag_name) @string.special
+    (attribute
+      (attribute_name) @attribute)+))
+  (#eq? @string.special "super"))
+
+(attribute_name) @attribute
+
+[
+  "\""
+  (attribute_value)
+] @string
+
+[
+  "<"
+  ">"
+  "</"
+  "/>"
+  "<!"
+] @punctuation.bracket
+
+"=" @punctuation.delimiter
--- a/queries/supermd/highlights.scm
+++ b/queries/supermd/highlights.scm
@@ -1,0 +1,44 @@
+(atx_heading
+  (inline) @markup.heading)
+
+(setext_heading
+  (paragraph) @markup.heading)
+
+[
+  (atx_h1_marker)
+  (atx_h2_marker)
+  (atx_h3_marker)
+  (atx_h4_marker)
+  (atx_h5_marker)
+  (atx_h6_marker)
+  (setext_h1_underline)
+  (setext_h2_underline)
+] @punctuation.special
+
+[
+  (link_title)
+  (indented_code_block)
+  (fenced_code_block)
+] @markup.raw
+
+(fenced_code_block_delimiter) @punctuation.delimiter
+
+(link_destination) @markup.link.url
+
+(link_label) @markup.link.label
+
+[
+  (list_marker_plus)
+  (list_marker_minus)
+  (list_marker_star)
+  (list_marker_dot)
+  (list_marker_parenthesis)
+  (thematic_break)
+] @punctuation.special
+
+[
+  (block_continuation)
+  (block_quote_marker)
+] @punctuation.special
+
+(backslash_escape) @string.escape
--- a/queries/ziggy/highlights.scm
+++ b/queries/ziggy/highlights.scm
@@ -1,0 +1,41 @@
+[
+  (true)
+  (false)
+] @constant.builtin
+
+(null) @constant.builtin
+
+[
+  (integer)
+  (float)
+] @number
+
+(struct_field
+  key: (_) @keyword)
+
+(struct
+  name: (_) @type)
+
+(tag) @function
+
+[
+  (string)
+  (line_string)*
+] @string
+
+(comment) @comment
+
+(escape_sequence) @string.escape
+
+"," @punctuation.delimiter
+
+[
+  "["
+  "]"
+  "{"
+  "}"
+  "("
+  ")"
+] @punctuation.bracket
+
+(top_comment) @comment
--- a/lockfile.json
+++ b/lockfile.json
@@ -886,5 +886,8 @@
   },
   "zig": {
     "revision": "21e2218e0ec7f4e3c0640d16bf8c67e6f0a61e18"
+  },
+  "ziggy": {
+    "revision": "c66f47bc632c66668d61fa06eda112b41d6e5130"
   }
 }
--- a/lockfile.json
+++ b/lockfile.json
@@ -725,6 +725,9 @@
   "supercollider": {
     "revision": "affa4389186b6939d89673e1e9d2b28364f5ca6f"
   },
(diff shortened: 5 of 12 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -2101,6 +2102,48 @@ list.supercollider = {
...
+
+list.supermd = {
+  install_info = {
+    url = "https://github.com/kristoff-it/supermd",
+    includes = { "tree-sitter/supermd/src" },
+    files = {
+      "tree-sitter/supermd/src/parser.c",
+      "tree-sitter/supermd/src/scanner.c",
+    },
+    branch = "main",
```
clason, line 2127:
```suggestion
```
You only need to specify branch names that are not the default.

lua/nvim-treesitter/parsers.lua
```diff
@@ -2101,6 +2102,48 @@ list.supercollider = {
...
+  },
+  filetype = "superhtml",
+  maintainers = { "@rockorager" },
+}
+
+list.supermd = {
+  install_info = {
+    url = "https://github.com/kristoff-it/supermd",
+    includes = { "tree-sitter/supermd/src" },
```
clason, line 2122:
```suggestion
```
That doesn't work. Parsers must be self-contained.

lua/nvim-treesitter/parsers.lua
```diff
@@ -2101,6 +2102,48 @@ list.supercollider = {
...
+    files = {
+      "tree-sitter/supermd/src/parser.c",
```
clason, line 2124:
```suggestion
      "src/parser.c",
```
Use the `location` key instead to specify subdirectories.

lua/nvim-treesitter/parsers.lua
```diff
@@ -35,6 +35,7 @@ for ft, lang in pairs {
   typ = "typst",
   pl = "perl",
   uxn = "uxntal",
+  ["ziggy-schema"] = "ziggy_schema",
```
clason, line 38:
```suggestion
```
Duplicate of the `filetype` key below (which is the correct approach here).

lua/nvim-treesitter/parsers.lua
```diff
@@ -2101,6 +2102,48 @@ list.supercollider = {
...
+  install_info = {
+    url = "https://github.com/kristoff-it/supermd",
+    includes = { "tree-sitter/supermd-inline/src" },
+    files = {
+      "tree-sitter/supermd-inline/src/parser.c",
+      "tree-sitter/supermd-inline/src/scanner.c",
+    },
+    branch = "main",
+  },
+  filetype = "supermd_inline",
```
clason, line 2143: Is this in fact a separate filetype, or a second-stage parser like `markdown-inline` (in which case it should be omitted)?

clason (commented):
You also need to add corresponding `lockfile.json` entries, otherwise CI won't run.

clason:
> The initial state of the queries are copies of those provided in the Zine repo.

That's probably not going to work; please make sure you use the captures from CONTRIBUTING.md (and only those).

queries/ziggy/highlights.scm
```diff
@@ -0,0 +1,47 @@
...
+  (string) 
+  (line_string)*
+] @string
+
+
+(comment) @comment.line
+
+(escape_sequence) @constant.character.escape
+
+(ERROR) @error
```
clason, line 32:
```suggestion
```
We don't capture `(ERROR)` nodes; they're only useful for debugging and will lead to christmas trees when editing.

clason (commented):
Please run the `./scripts/format-queries.lua queries` script to appease the linter.

lua/nvim-treesitter/parsers.lua
```diff
@@ -2101,6 +2102,48 @@ list.supercollider = {
...
+      "tree-sitter-superhtml/src/parser.c",
+      "tree-sitter-superhtml/src/scanner.c",
+    },
+    branch = "main",
+  },
+  filetype = "superhtml",
+  maintainers = { "@rockorager" },
+}
+
+list.supermd = {
```
lucario387, line 2119: The metadata of this repo's package.json is still that of `tree-sitter-markdown`. You should change it.

Addtionall question: The grammar code seems to have a lot of duplicates with Markdown parser's one. You can probably extend from it, some examples can be:
- https://github.com/tree-sitter-grammars/tree-sitter-vue/blob/main/grammar.js
- https://github.com/tree-sitter/tree-sitter-cpp/blob/master/grammar.js
- https://github.com/tree-sitter/tree-sitter-typescript/blob/master/common/define-grammar.js

lua/nvim-treesitter/parsers.lua
rockorager: > You only need to specify branch names that are not the default.

I had originally tried this but get this error when installing:

```
nvim-treesitter[ziggy]: Failed to execute the following command:
{
  cmd = "mv",
  opts = {
    args = { "-f", "tree-sitter-ziggy-tmp/ziggy-master", "tree-sitter-ziggy" },
    cwd = "/home/tim/.local/share/nvim",
    stdio = {
      [2] = <userdata 1>,
      [3] = <userdata 2>
    }
  }
}
mv: cannot stat 'tree-sitter-ziggy-tmp/ziggy-master': No such file or directory
```

Specifying the branch resolves this. Any ideas?

lua/nvim-treesitter/parsers.lua
rockorager: It's a second stage parser. Do I specify the grammar in the `supermd` table then?

lua/nvim-treesitter/parsers.lua
clason: I can also recommend https://github.com/tree-sitter-grammars/template for a repo layout and workflows. It's very important to have grammars be as consistent as possible for editors to be able to include them.

lua/nvim-treesitter/parsers.lua
clason: No, just omit the "fake" `filetype` and let injections handle it (same as for markdown; check their entries and queries).

lua/nvim-treesitter/parsers.lua
clason: As I wrote above, you need to add lockfile entries. `branch` in this case means "just use HEAD YOLO", which clearly is not ideal ;)

lua/nvim-treesitter/parsers.lua
clason: (The key also does nothing since it's not used.)

clason:
To be honest, it would have been easier to open separate PRs for the different (unrelated) parsers -- and to start with one and see how it works, then tackle the others in turn. But what's done is done.

rockorager:
> To be honest, it would have been easier to open separate PRs for the different (unrelated) parsers -- and to start with one and see how it works, then tackle the others in turn. But what's done is done.

I think this is still a good idea...Will be easier to get it right for the rest once I see how this is done for a single one. Sound ok if I close this and open a new one with a single parser and queries?

clason:
It's not necessary, but if you prefer, just drop the commits (`git rebase -i HEAD~5`) and force-push. I'd make sure to incorporate the current feedback (locally) first.

queries/superhtml/highlights.scm
```diff
@@ -0,0 +1,49 @@
+(erroneous_end_tag_name) @tag.error
```
clason, line 1: We only highlight _correct_ code; we explicitly do not do any linting.

(Consider if it makes sense to `; inherit` the  `html` or `html_tags` queries instead of duplicating them.)

queries/supermd/highlights.scm
clason, line 1: Same here; these look very similar to Markdown.

queries/supermd/highlights.scm
```diff
@@ -0,0 +1,46 @@
...
+
+[
+  (link_title)
+  (indented_code_block)
+  (fenced_code_block)
+] @text.literal
+
+(fenced_code_block_delimiter) @punctuation.delimiter
+
+(code_fence_content) @none
```
clason, line 26: That doesn't do anything in Neovim...

lua/nvim-treesitter/parsers.lua
rockorager: I'll see if upstream will consider extending the existing markdown grammar. Is this a dealbreaker for inclusion?

clason: Not really; we care more about the queries (and that `tree-sitter generate` and `tree-sitter build` work out of the box). But it's a good suggestion, and _if_ it is followed may lead to possibly breaking changes in the queries -- and it would be good to avoid that (i.e., wait until the "final" queries can be added here).

queries/supermd/highlights.scm
rockorager: > Same here

Referring to "this doesn't do anything in neovim"?

lua/nvim-treesitter/parsers.lua
clason: (But another indication why separate PRs for separate languages is a good idea ;))

queries/supermd/highlights.scm
clason: Referring to "consider inheriting them from `markdown` resp. `markdown_inline`".

lua/nvim-treesitter/parsers.lua
rockorager: > The metadata of this repo's package.json is still that of `tree-sitter-markdown`. You should change it.

https://github.com/kristoff-it/supermd/pull/1

queries/supermd/highlights.scm
rockorager: One can inherit queries? Do you have an example I can look at?

rockorager:
I believe all the major issues here are resolved. The only remaining previously mentioned is for upstream to consider extending the existing markdown parser (ref: https://github.com/kristoff-it/supermd/issues/2)

I can autosquash all those fixups if you prefer. There are a bunch, so I won't be suprised if you have more comments :)

Thanks!

lua/nvim-treesitter/parsers.lua
```diff
@@ -2101,6 +2101,43 @@ list.supercollider = {
...
 
+list.superhtml = {
+  install_info = {
+    url = "https://github.com/kristoff-it/superhtml",
+    files = {
+      "tree-sitter-superhtml/src/parser.c",
+      "tree-sitter-superhtml/src/scanner.c",
+    },
+  },
+  filetype = "superhtml",
```
clason, line 2112:
```suggestion
```

This is only needed if the filetype is different from the language name.

lua/nvim-treesitter/parsers.lua
```diff
@@ -2101,6 +2101,43 @@ list.supercollider = {
   maintainers = { "@madskjeldgaard" },
 }
 
+list.superhtml = {
+  install_info = {
+    url = "https://github.com/kristoff-it/superhtml",
+    files = {
+      "tree-sitter-superhtml/src/parser.c",
```
clason, line 2108: Please use the `location` key instead; this is not the correct way to specify non-standard locations (can lead to race conditions with parallel install, for one thing).

lua/nvim-treesitter/parsers.lua
```diff
@@ -2552,6 +2589,24 @@ list.zig = {
   maintainers = { "@amaanq" },
 }
 
+list.ziggy = {
+  install_info = {
+    url = "https://github.com/kristoff-it/ziggy",
+    files = { "tree-sitter-ziggy/src/parser.c" },
```
clason, line 2595: same (use `location`)

lua/nvim-treesitter/parsers.lua
```diff
@@ -2552,6 +2589,24 @@ list.zig = {
...
+  },
+  filetype = "ziggy",
```
clason, line 2597:
```suggestion
```

lua/nvim-treesitter/parsers.lua
```diff
@@ -2552,6 +2589,24 @@ list.zig = {
...
+  maintainers = { "@rockorager" },
+}
+
+list.ziggy_schema = {
+  install_info = {
+    url = "https://github.com/kristoff-it/ziggy",
+    files = { "tree-sitter-ziggy-schema/src/parser.c" },
+  },
+  filetype = "ziggy-schema",
```
clason, line 2606:
```suggestion
```
The filetype is actually called `ziggy_schema`.

rockorager:
CI failure looks unrelated.

> Error: parser for uxntal is not installed

queries/supermd/highlights.scm
clason: Yes, `cpp` and `c` (and `cuda` etc.)

kristoff-it:
Thank you all for this work, just commenting on the point about extending the original TS grammar: I'm currently test-driving semantic changes to markdown in SuperMD and in the meantime not breaking compatibility with it, but in the very near future I plan to deviate from markdown more significantly also parser-wise.

This initial copy-pasted version of the grammar is just so that users can have a good time using the file format (in the context of [Zine](https://zine-ssg.io)) and help me gather feedback.

clason:
Hmm... We do expect parser to be relatively stable before including them (these days). Breaking changes are a massive pain for users _and_ maintainers.

clason:
@rockorager Given the comment about SuperMD, I would indeed like to ask you to break this up into separate PRs for SuperHTML, SuperMD (when it's done), and Ziggy. This way we can at least merge two of the PRs now.

clason:
@rockorager friendly ping

rockorager:
@clason Do you want two (new) separate PRs, or would it be ok to remove SuperMD from this one?

clason:
If it's not too much work, three separate PRs (of which this can be one) for each family (superhtml, supermd, and ziggy) would be preferable.

rockorager:
#7197 and #7198

PR #7131, opened 2024-09-09
feat(latex): capture conditionals

jdujava:
Before:
image
After:
image

Can possibly capture commands which are not conditionals in the purest sense shown above, but generally it works pretty well. In case someone has macro/command starting on `if` which doesn't have anything with conditionals, and is bothered by such false positive capture, they are free to overwrite it with another tailor-made capture.


clason:
I'm always a bit hesitant to add pattern matching, as that has a serious effect on performance. But I guess people are used to LaTeX syntax highlighting being more leisurely...

PR #7139, opened 2024-09-11
feat(ruby): improve require highlighting

jeromedalbert:
1. Highlight import (like `require`) when they are not top level.

    Before:

    before

    After:

    after

2. Add `autoload` and `gem` to the list of imports


clason:
@TravonteD

PR #7140, opened 2024-09-11
feat(ruby): detect shebang directive

jeromedalbert:
Detect the shebang directive similarly to what Bash [does](https://github.com/nvim-treesitter/nvim-treesitter/blob/b6a6d8997c46dc15682020ce4fddc5a89ee1ac0d/queries/bash/highlights.scm#L236-L239).

### DIFF
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -208,6 +208,11 @@
 
 (comment) @comment @spell
 
+((program
+  .
+  (comment) @keyword.directive @nospell)
+  (#lua-match? @keyword.directive "^#!/"))
+
 (program
   (comment)+ @comment.documentation
   (class))


clason (changes requested):

clason:
@TravonteD

ribru17:
Why not only apply highlights if the comment *is* a shebang? Seems like the shebang is still a comment (as well as a message to the kernel)

jeromedalbert:
I was trying to highlight comments and shebangs differently. Since then I figured it was mostly pointless and would go against how other languages like Bash highlight shebangs.

I updated the PR by copy/pasting the Bash [shebang highlights](https://github.com/nvim-treesitter/nvim-treesitter/blob/b6a6d8997c46dc15682020ce4fddc5a89ee1ac0d/queries/bash/highlights.scm#L236-L239) and it works well.

clason (approved):

PR #7141, opened 2024-09-11
feat(ruby): detect more keywords

jeromedalbert:
Detect more keywords. Inspired from the Vim [syntax file for Ruby](https://github.com/vim-ruby/vim-ruby/blob/master/syntax/ruby.vim).

- `BEGIN` and `END` ([doc](https://ruby-doc.org/3.3.5/syntax/miscellaneous_rdoc.html#label-BEGIN+and+END))
- `proc`, `lambda`, etc
- `alias_method`, `define_method`, etc
- `abort`, `exit`, `loop`, etc


queries/ruby/highlights.scm
```diff
@@ -38,6 +46,10 @@
 (method
   "end" @keyword.function)
 
+(call
+  method: (identifier) @keyword.function
+  (#match? @keyword.function "(alias|define|define_singleton|remove|undef)_method"))
```
clason, line 51: Sorry. that's inacceptable. Use `#any-of?` here.

clason (changes requested):

queries/ruby/highlights.scm
```diff
@@ -16,6 +16,14 @@
   "then"
 ] @keyword
 
+(begin_block) @keyword
+
+(end_block) @keyword
```
ribru17, line 21:
```suggestion
[ 
  "BEGIN"
  "END"
] @keyword
```
Otherwise the entire block will be given the keyword highlight, rather than just the keywords themselves

queries/ruby/highlights.scm
```diff
@@ -16,6 +16,14 @@
...
+
+(call
+  method: (identifier) @keyword
+  (#any-of? @keyword "proc" "lambda" "caller" "callcc"))
```
ribru17, line 25: These queries get applied before the `@function.call` queries, and are thus overridden. They should work properly if you move them below those queries. Also it seems like maybe the more appropriate capture group would be `@function.builtin`?

queries/ruby/highlights.scm
```diff
@@ -71,6 +84,10 @@
   "ensure"
 ] @keyword.exception
 
+(call
+  method: (identifier) @keyword
+  (#any-of? @keyword "abort" "at_exit" "exit" "fork" "loop" "trap"))
```
ribru17, line 89: These also need to be moved lower, since they are being overridden, and also probably `@function.builtin`

queries/ruby/highlights.scm
```diff
@@ -38,6 +46,11 @@
 (method
   "end" @keyword.function)
 
+(call
+  method: (identifier) @keyword.function
+  (#any-of? @keyword.function
+    "alias_method" "define_method" "define_singleton_method" "remove_method" "undef_method"))
```
ribru17, line 52: Also need to be moved below- and seems like these should also be `@function.builtin`?

PR #7143, opened 2024-09-11
feat(markdown): Conceal backslashes in escape sequences

cpmsmith:
| | `conceallevel=0` | `conceallevel=3` |
| ---:| --- | --- |
| **Before** | Screenshot 2024-09-11 at 13 18 17 | Screenshot 2024-09-11 at 13 18 30 |
| **After** | Screenshot 2024-09-11 at 13 16 55 | Screenshot 2024-09-11 at 13 17 17 |

## Things to note
- For the conceal and offset to work, I needed to query for `(_ (backslash_escape))`; just querying for `backslash_escape` didn't work. I'm not sure why, but I'm guessing it might be something to do with whatever technical constraints require `markdown` and `markdown_inline` to be separate.
- I've made the slightly unorthodox move of not highlighting the escaped characters themselves, only the backslashes. The idea here is to use concealing as a simple "preview" mode, and in the final output, these escaped characters would appear as normal characters, so highlighting them felt a bit off to me. 
  - I also added `@punctuation.delimiter` to unescaped delimiters, which ends up with this easy distinction between unescaped and escaped delimiters:
    image

### DIFF
--- a/queries/markdown_inline/highlights.scm
+++ b/queries/markdown_inline/highlights.scm
@@ -10,16 +10,19 @@
 (shortcut_link
   (link_text) @nospell)
 
-[
-  (backslash_escape)
-  (hard_line_break)
-] @string.escape
+(hard_line_break) @string.escape
+
+; Conceal backslashes in escape sequences
+(_
+  (backslash_escape) @string.escape
+  (#offset! @string.escape 0 0 0 -1)
+  (#set! conceal ""))
 
 ; Conceal codeblock and text style markers
 ([
   (code_span_delimiter)
   (emphasis_delimiter)
-] @conceal
+] @punctuation.delimiter
   (#set! conceal ""))
 
 ; Conceal inline links


ribru17:
I think this was already discussed [here](https://github.com/nvim-treesitter/nvim-treesitter/pull/6742), seems like we want to wait for upstream to drive this (since conceals are hard to configure)

cpmsmith:
Ha, I should've looked for closed PRs. Going off @clason's comment on the previous PR:

> The issue is that the markdown parser is used by Neovim as a renderer for LSP documentation, and it seems to me that such strings are something you would want to see literally

I guess "LSP documentation" is referring to hovers, and I actually originally looked into this _because_ of the current behaviour with hovers, though I imagine there are a lot of variables. This is what I currently see when hovering `more_itertools.constrained_batches` using Pyright:

image

The source of this docstring is [here](https://github.com/more-itertools/more-itertools/blob/0cd42cf61132dae8ea363b7fda2f62d0164a1637/more_itertools/more.py#L4680-L4699), written in reStructuredText, and being converted into Markdown by (I assume) Pyright. Notably, the backslash in e.g. "max\_count" was added for the sake of Markdown presentation, and is not present in the original. I don't know why the asterisks on the emphasized ranges are also being escaped, considering the inline code ranges are being correctly converted.

All in all, Pyright is misbehaving a little bit here by escaping the asterisks, but given this output from the language server I expected the backslashes to be concealed.

When talking about letting upstream drive this, what is upstream in this case? Neovim, the language server, the protocol?

clason:
> All in all, Pyright is misbehaving a little bit here by escaping the asterisks, but given this output from the language server I expected the backslashes to be concealed.

A similar issue is in clangd (overzealous escaping of characters); _these_ should be fixed rather than papered over with conceals. (Because otherwise you might be removing _wanted_ escape characters -- if the documentation is talking about regular expressions, for example, and that is unacceptable. Conceals should be used _very_ sparingly and only in situations where you know they are always desirable.)

PR #7153, opened 2024-09-16
feat(highlights): capture wildcard patterns as `@character.special`

ValdezFOmar:
Some languages use `_` as a wildcard pattern to match anything, which should be highlighted as `@character.special`. Is possible that other languages use a different character for the same purpose, but I don't know of any.

I got a list of grammars that use an underscore with this command:

```sh
rg '"_"|underscore' queries/
```
and just checked which ones use it for pattern matching, with the following results:

- [x] `elm`
- [x] `func`
- [x] `gdscript`
- [x] `haskell`
- [x] `menhir`
- [x] `ocaml`
- [x] `python`
- [x] `query`
- [ ] `racket`: It technically uses `_` for pattern matching but this is
  just a regular symbol in the grammar, so it would need a special and
  complicated query for highlighting.
- [x] `rust`
- [x] `scala`

Grammars that use `_` for other purposes:

- `blueprint`
- `cue`
- `elvish`
- `fish`
- `hare`
- `lalrpop`
- `latex`
- `perl`
- `uxntal`
- `wit`
- `zig`

Please tell me if I miss anything.

### DIFF
--- a/queries/c_sharp/highlights.scm
+++ b/queries/c_sharp/highlights.scm
@@ -217,8 +217,9 @@
   "module"
   "this"
   "base"
-  (discard)
 ] @variable.builtin
+
+(discard) @character.special
 
 (constructor_declaration
   name: (identifier) @constructor)
--- a/queries/c_sharp/highlights.scm
+++ b/queries/c_sharp/highlights.scm
@@ -219,8 +219,6 @@
   "base"
 ] @variable.builtin
 
-(discard) @character.special
-
 (constructor_declaration
   name: (identifier) @constructor)
 
@@ -430,6 +428,11 @@
   "??="
   ".."
 ] @operator
+
+(list_pattern
+  ".." @character.special)
+
+(discard) @character.special
 
 [
   ";"
--- a/queries/elm/highlights.scm
+++ b/queries/elm/highlights.scm
@@ -67,7 +67,7 @@
   (lower_case_identifier) @variable)
 
 (anything_pattern
-  (underscore) @variable)
+  (underscore) @character.special)
 
 (record_base_identifier
   (lower_case_identifier) @variable)
@@ -88,7 +88,7 @@
 
 (function_declaration_left
   (anything_pattern
-    (underscore) @variable.parameter))
+    (underscore) @character.special))
 
 (function_declaration_left
   (lower_pattern
--- a/queries/gdscript/highlights.scm
+++ b/queries/gdscript/highlights.scm
@@ -129,7 +129,7 @@
   (#eq? @keyword.operator "new"))
 
 ; Match Pattern
-(underscore) @constant ; The "_" pattern.
+(underscore) @character.special ; The "_" pattern.
 
 (pattern_open_ending) @operator ; The ".." pattern.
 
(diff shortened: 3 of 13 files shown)


lucario387:
There is also ocaml and haskell underscores, doesn't seem like there is a query for that in here in the first place. Can you consider adding them as well
for ocaml it's something like

```ocaml
let foo = fun
  | 1 -> "bar"
  | _ -> "Whatever" (** wildcard **)
```

ValdezFOmar:
I tried this example for OCaml:

```ocaml
match foo with
  | [x, _] -> "bar"
  | _ -> "wildcard"
```

which produces this tree:

```scm
(compilation_unit ; [0, 0] - [3, 0]
  (expression_item ; [0, 0] - [2, 19]
    (match_expression ; [0, 0] - [2, 19]
      "match" ; [0, 0] - [0, 5]
      (value_path ; [0, 6] - [0, 9]
        (value_name)) ; [0, 6] - [0, 9]
      "with" ; [0, 10] - [0, 14]
      "|" ; [1, 2] - [1, 3]
      (match_case ; [1, 4] - [1, 19]
        pattern: (list_pattern ; [1, 4] - [1, 10]
          "[" ; [1, 4] - [1, 5]
          (tuple_pattern ; [1, 5] - [1, 9]
            (value_pattern) ; [1, 5] - [1, 6]
            "," ; [1, 6] - [1, 7]
            (value_pattern)) ; [1, 8] - [1, 9]
          "]") ; [1, 9] - [1, 10]
        "->" ; [1, 11] - [1, 13]
        body: (string ; [1, 14] - [1, 19]
          "\"" ; [1, 14] - [1, 15]
          (string_content) ; [1, 15] - [1, 18]
          "\"")) ; [1, 18] - [1, 19]
      "|" ; [2, 2] - [2, 3]
      (match_case ; [2, 4] - [2, 19]
        pattern: (value_pattern) ; [2, 4] - [2, 5]
        "->" ; [2, 6] - [2, 8]
        body: (string ; [2, 9] - [2, 19]
          "\"" ; [2, 9] - [2, 10]
          (string_content) ; [2, 10] - [2, 18]
          "\""))))) ; [2, 18] - [2, 19]
```

And it seems that currently the grammar doesn't expose `"_"` as a different node (which would explain why there wasn't a capture in the first place), so this needs to be handled in the parser first before we can highlight it. I'll open a PR in tree-sitter-ocaml repo for this.

ValdezFOmar:
@lucario387 I added highlights for haskell and ocaml.

I also notice that haskell uses `..` as a wildcard in pattern as well as rust, so I added those highlights too. I'll look if other languages also use `..` as a pattern.

ValdezFOmar:
I covered all that I could find, let me know if I'm still missing something.

ObserverOfTime (commented):
- [Java](https://docs.oracle.com/en/java/javase/22/language/unnamed-variables-and-patterns.html)
- [C#](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/patterns#discard-pattern)

ValdezFOmar:
@ObserverOfTime done

ObserverOfTime (approved):

lucario387 (approved):

PR #7158, opened 2024-09-18
feat(swift): fold queries

ribru17:


queries/swift/folds.scm
```diff
@@ -0,0 +1,13 @@
...
+  (function_body)
+  (do_statement)
+  (catch_block)
+  (if_statement)
+  (for_statement)
+  (property_declaration)
+  (switch_statement)
+  (switch_entry)
+  (import_declaration)+
+] @fold
```
lucario387, line 13: There is a lot more than just this, I'd assume.
Even the list in `swift/indents.scm` doesn't fully cover all foldable blocks

ribru17: Nice, thanks I have added most from there now. Some were duplicates so I removed a couple

PR #7161, opened 2024-09-18
highlight swift str_escaped_char as @string.escape

John-Colvin:


clason:
Thank you, but that's already included in https://github.com/nvim-treesitter/nvim-treesitter/pull/7157!

PR #7166, opened 2024-09-19
feat(fsharp): add parser and queries

Nsidorenco:
This adds support for the [fsharp parser](https://github.com/ionide/tree-sitter-fsharp) along with highlighting and injection queries. 

I have not been able to make satisfactory indentation queries so they are left out for now.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -617,6 +617,16 @@ list.fsh = {
   maintainers = { "@mgramigna" },
 }
 
+list.fsharp = {
+  install_info = {
+    url = "https://github.com/ionide/tree-sitter-fsharp",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
+    location = "fsharp",
+  },
+  maintainers = { "@nsidorenco" },
+}
+
 list.func = {
   install_info = {
     url = "https://github.com/amaanq/tree-sitter-func",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -621,7 +621,6 @@ list.fsharp = {
   install_info = {
     url = "https://github.com/ionide/tree-sitter-fsharp",
     files = { "src/parser.c", "src/scanner.c" },
-    branch = "main",
     location = "fsharp",
   },
   maintainers = { "@nsidorenco" },
--- a/queries/fsharp/highlights.scm
+++ b/queries/fsharp/highlights.scm
@@ -1,0 +1,372 @@
+[
+  (line_comment)
+  (block_comment)
+] @comment @spell
+
+((line_comment) @comment.documentation @spell
+  (#match? @comment.documentation "^///"))
+
+(const
+  [
+    (_) @constant
+    (unit) @constant.builtin
+  ])
+
+(primary_constr_args
+  (_) @variable.parameter)
+
+(class_as_reference
+  (_) @variable.parameter.builtin)
+
+((argument_patterns
+  (long_identifier
+    (identifier) @character.special))
+  (#match? @character.special "^\_.*"))
+
+(wildcard_pattern) @character.special
+
+(type_name
+  type_name: (_) @type.definition)
+
+[
+  (_type)
+  (atomic_type)
+] @type
+
+(member_signature
+  .
+  (identifier) @function.method
+  (curried_spec
+    (arguments_spec
+      "*"* @operator
+      (argument_spec
+        (argument_name_spec
+          "?"? @character.special
+          name: (_) @variable.parameter)))))
+
+(union_type_case) @constant
+
+(rules
+  (rule
+    pattern: (_) @constant
+    block: (_)))
+
+(identifier_pattern
+  .
+  (_) @constant
+  .
+  (_) @variable)
+
+(optional_pattern
+  "?" @character.special)
+
+(fsi_directive_decl
+  .
+  (string) @module)
+
+(import_decl
+  .
+  (_) @module)
+
+(named_module
+  name: (_) @module)
+
+(namespace
+  name: (_) @module)
+
+(module_defn
+  .
+  (_) @module)
+
+(ce_expression
+  .
+  (_) @constant.macro)
+
+(field_initializer
+  field: (_) @property)
+
+(record_fields
+  (record_field
+    .
+    (identifier) @property))
+
+(dot_expression
+  base: (_)? @module)
+
+(value_declaration_left
+  .
+  (_) @variable)
+
+(function_declaration_left
+  .
+  (_) @function
+  .
+  (_)* @variable.parameter)
+
+(member_defn
+  (method_or_prop_defn
+    [
+      (property_or_ident) @function
+      (property_or_ident
+        instance: (identifier) @variable.parameter.builtin
+        method: (identifier) @function.method)
+    ]
+    args: (_)* @variable.parameter))
+
+(application_expression
+  .
+  (_) @function.call
+  .
+  (_) @variable)
+
+((infix_expression
+  .
+  (_)
+  .
+  (infix_op) @operator
+  .
+  (_) @function.call)
+  (#eq? @operator "|>"))
+
+((infix_expression
+  .
+  (_) @function.call
+  .
+  (infix_op) @operator
+  .
+  (_))
+  (#eq? @operator "<|"))
+
+[
+  (xint)
+  (int)
+  (int16)
+  (uint16)
+  (int32)
+  (uint32)
+  (int64)
+  (uint64)
+  (nativeint)
+  (unativeint)
+] @number
+
+[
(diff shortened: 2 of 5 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -617,6 +617,16 @@ list.fsh = {
   maintainers = { "@mgramigna" },
 }
 
+list.fsharp = {
+  install_info = {
+    url = "https://github.com/ionide/tree-sitter-fsharp",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
```
clason, line 624:
```suggestion
```
Not needed if that's the default branch (but you do need to add a lockfile.json entry).

lua/nvim-treesitter/parsers.lua
```diff
@@ -617,6 +617,16 @@ list.fsh = {
...
+    location = "fsharp",
```
clason, line 625: That is a very weird layout.

queries/fsharp/highlights.scm
```diff
@@ -0,0 +1,372 @@
+[
+  (line_comment)
+  (block_comment)
+] @comment @spell
+
+((line_comment) @comment.documentation @spell
+  (#match? @comment.documentation "^///"))
```
clason, line 7: Please use lua-match and Lua patterns; regexp are unacceptably expensive.

queries/fsharp/injections.scm
```diff
@@ -0,0 +1,11 @@
...
+  (line_comment)
+  (block_comment_content)
+] @injection.content
+  (#set! injection.language "comment"))
+
+((line_comment) @injection.content
+  (#match? @injection.content "^///")
+  (#offset! @injection.content 0 3 0 0)
+  (#set! injection.language "xml")
+  (#set! injection.combined))
```
clason, line 11: Is that required? (It will parse all injections as one single document.)

clason (changes requested):

queries/fsharp/injections.scm
Nsidorenco: I think it is. A doctoring in fsharp looks like the following:
```fsharp
/// <summary>
/// text goes here
/// </summary>
let x = 42
```

So I need to combine all adjacent comment nodes to parse the embedded xml

lua/nvim-treesitter/parsers.lua
Nsidorenco: The repo also contains a parser for fsharp signature files. I just copied the layout of the typescript parser here tbh.

clason: Is the signature parser relevant here (as injections)?

queries/fsharp/injections.scm
clason: I think that can be handled with quantified queries @ribru17

lua/nvim-treesitter/parsers.lua
Nsidorenco: Eventually I would like it to be. But right now the signature file parser is very experimental.

queries/fsharp/highlights.scm
clason: (This applies to all occurrences; `#match?` is to be avoided at all costs.)

queries/fsharp/highlights.scm
```diff
@@ -0,0 +1,372 @@
...
+  "~"
+  "->"
+  "<-"
+  "&&"
+  "||"
+  ":>"
+  ":?>"
+  (infix_op)
+  (prefix_op)
+] @operator
```
ribru17, line 220: Needs 
```
(generic_type
  [
    "<"
    ">"
  ] @punctuation.bracket)
```
after the operator groups too pls

queries/fsharp/highlights.scm
```diff
@@ -0,0 +1,372 @@
...
+  (#eq? @operator "|>"))
+
+((infix_expression
+  .
+  (_) @function.call
+  .
+  (infix_op) @operator
+  .
+  (_))
+  (#eq? @operator "<|"))
```
ribru17, line 138: I think it would also be good to have
```
(function_declaration_left
  (identifier) @function)
```

queries/fsharp/highlights.scm
ribru17, line 1: I could be wrong but could all general `@variable` captures be replaced with just `(identifier) @variable`?

queries/fsharp/injections.scm
ribru17: I think not sadly, due to the `lua-match` usage

ribru17:
It looks like the parser may need a *tiny* bit of work, based on my very minimal testing

clason:
Let's wait for that then.

Nsidorenco:
Thank you for all the detailed suggestions on the queries @ribru17. It is nice to getting new eyes looking at it!
Could you perhaps elaborate on what problems you experienced with the parser when testing it out?

ribru17:
No problem! Maybe the problems aren't that severe but when parsing [this file from Github](https://github.com/yatli/fvim/blob/ee4316c505072cd22ec41f62f1b3081399bf308a/ui.fs) I got several errors which affected the highlighting. But I only tested with one other file, and that one was parsed perfectly

lucario387:
is fsharp similar enough to ocaml so that you can extend from ocaml and use it? Or is this a completely different language from ocaml but share some similar stuffs

queries/fsharp/highlights.scm
```diff
@@ -0,0 +1,372 @@
...
+  "("
+  ")"
+  "{"
+  "}"
+  "["
+  "]"
+  "[|"
+  "|]"
+  "{|"
+  "|}"
```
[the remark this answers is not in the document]
Nsidorenco: `[]` is a little funny since it actually a type constructor, e.g. `int[]` so it gets highlighted by the type queries

queries/fsharp/highlights.scm
Nsidorenco: I have previously had problems with this approach, since a lot of constructs wrap identifier in some way. 
For example, if a query matches `(long_identifier) @some_capture` then even though that comes later in the file the `(identifier) @variable` capture would override it, since it is more specific.

Nsidorenco:
@lucario387 they are both ML-based languages but at this point they mostly just share common concepts. The main difference is that parsing in fsharp is whitespace sensitive

queries/fsharp/highlights.scm
ribru17: Ah, I see- I think in this case a solution would be to replace `(long_identifier) @some_capture` with `(long_identifier (identifier) @some_capture)` (but I don't know if this would change the semantics of the capture too much- I think it wouldn't hopefully?)

Nsidorenco:
@ribru17 thanks for the feedback - I have ironed out the outliers in the repo! 
My own belief about the current state of the  parser is that is able to parse the vast majority of fsharp code, but since fsharp parsing is whitespace sensitive there is likely always someone why has figured out a funky way to indent something that I did not account for 😆 

Regardless, I think this is ready for another look when you have the time!

clason:
But all checks are failing now?

Nsidorenco:
Whoops, I was a bit too quick typing out a capture group. CI should be happy again now :)

queries/fsharp/injections.scm
Nsidorenco: I tried this out and it wasn't allowed because of the `lua-match`

queries/fsharp/highlights.scm
Nsidorenco: I have a concrete example where I find this a little troublesome - perhaps I'm just structuring it wrong so please give me your thoughts.

For application expressions I have the query:
```
(application_expression
   . (_) @function.call)
```
The first node of a `(application_expression)` can be any expression itself, which means I would enumerate out all the ways to reach the identifier. For example an application expression could look like:

```fsharp 
(SomeType.ListOfFunctions[1].Fun) x
```

Here, I would expect `Fun` to be highlighted as a `function.call`, but to capture the identifier in a query I would need something like:
```
(application_expression
   .
   (paren_expression (dot_expression (identifier) @function.call))
```

It is probably doable to map all the combinations of significant identifier nodes that would appear in the first position of an `application_expression`, but it seems rather cumbersome

ribru17: Ah, thanks for the clarification! That makes sense, in that case you're right probably best not to have such a broad capture

ribru17 (approved):

queries/fsharp/highlights.scm
```diff
@@ -0,0 +1,411 @@
...
+] @type
+
+(member_signature
+  .
+  (identifier) @function.method)
+
+(member_signature
+  (curried_spec
+    (arguments_spec
+      "*"* @operator
```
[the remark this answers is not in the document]
Nsidorenco: Agreed!
I'll keep the "?" in the group since it has a special meaning in this context

clason (approved):

PR #7172, opened 2024-09-22
feat(jsdoc): disable spelling for identifiers

kmoschcau:


queries/jsdoc/highlights.scm
```diff
@@ -19,6 +19,13 @@
 
 (identifier) @variable
 
+(tag
+  (identifier) @nospell)
+
+(tag
+  (optional_identifier
+    (identifier) @nospell))
+
```
ribru17, line 28: Maybe this should just be:
```suggestion
(identifier) @variable @nospell
```
that way it works with identifiers in `(path_expression)`, `(member_expression)`, etc.

kmoschcau: You are probably right. I changed it.

PR #7192, opened 2024-09-28
fix(log): use vim.notify for printing logs

seblj:
Problem:

When using `vim.api.nvim_echo` for printing out information as part of `:TSUpdate` and when installing/uninstalling, it could prompt `Press ENTER or type command to continue` if the screen is not wide enough. This is super frustrating. This happens with me all the time if I am using it on my macbook 14 inch with a split screen. I have also had it happening by using it fullscreen on my mac

Solution:

Use `vim.notify` which can be overridden and can therefore remove the problem. It is for example possible to use plugins like https://github.com/j-hui/fidget.nvim or https://github.com/rcarriga/nvim-notify (or possibly others) to not get this `Press ENTER or type command to continue`

### DIFF
--- a/lua/nvim-treesitter/log.lua
+++ b/lua/nvim-treesitter/log.lua
@@ -1,4 +1,5 @@
 local echo = vim.api.nvim_echo
+local notify = vim.notify
 
 -- TODO(lewis6991): write these out to a file
 local messages = {} --- @type {[1]: string, [2]: string?, [3]: string}[]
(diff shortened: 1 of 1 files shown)


clason:
Thank you; we discussed that before but rejected it because we prefer the current styling and want to wait for a better message interface. (I remind that the `main` branch is not yet released.)

seblj:
Alright, that's fair. I know it's not released yet, but I really like how the rewrite is done, and I am ready and prepared to deal with potential breaking changes. So I'll just continue using it

clason:
That's fine, of course :)

PR #7193, opened 2024-09-28
fix(markdown): don't set link metadata twice

ribru17:
Otherwise the same link will be opened twice when the cursor is on the link text itself.

I checked `grammar.js` before this change to make sure there are no other instances of `link_destination` that would no longer be covered by this change

### DIFF
--- a/queries/markdown_inline/highlights.scm
+++ b/queries/markdown_inline/highlights.scm
@@ -91,9 +91,6 @@
   (email_autolink)
 ] @markup.link.url @nospell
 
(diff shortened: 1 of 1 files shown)


clason:
Alternatively we could keep this and remove the specialized patterns above?

ribru17:
If we did that then links would no longer be openable when the cursor is on just the label text; we could capture everything as just `(_ (link_destination) (...metadata)) @_url` but that might be sketchy

PR #7195, opened 2024-09-29
fix(parsers): adapt to upstream change in Nvim 0.11

clason:
In Nvim 0.11, `vim.treesitter.lang.get_lang(filetype)` falls back to `filetype` by default, so filetypes no longer have to be registered for languages with the same name.

https://github.com/neovim/neovim/pull/30379

Also, bye-bye metatable!

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -58,12 +58,15 @@ end
 ---@field readme_name string|nil
 
 ---@type ParserInfo[]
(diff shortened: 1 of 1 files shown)


ribru17 (approved):

ribru17:
@clason Do we not need to register at all now? I am noticing this commit breaks (doesn't start) highlighting for parsers with names differing from the filetype name (like `git_config`)

clason:
Right, that was too simple. I will fix.

PR #7196, opened 2024-09-29
feat(tact): update parser and queries

novusnota:


queries/tact/highlights.scm
```diff
@@ -94,9 +95,8 @@
 (tlb_serialization
   "as" @keyword
   type: (identifier) @type.builtin
-  (#any-of? @type.builtin
-    "int8" "int16" "int32" "int64" "int128" "int256" "int257" "uint8" "uint16" "uint32" "uint64"
-    "uint128" "uint256" "coins" "remaining" "bytes32" "bytes64"))
+  (#match? @type.builtin
```
clason, line 98: Sorry, but that is inacceptable. Always prefer an explicit list if possible; otherwise use `#lua-match?` and Lua patterns. If necessary, use multiple patterns instead of shoving everything into a single complicated regex.

clason (changes requested):

queries/tact/highlights.scm
novusnota: Sure, will fix it in a moment

novusnota: Resolved, please take a look

clason: Do you really need that full generality (uint69)? It's not quite clear to me what you are trying to capture here.

novusnota: In the regexes there I expressed ranges `int1` - `int256` and `uint1` - `uint256`. Could've went for `u?int[0-9][0-9]?[0-9]?`, but that would have been too general (i.e. allowing `int0` and `int999`), so here we are :)

clason: So are you actually expecting uint37? Such patterns are extremely performance critical, so it's important to be flexible here. Again, it's much better to have a long list of a few dozen explicit values than to have a clever but expensive pattern.

clason: (Obviously I know nothing about tact, so treat them as honest questions.)

novusnota: Yeah, I do. But isn't this kind of regex rather cheap? It doesn't do lookaheads (expensive) or lookbehinds (_very_ expensive), just basic `optional?` and `[single character from here]`. Not even `*` for repeats or `()` groups as before

clason: Regex are never cheap in this context; even Lua patterns are a different order of magnitude, and that matters a lot when you're running in a hot loop (which this very much is, being part of the main display loop).

clason: And if you have to, why not simply do `uint[0-9]*`? What's the harm in highlighting `uint29886`?

(Tree-sitter is used for correctly highlighting _correct_ code, not for linting incorrect code.)

clason: (And you need separate patterns for that; you can't chain predicates this way.)

ribru17:
LGTM once @clason's suggestion is applied (of breaking into separate queries). Out of curiosity, any reason to remove `@function.builtin`?

queries/tact/highlights.scm
novusnota: > (Tree-sitter is used for correctly highlighting _correct_ code, not for linting incorrect code.)

Went with that idea and just removed all the patterns there. Incorrect code would still be highlighted now, but that is fixable on the linting/diagnostics side from, say, future LSPs

novusnota:
@ribru17 Thought that they don't make much sense to maintain as their (`@function.builtin`) highlighting rarely differs from the general function capture (`@function`)

ribru17 (approved):

clason (approved):

PR #7198, opened 2024-09-30
feat(superhtml): add superhtml parser and queries

rockorager:
SuperHTML is a templating language used by the zine static site
generator.

### DIFF
--- a/queries/superhtml/highlights.scm
+++ b/queries/superhtml/highlights.scm
@@ -1,0 +1,44 @@
+(doctype) @constant
+
+(comment) @comment
+
+(tag_name) @tag
+
+((tag_name) @string.special
+  (#any-of? @string.special "super" "extend"))
+
+(attribute_name) @attribute
+
+(attribute_value) @string
+
+((element
+  (start_tag
+    (attribute
+      (attribute_name) @_attribute
+      [
+        (attribute_value) @markup.link.url
+        (quoted_attribute_value
+          (attribute_value) @markup.link.url)
+      ]))
+  (element
+    (start_tag
+      (tag_name) @tag)))
+  (#eq? @tag "super")
+  (#eq? @_attribute "id"))
+
+(element
+  (start_tag
+    (tag_name) @string.special)
+  (#eq? @string.special "super"))
+
+"\"" @string
+
+[
+  "<"
+  ">"
+  "</"
+  "/>"
+  "<!"
+] @punctuation.bracket
+
+"=" @punctuation.delimiter
--- a/lockfile.json
+++ b/lockfile.json
@@ -728,6 +728,9 @@
   "supercollider": {
     "revision": "1a8ee0da9a4f2df5a8a22f4d637ac863623a78a7"
   },
+  "superhtml": {
+    "revision": "3edb67236291d45ba97a79cdac3b91692487a352"
+  },
   "surface": {
     "revision": "f4586b35ac8548667a9aaa4eae44456c1f43d032"
   },
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2113,6 +2113,18 @@ list.supercollider = {
   maintainers = { "@madskjeldgaard" },
 }
 
+list.superhtml = {
+  install_info = {
+    url = "https://github.com/kristoff-it/superhtml",
+    files = {
+      "src/parser.c",
+      "src/scanner.c",
+    },
+    location = "tree-sitter-superhtml",
+  },
+  maintainers = { "@rockorager" },
+}
+
 list.surface = {
   install_info = {
     url = "https://github.com/connorlay/tree-sitter-surface",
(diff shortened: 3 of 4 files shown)


rockorager:
CI failure looks unrelated:

image

queries/superhtml/highlights.scm
lucario387, line 1: Query order looks reversed for `tag_name` and `attribute_name`

rockorager: I ran `check-queries.lua` and didn't see an issue - what should be changed there?

clason:
Duplicate: #7206

queries/superhtml/highlights.scm
lucario387: The last query is applied here, which means the latter `tag_name` will take priority over the former, making the first one never used. Reversing the two queries capturing tag_name will let it be the conditional query that you might want (this is the default behavior since tree-sitter 0.22.0)

rockorager: Like this?

```diff
diff --git a/queries/superhtml/highlights.scm b/queries/superhtml/highlights.scm
index c45d21d75106..12e96ba5b013 100644
--- a/queries/superhtml/highlights.scm
+++ b/queries/superhtml/highlights.scm
@@ -2,11 +2,11 @@
 
 (comment) @comment
 
+(tag_name) @tag
+
 ((tag_name) @string.special
   (#any-of? @string.special "super" "extend"))
 
-(tag_name) @tag
-
 ((element
   (start_tag
     (attribute
```

clason: And same for `attribute_name` below.

0001-feat-ziggy-add-ziggy-parser-and-queries.patch
clason, line 1: I think these files should not be here...

queries/superhtml/highlights.scm
```diff
@@ -0,0 +1,47 @@
...
+
+((tag_name) @string.special
+  (#any-of? @string.special "super" "extend"))
+
+(attribute_name) @attribute
+
+((element
+  (start_tag
+    (attribute
+      (attribute_name) @attribute
```
lucario387, line 15: Please change this to `@_attribute` so it's not highlighted twice

queries/superhtml/highlights.scm
```diff
@@ -0,0 +1,47 @@
...
+    (start_tag
+      (tag_name) @tag)))
+  (#eq? @tag "super")
+  (#eq? @attribute "id"))
+
+((element
+  (start_tag
+    (tag_name) @string.special
+    (attribute
+      (attribute_name) @attribute)+))
```
lucario387, line 31: so is this. don't need to set `@attribute` since your general `(attribute_name) @attribute` query up above already cover it

queries/superhtml/highlights.scm
```diff
@@ -0,0 +1,47 @@
...
+  (#eq? @string.special "super"))
+
+[
+  "\""
+  (attribute_value)
```
lucario387, line 36: `(attribute_value)` needs to be up top, probably near `(attribute_name) @attribute`.

queries/superhtml/highlights.scm
```diff
@@ -0,0 +1,46 @@
...
+      [
+        (attribute_value) @markup.link.url
+        (quoted_attribute_value
+          (attribute_value) @markup.link.url)
+      ]))
+  (element
+    (start_tag
+      (tag_name) @tag)))
+  (#eq? @tag "super")
+  (#eq? @attribute "id"))
```
lucario387, line 27: As you changed `@attribute` to `@_attribute`, please also do so for this line

queries/superhtml/highlights.scm
```diff
@@ -0,0 +1,44 @@
...
+      [
+        (attribute_value) @markup.link.url
+        (quoted_attribute_value
+          (attribute_value) @markup.link.url)
+      ]))
+  (element
+    (start_tag
+      (tag_name) @tag)))
+  (#eq? @tag "super")
+  (#eq? @_attribute "id"))
```
lucario387, line 27: This is risky. Are you wanting to capture the tag_name and its corresponding attributes, or capturing elements whose have `id` attributes and have child `super` elements?

For example, with an HTML code of


```
<foo>
  <bar id="something">
  <bar1>
  <baz>
  <bez>
  <super>
</foo>
```
would super be highlighted as `@tag` here?

rockorager: Hmm I'm not 100% positive the original intent of the author, but based on usage of SuperHTML I believe it makes sense to want to capture elements who have `id` and a `<super>` child.

In your example, I believe `<super>` should be highlighted as `@tag`

kristoff-it: Yes, can confirm, it in this example super and the id attribute should be highlighted

lucario387: Can you temporarily comment this query out? I'm afraid this is too costly, considering this will capture all non top-level `element` nodes (which there is guaranteed to be a lot of them) before filtering.

lucario387 (approved):

PR #7202, opened 2024-09-30
fix(parsers): register explicit filetypes unconditionally

clason:
@ribru17 (but proper review this time ;))

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -58,15 +58,12 @@ end
 ---@field readme_name string|nil
 
 ---@type ParserInfo[]
-local list = {}
-if vim.fn.has "nvim-0.11" == 0 then
-  setmetatable(list, {
-    __newindex = function(table, parsername, parserconfig)
-      rawset(table, parsername, parserconfig)
-      ts.language.register(parsername, parserconfig.filetype or parsername)
-    end,
-  })
-end
+local list = setmetatable({}, {
+  __newindex = function(table, parsername, parserconfig)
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -58,15 +58,12 @@ end
...
-    __newindex = function(table, parsername, parserconfig)
-      rawset(table, parsername, parserconfig)
-      ts.language.register(parsername, parserconfig.filetype or parsername)
-    end,
-  })
-end
+local list = setmetatable({}, {
+  __newindex = function(table, parsername, parserconfig)
+    rawset(table, parsername, parserconfig)
+    ts.language.register(parsername, parserconfig.filetype or (vim.fn.has "nvim-0.11" == 0 and parsername))
```
ribru17, line 64:
```suggestion
    if parserconfig.filetype or vim.fn.has "nvim-0.11" == 0 then
      ts.language.register(parsername, parserconfig.filetype or parsername)
    end
```
Right now the right most expression gets evaluated as a boolean on nightly neovim, so this throws an error

ribru17 (approved):

clason:
I remember now why I tried to do it like that: I didn't want to have the `has('nvim-0.11')` check inside the metatable method, for performance reasons. But I'll guess people lazy-load anyway...

ribru17:
Interesting. Is it possible to just leave a note to remove that section after dropping 0.10 compat and then just keep it the way it was? Or are the unnecessary `register`s actually bad

clason:
I would have just pulled the check out. The extra register leads to duplicate entries in get_lang.

PR #7206, opened 2024-10-01
feat(angular): update highlight query

ninoDeme:
updated angular highlight query to support `@let` expressions

### DIFF
--- a/queries/angular/highlights.scm
+++ b/queries/angular/highlights.scm
@@ -56,6 +56,11 @@
   key: ((identifier) @variable.builtin
     (#eq? @variable.builtin "$implicit")))
 
+[
+  (control_keyword)
+  (special_keyword)
+] @keyword
+
 ((control_keyword) @keyword.repeat
   (#any-of? @keyword.repeat "for" "empty"))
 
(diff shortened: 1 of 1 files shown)


queries/angular/highlights.scm
```diff
@@ -68,6 +68,9 @@
 ((control_keyword) @keyword.exception
   (#eq? @keyword.exception "error"))
 
+((control_keyword) @keyword
+  (#eq? @keyword "let"))
```
clason, line 72: Wouldn't it be easier to capture `(control_keyword) @keyword` first and then override the special cases?

queries/angular/highlights.scm
```diff
@@ -68,6 +68,9 @@
...
+
 (special_keyword) @keyword
```
clason, line 74:
```suggestion
[
  (control_keyword)
  (special_keyword)
] @keyword
```
and move up

clason:
@dlvandenberg

queries/angular/highlights.scm
ninoDeme: makes sense

PR #7223, opened 2024-10-07
Fix(auto_install): Install parser if Filetype already known

nikolightsaber:
If setup function was called after `FileType` autocommand triggers, the installation of parser was not done.

This commit checks if filetype is already know before adding the autocommand.

-------------

I was trying to setup treesitter on VeryLazy event with lazy.  

After a clean install I noticed non of the parcers being installed. tracked down to this and tried to fix.

------------

Feel free to be critical about the code I have a feeling it can be better.


ribru17:
Currently, lazy loading of this plugin is not supported, so this sort of behavior is use-at-your-own-risk. Also based on the code it looks like if this block runs (when lazy loading, presumably), the following `FileType` autocmd is not even run at all? This seems undesirable, though I may be misunderstanding

nikolightsaber:
Indeed I forgot the case where you open a file of a different type then the first file you open.

Just fixed that, I still think it is worth trying to install filetype that is open the moment setup is called.

ribru17 (approved):

ribru17:
Nah, just needs approval from @clason :)

PR #7225, opened 2024-10-08
feat(cylc): add cylc parser and queries

elliotfontaine:
[elliotfontaine/tree-sitter-cylc](https://github.com/elliotfontaine/tree-sitter-cylc) (TS Grammar)
[cylc/cylc-flow](https://github.com/cylc/cylc-flow) (Targeted Language)

> Cylc (pronounced silk) is a general purpose workflow engine that also manages cycling systems very efficiently. It is used in production weather, climate, and environmental forecasting on HPC, but is not specialized to those domains.


elliotfontaine:
The tests fails because of captures that are used for `#eq?` / `#match?` predicates.

queries/cylc/folds.scm
```diff
@@ -0,0 +1,10 @@
+[
+  (multiline_string)
+  (multiline_graph_string)
+] @fold
+
+; any section
+(_
+  brackets_open: _
+  name: _?
+  brackets_close: _) @fold
```
lucario387, line 10: explicitly declaring all of the `*_section` will be better

lua/nvim-treesitter/parsers.lua
```diff
@@ -348,6 +348,15 @@ list.cue = {
   maintainers = { "@amaanq" },
 }
 
+list.cylc = {
+  install_info = {
+    url = "https://github.com/elliotfontaine/tree-sitter-cylc",
+    files = { "src/parser.c" },
+    branch = "main",
```
clason, line 355:
```suggestion
```
This is only needed for a non-default branch (regardless of the name).

queries/cylc/highlights.scm
```diff
@@ -0,0 +1,104 @@
...
+(task_parameter
+  "<" @punctuation.bracket
+  name: (_)? @keyword.directive
+  ","? @punctuation.delimiter
+  "="? @operator
+  selection: (_)? @keyword.directive
+  ">" @punctuation.bracket)
+
+(intercycle_annotation
+  "[" @punctuation.bracket
```
clason, line 73: These should just be captured in a simple list, not inside a complicated pattern.

(And in general your grammar overuses fields; if you want to capture `xtrigger: _`, then it would be better to name the node instead and capture `(xtrigger)` itself -- then you don't need all these complicated patterns that are just variations on a theme.)

clason:
And you can only use the captures groups listed in CONTRIBUTING.md (for their listed purpose); others are not allowed to have consistency for colorschemes.

clason:
> The tests fails because of captures that are used for `#eq?` / `#match?` predicates.

For "private use" captures, you can prefix them with an underscore (e.g., `@_key`).

queries/cylc/highlights.scm
```diff
@@ -0,0 +1,104 @@
+(include_statement
+  directive: _ @keyword.directive
+  path: _ @string.special.path)
+
+(comment) @comment
```
ribru17, line 5:
```suggestion
(comment) @comment @spell
```

queries/cylc/injections.scm
```diff
@@ -0,0 +1,22 @@
...
+  (#set! injection.language "comment"))
+
+((setting
+  key: (key) @key
+  (#match? @key "^script$|-script$|^script-")
+  value: (_
+    (string_content) @injection.content))
+  (#set! injection.language "bash"))
+
+; Requires no spacing around "=" in environment settings for proper highlighting.
```
ribru17, line 11: It seems like maybe this whole thing should be commented out then (or just removed), I have seen a lot of files with spaces between `=` (which is not valid bash) e.g. https://github.com/hjoliver/bomdr/blob/99354074846f738c6c13d8e747db7a7696855089/model1/flow.cylc#L5

Also there can be jinja expressions in here looks like, which is also not valid bash and should be handled more specifically

queries/cylc/highlights.scm
```diff
@@ -0,0 +1,104 @@
...
+  (recurrence) @number)
+
+(graph_task
+  xtrigger: _? @keyword.modifier
+  suicide: _? @keyword.modifier
+  name: _ @function)
+
+(task_parameter
+  "<" @punctuation.bracket
+  name: (_)? @keyword.directive
```
ribru17, line 66: Would task parameter names make more sense as `@variable.parameter`? Genuine question, I don't know cylc at all

queries/cylc/highlights.scm
```diff
@@ -0,0 +1,104 @@
...
+
+(multiline_graph_string
+  quotes_open: _ @string
+  quotes_close: _ @string)
+
+[
+  (graph_logical)
+  (graph_arrow)
+  (graph_parenthesis)
+] @keyword.operator
```
ribru17, line 54: I think `graph_arrow` should just be `@operator`, also `graph_parenthesis` sounds like it should be `@punctuation.bracket` instead? Not sure what graph logical does but maybe should just be `@operator` as well

queries/cylc/highlights.scm
```diff
@@ -0,0 +1,104 @@
...
+
+(task_output
+  ":" @tag
+  (nametag) @variable.builtin)
+
+(task_output
+  "?"? @tag)
+
+(setting
+  key: (key) @variable
```
ribru17, line 85: I think this should be `@property`? (the "key" in key value pairs)

clason:
@elliotfontaine friendly ping

queries/cylc/highlights.scm
```diff
@@ -0,0 +1,104 @@
...
+(datetime) @number.float
+
+[
+  (jinja2_expression)
+  (jinja2_statement)
+  (jinja2_comment)
+  (jinja2_shebang)
+] @keyword.directive
+
+(ERROR) @comment.error
```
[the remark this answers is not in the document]
elliotfontaine: Obviously this would be inappropriate if Cylc had a Language Server following LSP, but since it's not the case couldn't it be allowed as a placeholder for now?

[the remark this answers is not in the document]
clason: Nope, sorry. This will turn editing into blinkenlights. (Since `@comment.error` is a highlight group that is defined by default.)

elliotfontaine:
> @elliotfontaine friendly ping

Hi @clason !
Sorry, I didn't have much time to work on these fixes the past week. I'll update the PR in the coming days.

queries/cylc/highlights.scm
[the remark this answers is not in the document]
ribru17: Note that we used to highlight error nodes and made a decision to remove all of them (it's also difficult to properly assign priorities, sometimes error nodes father many other nodes, etc)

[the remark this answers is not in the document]
clason: Yes, but we used to capture them as `@error`, which was not a standard highlight group. Since we enforce standard groups for consistency now, that is no longer an option.

(You can always extend the queries here in your config or plugin, anyway.)

queries/cylc/folds.scm
lucario387: This is still not resolved yet. I'd prefer explicitly declaring each section like how you did it in indents.scm

queries/cylc/highlights.scm
lucario387: Missing changes here.

queries/cylc/injections.scm
elliotfontaine: Okay. I'll reimplement it if `tree-sitter` add support for the #3625 proposal.

clason: Wrong link?

queries/cylc/folds.scm
elliotfontaine: Sorry, I'm adding all these changes from VSCode and I resolved the conversation from there, while not having pushed the fixes yet.
I'm still learning the GitHub PR "best practices" workflow :/

clason: "Don't use VS Code" :)

elliotfontaine: **opens MS Word instead.*

queries/cylc/highlights.scm
elliotfontaine: task "parameters" in Cylc are basically used for templating. Less powerful than Jinja2 but less verbose. So I decided to use the same capture as Jinja2 nodes.
https://cylc.github.io/cylc-doc/latest/html/user-guide/writing-workflows/parameterized-tasks.html

queries/cylc/highlights.scm
elliotfontaine: > And in general your grammar overuses fields.

I'll make that a priority for the next version of the grammar. Sadly Tree-sitter lacks good resources for "style" and best practices so I'm learning by making these design mistakes 😕

elliotfontaine: @clason Do you have any rule of thumb on when to use fields *vs* when to use aliasing ? Is aliasing to a virtual node name (only existing through aliasing) bad practice, or is it ok ?

elliotfontaine:
@clason I followed your advice for the highlights query. It's not perfect but it should be a big step in the right direction.

clason (approved):

ObserverOfTime (approved):

queries/cylc/highlights.scm
ribru17: Sorry for the long delay. In this case I think maybe the jinja2 nodes and these should also be changed to `@variable.parameter`, or at least something besides `@keyword.directive` which doesn't make much sense to me

ribru17 (dismissed):

queries/cylc/highlights.scm
clason: @ribru17 So is there still something that should be changed in this PR?

ribru17: I think these should be changed to variable.parameter here. I can change jinja to match this in another pr (but do we even have jinja in this repo?)

elliotfontaine: Do what you think is best. I won't edit this one though as it would discard the 3 approvals 😅 
As you are reviewers, maybe it won't if either of you does it ?

clason (changes requested):
Should be `@variable.directive`.

clason:
Sorry, `@variable.parameter` like @ribru17 said!

clason (approved):

PR #7239, opened 2024-10-13
fix(slang): adapt to parser changes

amaanq:
The `auto` rule is removed from slang, so inheriting the C++ highlight queries is disallowed, instead C is inherited, with C++ and HLSL copied sans the `auto` rule.

### DIFF
--- a/queries/slang/highlights.scm
+++ b/queries/slang/highlights.scm
@@ -1,4 +1,308 @@
-; inherits: hlsl
+; inherits: c
+
+; cpp
+((identifier) @variable.member
+  (#lua-match? @variable.member "^m_.*$"))
+
+(parameter_declaration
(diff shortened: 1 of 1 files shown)


clason:
Can you explain _how_ the `auto` rule was removed?

amaanq:
> Can you explain _how_ the `auto` rule was removed?

`generate` filters out unused rules now, which fixes panics. slang overrides a rule that uses auto, hence the removal.

clason:
That sounds like a bigger issue with inherited languages?

amaanq:
how so? an inherited language might not allow one rule from the base language, thus it shouldn't generate the lex and parse states for this rule (making the query invalid)

clason:
I mean that we in general cannot assume that grammars that inherit another implies that the queries can (which we do across a number of languages).

PR #7240, opened 2024-10-14
refactor: replace vim.fn with alias fn for consistency

sys9kdr:
nvim-treesitter is aliasing `vim.fn` to `fn` as a local variable in many files. However, using vim.fn instead of fn in many places.

In this pr, I refactored for using fn.


clason:
Thank you, but the `master` branch is essentially frozen and will be superseded by `main` (which is already heavily refactored). This will make syncing query changes harder.

sys9kdr:
okay thx

PR #7260, opened 2024-10-18
feat(xresources): xresources parser and queries

ValdezFOmar:
Parser for `.Xresources` or `.Xdefaults` files.

### DIFF
--- a/queries/xresources/folds.scm
+++ b/queries/xresources/folds.scm
@@ -1,10 +1,10 @@
 [
-  (comment)+
   (elif_directive)
   (elifdef_directive)
   (else_directive)
   (if_directive)
   (ifdef_directive)
-  (include_directive)+
   (preprocessor_comment)
+  (comment)+
+  (include_directive)+
 ] @fold
--- a/queries/xresources/folds.scm
+++ b/queries/xresources/folds.scm
@@ -1,0 +1,10 @@
+[
+  (comment)+
+  (elif_directive)
+  (elifdef_directive)
+  (else_directive)
+  (if_directive)
+  (ifdef_directive)
+  (include_directive)+
+  (preprocessor_comment)
+] @fold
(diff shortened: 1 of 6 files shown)


queries/xresources/folds.scm
```diff
@@ -0,0 +1,10 @@
+[
+  (comment)+
+  (elif_directive)
+  (elifdef_directive)
+  (else_directive)
+  (if_directive)
+  (ifdef_directive)
+  (include_directive)+
+  (preprocessor_comment)
```
ribru17, line 9: Would you mind moving the quantified captures to the end of this alternation? You shouldn't have to, but I've noticed a weird but in tree-sitter where folds of different node kinds in the alternations are merged if they are a quantified one followed by a regular one...

queries/xresources/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+  "#elifndef"
+  "#else"
+  "#endif"
+  (directive)
+] @keyword.directive
+
+(identifier) @variable
+
+(define_directive
+  name: (identifier) @constant)
```
ribru17, line 28: Maybe `@constant` should be `@constant.macro`?

ribru17 (approved):

queries/xresources/highlights.scm
ValdezFOmar: I thought the same but the C queries use `@constant` for this and I wanted to keep them consistent with those (unless the C queries are in the wrong here).

ribru17: Oh thanks for checking that! Yeah IMO that seems wrong for the C queries to be doing that...

ValdezFOmar: I'll change the captures here and open another PR changing the captures for the C queries if that is the desired behavior.

PR #7261, opened 2024-10-18
fix: allow re-running setup and preserve options

dkarter:
# Motivation

I have a lot of plugins that depend on treesitter. In `lazy.nvim` I specify those in the `dependencies` option:

For example:
CleanShot 2024-10-18 at 18 15 54@2x

Since I split the plugin declaration to individual files (which is a common practice), each time one of the plugins that depends on treesitter loads, it resets the options passed to setup.

The existing implementation already allowed for this, but not for all options like `ignore_install` - so that setting was reset and eventually caused things marked as `ignore_install` to still be installed.

# Summary of Changes

- This PR keeps the options the user passed to setup and extends them if more options are passed in.
- I included a test to verify the intended behavior

# Testing Instructions

1. Set up treesitter to `ignore_install` a certain parser (e.g. `tmux`)
2. Set up another plugin and add treesitter as a dependency
3. Make sure the undesired parser is uninstalled (e.g. `:TSUninstall tmux`)
4. Open a file matching the ignored parser - after this PR it shouldn't be auto-installed anymore. 

> [!NOTE]
> I've been running this in a fork for a while and it's been working great for me.


clason:
Thank you, but the master branch is essentially frozen while we prepare (slowly) for the 1.0 rewrite, which massively simplifies the structure so this becomes unnecessary.

PR #7278, opened 2024-10-24
fix(rust): regex injections, highlights

ribru17:
**Problem:** Rust's regex injections were not applied since the string content is behind a `(string_content)` node, and the injections were applied to the parent node without the `include-children` directive.

**Solution:** Apply the injections to the string content. Also highlight them accordingly.

### DIFF
--- a/queries/rust/injections.scm
+++ b/queries/rust/injections.scm
@@ -49,7 +49,8 @@
     name: (identifier) @_new
     (#eq? @_new "new"))
   arguments: (arguments
-    (raw_string_literal) @injection.content)
+    (raw_string_literal
+      (string_content) @injection.content))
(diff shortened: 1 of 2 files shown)


queries/rust/injections.scm
```diff
@@ -71,7 +73,8 @@
     (#eq? @_new "new"))
   arguments: (arguments
     (array_expression
-      (raw_string_literal) @injection.content))
+      (raw_string_literal
+        (string_content)) @injection.content))
```
lucario387, line 77: This is still being applied to `raw_string_literal` and not string_content

ribru17: Oops🤦‍♂️thanks

PR #7286, opened 2024-10-25
fix(formatter): simplify code, remove odd line break

ribru17:
This commit simplifies much of the formatter code, removing the need for break statements, removing some unused captures, and merging some redundant `if` cases. The formatter output is the same except in some cases where a (seemingly unintentional) line break is now no longer added.


lucario387 (approved):
I think this is a backport of the optimization you did in ts_query_ls? Great work

ribru17:
Thank you! :)

clason:
@ribru17 thanks! could you take a look at the rebased version on `main`? I hope I didn't drop any changes we made there for nightly...

ribru17:
@clason Just tested, everything looks good to me!

PR #7292, opened 2024-10-28
feat(mustache): mustache parser and queries

TheLeoP:
Parser for `.mustache` files as defined by the [Mustache spec](https://github.com/mustache/spec)

### DIFF
--- a/queries/mustache/highlights.scm
+++ b/queries/mustache/highlights.scm
@@ -1,0 +1,17 @@
+[
+  (start_delimiter)
+  (end_delimiter)
+] @tag.delimiter
+
+(identifier) @tag.attribute
+
+"." @punctuation.delimiter
+
+[
+  "#"
+  "/"
+  "^"
+  ">"
+] @character
+
+(comment_statement) @comment
--- a/queries/mustache/highlights.scm
+++ b/queries/mustache/highlights.scm
@@ -1,9 +1,13 @@
 [
   (start_delimiter)
   (end_delimiter)
-] @tag.delimiter
+  "{"
+  "}"
+] @punctuation.bracket
 
-(identifier) @tag.attribute
+(identifier) @constant
+
+(partial_content) @constant.macro
 
 "." @punctuation.delimiter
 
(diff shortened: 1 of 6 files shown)


queries/mustache/highlights.scm
```diff
@@ -0,0 +1,17 @@
+[
+  (start_delimiter)
+  (end_delimiter)
+] @tag.delimiter
```
ribru17, line 4: Also `"{"` and `"}"` for things like `}}}`. Also should this be `@punctuation.bracket`?

ribru17:
The parser gives me double free error and crashes neovim for this file:

https://github.com/coq-community/templates/blob/e34c0cbf101ad8b4eaf67f90597091e0fcb339b0/index.md.mustache#L25

queries/mustache/highlights.scm
TheLeoP: > Also "{" and "}" for things like }}}

You are right. I'll add it

> should this be @punctuation.bracket?

I have no strong opinions about this. I could change it

TheLeoP:
> The parser gives me double free error and crashes neovim for this file:

Will check and fix

TheLeoP:
@ribru17 I've fixed some problems with the parser in the document that you sent me. I think now it's being parsed correctly.

But, I couldn't reproduce the double free error (nor the Neovim crash) with that file. I'm new to using C, so I'm a bit lost. Could you point me in the right direction in order to fix this issue?

TheLeoP:
I'm having trouble debugging an oom problem, I'll continue tomorrow

TheLeoP:
I think I've fixed the OOM errors (at least the fuzz CI action isn't finding them anymore).

Also, I may have been unable to reproduce the crashing parsing the mustache document you sent because instead of downloading it, I copied and pasted its content into Neovim and then saved it. One of the OOM errors happened when the file ended without a whitespace and Neovim inserting a newline at the end of the file may have prevented it from happening when I tested it earlier (but should also be fixed now).

ribru17:
I can try and help soon sorry in a busy spell at the moment

PR #7294, opened 2024-10-29
feat(thrift): small improvements

ribru17:
Less capture duplication by separating queries, include folds, move `@variable.member` to `@property`

### DIFF
--- a/queries/thrift/highlights.scm
+++ b/queries/thrift/highlights.scm
@@ -14,7 +14,7 @@
 
 ; Fields
 (field
-  (identifier) @variable.member)
+  (identifier) @property)
 
 ; Parameters
 (function_definition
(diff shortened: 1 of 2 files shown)


queries/thrift/highlights.scm
```diff
@@ -67,17 +67,15 @@
...
-  "{"
-  (identifier) @constant
-  "}")
+  type: (identifier) @type)
 
-; Builtin Types
-(primitive) @type.builtin
+(enum_definition
+  type: (identifier)
+  (identifier) @constant)
```
amaanq, line 74: think this can match identifiers in the enum name, e.g. `bar` in `foo.bar` as well as inside the enum

ribru17: Oops, thank you. Changed it to match after `"{"`

PR #7302, opened 2024-11-01
fix(rst): adornment should be captured as @markup.heading

SilverRainZ:
Left: before, Right: after.

<image src="https://github.com/user-attachments/assets/1289ec04-4340-472f-be2c-19cd873cddac"  width="45%" />
<image src="https://github.com/user-attachments/assets/94d9f84b-2e6e-430a-a385-f4cf4e52d617"  width="45%" />

### DIFF
--- a/queries/rst/highlights.scm
+++ b/queries/rst/highlights.scm
@@ -9,10 +9,7 @@
   (transition)
 ] @punctuation.special
 
-[
-  "bullet"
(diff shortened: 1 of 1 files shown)


clason:
It's a bit weird for this to be an anonymous node -- these are usually used for literals. Probably something that should be fixed (generally) in the grammar.

PR #7305, opened 2024-11-01
feat(runescript): add runescript parser and queries

InfiniteRain:
Add support for runescript/clientscript and the corresponding queries.

This is a scripting language that is used to develop 2004Scape: 

https://github.com/2004Scape/Server

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2624,6 +2624,22 @@ list.templ = {
   maintainers = { "@vrischmann" },
 }
 
+list.runescript = {
+  install_info = {
+    url = "https://github.com/2004Scape/tree-sitter-runescript",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@2004Scape" },
+}
+
+list.clientscript = {
+  install_info = {
+    url = "https://github.com/2004Scape/tree-sitter-runescript",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@2004Scape" },
+}
+
 local M = {
   list = list,
 }
--- a/queries/clientscript/highlights.scm
+++ b/queries/clientscript/highlights.scm
@@ -1,0 +1,99 @@
+; Variables
+
+[
+  (identifier)
+  (local_variable)
+  (game_variable)
+  (constant_variable)
+  (parameter_variable)
+  (declaration_variable)
+] @variable
+
+; Scripts
+
+(script
+  trigger: (identifier) @keyword.directive
+  name: (identifier) @function)
+
+; Calls
+
+(call
+  [
+   "~"
+   "@"
+  ]? @function
+  callee: (identifier) @function)
+
+; Literals
+
+((literal) @constant.builtin
+  (#match? @constant.builtin "^(true|false|null)$"))
+
+((literal) @number
+  (#match? @number "^(-?[0-9]+|[0-9]+_[0-9]+_[0-9]+_[0-9]+_[0-9]+)$"))
+
+((literal) @number
+  (#match?))
+
+(comment) @comment
+
+(string_fragment) @string
+
+(string_interpolation
+  [
+    "<"
+    ">"
+  ] @punctuation.special)
+
+(string_tag) @punctuation.special
+
+; Types
+
+(type) @type
+
+(parameter_type) @type
+
+; Tokens
+
+[
+  ";"
+  ":"
+  ","
+] @punctuation.delimeter
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+] @punctuation.bracket
+
+[
+  "*"
+  "/"
+  "+"
+  "-"
+  "&"
+  "|"
+  ">"
+  ">="
+  "<"
+  "<="
+  "="
+  "!"
+] @operator
+
+(arithmetic
+  "%" @module.operator)
+
+[
+ "default"
+ "if"
+ "else"
+ "while"
+ "calc"
+ "return"
+ "case"
+ (def_type_keyword)
+ (switch_type_keyword)
+] @keyword
--- a/queries/runescript/highlights.scm
+++ b/queries/runescript/highlights.scm
@@ -1,0 +1,99 @@
+; Variables
+
+[
+  (identifier)
+  (local_variable)
+  (game_variable)
+  (constant_variable)
+  (parameter_variable)
+  (declaration_variable)
+] @variable
+
+; Scripts
+
+(script
+  trigger: (identifier) @keyword.directive
+  name: (identifier) @function)
+
+; Calls
+
+(call
+  [
+   "~"
+   "@"
+  ]? @function
+  callee: (identifier) @function)
+
+; Literals
+
+((literal) @constant.builtin
+  (#match? @constant.builtin "^(true|false|null)$"))
+
+((literal) @number
+  (#match? @number "^(-?[0-9]+|[0-9]+_[0-9]+_[0-9]+_[0-9]+_[0-9]+)$"))
+
+((literal) @number
+  (#match?))
+
+(comment) @comment
+
+(string_fragment) @string
+
+(string_interpolation
+  [
+    "<"
+    ">"
+  ] @punctuation.special)
+
+(string_tag) @punctuation.special
+
+; Types
+
+(type) @type
+
+(parameter_type) @type
+
+; Tokens
+
+[
+  ";"
+  ":"
+  ","
+] @punctuation.delimeter
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+] @punctuation.bracket
+
+[
+  "*"
+  "/"
+  "+"
+  "-"
+  "&"
+  "|"
+  ">"
+  ">="
+  "<"
+  "<="
+  "="
+  "!"
+] @operator
+
+(arithmetic
+  "%" @module.operator)
+
+[
+ "default"
+ "if"
+ "else"
+ "while"
+ "calc"
+ "return"
+ "case"
+ (def_type_keyword)
+ (switch_type_keyword)
+] @keyword
--- a/lockfile.json
+++ b/lockfile.json
@@ -907,5 +907,11 @@
   },
   "ziggy_schema": {
     "revision": "42b6f5d7320340bc5903c4c29d34065e8517a549"
+  },
+  "runescript": {
+    "revision": "7b8727a29c1b9ee65b4fcf4792de128e555d4996"
+  },
+  "clientscript": {
+    "revision": "7b8727a29c1b9ee65b4fcf4792de128e555d4996"
   }
 }


clason (dismissed):
The two grammars look very similar, with nearly(?) identical queries. Wouldn't it be better for one query to `; inherit` the other?

clason:
Also, please add injections, at least for `comment`.

clason:
And please run `./scripts/format-queries.lua queries` to make the linter pass.

lua/nvim-treesitter/parsers.lua
```diff
@@ -2624,6 +2624,22 @@ list.templ = {
...
+  install_info = {
+    url = "https://github.com/2004Scape/tree-sitter-runescript",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@2004Scape" },
+}
+
+list.clientscript = {
+  install_info = {
+    url = "https://github.com/2004Scape/tree-sitter-runescript",
```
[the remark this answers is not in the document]
InfiniteRain: The two languages indeed inherit the same syntax, but are counted as two separate languages (legacy reasons that only the devs making the runescape engine back in 2004 know any answers to).

queries/clientscript/highlights.scm
```diff
@@ -0,0 +1,99 @@
...
+; Literals
+
+((literal) @constant.builtin
+  (#match? @constant.builtin "^(true|false|null)$"))
+
+((literal) @number
+  (#match? @number "^(-?[0-9]+|[0-9]+_[0-9]+_[0-9]+_[0-9]+_[0-9]+)$"))
+
+((literal) @number
+  (#match?))
```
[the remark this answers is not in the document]
InfiniteRain: mishap!

lua/nvim-treesitter/parsers.lua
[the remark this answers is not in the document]
clason: If they're truly identical, then you only need to add it once and handle this on the `vim.treesitter.language.register` level (see the table at the top of the file).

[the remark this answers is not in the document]
clason: (In case that wasn't clear from the test failures: you _cannot_ do that (add the same parser twice under different names); the language name here **must** be the same as the name the grammar itself specifies. Please drop the "clientscript" entry and (if appropriate) add an entry in the alias table.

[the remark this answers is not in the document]
InfiniteRain: Yes, it was clear. I will submit a revision when I have time (by eod).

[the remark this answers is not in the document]
clason: Ok, didn't want to push, take your time -- just explain a bit more.

queries/runescript/highlights.scm
```diff
@@ -0,0 +1,92 @@
...
+; Literals
+((literal) @constant.builtin
+  (#any-of? @constant.builtin "true" "false" "null"))
+
+[
+  (integer_literal)
+  (coord_literal)
+] @number
+
+(comment) @comment
```
clason, line 33:
```suggestion
(comment) @comment @spell
```

queries/runescript/highlights.scm
```diff
@@ -0,0 +1,92 @@
...
+  "!"
+] @operator
+
+(arithmetic
+  "%" @operator)
+
+[
+  "default"
+  "if"
+  "else"
```
clason, line 85: these should be `@keyword.conditional` (and similarly for the others; please use as specific a capture -- from the list in CONTRIBUTING.md -- as possible).

queries/runescript/highlights.scm
```diff
@@ -0,0 +1,92 @@
...
+  ]? @function
+  callee: (identifier) @function)
+
+; Literals
+((literal) @constant.builtin
+  (#any-of? @constant.builtin "true" "false" "null"))
+
+[
+  (integer_literal)
+  (coord_literal)
```
clason, line 30: Is that an integer or a float?

queries/runescript/highlights.scm
```diff
@@ -0,0 +1,92 @@
...
+(call
+  [
+    "~"
+    "@"
+  ]? @function
+  callee: (identifier) @function)
+
+; Literals
+((literal) @constant.builtin
+  (#any-of? @constant.builtin "true" "false" "null"))
```
clason, line 26: shouldn't `true` and `false` be `@boolean`?

queries/runescript/highlights.scm
InfiniteRain: Neither, it's a special literal type that represents a map coordinate, it's 5 integers separated by underscores.

queries/runescript/highlights.scm
InfiniteRain: Good point.

queries/runescript/highlights.scm
clason: Hmm, then it's not really a number.

InfiniteRain: I'd still like to keep it as such, if you don't mind. This is pretty much how it's done in the existing non-treesitter syntax highlighting, and we'd like to keep those closely matched.

clason: Sorry, but this is a non-goal for the queries in this repo -- it's much more important to be consistent across the different languages, so colorschemes can target the official standard captures without (too many) language-specific special cases.

@ribru17 ?

clason: Not sure; `@string.special` possibly, which is what is used elsewhere in similar (but not the same) situations.

InfiniteRain: Alright, let's go with that one then.

ribru17: Maybe if the grammar exposes child nodes as `(integer)` `"_"`  ... in the future we can highlight the numbers as numbers and then `@punctuation.delimiter` for underscores

InfiniteRain: > Maybe if the grammar exposes child nodes as (integer) "_" ... in the future we can highlight the numbers as numbers and then @punctuation.delimiter for underscores

That's a good idea for future iterations, but I think I will keep it as @string.special for now (until we change the parsing a little bit).

queries/runescript/highlights.scm
```diff
@@ -0,0 +1,100 @@
...
+  ":"
+  ","
+] @punctuation.delimiter
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+] @punctuation.bracket
```
ribru17, line 64: Also square brackets

queries/runescript/highlights.scm
```diff
@@ -0,0 +1,100 @@
...
+(literal
+  "null" @constant.builtin)
+
+(integer_literal) @number
+
+(coord_literal) @string.special
+
+(comment) @comment @spell
+
+(string_fragment) @string
```
ribru17, line 37: Also for `"\""`

queries/runescript/highlights.scm
ribru17, line 1: We should also highlight e.g. the `"%"` and `"^"` before variables, not sure if it should be `@operator`  or `@type` (docs say these symbols represent variable types?)

queries/runescript/highlights.scm
```diff
@@ -0,0 +1,100 @@
...
+(script
+  trigger: (identifier) @keyword.directive
+  subject: (identifier) @function)
+
+; Calls
+(call
+  [
+    "~"
+    "@"
+  ]? @function
```
ribru17, line 21: Seems like these should be `@operator`? Or am I wrong in this

queries/runescript/highlights.scm
```diff
@@ -0,0 +1,100 @@
...
+  callee: (identifier) @function)
```
ribru17, line 22: Should be `@function.call`

queries/runescript/highlights.scm
InfiniteRain: Square brackets are only used to define the script type (header), nowhere else. I don't think that the should be counted as normal punctuation here. (Feel free to correct me if you think I'm wrong).

queries/runescript/highlights.scm
clason: Yeah but we can do better :) (we have the technology!)

queries/runescript/highlights.scm
InfiniteRain: Same as above; those prefixes define the type of a call `~call()` vs `@call()` vs `call()`. I kept them the same color as the callee itself, but if you think they should be highlighted differently lmk.

clason:
Tangential: We really need a better (or, really, any) template for new parser PRs. Discussions would be much smoother if we asked for a link to the grammar, a syntax specification, and some example files (among other information).

queries/runescript/highlights.scm
clason: Type in what sense?

queries/runescript/highlights.scm
clason: Colorschemes can always choose to make them the same color, but if there are syntactically different elements, we should capture them differently -- otherwise colorschemes have no choice in the matter.

queries/runescript/highlights.scm
ribru17: Actually I just meant we need a highlight for regular double quotes here (as string), which must be `"\"" @string` in the query file. Kind of confusing! But yes we should highlight escape sequences as well as `@string.escape`

queries/runescript/highlights.scm
ribru17: Hmm I think it would still make sense to highlight them like this because they are still brackets, but if you prefer I could see maybe an argument for `@punctuation.special` maybe? cc @clason

queries/runescript/highlights.scm
InfiniteRain: Ahh, makes sense! Yes, I will add both

queries/runescript/highlights.scm
InfiniteRain: "~some_proc()" calls a procedure script (something that returns) and "@some_label()" jumps to a label (a script that doesn't return). I think in this case we can go with the way you proposed in the comment above.

queries/runescript/highlights.scm
InfiniteRain: I'm OK with going with `@punctuation.special`

clason:
It's just plonking a new file in `.github/ISSUE_TEMPLATES`; that doesn't affect the rewrite at all.

queries/runescript/highlights.scm
InfiniteRain: > docs say these symbols represent variable types?

Not types in the normal sense (int, string, etc), but more like "categories". I.e., player variables, global variables, local variables

InfiniteRain: > Colorschemes can always choose to make them the same color, but if there are syntactically different elements, we should capture them differently -- otherwise colorschemes have no choice in the matter.

Alright, I will add highlighting to those

queries/runescript/highlights.scm
clason: That is... special :) Hard to find a good bucket for that -- maybe `@attribute`(`.builtin`)? Otherwise `@punctuation.special` could work.

Basically, my thinking is this: if the grammar does not include these as part of the "function" node, they should not be treated as part of the function name but captured separately. _What_ they should be captured as is another question, which can be hard to answer for non-standard languages. Here we go with best-effort and more leeway to language maintainers.

queries/runescript/highlights.scm
clason: > > docs say these symbols represent variable types?
> 
> Not types in the normal sense (int, string, etc), but more like "categories". I.e., player variables, global variables, local variables

So these are labels, similarly to the `~` and `@` below? Then we should use the same (type of) capture: `@attribute` or `@punctuation.special`.

(Rule of thumb: capture different things differently while trying to be consistent both within and across languages.)

queries/runescript/highlights.scm
InfiniteRain: > That is... special :)

Brought to us by Andrew Gower himself at the dawn of the century :)

> maybe @attribute(.builtin)?

This one actually fits pretty well.

queries/runescript/highlights.scm
clason: > > docs say these symbols represent variable types?
> 
> Not types in the normal sense (int, string, etc), but more like "categories". I.e., player variables, global variables, local variables

But

> syntax wise it's similar to any C-style language, except variable names are prefixed with their type (eg. %varname is an integer).
> — Mod Chris E Game Engine Q&A

🤔

InfiniteRain: It's an oversimplification, the `%` prefix refers to a global variable. It can only store ints in a player file, and all other types except for strings are ints under the hood. Variables with this prefix still support runtime strings.

This language is hella weird, it has way more quirks than this if you actually start looking into it.

clason: Yeah, that's what I'm starting to see... (At least it doesn't have loobeans.)

ObserverOfTime:
Those are for issues, not pull requests. GitHub only supports a single template for pull requests.

clason:
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

clason:
@ribru17 good to go now?

@InfiniteRain please rebase (the constant unresolvable conflicts happen because you opened a pull request from your master branch; you should never do that for this and other reasons)

ObserverOfTime:
> https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository

But there's no selection menu.

queries/runescript/highlights.scm
```diff
@@ -0,0 +1,117 @@
...
+(game_variable
+  [
+    "%"
+    ".%"
+  ] @attribute.builtin
+  name: (identifier) @variable)
+
+(constant_variable
+  "^" @attribute.builtin
+  name: (identifier) @variable)
```
ribru17, line 17: Nit: all these `(identifier) @variable` highlights are unnecessary because of the first query

queries/runescript/highlights.scm
```diff
@@ -0,0 +1,117 @@
...
+
+(integer_literal) @number
+
+(coord_literal) @string.special
+
+(comment) @comment @spell
+
+[
+  (string_fragment)
+  (escape_sequence)
```
ribru17, line 47: This needs to be `@string.escape`

queries/runescript/highlights.scm
```diff
@@ -0,0 +1,117 @@
...
+  [
+    "%"
+    ".%"
+  ] @attribute.builtin
+  name: (identifier) @variable)
+
+(constant_variable
+  "^" @attribute.builtin
+  name: (identifier) @variable)
+
```
ribru17, line 18: Last thing, can we get

```scm
(parameter
  (local_variable
    name: (identifier) @variable.parameter))
```
here? Making sure it is after the identifier highlight so it overrides it

ribru17 (commented):
Good to go after changes addressed

queries/runescript/highlights.scm
InfiniteRain: Could you please elaborate further? I'm not sure I understand

InfiniteRain: Ah, yes. Makes perfect sense. Thank you :)

ribru17 (approved):
Thanks!

InfiniteRain:
> Thanks!

Thank you for the support :)

clason:
Thanks for your patience!

PR #7306, opened 2024-11-01
feat: sway programming language

ribru17:
https://github.com/FuelLabs/sway

### DIFF
--- a/queries/sway/highlights.scm
+++ b/queries/sway/highlights.scm
@@ -1,0 +1,336 @@
+(type_identifier) @type
+
+(identifier) @variable
+
+(field_identifier) @variable.member
+
+(escape_sequence) @string.escape
+
+(primitive_type) @type.builtin
+
+(boolean_literal) @boolean
+
+(integer_literal) @number
+
+(float_literal) @number.float
+
+(char_literal) @character
+
+; -------
+; Paths
+; -------
+(use_declaration
+  argument: (identifier) @module)
+
+(use_wildcard
+  (identifier) @module)
+
+(mod_item
+  name: (identifier) @module)
+
+(scoped_use_list
+  path: (identifier)? @module)
+
+(use_list
+  (identifier) @module)
+
+(use_as_clause
+  path: (identifier)? @module
+  alias: (identifier) @module)
+
+; ---
+; Remaining Paths
+; ---
+(scoped_identifier
+  path: (identifier)? @module
+  name: (identifier) @module)
+
+(scoped_type_identifier
+  path: (identifier) @module)
+
+[
+  "*"
+  "'"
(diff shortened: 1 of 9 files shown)


queries/sway/highlights.scm
```diff
@@ -0,0 +1,336 @@
...
+    (type_identifier) @constructor
+    (scoped_type_identifier
+      name: (type_identifier) @constructor)
+  ])
+
+; ---
+; Other PascalCase identifiers are assumed to be structs.
+; ---
+((identifier) @type
+  (#match? @type "^[A-Z]"))
```
clason, line 309: You should know better ;)

ribru17: Yeah, I was wondering about whether to remove though because e.g. rust has many such captures? I should mention this language is nearly identical to rust (but doesn't inherit it annoyingly). Seems weird to arbitrarily disallow these kinds of queries while many languages still have them

clason: I was referring to the `#match?` itself...

(I agree that we should remove the "capital convention" everywhere or nowhere -- with a personal preference for "everywhere"; the point of tree-sitter is that we can do better than that kind of drunk aiming.)

ribru17: Oh oops :laughing: thank you

ribru17:
Gonna add indents soon

queries/sway/highlights.scm
```diff
@@ -0,0 +1,336 @@
...
+  "&" @keyword.modifier)
+
+(self_parameter
+  "&" @keyword.modifier)
+
+; -------
+; Guess Other Types
+; -------
+((identifier) @constant
+  (#lua-match? @constant "^[A-Z][A-Z%d_]*$"))
```
clason, line 268: As talked about before, I _really_ want to do away with these "convention captures" -- with tree-sitter, we should be able to do better.

This is not specific to this language, of course, but I'm wondering if at some point we should stop adding new ones even before removing the old ones?

clason: And obviously not a blocker; just food for thought.

ribru17: Sure, that makes sense. Personally I think in that case we should stop allowing these queries for new languages, although maybe this could be a special case since the language is so similar to rust. Maybe would make more sense to remove them here at the time of removing the rust ones (which maybe would be when removing all of them?)

clason: Sure, as I said, don't want to block this!

clason:
@ribru17 `tree-sitter generate` fails for this grammar (bad regex); this is something we have to check (manually) before adding -- another item for the pr template!

PR #7309, opened 2024-11-02
fix(wgsl): indent for multiline type constructor and function call expression

ganhaque:


ribru17:
Would you mind adding a few new line tests just in case? You can see e.g. the swift example: https://github.com/nvim-treesitter/nvim-treesitter/blob/38c12cb9ac8e1c547a8472c87df9e9d7c8b1af0d/tests/indent/swift_spec.lua#L15

ribru17:
Also would be nice to say in the description of the PR that this change closes your issue :)

ganhaque:
This is not related to the parameter list indent issue.

I'm not familiar with the new line tests. What is the `on_line` in the new_line test for?

ribru17:
> This is not related to the parameter list indent issue.

Ah, my mistake

> I'm not familiar with the new line tests. What is the `on_line` in the new_line test for?

Basically just testing that, if you add the specified text after <on_line>, the indent is correct

PR #7312, opened 2024-11-02
chore: fix typing in parsers.lua

ribru17:
This commit gets rid of ~300 unnecessary Lua_LS warnings.


clason:
Is it worth that for the (essentially frozen) master branch?

ribru17:
Then let's not

clason:
(I would have accepted "yes", for the record.)

ribru17:
Meh I'd rather not introduce even a tiny bit more difficulty into your workflow. Warnings are just warnings :)

Issue #7308, opened 2024-11-02
bad indent in wgsl parameter list


ganhaque:
### Describe the bug

While working with .wgsl file and the parameter is spread across multiple lines, the first parameter is not indented. The subsequent parameters are indented correctly.

### To Reproduce

```wgsl
fn foo(
a: f32,
  b: f32
) {
  // ...
}
```

### Expected behavior

```wgsl
fn foo(
  a: f32,
  b: f32
) {
  // ...
}
```

### Output of `:checkhealth nvim-treesitter`

==============================================================================
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
- OK `node` found v22.7.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: cc (GCC) 14.2.1 20240805
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "6.10.6-arch1-1",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Mon, 19 Aug 2024 17:02:39 +0000"
} ~

Parser/Features         H L F I J
  - bash                ✓ ✓ ✓ . ✓
  - c                   ✓ ✓ ✓ ✓ ✓
  - cpp                 ✓ ✓ ✓ ✓ ✓
  - css                 ✓ . ✓ ✓ ✓
  - glsl                ✓ ✓ ✓ ✓ ✓
  - go                  ✓ ✓ ✓ ✓ ✓
  - html                ✓ ✓ ✓ ✓ ✓
  - javascript          ✓ ✓ ✓ ✓ ✓
  - json                ✓ ✓ ✓ ✓ .
  - lua                 ✓ ✓ ✓ ✓ ✓
  - markdown            ✓ . ✓ ✓ ✓
  - markdown_inline     ✓ . . . ✓
  - prisma              ✓ . ✓ . ✓
  - python              ✓ ✓ ✓ ✓ ✓
  - query               ✓ ✓ ✓ ✓ ✓
  - rust                ✓ ✓ ✓ ✓ ✓
[...]

PR #7319, opened 2024-11-04
fix(wgsl): indent first parameter of function definition

ribru17:
Closes https://github.com/nvim-treesitter/nvim-treesitter/issues/7308


queries/wgsl/indents.scm
```diff
@@ -1,10 +1,17 @@
...
   (struct_declaration)
 ] @indent.begin
 
+((parameter_list) @indent.begin
+  (#set! indent.immediate)
+  (#set! indent.start_at_same_line))
+
+(function_declaration
+  "(" @indent.begin
+  (#set! indent.immediate))
```
lucario387, line 13: Does it cover for things like
```
fn foo
(
   a,
   b,
   c
)
```

You may or may not cover for this, it's up to you. I feel like this is too much of an edge case to be caring about

ribru17: It will indent all the params properly in this case (no function body) but I think the `"("` indentation is not applied (so while editing users have to indent themselves). Maybe could be looked into in the future

ribru17: Actually I have just added support for this

lucario387 (approved):

Issue #7304, opened 2024-11-01
Imported identifiers starting with a capital are picked up as `@type` in JS modulles


rschristian:
### Describe the highlighting problem

It seems as though TypeScript grammar is being incorrectly applied onto JavaScript files, strangely, even when the TypeScript grammar isn't even installed (though I would like to have it installed). I'm not familiar enough with TreeSitter/nvim-treesitter to say why this is but it's rather confusing for a new user.

---

As best as I can tell, the JS grammar does not set the `@type` highlight (for obvious reasons), but the TS grammar does, and that's somehow being applied to JS files. There might be other overlaps as well, but this is the one I've found. Likely source of issue from TS:

>((identifier) @type
> (#match? @type "^[A-Z]"))

https://github.com/tree-sitter/tree-sitter-typescript/blob/73c4447796b251295b498227bad028d88dc1918b/queries/highlights.scm#L6-L7

I understand that there's limitations in what one can do with AST alone, and that while there are more preferred alternatives today (`import type { ... }` & `import { type ... }`), this is an understandable, but super iffy, guess in a TS module. However, in a JS module, this is not correct, types cannot be directly imported.

It's convention for JSX components to start with a capital letter, so that could be one alternative, as could classes or even certain constants (`const MY_VAR = ...`).

`:echo &filetype` returns `javascript` so nothing appears to be fishy with the file type detection.

### Example snippet that causes the problem

```js
import { Foo } from './foo.js';
       // ^ `:Inspect` yields `@type.javascript`
```

### Tree-sitter parsing result

```
(program ; [0, 0] - [1, 0]
  (import_statement ; [0, 0] - [0, 31]
    (import_clause ; [0, 7] - [0, 14]
      (named_imports ; [0, 7] - [0, 14]
        (import_specifier ; [0, 9] - [0, 12]
          name: (identifier)))) ; [0, 9] - [0, 12]
    source: (string ; [0, 20] - [0, 30]
      (string_fragment)))) ; [0, 21] - [0, 29]
```

### Example screenshot

![Image](https://github.com/user-attachments/assets/56201150-478d-48bb-bcc2-c1e7b6baa7dd)

(Sorry, cursor disappears -- it's over the `Foo` identifier)

### Expected behavior

That should likely be an identifier or a constant.

### Output of `:checkhealth nvim-treesitter`

```
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
[...]

PR #7322, opened 2024-11-04
refactor(ecma): Exclude all-caps constants from `@type` highlight

rschristian:
Part of #7304

It's a common convention in JS/TS/ecma to use screaming snake case (`FOO_BAR_BAZ`) for magic constants; for example, [`Number.MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) from the language spec or [`IS_NON_DIMENSIONAL`](https://github.com/preactjs/preact/blob/07dc9f324e58569ce66634aa03fe8949b4190358/src/constants.js#L15) from this ecosystem project. These are not type definitions, but do start with a capital letter, and so get incorrectly caught by the regex pattern here.

What I propose is that the regex expand to check if there's also a lower-case letter somewhere within the identifier name and if not, it won't apply the `@type` highlight to the identifier.

This seems like the most efficient regex, but I'm no expert here. Hints/advice certainly appreciated if there's a better way to do this too.

I'd like to note, however, going from an identifier to a `@type` is fatally flawed -- it cannot be done per the language spec using such little context. Scope and preceding/following tokens are needed to do it right.

### DIFF
--- a/queries/ecma/highlights.scm
+++ b/queries/ecma/highlights.scm
@@ -22,7 +22,7 @@
 ; Special identifiers
 ;--------------------
 ((identifier) @type
-  (#lua-match? @type "^[A-Z]"))
+  (#lua-match? @type "^[A-Z].+?[a-z]"))
 
 ((identifier) @constant
   (#lua-match? @constant "^_*[A-Z][A-Z%d_]*$"))


queries/ecma/highlights.scm
```diff
@@ -22,7 +22,7 @@
 ; Special identifiers
 ;--------------------
 ((identifier) @type
-  (#lua-match? @type "^[A-Z]"))
+  (#lua-match? @type "^[A-Z].+?[a-z]"))
```
ribru17, line 25: For a lua pattern with lazy quantifiers I believe this would need to use `.-`? Although it means zero or more, not one or more, so maybe the pattern would need a slight adjustment

clason:
Closing as stale.

PR #7325, opened 2024-11-05
feat(yaml): add language injection from comments

zlkn:
Added dynamic injection from comment into yaml value in order to support compatible with [intellij language injection](https://www.jetbrains.com/help/idea/using-language-injections.html#use-language-injection-comments).

This feature usable for yaml config contain inline config.

For example
- Envoy proxy configuration may contain [lua code](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter) 

- ArgoCD Application config can contain yaml [values for helm chart](https://argo-cd.readthedocs.io/en/latest/user-guide/application-specification/)

### DIFF
--- a/queries/yaml/injections.scm
+++ b/queries/yaml/injections.scm
@@ -78,3 +78,25 @@
           (block_scalar) @injection.content
           (#set! injection.language "promql")
           (#offset! @injection.content 0 1 0 0))))))
+
+; Support Idea-style injection: # language=yaml
+; Inject language at toplevel yaml node
+((comment) @injection.language
+  (#gsub! @injection.language "#%s*language=%s*([%w%p]+)%s*" "%1")
+  .
+  (block_mapping_pair
+    value: (block_node
+      (block_scalar) @injection.content
+      (#offset! @injection.content 0 1 0 0))))
+
+; Inject language in yaml body
+(block_mapping_pair
+  (comment) @injection.language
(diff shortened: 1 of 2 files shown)


clason:
Thank you! But I am not sure we want to add such non-standard extensions to the default queries. (Users or plugins can always add them via `; extends`.)

@ribru17 ?

ribru17:
Yeah I was wondering how common this would be, but I guess if IntelliJ has them like this it shouldn't be an issue. I am fine with this personally but I would prefer also if it worked with something like the following:

```yaml
    healthcheck:
      # language=sh
      test: wget --no-verbose --spider --tries=1 http://localhost:7880 || exit 1
      interval: 10s
      timeout: 5s
      retries: 5
```

zlkn:
Thank you for taking the time to consider my request!

After experimenting with Neovim's language injection support, I found some plugins that achieve similar results while supporting more languages.

From a user's perspective, I would expect this functionality to be available across all compatible languages. I'll continue working on expanding the idea to cover more languages. For now, I suggest closing this request

zlkn:
Thank you for taking the time to consider my request!

After experimenting with Neovim's language injection support, I found some plugins that achieve similar results while supporting more languages.

From a user's perspective, I would expect this functionality to be available across all compatible languages. I'll continue working on expanding the idea to cover more languages. For now, I suggest closing this request

PR #7340, opened 2024-11-08
refactor(unison): update and improve queries for latest grammar

zetashift:
Also adds indents and folds in separate commits!

### DIFF
--- a/queries/unison/highlights.scm
+++ b/queries/unison/highlights.scm
@@ -12,17 +12,15 @@
 (literal_boolean) @boolean
 
 ; Keywords
+(type_kw) @keyword.type
+
 [
   (kw_forall)
-  (unique)
-  (type_kw)
   (do)
   (kw_let)
   (ability)
   (where)
 ] @keyword
-
-(kw_equals) @keyword.operator
 
 (structural) @keyword.modifier
 
@@ -30,15 +28,16 @@
 
 (type_constructor) @constructor
 
-(doc_block) @comment.documentation
+((doc_block) @comment.documentation @spell
+  (#set! priority 90))
 
 [
   (operator)
   (pipe)
   (arrow_symbol)
-  ">"
   (or)
-] @keyword.operator
+  (kw_equals)
+] @operator
 
 [
   "if"
@@ -53,23 +52,34 @@
 
 (pattern) @variable
 
-(constructor_or_variable_pattern) @type
-
 (use_clause) @keyword.import
 
 ; Types
 (record_field
-  name: (wordy_id) @variable
-  type: (wordy_id) @type)
+  (field_name) @variable.member
+  type: (regular_identifier) @type)
 
 (type_name) @type
+
+(type_declaration
+  (regular_identifier) @type)
+
+(ability_name
+  (path)? @module
+  (regular_identifier) @type)
 
 (ability_declaration
   (ability_name) @type
   (type_argument) @variable.parameter)
 
+(constructor
+  (constructor_name) @constructor)
+
+(constructor
+  type: (regular_identifier) @type)
+
 (effect
-  (wordy_id) @attribute) ; NOTE: an effect is a special type
+  (regular_identifier) @attribute) ; NOTE: an effect is a special type
 
 ; Namespaces
 (path) @module
@@ -79,22 +89,21 @@
 ; Terms
 (type_signature
   term_name: (path) @module
-  term_name: (wordy_id) @variable)
+  term_name: (regular_identifier) @variable)
 
 (type_signature
-  term_name: (wordy_id) @variable)
+  term_name: (regular_identifier) @variable)
 
 (term_type) @type
 
-(function_application
-  function_name: (path)
-  function_name: (wordy_id) @function)
+(term_definition
+  name: (path) @module)
 
 (term_definition
-  name: (wordy_id) @variable)
+  name: (regular_identifier) @variable)
 
 (term_definition
-  param: (wordy_id) @variable.parameter)
+  param: (regular_identifier) @variable.parameter)
 
 ; Punctuation
 [
@@ -111,5 +120,6 @@
   "]"
 ] @punctuation.bracket
 
-(test_watch_expression
-  (wordy_id) @keyword.directive)
+(watch_expression) @keyword.directive
+
+(test_watch_expression) @keyword.directive
--- a/.github/workflows/update-lockfile.yml
+++ b/.github/workflows/update-lockfile.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Update parsers
         env:
-          SKIP_LOCKFILE_UPDATE_FOR_LANGS: "gnuplot,unison"
+          SKIP_LOCKFILE_UPDATE_FOR_LANGS: "gnuplot"
         run: |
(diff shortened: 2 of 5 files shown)


queries/unison/highlights.scm
```diff
@@ -12,33 +12,30 @@
...
-  (unique)
-  (type_kw)
   (do)
   (kw_let)
   (ability)
   (where)
 ] @keyword
 
-(kw_equals) @keyword.operator
+(kw_equals) @operator
```
clason, line 25: I think that should stay as it was (and similar below); operator is reserved for symbols.

queries/unison/highlights.scm
```diff
@@ -12,33 +12,30 @@
...
 
 (structural) @keyword.modifier
 
 (unique) @keyword.modifier
 
 (type_constructor) @constructor
 
-(doc_block) @comment.documentation
```
clason, line 33: This should stay as well, in case people do not want injections.

clason: If needed, set a lower priority here.

zetashift: How do I set a lower priority?

clason: With #set! priority directive (on mobile right now)

zetashift: Thank you! I found some examples and took it from there.

queries/unison/indents.scm
```diff
@@ -0,0 +1,19 @@
...
+  (cases)
+  (kw_equals)
+  (kw_else)
+  (do)
+] @indent.begin
+
+(literal_list) @indent.begin
+
+(tuple_or_parenthesized
+  "(") @indent.begin
```
lucario387, line 12: Shouldn't there also be a 
```
(tuple_or_parenthesized
  "}" @indent.branch @indent.end)
```
to go with this

zetashift: There should! However this made an indent action (e.g. pressing `o`) slower, there was a noticeable delay when that query is used as well as:
```scheme
(literal_list
  ")" @indent.branch @indent.end)
```

So I opted to include only the starting queries

queries/unison/highlights.scm
```diff
@@ -30,15 +30,15 @@
...
+((doc_block) @comment.documentation
```
clason, line 33:
```suggestion
((doc_block) @comment.documentation @spell
```

queries/unison/highlights.scm
```diff
@@ -111,5 +121,6 @@
   "]"
 ] @punctuation.bracket
 
-(test_watch_expression
-  (wordy_id) @keyword.directive)
+(watch_expression) @label
```
clason, line 124: Is this an intended change of the capture? Can you comment on that?

queries/unison/indents.scm
clason: Yes, but won't that break indentation (in some situations)? Correctness > performance

If the performance is unacceptable, it's better not to touch the construct at all.

queries/unison/highlights.scm
zetashift: It wasn't! I tried this out because I was unsure about what an watch expression exactly is wrt the mapping to nodes. But  I believe `keyword.directive` is more correct.

queries/unison/indents.scm
zetashift: In the (limited) tests I did I didn't see weird behaviour.

However you are right, I'd rather see it being correct. I've removed the queries, because the performance hit was disturbing, and adding them back in shouldn't be that hard.

clason:
Needs a rebase (and squash) now, sorry.

@ribru17 could you take a last look?

clason:
Rebase, not merge...

zetashift:
> Rebase, not merge...

Whoops habits die hard.

> Oh, @zetashift, cool that you did some stuff re the doc block Markdown! I've shied away from having the parser do anything with docblocks besides mark as one leaf because when I started this project, the API for doc blocks was changing, getting new features, and I was waiting for things to settle down there. Maybe I'll take another look now, especially since I think one can write a grammar that allows one language to be embedded in another.

@kylegoetz yeah hopefully the injections stuff make dealing with markdown easier!

queries/unison/highlights.scm
ribru17, line 1: Would you mind also moving `(kw_equals)` just to `@operator`?

queries/unison/indents.scm
ribru17, line 1: Would you mind adding some tests for this? Don't have to be very large. Also these indents can change the syntax tree as of now, e.g. running `=ip` on the following code block:

```
base58Encode ws =
  x = bigEndian ws
  base58 : Integer -> Optional (Integer, Text)
  base58 a =
    if a == 0
    then Optional.None
    else match divmod a 58 with
           (d, m) -> Optional.Some (d, charAt m codeString)
  foldLeft "" Text.concatenate (unfoldRight base58 x)
```

clason:
I think I would prefer it if we could split the indent commit out from this PR and stick to only updating the queries and grammar; that makes it easier to review, test, and merge the clear fixes quickly. I don't want to hold up updating the parser on getting the indents just right.

(In principle, this applies to folds as well, but those are simple enough to throw in anywhere. Should be left as a separate "feat" commit when rebasing, though.)

queries/unison/highlights.scm
zetashift: I don't mind! But I got the feedback from @clason that `@operator` is only for reserved keywords:
https://github.com/nvim-treesitter/nvim-treesitter/pull/7340#discussion_r1835085974

queries/unison/indents.scm
zetashift: I'll add tests(I'm not sure tho, in a follow up PR).

queries/unison/highlights.scm
ribru17: Ah, I believe that was a mistake then, sorry. Probably confusing because usually grammars expose that as e.g. `"="` rather than a named node, but here that is always a symbol (`=`) and we use `@keyword.operator` for english words that are also considered operators. Sorry about that! Hopefully the grammar uses some different node naming conventions down the line to prevent confusion

clason: Yep, my bad, sorry! @ribru17 is completely right; I should have checked the node contents before commenting!

queries/unison/highlights.scm
```diff
@@ -12,33 +12,33 @@
...
-  (unique)
-  (type_kw)
   (do)
   (kw_let)
   (ability)
   (where)
 ] @keyword
 
-(kw_equals) @keyword.operator
+(kw_equals) @operator
```
clason, line 25: Could you keep this with the `@operator` list below?

clason:
Thanks! Can you squash the last commit into the first one? (Poor `@operator` won't get pinged constantly that way -- make sure to always quote captures for that reason!)

ribru17:
They knew the risks... :boom:

zetashift:
> Thanks! Can you squash the last commit into the first one? (Poor `@operator` won't get pinged constantly that way -- make sure to always quote captures for that reason!)

Omg! Didn't mean to do that, thank you I'll clean up the history

clason:
Thank you!

PR #7344, opened 2024-11-10
feat(julia): update parser and queries

clason:
Queries synced with upstream (preserving obvious local additions, but there could be regressions; most but not all of the diff are reshuffles).

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -1,9 +1,10 @@
 ; Identifiers
 (identifier) @variable
 
-; ;; If you want type highlighting based on Julia naming conventions (this might collide with mathematical notation)
-; ((identifier) @type
-;   (match? @type "^[A-Z][^_]"))  ; exception: Highlight `A_foo` sort of identifiers as variables
+(field_expression
+  (identifier) @variable.member .)
+
+; Symbols
 (quote_expression
   ":" @string.special.symbol
   [
@@ -11,10 +12,7 @@
     (operator)
   ] @string.special.symbol)
 
-(field_expression
-  (identifier) @variable.member .)
-
-; calls
+; Function calls
 (call_expression
   (identifier) @function.call)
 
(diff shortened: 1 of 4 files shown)


queries/julia/highlights.scm
```diff
@@ -194,124 +142,134 @@
...
+  "."
+  "..."
+  "::"
+] @punctuation.special
 
 [
   ","
-  "."
   ";"
-  "::"
```
ribru17, line 309: I feel like this should be left as `@punctuation.delimiter`, looks like most other languages use that highlight for this double colon thing. Unless there was a specific reason

ribru17 (approved):

ribru17:
Generating the parser gives me several warnings (and takes a long time) btw:

```
Warning: rule _pair_operator is just a `seq` or `choice` rule with a single element. This is unnecessary.
Warning: rule _lazy_or_operator is just a `seq` or `choice` rule with a single element. This is unnecessary.
Warning: rule _lazy_and_operator is just a `seq` or `choice` rule with a single element. This is unnecessary.
Warning: rule _pipe_right_operator is just a `seq` or `choice` rule with a single element. This is unnecessary.
Warning: rule _pipe_left_operator is just a `seq` or `choice` rule with a single element. This is unnecessary.
Warning: rule _rational_operator is just a `seq` or `choice` rule with a single element. This is unnecessary.
Warning: rule _tilde_operator is just a `seq` or `choice` rule with a single element. This is unnecessary.
```

will make a PR

clason:
> Generating the parser gives me several warnings (and takes a long time) btw:

Yes, I know.

> will make a PR. Also the built parser is about 6 mb with many states 😬

Yes, I know, but that was the case already before the update. The problem (besides over-emphasis on reproducing the official syntax) is the huge list of Unicode symbols in the grammar.

queries/julia/highlights.scm
clason: Upstream just doesn't like our queries. 🤷

PR #7345, opened 2024-11-11
fix: formatter idempotency for nested groupings

ribru17:
Currently, for something like `(((((node))))) @cap`, the formatter will produce `((node)) @cap`, and then after another pass it will produce `(node) @cap`. This commit makes it so that all extraneous parentheses are removed on the first pass.

### DIFF
--- a/scripts/format-queries.lua
+++ b/scripts/format-queries.lua
@@ -248,6 +248,7 @@ local format_queries = [[
   "(" @format.remove
   .
   [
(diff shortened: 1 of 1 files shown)


lucario387 (approved):
Thanks. I'm pretty sure I didn't left that in because I didn't know how to handle it with my formatter code of that time. Now it seems like the issue is resolved.

Issue #7326, opened 2024-11-05
Trailing forward slash in jsdoc delimiter


rschristian:
### Describe the highlighting problem

The trailing forward slash in a JSDoc comment, be it single or multi-line comment, has `@punctuation.delimiter` applied which I'm not sure is correct. At the very least, it is inconsistent as the leading forward slash does not receive it.

Looks to be caused by #6864

### Example snippet that causes the problem

```js
/** this is a comment */
```

### Tree-sitter parsing result

```
(program ; [0, 0] - [1, 0] javascript
  (comment ; [0, 0] - [0, 24] javascript
    (source ; [0, 0] - [0, 24] comment
      "text" ; [0, 0] - [0, 1] comment
        "/") ; [0, 0] - [0, 1] comment
      "text" ; [0, 1] - [0, 3] comment
      "text" ; [0, 4] - [0, 8] comment
      "text" ; [0, 9] - [0, 11] comment
      "text" ; [0, 12] - [0, 13] comment
      "text" ; [0, 14] - [0, 21] comment
      "text" ; [0, 22] - [0, 23] comment
      "text" ; [0, 23] - [0, 24] comment
        "/")))) ; [0, 23] - [0, 24] comment
```

### Example screenshot

Leading Slash            |  Trailing Slash
:-------------------------:|:-------------------------:
![leading slash](https://github.com/user-attachments/assets/453b4939-ce52-4a3e-9430-dbf59774721f)  | ![trailing slash](https://github.com/user-attachments/assets/f1b98408-d2eb-49c1-9f9e-d540bc6d1f3d)

### Expected behavior

_I don't think_ that should count as a delimiter, so removing that highlight would be an option, else, also applying it to the leading slash would equally be reasonable.

### Output of `:checkhealth nvim-treesitter`

```
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- WARNING `tree-sitter` executable not found (parser generator, only needed for :TSInstallFromGrammar, not required for :TSInstall)
- OK `node` found v20.18.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: cc (GCC) 14.2.1 20240910
- OK Neovim was compiled with tree-sitter runtime ABI version 14 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
[...]

PR #7351, opened 2024-11-13
refactor(jsdoc): Remove delimiter highlights

rschristian:
As best as I can tell, none of these delimiter highlights work. I don't quite understand why, especially as the bracket highlights have no problem at all, but I haven't been able to get them to be picked up except where they shouldn't be (#7326).

```
/**
 * @param {string} [bar]
 *                 ^ Correctly picked up as @punctuation.bracket
 * @return {string}
 *         ^ Also picked up correct as @punctuation.bracket
 *
 *
 * None of the following, however, are picked up as @punctuation.delimiter
 * : / . # ~
 */
  ^ This is picked up as @punctuation.delimiter but it shouldn't be
```

Not quite sure if this wasn't tested, if something changed, or what, but given the circumstances this seemed like a reasonable suggestion?

### DIFF
--- a/queries/jsdoc/highlights.scm
+++ b/queries/jsdoc/highlights.scm
@@ -9,14 +9,6 @@
   "]"
 ] @punctuation.bracket
 
-[
-  ":"
-  "/"
-  "."
-  "#"
-  "~"
-] @punctuation.delimiter
-
 (identifier) @variable @nospell
 
 (tag


lucario387:
This is not the correct fix. It should be applying offset for the injection query in https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/ecma/injections.scm#L1-L3

rschristian:
I'm sorry, but I don't understand what that means. Not a regular contributor, quite unfamiliar with everything about treesitter yet.

If your intended fix would have the delimiters show up, and the last trailing forward slash would no longer be picked up as a delimiter, I'll happily close this out for someone else to contribute.

ribru17:
See `:h treesitter-directive-offset!`, this will prevent trailing slash from getting picked up

rschristian:
Is #7326 wrong then? They seemed pretty convinced it was an issue with the parsing.

rschristian:
That might be so, but as I tried to outline, none of the delimiters are picked up at the moment. The trailing slash is the only one that "works", and it works incorrectly. As such, it seems like the better move is to remove them entirely? Unless someone wanted to fix them all.

https://github.com/nvim-treesitter/nvim-treesitter/issues/7326#issuecomment-2457042752 also seemed pretty convinced nothing could be done about the parsing here?

ribru17:
If none of the other delimiters work, it would take a PR to the jsdoc parser. Do you have a code example of something that doesnt work? I know you outlined delimiters but I am not familiar enough with jsdoc to know how they would be used

ribru17:
In any case I don't think highlights should be removed, slash delimiters _will_ work correctly when the offset is applied, and once the parser is updated to expose the other delimiters we can highlight them properly as well

rschristian:
> Do you have a code example of something that doesnt work? I know you outlined delimiters but I am not familiar enough with jsdoc to know how they would be used

The PR description contains a comment that you can copy/paste showing none of the delimiters are picked up in the comment body but that's part of the issue I suppose -- there are no defined delimiters when it comes to jsdoc. The [jsdoc reference pages](https://jsdoc.app/) don't use any of those symbols beyond what you'd find in standard comment text and they are not highlighted any differently when they are used. The [first example block here](https://jsdoc.app/tags-inline-link#examples) contains `#`, `:`, `/`, and `.`, yet they receive the same "comment" highlight from hljs. They are not treated differently from standard comment text.

However, if you're not familiar with jsdoc then I'm a bit confused as to why [you added these highlights](https://github.com/nvim-treesitter/nvim-treesitter/pull/6864) in the first place? Honest question, not trying to accuse or anything, I just don't see any justification or reason for the change. I assumed it was a rather blind or personal addition? I've never seen any other highlighter match this.

It's fine if this is something for me to maintain on my end, I'm happy enough to keep my own corrected highlights, just feels like there's been a few odd decisions to get to this point.

ribru17:
Ah, I understand. Ok so the delimiters you posted above will not (always) apply since they are only exposed as actual nodes in member expressions of the grammar:  https://github.com/tree-sitter/tree-sitter-jsdoc/blob/b253abf68a73217b7a52c0ec254f4b6a7bb86665/grammar.js#L127 

This is intentional since (according to the grammar) they only have semantic meaning here. We also only should consider `/` to have semantic meaning in a path expression, also detected by the grammar. Unfortunately the grammar also exposes `/` in the start and end blocks, so we need to only apply queries to `/` characters within a path expression. I will make a PR for this

As for the other delimiters, if you see them in a true member expression they will be given proper highlights as intended. If you see these delimiters in a random text block they will not be highlighted (since they have no semantic meaning there) and this is also intended. I will remove the extra `/` highlights from the start/end now. Is this a valid change? Hopefully I did not miss anything, I think I understand your point more now

rschristian:
> As for the other delimiters, if you see them in a true member expression they will be given proper highlights as intended. If you see these delimiters in a random text block they will not be highlighted (since they have no semantic meaning there) and this is also intended. I will remove the extra `/` highlights from the start/end now.

Sorry, can you clarify what you mean? `/`, at the moment, **will** be picked up in comment text blocks, this isn't limited to just the start/end of the comment. The rest of the symbols, however, are not.

```js
/** @type {string} / */
                // ^  ^ Both of these are picked up as @punctuation.delimiter.jsdoc
/** @type {string} # . */
                // ^ ^ Both of these are ignored
```

With potential exceptions to the start/end nodes (though they don't appear to be highlighted separately today), these all should, to my knowledge, be highlighted the same. However, `/` seems to be the only one every picked up by that highlight (and only following the start node at that), I cannot find a situation in which the others will be. Maybe I'm missing something, but from typical JSDoc use, I can't think of any situation in which you could or would use these as anything other than standard comment text, they shouldn't ever have semantic meaning.

---

Thanks for bearing with me here, trying to learn my way around but still very new to this.

ribru17:
In the first example, that is due to a parsing error... another thing that requires parser PR for robustness :) But as for the other delimiters, try this example and you will find they are highlighted:

/**
 * @alias Apple.Core
 * @event module:foo/bar.event:MyEvent
 * @fires Hurl#snowball
 * @exports hello/world
 */

similarly, if we apply the change I was mentioning of only highlighting slashes when inside a path expression, even the above error will be handled (at least from the highlight side, it will still produce an erroneous tree which should be fixed eventually)

ribru17:
Of course, thanks for the thorough explanations :)

PR #7363, opened 2024-11-15
fix(zig): begin indent at `struct_declaration`

ofseed:
The curly braces contained in the `struct_declaration` are not included by the `block`. Indentation should also start here.

## Example:
image

Before this PR:
image

After this PR:
image


clason:
Before/after examples go a long way ;)

ofseed:
Edited with an example attached :)

PR #7382, opened 2024-11-21
fix(nu): add scanner.c

mrdgo:
Relevant, as soon as https://github.com/nushell/tree-sitter-nu/pull/117 is merged

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1472,7 +1472,7 @@ list.nqc = {
 list.nu = {
   install_info = {
     url = "https://github.com/nushell/tree-sitter-nu",
-    files = { "src/parser.c" },
+    files = { "src/parser.c", "src/scanner.c" },
   },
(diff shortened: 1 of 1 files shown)


clason:
Thank you for the heads up; you will need to combine this change with the lockfile update whenever that PR is landed.

Since that doesn't look to be happening very soon, I'll take the liberty of closing this PR; I'll just include the change in the automated lockfile update when it happens, now that I know about.

mrdgo:
So we don't actually break things whenever we make breaking changes to the parser. Thanks, I was worried about that :)

clason:
No, the automated update is only merged if CI passes, which means no breaking changes. Otherwise I have to manually either fix things or drop the update (so breaking changes are not completely "free").

Adding a scanner is pretty trivial to adjust for; the really tricky changes are those that drop nodes or change how patterns work, then I either have to rely on bundled queries (in the parser repo) to see what changes are required, or rely on the query maintainer to (eventually) make a PR and unblock the updates.

PR #7384, opened 2024-11-21
Additions to Robot highlights

Hubro:
Related issue: https://github.com/Hubro/tree-sitter-robot/issues/8

### DIFF
--- a/queries/robot/highlights.scm
+++ b/queries/robot/highlights.scm
@@ -65,6 +65,10 @@
   (else_statement
     "ELSE" @keyword.conditional))
 
+(inline_if_statement
+  (inline_else_statement
+    "ELSE" @keyword.conditional))
+
 [
   "TRY"
   "EXCEPT"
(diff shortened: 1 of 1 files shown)


queries/robot/highlights.scm
```diff
@@ -77,3 +81,50 @@
...
     "ELSE" @keyword.exception))
+
+; Highlight text chunks in "Documentation" settings with @spell
+
+(setting_statement
+  name: (setting_name) @_name
+  (arguments
+    (argument
+      (text_chunk) @spell))
+  (#match? @_name "^Documentation$"))
```
clason, line 92: For performance reasons, `#match?` should not be used in our queries. You should replace this with `#eq?`.

clason: And `@spell` does not highlight anything. Don't you want to add `@string.documentation` as well for actual highlights?

clason:
Run `./scripts/format-queries.lua queries/robot` to appease the linter.

queries/robot/highlights.scm
Hubro: Setting names in Robot are case insensitive, so I was hoping there was a way to match "documentation" case insensitively with `match`. Is that possible?

I know `@spell` doesn't highlight anything, but as far as I understood, it should still be put in the highlight query file (?). The issue I'm addressing is marking documentation keywords for spell checking. I didn't know about `@string.documentation`, it makes sense to add that too.

clason: > Is that possible?

Yes, but not like this, and then there are still better ways. do you expect you need to support the full `DOcumenTATION` combinatorial explosion? If not, use `#any-of?` for the reasonable variants; otherwise use `#lua-match?` with a `[dD][oO]...` Lua pattern.

clason: > I know @spell doesn't highlight anything, but as far as I understood, it should still be put in the highlight query file (?). The issue I'm addressing is marking documentation keywords for spell checking. 

Yes, that is correct. I was just replying to the comment, which explicitly mentions "highlighting".

clason: Pro-tip: always back-tick captures in commit messages; otherwise the poor user named spell will get pinged on any force-push.

Hubro: > > Is that possible?
> 
> Yes, but not like this, and then there are still better ways. do you expect you need to support the full `DOcumenTATION` combinatorial explosion? If not, use `#any-of?` for the reasonable variants; otherwise use `#lua-match?` with a `[dD][oO]...` Lua pattern.

I suppose it's not really reasonable for anybody to write Documentation like that, so I'm settling for `#any-of?`.

clason: Can always revisit if someone comes up with a reasonable request. The main point is to avoid `#match?` like the plague (because it spins up Vim's own regex engine) and to prefer literal matches over patterns wherever possible.

queries/robot/highlights.scm
```diff
@@ -77,3 +81,49 @@
...
   (else_statement
     "ELSE" @keyword.exception))
+
+; Extra captures for "Documentation" settings
+(setting_statement
+  name: (setting_name) @_name
+  (arguments
+    (argument
+      (text_chunk) @spell @string.documentation))
+  (#any-of? @_name "Documentation" "documentation" "DOCUMENTATION"))
```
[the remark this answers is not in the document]
Hubro: I don't think they can be condensed to 2 queries, at least not to my limited knowledge of Tree-sitter queries. There are 3 different outer tag names (`setting_statement`, `keyword_setting`, `test_case_setting`).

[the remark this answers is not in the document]
clason: But maybe 3 or 4?

clason:
@Hubro If there's any way to reduce the number of patterns, that would be worth it.

Otherwise please rebase and squash your commits and I'll merge.

Hubro:
> @Hubro If there's _any_ way to reduce the number of patterns (even just by 2), that would be worth it.
> 
> Otherwise please rebase and squash your commits and I'll merge.

Since there are 3 different parent nodes and the target nodes are at 2 different depths, I'm unaware of any way to use less than 6 queries.

I've rebased and squashed.

PR #7387, opened 2024-11-22
fix: only highlight JSDoc slashes in paths

ribru17:
Otherwise the final slash in a doc comment `*/` will also be highlighted.

Supersedes #7351


rschristian:
Patched this in quickly and it looks great! Fixes both issues regarding `/` (comment text & trailing).

Thanks!

PR #7390, opened 2024-11-22
fix(julia): fix injection queries

fredrikekre:
This patch fixes the julia `(string_literal)` injection queries after
the breaking changes in
https://github.com/tree-sitter/tree-sitter-julia/pull/153. The queries
are simplified by the fact that string content is now directly available
as a separate `(content)` child node.


fredrikekre:
Thanks. Just realized that this also fixes docstring injection for single-line docstrings like
```julia
"this is **docs**"
function foo()
end
```

PR #7392, opened 2024-11-22
fix(julia): treat :: as operator

fredrikekre:
This patch removes `::` from `@punctuation.delimiter` and instead treat
it as `@operator` within `(unary_typed_expression)` and
`(typed_expression)`. I believe these are the only places where `::` can
show up, so maybe it is simpler to just capture it as `@operator`
globally `::`. (Note that `<:` and `>:` already are `@operator`s
globally since they are parsed into the `(operator)` node.)

Edit: I split out the non-controversial parts into #7396

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -281,6 +281,14 @@
   "::"
 ] @punctuation.delimiter
 
+; Treat `::` as operator in type contexts, see
+; https://github.com/nvim-treesitter/nvim-treesitter/pull/7392
+(typed_expression
+  "::" @operator)
+
+(unary_typed_expression
+  "::" @operator)
+
 [
   "("
   ")"
--- a/tests/query/highlights/julia/test.jl
+++ b/tests/query/highlights/julia/test.jl
@@ -2,13 +2,13 @@ function load_data(::Symbol; ::Int) :: Tuple
 # <- @keyword.function
 #        ^ @function.call
 #                 ^ @punctuation.bracket
-#                  ^^ @punctuation.delimiter
+#                  ^^ @operator
 #                    ^ @type.builtin
 #                          ^ @punctuation.delimiter
-#                            ^^ @punctuation.delimiter
+#                            ^^ @operator
 #                              ^^^ @type.builtin
 #                                 ^ @punctuation.bracket
-#                                   ^^ @punctuation.delimiter
+#                                   ^^ @operator
 #                                      ^ @type.builtin
     dataset = CIFAR10(; Tx = Float32, split = split)
 #   ^^^^^^^ @variable


clason:
(Test failure is related, and please run `./scripts/format-queries.lua queries/julia`.)

fredrikekre:
I see. Personally I find it weird that `::`, `<:`, and `::` are not treated equally since they are so tightly coupled, and by Julias parsers they are all operators).

clason:
Well, the "semantic" (yes, yes, I know) meaning is relevant here so if there's a strong argument that Julia `::` is not like the other `::`, that would be fine. We'd just need to document this somewhere explicitly so this doesn't get changed (again)  next time someone touches those queries.

fredrikekre:
Btw, can I ask for some clarification on how https://github.com/tree-sitter/tree-sitter-julia/blob/master/queries/highlights.scm and https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/julia/highlights.scm interact, it seems like they are merged?

clason:
They don't interact at all. I manually adapted to the breaking parser changes (since nobody else seemed to care) by comparing the upstream queries to infer the necessary changes.

clason:
Can you add (to the commit message) some reference why these are operators in this context? (Again, this would help guard against regressions. We don't require that normally, but since there's already been some back-and-forth about this, I feel it would be worth it in this case.)

And how sure are you that `::` can never occur anywhere else? If there's doubt, I'd keep the "global" highlight of `::` as `@punctuation.delimiter` but add the specific patterns _after_ that so they have precedence.

And the tests need to be adapted.

fredrikekre:
Rewrote the commit message. The only use of `::` in julia is in type contexts, but keeping it as punctuation as fallback doesn't hurt.

fredrikekre:
Looks like the fallback takes priority, should I move the punctuation queries up altogether or just the `::` query?

clason:
> Looks like the fallback takes priority, should I move the punctuation queries up altogether or just the `::` query?

Yes, that's what I was talking about :) I think it's better to move the specific queries (right) afterwards to make clear this is a specific override (you can add a comment for extra credit, if you want); less churn that way.

fredrikekre:
Ah, I missed that, will do. Why do tests pass though?

clason:
That's a good question I don't have an answer to, unfortunately.

fredrikekre:
Alright, well, pushed the change and confirmed it works locally.

fredrikekre:
Thanks for the quick reviews/merges. After https://github.com/nvim-treesitter/nvim-treesitter/pull/7390, https://github.com/nvim-treesitter/nvim-treesitter/pull/7391, https://github.com/nvim-treesitter/nvim-treesitter/pull/7392, https://github.com/nvim-treesitter/nvim-treesitter/pull/7395 and https://github.com/nvim-treesitter/nvim-treesitter/pull/7396 I don't see any more highlight issues in the julia code bases I have scrolled through :tada:

clason:
Thank you for the PRs! If you want, I can put you down as maintainer of the queries here :)

clason:
(Now for fixing the parser itself; the performance issues are rough...)

fredrikekre:
> I can put you down as maintainer of the queries here :)

Sure, I have no experience with the parser though.

clason:
It's only about the queries in this repo (although help with adaptation to breaking parser changes is much appreciated); the parser is a completely separate project we don't have any connection with.

goerz:
Just to piggy-back on this and https://github.com/nvim-treesitter/nvim-treesitter/pull/7392#issuecomment-2494155184 in particular, I still would like to clarify the maintenance of the Julia support in `nvim-treesitter` relative to the upstream [tree-sitter-julia](https://github.com/tree-sitter/tree-sitter-julia) repo.

* If someone makes changes to the parser in `tree-sitter-julia`, how does that change get incorporated into `nvim-treesitter`? It seems like the [lockfile.json](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lockfile.json) get updated by a bot. When does that happen? For every commit on `tree-sitter-julia`, or just when `tree-siter-julia` tags a release?
* Is there a reason for the [`queries` in `tree-sitter-julia`](https://github.com/tree-sitter/tree-sitter-julia/tree/master/queries) (and an upstream repo in general) not to be identical with the [`queries` in `nvim-treesitter`](https://github.com/nvim-treesitter/nvim-treesitter/tree/master/queries/julia)? If a bot updates the link to the parser, it feels like it should copy the queries from the upstream repo as well. Otherwise, everything relies on a vigilant maintainer, i.e., @fredrikekre now, since @savq [doesn't want to maintain the `nvim-treesitter` queries going forward](https://github.com/tree-sitter/tree-sitter-julia/pull/153#issuecomment-2460843833). I suppose queries are editor-specific, while the parser is general, but then why does `tree-sitter-julia` and upstream repos in general contain queries at all? How do the other supported languages handle this?
* If the system relies on manually coordinating updates between `nvim-treesitter` and upstream repos, that seems to me like it only really works if the maintainer for the queries in `nvim-treesitter` is also the maintainer of the upstream repo. For `tree-sitter-julia`, it seems that @savq still wants to maintain the parsers and the queries there. But shouldn't @fredrikekre be at least a co-maintainer of `tree-sitter-julia` as well, if he is going to take on the responsibility of keeping the queries in sync between the two repositories?
* I might be interested in contributing some work on the queries in the future, like my [fix for the markdown injection into docstrings that start with `@doc raw"""`](https://github.com/goerz/neovimrc/commit/483721f0a387ca9df492adc97eb5fc4175a3634d) (which will have to be modified to work with the revised version of the parser in the current `master`). Where should I make PRs for that, and who should I ping as a maintainer to look at them?

It might be good to add some information in [CONTRIBUTING.md](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md) to explain how parsers and queries are maintained between `nvim-treesitter` and upstream repos, whether the queries should be identical, etc.

clason:
All this is off-topic here, and discussions should happen in the nvim-treesitter Matrix room. But briefly:

It's simple: Queries for Neovim are maintained here **and only here**. (Parsers can have their own queries for whatever reason, and that is useful background information, but they can't be used as-is in general.) The reason is that queries are specific to the editor, and we strive for consistency across languages to make it possible for colorschemes to work well. (Please extend us the courtesy to assume that we know what we are doing, and have discussed all these issues -- ad nauseam -- already.)

And the maintainer is the query maintainer, end of story. We have no control over parsers, so those you have to talk to directly.

The bot runs daily (but I'll reduce that frequency in the near future).

And you make PRs to whatever project you use.

PR #7393, opened 2024-11-22
fix(astro): highlight self-closing component tags

net-solution:
For Astro, adds highlighting to self-closing custom component tags, e.g `<Navigation />`.

Before this would be highlighted like a regular HTML tag, unless it was not self closing.


clason:
@virchau13

PR #7395, opened 2024-11-22
fix(julia): fix macro identifier highlight query

fredrikekre:
This fixes the macro identifier highlight query after the breaking
changes in https://github.com/tree-sitter/tree-sitter-julia/pull/153.
See also
https://github.com/tree-sitter/tree-sitter-julia/pull/158/commits/9f2fb3b9a8c73ef0d2ec005c97f91274b492d48d.

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -34,7 +34,9 @@
   (#any-of? @_pipe "|>" ".|>"))
 
 ; Macros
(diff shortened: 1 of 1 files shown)


clason:
Can you show a before/after example? I don't see how that change would make a difference.

fredrikekre:
Before:
image

After:
image

PR #7414, opened 2024-11-28
feat(inko): highlight "inline" keyword

yorickpeterse:
The parser added support for the "inline" keyword in a few places. This ensures a highlight is present for the keyword.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -354,7 +354,7 @@
     "revision": "962568c9efa71d25720ab42c5d36e222626ef3a6"
   },
   "inko": {
-    "revision": "aecabede39b0db05678e2d4686258d4f71b00a51"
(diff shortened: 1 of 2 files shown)


lockfile.json
```diff
@@ -354,7 +354,7 @@
     "revision": "962568c9efa71d25720ab42c5d36e222626ef3a6"
   },
   "inko": {
-    "revision": "aecabede39b0db05678e2d4686258d4f71b00a51"
+    "revision": "a6fb0b0bfee5dede688968fd91282da799e5aa8e"
```
yorickpeterse, line 357: If preferred, I can just wait until the bot updates the lockfile and then rebase this PR.

clason: This is fine, thank you (keeps changes atomic). I'm trying to move away from the constant update churn towards more meaningful updates.

PR #7415, opened 2024-11-28
feat(julia): highlight interpolation operator

fredrikekre:
This patch adds `$` to the `@punctuation.special` capture group. `$` is
used for string interpolation (`"hello $name"`) and expression
interpolation (`:(hello $name)`) and for both of these cases the
`@punctuation.special` category seem appropriate.

### DIFF
--- a/queries/julia/highlights.scm
+++ b/queries/julia/highlights.scm
@@ -298,6 +298,15 @@
   "}"
 ] @punctuation.bracket
 
+; Interpolation
+(string_interpolation
+  .
+  "$" @punctuation.special)
+
+(interpolation_expression
+  .
+  "$" @punctuation.special)
+
 ; Keyword operators
 ((operator) @keyword.operator
   (#any-of? @keyword.operator "in" "isa"))


clason:
Just to make sure: That is the only context where this anonymous node can appear?

fredrikekre:
I thought so, but actually it can be used as an infix binary operator:
```
julia> ($)(a, b) = a + b
$ (generic function with 1 method)

julia> 1 $ 2
3

julia> ($)(1, 3)
4
```
Limited this patch to the relevant interpolation nodes. Note that the parser errors for these examples (https://github.com/tree-sitter/tree-sitter-julia/issues/161).

clason:
Makes sense. As a general note, we need to consider not only correctness but also performance; such patterns are not exactly free, and Julia is already one of the slowest languages in the bunch (`./scripts/check-queries.lua` prints timings). To be fair, most of this is due to parser complexity, but I definitely see usability issues (lag when opening Julia files, which is something people tend to complain about in other languages, even where I don't see any).

PR #7418, opened 2024-11-29
fix(rust) + feat(rust): update queries and add tags

flippette:
I've replaced `queries/rust/highlights.scm` with the one from upstream `tree-sitter-rust`, this one supports raw references and `gen` blocks.

I'm unsure about `queries/rust/injections.scm` though, feedback is welcome.


flippette:
I'm just now aware that the queries are supposed to be different, I'll open a different PR.

clason:
And tags queries are not used in Neovim, see https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md.

PR #7419, opened 2024-11-29
feat(rust): add gen/yield syntax and raw references

flippette:


amaanq (approved):

ribru17:
Can you remove `"yield"` from `@keyword.return` if you feel this capture is better suited

flippette:
`yield` is more like `return`, now that I think about it... I'll adjust the queries.

Issue #3837, opened 2022-11-20
improve documentation / tooling for contributions


kevinushey:
As part of trying to update the R tree-sitter grammar, I had trouble figuring out how to replicate failing CI tests locally (as the grammar changes would require some changes to the queries bundled in this repository).

It would be helpful if there were some of the following:

1. Helper scripts to make it easier to bootstrap the requisite testing environment;
2. Documentation as to how one could prepare their local environment.

There are some helper scripts in (1), but they're not documented and seem somewhat tied to pre-requisites that only happen in CI. See e.g.

https://github.com/nvim-treesitter/nvim-treesitter/blob/master/.github/workflows/tests.yml#L33-L46

Some things that would be helpful to see:

- What version of neovim should I install for running tests? How should I install it?
- What neovim plugins are required to run tests? (e.g. https://github.com/kevinushey/nvim-treesitter/blob/6bf1950ada77bd411163ca071a46bb3552c3e1fa/scripts/minimal_init.lua#L2 implies plenary.nvim is required; how should we install it?)
- While I'm looking there; should https://github.com/kevinushey/nvim-treesitter/blob/6bf1950ada77bd411163ca071a46bb3552c3e1fa/scripts/minimal_init.lua#L1 prepend to the runtime path rather than append? (so the queries etc. defined in the repository take precedence over other things that might be on runtimepath)
- How should I run a specific suite of tests? How does this differ for highlight vs. indentation tests?

I'm guessing most of this is obvious for someone familiar with the neovim ecosystem, but for those of us who are not such instructions would be incredibly helpful.

clason:
These are valid points, thank you for raising them!

PR #7425, opened 2024-12-02
feat(slim): add parser, queries and tests

theoo:


queries/slim/injections.scm
```diff
@@ -0,0 +1,44 @@
...
+ (#set! injection.language "less")
+ (#set! injection.combined))
+
+((scss_lang) @injection.content
+ (#set! injection.language "scss")
+ (#set! injection.combined))
+
+((coffee_lang) @injection.content
+ (#set! injection.language "coffee")
+ (#set! injection.combined))
```
clason, line 34:
```suggestion
```

We don't have a parser for coffeescript.

clason (changes requested):
Please read CONTRIBUTING.md; you are only(!) allowed to use captures listed there. Please use the most specific matching capture (prefer `@keyword.import` over just `@keyword` etc.)

clason:
You can (and need to) run `./scripts/format-queries.lua queries/slim`.

queries/slim/highlights.scm
```diff
@@ -0,0 +1,64 @@
...
+  (tag_name) @markup.italic
+  (#any-of? @markup.italic "i" "em")
+)
+(
+  (tag_name) @markup.link
+  (#any-of? @markup.link "a")
+)
+(
+  (tag_name) @markup.heading
+  (#any-of? @markup.heading "header" "h1" "h2" "h3" "h4" "h5" "h6")
```
ribru17, line 59: We actually have more specific headings for each of these (e.g. `@markup.heading.3`)

queries/slim/injections.scm
```diff
@@ -0,0 +1,44 @@
...
+
+((markdown_lang) @injection.content
+ (#set! injection.language "markdown"))
+
+((textile_lang) @injection.content
```
ribru17, line 39: We also don't have `textile` (yet? :))

queries/slim/highlights.scm
```diff
@@ -0,0 +1,64 @@
...
+ ] @string
+
+(bracket) @punctuation.bracket
+(delimiter) @tag.delimiter
+(parameter_name) @variable.parameter
+(parameter_value) @tag.attribute
+
+[
+ (text_inline)
+ (text_block)
```
ribru17, line 24: Maybe this one should be `@markup.raw.block`?

queries/slim/highlights.scm
```diff
@@ -0,0 +1,64 @@
+; Inspired by https://github.com/tree-sitter/tree-sitter/discussions/2025
+; and https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights
+
+(comment) @comment
```
ribru17, line 4:
```suggestion
(comment) @comment @spell
```

queries/slim/injections.scm
ribru17, line 1: Please also add comment injections

theoo:
Thanks for your feedback and patience. I tried to run `scripts/run_tests.sh` without success. The _minimal_init.lua_ config file has reference to a missing file. _plugin/plenary.vim_ , therefore the script exists with something like `PlenaryBustedDirectory is not an editor command`.

An existing issue might be related: https://github.com/nvim-treesitter/nvim-treesitter/issues/3837

I wasn't able to fix it. Should I open a ticket for that?

clason:
No, you need to install plenary to run the tests.

https://github.com/nvim-lua/plenary.nvim

theoo:
It's installed through _my setup_ but the script is loading an alternative config/init file:
```bash
run() {
    nvim --headless --noplugin -u scripts/minimal_init.lua \
        -c "PlenaryBustedDirectory $1 { minimal_init = './scripts/minimal_init.lua' }"
}
```
And the minimal_init.lua file references a missing file, it seems:
```lua
vim.opt.runtimepath:append "."
vim.cmd.runtime { "plugin/plenary.vim", bang = true } -- there is no such file
vim.cmd.runtime { "plugin/nvim-treesitter.lua", bang = true }
```
So unless I'm missing something important about neovim paths, I won't be able to load plenary that way, when installed with Lazy.

Anyways, I will just remove the highlight test. It works just fine with `tree-sitter test` but the comments structure is odd with slim, so I can see where it breaks in the _highlights_spec.lua_ file.

clason:
>  when installed with Lazy.

Yep, that's the issue.

ribru17:
Pretty sure you need to set `ft=slim` with a modeline in the injection test files, otherwise tests wont pass

clason:
@theoo friendly ping. (I don't think the tests are actually valuable enough here, anyway.)

queries/slim/highlights.scm
```diff
@@ -0,0 +1,79 @@
+; Inspired by https://github.com/tree-sitter/tree-sitter/discussions/2025
+; and https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights
```
clason, line 2:
```suggestion
```
Not needed here.

theoo:
TLDR; I have a tab (1 of 8 :-) only) in my browser for that and I will fix it ASAP, probably in the coming days.

Sorry for the delay, and thank you for the patience. Something came up.

I'm not able to run the test locally due to how Plenary is loaded in my config (Lazy), and I didn't take the time to write an alternative minimal neovim config just for to run test. It's obiously not making sense to continue pushing arbitrary changes in the hope test wil pass in the pipeline, I must run tests locally before submitting the next changes.

clason:
You can just remove the tests; they are not a hard requirement.

theoo:
I have not been able to run tests locally, so I removed injections tests (cannot debug here because I don't even understand the error and suggested to do so).

Setup: Archlinux, Neovim v0.10.3, latest Plenary (3707cdb) loaded without plugins manager.

Issues: 
- `highlight-assertions` fails to compile
- `vim/_editor.lua:0: command line..script nvim_exec2() called at command line:0..BufEnter Autocommands for "<buffer=1>": Vim(append):Error executing lua callback: /usr/share/nvim/runtime/lua/vim/treesitter/language.lua:107: no parser for 'scheme' language`

I was wondering if something speaks agains packing tests in docker? I didn't find a list of dependencies, but I can help writing the Dockerfile if needed.

ribru17:
Local runs are indeed a mess; that said, it looks like tests would have failed because:

1. languages need an `@` (e.g. `^ @javascript`)
2. since vim has no slim filetype, every test injection file would need a modeline like `// vim: ft=slim` or equivalent

clason:
And thanks for the offer, but for 1.0 (`main` branch WIP) we plan to rework the tests to avoid the private crate (and plenary) anyway.

README.md
```diff
@@ -408,6 +408,7 @@ We are looking for maintainers to add more parsers and to write query files for
 - [x] [scss](https://github.com/serenadeai/tree-sitter-scss) (maintained by @elianiva)
 - [x] [sflog](https://github.com/aheber/tree-sitter-sfapex) (maintained by @aheber, @xixiaofinland)
 - [x] [slang](https://github.com/theHamsta/tree-sitter-slang) (experimental, maintained by @theHamsta)
+- [x] [slim](https://gitlab.com/theoreichel/tree-sitter-slim) (maintained by @theoreichel)
```
clason, line 411: This file is autogenerated, so it should match the parser information.

lua/nvim-treesitter/parsers.lua
```diff
@@ -2042,6 +2042,14 @@ list.slang = {
   experimental = true,
 }
 
+list.slim = {
+  install_info = {
+    url = "https://gitlab.com/theoreichel/tree-sitter-slim",
+    files = { "src/parser.c", "src/scanner.c" },
+  },
+  maintainers = { "@theoo", "@theoreichel" },
```
clason, line 2050: That's the same person, right, just github and gitlab user names? If so, please remove the Gitlab username; the point of this information is to record who should be pinged on issues and PRs related to these queries(!), and that would obviously not work across code forges.

README.md
theoo: Changes removed as the file is generated.

theoo:
@ribru17 Thank you for the explanation. I'll be happy to provide injection tests (or any useful tests for that matter). @clason mentioned that they are not mandatory so should I open a new PR once this one is (eventually) approved, or would you rather have them in this very PR?

I feel uncomfortable to do this if I cannot run tests locally.

lua/nvim-treesitter/parsers.lua
```diff
@@ -2044,10 +2044,11 @@ list.slang = {
 
 list.slim = {
   install_info = {
-    url = "https://gitlab.com/theoreichel/tree-sitter-slim",
+    url = "https://github.com/theoo/tree-sitter-slim",
```
clason, line 2047: (It's fine to point to your gitlab repo here, by the way; it's just about the maintainers entry which needs to be a user on Github -- the "owner" is clear from the repo URL.)

lua/nvim-treesitter/parsers.lua
```diff
@@ -2044,10 +2044,11 @@ list.slang = {
...
     files = { "src/parser.c", "src/scanner.c" },
   },
-  maintainers = { "@theoo", "@theoreichel" },
+  maintainers = { "@theoo" },
+  experimental = true,
```
clason, line 2051: Sorry, but we don't accept experimental parsers anymore. If the grammar is not suitable for daily use (because the syntax is not fully covered or breaking changes are expected), then it's better to wait.

If it is, then it's not experimental ;)

queries/slim/highlights.scm
ribru17: I think after removing these comments and adding 

```scm
[
  "#{"
  "}"
] @punctuation.special
```

this should be good to go

theoo: Thanks for the good suggestion. I updated it upstream as well.

theoo:
Thank you for your support and help!

PR #7450, opened 2024-12-07
Enable syntax highlighting in docs

thetic:


clason:
Thank you, but the `master` branch is essentially frozen. I'll have to rewrite the docs completely for the 1.0 release from `main` anyway, and this makes it harder because it introduces conflicts when rebasing.

PR #7475, opened 2024-12-17
fix: perf issue of nu/highlight.scm

blindFS:
Related issue: https://github.com/nushell/tree-sitter-nu/issues/170

### DIFF
--- a/queries/nu/highlights.scm
+++ b/queries/nu/highlights.scm
@@ -356,7 +356,9 @@ key: (identifier) @property
 (comment) @comment @spell
 
 ((comment) @comment.documentation @spell
+  .
   (decl_def))
 
 ((parameter)
+  .
   (comment) @comment.documentation @spell)


ribru17:
Thanks, this is a great change. Just one question: for the parameter doc comment change, I think it would have to be `(parameter (comment) @...)`, at least based on the example they give in the manual for documenting functions:

```nushell
# Greet guests along with a VIP
#
# Use for birthdays, graduation parties,
# retirements, and any other event which
# celebrates an event # for a particular
# person.
def vip-greet [
  vip: string        # The special guest
   ...names: string  # The other guests
] {
  for $name in $names {
    print $"Hello, ($name)!"
  }

  print $"And a special welcome to our VIP today, ($vip)!"
}
```

the comment nodes are children of the parameters here, not adjacent to them. Same thing when you remove the type specifiers

blindFS:
@ribru17 Oh, nice catch. Didn't check that.

PR #7480, opened 2024-12-18
feat: add injection for ruby debugger command strings

DerekStride:
## What

The [ruby debugger](https://github.com/ruby/debug?tab=readme-ov-file#bindingbreak-method) allows passing a String that gets executed as ruby code before dropping into the debugger.

This PR adds support for marking the string content as ruby code.

```ruby
binding.b(pre: "puts 'run this ruby code before dropping me into the debugger'")
binding.break(do: "puts 'run this ruby code and continue executing'")
```

Screenshot 2024-12-18 at 11 17 26

### DIFF
--- a/queries/ruby/injections.scm
+++ b/queries/ruby/injections.scm
@@ -9,3 +9,15 @@
 (regex
   (string_content) @injection.content
   (#set! injection.language "regex"))
+
+((call
+  receiver: (identifier) @_receiver
+  method: (identifier) @_method
+  arguments: (argument_list
+    (pair
+      key: (hash_key_symbol)
+      value: (string
+        (string_content) @injection.content))))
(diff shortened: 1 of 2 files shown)


queries/ruby/injections.scm
```diff
@@ -9,3 +9,15 @@
...
+  receiver: (identifier) @_receiver
+  method: (identifier) @_method
+  arguments: (argument_list
+    (pair
+      key: (hash_key_symbol)
+      value: (string
+        (string_content) @injection.content))))
+  (#eq? @_receiver "binding")
+  (#any-of? @_method "b" "break")
+  (#set! injection.language "ruby"))
```
clason, line 23: You can use `injection.self, I believe.

PR #7490, opened 2024-12-21
fix(install): add -std=c99 flag

jiangyinzuo:
My gcc version is `gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)`. I can not compile parsers without `-std=c99` flag.


clason:
Sorry, your `gcc` version is just too old, we don't support that. (As you can see, this flag breaks compilation on other, supported, systems.)

PR #7481, opened 2024-12-18
Add support for crystal language.

hugopl:
The parser still under (active) development, but it's pretty usable, fast and doesn't crashed since I started using it.

- I use it daily for work and pet projects
- Indent not added yet because I didn't find the documentation about the captures, but I should add it soon.
- Locals not added because I didn't test it yet.

image

PR #7497, opened 2024-12-24
Improve parser build reliability

keidax:
While tracking down the cause of Windows-specific CI failures on https://github.com/nvim-treesitter/nvim-treesitter/pull/7481, I found two improvements to make.

1. `:TSInstallSync` didn't report failures on Windows because it was checking the return code of `popd` instead of the actual command.
2. The root cause of the failure is a `_Static_assert` statement, which is a C11 feature. This project should set the C standard version to C11, since [tree-sitter always does that](https://github.com/tree-sitter/tree-sitter/blob/201b41cf11fb217a1f1ce03ea25b83e62b7b48cb/cli/loader/src/lib.rs#L821).

### DIFF
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -92,6 +92,7 @@ function M.select_compiler_args(repo, compiler)
       "/Isrc",
       repo.files,
       "-Os",
+      "/std:c11",
       "/utf-8",
       "/LD",
     }
@@ -105,6 +106,7 @@ function M.select_compiler_args(repo, compiler)
       "-Isrc",
       "-shared",
       "-Os",
+      "-std=c11",
     }
   else
     local args = {
@@ -113,6 +115,7 @@ function M.select_compiler_args(repo, compiler)
       "-I./src",
       repo.files,
       "-Os",
+      "-std=c11",
     }
     if fn.has "mac" == 1 then
       table.insert(args, "-bundle")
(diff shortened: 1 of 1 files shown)


lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -353,9 +356,9 @@ end
 function M.make_directory_change_for_command(dir, command)
   if fn.has "win32" == 1 then
     if string.find(vim.o.shell, "cmd") ~= nil then
-      return string.format("pushd %s & %s & popd", cmdpath(dir), command)
+      return string.format("pushd %s & %s", cmdpath(dir), command)
```
ObserverOfTime, line 359:
```suggestion
      return string.format("cd %s & %s", cmdpath(dir), command)
```

lua/nvim-treesitter/shell_command_selectors.lua
```diff
@@ -353,9 +356,9 @@ end
...
     else
-      return string.format("pushd %s ; %s ; popd", cmdpath(dir), command)
+      return string.format("pushd %s ; %s", cmdpath(dir), command)
```
ObserverOfTime, line 361:
```suggestion
      return string.format("cd %s ; %s", cmdpath(dir), command)
```

ObserverOfTime (approved):
If we don't need `popd` then `cd` is enough.

keidax:
@ObserverOfTime replacing `pushd` with `cd` is causing all the Windows build to fail with "The system cannot find the path specified."

I don't have a Windows machine to test on, and I'm not sure what the syntax difference is between `pushd` and `cd`. Any thoughts?

ObserverOfTime:
Go back to `pushd` ¯\\\_(ツ)\_/¯

clason:
Thanks, I'll merge this for now even though this whole codepath will be removed in 1.0.

PR #7500, opened 2024-12-25
changed the ordering of compilers on windows to something more sensib…

### LINKED ISSUE #7499: Different ordering of compilers based on the current platform
I have recently run into a problem with treesitter on windows specifically. The way treesitter chooses its compilers is it just pulls the first one from the environment the order is found in the below code snippet

`M.compilers = { vim.fn.getenv "CC", "cc", "gcc", "clang", "cl", "zig" }`
Any Windows users that are using something like MYSYS or CYGWIN may have gcc on the path such as myself and tree-sitter will use cc (common alias for gcc) or gcc first. The parsers I have tested c, cpp when compiled with gcc on Windows just cause a flat-out crash of neovim when attempting to open up any .cpp or .c file. Admittedly I haven't tested every single parser but I think is down to compiling using the platform's incorrect compiler.

More sensible behaviour would be to check for the Windows platform and reorder the M.compilers to something more sensible like 
`M.compilers = { vim.fn.getenv "cl", "clang", ...}`

Poundednote:
fixes #7499 in which the default C compiler was set to gcc due to having MYSY2/mingw installed and gcc in the path causing parsers to be compiled with gcc and causing crashes on the Windows version of neovim

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -19,7 +19,7 @@ local lockfile = {}
 
 if fn.has "win32" == 1 then
   M.compilers = { vim.fn.getenv "CC", "cl", "clang", "cc", "gcc", "zig" }
-else 
+else
   M.compilers = { vim.fn.getenv "CC", "cc", "gcc", "clang", "cl", "zig" }
 end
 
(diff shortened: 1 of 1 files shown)


clason:
Thank you, but the `master` branch is effectively frozen. Until then, just set the compiler manually.

PR #7504, opened 2024-12-27
feat(ghostty): add parser & queries for Ghostty configuration files

bezhermoso:
- Register [tree-sitter-ghostty](https://github.com/bezhermoso/tree-sitter-ghostty) parser for [Ghostty](https://ghostty.org) configuration files
- Add queries for syntax highlighting


clason:
Sorry, we no longer accept experimental parsers. And captures should correspond to the syntax elements, not how they should look.

PR #7510, opened 2024-12-29
Fix/leo highlight highlight for casting, record initialization and indents

r001:
**highlights**:
Record assignment highlighting added
Casting expression highlighting added

**indents**:
Fixed indents as the "}" did not indent well before, now all perfect


clason:
Why did you close this?

PR #7512, opened 2024-12-29
feat(yaml): highlighting to help deal with yaml's Norway problem

chrisgrieser:
**Problem**  
In yaml there are some strings that, when not unquoted, are evaluated as boolean values instead of strings, for example `off` or `no`. This is also known as [yaml's Norway problem](https://www.bram.us/2022/01/11/yaml-the-norway-problem/), and results in easy to overlook bugs since the syntax highlighting of those tokens as string is quite misleading.

**Solution**  
The PR changes the highlight group of those strings to `@boolean`. It does not affect quoted strings. The specific strings this highlight-group-change is triggered on are taken from [yamllint's related `truthy` rule](https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy).

---

before
Showcase

after
Showcase

### DIFF
--- a/queries/yaml/highlights.scm
+++ b/queries/yaml/highlights.scm
@@ -82,7 +82,7 @@
 ] @punctuation.special
 
 ; help deal with for yaml's norway problem https://www.bram.us/2022/01/11/yaml-the-norway-problem/
-; values based on https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy
+; only using `true` and `false`, since Treesitter parser targets YAML spec 1.2 https://github.com/nvim-treesitter/nvim-treesitter/pull/7512#issuecomment-2565397302
 (block_mapping_pair
   value: (block_node
     (block_sequence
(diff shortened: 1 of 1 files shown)


clason:
Hmm, this seems not to be standardized. I wonder if it might not be better to stick to the minimal list guaranteed to be truthy?

chrisgrieser:
Somewhat. From the linked yamllint docs:

> In YAML 1.2, only capitalized / uppercased combinations of true and false are considered truthy, whereas in YAML 1.1 combinations of yes, no, on and off are too.

I am really not knowledgeable about yaml versions, but apparently a lot of stuff still runs on 1.1. Had the `yes` problem with GitHub actions not too long ago, for example.

---

However, there is also an argument to be made to make this change regardless.
1. If the user's yaml processor is on 1.1, the change of this PR will be very useful in preventing bugs. If the user's yaml processor is on 1.2, the user adds quotes to their `yes`/`no`/… strings without it being necessary. The benefit for the former case in my view outweighs the downside of the latter case by far.
2. In addition, it will often not be clear to the user whether a yaml file will be processed by 1.1 or 1.2, and even if processed by 1.2, the file might get shared with another user etc.—simply for portability and to be on the safe side, it is thus reasonable to quote the `yes`/`no`/… strings even if on 1.2

clason:
Yes, the best way is not clear, that's why I am asking. As another data point, the [grammar](https://github.com/tree-sitter-grammars/tree-sitter-yaml) explicitly references the YAML version 1.2 specification, so I would expect this to be the versions that queries target as well. @amaanq ?

(Strings being incorrectly highlighted as booleans can lead to bugs, too.)

Things are muddied further by the fact that this can be configured via `allowed-values`...

ObserverOfTime:
The grammar uses the YAML 1.2 [core schema](https://yaml.org/spec/1.2.2/#103-core-schema) by default, with a build-time option to use the more restrictive [JSON schema](https://yaml.org/spec/1.2.2/#json-schema) instead. The old YAML 1.1 schema is not currently supported.

clason:
@chrisgrieser Sorry for the lack of response, but it seems there's no progress on adding YAML 1.1 support to the parser.

So we consider this parser to target YAML 1.2 exclusively. If you update the PR to only include the cases that are problematic _in that specific schema_, I'll merge it. Other cases can be revisited at a later date.

chrisgrieser:
@clason Sure, have done so. There aren't many cases left then, but I guess it's still better than nothing.

clason:
Indeed.

PR #7522, opened 2025-01-02
fix(php): properly highlight grouped imported functions and constants

calebdw:
Hello!

This fixes highlighting for grouped functions and constants like:

```php
use function Foo\{bar, baz};
use const Foo\{FOO, BAR};
```

Thanks!


calebdw:
Thanks!

PR #7526, opened 2025-01-04
fix(rbs): fix highlight queries of rbs that is missing some definitions

joker1007:
Some rbs highlight definitions are missing.
This is reported by https://github.com/joker1007/tree-sitter-rbs/issues/15
I apply the fixes to this repository.

### DIFF
--- a/queries/rbs/highlights.scm
+++ b/queries/rbs/highlights.scm
@@ -54,9 +54,24 @@
   (method_name
     [
       (identifier)
+      (identifier_suffix)
       (constant)
+      (constant_suffix)
       (operator)
       (setter)
+      (constant_setter)
+    ] @function.method))
+
+(attribute_member
+  (method_name
+    [
+      (identifier)
+      (identifier_suffix)
+      (constant)
+      (constant_suffix)
+      (operator)
+      (setter)
+      (constant_setter)
     ] @function.method))
 
 [
@@ -114,6 +129,10 @@
 
 (type
   (integer_literal) @number)
+
+(type
+  (record_type
+    key: (record_key) @string.special.symbol))
 
 ; Operators
 [


queries/rbs/highlights.scm
```diff
@@ -115,6 +130,10 @@
 (type
   (integer_literal) @number)
 
+(type
+  (record_type
+    key: (record_key) @string.special.symbol))
```
clason, line 135: if this is a key in a key/value pair, it should be `@property`

queries/rbs/highlights.scm
```diff
@@ -54,9 +54,24 @@
...
       (identifier)
+      (identifier_suffix)
       (constant)
+      (constant_suffix)
       (operator)
       (setter)
+      (constant_setter)
+    ] @function.method))
+
+(attribute_member
```
clason, line 65: can these be combined via alternations (`[method_member attribute member]`)?

queries/rbs/highlights.scm
joker1007: in RBS, the key means the Hash table's key in Ruby language. The key class is Symbol.
I think `@string.special.symbol` represents the key semantics.

clason: Do you have an example for me?

queries/rbs/highlights.scm
joker1007: I tried to combine these like the following.

```
(([(method_member) (attribute_member)]
  (method_name
    [
     (identifier)
     (identifier_suffix)
     (constant)
     (constant_suffix)
     (operator)
     (setter)
     (constant_setter)
    ] @method)))
```

But tree-sitter cannot highlight properly.
Is there a mistake?

clason: Yeah, that has a different (incorrect) meaning (missing nesting). Nevermind then, thanks for checking!

queries/rbs/highlights.scm
joker1007: I show a example.

foo.rbs: this is a type signature file.
```rbs
class Foo
  @record: {a: String, b: Integer}
end
```

foo.rb: this is an actual Ruby code.
```ruby
class Foo
  def initialize
    @record = {a: "string value", b: 1234}
  end

  def get_a
    @record[:a]
  end
end
```

In this case, the record key in the rbs file means the fixed-key Hash table type in Ruby code.

In Ruby code, the Symbol class is used as a key to refer to this value, which is represented as `@string.special.symbol` in Ruby syntax highlighting.

Since they refer to the same thing semantically, it is easier for programmers to understand if they are highlighted the same way.

clason:
Thanks!

PR #7534, opened 2025-01-07
feat(koto): Update highlighting for 0.15

irh:
Adds some missing highlight queries for `0.15`.

- `->` and `?` operators
- `let` keyword
- Type hints


clason:
You can just run `./scripts/format-queries.lua queries/koto` to make the linter happy.

irh:
@clason Thanks 👍

PR #7542, opened 2025-01-10
feat(editorconfig): highlight builtin properties

yochem:


yochem:
Whoops, `@property.builtin` is not supported. The `.builtin` 'subproperty' seems the most fitting to me, but I also don't want this to be a breaking change by deviating from `@property`. And it's also weird to have for example `@property` and `@constant.builtin`. What would you recommend?

clason:
I would recommend doing nothing then (here; in your config you can do whatever you like ;)).

clason:
(For editorconfig, "non-builtin" properties don't exists afaik, so this doesn't really help. As always, we want to highlight correct code correctly; we don't care about incorrect since we're not implementing a linter here.)

yochem:
Thanks for the explanation! It is indeed true that the format allows arbitrary properties, but I thought a _possibility_ to quickly see which keys are "officially supported"[1] would be useful, since something similar is done in the [editorconfig syntax file](https://github.com/neovim/neovim/blob/master/runtime/syntax/editorconfig.vim). Sorry for the noice :)

[1]: Would've been better to use https://spec.editorconfig.org/#supported-pairs for it then

clason:
Ah, ok. Yes, that _would_ make sense then, but not worth adding new captures just for that. I'm sure there's a language server with semantic highlighting, though ;)

PR #7543, opened 2025-01-10
Update injections.scm

bcardarella:
Match the Elixir treesitter updates for LVN sigil support: https://github.com/elixir-lang/tree-sitter-elixir/commit/53458546e3bb717beee1d15df30724c81eb41d1c

If possible to also bump up the heex treesitter packaged. 0.7.0 was recently released and that adds file alias support for neex -> HEEX syntax highlighting: https://github.com/phoenixframework/tree-sitter-heex/commit/9359017bd0dc6b023044713aa215544885663637

If you ned a PR for that I can do it, just not sure where to add the extension alias

### DIFF
--- a/queries/elixir/injections.scm
+++ b/queries/elixir/injections.scm
@@ -21,7 +21,7 @@
 (sigil
   (sigil_name) @_sigil_name
   (quoted_content) @injection.content
-  (#match? @_sigil_name "^(H|LVN)$")
+  (#any-of? @_sigil_name "H" "LVN")
   (#set! injection.language "heex"))
 
 ; Surface
--- a/queries/elixir/injections.scm
+++ b/queries/elixir/injections.scm
@@ -21,7 +21,7 @@
 (sigil
   (sigil_name) @_sigil_name
   (quoted_content) @injection.content
-  (#eq? @_sigil_name "H")
+  (#match? @_sigil_name "^(H|LVN)$")
   (#set! injection.language "heex"))
 
 ; Surface


clason:
Parsers are automatically updated once daily. Unless that is a breaking change, best to wait for that and then make query changes afterwards. 

Filetypes are irrelevant here; these are handled by Vim (not Neovim!) so any change needs to be made there.

queries/elixir/injections.scm
```diff
@@ -21,7 +21,7 @@
 (sigil
   (sigil_name) @_sigil_name
   (quoted_content) @injection.content
-  (#eq? @_sigil_name "H")
+  (#match? @_sigil_name "^(H|LVN)$")
```
clason, line 24: No. Use #any-of?, or #lua-match? if the anchors are strictly necessary.

clason (dismissed):

queries/elixir/injections.scm
bcardarella: if I remove the anchors is `match` ok to use or should I switch to `any-of`?

clason:
We are not involved in this project. You can see which parser we track (the default branch of) in parsers.lua

bcardarella:
@clason I did, that is the project you're tracking https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/parsers.lua#L514

clason:
Then we are tracking HEAD and I don't understand your point.

bcardarella:
HEAD incudes the change that I PR'd and that change is not in this repo.

queries/elixir/injections.scm
bcardarella: Can you provide me a direct answer on what solution you want?

clason:
Please be more explicit about what you mean. If you refer to queries, these are _never_ synced; queries are editor-dependent and are manually updated here.

clason:
@connorlay

bcardarella:
I'm happy to do whatever work is necessary to get this lib caught up with the Elixir release. If you want me to copy in what's been released there I can do that, but that release has the `match` syntax which you said you don't want. If you want me to use `any-of` instead I'll implement that. I'm not familiar with your standards and just need some direction here and I'm happy to do the work.

queries/elixir/injections.scm
clason: #any-of?, if the anchor is not strictly necessary

clason:
What _is_ your background, then, if I might ask? Why are you making this PR?

bcardarella:
I'm making the PR to bring this project in line with recent changes that were accepted and released in the elixir tree-sitter repo.

clason:
But why? What is your personal investment here? 

Obviously we have an impedance mismatch so I need to recalibrate.

bcardarella:
The LVN sigil this PR aliases is for the LiveView Native framework which has a template syntax that is the same as HEEx

clason:
That does not answer my question. 

But never mind, this is wasting both our time. Use any-of, period.

bcardarella:
Thank you Clay, much appreciated. just at dinner with my family. I’ll
update PR when I get back home.


----------------------
Brian Cardarella
Founder, DockYard

bcardarella:
@connorlay should I PR the elixir repo as well with the same syntax?

queries/elixir/injections.scm
ribru17: :point_up: Especially since there are start and end anchors, making this equivalent to `#any-of?`

bcardarella: @ribru17 yup, just updated. If there are any other required changes please let me know.

PR #7545, opened 2025-01-11
feat: add razor

tris203:
Adds `razor` parser

A superset of C#/HTML for .NET web development


queries/razor/highlights.scm
```diff
@@ -0,0 +1,662 @@
+; inherits: html_tags
+
+(razor_comment) @comment @spell
+
+(razor_page_directive) @constant.macro
```
clason, line 5: Please group common captures.

queries/razor/highlights.scm
```diff
@@ -0,0 +1,662 @@
...
+
+(razor_html_attribute) @function
+
+(razor_attribute_value
+  (identifier) @function)
+
+; HTML RULES
+(tag_name) @tag
+
+; (erroneous_end_tag_name) @error ; we do not lint syntax errors
```
clason, line 23:
```suggestion
```

clason:
Please add injections, at least for comments. (Folds shouldn't be hard, either.)

clason:
Wouldn't it make more sense to inherit from csharp queries, since you extend that grammar?

(Also, make sure  `tree-sitter build` and `tree-sitter generate` works without errors. Generation needs to happen from `grammar.json` so users don't need `npm`, so make sure to keep that file committed and updated.)

tris203:
> Wouldn't it make more sense to inherit from csharp queries, since you extend that grammar?
> 
> (Also, make sure `tree-sitter build` and `tree-sitter generate` works without errors. Generation needs to happen from `grammar.json` so users don't need `npm`, so make sure to keep that file committed and updated.)

I didn't do that, as it doesn't fully inherit the `c_sharp` grammar. For example, there are no preprocessor rules

tris203:
i think that is all addressed @clason 

let me know if you want to see anything else, or if I can mark this ready :) 

Thanks for your guidance, I'm quite green in the lisp style syntax

ribru17:
Looks like the parser can't handle colons (`:`), take this file for example:

```razor
@{
var Shows = ShowDatabase.Recent();
Response.ContentType = "application/atom+xml";
}
<?xml version="1.0" encoding="UTF-8"?>
<feed xmlns:itunes="http://www.itunes.com/dtds/podcast-1.0.dtd" xmlns="http://www.w3.org/2005/Atom">
  <title type="text">This Developer's Life</title>
  <generator uri="http://asp.net/webmatrix">Web Matrix</generator>
  <id>tag:thisdeveloperslife.com,2011:/</id>
  <link rel="self" href="http://thisdeveloperslife.com/rss" />
  <link rel="alternate" href="http://thisdeveloperslife.com" />
  <subtitle type="text">Stories About Developers and Their Lives</subtitle>
  <itunes:image>http://www.thisdeveloperslife.com/images/logo_full1400.jpg</itunes:image>
  <updated>@Sugar.ToRFC3339(Shows.Last().PublishedAt)</updated>
  @foreach(var show in Shows){
  <entry>
    <title>@show.Title</title>
    <author>
      <name>Rob Conery, Scott Hanselman</name>
      <uri>http://thisdeveloperslife.com</uri>
    </author>
    <link type="text/html" rel="alternate" href="http://thisdeveloperslife.com/@Sugar.RelativeUrl(show)" />
    <id>tag:thisdeveloperslife.com,@show.PublishedAt.ToString("yyyy-MM-dd"):/@Sugar.RelativeUrl(show)</id>
    <content type="html">
      @ShowDatabase.FeedContent(show)
    </content>
    <published>@Sugar.ToRFC3339(show.PublishedAt)</published>
    <updated>@Sugar.ToRFC3339(show.PublishedAt)</updated>
  </entry>
  }
</feed>
<!-- vim: set ft=razor: -->
```

(ref: https://github.com/shanselman/thisdeveloperslife/blob/71975849295651aa4948d4cf49361947e628991f/rss.cshtml)

ribru17:
It also errors with html tags like this: `<input type="text" name="UserName" required />` (due to the `required`). Could this  inherit from or inject html?

tris203:
> It also errors with html tags like this: `<input type="text" name="UserName" required />` (due to the `required`). Should this inject html instead of trying to parse it itself?

I want to do that, but the grammr extends the `c_sharp` grammar, and I don't think you can extend from two grammars? I especially don't know how that wold work as both `c_sharp` and `html` use custom scanners.

> @section Main
> {
> <h2>Log On</h2>
> }
> ```
I don't think that is valid, as `section` isn't a razor keyword
https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-9.0

here is a link to the syntax overview

ribru17:
> I don't think that is valid, as `section` isn't a razor keyword

Hmm, I do see it here though: https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-9.0#razor-keywords (got there from a link here:

> When an @ symbol is followed by a [Razor reserved keyword](https://learn.microsoft.com/en-us/aspnet/core/mvc/views/razor?view=aspnetcore-9.0#razor-reserved-keywords), ...

> I want to do that, but the grammr extends the `c_sharp` grammar, and I don't think you can extend from two grammars?

I think you could do this without extending the grammar, maybe just broadly capture HTML tags as tags (trying not to get ERROR nodes) and then from an injection query inject HTML into there. I think markdown does something similar.

tris203:
see my updated comment about `@section` you were right, i just don't parse that yet

> I think you could do this without extending the grammar, maybe just broadly capture HTML tags as tags (trying not to get ERROR nodes) and then from an injection query inject HTML into there. I think markdown does something similar.

There is also the issue that you can interweave the HTML and razor, similar to JSX
```razor
<button class="btn btn-primary" @onclick="IncrementCount">Click me</button>
```
```razor
                    <td>@forecast.Date.ToShortDateString()</td>
```

So I don't think capturing it at the outer level will work with enough resolution

ribru17:
Will C# always start with @ inside HTML? You could take a look at https://github.com/tree-sitter/tree-sitter-embedded-template/blob/master/grammar.js (for .erb) which seems really similar to this

tris203:
> Will C# always start with @ inside HTML? You could take a look at [tree-sitter/tree-sitter-embedded-template@`master`/grammar.js](https://github.com/tree-sitter/tree-sitter-embedded-template/blob/master/grammar.js?rgh-link-date=2025-01-21T21%3A46%3A23Z) (for .erb) which seems really similar to this
> 
> Edit: then we could just do something like this: [`07bd1b5`/queries/embedded_template/injections.scm](https://github.com/nvim-treesitter/nvim-treesitter/blob/07bd1b53bf465e42d53253b48b7437892d6c45e8/queries/embedded_template/injections.scm)

yes, but so will other things, like a list with a loop inside it
```razor
<ul>
  @foreach(var fruit in Fruits){
<li>@fruit</li>
}
</ul>
}
```

ribru17:
That makes sense. Imo the best thing to do would be to emulate something like erb and just parse sections into either html or code, and apply injections from there (or parse all code from within razor, since like you say it is not a true C sharp subset)

tris203:
> It also errors with html tags like this: `<input type="text" name="UserName" required />` (due to the `required`). Should this inject html instead of trying to parse it itself?

just pushed a fix for this.

I think the best route is to try and parse it all within razor, the fact its not a full subset of either HTML or csharp and kind of interweaves the two. Its probably most like JSX or templ

ribru17:
In that case I trust your judgement, though I do think the parser needs a tiny bit more polishing before we can merge this

clason:
Agreed; this would have been a case for marking a parser as "experimental", but we no longer accept those.

If the required changes are bigger or will take longer, it's probably better to close this and open a new PR when you're ready.

tris203:
I'll aim to have it done by the end of the weekend

queries/razor/highlights.scm
```diff
@@ -0,0 +1,689 @@
+(html_text) @none @spell
```
clason, line 1:
```suggestion
(html_text) @spell
```
`@none` does nothing in Neovim.

queries/razor/highlights.scm
```diff
@@ -0,0 +1,689 @@
...
+
+(namespace_declaration
+  name: (identifier) @module)
+
+(file_scoped_namespace_declaration
+  name: (identifier) @module)
+
+(qualified_name
+  (identifier) @module
+  (#not-has-ancestor? @module method_declaration)
```
clason, line 219: Those can be quite expensive.

queries/razor/highlights.scm
```diff
@@ -0,0 +1,689 @@
...
+; HTML RULES
+(tag_name) @tag
+
+(comment) @comment @spell
+
+(html_attribute_name) @tag.attribute
+
+((html_attribute
+  (html_attribute_value) @string)
+  (#set! priority 99))
```
clason, line 64: Can you comment on why that is needed? You're not injecting anything here, so priority should not matter. (If it does, you need to re-order the queries.)

ribru17:
Can you follow up on why it was not possible to parse the html in the way discussed earlier?

tris203:
> Can you follow up on why it was not possible to parse the html in the way discussed earlier?

I can, and I have
This is way cleaner, it just took me a while to see it

Thanks for all of your guidance, I have learnt a tonne

queries/razor/highlights.scm
```diff
@@ -0,0 +1,58 @@
+; inherits: c_sharp
+
+[
+  (razor_comment)
```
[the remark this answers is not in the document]
tris203: This isn't valid, inside a razor block is c# style comments
```razor
@{
// This is a comment.
var Shows = ShowDatabase.Recent();
Response.ContentType = "application/atom+xml";
}
```

queries/razor/injections.scm
```diff
@@ -0,0 +1,13 @@
+((html_comment) @injection.content
+  (#set! injection.language "comment"))
+
+((razor_comment) @injection.content
+  (#set! injection.language "comment"))
+
+((comment) @injection.content
+  (#set! injection.language "comment"))
```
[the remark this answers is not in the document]
tris203: fixed

ribru17 (approved):
Looks great. Thanks a ton for your patience!

PR #7547, opened 2025-01-11
feat(nix): Fix function highlights when part of a set

bodby:
Functions in Nix are only highlighted when they aren't part of a set. This may be intentional since GitHub highlights them the same way (see below), though.

```nix
# This gets highlighted as a function.
someFunc 32;

# These do not. 'someOtherFunc' gets highlighted with @variable.member.nix only.
lib.someOtherFunc 48;
lib.hello.someOtherFunc { };
```

There probably is a more efficient way to do this, but I don't know Scheme. More specifically, I couldn't figure out how to select the last node and had to write two different queries (one for `a.b.func` and one for `a.func`).

### DIFF
--- a/queries/nix/highlights.scm
+++ b/queries/nix/highlights.scm
@@ -88,7 +88,7 @@
     (select_expression
       attrpath: (attrpath
         (_)
-        attr: (identifier) @function.call))
+        attr: (identifier) @function.call .))
   ])
 
 ; basic identifiers
--- a/queries/nix/highlights.scm
+++ b/queries/nix/highlights.scm
@@ -79,6 +79,43 @@
   function: (variable_expression
     name: (identifier) @function.call))
 
+; basic identifiers
+(variable_expression) @variable
+
+(variable_expression
+  name: (identifier) @keyword.import
+  (#eq? @keyword.import "import"))
+
+(variable_expression
+  name: (identifier) @boolean
+  (#any-of? @boolean "true" "false"))
+
+; string interpolation (this was very annoying to get working properly)
+(interpolation
+  "${" @punctuation.special
+  (_)
+  "}" @punctuation.special) @none
+
+(select_expression
+  expression: (_) @_expr
+  attrpath: (attrpath
+    attr: (identifier) @variable.member)
+  (#not-eq? @_expr "builtins"))
+
+(attrset_expression
+  (binding_set
+    (binding
+      .
+      (attrpath
+        (identifier) @variable.member))))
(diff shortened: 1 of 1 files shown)


bodby:
This also breaks in one of my files:

```nix
# ...
mkSystem =
  hostnameSet:
  nixpkgs.lib.attrsets.genAttrs' (system: { });
  #           ^^^^^^^^
  # 'attrsets' gets highlighted when it shouldn't; only genAttrs' should be.
```

clason:
Yeah, that won't work without grammar support.

bodby:
I got it to work by adding an extra dot so it only matches the last `attr` node. (At least, it seems to highlight all the functions in my Nix files properly.)

This really isn't elegant, but I have no idea how to develop grammars.

clason:
Grammars are a DSL based on JS; you don't have to interface with node itself, `tree-sitter` CLI handles all that.

ribru17:
These highlights are being overridden by the `@variable.member.nix` highlights for me; maybe we can move them lower (while being careful to make the `.builtin` versions still have higher precedence)

bodby:
To before https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/nix/highlights.scm#L93?

bodby:
Don't know what I'm doing, it's late and I have no way to properly test this quickly. Sorry.

bodby:
Seems to work now ~~on my machine~~.

Red is a builtin function, and blue is normal. All my other files look fine too.

ribru17 (approved):
Thanks!

bodby:
Not sure if I should open a new PR for this but I made the query 2 lines shorter (wow).

https://github.com/bodby/nvim-btw/blob/master/nvim/after/queries/nix/highlights.scm

Do queries somehow affect performance much if they're written differently?

ribru17:
What about just:

```
function: (select_expression
  attrpath: (attrpath
    attr: (identifier) @function.call .))
```

seems to work for me, but I haven't tested it very rigorously

bodby:
I think I had that initially but it would highlight extras, e.g.

```nix
set1.set2.set3.function { };
#    ^^^^^^^^^^^^^^^^^^
# All of these get highlighted, and not just 'function'.
```

bodby:
Seems to work for me too. Is it possible to merge a PR if it's already merged?

On a side note, what do you call the `(_)`?

bodby:
Opened another in #7579 then.

PR #7562, opened 2025-01-17
feat(css): add @layer to folds query

mparq:
Allows folding [css @layer](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) blocks.

Not sure if it's best to add here or to try to add a new layer_node in treesitter-css. First time playing with treesitter, just wanted to add this because I've been working with css layers recently, let me know if there's a better way to do this!


clason:
@amaanq ?

amaanq:
No, it should be folding `(block)`, that's the actual node spanning multiple lines

mparq:
@amaanq that makes sense to me, and that's what I originally expected from the folds query, but I figured there was a reason why `(block)` wasn't already being used, as `rule_set` also contains `block`? Happy to update but just wanted to check first why it wasn't already there.

mparq:
I tested it and `(block)` seems to work well for my personal use case. It includes layers as well as other things which were not folding before like `@media` query blocks. I also see why `rule_set` was used over `block` before because otherwise, folds don't work well for selectors which span multiple lines

```css
h1, /** <-- (rule_set) folds here */
p,
span,
div { /** <-- (block) folds here*/
}
```

this ends up adding an extra fold for the block within a ruleset which wasn't there before but I'm guessing that's okay. Maybe there are similar situations for other blocks which can have new lines between the start of the region and start of the block, but those wouldn't have been folding in the first place anyway so the behavior seems at least better than before.

amaanq:
Tbh, not really sure why `block` wasn't added before, I don't think I wrote the query initially so I couldn't say why. Good call on the block folding for selectors spanning multiple lines - if you're up for it, you can restrict the parent nodes of the block node to not include `rule_set`, which would be `media_statement`, `keyframe_block`, `supports_statement`, and `at_rule`.

mparq:
Thanks! I added those with a comment. I ended up just using all the top-level statements directly since it seems they can all span multiple lines before the block node, and this follows pre-existing ux with rule_set.

PR #7568, opened 2025-01-18
feat(hyprlang): add missing highlights

ValdezFOmar:
Before:
image

After:
image

### DIFF
--- a/queries/hyprlang/highlights.scm
+++ b/queries/hyprlang/highlights.scm
@@ -16,7 +16,7 @@
   (name) @module)
 
 (section
-  device: (device_name) @type)
+  device: (device_name) @string.special)
 
 (variable) @variable
 
@@ -24,7 +24,13 @@
 
 (boolean) @boolean
 
-(mod) @constant
+[
+  "deg"
+  (string)
+  (string_literal)
+] @string
+
+(mod) @constant.builtin
 
 [
   "rgb"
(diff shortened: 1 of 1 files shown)


queries/hyprlang/highlights.scm
```diff
@@ -16,15 +16,21 @@
   (name) @module)
 
 (section
-  device: (device_name) @type)
+  device: (device_name) @string.special)
```
clason, line 19: Can you comment on this change?

ValdezFOmar: I don't find the `type` capture to be appropriate. `(device_name)` is the name of the device as reported by `hyprctl devices`, so it's not which type of device it is, but a name to target that specific piece of hardware ([docs](http://wiki.hyprland.org/0.28.0/Configuring/Keywords/#per-device-input-configs)).

clason: `@string.special.path` then?

clason: But agreed, `@type` seems off.

ValdezFOmar: I looked at the example sway config and found its [input configuration](https://github.com/swaywm/sway/blob/30c858423dcdfb82d768a5025b5f00c19b250322/config.in#L45-L57) is similar, but the syntax does actually look like a typical string.
Another thing to take into account its that `device:name` is the old syntax for Hyprland, and the [new syntax](https://wiki.hyprland.org/Configuring/Keywords/#per-device-input-configs) would recognized the device name as a `string` node. Maybe `@string.special` is not that bad of an option?

clason: Alright, thanks for checking!

PR #7571, opened 2025-01-19
fix(inc-selection): parse injection

tomtomjhj:
Fixes https://github.com/neovim/neovim/issues/32093


tomtomjhj:
@wurli added some more fixes. can try it again?

wurli:
Hello, thanks and nice work! Things seem to be working correctly; running `vim.treesitter.stop()` turns off treesitter highlighting, but incremental selection continues to work as expected 🥳

Only issue I noticed is that I get an error if I try to use incremental selection with the cursor on the first 'E' of the final `EOF` in the example you gave:

``` lua
vim.cmd[=[
echo 'hello' 'world!'
lua << EOF
vim.cmd[[echo 'hello' 'world!']]
EOF
]=]
```

```
E5108: Error executing lua: ...eovim/0.10.2_1/share/nvim/runtime/lua/vim/treesitter.lua:167: attempt to index local 'node_or_range' (a nil value)
stack traceback:
	...eovim/0.10.2_1/share/nvim/runtime/lua/vim/treesitter.lua:167: in function 'get_node_range'
	...im/lazy/nvim-treesitter/lua/nvim-treesitter/ts_utils.lua:279: in function 'update_selection'
	...treesitter/lua/nvim-treesitter/incremental_selection.lua:20: in function <...treesitter/lua/nvim-treesitter/incremental_selection.lua:15>
```

This occurs regardless of whether I've run `vim.treesitter.stop()`.

tomtomjhj:
> Only issue I noticed is that I get an error if I try to use incremental selection with the cursor on the first 'E' of the final EOF in the example you gave:

turns out this was an already existing problem unrelated to the current pr. anyway, I added the second commit to fix that problem.

wurli:
Awesome, thanks for fixing and for looking at this issue so quickly!

clason:
Thanks, I will merge this for now, but the functionality will still be removed in 1.0. People interested in this functionality should consider breaking it out (and maintaining it) as a standalone plugin that only uses Neovim APIs.

PR #7572, opened 2025-01-19
feat(hyprlang): bash injections for exec keybindings

ValdezFOmar:
Example:

```
bind = SHIFT, Print, exec, slurp | grim -g - - | wl-copy
bind = $mod, P, exec, uwsm app -- hyprpicker --autocopy --format=hex
```

Before:
image

After:
image

### DIFF
--- a/queries/hyprlang/injections.scm
+++ b/queries/hyprlang/injections.scm
@@ -4,3 +4,18 @@
 (exec
   (string) @injection.content
   (#set! injection.language "bash"))
+
+((keyword
+  (name) @_bind
+  (params
+    .
+    (_)
+    .
+    (_)
+    .
+    (string) @_exec
+    .
+    (string) @injection.content))
+  (#lua-match? @_bind "^bind")
+  (#lua-match? @_exec "^%s*exec%s*$")
+  (#set! injection.language "bash"))


queries/hyprlang/injections.scm
```diff
@@ -4,3 +4,18 @@
...
+    .
+    (_)
+    .
+    (_)
+    .
+    (string) @_exec
+    .
+    (string) @injection.content))
+  (#lua-match? @_bind "^bind")
+  (#lua-match? @_exec "^%s*exec%s*$")
```
amaanq, line 20: why not just check if its equal to exec? (same with bind?)

clason: Yeah, never to a pattern when an exact match will do.

ValdezFOmar: In the case of `bind`, is because it accepts trailing characters as a form of flags ([docs](https://wiki.hyprland.org/Configuring/Binds/#bind-flags))

In the case of `exec`, is because the parser includes the surrounding white space as part of the `string` node, so it needs to be accounted for.

amaanq: Makes sense, thanks

PR #7573, opened 2025-01-19
feat(jinja2): add jinja2

cathaysia:
continue of #7038


lua/nvim-treesitter/parsers.lua
```diff
@@ -1119,6 +1119,24 @@ list.javascript = {
   maintainers = { "@steelsojka" },
 }
 
+list.jinja2 = {
+  install_info = {
+    url = "https://github.com/cathaysia/tree-sitter-jinja2",
```
clason, line 1124: Didn't you rename these?

cathaysia: What to rename?

clason:
You cannot name your parser "tree-sitter-jinja2" if you declare the name to be "jina" in your grammar: 
https://github.com/cathaysia/tree-sitter-jinja2/blob/9331e21b421d833755e5c752d237f003c31b1b6b/tree-sitter-jinja/grammar.js#L4

Please pick either "jinja" or "jinja2" and stick to it everywhere. (For the record, the Vim filetype is "jinja".)

cathaysia:
> You cannot name your parser "tree-sitter-jinja2" if you declare the name to be "jina" in your grammar: https://github.com/cathaysia/tree-sitter-jinja2/blob/9331e21b421d833755e5c752d237f003c31b1b6b/tree-sitter-jinja/grammar.js#L4
> 
> Please pick either "jinja" or "jinja2" and stick to it everywhere. (For the record, the Vim filetype is "jinja".)

solved

clason:
Tests are failing. Please follow CONTRIBUTING.md.

clason:
Proper review of the queries, which I don't have time for at the moments.

queries/jinja/highlights.scm
```diff
@@ -0,0 +1,176 @@
...
+[
+  "("
+  ")"
+  "["
+  "]"
+  "<"
+  ">"
+] @punctuation.bracket
+
+(binary_operator) @operator
```
ribru17, line 51: Also `"="` (which maybe should be considered binary operator?)

queries/jinja/highlights.scm
```diff
@@ -0,0 +1,176 @@
...
+(function_call
+  (identifier) @function.call)
+
+(arg
+  (identifier) @variable.parameter)
+
+(expression
+  (expression) @variable.member
+  "."
+  (expression)+ @variable.member)
```
ribru17, line 125: I think this should be something like

```
(expression
  "."
  (expression) @variable.member)
```
since the first variable being indexed is not necessarily a member variable, and also the `+` is unneeded here

Similar story for the query below

ribru17:
Can you give an example of when `jinja_inline` would be applied? The queries seem almost identical, makes me thing maybe jinja should just inject itself into these places

cathaysia:
> Can you give an example of when jinja_inline would be applied?

Here is a example:

```jinja
{#
    block comment
#}
{%
if true
%}

{% endif %}

## inline comment
# set something = false
```

Here is a render result:

image

> The queries seem almost identical.

Yes, jinja and jinja_inline are very seem, expect that jinja allow NEW_LINE between expression: https://github.com/cathaysia/tree-sitter-jinja/blob/851741d96195ea10133885841d8215ebfe45b238/tree-sitter-jinja/grammar.js#L5

and jinja_inline don't allow NEW_LINE between expression:

https://github.com/cathaysia/tree-sitter-jinja/blob/851741d96195ea10133885841d8215ebfe45b238/tree-sitter-jinja_inline/grammar.js#L5 

> makes me thing maybe jinja should just inject itself into these places.

Unless we can inject specific rule, you can only handle extra rules manually. This is really troublesome.

queries/jinja/highlights.scm
cathaysia: solved by [6ee7c483d353c94737c8000ffbb745f6cce5021b](https://github.com/cathaysia/tree-sitter-jinja/commit/6ee7c483d353c94737c8000ffbb745f6cce5021b)

queries/jinja/highlights.scm
```diff
@@ -0,0 +1,176 @@
...
+"debug" @function.builtin
+
+; TODO: only match raw
+(raw_start) @keyword
+
+(raw_end) @keyword
+
+(raw_body) @markup.raw.block @nospell
+
+(identifier) @variable.parameter
```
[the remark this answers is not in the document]
cathaysia: There are two queries:

```lisp
; func(name)
;            ^ ident 
(arg

; func(name=true)
;             ^ident
(arg
  (expression
    (binary_expression
      (unary_expression
        (primary_expression
          (identifier) @variable.parameter)))))
```

cathaysia:
🤔

cathaysia:
Image a grammar like this:

```js
{
    extras: $=> [/\s/, $.comment],
    rules: $=> {
        control: $=> seq('{%', $.expression, '%}'),
        inline_control: $=> seq('# ', $.expression),
    }
}
```

so it will parse a statement like this:

```jinja
{%
 if true
%}
```

and

valid:
```jinja

# if true
```

invalid:
```jinja
# 
if true
```

How do you solve the `invalid`?

ribru17:
Not fully sure of the question, but if I understand it correctly you can just change `inline_control` to:

`inline_control: $=> seq('#', token.immediate(/ /), $.expression),`

and then just 

```query
; (in jinja_inline/highlights.scm, don't actually include this line)
(node) @capture
; ... etc. (put the normal queries here)
```

```query
; (in jinja/highlights.scm, don't actually include this line)
; inherits: inja_inline
"{%" @capture
; ... etc.
```

cathaysia:
oh, sorry. I give a bad example. The problem is `$.expression`.

for this example:

```jinja
{% for x 
in items
%}
```

This is a valid grammar, because **NEWLINE** can appears in the expression.

but for a inline statement:

```
# for x
in items
```

is invalid, because NEWLINE cannot appears in the expression.

ribru17:
Yeah that's fine how you have it now, I don't mean to change the grammar in that way. I just mean inherit the *queries*, so e.g. write one larger query file in jinja_inline then just inherit that file in the regular jinja highlights file

cathaysia:
Ok, I know your point. `inherit` a query is a way that I miss it. I cannot find it in the tree-sitter's document, it this a extension of nvim?

and it's grammar is:

> ; (in jinja/highlights.scm, don't actually include this line)
; inherits: inja_inline
"{%" @capture
; ... etc.

?

ribru17:
Exactly, you can also see `:h treesitter-query-modeline-inherits` or some examples in this repo: https://github.com/nvim-treesitter/nvim-treesitter/blob/824bf8455fd670186986fb3eb46fcaa4c5460949/queries/cpp/folds.scm#L1

cathaysia:
ready for merge

ribru17 (approved):

clason:
Needs injections at least for `comment`. And can you add at least basic folds?

cathaysia:
> Needs injections at least for comment.

Where I need these injections? And it works well now.
image

> And can you add at least basic folds

I am not familiar with fold. Currently, jinja only parses the expression, not the code block. So it may not be possible to add fold.

For example, the follow statements:

```jinja
{% if true %}
{% endif %}

```

will be parse as:

```schema
(definition 
  (control 
    (statement 
      (if_expression)))) 
(definition 
  (control 
    (statement)))) 
```

clason:
The comment parser is responsible for, e.g., highlighting "TODO" notes in comments. Look at other languages; virtually all have them.

cathaysia:
🥳

jamilraichouni:
First for working on support for jinja!
Quick note: Just updated nvim-treesitter and opened a template with nvim and got:

Screenshot 2025-02-09 at 05 12 07

after i closed neovim and tried again, this message did not appear again.

PR #7579, opened 2025-01-22
fix(nix): Shorter sub-attribute function highlight query

bodby:
See #7547.


queries/nix/highlights.scm
```diff
@@ -117,16 +117,9 @@
...
-        .
-        attr: (identifier) @function.call .))
-    (select_expression
-      attrpath: (attrpath
-        (_)
-        attr: (identifier) @function.call .))
-  ])
+  function: (select_expression
+    attrpath: (attrpath
+      attr: (identifier) @function.call .)))
```
ribru17, line 122: What about making the entire thing just

```
function: (select_expression
  attrpath: (attrpath
    attr: (identifier) @function.call .))
```

(one line shorter)

bodby: I thought you had to start with a node.

ribru17 (approved):

PR #7594, opened 2025-01-28
feat(djot): update djot queries

treeman:
Updates the djot grammar to breaking changes introduced in https://github.com/treeman/tree-sitter-djot/pull/56

### DIFF
--- a/queries/djot/highlights.scm
+++ b/queries/djot/highlights.scm
@@ -1,28 +1,28 @@
 (heading) @markup.heading
 
 ((heading
-  (marker) @heading.marker) @markup.heading.1
-  (#eq? @heading.marker "# "))
+  (marker) @_heading.marker) @markup.heading.1
+  (#eq? @_heading.marker "# "))
 
 ((heading
-  (marker) @heading.marker) @markup.heading.2
-  (#eq? @heading.marker "## "))
+  (marker) @_heading.marker) @markup.heading.2
+  (#eq? @_heading.marker "## "))
 
 ((heading
-  (marker) @heading.marker) @markup.heading.3
-  (#eq? @heading.marker "### "))
+  (marker) @_heading.marker) @markup.heading.3
+  (#eq? @_heading.marker "### "))
 
 ((heading
-  (marker) @heading.marker) @markup.heading.4
-  (#eq? @heading.marker "##### "))
+  (marker) @_heading.marker) @markup.heading.4
+  (#eq? @_heading.marker "##### "))
 
 ((heading
-  (marker) @heading.marker) @markup.heading.5
-  (#eq? @heading.marker "###### "))
+  (marker) @_heading.marker) @markup.heading.5
+  (#eq? @_heading.marker "###### "))
 
 ((heading
-  (marker) @heading.marker) @markup.heading.6
-  (#eq? @heading.marker "####### "))
+  (marker) @_heading.marker) @markup.heading.6
+  (#eq? @_heading.marker "####### "))
 
 (thematic_break) @string.special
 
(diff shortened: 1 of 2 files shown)


treeman:
Ping @NoahTheDuke

queries/djot/highlights.scm
```diff
@@ -1,14 +1,22 @@
-(heading1) @markup.heading.1
+(heading) @markup.heading
 
-(heading2) @markup.heading.2
+((heading) @markup.heading.1
+  (#match? @markup.heading.1 "^# "))
```
clason, line 4: Don't use `#match?` -- prefer `#eq?`, or `#lua-match?` if you must. (Would have been better to have kept the nodes in the grammar...)

clason:
Also, I feel this is a regression -- it makes much more sense to capture this in the grammar (like Markdown is doing).

clason:
I do appreciate the PR, by the way! ❤️

queries/djot/highlights.scm
```diff
@@ -148,50 +156,21 @@
...
 ] @string.special
 
-; We need to target tokens specifically because `{=` etc can exist as fallback symbols in
-; regular text, which we don't want to highlight or conceal.
-(highlighted
-  [
-    "{="
-    "=}"
-  ] @punctuation.delimiter
-  (#set! conceal ""))
```
clason, line 158: Is the removal of the conceals intentional?

treeman: They're moved into the other delimiter definitoins (not shown in the diff on GitHub):

```
([
  (emphasis_begin)
  (emphasis_end)
  (strong_begin)
  (strong_end)
  (superscript_begin)
  (superscript_end)
  (subscript_begin)
  (subscript_end)
  (highlighted_begin) <- used to be `{=`
  (highlighted_end)
  (insert_begin)
  (insert_end)
  (delete_begin)
  (delete_end)
  (verbatim_marker_begin)
  (verbatim_marker_end)
  (math_marker)
  (math_marker_begin)
  (math_marker_end)
  (raw_inline_attribute)
  (raw_inline_marker_begin)
  (raw_inline_marker_end)
] @punctuation.delimiter
  (#set! conceal ""))

```

queries/djot/highlights.scm
treeman: Hmm, I made this change to support arbitrary headings and I was quite pleased with how it simplified the grammar definition.. Previously we used the same setup as in markdown, but that was hardcoded to 6 heading levels.

Is it a big problem to use `#lua-match?` Something like this with `#eq?` might work too:

```
((heading
  (marker) @heading.marker) @markup.heading.1
  (#eq? @heading.marker "# "))
```

What would be the best design here in your opinion?

treeman: As an aside, it wasn't clear to me from Neovim's help pages if we should prefer `#match?` or `#lua-match?` in our queries. Not a big deal to change of course.

clason: If more than six levels are useful, then I see the point ;)

If you can get away with `#eq?`, that will help performance; if not, `#lua-match?` is OK, too (but `#eq?` is even faster). `#match?` spins up the Vim regex engine, which is _really_ slow by comparison.

queries/djot/highlights.scm
```diff
@@ -1,14 +1,28 @@
-(heading1) @markup.heading.1
+(heading) @markup.heading
 
-(heading2) @markup.heading.2
+((heading
+  (marker) @heading.marker) @markup.heading.1
+  (#eq? @heading.marker "# "))
```
clason, line 5:
```suggestion
  (marker) @_heading.marker) @markup.heading.1
  (#eq? @_heading.marker "# "))
```

(We dropped that as an official capture (for color schemes).)

treeman: Damn you're fast!

PR #7595, opened 2025-01-28
feat(typst): capture multilevel headings

DimitrisDimitropoulos:
In this PR multi-level headings are captured. Generally, the implementation is similar with markdown, which inspired typst to a great extend. Lastly, if there are any recommendations I am totally open.

### DIFF
--- a/queries/typst/highlights.scm
+++ b/queries/typst/highlights.scm
@@ -84,8 +84,23 @@
 ; text
 (text) @spell
 
-; TODO: use multi level headings
-(heading) @markup.heading
+(heading
+  "=" @markup.heading.1) @markup.heading.1
+
+(heading
+  "==" @markup.heading.2) @markup.heading.2
+
+(heading
+  "===" @markup.heading.3) @markup.heading.3
+
+(heading
+  "====" @markup.heading.4) @markup.heading.4
+
+(heading
+  "=====" @markup.heading.5) @markup.heading.5
+
+(heading
+  "======" @markup.heading.6) @markup.heading.6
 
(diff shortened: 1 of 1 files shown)


DimitrisDimitropoulos:
> Well, in Markdown these are captured by the grammar, not the query (which is the much better approach).

I didn't know that. I get the point you are trying to make, if you think this should not be included feel free to close this PR.

clason:
Basically, they just need to expose the aliases:
https://github.com/uben0/tree-sitter-typst/blob/26dfb4b94a99cc9d6044342b7ad0ba761da77d60/grammar.js#L168-L173

@uben0

ribru17:
Why not just something like:

```scm
(heading
  "===" @markup.heading.3
  (text) @markup.heading.3)
```

etc.? Seems like all the grammar info is already there

DimitrisDimitropoulos:
The grammar is indeed there, however it is not exposed and the CI throws some errors, in [this](https://github.com/nvim-treesitter/nvim-treesitter/commit/99582b45251f63d16ac3e3dccb3e28ff61711568#diff-c85e3f50e59be4e3d64b4733e75c1911a162cff5fc46a64e332ff585e6f06542R91) previous commit.

ribru17:
The nodes are exposed though, I think CI was failing because `@markup.heading.marker` is not a valid capture name. Can you try again this way?

clason:
Yes, because you used an invalid capture name.

DimitrisDimitropoulos:
I think now it's fine

PR #7604, opened 2025-01-31
docs(readme): use vim.uv instead of vim.loop

calebgl:
While checking my config, I stumbled into a warning 

```lua
local ok, stats = pcall(vim.loop.fs_stat, vim.api.nvim_buf_get_name(buf)) -- Undefined field fs_stat
```

Looking into the Neovim help documentation (`:h vim.loop`), I found the following note:

```txt
• *vim.loop*				Use |vim.uv| instead.
```

This PR updates the README to reflect this change and encourage using `vim.uv` instead of the deprecated `vim.loop`.


clason:
Thank you, but the README needs to be completely rewritten anyway (and `vim.loop` still works for quite a while). So I'd prefer not to make that change to avoid conflicts with the ongoing work on the `main` branch.

PR #7607, opened 2025-02-01
more perl syntax

rabbiveesh:
- **feat(perl): highlight newer nodes**
- **fix(perl): injections for s///e should fully re-parse**
- **feat(perl): bump the parser**

This is bundled with some new features i just did for the parser. only open question i
have is if globs (`*foo{THING}`) which represent package tables should get the highlight
for `variable` or `module`, b/c it kinda is a namespace (tho pivoted from the usual
meaning). For now i was conservative and made it variable

### DIFF
--- a/queries/perl/highlights.scm
+++ b/queries/perl/highlights.scm
@@ -243,70 +243,43 @@
   ; highlights punc vars and also numeric only like $11
   (#lua-match? @variable.builtin "^%A+$"))
 
-(scalar) @variable
-
-(scalar_deref_expression
-  [
-    "$"
-    "*"
-  ] @variable)
-
 [
+  (scalar)
   (array)
-  (arraylen)
+  (hash)
+  (glob)
+  ; arraylen's sigil is kinda special b/c it's not a data type
+  (arraylen
+    "$#" @operator)
 ] @variable
 
-(array_deref_expression
+; all post deref sigils highlighted as operators, and the unrolly star is a special char
+(postfix_deref
   [
+    "$"
     "@"
-    "*"
-  ] @variable)
-
-(arraylen_deref_expression
-  [
-    "$#"
-    "*"
-  ] @variable)
-
-(hash) @variable
-
-(hash_deref_expression
-  [
     "%"
     "*"
-  ] @variable)
+    "$#"
+  ] @operator
+  "*" @character.special)
 
+(slices
+  [
+    arrayref: _
+    hashref: _
+  ]
+  [
+    "@"
+    "%"
+  ] @operator)
+
+; except for subref deref, b/c that's actually a function call
 (amper_deref_expression
   [
     "&"
     "*"
-  ] @variable)
-
-(glob) @variable
-
-(glob_deref_expression
-  "*" @variable)
-
-(glob_slot_expression
-  "*" @variable)
-
-(array_element_expression
-  array: (_) @variable)
-
-(slice_expression
-  array: (_) @variable)
-
-(keyval_expression
-  array: (_) @variable)
-
-(hash_element_expression
-  hash: (_) @variable)
-
-(slice_expression
-  hash: (_) @variable)
-
-(keyval_expression
-  hash: (_) @variable)
+  ] @function.call)
 
 ; mark hash or glob keys that are any form of string in any form of access
 (_
--- a/queries/perl/highlights.scm
+++ b/queries/perl/highlights.scm
@@ -278,13 +278,15 @@
 
 (amper_deref_expression
   [
-   "&"
-   "*"
-   ] @variable)
+    "&"
+    "*"
+  ] @variable)
 
 (glob) @variable
+
 (glob_deref_expression
   "*" @variable)
+
 (glob_slot_expression
   "*" @variable)
 
--- a/queries/perl/highlights.scm
+++ b/queries/perl/highlights.scm
@@ -262,6 +262,12 @@
     "*"
   ] @variable)
 
+(arraylen_deref_expression
+  [
+    "$#"
+    "*"
+  ] @variable)
+
 (hash) @variable
 
 (hash_deref_expression
@@ -269,6 +275,18 @@
     "%"
     "*"
   ] @variable)
+
+(amper_deref_expression
+  [
+   "&"
+   "*"
+   ] @variable)
+
+(glob) @variable
+(glob_deref_expression
+  "*" @variable)
+(glob_slot_expression
+  "*" @variable)
 
 (array_element_expression
   array: (_) @variable)
--- a/queries/perl/highlights.scm
+++ b/queries/perl/highlights.scm
@@ -308,6 +308,20 @@
 (keyval_expression
   hash: (_) @variable)
 
+; mark hash or glob keys that are any form of string in any form of access
+(_
+  "{"
+  [
+    (autoquoted_bareword)
+    (_
+      (string_content))
+  ] @variable.member
+  "}")
+
+; mark stringies on the LHS of a fat comma as a hash key, b/c that's usually what it
+; denotes somewhat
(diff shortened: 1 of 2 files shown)


queries/perl/highlights.scm
```diff
@@ -262,6 +262,12 @@
     "*"
   ] @variable)
 
+(arraylen_deref_expression
+  [
+    "$#"
+    "*"
+  ] @variable)
```
clason, line 269: That looks wrong to me. Can you give some context examples?

clason: (These are usually highlighted as `@operator`.)

rabbiveesh: yes, it's postfix array length `$arrayref->$#*`.

the glob deref ones are `$globref->*{SCALAR}` and `$globref->**` too

queries/perl/highlights.scm
```diff
@@ -270,6 +276,20 @@
     "*"
   ] @variable)
 
+(amper_deref_expression
+  [
+    "&"
+    "*"
+  ] @variable)
+
+(glob) @variable
```
clason, line 285: Same for globs in regular expressions.

queries/perl/highlights.scm
clason: Then `@property`  is more appropriate, I think.

queries/perl/highlights.scm
rabbiveesh: ya, as i posted above, this isn't globs as in fileglob, this is a horrific perl construct called a glob, which represents a symbol table for a particular global variable (for example `*foo{SCALAR}` is a different way to get at the global variable `$foo`).

clason: That still sounds like an `@operator` to me?

rabbiveesh: ya, i'm not really sure how to think about it; the rest of the sigils i have highlighted as `variable`, b/c it's more useful to conceptualize them as part of the variable's name.
I guess with globs we are doing a different thing; so you would vote the `*` gets `@operator`, the identifier gets `@variable` and the `{SCALAR}` gets `@property`?

clason: That's what sounds most appropriate to me, given how other languages operate. (The point here is to group syntax elements into cross-language semantic capture categories, _not_ to assign "matching colors" directly -- that's the job of the colorscheme, which can special-case languages with the `@operator.perl` specialization.)

I'll let @ribru17 weigh in, though.

rabbiveesh: i thought about it more, and i disagree.
the following is a valid usage of a glob in perl `*glob = *some_other_glob` - in this operation, the `*` is just acting as a sigil, and thus is just part of the variable name (since we do not in general treat sigils as operators).

I think it makes sense to leave it the way that it is, tho i give you that `@property` makes sense for the `{SCALAR}` part. given that, i guess it also makes sense to do the same for regular hash access.

queries/perl/highlights.scm
rabbiveesh: i've been thinking about this and also i disagree; you don't conceptualize sigils as an operator, and you don't view deref as accessing a property. it just doesn't fit the mental model (i believe we had a similar discussion when i first commited my parser)

rabbiveesh:
Hey, i've left my arguments in the comments, and i have also added in highlighting of hash keys in both defiinition and access as `@variable.member` (i saw that ecma uses that instead of `@property`)

queries/perl/highlights.scm
ribru17: Imho all the sigil characters should be `@character.special` or `@punctuation.special`; let me know if you disagree (I know nothing about perl), but stackoverflow seems to think that they are not to be considered part of the variable name, but rather show contex

ribru17: Also if we do this we should only highlight `(glob (varname) @variable)`

queries/perl/highlights.scm
ribru17: Or maybe just `$#` as operator and `*` as a sigil (something `.special`) is more accurate?

queries/perl/highlights.scm
rabbiveesh: i disagree about sigils in general; as a perl programmer we do think of them as part of the variable's name (even tho it's more complicated than that; in definitions it's certainly part of the name; when using a variable it's more complicated and it acts more as an operator, but i would say most people treat it as part of the name.

in any case, the other languages that use sigils, like ruby + php, both seem to keep them as part of an atomic token with the variable name.

Previous discussion on this topic starts around here https://github.com/nvim-treesitter/nvim-treesitter/pull/5222#issuecomment-1686935277

rabbiveesh: i'm mostly against it, but i think i could get used to it.

I do think that it would be a separate issue to pursue and shouldn't block this PR (unless you specifically want the conceptual distinction with the `->**` business)

clason: (I'm hesitant to just go with your proposal since "think of them as" and "treat it as" implies that they are in fact not, and respecting the actual language syntax -- and providing cross-language abstractions for that -- are the project goals here.)

ribru17: True. But I am also fine to not let that block this PR. In my opinion this is good to go once the array postfix length gets some sort of operator indication

clason:
If you prefer, you could pull your uncontroversial changes out into a separate PR (or vice versa) so those can get merged more quickly. I don't want to drag this out unnecessarily either way, though.

queries/perl/highlights.scm
rabbiveesh: i think that as was discussed in the other PR, either all sigils should get indicated or none; up until now we opted for none + it seems like we'd like some, so i think i'll just quickly update all sigils.

We've settled on `@character.special` then

clason: > i think that as was discussed in the other PR, either all sigils should get indicated or none

I agree; and if the other occurrences are more clearly not to be distinguished, I am also OK with none.

(It's unfortunate that such reasoned decisions are documented -- poorly -- on Github PR discussions. I'll have to think how to best record "language query design strategies" for better reference. Perl certainly needs it ;))

rabbiveesh:
i've been playing around with some different ideas (colorscheme is `evening`)

image
here's highlighting postfix sigils as `@operator` and the star afterwards as `@character.special` (which makes the most sense to me)

This includes prefix sigils that are "doing" soimething as `operator` (arraylen and all forms of hash/array acess) - i'm on the fence on if this goes too far.
image

And here it is with ALL sigils marked as `operator` - it doesn't sit well b/c even in declaration they're marked as operators, and there they don't do anything
image

And here's that walked back one drop to not do it do declaration
image

lmk which you guys think makes the most sense @clason and @ribru17

rabbiveesh:
@ribru17 could i get your opinion on what i posted above? don't want this to just languish here

ribru17:
Personally I think the first option looks best :+1: though I noticed `&*` never got highlights, is this related to the supertypes bug you mentioned a while back?

rabbiveesh:
> Personally I think the first option looks best 👍 though I noticed `&*` never got highlights, is this related to the supertypes bug you mentioned a while back?

no; `&*` is actually a function call so it highlights as `function.call`

rabbiveesh:
Sorry, I haven't gotten back to this since you last commented bc werk. I
will push those changes soon

rabbiveesh:
sorry for the delay; i've pushed the requested changes, including the `$#` of arraylen as operator, similar to lua

clason:
Thanks!

rabbiveesh:
thank you for all the careful consideration you have for quality editing
experience.

PR #7608, opened 2025-02-02
feat(hyprlang): bash injection for bindings with descriptions

### LINKED ISSUE #7596: hyprlang: edge case in bash injections for exec keybinding
The `bind` keyword for hyprland usualy takes 4 arguments and the bash injection works fine here
```
bind = MODS, key, exec, params
```

But its takes 5 arguments when adding the [description flag](https://wiki.hyprland.org/Configuring/Binds/#description). In this case bash injection fails
```
bindd = MODS, key, description, exec, params
```
---
One thing I think I need to mention is that when I tried the [injections.scm](https://github.com/tree-sitter-grammars/tree-sitter-hyprlang/blob/master/queries/hyprlang/injections.scm) query from there, it workded as expected

ValdezFOmar:
Fixes #7596 by adding an optional node. There is still an edge case if the keybind description is exactly `exec`, but I don't expect people to write their config like that. Solved by anchoring the injection injection to the last node.

```
# edge case
bindd = MODS, key, exec, exec, params
```

### DIFF
--- a/queries/hyprlang/injections.scm
+++ b/queries/hyprlang/injections.scm
@@ -13,6 +13,8 @@
     .
     (_)
     .
+    (_)?
+    .
     (string) @_exec
     .
     (string) @injection.content))


queries/hyprlang/injections.scm
```diff
@@ -8,11 +8,6 @@
...
   (name) @_bind
   (params
-    .
-    (_)
-    .
-    (_)
-    .
     (string) @_exec
     .
     (string) @injection.content))
```
Juhan280, line 13: Can't we prevent that edge case by anchoring the bash injection to the last node?

```suggestion
    (string) @injection.content .))
```

Juhan280:
Btw, do we actually need to relax the injection query? Can't we get away by add one extra optional node? This way if anybody wrote less arguments it the injection will not happen. Just like the current behaviour.

```diff
(params
    . 
    (_)
    . 
    (_) 
    . 
+   (_)? 
+   . 
    (string) @_exec
```

PR #7615, opened 2025-02-03
fix(gleam): indent on unqualified imports rather than import

rawhat:
The current `import` indent causes an indentation to trigger after a line with `import ...`.  I _think_ the more correct is to only begin indentation when doing an `unqualified_imports`, i.e. 

```gleam
// don't  indent from _just_ this
import my_library.{
  // but indent for this
  type Something
}
```

Obviously could very well be wrong about this, but I've been getting indentation bugs with what's on `master`.  It does
```gleam
import gleam/result
  |
```
which feels wrong.

### DIFF
--- a/queries/gleam/indents.scm
+++ b/queries/gleam/indents.scm
@@ -7,7 +7,6 @@
   (constant)
   (external_function)
   (function)
-  (import)
   (let)
   (list)
   (constant)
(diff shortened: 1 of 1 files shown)


clason:
@amaanq

clason:
No maintainer, so let's just merge it and see who else complains. (Indents are best-effort only, after all.)

PR #7624, opened 2025-02-06
Bug-fix cmdpath function in shell_command_selectors.lua

chrispetkau:
Surround all paths submitted to the command line with double-quotes so as to accommodate spaces in those paths.

### DIFF
--- a/lua/nvim-treesitter/shell_command_selectors.lua
+++ b/lua/nvim-treesitter/shell_command_selectors.lua
@@ -7,7 +7,7 @@ local uv = vim.uv or vim.loop
 ---@param p string
 ---@return string
 local function cmdpath(p)
-  p = '\"' .. p .. '\"'
+  p = '"' .. p .. '"'
   if vim.opt.shellslash:get() then
     local r = p:gsub("/", "\\")
     return r


clason:
What concrete bug is fixed by this?

chrispetkau:
For each parser that is not yet installed I get a log message like below for each, on startup. I have 'ensure_installed' and 'auto_install' set in the treesitter settings.
If I open a file that uses a parser, the parser does get installed (e.g. open an .rs and the Rust parser gets installed).
The problem seems to be the space in the path in cwd.

> Downloading tree-sitter-query...
> fatal: destination path 'tree-sitter-query' already exists and is not an empty directory.
> Error during download, please verify your internet connection
> Failed to execute the following command:
> {
>   cmd = "git",
>   err = "Error during download, please verify your internet connection",
>   info = "Downloading tree-sitter-query...",
>   opts = {
>     args = { "clone", "https://github.com/neovim/tree-sitter-vimdoc", "tree-sitter-vimdoc", "--filter=blob:none", },
>     cwd = "C:\\Users\\Chris Petkau\\AppData\\Local\\nvim-data"
>   }
> }

PR #7626, opened 2025-02-07
feat: add enforce

simonvic:
Add `enforce` parser.

> Enforce Script is the language that is used by the [Enfusion](https://community.bistudio.com/wiki/Enfusion) engine first introduced in [DayZ](https://community.bistudio.com/wiki/Category:DayZ) Standalone. It is a Object-Oriented Scripting Language (OOP) that works with objects and classes and is similar to C# programming language.

[Official documentation](https://community.bistudio.com/wiki/DayZ:Enforce_Script_Syntax)

### DIFF
--- a/queries/enforce/highlights.scm
+++ b/queries/enforce/highlights.scm
@@ -1,27 +1,18 @@
-(comment_line) @comment @spell
-(comment_block) @comment.block @spell
+[
+ (comment_line)
+ (comment_block)
+ ] @comment @spell
 
-(doc_line) @comment.documentation @spell
-(doc_block) @comment.documentation.block @spell
+[
+ (doc_line)
+ (doc_block)
+ ] @comment.documentation @spell
 
 (literal_bool) @boolean
 (literal_int) @number
 (literal_float) @number.float
 (literal_string) @string
 (escape_sequence) @string.escape
-
-
-; vectors
-; TODO: what about arrays of vectors?
-(decl_variable
-  (type_vector)
-  (literal_string) @vector
-  )
-
-(formal_parameter
-  type: (type_vector)
-  default: (literal_string) @vector
-  )
 
 (identifier) @variable
 
@@ -118,34 +109,11 @@
 
 ; Constructor and deconstructor (function with same name of the class)
 (decl_class
-  typename: (identifier) @foo
+  typename: (identifier) @_classname
   body: (class_body
     (decl_method
-	    name: (identifier) @constructor
-	    (#eq? @constructor @foo)
+      name: (identifier) @constructor
+      (#eq? @constructor @_classname)
       )
     )
   )
-
-; TODO: mark invalid deconstructor as error?
-(decl_class
-  typename: (identifier) @foo
-  body: (class_body
-    (decl_method
-      "~"
-	    name: (identifier) @constructor.deconstructor
-	    (#eq? @constructor.deconstructor @foo)
-      )
-    )
-  )
-
-; Dead code
-(block
-  (_)*
-  (return)
-  (_)* @deadcode (#set! "priority" 110)
-  )
-
-(ERROR) @error
-
-; TODO: string and print format injection
--- a/queries/enforce/highlights.scm
+++ b/queries/enforce/highlights.scm
@@ -113,9 +113,9 @@
 ] @keyword
 
 [
- "new"
- "delete"
- ] @keyword.operator
+  "new"
+  "delete"
+] @keyword.operator
 
 "return" @keyword.return
 
--- a/queries/enforce/highlights.scm
+++ b/queries/enforce/highlights.scm
@@ -20,6 +20,9 @@
 
 (identifier) @variable
 
+(formal_parameter
+  name: (identifier) @variable.parameter)
+
 ((identifier) @constant
   (#lua-match? @constant "^[A-Z_][A-Z%d_]+$"))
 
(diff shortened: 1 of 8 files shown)


clason:
Please read https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md. You **cannot** use any but the listed captures.

queries/enforce/highlights.scm
```diff
@@ -0,0 +1,177 @@
...
+;   )
+
+; Dead code
+(block
+  (_)*
+  (return)
+  (_)* @deadcode (#set! "priority" 110)
+  )
+
+(ERROR) @error
```
clason, line 175:
```suggestion

```
We don't capture `ERROR` nodes for highlighting.

clason:
How stable is the parser? If it's still under development, we'd prefer to wait until it is done before adding it.

simonvic:
> How stable is the parser? If it's still under development, we'd prefer to wait until it is done before adding it.

The parser should be ready; it successfully parses the entire DayZ code base (~150k loc).
Though, I should mention that, given the many quirks of the language and its lack of documentation, this parser could be not 100% up to spec (e.g. in some contexts the real parser could allow unbalanced parenthesis, missing punctuation etc.)

clason:
That's ok; the point is that it should parse correct code correctly; we don't care about (and don't want to invest any effort on) what it does with incorrect code.

ribru17:
Also please format the query files :) See https://github.com/nvim-treesitter/nvim-treesitter/blob/master/scripts/format-queries.lua or https://github.com/ribru17/ts_query_ls

queries/enforce/highlights.scm
```diff
@@ -0,0 +1,177 @@
...
+  invoked: (identifier) @function.method.call
+  )
+
+; Constructor and deconstructor (function with same name of the class)
+(decl_class
+  typename: (identifier) @foo
+  body: (class_body
+    (decl_method
+	    name: (identifier) @constructor
+	    (#eq? @constructor @foo)
```
clason, line 127: Prefix "private" captures with underscore if they are not meant to be highlighted but only used in predicates.

clason:
Gotta lint ALL THE QUERIES ;)

simonvic:
oops, now it should be good :D
I cant't make the `format-queries.lua` script work, so I've used the language server for now :)

Thanks for the very quick answer and help btw!
*(this is actually my very first public PR)*

queries/enforce/highlights.scm
```diff
@@ -0,0 +1,183 @@
...
+  (method_modifier)
+  (class_modifier)
+  (field_modifier)
+  (formal_parameter_modifier)
+] @keyword.modifier
+
+"ref" @type
+
+(decl_class
+  typename: (identifier) @type)
```
ribru17, line 155: Maybe also:

```scheme
(decl_class
  superclass: (superclass
    typename: (identifier) @type))
```

queries/enforce/highlights.scm
```diff
@@ -0,0 +1,183 @@
...
+
+(type_identifier
+  (identifier) @type)
+
+(type_primitive) @type.builtin
+
+[
+  (super)
+  (this)
+  (literal_null)
```
ribru17, line 168: Most places have `NULL` has `@constant.builtin`, so I think this should be changed, unless you feel variable is better

queries/enforce/highlights.scm
```diff
@@ -0,0 +1,183 @@
...
+  "continue"
+  "break"
+  "switch"
+  "case"
+  "typedef"
+  "delete"
+  "default"
+  "extends"
+  "new" ; TODO: is it operator?
+  "auto" ; TODO: is it type?
```
ribru17, line 123: C++ uses `@type.builtin`, maybe we should follow that example

queries/enforce/highlights.scm
```diff
@@ -0,0 +1,183 @@
...
+[
+  "continue"
+  "break"
+  "switch"
+  "case"
+  "typedef"
+  "delete"
+  "default"
+  "extends"
+  "new" ; TODO: is it operator?
```
ribru17, line 122: Typically `@keyword.operator` for this

queries/enforce/highlights.scm
```diff
@@ -0,0 +1,183 @@
...
+    "\""
+    "\""
+  ] @punctuation.delimiter)
+
+[
+  "continue"
+  "break"
+  "switch"
+  "case"
+  "typedef"
```
ribru17, line 118: This should be `@keyword.type`

queries/enforce/highlights.scm
```diff
@@ -0,0 +1,183 @@
...
+  [
+    "\""
+    "\""
+  ] @punctuation.delimiter)
+
+[
+  "continue"
+  "break"
+  "switch"
+  "case"
```
ribru17, line 117: These two should be `@keyword.conditional`

queries/enforce/highlights.scm
```diff
@@ -0,0 +1,183 @@
...
+
+(literal_string
+  [
+    "\""
+    "\""
+  ] @punctuation.delimiter)
+
+[
+  "continue"
+  "break"
```
ribru17, line 115: These two should be `@keyword.repeat`

queries/enforce/indents.scm
```diff
@@ -0,0 +1,30 @@
+[
+  (block)
```
ribru17, line 2: This one is tricky: I think instead of having `(block)` here we should have `(decl_method)`, `(for)`, `(if)`, and etc. for anything that can have a `block` within it. As it stands currently, indentation for a method like this:

```java
    void PM_Merchant_Base()
    {
        TStringArray attachmentSlots = new TStringArray;
        GetAttachmentSlots(this, attachmentSlots);
        for (int i = 0; i < attachmentSlots.Count(); i++)
        {
            m_MerchantSlots.Set(attachmentSlots.Get(i), InventorySlots.GetSlotIdFromString(attachmentSlots.Get(i)));
        }
    }
```

becomes this:

```java
    void PM_Merchant_Base()
{
        TStringArray attachmentSlots = new TStringArray;
        GetAttachmentSlots(this, attachmentSlots);
        for (int i = 0; i < attachmentSlots.Count(); i++)
    {
            m_MerchantSlots.Set(attachmentSlots.Get(i), InventorySlots.GetSlotIdFromString(attachmentSlots.Get(i)));
        }
    }
```

due to the opening braces `{` starting on another line

simonvic: good point... I'll have to take a closer look at this and do some more testing since `block` is also allowed as a statement i.e.
```cs
class Foo {
	void foo() {
		// foo
		{
			// foo
		}
		// foo
		{
			// foo
		}
	}
}

```

ribru17 (approved):

PR #7630, opened 2025-02-08
Add check for locked parser on Windows

GPetrites:
On Windows, as soon as parsers are loaded into Neovim, a lock is placed on the folder containing the parser. Subsequent efforts to update the parser will delete the contents of the folder but will fail to delete the folder, leaving the parser unusable. The only way to release the locks is by closing all terminals used to run Neovim.

This change adds a simple mechanism to test if a parser folder is locked prior to performing the delete and install by attempting to rename the folder if it exists. If the rename fails, a warning is logged and the install is cancelled, leaving the parser in its current state.

- Implemented a check for locked folders on Windows in `run_install` function.
- Introduced `select_install_test_lock` function to generate commands for testing folder locks on Windows.
- Added optional `hide_errors` parameter to `iter_cmd_sync` function to allow suppressing error messages during lock tests.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -286,7 +286,8 @@ end
 
 ---@param cmd_list Command[]
 ---@return boolean
-local function iter_cmd_sync(cmd_list)
+local function iter_cmd_sync(cmd_list, hide_errors)
+  hide_errors = hide_errors or false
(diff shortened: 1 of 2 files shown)


clason:
Thank you, but this is already fixed on the `main` branch; `master` is pretty much frozen except for really small bugfixes (which this is not).

PR #7631, opened 2025-02-08
feat(cpp): add support for modules

jdrouhard:


clason:
It's not that simple. (And, if I may be so bold, your uncoordinated tree-sitter-c change and release are ill-received.)

jdrouhard:
> It's not that simple. (And, if I may be so bold, your uncoordinated tree-sitter-c change and release are ill-received.)

Ok? I apologized to @amaanq and realized I made a blunder... this PR is just some real simple extensions to the existing query files. Closing this just because I hurt some feelings doesn't seem like a great approach, either.

clason:
No, I am closing it because it doesn't work, as you can see from the failing tests.

jdrouhard:
Ok, I'll open a new PR when the dependent parsers are updated to use a more recent cpp grammar. Sorry again.

PR #7632, opened 2025-02-09
feat(java): add injections for markdown documentation comments

rmuir:
Since Java 23, java documentation supports markdown.

It is supported by a `///` line comment followed by mandatory white space: https://openjdk.org/jeps/467

Before:
markdown_before

After:
markdown_after

I wasn't able to figure out a way to use `injection.combined` alongside `#offset!` such that the prefix was removed from ALL lines in the "block" (not just the first one) to treat as one markdown document, which is what I wanted. As a workaround it uses `markdown_inline` for now, which seems ok a start. Any ideas?

Thank you for nvim-treesitter.

### DIFF
--- a/queries/java/injections.scm
+++ b/queries/java/injections.scm
@@ -8,6 +8,18 @@
   (#lua-match? @injection.content "/[*][!<*][^a-zA-Z]")
   (#set! injection.language "doxygen"))
 
+; markdown-style javadocs: https://openjdk.org/jeps/467
+((line_comment) @injection.content
+  (#lua-match? @injection.content "^///%s")
+  (#offset! @injection.content 0 4 0 0)
+  (#set! injection.language "markdown_inline"))
+
+; markdown-style javadocs: https://openjdk.org/jeps/467
+((line_comment) @injection.content
+  (#lua-match? @injection.content "^///%s+[@]")
+  (#offset! @injection.content 0 4 0 0)
+  (#set! injection.language "doxygen"))
+
 ((method_invocation
   name: (identifier) @_method
   arguments: (argument_list


ribru17 (approved):
LGTM, I've opened https://github.com/tree-sitter/tree-sitter/issues/4227 in response to your comment

rmuir:
@ribru17 thank you for debugging it! I'm a newbie, but it seems from the linked issue that I need to ensure a newline is included? I will fight with the injections queries again, to see if there is any way to solve it with that in mind. Combined injection would be really good, so that e.g. fenced code snippets and other multi-line elements in the documentation display correctly.

ribru17:
Sorry for getting back to you so late, but yes the newline fixes it for me! Something like this was mostly working:

```query
; markdown-style javadocs: https://openjdk.org/jeps/467
((line_comment) @injection.content
  (#lua-match? @injection.content "^///%s")
  (#offset! @injection.content 0 4 0 1)
  (#set! injection.combined)
  (#set! injection.language "markdown"))
```

the change to `#offset!` to include the newline made everything good for the most part. But check it yourself, I got some `end_col out of bounds` errors in the highlighter, not sure if it was due to this or to me going overboard with some other stress testing I was doing at the time...

rmuir:
> the change to `#offset!` to include the newline made everything good for the most part. But check it yourself, I got some `end_col out of bounds` errors in the highlighter, not sure if it was due to this or to me going overboard with some other stress testing I was doing at the time...

@ribru17 Same behavior here: this solves the multi-line markdown problem but brings in bounds-checks errors. trying to inspect the area with `InspectTree` will reproduce the bounds-check error every time (and prevent inspecting)

ribru17:
Dang, so annoying. We really need a way to tell the set_extmarks call to ignore the newline, while still keeping it as part of the node range...

rmuir:
It seems to me the newline is only a partial solve, that works for the following line.  You can see problems easily with a two-line code block in java:
```
/// Example:
/// ```java
/// String x = "foobar";
/// x.hashCode();
/// ```
```

I believe you encounter the same problem, if you wanted to apply highlighting to lua documentation in the neovim source code itself. For example if i open `languagetree.lua` with this injection:
```
; markdown in lua documentation (just the --- case)
(comment
   content: (_) @injection.content
  (#lua-match? @injection.content "^[-]%s")
  (#offset! @injection.content 0 1 0 1)
  (#set! injection.combined)
  (#set! injection.language "markdown"))
```

You will think you have it solved with lua snippets highlighted wonderfully, but then add a second line of lua to one of them.

ribru17:
I believe that is another issue actually... related to #21310. I am looking into reviving that PR, I have a development branch now which fixes the issue you say. I'll aim to have it up soon

PR #7650, opened 2025-02-13
feat(nu): new attribute syntax and add folds.scm

blindFS:
Add support of the new attribute syntax.

https://github.com/nushell/tree-sitter-nu/pull/182

### DIFF
--- a/queries/nu/highlights.scm
+++ b/queries/nu/highlights.scm
@@ -278,19 +278,12 @@ key: (identifier) @property
 (param_cmd
   (cmd_identifier) @string)
 
-(param_long_flag
-  (long_flag_identifier) @attribute)
-
-(param_short_flag
-  (param_short_flag_identifier) @attribute)
-
-(short_flag
-  (short_flag_identifier) @attribute)
-
-(long_flag_identifier) @attribute
-
-(attribute
-  (attribute_identifier) @attribute)
+[
+  (attribute_identifier)
+  (long_flag_identifier)
+  (param_short_flag_identifier)
+  (short_flag_identifier)
+] @attribute
 
 (scope_pattern
   (wild_card) @function)
--- a/queries/nu/highlights.scm
+++ b/queries/nu/highlights.scm
@@ -244,6 +244,9 @@ file_path: (val_string) @variable.parameter
 (returns
   "->" @punctuation.delimiter)
 
+(attribute
+  "@" @punctuation.special)
+
 [
   "("
   ")"
@@ -285,6 +288,9 @@ key: (identifier) @property
   (short_flag_identifier) @attribute)
 
 (long_flag_identifier) @attribute
+
+(attribute
+  (attribute_identifier) @attribute)
 
 (scope_pattern
   (wild_card) @function)
--- a/queries/nu/highlights.scm
+++ b/queries/nu/highlights.scm
@@ -244,9 +244,6 @@ file_path: (val_string) @variable.parameter
 (returns
   "->" @punctuation.delimiter)
 
-(attribute
-  "@" @attribute)
-
 [
   "("
   ")"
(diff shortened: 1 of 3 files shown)


ribru17:
In other languages we capture `"@"` also as `@attribute` rather than `@punctuation.special`, but other than that LGTM

queries/nu/highlights.scm
```diff
@@ -286,6 +289,9 @@ key: (identifier) @property
 
 (long_flag_identifier) @attribute
 
+(attribute
+  (attribute_identifier) @attribute)
+
```
clason, line 294: Can't you just do `(attribute) @attribute` then?

blindFS: Do you mean `(attribute_identifier) @attribute` ?

clason: No, I meant what I wrote. Asked another way: what is left of the `(attribute)` node that you are _not_ capturing in these two patterns?

(It's always good to show an example -- code snippet, parsed tree, and before/after screenshot -- for any proposed query change.)

blindFS: @clason 
image

The `identifier` here is `example`, the rest is also in node `attribute`

blindFS:
@ribru17 Thanks for the tip, always good to follow the convention, changed in the last commit.

queries/nu/highlights.scm
clason: Then probably
```lisp
(attribute
  "@" @attribute
  (attribute_identifer) @attribute)
```
(please test first). If that doesn't work, at least keep the two patterns next to each other.

blindFS: That makes sense, one problem, there are similar issues for `long_flag` , `short_flag`,  `param_long_flag` etc. Do I need to change them in this PR?

clason: No; if you want to clean them up, that would be welcome, but should be a separate PR.

clason: Oh, unless you're talking about the `"@"` highlight not being consistently applied otherwise.

In that case, it would be better to refactor things here -- but I think then the right approach would be to do
```
(attribute
  "@" @attribute)
  
[
  (attribute_identifier)
  (long_flag_identifier)
  ...
] @attribute
```

(assuming this still does the right thing).

queries/nu/highlights.scm
```diff
@@ -286,6 +289,9 @@ key: (identifier) @property
 
 (long_flag_identifier) @attribute
```
clason, line 290: Thinking about it; isn't this redundant with line 282?

blindFS: Definitely, there's a lot to refactor.

clason: Then it would be appreciated, if you're up for it. (Not the whole query, just the `@attribute` patterns.)

If not, that's ok too, but then keep the new `"@"` pattern next to the other `@attribute` ones. (I am assuming that you're not relying on the ordering here to have this overwritten later.)

queries/nu/highlights.scm
```diff
@@ -275,16 +278,12 @@ key: (identifier) @property
...
-(short_flag
-  (short_flag_identifier) @attribute)
-
-(long_flag_identifier) @attribute
+[
+  (attribute_identifier)
+  (long_flag_identifier)
+  (param_short_flag_identifier)
+  (short_flag_identifier)
+] @attribute
```
blindFS, line 287: @clason Does this look good?

clason: Almost (assuming you've tested against regressions, which I cannot since I don't know the language).

But the `"@"` should still be moved to after line 280.

blindFS: done, feels odd to me though, I mean why anonymous nodes are not put together.

blindFS: cool, I'm OK with any way.

ribru17 (approved):

PR #7657, opened 2025-02-15
feat: added tree-sitter-kcl

bertbaron:
This PR adds highlighting support for the [KCL language](https://www.kcl-lang.io/)

Please let me know if any other action is needed to get this added to the project.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1211,6 +1211,15 @@ list.just = {
   maintainers = { "@Hubro" },
 }
 
+list.kcl = {
+  install_info = {
+    url = "https://github.com/kcl-lang/tree-sitter-kcl",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
+  },
+  maintainers = { "@bertbaron"},
+}
+
 list.kconfig = {
   install_info = {
     url = "https://github.com/amaanq/tree-sitter-kconfig",
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1215,7 +1215,6 @@ list.kcl = {
   install_info = {
     url = "https://github.com/kcl-lang/tree-sitter-kcl",
     files = { "src/parser.c", "src/scanner.c" },
-    branch = "main",
   },
   maintainers = { "@bertbaron"},
 }
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1216,7 +1216,7 @@ list.kcl = {
     url = "https://github.com/kcl-lang/tree-sitter-kcl",
     files = { "src/parser.c", "src/scanner.c" },
   },
-  maintainers = { "@bertbaron"},
+  maintainers = { "@bertbaron" },
 }
 
 list.kconfig = {
--- a/queries/kcl/highlights.scm
+++ b/queries/kcl/highlights.scm
@@ -1,0 +1,166 @@
+(identifier) @variable
+
+; Reset highlighting in string interpolations
+(interpolation) @none
+
+(import_stmt
+  (dotted_name
+    (identifier) @namespace))
+
+(import_stmt
+  (dotted_name
+    (identifier) @namespace)
+  (identifier) @namespace)
+
+(basic_type) @type
+
+(schema_type
+  (dotted_name
+    (identifier) @type))
+
+(schema_type
+  (dotted_name
+    (identifier) @namespace
+    (identifier) @type))
+
+(schema_expr
+  (identifier) @type)
+
+(protocol_stmt
+  (identifier) @type)
+
+(rule_stmt
+    (identifier) @type)
+
+(schema_stmt
+  (identifier) @type)
+
+(lambda_expr
+  (typed_parameter (identifier) @parameter))
+
+(lambda_expr
+  (identifier) @parameter)
+
+(selector_expr
+  (select_suffix
+    (identifier) @property))
+
+(comment) @comment
+(string) @string
+(escape_sequence) @string.escape
+
+(schema_stmt
+  body: (block
+    .
+    (string
+      (string_content) @string.documentation)))
+
+(decorator
+  (identifier) @attribute)
+
+(call_expr
+  function: (identifier) @function)
+
+(call_expr
+  function: (selector_expr
+    (select_suffix
+      (identifier) @function)))
+
+[
+  (integer)
+  (float)
+] @number
+
+[
+  (true)
+  (false)
+  (none)
+  (undefined)
+] @constant.builtin
+
+[
+  "all"
+  "any"
+  "assert"
+  "as"
+  "check"
+  "elif"
+  "else"
+  "filter"
+  "for"
+  "if"
+  "import"
+  "lambda"
+  "map"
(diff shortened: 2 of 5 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -1211,6 +1211,15 @@ list.just = {
   maintainers = { "@Hubro" },
 }
 
+list.kcl = {
+  install_info = {
+    url = "https://github.com/kcl-lang/tree-sitter-kcl",
+    files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
```
clason, line 1218: Not needed since it's the default branch.

You do need to add a lockfile entry, though.

queries/kcl/highlights.scm
```diff
@@ -0,0 +1,166 @@
+(identifier) @variable
+
+; Reset highlighting in string interpolations
+(interpolation) @none
```
clason, line 4: This does not work in Neovim. (You cannot use upstream queries unchanged; please see CONTRIBUTING.MD.)

queries/kcl/injections.scm
```diff
@@ -0,0 +1,33 @@
...
+    .
+    (string
+      (string_content) @injection.content))
+  (#eq? @_regex "regex")
+  (#set! injection.language "regex"))
+
+((comment) @injection.content
+  (#set! injection.language "comment"))
+
+; Unfortunately, rst doesn't work with the indented docstrings, markdown_inline at least provides some highlighting
```
clason, line 27: Sorry, but that's a bad idea. All manner of things use injection queries, and rely on the language being correct.

clason:
You can run `./scripts/format-queries.lua` (or the ts_query_ls language server) to format the queries.

queries/kcl/injections.scm
bertbaron: Ok, I removed it. It's a shame that #2212 is not solved, all auto-generated docstrings (from imported xrd's etc.) are in Restructured Text but injection does not work because of the indentation. Markdown_inline works actually quite well due to the similarities between rst and markdown although it would only highlight part of it (italic, bold, links, ... but not headings)

queries/kcl/highlights.scm
bertbaron: Thanks for the quick response. I actually wrote the upstream query file myself and tested with Neovim and it seem to work quite well, but I changed it to only use the documented tags.

queries/kcl/injections.scm
clason: If it's rst, it should be injected as rst and nothing else. I don't quite see the relevance of the linked issue?

queries/kcl/highlights.scm
bertbaron: Ah, ok, I was this in the python highlighter and assumed it would have affect maybe for 'unhighlighted' parts of the interpolated expression, but didn't test. Removed the line.

queries/kcl/injections.scm
bertbaron: The doc-comment is indented, so each line has a whitespace prefix. Indented rst implies a quote I think, at least it is not or almost not highlighted because of the indentation. I thought that #2212 refers to a similar issue with markdown. But anyway, you are probably right that it's better to not inject anything than injecting something else that just partly works.

queries/kcl/highlights.scm
clason: Yes, different highlighters work differently, that was my point about queries being editor specific. 

I should probably tighten up CI to no longer accept `@none` captures.

bertbaron:
I think I resolved all issues now so I resolved the conversations.

queries/kcl/highlights.scm
```diff
@@ -0,0 +1,167 @@
...
+
+(call_expr
+  function: (selector_expr
+    (select_suffix
+      (identifier) @function)))
+
+[
+  (integer)
+  (float)
+] @number
```
clason, line 72:
```suggestion
(integer) @number
(float) @number.float
```

Try to be as specific as possible.

queries/kcl/highlights.scm
```diff
@@ -0,0 +1,167 @@
...
+
+[
+  (true)
+  (false)
```
clason, line 76: These should be `@boolean`, no?

queries/kcl/highlights.scm
```diff
@@ -0,0 +1,167 @@
...
+[
+  "all"
+  "any"
+  "assert"
+  "as"
+  "check"
+  "elif"
+  "else"
+  "filter"
+  "for"
```
clason, line 90: `@keyword.repeat` (and `@keyword.conditional` for others).

Again, use subcaptures wherever possible.

queries/kcl/highlights.scm
```diff
@@ -0,0 +1,167 @@
...
+    (identifier) @variable.parameter))
+
+(lambda_expr
+  (identifier) @variable.parameter)
+
+(selector_expr
+  (select_suffix
+    (identifier) @property))
+
+(comment) @comment
```
clason, line 46:
```suggestion
(comment) @comment @spell
```

queries/kcl/highlights.scm
```diff
@@ -0,0 +1,180 @@
...
+"check" @keyword
+
+[
+  "("
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
+] @punctuation.bracket
```
ribru17, line 119: Also needs `@punctuation.delimiter` for `"->"`, `"."`, `","`, `"?:"`, etc

queries/kcl/highlights.scm
```diff
@@ -0,0 +1,180 @@
...
+  "<"
+  ">"
+  "~"
+  "<="
+  ">="
+  "=="
+  "!="
+  "@"
+  "="
+  ":"
```
ribru17, line 147: This should be changed to `@punctuation.delimiter` imo

bertbaron:
Just added folds queries since this PR is still open. I hope it can be merged soon though.

bertbaron:
Thanks for the feedback. I made some changes and also added some operators that were missing. I'm not sure of some of them though. In particular:

- `->` is used for lambda return type (`my_fn = lambda x:int, y:int -> int { x + y }`)
- `@` is used for decorators (`@deprecated`)
- `?:` is used for optional attributes (`   age?: int`)*
- `?.` is a compound chaining operator `print(person?.name)`*
- `?` is similar for arrays: `print(some.array?[2])`*

*The tree-sitter grammar for these deviates from the [official spec](https://www.kcl-lang.io/docs/reference/lang/spec/kcl-spec) where `?`, `.` and `:` are different tokens and whitespace is allowed in between. Not sure if I should try to fix the grammar for this, but if I do then I will of course also fix the highlighting in another PR.

ribru17:
Thanks, that makes sense. In that case I think `"@"` should be `@attribute`, same with the text `deprecated` from `@deprecated` (if not already). It would be nice if the `?` part was captured separately from e.g. `:`, but not a blocker :+1: `->` can go either way with `@operator` or `@punctuation.delimiter`, we have both scattered a lot in the code base (we should consolidate them and pick one for things like return value type specification)

bertbaron:
Using `@attribute` now for `@`, just like the attribute itself. I will address the grammar changes in a new PR, probably together with indents.

ribru17 (approved):

PR #7659, opened 2025-02-16
feat(typst): add more folds

har7an:
for 'for' blocks and 'if/else' expressions.

Actually I would like the folds to be more fine grained, but the treesitter grammar doesn't give me the information I need for that. I can't seem to make it fold individual if/else branches (see below, it folds the entire expression) nor can I distinguis multiline content blocks like this one:

```
#[
    Inlined content here
]
```

Perhaps I'll invest some time into that another day, but for the moment being able to fold large `for`/`if/else` expressions is a big win in my opinion.

## folds for `if/else`

image

## folds for `for`

image

### DIFF
--- a/queries/typst/folds.scm
+++ b/queries/typst/folds.scm
@@ -4,4 +4,8 @@
   (show)
   (call)
   (section)
+  (for)
+  (branch)
+  (content)
+  (raw_blck)
 ] @fold


har7an:
Actually I just remembered users can configure for themselves how small of a chunk they want folded (`'foldminlines'`), so here we go:

## Folds for content

image

image

But not for single-line content with `vim.opt.foldminlines = 1`:

image


## folds for raw markup

image

clason:
@uben0, @RaafatTurki

clason:
@uben0, @RaafatTurki

har7an:
@clason Cool, thanks.

Out of sheer curiosity: What are the tasks of maintaining a tree-sitter grammar integration here/how does one apply? I'm regularly working with typst at work, so I've sort of got an interest in keeping this maintained.

clason:
Reviewing issues and PRs regarding the queries; monitoring parser development and being proactive about adapting the queries to breaking parser changes.

PR #7662, opened 2025-02-17
fix(install): early return if parser_info does not exist

seblyng:
Every now and then when a parser is renamed/removed or something, it will fail with `trying to index `parser_config` a nil value because we are not just continuing to execute even if it is nil or not

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -91,6 +91,7 @@ local function get_parser_install_info(lang)
 
   if not parser_config then
     log.error('Parser not available for language "' .. lang .. '"')
+    return
   end
 
   return parser_config.install_info


clason:
Thank you, but `main` is WIP and not yet done. (And as you see, this change fails tests.)

seblyng:
I know that it is WIP and therefore wanted to contribute to iron things out before it is made as default. But I assume you don't want contributions to main then? I could look into fixing the tests along with this change

clason:
That is appreciated, but `main` is indeed still in flux (I'm looking to rebase the install on async v2 when that is done, and I don't know how much will have to change for that).

But if you can keep it at a one line fix, I can include it for now.

clason:
To be clear, the tests are correct; this change is actually breaking for the async workflow.

clason:
Those are clearly unrelated and will be cleaned up before this becomes the default branch.

seblyng:
Are you sure the tests failing is related? I just tried to commit a single comment change and the same tests are failing

clason:
The validation test is indeed unrelated; please rebase. (Did I mention that this branch is not ready for PRs yet?)

seblyng:
Done

clason:
Ok, the test failure is a known regression, so this does look good. I assume you tested it locally, also when updating/uninstalling/reinstalling?

seblyng:
Yes I did

seblyng:
Btw while I have you here. Is there any ETA on making `main` the default branch? Just curious😅

PR #7663, opened 2025-02-17
Remove the "class" keyword

yorickpeterse:
This keyword is deprecated as of 0.18.1 (released on February 12, 2025) and is being removed as part of the upcoming 0.19.0 release.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -363,7 +363,7 @@
     "revision": "962568c9efa71d25720ab42c5d36e222626ef3a6"
   },
   "inko": {
-    "revision": "0b08a8f976456a9271f70d4682143328d7224115"
+    "revision": "4d057a5c24d715b211c3c82de2526f041b4af66f"
   },
   "ipkg": {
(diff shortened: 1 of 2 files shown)


clason:
Hmm, are you sure there's no code in the wild legitimately using thhst keyword?

yorickpeterse:
@clason I doubt it: Inko moves pretty fast (though I try to keep syntax changes to a minimal), so most users probably follow `main` already. The stable releases are more meant for package managers and people that just want to quickly try things out.

The current 0.18.1 release already automatically reformats `class` to `type` and all documentation uses the new keyword. I suspect that the number of users that use the tree sitter grammar _and_ are not yet using 0.18.1 is zero :smile:

clason:
It's a breaking change nonetheless, and these should never be made lightly.

Since I assume you've already removed the node in your grammar (hence breaking queries), I have no choice but to merge this.

Nevertheless, it does indicate that the language is not stable yet, which would these days be grounds to reject it from nvim-treesitter.

PR #7669, opened 2025-02-19
feat(promql): do not use regex for string labels

danielleontiev:
In Prometheus, label values are treated as strings when used with the `=` and `!=` operators, and as regular expressions when used with the `=~` and `!~` operators.

Injecting and then highlighting all label values as regex leads to a situation where entirely valid **string** label values containing regex special characters are mistakenly parsed and highlighted as regex. This results in syntax errors, causing labels to be highlighted incorrectly.

For example, in `foo{bar=~"[a-z]{1,3}"}`, `{` and `}` are regex special characters, so regex highlighting is expected. However, in `foo{path="/foo/{id}"}`, `{` and `}` are just part of the string and have no special meaning, so the whole value should be highlighted as a string.

---

Before:

before

After:

after

### DIFF
--- a/scripts/minimal_init.lua
+++ b/scripts/minimal_init.lua
@@ -13,7 +13,6 @@ vim.filetype.add {
     usda = "usd",
     wgsl = "wgsl",
     w = "wing",
-    promql = "promql",
   },
 }
 
--- a/scripts/minimal_init.lua
+++ b/scripts/minimal_init.lua
@@ -13,6 +13,7 @@ vim.filetype.add {
     usda = "usd",
     wgsl = "wgsl",
     w = "wing",
+    promql = "promql",
   },
 }
 
--- a/tests/query/highlights/promql/regex.promql
+++ b/tests/query/highlights/promql/regex.promql
@@ -1,0 +1,8 @@
+foo{path=~"^foo$"}[5m] or
+#          ^ @string.regexp
+foo{path!~"[a-zA-Z0-9]{1,3}"}[5m] or
+#          ^ @string.regexp
+foo{path="/api/users/{userId}"}[5m] or
+#         ^ @string
+foo{path!="/api/users/{userId}"}[5m]
+#          ^ @string
--- a/tests/query/highlights/promql/regex.promql
+++ b/tests/query/highlights/promql/regex.promql
@@ -1,4 +1,3 @@
-# vim: ft=promql
 foo{path=~"^foo$"}[5m] or
 #          ^ @string.regexp
 foo{path!~"[a-zA-Z0-9]{1,3}"}[5m] or
@@ -7,3 +6,5 @@ foo{path="/api/users/{userId}"}[5m] or
 #         ^ @string
 foo{path!="/api/users/{userId}"}[5m]
 #          ^ @string
+
+# vim: ft=promql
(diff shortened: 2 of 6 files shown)


scripts/minimal_init.lua
```diff
@@ -13,6 +13,7 @@ vim.filetype.add {
     usda = "usd",
     wgsl = "wgsl",
     w = "wing",
+    promql = "promql",
```
clason, line 16:
```suggestion
```
Not needed.

danielleontiev: without it, `.promql` files are not recognized as `promql` filetype and newly added unit tests involving `.promql` files fail. On the other hand, unit tests with the `.yaml` files containing `promql` queries finish successfully because `yaml` injections are correctly recognized being `promql`. Is there a way to workaround the issue?

danielleontiev: Fixed using `vim: ft=promql` in test files

clason: Yes, that's cleaner.

clason: Ideally, someone would contribute filetype detection for promql files to Vim. (No need for syntax or other files; just detection is enough.)

tests/query/highlights/promql/regex.promql
```diff
@@ -0,0 +1,9 @@
+# vim: ft=promql
```
clason, line 1: Modelines go at the end of the file (separated by a blank line).

clason:
@MichaHoffmann

scripts/minimal_init.lua
danielleontiev: Interestingly, I never encountered standalone `*.promql` files in the wild, they always live as a part of some yaml configuration or template. There are few examples on GitHub (https://github.com/search?q=path%3A*.promql%20NOT%20is%3Aarchived%20NOT%20is%3Afork&type=code), but probably it's not the best metric because it's not very common to see open-sourced infrastructure code on GitHub anyway

tests/query/highlights/promql/regex.promql
danielleontiev: done

scripts/minimal_init.lua
clason: Makes sense; then the modeline is indeed the best approach.

MichaHoffmann: So this was initially written for Alertmanager config files; i think i might have added the FT just for the tests maybe.

MichaHoffmann (commented):
Beautiful, LGTM. Thank you for the PR; Awesome to see that someone else is using this too!

MichaHoffmann:
@clason I wonder ~ Do you have an Idea how to make the injections work with templated yaml?  I have a usecase where we have templated yaml and i would love to still have the injections for PromQL

clason:
I don't, sorry. Templates don't really work with tree-sitter (they invert the injection logic tree-sitter is built around), and the whole `injection.combined` thing is not really working.

MichaHoffmann:
> I don't, sorry. Templates don't really work with tree-sitter (they invert the injection logic tree-sitter is built around), and the whole `injection.combined` thing is not really working.

Ah thats what i feared, thank you!

PR #7671, opened 2025-02-20
feat(pascal): Add asm language injection for Pascal

Isopod:
I noticed there is an asm grammar now, so we can uncomment this.


Isopod:
I just tested it before creating this PR. Haven't noticed any problems. The asm grammar might not be exactly the same as the inline asm grammar used by Delphi/FPC, but this grammar seems to be very general anway. At least it highlights numbers and comments correctly, so it is definitely an improvement.

PR #7688, opened 2025-02-26
fix(kdl): Highlight query order

paul-hansen:
Fixes the highlighting queries for KDL so everything isn't highlighted as a variable due to the order of the queries previously meaning several queries were ignored.

I also changed the node names to be highlighted as `@tag` since KDL is an xml-like format so @tag seemed more appropriate than `@type` and it differentiates it from the type annotations.

This does NOT update to the latest KDL version 2, that would need grammar changes. It does "mostly" work with
KDL 2 though (notably unquoted strings seem to throw the grammar off.).

## Before

(Everything that's white is `@variable` due to the broad `(identifier) @variable` that was after the queries that were targeting specific identifiers)
Screenshot_20250226_134336

## After
Screenshot_20250226_134612

### DIFF
--- a/queries/kdl/highlights.scm
+++ b/queries/kdl/highlights.scm
@@ -1,17 +1,18 @@
 ; Types
+; Variables
+(identifier) @variable
+
+; Nodes
 (node
+  (identifier) @tag)
+
+; Type annotation
+(type
   (identifier) @type)
-
-(type) @type
-
-(annotation_type) @type.builtin
 
 ; Properties
 (prop
   (identifier) @property)
-
-; Variables
-(identifier) @variable
 
 ; Operators
 [


clason:
@amaanq

paul-hansen:
Fixed formatting.

KDL examples from screenshots were pulled from the KDL V1 spec examples (since the grammars haven't been updated for v2 yet):
https://kdl.dev/spec-v1/

<details>
  <summary>Click to show KDL examples for copy pasting</summary>

```kdl
my-node 1 2 3 "a" "b" "c"
foo 1 key="val" 3 {
    bar
    (role)baz 1 2
}
parent {
    child1
    child2
}
  
</details>

parent { child1; child2; }
node (u8)123
node prop=(regex)".*"
(published)date "1970-01-01"
(contributor)person name="Foo McBar"
```

PR #7693, opened 2025-02-27
feat(blade): add laravel blade grammar

calebdw:
Hello!

This add the [Laravel Blade](https://github.com/EmranMR/tree-sitter-blade) grammar to the list

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,7 @@ We are looking for maintainers to add more parsers and to write query files for 
 - [x] [bibtex](https://github.com/latex-lsp/tree-sitter-bibtex) (maintained by @theHamsta, @clason)
 - [x] [bicep](https://github.com/amaanq/tree-sitter-bicep) (maintained by @amaanq)
 - [x] [bitbake](https://github.com/amaanq/tree-sitter-bitbake) (maintained by @amaanq)
+- [x] [blade](https://github.com/EmranMR/tree-sitter-blade) (maintained by @calebdw)
 - [x] [blueprint](https://gitlab.com/gabmus/tree-sitter-blueprint.git) (experimental, maintained by @gabmus)
 - [x] [bp](https://github.com/ambroisie/tree-sitter-bp) (maintained by @ambroisie)
 - [x] [c](https://github.com/tree-sitter/tree-sitter-c) (maintained by @amaanq)
--- a/queries/blade/highlights.scm
+++ b/queries/blade/highlights.scm
@@ -1,0 +1,15 @@
+([
+  (directive)
+  (directive_start)
+  (directive_end)
+] @tag
+  (#set! priority 101))
+
+([
+  (bracket_start)
+  (bracket_end)
+] @tag.delimiter
+  (#set! priority 101))
+
+((comment) @comment @spell
+  (#set! priority 101))
--- a/queries/blade/injections.scm
+++ b/queries/blade/injections.scm
@@ -1,0 +1,15 @@
+((text) @injection.content
+  (#set! injection.combined)
+  (#set! injection.language php))
+
+((text) @injection.content
+  (#has-ancestor? @injection.content "envoy")
+  (#set! injection.combined)
+  (#set! injection.language bash))
+
+((php_only) @injection.content
+  (#set! injection.combined)
+  (#set! injection.language php_only))
+
+((parameter) @injection.content
+  (#set! injection.language php_only))
--- a/lockfile.json
+++ b/lockfile.json
@@ -44,6 +44,9 @@
   "bitbake": {
     "revision": "a5d04fdb5a69a02b8fa8eb5525a60dfb5309b73b"
   },
+  "blade": {
+    "revision": "bcdc4b01827cac21205f7453e9be02f906943128"
+  },
   "blueprint": {
     "revision": "60ba73739c6083c693d86a1a7cf039c07eb4ed59"
   },
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -191,6 +191,14 @@ list.bitbake = {
   maintainers = { "@amaanq" },
 }
 
+list.blade = {
+  install_info = {
+    url = "https://github.com/EmranMR/tree-sitter-blade",
+    files = { "src/parser.c" },
+  },
+  maintainers = { "@calebdw" },
+}
+
 list.blueprint = {
   install_info = {
     url = "https://gitlab.com/gabmus/tree-sitter-blueprint.git",
--- a/queries/blade/folds.scm
+++ b/queries/blade/folds.scm
@@ -1,0 +1,14 @@
+[
+  (authorization)
+  (conditional)
+  (envoy)
+  (fragment)
+  (livewire)
+  (loop)
+  (once)
+  (php_statement)
+  (section)
+  (stack)
+  (switch)
+  (verbatim)
+] @fold
--- a/queries/blade/indents.scm
+++ b/queries/blade/indents.scm
@@ -1,0 +1,3 @@
+(directive_start) @indent.begin
+
+(directive_end) @indent.end


calebdw:
@clason,

Is there a way to have multiple injection files? I have some injections that would be great to include with this grammar, but they technically need to extend the `html` queries.

clason:
You can `; inherits: ...` injections, but you can only have one file per language. Note that templating languages are a poor fit for treesitter in general (and I have yet to see one where it works even remotely well...)

README.md
```diff
@@ -188,6 +188,7 @@ We are looking for maintainers to add more parsers and to write query files for
 - [x] [bibtex](https://github.com/latex-lsp/tree-sitter-bibtex) (maintained by @theHamsta, @clason)
 - [x] [bicep](https://github.com/amaanq/tree-sitter-bicep) (maintained by @amaanq)
 - [x] [bitbake](https://github.com/amaanq/tree-sitter-bitbake) (maintained by @amaanq)
+- [x] [blade](https://github.com/EmranMR/tree-sitter-blade) (experimental, maintained by @calebdw)
```
clason, line 191: Sorry, but we no longer accept experimental grammars. Either it is ready for production use, or it needs to cook outside of nvim-treesitter. (We have too many languages now that frequent breaking changes are too much to handle for me.)

queries/blade/highlights.scm
clason, line 1: That looks a bit meager given the grammar. Is this work in progress, or how is this supposed to work?

README.md
calebdw: I understand, I've held off on adding this grammar for that reason---but this has been around since 2023 and has been pretty stable so figured it's time to add it.

Wasn't sure what the experimental flag was for so just added it to be safe but I'll remove

queries/blade/injections.scm
```diff
@@ -0,0 +1,16 @@
+((text) @injection.content
+  (#not-has-ancestor? @injection.content "envoy")
```
clason, line 2: Those are relatively expensive. It may be better to default to the more common one and override it in a second stanza.

queries/blade/highlights.scm
calebdw: This grammar primarily serves to inject php (which in turn injects html) into the relevant portions of the template. As such there's very little that the actual grammar needs to highlight---but I did remember that I could highlight the brackets (`{{`, `{!!`) so I added that

queries/blade/injections.scm
calebdw: so you mean like this?

```scheme
((text) @injection.content
  (#set! injection.combined)
  (#set! injection.language php))

((text) @injection.content
  (#has-ancestor? @injection.content "envoy")
  (#set! injection.combined)
  (#set! injection.language bash))
```

queries/blade/highlights.scm
clason: It's a pity you can't have Pull Request Templates on Github; otherwise this would include a request for code samples and screenshots...

queries/blade/injections.scm
clason: Yes.

clason:
You probably want to add folds for the tags?

queries/blade/injections.scm
clason, line 1: One thing to keep in mind is that combined injections are currently broken in Neovim (they "leak out" of the injected nodes if they span multiple lines).

calebdw: Hmm, not sure there's much we can do about that then...

clason: Right, and I'd prefer _not_ providing a parser and queries that are known not to work properly...

But surely you've test-driven these queries yourself for a bit before contributing them here, so you can answer that: Do they work properly (with nested injections) or not?

calebdw: Yes, I've been using these queries for a while and they appear to be injecting correctly:

image

clason: Are they always inline? Or can they span multiple lines?

calebdw: They can span multiple lines, but turning off `@injection.combined` definitely breaks highlighting:

image

clason: Yes, combined is needed. The question is if the highlight is correct _between_ tags with multiline injected tags.

calebdw: > The question is if the highlight is correct between tags with multiline injected tags.

I'm not quite sure what this means, but I believe so. The example below shows two multiline `text` nodes that are separated be the `{{ ...  }}` node and highlighting appears to be working correctly:

image

clason: I'm asking about the other way around: if the  `{{ ... }}` (which are injected here) can span multiple lines, and what happens if they do.

calebdw: Still seems to work:

image

clason: Just to make sure: can there be multiple lines of _content_, and what happens if you sandwich normal code between two such tags?

calebdw: Still not sure what exactly you're getting at, but here I show multiple multiline injections going back and forth and everything looks correct:

image

clason: Then it's probably fine.

clason:
Have you tested `tree-sitter generate` and `tree-sitter build` work with the latest version (0.25.2) without warnings?

calebdw:
So ABI 15 parsers will work if I'm on Nvim nightly?

Basically I need to support v0.25.2 but commit the parser generated with `--abi 14`?

clason:
> So ABI 15 parsers will work if I'm on Nvim nightly?

Yes, and only on nightly. (`:TSInstallFromGrammar` will use the latest ABI your Nvim supports -- that's the point of this feature.)

> Basically, in order for this parser to be supported in nvim-treesitter I need to support v0.25.2 but commit the parser generated with --abi 14?

Yes, the parser should not trigger warnings on v0.25.2. It's OK to know that it doesn't (hence my asking you to test it -- it's another checkmark I would add to a template if we could have one...) but you don't actually have to use it for the committed parser.

calebdw:
> No, that was specifically about parsers generated with ABI 15. I want to make sure that :TSInstallFromGrammar blade works for people who have the latest tree-sitter CLI installed (and use Nvim nightly, which will generate with ABI 15 in this case).

I added support for tree-sitter v0.25 to the grammar and updated the commit hash

clason:
you didn't have to do that; I just wanted to know if there are any warnings (since 0.25 added a bunch of new ones, which can turn to errors quickly...)

calebdw:
> I just wanted to know if there are any warnings (since 0.25 added a bunch of new ones, which can turn to errors quickly...)

The warnings were related to not having a `tree-sitter.json` which I added, so no warnings now :).


> but you can only have one file per language.

On another note, the `tree-sitter.json` files does allow specifying multiple query files (e.g., from javascript):

```json
      "highlights": [
        "queries/highlights.scm",
        "queries/highlights-jsx.scm",
        "queries/highlights-params.scm"
      ],
```

would it not be possible to support multiple files in nvim-treesitter, provided that the filenames match?

clason:
No. We do not trust anybody's queries but our own. (Can't, really.) 

So all those extraneous files are meaningless to us; only parser.c and scanner.c matter.

clason:
If you need to split things up, we have the `; inherits ` mechanism for that (check ecma and friends).

calebdw:
> No. We do not trust anybody's queries but our own. (Can't, really.)

I'm not saying to use other queries, just allowing multiple query files (can even determine which type it is based on the prefix like `highlight-*.scm`).

> If you need to split things up, we have the ; inherits  mechanism for that (check ecma and friends).

This won't work in this use-case. I need to extend the html queries but when I place them into the regular `injections.scm` I get errors that there's no such nodes in the `blade` grammar:

```scheme
;; extends html

; AlpineJS attributes
(attribute
  (attribute_name) @_attr
    (#lua-match? @_attr "^x%-%l")
    (#not-any-of? @_attr "x-teleport" "x-ref" "x-transition")
  (quoted_attribute_value
    (attribute_value) @injection.content)
  (#set! injection.language "javascript"))
(attribute
  (attribute_name) @_attr
    (#lua-match? @_attr "^[:@]%l")
  (quoted_attribute_value
    (attribute_value) @injection.content)
  (#set! injection.language "javascript"))

; Blade escaped JS attributes
; <x-foo ::bar="baz" />
(element
  (_
    (tag_name) @_tag
      (#lua-match? @_tag "^x%-%l")
  (attribute
    (attribute_name) @_attr
      (#lua-match? @_attr "^::%l")
    (quoted_attribute_value
      (attribute_value) @injection.content)
    (#set! injection.language "javascript"))))

; Blade PHP attributes
; <x-foo :bar="$baz" />
(element
  (_
    (tag_name) @_tag
      (#lua-match? @_tag "^x%-%l")
    (attribute
      (attribute_name) @_attr
        (#lua-match? @_attr "^:%l")
      (quoted_attribute_value
        (attribute_value) @injection.content)
      (#set! injection.language "php_only"))))
```

clason:
Yeah, that won't work. But _why_ do you want to split up queries for a single language? I do not see any concrete benefits from that.

calebdw:
> But why do you want to split up queries for a single language? I do not see any concrete benefits from that.

Without a major overhaul of the blade grammar (such as having it extend from the html grammar), there's no way to get PHP and JS injected into these nodes without adding them to the html queries (as far as I am aware). I don't particularly want to add these to the `html` injections (although I suppose that is a possibility) given that these are specialized to the Laravel Blade / AlpineJs ecosystems.

Right now the query lives in my `nvim/after/queries/html`, but obviously it's not great for everyone to have to add these same queries to their own dotfiles.

clason:
Yeah, I don't have a good answer for that (and simply allowing multiple queries per language won't help either). This is just part and parcel of why I keep saying that template languages are poorly served by tree-sitter, and there is nothing we can do about this. This is something that requires fundamental change upstream to support properly.

queries/blade/injections.scm
ribru17, line 1: Should this also have comment injections? Unsure if the higher comment priority would override them

ribru17: ah darn

ribru17 (approved):

PR #7707, opened 2025-03-05
feat: add tera

uncenter:
Adds support for the [Tera](https://keats.github.io/tera/) templating language, using my https://github.com/uncenter/tree-sitter-tera parser/grammar. Please let me know if I have made any mistakes in adding this, I was struggling a bit due to the lack of clear documentation on this process.

### DIFF
--- a/queries/tera/highlights.scm
+++ b/queries/tera/highlights.scm
@@ -23,6 +23,7 @@
 [
   "."
   ","
+  "::"
 ] @punctuation.delimiter
 
 [
@@ -55,7 +56,6 @@
   "{{"
   "-}}"
   "{{-"
-  "::"
 ] @punctuation.bracket
 
 ; Tags
--- a/queries/tera/highlights.scm
+++ b/queries/tera/highlights.scm
@@ -119,7 +119,7 @@
       name: (identifier) @variable.parameter)))
 
 (call_expression
-  scope: (identifier)? @namespace
+  scope: (identifier)? @module
   name: (identifier) @function.call)
 
 (call_expression
@@ -155,4 +155,4 @@
 ; Namespaces
 ;-----------
 (import_statement
-  scope: (identifier) @namespace)
+  scope: (identifier) @module)
--- a/queries/tera/highlights.scm
+++ b/queries/tera/highlights.scm
@@ -24,6 +24,7 @@
   "."
   ","
   "::"
+  (frontmatter_delimiter)
 ] @punctuation.delimiter
 
 [
--- a/queries/tera/highlights.scm
+++ b/queries/tera/highlights.scm
@@ -60,7 +60,7 @@
 
 ; Tags
 ;-----------
-(comment_tag) @comment
+(comment_tag) @comment @spell
 
 ; Keywords
 ;-----------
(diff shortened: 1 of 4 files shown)


clason:
Please read CONTRIBUTING.md for the list of allowed captures (be as specific as possible!). You should also add `comment` injections (if your language has comments).

uncenter:
> Please read CONTRIBUTING.md for the list of allowed captures (be as specific as possible!).

I've tried my best to follow and use those guidelines. Can you point out specifically where I can improve?

clason:
Already fixed -- CI will yell at you if you don't :)

queries/tera/highlights.scm
```diff
@@ -0,0 +1,158 @@
...
+(comment_tag) @comment
+
+; Keywords
+;-----------
+[
+  "if"
+  "elif"
+  "else"
+  "endif"
+] @keyword.control.conditional
```
clason, line 72: You're using Helix captures -- we don't have `@keyword.control` but capture `@keyword.conditional` and friends directly, see https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#keywords

uncenter:
> Already fixed -- CI will yell at you if you don't :)

Gotcha. I see a few improvements after a second look, will implement shortly.
---

If I have a method call that is for a builtin method, should I use `@function.method` or `@function.builtin`?

clason:
We don't have a builtin version `of @function.method.call`, so it'll have to be that.

clason:
`@function.method.call`, as I wrote ;)

queries/tera/highlights.scm
```diff
@@ -0,0 +1,147 @@
...
+  "}}"
+  "{{"
+  "-}}"
+  "{{-"
+  "::"
+] @punctuation.bracket
+
+; Tags
+;-----------
+(comment_tag) @comment
```
clason, line 63:
```suggestion
(comment_tag) @comment @spell
```

(and elsewhere this makes sense)

uncenter: re: where else it would make sense; reading through https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#non-highlighting-captures, should the basic `string` node be spellchecked? It is the only type of string Tera has.

MuntasirSZN:
@uncenter Add a query to add commentstring. See: https://neovim.io/doc/user/treesitter.html#_treesitter-highlight-commenstring

MuntasirSZN:
@clason Is there anything that needs to be done?

queries/tera/highlights.scm
```diff
@@ -0,0 +1,152 @@
...
+  "]"
+  "{%"
+  "%}"
+  "-%}"
+  "{%-"
+  "}}"
+  "{{"
+  "-}}"
+  "{{-"
+  "::"
```
ribru17, line 58: I think this one should be `@punctuation.delimiter`

uncenter: Oh yep good catch, thank you.

ribru17 (approved):
Maybe just needs `(frontmatter_delimiter) @punctuation.special`, then should be good to go

PR #7712, opened 2025-03-07
fix(python): don't highlight attribute name as builtin

disrupted:
assigned higher highlight priority to `@variable.member` over `@type.builtin` to correctly highlight `type` since it is a regular attribute in this context

### Before
screen 2025-03-07 at 09 28 41

### After
screen 2025-03-07 at 09 24 45

### DIFF
--- a/queries/python/highlights.scm
+++ b/queries/python/highlights.scm
@@ -25,7 +25,8 @@
 
 ((attribute
   attribute: (identifier) @variable.member)
-  (#lua-match? @variable.member "^[%l_].*$"))
+  (#lua-match? @variable.member "^[%l_].*$")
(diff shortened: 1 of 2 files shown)


clason:
Isn't it enough to move the `@variable.member` overrides _after_ the `@type.builtin` pattern(s)?

(Never use priority if order alone will do.)

disrupted:
I did not know that the order matters. I'll change it

eeeXun:
After this update. The member function is highlight as member variable. Is this intended behavior?

Before:
image

After:
image

polirritmico:
> After this update. The member function is highlight as member variable. Is this intended behavior?

I've just reported the same Here #7727

PR #7714, opened 2025-03-08
feat(indent): injected language shiftwidth

ruuzia:
This pull request changes the tab size for lines in injected trees to use `nvim_get_option_value()` to get the default shiftwidth for the language's filetype. 

### Context
Nvim-treesitter's indent module supports injected language trees, where one language is embedded within another language. However, the indent size is currently always calculated using the current buffer's `vim.bo.shiftwidth`. Within an injected tree, the language is different from the language of the outer buffer, so it doesn't make sense to use the buffer's value for 'shiftwidth'. In the future, it may be desirable to add more robust configuration for indent sizes, however I believe this is an improvement for the default indentation behavior.

Test case included.

### DIFF
--- a/lua/nvim-treesitter/indent.lua
+++ b/lua/nvim-treesitter/indent.lua
@@ -189,6 +189,10 @@ function M.get_indent(lnum)
   if root_start ~= 0 then
     -- injected tree
     indent = vim.fn.indent(root:start() + 1)
+    -- With an injected tree, we have a different language than the filetype of the current buffer,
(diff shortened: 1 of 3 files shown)


clason:
Thank you, but the master branch is frozen while we prepare for the Nvim-treesitter 1.0 release.

If you are interested in adding this to the `main` branch, feel free (but be aware of regressions).

llakala:
Would you be open to adding this to the main branch? If not, I may take this PR up.

ruuzia:
Sure thing! Thanks for reaching out, @llakala

PR #7715, opened 2025-03-08
fix(rust): inject html and json in macro invocations

ginnyTheCat:
This properly injects html and json in `json!`, `serde_json::json!` and `html!` etc. macro invocations. Before rust was being injected at the same time which caused the injection to break.

Before:
rust macro invocations with improper injection highlighting

After:
rust macro invocations with working injection highlighting


ginnyTheCat:
That was faster than expected. Thx :heart:

PR #7719, opened 2025-03-09
feat(roc)!: bump parser and queries

clason:
Adapt to https://github.com/faldor20/tree-sitter-roc/pull/30

Also synchronize queries with upstream and add indents


clason:
@nat-418

PR #7731, opened 2025-03-12
feat(tact): update parser and queries

novusnota:
There were a couple of breaking changes in the queries, so here I am

### DIFF
--- a/queries/tact/highlights.scm
+++ b/queries/tact/highlights.scm
@@ -2,6 +2,10 @@
 ; --------
 (identifier) @variable
 
+(destruct_bind
+  name: (identifier) @comment
+  bind: (identifier) @variable)
+
 ; variable.builtin
 ; ----------------
 (self) @variable.builtin
@@ -51,15 +55,23 @@
   "<"
   "<="
   "<<"
+  "<<="
   ">"
   ">="
   ">>"
+  ">>="
   "&"
+  "&="
   "|"
+  "|="
   "^"
+  "^="
   "&&"
+  "&&="
   "||"
+  "||="
   "->"
+  ".."
 ] @operator
 
 ; constructor
@@ -70,6 +82,9 @@
 (initOf
   name: (identifier) @constructor)
 
+(codeOf
+  name: (identifier) @constructor)
+
 ; type
 ; ----
 (type_identifier) @type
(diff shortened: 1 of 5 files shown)


queries/tact/highlights.scm
```diff
@@ -2,6 +2,10 @@
 ; --------
 (identifier) @variable
 
+(destruct_bind
+  name: (identifier) @comment
```
clason, line 6: That seems wrong?

novusnota: No, this was done intentionally to visually shadow the previous name (of the field) and draw attention to the new name (of the variable)

queries/tact/highlights.scm
```diff
@@ -277,6 +298,35 @@
...
+
+(asm_integer) @number
+
+(asm_string) @string
+
+(asm_control_register) @string.special.symbol
+
+(asm_stack_register) @string.special.symbol
+
+(asm_hex_bitstring) @function.macro
```
clason, line 313: I'm also not 100% certain about these ones -- can you show a code example?

novusnota: Sure, one moment :)

queries/tact/highlights.scm
clason: I don't think we should do this. Can you give a code example to show how what this pattern captures? Then I can suggest alternatives.

novusnota: > I don't think we should do this. Can you give a code example to show how what this pattern captures? Then I can suggest alternatives.

```tact
struct Example { field: Int }

fun naming(s: Example) {
    let Example { field: varFromField } = s;
    //                   ------------     ↑
    //                   ↑                |
    //                   |                instance of Example struct, received
    //                   |                as a parameter of the function "naming"
    //                   definition of "varFromField" variable, derived
    //                   from the field "field" in Example Struct
}
```

And it's nice (IMO) to highlight the `field` as comment there to show that only the `varFromField` is accessible, not the `field` itself

queries/tact/highlights.scm
novusnota: From the test suite:

```
// Stack register
s0 PUSH
// <- string.special.symbol
// ^ function.call

// ...

// hex bitstring
x{DEADBEEF_} PUSHSLICE
// <- function.macro
// ^^^^^^^^^ function.macro
//           ^ function.call
```

The former is a `string.special.symbol`, because it does not resolve to a value per se — it is its own value.

The latter is a `function.macro`, because its preprocessed into a specific data type and placed on the stack, for the following `function` PUSHSLICE to consume it

clason:
You need to generate with `--abi 14`, or we can't use the new version.

novusnota:
> You need to generate with `--abi 14`

Done!

ribru17 (approved):

novusnota:
Awesome, thanks! Forgot to say it for the rapid response and mindful suggestions :)

Issue #7735, opened 2025-03-13
tree-sitter-lua-master: No such file or directory


Anderwafe:
### Describe the bug

nvim-treesitter, as i think, creates some directories for parser source files. And if source code is downloaded from git, it creates directory "tree-sitter-...-{main branch name}". In the source code of nvim-treesitter main branch name "master" is expected, but "tree-sitter-lua" contains single branch with name "main". So, nvim-treesitter search for "tree-sitter-lua-master" directory, which has name "tree-sitter-lua-main"

Output of `:TSInstall lua`:
nvim-treesitter[lua]: Failed to execute the following command:                                            
{                                                                                                         
  cmd = "mv",                                                                                             
  opts = {                                                                                                
    args = { "-f", "tree-sitter-lua-tmp/tree-sitter-lua-master", "tree-sitter-lua" },                     
    cwd = "/home/user/.local/share/nvim",                                                            
    stdio = {                                                                                             
      [2] = <userdata 1>,                                                                                 
      [3] = <userdata 2>                                                                                  
    }                                                                                                     
  }                                                                                                       
}                                                                                                         
mv: не удалось выполнить stat для 'tree-sitter-lua-tmp/tree-sitter-lua-master': Нет такого файла или каталога

### To Reproduce

1. Start 'nvim'
2. `:TSInstall lua`

### Expected behavior

Install lua parser

### Output of `:checkhealth nvim-treesitter`

```markdown
==============================================================================
nvim-treesitter: require("nvim-treesitter.health").check()

Installation ~
- OK `tree-sitter` found 0.25.3 (2a835ee029dca1c325e6f1c01dbce40396f6123e) (parser generator, only needed for :TSInstallFromGrammar)
- OK `node` found v23.9.0 (only needed for :TSInstallFromGrammar)
- OK `git` executable found.
- OK `cc` executable found. Selected from { vim.NIL, "cc", "gcc", "clang", "cl", "zig" }
  Version: cc (GCC) 14.2.1 20250207
- OK Neovim was compiled with tree-sitter runtime ABI version 15 (required >=13). Parsers must be compatible with runtime ABI.

OS Info:
{
  machine = "x86_64",
  release = "6.13.6-arch1-1",
  sysname = "Linux",
  version = "#1 SMP PREEMPT_DYNAMIC Fri, 07 Mar 2025 20:19:00 +0000"
} ~

Parser/Features         H L F I J
[...]

PR #7738, opened 2025-03-14
parsers.lua: use `main` branch

mathstuf:
Upstream migrated the default branch.

Fixes: #7735

---
Cc: @Anderwafe

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1338,6 +1338,7 @@ list.lua = {
   install_info = {
     url = "https://github.com/MunifTanjim/tree-sitter-lua",
     files = { "src/parser.c", "src/scanner.c" },
+    branch = "main",
   },
   maintainers = { "@muniftanjim" },
 }


clason:
Sorry, but we track the default branch automatically. So this is unnecessary and harmful if the branch changes again.

mathstuf:
This was needed when I hit the same problem as in #7735…so "tracking the default branch" isn't actually happening in all cases.

clason:
Then you should find out why it doesn't happen for you when it works for everybody else.

mathstuf:
All I see is `revision = revision or repo.branch or "master"` in `nvim-treesitter/shell_command_selectors.lua`. I see nowhere that it does `git ls-remote --symref "https://github.com/MunifTanjim/tree-sitter-lua" HEAD` to get the name of the default branch from the repo itself. At that line, `repo` is the `install_info` object I'm editing here. I see where it does get the second line's refname if `branch` isn't set when writing the *lockfile* in `nvim-treesitter/install.lua`, but this is not involved in the installation process.

clason:
It installs by commit hash (tracked in lockfile.json), which works just fine for nearly everybody. So the actual fix requires finding out what precisely goes wrong in your(?) case.

mathstuf:
Ah, I see now. That file is missed in my setup because it was outside of the directories it knows are nvim contents (though I taught it about `queries`), so it wasn't reachable from the actual location (everything is symlinked back to the "source" of the plugin, but `lockfile.json` was not). Ok. Maybe a warning should be raised if the lockfile cannot be found?

clason:
Sorry, but if you are not using a standard installation, that's on you. We can't guard against every possible shenanigans.

mathstuf:
I understand, but I'm not the only one to have hit this (I didn't file #7735, but found it when I came to search about it here). Perhaps a warning when the lockfile logic ends up not providing a revision is sufficient? That would at least help alert those with non-standard things that something is up.

clason:
No, sorry, that is required for custom parsers to work. Again, if (generic) you don't use a standard installation, you better know what you're doing. You can't expect other people to cater to you.

Issue #1622, opened 2021-07-24
Inclusive and exclusive make-range!


Jacfger:
**Is your feature request related to a problem? Please describe.**
The problem actually comes from matching the functionality of an nvim plugin ```nvim-treesitter-textobjects``` to ```targets.vim```. While defining textobjects using treesitter queries mostly works, there are some limitation around capturing whitespace. One example is in the implementation of ```@parameter.inner``` and ```@parameter.outer```, which is explained below.
I therefore wanted to request a predicates such that it is possible to do exclusive make-range (i.e. make-range but excluding either the start node or end node, or both.)

Now in the implementation of the argument text objects would be like this, (`ia` and `aa` would be equals to `@parameter.inner` and `@parameter.outer`)
```
    x = ( aaaaaa, bbbb  , dddd   )
          └─ia─┘
    x = ( aaaaaa, bbbb  , dddd   )
          └─aa──┘
```
However, in ideal case, I think modifying to this would make more sense.
```
    x = ( aaaaaa, bbbb  , dddd   )
         └─ ia─┘
    x = ( aaaaaa, bbbb  , dddd   )
         └── aa──┘
```
In both case, an exclusive version of the ```make-range!``` predicates is necessary because the parsing tree in common programming language (e.g. c/c++, python) will ignore spaces. Without a ```make-range!``` that excludes the starting node (or ending node), it is not possible to remove the space of the argument adequately. 

**Describe the solution you'd like**
A ```make-range!``` predicate that could exclude either the start node or end node, or both.

**Additional context**
Related issue in textobjects repository: https://github.com/nvim-treesitter/nvim-treesitter-textobjects/issues/69

PR #7742, opened 2025-03-15
feat(query): add exclusive make-range! variants

nathanregner:
Related: #1622

## Motivation

I'd like to write a textobject query to match inside html tags. I'm able to get most of the way there with this query:

```query
(element
  (start_tag)
  .
  (_) @_start
  (_)? @_end
  .
  (end_tag)
  (#make-range! "tag.inner" @_start @_end))
```

But it doesn't match empty tags, as there are no inner nodes:

```html
<div></div>
```
Having a `make-range-exclusive!` directive would make this possible.

```query
(element
  (start_tag) @_start
  (end_tag) @_end
  (#make-range-exclusive! "tag.inner" @_start @_end))
```

## API

I'm not set on the exact naming/API. Alternatives:

- Add optional parameter(s) to the existing `make-range!` directive:
  ```query
  (#make-range! "@capture" @_start @end "[)") ; include start, exclude end
  ```
- Use [mathematical terminology](https://en.wikipedia.org/wiki/Interval_%28mathematics%29#Definitions_and_terminology)
  - make-range-open
  - make-range-left-open
  - make-range-right-open

### DIFF
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -231,6 +231,14 @@ function M.insert_to_path(object, path, value)
   curr_obj[path[#path]] = value
 end
 
+---@type table<string, TSRangeFromNodesOpts>
+local range_directives = {
+  ["make-range!"] = {},
+  ["make-range-exclude-end!"] = { exclude_end = true },
+  ["make-range-exclude-start!"] = { exclude_start = true },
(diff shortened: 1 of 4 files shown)


clason:
Thank you, but this is deprecated and will be removed in 1.0.

nathanregner:
Makes sense! I'll create a PR in the textobjects repo :smiley:

PR #7749, opened 2025-03-19
feat(javadoc): add parser and queries

rmuir:
Previously java was configured to use doxygen parser for documentation comments, but javadocs are not doxygen. Inline tags have a different syntax, block tags are not recognized, and doxygen creates a lot of errors during highlighting.

Add parser for javadoc comments, with queries for highlights and injections.

### DIFF
--- a/queries/javadoc/highlights.scm
+++ b/queries/javadoc/highlights.scm
@@ -1,0 +1,98 @@
+[
+ (tag_name)
+ "include"
+ "exclude"
+] @nospell @keyword
+
+(identifier) @nospell @variable
+
+(fragment) @nospell @variable.member
+
+(parameter
+  name: (identifier) @variable.parameter)
+
+(param_tag
+  parameter_name: (identifier) @variable.parameter)
+
+[
+ (boolean_type)
+ (integral_type)
+ (floating_point_type)
+] @nospell @type.builtin
+
+(module
+  (identifier) @module)
+
+(type
+  (identifier) @type)
+
+(type_parameter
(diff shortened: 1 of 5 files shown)


queries/javadoc/highlights.scm
```diff
@@ -0,0 +1,98 @@
...
+
+(type_parameter
+  (identifier) @type)
+
+(method
+  (identifier) @function)
+
+(((method
+    (identifier) @constructor)
+    (#match? @constructor "^[A-Z]")))
```
clason, line 37: Is this really necessary? We are trying to get away from convention-driven highlighting and prefer to rely on the actual grammar.

Also, `#match?` is never acceptable in Neovim for performance reasons; use `#lua-match?` with Lua patterns instead.

rmuir: no, not necessary. I was attempting to be consistent with parent language highilghts (java).

as far as the `#match?`, I can replace any that are left, after removing "conventions" matches. I only used them because they work with tree-sitter CLI.

clason: Yes, different editors behave differently, depending on the regex support. For Neovim, regex are particularly expensive since they cross the language barrier to Nvim core.

clason:
Please run `scripts/format-queries.lua queries/javadoc`.

clason:
Can you please add (simple) indentation queries? The java indentation tests fail otherwise due to the new injections.

ribru17 (approved):
Nicely done

PR #7750, opened 2025-03-19
fix(python): reorder builtin functions after other function calls

dav-cho:
`@functon.method.call` currently overrides `@functon.builtin`.

Reorder builtin functions after other function calls.


clason:
Please try to use order for this, not explicit priority.

dav-cho:
Thank you 🙂

PR #7768, opened 2025-03-27
fix(compat): nightly deprecation

clason:
Working toward https://github.com/neovim/neovim/pull/33070

@kiyoon may need changes in nvim-treesitter-textobjects as well (on the `main` branch).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,6 +20,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         cc: [gcc]
+        nvim_tag: [stable, nightly]
 
     name: Run tests
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


clason:
Ok, those tests fail on `master`, too, so we're not even exercising this function. That'll make this harder.

kiyoon:
If I understand correctly, it will change the output format of the iter_matches() and thus the nvim-treesitter side of function? I think it will then break many plugins, so is it possible to make it compatible and keep the old behaviour?

clason:
The old behavior was broken, so no. Most plugins should have already adjusted to the correct behavior (which was opt-out in nightly/0.11!); if not, they should.

clason:
At least for the `ts_util` test, the "passed" text _is_ the expected text, _not_ what it claims is "expected". This leads me to suspect that it's actually `plenary.busted` that's... busted on Nvim 0.11.

Another reason to replace it for 1.0 (which I hereby declare as what will get me to trigger the 1.0 migration process).

clason:
Not going to waste time on this. The `master` branch is effectively frozen and will not support Nvim 0.12 (where this deprecation will happen); Nvim 0.11 is the last "supported" version. All efforts should now switch to the `main` branch (starting with fixing the test infrastructure, which has bitrotted badly).

PR #7775, opened 2025-03-29
fix(dhall): highlight several missing builtins

savanto:
The dhall language `highlights.scm` queries are missing several builtins. I used the [language reference](https://docs.dhall-lang.org/references/Built-in-types.html) and the [treesitter implementation](https://github.com/jbellerb/tree-sitter-dhall/blob/master/queries/highlights.scm) as references:

- highlight builtin type [`Bool`](https://docs.dhall-lang.org/references/Built-in-types.html#bool)
- highlight import [`... as Bytes`](https://docs.dhall-lang.org/references/Built-in-types.html#keyword-as-bytes) and the bytes literal (ie. [`0x"deadbeef"`](https://docs.dhall-lang.org/references/Built-in-types.html#keyword-as-bytes))
- highlight builtin functions [`Date/show`](https://docs.dhall-lang.org/references/Built-in-types.html#function-date-show), [`Time/show`](https://docs.dhall-lang.org/references/Built-in-types.html#function-time-show), [`TimeZone/show`](https://docs.dhall-lang.org/references/Built-in-types.html#function-timezone-show)

### DIFF
--- a/queries/dhall/highlights.scm
+++ b/queries/dhall/highlights.scm
@@ -1,4 +1,3 @@
-; Text
 ; Imports
 (missing_import) @keyword.import
 
@@ -12,6 +11,7 @@
 ] @string.special
 
 [
+  (import_as_bytes)
   (import_as_location)
   (import_as_text)
 ] @type
@@ -92,6 +92,7 @@
 
 (builtin
   [
+    "Bool"
     "Natural"
     "Natural/build"
     "Natural/fold"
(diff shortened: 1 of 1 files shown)


clason:
I think the byte literals should be highlighted differently. Could you please check how other queries define this?

clason:
Ah, looks like the parser doesn't expose the `0x`. 

Still, wouldn't `@number` be more appropriate?

PR #7794, opened 2025-04-03
fix(doc): Update tree-sitter link in README

jrwrigh:


clason:
Thank you, but the README has to be (and partly already is) rewritten fully, so this will just introduce conflicts  for very small (though positive) gains.

jrwrigh:
Fair enough. Might as well point this to wherever (or whoever) is rewriting the README? (Assuming that line is in the rewrite of course).

clason:
Nah, the whole documentation needs to be rewritten from the ground up. If if this link somehow manages to survive the process (which I doubt since it's not really relevant), you'd be more than welcome to open a PR after the 1.0 release.

PR #7801, opened 2025-04-05
feat(html): use gsub for mimetype lookup

clason:
Swap out `module` and `importmap` (which are not "mimetypes") in favor of explicit patterns.


clason:
@przepompownia

runtime/queries/html_tags/injections.scm
```diff
@@ -39,9 +39,9 @@
       (attribute_name) @_attr
       (#eq? @_attr "type")
       (quoted_attribute_value
-        (attribute_value) @_type)))
+        (attribute_value) @injection.language)))
   (raw_text) @injection.content
-  (#set-lang-from-mimetype! @_type))
+  (#gsub! @injection.language "(.-)/(.+)" "%2"))
```
przepompownia, line 44: I appreciate the simplicity, but it comes at the cost of lack of `script type="module"` support.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#applying_the_module_to_your_html

Please keep in mind that I didn't do frontend jobs for years (it was an even bigger mess back then), so let me comment the `importmap` case by reference to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#importing_modules_using_import_maps (TLDR: we can probably inject it as `json`).

clason: Yes, as I have written, I consider this an acceptable regression _out-of-the-box_. (Not least because it's not actually a mimetype, so the directive was misused.)

You can still make it work by registering -- or, better yet, adding a separate injection query for it.

clason: Added the explicit patterns for the no longer covered test cases.

przepompownia: It would be no problem for me to add them on the user side, but thanks for adding them here.

PR #7804, opened 2025-04-06
feat(yaml): inject bash for Taskfiles

Anthony-Fiddes:
This change would inject bash for yaml [Taskfiles](https://taskfile.dev/) like is currently done for yaml Gitlab/Github pipeline configs. I found it helpful in my config so I thought it could be nice to share with others (especially since it's a simple change to the current injections).

Before:

Screenshot 2025-04-06 at 7 05 08 PM

After: 

Screenshot 2025-04-06 at 7 03 23 PM


clason:
Thanks!

PR #7807, opened 2025-04-09
docs: prettify

e-kwsm:


clason:
Thank you, but the master branch is frozen, and all this will have to be rewritten from scratch for 1.0.

PR #7808, opened 2025-04-11
Document `textobjects` support

jyn514:
This is really just a quick draft; I figure it's better to have something short and a link to the full reference than nothing at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ To make sure a parser is at the latest compatible version (as specified in `nvim
 
 ## Modules
 
-Each module provides a distinct tree-sitter-based feature such as [highlighting](#highlight), [indentation](#indentation), or [folding](#folding); see [`:h nvim-treesitter-modules`](doc/nvim-treesitter.txt) or ["Available modules"](#available-modules) below for a list of modules and their options.
+Each module provides a distinct tree-sitter-based feature such as [highlighting](#highlight), [indentation](#indentation), [textobjects](#text-objects), or [folding](#folding); see [`:h nvim-treesitter-modules`](doc/nvim-treesitter.txt) or ["Available modules"](#available-modules) below for a list of modules and their options.
 
 Following examples assume that you are configuring neovim with lua. If you are using vimscript, see `:h lua-heredoc`.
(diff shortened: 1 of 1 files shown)


jyn514:
hmmmmmmm
image

clason:
Textobjects are a separate plugin and not maintained here. The "module" integration will be removed, to boot.

jyn514:
@clason hmm, i am confused: textobjects are in the nvim-treesitter organization, right? i don't have opinions on whatever "modules" are, but does it make sense to document how things work today, and then those docs can be updated whenever module integration is removed?

jyn514:
also, the docs [mention "textobjects.scm"](https://github.com/nvim-treesitter/nvim-treesitter/tree/master?tab=readme-ov-file#adding-queries) - is that outdated?

clason:
> textobjects are in the nvim-treesitter organization, right? 

Yes, but it's still a separate plugin. There is no one "nvim-treesitter meta-plugin".

> i don't have opinions on whatever "modules" are, 

Well, that's necessary to understand, though, for this discussion.

> but does it make sense to document how things work today, and then those docs can be updated whenever module integration is removed?

No, because they will be removed soon(tm), so it doesn't make sense to add it now only to introduce merge conflicts in the ongoing removal work.

I'd recomment reading the pinned "nvim-treesitter 1.0" issue.

> also, the docs mention "textobjects.scm" - is that outdated?

Yes. The whole docs need to be (and currently are in the process of being) rewritten from scratch.

jyn514:
> No, because they will be removed soon(tm), so it doesn't make sense to add it now only to introduce merge conflicts in the ongoing removal work.

this worries me :sweat_smile: this is exactly what rustfmt tried to do (at one point there was a planned "rustfmt 2.0"!) and that ended up taking much longer than expected and in the meantime 1.0 got neglected ...

clason:
Well, that's open source for you. You get what you pay for.

The current architecture has accumulated so much technical debt that it's unmaintainable in the long run. So you either get a rebooted 1.0 whenever it's ready, or an archived plugin.

PR #7809, opened 2025-04-13
feat: `@constructor.call`

tomtomjhj:


clason:
Thank you, but it's not clear why this change is needed. I remind you that the list of captures is now documented by Neovim itself, so we can no longer make changes here without considering upstream.

So the question is: Is this really worth it?

tomtomjhj:
I like to highlight function definition more noisily than invocation so that they clearly show the structure of the code, like this:
image.
(I believe this is not completely niche use case as there is another popular-ish colorscheme that does similar thing, though with custom captures: https://github.com/p00f/alabaster.nvim.)

Given that, it is reasonable to apply the same scheme to constructors.

In a related to note, I plan to propose `*.definition` suffix (note that there already is `@type.definition`), so that `@function` (and the like) is not overloaded for definition and invocation, and `@function` is used only for generic fallback when the query is not precise enough (e.g., https://github.com/nvim-treesitter/nvim-treesitter/pull/3959#pullrequestreview-1218830075)

clason:
Again, adding new captures comes with significant friction -- upstream is now the source of truth, _not_ this project! So there needs to be a strong reason for adding it beyond "_I_ prefer it" and "we already have a another similar capture" (that is used much more widely.

tomtomjhj:
Since `*.call` and `*.definition` are suffixes, there is no strict need to apply/handle them. Query and colorscheme authors can use the generic fallback without suffix if they don't care about differentiating calls and invocations.

PR #7810, opened 2025-04-13
fix(installer): remove deprecated `--no-bindings` cli flag

js0ny:
See https://github.com/tree-sitter/tree-sitter/pull/4292

The flag will let the installation of latex parser error.

```plaintext
nvim-treesitter[latex]: Error during "tree-sitter generate"                                                            
The --no-bindings flag is no longer used and will be removed in v0.25.0thread 'main' panicked at /usr/share/cargo/regis
try/tree-sitter-generate-0.24.7/src/render.rs:1703:5:                                                                  
This version of Tree-sitter can only generate parsers with ABI version 13 - 14, not 15                                 
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -360,10 +360,17 @@ local function run_install(cache_folder, install_folder, lang, repo, with_sync, 
   else
     if not M.ts_generate_args then
       local ts_cli_version = utils.ts_cli_version()
-      if ts_cli_version and vim.split(ts_cli_version, " ")[1] > "0.20.2" then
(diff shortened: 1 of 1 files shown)


clason:
This is just a warning and unrelated to the actual error, which comes below. You need a newer version of tree-sitter.

js0ny:
> This is just a warning and unrelated to the actual error, which comes below. You need a newer version of tree-sitter.

I tested with a newer tree-sitter CLI and the problem doesn't occur. Therefore, I'm closing this PR.

PR #7811, opened 2025-04-13
fix(brightscript): Updating parser filetype for brightscript: brs

ajdelcimmuto:
After testing out the recently merged brightscript parser, I noticed I missed the specification of filetype. Filetypes for brightscript should be identified as `brs`. The parser / highlights was not working out of the box for me, and I had to set it explicitly in my lua config like this:
 `vim.treesitter.language.register('brightscript', 'brs')`
 
 I am hoping this change will allow for treesitter parser to work out of the box for `brs` filetype.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -221,6 +221,7 @@ list.brightscript = {
     url = "https://github.com/ajdelcimmuto/tree-sitter-brightscript",
     files = { "src/parser.c" },
   },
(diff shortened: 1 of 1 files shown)


clason:
There's no (upstream) `brs` filetype, though?

clason:
We only support "official" (Vim) filetypes; those set in custom plugins (or config) should be registered there as well.

If/when upstream adds brightscript filetype detection, we can and will add that.

PR #7814, opened 2025-04-16
feat(typespec): update typespec highlights and indents, add folds

happenslol:
The language added values and a few other changes a while ago, and I just now got around to adding that to the parser. This also updates the revision, which I see is normally done by a bot, so tell me if I should handle that differently.

The updated highlights, indents and folds come directly from the `tree-sitter-typespec` repository itself.

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -861,7 +861,7 @@
     "revision": "75b3874edb2dc714fb1fd77a32013d0f8699989f"
   },
   "typespec": {
-    "revision": "42fb163442ef2691b9b720fb4e4e846809415d18"
+    "revision": "0a0575a4754a089efdca4248a7c31bb7cfd77921"
   },
   "typoscript": {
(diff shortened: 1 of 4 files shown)


clason:
We stopped all automatic updates since more and more parsers switched to ABI 15, which is not compatible with supported Neovim versions. 

If this parser is affected, I am afraid we will have to shelve this PR until after we have dropped support for Nvim 0.10 (which won't happen on the master branch).

clason:
Yuo, sorry!

happenslol:
> If this parser is affected, I am afraid we will have to shelve this PR until after we have dropped support for Nvim 0.10 (which won't happen on the master branch).

How can I find out whether this parser is affected? I'm also the maintainer of the parser itself, so if I can open a branch and downgrade the ABI I'd be down to do that if it helps get this merged.

clason:
Well, if you are the maintainer, you should know the ABI of your parser...

It's best to wait until the rewrite (on the main branch) is complete, at which point this PR will be very welcome.

PR #7818, opened 2025-04-17
error reporting: say what query and language it was when failing to load queries

lf-:
This is the bare minimum to not make it exceedingly painful to debug query load failures.

Fixes: https://github.com/neovim/neovim/issues/33362


clason:
Thank you, but this is dead code (removed in the 1.0 rewrite). All query handling is done by Neovim APIs, and that is where fixes need to be made.

PR #7822, opened 2025-04-21
feat(cobweb): add support for cobweb

FabricSoul:


clason:
Thank you, but the master branch is frozen; we currently do not accept new parser contribtions.

Issue #7828, opened 2025-04-27
Mark Java's `instanceof` pattern variables as `@local.definition.var`


Selebrator:
Java's instanceof pattern matching contains a variable declaration. Take for example `o instanceof String s`.
In this example the variable `s` of type `String` is declared. Therefore, `s` should be marked as a variable declaration `@local.definition.var`. It is not.

**Scopes**
Ideally the variable declaration is also assigned the correct scope. But this seems complicated, because the scope depends on control-flow. The scope specification can be found [here](https://docs.oracle.com/javase/specs/jls/se24/html/jls-6.html#jls-6.3.2). These are some examples illustrating the scopes:
```java
if (e instanceof String s1) {
  System.out.println(s1); // valid
} else {
  System.out.println(s1); // error: s1 not in scope
}
```
```java
if (!(e instanceof String s2)) {
  System.out.println(s2); // error: s2 not in scope
} else {
  System.out.println(s2); // valid
}
```
```java
if (!!(e instanceof String s3)) {
  System.out.println(s3); // valid
} else {
  System.out.println(s3); // error: s3 not in scope
}
```
```java
if (!(e instanceof String s4)) {
  System.out.println(s4); // error: s4 not in scope
  return;
}
System.out.println(s4); // valid
```

PR #7829, opened 2025-04-27
feat(java) mark instanceof pattern variables as @local.definition.var

Selebrator:
This addresses part of #7828. Specifically, it marks the variable declaration as such. But I did not manage to assign the correct scope.


clason:
The `master` branch is frozen; any changes need to be made to (and from) the `main` branch.

PR #7835, opened 2025-04-30
ci: renovate

clason:
# What changed

* separate two workflows:
  - **lint**: Lua files (stylua, luals), query files (valid captures, predicates, directives using tsqueryls), docs (`SUPPORTED_LANGUAGES.md`) -- does not need parser installation
  - **tests**: parsers (ABI compatibility), query files (tsqueryls on Linux/macOS; nvim on Windows), highlight and indent tests (separated for better readability) -- needs parser installation (but only once)
* remove "update readme" workflow (replaced by check)
* switch to `highlight-assertions` fork with ABI 15 support
* run lints/tests through Makefile for easy local checks (Linux/macOS only)
* fix broken gleam tests (parser update introduced`ERROR`s)

# What needs to be done (help needed)

- [x] fix missing `locals.lua` annotations
- [x] fix failing ~~YAML and~~ Angular indents tests
- [x] ~only run tests if lint passes?~ (deemed not worth it)
- [ ] remove `ci-install.sh`? not happy about the duplicate Nvim download, but needed on Windows (and makes testing multiple Nvim versions simpler)

Windows tests are quite flaky; possibly the runners can't handle the load when building in parallel. Should we just drop Windows from CI? We don't really provide first-class support since nobody on the team uses Windows...

# Followups

* replace `highlight-assertions` crate with nvim implementation (shouldn't be hard)
* replace `plenary.test_harness` with `nvim-test` (once it no longer relies on busted), improve output
* add basic `locals` test (or remove module)


clason:
@dlvandenberg I suspect that the indentation test failures for angular are actually due to query problems (possibly because we inherit from `html_tags`). Could you maybe take a look?

ribru17:
Wow, lua_ls in the editor gave inconsistent results compared to when it was run as a CI check

PR #7837, opened 2025-05-01
feat(gdscript): rm underscore rule highlight

PrestonKnopp:


clason:
It looks like this is an adaptation to a breaking parser change (which would have been good to mention in the description).

Since we haven't pulled in the new parser on `master` (and never will, since that branch is now frozen), I'll close this PR. (I've updated parser and query manually on the `main` branch, though.)

PrestonKnopp:
Oops, I'll mention breaking parser changes in the description and target `main` for future PR's.

Thank you!

PR #7839, opened 2025-05-03
fix: appease lua_ls locals.lua warnings

ribru17:
Some of these warnings expose truly problematic code, but this commit doesn't attempt to clean up that mess, just the warnings (in as sensible a way as possible)

cc @clason, lmk if I should push this to ci/make

### DIFF
--- a/lua/nvim-treesitter/locals.lua
+++ b/lua/nvim-treesitter/locals.lua
@@ -18,15 +18,7 @@ end
 ---@param node TSNode
 ---@return TSNode result
 local function get_root_for_node(node)
-  local parent = node ---@type TSNode?
-  local result = node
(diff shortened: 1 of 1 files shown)


clason:
yes please, just push to that PR

lua/nvim-treesitter/locals.lua
```diff
@@ -18,15 +18,7 @@ end
...
-  local parent = node ---@type TSNode?
-  local result = node
-
-  while parent ~= nil do
-    result = parent
-    parent = result:parent()
-  end
-
-  return result
+  return node:tree():root()
```
clason, line 21: Maybe just inline that then?

PR #7841, opened 2025-05-03
ci: only run query checks if lints succeed

ribru17:


clason:
That's not quite what I had in mind.

Right now, the jobs are deliberately separated into those that require installed parsers and those that don't.

The idea was to run the test workflow (as is) only if the lint workflow (as is) passes.

ribru17:
Yeah, I didn't know if there was a way to have dependencies across yaml files like that. Maybe there is but I wouldn't know of it, need dun

ribru17:
Up to you on what to do with this. Looks like it correctly halts parser generation and build checks on lint failures, but ofc the separation is not as clean

clason:
Yeah, this is a pure GHA question. I seem to remember him looking into this for Neovim and concluding that it's not worth it.

ribru17:
Not worth it to keep the separation, or not worth it to short circuit other workflows at the cost of less separation?

clason:
Not worth the complexity tying one workflow to another. The separation is still reasonable (and sufficient for fast turnaround, IMO).

PR #7843, opened 2025-05-04
feat(locals)!: remove locals module

clason:
Problem: `locals.lua` was neither used nor tested, and providing it goes against the goal of nvim-treesitter not being a required dependency for other plugins.

Solution: Remove the module and document that the queries are provided as-is for limited backward compatibility only.


ribru17 (approved):
:rocket:

PR #7845, opened 2025-05-04
feat!: update tier 1 parsers to versions

clason:


scripts/update-parsers.lua
```diff
@@ -1,4 +1,12 @@
 #!/usr/bin/env -S nvim -l
+
+local tier = nil ---@as integer
+for i = 1, #_G.arg do
+  if _G.arg[i]:find('^%-%-tier=') then
+    tier = tonumber(_G.arg[i]:match('=(%d+)'))
```
ribru17, line 6: Would you mind putting a comment somewhere over here? This paragraph was confusing to me at first glance

CONTRIBUTING.md
```diff
@@ -36,6 +36,11 @@ zimbu = {
 
 **Note:** The "maintainers" here refers to the person maintaining the **queries** in `nvim-treesitter`, not the parser maintainers (who likely don't use Neovim). The maintainers' duty is to review issues and PRs related to the query and to keep them updated with respect to parser changes.
 
+**Note:** To qualify for Tier 1 ("stable"), a the parser needs to
```
ribru17, line 39:
```suggestion
**Note:** To qualify for Tier 1 ("stable"), a parser needs to:
```

scripts/update-parsers.lua
clason: The goal here is to run updates for stable and unstable parsers in parallel so that one doesn't block another. (But we can't run CI for this until `main` is the default branch.)

CONTRIBUTING.md
clason: This is basically just note taking right now; I need to make a full pass through the documentation next.

scripts/update-parsers.lua
```diff
@@ -1,4 +1,18 @@
 #!/usr/bin/env -S nvim -l
+-- Update parsers to latest version (tier 1, stable) or commit (tier 2, unstable)
+--
+-- Usage:
+-- nvim -l update-parsers.lua            # update all (stable and unstable) parsers
+-- nvim -l update-parsers.lua --tier=1   # update stable parsers to latest version
+-- nvim -l update-parsers.lua --tier=2   # update unstable parsers to latest commit
+
+local tier = nil ---@as integer
```
ribru17, line 9: Is `@as` needed here? `make checklua` ran fine without it for me. If it is needed, maybe `---@type integer?` makes more sense?

ribru17 (approved):

Issue #7842, opened 2025-05-04
Broken highlighting for combined injections


calebdw:
### Describe the bug

Hello!

We've been getting multiple reports that highlighting no longer works as it should (https://github.com/EmranMR/tree-sitter-blade/issues/108, https://github.com/EmranMR/tree-sitter-blade/issues/104).

The `blade` parser injects php into the combined `text` nodes:

```fennel
((text) @injection.content
    (#set! injection.combined)
    (#set! injection.language php))
```

This used to work, and nothing has been updated on the `blade` parser side that would cause this. I suspect that this has something to do with the recent combined injection changes that I remember hearing about (but haven't been following too closely).

If you look at the below image, the html highlighting stops after the `{{ ... }}` php code on line 45:

![Image](https://github.com/user-attachments/assets/824dcec1-969e-4dab-868b-5531ff21de2e)

Thanks!!

CC: @ribru17 

### To Reproduce

see above

### Output of `:checkhealth nvim-treesitter`

```markdown
n/a
```

### Output of `nvim --version`

```text
12.x
```

ribru17:
Weird... would you mind pasting/linking that file so I can test?

ribru17:
Nvm I see the minimal repro from that issue 👍

ribru17:
Ok so this shouldn't have to do with recent combined injection changes since the original post said it was happening since 0.10.4 even, but switching the query to the following seems to fix things for me:

```scm
((text) @injection.content
  (#set! injection.combined)
  (#set! injection.language html))

((text) @injection.content
  (#has-ancestor? @injection.content "envoy")
  (#set! injection.combined)
  (#set! injection.language bash))

((php_only) @injection.content
  (#set! injection.combined)
  (#set! injection.language php_only))

((parameter) @injection.content
  (#set! injection.language php_only))
```

the issue is injecting the combined php which immediately injects combined html. Ideally this would work but this patch seems sufficient. Ideally again we could just extend the php queries here but it looks like this can't be done because `php_only` has some nodes that blade doesn't recognize

calebdw:
> the issue is injecting the combined php which immediately injects combined html

I've used the same queries for well over a year and this has worked fine

ribru17:
[...]

PR #7846, opened 2025-05-06
fix(blade): fix blade highlighting

calebdw:
Injecting php was broken upstream so this is a temporary fix until upstream fixes it.

See https://github.com/nvim-treesitter/nvim-treesitter/issues/7842

CC: @ribru17

### DIFF
--- a/queries/blade/injections.scm
+++ b/queries/blade/injections.scm
@@ -1,6 +1,6 @@
 ((text) @injection.content
   (#set! injection.combined)
-  (#set! injection.language php))
+  (#set! injection.language html))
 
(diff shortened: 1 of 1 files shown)


calebdw:
Thanks!

clason:
FYI: Master is frozen, so this should not have been merged (I am trying to stabilize `main`, which includes stopping to force-push -- and rebasing on master requires that). I'll pull this in this time, but I guess I should lock the branch to prevent this in the future.

ribru17:
Oops, sorry about that

clason:
No worries; no harm done (yet)!

Issue #2397, opened 2022-01-29
Query files (and other Lisp dialects) are missing indentation support


theHamsta:
**Is your feature request related to a problem? Please describe.**
It would be nice to use the new indentation features like `@aligned_indent` etc. to indent query files.

**Describe the solution you'd like**
`queries/query/fold.scm` should be added.

**Describe alternatives you've considered**
Use native lispindent

**Additional context**
N/A

clason:
The best way to move this along is to make a PR.

MovieMaker93:
Is the issue stale? If not, could you provide more context? 
I see in the repo the indent.scm file, should I add something?
I'd like to contribute, but I need some guidance first (I am new to Neovim's contribution).
Thanks

lucario387:
It still is missing, support aren't fully in place yet.

VioletJewel:
I've created an indentation for `fennel` - the friendly lisp - but there are a few problems because of limitations using AST.

The issues are as follows (see below for examples):

1. it is impossible (afaik) to get either indent in a `list` *after* the first node if another node follows in the same line or default to an indentation of `2`.
2. there are times when you'd want indentation while typing but if there are no children of the current node, then there is no way to properly indent

```fennel
;;; problems
(foo
  bar qux) ;; 1. this works with my indentation!

(foo bar
     qux) ;; 1. this is impossible to coordinate while maintaining above!

(let [
] nil) ;; 2. this should be indented 6, but there are no children, so it resorts to previous indent!
```

Here is the indentation:

<details>
<summary><b><i>queries/fennel/indents.scm</i></b></summary>

```scheme
;;; lists

((list)
 @indent.align
 (#set! indent.open_delimiter "(")
 (#set! indent.close_delimiter ")")
 (#set! indent.increment 2))

([(table) (table_binding) (table_metadata)]
 @indent.align
 (#set! indent.open_delimiter "{")
 (#set! indent.close_delimiter "}"))

([(sequence) (sequence_arguments) (let_vars)]
 @indent.align
 (#set! indent.open_delimiter "[")
 (#set! indent.close_delimiter "]"))

[")" "}" "]"] @indent.branch
[")" "}" "]"] @indent.end

;;; lists special cases

((case_pair
   lhs: (list_binding)
   @indent.align)
 (#set! indent.open_delimiter "(")
 (#set! indent.close_delimiter ")"))

((list
   call: (symbol) @_call)
 @indent.align
[...]

PR #7847, opened 2025-05-08
feat(Clojure) Add indentation queries for Clojure

garrett-hopper:
This adds indentation support for Clojure based on Tonsky's simplified indentation rules: [Better Clojure formatting](https://tonsky.me/blog/clojurefmt).

There were several `indent.lua` changes necessary to make this work.

# Search parents for `@indent.end`

Previously only the very last leaf node of the previous line was checked for `@indent.end`. Due to many Clojure forms being both `@indent.start` and `@indent.end` (to handle optional newlines between the symbol and the form it applies to), it was necessary to make the code search for `@indent.end` in ancestors of the final node on the previous line.

# Merge metadata from multiple capture

Some of the queries were easier to implement (and the conditional `(list_lit ((meta_lit)* . (sym_lit)) (#set! indent.increment 2))` required) using multiple of the same captures which included different `#set!` metadata directives. `vim.tbl_extend` was used to merge the metadata rather than override when multiple of the same capture name match the same node.

# `indent.propagate`

A new `indent.propagate` metadata flag was added with support for both `@indent.begin` as well as `@indent.align`. It signifies that the indentation of a given node should not prevent parent nodes from being processed.

```clojure
(defn _ [a
         b]<CR>
  true)
```

Entering a new line after `b]` would normally cause the `vec_lit` to be seen and processed for indentation. Without `indent.propagate`, the cursor would end up at column 0. With it, the `list_lit` of the `defn` is processed and the cursor is placed on column 2 instead.

# `indent.align_hanging`

Without this, entering a newline after a `(` (or other collection open delimiter) would cause the cursor to be indented by `indent_size` by the `o_is_last_in_line` logic rather than being aligned with the opening delimiter. `indent.align_hanging` metadata flag on `@indent.align` captures causes that logic to be skipped. It's rare to find committed Clojure code that has a hanging open delimiter at the end of a line, but it does occasionally happen while developing. (And the alternative feels very wrong if you do happen to do it)

# `@indent.open_delimiter`/`@indent.close_delimiter`

In addition to the `indent.open_delimiter` and `indent.close_delimiter` metadata values attached to the `@indent.align` capture, I've added support for having separate `@indent.open_delimiter` and `@indent.close_delimiter` to directly capture the anonymous node of the opening and closing delimiters. While the metadata versions with explicit strings works in most cases, it does not allow for any preview `meta: (meta_lit)` nodes that the Clojure grammar includes within a node.

For example,

```clojure
^:example
{:a 1
 :b 2}
```

is parsed as

```scheme
(source
  (map_lit
    meta: (meta_lit
      marker: "^"
      value: (kwd_lit
        marker: ":"
        name: (kwd_name)))
    open: "{"
    value: (kwd_lit
      marker: ":"
      name: (kwd_name))
    value: (num_lit)
    value: (kwd_lit
      marker: ":"
      name: (kwd_name))
    value: (num_lit)
    close: "}"))
```

because the `^:example` metadata is a child of the `list_lit` node, indentation run on the second `{:a 1` line sees it as the previous line and thinks it needs to indent the line. Having `{` and `}` set as the delimiters may be made to work in this case, however Clojure allows any arbitrary form to be used as metadata.

```clojure
^{:a 1
  :b 2}
{:c 3
 :d 4}
```

I was unable to find any good alternatives which would make it clear that indentation should not begin for the second `map_lit` until after the 3rd line. (The first `map_lit` that's nested within the `meta_lit` has its own align indentation too of course.)

The `lnum - 1 > o_srow` check is to explicitly not do any `@indent.align` indentation until the line following the open delimiter. (To avoid indenting any preceding metadata that happens to be part of the node)

`find_delimiter` was rewritten to accept the entire query result so that it could reference it while it iterates through the node's children searching for a matching delimiter.


garrett-hopper:
Relates to #2397

garrett-hopper:
Looks like the test that is failing is [tests/indent/capnp/test.capnp#L133-L133](https://github.com/nvim-treesitter/nvim-treesitter/blob/12453e0e3d2e82a4ae813f1540315c7cf2df2363/tests/indent/capnp/test.capnp#L133-L133
) (and [test.capnp#L719-L719](https://github.com/nvim-treesitter/nvim-treesitter/blob/12453e0e3d2e82a4ae813f1540315c7cf2df2363/tests/indent/capnp/test.capnp#L719-L719
)) which is due to [queries/capnp/indents.scm#L30-L30](https://github.com/nvim-treesitter/nvim-treesitter/blob/57a8acf0c4ed5e7f6dda83c3f9b073f8a99a70f9/queries/capnp/indents.scm#L30-L30
)

```scheme
(concatenated_string) @indent.align
```

which doesn't have any `indent.open_delimiter` or `indent.close_delimiter` defined.

clason:
Thank you, but the master branch is frozen and we are already discussing a complete rewrite of the indentation module for the `main` branch.

garrett-hopper:
Doh 😮
Where can I follow progress on that?
Consider this a few feature requests to help support Lisps in the new version I guess. 🙂

garrett-hopper:
@clason, any value in rebasing this to merge into `main`? Seems like it'd still apply as that stands now, although the `indents.scm` does depend on the extra capture/metadata I added. 🤔 I saw Helix/Zen compatibility mentioned in the readme regarding the rewrite, so having different features may be undesirable. (Is it expected that all indent queries will need to be rewritten?)

PR #7854, opened 2025-05-13
fix(queries): remove unnecessary escapes

clason:
Some of these may actually need _additional_ backslashes to work as intended.

### DIFF
--- a/runtime/queries/powershell/highlights.scm
+++ b/runtime/queries/powershell/highlights.scm
@@ -142,7 +142,7 @@
     "$true"))
 
 ((variable) @variable.builtin
-  (#lua-match? @variable.builtin "^\$env:"))
+  (#lua-match? @variable.builtin "^$env:"))
 
 (data_name
   (simple_name) @constant)
--- a/runtime/queries/sql/highlights.scm
+++ b/runtime/queries/sql/highlights.scm
@@ -42,7 +42,7 @@
   (#lua-match? @number "^%d+$"))
 
 ((literal) @number.float
-  (#lua-match? @number.float "^[-]?%d*\.%d*$"))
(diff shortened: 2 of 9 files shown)


runtime/queries/sql/highlights.scm
```diff
@@ -42,7 +42,7 @@
   (#lua-match? @number "^%d+$"))
 
 ((literal) @number.float
-  (#lua-match? @number.float "^[-]?%d*\.%d*$"))
```
ribru17, line 45: In this case, the backslash should be changed to a `%`. Looks like whoever put this here meant to escape the `.` (as in a floating point number) but forgot lua patterns use `%` instead of `\` (bug, squashed!)

runtime/queries/powershell/highlights.scm
```diff
@@ -142,7 +142,7 @@
     "$true"))
 
 ((variable) @variable.builtin
-  (#lua-match? @variable.builtin "^\$env:"))
+  (#lua-match? @variable.builtin "^$env:"))
```
clason, line 145: Probably here, too, right? (`%$`)

runtime/queries/systemtap/highlights.scm
```diff
@@ -43,7 +43,7 @@
     "sprintln"))
 
 ((identifier) @variable.builtin
-  (#lua-match? @variable.builtin "^\$+[0-9A-Z_a-z]+\$*$"))
+  (#lua-match? @variable.builtin "^$+[0-9A-Z_a-z]+$*$"))
```
clason, line 46: and here

ribru17: Nah, those are fine, I tested and `$` is only recognized as a special char if it is the last one in the pattern

ribru17 (approved):

PR #7857, opened 2025-05-15
Chore: Consistency tweaks in parsers.lua

david:
First of all, please forgive me for the crap PR. I really appreciate your work and hope this doesn't take too much of your time.

Believe it or not, I'm writing a small tool that reads this file and transforms it into a JSON structure, but the inconsistencies this PR "fixes" would require me to write more code to handle them. Since they do seem to be actual inconsistencies with the structure/formatting of the rest of the file, would you consider merging? 

Much appreciation in advance.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2256,10 +2256,7 @@ list.supercollider = {
 list.superhtml = {
   install_info = {
     url = "https://github.com/kristoff-it/superhtml",
-    files = {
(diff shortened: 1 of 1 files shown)


clason:
But why do you need to write a tool for that? Isn't that just `vim.json.encode(require'nvim-treesitter.parsers')`?

david:
Oh, dear. I saw that lua didn't come with a JSON library, but didn't think to see if nvim had an api for it. This is much better, thanks!

PR #7861, opened 2025-05-17
feat(bash): highlight variable argument for `printf -v`

ValdezFOmar:
Example:

```bash
printf -v var 'hello\nworld'
        # ^^^ this gets highlighted as @variable
```


clason:
Thank you, but the `master` branch is effectively frozen as part of the transition to `main`. I'm planning to make that switch official today, after which you can reopen a PR against `main`.

PR #7865, opened 2025-05-18
docs(readme)!: announce archiving of master branch

clason:


adelarsq:
@clason On the main branch it appears that 'nvim-treesitter.configs' module was removed. Which will be the setup for that rightnow?

clason:
Please read the README; it describes everything that is officially supported. (`main` is _not_ (intended to be) a drop-in replacement for `master`; if you are not willing to adapt your config, stick to `master`.)

adelarsq:
@clason Got it. My config is ok, but it will be necessary some changes on some third party plugins that are using the old way. Thanks for the clarification.

clason:
Third-party plugins should not be referencing `nvim-treesitter` at all; if they are, they are not compatible with the `main` branch and either need to be rewritten, or users need to stay on `master` (which is not going anywhere, for that exact reason).

PR #7869, opened 2025-05-20
Update to Gleam latest version

lpil:
The version previously locked to didn't support all Gleam syntax of Gleam v1.11.0.

I'm unfamiliar with this repo, so apologies if this isn't the correct process for updating this.


lpil:
Ah I see! Thank you. I was confused by `master` being specified as the default branch on github.

clason:
Yes, that is part of the migration process -- after a grace period, I will switch the default branch to `main`, but doing this now would be massively breaking.

PR #7887, opened 2025-05-26
feat(parsers): Add qf(quickfix)

OXY2DEV:
# Name of language

[qf (vim quickfix list)](https://github.com/neovim/neovim/blob/a6e2c2234741d25fbff7fc1b6ecca1a2a86bf01c/runtime/syntax/qf.vim)

>[!NOTE]
> I added an extra node to the grammar as normally you can't specify the language for the item text. So, the format looks like this now.
> ```txt
> <filename>|<row> col <col>|>!<language>!< <text>
> ```
> This is completely optional though!

<details>
<summary>Representative code sample</summary>

```
filename.text | 10 col 5 | Some messages.
lua_file.lua | 12-5 col 2-3 | >!lua!< local some_var = 10;
main.c | 5 col 5-6 warning | some_func(;
diagnostic.lua | 10 | diagnostic.text = "hi"
piped_\|filename.lua | 1-5 col 2-7 | local some_var = 10;
```

</details>

## Parser repo

https://github.com/OXY2DEV/tree-sitter-qf

<details>
<summary>Parsed tree for code sample</summary>

```
(quickfix_list [0, 0] - [4, 0]
  (quickfix_item [0, 0] - [0, 41]
    (filename [0, 0] - [0, 13])
    (range [0, 16] - [0, 24]
      (row [0, 16] - [0, 18]
        (value [0, 16] - [0, 18]))
      (col [0, 23] - [0, 24]
        (value [0, 23] - [0, 24])))
    (code_block [0, 26] - [0, 41]
      (content [0, 26] - [0, 41])))
  (quickfix_item [1, 0] - [1, 58]
    (filename [1, 0] - [1, 12])
    (range [1, 15] - [1, 27]
      (row [1, 15] - [1, 19]
        (from [1, 15] - [1, 17])
        (to [1, 18] - [1, 19]))
      (col [1, 24] - [1, 27]
        (from [1, 24] - [1, 25])
        (to [1, 26] - [1, 27])))
    (code_block [1, 29] - [1, 58]
      (language_delimiter [1, 29] - [1, 37])
      (content [1, 37] - [1, 58])))
  (quickfix_item [2, 0] - [2, 40]
    (filename [2, 0] - [2, 6])
    (range [2, 9] - [2, 26]
      (row [2, 9] - [2, 10]
        (value [2, 9] - [2, 10]))
      (col [2, 15] - [2, 18]
        (from [2, 15] - [2, 16])
        (to [2, 17] - [2, 18]))
      (item_type [2, 19] - [2, 26]))
    (code_block [2, 28] - [2, 40]
      (content [2, 28] - [2, 40])))
  (quickfix_item [3, 0] - [3, 44]
    (filename [3, 0] - [3, 14])
    (range [3, 17] - [3, 19]
      (row [3, 17] - [3, 19]
        (value [3, 17] - [3, 19])))
    (code_block [3, 21] - [3, 44]
      (content [3, 21] - [3, 44]))))
```

</details>

## Queries

Source of queries: "written from scratch"

<details>
<summary>Screenshots of code sample</summary>

Screenshot 2025-05-29 at 8 20 50 AM

</details>

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1765,7 +1765,7 @@ return {
   },
   qf = {
     install_info = {
-      revision = 'v1.2.0',
+      revision = '6a9d1deb934749720aa3bc525f1506f7d2ae4d0a',
       url = 'https://github.com/OXY2DEV/tree-sitter-qf',
     },
     maintainers = { '@OXY2DEV' },


runtime/queries/qf/highlights.scm
```diff
@@ -0,0 +1,26 @@
+"|" @punctuation.delimiter
+
+" col " @keyword
```
clason, line 3: Why the whitespace? That is usually not highlighted.

clason:
Yeah, I have a sneaking suspicion that the `errorformat` is just too flexible to make a general quickfix parser feasible... (Cries in vimdoc.)

OXY2DEV:
> Also just a heads up, the parser behaves weirdly with the following text:

Can you elaborate? Does it not look something like this?

Screenshot_2025-05-29-00-10-47-263_com termux-edit

>[!NOTE]
> Tree-sitter won't highlight correctly if the contents of the quickfix list changes(e.g. you do `setqflist(a)`, `:copen`, `setqflist(b)`, `:copen`). So, that may be what's happening.

OXY2DEV:
> Maybe we can omit that injection here and leave it for the plugin to ; extend with?

So, do I add it to the parser's README that it exists? Or just remove the node entirely?

OXY2DEV:
> While you're at it, maybe the - in row/column number ranges can be parsed as a separate token for more granular highlights? 

Should be easy to implement.

clason:
> > Maybe we can omit that injection here and leave it for the plugin to ; extend with?
> 
> 
> 
> So, do I add it to the parser's README that it exists? Or just remove the node entirely?

That is up to you, depending on your plans for this parser (whether there's a "better-qf.nvim" that goes with it). It's not a blocker, but until this is officially supported by Neovim, I think it's better to omit it here. (We are pretty strict about aligning with Neovim core these days.)

Adding a snippet to the parser README sounds like the best option for now. (Can always revisit later!)

lua/nvim-treesitter/parsers.lua
```diff
@@ -1763,6 +1763,15 @@ return {
     maintainers = { '@stsewd', '@theHamsta' },
     tier = 2,
   },
+  qf = {
+    install_info = {
+      revision = 'v1.2.0',
+      url = 'https://github.com/OXY2DEV/tree-sitter-qf',
+    },
+    maintainers = { '@OXY2DEV' },
+    tier = 1,
```
clason, line 1772: I think it's better to keep this at tier 2 for now, if the parser isn't stable yet.

lua/nvim-treesitter/parsers.lua
```diff
@@ -1763,6 +1763,15 @@ return {
     maintainers = { '@stsewd', '@theHamsta' },
     tier = 2,
   },
+  qf = {
+    install_info = {
+      revision = 'v1.2.0',
```
clason, line 1768: should be latest commit on the main branch then

(the point of tier 2 is to make it easier to iterate so you don't have to make a release for every change)

OXY2DEV:
> There's also the issue with this breaking the `gO` toc windows which needs to be handled somehow.

Can't `ftplugin/qf.vim` be turned into `ftplugin/qf.lua`(or just add it as a separate file if possible) to handle that?

I have been using [this](https://github.com/OXY2DEV/nvim/blob/main/after/ftplugin/qf.lua#L27-L61) to hide the columns in them.

If that PR gets merged into Noevim we could just use an autocmd somewhere.

Or is there some other issues that I am not getting?

PR #7895, opened 2025-05-27
fix(haskell): properly set inline-python injection language

mrcjkb:
The existing query only detects the quasiquote, but doesn't actually set the language to python.


clason:
Thanks, but the master branch is locked; all future changes need to be made to `main`.

mrcjkb:
Yep, I'm aware of the 1.0 roadmap :)

Issue #7872, opened 2025-05-21
On `main`, when installing certain parsers a message is always logged even when there's no work to be done


yorickpeterse:
### Describe the bug

Using the `main` branch, when calling `install()` with multiple parsers, the log message `Installed A/B languages` is always logged, even if the parsers are installed. Curiously, this only seems to happen when certain parsers are used and only when a certain number is used, though I haven't quite figured out the specifics.

### To Reproduce

Create `test.lua` as follows:

```lua
vim.cmd.packadd("nvim-treesitter")

local ts = require("nvim-treesitter")

ts.setup()
ts.install({
	"bash",
	"c",
	"css",
	"fish",
	"gitcommit",
	"go",
	"html",
	"inko",
	"javascript",
	"json",
	"llvm",
	"lua",
	"markdown",
	"markdown_inline",
	"python",
	"query",
	"ruby",
	"rust",
	"toml",
	"typst",
	"vimdoc",
	"yaml",
})
```

Make sure that nvim-treesitter is installed. Next, start Neovim as follows:

```
nvim -u /tmp/test.lua
```

If you then close and run open Neovim again, the log message `[nvim-treesitter]: Installed 26/26 languages` is written in spite of all parsers already being installed.

### Expected behavior

No log message is written when there's nothing to do.

### Output of `:checkhealth nvim-treesitter`

```markdown
==============================================================================
nvim-treesitter:                     require("nvim-treesitter.health").check()

Requirements ~
- ✅ OK Neovim was compiled with tree-sitter runtime ABI version 15 (required >=13).
- ❌ ERROR tree-sitter CLI v0.25.4 is required
- ✅ OK node 23.11.1 (/usr/sbin/node)
- ✅ OK tar 1.35.0 (/usr/sbin/tar)
- ✅ OK curl 8.13.0 (/usr/sbin/curl)
  curl 8.13.0 (x86_64-pc-linux-gnu) libcurl/8.13.0 OpenSSL/3.5.0 zlib/1.3.1 brotli/1.1.0 zstd/1.5.7 libidn2/2.3.7 libpsl/0.21.5 libssh2/1.11.1 nghttp2/1.65.0 nghttp3/1.9.0
  Release-Date: 2025-04-02
  Protocols: dict file ftp ftps gopher gophers http https imap imaps ipfs ipns mqtt pop3 pop3s rtsp scp sftp smb smbs smtp smtps telnet tftp ws wss
[...]

PR #7907, opened 2025-05-28
feat(install): Add a `quite` option to the `install()` method

oriori1703:
This flag skips the "[nvim-treesitter]: Installed X/X languages" message.

fixes https://github.com/nvim-treesitter/nvim-treesitter/issues/7872


clason:
Sorry, that is an intrusive papering over. Thank you, but no thank you.

PR #7910, opened 2025-05-29
feat(install): support custom queries

clason:
This allows specifying a `queries` dir in a (custom) parser's `install_info` field. When installing the parser, this directory is
* linked to the `install_dir` if `repo.path` is set (local checkout/project);
* copied to the `install_dir` otherwise.


clason:
async `rmpath` seems to be breaking Windows 😞

clason:
@simonvic feel free to test this out (both with local and remote custom parsers)!

PR #7914, opened 2025-05-30
feat(haskell) injections for inline-c-cpp quasiquotes

aavogt:
image
the pattern for antiquotes ( `$solid:solid` `$pnt:p`) helps but for some reason antiquotes in the the first statement after `double {` are not highlighted in the same way and it seems like there are other errors though I don't know c++.


clason:
thank you, but the `master` branch is frozen. And your pattern seems to be taken from other queries -- be advised that you can't simply copy them but must make sure they work in the target editor. (For example, Neovim does not support `injection.pattern`.)

PR #7915, opened 2025-05-31
fix(git_config): match full-lowercase Git config variables

ccjmne:
Git config's sections, variable names, and (often) subsections are case-insensitive.

Source: https://github.com/git/git/blob/master/Documentation/config.adoc (`git help config`)

---

Instead of doing a full case-insensitive matching in the (few) places where we'd put the non-normalised variable names, I went for `#any-of?` because:
1. What are the chances anybody would use any casing other than `lowerCamelCase` or `lowercase`?
2. Nobody wants to read `(#lua-match? @keyword.import "^[iI][nN][cC][lL][uU][dD][eE][Ii][fF]$"))`
3. `#any-of` is generally cheaper (read: faster) to run

Oh, and it's a rule in the `CONTRIBUTING` guildelines!
I also ran `luacheck` and `stylua`.

### DIFF
--- a/runtime/queries/git_config/highlights.scm
+++ b/runtime/queries/git_config/highlights.scm
@@ -7,7 +7,7 @@
 ((section_header
   (section_name) @keyword.import
   (subsection_name))
-  (#eq? @keyword.import "includeIf"))
+  (#any-of? @keyword.import "includeIf" "includeif"))
 
 (variable
   (name) @property)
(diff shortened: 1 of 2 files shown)


ccjmne:
Sorry for the few force-pushes.
I originally didn't pay attention to the Conventional Commit usage, nor to the fact I was targeting the wrong branch (`master`), from the wrong branch (`master`).

PR #7938, opened 2025-06-06
feat(install): allow for absolute query paths

dsully:


clason:
Sorry, but why?

dsully:
> Sorry, but why?

I have an odd use case in that queries in https://github.com/apple/pkl-neovim than in https://github.com/apple/tree-sitter-pkl, so I need to do this:

```lua
---@type ParserInfo
parsers.pkl = {
    install_info = {
        url = "https://github.com/apple/tree-sitter-pkl",
        queries = vim.fs.joinpath(root, "pkl-neovim", "queries", "pkl"),
        used_by = { "pcf" },
    } --[[@as InstallInfo]],
    filetype = "pkl",
    maintainers = {},
    tier = 3,
}
```

clason:
Sorry, that's not something I want to support. If the queries live outside the grammar repo, just `vim.opt.rtp:prepend()` the directory.

dsully:
Thanks @clason - should that be added to the README?

clason:
It's standard Neovim functionality; I don't see the necessity of duplicating `:h vim.treesitter`.

PR #7940, opened 2025-06-08
fix(query): parse before access trees

wodesuck:
When using [nvim-treesitter-textobjects-lsp-interop](https://github.com/nvim-treesitter/nvim-treesitter-textobjects?tab=readme-ov-file#textobjects-lsp-interop) to peek definition code, if the definition locate at a unopened file, `get_capture_matches_recursively` will return empty because the file have not been parse, and we can't get the range of the definition code. Add the missing `parse:parse()` to fix it.

### DIFF
--- a/lua/nvim-treesitter/query.lua
+++ b/lua/nvim-treesitter/query.lua
@@ -437,6 +437,7 @@ function M.get_capture_matches_recursively(bufnr, capture_or_fn, query_type)
   local matches = {}
 
   if parser then
+    parser:parse()
     parser:for_each_tree(function(tree, lang_tree)
       local lang = lang_tree:lang()
       local capture, type_ = type_fn(lang, tree, lang_tree)


clason:
Thank you, but the `master` branch is frozen. Plugins should not depend on nvim-treesitter and use the Neovim core API directly.

wodesuck:
So I should modify in `main` branch or `nvim-treesitter-textobjects`?

clason:
Neither. (`nvim-treesitter-textobjects` already has a compatible `main` branch that does not rely on nvim-treesitter).

wodesuck:
It seems that `main` branch of `nvim-treesitter-textobjects` doesn't have `lsp_interop`, will it be supported in the future?

clason:
Nope, not planned.

wodesuck:
So sad, I love that feature 😢

clason:
Extract it into a stand-alone plugin then.

PR #7946, opened 2025-06-12
fix: use `@tag` for css universal selector

uncenter:
Right now, the "*" `universal_selector` isn't actually even highlighted as `character.special`, but as `operator`, because the "*" a few lines below overrides it. I began to fix this by moving the line below the line setting * as operator, but then I thought that `tag` would make more sense as a capture than a special character. `tag` in CSS highlighting is almost the generic selector color and I think it is fitting, other uses of `character.special` (like for the `nesting_selector` feel different.

### DIFF
--- a/queries/css/highlights.scm
+++ b/queries/css/highlights.scm
@@ -27,10 +27,7 @@
   (feature_name)
 ] @property
 
-[
-  (nesting_selector)
-  (universal_selector)
-] @character.special
+(nesting_selector) @character.special
 
 (function_name) @function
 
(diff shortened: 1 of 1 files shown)


clason:
thank you, but the master branch is frozen (and this breaks tests)

uncenter:
Should I just change the target branch to main or are you rejecting this change for other reasons?

clason:
Well, all tests were failing, so that would need to be fixed on `main` as well. Other than that, I defer to the language maintainer whether this change makes sense.

PR #7948, opened 2025-06-13
fix: use `@tag` for css universal selector

uncenter:
https://github.com/nvim-treesitter/nvim-treesitter/pull/7946

### DIFF
--- a/runtime/queries/css/highlights.scm
+++ b/runtime/queries/css/highlights.scm
@@ -27,10 +27,7 @@
   (feature_name)
 ] @property
 
-[
-  (nesting_selector)
-  (universal_selector)
-] @character.special
+(nesting_selector) @character.special
 
 (function_name) @function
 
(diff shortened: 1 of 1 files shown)


clason:
A screenshot before/after would go a long way.

clason:
@TravonteD

clason:
>  tag in CSS highlighting is almost the generic selector color and I think it is fitting

That is a very weak argument (if I understand it correctly). The captures are chosen semantically, not by color (which differ wildly based on the colorscheme). If the universal selector follows the same grammar as the nesting selector, they need to be the same capture. (People can always customize their queries if they feel differently, but _our_ captures prioritize consistency, both within and -- more importantly -- across languages.)

uncenter:
Yeah it's a weak argument for most captures, none of them really feel like a fit to me.

> If the universal selector follows the same grammar as the nesting selector

How do you mean the same grammar as each other? I'm not understanding this.

PR #7954, opened 2025-06-18
feat(vue): add template commentstring metadata

guilhas07:


clason:
Sorry, the `master` branch is frozen. All changes must go to the `main` branch.

clason:
And we do not set options in our queries, period -- these should be local user modifications.

guilhas07:
> Sorry, the master branch is frozen. All changes must go to the main branch.

Sorry didn't notice. 

> And we do not set options in our queries, period -- these should be local user modifications.

I saw other queries setting the comment string (https://github.com/nvim-treesitter/nvim-treesitter/pull/7704)

Given that inside a <script> and <style> tag I can comment a line (I think this is due to @injection.language), I would expect the same to be applied to the vue file when commenting html. 

IMO given that the default vue content is html this should be the default functionality. If there is a more correct approach please lmk.

clason:
Ah, I missed that. If we've already started, then there's no reason not to continue.

PR #7957, opened 2025-06-19
Adds a Tree-sitter parser for Seed7 and removes ada.

Aliqyan-21:
- Parser: https://github.com/aliqyan-21/tree-sitter-seed7
- Features: Supports includes, constants, functions, variables, control structures (if, for, while, repeat, case with when), expressions (identifiers, strings, integers, function calls, concatenation, comparisons, assignments, arithmetic, set literals, ranges, set unions), types (proc, string, integer), and comments.
- Includes highlights.scm for Neovim syntax highlighting.
- Filetype: sd7

### DIFF
--- a/lockfile.json
+++ b/lockfile.json
@@ -722,6 +722,9 @@
   "scss": {
     "revision": "c478c6868648eff49eb04a4df90d703dc45b312a"
   },
+  "seed7": {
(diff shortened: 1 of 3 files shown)


clason:
No thank you.

PR #7959, opened 2025-06-19
Add seed7 parser

Aliqyan-21:
Seed7 parser for syntax highlighting

seed7 is an extensible and general purpose programming language described at the [Seed7 Homepage](http://seed7.sourceforge.net/index.htm).

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,9 @@
   </p>
 </div>
 
+>[!CAUTION]
+> The `master` branch is frozen and provided for backward compatibility only. All future updates happen on the [`main` branch](https://github.com/nvim-treesitter/nvim-treesitter/blob/main/README.md), which will become the default branch in the future.
+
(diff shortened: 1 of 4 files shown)


Aliqyan-21:
I apologize for the mistakes of pushing it to master branch previously this is my first time contributing to nvim-treesitter, just wanted to add support for syntax highlighting for a this incredible language.

clason:
I am saying this with respect, but the PRs are not at the standard we expect at this point. It would be better to develop the language and parser independently until it is stable enough for inclusion here.

clason:
Rule of thumb: If Vim does not recognize the filetype, it's probably too niche for us.

Aliqyan-21:
I understand, seed7 is not that popular and thus vim also does not recognise the file type too, u are right Mr. @clason, we will do as you say, thank you for your time.

Have a great day.

PR #7976, opened 2025-06-29
feat(php): fold match_expression

przepompownia:


clason:
You know the drill ;)

@ribru17 do we want to impose some sort of consistency for folds across languages, i.e., a list of common constructs that should -- or should not -- be folded? Too much folding can be annoying and counterproductive, after all.

przepompownia:
Maybe it's not as useful as for folding complex arrays, but sometimes allows to fold long `match` arm lists (https://github.com/phpactor/phpactor/blob/af9a18ab08202db60a65c8169b0e80ff6409a267/lib/Indexer/Adapter/ReferenceFinder/IndexedNameSearcher.php#L66-L81).

image

I am assuming that if we have folded `switch` arms, `match` could be folded too (or both shouldn't). 

(I use `foldlevel=99` and usually fold with `za`.)

ribru17:
> You know the drill ;)
> 
> @ribru17 do we want to impose some sort of consistency for folds across languages, i.e., a list of common constructs that should -- or should not -- be folded? Too much folding can be annoying and counterproductive, after all.

I would be in favor of this,  sounds like a good thing to document. Off-topic, but eventually it would also be cool to have something like `(#set! foldkind "comments")` similar to LSP, to allow folds to be toggled based on their kind.

clason:
Hmm... I feel that needs to handled upstream first. Vim folds are all-or-nothing; this would require adding a new configuration infrastructure on top (and you know that upstream just _loooves_ doing that).

If we _do_ change this -- after due upstream discussion -- we should just do `@fold.comment` and similar; no point in adding additional metadata on top.

For now, let's start with adding guidance to `CONTRIBUTING.MD`, beginning with the noncontroversial classes.

As I don't use folds (at all), I'll have to let someone else take the lead on this ;)

PR #7977, opened 2025-06-29
feat(D): Add missing string highlights

baruch:
I merely copied the fix from https://github.com/gdamore/tree-sitter-d/issues/54 and applied it here so it will be available upstream.


clason:
The `master` branch is frozen; PRs need to go to the `main` branch.

baruch:
So why is the master branch the default one?

clason:
Please read the pinned issue (or the README).

PR #7979, opened 2025-06-29
docs: specify fold dos and don'ts

ribru17:
Open to changing these rules if someone disagrees with the direction.


CONTRIBUTING.md
```diff
@@ -345,12 +345,27 @@ The valid captures are:
 
 ### Folds
 
-You can define folds for a given language by adding a `folds.scm` query. This is implemented in Neovim. The only valid capture is
+You can define folds for a given language by adding a `folds.scm` query. This is implemented in Neovim. The only valid capture is `@fold`:
 
 ```query
-@fold ; fold this node
+(function_definition) @fold ; fold this node
```
clason, line 351: We still scrape this file for CI on Windows, so I think this needs to stay as-is.

clason:
Maybe we could refer to highlight captures as classes that could contain candidates for folding?

CONTRIBUTING.md
ribru17: yuck that is brittle

clason: Yep, which is why we switched to ts_query_ls (but Windows is brittle, period). 😇

clason: Oh, wait, I removed that part from the scripts completely since we have a separate lint step now. So you're good!

PR #7983, opened 2025-07-02
feat(cpp): add export, import, module highlighting

Soupertonic:
Adds highlighting for the C++20 `export`, `import`, and `module` keywords using the `@keyword.import` group. These keywords are parsed in the Treesitter grammar but were not previously captured for highlighting.

<small>_(Note: This is a resubmission because the original PR #7982 was based on the frozen master branch.)_</small>

### Retrobox
 <table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

### Tokyo Night (Night Variant)
 <table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

### DIFF
--- a/runtime/queries/cpp/highlights.scm
+++ b/runtime/queries/cpp/highlights.scm
@@ -225,6 +225,12 @@
 ] @keyword.type
 
 [
+  "export"
+  "import"
+  "module"
+] @keyword.import
+
+[
   "co_await"
   "co_yield"
   "co_return"


clason:
Sorry, we can't do that, Dave. Some languages -- in this case `arduino` and `cuda` --  are based on cpp and would need to be updated to support the same constructs.

clason:
@ribru17 this is why I keep the nvim query check around ;)

ribru17:
Rats, nice catch.  Thought it was a flake!

(This will be a good reference for me to fix this last inconsistency in the query lsp)

clason:
Yeah, following `; inherits` is crucial for us.

clason:
(Unless we stop doing that completely, which I am more and more in favor of.)

PR #7991, opened 2025-07-07
feat(vim): highlight "trim" in script heredocs

clason:
Allow the optional "trim", as documented in ":help :lua-heredoc".

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2499,7 +2499,7 @@ return {
   },
   vim = {
     install_info = {
-      revision = 'ccc312e878aa84f32d180b8528a3585c7b86a545',
+      revision = 'a93e834bea0975ec0ccb3f6d18540e9bd8170a4d',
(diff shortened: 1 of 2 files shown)


calebdw:
@clason, any idea why this is causing the following failure?

image

I'm on the latest nvim `master` and nvim-treesitter `main`, I've removed this plugin and reinstalled, I've manually executed `TSUpdate` and `TSInstall vim`, but it keeps erroring out

clason:
Please check all your parsers, and make sure the one from nvim-treesitter has priority on rtp. Also try uninstalling and reinstalling the vim parser.

calebdw:
Hmm, it looks like a manual `TSUninstall vim`/`TSInstall vim` did the trick, but it's weird that it wasn't automatically updated...

> make sure the one from nvim-treesitter has priority on rtp

Not sure what you mean here exactly, I only specify the `vim` parser with all my other parsers: https://github.com/calebdw/dotfiles/blob/master/.config/nvim/lua/calebdw/plugins/treesitter.lua

PR #7996, opened 2025-07-09
fix(install): clear status after install

MeanderingProgrammer:
## Details

Currently once the `install_status` is set for a language installing will never proceed, even when `force` is set. For my current setup this prevents updating, some info below in the `Updating` section if it matters.

Regardless of my specific issue I believe the notion of `force` should trigger an install / update when possible. To make this happen I've updated `install_status` so that it is only set for a language while an installation is in progress. After the task is completed `install_status` is cleared for the language. As a consequence of this `install_status` should really only be accessed in the `install_lang` function and can no longer be used to get the result of an installation.

### Updating

What I currently do is always trigger `install` for a list of languages in my config using:

```lua
require('nvim-treesitter').install({ ... })
```

This way if I add any new languages they are automatically installed. As a consequence of this all the languages end up with an `install_status` of `installed`.

After this I'll update my plugins using `lazy.nvim`, this triggers the `:TSUpdate` command, which will not actually update any languages (because of the `install_status`) but output a summary `Installed n/n languages`.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -456,7 +456,6 @@ local install_status = {} ---@type table<string,InstallStatus?>
 ---@return InstallStatus status
 local function install_lang(lang, cache_dir, install_dir, force, generate)
   if not force and vim.list_contains(config.get_installed(), lang) then
-    install_status[lang] = 'installed'
     return 'installed'
   end
 
@@ -475,6 +474,7 @@ local function install_lang(lang, cache_dir, install_dir, force, generate)
 
   local status = install_status[lang]
   assert(status and status ~= 'installing')
+  install_status[lang] = nil
   return status
 end
 
@@ -570,7 +570,6 @@ end)
 ---@return string? err
 local function uninstall_lang(logger, lang, parser, queries)
   logger:debug('Uninstalling ' .. lang)
-  install_status[lang] = nil
 
   if fn.filereadable(parser) == 1 then
     logger:debug('Unlinking ' .. parser)


clason:
Thank you. But isn't the issue rather that `install_status` is never reset when an(y) operation concludes? I think doing that (in the proper place(s)) would be the better fix.

@lewis6991

MeanderingProgrammer:
Happy to make any changes, I'm definitely lacking in context on how you were planning to use `install_status`.

Here's my suggestion just based on the current code.

First I don't think this chunk of code does what it's intended to do:

```lua
if install_status[lang] == 'installing' then
  vim.wait(INSTALL_TIMEOUT, function()
    return install_status[lang] ~= 'installing'
  end)
  install_status[lang] = 'timeout'
end
```

Since it doesn't use the output of `vim.wait` it'll end up setting status to `timeout` regardless of what happens, so even if the `install_status` changes within the `INSTALL_TIMEOUT` the return value will be `timeout`.

So I'll update this bit and from there I think the only thing that needs to be tracked is which languages are currently being installed.

```lua
local installing = {} ---@type table<string,boolean?>

---@async
---@param lang string
---@param cache_dir string
---@param install_dir string
---@param force? boolean
---@param generate? boolean
---@return boolean success
local function install_lang(lang, cache_dir, install_dir, force, generate)
  if not force and vim.list_contains(config.get_installed(), lang) then
    return true
  elseif installing[lang] then
    local success = vim.wait(INSTALL_TIMEOUT, function()
      return installing[lang] == nil
    end)
    return success
  else
    installing[lang] = true
    local err = try_install_lang(lang, cache_dir, install_dir, generate)
    installing[lang] = nil
    return err == nil
  end
end
```

And 2 other small changes that rely on `install_status` and the output of `install_lang`.

clason:
Hmm. I am all for simplifications, but we do need to properly catch failures (including timeouts, which need not necessarily be given a separate status). This is crucial for CI.

MeanderingProgrammer:
That makes sense if there is a plan to use the different states in the future.

Keeping around all the statuses and clearing `install_status` upon completion would look something like:

```lua
local function install_lang(lang, cache_dir, install_dir, force, generate)
  if not force and vim.list_contains(config.get_installed(), lang) then
    -- no longer sets 'installed' for already installed languages
    return 'installed'
  end

  if install_status[lang] then
    if install_status[lang] == 'installing' then
      vim.wait(INSTALL_TIMEOUT, function()
        return install_status[lang] ~= 'installing'
      end)
      -- should probably be conditioned on the result of vim.wait, but keeping as is
      -- unsure how the condition should function since at this point in time we can't use
      -- install_status to check if the task that was running succeeded or failed
      install_status[lang] = 'timeout'
    end
  else
    install_status[lang] = 'installing'
    local err = try_install_lang(lang, cache_dir, install_dir, generate)
    install_status[lang] = err and 'failed' or 'installed'
  end

  local status = install_status[lang]
  assert(status and status ~= 'installing')

  -- reset install_status before returning the result
  -- vim.wait function should still work since nil, 'failed', & 'installed' 
  -- all cause callback to return true
  -- install_status no longer stores the result of the previous execution,
  -- maybe a history table would be useful?
  install_status[lang] = nil

  return status
end
```

The only changes are:

- not setting `install_status[lang] = 'installed'` in the first if
- setting `install_status[lang] = nil` before returning

clason:
Another (simpler) idea would be to just reset `install_status = {}` at the top of the main (local) `install` function. Could you test whether that works for you?

MeanderingProgrammer:
I think that would work for me and I can test it but I think that could lead to edge cases for you guys. I'm not sure if I fully understand how the async boundaries work, but as an absurd example if I did:

```lua
for _ = 1, 10 do
  require('nvim-treesitter').install({ 'lua' })
end
```

Can multiple of these install tasks end up running concurrently? My current understanding is they can. More realistically this would happen with a call to install and a call to update that happen to be close together or something like that.

But if the tasks happen to run such that:

- `install_status` is cleared from first call to install
- `lua` `install_status` is `nil` so set to `installing` and task is scheduled
- `install_status` is cleared from second call to install
- `lua` `install_status` is `nil` again set to `installing` and task is scheduled
- `lua` install task breaks for lots of potential reasons by modifying shared resources

clason:
That is indeed an (unlikely) edge case. It's ok if this would install multiple times, but you are right they should not intersect. So let's go with your approach.

MeanderingProgrammer:
I assumed you meant to go with the simpler of my approaches, have updated with those changes, thanks for taking a look!

I have some suggestions on simplifying things a bit, but it's based on my limited understanding:

- track whether a language is `installing` in its own table
- remove the `installing` status
- updating the `timeout` status to only occur when a `timeout` is reached

```lua
---@alias InstallStatus
--- | 'installed'
--- | 'failed'
--- | 'timeout'

local installing = {} ---@type table<string,boolean?>
local install_status = {} ---@type table<string,InstallStatus?>

---@async
---@param lang string
---@param cache_dir string
---@param install_dir string
---@param force? boolean
---@param generate? boolean
---@return InstallStatus status
local function install_lang(lang, cache_dir, install_dir, force, generate)
  local status
  if not force and vim.list_contains(config.get_installed(), lang) then
    -- TODO: if installing[lang] == true, does it make sense to wait for the install to complete
    -- and report the result or to just report 'installed'?
    -- Essentially should this if statement be moved below the next one?
    status = 'installed'
  elseif installing[lang] then
    local finished = vim.wait(INSTALL_TIMEOUT, function()
      return not installing[lang]
    end)
    status = finished and install_status[lang] or 'timeout'
  else
    installing[lang] = true
    local err = try_install_lang(lang, cache_dir, install_dir, generate)
    status = err and 'failed' or 'installed'
    installing[lang] = nil
  end
  install_status[lang] = status
  return status
end
```

clason:
Sorry for the radio silence; I was hoping for input from the person who wrote the code but I guess that won't come.

I have looked a bit into this myself and I now agree that `install_status` is actually unnecessary as the states are never really used (in the sense of persistent states). So I think your suggestion from https://github.com/nvim-treesitter/nvim-treesitter/pull/7996#issuecomment-3056215425 is actually the way to go here.

Do you still want to make that change?

clason:
(I still merged this PR, as it fixes an important issue, but the further simplification would still be welcome.)

MeanderingProgrammer:
No worries, thanks for merging the change!

Posted a PR with the further simplification: https://github.com/nvim-treesitter/nvim-treesitter/pull/8021

PR #8004, opened 2025-07-15
ci: ensure parsers have a supported ABI version

ribru17:
I assume we already check this elsewhere, is it worth removing that logic?


clason:
I assumed that's the default?

> I assume we already check this elsewhere, is it worth removing that logic?

Yes, we check it in `check-parsers`, which is a quick "can Nvim load the parser?" step. I would like to keep it since it covers the actual thing I want to check. (I thought the check is more useful for grammar repos that want to lock the ABI for some reason...)

Of course, there's no reason not to add that to the query check as well.

ribru17:
> I assumed that's the default?

The default assumes all parser ABIs are supported; that is, if no ABIs are specified then the check will not be performed

clason:
Can't hurt; and we might decide to yeet (or regenerate) ABI 13 parsers down the line.

PR #8006, opened 2025-07-16
fix(ipkg) ipkg no longer exists and causes error when installing

keagster:


clason:
The master branch is frozen and will not be updated anymore.

PR #8009, opened 2025-07-18
fix(jinja): assign highlights to all variables and functions

rmuir:
The majority of jinja variables aren't assigned any highlights at all. Assign `@variable` to all identifiers.

Jinja filters without parameters are not highlighted as functions: add a query for these.

Refine the existing `@variable.member` to only capture identifiers instead of broader nodes.

Background: I configure yaml to inject this parser for ansible highlighting. But due to the missing highlight captures, most jinja is instead highlighted as yaml string (green), which is confusing. Additionally you can see another bug `abc.list_two` with the overly broad `@variable.member` capture.

Before:
jinja_before

After:
jinja_after

cc: @cathaysia (and thank you for this parser)

### DIFF
--- a/runtime/queries/jinja_inline/highlights.scm
+++ b/runtime/queries/jinja_inline/highlights.scm
@@ -71,11 +71,7 @@
   "as"
 ] @keyword.import
 
-(import_statement
-  (identifier) @variable)
-
-(import_as
-  (identifier) @variable)
+(identifier) @variable
 
 [
   "if"
@@ -109,11 +105,33 @@
 
 (expression
   "."
-  (expression)+ @variable.member)
+  (expression
+    (binary_expression
+      .
+      (unary_expression
+        (primary_expression
+          (identifier) @variable.member)))))
+
+(expression
+  "."
+  (expression
+    (binary_expression
+      (binary_expression
+        (unary_expression
+          (primary_expression
+            (identifier) @variable.member))))))
 
 (assignment_expression
   "."
   (identifier)+ @variable.member)
+
+; jinja filters
+(binary_expression
+  (binary_operator
+    "|")
+  (unary_expression
+    (primary_expression
+      (identifier) @function.call)))
 
 (inline_trans
   "_" @function.builtin)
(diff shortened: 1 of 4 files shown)


cathaysia:
>  I configure yaml to inject this parser for ansible highlighting

Maybe this is helpful for you? https://github.com/cathaysia/nvim-jinja

rmuir:
@cathaysia thank you, that may be helpful for `yaml.j2` files. 

In the case here, files are yaml (e.g. ansible playbooks and task files). These are not templated files, just plain yaml, but they support jinja snippets injected inside the string values. Ansible files of this type are commonly used in the devops world. Example playbook: https://github.com/apache/lucene/blob/main/dev-tools/aws-jmh/playbook.yml

So it is a bit of a different problem than the templated `yaml.j2` one, I solve it with ordinary injections.

<details>
<summary>queries/yaml/injections.scm</summary>

```
; extends

; jinja injections inside quoted yaml values
(block_mapping_pair
  value: (flow_node
    [
      (double_quote_scalar)
      (single_quote_scalar)
    ] @injection.content
    (#lua-match? @injection.content "{[%%{].*[%%}]}")
    (#set! injection.language "jinja")
    (#offset! @injection.content 0 1 0 -1)))

; jinja injections inside yaml unquoted string value
(block_mapping_pair
  value: (flow_node
    (plain_scalar
      (string_scalar) @injection.content
      (#lua-match? @injection.content "{[%%{].*[%%}]}")
      (#set! injection.language "jinja"))))

; jinja injections inside yaml block string value
(block_mapping_pair
  value: (block_node
    (block_scalar) @injection.content
    (#lua-match? @injection.content "{[%%{].*[%%}]}")
    (#set! injection.language "jinja")))
```
</details>

ribru17 (approved):

runtime/queries/jinja_inline/highlights.scm
```diff
@@ -109,12 +105,34 @@
...
   (identifier)+ @variable.member)
 
+; jinja filters
+(binary_expression
+  (binary_operator
+    "|")
+  (unary_expression
+    (primary_expression
+      (identifier) @function.call)))
+
```
clason, line 135: This pattern surfaced a bug in the jinja_inline parser when it's generated from grammar (which seems not to have happened for quite a while on the grammar repo).

Since this breaks tests on, e.g., tree-sitter, I'll have to remove this until the upstream bug is fixed.

PR #8014, opened 2025-07-20
fix(nix): point parser to new upstream

zimbatm:
The repo moved location

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1537,7 +1537,7 @@ list.ninja = {
 
 list.nix = {
   install_info = {
-    url = "https://github.com/cstrahan/tree-sitter-nix",
+    url = "https://github.com/nix-community/tree-sitter-nix",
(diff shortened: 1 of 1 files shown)


clason:
Thank you. The master branch is frozen, but I'll make the change on the main branch when I next update it. (Github is very good with redirects, so there's no urgency.)

clason:
Oh, and you may want to let upstream know that ABI 13 is extremely outdated and probably no longer be supported by tree-sitter in the near future. When that happens, we'll drop such parsers as well.

zimbatm:
Thanks for the heads up about the ABI. I have a update to tree-sitter 0.25 in the works, so that should be coming soon.

For the base branch, if you change it in the repo setting, then fresh checkouts will pick that up. I should have read more carefully the README of course. Pushed #8015 in case that's helpful, I won't be offended if you close it too.

clason:
Yes, I know, but then everybody who downloads the repo will suddenly get the incompatible rewrite, which is massively breaking. I don't want to do that (yet).

PR #8015, opened 2025-07-20
fix(nix): update repo url

zimbatm:
Also fix the upstream location.


clason:
We have tooling for such parser updates; a manual PR is only needed (and wanted) if it requires or includes changes to the _queries_.

And you need to run `make docs` to update the parser list with the new location and maintainer.

zimbatm:
got it. with the update commit you just pushed and this change we should be all up2date.

PR #8019, opened 2025-07-20
Create folds.scm for glimmer_javascript

NullVoxPopuli:


clason:
(You can add both files in one PR)

PR #8021, opened 2025-07-20
refactor(install): replace status enum with boolean

MeanderingProgrammer:
## Details

Follow up to: https://github.com/nvim-treesitter/nvim-treesitter/pull/7996

Currently in the process of installing a parser the status is tracked via the `install_status` table. However that status is not used beyond simply being persisted. Rather than tracking the status now we only track which languages are in the process of being installed to avoid running multiple installs for the same language at the same time.

The result of the `install_lang` function is similarly changed to a boolean indicating whether the installation step succeeded.

> [!NOTE]
> The case of waiting for a language to finish installing is slightly special.
> The return value does not indicate the successful installation of a language, rather it means the install step completed within the timeout, but it may have actually failed.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -439,13 +439,7 @@ local function try_install_lang(lang, cache_dir, install_dir, generate)
   logger:info('Language installed')
 end
 
----@alias InstallStatus
---- | 'installing'
---- | 'installed'
---- | 'failed'
---- | 'timeout'
-
-local install_status = {} ---@type table<string,InstallStatus?>
+local installing = {} ---@type table<string,boolean?>
 
 ---@async
 ---@param lang string
@@ -453,29 +447,21 @@ local install_status = {} ---@type table<string,InstallStatus?>
 ---@param install_dir string
 ---@param force? boolean
 ---@param generate? boolean
----@return InstallStatus status
+---@return boolean success
 local function install_lang(lang, cache_dir, install_dir, force, generate)
   if not force and vim.list_contains(config.get_installed(), lang) then
-    return 'installed'
-  end
-
-  if install_status[lang] then
-    if install_status[lang] == 'installing' then
-      vim.wait(INSTALL_TIMEOUT, function()
-        return install_status[lang] ~= 'installing'
-      end)
-      install_status[lang] = 'timeout'
-    end
+    return true
+  elseif installing[lang] then
+    local success = vim.wait(INSTALL_TIMEOUT, function()
+      return not installing[lang]
+    end)
+    return success
   else
-    install_status[lang] = 'installing'
+    installing[lang] = true
     local err = try_install_lang(lang, cache_dir, install_dir, generate)
-    install_status[lang] = err and 'failed' or 'installed'
+    installing[lang] = nil
+    return not err
   end
-
-  local status = install_status[lang]
-  assert(status and status ~= 'installing')
-  install_status[lang] = nil
-  return status
 end
 
 --- Reload the parser table and user modifications in case of update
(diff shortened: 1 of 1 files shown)


clason:
Thanks!

clason:
So looks like this did in fact break the use of nvim-treesitter in various CI jobs -- some parsers simply fail to be installed, without any error message; see, e.g., https://github.com/tree-sitter/tree-sitter/actions/runs/16449276808/job/46581455909?pr=4634

It looks like it times out now where it didn't before. Clearly, that's not good...

Unfortunately, I can't reproduce this locally -- could you please take a look?

MeanderingProgrammer:
Crap, yes, I can definitely look into this!

Just to make sure I'm on the same page, are you seeing consistent failures, or after this change the builds in general are more flaky?

It does look like some more recent runs that include this change (like the ones triggered by https://github.com/nvim-treesitter/nvim-treesitter/commit/5d539943c6a8b7e4aa7c77c7b79517f07be5454f) are succeeding.

clason:
Yes, it's flaky, but it definitely did not fail that much (and consistently) before this commit. 

Hence my suspicion that it's timeout related. 

(We don't see it here since I have throttled the installation, which prevents the situation it seems. That is an easy fix, but I would still like to understand the reason for the regression.)

MeanderingProgrammer:
Since I'm also not able to reproduce the behavior, I can only guess at what's going on but I'm fairly confident in it.

I think the issue stems from a combination of:

1. handling timeouts as a failure (new behavior) 
2. dependencies being added multiple times to the list of languages to install causing timeouts (existing behavior)

So for the first part that's strictly because of this change. Before if `INSTALL_TIMEOUT` was ever reached the `install_status` was set to `timeout` and returned. The check on the return value was `status ~= 'failed'`. So timeouts were likely being hit before but they ended up counting as successes.

The second part is what I believe causes the timeout to be reached and it's because in the call to `norm_languages` dependencies are added as such:

```lua
  if not (skip and skip.dependencies) then
    for _, lang in pairs(languages) do
      if parsers[lang] and parsers[lang].requires then
        vim.list_extend(languages, parsers[lang].requires)
      end
    end
  end
```

This means that shared dependencies are added potentially many times over. For example when we already have the list of all languages `ecma` is going to be in that list. Then we hit `tsx` which adds `ecma` again and `typescript`. Then we hit `typescript` not once but twice (once from the original list and again because of `tsx`) which then adds `ecma` 2 more times and as a result the languages list will contain `ecma` 4 times. At that point for `ecma` one task can be stuck waiting for installing to be `nil` but it ends up timing out because additional install tasks for language get scheduled just as others finish.

So as a summary I believe tasks were timing out before this change, however previously those timeouts counted as successes. The timeouts are not actually caused by things taking too long, but due to scheduling install tasks for the same language multiple times.

I think the fix for this is to de-duplicate the list of languages to install which I'll send out a PR for shortly.

clason:
Good call on the non-parser languages!

MeanderingProgrammer:
Here's the PR: https://github.com/nvim-treesitter/nvim-treesitter/pull/8036

Tests pass on this one, which is a good signal.

clason:
Unfortunately not, since the throttling prevents the flake. If you want to check, remove (in a separate commit to be dropped later) the `--max-jobs=10` flag in the `test-core` workflow.

MeanderingProgrammer:
Makes sense: https://github.com/nvim-treesitter/nvim-treesitter/pull/8037

clason:
I mean to add this to your PR to verify that it fixes the issue ;)

MeanderingProgrammer:
Ah, I made it a separate PR with both changes to kick off the tests, and linked the results in the PR with the fix.

Did you still want me to launch it directly from: https://github.com/nvim-treesitter/nvim-treesitter/pull/8036?

clason:
ah, ok, that works too

PR #8024, opened 2025-07-21
fix(jinja): fix crashing query pattern

rmuir:
The pattern used to match jinja filters mysteriously works correctly and passes tests with the current parser, but trips a treesitter "impossible" query assert if you regenerate the jinja_inline parser.

Replace with pattern that passes with both current parser and regenerated jinja_inline parser.

See https://github.com/cathaysia/tree-sitter-jinja/issues/37


runtime/queries/jinja_inline/highlights.scm
```diff
@@ -125,14 +125,14 @@
...
-; ; jinja filters
-; (binary_expression
-;   (binary_operator
-;     "|")
-;   (unary_expression
-;     (primary_expression
-;       (identifier) @function.call)))
+; jinja filters
+(binary_expression
+  (binary_operator) @_operator
```
clason, line 130:
```suggestion
  "|" @_operator
```
Wouldn't that be simpler here?

clason:
I'm not comfortable making _any_ changes here until I understand the situation.

rmuir:
@clason I agree: converted to draft. I just feel responsible for adding the crashing query here...

clason:
Honestly, after looking through the grammar, the only change I'd be comfortable with is dropping it entirely...

clason:
Can you rebase on main? The parsers were updated.

**EDIT**: Did it myself.

PR #8036, opened 2025-07-24
fix(install): skip duplicate dependencies

MeanderingProgrammer:
## Details

Follow up to: https://github.com/nvim-treesitter/nvim-treesitter/pull/8021

More info on why duplicate dependencies can cause problems: https://github.com/nvim-treesitter/nvim-treesitter/pull/8021#issuecomment-3114962375

Tests without `--max-jobs`:  https://github.com/nvim-treesitter/nvim-treesitter/actions/runs/16518667608

### DIFF
--- a/lua/nvim-treesitter/config.lua
+++ b/lua/nvim-treesitter/config.lua
@@ -164,7 +164,8 @@ function M.norm_languages(languages, skip)
     end
   end
 
-  return languages
+  table.sort(languages)
+  return vim.fn.uniq(languages) --[=[@as string[]]=]
 end
 
 return M


clason:
Why not simply deduplicate the `languages` list before returning it? That would also guard against user error like `:TSInstall lua vim lua`.

Unfortunately, nvim doesn't have a function for that in its stdlib, but you can find lots of snippets on the interwebs. Since this is not performance-critical, you could even do this with a `table.sort` plus `vim.fn.uniq`?

MeanderingProgrammer:
> Why not simply deduplicate the languages list before returning it? That would also guard against user error like :TSInstall lua vim lua.

Makes sense, updated.

clason:
Can you repeat the test for CI (here)?

lua/nvim-treesitter/config.lua
```diff
@@ -157,14 +167,14 @@ function M.norm_languages(languages, skip)
...
-    for _, lang in pairs(languages) do
+    for _, lang in ipairs(languages) do
       if parsers[lang] and parsers[lang].requires then
         vim.list_extend(languages, parsers[lang].requires)
       end
     end
   end
 
-  return languages
+  return unique(languages)
```
clason:
```suggestion
  table.sort(languages)
  vim.fn.uniq(language)
  return languages
```

Less code, and we get a nice sorted list as a bonus ;)

(Only if this works, of course -- I haven't tested it.)

clason:
Please see the edited comment (and you don't need `ipairs` if you sort later).

lua/nvim-treesitter/config.lua
clason: Hmm, that doesn't work locally; I still get "352/368 languages installed". It also looks to be slower (since sorting means you hit split parser repos in parallel).

MeanderingProgrammer: It should work once `vim.fn.uniq(language)` is fixed to `vim.fn.uniq(languages)`

clason: No, I tested that locally (without the typo). But you're right, it _should_ work.

MeanderingProgrammer: Huh, that was unexpected

clason: Yup. Looks like there's a bug in `vim.fn.uniq`, it just doesn't work.

So let's go with your original approach, and I'll report this upstream -- maybe this will _finally_ convince people to add a `vim.list_unique` in Neovim!

Sorry for the runaround.

MeanderingProgrammer: It looks like the returned list from `local languages = vim.fn.uniq(languages)` is correct, but the in place modification part is somehow not working? Very strange.

clason: Well, that's sort of expected -- you can't modify table elements across the Vimscript-Lua bridge.

The issue is that the annotation is incorrect in this case, which breaks our type checking. So no matter which way you turn, CI fails ;)

clason: Ah, but wait, maybe we can brute-force it

MeanderingProgrammer: Yeah, I took the "trust me it's definitely a string[]" annotation approach

lua/nvim-treesitter/config.lua
```diff
@@ -157,14 +157,15 @@ function M.norm_languages(languages, skip)
   end
 
   if not (skip and skip.dependencies) then
-    for _, lang in pairs(languages) do
+    for _, lang in ipairs(languages) do
```
clason:
```suggestion
    for _, lang in pairs(languages) do
```

lua/nvim-treesitter/config.lua
```diff
@@ -157,14 +157,15 @@ function M.norm_languages(languages, skip)
...
       if parsers[lang] and parsers[lang].requires then
         vim.list_extend(languages, parsers[lang].requires)
       end
     end
   end
 
-  return languages
+  table.sort(languages)
+  return vim.fn.uniq(languages) ---@type string[]
```
clason, line 168:
```suggestion
  return vim.fn.uniq(languages) --[=[@as string[]]=]
```
is I think more correct?

clason:
```suggestion
  return vim.fn.uniq(languages) --[[ @as string[] ]]
```
is I think more correct

lua/nvim-treesitter/config.lua
clason: Yeah, that looks to be a luals bug with union types. (emmyluals has a different issue with this...)

lua/nvim-treesitter/config.lua
MeanderingProgrammer: Why change this? `ipairs` is the normal way of iterating over integer indexed tables. I mean you can technically use `pairs` in all cases so long as order isn't important.

MeanderingProgrammer:
> Can you repeat the test for CI (here)?

Results of CI without `--max-jobs`: https://github.com/nvim-treesitter/nvim-treesitter/actions/runs/16518667608

lua/nvim-treesitter/config.lua
clason: Well, the question _is_ "why change this?" -- you made that change ;)

And the argument is indeed that order is not important because we sort later anyway.

It's not a big deal, but I prefer for a `fix` to make the minimal change necessary.

MeanderingProgrammer: Was something I modified when the parts I was changing were specific to dependencies. Figured why not leave it, but fair enough.

clason:
Thanks!

PR #8051, opened 2025-08-01
feat(systemverilog)!: update parser and queries

clason:


clason:
Systemverilog added a test file corpus which contains symlinks, breaking extraction on Windows. @gmlarumbe

PR #8065, opened 2025-08-10
feat(orca): add support for ORCA input files

kszenes:
ORCA is a software package to perform quantum chemistry simulations: https://www.faccts.de/orca/.
Note that the filetype `inp` has been added such that `*.inp` files would be matched.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1622,19 +1622,19 @@ list.odin = {
 }
 
 -- Define filetype for ORCA input files
-vim.filetype.add({
-	extension = {
-		inp = "inp",
-	},
-})
+vim.filetype.add {
+  extension = {
+    inp = "inp",
+  },
+}
 
 list.orca = {
   install_info = {
     url = "https://github.com/kszenes/tree-sitter-orca.git",
-    files = { "src/parser.c" }
+    files = { "src/parser.c" },
   },
   filetype = "inp",
-  maintainers = { "@kszenes" }
+  maintainers = { "@kszenes" },
 }
 
 list.pascal = {
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -1621,6 +1621,22 @@ list.odin = {
   maintainers = { "@amaanq" },
 }
 
+-- Define filetype for ORCA input files
+vim.filetype.add({
+	extension = {
+		inp = "inp",
+	},
+})
+
+list.orca = {
+  install_info = {
+    url = "https://github.com/kszenes/tree-sitter-orca.git",
+    files = { "src/parser.c" }
+  },
+  filetype = "inp",
+  maintainers = { "@kszenes" }
+}
+
 list.pascal = {
   install_info = {
     url = "https://github.com/Isopod/tree-sitter-pascal",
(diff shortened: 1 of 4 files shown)


lua/nvim-treesitter/parsers.lua
```diff
@@ -1621,6 +1621,22 @@ list.odin = {
   maintainers = { "@amaanq" },
 }
 
+-- Define filetype for ORCA input files
+vim.filetype.add {
+  extension = {
+    inp = "inp",
```
ribru17, line 1627: This should be done in the vim repository rather than here (by adding filetype detection)

queries/orca/highlights.scm
```diff
@@ -0,0 +1,276 @@
+(comment) @comment
+
+(integer) @number
+
+(float) @float
```
ribru17, line 5:
```suggestion
(float) @number.float
```

queries/orca/highlights.scm
```diff
@@ -0,0 +1,276 @@
...
+
+(element) @string
+
+(simple_line) @variable.parameter @markup.strong
```
ribru17, line 9: I think `@markup.strong` should be removed, unless this is really meant to convey a bold markup item. Can you explain why this has two captures?

queries/orca/highlights.scm
```diff
@@ -0,0 +1,276 @@
...
+
+(quoted_string) @string
+
+(geom_line_types) @type
+
+"xyz" @type
+
+"int" @type
+
+"gzmt" @type
```
ribru17, line 23:
```suggestion
[
  (geom_line_types)
  "xyz"
  "int"
  "gzmt" 
] @type
```

queries/orca/highlights.scm
```diff
@@ -0,0 +1,276 @@
...
+
+"," @punctuation.delimiter
+
+; General word values (LAST - lower priority)
+(value_atom
+  (word) @number) ; Words used as values get same color as other values
+
+; Internal coordinate highlighting
+(int_line
+  connect1: (integer) @operator)
```
ribru17, line 94: Is this right? I would think integers should be `@number`

queries/orca/highlights.scm
```diff
@@ -0,0 +1,276 @@
...
+
+(compound_additive_expr
+  "-" @operator)
+
+; Assignment operators
+(compound_assignment
+  "=" @operator)
+
+(compound_array_assignment
+  "=" @operator)
```
ribru17, line 250: Instead of having many patterns for all possible sightings of e.g. an `=`, it is better to have one pattern for e.g. all operators:

```
[
  "="
  "+"
  "/"
  ; ...
] @operator
```

the alteration syntax `[ ]` makes this easy :+1: 

(Same with punctuation.delimiter, etc.)

queries/orca/highlights.scm
```diff
@@ -0,0 +1,276 @@
...
+(compound_for_loop) @keyword
+
+(compound_if_block) @keyword
+
+; For loop keywords and components
+(compound_for_loop
+  (variable_name) @function)
+
+; If/else condition components
+(compound_condition) @conditional
```
ribru17, line 199: This is not a valid capture

clason:
You have opened this PR against the frozen (and locked) master branch. Please open a new one against "main" (but first read the README and CONTRIBUTING guide on that branch!)

PR #8066, opened 2025-08-10
fix(typescript): support type param for styled components

svirschevskiy:
Added support for type arguments for styled components. Also inherit tsx injections from typescript. 
Seems like tests are failing in the `main` branch, so I skipped them.

| Before    | After |
| --------      | ------- |
| Screenshot_2025-08-10_19-09-43  | Screenshot_2025-08-10_19-10-32    |

### DIFF
--- a/runtime/queries/tsx/injections.scm
+++ b/runtime/queries/tsx/injections.scm
@@ -1,1 +1,1 @@
-; inherits: ecma,jsx
+; inherits: typescript,jsx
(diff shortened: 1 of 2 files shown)


ribru17 (approved):

ribru17:
Sorry for the delay

Decodetalkers:
I find that the injections not work on my platform. I have already install the treesitter css , tsx, and typescript.. but

image
it works before, only after I switch to the main branch..

sorry, I see, I need to install the language styled

PR #8073, opened 2025-08-13
Allow comments to be folded for ecma and ecma derivations

NullVoxPopuli:


queries/ecma/folds.scm
```diff
@@ -1,4 +1,5 @@
 [
+  (comment)
```
lucario387, line 2: This should be `(comment)+` or be limited to `/** */` comments, I think. @ribru17

clason:
The master branch is frozen; all further PRs need to be made to the `main` branch (but read the CONTRIBUTING _on that branch_  first).

PR #8074, opened 2025-08-13
allow comments to be folded for ecma languages 

NullVoxPopuli:


clason:
Please address @ribru17's comment.

PR #8077, opened 2025-08-15
feat(cpp): Add highlights capture group for lambda brackets

MartyMcFlyInTheSky:
When creating a colortheme I noticed that there's no way to target lambda capture brackets in c++, so I added this capability under the @lambda.capture.bracket specifier.

This for example sets lambda capture brackets to orange:

```
    -- highlight lambda capture brackets
    vim.api.nvim_set_hl(0, "@lambda.capture.bracket", { fg = "#ff8400"})
```

### DIFF
--- a/queries/cpp/highlights.scm
+++ b/queries/cpp/highlights.scm
@@ -30,10 +30,10 @@
   declarator: (field_identifier) @function.method)
 
 ; lambdas
-((lambda_expression
+(lambda_expression
   captures: (lambda_capture_specifier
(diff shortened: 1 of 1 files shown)


clason:
Sorry, but queries here are only allowed to use the documented captures/highlight groups in CONTRIBUTING (or `:h treesitter-highlight-groups`). Such customizations belong in your personal config (you can extend the queries by adding the same file in your `.config/nvim` path with `;; extends` in the first line).

PR #8084, opened 2025-08-19
feat(install): link queries with relative path

henry-hsieh:
The queries are linked with absolute paths. This makes the environment less portable. Therefore, I modified the `do_link_queries` to link with relative path in first priority. If the common base of `query_dir` and `query_src` is root ("/" or "%a:/"), use the absolute path as usual.

The fnamemodify() and vim.fs.relpath() cannot be directly used because the `query_src` is usually not the subdirectory of `query_dir`. Therefore, I used `vim.fs.parents()` to find the command base of these two directories.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -320,6 +320,21 @@ end
 ---@return string? err
 local function do_link_queries(logger, query_src, query_dir)
   uv_unlink(query_dir)
+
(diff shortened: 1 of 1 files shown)


clason:
Thank you, but I want to keep the code simple. Portability is not a priority.

henry-hsieh:
Sorry to heard that. I'll do another approach. Thanks!

PR #8095, opened 2025-08-29
docs: Change query legend to match table

OninDynamics:
It's a really tiny and insignificant change to the SUPPORTED_LANGUAGES.md file.


clason:
Thanks, but that file is autogenerated.

PR #8099, opened 2025-08-30
chore: bump Agda parser

foxyseta:
This includes a patch for the crash described in https://github.com/tree-sitter/tree-sitter-agda/issues/24.


clason:
Thank you, but parser updates are batched. PRs are only needed if there's a breaking change that requires query adaptation.

(And the `master` branch is locked.)

PR #8106, opened 2025-09-06
feat(snakemake): update queries

osthomas:
This PR updates indents, locals and highlights queries for snakemake, and adds tests for indent queries.

PS: To run tests for python/snakemake only, I deleted all *_spec.lua files except for snakemake_spec.lua via `find tests/ -depth -name '*python*' -prune -o -name '*_spec*' -print -delete`. Is there a less drastic way to select which tests to run via `make tests`?

### DIFF
--- a/runtime/queries/snakemake/highlights.scm
+++ b/runtime/queries/snakemake/highlights.scm
@@ -63,20 +63,20 @@ body: (_
 ; directive labels in wildcard context
 ((wildcard
   (identifier) @label)
-  (#any-of? @label "input" "log" "output" "params" "resources" "threads" "wildcards"))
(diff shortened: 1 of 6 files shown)


clason:
Yes, you can specify via environment variables, just like for busted. E.g.,
```bash
make tests TESTS=indent TEST_FILE=snakemake_spec.lua
```

osthomas:
> Yes, you can specify via environment variables, just like for busted. E.g.,
> 
> ```shell
> make tests TESTS=indent TEST_FILE=snakemake_spec.lua
> ```

This unfortunately doesn't work, at least on my end. It still runs all tests.

clason:
Sorry, should have been just `make tests TESTS=indent/snakemake_spec.lua`

PR #8109, opened 2025-09-08
Teach protobuf parser about editions

stefanvanburen:
Handled similarly to `syntax = "proto{2,3}"` lines.

Ref: https://protobuf.dev/editions/overview/
Ref: https://buf.build/blog/protobuf-editions-are-here


stefanvanburen:
ah, I realize I need to add this stuff to the [upstream grammar repo](https://github.com/treywood/tree-sitter-proto), which doesn't look too maintained. ah well!

PR #8120, opened 2025-09-14
Support alternative git host with different archive structure

madmaxieee:
Some git hosting services name the directory inside their tarball archive differently. For example [codeberg](https://codeberg.org/), unlike github, does not add the revision suffix to the direcotry in the tarball.

I had problem installing this parser:
```lua
require("nvim-treesitter.parsers").d2 = {
  tier = 0,
  install_info = {
    url = "https://codeberg.org/p8i/tree-sitter-d2.git",
    queries = "queries",
  },
}
```

Tarball archives download from GitHub's main branch would extract into a directory named `tree-sitter-d2-main`, but tarball archives from Codeberg would not have the `-main` suffix. Causing the current download flow to fail to locate the extracted directory and not give clear error message.

I added a list of candidates that we may search for the extracted directory using `vim.uv.fs_lstat`, including the original `repo_project_name .. '-' .. dir_rev` and the new `repo_project_name`, to solve this problem.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -261,7 +261,27 @@ local function do_download(logger, url, project_name, cache_dir, revision, outpu
   do -- Move tmp dir to output dir
     local dir_rev = revision:find('^v%d') and revision:sub(2) or revision
     local repo_project_name = url:match('[^/]-$')
-    local extracted = fs.joinpath(tmp, repo_project_name .. '-' .. dir_rev)
+
+    ---@type string?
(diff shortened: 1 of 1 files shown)


clason:
I know, but these are explicitly not supported -- this is too brittle, and all supported parsers are at least mirrored on github or gitlab, so this is not worth the code complexity.

(Support for installing unsupported parsers is a byproduct, not a primary design goal.)

clason:
(For a more full-featured custom installer, see https://github.com/lewis6991/ts-install.nvim.)

PR #8124, opened 2025-09-16
docs: Clarify tree-sitter-cli requirement

savq:
Link to the tree-sitter-cli README directly, instead of the tree-sitter top-level README to avoid confusion.


clason:
Seems like most distros have split out the CLI package under this name, so this is probably indeed better now. (Although the CLI readme is so anemic, it won't be much more helpful than the top-level one.)

PR #7788, opened 2025-04-02
fix(python): Highlight all types of docstrings

jbdyn:
I read [PEP 257](https://peps.python.org/pep-0257/#what-is-a-docstring) again. It says:

> A docstring is a string literal that occurs as the first statement in a module, function, class, or method definition. Such a docstring becomes the `__doc__` special attribute of that object.

but also in the paragraph after the next:

> String literals occurring elsewhere in Python code may also act as documentation.

So, *every* string literal can be treated as a docstring (and is valid for Python as long as the indentation is correct), even with single quotes.

Since we are not distinguishing between
- docstrings that become the `__doc__` special attribute,
- package / module / class / function / attribute docstrings and 
- additional docstrings,

my suggestions is to just say: Every standalone string in a Python script is a documenting string (but not necessarily *the* docstring).

The highlight query then just boils down to

```
(expression_statement
  (string
    (string_content) @spell) @string.documentation)
```

If one would want to distinguish between the different types of docstrings, then this is obviously not sufficient, but I personally cannot imagine how one would benefit from that.
Differentiating between comments and docstrings, yes, sure, but not between types of docstrings.

Fixes #7786

PR #8131, opened 2025-09-18
fix(python): correct highlighting of docstrings

ValdezFOmar:
The node `expression_statement` was changed to a supertype in the python parser, which caused a bug where the docstring query captures all strings as `@string.documentation`.
This commit restores the behavior as it was discussed in #7788.

I tested this visually (I don't know how to run the tests) with `tests/query/highlights/python/docstrings.py`.

Minimal reproduction:

```lua
vim.opt.runtimepath:append('.')
require('nvim-treesitter').setup({ install_dir = vim.fn.getcwd() .. '/site' })
require('nvim-treesitter').install({ 'python' }):wait()

vim.api.nvim_set_hl(0, '@string.documentation.python', { fg = '#FF0000' })

vim.api.nvim_create_autocmd('FileType', {
  pattern = 'python',
  callback = function()
    pcall(vim.treesitter.start)
  end,
})
```

Before this commit:
image

After this commit:
image

### DIFF
--- a/runtime/queries/python/highlights.scm
+++ b/runtime/queries/python/highlights.scm
@@ -80,7 +80,11 @@
 ] @string.escape
 
 ; doc-strings
-(expression_statement
+(module
+  (string
+    (string_content) @spell) @string.documentation)
+
+(block
   (string
     (string_content) @spell) @string.documentation)
 


clason:
> I tested this visually (I don't know how to run the tests) 

`make tests`

And your fix doesn't seem correct(ly committed) -- CI fails due to invalid pattern.

clason:
And, yes, the supertype change was a pretty intrusive change -- and one that wasn't caught sufficiently by CI since it doesn't actually break queries, it just makes them work incorrectly. When you have fixed the query, please add corresponding tests so we can guard against future regressions.

ValdezFOmar:
I investigated a bit and it seems that `tree-sitter-snakemake` depends on a old version of `tree-sitter-python` (though the breaking change is still not released on any version), which produces the "impossible pattern" error because snakemake expects a `expression_statement` to be the parent of a docstring.
I built the snakemake parser using the latest commit from the python parser and was able to run the tests with no errors.

clason:
Can you make a PR to snakemake (if necessary, updating the queries there) and link it here?

clason:
@osthomas

ValdezFOmar:
~Yes, I'll open a PR there~. Wasn't fast enough :p.

osthomas:
tree-sitter-snakemake was now updated for tree-sitter-python 0.25.0.

Queries should be unaffected, because they inherit from Python, which is what caused this failure in the first place.

Would it be better to inline the Python queries in the Snakemake-specific ones to prevent such version mismatches?

clason:
> tree-sitter-snakemake was now updated for tree-sitter-python 0.25.0.

👍 

> Queries should be unaffected, because they inherit from Python, which is what caused this failure in the first place.

No, the issue in this specific case is more subtle: the Python grammar changed the _meaning_ of nodes, which required changes to keep the semantics. The same may (or may not) be true for the snakemake (specific) queries. (Since we don't have good highlight test coverage for snakemake, we wouldn't catch such regressions in our CI.)

> Would it be better to inline the Python queries in the Snakemake-specific ones to prevent such version mismatches?

I don't think that would make much difference; as long as the _grammars_ depend on each other, you'll run into issues one way or another. The real solution is to make each grammar _and_ queries stand-alone instead of the current house of cards. (This is a general problem; Python/Snakemake is child's play compared to the mess that is the C family...)

clason:
@ValdezFOmar can you bump the snakemake parser in this PR (as a separate commit _before_ the Python fix)?

ValdezFOmar:
> tree-sitter-snakemake was now updated for tree-sitter-python 0.25.0.

I don't think that is enough, the breaking change was made after that released and there is no released with that commit :(.

osthomas:
Right, sorry, I got confused. This actually does not address the issue here.

I think a release of tree-sitter-python would be good at this point, so that the dependency can be updated.

clason:
> I think a release of tree-sitter-python would be good at this point, so that the dependency can be updated.

Hmm, I wouldn't expect that in the near future (and I have no influence over those parsers).

I'll have to decide how to handle this; probably I'll revert the bump and mark the Python parser as "unmaintained" (which locks it).

osthomas:
I've updated the parser against the tree-sitter-python dev version, maybe use this tree-sitter-snakemake branch for now?

https://github.com/osthomas/tree-sitter-snakemake/tree/ts-python-0.25.0dev

clason:
Thank you, but I prefer not to use branches. We want to move away from "experimental"/"work in progress" parsers.

Ideally, everything would be following proper semver practices and coordinated releases so we could avoid such problems entirely, but alas.

clason:
I'm reverting the breaking change (#8138) and pinning the parser, so this is no longer necessary.

Keep an eye out for the next release, which will pull in the breaking change again. @osthomas I am assuming (and strongly encouraging) that your parser follows the Python releases closely, so we will need to bump both in unison to avoid regressions.

PR #8133, opened 2025-09-18
fix(c_sharp): missing highlight for parameter modifier

sharpchen:


clason:
It's always helpful to add context to your PRs (and commit descriptions) -- e.g., a code sample and before/after screenshots so maintainers who don't write C# can evaluate the change (and in particular whether the capture is used consistently).

sharpchen:
Ok, I added screenshot and sample code

amaanq:
crap, misclicked on mobile

PR #8140, opened 2025-09-19
Adding cedars policy language with queries

edmondop:
Name of language
[cedars](https://www.cedarpolicy.com/)

Representative code sample
Parser repo
https://github.com/chrnorm/tree-sitter-cedar

Parsed tree for code sample
Queries
Source of queries: "written from scratch"

Screenshots of code sample

Screenshot Treesitter Cedar

### DIFF
--- a/queries/cedar/folds.scm
+++ b/queries/cedar/folds.scm
@@ -1,0 +1,6 @@
+[
+  (policy)
+  (record_literal)
+  (set_literal)
+  (condition)
+] @fold
--- a/queries/cedar/highlights.scm
+++ b/queries/cedar/highlights.scm
@@ -1,0 +1,51 @@
+
+; Main policy structure
+(policy) @keyword
+(effect) @keyword
+(scope) @keyword
+
+; Expressions
+(binary_expression) @operator
+(unary_expression) @operator
+(call_expression) @function.call
+(ext_fun_call) @function.call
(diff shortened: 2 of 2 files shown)


clason:
Missing the actual parser

clason:
And please use as specific captures as possible.

clason:
And before you waste your time, that parser does not look maintained; we have stricter quality standards these days.

edmondop:
Thanks, can you tell me what are the characteristics for the parser? I can create a new treesitter parser I can maintain

clason:
Well, if the last commit (of ten) was a year ago, that is a pretty definitive sign...

(This is not academic; tree-sitter itself makes relatively frequent breaking changes, which parsers need to adapt to or become unusable -- which then breaks things for us.)

edmondop:
I am starting to get involved in Cedar, and I use Neovim, so I really care about getting this. However the grammar hasn't changed, what makes you expect that the parser would change? I am happy to address all other issues, but if the grammar doesn't chamge, a parse r wouldn't be updated for a year

clason:
See above: tree-sitter itself changes. I need to be able to rely on an active maintainer reacting to such changes in a timely manner, otherwise it will clog our CI. And I don't want that. 

So without an active parser maintainer, I don't want the parser. We have enough on our plate already.

PR #8160, opened 2025-09-30
regex failed to match last segment of url

bramadityaw:
I tried to install a parser through a git url, and it failed with the following error.

`Could not rename temp: ENOENT no such file`

It tried to rename <repo_project_name ..  dir_rev> into <output_dir>, but the call to `match` before it returned nil, which is the value of repo_project_name that got turned into an empty string.

The changed regex now matches the last url segment.

### DIFF
--- a/lua/nvim-treesitter/install.lua
+++ b/lua/nvim-treesitter/install.lua
@@ -260,7 +260,7 @@ local function do_download(logger, url, project_name, cache_dir, revision, outpu
 
   do -- Move tmp dir to output dir
     local dir_rev = revision:find('^v%d') and revision:sub(2) or revision
-    local repo_project_name = url:match('[^/]-$')
+    local repo_project_name = url:match('[^/]+$')
     local extracted = fs.joinpath(tmp, repo_project_name .. '-' .. dir_rev)
     logger:debug('Moving %s to %s/...', extracted, output_dir)
     local err = uv_rename(extracted, output_dir)


clason:
Can you please show exactly what you tried and what failed? The current code works fine for the hundreds of supported parsers.

bramadityaw:
I tried to install the parser for the language Koka through a git https URL, with the following install_info
```lua
{
  install_info = {
    url = 'https://github.com/bramadityaw/tree-sitter-koka',
    revision = '31fb0c4e76025da934075ed83bba7d51567e95c8',
    branch = 'issue-7',
    queries = 'queries',
  },
}
```
When the plugin downloads the tarball, it tries to move the path stored in `extracted`, which should be 'tree-sitter-koka-tmp/tree-sitter-koka-31fb0c4e76025da934075ed83bba7d51567e95c8' to output_dir.

the problem however is that `repo_project_name`, which is part of the path in `extracted` is nil, since the regex found no match. you can see it in [this regex101 fiddle](https://regex101.com/r/XbhpAF/1). `extracted` now becomes 'tree-sitter-koka-tmp/-31fb0c4e76025da934075ed83bba7d51567e95c8'.

i modified the regex to match the last url segment of my github URL, as shown [here](https://regex101.com/r/9wPJCo/1)

clason:
So the issue is the dash in the branch name? Can you explain how the change works to prevent this problem (without creating new ones)? (Ideally in the commit message, which should use conventional commits (`fix(install): ...`).

clason:
Also, note that Lua does _not_ use regular expressions but (much simpler) Lua patterns.

bramadityaw:
I am sorry, it seems that this error occured when I switched to neovim nightly a few weeks ago, which of course would be buggy. I have since switched to v0.11.4, and this plugin works as intended. Feel free to close this pull request.

clason:
(I am using nightly as well and have not observed any such issues, fwiw.)

PR #8166, opened 2025-10-02
feat(inko): update parser and highlights

yorickpeterse:
This commit includes syntax support for a few new syntax elements, and updates the highlights queries to highlight two new expression keywords.


runtime/queries/inko/highlights.scm
```diff
@@ -51,6 +51,8 @@
 ; Keywords
 [
   "as"
+  "async"
+  "await"
```
ribru17: Would you mind capturing these as `@keyword.coroutine` instead?

PR #8176, opened 2025-10-07
feat(inko): update parser for let pattern matching

yorickpeterse:
Commit [9d7ed4](https://github.com/inko-lang/tree-sitter-inko/commit/9d7ed4f6c0ea2a8f846f3bb00e33ab21ec9ca379) of the Inko tree-sitter grammar introduces support for pattern matching in `let` expressions. This requires some corresponding changes to the "local" queries to correctly define local variables. This is done by simply defining locals for all "identifier_pattern" nodes, instead of only doing this for "define_variable" nodes.

### DIFF
--- a/runtime/queries/inko/locals.scm
+++ b/runtime/queries/inko/locals.scm
@@ -13,7 +13,7 @@
 (argument
   name: _ @local.definition.parameter)
 
-(define_variable
+(identifier_pattern
   name: _ @local.definition.var)
 
 (define_constant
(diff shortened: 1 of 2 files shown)


runtime/queries/inko/locals.scm
```diff
@@ -13,7 +13,7 @@
 (argument
   name: _ @local.definition.parameter)
 
-(define_variable
+(identifier_pattern
   name: _ @local.definition.var)
```
yorickpeterse, line 17: Just to further clarify: Inko has `let` and `match` that can introduce local definitions through these `identifier_pattern` nodes. Using this approach means both cases are covered, regardless of how complex the patterns are (e.g. something like `(define_variable pattern: (tuple_pattern (identifier_pattern) ...))`).

PR #8179, opened 2025-10-10
feat(tmux)!: update parser and highlights

stevenxxiu:


clason:
`make docs`

clason:
Oh, and breaking changes should be `feat(tmux)!` (ideally with a short description of which nodes were removed/renamed) so other plugins that rely on the parser get a heads-up.

PR #8184, opened 2025-10-11
feat(cpp): keyword highlights for `export`, `module`

ribru17:


ribru17:
So that's why these weren't already here...

clason:
Stupid C family 😠

clason:
Not sure there is much that can be done with this PR...

PR #8186, opened 2025-10-11
feat(vue): delimiter highlight touch-ups

ribru17:


runtime/queries/vue/highlights.scm
```diff
@@ -5,7 +5,15 @@
   "]"
 ] @punctuation.bracket
 
-(interpolation) @punctuation.special
+[
+  ":"
```
lucario387, line 9: They aren't exactly delimiters though, `:` (shortform of `v-bind:`) are for props propagation in Vue components, while `.` is either a different form of `:` or shorthand for `.props`

https://vuejs.org/api/built-in-directives.html#v-bind

ribru17: Ah, I see. Does punctuation.special make more sense for them? Or character.special?

lucario387: I'd say `@character.special` make more sense.

PR #8191, opened 2025-10-13
Fix get_capture_matches_recursively function

error1707:
Hi! 

I think I found a bug.
I was using the `nvim-treesitter-textobjects` plugin and tried to set up the `lsp_interop` module, but no matter how I configured it, peeking at a definition would only show the first line.
After hours of print debugging, I think I finally managed to find a one-line solution to this issue.

I should mention that I've only been using Neovim for about two weeks and have never contributed to a plugin before. This is just a solution I came up with, and I'm not sure if it's the correct one.


clason:
Thank you, but these functions have all been removed on the maintained `main` branch.

PR #8192, opened 2025-10-14
feat(starlark): build targets as locals

lf-:
Problem: I have very big BUILD files I want to navigate quickly. I have Telescope, so I can use `:Telescope treesitter` for an outline, but build targets aren't emitted as definitions.

Solution: Add a query that catches build targets.

One thing I'm unclear on is:
- I want to filter to just build targets.
- Build targets are absolutely not any of the mentioned types in the CONTRIBUTING file.
- Build targets do not actually create identifiers in buck2/bazel: they create objects which are observable to the outside world and which the user generally *cares* about, but they aren't in the same namespace as normal identifiers.

  This means that they *especially* don't fit into any of the existing categories.

I would like guidance as to whether this is the right approach or if I should call the capture group something else.

### DIFF
--- a/.tsqueryrc.json
+++ b/.tsqueryrc.json
@@ -156,6 +156,7 @@
       "local.definition.namespace": "modules or namespaces",
       "local.definition.import": "imported names",
       "local.definition.associated": "the associated type of a variable",
+      "local.definition.target": "build targets",
       "local.scope": "scope block",
       "local.reference": "identifier reference"
     }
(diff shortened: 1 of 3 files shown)


clason:
I don't think we want to extend the upstream categories here (`locals` are pretty much deprecated).

You can always put such a custom query into your own `; extend` query file.

lf-:
hm. this is a tough position then: because the built in ontology is bad, it follows that users get a bad experience by default. i can put it in a downstream plugin, i just think it is an unfortunate alternative.

i might just eventually do it in the lsp instead.

clason:
yes, and if the fault is with the builtin ontology, it has to be fixed whereit is built in—and that is upstream tree-sitter.

The core goal of this project is _consistency_; anything that is specific to a single language belongs to specific plugin.

clason:
We can revisit this when and if upstream adds this to the documented `locals` captures.

PR #8193, opened 2025-10-14
Add support for babel/text <script> syntax highlighting in HTML

markvincze:
The `text/babel` script type is used with `@babel/standalone`: https://babeljs.io/docs/babel-standalone
This was not recognized by treesitter to use the JavaScript syntax highlighter, so the code looked like this:

image

Adding `"text/babel"` to `html_script_type_languages` solved the issue.


markvincze:
Declined as I realized I did this change on the deprecated `master` branch.
I'll try to see if the same issue happens on `main` as well and fix it there.

PR #8194, opened 2025-10-14
use gitlab api properly

pedrohgmacedo:


clason:
No, sorry. I don't want the codeforge leaking. As I mentioned on the issue you commented on, the proper fix is on your side. (There's only four parsers (only) hosted on gitlab  -- most of which have not been updated in years -- out of the > 300 parser we support, so this is definitely not worth the extra complexity.)

pedrohgmacedo:
> I don't want the codeforge leaking.

[It already leaks tho.](https://github.com/nvim-treesitter/nvim-treesitter/blob/0606c7a9dcaa5c5beee0b0f09043e9fdd1ba0a68/lua/nvim-treesitter/install.lua#L208)

clason:
Yes. In one place. I draw the line at two (and am in fact looking at dropping gitlab altogether).

pedrohgmacedo:
dropping gitlab and using one of the sources on github would solve the problem too, tbh.

clason:
If there's a github mirror, feel free to make a PR to switch. (It should be maintained, though, and not be stuck at ABI 13 which is not long for this world).

pedrohgmacedo:
I saw that someone else commented on the bug, confirming my hypothesis. It was then locked with the justification that it will be removed.  

The reason why this PR leaks in two places is actually because the original code assumes that the tarball has a top level subdirectory name that happens to be different than the gitlab api. 

However, [tar has a parameter, `strip-components`](https://man7.org/linux/man-pages/man1/tar.1.html), that could be used to work around that, and allow 'leaking' only in one place. However, the tarball extraction logic would need updating. 

Would you be interested in a PR using that? 
It would satisfy the requirement of leaking only in one place and would simplify a bit the extraction logic.

clason:
No, sorry. This is not the only issue with gitlab (nor that parser, which _will_ be removed). If the logic can be simplified _unconditionally_, I'd be interested (but be aware that not all platforms have the same version of `tar`, so this could be unavailable on, say, Windows, which would make this a no-go).

pedrohgmacedo:
> the logic can be simplified _unconditionally_

the logic can certainly be be improved unconditionally if it uses the features of the tools it uses smartly. 

> not all platforms have the same version of tar, so this could be unavailable on, say, Windows

windows (mac as well) uses bsdtar, which also supports this option [since at least 2004](https://man.freebsd.org/cgi/man.cgi?query=bsdtar&apropos=0&sektion=0&manpath=FreeBSD+6.0-RELEASE&arch=default&format=html#:~:text=also%20specified.%0A%0A%20%20%20%20%20%20%20%2D%2D-,strip%2Dcomponents).

> nor that parser, which will be removed

i'm only interested in fixing my problem tbh.

PR #8196, opened 2025-10-15
Switch tree-sitter-t32 repository to GitHub

xasc:
Switches the grammar repository from GitLab to GitHub.

On a related note, I have a question about this part from *CONTRIBUTING.md*:

> - provide WASM release artifacts;

The WASM modules for *tree-sitter-t32* are currently in the [GitLab package registry](https://gitlab.com/xasc/tree-sitter-t32/-/releases). Do they need to move, too?


clason:
The wasm parser is... aspirational at this point. Upstream hasn't got the story straight yet, so don't worry about it for now.

PR #8197, opened 2025-10-15
fix(python): regex injection not working for concatenated strings

AbysmalBiscuit:
Hello,

Currently the Python regex injection query doesn't handle concatenated strings.
This PR fixes this by adding a new query to inject regex into concatenated strings.

Thank you for taking the time to review my PR. 😊

## Context

Splitting a regex pattern into multiple strings in Python can be beneficial for legibility, especially when you have to use very long regex expressions.

Example code:

```python
import re

re_test: re.Pattern[str] = re.compile(r"^(?P<year>\d{4}) (?P<day>\d) \w\s{,3}")
re_test: re.Pattern[str] = re.compile(
    r"^(?P<year>\d{4}) "
    r"(?P<day>\d) \w\s{,3}"
)
```

Both instances create equivalent `re.Pattern` objects.

## Before

Here is how it gets highlighted currently:
image

The concatenated string in the second `re.compile` is not detected as regex.

<details><summary>Screenshot with Inspect tree view (without new query)</summary>
<p>

image

</p>
</details> 

## After

Including the new additional query will take into account concatenated strings, so they get detected as regex:
image

The concatenated string in the second `re.compile` is now detected as regex.

<details><summary>Screenshot with Inspect tree view (with new query)</summary>
<p>

image

</p>
</details>

### DIFF
--- a/runtime/queries/python/highlights.scm
+++ b/runtime/queries/python/highlights.scm
@@ -426,11 +426,8 @@
     object: (identifier) @_re)
   arguments: (argument_list
     (concatenated_string
-      [
-        (string
-          (string_content) @string.regexp)
-        (comment)
-      ]))
+      (string
+        (string_content) @string.regexp)))
   (#eq? @_re "re"))
 
 ; Decorators


clason:
Should these be `injection.combine`d?

ribru17:
A scoped/quantified injection would be more performant, but only supported on 0.11.3+

ribru17:
(We should also give the associated `@string.regexp` highlight group)

clason:
> A scoped/quantified injection would be more performant, but only supported on 0.11.3+

That's fine; the `main` branch is 0.11.4 only.

AbysmalBiscuit:
Hey @ribru17 and @clason, I updated the injection to use `@injection.combined` and updated the highlights query to set `@string.regexp` for concatenated strings.

I added the small example I gave as a test file.

ribru17:
Thanks! I think we should prefer scoped injections here, though. Something like the following:

```tsq
(call
  function: (attribute
    object: (identifier) @_re)
  arguments: (argument_list
    .
    (concatenated_string
      (string
        (string_content) @injection.content)+))
  (#eq? @_re "re")
  (#set! injection.language "regex"))
```

This one won't account for comments between strings, but I don't think that would be a common case. If you disagree we can maybe go with a more complex pattern to account for this. Like this:

```tsq
(call
  function: (attribute
    object: (identifier) @_re)
  arguments: (argument_list
    .
    (concatenated_string
      [
        (string
          (string_content) @injection.content)
        (comment)
      ]+))
  (#eq? @_re "re")
  (#set! injection.language "regex"))
```

AbysmalBiscuit:
@ribru17 thanks for the suggestions!

I think comments around the strings are indeed uncommon, but I can imagine them being useful.

I updated the injection to be scoped and to handle comments around the strings.
I also updated the highlights for `@string.regex` to be scoped as well.

ribru17:
Any time! And sorry to do this to you, but would you mind splitting up the regular string and concatenated string patterns into two unique patterns? The singular pattern looks a bit convoluted. Also, imo, we would remove the preceding `(comment)*` and the start anchor in the patterns

Also, would you mind making the highlight pattern split, and also actually remove the quantifier? Those ones don't need to be quantified/scoped and it usually makes query iteration slightly slower

AbysmalBiscuit:
Hey, I updated the patterns to split them for both the injections and highlights.
I also removed the quantifier from the highlights query.

Sorry for the delay on this one.

clason:
Looks like this is blocked by the `starlark` parser, which inherits the python injections.

AbysmalBiscuit:
Hey, I had a look and to the best of my understanding it doesn't seem like `starlark` has a `re` module.

I updated the starlark injections to not inherit from python and added test files to make sure it works correctly.

I found [this third party project](https://github.com/magnetde/starlark-re) that adds it. If you think it's necessary I can add the non-concatenated regex injection query to the starlark injections.

runtime/queries/python/highlights.scm
```diff
@@ -417,11 +417,22 @@
...
+(call
+  function: (attribute
+    object: (identifier) @_re)
+  arguments: (argument_list
+    (concatenated_string
+      [
+        (string
+          (string_content) @string.regexp)
+        (comment)
+      ]))
```
ribru17:
```suggestion
      (string
        (string_content) @string.regexp)))
```

ribru17 (approved):
LGTM! One final nit:

PR #8199, opened 2025-10-15
feat(parsers)!: updated wit queries and grammar to v1.2.0

mkatychev:
Updated to match new GH release:
https://github.com/bytecodealliance/tree-sitter-wit/releases/tag/v1.2.0

### DIFF
--- a/runtime/queries/wit/highlights.scm
+++ b/runtime/queries/wit/highlights.scm
@@ -6,6 +6,32 @@
 
 (version) @string.special
 
+(use_path
+  [
+    "@"
+    "/"
+  ] @punctuation.delimiter)
+
+(package_decl
+  [
+    "@"
+    "/"
+  ] @punctuation.delimiter)
+
+; feature gates with leading `@`
+(_
+  .
+  "@" @punctuation.special
+  .
+  [
+    "since"
+    "unstable"
+    "deprecated"
+  ] @attribute.builtin)
+
+(unstable_gate
+  feature: (id) @string)
+
 (world_item
   name: (id) @module)
 
@@ -53,8 +79,7 @@
 (flags_items
   name: (id) @type)
 
-(body
-  (id) @variable.member)
+(flags_field) @variable.member
 
 (variant_items
   name: (id) @type)
@@ -65,8 +90,7 @@
 (enum_items
   name: (id) @type)
 
-(enum_case
-  name: (id) @constant)
+(enum_case) @constant
 
 (resource_item
   name: (id) @type)
@@ -89,6 +113,9 @@
 (use_path
   (id) @module)
 
+(alias_item
+  (id) @module)
+
 (use_names_item
   (id) @module)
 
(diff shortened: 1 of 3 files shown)


clason:
@ribru17 could you take a quick look at the new captures?

runtime/queries/wit/highlights.scm
```diff
@@ -129,27 +161,31 @@
...
+] @type
 
-[
-  "@"
-  "_"
-] @punctuation.special
+"_" @variable.parameter.builtin
 
 [
-  "/"
```
ribru17, line 145: Looks like all the new `"/"` highlights are also `@punctuation.delimiter`- why not remove them and keep this here?

runtime/queries/wit/highlights.scm
```diff
@@ -129,27 +161,31 @@
...
+[
   "tuple"
   "list"
   "option"
   "result"
   "borrow"
-] @type.builtin
+  "future"
+  "stream"
+] @type
```
ribru17, line 174: These are also builtin types, no? Builtin types need not be primitive types necessarily

runtime/queries/wit/highlights.scm
```diff
@@ -89,6 +113,9 @@
 (use_path
   (id) @module)
 
+(alias_item
+  alias: (id) @module)
```
ribru17: Why not `path: (id)` as well? Then we can just make this

```tsq
(alias_item
  (id) @module)
```

runtime/queries/wit/highlights.scm
mkatychev: I was looking more at the way rust handles it and wanted to distinguish primitive types from those that can wrap other types:
https://github.com/tree-sitter/tree-sitter-rust/blob/master/queries/highlights.scm#L4

All types that have the `@type` decorator here can wrap a `T`

Image

runtime/queries/wit/highlights.scm
mkatychev: good point

runtime/queries/wit/highlights.scm
mkatychev: I recall the problem being that doc comments would pick up on the extra slash as a delimiter:

Image

Due to the logic here:
https://github.com/bytecodealliance/tree-sitter-wit/blob/a80c1f47baa5bfb13e2b5f49aa5304e3dab94948/grammar.js#L321-L321

EDIT: if I recall this visual bug presents itself with the current nvim-treesitter WIT config, maybe there's a cleaner way to handle this in the grammar?

clason:
@mkatychev sorry for dragging this out, but I hope we can get this in shortly. Mind rebasing on latest `main`?

@ribru17 could please you take a final look?

mkatychev:
No worries, there's pending changes but some of it merits a brief discussion.

clason:
This has been blocking the queue long enough; any outstanding fixes can be made in a follow-up PR.

ribru17:
Sorry for forgetting about this; LGTM :+1:

PR #8204, opened 2025-10-17
Add automatic installation section to README

HampusHauffman:
Added instructions for automatic installation of nvim-treesitter with Lua configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,56 @@ These queries can be used to look up definitions and references to identifiers i
 
 # Advanced setup
 
+## Automatic installation
+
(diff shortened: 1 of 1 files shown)


clason:
Thank you, but no; this feature was intentionally removed because it is not a use case I want to support.

clason:
(At some point, we'll reopen the wiki, and then such snippets can be added there.)

HampusHauffman:
> Thank you, but no; this feature was intentionally removed because it is not a use case I want to support.

No sweat! Thanks for the quick response. I'll be sure to add it to the wiki if/when that gets added!

PR #8205, opened 2025-10-17
feat(perl): Add language injection based on heredoc delimiter

mrsdizzie:
Similar to other languages.

Before:
Screenshot 2025-10-17 at 10 10 28

After:
Screenshot 2025-10-17 at 10 10 53


clason:
@rabbiveesh

clason:
Make sure the nesting is correct if you add directives (compare, as you write, other languages).

rabbiveesh:
Looks good to me

PR #8210, opened 2025-10-19
feat: add token for match import special characters

nickshiro:
## Problem
In import statements, keywords like import are tokenized as `@keyword.import`. However, symbols such as * (e.g., in `import * as all from 'package'` in JavaScript) are not assigned a dedicated token. This makes it impossible to consistently distinguish these symbols from other characters like _, which are matched as `@character.special` and prevents proper highlighting in themes.

## Solution
Introduce a new token `@character.import` specifically for symbols used in import constructs.

### DIFF
--- a/.tsqueryrc.json
+++ b/.tsqueryrc.json
@@ -37,6 +37,7 @@
 
       "character": "character literals",
       "character.special": "special characters (e.g. wildcards)",
+      "character.import": "special characters in import statements",
 
       "boolean": "boolean literals",
       "number": "numeric literals",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -149,6 +149,7 @@ The valid captures are listed below.
 
 @character              ; character literals
 @character.special      ; special characters (e.g. wildcards)
+@character.import       ; special characters in import statements
 
 @boolean                ; boolean literals
 @number                 ; numeric literals
--- a/runtime/queries/ecma/highlights.scm
+++ b/runtime/queries/ecma/highlights.scm
@@ -308,15 +308,15 @@
 ; Imports
 ;----------
 (namespace_import
-  "*" @character.special
+  "*" @character.import
   (identifier) @module)
 
 (namespace_export
-  "*" @character.special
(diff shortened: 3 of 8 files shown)


ribru17:
Adding a new capture group requires a pretty strong justification- I don't personally agree with this (yet?). What about `@character.special` is insufficient? Can you give some examples of code and how a separate capture group makes highlighting better? I don't really see the benefit. If you really think this capture should be changed, would `@module` work over `@character.special`?

---

If anything, I think `@character.special.import` would be a better fit for a new group

nickshiro:
Thank you for your response. Let me clarify the reasoning behind this proposal.

Using the JavaScript (ECMA) grammar as an example, consider the following construct:
`import * as namespace from 'module'`

In this case, the `*` symbol is currently tokenized as `@character.special`, while the keywords `import`, `as`, and `from` are correctly tokenized as `@keyword.import`.

The distinction for keywords allows themes to highlight import and export statements consistently and distinctly from regular code — which is both correct and convenient. However, the asterisk (`*`), which is an integral part of the import statement, remains grouped under the general `@character.special token`. This means it is treated the same way as symbols used in unrelated contexts, such as regular expressions.

This results in a small but noticeable inconsistency: keywords in the import statement have a dedicated import-related scope, while the symbol that is also part of that construct does not.

Introducing a new token such as `@character.import` (or alternatively `@character.special.import`) would allow for more consistent and precise highlighting of import-related syntax across languages.

clason:
Sorry, that reasoning is too weak. We value consistency across languages (necessary for colorschemes to be designed without the explosion of language-specific groups that plague the legacy syntax). I also remind you that the highlight groups are defined by Neovim itself, not here (we just manage the queries), so this is not a simple change we can make.

You do know that you can customize the queries to your heart's content, right?

nickshiro:
Yes, I know how to write a queries. I just wanted to set a standard. Is it not possible even with `@character.special.import?`

clason:
No, sorry. As I wrote, the highlight groups are to be considered set in stone for now.

PR #8211, opened 2025-10-20
fix(nu): inefficient and incorrect highlight query of `collection_type`

blindFS:
Previously causing issues with verbose collection types.

https://github.com/nushell/tree-sitter-nu/pull/230

### DIFF
--- a/runtime/queries/nu/highlights.scm
+++ b/runtime/queries/nu/highlights.scm
@@ -344,7 +344,7 @@ key: (identifier) @property
   [
     "record"
     "table"
-  ] @type.enum)
+  ] @type.builtin)
 
 (collection_type
   key: (_) @variable.parameter)


runtime/queries/nu/highlights.scm
```diff
@@ -344,14 +344,19 @@ key: (identifier) @property
...
-  "<" @punctuation.bracket
-  key: (_) @variable.parameter
+  ] @type.builtin)
+
+(collection_type
+  key: (_) @variable.parameter)
+
+(collection_type
   [
-    ","
```
clason, line 351: Where did this go?

blindFS: It's captured in another layer of anonymous node named `_entry_separator`, a change made 11 months ago in https://github.com/nushell/tree-sitter-nu/pull/166 I believe.

So basically it's an impossible pattern

image

Maybe there's a bug in the linter for not detecting this without this PR.

clason: I don't see that error on your PR, either. (Maybe you're on an older version of the LSP?)

@ribru17?

blindFS: > I don't see that error on your PR, either. (Maybe you're on an older version of the LSP?)

You mean after adding back "," ? I suppose an error is expected.

clason: I mean just adding that pattern in your screenshot. If it's not an error now, I don't see how it could be an error on your PR.

clason: (This is not about your changes; those look good to me and I will merge them later.)

ribru17: That's really weird. I'll look into this soon

ribru17: I'm unable to trigger the "impossible pattern" diagnostic for some reason, no clue why. `tree-sitter query` shows that it is an invalid structure but neither the LSP nor neovim linter shows this, for me

PR #8216, opened 2025-10-21
Add zsh support

georgeharker:
Added support for a new zsh parser which better recognizes zsh specific shell features.


clason:
Sorry, but I don't want any CringePilot on this project. If you can open a PR without triggering it, go ahead.

clason:
Also, this change is worthless without queries.

georgeharker:
queries are in the queries directory, can't believe it's so hard to disable copilot.  Will try with it off entirely for reviews.

clason:
Yeah, sorry about that. (Not really surprised though.)

But we do _not_ use upstream queries, as a point of principle, so that's not acceptable for a tracked parser. (I just added this for custom parsers.)

PR #8217, opened 2025-10-21
Add Zsh support

georgeharker:
Added support for a new zsh parser which better recognizes zsh specific shell features

attempting to create this PR without copilot automated reviews


clason:
Nope.

clason:
(And, again, no queries, no point.)

clason:
The queries need to be committed to _this repository_ since they need to conform to specifications.

PR #8218, opened 2025-10-21
Add Zsh support

georgeharker:
Added support for a new zsh parser which better recognizes zsh specific shell features

attempting to create this PR without copilot automated reviews - disabled copilot reviews for my entire account.

### DIFF
--- a/lua/nvim-treesitter/parsers.lua
+++ b/lua/nvim-treesitter/parsers.lua
@@ -2708,6 +2708,7 @@ return {
       location = 'tree-sitter-zsh',
       revision = '9340bfd8c8d556aea1d8feb41b32600ba9f3cc33',
       url = 'https://github.com/georgeharker/tree-sitter-zsh',
+      queries = 'queries'
     },
     maintainers = { '@georgeharker' },
     tier = 2,


clason:
Again, you need to add queries _to the PR_. This is non-negotiable.

clason:
And, sorry, we no longer accept WIP parsers. Please let this cook for a bit until it's stable before adding it here.

lua/nvim-treesitter/parsers.lua
```diff
@@ -2703,4 +2703,14 @@ return {
     maintainers = { '@rockorager' },
     tier = 2,
   },
+  zsh_schema = {
+    install_info = {
+      location = 'tree-sitter-zsh',
```
clason, line 2708: This is wrong, for the record.

georgeharker:
queries are in the queries directory and support similar to bash grammar.  Will look to improve coverage and specificity, but they match what tree-sitter-bash does right now.

georgeharker:
ok, will let it cook a while - was hoping to get additional eyes on it to help ensure stability

clason:
Sorry, maintenance load is too high for that. (And, again, the point of this project is curated and tested queries, so we will **never** simply install upstream queries. No CI, no dice.)

clason:
(If you want more eyes, post on Matrix or Discord asking for beta testers.)

PR #8225, opened 2025-10-24
feat(kitty): Update parser

OXY2DEV:
Parser revision changed for the following bug fixes,

- Case-insensitive key names(e.g. `CTRL`, `Ctrl`, `ctrl`).
- Made `=` optional for `--type` flag in `launch` action.

Ref: OXY2DEV/tree-sitter-kitty#6


clason:
Simple parser bumps are automated (and throttled); individual PRs are only necessary and welcome if queries need to be adapted to match.

PR #8226, opened 2025-10-25
fix: use proper `; inherits` syntax in queries

ribru17:
Some queries don't add a colon after the `inherits` keyword, which nvim could handle but `ts_query_ls` could not, causing it to give incomplete diagnostics.

### DIFF
--- a/runtime/queries/html/injections.scm
+++ b/runtime/queries/html/injections.scm
@@ -1,4 +1,5 @@
-; inherits html_tags
+; inherits: html_tags
+
 (element
   (start_tag
     (tag_name) @_py_script)
--- a/runtime/queries/starlark/injections.scm
+++ b/runtime/queries/starlark/injections.scm
@@ -1,1 +1,1 @@
-; inherits python
+; inherits: python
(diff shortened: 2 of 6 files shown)


clason:
Can we add this to the lint/format in `ts_query_ls` so we don't regress?

ribru17:
Sure, I'll give a lint for specifically the no-colon case. Hopefully that is enough without overdoing it

ribru17:
Maybe we should take this as an opportunity to decide on how much parity we want with nvim? Nvim supports parentheses around import names, indicating they should not be recursively imported. The LSP does not support this. I believe nvim also trims spaces from import names (which the lsp does not, but it would be a quick fix). 

Neovim doesn't support numbers in import names (which it should, because we have several parsers with numbers in their names, like `c3`)

If the LSP is meant to match neovim exactly, should they meet in the middle? I don't want to tailor it too closely to one editor (the best one), but it doesn't really matter right now because nobody else has really taken an interest :pensive:

clason:
I don't think much thought has gone into the format in Neovim. If there's any (reasonable) outside guidance, we'd be happy to follow suit. (Choice is the refuge of people who can't make up their mind.)

clason:
(And not supporting numbers is clearly a bug.)

PR #8222, opened 2025-10-23
queries(apex): Use Javadoc for comment injections

imawaki:
Apex codebases commonly use Javadoc-style comments, similar to Java (Apex is Salesforce's object-oriented language).

This updates the injection query to capture `javadoc` nodes instead of the generic `comment` for better highlighting and parsing accuracy.

PR #8232, opened 2025-10-26
queries(apex): Use Javadoc for comment injections

imawaki:
Apex codebases commonly use Javadoc-style comments, similar to Java (Apex is Salesforce's object-oriented language).

This updates the injection query to capture javadoc nodes instead of the generic comment for better highlighting and parsing accuracy.

fixes; https://github.com/nvim-treesitter/nvim-treesitter/pull/8222


imawaki-jgcd:
Thank you!

PR #8235, opened 2025-10-27
feat(julia)!: update parser and queries

ChrHorn:
ref: https://github.com/tree-sitter-grammars/tree-sitter-julia/pull/6

@clason 
You have to check on your end. I don't really use neovim.


clason:
Thanks, but simply copying the queries is not enough since we have slightly different capture conventions. I'll just take it from here.

PR #8238, opened 2025-10-28
Diff injections

folke:
This PR adds injections for the changes in a diff

## Before
image

## After
image


clason:
Besides the wrong branch, I feel the usefulness (or even harmfulness) of this depends _a lot_ on the specific colorscheme? I could imagine this making diffs completely unreadable (e.g., with the default Nvim colorscheme).

PR #8239, opened 2025-10-28
feat(diff): add language injection for changes

folke:
This PR adds injections for the changes in a diff

## Before
image

## After
image

### DIFF
--- a/runtime/queries/diff/injections.scm
+++ b/runtime/queries/diff/injections.scm
@@ -1,2 +1,10 @@
 ((comment) @injection.content
   (#set! injection.language "comment"))
+
+(block
+  (new_file
+    (filename) @injection.filename)
+  (hunks
+    (hunk
+      changes: (changes) @injection.content
+      (#set! injection.include-children))))


clason:
Copying my comment from the previous attempt:

I feel the usefulness (or even harmfulness) of this depends a lot on the specific colorscheme? I could imagine this making diffs completely unreadable (e.g., with the default Nvim colorscheme).

folke:
Isn't that with everything though?
Every changes potentially needs colorschemes to update as well.

Delta is a widely used git diff pager, that does pretty much what this injection does.

However, if this is not something you're interested in, let me know and I'll close the PR

clason:
> Isn't that with everything though?
> Every changes potentially needs colorschemes to update as well.

Well, the intent of this project is to standardize the captures to such an extent that colorschemes only need to consider a finite set of documented (upstream!) highlight groups.

But this is very different: you're not changing a single color, but fundamentally how diffs are highlighted -- breaking all previous assumptions. (Delta is an unfair comparison because it comes with its own colorschemes _that are built on the assumption that diffs are syntax highlighted_.)

I really invite you to test this with `:colorscheme default` and see if this is actually an improvement.

So, yes, I think this is too intrusive to add as a default (but don't get me wrong, it's neat that you can do that; maybe we could make a wiki page for such customization ideas that we don't want to enable by default for one reason or another).

clason:
Or in other words: If you feel strongly enough about this, you need to push for this change upstream (essentially _mandating_ that `@diff.*` highlight groups use `bg` highlighting). There have been requests for bundling the `diff` parser (which is currently not in any shape for it, but I could try to fix that if there's sufficient demand).

folke:
no, that's ok. Just added these injections to `snacks.nvim`.

I'll assume that anyone using snacks is not using the default colorscheme, since that will not be great for some snacks things anyway. I'll see if I get any objections about this...

clason:
That sounds like a good approach, yes (and maybe document the expectation on the colorscheme?) If upstream decides they would like that, too, we can revisit this.

PR #8240, opened 2025-10-29
Add zsh support

georgeharker:
Adds a zsh grammar which parses zsh more completely than the bash grammar or old zsh grammar.

Adds nvim hilight, fold queries

Add tests for highlights

### DIFF
--- a/runtime/queries/zsh/highlights.scm
+++ b/runtime/queries/zsh/highlights.scm
@@ -213,16 +213,16 @@
 
 (array_at) @variable.builtin
 
-(expansion_flags)  @attribute.builtin
+(expansion_flags) @attribute.builtin
 
-(expansion_style)  @attribute.builtin
+(expansion_style) @attribute.builtin
 
 (expansion_pattern
-  pattern: ("#" @attribute.builtin))
+  pattern: "#" @attribute.builtin)
 
 (expansion_modifier) @attribute.builtin
 
-(simple_variable_name)  @variable
+(simple_variable_name) @variable
 
 (glob_pattern) @string.regexp
 


runtime/queries/zsh/highlights.scm
```diff
@@ -0,0 +1,277 @@
...
+(declaration_command
+  (word) @variable.parameter)
+
+(unset_command
+  (word) @variable.parameter)
+
+(number) @number
+
+((word) @number
+  (#lua-match? @number "^[0-9]+$"))
```
clason, line 193: Why is this necessary if your grammar supports `(number)`?

runtime/queries/zsh/highlights.scm
```diff
@@ -0,0 +1,277 @@
...
+((command
+  name: (command_name
+    (word) @_printf)
+  .
+  argument: (word) @_v
+  .
+  argument: (word) @variable)
+  (#eq? @_printf "printf")
+  (#eq? @_v "-v")
+  (#lua-match? @variable "^[a-zA-Z_][a-zA-Z0-9_]*$"))
```
clason, line 281: again, shouldn't that be handled on the level of the grammar?

runtime/queries/zsh/highlights.scm
```diff
@@ -0,0 +1,277 @@
...
+
+[
+  (regex)
+  (extglob_pattern)
+] @string.regexp
+
+((program
+  .
+  (comment) @keyword.directive @nospell)
+  (#lua-match? @keyword.directive "^#![ \t]*/"))
```
clason, line 294: same as above

tests/query/highlights/zsh/test.zsh
clason, line 1: No need to duplicate upstream tests.

runtime/queries/zsh/highlights.scm
georgeharker: in shell grammars, the tokenization is on word (only in arithmetic contexts will numbers be specifically recognized).  For highlighting purposes however we would like to show numbers in a specific highlight group.

See https://github.com/nvim-treesitter/nvim-treesitter/blob/645f42e85d8665c91a9911c3896afb57d6b8a923/runtime/queries/bash/highlights.scm on which this is based.

runtime/queries/zsh/highlights.scm
georgeharker: this is solving particular highlights for printf - the shell grammar doesn't care.  But for feature parity with bash, we try to highlight printf specifically.

runtime/queries/zsh/highlights.scm
georgeharker: This deals with shebang constructs at the top of the script (eg #!/bin/sh - we don't want these spell checked, whereas general comments are).

tests/query/highlights/zsh/test.zsh
georgeharker: Would you prefer they be removed entirely?  There are no upstream tests for highlight groups - just for the grammar itself.

georgeharker:
Had some positive feedback from non integrated launch of this, with the major feedback being please get this into nvim-treesitter - hence the PR.

clason:
Let me be blunt (to explain where I'm coming from, not to shut this down): By upstreaming this, you are effectively passing on all maintenance to us -- meaning, all issues will be reported here first and we need to triage them. You can imagine that this quickly adds up for >300 languages... In addition, every breaking change on your grammar will break our automation and require manual intervention to either fix or repeatedly postpone.

So I don't care how many people want to have a convenient one-click install; all I care is how much work this is going to create for me. Specifically, experience shows that new parsers often either go through a phase of heavy development with frequent breaking changes or are abandoned quickly in a half-finished state. So my question is where this grammar sits on the development spectrum: Is it stable, i.e., no further significant changes are planned, everything is nicely covered by regression tests, and all that will happen is the usual maintenance trickle of semver'ed releases, and how long do you think you'll be able to (quickly, within days) react to pings here?

clason:
Or to put this more constructively: If the parser is stable, consider marking it as tier 1 and track release versions, not every commit.

georgeharker:
I hear you.  It's definitely not my intention to add workload for you all.

The grammar is stable in that i've spent a chunk of time debugging it, it's based on the previously stable bash grammar, and whilst significant differences have been introduced against the bash codebase, I have endeavored to have each difference well tested in the upstream grammar repo.  Others have also downloaded and installed (250 clones according to GitHub metrics - though I'm sure there's plenty of interpretation to be done on GitHub metrics).  

There could - of course - be bugs.  The language itself is a fixed target (not an evolving thing), which I believe i've covered well. I've added extensive tests over and above the bash grammar tests.  Given the closeness of the queries to the bash grammar's queries despite the underlying implementation having significant differences, I would estimate that most issues would be of the form of ensuring parsing accepts / recognizes something rather than reworking the entire structure - certainly wrt. the highlighting / queries.

I am aiming to address bugs on the grammar and highlighting promptly.

Couple of questions:
- my understanding of tier may be incorrect - I understood tier 1 to be core and completely stable, tier 2 to be robust and well supported but not necessarily static.
- I understood the intent of version locks by commit hash to preclude changes from the grammar breaking nvim-treesitter releases.  I am curious if you support semver tags, but regardless any changes upstream should not break the nvim-treesitter releases - is that correct?
- what mechanisms are there for directing bug reports to the originating grammar repo?  I'm happy to remain connected here and address issues and PRs on the grammar repo.
- should the highlight tests remain part of nvim-treesitter, was trying to be complete.
- you asked if the parser is sufficiently stable to be tier 1 and track release versions.  I'm happy to do that - though looking at the examples for both it felt like tier 2 was correct.  I am certainly happy to switch to a tag release reference though I see none other did that.  The grammar is semver release tagged.

I most certainly would not be looking to push every commit to nvim-treesitter, but rather any major address of an issue found / large improvements / significant release

clason:
That all sounds good.

To your questions:

1. the default (on tier <= 2) is to indeed track _every commit_, although that is not done manually but roughly once a week (there's a script for bumping all parsers, and a CI workflow that is currently disabled because the `main` branch is not default yet).
2. of course, we run those changes through CI to see that nothing breaks -- if not, I'll merge the update, otherwise manual work is required.
3. Tier 1 means "stable" in the sense that users only want to see stable releases instead of tracking HEAD (since configs can be overridden) -- this is a new concept on `main` which I'm trying to push, so none of the legacy entries taken from `master` have been upgraded. For new parsers, I prefer starting with that -- if it fits.
4. The mechanism for directing bug reports is me closing the issue with "upstream problem; please report there". Of course, if you actively monitor this repo, you can get there first and be more helpful ;) But we don't have any automation for that (for >300 languages with more constantly being added, that would not scale well).

clason:
Oh, and the highlight test can stay. It'll be useful to guard against regressions.

georgeharker:
Great! I've updated the PR to do a named/tagged revision and be tier 1.

clason:
Needs a rebase (and a review by @ribru17), otherwise looks good to me.

georgeharker:
Thanks!

Any thoughts @ribru17 ?

runtime/queries/zsh/highlights.scm
ribru17, line 1: A few nits: it would be nice to have highlights for `"#"`, `(array_at)`, `(expansion_flags)`, `(simple_variable_name)`, and `(glob_pattern)`. Whatever you think is most appropriate for all of them

ribru17 (approved):
LGTM overall

runtime/queries/zsh/highlights.scm
georgeharker: Great - I've added these and a couple more (assuming hash in variable name, comments were already picked up)!

ribru17 (approved):

georgeharker:
Thanks @clason @ribru17 !

PR #8244, opened 2025-10-29
feat!(gdshader): replace gdshader grammer and queries

airblast-dev:
Replaces the current gdshader grammer with the one I wrote.

I wanted to get syntax highlighting for my gdshader files but noticed a bunch of nodes being ERROR's. Even float and integer literals weren't being parsed correctly. 
I originally intended to fix the current grammar but felt like the effort to fix, update, and then get it reviewed wasn't worth it as the changes required would be breaking anyways.
As a result I wrote my own grammar. The project has a bunch of tests compared to the current one and follows the GLSL ES 3.0 specification (which gdshader is based on). I also manually tested the grammar with a few shaders from https://godotshaders.com/.

Regardless of this getting merged or not; I intend to use this and thought that there might be some interest in replacing the current grammar.

Please let me know if I should open an issue first. I thought it would easier to create PR as I have already put the effort in writing a new grammar.
Link to grammar: https://github.com/airblast-dev/tree-sitter-gdshader

### DIFF
--- a/tests/query/highlights/gdshader/test.gdshader
+++ b/tests/query/highlights/gdshader/test.gdshader
@@ -1,0 +1,26 @@
+shader_type spatial;
+
+render_mode unshaded, wireframe;
+// <- @keyword
+//          ^^^^^^^^ @keyword
+//                    ^^^^^^^^^ @keyword
+
+/** My Doc comment*/
+//^^^^^^^^^^^^^^^^^^ @comment.documentation
+uniform float MY_VALUE: interesting_hint = 1.0;
+//<- @keyword 
+//      ^^^^^ type.builtin
+//            ^^^^^^^^ constant
+//                    ^ @punctuation.delimiter
+//                      ^^^^^^^^^^^^^^^^ @attribute
+//                                       ^ @operator
+//                                         ^^^ @number.float
+//                                            ^ punctuation.delimiter
+
+void vertex() {
+// <- type.builtin
+//   ^^^^^^ @function
+  do {} while (123< 2231);
+//^^ keyword.repeat
+//      ^^^^^ keyword.repeat
+}


clason:
Should be `feat!` (breaking changes never fix anything ;))

I'm not averse to changing, but there needs to be a significant demonstrated benefit (to prevent this from devolving into "my parser is better!" shouting or flip-flop PRs). Can you give a few examples where your parser gives a clear improvement? (I also accept "my parser parses this corpus without error, while the other one has X errors".)

runtime/queries/gdshader/highlights.scm
```diff
@@ -1,142 +1,298 @@
...
-(builtin_function) @function.builtin
+; technically an injection site for bbcode TODO: maybe add it?
+((comment) @comment.documentation @spell
+  .
+  [
+    (struct_definition)
+    (declaration)
+    (function_definition)
+  ]
+  (#match? @comment.documentation "/\\*\\*[\\s\\S]*?\\*/"))
```
clason: Please change to `#lua-match` (so we don't need to spin up the Vim regex engine).

Although it would be even better if that could be handled on the level of the grammar. 

Also, you can at least add this to the `comment` injection.

runtime/queries/gdshader/highlights.scm
```diff
@@ -1,142 +1,298 @@
...
+    "unpackSnorm2x16" "packUnorm4x8" "unpackUnorm4x8" "packSnorm4x8" "unpackSnorm4x8"
+    ; bitwise builtins
+    "bitfieldExtract" "bitfieldInsert" "bitfieldReverse" "bitCount" "findLSB" "findMSB" "imulExtend"
+    "umulExtend" "uaddCarry" "usubBorrow" "ldexp" "frexp"))
 
-(member_expr
-  member: (ident) @property)
+; assume all uppercase variables as constants
+((identifier) @constant
+  (#match? @constant "^[A-Z][A-Z_0-9]*$"))
```
clason: same: `#lua-match?`

runtime/queries/gdshader/highlights.scm
```diff
@@ -1,142 +1,298 @@
...
+  (#any-of? @constant.builtin
+    ; common / declared in all shader types
+    "PI" "E" "TAU" "CURRENT_RENDERER" "RENDERER_COMPATIBILITY" "RENDERER_MOBILE"
+    "RENDERER_FORWARD_PLUS"; TODO: add more builtins
+    ))
 
-(call_expr
-  function: (builtin_type) @function.call)
+((identifier) @variable.builtin
+  (#any-of? @variable.builtin "TIME"))
```
clason:
```suggestion
  (#eq? @variable.builtin "TIME"))
```

airblast-dev:
Included both the tree and highlights in the comparison.

New Grammar
image

Old Grammar
image

These are just a few that I remember not working in the current grammar.

There is also one more thing that bothers me that might be subjective: There's a bunch of `maybe_*` `invalid_*` nodes which aren't needed at all and just cause confusion.

Will check the suggestions in a bit.

Edit:
In case someone wants to try the grammar out with the same snippet as above.
```gdshader
shader_type spatial;

render_mode unshaded;

uniform uint counter = 1U;
uniform float angle = 1.0E-3;
const int my_array[32] = int[](my_args);

void func() {
  do {
    do_thing();
  } while(1 < 5);
}
```

runtime/queries/gdshader/highlights.scm
ribru17: Maybe we should mandate any-of? take 2+ args? So we get a lint for this and can direct people to use eq

runtime/queries/gdshader/highlights.scm
airblast-dev: > Also, no injections?

The doc string of gdshader uses bbcode and to my knowledge there isn't any tree-sitter grammar for it.
(I also have no idea how bbcode works so I didn't try too hard tbh)

> Although it would be even better if that could be handled on the level of the grammar.

From a syntax standpoint its just a multiline comment with an extra "*" at its start. Its even considered a normal comment if there isn't a definition or declaration under it so I thought adding a capture group is correct. I am fairly new to writing grammars so let me know if I should still change it.

> Please change to #lua-match (so we don't need to spin up the Vim regex engine).

Will do. I noticed that the queries here used `#lua-match` but never understood why.

clason: > The doc string of gdshader uses bbcode and to my knowledge there isn't any tree-sitter grammar for it.

Yeah, I got that -- my question was whether the `comment` parser was not strictly better than nothing?

> Its even considered a normal comment if there isn't a definition or declaration under it so I thought adding a capture group is correct.

Do you have an example code snippet for that?

> Will do. I noticed that the queries here used #lua-match but never understood why.

Yeah, the naming is not great. `#match` should really be `#vim-match` in queries, but we added the `#match` alias for compatibility (at significant performance cost) with upstream queries.

clason: Note that Lua patterns are not regular expressions, so not every regex can be translated into a Lua pattern (hence no lint). The vast majority can, though.

airblast-dev: > ...my question was whether the comment parser was not strictly better than nothing?

I mean it is kind of intended to be better than nothing as there is no language we can inject here. I am still open to ideas to improve it if there is a feasible way to do it.

> Do you have an example code snippet for that?

I am not sure how I can provide a snippet. What I meant is the Godot editor highlights it as a doc comment but ignores the fact that it doesn't document anything. As a result its just a comment with an extra star. 

Though it probably makes sense to group it as a comment.documentation regardless of a declaration being under it.

> Also, you can at least add this to the comment injection.

(from your first suggestion)
Isn't that already done because of the `(comment)` node in `injections.csm`?

clason: Isn't that already done because of the (comment) node in injections.csm?

Ah, you are right, I missed that. Sorry!

airblast-dev: Replaced this and others with `#lua-match?`.
Let me know if there is anything else I should do here.

runtime/queries/gdshader/highlights.scm
```diff
@@ -1,142 +1,298 @@
...
-] @keyword.conditional
+"#define" @keyword.directive.define
 
 [
-  "#"
-  "include"
-] @keyword.directive
+  ";"
+  "."
+] @punctuation.delimiter
```
ribru17, line 56: Also `":"`, `","`

runtime/queries/gdshader/highlights.scm
```diff
@@ -1,142 +1,298 @@
...
+    (preproc)
+  ]*
+  .
+  (shader_type_statement
+    (shader_type) @_shader_type)
+  (#eq? @_shader_type "sky")
+  (_)*
+  (render_mode_statement
+    (render_mode) @keyword)
+  (#any-of? @keyword "use_half_res_pass" "use_quarter_res_pass" "disable_fog"))
```
ribru17, line 173:
```suggestion
(source_file
  (shader_type_statement
    (shader_type) @_shader_type)
  (#eq? @_shader_type "sky")
  (render_mode_statement
    (render_mode) @keyword)
  (#any-of? @keyword "use_half_res_pass" "use_quarter_res_pass" "disable_fog"))
```

we don't need to match the whole structure

runtime/queries/gdshader/highlights.scm
```diff
@@ -1,142 +1,298 @@
...
-  name: (ident) @type)
+(function_definition
+  declarator: (identifier) @function)
 
-(struct_member
-  name: (ident) @property)
+(call_expression
+  function: (identifier) @function.call
+  arguments: (argument_list
+    (identifier)? @variable))
```
ribru17:
```suggestion
(call_expression
  function: (identifier) @function.call)
```

we already capture all identifiers as variables above

runtime/queries/gdshader/highlights.scm
```diff
@@ -1,142 +1,298 @@
...
-(builtin_function) @function.builtin
+; technically an injection site for bbcode TODO: maybe add it?
+((comment) @comment.documentation @spell
+  .
+  [
+    (struct_definition)
+    (declaration)
+    (function_definition)
+  ]
+  (#lua-match? @comment.documentation "^/%*%*.*%*/$"))
```
ribru17:
```suggestion
  (#lua-match? @comment.documentation "^/%*%*"))
```
nit: this should do the same thing

runtime/queries/gdshader/highlights.scm
```diff
@@ -1,142 +1,298 @@
...
+      ])))
+
+(preproc_include
+  path: (_) @string)
+
+(preproc_undef
+  argument: (identifier) @constant)
 
-(builtin_variable) @constant.builtin
+(comment) @comment
```
ribru17:
```suggestion
(comment) @comment @spell
```

runtime/queries/gdshader/highlights.scm
```diff
@@ -1,142 +1,298 @@
...
+  [
+    (comment)
+    (preproc)
+  ]*
+  .
+  (shader_type_statement
+    (shader_type) @_shader_type)
+  (#eq? @_shader_type "spatial")
+  (_)*
+  (render_mode_statement
```
ribru17: Can also remove these `*` quantifiers & nodes since they aren't necessary for the pattern structure, also in all other places if any

runtime/queries/gdshader/highlights.scm
```diff
@@ -1,142 +1,298 @@
...
+] @keyword.conditional
+
 "struct" @keyword.type
 
 [
-  (precision_qualifier)
-  (interpolation_qualifier)
-] @keyword.modifier
+  "continue"
+  "break"
```
ribru17: These are usually `@keyword.repeat`

runtime/queries/gdshader/highlights.scm
```diff
@@ -1,142 +1,298 @@
...
+  "^="
+  ">>="
+  "<<="
   "--"
+  "++"
 ] @operator
 
-(boolean) @boolean
+(comma_expression
+  operator: "," @operator)
```
ribru17, line 104: maybe this should be removed in favor of a global `","` `@punctuation.delimiter` capture?

runtime/queries/gdshader/highlights.scm
airblast-dev: Simplified them like you suggested.

runtime/queries/gdshader/highlights.scm
airblast-dev: Added them

runtime/queries/gdshader/highlights.scm
airblast-dev: I think this was left over from something else. Not sure what I was doing here.

runtime/queries/gdshader/highlights.scm
airblast-dev: Done but with a slightly different pattern to avoid grouping `/**/` as a doc comment which your pattern would match.

runtime/queries/gdshader/highlights.scm
airblast-dev: I don't think it should be removed as in this context it is an operator and not a delimiter.

Not the same but https://en.wikipedia.org/wiki/Comma_operator and the GLSL ES 3.0 spec seems to also support this.

runtime/queries/gdshader/highlights.scm
airblast-dev: Moved them to keyword.repeat.

airblast-dev:
Added highlight tests for queries unique to `nvim-treesitter`.

ribru17 (approved):

PR #8267, opened 2025-11-08
ci: add check for downstream queries

clason:
Problem: Parser updates may break downstream queries.

Solution: Add workflow that clones and validates queries in

* https://github.com/nvim-treesitter/nvim-treesitter-textobjects#main
* https://github.com/nvim-treesitter/nvim-treesitter-context

on changes to the parsers.lua table. (Workflow should not be required to
pass; the purpose is to give a heads-up that follow-up PRs are needed.)

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -1,0 +1,58 @@
+name: Tests
+
+on:
+  pull_request:
+    branches:
+      - "main"
+    paths:
+      - "lua/nvim-treesitter/parsers.lua"
+  workflow_dispatch:
+
+jobs:
+  test-downstream:
+    name: Check downstream queries
+    runs-on: ubuntu-latest
+    env:
+      NVIM: "nvim"
+    steps:
+      - uses: actions/checkout@v5
+      - uses: tree-sitter/setup-action/cli@v1
+
+      - name: Install and prepare Neovim
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/downstream.yml
```diff
@@ -0,0 +1,58 @@
+name: Tests
+
+on:
+  pull_request:
+    branches:
+      - "main"
+    # paths:
```
clason: Note to self: don't forget to uncomment before merging.

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,58 @@
...
+        uses: actions/checkout@v5
+        with:
+          repository: nvim-treesitter/nvim-treesitter-textobjects
+          ref: main
+          path: .tests/nvim-treesitter-textobjects
+          sparse-checkout: queries
+
+      - name: Check textobjects
+        working-directory:  .tests/nvim-treesitter-textobjects/
+        run: ../../ts_query_ls check queries/
```
clason, line 43: @ribru17 If I run `ts_query_ls` from the nvim-treesitter checkout root, it won't find any `; inherit`ed modules. Why dis?

clason: Answer: because the `workspace` (where modules are looked for) is implicitly the current working directory. Use `--workspace` to override that.

PR #8272, opened 2025-11-10
feat(parsers): update sql

DerekStride:
## What

Updating the sql parser to the latest & adding some new keyword highlights from recent PRs

### DIFF
--- a/runtime/queries/sql/highlights.scm
+++ b/runtime/queries/sql/highlights.scm
@@ -328,8 +328,9 @@
   (keyword_storage)
   (keyword_compression)
   (keyword_duplicate)
-  (keyword_while)
 ] @keyword
+
+(keyword_while) @keyword.repeat
 
 [
   (keyword_restrict)


runtime/queries/sql/highlights.scm
```diff
@@ -149,6 +152,7 @@
   (keyword_constraint)
   (keyword_force)
   (keyword_use)
+  (keyword_include)
```
clason, line 155: `@keyword.import`?

(try to be as specific as possible)

DerekStride: This is for a covering index:

```sql
CREATE INDEX idx_foo ON bar (baz_id) INCLUDE (name, title)
```

runtime/queries/sql/highlights.scm
```diff
@@ -302,6 +309,26 @@
...
+  (keyword_program)
+  (keyword_quote)
+  (keyword_stdin)
+  (keyword_extended)
+  (keyword_main)
+  (keyword_plain)
+  (keyword_storage)
+  (keyword_compression)
+  (keyword_duplicate)
+  (keyword_while)
```
ribru17: This looks like it should be `@keyword.repeat`

ribru17 (approved):

ribru17 (approved):

PR #8276, opened 2025-11-11
fix(t32)!: update parser and queries

xasc:
`trace32_hll_variable` has been removed from the grammar. The queries are now capturing `symbol` instead. `symbol` is highlighted as variable or constant.

### DIFF
--- a/runtime/queries/t32/highlights.scm
+++ b/runtime/queries/t32/highlights.scm
@@ -102,8 +102,6 @@
 
 (path) @string.special.path
 
-(symbol) @string.special.symbol
-
 [
   (character)
   (hll_char_literal)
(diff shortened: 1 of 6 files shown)


runtime/queries/t32/highlights.scm
```diff
@@ -168,6 +166,12 @@
   (identifier) @constant.builtin)
   (#lua-match? @constant.builtin "^[%%/][%l%u][%l%u%d.]*$"))
 
+((symbol) @constant
+  (#match? @constant "^\\\\\\\\\\\\[^\\\\]*(\\\\\\\\[^\\\\]*)?(\\\\[^\\\\]*)?$"))
```
clason: please use `#lua-match?` for performance reasons

xasc: Sorry, I forgot to replace the predicate after copying.

PR #8277, opened 2025-11-12
Update rbs parser and ruby queries

joker1007:

image

### DIFF
--- a/runtime/queries/ruby/injections.scm
+++ b/runtime/queries/ruby/injections.scm
@@ -1,6 +1,10 @@
 ((comment) @injection.content
   (#set! injection.language "comment"))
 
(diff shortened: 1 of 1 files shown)


runtime/queries/ruby/injections.scm
```diff
@@ -1,6 +1,10 @@
 ((comment) @injection.content
   (#set! injection.language "comment"))
 
+((comment) @injection.content
+  (#match? @injection.content "(^#:|\\@rbs)")
```
clason, line 5: needs to be converted to `#lua-match?`

clason:
You can't simply change the target -- you need to actually branch your PR from the target branch.

PR #8278, opened 2025-11-12
Update rbs parser and ruby queries

joker1007:
I addressed the review comment.
I use `lua-match?` instead of `match?`.

And re-submit https://github.com/nvim-treesitter/nvim-treesitter/pull/8277

---
(Original description)

image


joker1007:
fixed some problems that I forgot to support.

joker1007:
I'm sorry.
I found more edge cases. I will fix parser implementation and submit this pull request again.

PR #8281, opened 2025-11-13
bump zsh grammar, fix regex with expansion

georgeharker:
regex with expansions will cause error if injected regex grammar used, disallow

It appears I managed to create a branch on nvim-treesitter/fix-zsh-regex-highlights due to an error in git tracking setup ( i had no idea I had perms, but I do not have perms for a deletion ).  Can you please delete that branch, and sincere apologies for that.

### DIFF
--- a/runtime/queries/zsh/injections.scm
+++ b/runtime/queries/zsh/injections.scm
@@ -2,7 +2,7 @@
   (#set! injection.language "comment"))
 
 ((regex) @injection.content
-  (#not-lua-match? @injection.content "\${.*}")
+  (#not-lua-match? @injection.content "%${.*}")
   (#set! injection.language "regex"))
 
 (heredoc_redirect


clason:
You didn't :) (Just on your fork maybe?)

No need to bump the grammar unless the query changes require it, btw.)

runtime/queries/zsh/injections.scm
```diff
@@ -2,6 +2,7 @@
   (#set! injection.language "comment"))
 
 ((regex) @injection.content
+  (#not-match? @injection.content "\\$\\{.*\\}")
```
clason: `#lua-match?` please

georgeharker: That's weird, my GitHub is showing a branch on nvim-treesitter as well as in my fork.

Have updated to use not-lua-match.

The query changes don't strictly require it though I did fix related issues for cleaner parsing.

PR #8282, opened 2025-11-13
Update rbs parser and ruby queries to support rbs-inline syntax

joker1007:

image

### DIFF
--- a/runtime/queries/rbs/highlights.scm
+++ b/runtime/queries/rbs/highlights.scm
@@ -174,16 +174,11 @@
   ":"
 ] @punctuation.delimiter
 
-; RBS Inline prefix (keep comment color)
-(inline_prefix) @comment
-
 ; RBS Inline syntax
 (inline_class_annotation) @keyword
 
 (inline_doc
   (var_name) @variable.parameter)
-
-(inline_doc_comment) @comment
 
 (inline_generic) @keyword
 


runtime/queries/rbs/highlights.scm
```diff
@@ -160,3 +172,20 @@
   ","
   "."
 ] @punctuation.delimiter
```
ribru17, line 175: nit: `":"` as well

runtime/queries/rbs/highlights.scm
```diff
@@ -160,3 +172,20 @@
...
+; RBS Inline prefix (keep comment color)
+(inline_prefix) @comment
+
+; RBS Inline syntax
+(inline_class_annotation) @keyword
+
+(inline_doc
+  (var_name) @variable.parameter)
+
+(inline_doc_comment) @comment
```
ribru17: Maybe `@comment.documentation` is more fitting? Also, is this capture necessary if ruby itself already captures the text like this?

runtime/queries/rbs/highlights.scm
```diff
@@ -160,3 +172,20 @@
   ","
   "."
 ] @punctuation.delimiter
+
+; RBS Inline prefix (keep comment color)
+(inline_prefix) @comment
```
ribru17: Same here, is this necessary?

runtime/queries/rbs/highlights.scm
joker1007: Thanks!
I added `":"`.

runtime/queries/rbs/highlights.scm
joker1007: As you say, while it was certainly useful for testing highlighting individually with the RBS parser, I don't think it's particularly important for actual use in a text editor.

I delete this, and below too.

joker1007:
@ribru17 
Thank you very much for your review.
I addressed your comments.

ribru17 (approved):

PR #8283, opened 2025-11-14
 chore: remove format-queries script

clason:
The source of truth for formatting (according to make formatquery and
 CI) is now ts_query_ls, so remove the no longer required script to
 prevent divergence and bitrotting.

(This script also triggers a bunch of warnings with emmyluals, adding friction to the eventual migration.)

@lucario387 Thank you for that contribution! Without it, we would still be struggling in the mud of badly formatted treesitter queries!


clason:
Not just partly! The LSP took your format,

PR #8289, opened 2025-11-19
fix(fsharp): correct use of supertypes in `(_type)` patterns

clason:
Problem: These patterns were impossible, since children need to be children of
every subtype of a supertype to be captured in this way. As subtypes
could appear as children themselves, the query code silently "skipped
over" the supertype restriction in the pattern. This was fixed in
tree-sitter v0.26.0, which now (correctly) flags these patterns as
"impossible".

Solution: Add the appropriate child nodes explicitly.

Thanks @willlillis for the fix and the explanation!

### DIFF
--- a/runtime/queries/fsharp/highlights.scm
+++ b/runtime/queries/fsharp/highlights.scm
@@ -376,8 +376,9 @@
   (unit) @function.call)
 
 ((_type
-  (long_identifier
-    (identifier) @type.builtin))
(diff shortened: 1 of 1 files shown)


clason:
Could be, but my goal here is to make sure it doesn't break on 0.26.0 -- if actual F# users encounter a case that is missing, they are welcome to make a PR to add it.

PR #8291, opened 2025-11-20
ci: add GHA dependabot

clason:


clason:
Looks like GH has having a day. Since dependabot won't run from a non-default branch anyway, it's not worth doing now.

PR #8296, opened 2025-11-24
improve highlights for wgsl (add operators and lower priority variable)

benbaarber:
@variable at priority 100 was showing over function calls and types among other things. putting it at 99 is better behavior. also, added 8 compound assignment operators to highlight as operators. 

should note as well that the tree-sitter grammar being used is no longer maintained and is now quite behind the wgsl language spec. i am currently extending the query with stuff like `((identifier) @keyword (#eq? @keyword "const"))` to highlight keywords that are not included in this grammar. Since it does seem a bit hacky, I didn't include those changes in this PR, but let me know if that's something you want and I can add them.

### DIFF
--- a/runtime/queries/wgsl/highlights.scm
+++ b/runtime/queries/wgsl/highlights.scm
@@ -1,4 +1,5 @@
-((identifier) @variable (#set! priority 99))
+((identifier) @variable
+  (#set! priority 99))
 
 (int_literal) @number
 


clason:
Many parsers tracked here have long been abandoned. A Great Pruning is necessary before the 1.0 release.

benbaarber:
Looks like there is an official grammar for wgsl that is now actively maintained. should replace the dead one with this!

https://github.com/gpuweb/tree-sitter-wgsl

I might take a crack at it if i have the time.

clason:
Sorry, but that doesn't look very active either (12 commits, 1y to 3mo ago).

It might still be better, but that would have to be objectively demonstrated (by comparing parse errors on a representative corpus, say).

benbaarber:
You're right, after trying it out this one is actually worse lol, throws the `@type` token out there for basically any identifier. Maybe ill try forking the current grammar and updating it to the modern wgsl spec

PR #8301, opened 2025-11-27
fix(lua-ls): correct workspace.library format in .luarc.json

2400060033:
Updated .luarc.json to use the correct workspace.library object format required by LuaLS. Previously it was an array, which caused LuaLS to not recognize runtime libraries. No other settings were changed.


clason:
This is AI slop bot spam; you are blocked from this org (as well).

PR #8313, opened 2025-12-04
From sql/highlights.scm, remove @type capture

adaitche:
The capture `@type` is too generic and captures many different things, e.g. function calls and table references. It doesn't capture builtin types, which are captured by `@type.builtin`. See screenshots below for examples.

The outcome is that function calls are highlighted as types. Same for table names.

I guess one can make an argument that tables are composite types in SQL, but it doesn't feel right to highlight them in the same way as e.g. `DOUBLE`.

To fix this issue, just remove the `@type` capture rule.

# Examples

Screenshot 2025-12-04 at 11 12 28
Screenshot 2025-12-04 at 11 12 47
Screenshot 2025-12-04 at 11 11 55

### DIFF
--- a/queries/sql/highlights.scm
+++ b/queries/sql/highlights.scm
@@ -12,9 +12,6 @@
   (keyword_array)
 ] @function.call
 
-(object_reference
-  name: (identifier) @type)
-
 (relation
   alias: (identifier) @variable)
 


adaitche:
Note, CI signals fail for unrelated reason: `Error: parser for ipkg is not installed`

clason:
CI fails because you opened the PR to the locked master branch; the active/maintained branch is `main` (as explained in the README).

But before you open a new PR, I want @DerekStride to comment on this change, which looks intrusive.

DerekStride:
This looks like it's cause by a refactor where `object_reference` was used in more places than the previous `table_reference`.

> I guess one can make an argument that tables are composite types in SQL, but it doesn't feel right to highlight them in the same way as e.g. DOUBLE

We can restrict the highlight to the following to achieve a similar result as you mentioned above.

```sexp
(relation
  (object_reference
    name: (identifier) @type))
```

PR #8314, opened 2025-12-04
[ruby] Ensure string-symbols are highlighted properly

CKolkey:
Ensure the entire symbol is highlighted properly, not just the leading `:"` characters.

set priority for closing `"` character so it has symbol highlight applied, instead of string highlight

Before:
Screenshot 2025-12-04 at 12 43 40

After:
Screenshot 2025-12-04 at 12 43 11

### DIFF
--- a/queries/ruby/highlights.scm
+++ b/queries/ruby/highlights.scm
@@ -193,11 +193,12 @@
   (hash_key_symbol)
 ] @string.special.symbol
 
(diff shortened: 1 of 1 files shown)


clason:
Please open your PR against the active `main` branch. Also, avoid explicit priorities and use pattern order if possible (highlights are always applied for all matching patterns in the order within the query).

PR #8315, opened 2025-12-04
In sql/highlights.scm improve @type capture

adaitche:
`@type` currently captures function calls, resulting in the wrong color assigned to them (the one of `@type` instead of `@function.call`).

Also `@type` doesn't capture CTE names. But to be consistent with the behavior that `@type` refers to tables it arguably should capture CTE names.

This PR is an improved version of #8313.

# Example

image

# Alternative

The alternative 
```
(relation
  (object_reference
    name: (identifier) @type))
```
was not chosen because it doesn't capture the table name if it is used to qualify a column name: `table.column`.

cc @DerekStride

### DIFF
--- a/runtime/queries/sql/highlights.scm
+++ b/runtime/queries/sql/highlights.scm
@@ -13,8 +13,12 @@
   (keyword_object_id)
 ] @function.call
 
-(object_reference
-  name: (identifier) @type)
+((object_reference
+  name: (identifier) @type) @_obj_ref
+  (#not-has-parent? @_obj_ref invocation))
+
+(cte
+  (identifier) @type)
 
 (relation
   alias: (identifier) @variable)


runtime/queries/sql/highlights.scm
```diff
@@ -13,8 +13,12 @@
   (keyword_object_id)
 ] @function.call
 
-(object_reference
-  name: (identifier) @type)
+((object_reference
+  name: (identifier) @type) @_obj_ref
+  (#not-has-parent? @_obj_ref invocation))
```
clason, line 18: Be aware that this predicate involves a non-local condition (requiring walking up the entire tree), so this can have significant performance implications.

If it's necessary for correctness, then that needs to be accepted, but you should test this to make sure the impact (especially on typing latency) is not severe.

PR #8316, opened 2025-12-05
feat(parsers)!: remove gitlab-hosted blueprint, fusion, jsonc

clason:
Problem: Gitlab has too frequent outages, which break automation.

Solution: Drop all Gitlab-hosted parsers (two of which have been
unmaintained for years, with jsonc being made obsolete by json
supporting comments now).

---

Enough is enough. If someone wants to take over maintenance of
fusion (with at least a Github mirror) or is willing to
set up and maintain a Github mirror for blueprint, a resubmission
is welcome.

---

(Yes, I know this is mostly Cloudflare's fault, but that doesn't change
the effect it has on us.)

### DIFF
--- a/SUPPORTED_LANGUAGES.md
+++ b/SUPPORTED_LANGUAGES.md
@@ -148,7 +148,7 @@ html_tags (queries only)[^html_tags] | unstable | `H IJ ` | @TravonteD
 [jinja_inline](https://github.com/cathaysia/tree-sitter-jinja)[^jinja_inline] | unstable | `H  J ` | @cathaysia
 [jq](https://github.com/flurie/tree-sitter-jq) | unstable | `H  JL` | @ObserverOfTime
 [jsdoc](https://github.com/tree-sitter/tree-sitter-jsdoc) | unstable | `H    ` | @steelsojka
-[json](https://github.com/tree-sitter/tree-sitter-json) | unstable | `HFI L` | @steelsojka
+[json](https://github.com/tree-sitter/tree-sitter-json) | unstable | `HFIJL` | @steelsojka
 [json5](https://github.com/Joakker/tree-sitter-json5) | unstable | `H  J ` | @Joakker
 [jsonnet](https://github.com/sourcegraph/tree-sitter-jsonnet) | unstable | `HF JL` | @nawordar
 jsx (queries only)[^jsx] | unstable | `HFIJ ` | @steelsojka
--- a/plugin/filetypes.lua
+++ b/plugin/filetypes.lua
@@ -24,6 +24,7 @@ local filetypes = {
   ini = { 'confini', 'dosini' },
   janet_simple = { 'janet' },
   javascript = { 'javascriptreact', 'ecma', 'ecmascript', 'jsx', 'js' },
+  json = { 'jsonc' },
   glimmer_javascript = { 'javascript.glimmer' },
   latex = { 'tex' },
   linkerscript = { 'ld' },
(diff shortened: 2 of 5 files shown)


clason:
@kiyoon this will affect nvim-treesitter-textobjects; you'd have to drop jsonc as well.

kiyoon:
okay I just delete the queries ?

clason:
Yep (and whatever else in documentation or tests you have for that).

I'll leave this PR up for a short while, though -- maybe someone feels invested enough to adopt the JSONC parser ;)

(It's an embarrassingly simple modification of the `json` parser -- although it still needs active maintenance, of course). This is the grammar in its entirety:
```js
const json = require("tree-sitter-json/grammar");

module.exports = grammar(json, {
  name: "jsonc",

  extras: ($, original) => [...original, $.comment],

  rules: {
    document: ($, original) => optional(original),

    comment: ($) =>
      token(
        choice(seq("//", /.*/), seq("/*", /[^*]*\*+([^/*][^*]*\*+)*/, "/"))
      ),
  },
});
```

CONTRIBUTING.md
```diff
@@ -42,6 +42,9 @@ zimbu = {
 >[!IMPORTANT]
 > The "maintainers" here refers to the person maintaining the **queries** in `nvim-treesitter`, not the parser maintainers (who likely don't use Neovim). The maintainers' duty is to review issues and PRs related to the query and to keep them updated with respect to parser changes.
 
+>[!IMPORTANT]
+> Due to reliability issues with smaller codeforges, only Github-hosted parsers are currently eligible for inclusion. (The development may happen elsewhere, but there must at least exist a Github mirror to pull the source from.) We are monitoring the situation and hope to support more codeforges again in the future.
```
acid-bong, line 46: so is it about excluding only Gitlab or disallowing other forges as well? this paragraph conflicts with the rationale in the PR description

clason: We already dropped support for all other forges before; Gitlab was grandfathered in, but the maintenance cost is no longer worth it. I took the liberty of making the stance (and rationale) explicit while I was at it.

clason: Also, to be clear, this only affects the parsers we officially support (installation out of the box, bundled and maintained queries) -- hence this note is in the CONTRIBUTING, not the README/docs. You can still specify Gitlab (and, last I checked, Forgejo) repositories for custom parsers, since then any outage is your problem, not mine.

clason: The problem here is that any flakiness triggers CI failures, not just in this project but across the tree-sitter ecosystem (which relies on nvim-treesitter for "crater tests" since it's the largest collection of curated queries on the internet). We simply can't afford the flakiness, no matter the personal opinion on Github vs. alternatives.

brianhuster:
It seems to me that the JSONC parser is not necessary at all (which is why it is unmaintained), since the JSON parser also support Javascript comments

https://github.com/tree-sitter/tree-sitter-json/blob/001c28d7a29832b06b0e831ec77845553c89b56d/grammar.js#L94-L100

https://github.com/tree-sitter/tree-sitter-json/commit/124edff33adfa209247c03119285feec8c4aca43

clason:
Thank you, that is very helpful! Indeed, the json parser (now) works fine for jsonc files as well, so we can just add the comment captures to the json queries and set up the alias on startup.

ObserverOfTime:
We also have the json5 parser anyway.

z3ntu:
I was using blueprint :(

clason:
Then consider taking over maintenance and setting up a Github mirror.

PR #8317, opened 2025-12-05
[ruby] Fix highlighting for string-literal symbols.

CKolkey:
Before:
Screenshot 2025-12-04 at 12 43 40

After:
Screenshot 2025-12-04 at 12 43 11

In ruby, a symbol literal can be defined with a leading colon, like this `:hello`. If you want the symbol to contain certain characters (numbers, spaces, dashes, etc), quotation marks can be used like a string, but with a leading colon. This is still a symbol literal, not a string.

The current highlight group applied to the "string" content is for strings, but it would be more accurate to apply the symbol highlight group.

Previous PR: https://github.com/nvim-treesitter/nvim-treesitter/pull/8314

### DIFF
--- a/runtime/queries/ruby/highlights.scm
+++ b/runtime/queries/ruby/highlights.scm
@@ -186,16 +186,13 @@
 [
   (bare_symbol)
   (simple_symbol)
-  (delimited_symbol)
   (hash_key_symbol)
 ] @string.special.symbol
 
 (delimited_symbol
-  (string_content) @string.special.symbol)
-
-; Ensure "symbol" highlight has priority over "string" highlight for closing quote character.
-((delimited_symbol) @string.special.symbol
-  (#set! priority 101))
+  ":\"" @string.special.symbol
+  (string_content) @string.special.symbol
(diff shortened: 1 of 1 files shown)


runtime/queries/ruby/highlights.scm
```diff
@@ -190,6 +190,13 @@
   (hash_key_symbol)
 ] @string.special.symbol
 
+(delimited_symbol
+  (string_content) @string.special.symbol)
+
+; Ensure "symbol" highlight has priority over "string" highlight for closing quote character.
+((delimited_symbol) @string.special.symbol
+  (#set! priority 101))
```
ribru17: This pattern makes the one above it redundant. What we really need is just the following:

```
(delimited_symbol
  ":\"" @string.special.symbol
  (string_content) @string.special.symbol
  "\"" @string.special.symbol)
```

and then we can remove the `(delimited_symbol)` node from the alternation above.

CKolkey: Great suggestion, thanks

PR #8320, opened 2025-12-06
feat(parsers): update angular, c3, desktop, elm, forth, inko, mlir, nix, pkl, powershell, templ, xresources

clason:


clason:
@yorickpeterse for some reason the update https://github.com/inko-lang/tree-sitter-inko/commit/fc37d05c363ccf7f515354c5a47d72b2d3ff555e to the inko parser broke parsing the following pattern:

```lisp
(if
    alternative: (else
      (block) @conditional.inner))
```

I don't see why that should be the case?

(By the way, would you consider making actual releases, using the tree-sitter workflows, in addition to tagging from time to time? I would prefer to move away from following every commit to tracking versioned releases.)

yorickpeterse:
@clason I think that pattern itself is wrong. It translates to the syntax `if {} else {}` but that was never correct to begin with, as an `if` requires a condition. I think the reason this worked before is due to the TS grammar being more relaxed compared to Inko's own parser, with the recent changes changing that a bit.

I looked at the existing queries for Inko but can't find anything that resembles your query. Where may I be able to find it?

As for tagging: my bad for not doing that earlier. I've set up the necessary tooling so it should be good from now on. The latest tag is v0.3.0.

clason:
@yorickpeterse sorry, that information got hidden when I merged the PR -- it's actually on nvim-treesitter-textobjects:
https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/5ca4aaa6efdcc59be46b95a3e876300cfead05ef/queries/inko/textobjects.scm#L114-L116

If you see how that can be fixed, a PR would be more than welcome!

(The situation is a bit complicated because this pattern actually crashes `ts_query_ls`, which... shouldn't be able to happen.)

> As for tagging: my bad for not doing that earlier. I've set up the necessary tooling so it should be good from now on. The latest tag is v0.3.0.

No worries! But I actually was talking about proper releases like https://github.com/tree-sitter-grammars/tree-sitter-query/releases/tag/v0.8.0, not just a tag! (You can steal the release workflow from there.)

yorickpeterse:
@clason Ah gotcha, I'll take a look at the query and adopt the release workflow.

yorickpeterse:
OK I'm working on the TS queries. I'm running into the assertion failure described at https://github.com/tree-sitter/tree-sitter/issues/4872, which is fixed in https://github.com/tree-sitter/tree-sitter/pull/4877. It's not released yet though so I'll continue trying to find a way around this.

clason:
Well, you can still test, right? If it stops crashing, the pattern is fixed ;)

But I can help test; I don't recall whether building ts_query_ls with 0.26 needs changes.

(In any case, you can use Nvim's built-in query linter for this; nightly includes a tree-sitter commit with the fix.)

yorickpeterse:
@clason I currently have things set up to a point where all the conditionals work, except adding a query for `else` crashes it. It seems it doesn't really matter how you structure the query. The diff is this:

```diff
diff --git a/queries/inko/textobjects.scm b/queries/inko/textobjects.scm
index 2a67883..a494159 100644
--- a/queries/inko/textobjects.scm
+++ b/queries/inko/textobjects.scm
@@ -76,19 +76,20 @@
 
 ; Conditionals
 (if
-  alternative: (_
-    (_) @conditional.inner)?) @conditional.outer
+ consequence: (_) @conditional.inner)
 
 (if
-  alternative: (else
-    (block) @conditional.inner))
+ condition: (_) @conditional.inner)
 
-(if
-  consequence: (block)? @conditional.inner)
+(else_if
+  consequence: (_) @conditional.inner)
 
-(if
+(else_if
   condition: (_) @conditional.inner)
 
+;(else
+;  body: (_) @conditional.inner)
+
 (case) @conditional.inner
 
 (match) @conditional.outer
```

The problematic query is the following:

```
(else
  body: (_) @conditional.inner)
```

If you nest it inside an `if` (so `(if alternative: (else ...))`) it makes no difference. Looking at the issue it seems it may be due certain patterns not being reachable (at least based on the PR description). I'll see if adjusting the grammar (e.g. the precedence) helps.

clason:
Yeah, tree-sitter 0.26 tightened up some previously silently skipped invalid patterns. (There's going to be a proper release within the next week or so, so now is the time to get prepared for it rolling out ;))

PR #8323, opened 2025-12-06
feat(fortran)!: update parser and queries

clason:
Breaking change: `(do_loop_statement)` renamed to `(do_loop)`


clason:
@lewis6991 this is a breaking change for context (with a less trivial fix).

PR #8324, opened 2025-12-06
Allow downloading with git instead of curl

ndhansen:
_(Yes, I know git uses curl under the hood, bear with me here)_

This is a bit of a weird one. I'm running this plugin in an environment which is heavily locked down. Only cloning from git works, downloading archives does not work. To use this plugin I forked and added a config option to use git to obtain the code, rather than using curl to download the archive.

I totally understand if this PR is rejected, arguably I'm adding code bloat with redundant functionality to side-step a limitation beyond the scope of this tool. But I felt like it might be worth trying to upstream this change, in case there is actually a good purpose for this.

### DIFF
--- a/lua/nvim-treesitter/config.lua
+++ b/lua/nvim-treesitter/config.lua
@@ -4,10 +4,12 @@ M.tiers = { 'stable', 'unstable', 'unmaintained', 'unsupported' }
 
 ---@class TSConfig
 ---@field install_dir string
+---@field prefer_git boolean
 
 ---@type TSConfig
 local config = {
   install_dir = vim.fs.joinpath(vim.fn.stdpath('data'), 'site'),
+  prefer_git = false,
 }
 
 ---Setup call for users to override configuration configurations.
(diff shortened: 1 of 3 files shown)


clason:
Yes, sorry, we removed all these options intentionally since the combinations were just too much to maintain. 

My suggestion for such situations is to just pull in parsers offline and then move them to your config manually.  A locked down system shouldn't constantly update, anyway ;)

ndhansen:
@clason If I changed how the plugin pulls code from `curl`-ing the archive endpoint to `curl`-ing the regular git endpoints, would that be accepted? Arguably that would leave the code largely the same, but extend it to git servers beyond just github and gitlab.

clason:
I'm not against the change if it's the same code complexity and works with any git reference (including tags and releases!) Supporting non-Github servers is a non-goal, though, sorry.

But be aware that the end goal is to pull actual release artifacts rather than the full repo (once upstream finally gets that story sorted out); that's the reason for the current endpoint.

ndhansen:
Took a quick look, and while it would be possible it would add a lot of complexity and add a lot of potential points where it could break. I would have to craft a custom request body binary (see below), and if github or gitlab changed to a newer git protocol version at some future point it would break in far more complicated ways than it would break now.

I'll abandon this approach and go with the far more reasonable "download the artifacts and install from local" approach suggested above. Thanks!

```
0011 command=fetch
001a agent=git/2.52.0-Linux
0016 object-format=sha1
0001 
000d thin-pack
000f no-progress
000d ofs-delta
0032 want <commit hash here>.
0009 done.
0000
```

clason:
Thanks for looking into this! And sorry we can't cover this.

(If you go with the "download offline" route, you do not need to install _at all_ -- in fact, if you don't use indents, you do not need this plugin, period -- just put the compiled parsers and the queries downloaded from here somewhere in your runtimepath and everything will Just Work(tm). It's a common misconception that you _need_ nvim-treesitter to use treesitter in any way.)

PR #8325, opened 2025-12-06
[gdscript] abstract annotation highlight

rrddrose:
Highlight abstract class annotation introduced in Godot 4.5


clason:
I asked you to please keep the list alphabetical.

clason:
Also, these lists are (auto?)generated from the actual upstream data (see the comment above the list), and this manual addition breaks that.

Hence I would prefer it if you could fully sync the list and update the commit sha to match.

rrddrose:
I have re-ordered the data and updated the commit hash and references that introduced the changes in Godot.

PR #8326, opened 2025-12-06
ci(test): add parsers cache

clason:
We originally said that the new async parser installation is fast enough not to _need_ a cache, but it's still good not to needlessly waste cycles. And as we move towards less frequent parser updates (e.g., by switching more and more parsers to version updates -- or no updates at all), the cache will pay off more and more.


clason:
Nvm, forgot that here we also install the queries, so we'd need to cache them as well -- which means the cache is invalidated on pretty much every PR.

PR #8327, opened 2025-12-06
feat(inko): update to 0.4.0

yorickpeterse:
This changes the structure of "if" expressions so the queries for nvim-treesitter-textobjects can be changed as to not cause any crashes.


clason:
You can use the tag (and set `tier = 1`) instead of the sha.

clason:
and `make docs` ;)

yorickpeterse:
The textobject tests will continue to fail until https://github.com/nvim-treesitter/nvim-treesitter-textobjects/pull/837 is merged. That however depends on this PR merging, so this is a bit of an unfortunate circular dependency.

clason:
Nope, not circular. The downstream check is just a canary and doesn't block the merge.

PR #8331, opened 2025-12-08
feat(bash): highlight arguments of trap and kill builtin commands

ValdezFOmar:
Before:
image

After:
image


runtime/queries/bash/injections.scm
```diff
@@ -76,3 +76,15 @@
...
   (#set! injection.include-children)
   (#set! injection.language "readline"))
+
+((command
+  name: (command_name) @_command
+  .
+  argument: [
+    (string)
+    (raw_string)
+  ] @injection.content)
```
clason, line 86: I think you can just use `@injection.self` here?

